PR #6830, opened 2026-05-22
[OPIK-6652] [FE][BE] Route optimization studio LLM calls through backend gateway

jverre:
## Details

image

image

image

Route all optimization studio LLM calls through the Opik backend's existing `/v1/private/chat/completions` gateway endpoint instead of using hardcoded per-provider API keys.

Previously, the Python backend passed `OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, and `OPENROUTER_API_KEY` as env vars to the optimizer subprocess. This limited optimization studio to only providers whose keys were configured in docker-compose.

Now, LLM calls go through the same gateway the playground uses. The backend resolves provider keys from the workspace DB, so optimization studio automatically supports every provider configured in the UI (OpenAI, Anthropic, Gemini, Vertex AI, Bedrock, OpenRouter, Ollama, Custom).

**Python backend changes:**
- Replace `LLM_API_KEYS` dict with `OPIK_GATEWAY_BASE_URL` pointing at the backend's completions endpoint
- Set `OPENAI_API_BASE` env var so LiteLLM routes through the gateway
- Prepend `openai/` to model strings so LiteLLM uses its OpenAI handler (stripped before the HTTP request)
- Remove per-provider API key env vars from docker-compose
- Pass `stream=False` explicitly via model params and metric models to avoid null stream issues
- Improve subprocess error logging to surface stdout (where optimizer_runner outputs error JSON)

**Java backend changes:**
- Fix NPE in `LlmProviderAnthropicMapper` when `stream` field is null — default to `false`

**Frontend changes:**
- Update v2 `OptimizationModelSelect` to use the same `useProviderKeys` + `useModelOptions` hooks as the playground
- Show all configured providers instead of a hardcoded OpenAI/Anthropic list
- Add "Manage AI providers" link to the model selector dropdown

## Additional changes (Daniel)

- **Anthropic mapper regression test** (`AnthropicMappersTest.java`): parameterized test covering the null-stream NPE fix (true / false / null cases)
- **Helm chart docs**: rephrase `secretRefs` guidance — LLM keys are now configured via workspace UI, not env vars; regenerate chart README
- **executor_isolated**: extract `_parse_last_json_line` helper to dedupe stdout JSON parsing across success/failure branches
- **OptimizationModelSelect trigger**: show placeholder when the stored model value has no matching configured provider (avoids displaying orphaned selections that the user can't actually use)
- **FE import order**: align with project convention (React → external libs → `@/` aliases)

## Migration notes

- **Self-hosted users with env-var-set LLM keys** will need to configure providers via workspace AI Providers settings post-upgrade. The v2 selector surfaces this explicitly with a "No configured providers" empty state and a "Manage AI providers" link.
- **comet.com hosted users**: `OPENAI_API_KEY` is injected into the python-backend pod via AWS Secrets Manager (`opik-python-backend` secret in `values-production-opik.yaml`). After this PR, hosted optimization-studio users will need workspace-configured providers. Worth coordinating with infra/ops before shipping.
- **v1 optimization studio**: the v1 UI is unchanged but its backend path now requires workspace-configured providers. v1 users will see the old hardcoded OpenAI/Anthropic dropdown but optimization jobs will fail unless providers are configured. Worth a release-note callout, or a follow-up to add a v1 warning banner.

## Follow-up tickets

- [OPIK-6655](https://comet-ml.atlassian.net/browse/OPIK-6655) — `[FE]` Unify `OptimizationModelSelect` and `PromptModelSelect` into a single component (also addresses the empty-state edge case raised in review)
- [OPIK-6656](https://comet-ml.atlassian.net/browse/OPIK-6656) — `[INFRA]` Run helm-docs locally via `.hooks/pre-commit` when chart files change

## Change checklist
- [x] Python backend gateway routing
- [x] Frontend model selector parity with playground
- [x] Remove unused docker-compose env vars
- [x] Fix Anthropic mapper null stream NPE
- [x] Improve subprocess error logging
- [x] Anthropic mapper regression test (parameterized)
- [x] Helm chart docs updated for workspace-based provider config

## Issues
https://comet-ml.atlassian.net/browse/OPIK-6652

## AI-WATERMARK
ai-generated

## Testing
- [x] Run optimization with an Anthropic model (claude-opus-4-5) — verified working end-to-end
- [x] Run optimization with an OpenAI model (e.g. gpt-4o-mini)
- [x] Verify model selector shows all configured providers
- [x] Verify "Manage AI providers" link works from optimization studio
- [x] Verify Anthropic + GEval combo end-to-end on test env
- [x] Verify selector trigger shows placeholder (not orphaned value) when no provider matches the stored value

## Documentation
N/A

[OPIK-6655]: https://comet-ml.atlassian.net/browse/OPIK-6655?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/opik-python-backend/src/opik_backend/studio/config.py
```diff
@@ -5,12 +5,10 @@
...
-# LLM API Keys
-LLM_API_KEYS = {
-    "OPENAI_API_KEY": os.getenv("OPENAI_API_KEY"),
-    "ANTHROPIC_API_KEY": os.getenv("ANTHROPIC_API_KEY"),
-    "OPENROUTER_API_KEY": os.getenv("OPENROUTER_API_KEY"),
-}
+# Gateway base URL for LLM calls — points at the Opik backend's
+# OpenAI-compatible completions endpoint so that all provider/key
+# resolution happens in the Java backend (same path as the playground).
+OPIK_GATEWAY_BASE_URL = f"{OPIK_URL}/v1/private" if OPIK_URL else None
```
JetoPistola: 🚫 **blocker** | Configuration

If `OPIK_URL_OVERRIDE` is unset, `OPIK_GATEWAY_BASE_URL` is `None`, and the caller in `optimizer.py` skips setting `OPENAI_API_BASE` but still sets `OPENAI_API_KEY = "opik-local"`. LiteLLM will then attempt to hit the **real OpenAI API** with the literal key `opik-local`, producing a confusing 401 from OpenAI instead of a clear configuration error.

In docker-compose this is guarded (`OPIK_URL_OVERRIDE` defaults to `http://backend:8080`), but k8s/dev deployments — including likely setups for the NVIDIA POC — can hit this silently. Suggest failing fast with a clear error when `OPIK_GATEWAY_BASE_URL` is unset, or at minimum logging a prominent warning at module import.

apps/opik-python-backend/src/opik_backend/jobs/optimizer_runner.py
```diff
@@ -153,6 +153,18 @@ def main():
         context = OptimizationJobContext.from_job_message(job_message)
         config = OptimizationConfig.from_dict(job_message.get("config", {}))
 
+        # Force LiteLLM to route through its OpenAI handler so it uses
+        # OPENAI_API_BASE (the Opik gateway).  LiteLLM strips the "openai/"
+        # prefix before sending the request, so the backend receives the
+        # original model string (e.g. "vertex_ai/gemini-2.5-flash").
+        if not config.model.startswith("openai/"):
```
JetoPistola: 💡 **suggestion** | Robustness

The `openai/` prefix is prepended unconditionally when not already present, so `vertex_ai/gemini-2.5-flash` becomes `openai/vertex_ai/gemini-2.5-flash`. The comment correctly notes LiteLLM strips the `openai/` prefix for its OpenAI handler — but consider logging the final model string so debugging a misrouted call (if LiteLLM ever changes behavior) is easier.

Also: this prefix manipulation lives in the runner, but the workaround is really about routing — would be worth a brief note in `process_optimizer_job` (where the gateway env vars are set) pointing at this spot, so the two pieces stay in sync.

apps/opik-python-backend/src/opik_backend/executor_isolated.py
```diff
@@ -331,8 +331,18 @@ def _execute_in_subprocess(
...
+                # Try to parse error JSON from stdout (optimizer_runner outputs it there)
+                try:
+                    lines = [line for line in stdout.split('\n') if line.strip()]
+                    if lines:
+                        result = json.loads(lines[-1])
+                        if isinstance(result, dict) and "error" in result:
+                            return result
+                except (json.JSONDecodeError, ValueError):
+                    pass
                 return {
```
JetoPistola: 💡 **suggestion** | Error Handling

When a subprocess fails and the last stdout line is valid JSON with an `error` key, the code returns it directly. But downstream callers may expect a `code` field (the surrounding fallback path always sets it to `500`). If `optimizer_runner` emits `{"error": "..."}` without a `code`, callers checking `result["code"]` will get a `KeyError`.

```suggestion
                try:
                    lines = [line for line in stdout.split('\n') if line.strip()]
                    if lines:
                        result = json.loads(lines[-1])
                        if isinstance(result, dict) and "error" in result:
                            result.setdefault("code", 500)
                            return result
                except (json.JSONDecodeError, ValueError):
                    pass
```

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/antropic/LlmProviderAnthropicMapper.java
```diff
@@ -56,7 +56,7 @@ interface LlmProviderAnthropicMapper {
     AssistantMessage toAssistantMessage(@NonNull AnthropicContent content);
 
     @Mapping(expression = "java(request.model())", target = "model")
-    @Mapping(expression = "java(request.stream())", target = "stream")
+    @Mapping(expression = "java(Boolean.TRUE.equals(request.stream()))", target = "stream")
```
JetoPistola, line 59: 💡 **suggestion** | Readability

`Boolean.TRUE.equals(request.stream())` works but reads a bit oddly given the surrounding mapping expressions. The root cause is that the `stream` target on `AnthropicCreateMessageRequest` is a primitive `boolean`, so a null `request.stream()` NPEs when MapStruct unboxes. A cleaner long-term fix would be making the target field `Boolean` (or defaulting it to `false` at the model level), but the current patch is fine as a localized fix.

One question worth flagging on the PR: is this the same NPE responsible for the intermittent Anthropic-in-playground production bug? If so, this one-line fix may be worth extracting into its own PR so it can ship independently of the optimization-studio changes.

apps/opik-python-backend/src/opik_backend/studio/metrics.py
```diff
@@ -350,12 +351,14 @@ def _build_geval_metric(params: Dict[str, Any], model: str, **kwargs) -> Callabl
         _has_template_placeholders(eval_criteria_template)
     )
     
+    llm_model = LiteLLMChatModel(model_name=model, stream=False)
```
JetoPistola, line 354: ❓ **question** | Defense-in-depth

Now that the Anthropic mapper NPE is fixed and `optimizer_runner.py` sets `model_params["stream"] = False` for the main model, is the explicit `LiteLLMChatModel(model_name=model, stream=False)` in the metric models still strictly necessary, or is this defensive duplication? Same question applies to L418. Worth a short comment if intentional, or removable if redundant.

apps/opik-frontend/src/v2/pages-shared/optimizations/OptimizationModelSelect/OptimizationModelSelect.tsx
```diff
@@ -103,126 +109,198 @@ const OptimizationModelSelect: React.FC<OptimizationModelSelectProps> = ({
...
+              withoutCheck
+              className="comet-body-s h-10 justify-center hover:bg-primary-foreground"
+            >
+              <div className="flex w-full items-center gap-2">
+                {FreeModelIcon && <FreeModelIcon className="size-4 shrink-0" />}
+                <span>{filteredFreeModel.label}</span>
+              </div>
+            </SelectItem>
+          )}
+          {filteredGroups.map((group) => (
```
JetoPistola, line 140: 🧹 **nit** | Style

The inline object return is long for one line — could be formatted across lines for consistency with the other branch.

```suggestion
    if (freeModelOption && value === freeModelOption.value) {
      return {
        icon: freeModelOption.icon,
        label: freeModelOption.label,
        title: freeModelOption.label,
      };
    }
```

JetoPistola: Fixed in 83006fe — auto-formatted by ESLint/Prettier (the lint commit also expanded the inline object to multi-line as suggested).

apps/opik-python-backend/src/opik_backend/jobs/optimizer.py
```diff
@@ -114,11 +114,12 @@ def process_optimizer_job(*args: Any, **kwargs: Any) -> Dict[str, Any]:
...
-        # Pass LLM API keys and Opik configuration
-        env_vars = {
-            **LLM_API_KEYS,  # OPENAI_API_KEY, ANTHROPIC_API_KEY, etc.
-        }
+        # Route all LLM calls through the Opik backend gateway.
+        # LiteLLM reads OPENAI_API_BASE and uses it as an OpenAI-compatible endpoint.
+        env_vars = {}
+        if OPIK_GATEWAY_BASE_URL:
+            env_vars["OPENAI_API_BASE"] = OPIK_GATEWAY_BASE_URL
+        env_vars["OPENAI_API_KEY"] = context.opik_api_key or "opik-local"
```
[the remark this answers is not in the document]
JetoPistola: Skipping — this is a deliberate design decision, not an oversight. The PR moves opt-studio onto the same workspace-provider resolution path the playground uses, which removes a parallel env-var config plane that was a stopgap (env vars required restarts; the workspace UI doesn't).

For users currently relying on backend env vars: post-PR, the optimization-studio model selector explicitly shows "No configured providers" with a one-click "Manage AI providers" link when nothing is set up, so the migration surfaces visibly in-context rather than as a silent break. The customer-facing helm chart docs have been updated in 3c65882 to direct operators to the UI.

Worth confirming with infra/ops that comet.com hosted opt-studio users are aware of the change before shipping — but the migration path is discoverable, not silent.

apps/opik-python-backend/src/opik_backend/jobs/optimizer_runner.py
```diff
@@ -153,6 +153,13 @@ def main():
...
+            config.model = f"openai/{config.model}"
```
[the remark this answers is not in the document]
JetoPistola: Skipping — same rationale as the [optimizer.py:122 thread](https://github.com/comet-ml/opik/pull/6830#discussion_r3288840433): the env-var path is being intentionally retired in favor of workspace-provider resolution, which the playground already uses. Provider keys reach the optimizer subprocess via the new gateway routing (`OPIK_API_KEY` → `/v1/private/chat/completions` → Java backend resolves workspace provider). Not a silent break — the FE selector surfaces the empty state explicitly.

apps/opik-frontend/src/v2/pages-shared/optimizations/OptimizationModelSelect/OptimizationModelSelect.tsx
```diff
@@ -103,126 +109,198 @@ const OptimizationModelSelect: React.FC<OptimizationModelSelectProps> = ({
...
 
-  const selectedModelLabel =
-    selectedGroup?.options.find((o) => o.value === value)?.label ?? value;
-  const displayTitle = selectedGroup
-    ? `${selectedGroup.label} ${selectedModelLabel}`
-    : selectedModelLabel;
+  const hasNoProviders =
+    !freeModelOption && configuredProvidersList.length === 0;
+  const hasNoResults =
+    !filteredFreeModel && filteredGroups.length === 0 && filterValue !== "";
```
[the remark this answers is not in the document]
JetoPistola: Skipping in this PR — same edge case exists in `PromptModelSelect` (the playground selector this PR mirrors via the shared `useModelOptions` hook). Fixing it only here would re-introduce the divergence this PR is trying to eliminate.

Tracking the full unification as [OPIK-6655](https://comet-ml.atlassian.net/browse/OPIK-6655) — that ticket subsumes this edge case by lifting the empty-state check into the shared hook so both selectors stay in sync automatically. Not blocking the POC.

apps/opik-python-backend/src/opik_backend/studio/config.py
JetoPistola: Skipping — on reflection, this isn't actually a blocker. `OPIK_URL_OVERRIDE` is set with a hardcoded default in every production deployment path:

- `deployment/docker-compose/docker-compose.yaml` → `http://backend:8080`
- `deployment/docker-compose/docker-compose.override.yaml` → `http://host.docker.internal:8080`
- `deployment/helm_chart/opik/values.yaml:270` → `http://opik-backend:8080` (hardcoded)
- comet.com production helm inherits the same default
- All E2E CI workflows → `http://localhost:8080`

The only way to hit the misleading-401 failure mode is to deliberately remove the variable from a deployment that ships with it set, which is a self-inflicted misconfig outside the normal contract. Downgrading from blocker to non-issue.

apps/opik-python-backend/src/opik_backend/jobs/optimizer_runner.py
JetoPistola: Skipping — the existing 3-line comment at L156–160 already explains the prefix manipulation clearly (why we prepend `openai/`, that LiteLLM strips it before the HTTP request). Adding an extra debug log of the final model string would generate noise on every optimization run for a hypothetical LiteLLM behavior change. Polish-tier, not worth the maintenance burden.

apps/opik-python-backend/src/opik_backend/executor_isolated.py
JetoPistola: Skipping — re-checked the actual caller in `optimizer.py:180–182` and it only reads `"error" in result` and `result.get("error", ...)`. No consumer references `result["code"]`, so the missing field doesn't trigger a `KeyError` downstream. My original suggestion was based on a hypothetical caller pattern that doesn't exist in this codebase. Adding `setdefault("code", 500)` would be unused defensive code.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/antropic/LlmProviderAnthropicMapper.java
JetoPistola: Skipping — on closer look, the fix is correct as-is and not a behavior change for the non-buggy case. The only behavior difference is in the null case:

| `request.stream()` | Before | After |
|---|---|---|
| `Boolean.TRUE` | `true` | `true` |
| `Boolean.FALSE` | `false` | `false` |
| `null` | **NPE** | `false` |

Treating null as `false` matches Anthropic's API default (non-streaming) and is consistent with how other LLM providers in the codebase handle missing `stream` fields. Strict bug fix, not a semantic change. Changing the target type to `Boolean` would require touching langchain4j-provided types, which is more overhead than the localized fix is worth.

One question worth your call though: is this NPE the same root cause as the intermittent Anthropic-in-playground production bug mentioned in the PR description? If yes, the one-line fix could be cherry-picked into a standalone PR so the playground bug fix ships independently of the optimization-studio routing changes.

apps/opik-python-backend/src/opik_backend/studio/metrics.py
JetoPistola: Skipping — re-checked the code and these `LiteLLMChatModel` instances are for the metric/judge models (GEval, JSON Schema Validator), not the main optimization model. They construct their own `LiteLLMChatModel` and don't inherit `stream=False` from `optimizer_runner.py`'s `config.model_params`. So the explicit setting here isn't duplicative with the main-model setting — it's required for the metric path to send a non-null `stream` field through the gateway.

apps/opik-python-backend/src/opik_backend/executor_isolated.py
```diff
@@ -331,8 +331,18 @@ def _execute_in_subprocess(
...
+                    f"stdout: {stdout}. stderr: {stderr}"
                 )
+                # Try to parse error JSON from stdout (optimizer_runner outputs it there)
+                try:
+                    lines = [line for line in stdout.split('\n') if line.strip()]
+                    if lines:
+                        result = json.loads(lines[-1])
+                        if isinstance(result, dict) and "error" in result:
+                            return result
+                except (json.JSONDecodeError, ValueError):
```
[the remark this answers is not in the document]
JetoPistola: Partially addressing — applying the dedup suggestion (extracting a shared helper for the stdout JSON parsing across success/failure paths). That fix is incoming in a follow-up commit.

Skipping the log-verbosity portion (moving stdout/stderr to DEBUG): the PR's stated goal in 92a3512 was specifically to *surface* the subprocess stdout in error logs because optimizer_runner outputs error JSON there and the previous logging was hiding optimizer failures. Lowering it back to DEBUG would partially undo what the PR deliberately added. Can revisit as a separate ticket if optimizer-error log volume becomes a real problem in production.

apps/opik-frontend/src/v2/pages-shared/optimizations/OptimizationModelSelect/OptimizationModelSelect.tsx
```diff
@@ -9,20 +9,23 @@ import {
   SelectTrigger,
   SelectValue,
 } from "@/ui/select";
-import { ChevronRight, Search } from "lucide-react";
+import { ChevronRight, Search, Settings2 } from "lucide-react";
+import { ListAction } from "@/ui/list-action";
```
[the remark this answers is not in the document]
JetoPistola: Fixed in 18711782a — reordered imports to follow React → external libs (`lucide-react`) → `@/` aliases.

apps/opik-python-backend/src/opik_backend/executor_isolated.py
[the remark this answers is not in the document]
JetoPistola: Dedup fix landed in 18711782a — extracted `_parse_last_json_line` static method on `IsolatedSubprocessExecutor` that both success and failure branches now use. Helper returns `(result, error_message)` so the success-branch error string parity is preserved.

(Still skipping the log-verbosity change per the rationale in the prior reply.)

CometActions:

### Access Information
- **URL:** https://pr-6830.dev.comet.com
- **Namespace:** pr-6830
- **Version:** 2.0.47-6830-merge-2144
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-6830%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

CometActions:

### Access Information
- **URL:** https://pr-6830.dev.comet.com
- **Namespace:** pr-6830
- **Version:** 2.0.47-6830-merge-2145
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-6830%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/antropic/LlmProviderAnthropicMapper.java
JetoPistola: Correction to my earlier note: the cherry-pick suggestion was unnecessary. `LlmProviderAnthropicMapper` is shared infrastructure called from every Anthropic code path — `ChatCompletionsResource` (playground + this PR's gateway path), `OnlineScoringLlmAsJudgeScorer` (online evals), and any future consumer via `LlmProviderFactory`. The one-line fix in this PR already covers all of them. If the intermittent Anthropic-in-playground prod bug is the null-stream NPE, this PR fixes it as-is. No standalone PR needed.

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-6830`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-6830`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

alexkuzmik (dismissed):

aadereiko (dismissed):
FE LGTM!

apps/opik-frontend/src/v2/pages-shared/optimizations/OptimizationModelSelect/OptimizationModelSelect.tsx
```diff
@@ -103,126 +99,200 @@ const OptimizationModelSelect: React.FC<OptimizationModelSelectProps> = ({
...
-        >
-          <SelectValue placeholder="Select an LLM model">
-            <div className="flex items-center gap-2">
-              {selectedGroup && (
-                <selectedGroup.icon className="min-w-3.5 text-foreground" />
+  const renderOptions = () => {
+    if (hasNoProviders) {
+      return (
+        <div className="comet-body-s flex h-20 items-center justify-center text-muted-slate">
+          No configured providers
```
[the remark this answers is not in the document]
JetoPistola: Skipping — this is the same finding already tracked as [OPIK-6655](https://comet-ml.atlassian.net/browse/OPIK-6655) (unify `OptimizationModelSelect` and `PromptModelSelect` into a single component). Deferred deliberately to keep this PR's scope tight to the gateway-routing change. The shared `useModelOptions` hook handles the data layer; OPIK-6655 covers lifting the rendering shell into a shared component so both selectors stay in sync.

aadereiko (dismissed):
FE LGTM

alexkuzmik (dismissed):

apps/opik-python-backend/src/opik_backend/jobs/optimizer_runner.py
```diff
@@ -153,6 +153,18 @@ def main():
         context = OptimizationJobContext.from_job_message(job_message)
         config = OptimizationConfig.from_dict(job_message.get("config", {}))
 
+        # Force LiteLLM to route through its OpenAI handler so it uses
```
ldaugusto: This description isn't bad, but I had to read it a few times to actually understand what's going on -- like I wasn't aware this is a regular LiteLLM usage, just prefix 'openai/' so it can work with any OpenAI-compatible provider, like us.

I'd rather have this core part a bit more clear.

CometActions:

### Access Information
- **URL:** https://pr-6830.dev.comet.com
- **Namespace:** pr-6830
- **Version:** 2.0.49-6830-merge-2174
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-6830%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

apps/opik-python-backend/src/opik_backend/studio/config.py
ldaugusto: Not a biggie as we will be doing this, but maybe add some basic validation/cleaning here, like someone can end URLs with '/' which would lead to double /

apps/opik-python-backend/src/opik_backend/executor_isolated.py
```diff
@@ -223,6 +223,20 @@ def _prepare_environment(self, env_vars: Optional[dict] = None) -> dict:
         env["LOG_FORMAT"] = 'json'
         return env
 
+    @staticmethod
+    def _parse_last_json_line(stdout: str) -> tuple[Optional[dict], Optional[str]]:
```
ldaugusto, line 227: this test could use more testing under edge cases to make sure it behaves as expected

ldaugusto (dismissed):
The solution makes sense for us, particuarly reusing what we already have.

I have some minor suggestions, but none are blockers.

apps/opik-python-backend/src/opik_backend/jobs/optimizer_runner.py
JetoPistola: Fixed in 3d909d64f — rephrased the comment to lead with the intent ("treat the Opik backend as an OpenAI-compatible endpoint") and walk through the prefix-and-strip mechanic from that framing. Should read more clearly on a first pass.

apps/opik-python-backend/src/opik_backend/studio/config.py
JetoPistola: Fixed in 3d909d64f — added `rstrip("/")` on `OPIK_URL` before composing the gateway URL, with a brief inline comment explaining why. Trailing-slash configs no longer produce double-slash URLs.

apps/opik-python-backend/src/opik_backend/executor_isolated.py
JetoPistola: Fixed in 3d909d64f — added a `TestParseLastJsonLine` class to `test_executor_isolated.py` covering:

- Single valid JSON line → parsed dict
- Multiple lines → picks the last non-empty
- Trailing blank lines → ignored
- Empty string → `(None, "No output produced by subprocess")`
- Whitespace-only stdout → same as above
- Malformed JSON on last line → `(None, "Invalid JSON response from subprocess...")`
- Non-dict JSON (e.g. array) → returns it unchanged; the test documents that dict-shape enforcement is the caller's job

7 new tests, all pass alongside the existing 19.

alexkuzmik (approved):

PR #6013, opened 2026-03-31
[OPIK-5223][FE]: kpi cards;

aadereiko:
## Details
Added KPI cards and Graphs to v2 of Logs page;
image
image
image
image

## Change checklist
- [ X] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-5223

## AI-WATERMARK

  - Tools: Calude Code
  - Model(s): Opus 
  - Human verification: Yes

## Testing
- Manually
- `npm run lint`
- `npm run typecheck`

PR #6866, opened 2026-05-26
[OPIK-6640] [BE][FE] fix: return KPI error rate as percentage (was raw count)

awkoy:
## Summary

- Backend SQL in `KpiCardDAO` (trace + span queries) now returns the errored-row **percentage** (`errored * 100.0 / total`, guarded against zero) instead of a raw count. Same formula already used by `ProjectMetricsDAO.error_rate` for the time-series chart.
- Document the new contract on `KpiMetricType.ERRORS` and update the OpenAPI example to a fractional value (2.5 instead of 320).
- Tighten the frontend formatter (2 decimals when < 1%, 1 decimal otherwise) and hide the relative-change delta on the errors card, since a `5% → 6%` rate change rendering as `+20%` was confusing.

## Why

The KPI errors card was rendering values like **"312.0%"** because the backend returned a count (e.g. 312) and the frontend labelled the card "Error rate" and appended `%`. The count vs. rate contract had drifted between PR #5884 (backend, count semantics) and PR #6013 (frontend label).

Jira: [OPIK-6640](https://comet-ml.atlassian.net/browse/OPIK-6640)

## Test plan

- [x] `mvn test -Dtest=KpiCardsResourceTest` — 85/85 pass (assertions and `assertFilteredMetrics` helper updated to expect rate)
- [x] Frontend `tsc --noEmit` passes
- [ ] Smoke: verify the errors card on a real project shows e.g. `2.5%` instead of `312.0%`
- [ ] Smoke: verify the time-series chart and the KPI card now agree (both use the same formula)
- [ ] Smoke: verify cards in compact/no-label/icon-only layouts still hide their delta (existing `cardMode !== "full"` behavior preserved)

[OPIK-6640]: https://comet-ml.atlassian.net/browse/OPIK-6640?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/metrics/KpiCardResponse.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/metrics/KpiCardResponse.java
@@ -31,6 +31,7 @@ public record KpiCardResponse(List<KpiMetric> stats) {
     @RequiredArgsConstructor
     public enum KpiMetricType {
         COUNT("count"),
+        /** Errors KPI — value is a percentage in the range [0, 100], with 0 returned when the period has no entities. */
         ERRORS("errors"),
         AVG_DURATION("avg_duration"),
         TOTAL_COST("total_cost");
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/metrics/KpiCardResponse.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/metrics/KpiCardResponse.java
@@ -6,6 +6,7 @@ import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonValue;
 import com.fasterxml.jackson.databind.PropertyNamingStrategies;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import io.swagger.v3.oas.annotations.media.Schema;
 import lombok.Builder;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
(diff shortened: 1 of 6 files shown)


andrescrz (changes requested):
Hi @awkoy before we go into the low level details of the change. The API would still work but it changes its contract, by returning a percentage.

Have you audited who are the users of this? It's just Comet internal? Only FE? Both FE and TS and Python SDKs?

Please provide more information.

awkoy:
Hi @andrescrz — audit results:

| Consumer | Status |
|---|---|
| Opik FE (`MetricsSummary.tsx`) | only real consumer; updated in this PR |
| Python SDK | Fern-generated stub only — no wrapper / example / test / docs reference (verified across `sdks/python/{src,tests}` outside `rest_api/`) |
| TypeScript SDK | same — Fern-generated stub only, no usage outside `rest_api/` |
| `ollie-assist` | no usage |
| Docs | auto-generated OpenAPI reference page only |

The endpoint is exposed under `/v1/private/...` so it is reachable via both SDK clients, but there's no higher-level adoption surface around it.

**Wire shape (field name, `Double` type, nullability) doesn't change** — only the numeric meaning of the `errors` metric. And the count semantics was already self-inconsistent: the FE labels it "Error rate" with `%`, and the sibling `/metrics` endpoint already returns this as a rate via the same `errored * 100 / total` formula (`ProjectMetricsDAO.error_rate`). The fix aligns the two.

If SDK-contract stability outweighs the consistency cost, I can switch to an additive change instead (add a new `error_rate` metric type, keep `errors` as a count). Let me know which you'd prefer.

andrescrz:
> Hi @andrescrz — audit results:
> 
> Consumer	Status
> Opik FE (`MetricsSummary.tsx`)	only real consumer; updated in this PR
> Python SDK	Fern-generated stub only — no wrapper / example / test / docs reference (verified across `sdks/python/{src,tests}` outside `rest_api/`)
> TypeScript SDK	same — Fern-generated stub only, no usage outside `rest_api/`
> `ollie-assist`	no usage
> Docs	auto-generated OpenAPI reference page only
> The endpoint is exposed under `/v1/private/...` so it is reachable via both SDK clients, but there's no higher-level adoption surface around it.
> 
> **Wire shape (field name, `Double` type, nullability) doesn't change** — only the numeric meaning of the `errors` metric. And the count semantics was already self-inconsistent: the FE labels it "Error rate" with `%`, and the sibling `/metrics` endpoint already returns this as a rate via the same `errored * 100 / total` formula (`ProjectMetricsDAO.error_rate`). The fix aligns the two.
> 
> If SDK-contract stability outweighs the consistency cost, I can switch to an additive change instead (add a new `error_rate` metric type, keep `errors` as a count). Let me know which you'd prefer.

Thanks for the audit. With all this in mind, we can change the semantics of the endpoint. Reviewing now.

andrescrz (approved):
LGTM.

PR #6867, opened 2026-05-26
[OPIK-6696] [BE] feat(datasets): copy_from coords on createOrUpdateDatasetItems + applyDatasetItemChanges

JetoPistola:
## Details

image

**TL;DR — moves the `INSERT FROM SELECT` that copies unchanged rows from "destination's just-minted prior version" to "a caller-supplied stable upstream version."** That's the structural fix for the multi-replica read-after-write window that drives OPIK-6674's data loss and the `qa-v1-stress-*` regressions. Default behavior is unchanged when the new fields are null.

The change is exposed on the wire as two paired optional fields — `copy_from_dataset_id` + `copy_from_version_id` — on both `DatasetItemBatch` (PUT `/v1/private/datasets/items`) and `DatasetItemChanges` (POST `/v1/private/datasets/{id}/items/changes`). When set, the BE redirects all three reads that touched destination v_(i-1) to read from the caller-supplied source instead: the COPY (`copyUnchangedItems`), the edit-via-SELECT-INSERT (`editItemsViaSelectInsert`), and the classification SELECT (`getItemIdsAndHashes`). Payload stays O(delta) — the two UUIDs are the only addition.

The structural problem: `opik migrate` chains version writes against a destination that may not have replicated yet. `applyDelta`'s `copyUnchangedItems` step silently drops rows that haven't reached the SELECT-serving replica. `applyDatasetItemChanges` additionally runs `editItemsViaSelectInsert` and `countRowsInVersion` against the same destination version with the same exposure. For migrate the fix is straightforward — point copy-from at source v_i, which was written long ago and is fully replicated.

**Design pivot from earlier commits:** the first commits on this branch shipped a `snapshot: Boolean` field that made the payload the source of truth for the new version's row set. That doesn't scale for migrate — the SDK can't send 10k items on every call to keep an existing 10k+1k version. The latest commits replace `snapshot` with the copy-from coordinates approach: payload stays delta-sized, BE points at a stable upstream version for carry-forwards.

Implementation highlights:
- `DatasetItemBatch` / `DatasetItemChanges` gain `copyFromDatasetId` + `copyFromVersionId` (nullable, must be set together — validator + inline check enforce the pairing).
- When the new fields are set, the service validates the source version exists in the workspace AND belongs to `copy_from_dataset_id` — returns 404 on miss. Prevents silently writing an empty-carry-forwards version from a bogus or cross-workspace pair.
- `DatasetItemVersionDAO.applyDelta`, `editItemsViaSelectInsert`, `copyVersionItems` signatures take explicit source vs. target dataset_id.
- `COPY_VERSION_ITEMS` and `EDIT_ITEM_VIA_SELECT_INSERT` SQL templates project `:targetDatasetId` (was `src.dataset_id`, a latent bug that worked only because source always equaled target).
- `createVersionWithDelta`'s classification SELECT (`getItemIdsAndHashes`) also reads from copy-from coords — reading from a stale destination base could mis-route an "edit" as an "add" and produce duplicate stable IDs.
- Existing intra-dataset callers (`DatasetVersionService`, filter-based batch delete / update) pass `datasetId` as both source and target → behavior unchanged.

## Change checklist
- [x] User facing
- [x] Documentation update

## Issues

- OPIK-6696
- Blocks: OPIK-6697 (`[SDK] feat(migrate): use snapshot mode for dataset version replay`) — SDK side will set the copy-from coords when chaining replay calls

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Scope: full implementation (DAO + service + DTOs + tests). Design pivoted mid-PR after the user surfaced the O(N²) scaling problem with the original snapshot-field approach
- Human verification: code review, scope refinement, test design review, local test run

## Testing

Ran the new `CopyFromSourceVersion` nested test class locally:

```
mvn test -Dtest='DatasetsResourceTest$CopyFromSourceVersion'
```

All 4 tests pass in ~1.8s:
- `createOrUpdate__copyFromSource__unchangedRowsCopiedFromSource` — a small delta sent to destination v_2 with copy_from pointing at a separate source dataset/version → destination v_2 contains the delta plus all of source's items.
- `createOrUpdate__noCopyFrom__carriesForwardFromDestination` — regression guard for the default path.
- `applyChanges__copyFromSource__readsFromSource` — same shape on the `/items/changes` endpoint, with a delete and an add, asserting source items appear in the destination and the destination's seed item is correctly deleted.
- `createOrUpdate__copyFromMissingVersion__returns404` — bogus `copy_from_version_id` returns 404 instead of silently writing an empty version.

Spotless clean. No frontend or SDK files changed. The migrate `qa-v1-stress-*` E2E depends on the SDK side (OPIK-6697) adopting the new fields; that will land on top.

## Documentation

N/A — the OpenAPI surface picks up the new fields automatically from Swagger annotations on `DatasetItemBatch` and `DatasetItemChanges`. The Fern regen workflow will publish them to the Python/TS SDKs on merge. Both endpoints' `@Operation` descriptions were updated to describe the copy-from semantics.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/DatasetItemBatch.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/DatasetItemBatch.java
@@ -35,7 +35,9 @@ public record DatasetItemBatch(
         @JsonView({
                 DatasetItem.View.Write.class}) @Schema(description = "Optional batch group ID to group multiple batches into a single dataset version. If null, mutates the latest version instead of creating a new one.") UUID batchGroupId,
         @JsonView({
-                DatasetItem.View.Write.class}) @Schema(description = "If true, the new version's row set equals exactly the items in this batch. The server skips the INSERT FROM SELECT that copies unchanged rows from the previous version, avoiding multi-replica read-after-write inconsistency. Items absent from the payload are absent from the new version (deletions implicit). Ignored when batch_group_id is null.") Boolean snapshot)
+                DatasetItem.View.Write.class}) @Schema(description = "OPIK-6696. Optional. Dataset to read carry-forward rows from when materializing the new version. Required together with copy_from_version_id. When both are set, the INSERT FROM SELECT that copies unchanged rows reads from this (dataset, version) pair instead of the destination dataset's prior version, avoiding the multi-replica read-after-write window. When null, the existing behavior applies (reads from the destination's prior version).") UUID copyFromDatasetId,
+        @JsonView({
+                DatasetItem.View.Write.class}) @Schema(description = "OPIK-6696. Optional. Version within copy_from_dataset_id to read carry-forward rows from. Required together with copy_from_dataset_id.") UUID copyFromVersionId)
         implements
             RateEventContainer {
 
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/DatasetItemBatch.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/DatasetItemBatch.java
@@ -34,10 +34,13 @@ public record DatasetItemBatch(
         @JsonView({DatasetItem.View.Write.class}) @NotNull @Size(min = 1, max = 1000) @Valid List<DatasetItem> items,
         @JsonView({
                 DatasetItem.View.Write.class}) @Schema(description = "Optional batch group ID to group multiple batches into a single dataset version. If null, mutates the latest version instead of creating a new one.") UUID batchGroupId,
+        // OPIK-6696: when both copy_from_* coordinates are set, the INSERT FROM SELECT that copies
+        // unchanged rows reads from this (dataset, version) pair instead of the destination dataset's
+        // prior version, avoiding the multi-replica read-after-write window.
         @JsonView({
-                DatasetItem.View.Write.class}) @Schema(description = "OPIK-6696. Optional. Dataset to read carry-forward rows from when materializing the new version. Required together with copy_from_version_id. When both are set, the INSERT FROM SELECT that copies unchanged rows reads from this (dataset, version) pair instead of the destination dataset's prior version, avoiding the multi-replica read-after-write window. When null, the existing behavior applies (reads from the destination's prior version).") UUID copyFromDatasetId,
+                DatasetItem.View.Write.class}) @Schema(description = "Optional. Dataset to read carry-forward rows from when materializing the new version. Required together with copy_from_version_id. When null, carry-forward rows are read from the destination dataset's prior version.") UUID copyFromDatasetId,
         @JsonView({
-                DatasetItem.View.Write.class}) @Schema(description = "OPIK-6696. Optional. Version within copy_from_dataset_id to read carry-forward rows from. Required together with copy_from_dataset_id.") UUID copyFromVersionId)
+                DatasetItem.View.Write.class}) @Schema(description = "Optional. Version within copy_from_dataset_id to read carry-forward rows from. Required together with copy_from_dataset_id.") UUID copyFromVersionId)
         implements
             RateEventContainer {
 
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/DatasetItemBatch.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/DatasetItemBatch.java
@@ -33,7 +33,9 @@ public record DatasetItemBatch(
                 DatasetItem.View.Write.class}) @Schema(description = "Optional. Associates the batch with a project by ID. Takes precedence over project_name.") UUID projectId,
         @JsonView({DatasetItem.View.Write.class}) @NotNull @Size(min = 1, max = 1000) @Valid List<DatasetItem> items,
         @JsonView({
-                DatasetItem.View.Write.class}) @Schema(description = "Optional batch group ID to group multiple batches into a single dataset version. If null, mutates the latest version instead of creating a new one.") UUID batchGroupId)
+                DatasetItem.View.Write.class}) @Schema(description = "Optional batch group ID to group multiple batches into a single dataset version. If null, mutates the latest version instead of creating a new one.") UUID batchGroupId,
+        @JsonView({
+                DatasetItem.View.Write.class}) @Schema(description = "If true, the new version's row set equals exactly the items in this batch. The server skips the INSERT FROM SELECT that copies unchanged rows from the previous version, avoiding multi-replica read-after-write inconsistency. Items absent from the payload are absent from the new version (deletions implicit). Ignored when batch_group_id is null.") Boolean snapshot)
         implements
             RateEventContainer {
 
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/DatasetItemChanges.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/DatasetItemChanges.java
@@ -44,9 +44,12 @@ public record DatasetItemChanges(
 
         @JsonView(DatasetItemChanges.View.Write.class) @Schema(description = "When true, clears the version-level execution policy (removes inherited policy from base version)") Boolean clearExecutionPolicy,
 
-        @JsonView(DatasetItemChanges.View.Write.class) @Schema(description = "OPIK-6696. Optional. Dataset to read carry-forward rows from when materializing the new version. Required together with copy_from_version_id. When both are set, the INSERT FROM SELECT that copies unchanged rows reads from this (dataset, version) pair instead of the destination dataset's prior version, avoiding the multi-replica read-after-write window. When null, the existing behavior applies (reads from the destination's prior version).") UUID copyFromDatasetId,
+        // OPIK-6696: when both copy_from_* coordinates are set, the INSERT FROM SELECT that copies
+        // unchanged rows reads from this (dataset, version) pair instead of the destination dataset's
+        // prior version, avoiding the multi-replica read-after-write window.
+        @JsonView(DatasetItemChanges.View.Write.class) @Schema(description = "Optional. Dataset to read carry-forward rows from when materializing the new version. Required together with copy_from_version_id. When null, carry-forward rows are read from the destination dataset's prior version.") UUID copyFromDatasetId,
 
-        @JsonView(DatasetItemChanges.View.Write.class) @Schema(description = "OPIK-6696. Optional. Version within copy_from_dataset_id to read carry-forward rows from. Required together with copy_from_dataset_id.") UUID copyFromVersionId) {
+        @JsonView(DatasetItemChanges.View.Write.class) @Schema(description = "Optional. Version within copy_from_dataset_id to read carry-forward rows from. Required together with copy_from_dataset_id.") UUID copyFromVersionId) {
 
     public static class View {
         public static class Write {
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/DatasetItemChanges.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/DatasetItemChanges.java
@@ -42,7 +42,11 @@ public record DatasetItemChanges(
 
         @JsonView(DatasetItemChanges.View.Write.class) @Valid @Schema(description = "Optional default execution policy for items in this version") ExecutionPolicy executionPolicy,
 
-        @JsonView(DatasetItemChanges.View.Write.class) @Schema(description = "When true, clears the version-level execution policy (removes inherited policy from base version)") Boolean clearExecutionPolicy) {
+        @JsonView(DatasetItemChanges.View.Write.class) @Schema(description = "When true, clears the version-level execution policy (removes inherited policy from base version)") Boolean clearExecutionPolicy,
+
+        @JsonView(DatasetItemChanges.View.Write.class) @Schema(description = "OPIK-6696. Optional. Dataset to read carry-forward rows from when materializing the new version. Required together with copy_from_version_id. When both are set, the INSERT FROM SELECT that copies unchanged rows reads from this (dataset, version) pair instead of the destination dataset's prior version, avoiding the multi-replica read-after-write window. When null, the existing behavior applies (reads from the destination's prior version).") UUID copyFromDatasetId,
+
+        @JsonView(DatasetItemChanges.View.Write.class) @Schema(description = "OPIK-6696. Optional. Version within copy_from_dataset_id to read carry-forward rows from. Required together with copy_from_dataset_id.") UUID copyFromVersionId) {
 
     public static class View {
         public static class Write {
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/DatasetsResource.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/DatasetsResource.java
@@ -488,9 +488,11 @@ public class DatasetsResource {
         }
     }
 
-    // OPIK-6696: the copy_from_* coordinates let callers pin the carry-forward read source to a
-    // specific (dataset, version) pair, avoiding the multi-replica read-after-write window when
-    // chaining version writes against a destination that may not have replicated yet.
+    /**
+     * OPIK-6696: the copy_from_* coordinates let callers pin the carry-forward read source to a
+     * specific (dataset, version) pair, avoiding the multi-replica read-after-write window when
+     * chaining version writes against a destination that may not have replicated yet.
+     */
     @PUT
     @Path("/items")
     @Operation(operationId = "createOrUpdateDatasetItems", summary = "Create/update dataset items", description = """
@@ -617,9 +619,11 @@ public class DatasetsResource {
         return Response.status(Response.Status.ACCEPTED).build();
     }
 
-    // OPIK-6696: the copy_from_* coordinates let callers pin the carry-forward (and edit-via-SELECT-INSERT)
-    // read source to a specific (dataset, version) pair, avoiding the multi-replica read-after-write window
-    // when chaining version writes against a destination that may not have replicated yet.
+    /**
+     * OPIK-6696: the copy_from_* coordinates let callers pin the carry-forward (and edit-via-SELECT-INSERT)
+     * read source to a specific (dataset, version) pair, avoiding the multi-replica read-after-write window
+     * when chaining version writes against a destination that may not have replicated yet.
+     */
     @POST
     @Path("/{id}/items/changes")
     @Operation(operationId = "applyDatasetItemChanges", summary = "Apply changes to dataset items", description = """
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/DatasetsResource.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/DatasetsResource.java
@@ -488,6 +488,9 @@ public class DatasetsResource {
         }
     }
 
+    // OPIK-6696: the copy_from_* coordinates let callers pin the carry-forward read source to a
+    // specific (dataset, version) pair, avoiding the multi-replica read-after-write window when
+    // chaining version writes against a destination that may not have replicated yet.
     @PUT
     @Path("/items")
     @Operation(operationId = "createOrUpdateDatasetItems", summary = "Create/update dataset items", description = """
@@ -495,11 +498,9 @@ public class DatasetsResource {
             Each item's 'id' field is the stable identifier and upsert key.
             Provide it to update an existing item, or omit it to create a new one.
 
-            OPIK-6696: set 'copy_from_dataset_id' and 'copy_from_version_id' together to read carry-forward
-            rows from the supplied (dataset, version) pair instead of the destination's prior version.
-            This avoids the multi-replica read-after-write window when chaining version writes against a
-            destination that may not have replicated yet. When the fields are null, the existing behavior
-            applies (carry-forward reads from the destination's prior version).""", responses = {
+            Set 'copy_from_dataset_id' and 'copy_from_version_id' together to read carry-forward rows
+            from the supplied (dataset, version) pair instead of the destination's prior version. When
+            the fields are null, carry-forward rows are read from the destination's prior version.""", responses = {
             @ApiResponse(responseCode = "204", description = "No content"),
     })
     @RateLimited
@@ -616,6 +617,9 @@ public class DatasetsResource {
         return Response.status(Response.Status.ACCEPTED).build();
     }
 
+    // OPIK-6696: the copy_from_* coordinates let callers pin the carry-forward (and edit-via-SELECT-INSERT)
+    // read source to a specific (dataset, version) pair, avoiding the multi-replica read-after-write window
+    // when chaining version writes against a destination that may not have replicated yet.
     @POST
     @Path("/{id}/items/changes")
     @Operation(operationId = "applyDatasetItemChanges", summary = "Apply changes to dataset items", description = """
@@ -628,11 +632,10 @@ public class DatasetsResource {
 
             Use `override=true` query parameter to force version creation even with stale baseVersion.
 
-            OPIK-6696: set 'copy_from_dataset_id' and 'copy_from_version_id' together on the request body
-            to read carry-forward rows (and the edit-via-SELECT-INSERT source rows) from the supplied
-            (dataset, version) pair instead of the destination's prior version. This avoids the
-            multi-replica read-after-write window when chaining version writes against a destination that
-            may not have replicated yet. When the fields are null, the existing behavior applies.
+            Set 'copy_from_dataset_id' and 'copy_from_version_id' together on the request body to read
+            carry-forward rows from the supplied (dataset, version) pair instead of the destination's
+            prior version. When the fields are null, carry-forward rows are read from the destination's
+            prior version.
             """, responses = {
             @ApiResponse(responseCode = "201", description = "Version created successfully", content = @Content(schema = @Schema(implementation = DatasetVersion.class))),
             @ApiResponse(responseCode = "400", description = "Bad Request", content = @Content(schema = @Schema(implementation = ErrorMessage.class))),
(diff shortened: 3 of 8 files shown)


apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
```diff
@@ -1621,6 +1621,22 @@ private Mono<DatasetVersion> insertItemsIntoVersion(DatasetItemBatch batch, UUID
...
+                        // payload rows are counted as "added"; callers in snapshot mode are
+                        // expected not to re-submit the same stable id within a batch group.
+                        if (Boolean.TRUE.equals(batch.snapshot())) {
+                            log.info("Inserting into version '{}' in snapshot mode: items='{}'",
+                                    versionId, normalizedItems.size());
+                            return versionDao
+                                    .insertItems(datasetId, versionId, normalizedItems, workspaceId, userName)
+                                    .then(Mono.fromCallable(() -> {
+                                        updateVersionCountsForInsert(versionId, workspaceId,
+                                                normalizedItems.size(), 0, userName);
```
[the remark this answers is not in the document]
JetoPistola: Skipping — this comment is against the now-removed `snapshot: Boolean` field. After the design pivot at commit 06e92dd2, the snapshot-mode code path no longer exists; the new shape is `copy_from_dataset_id` + `copy_from_version_id`. See the PR description for the full rationale.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
```diff
@@ -1901,6 +1925,59 @@ private Mono<DatasetVersion> createVersionWithDelta(UUID datasetId, UUID baseVer
...
+        log.info("Creating snapshot-mode version for dataset '{}', baseVersion '{}', itemCount '{}'",
+                datasetId, baseVersionId, items.size());
+
+        UUID newVersionId = idGenerator.generateId();
+
+        List<DatasetItem> snapshotItems = items.stream()
+                .map(item -> {
+                    UUID stableId = item.datasetItemId() != null
+                            ? item.datasetItemId()
+                            : (item.id() != null ? item.id() : idGenerator.generateId());
```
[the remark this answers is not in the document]
JetoPistola: Skipping — this comment refers to the now-removed `createVersionAsSnapshot` method. After the design pivot at commit 06e92dd2 the snapshot path was removed entirely; carry-forward rows are now read from caller-supplied source coordinates instead. The duplication you flagged no longer exists.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
```diff
@@ -1437,13 +1444,21 @@ public Mono<DatasetVersion> applyDeltaChanges(@NonNull UUID datasetId,
...
                     .collect(Collectors.toSet());
 
+            // OPIK-6696: when copy-from coordinates are supplied, the COPY (and supporting reads:
+            // pool sizing, edit-via-SELECT-INSERT) all run against the (dataset, version) pair the
+            // caller supplied instead of the destination's just-minted prior version. For migrate
+            // this is source v_i — stable and fully replicated, avoiding the multi-replica
+            // read-after-write window.
+            UUID copyDatasetId = changes.copyFromDatasetId() != null ? changes.copyFromDatasetId() : datasetId;
+            UUID copyVersionId = changes.copyFromVersionId() != null ? changes.copyFromVersionId() : baseVersionId;
+
```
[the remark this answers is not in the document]
JetoPistola: Skipping — extracting a helper for two pairs of one-line null-coalesces (`x != null ? x : defaultX`) introduces either a new record type or a tuple return for less code than it removes. Project code-style rule explicitly discourages helper functions for one-time/single-use abstractions. The two call sites stay inline.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
```diff
@@ -1437,13 +1444,21 @@ public Mono<DatasetVersion> applyDeltaChanges(@NonNull UUID datasetId,
...
             // OPIK-6390: size the unchanged-UUID pool from a live ClickHouse count instead of the
             // MySQL items_total, which can drift below the actual row count and silently truncate
             // the copy. Excludes the same ids the subsequent applyDelta excludes from the COPY.
             Set<UUID> excludedFromCopy = new HashSet<>(deletedIds);
             excludedFromCopy.addAll(editedDatasetItemIds);
 
-            return versionDao.countRowsInVersion(datasetId, baseVersionId, excludedFromCopy, null, workspaceId)
+            return versionDao.countRowsInVersion(copyDatasetId, copyVersionId, excludedFromCopy, null, workspaceId)
```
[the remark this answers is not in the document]
JetoPistola: Fixed in 6e23fbf77 — applyDeltaChanges now calls `versionService.getVersionById(workspaceId, copyDatasetId, copyVersionId)` upfront when `copy_from_dataset_id` is supplied. Throws `NotFoundException` (404) on missing or cross-workspace pairs; also verifies the resolved version's `datasetId` matches `copy_from_dataset_id` (404 if not). Covered by new test `createOrUpdate__copyFromMissingVersion__returns404`.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
```diff
@@ -1816,14 +1836,24 @@ private Mono<DatasetVersion> createFirstVersion(UUID datasetId, List<DatasetItem
...
+                datasetId, baseVersionId, items.size(), copyDatasetId, copyVersionId);
 
         UUID newVersionId = idGenerator.generateId();
 
-        // Get existing item IDs from the base version to determine adds vs edits
-        return versionDao.getItemIdsAndHashes(datasetId, baseVersionId)
+        // Get existing item IDs from the copy-from version to determine adds vs edits
+        return versionDao.getItemIdsAndHashes(copyDatasetId, copyVersionId)
                 .collectList()
                 .flatMap(existingItems -> {
```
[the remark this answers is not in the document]
JetoPistola: Fixed in 6e23fbf77 — same validation added to `createVersionWithDelta`. When `copy_from_dataset_id` is supplied, calls `versionService.getVersionById(workspaceId, copyDatasetId, copyVersionId)` before `getItemIdsAndHashes` and the COPY. Returns 404 on missing / cross-workspace / wrong-dataset version. New test asserts the 404 behavior.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -3010,12 +3035,13 @@ private Mono<Long> getCount(DatasetItemSearchCriteria criteria, UUID versionId)
...
+            List<DatasetItemFilter> excludeFilters, @NonNull List<UUID> uuids) {
 
         log.info(
-                "Copying items from version '{}' to version '{}' for dataset '{}', excludeFilters='{}', uuidPoolSize='{}'",
-                sourceVersionId, targetVersionId, datasetId,
+                "Copying items from (dataset '{}', version '{}') to (dataset '{}', version '{}'), excludeFilters='{}', uuidPoolSize='{}'",
+                sourceDatasetId, sourceVersionId, targetDatasetId, targetVersionId,
                 excludeFilters != null ? excludeFilters.size() : 0, uuids.size());
 
         return Mono.deferContextual(ctx -> {
```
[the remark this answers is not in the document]
JetoPistola: Fixed in 6e23fbf77 — demoted both the entry-point and success log lines in `copyVersionItems` from INFO to DEBUG. The service layer (`applyDelta` / `applyDeltaChanges`) remains the single source of truth for operational INFO telemetry; the DAO now contributes diagnostics at DEBUG only.

CometActions:

### Access Information
- **URL:** https://pr-6867.dev.comet.com
- **Namespace:** pr-6867
- **Version:** 2.0.49-6867-merge-2175
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-6867%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-6867`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -1830,7 +1850,8 @@ INSERT INTO dataset_item_versions (
             SELECT
                 :newId as id,
                 src.dataset_item_id,
-                src.dataset_id,
+                -- OPIK-6696: target dataset_id, not source's (cross-dataset edit-via-SELECT-INSERT)
+                :targetDatasetId as dataset_id,
```
thiagohora: Void adding a comment in the middle of the query. Add it on Java only

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -1912,7 +1933,11 @@ INSERT INTO dataset_item_versions (
                    arrayElement(:uuids, src.rn),
                    toString(generateUUIDv7())) AS id,
                 src.dataset_item_id,
-                src.dataset_id,
+                -- OPIK-6696: target dataset_id, not source's. When copy_from_dataset_id differs from
+                -- the destination, the read source is a different dataset (e.g. migrate replay reads
+                -- from the source workspace's dataset and writes into the destination workspace's
+                -- dataset). The inserted rows must carry the destination dataset_id.
```
thiagohora: Same

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/DatasetsResource.java
```diff
@@ -493,7 +493,13 @@ public ChunkedOutput<JsonNode> streamDatasetItems(
...
             Create/update dataset items based on dataset item id.
             Each item's 'id' field is the stable identifier and upsert key.
-            Provide it to update an existing item, or omit it to create a new one.""", responses = {
+            Provide it to update an existing item, or omit it to create a new one.
+
+            OPIK-6696: set 'copy_from_dataset_id' and 'copy_from_version_id' together to read carry-forward
+            rows from the supplied (dataset, version) pair instead of the destination's prior version.
+            This avoids the multi-replica read-after-write window when chaining version writes against a
+            destination that may not have replicated yet. When the fields are null, the existing behavior
+            applies (carry-forward reads from the destination's prior version).""", responses = {
```
thiagohora: The internal ticket id shouldn't be in the published API docs. This `@Operation` description ships into the OpenAPI spec and from there into the generated Python/TS SDK docs, so end users would see `OPIK-6696:` and the replication-internals jargon.

Suggest keeping the description limited to the user-facing `copy_from_*` contract and moving the ticket reference + multi-replica read-after-write rationale into a plain `//` code comment above the `@Operation`.

Same applies to the `applyDatasetItemChanges` `@Operation` (line 631).

apps/opik-backend/src/main/java/com/comet/opik/api/DatasetItemBatch.java
```diff
@@ -33,7 +33,11 @@ public record DatasetItemBatch(
...
         @JsonView({DatasetItem.View.Write.class}) @NotNull @Size(min = 1, max = 1000) @Valid List<DatasetItem> items,
         @JsonView({
-                DatasetItem.View.Write.class}) @Schema(description = "Optional batch group ID to group multiple batches into a single dataset version. If null, mutates the latest version instead of creating a new one.") UUID batchGroupId)
+                DatasetItem.View.Write.class}) @Schema(description = "Optional batch group ID to group multiple batches into a single dataset version. If null, mutates the latest version instead of creating a new one.") UUID batchGroupId,
+        @JsonView({
+                DatasetItem.View.Write.class}) @Schema(description = "OPIK-6696. Optional. Dataset to read carry-forward rows from when materializing the new version. Required together with copy_from_version_id. When both are set, the INSERT FROM SELECT that copies unchanged rows reads from this (dataset, version) pair instead of the destination dataset's prior version, avoiding the multi-replica read-after-write window. When null, the existing behavior applies (reads from the destination's prior version).") UUID copyFromDatasetId,
```
thiagohora: Same here: the `OPIK-6696.` prefix and the `INSERT FROM SELECT ... multi-replica read-after-write window` internals are part of the `@Schema` description, which is published to the OpenAPI surface and propagates to the generated SDK field docs.

Keep the `@Schema` to the user-facing contract (what the field does, that it must be set together with `copy_from_version_id`, the null-default behavior) and move the ticket id + implementation rationale into a code comment.

Applies to `copyFromVersionId` (line 40) as well.

apps/opik-backend/src/main/java/com/comet/opik/api/DatasetItemChanges.java
```diff
@@ -42,7 +42,11 @@ public record DatasetItemChanges(
 
         @JsonView(DatasetItemChanges.View.Write.class) @Valid @Schema(description = "Optional default execution policy for items in this version") ExecutionPolicy executionPolicy,
 
-        @JsonView(DatasetItemChanges.View.Write.class) @Schema(description = "When true, clears the version-level execution policy (removes inherited policy from base version)") Boolean clearExecutionPolicy) {
+        @JsonView(DatasetItemChanges.View.Write.class) @Schema(description = "When true, clears the version-level execution policy (removes inherited policy from base version)") Boolean clearExecutionPolicy,
+
+        @JsonView(DatasetItemChanges.View.Write.class) @Schema(description = "OPIK-6696. Optional. Dataset to read carry-forward rows from when materializing the new version. Required together with copy_from_version_id. When both are set, the INSERT FROM SELECT that copies unchanged rows reads from this (dataset, version) pair instead of the destination dataset's prior version, avoiding the multi-replica read-after-write window. When null, the existing behavior applies (reads from the destination's prior version).") UUID copyFromDatasetId,
```
thiagohora: Same `OPIK-6696.`-in-`@Schema` issue as `DatasetItemBatch`: drop the ticket prefix and the internal rationale from the published description and move them into a code comment, leaving the `@Schema` user-facing. Applies to both `copyFromDatasetId` (line 47) and `copyFromVersionId` (line 49).

thiagohora (commented):
A few spots leak the internal ticket id (`OPIK-6696`) and implementation rationale into the published API surface — the Swagger `@Operation`/`@Schema` descriptions. These flow into the OpenAPI spec and the generated Python/TS SDK docs, so end users would see them.

Recommend keeping the descriptions user-facing and moving the ticket reference + multi-replica read-after-write rationale into plain code comments. Non-blocking. Details inline.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
thiagohora: Fixed in a79f414d09 — removed the inline `--` comment from inside `EDIT_ITEM_VIA_SELECT_INSERT` and moved the rationale to a Java `//` comment above the constant.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
thiagohora: Fixed in a79f414d09 — removed the inline `--` comment from inside `COPY_VERSION_ITEMS` and folded the rationale into the Java `//` comment block above the constant.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/DatasetsResource.java
thiagohora: Fixed in a79f414d09 — dropped the `OPIK-6696:` prefix and the multi-replica read-after-write rationale from both the `createOrUpdateDatasetItems` and `applyDatasetItemChanges` `@Operation` descriptions; they now state only the user-facing `copy_from_*` contract. The ticket id + internals moved into `//` comments above each method.

apps/opik-backend/src/main/java/com/comet/opik/api/DatasetItemBatch.java
thiagohora: Fixed in a79f414d09 — stripped `OPIK-6696.` and the INSERT-FROM-SELECT/multi-replica internals from the `@Schema` descriptions of `copyFromDatasetId` and `copyFromVersionId`; the published descriptions are now user-facing only, with the rationale in a `//` comment above the fields.

apps/opik-backend/src/main/java/com/comet/opik/api/DatasetItemChanges.java
thiagohora: Fixed in a79f414d09 — same treatment for `DatasetItemChanges`: the `copyFromDatasetId`/`copyFromVersionId` `@Schema` descriptions are now user-facing only, ticket id + internals moved to a `//` comment above the fields.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/DatasetsResource.java
```diff
@@ -488,12 +488,19 @@ public ChunkedOutput<JsonNode> streamDatasetItems(
         }
     }
 
+    // OPIK-6696: the copy_from_* coordinates let callers pin the carry-forward read source to a
+    // specific (dataset, version) pair, avoiding the multi-replica read-after-write window when
+    // chaining version writes against a destination that may not have replicated yet.
```
andrescrz: Nit: all these better as javadocs. Applies everywhere in the PR.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
```diff
@@ -1369,6 +1370,12 @@ public Mono<DatasetVersion> applyDeltaChanges(@NonNull UUID datasetId,
             log.info("Applying delta changes for dataset '{}', baseVersion '{}', override '{}'",
                     datasetId, changes.baseVersion(), override);
 
+            // OPIK-6696: copy_from_dataset_id and copy_from_version_id must be set together.
+            if ((changes.copyFromDatasetId() == null) != (changes.copyFromVersionId() == null)) {
+                return Mono.error(new BadRequestException(
+                        "copy_from_dataset_id and copy_from_version_id must be provided together"));
+            }
```
andrescrz, line 1379: Minor: double check all the entry points to here. If we're always covered at the endpoint level, you don't need the extra validation here. If you weren't covered, you should aim to be covered there instead of here. Once all that is covered, this validation wouldn't be needed. But ok to keep it if you wish, it's just some extra validation in the code and the `BadRequestException` is thrown a the appropriate place (the service level).

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
```diff
@@ -1437,13 +1444,32 @@ public Mono<DatasetVersion> applyDeltaChanges(@NonNull UUID datasetId,
...
+            // this is source v_i — stable and fully replicated, avoiding the multi-replica
+            // read-after-write window.
+            UUID copyDatasetId = changes.copyFromDatasetId() != null ? changes.copyFromDatasetId() : datasetId;
+            UUID copyVersionId = changes.copyFromVersionId() != null ? changes.copyFromVersionId() : baseVersionId;
+
+            // Validate caller-supplied copy-from coords resolve to a real (workspace-scoped) version
+            // on the named dataset. Without this, a bogus or cross-workspace pair would silently
+            // produce a new version with zero carry-forwards instead of returning an error.
+            if (changes.copyFromDatasetId() != null) {
+                DatasetVersion sourceVersion = versionService.getVersionById(workspaceId, copyDatasetId, copyVersionId);
```
andrescrz: Minor: double check the condition here, it's not aligned with what is actually used: `copyFromDatasetId` vs `copyDatasetId`

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
```diff
@@ -1816,14 +1847,36 @@ private Mono<DatasetVersion> createFirstVersion(UUID datasetId, List<DatasetItem
...
+        // on the named dataset. Without this, a bogus or cross-workspace pair would silently produce
+        // a new version with all incoming items classified as adds (the classification SELECT would
+        // return empty) and no carry-forwards (the COPY would also return empty).
+        if (copyFromDatasetId != null) {
+            DatasetVersion sourceVersion = versionService.getVersionById(workspaceId, copyDatasetId, copyVersionId);
+            if (!sourceVersion.datasetId().equals(copyDatasetId)) {
+                return Mono.error(new NotFoundException(
+                        "Version '%s' does not belong to dataset '%s'".formatted(copyVersionId, copyDatasetId)));
+            }
+        }
```
andrescrz: Same here about the misalignment between the condition and the value really used. 

In addition, if this is exactly copied and pasted, this is one of those place where you should DRY and encapsulate the validation in small aux methods.

andrescrz (commented):
Generally in the good direction, let's iterate on it before moving forward.

apps/opik-backend/src/main/java/com/comet/opik/api/validation/DatasetItemBatchValidator.java
```diff
@@ -8,6 +8,21 @@ public class DatasetItemBatchValidator implements ConstraintValidator<DatasetIte
...
+        // OPIK-6696: copy_from_dataset_id and copy_from_version_id must be set together (or both null).
+        boolean hasCopyFromDataset = datasetItemBatch.copyFromDatasetId() != null;
+        boolean hasCopyFromVersion = datasetItemBatch.copyFromVersionId() != null;
+        if (hasCopyFromDataset != hasCopyFromVersion) {
+            context.disableDefaultConstraintViolation();
+            context.buildConstraintViolationWithTemplate(
+                    "copy_from_dataset_id and copy_from_version_id must be provided together")
+                    .addConstraintViolation();
+            return false;
+        }
```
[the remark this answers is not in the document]
thiagohora: Fixed in 0a9af18b70 — extracted the shared check into `DatasetItemBatchValidator.isCopyFromPairConsistent(...)` + the `COPY_FROM_PAIR_MESSAGE` constant. The `DatasetItemBatch` validator and the `DatasetItemChanges` service-layer guard (`applyDeltaChanges`) now call the same method and reuse the same error text, so the rule and message stay in sync.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/DatasetsResource.java
thiagohora: Fixed in 0a9af18b70 — converted the rationale comments above `createOrUpdateDatasetItems` and `applyDatasetItemChanges` to `/** javadoc */`. Scoped to the endpoint methods; left the `--`/`//` notes inside the private SQL-constant blocks in the DAO as line comments, since javadoc on a `private static final String` constant doesn't render as API docs.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
thiagohora: Kept — but refactored in 0a9af18b70. Unlike `DatasetItemBatch` (which has the `@DatasetItemBatchValidation` class-level validator), `DatasetItemChanges` has no class-level validator, so this service-layer check is the only pair-wise guard for the `/items/changes` endpoint. It now delegates to the shared `DatasetItemBatchValidator.isCopyFromPairConsistent(...)` so it can't drift from the batch endpoint's rule. Happy to add a dedicated DTO validator for symmetry in a follow-up if you'd prefer endpoint-level coverage.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
thiagohora: Fixed in 0a9af18b70 — the resolve+validate logic moved into a shared `resolveAndValidateCopyFrom(...)` helper that validates the caller-supplied `copyFromDatasetId`/`copyFromVersionId` directly (the values it actually keys the lookup on), instead of gating on `copyFromDatasetId` while operating on the resolved `copyDatasetId`/`copyVersionId`. Condition and values are now aligned. (Functionally it was equivalent before — inside the guard the resolved values always equalled the caller-supplied ones — but the new form is unambiguous.)

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
thiagohora: Fixed in 0a9af18b70 — the duplicated block in `createVersionWithDelta` and `applyDeltaChanges` is now the single `resolveAndValidateCopyFrom(workspaceId, copyFromDatasetId, copyFromVersionId, fallbackDatasetId, fallbackVersionId)` helper returning a `CopyFromCoordinates` record. That removes the copy-paste and fixes the same condition/value alignment as the sibling thread.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
```diff
@@ -1815,15 +1839,51 @@ private Mono<DatasetVersion> createFirstVersion(UUID datasetId, List<DatasetItem
                 });
     }
 
+    private record CopyFromCoordinates(UUID datasetId, UUID versionId) {
```
andrescrz, line 1842: Nit: in these cases you can use `Pair` class from Apache commons to avoid creating this kind of aux records for method return values.

andrescrz (approved):
LGTM, well done!

PR #6871, opened 2026-05-26
[OPIK-6703] [BE] fix: zero-rows retry guard + v2-client INSERT...SELECT on dataset_item_versions

thiagohora:
## Details

INSERT...SELECT on `dataset_item_versions` can return 0 rows when the SELECT lands on a ClickHouse replica that hasn't yet seen the source version's parts (cross-replica race under `async_insert` with no quorum). With the previous r2dbc path, the row count was read from an interim `Progress` event and was unreliable for this query shape (`clickhouse-java#2860`) — the resulting truncated INSERT was silently committed with a wrong `items_total`.

This PR moves the COPY and EDIT paths in `DatasetItemVersionDAOImpl` to the v2 ClickHouse client (authoritative `written_rows` from `Statistics`) and wraps them in a configurable retry-on-zero-rows guard that surfaces a typed `ZeroRowsWrittenException` on exhaustion instead of committing the truncated state.

Key changes:
- Application-side retry guard, configurable via `datasetVersioning.zeroRowsRetry.{maxAttempts,minBackoffMillis,maxBackoffMillis}`.
- `copyVersionItems` / `copyUnchangedItems` / `editItemsViaSelectInsert` routed through the v2 ClickHouse client. Retries are idempotent: caller-supplied UUIDs + ReplacingMergeTree dedup on the sort key.
- New v2-client entry points on `FilterQueryBuilder` (`toAnalyticsDbFiltersV2Client`, `bindV2Client`, public `formatStringArrayLiteral`) sharing logic with the r2dbc `bind` path via a private `bindUsing(BiConsumer)` core.
- Injection-hardened array literal formatter (backslashes doubled first, then single quotes doubled — ClickHouse accepts both `''` and `\'` escapes).
- Non-blocking query path via `Mono.fromFuture(Supplier)` with outer `subscribeOn(Schedulers.boundedElastic())` so `clickHouseClient.query()`, SQL/params build, and the blocking `QueryResponse.close()` stay off the reactive event loop while the in-flight query doesn't park a worker.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-6703
- OPIK-6674

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): claude-opus-4-7
- Scope: implementation, unit tests, this PR description
- Human verification: design review across iterations; integration tests (112/112 DatasetVersionResourceTest) and unit tests (14/14 FilterQueryBuilderV2ClientTest) executed locally

## Testing

```
mvn -B test -Dtest='DatasetVersionResourceTest'        # 112/112 pass
mvn -B test -Dtest='FilterQueryBuilderV2ClientTest'    # 14/14 pass
```

Scenarios validated:
- Happy path: `copyVersionItems`, `copyUnchangedItems` with and without `excludedIds` / `excludeFilters`.
- `editItemsViaSelectInsert` across combinations of `data` / `tags` / `description` / `evaluators` / `execution_policy` / `clearExecutionPolicy`.
- `batchUpdateByFilters` path (`PatchItemsWithVersioning`).
- Apply-delta full flow (add + edit + delete + unchanged copy) via `ApplyDatasetItemChanges`.
- Injection vectors: single-quote and backslash+single-quote payloads on the array-literal formatter.
- Retry idempotency: same `newRowIds` reused across attempts; ReplacingMergeTree collapses duplicates on `(workspace_id, dataset_id, dataset_version_id, id)`.

The previously flaky `CreateFromTracesAndSpans` test (which was masking the silent-truncation behaviour upstream) now passes consistently.

## Documentation

No user-facing or public API changes. New config block `datasetVersioning.zeroRowsRetry` is documented inline in `config.yml` with defaults and env-var override pattern.


apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -3008,6 +3023,66 @@ private Mono<Long> getCount(DatasetItemSearchCriteria criteria, UUID versionId)
...
+        int maxAttempts = retryCfg.getMaxAttempts();
+        Duration minBackoff = Duration.ofMillis(retryCfg.getMinBackoffMillis());
+        Duration maxBackoff = Duration.ofMillis(retryCfg.getMaxBackoffMillis());
+        return operation
+                .flatMap(actual -> {
+                    if (actual == 0L) {
+                        return Mono.error(new ZeroRowsWrittenException(
+                                "%s returned 0 rows; expected %s".formatted(operationLabel, expectedRows)));
+                    }
+                    return Mono.just(actual);
```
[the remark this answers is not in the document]
thiagohora: Acknowledged but not changing — design choice. The retry guard is intentionally narrow to OPIK-6674's catastrophic-zero signature: `actual == 0L` for the whole call. A stricter `actual < expectedRows` check would false-positive for legitimate causes — in the COPY path the expected count is derived from the conservative pool sizing (`pool.size() / UUID_POOL_MULTIPLIER`), but the actual carry-forward can legitimately be lower when callers pass `excludeFilters` or `excludedIds`. Detecting partial truncation requires a separate mechanism beyond the scope of this PR.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -3018,51 +3093,97 @@ public Mono<Long> copyVersionItems(@NonNull UUID datasetId, @NonNull UUID source
...
+    private static String formatUuidArrayParam(Collection<UUID> ids) {
+        StringBuilder sb = new StringBuilder("[");
+        boolean first = true;
+        for (UUID id : ids) {
+            if (!first) sb.append(',');
+            sb.append('\'').append(id).append('\'');
+            first = false;
+        }
+        return sb.append(']').toString();
     }
```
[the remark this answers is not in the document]
thiagohora: Fixed in 0cf2fa5359 — `formatUuidArrayParam` dropped; replaced with a thin `uuidsToArrayLiteral` delegate that calls `FilterQueryBuilder.formatStringArrayLiteral(ids.stream().map(UUID::toString).toList())`. `UUID.toString()` is `[0-9a-f-]` only so the per-element escape is a no-op.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -3122,124 +3253,117 @@ public Mono<Long> editItemsViaSelectInsert(@NonNull UUID datasetId, @NonNull UUI
 
         long itemCount = editedItems.size();
 
-        return Mono.deferContextual(ctx -> {
-            String workspaceId = ctx.get(RequestContext.WORKSPACE_ID);
-            String userName = ctx.get(RequestContext.USER_NAME);
+        return retryIfZeroRows(
+                executeEditItemsViaSelectInsert(datasetId, baseVersionId, newVersionId, editedItems, newRowIds),
+                itemCount, "editItemsViaSelectInsert")
+                .map(actualSum -> itemCount);
```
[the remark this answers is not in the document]
thiagohora: Same rationale as the COPY-path comment above — "retry only on zeros" policy. A per-item zero is not necessarily a stale-replica miss: the EDIT path uses `INSERT ... SELECT` against a `LIMIT 1` lookup keyed by `dataset_item_id`, so a missing source row (item not in the base version) legitimately produces 0 rows for that item. Triggering retry on a per-item zero would false-positive on legitimately-absent items. The sum-based check trades full coverage of partial-truncation for fewer false retries; broader detection is a separate effort.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -3122,124 +3253,114 @@ public Mono<Long> editItemsViaSelectInsert(@NonNull UUID datasetId, @NonNull UUI
...
+            // Array(String) params must be ClickHouse array literals — the v2 client serialises
+            // Map values via String.valueOf and would otherwise emit Java's unquoted [a, b] form.
+            params.put("data_keys", FilterQueryBuilder.formatStringArrayLiteral(dataAsStrings.keySet()));
+            params.put("data_values", FilterQueryBuilder.formatStringArrayLiteral(dataAsStrings.values()));
+        }
+        if (edit.description() != null) {
+            params.put("description", edit.description());
+        }
+        if (edit.tags() != null) {
+            params.put("tags", FilterQueryBuilder.formatStringArrayLiteral(edit.tags()));
```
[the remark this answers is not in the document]
thiagohora: Not changing — this matches pre-existing behavior. The previous r2dbc path (`statement.bind("tags", edit.tags().toArray(new String[0]))`) also NPEs on null elements during driver-level binding. Null-validation for `DatasetItemEdit.tags` belongs at the API boundary (DTO validation), not the DAO; adding it here would mask data-quality bugs further upstream.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -3018,53 +3093,98 @@ public Mono<Long> copyVersionItems(@NonNull UUID datasetId, @NonNull UUID source
...
-            String query = template.render();
-
-            // Convert UUIDs to String array for ClickHouse binding
-            String[] uuidStrings = uuids.stream()
-                    .map(UUID::toString)
-                    .toArray(String[]::new);
+        return retryIfZeroRows(
+                executeCopyVersionItems(datasetId, sourceVersionId, targetVersionId, uuids,
+                        null /* excludedIds */, excludeFilters),
+                expectedRowsFromPool(uuids), "copyVersionItems");
```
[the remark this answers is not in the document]
thiagohora: UUID-array rendering is exercised end-to-end by `DatasetVersionResourceTest` (112/112 pass, including `DeleteItemsWithVersioning`, `PatchItemsWithVersioning`, `ApplyDatasetItemChanges`, `BatchVersioningTests`) and `DatasetsCsvUploadResourceTest#uploadCsvFile__largeBatch` (which carries forward 2000 UUIDs). A dedicated unit test for `retryIfZeroRows` semantics is a known gap — captured as a follow-up in OPIK-6703.

apps/opik-backend/src/main/java/com/comet/opik/domain/filter/FilterQueryBuilder.java
```diff
@@ -1070,24 +1121,80 @@ public static Statement bind(
...
+     * (backslash, single-quote, backtick, newline, tab, etc.). This closes injection vectors
+     * like {@code x';DROP TABLE...} and {@code x\';...} by emitting {@code \'} and {@code \\}.
+     *
+     * @throws NullPointerException if {@code values} is null or contains a null element
+     */
+    public static String formatStringArrayLiteral(@NonNull Collection<String> values) {
+        StringBuilder sb = new StringBuilder("[");
+        boolean first = true;
+        for (String v : values) {
+            Objects.requireNonNull(v, "ClickHouse Array(String) literal element must not be null");
```
BorisTkachenko: Might be better to add validation to the method signature instead of
`Objects.requireNonNull(v, "ClickHouse Array(String) literal element must not be null");`

To follow existing pattern:
`public static String formatStringArrayLiteral(@NonNull Collection<@NonNull String> values) {`

apps/opik-backend/src/main/java/com/comet/opik/domain/filter/FilterQueryBuilder.java
```diff
@@ -1033,6 +1047,43 @@ public static Statement bind(
...
+     * <p>Multi-value operator values are pre-rendered as a ClickHouse array literal string
+     * (e.g. {@code ['a','b']}) because the v2 client serialises Map values via
+     * {@code String.valueOf}, which would otherwise emit an unquoted Java array {@code [a, b]}.
+     *
+     * <p>Pair with {@link #toAnalyticsDbFiltersV2Client} for the matching SQL fragment.
+     */
+    public static void bindV2Client(
+            @NonNull Map<String, Object> params,
+            @NonNull List<? extends Filter> filters,
+            @NonNull FilterStrategy filterStrategy) {
```
BorisTkachenko: Name here is confusing, since we do not bind anything here. We just parse strings and populate `params` Map with it. While the other `bind()` method actually binds them to a `Statement`

BorisTkachenko (dismissed):
Left a few comments, both not critical. Great job.

andrescrz (changes requested):
Why a defensive improvement needs to change the queries and their templating? Instead of using ST and bind as usual?

thiagohora:
@andrescrz fair question — the query/templating change isn't cosmetic, and the `FilterQueryBuilder` change in particular is a knock-on of one hard requirement: getting a **trustworthy written-row count**.

The chain:

1. **Root cause** — the carry-forward `INSERT … SELECT` can write 0 rows when the SELECT lands on a stale replica, and we silently commit a truncated version. The guard's whole job is to *detect* that, which means we must know how many rows the INSERT actually wrote.

2. **r2dbc can't tell us reliably** — `Result.getRowsUpdated()` for `INSERT … SELECT` is populated from an interim `Progress` event, not the final statistics, so it can report 0 on a successful write (and vice-versa). That's the unreliable signal at the heart of the bug; filed upstream as `ClickHouse/clickhouse-java#2860`. The guard can't be built on it.

3. **v2 client gives an authoritative count** — `QueryResponse.getWrittenRows()` comes from the server `Statistics`. That's why the COPY/EDIT execution moved to the v2 client, and the v2 client only understands `{name:Type}` placeholders, not r2dbc's `:name`.

4. **Why the filter had to change** — the COPY query carries an optional `excludeFilters` clause, and that clause is generated by `FilterQueryBuilder`, which emits `:name` placeholders. Once COPY runs on the v2 client, that fragment has to be `{name:Type}`. Rather than fork the operator/strategy logic, `toAnalyticsDbFiltersV2Client` / `bindV2Client` delegate to the **same internal core** (`bindUsing`) as the existing r2dbc `bind()` path — it's a thin syntax adapter over shared logic, not a second implementation.

So: the templating churn is the cost of a reliable row count, and the filter change is the cost of keeping the existing `excludeFilters` feature working on that count-reliable path without duplicating its logic.

We did weigh the lighter alternative (keep r2dbc + ST + `bind()` untouched, verify with a separate `SELECT count()` on the new version, retry on a false zero). It's a smaller diff but adds a round-trip per INSERT and still leaves the EDIT path on the unreliable `getRowsUpdated()`. Happy to switch to that if you'd prefer the smaller blast radius over the per-call extra query — your call.

apps/opik-backend/src/main/java/com/comet/opik/domain/filter/FilterQueryBuilder.java
```diff
@@ -1033,6 +1047,43 @@ public static Statement bind(
...
+        bindUsing((name, value) -> {
+            if (value instanceof String[] arr) {
+                params.put(name, formatStringArrayLiteral(Arrays.asList(arr)));
```
andrescrz, line 1073: Question: how maintainable this is for other Collections in the future. I noticed that in some parts of the code we invoke `toArray` e.g: `toArray(UUID[]::new)` to bind collections, usually list/set. 

However, people has replicated this code in many place but it's not really necessary. The R2DB client seems to properly bind this by just passing the collection. There are parts of the code working perfectly find without the `toArray` call.

Please investigate this a bit and let me know more about this. I'm concerned about maintainability for other collections, for making widespread the `toArray` pattern if not really needed.

apps/opik-backend/src/main/java/com/comet/opik/domain/filter/FilterQueryBuilder.java
```diff
@@ -1033,6 +1047,43 @@ public static Statement bind(
...
+        }, filters, filterStrategy);
+    }
+
+    /**
+     * Core binding logic shared by the r2dbc {@link Statement} and the v2-client
+     * {@code Map<String, Object>} entry points above. Iterates filters and emits param
+     * (name, value) pairs through the supplied {@code binder}.
+     */
+    private static void bindUsing(
+            @NonNull java.util.function.BiConsumer<String, Object> binder,
```
andrescrz: Nit: use imports.

apps/opik-backend/src/main/java/com/comet/opik/domain/filter/FilterQueryBuilder.java
```diff
@@ -1070,24 +1121,80 @@ public static Statement bind(
...
+     * longer name first.
+     *
+     * <p>Package-private: {@link #toAnalyticsDbFiltersV2Client} is the public entry point
+     * that bundles this rewrite with the SQL generation.
+     *
+     * @param sql     SQL fragment containing {@code :name} placeholders
+     * @param filters filters that were used to build {@code sql}; their order determines
+     *                the param indices ({@code :filter0}, {@code :filter1}, …)
+     */
+    static String rewritePlaceholdersForV2Client(@NonNull String sql, @NonNull List<? extends Filter> filters) {
```
andrescrz, line 1163: Minor: make private or mark as visible for testing with Guava annotation. If effectively private, you likely want to revisit the argument validations.

apps/opik-backend/src/main/java/com/comet/opik/domain/filter/FilterQueryBuilder.java
```diff
@@ -1070,24 +1121,80 @@ public static Statement bind(
...
+        String out = sql;
+        for (int i = filters.size() - 1; i >= 0; i--) {
+            Filter filter = filters.get(i);
+            String filterType = Operator.MULTI_VALUE_OPERATORS.contains(filter.operator())
+                    ? "Array(String)"
+                    : "String";
+            out = out.replace(":dynamicJsonPath" + i, "{dynamicJsonPath" + i + ":String}");
+            out = out.replace(":dynamicField" + i, "{dynamicField" + i + ":String}");
+            out = out.replace(":filterKey" + i, "{filterKey" + i + ":String}");
+            out = out.replace(":filter" + i, "{filter" + i + ":" + filterType + "}");
```
andrescrz, line 1173: Same concern about maintainability here, not only for the array, but for all keys such as `dynamicJsonPath`. These are utility methods that should be generic and neutral. It'd be desirable to not to rely on how queries are actually written and their specific field names to bind.

apps/opik-backend/src/main/java/com/comet/opik/domain/filter/FilterQueryBuilder.java
```diff
@@ -1070,24 +1121,80 @@ public static Statement bind(
...
+     * Java array {@code [a, b]} that the server rejects.
+     *
+     * <p>Per-element escaping delegates to {@link ClickHouseUtil#escape(String)}, the upstream
+     * ClickHouse JDBC helper that handles the full C-style escape set the server accepts
+     * (backslash, single-quote, backtick, newline, tab, etc.). This closes injection vectors
+     * like {@code x';DROP TABLE...} and {@code x\';...} by emitting {@code \'} and {@code \\}.
+     *
+     * @throws NullPointerException if {@code values} is null or contains a null element
+     */
+    public static String formatStringArrayLiteral(@NonNull Collection<String> values) {
```
andrescrz: Related to my other question. If this handles a Collection, maybe the related `instanceof` check can be extended. Anyway, I'd rather get rid of formatting our data when binding. 

If that's really required, it'd be a big gap on the V2 client.

I'd explore the option to delegate any data conversion to the client. Let's try to find a way for that and avoid pre-processing on our end.

apps/opik-backend/src/main/java/com/comet/opik/domain/filter/FilterQueryBuilder.java
```diff
@@ -1070,24 +1121,80 @@ public static Statement bind(
...
+        StringBuilder sb = new StringBuilder("[");
```
andrescrz: Mentioned before but I'd rather not implement this method. But if we have no choice, let's not implement it manually. There are plenty of 1st and 3rd party libraries supporting this logic to concat strings with a separator char and prefix/suffix. Please explore the options: 
- Java built-in libraries:`Collectors.joining`, `String.join` `StringJoiner`
- Apache Commons: `StringUtils.join`. 
- Guava: `Joiner`.

apps/opik-backend/src/main/java/com/comet/opik/domain/filter/FilterQueryBuilder.java
andrescrz: This would be great. Similar works for Jakarta validations but not sure if Lombok supports that, but we could test. You can alternatively use Guava preconditions for this.

Anyhow, I'd move these validations to first thing in the method.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -3008,6 +3023,66 @@ private Mono<Long> getCount(DatasetItemSearchCriteria criteria, UUID versionId)
         });
     }
 
+    // OPIK-6674: zero-rows retry guard for INSERT...SELECT on dataset_item_versions.
```
andrescrz: Defining all these classes (exception, retry logic etc.) is out the responsibility of this particular DAO. 

That should be defined as client infrastructure auxiliary code and the DAO should be an user of that. 

This might be used in the future for other DAOs, and we don't want to be coupled.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -3008,6 +3023,66 @@ private Mono<Long> getCount(DatasetItemSearchCriteria criteria, UUID versionId)
...
+        public ZeroRowsWrittenException(String message) {
+            super(message);
+        }
+    }
+
+    /**
+     * Reverses {@link FilterUtils#generateUuidPool}'s sizing (count × UUID_POOL_MULTIPLIER, floor 1)
+     * to recover the expected row count. A pool size ≤ 1 corresponds to count 0 (the floor kicked in).
+     */
+    private static int expectedRowsFromPool(List<UUID> pool) {
```
andrescrz: Similarly, likely belongs to the same class whose behaviour is reversing. That's `FilterUtils`.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -3008,6 +3023,66 @@ private Mono<Long> getCount(DatasetItemSearchCriteria criteria, UUID versionId)
...
+                    return Mono.just(actual);
+                })
+                .retryWhen(Retry
+                        .backoff(maxAttempts, minBackoff)
+                        .maxBackoff(maxBackoff)
+                        .filter(ZeroRowsWrittenException.class::isInstance)
+                        .doBeforeRetry(rs -> log.warn(
+                                "[{}] returned 0 rows; retrying (attempt {}/{})",
+                                operationLabel, rs.totalRetries() + 1, maxAttempts))
+                        .onRetryExhaustedThrow((spec, signal) -> signal.failure()));
```
andrescrz: These retry implementation is based on `reactor`, whereas you introduce the V2 client. I understand that mostly for capturing properly the number of inserted rows.

However, this isn't quite aligned. Maybe the V2 client offers configurable retry policy. 

Anyway, this would be fine to keep if V2 client has no proper retry policy or not granular enought, specially considering that we need high level logic for this case.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/DatasetVersioningConfig.java
```diff
@@ -0,0 +1,57 @@
...
+    public static class ZeroRowsRetry {
+
+        /**
+         * Number of retry attempts when an INSERT...SELECT writes zero rows
+         * while the input set was non-empty. Set to 0 to disable retries
+         * (the guard still surfaces the failure as an exception). Total
+         * call count is {@code maxAttempts + 1} (initial + retries).
+         */
+        @JsonProperty
+        @Min(0) @Max(10) private int maxAttempts = 3;
```
andrescrz: Minor: I think we should generally stop this pattern of defining defaults in the Java code, so it's all centralised in the config files (prod and test).

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/DatasetVersioningConfig.java
```diff
@@ -0,0 +1,57 @@
...
+
+        /**
+         * Minimum backoff between retries, in milliseconds. The actual delay
+         * grows exponentially from this value up to {@link #maxBackoffMillis}.
+         */
+        @JsonProperty
+        @Min(0) @Max(60000) private long minBackoffMillis = 100;
```
andrescrz: Minor: these two should better be defined as durations.

andrescrz (commented):
I understand now why the change on the queries. This is a major change, I left more specific feedback before delivering it. 

Let me know if any question.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
thiagohora: Done in f67e71ecd5 — moved out of the DAO into reusable client infrastructure: \`infrastructure/db/ZeroRowsWrittenException\` + \`ZeroRowsRetryPolicy\`, provided via \`DatabaseAnalyticsModule\`. The DAO now just injects and calls the policy, so other DAOs issuing INSERT...SELECT on replicated tables can reuse it.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
thiagohora: Done in f67e71ecd5 — moved \`expectedRowsFromPool\` into \`FilterUtils\`, right next to \`generateUuidPool\` whose sizing it reverses.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
thiagohora: Kept the reactor-based retry and documented why in \`ZeroRowsRetryPolicy\`: the v2 client's \`setMaxRetries\`/\`retryOnFailures\` only cover transport-level \`ClientFaultCause\`s (connection errors, 503s). They have no concept of "the INSERT succeeded at the HTTP layer but wrote 0 rows", which is the business-level condition we must detect — so it has to be high-level logic on our side.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/DatasetVersioningConfig.java
thiagohora: Done in f67e71ecd5 — \`DatasetVersioningConfig\` is now a record with no Java-code defaults; the values live only in \`config.yml\` and \`config-test.yml\`.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/DatasetVersioningConfig.java
thiagohora: Done in f67e71ecd5 — \`minBackoff\`/\`maxBackoff\` are now dropwizard \`Duration\` with \`@MinDuration\`/\`@MaxDuration\` validation; yml uses \`100ms\`/\`1s\`.

apps/opik-backend/src/main/java/com/comet/opik/domain/filter/FilterQueryBuilder.java
thiagohora: Done in f67e71ecd5 — \`bindUsing\` now uses an imported \`BiConsumer\`.

apps/opik-backend/src/main/java/com/comet/opik/domain/filter/FilterQueryBuilder.java
thiagohora: Done in f67e71ecd5 — marked \`@VisibleForTesting\` (Guava). It is only reached by the public \`toAnalyticsDbFiltersV2Client\` plus the unit test.

apps/opik-backend/src/main/java/com/comet/opik/domain/filter/FilterQueryBuilder.java
thiagohora: Done in f67e71ecd5 — \`formatStringArrayLiteral\` now builds via \`Collectors.joining(",", "[", "]")\` instead of a hand-rolled StringBuilder. Per-element escaping still delegates to the upstream \`ClickHouseUtil.escape\`. See the investigation note on the sibling thread for why we still need to render the literal ourselves.

apps/opik-backend/src/main/java/com/comet/opik/domain/filter/FilterQueryBuilder.java
thiagohora: Investigated — it is a real gap in the v2 client, not something we can delegate. At the HTTP boundary the v2 client serializes every statement param via \`String.valueOf(entry.getValue())\` (\`HttpAPIClientHelper\` line 534), so passing a \`Collection\`/array yields Java's \`[a, b]\` form, which ClickHouse rejects for \`Array(String)\`. There is no typed/collection binding in client-v2 0.9.0. So for the v2 path we have to render the \`['a','b']\` literal ourselves. I documented this on \`formatStringArrayLiteral\`. (Worth a follow-up upstream issue if we want typed array params.)

apps/opik-backend/src/main/java/com/comet/opik/domain/filter/FilterQueryBuilder.java
thiagohora: You're right about r2dbc — confirmed it binds collections directly (e.g. \`CommentDAO\` binds \`Set<UUID> entityIds\`, \`ExperimentDAO\` binds \`Set<UUID> experimentIds\`, both without \`toArray\`). So the \`toArray\` is unnecessary on the r2dbc path. The catch: that \`toArray\` lives in the shared \`bindUsing\` core, which feeds BOTH the r2dbc \`bind(Statement)\` path (used by traces/spans/experiments/... many queries) AND the v2 \`populateV2ClientParams\` path (which needs the formatted literal per the v2 gap above). Dropping it would shift ENUM IN/NOT_IN binding behavior across every filter-using query, so I scoped it out of this PR. Happy to do it as a focused follow-up if you want.

apps/opik-backend/src/main/java/com/comet/opik/domain/filter/FilterQueryBuilder.java
thiagohora: Acknowledged. The placeholder rewrite (\`rewritePlaceholdersForV2Client\`) does have to know the param-name convention (\`filter\`/\`filterKey\`/\`dynamicField\`/\`dynamicJsonPath\`) because it maps each to its ClickHouse type (\`Array(String)\` vs \`String\`). It shares the exact same names with the r2dbc \`bind\` path via the common \`bindUsing\` core, so they cannot drift. Making it fully field-name-agnostic would mean parsing arbitrary \`:identifier\` tokens and inferring types another way — a larger change I would rather not fold into this PR. Open to a follow-up.

apps/opik-backend/src/main/java/com/comet/opik/domain/filter/FilterQueryBuilder.java
thiagohora: Done in f67e71ecd5 — renamed to \`populateV2ClientParams\` to reflect that it populates the params Map rather than binding to a Statement.

apps/opik-backend/src/main/java/com/comet/opik/domain/filter/FilterQueryBuilder.java
thiagohora: Done in f67e71ecd5 — signature is now \`@NonNull Collection<@NonNull String> values\`, and the inline \`Objects.requireNonNull\` is gone.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/ZeroRowsRetryPolicy.java
```diff
@@ -0,0 +1,60 @@
...
+     * as {@link ZeroRowsWrittenException} and retried with exponential backoff. After the
+     * attempts are exhausted the exception propagates instead of resolving to the truncated 0.
+     *
+     * <p>When {@code expectedRows <= 0} the guard is a no-op: 0 written rows is a valid outcome.
+     */
+    public Mono<Long> retryOnZeroRows(@NonNull Mono<Long> operation, long expectedRows,
+            @NonNull String operationLabel) {
+        if (expectedRows <= 0L) {
+            return operation;
+        }
```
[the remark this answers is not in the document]
thiagohora: Added in 0215808eda — \`ZeroRowsRetryPolicyTest\` covers all four paths with StepVerifier: \`expectedRows<=0\` bypass, non-zero pass-through, 0-rows-then-success (retries then resolves), and exhaustion surfacing \`ZeroRowsWrittenException\`. The DAO COPY/EDIT integration is already exercised by \`DatasetVersionResourceTest\` (112) + \`DatasetsCsvUploadResourceTest\` large-batch.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/OpikConfiguration.java
```diff
@@ -123,6 +123,9 @@ public class OpikConfiguration extends JobConfiguration {
     @Valid @NotNull @JsonProperty
     private DatasetVersioningMigrationConfig datasetVersioningMigration = new DatasetVersioningMigrationConfig();
 
+    @Valid @NotNull @JsonProperty
+    private DatasetVersioningConfig datasetVersioning = DatasetVersioningConfig.builder().build();
+
```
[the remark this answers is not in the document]
thiagohora: Already present — \`OpikConfiguration\` line 126-127 has \`@Valid @NotNull @JsonProperty\` on \`datasetVersioning\`, so Dropwizard does deserialize the block and \`@Valid\` cascades into \`ZeroRowsRetry\` (enforcing \`@Max(10)\` etc.). The \`builder().build()\` default leaving \`zeroRowsRetry\` null is the same pattern as the existing \`AlertProjectMigrationConfig\`/\`DatasetProjectMigrationConfig\` defaults; if YAML ever omitted the block, \`@Valid @NotNull\` would fail config validation at startup (fail-fast) rather than NPE in \`getZeroRowsRetryPolicy\`.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/DatabaseAnalyticsModule.java
```diff
@@ -55,6 +55,14 @@ public Client getClickHouseClient() {
         return clickHouseClient;
     }
 
+    @Provides
+    @Singleton
+    public ZeroRowsRetryPolicy getZeroRowsRetryPolicy(OpikConfiguration configuration) {
+        var retry = configuration.getDatasetVersioning().zeroRowsRetry();
+        return new ZeroRowsRetryPolicy(retry.maxAttempts(),
+                retry.minBackoff().toJavaDuration(), retry.maxBackoff().toJavaDuration());
+    }
```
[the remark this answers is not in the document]
thiagohora: Both \`config.yml\` and \`config-test.yml\` do define \`datasetVersioning.zeroRowsRetry\` (added in this PR — \`maxAttempts\`, \`minBackoff\`, \`maxBackoff\`), so startup binds non-null values. And as above, \`@Valid @NotNull\` on the field makes a missing block a fail-fast validation error, not a runtime NPE.

apps/opik-backend/src/main/java/com/comet/opik/domain/filter/FilterQueryBuilder.java
andrescrz: Thanks for the explanation. Ok to scope it out of this PR.

apps/opik-backend/src/main/java/com/comet/opik/domain/filter/FilterQueryBuilder.java
andrescrz: Sounds good. Fine to go in a separated PR.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/DatabaseAnalyticsModule.java
```diff
@@ -55,6 +55,14 @@ public Client getClickHouseClient() {
         return clickHouseClient;
     }
 
+    @Provides
+    @Singleton
+    public ZeroRowsRetryPolicy getZeroRowsRetryPolicy(OpikConfiguration configuration) {
+        var retry = configuration.getDatasetVersioning().zeroRowsRetry();
+        return new ZeroRowsRetryPolicy(retry.maxAttempts(),
+                retry.minBackoff().toJavaDuration(), retry.maxBackoff().toJavaDuration());
```
andrescrz, line 63: Nit: always favour builders over constructors.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/ZeroRowsRetryPolicy.java
```diff
@@ -0,0 +1,60 @@
...
+ * decoupled from any specific config class — callers pass the resolved retry parameters.
+ */
+@Slf4j
+public class ZeroRowsRetryPolicy {
+
+    private final int maxAttempts;
+    private final Duration minBackoff;
+    private final Duration maxBackoff;
+
+    public ZeroRowsRetryPolicy(int maxAttempts, @NonNull Duration minBackoff, @NonNull Duration maxBackoff) {
```
andrescrz, line 30: Minor: boiler-plate constructor better with Lombok.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/ZeroRowsWrittenException.java
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * Raised when a ClickHouse {@code INSERT ... SELECT} reports zero written rows while the
+ * caller expected a non-empty result — typically a cross-replica race where the SELECT
+ * ran against a replica that did not yet have the source parts visible.
+ *
+ * <p>Used as the retry signal by {@link ZeroRowsRetryPolicy}. See OPIK-6674.
+ */
+public class ZeroRowsWrittenException extends RuntimeException {
+
+    public ZeroRowsWrittenException(String message) {
```
andrescrz, line 12: Nit: you could use Lombok `@StandardException`. Still experimental, but it should be ok.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/DatasetVersioningConfig.java
```diff
@@ -0,0 +1,38 @@
...
+@Builder(toBuilder = true)
+public record DatasetVersioningConfig(
+        @Valid @NotNull ZeroRowsRetry zeroRowsRetry) {
+
+    @Builder(toBuilder = true)
+    public record ZeroRowsRetry(
+            @Min(0) @Max(10) int maxAttempts,
+            @NotNull @MinDuration(value = 0, unit = TimeUnit.MILLISECONDS) //
+            @MaxDuration(value = 1, unit = TimeUnit.MINUTES) Duration minBackoff,
+            @NotNull @MinDuration(value = 0, unit = TimeUnit.MILLISECONDS) //
```
andrescrz, line 35: Nit: should the min durations be 1 millisecond instead of 0?

andrescrz (dismissed):
Only a few optional minor and nits on my end. Well done! This is good to go.

apps/opik-backend/src/main/java/com/comet/opik/domain/filter/FilterQueryBuilder.java
thiagohora: Thanks, I will address all remaining items on the follow-up PR.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/DatabaseAnalyticsModule.java
thiagohora: Thanks, I will address all remaining items on the follow-up PR.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -3048,54 +3073,83 @@ public Mono<Long> copyVersionItems(@NonNull UUID sourceDatasetId, @NonNull UUID
...
+
+            Segment segment = startSegment(DATASET_ITEM_VERSIONS, CLICKHOUSE, "copy_version_items");
+            return Mono.fromFuture(() -> clickHouseClient.query(sql, params, settings))
+                    .flatMap(response -> Mono.fromCallable(() -> {
+                        try (response) {
+                            long written = response.getWrittenRows();
+                            log.info(
+                                    "Copied '{}' items from (dataset '{}', version '{}') to (dataset '{}', version '{}')",
+                                    written, sourceDatasetId, sourceVersionId, targetDatasetId, targetVersionId);
+                            return written;
```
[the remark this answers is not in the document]
thiagohora: Skipping — these copy/edit row counts are deliberate version-mutation events we keep at INFO for production traceability of `dataset_item_versions` writes (one line per COPY/EDIT operation, not per row). Leaving the level and placement as-is.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -3159,135 +3223,120 @@ public Mono<Long> editItemsViaSelectInsert(@NonNull UUID sourceDatasetId, @NonNu
...
-                    UUID newRowId = newRowIds.get(i);
+        return makeMonoContextAware((userName, workspaceId) -> {
+            Segment segment = startSegment(DATASET_ITEM_VERSIONS, CLICKHOUSE, "edit_items_via_select_insert");
+            return Flux.range(0, editedItems.size())
+                    .concatMap(i -> executeEditOneItem(editedItems.get(i), newRowIds.get(i),
+                            sourceDatasetId, sourceVersionId, targetDatasetId, newVersionId, userName, workspaceId))
+                    .reduce(0L, Long::sum)
+                    .doOnSuccess(actualSum -> log.info(
+                            "Edited '{}' items via SELECT INSERT into (dataset '{}', version '{}') (actual rows written: {})",
+                            editedItems.size(), targetDatasetId, newVersionId, actualSum))
```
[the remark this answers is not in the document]
thiagohora: Skipping — these copy/edit row counts are deliberate version-mutation events we keep at INFO for production traceability of `dataset_item_versions` writes (one line per COPY/EDIT operation, not per row). Leaving the level and placement as-is.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -3048,54 +3073,83 @@ public Mono<Long> copyVersionItems(@NonNull UUID sourceDatasetId, @NonNull UUID
...
-            String query = template.render();
-
-            // Convert UUIDs to String array for ClickHouse binding
-            String[] uuidStrings = uuids.stream()
-                    .map(UUID::toString)
-                    .toArray(String[]::new);
+        return zeroRowsRetryPolicy.retryOnZeroRows(
+                executeCopyVersionItems(sourceDatasetId, sourceVersionId, targetDatasetId, targetVersionId, uuids,
+                        null /* excludedIds */, excludeFilters),
+                FilterUtils.expectedRowsFromPool(uuids), "copyVersionItems");
```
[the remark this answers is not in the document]
thiagohora: Fixed in 80c135a6ed — gated the zero-rows guard on `excludeFilters`. The UUID pool (hence `expectedRowsFromPool`) is sized for the worst case (all items), so it was always `> 0`; but the delete and batch-update carry-forward callers pass real filters that can legitimately exclude every source row. Now `expectedRows` is `expectedRowsFromPool(uuids)` only when `excludeFilters` is empty (the unfiltered carry-forward path OPIK-6674 protects); when filters are present we pass `0`, which bypasses the guard so a valid "copy none" no longer retries into `ZeroRowsWrittenException`. Good catch.

andrescrz (approved):

PR #6873, opened 2026-05-26
[OPIK-6626] [BE] feat: support multiple environments per prompt version

BorisTkachenko:
## Details

Extends the prompt version environment model from a single `environment` field to a `Set<String> environments`, allowing one version to be pinned to multiple environments simultaneously.

**Schema changes:**
- New `prompt_version_envs` table (migration `000077`) tracks environment-to-version mappings with soft-delete via `ended_at`, replacing the `environment` column on `prompt_versions`
- Indexes added for efficient lookup by workspace+prompt+environment and by workspace+version

**API changes:**
- `PromptVersion.environment` → `environments: Set<String>` (nullable, max 100 entries, each validated against `Environment.NAME_PATTERN`)
- `PromptVersionEnvironmentUpdate.environment` → `environments: Set<String>` (non-null; empty set clears all; previously nullable string mapped to clear behavior)
- Validation updated: mask versions reject non-empty `environments` sets

**Service / DAO changes:**
- `PromptService.setVersionEnvironment` accepts `Set<String>` and performs: close all current env assignments for the version, close any existing ownerships for the incoming envs on the same prompt, then batch-insert the new assignments
- `createVersionWithEnvironment` checks for conflicts via `findTakenEnvironments` (set-based) instead of single-version lookup; raises `409` listing all conflicting env names
- `savePromptVersion` now also calls `saveEnvironments` in the same transaction
- Queries that return `PromptVersion` objects use a correlated subquery (`JSON_ARRAYAGG`) to populate `environments` from the new table
- `EnvironmentService.findExistingNames` added to validate the full set of requested envs before assignment

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues
- OPIK-6626

## Testing

Integration tests in `PromptResourceTest` updated to cover:
- Creating a version with multiple environments
- Setting multiple environments via `PUT /v1/private/prompts/versions/{id}/environment`
- Conflict detection when two versions claim overlapping environments
- Clearing all environments with an empty set
- Mask version rejection when `environments` is non-empty
- Environment existence validation (404 on unknown names)

## Documentation

No public documentation changes — internal API behavior change. Callers previously passing a single `environment` string must migrate to `environments: [<name>]` (or `environments: []` to clear).

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/PromptVersion.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/PromptVersion.java
@@ -50,7 +50,7 @@ public record PromptVersion(
         @JsonView({PromptVersion.View.Public.class, Prompt.View.Detail.class,
                 PromptVersion.View.Detail.class}) @Schema(description = "version type discriminator; defaults to prompt_version") PromptVersionType versionType,
         @JsonView({PromptVersion.View.Public.class, Prompt.View.Detail.class,
-                PromptVersion.View.Detail.class}) @Nullable @Pattern(regexp = Environment.NAME_PATTERN, message = Environment.NAME_PATTERN_MESSAGE) @Size(max = 150, message = "cannot exceed 150 characters") String environment,
+                PromptVersion.View.Detail.class}) @Nullable @Size(max = 100, message = "cannot exceed 100 environments") Set<@Pattern(regexp = Environment.NAME_PATTERN, message = Environment.NAME_PATTERN_MESSAGE) @Size(max = 150, message = "cannot exceed 150 characters") String> environments,
         @JsonView({PromptVersion.View.Public.class, Prompt.View.Detail.class,
                 PromptVersion.View.Detail.class}) String changeDescription,
         @JsonView({PromptVersion.View.Public.class, Prompt.View.Detail.class,
@@ -102,7 +102,7 @@ public record PromptVersion(
     }
 
     @JsonIgnore
-    @AssertTrue(message = "environment cannot be set on a mask version") public boolean isEnvironmentCompatibleWithVersionType() {
-        return environment == null || versionType() != PromptVersionType.MASK;
+    @AssertTrue(message = "environments cannot be set on a mask version") public boolean isEnvironmentCompatibleWithVersionType() {
+        return (environments == null || environments.isEmpty()) || versionType() != PromptVersionType.MASK;
     }
 }
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/PromptVersion.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/PromptVersion.java
@@ -15,6 +15,7 @@ import jakarta.validation.constraints.NotBlank;
 import jakarta.validation.constraints.Pattern;
 import jakarta.validation.constraints.Size;
 import lombok.Builder;
+import org.apache.commons.collections4.CollectionUtils;
 import org.jdbi.v3.json.Json;
 
 import java.time.Instant;
@@ -103,6 +104,6 @@ public record PromptVersion(
 
     @JsonIgnore
     @AssertTrue(message = "environments cannot be set on a mask version") public boolean isEnvironmentCompatibleWithVersionType() {
-        return (environments == null || environments.isEmpty()) || versionType() != PromptVersionType.MASK;
+        return CollectionUtils.isEmpty(environments) || versionType() != PromptVersionType.MASK;
     }
 }
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/PromptVersionEnvironmentUpdate.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/PromptVersionEnvironmentUpdate.java
@@ -4,20 +4,23 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.databind.PropertyNamingStrategies;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import io.swagger.v3.oas.annotations.media.Schema;
-import jakarta.annotation.Nullable;
+import jakarta.validation.constraints.NotNull;
 import jakarta.validation.constraints.Pattern;
 import jakarta.validation.constraints.Size;
 import lombok.Builder;
+
+import java.util.Set;
 
 @Builder(toBuilder = true)
 @JsonIgnoreProperties(ignoreUnknown = true)
 @JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
 @Schema(description = """
-        Set or clear the environment owned by a prompt version.
-        - non-null name: maps the environment to this version; if another version of the same prompt currently owns the environment, ownership is moved atomically.
-        - null: clears the environment from this version.
-        The environment must already exist in the workspace registry; unknown names return 404.
+        Replace the full set of environments assigned to a prompt version.
+        The provided set becomes the new complete set of environments for this version.
+        - Non-empty set: assigns each environment to this version; if another version of the same prompt currently owns any of those environments, ownership is moved atomically.
+        - Empty set: clears all environments from this version.
+        All environments must already exist in the workspace registry; unknown names return 404.
         """)
 public record PromptVersionEnvironmentUpdate(
-        @Nullable @Pattern(regexp = Environment.NAME_PATTERN, message = Environment.NAME_PATTERN_MESSAGE) @Size(max = 150, message = "cannot exceed 150 characters") String environment) {
+        @NotNull @Size(max = 100, message = "cannot exceed 100 environments") Set<@Pattern(regexp = Environment.NAME_PATTERN, message = Environment.NAME_PATTERN_MESSAGE) @Size(max = 150, message = "cannot exceed 150 characters") String> environments) {
 }
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/EnvironmentDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/EnvironmentDAO.java
@@ -61,6 +61,10 @@ interface EnvironmentDAO {
     @SqlQuery("SELECT COUNT(*) FROM environments WHERE workspace_id = :workspaceId AND name = :name")
     long countByName(@Bind("workspaceId") String workspaceId, @Bind("name") String name);
 
+    @SqlQuery("SELECT name FROM environments WHERE workspace_id = :workspaceId AND name IN (<names>)")
+    Set<String> findExistingNames(@Bind("workspaceId") String workspaceId,
+            @BindList("names") Set<String> names);
+
     @SqlQuery("SELECT COUNT(*) FROM environments WHERE workspace_id = :workspaceId")
     long countByWorkspace(@Bind("workspaceId") String workspaceId);
 
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/EnvironmentDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/EnvironmentDAO.java
@@ -68,10 +68,6 @@ interface EnvironmentDAO {
         return findNames(workspaceId, null);
     }
 
-    default Set<String> findExistingNames(String workspaceId, Set<String> names) {
-        return findNames(workspaceId, names);
-    }
-
     @SqlQuery("SELECT COUNT(*) FROM environments WHERE workspace_id = :workspaceId AND name = :name")
     long countByName(@Bind("workspaceId") String workspaceId, @Bind("name") String name);
 
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/EnvironmentDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/EnvironmentDAO.java
@@ -4,12 +4,15 @@ import com.comet.opik.api.Environment;
 import com.comet.opik.infrastructure.db.UUIDArgumentFactory;
 import org.jdbi.v3.sqlobject.config.RegisterArgumentFactory;
 import org.jdbi.v3.sqlobject.config.RegisterConstructorMapper;
+import org.jdbi.v3.sqlobject.customizer.AllowUnusedBindings;
 import org.jdbi.v3.sqlobject.customizer.Bind;
 import org.jdbi.v3.sqlobject.customizer.BindList;
 import org.jdbi.v3.sqlobject.customizer.BindMethods;
+import org.jdbi.v3.sqlobject.customizer.Define;
 import org.jdbi.v3.sqlobject.statement.SqlBatch;
 import org.jdbi.v3.sqlobject.statement.SqlQuery;
 import org.jdbi.v3.sqlobject.statement.SqlUpdate;
+import org.jdbi.v3.stringtemplate4.UseStringTemplateEngine;
 
 import java.util.List;
 import java.util.Optional;
@@ -55,15 +58,22 @@ interface EnvironmentDAO {
     @SqlQuery("SELECT * FROM environments WHERE workspace_id = :workspaceId ORDER BY created_at ASC")
     List<Environment> findAll(@Bind("workspaceId") String workspaceId);
 
-    @SqlQuery("SELECT name FROM environments WHERE workspace_id = :workspaceId")
-    List<String> findAllNames(@Bind("workspaceId") String workspaceId);
+    @SqlQuery("SELECT name FROM environments WHERE workspace_id = :workspaceId <if(names)> AND name IN (<names>) <endif>")
+    @UseStringTemplateEngine
+    @AllowUnusedBindings
+    Set<String> findNames(@Bind("workspaceId") String workspaceId,
+            @Define("names") @BindList(onEmpty = BindList.EmptyHandling.NULL_VALUE, value = "names") Set<String> names);
+
+    default Set<String> findAllNames(String workspaceId) {
+        return findNames(workspaceId, null);
+    }
+
+    default Set<String> findExistingNames(String workspaceId, Set<String> names) {
+        return findNames(workspaceId, names);
+    }
 
     @SqlQuery("SELECT COUNT(*) FROM environments WHERE workspace_id = :workspaceId AND name = :name")
     long countByName(@Bind("workspaceId") String workspaceId, @Bind("name") String name);
-
-    @SqlQuery("SELECT name FROM environments WHERE workspace_id = :workspaceId AND name IN (<names>)")
-    Set<String> findExistingNames(@Bind("workspaceId") String workspaceId,
-            @BindList("names") Set<String> names);
 
     @SqlQuery("SELECT COUNT(*) FROM environments WHERE workspace_id = :workspaceId")
     long countByWorkspace(@Bind("workspaceId") String workspaceId);
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/EnvironmentService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/EnvironmentService.java
@@ -19,6 +19,7 @@ import jakarta.ws.rs.core.Response;
 import lombok.NonNull;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.jdbi.v3.core.statement.UnableToExecuteStatementException;
 import reactor.core.publisher.Mono;
@@ -203,7 +204,7 @@ class EnvironmentServiceImpl implements EnvironmentService {
 
     @Override
     public Set<String> findExistingNames(@NonNull Set<String> names, @NonNull String workspaceId) {
-        if (names.isEmpty()) {
+        if (CollectionUtils.isEmpty(names)) {
             return Set.of();
         }
         return template.inTransaction(READ_ONLY, handle -> handle.attach(EnvironmentDAO.class)
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/EnvironmentService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/EnvironmentService.java
@@ -203,12 +203,12 @@ class EnvironmentServiceImpl implements EnvironmentService {
     }
 
     @Override
-    public Set<String> findExistingNames(@NonNull Set<String> names, @NonNull String workspaceId) {
+    public Set<String> findExistingNames(Set<String> names, @NonNull String workspaceId) {
         if (CollectionUtils.isEmpty(names)) {
             return Set.of();
         }
         return template.inTransaction(READ_ONLY, handle -> handle.attach(EnvironmentDAO.class)
-                .findExistingNames(workspaceId, names));
+                .findNames(workspaceId, names));
     }
 
     private void trackEnvironmentCreatedEvent(Environment environment, String workspaceId, String userName,
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/EnvironmentService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/EnvironmentService.java
@@ -57,6 +57,8 @@ public interface EnvironmentService {
 
     boolean existsByName(String name, String workspaceId);
 
+    Set<String> findExistingNames(Set<String> names, String workspaceId);
+
     Environment update(UUID id, EnvironmentUpdate environmentUpdate);
 
     Environment get(UUID id);
@@ -197,6 +199,15 @@ class EnvironmentServiceImpl implements EnvironmentService {
             var repository = handle.attach(EnvironmentDAO.class);
             return repository.countByName(workspaceId, name) > 0;
         });
+    }
+
+    @Override
+    public Set<String> findExistingNames(@NonNull Set<String> names, @NonNull String workspaceId) {
+        if (names.isEmpty()) {
+            return Set.of();
+        }
+        return template.inTransaction(READ_ONLY, handle -> handle.attach(EnvironmentDAO.class)
+                .findExistingNames(workspaceId, names));
     }
 
     private void trackEnvironmentCreatedEvent(Environment environment, String workspaceId, String userName,
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/PromptDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/PromptDAO.java
@@ -55,70 +55,81 @@ public interface PromptDAO {
     void save(@Bind("workspace_id") String workspaceId, @BindMethods("bean") Prompt prompt);
 
     @SqlQuery("""
+            WITH pv_for_prompt AS (
+                SELECT pv.*
+                FROM prompt_versions pv
+                WHERE pv.prompt_id = :id AND pv.workspace_id = :workspace_id
+            ), active_envs AS (
+                SELECT pve.version_id, pve.environment
+                FROM prompt_version_envs pve
+                INNER JOIN pv_for_prompt pfp ON pfp.id = pve.version_id
+                WHERE pve.workspace_id = :workspace_id AND pve.ended_at IS NULL
+            ), ver_envs AS (
+                SELECT version_id, JSON_ARRAYAGG(environment) AS environments
+                FROM active_envs
+                GROUP BY version_id
+            )
             SELECT
-                *,
+                p.*,
                 (
-                    SELECT COUNT(pv.id)
-                    FROM prompt_versions pv
-                    WHERE pv.prompt_id = p.id
-                    AND pv.workspace_id = p.workspace_id
-                    AND pv.version_type = 'prompt_version'
+                    SELECT COUNT(pfp.id)
+                    FROM pv_for_prompt pfp
+                    WHERE pfp.version_type = 'prompt_version'
                 ) AS version_count,
                 (
                     SELECT JSON_OBJECT(
-                        'id', pv.id,
-                        'prompt_id', pv.prompt_id,
-                        'commit', pv.commit,
-                        'version_number', pv.version_number,
-                        'template', pv.template,
-                        'metadata', pv.metadata,
-                        'change_description', pv.change_description,
-                        'type', pv.type,
-                        'version_type', pv.version_type,
-                        'environments', (SELECT JSON_ARRAYAGG(pve.environment) FROM prompt_version_envs pve WHERE pve.workspace_id = pv.workspace_id AND pve.version_id = pv.id AND pve.ended_at IS NULL),
-                        'tags', pv.tags,
-                        'created_at', pv.created_at,
-                        'created_by', pv.created_by,
-                        'last_updated_at', pv.last_updated_at,
-                        'last_updated_by', pv.last_updated_by
+                        'id', pfp.id,
+                        'prompt_id', pfp.prompt_id,
+                        'commit', pfp.commit,
+                        'version_number', pfp.version_number,
+                        'template', pfp.template,
+                        'metadata', pfp.metadata,
+                        'change_description', pfp.change_description,
+                        'type', pfp.type,
+                        'version_type', pfp.version_type,
+                        'environments', ve.environments,
+                        'tags', pfp.tags,
+                        'created_at', pfp.created_at,
+                        'created_by', pfp.created_by,
+                        'last_updated_at', pfp.last_updated_at,
+                        'last_updated_by', pfp.last_updated_by
                     )
-                    FROM prompt_versions pv
-                    WHERE pv.prompt_id = p.id
-                    AND pv.workspace_id = p.workspace_id
-                    AND pv.version_type = 'prompt_version'
-                    ORDER BY pv.id DESC
+                    FROM pv_for_prompt pfp
+                    LEFT JOIN ver_envs ve ON ve.version_id = pfp.id
+                    WHERE pfp.version_type = 'prompt_version'
+                    ORDER BY pfp.id DESC
                     LIMIT 1
                 ) AS latest_version
                 <if(mask_id || environment)>
                 ,
                 (
                     SELECT JSON_OBJECT(
-                        'id', pv.id,
-                        'prompt_id', pv.prompt_id,
-                        'commit', pv.commit,
-                        'version_number', pv.version_number,
-                        'template', pv.template,
-                        'metadata', pv.metadata,
-                        'change_description', pv.change_description,
-                        'type', pv.type,
-                        'version_type', pv.version_type,
-                        'environments', (SELECT JSON_ARRAYAGG(pve.environment) FROM prompt_version_envs pve WHERE pve.workspace_id = pv.workspace_id AND pve.version_id = pv.id AND pve.ended_at IS NULL),
-                        'tags', pv.tags,
-                        'created_at', pv.created_at,
-                        'created_by', pv.created_by,
-                        'last_updated_at', pv.last_updated_at,
-                        'last_updated_by', pv.last_updated_by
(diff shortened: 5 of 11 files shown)


apps/opik-backend/src/main/java/com/comet/opik/api/PromptVersionEnvironmentUpdate.java
```diff
@@ -4,20 +4,23 @@
...
-        The environment must already exist in the workspace registry; unknown names return 404.
+        Replace the full set of environments assigned to a prompt version.
+        The provided set becomes the new complete set of environments for this version.
+        - Non-empty set: assigns each environment to this version; if another version of the same prompt currently owns any of those environments, ownership is moved atomically.
+        - Empty set: clears all environments from this version.
+        All environments must already exist in the workspace registry; unknown names return 404.
         """)
 public record PromptVersionEnvironmentUpdate(
-        @Nullable @Pattern(regexp = Environment.NAME_PATTERN, message = Environment.NAME_PATTERN_MESSAGE) @Size(max = 150, message = "cannot exceed 150 characters") String environment) {
+        @NotNull @Size(max = 100, message = "cannot exceed 100 environments") Set<@Pattern(regexp = Environment.NAME_PATTERN, message = Environment.NAME_PATTERN_MESSAGE) @Size(max = 150, message = "cannot exceed 150 characters") String> environments) {
```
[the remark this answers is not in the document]
BorisTkachenko: We do not need to support backward compatibility here. This feature is not yet in production.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000077_add_prompt_version_envs_table.sql
```diff
@@ -0,0 +1,19 @@
...
+    created_at   TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
+    created_by   VARCHAR(255) NOT NULL DEFAULT '',
+    ended_at     TIMESTAMP(6) NULL DEFAULT NULL,
+    PRIMARY KEY (id),
+    INDEX idx_pve_env_lookup (workspace_id, prompt_id, environment, ended_at),
+    INDEX idx_pve_version (workspace_id, version_id, ended_at)
+);
+
+ALTER TABLE prompt_versions DROP INDEX idx_prompt_versions_workspace_prompt_environment;
+ALTER TABLE prompt_versions DROP COLUMN environment;
```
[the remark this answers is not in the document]
BorisTkachenko: It could be forward-only since this is not in production anyway, so we don't need to roll back.

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptService.java
```diff
@@ -332,16 +332,21 @@ public PromptVersion createPromptVersion(@NonNull CreatePromptVersion createProm
...
+                ? Set.of()
+                : createPromptVersion.version().environments().stream()
+                        .filter(StringUtils::isNotBlank)
+                        .map(String::trim)
+                        .collect(java.util.stream.Collectors.toSet());
 
-        if (environment != null) {
-            environmentService.bulkCreate(Set.of(environment), workspaceId, userName);
+        if (!environments.isEmpty()) {
+            environmentService.bulkCreate(environments, workspaceId, userName);
```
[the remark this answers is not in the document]
BorisTkachenko: No need, it will make code less readable

apps/opik-backend/src/main/java/com/comet/opik/api/PromptVersion.java
```diff
@@ -102,7 +102,7 @@ public PromptVersionType versionType() {
     }
 
     @JsonIgnore
-    @AssertTrue(message = "environment cannot be set on a mask version") public boolean isEnvironmentCompatibleWithVersionType() {
-        return environment == null || versionType() != PromptVersionType.MASK;
+    @AssertTrue(message = "environments cannot be set on a mask version") public boolean isEnvironmentCompatibleWithVersionType() {
+        return (environments == null || environments.isEmpty()) || versionType() != PromptVersionType.MASK;
```
thiagohora: NIT: Use is empty from CollectionUtils

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptDAO.java
```diff
@@ -75,7 +75,7 @@ SELECT JSON_OBJECT(
                         'change_description', pv.change_description,
                         'type', pv.type,
                         'version_type', pv.version_type,
-                        'environment', pv.environment,
+                        'environments', (SELECT JSON_ARRAYAGG(pve.environment) FROM prompt_version_envs pve WHERE pve.workspace_id = pv.workspace_id AND pve.version_id = pv.id AND pve.ended_at IS NULL),
```
thiagohora: Can you check the query plan? I think this will run once per row

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptService.java
```diff
@@ -332,16 +332,21 @@ public PromptVersion createPromptVersion(@NonNull CreatePromptVersion createProm
...
+        Set<String> environments = createPromptVersion.version().environments() == null
+                ? Set.of()
+                : createPromptVersion.version().environments().stream()
+                        .filter(StringUtils::isNotBlank)
+                        .map(String::trim)
+                        .collect(java.util.stream.Collectors.toSet());
 
-        if (environment != null) {
-            environmentService.bulkCreate(Set.of(environment), workspaceId, userName);
+        if (!environments.isEmpty()) {
```
thiagohora: NIT: Use CollectionUtils

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptService.java
```diff
@@ -756,36 +763,46 @@ public int updateVersions(@NonNull PromptVersionBatchUpdate update) {
...
 
     @Override
-    public void setVersionEnvironment(@NonNull UUID versionId, String environment) {
+    public void setVersionEnvironment(@NonNull UUID versionId, Set<String> environments) {
         String workspaceId = requestContext.get().getWorkspaceId();
-        String env = StringUtils.trimToNull(environment);
+        String userName = requestContext.get().getUserName();
+
+        Set<String> envs = environments == null
+                ? Set.of()
```
thiagohora, line 777: Same

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptService.java
```diff
@@ -756,36 +763,46 @@ public int updateVersions(@NonNull PromptVersionBatchUpdate update) {
...
-
-        if (env != null && !environmentService.existsByName(env, workspaceId)) {
-            throw new NotFoundException(
-                    "Environment '%s' does not exist in the workspace".formatted(env));
+        if (!envs.isEmpty()) {
+            Set<String> existing = environmentService.findExistingNames(envs, workspaceId);
+            if (existing.size() != envs.size()) {
+                Set<String> missing = new java.util.HashSet<>(envs);
+                missing.removeAll(existing);
+                throw new NotFoundException(
```
thiagohora: I think here, this should be a 409

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptVersionDAO.java
```diff
@@ -56,16 +56,58 @@ INSERT INTO prompt_versions (
...
                 :bean.tags,
                 :bean.createdBy,
                 :workspace_id
             )
             """)
     void save(@Bind("workspace_id") String workspaceId, @BindMethods("bean") PromptVersion prompt);
 
+    @SqlBatch("""
+            INSERT INTO prompt_version_envs (id, workspace_id, prompt_id, version_id, environment, created_by)
+            VALUES (UUID(), :workspace_id, :prompt_id, :version_id, :environment, :created_by)
```
thiagohora: This is a UUIDv1, we use our v7 ids

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptVersionDAO.java
```diff
@@ -56,16 +56,58 @@ INSERT INTO prompt_versions (
...
+            SELECT environment FROM prompt_version_envs
+            WHERE workspace_id = :workspace_id AND prompt_id = :prompt_id AND environment IN (<environments>) AND ended_at IS NULL
+            """)
+    Set<String> findTakenEnvironments(@Bind("prompt_id") UUID promptId, @Bind("workspace_id") String workspaceId,
+            @BindList("environments") Set<String> environments);
+
+    @SqlQuery("""
+            SELECT pv.*,
+                p.template_structure,
+                (SELECT JSON_ARRAYAGG(pve.environment) FROM prompt_version_envs pve WHERE pve.workspace_id = pv.workspace_id AND pve.version_id = pv.id AND pve.ended_at IS NULL) AS environments
```
thiagohora: Again, I would check the cost of those queries. Just to make sure we don't stress MySQL

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptService.java
```diff
@@ -756,36 +763,46 @@ public int updateVersions(@NonNull PromptVersionBatchUpdate update) {
...
 
-        UUID promptId = version.promptId();
-
-        if (env != null && !environmentService.existsByName(env, workspaceId)) {
-            throw new NotFoundException(
-                    "Environment '%s' does not exist in the workspace".formatted(env));
+        if (!envs.isEmpty()) {
+            Set<String> existing = environmentService.findExistingNames(envs, workspaceId);
+            if (existing.size() != envs.size()) {
+                Set<String> missing = new java.util.HashSet<>(envs);
```
thiagohora: Use proper import format

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/PromptResourceTest.java
```diff
@@ -5538,7 +5681,7 @@ private PromptVersion createRegularVersion(Prompt prompt) {
             var version = factory.manufacturePojo(PromptVersion.class).toBuilder()
                     .createdBy(USER)
                     .versionType(PromptVersionType.PROMPT_VERSION)
-                    .environment(null)
+                    .environments(null)
                     .build();
```
thiagohora, line 5616: Centralize this since it repeats many times

thiagohora (commented):
A few follow-ups before moving forward

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/PromptResourceTest.java
```diff
@@ -5165,14 +5163,135 @@ void getPromptByIdWithBothEnvAndMaskReturnsBadRequest() {
...
+            String env2 = uniqueEnvName("create-b");
+            var version = factory.manufacturePojo(PromptVersion.class).toBuilder()
+                    .createdBy(USER).versionType(PromptVersionType.PROMPT_VERSION)
+                    .environments(Set.of(env1, env2)).build();
+
+            var saved = createPromptVersion(
+                    createPromptVersionRequest(prompt.name(), version, prompt.templateStructure()),
+                    API_KEY, TEST_WORKSPACE);
+
+            assertThat(saved.environments()).containsExactlyInAnyOrder(env1, env2);
```
thiagohora: These new environment tests assert only the `environments` property (`assertThat(saved.environments())...`), reshaping the returned `PromptVersion` down to a single `Set` for the comparison. This file already has whole-object helpers — `getPromptVersionAndAssert(...)` and `retrievePromptVersionAndAssert(...)` (around L3100–L3140) — that compare the full `PromptVersion` via `usingRecursiveComparison` with the `createdAt` Instant comparator.

Prefer asserting the whole returned/fetched object so regressions in any other field on the create/patch/retrieve paths are also caught, e.g.:

```java
getPromptVersionAndAssert(
        saved.id(),
        saved.toBuilder().environments(Set.of(env1, env2)).build(),
        API_KEY, TEST_WORKSPACE);
```

For the clear/move cases the expected `environments` is `null` (the column mapper maps an empty `JSON_ARRAYAGG` result to `null`, not an empty set), so build the expected with `.environments(null)`. This applies to the create, multi-env, patch, clear, move and retrieve tests in this block.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000077_add_prompt_version_envs_table.sql
```diff
@@ -0,0 +1,19 @@
...
+    environment  VARCHAR(150) NOT NULL,
+    created_at   TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
+    created_by   VARCHAR(255) NOT NULL DEFAULT '',
+    ended_at     TIMESTAMP(6) NULL DEFAULT NULL,
+    PRIMARY KEY (id),
+    INDEX idx_pve_env_lookup (workspace_id, prompt_id, environment, ended_at),
+    INDEX idx_pve_version (workspace_id, version_id, ended_at)
+);
+
+ALTER TABLE prompt_versions DROP INDEX idx_prompt_versions_workspace_prompt_environment;
```
thiagohora, line 20: Dropping `idx_prompt_versions_workspace_prompt_environment` removes the **DB-level** single-owner-per-environment guarantee. After this change the invariant ("at most one active version owns a given environment per prompt") is enforced only by the application `withPromptVersionLock`. That's a real loss of defense-in-depth, and it can — and I'd argue should — still be enforced in the schema.

MySQL has no partial indexes, but a generated stored column gives the same effect with soft-delete:

```sql
ALTER TABLE prompt_version_envs
  ADD COLUMN active_env VARCHAR(150)
    GENERATED ALWAYS AS (IF(ended_at IS NULL, environment, NULL)) STORED,
  ADD UNIQUE INDEX uq_pve_active_env (workspace_id, prompt_id, active_env);
```

Closed rows get `active_env = NULL`, and MySQL allows multiple NULLs in a unique index, so only **active** rows are constrained — one per `(workspace_id, prompt_id, environment)`.

I checked the write paths and they're already compatible with this constraint (it would never reject a legitimate operation):
- `setVersionEnvironment` runs `closeVersionEnvironments` → `closeEnvOwnershipsForPrompt(envs)` → `saveEnvironments(...)` in a single WRITE transaction. The closes set `ended_at` first (recomputing `active_env` to NULL for the prior owners), so the INSERT of the new active rows can't collide.
- `createVersionWithEnvironment` rejects with 409 via `findTakenEnvironments` before inserting, so it only inserts when the envs are free.

So at commit time there is never more than one active row per `(ws, prompt, env)`, and the unique index becomes a safety net that also closes the lock's cross-replica race window (the existence check and insert are in separate transactions). Worth adding to this migration.

thiagohora (commented):
Two review notes (non-blocking):

1. **Tests** — the new environment tests assert only `.environments()` rather than the whole `PromptVersion`. Reuse the existing `getPromptVersionAndAssert` / `usingRecursiveComparison` helpers so other fields are covered too. (inline)

2. **Single-owner invariant** — dropping the unique index moves this guarantee from the DB to the app lock. I verified the write paths and a DB-level unique constraint via a generated `active_env` column is fully compatible and would restore defense-in-depth. (inline)

andrescrz (changes requested):
There's a change in the contract for `environments`. I agree that it's for good reasons and a good one on the long term. But it breaks backwards compatibility. Can we do that? How do you ensure that we don't break existing clients? Are our clients only internal SDKs or FE? Or do we potentially have external clients?

BorisTkachenko:
> There's a change in the contract for `environments`. I agree that it's for good reasons and a good one on the long term. But it breaks backwards compatibility. Can we do that? How do you ensure that we don't break existing clients? Are our clients only internal SDKs or FE? Or do we potentially have external clients?

@andrescrz This feature is not yet in production, so the current environment is not being used by Frontend or SDK. Breaking contract is safe here.

apps/opik-backend/src/main/java/com/comet/opik/domain/EnvironmentDAO.java
```diff
@@ -61,6 +61,10 @@ void update(@Bind("id") UUID id,
     @SqlQuery("SELECT COUNT(*) FROM environments WHERE workspace_id = :workspaceId AND name = :name")
     long countByName(@Bind("workspaceId") String workspaceId, @Bind("name") String name);
 
+    @SqlQuery("SELECT name FROM environments WHERE workspace_id = :workspaceId AND name IN (<names>)")
+    Set<String> findExistingNames(@Bind("workspaceId") String workspaceId,
+            @BindList("names") Set<String> names);
```
andrescrz: Minor: existing method already without the names filtering. I'd centralise the query there making that optional with ST, and then maybe ok to have an overloaded method without names param.

apps/opik-backend/src/main/java/com/comet/opik/domain/EnvironmentService.java
```diff
@@ -199,6 +201,15 @@ public boolean existsByName(@NonNull String name, @NonNull String workspaceId) {
         });
     }
 
+    @Override
+    public Set<String> findExistingNames(@NonNull Set<String> names, @NonNull String workspaceId) {
+        if (names.isEmpty()) {
```
andrescrz: Minor: if you handle empty, you should handle value instead of validating and throwing a NPE. I recommend using a null-safe check with CollectionUtils.

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptService.java
```diff
@@ -332,16 +332,21 @@ public PromptVersion createPromptVersion(@NonNull CreatePromptVersion createProm
         Prompt prompt = getOrCreatePrompt(workspaceId, createPromptVersion.name(), userName, templateStructure,
                 projectId);
 
-        String environment = StringUtils.trimToNull(createPromptVersion.version().environment());
+        Set<String> environments = createPromptVersion.version().environments() == null
+                ? Set.of()
+                : createPromptVersion.version().environments().stream()
+                        .filter(StringUtils::isNotBlank)
+                        .map(String::trim)
```
andrescrz: Nit: better strip.

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptService.java
```diff
@@ -332,16 +332,21 @@ public PromptVersion createPromptVersion(@NonNull CreatePromptVersion createProm
...
+                        .collect(java.util.stream.Collectors.toSet());
```
andrescrz: Nit: imports

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptService.java
```diff
@@ -520,6 +525,8 @@ private PromptVersion savePromptVersion(String workspaceId, PromptVersion prompt
             }
 
             promptVersionDAO.save(workspaceId, toSave);
+            promptVersionDAO.saveEnvironments(workspaceId, toSave.promptId(), toSave.id(),
```
andrescrz: Let's double check if correct to have a separated `saveEnvironments` call from the `save` here.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000077_add_prompt_version_envs_table.sql
```diff
@@ -0,0 +1,19 @@
+--liquibase formatted sql
+--changeset boryst:000077_add_prompt_version_envs_table
+
+CREATE TABLE prompt_version_envs (
```
andrescrz, line 4: Given the low number of envs (100). I'm not sure if a normalised schema here is beneficial over simply storing an array of values on the version. Likely, the number isn't going to increase.

I believe this needs more investigation for a proper decision.

andrescrz (commented):
No concern about the backwards compatibility after your explanation.

I gave a more specific review.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000077_add_prompt_version_envs_table.sql
BorisTkachenko: The reason for a separate table is as following:
- Ownership tracking across versions. The ended_at column is what makes the "move environment" operation possible without losing history, was discussed with product team that history might be needed later. With an array on the version row, moving env=production from v2 to v3 means patching two rows (v2.environments -= production, v3.environments += production). With the current schema you soft-delete the old ownership row and insert a new one atomically — the history is preserved.

- DB-level uniqueness. The generated column + unique index enforces "only one active version owns this environment per prompt" at the database level. With a JSON array you can't do that — you'd be back to app-level enforcement only.

- Indexed ownership lookups. "Which version currently owns environment X for prompt Y?" is a single indexed point lookup against prompt_version_envs WHERE environment = :env AND ended_at IS NULL. With a JSON array on the version you'd need JSON_CONTAINS which is not indexable in MySQL.

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptService.java
BorisTkachenko: We are writing to two different tables, and this is transactional, so two calls are correct here.

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptDAO.java
BorisTkachenko: Updated the query

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptDAO.java
```diff
@@ -55,70 +55,81 @@ boolean hasVersion1Prompts(
...
+            WITH pv_for_prompt AS (
+                SELECT pv.*
+                FROM prompt_versions pv
+                WHERE pv.prompt_id = :id AND pv.workspace_id = :workspace_id
+            ), active_envs AS (
+                SELECT pve.version_id, pve.environment
+                FROM prompt_version_envs pve
+                INNER JOIN pv_for_prompt pfp ON pfp.id = pve.version_id
+                WHERE pve.workspace_id = :workspace_id AND pve.ended_at IS NULL
+            ), ver_envs AS (
```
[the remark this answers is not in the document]
BorisTkachenko: No need

apps/opik-backend/src/main/java/com/comet/opik/domain/EnvironmentDAO.java
```diff
@@ -55,8 +58,19 @@ void update(@Bind("id") UUID id,
...
+    Set<String> findNames(@Bind("workspaceId") String workspaceId,
+            @Define("names") @BindList(onEmpty = BindList.EmptyHandling.NULL_VALUE, value = "names") Set<String> names);
+
+    default Set<String> findAllNames(String workspaceId) {
+        return findNames(workspaceId, null);
+    }
+
+    default Set<String> findExistingNames(String workspaceId, Set<String> names) {
+        return findNames(workspaceId, names);
+    }
```
andrescrz: Very nit: this method doesn't do anything except the rename. You could just call `findNames`, the filter is obvious from the `names` params.

apps/opik-backend/src/main/java/com/comet/opik/domain/EnvironmentService.java
```diff
@@ -199,6 +202,15 @@ public boolean existsByName(@NonNull String name, @NonNull String workspaceId) {
         });
     }
 
+    @Override
+    public Set<String> findExistingNames(@NonNull Set<String> names, @NonNull String workspaceId) {
```
andrescrz: Very nit: still the redundant `NonNull` validation for `names`. Cases NPE which can be avoided by returning the empty set in the null-safe isEmpty below.

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptVersionDAO.java
```diff
@@ -56,25 +56,93 @@ INSERT INTO prompt_versions (
...
+            @Bind("id") List<UUID> ids,
+            @Bind("workspace_id") String workspaceId,
+            @Bind("prompt_id") UUID promptId,
+            @Bind("version_id") UUID versionId,
+            @Bind("environment") List<String> environments,
+            @Bind("created_by") String createdBy);
+
+    default void saveEnvironments(List<UUID> ids, String workspaceId, UUID promptId, UUID versionId,
+            Set<String> environments, String createdBy) {
+        if (environments == null || environments.isEmpty()) {
```
andrescrz: Very nit: as usual, better use collection utils for a null-safe is empty check.

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptVersionDAO.java
```diff
@@ -56,25 +56,93 @@ INSERT INTO prompt_versions (
...
-            <if(search)> AND (pv.template LIKE CONCAT('%', :search, '%') OR pv.change_description LIKE CONCAT('%', :search, '%')) <endif>
-            <if(filters)> AND <filters> <endif>
+            SELECT environment FROM prompt_version_envs
+            WHERE workspace_id = :workspace_id AND version_id = :version_id AND ended_at IS NULL
+            """)
+    Set<String> findVersionEnvironments(@Bind("version_id") UUID versionId, @Bind("workspace_id") String workspaceId);
+
+    @SqlUpdate("""
+            UPDATE prompt_version_envs
+            SET ended_at = CURRENT_TIMESTAMP(6)
```
andrescrz: I believe you should track the typical update metadata: last_updated_at and last_updated_by. Any reason for not doing this in the new `prompt_version_envs` table?

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptVersionDAO.java
```diff
@@ -56,25 +56,93 @@ INSERT INTO prompt_versions (
...
+            """)
+    void closeVersionEnvironments(@Bind("version_id") UUID versionId, @Bind("workspace_id") String workspaceId);
+
+    @SqlUpdate("""
+            UPDATE prompt_version_envs
+            SET ended_at = CURRENT_TIMESTAMP(6)
+            WHERE workspace_id = :workspace_id AND version_id = :version_id AND environment IN (<environments>) AND ended_at IS NULL
+            """)
+    void closeVersionEnvironmentsForNames(@Bind("version_id") UUID versionId, @Bind("workspace_id") String workspaceId,
+            @BindList("environments") Set<String> environments);
```
andrescrz: Same here about last_updated_at and last_updated_by. In addition, the only difference is the filter on `environment`. You should centralise that in a single method and use ST templating as usual.

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptVersionDAO.java
```diff
@@ -56,25 +56,93 @@ INSERT INTO prompt_versions (
...
+
+    @SqlUpdate("""
+            UPDATE prompt_version_envs
+            SET ended_at = CURRENT_TIMESTAMP(6)
+            WHERE workspace_id = :workspace_id AND prompt_id = :prompt_id AND environment IN (<environments>) AND ended_at IS NULL
+            """)
+    void closeEnvOwnershipsForPrompt(@Bind("prompt_id") UUID promptId, @Bind("workspace_id") String workspaceId,
+            @BindList("environments") Set<String> environments);
```
andrescrz: Again, one more duplication.

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptVersionDAO.java
```diff
@@ -56,25 +56,93 @@ INSERT INTO prompt_versions (
...
+
+    @SqlQuery("""
+            SELECT environment FROM prompt_version_envs
+            WHERE workspace_id = :workspace_id AND prompt_id = :prompt_id AND environment IN (<environments>) AND ended_at IS NULL
+            """)
+    Set<String> findTakenEnvironments(@Bind("prompt_id") UUID promptId, @Bind("workspace_id") String workspaceId,
+            @BindList("environments") Set<String> environments);
```
andrescrz: Another duplication.

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptVersionDAO.java
```diff
@@ -152,7 +222,9 @@ PromptVersion findByCommit(@Bind("prompt_id") UUID promptId, @Bind("commit") Str
             @Bind("workspace_id") String workspaceId);
 
     @SqlQuery("""
-            SELECT pv.*, p.template_structure
+            SELECT pv.*,
+                p.template_structure,
+                (SELECT JSON_ARRAYAGG(pve.environment) FROM prompt_version_envs pve WHERE pve.workspace_id = pv.workspace_id AND pve.version_id = pv.id AND pve.ended_at IS NULL) AS environments
```
andrescrz: Minor: seems duplicated as well.

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptVersionDAO.java
```diff
@@ -223,28 +295,15 @@ int update(@Bind("workspace_id") String workspaceId,
...
+            SELECT pv.*,
+                p.template_structure,
+                (SELECT JSON_ARRAYAGG(pve2.environment) FROM prompt_version_envs pve2 WHERE pve2.workspace_id = pv.workspace_id AND pve2.version_id = pv.id AND pve2.ended_at IS NULL) AS environments
             FROM prompt_versions pv
             INNER JOIN prompts p ON pv.prompt_id = p.id
-            WHERE pv.prompt_id = :prompt_id AND pv.environment = :environment AND pv.workspace_id = :workspace_id
-            AND pv.version_type = 'prompt_version'
+            INNER JOIN prompt_version_envs pve ON pve.workspace_id = pv.workspace_id AND pve.version_id = pv.id
+            WHERE pv.prompt_id = :prompt_id AND pve.environment = :environment AND pv.workspace_id = :workspace_id
+            AND pv.version_type = 'prompt_version' AND pve.ended_at IS NULL
```
andrescrz: Minor as probably pre-existing: seems like duplication.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000077_add_prompt_version_envs_table.sql
```diff
@@ -0,0 +1,21 @@
...
+    id           CHAR(36)     NOT NULL,
+    workspace_id VARCHAR(255) NOT NULL,
+    prompt_id    CHAR(36)     NOT NULL,
+    version_id   CHAR(36)     NOT NULL,
+    environment  VARCHAR(150) NOT NULL,
+    created_at   TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
+    created_by   VARCHAR(255) NOT NULL DEFAULT '',
```
andrescrz, line 11: Mentioned in another comment. There are update semantics. We likely require the typical metadata for that: last_updated at and by.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000077_add_prompt_version_envs_table.sql
andrescrz, line 1: This is missing the rollback.

andrescrz (commented):
The PR still has some blocking gaps. Please take a look at the specific comments. I leave the other reviewer @thiagohora to double check the queries as result of the plans review on `PromptDAO`.

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptVersionDAO.java
BorisTkachenko: No this is a desired behavior. With our typical update metadata, we will not be able to track the history. It will be just updated in place.

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptVersionDAO.java
BorisTkachenko: Same as above. We don't want to introduce here last_updated_*, as in this case, history cannot be tracked for a specific environment.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000077_add_prompt_version_envs_table.sql
BorisTkachenko: We never update an environment in a regular way

andrescrz: Not a blocker on my end, but you won't capture who closed the environment.

andrescrz (approved):
On my end PR looks good and can go through. I recommend @thiagohora double checks the latest `PromptDAO` queries, as he requested the execution plan.

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptDAO.java
BorisTkachenko: After updating the query, I checked the query plan. All CTEs are run only once, and everything seems okay. 
```
| -> Table scan on p  (cost=46 rows=435)
-> Select #2 (subquery in projection; run only once)
    -> Aggregate: count(pv.id)  (cost=116 rows=1)
        -> Filter: (pv.version_type = 'prompt_version')  (cost=79.9 rows=366)
            -> Table scan on pv  (cost=79.9 rows=731)
-> Select #4 (subquery in projection; run only once)
    -> Limit: 1 row(s)
        -> Sort: id DESC, limit input to 1 row(s) per chunk
            -> Stream results  (cost=46.5 rows=366)
                -> Left hash join (<hash>(ve.version_id)=<hash>(pv.id)), extra conditions: (ve.version_id = pv.id)  (cost=46.5 rows=366)
                    -> Filter: (pv.version_type = 'prompt_version')  (cost=79.9 rows=366)
                        -> Table scan on pv  (cost=79.9 rows=731)
                    -> Hash
                        -> Table scan on ve  (cost=3.1..3.1 rows=1)
                            -> Materialize CTE ver_envs if needed  (cost=0.585..0.585 rows=1)
                                -> Group aggregate: json_arrayagg(pve.environment)  (cost=0.485 rows=1)
                                    -> Nested loop inner join  (cost=0.385 rows=1)
                                        -> Sort: version_id  (cost=0.26 rows=1)
                                            -> Filter: (pve.ended_at is null)  (cost=0.26 rows=1)
                                                -> Index lookup on pve using uq_pve_active_env (workspace_id='default')  (cost=0.26 rows=1)
                                        -> Single-row covering index lookup on pv using PRIMARY (id=pve.version_id)  (cost=1.25 rows=1)
-> Select #9 (subquery in projection; run only once)
    -> Left hash join (<hash>(ve.version_id)=<hash>(pv.id)), extra conditions: (ve.version_id = pv.id)  (cost=46.5 rows=366)
        -> Filter: (pv.version_type = 'prompt_version')  (cost=79.9 rows=366)
            -> Table scan on pv  (cost=79.9 rows=731)
        -> Hash
            -> Table scan on ve  (cost=3.1..3.1 rows=1)
                -> Materialize CTE ver_envs if needed (query plan printed elsewhere)  (cost=0.585..0.585 rows=1)
 |
```

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000077_add_prompt_version_envs_table.sql
BorisTkachenko: Actually, we will be able to capture it. Whoever created the next environment is the one who closed this. So we will be able to retrieve this info.

thiagohora (approved):

PR #6884, opened 2026-05-27
[OPIK-6701] [BE] feat: assign ambiguous datasets to their dominant project

andrescrz:
## Details

The dataset project migration job previously skipped datasets whose referencing experiments spanned multiple projects ("ambiguous") and trapped the workspace on V1. Under forced V2 that left those datasets invisible in project-scoped surfaces (perceived data loss). This replaces skip-and-trap with a deterministic **dominant-project assignment**: each such dataset is assigned to the project with the most referencing experiments — ties broken by most recent experiment activity, then a stable `project_id` ordering — so every non-demo orphan dataset is migrated and affected workspaces can promote to V2. (~5% of datasets landed in the ambiguous bucket in production.)

- ClickHouse `COMPUTE_DATASET_PROJECT_MAPPING` now returns the dominant project per dataset plus a sorted per-project breakdown, in a single per-workspace query (no fan-out, comparable cost to before).
- The service folds formerly-ambiguous datasets into the certain bucket; certain-deleted and no-inference datasets still route to the Default Project (auto-created if missing). A deleted dominant project falls back to the Default Project.
- The workspace-trap mechanism for this job is removed end-to-end: a Liquibase migration drops the trap columns and index, and the DAO/service methods, record fields, and the `cycle.trapped_workspaces` gauge are deleted. Previously-trapped workspaces re-enter the queue automatically once the columns are dropped.
- New `datasets.assigned_to_dominant_project` counter (tagged by `distinct_project_count`, capped to bound label cardinality) plus a structured per-assignment log naming the chosen project and the per-project experiment counts.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- OPIK-6701

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.7
- Scope: implementation and test authoring, with human direction and manual refinements
- Human verification: code review + integration tests

## Testing

Extended `DatasetProjectMigrationServiceTest` (integration-level, real ClickHouse/MySQL/Redis via testcontainers) and kept the `DatasetProjectMigrationJobTest` smoke test unchanged.

- Command: `mvn test -Dtest='DatasetProjectMigrationServiceTest,DatasetProjectMigrationJobTest'` — all green (15 tests).
- Scenarios validated: dominant-by-count, recency tiebreaker, exclusion of empty-`project_id` (unmigrated) experiments, deleted-dominant fallback to Default Project, certain / certain-deleted / no-inference buckets, Default Project auto-create, env-excluded workspaces, re-run idempotency, and a mixed-bucket workspace promoting to V2.
- Also ran `WorkspaceVersionResourceTest` to confirm the `workspaces` column drop does not break the version read path.
- Not run locally: full backend suite and quality checks — deferred to CI.

## Documentation

- The operational runbook for this migration job was updated to describe the post-change behavior (dominant-project assignment, no trap, CLI as an opt-in copy tool).
- The monitoring dashboard update is tracked separately (it lives in the monitoring repository).
- No in-repo documentation changes.


apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetProjectMigrationService.java
```diff
@@ -302,61 +271,84 @@ private Mono<Boolean> classifyAndMigrate(
...
                     var allMappings = new ArrayList<>(validatedCertain);
-                    // distinctProjectCount is a query-side field; for the synthesized
-                    // mappings it's unused downstream, so 1 is a safe filler.
+                    // distinctProjectCount and projectBreakdown are query-side fields; for the
+                    // synthesized Default Project mappings they're unused downstream, so 1 and an
+                    // empty breakdown are safe fillers.
                     for (var id : certainDeletedIds) {
                         allMappings.add(DatasetProjectMapping.builder()
-                                .datasetId(id).projectId(defaultProjectId).distinctProjectCount(1L).build());
+                                .datasetId(id).projectId(defaultProjectId).distinctProjectCount(1L)
```
ldaugusto, line 332: For the noinference the best is to move to Default Project, but for when the dominant project was deleted, couldn't we try the 2nd in the rank? Or was it decided it's not worthy the effort?

ldaugusto (commented):
I have one question to make sure, but it looks good to go

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetProjectMigrationService.java
```diff
@@ -302,61 +271,84 @@ private Mono<Boolean> classifyAndMigrate(
...
+                                .projectBreakdown("").build());
```
thiagohora, line 333: **Nit (non-blocking):** the `projectBreakdown("")` fillers here (and on the no-inference branch below) are dead values — the write path (`DatasetDAO.batchSetProjectId`) binds only `:projectId`/`:datasetId`, and `recordDominantAssignments` reads `projectBreakdown` only when `distinctProjectCount > 1` (these are `1L`). They exist purely to satisfy the `@NonNull` on the record component. Dropping `@NonNull` on `DatasetProjectMapping.projectBreakdown` would let these be left unset. Harmless as-is; mentioning for a possible follow-up.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetProjectMapping.java
```diff
@@ -5,8 +5,21 @@
...
+ * example {@code "p1=5,p2=3,p3=1"}) and is included in the log entry written for each dominant
+ * assignment. It is empty for the rows the service builds for Default Project assignments.
+ */
 @Builder(toBuilder = true)
-public record DatasetProjectMapping(@NonNull UUID datasetId, @NonNull UUID projectId, long distinctProjectCount) {
+public record DatasetProjectMapping(
+        @NonNull UUID datasetId,
+        @NonNull UUID projectId,
+        long distinctProjectCount,
+        @NonNull String projectBreakdown) {
```
thiagohora, line 24: **Nit (non-blocking):** `@NonNull` here is what forces the `projectBreakdown("")` fillers in `DatasetProjectMigrationService` for the synthesized Default-Project mappings. Since the dominant-assignment log already guards on `distinctProjectCount > 1`, this could be made nullable to drop those fillers. Not blocking.

thiagohora (approved):
Correct and well-tested. Only a few small points, but no blockers

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetProjectMigrationService.java
andrescrz: Good question — it was a deliberate scope decision, not an oversight. The ClickHouse query returns only the single dominant project per dataset (`ranked[1]`), so the service never sees the runner-ups; a deleted dominant therefore flows through the existing certain-deleted path → Default Project, matching the ticket's certain-deleted policy. A 2nd-rank fallback is feasible (we already compute the full `ranked` array) but would mean returning the ranked list and validating it in rank order — more moving parts for a case that's rare (the specifically-chosen dominant deleted while a runner-up survives). Happy to add it as a follow-up if we see it happen in practice.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetProjectMapping.java
andrescrz: Intentional — we chose to keep `projectBreakdown` `@NonNull` to favour non-nullable fields, since the query path always yields a non-null breakdown. The empty-string fillers on the synthesized Default-Project mappings are the accepted cost (we also looked at `@Builder.Default`, but it isn't usable on a record component). Noting as a possible follow-up if we'd rather trade the `@NonNull` guarantee to drop the fillers.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetProjectMigrationService.java
andrescrz: Intentional, paired with the `@NonNull` on `DatasetProjectMapping.projectBreakdown` — kept non-nullable on purpose, with these empty-string fillers as the accepted cost on the synthesized Default-Project mappings. Left as a possible follow-up rather than changed here.

PR #6893, opened 2026-05-28
[OPIK-4909] [BE] fix: use pre-computed gen_ai.usage.cost from OTel

### LINKED ISSUE #5620: [Bug] LiteLLM OTel: gen_ai.usage.cost (pre-computed cost) is silently dropped and ignored
## Summary

LiteLLM computes the exact cost of each request using its own pricing engine and sends it as a float OTel attribute (`gen_ai.usage.cost`). Opik silently drops this value and re-derives cost from its own bundled pricing table instead. This is why other observability platforms (PostHog, Langfuse) show the correct LiteLLM cost while Opik does not — they consume `gen_ai.usage.cost` directly.

**Originating report:** https://cometml.slack.com/archives/C02HPRHR2F9/p1772947523388679

## Root Cause

`gen_ai.usage.*` attributes route to the USAGE outcome in `GenAIMappingRules.java`. The usage map is typed `Map<String, Integer>`, so `extractUsageField()` only handles integer and string OTel values:

```java
// OpenTelemetryMappingUtils.java:88-101
public static void extractUsageField(Map<String, Integer> usage, ..., AnyValue value) {
    if (value.hasIntValue()) {
        // handle integer tokens
    } else if (value.hasStringValue()) {
        // try Integer.parseInt, then JSON object parsing
    }
    // DOUBLE_VALUE (float) falls through — silently ignored
}
```

`gen_ai.usage.cost = 0.001234` is a `DOUBLE_VALUE` in OTel proto. It is never read.

Even if LiteLLM sends it as a string `"0.001234"`, `Integer.parseInt("0.001234")` throws `NumberFormatException`, and the JSON path returns `cost = 0` (float truncated to int via `.intValue()`). Either way, the value is lost.

Opik then calls `CostService.calculateCost()` independently, which may produce a different (and incorrect) result.

## Suggested Fix

Detect `gen_ai.usage.cost` (and potentially `llm.usage.cost`) before the integer-only usage path and set it directly as the span's `totalEstimatedCost`, bypassing Opik's recalculation:

1. Add a new `Outcome.COST` in `OpenTelemetryMappingRule` (or handle it as a special case in `enrichSpanWithAttributes()`).
2. Map `gen_ai.usage.cost` → `Outcome.COST` in `GenAIMappingRules`.
3. In `OpenTelemetryMapper.enrichSpanWithAttributes()`, read the float value and call `spanBuilder.totalEstimatedCost(BigDecimal.valueOf(value.getDoubleValue()))` with `totalEstimatedCostVersion = ""` so it is treated as a manually-set (authoritative) cost and not overwritten by Opik's calculator.

This matches the behaviour of Langfuse and PostHog and would make costs consistent across all observability platforms for LiteLLM users.

## Relevant Files

- `apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java` — `extractUsageField()` (line 88)
- `apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/otel/GenAIMappingRules.java` — `gen_ai.usage.*` rule (line 49)
- `apps/opik-backend/src/main/java/com/comet/opik/domain/OpenTelemetryMapper.java` — `enrichSpanWithAttributes()` (line 85)
- `apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingRule.java`

YarivHashaiComet:
## Details

opik-4909-cost-flow

Pre-computed `gen_ai.usage.cost` from LiteLLM (and similar OTel integrations) was silently dropped because it matched the broad `gen_ai.usage.` prefix rule and got routed to the integer-only usage map; Opik then recalculated cost from its own price table, diverging from what LiteLLM/Langfuse/PostHog report.

- Add new `Outcome.COST` and register an exact-match rule `gen_ai.usage.cost → COST` **before** the `gen_ai.usage.` prefix rule (`findRule()` returns the first match, so ordering is critical).
- `extractCost(AnyValue)` reads `DOUBLE_VALUE` / `INT_VALUE` / parseable `STRING_VALUE` into a `BigDecimal` and sets `spanBuilder.totalEstimatedCost(...)`; `SpanDAO` then stores the value as authoritative (`total_estimated_cost_version = ""`) and skips `calculateCost`.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #5620
- OPIK-4909

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Scope: implementation of the 4 backend edits, 2 unit tests, manual E2E verification (local + prod), and this PR description.
- Human verification: PR author reviewed the diff locally, ran the test suite, and performed manual end-to-end validation against both a local backend (with fix) and production (without fix) confirming the divergence.

## Testing

How you tested:

- `mvn test -Dtest='OpenTelemetryMapperTest,OpenTelemetryMappingUtilsTest'` (run from `apps/opik-backend`) — 51 + 40 tests, 0 failures, incl. 2 new `gen_ai.usage.cost` tests.
- `mvn spotless:apply` — clean, no formatting violations.
- Manual E2E **with fix** (local backend started via `./scripts/dev-runner.sh --restart`): POST an OTel span with `gen_ai.usage.cost = 0.424242`, `gen_ai.request.model = gpt-4o`, 100 prompt + 50 completion tokens to `http://localhost:8080/v1/private/otel/v1/traces`. Stored span: `total_estimated_cost = 0.424242`, `total_estimated_cost_version = ""` (authoritative); usage map = `{prompt_tokens: 100, completion_tokens: 50, total_tokens: 150}` (no `cost` key — confirms it did not leak into the integer usage map).
- Manual E2E **without fix** (production, against author's personal workspace): identical payload → stored `total_estimated_cost = 0.000750`, exactly Opik's bundled gpt-4o recalculation (`100 × $2.50/1M + 50 × $10/1M = $0.000750`). This confirms the bug today and that the fix changes the behavior.
- Scenarios validated: cost as `DOUBLE_VALUE`, cost as numeric `STRING_VALUE`, cost does not pollute the usage map, sibling token attributes preserved, `total_tokens` auto-computation still works.
- Environment: local process mode (backend Java + frontend Vite) via `dev-runner.sh`, plus a cloud comparison against production over HTTPS.
- Not run: the full containerized integration test suite for the backend (Testcontainers heavyweight; this change is purely additive in the attribute mapper and is fully covered by the unit tests above).

Video evidence: N/A (backend-only data-mapping fix; values verified by API readback shown above).

## Documentation

No user-facing documentation changes required. The OTel integration page already lists `gen_ai.usage.cost` as a supported attribute; this PR makes that attribute respected rather than dropped, with no API surface change.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java
@@ -112,11 +112,20 @@ public class OpenTelemetryMappingUtils {
      */
     public static Optional<BigDecimal> extractCost(@NonNull AnyValue value) {
         return switch (value.getValueCase()) {
-            case DOUBLE_VALUE -> Optional.of(BigDecimal.valueOf(value.getDoubleValue()));
+            case DOUBLE_VALUE -> {
+                // BigDecimal.valueOf throws NumberFormatException for NaN / Infinity, which would
+                // abort span enrichment for the whole batch. Treat non-finite values as unsupported.
+                double d = value.getDoubleValue();
+                if (!Double.isFinite(d)) {
+                    log.warn("Non-finite cost value, skipping: '{}'", d);
+                    yield Optional.empty();
+                }
+                yield Optional.of(BigDecimal.valueOf(d));
+            }
             case INT_VALUE -> Optional.of(BigDecimal.valueOf(value.getIntValue()));
             case STRING_VALUE -> parseCostString(value.getStringValue());
             default -> {
-                log.warn("Unsupported value type for cost extraction: {}", value.getValueCase());
+                log.warn("Unsupported value type for cost extraction: '{}'", value.getValueCase());
                 yield Optional.empty();
             }
         };
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java
@@ -112,17 +112,8 @@ public class OpenTelemetryMappingUtils {
      */
     public static Optional<BigDecimal> extractCost(@NonNull AnyValue value) {
         return switch (value.getValueCase()) {
-            case DOUBLE_VALUE -> {
-                // BigDecimal.valueOf throws NumberFormatException for NaN / Infinity, which would
-                // abort span enrichment for the whole batch. Treat non-finite values as unsupported.
-                double d = value.getDoubleValue();
-                if (!Double.isFinite(d)) {
-                    log.warn("Non-finite cost value, skipping: '{}'", d);
-                    yield Optional.empty();
-                }
-                yield Optional.of(BigDecimal.valueOf(d));
-            }
-            case INT_VALUE -> Optional.of(BigDecimal.valueOf(value.getIntValue()));
+            case DOUBLE_VALUE -> parseCostDouble(value.getDoubleValue());
+            case INT_VALUE -> parseCostInt(value.getIntValue());
             case STRING_VALUE -> parseCostString(value.getStringValue());
             default -> {
                 log.warn("Unsupported value type for cost extraction: '{}'", value.getValueCase());
@@ -131,11 +122,31 @@ public class OpenTelemetryMappingUtils {
         };
     }
 
+    private static Optional<BigDecimal> parseCostDouble(double value) {
+        // BigDecimal.valueOf throws NumberFormatException for NaN / Infinity. Wider catch keeps
+        // cost extraction non-fatal regardless of how the conversion fails in the future.
+        try {
+            return Optional.of(BigDecimal.valueOf(value));
+        } catch (RuntimeException e) {
+            log.warn("Failed to parse cost double value '{}'", value);
+            return Optional.empty();
+        }
+    }
+
+    private static Optional<BigDecimal> parseCostInt(long value) {
+        try {
+            return Optional.of(BigDecimal.valueOf(value));
+        } catch (RuntimeException e) {
+            log.warn("Failed to parse cost int value '{}'", value);
+            return Optional.empty();
+        }
+    }
+
     private static Optional<BigDecimal> parseCostString(String stringValue) {
         try {
-            return Optional.of(new BigDecimal(stringValue.trim()));
-        } catch (NumberFormatException e) {
-            log.debug("Failed to parse cost string value '{}' as a number", stringValue);
+            return Optional.of(new BigDecimal(stringValue.strip()));
+        } catch (RuntimeException e) {
+            log.warn("Failed to parse cost string value '{}' as a number", stringValue);
             return Optional.empty();
         }
     }
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java
@@ -9,9 +9,11 @@ import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 
 import java.io.UncheckedIOException;
+import java.math.BigDecimal;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.stream.Stream;
 
 /**
@@ -97,6 +99,35 @@ public class OpenTelemetryMappingUtils {
                 // extracting from a JSON object
                 tryExtractUsageFromJsonObject(usage, key, value.getStringValue());
             }
+        }
+    }
+
+    /**
+     * Extracts a pre-computed cost value from an AnyValue. Some integrations (e.g. LiteLLM) send the
+     * exact request cost as a float attribute; we use it as the authoritative cost instead of recalculating.
+     * Supports double, integer, and numeric string values.
+     *
+     * @param value the AnyValue containing the cost
+     * @return the parsed cost, or empty if the value is not a parseable number
+     */
+    public static Optional<BigDecimal> extractCost(@NonNull AnyValue value) {
+        return switch (value.getValueCase()) {
+            case DOUBLE_VALUE -> Optional.of(BigDecimal.valueOf(value.getDoubleValue()));
+            case INT_VALUE -> Optional.of(BigDecimal.valueOf(value.getIntValue()));
+            case STRING_VALUE -> parseCostString(value.getStringValue());
+            default -> {
+                log.warn("Unsupported value type for cost extraction: {}", value.getValueCase());
+                yield Optional.empty();
+            }
+        };
+    }
+
+    private static Optional<BigDecimal> parseCostString(String stringValue) {
+        try {
+            return Optional.of(new BigDecimal(stringValue.trim()));
+        } catch (NumberFormatException e) {
+            log.debug("Failed to parse cost string value '{}' as a number", stringValue);
+            return Optional.empty();
         }
     }
 
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java
@@ -127,8 +127,8 @@ public class OpenTelemetryMappingUtils {
         // cost extraction non-fatal regardless of how the conversion fails in the future.
         try {
             return Optional.of(BigDecimal.valueOf(value));
-        } catch (RuntimeException e) {
-            log.warn("Failed to parse cost double value '{}'", value);
+        } catch (RuntimeException exception) {
+            log.warn("Failed to parse cost double value '{}'", value, exception);
             return Optional.empty();
         }
     }
@@ -136,8 +136,8 @@ public class OpenTelemetryMappingUtils {
     private static Optional<BigDecimal> parseCostInt(long value) {
         try {
             return Optional.of(BigDecimal.valueOf(value));
-        } catch (RuntimeException e) {
-            log.warn("Failed to parse cost int value '{}'", value);
+        } catch (RuntimeException exception) {
+            log.warn("Failed to parse cost int value '{}'", value, exception);
             return Optional.empty();
         }
     }
@@ -145,8 +145,8 @@ public class OpenTelemetryMappingUtils {
     private static Optional<BigDecimal> parseCostString(String stringValue) {
         try {
             return Optional.of(new BigDecimal(stringValue.strip()));
-        } catch (RuntimeException e) {
-            log.warn("Failed to parse cost string value '{}' as a number", stringValue);
+        } catch (RuntimeException exception) {
+            log.warn("Failed to parse cost string value '{}' as a number", stringValue, exception);
             return Optional.empty();
         }
     }
--- a/apps/opik-backend/src/test/java/com/comet/opik/domain/OpenTelemetryMapperTest.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/domain/OpenTelemetryMapperTest.java
@@ -225,56 +225,13 @@ class OpenTelemetryMapperTest {
     void genAiUsageCostExtractedIntoTotalEstimatedCost(String label, AnyValue costValue, BigDecimal expected) {
         // Pre-computed cost from gen_ai.usage.cost lands on the span's totalEstimatedCost for every
         // supported value shape. Malformed values (non-finite doubles, unparseable strings) are
-        // skipped so they never abort span enrichment.
+        // skipped so they never abort span enrichment. Sibling token attributes still populate the
+        // integer-only usage map without a 'cost' key — guarding against the original bug where the
+        // double cost was routed there and silently dropped.
         var attributes = List.of(
                 KeyValue.newBuilder()
                         .setKey("gen_ai.usage.cost")
                         .setValue(costValue)
-                        .build());
-
-        var spanBuilder = Span.builder()
-                .id(UUID.randomUUID())
-                .traceId(UUID.randomUUID())
-                .projectId(UUID.randomUUID())
-                .startTime(Instant.now());
-
-        OpenTelemetryMapper.enrichSpanWithAttributes(spanBuilder, attributes, null, null);
-
-        var span = spanBuilder.build();
-
-        if (expected == null) {
-            assertThat(span.totalEstimatedCost()).isNull();
-        } else {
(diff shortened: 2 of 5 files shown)


apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java
```diff
@@ -100,6 +102,44 @@ public static void extractUsageField(@NonNull Map<String, Integer> usage, @NonNu
...
+                double d = value.getDoubleValue();
+                if (!Double.isFinite(d)) {
+                    log.warn("Non-finite cost value, skipping: '{}'", d);
+                    yield Optional.empty();
+                }
+                yield Optional.of(BigDecimal.valueOf(d));
+            }
+            case INT_VALUE -> Optional.of(BigDecimal.valueOf(value.getIntValue()));
+            case STRING_VALUE -> parseCostString(value.getStringValue());
+            default -> {
```
andrescrz, line 118: Minor: The `ValueCase` enum might have some extra value that maybe we can extract the cost from, such as `STRING_VALUE_STRINDEX`. Feel free to investigate and extend this switch statement to cover it.

apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java
```diff
@@ -100,6 +102,44 @@ public static void extractUsageField(@NonNull Map<String, Integer> usage, @NonNu
...
+        return switch (value.getValueCase()) {
+            case DOUBLE_VALUE -> {
+                // BigDecimal.valueOf throws NumberFormatException for NaN / Infinity, which would
+                // abort span enrichment for the whole batch. Treat non-finite values as unsupported.
+                double d = value.getDoubleValue();
+                if (!Double.isFinite(d)) {
+                    log.warn("Non-finite cost value, skipping: '{}'", d);
+                    yield Optional.empty();
+                }
+                yield Optional.of(BigDecimal.valueOf(d));
```
andrescrz: Minor: feel free to encapsulate the `double` parsing logic to an small aux method, like you do for string, and add a try catch there with a warning when unable to parse. 

That way you make this more endurable. But also you can simplify the logic and simply attempt to covert or warn and return empty otherwise, so you don't need to call `isFinite` etc.

apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java
```diff
@@ -100,6 +102,44 @@ public static void extractUsageField(@NonNull Map<String, Integer> usage, @NonNu
...
+            }
+            case INT_VALUE -> Optional.of(BigDecimal.valueOf(value.getIntValue()));
```
andrescrz: Nit: I believe all these methods don't throw any exception, according to their Javadocs. But feel free to also encapsulate this int conversion in a method with a try-catch and a warning, to make it more endurable for the future.

apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java
```diff
@@ -100,6 +102,44 @@ public static void extractUsageField(@NonNull Map<String, Integer> usage, @NonNu
...
+                log.warn("Unsupported value type for cost extraction: '{}'", value.getValueCase());
+                yield Optional.empty();
+            }
+        };
+    }
+
+    private static Optional<BigDecimal> parseCostString(String stringValue) {
+        try {
+            return Optional.of(new BigDecimal(stringValue.trim()));
+        } catch (NumberFormatException e) {
```
andrescrz: Minor: to have higher security that the cost extraction never disrupts the normal execution, you could widen this catch to RuntimeException.

apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java
```diff
@@ -100,6 +102,44 @@ public static void extractUsageField(@NonNull Map<String, Integer> usage, @NonNu
...
+            log.debug("Failed to parse cost string value '{}' as a number", stringValue);
```
andrescrz: Minor: not aligned with the rest of the code, better at warn level.

apps/opik-backend/src/test/java/com/comet/opik/domain/OpenTelemetryMapperTest.java
```diff
@@ -216,6 +220,87 @@ void testGenAiCostAttributesMapsToMetadata() {
...
+                KeyValue.newBuilder()
+                        .setKey("gen_ai.usage.cost")
+                        .setValue(costValue)
+                        .build());
+
+        var spanBuilder = Span.builder()
+                .id(UUID.randomUUID())
+                .traceId(UUID.randomUUID())
+                .projectId(UUID.randomUUID())
+                .startTime(Instant.now());
```
andrescrz: Minor: let's use Lombok to build POJOs, then simplify override the desired values with toBuilder etc.

apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java
```diff
@@ -100,6 +102,44 @@ public static void extractUsageField(@NonNull Map<String, Integer> usage, @NonNu
...
+            default -> {
+                log.warn("Unsupported value type for cost extraction: '{}'", value.getValueCase());
+                yield Optional.empty();
+            }
+        };
+    }
+
+    private static Optional<BigDecimal> parseCostString(String stringValue) {
+        try {
+            return Optional.of(new BigDecimal(stringValue.trim()));
```
andrescrz: Minor: in modern java, better go with strip.

apps/opik-backend/src/test/java/com/comet/opik/domain/OpenTelemetryMapperTest.java
```diff
@@ -216,6 +220,87 @@ void testGenAiCostAttributesMapsToMetadata() {
...
+        }
+    }
+
+    static Stream<Arguments> genAiUsageCostCases() {
+        return Stream.of(
+                Arguments.of("double", AnyValue.newBuilder().setDoubleValue(0.001234).build(),
+                        new BigDecimal("0.001234")),
+                Arguments.of("int", AnyValue.newBuilder().setIntValue(1).build(),
+                        new BigDecimal("1")),
+                Arguments.of("numeric string", AnyValue.newBuilder().setStringValue("0.005").build(),
```
andrescrz, line 271: Minor: missing test case, stripping string value and still succeeding.

apps/opik-backend/src/test/java/com/comet/opik/domain/OpenTelemetryMapperTest.java
```diff
@@ -216,6 +220,87 @@ void testGenAiCostAttributesMapsToMetadata() {
...
+                        AnyValue.newBuilder().setStringValue("not a number").build(), null),
+                Arguments.of("NaN", AnyValue.newBuilder().setDoubleValue(Double.NaN).build(), null),
+                Arguments.of("positive infinity",
+                        AnyValue.newBuilder().setDoubleValue(Double.POSITIVE_INFINITY).build(), null),
+                Arguments.of("negative infinity",
+                        AnyValue.newBuilder().setDoubleValue(Double.NEGATIVE_INFINITY).build(), null));
+    }
+
+    @Test
+    void testGenAiUsageCostDoesNotPolluteUsageMap() {
```
andrescrz: Nit: I believe the regression doesn't differ much from the existing normal test case. I'd just fit the assertions there.

andrescrz (dismissed):
Many minor comments, I optionally recommend a quick 2nd iteration, but not a blocker.

apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java
YarivHashaiComet: Addressed: extracted `parseCostDouble` as a dedicated aux method with a `try/catch` around `BigDecimal.valueOf`. Dropped the explicit `Double.isFinite` check since the catch now handles `NaN`/`Infinity` uniformly (they raise `NumberFormatException` from `BigDecimal.valueOf` → empty Optional). Same pattern as `parseCostString`.

apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java
YarivHashaiComet: Addressed: extracted `parseCostInt(long)` aux method with a `try/catch` for consistency with the other branches. As you noted the JDK doesn't currently throw here, but the wrapper keeps the surface uniform and future-proof.

apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java
YarivHashaiComet: Addressed: widened catches in all three `parseCost*` helpers to `RuntimeException` so any unexpected runtime failure during cost extraction yields `Optional.empty()` instead of aborting span enrichment.

apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java
YarivHashaiComet: Addressed: bumped the parse-failure log in `parseCostString` from `debug` to `warn`, aligning with the unsupported-value-type branch and the new double/int helpers.

apps/opik-backend/src/test/java/com/comet/opik/domain/OpenTelemetryMapperTest.java
YarivHashaiComet: Addressed: extracted a `newSpanBuilder()` helper that returns a `Span.SpanBuilder` with the common id/traceId/projectId/startTime defaults. The `gen_ai.usage.cost` parameterized test now uses it, so each case only overrides what it cares about.

apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java
YarivHashaiComet: Addressed: switched to `String.strip()` (Unicode-aware).

apps/opik-backend/src/test/java/com/comet/opik/domain/OpenTelemetryMapperTest.java
YarivHashaiComet: Addressed: added a `"numeric string with whitespace"` case (`"  0.005  "` → `0.005`) to `genAiUsageCostCases` to cover the `strip()` path.

apps/opik-backend/src/test/java/com/comet/opik/domain/OpenTelemetryMapperTest.java
YarivHashaiComet: Addressed: folded the regression assertions into the parameterized test. Each `gen_ai.usage.cost` case now also carries `prompt_tokens`/`completion_tokens` siblings and asserts the `usage` map has no `cost` key while preserving the token entries — so the no-pollution guard is exercised across all value shapes (double, int, numeric/whitespace string, malformed string, non-finite doubles). The standalone `testGenAiUsageCostDoesNotPolluteUsageMap` was removed.

apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java
```diff
@@ -100,6 +102,55 @@ public static void extractUsageField(@NonNull Map<String, Integer> usage, @NonNu
...
+        // cost extraction non-fatal regardless of how the conversion fails in the future.
+        try {
+            return Optional.of(BigDecimal.valueOf(value));
+        } catch (RuntimeException e) {
+            log.warn("Failed to parse cost double value '{}'", value);
+            return Optional.empty();
+        }
+    }
+
+    private static Optional<BigDecimal> parseCostInt(long value) {
```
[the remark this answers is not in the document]
YarivHashaiComet: Partially addressed in commit 9b0a796:

- ✅ Now passing the caught exception as the last `log.warn` arg in all three `parseCost*` helpers, preserving the stack trace.
- ❌ Keeping the catch at `RuntimeException` (not narrowing to `NumberFormatException` / `IllegalArgumentException`). That width was explicitly requested by andrescrz on this thread in [comment 3324055545](https://github.com/comet-ml/opik/pull/6893#discussion_r3324055545) — *"to have higher security that the cost extraction never disrupts the normal execution, you could widen this catch to RuntimeException"*. Narrowing now would revert an already-addressed reviewer ask.

apps/opik-backend/src/test/java/com/comet/opik/domain/OpenTelemetryMapperTest.java
```diff
@@ -216,6 +220,72 @@ void testGenAiCostAttributesMapsToMetadata() {
...
+                        AnyValue.newBuilder().setStringValue("not a number").build(), null),
+                Arguments.of("NaN", AnyValue.newBuilder().setDoubleValue(Double.NaN).build(), null),
+                Arguments.of("positive infinity",
+                        AnyValue.newBuilder().setDoubleValue(Double.POSITIVE_INFINITY).build(), null),
+                Arguments.of("negative infinity",
+                        AnyValue.newBuilder().setDoubleValue(Double.NEGATIVE_INFINITY).build(), null));
+    }
+
+    private static Span.SpanBuilder newSpanBuilder() {
+        return Span.builder()
```
andrescrz: Ni: I'd use Podam to build this.

andrescrz (dismissed):
LGTM

apps/opik-backend/src/test/java/com/comet/opik/domain/OpenTelemetryMapperTest.java
YarivHashaiComet: Addressed in commit ec61d9d: `newSpanBuilder()` now builds the baseline via `PodamFactoryUtils.newPodamFactory().manufacturePojo(Span.class).toBuilder()`, with `totalEstimatedCost` explicitly nulled out — `enrich` only writes that field on a successful parse, so the malformed/non-finite cost cases would otherwise inherit PODAM's random `BigDecimal`. All 8 cost cases pass.

andrescrz (approved):

PR #5569, opened 2026-03-09
[OPIK-4762] [BE] feat: add dashboard type and scope columns

andriidudar:
## Details

Add `type` and `scope` columns to the dashboards table to classify dashboards by their data source and accessibility context.

**Type** (`multi-project` | `experiments`): what data the dashboard works with.
**Scope** (`workspace` | `insights`): where the dashboard is accessible.

### Changes

- **Enums**: `DashboardType` and `DashboardScope` with JSON serialization, JDBI argument factories and column mappers
- **Migration**: `000053_add_dashboard_type_scope.sql` — adds `type` and `scope` columns with `NOT NULL DEFAULT`
- **DTO**: `Dashboard` and `DashboardUpdate` records updated with new fields
- **DAO**: Insert/update queries include type and scope (COALESCE for PATCH)
- **Filter infrastructure**: `DashboardField`, `DashboardFilter`, `FilterStrategy.DASHBOARD` — enables filtering by type, scope, name, id, timestamps, created_by, last_updated_by
- **Resource**: `GET /v1/private/dashboards` now accepts `?filters=` query param (same pattern as Alerts)
- **Sorting**: `type` added to sortable fields
- **Service**: Defaults type/scope on create via `Optional.ofNullable().orElse()` pattern
- **Tests**: 6 new test cases covering create with/without type+scope, update type/scope, filter by type/scope/both

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues
- OPIK-4762

## Testing
- New integration tests in `DashboardsResourceTest`:
  - Create dashboard with explicit type and scope
  - Create dashboard without type/scope (defaults applied)
  - Update dashboard type via PATCH
  - Update dashboard scope via PATCH
  - Filter dashboards by type
  - Filter dashboards by scope
  - Filter dashboards by both type and scope
- Existing dashboard tests continue to pass (backward compatible)

## Documentation
N/A — internal API change, no public documentation updates needed.

PR #6905, opened 2026-05-28
[OPIK-6666] [FE] feat: add /oauth/consent UI for MCP OAuth flow

LifeXplorer:
## Details

Adds the user-facing consent screen that completes the MCP OAuth Authorization Server flow. opik-backend's `GET /oauth/authorize` validates the request + session and 302-redirects the browser to `${baseUrl}/oauth/consent`; this PR adds the SPA route + page that renders the consent card and posts the user's decision back.

Stacked on top of the backend AS work. **Merge order:** comet-backend (`/opik/auth-by-username`, [#5569](https://github.com/comet-ml/comet-backend/pull/5569)) → opik backend AS → this PR. Now retargeted onto the backend base branch.

- `src/api/api.ts` — `OAUTH_REST_ENDPOINT` and `OAUTH_AUTHORIZE_CONTEXT_KEY`.
- `src/api/oauth/types.ts` — `OAuthAuthorizeContext`, `OAuthConsentRequest`, `OAuthConsentResponse` (snake_case to match the backend's global `SnakeCaseStrategy`).
- `src/api/oauth/useOAuthAuthorizeContext.ts` — TanStack Query hook for `GET /api/oauth/authorize/context` (`retry: false`, 4xx is terminal).
- `src/api/oauth/useOAuthConsentMutation.ts` — TanStack Mutation for `POST /api/oauth/authorize`; returns `{redirect_to}`.
- `src/plugins/comet/WorkspacePreloader.tsx` — **captures the return target for the unauthenticated MCP OAuth path.** When an unauthenticated user is bounced to login, `persistMcpOAuthReturn()` reads a **same-host** `?returnTo=` (carried by opik-backend) and persists it into `localStorage['mcpOAuthRedirectURL']` before the redirect. A dedicated key (not `redirectURLAfterLogin`, which the SSO modal overwrites) so the target survives both email and SSO logins. The **consume half** runs on the Comet login host post-login — comet-ml/comet-react#7733 — using the identical key literal and the same same-host guard.
- `src/shared/OAuthConsentPage/OAuthConsentPage.tsx` — consent card: client name/logo + radio-group workspace picker + Allow/Deny.
- `src/v1/router.tsx` / `src/v2/router.tsx` — root-level `/oauth/consent` route in both routers (lazy import from `src/shared/`), because `WorkspaceVersionGate` picks v1 vs v2 before any workspace is resolved and the consent URL is pre-workspace.
- `deployment/docker-compose/nginx_default_local.conf` — local-dev nginx routing for the OAuth surfaces (consent SPA route, AS endpoints, RFC 8414 discovery URL).

Behaviour: **Allow** → POSTs the original OAuth params + chosen workspace + CSRF token → navigates to `redirect_to`. **Deny** → `<redirect_uri>?error=access_denied&state=<state>` (RFC 6749 §4.1.2.1). **Missing params / context error** → terminal error card; does NOT auto-follow the untrusted `redirect_uri`.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- OPIK-6666 — https://comet-ml.atlassian.net/browse/OPIK-6666

## AI-WATERMARK

AI-WATERMARK: yes

- If yes:
  - Tools: Claude Code
  - Model(s): Claude Opus 4.8
  - Scope: FE consent page, OAuth API hooks/types, return-target capture in WorkspacePreloader, v1/v2 routes, local nginx routing; merge/conflict resolution onto the new backend base branch.
  - Human verification: Author review; end-to-end verification on a local dev cluster against the backend branches.

## Testing

- To be verified end-to-end on a local dev cluster against the two backend branches (comet-backend + opik backend AS), exercising the Allow / Deny / missing-params paths.
- Merge onto the new base resolved in favor of the finalized backend; FE request/response types verified field-compatible with the base DTOs (`AuthorizeContext`, `ConsentRequest`, `ConsentResponse`, `WorkspaceInfo`). Full local build not yet run on the merge result.

Follow-ups (out of scope here): Vitest unit + Playwright E2E coverage.

## Documentation

Inline English strings only (no i18n yet). No external/Fern docs change in this PR.

Follow-ups (out of scope):
- 401-specific login redirect (`${baseUrl}/login?return_to=…`) instead of the generic terminal-error card.
- i18n for the consent page strings.
- Single-user OSS banner once the backend's `AuthorizeContext` exposes a `mode` flag.


apps/opik-frontend/src/shared/OAuthConsentPage/OAuthConsentPage.tsx
```diff
@@ -0,0 +1,236 @@
...
+    response_type: q.get("response_type") as string,
+    code_challenge: q.get("code_challenge") as string,
+    code_challenge_method: q.get("code_challenge_method") as string,
+    resource: q.get("resource") as string,
+    state: q.get("state"),
+  };
+};
+
+const denyAndRedirect = (redirectUri: string, state: string | null) => {
+  const url = new URL(redirectUri);
```
LifeXplorer: **[medium] denyAndRedirect throws on malformed redirect_uri and crashes the page**

denyAndRedirect calls `new URL(redirectUri)` on the raw `redirect_uri` query param with no try/catch. `redirect_uri` comes straight from `window.location.search` (parseParams only checks it is non-blank, not that it is a valid absolute URL). If the value is relative or malformed, `new URL()` throws synchronously inside the Deny click handler, producing an uncaught exception and a broken/blank consent page with no recovery path. The sibling code in src/plugins/comet/WorkspacePreloader.tsx (line 55) wraps the same `new URL()` call in try/catch precisely for this reason.

💡 Wrap the URL construction in try/catch (mirroring WorkspacePreloader) and fall back to rendering a ConsentError / toast when the redirect_uri cannot be parsed, instead of letting the exception bubble up.

_general finding_

LifeXplorer (commented):

9 findings — 0 high · 1 medium · 8 low. Suppressed by team conventions: see `suppressed.md`.

apps/opik-frontend/src/api/oauth/types.ts
```diff
@@ -0,0 +1,28 @@
+export type OAuthWorkspaceInfo = {
+  id: string;
+  name: string;
+};
+
+export type OAuthAuthorizeContext = {
+  client_name: string;
+  client_logo_uri: string | null;
```
LifeXplorer: **[low] client_logo_uri should be optional, not just nullable**

The backend AuthorizeContext record (apps/opik-backend/.../oauth/AuthorizeContext.java) is annotated with @JsonInclude(JsonInclude.Include.NON_NULL). When clientLogoUri is null the field is OMITTED from the JSON entirely, so on the wire it arrives as `undefined`, never as `null`. Declaring `client_logo_uri: string | null` does not model the absent case: any code that relies on the property existing (e.g. `'client_logo_uri' in data`, or destructuring with a `null` default `const { client_logo_uri = null }`) would be mistyped. The consumer OAuthConsentPage.tsx happens to use a truthy check so it is safe today, but the type is contractually inaccurate. Per the enforced rule, fields the backend can omit must be marked optional.

💡 Mark the field optional and nullable: `client_logo_uri?: string | null;`. For consistency, the other NON_NULL-serialized fields (workspaces, csrf_token) can also be absent and are accessed with `?? []` / used directly downstream, so consider `workspaces?: OAuthWorkspaceInfo[];` and `csrf_token?: string;` as well to match the real wire contract.

_rule: `frontend-api-contract-keep-frontend-typescript-types-fully-aligned-with-the-backen`_

apps/opik-frontend/src/api/oauth/useOAuthConsentMutation.ts
```diff
@@ -0,0 +1,18 @@
+import { useMutation } from "@tanstack/react-query";
+import api, { OAUTH_REST_ENDPOINT } from "@/api/api";
+import { OAuthConsentRequest, OAuthConsentResponse } from "@/api/oauth/types";
+
+const useOAuthConsentMutation = () =>
```
LifeXplorer, line 5: **[low] Mutation hook has no default onError; full server error detail is never surfaced**

Unlike every other mutation hook in src/api (e.g. useTraceCommentsBatchDeleteMutation, useThreadBatchDeleteMutation, useInsightsViewBatchDeleteMutation), this hook defines no default onError handler. Error handling is delegated entirely to the single consumer (OAuthConsentPage.handleAllow), which shows a static 'The server rejected the consent' message and discards error.response.data.message / errors[]. As a result the actual backend reason (invalid CSRF, expired session, unregistered client, etc.) is never shown to the user, making consent failures hard to diagnose. The sibling useOAuthAuthorizeContext also omits a default handler, so the omission is consistent within the oauth/ dir but diverges from the rest of the API layer.

💡 Either add a default onError in this hook that joins error.response.data.errors and error.response.data.message with an error.message fallback (matching the team's get(error, ['response','data','message'], error.message) toast pattern), or have the consumer's onError extract and display those server fields instead of a fixed string. Keep the same extraction consistently across both oauth hooks.

_rule: `frontend-error-handling-extract-full-server-error-details-in-mutation-onerror-toasts`_

apps/opik-frontend/src/plugins/comet/WorkspacePreloader.tsx
```diff
@@ -42,6 +42,24 @@ const redirectToEM = () => {
...
+// MCP OAuth return survives both email and SSO logins. comet-react consumes it post-login.
+const MCP_OAUTH_REDIRECT_URL_KEY = "mcpOAuthRedirectURL";
+
+// Persist the MCP OAuth authorize URL (carried as ?returnTo= by opik-backend) before
+// bouncing to login, so the user returns there after authenticating.
+const persistMcpOAuthReturn = () => {
+  const returnTo = new URLSearchParams(window.location.search).get("returnTo");
+  if (!returnTo) return;
+  try {
+    if (new URL(returnTo).host === window.location.host) {
```
LifeXplorer: **[low] returnTo host check ignores protocol**

persistMcpOAuthReturn validates only that new URL(returnTo).host matches window.location.host before persisting it to localStorage. The protocol/scheme is not checked, so a value like 'javascript:...' would not survive (URL.host would be empty and not match), but a downgraded 'http://<host>' on an https app would still be accepted and later consumed by comet-react as a redirect target. Since the stored value is read elsewhere (external comet-react) and used to navigate, the host-only check is the single trust boundary here.

💡 Compare against the full origin instead of host (e.g. new URL(returnTo).origin === window.location.origin) so protocol and port are validated together, closing scheme-downgrade and port-mismatch gaps. AssistantSidebar.tsx already uses new URL(probeUrl).origin for the analogous check.

_general finding_

apps/opik-frontend/src/shared/OAuthConsentPage/OAuthConsentPage.tsx
```diff
@@ -0,0 +1,236 @@
...
+import { Label } from "@/ui/label";
+import { RadioGroup, RadioGroupItem } from "@/ui/radio-group";
+import Loader from "@/shared/Loader/Loader";
+import { useToast } from "@/ui/use-toast";
+
+type ParsedParams = {
+  client_id: string;
+  redirect_uri: string;
+  response_type: string;
+  code_challenge: string;
```
LifeXplorer: **[low] response_type is required and parsed but never used**

`response_type` is declared on ParsedParams, listed in the `required` array, and extracted into the parsed object, but it is never read afterward — it is not sent in the consent mutation body (OAuthConsentRequest has no response_type field) nor used in the deny redirect. It is effectively dead data; its only effect is to reject the request when absent.

💡 If the field is genuinely needed as a presence gate, keep it in `required` but drop it from the ParsedParams type/object; if it is not needed at all, remove it entirely. Otherwise forward it where the backend expects it.

_rule: `frontend-style-maintainability-remove-dead-code-eliminate-branches-where-both-paths-are-id`_

apps/opik-frontend/src/shared/OAuthConsentPage/OAuthConsentPage.tsx
```diff
@@ -0,0 +1,236 @@
...
+        state: params.state,
+        workspace_id: workspace.id,
+        workspace_name: workspace.name,
+        csrf: data.csrf_token,
+      },
+      {
+        onSuccess: ({ redirect_to }) => {
+          window.location.href = redirect_to;
+        },
+        onError: () => {
```
LifeXplorer: **[low] onError toast uses a fixed string and discards server error detail**

The consent mutation onError handler shows a hardcoded description and ignores the server response (error.response.data.errors / message). When the backend rejects the consent (e.g. expired CSRF, invalid PKCE challenge, unregistered redirect), the user gets a generic 'server rejected the consent' message with no actionable detail, and the real cause is dropped. The team convention is to join error.response.data.errors and combine with error.response.data.message, falling back to error.message and then a static string.

💡 Type the onError param and build the description from error.response.data.errors / message with the current static string only as the final fallback.

_rule: `frontend-error-handling-extract-full-server-error-details-in-mutation-onerror-toasts`_

deployment/docker-compose/nginx_default_local.conf
```diff
@@ -58,6 +58,53 @@ server {
...
+    # from the SPA bundle. Listed BEFORE the /oauth/ prefix block so nginx's
+    # exact-match precedence keeps consent on the FE instead of the backend.
+    location = /oauth/consent {
+        try_files /dev/null /index.html;
+    }
+
+    # AS endpoints (/oauth/authorize, /oauth/authorize/context, /oauth/token,
+    # /oauth/revoke, /oauth/register). Backend's Jersey resources mount these
+    # paths directly (no /api prefix in @Path), so no rewrite is needed.
+    location /oauth/ {
```
LifeXplorer: **[low] Local /oauth/ block diverges from canonical helm config (missing proxy_redirect off, plain prefix vs ^~)**

The production/helm equivalent (deployment/helm_chart/opik/templates/configmap-frontend-nginx.yaml lines 165-190) routes /oauth/* through a named location with 'proxy_redirect off' and uses 'location ^~ /oauth/' for the prefix. This local config omits 'proxy_redirect off' and uses a plain 'location /oauth/' prefix. The OAuth authorize flow (OAuthAuthorizeResource) responds almost entirely with 302 redirects whose Location is built from config.getBaseUrl() (absolute, deployment-root), so nginx's default proxy_redirect should not rewrite them in practice — hence low. But the two configs drifting apart is a maintenance hazard: a future change verified only against local will not reflect prod behavior. Keeping them aligned (and explicit) avoids surprises.

💡 Mirror the helm pattern: add 'proxy_redirect off;' to the /oauth/ block, and consider 'location ^~ /oauth/' so it cannot be pre-empted if a regex location is ever added. At minimum add a comment noting that this local file intentionally mirrors configmap-frontend-nginx.yaml and both must be kept in sync.

_general finding_

deployment/docker-compose/nginx_default_local.conf
```diff
@@ -58,6 +58,53 @@ server {
...
+    # OAuth surfaces live at the deployment's natural root namespace, sibling
+    # to the data API at /api/* — NOT nested under it. The issuer URL is the
+    # deployment root (e.g. http://localhost:5173 locally; https://host/opik
+    # in prod). This matches industry OAuth conventions (GitHub /login/oauth/*,
+    # Auth0 /oauth/*) and avoids the consent-redirect dead-end you get when
+    # the issuer points at the API namespace (consent_redirect = baseUrl +
+    # "/oauth/consent" would land at /api/oauth/consent which doesn't exist
+    # on the FE).
+    #
+    # Exact match for the consent UI — React Router renders /oauth/consent
```
LifeXplorer, line 70: **[low] Comment's claim that block ordering enforces consent-on-FE is inaccurate**

The comment states the exact-match consent block is 'Listed BEFORE the /oauth/ prefix block so nginx's exact-match precedence keeps consent on the FE'. In nginx, an exact match (location =) always wins over a prefix match regardless of the textual order of the blocks; ordering of these two locations is not what makes it work. The behavior is correct, but the stated reasoning is wrong and could mislead a future maintainer into thinking reordering would break routing.

💡 Reword to: 'Exact-match (= ) locations always take precedence over prefix matches in nginx regardless of order, so the consent UI is served from the SPA while all other /oauth/* paths proxy to the backend.' Drop the 'Listed BEFORE ... so ... precedence' phrasing.

_general finding_

deployment/docker-compose/nginx_default_local.conf
```diff
@@ -58,6 +58,53 @@ server {
...
+    # exact-match precedence keeps consent on the FE instead of the backend.
+    location = /oauth/consent {
+        try_files /dev/null /index.html;
+    }
+
+    # AS endpoints (/oauth/authorize, /oauth/authorize/context, /oauth/token,
+    # /oauth/revoke, /oauth/register). Backend's Jersey resources mount these
+    # paths directly (no /api prefix in @Path), so no rewrite is needed.
+    location /oauth/ {
+        proxy_pass http://backend;
```
LifeXplorer, line 84: **[low] Proxy header block duplicated a third time**

The full proxy_set_header / proxy_*_timeout / http_version upgrade block is now copy-pasted across three locations (@api, /oauth/, and partially the well-known block). The well-known block additionally drops the proxy_read/connect/send_timeout and http_version/Upgrade lines present in the other two, so the three are subtly inconsistent. This is a maintainability nit (nginx lacks first-class includes here without a separate snippet file), but the inconsistency between the blocks is the kind of thing that causes 'works for /oauth but not /.well-known' confusion later.

💡 If keeping inline, at least make the three blocks intentionally consistent (or document why the well-known proxy intentionally omits the timeouts/upgrade headers since metadata is a small GET). Optionally factor common proxy directives into an included snippet to avoid drift.

_general finding_

apps/opik-frontend/src/v1/router.tsx
```diff
@@ -130,6 +130,19 @@ const pairingRouteOssAlias = createRoute({
...
+// is selected, so both V1 and V2 routers must serve it — WorkspaceVersionGate picks the
+// app from cached defaults and the consent page itself is independent of either layout.
+const OAuthConsentPage = lazy(
+  () => import("@/shared/OAuthConsentPage/OAuthConsentPage"),
+);
+const oauthConsentRoute = createRoute({
+  getParentRoute: () => rootRoute,
+  path: "/oauth/consent",
+  component: OAuthConsentPage,
+});
```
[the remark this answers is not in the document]
LifeXplorer: Declined for now. The v1/v2 routers are deliberately separate trees, and this is ~5 lines of `createRoute` per side where the parent route and registration array differ. Extracting a shared `createOAuthConsentRoute(rootRoute)` helper for a single route would add an abstraction without removing meaningful duplication, so I've left it inline to match the rest of both routers.

deployment/docker-compose/nginx_default_local.conf
```diff
@@ -58,6 +58,53 @@ server {
...
+    # deployment root (e.g. http://localhost:5173 locally; https://host/opik
+    # in prod). This matches industry OAuth conventions (GitHub /login/oauth/*,
+    # Auth0 /oauth/*) and avoids the consent-redirect dead-end you get when
+    # the issuer points at the API namespace (consent_redirect = baseUrl +
+    # "/oauth/consent" would land at /api/oauth/consent which doesn't exist
+    # on the FE).
+    #
+    # Exact match for the consent UI — React Router renders /oauth/consent
+    # from the SPA bundle. Listed BEFORE the /oauth/ prefix block so nginx's
+    # exact-match precedence keeps consent on the FE instead of the backend.
```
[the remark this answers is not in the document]
LifeXplorer: Done in b3ead38. Added the three OAuth blocks (`= /oauth/consent`, `^~ /oauth/`, `= /.well-known/oauth-authorization-server`) before the SPA catch-all in both `nginx_guardrails_local.conf` and `nginx_local_be_local.conf`, mirroring `nginx_default_local.conf`. Helm + the default config were already covered.

apps/opik-frontend/src/plugins/comet/WorkspacePreloader.tsx
```diff
@@ -42,6 +42,24 @@ const redirectToEM = () => {
...
+      window.localStorage.setItem(MCP_OAUTH_REDIRECT_URL_KEY, returnTo);
```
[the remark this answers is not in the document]
LifeXplorer: Done in b3ead38. `persistMcpOAuthReturn` now compares `new URL(returnTo).origin === window.location.origin`, so protocol and port are validated alongside host (closes the http-downgrade gap).

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthRegisterResource.java
```diff
@@ -0,0 +1,91 @@
...
+        return Response.created(URI.create("/admin/mcp-oauth-clients/" + client.clientId()))
+                .entity(body)
+                .build();
+    }
+
+    // First X-Forwarded-For hop set by the fronting nginx; direct remote address otherwise.
+    private static String clientIp(HttpServletRequest request) {
+        String forwarded = request.getHeader("X-Forwarded-For");
+        if (StringUtils.isNotBlank(forwarded)) {
+            return forwarded.split(",")[0].trim();
```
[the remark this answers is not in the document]
LifeXplorer: Already addressed in 7464dd4: `clientIp()` keys the throttle off the right-most `X-Forwarded-For` hop (nginx's `$proxy_add_x_forwarded_for` peer), so the caller-controlled left-most hops can't influence the DCR rate-limit bucket. No further change.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/McpOAuthConfig.java
```diff
@@ -0,0 +1,76 @@
...
+    // Grace window during which a just-rotated refresh token is still accepted
+    @Valid @JsonProperty
+    @NotNull private Duration refreshRotationGrace = Duration.ofSeconds(30);
+
+    // DCR (RFC 7591) is deliberately unauthenticated, so cap registrations per source IP per window.
+    @Valid @JsonProperty
+    private long registrationRateLimit = 20;
+
+    @Valid @JsonProperty
+    @NotNull private Duration registrationRateLimitDuration = Duration.ofHours(1);
```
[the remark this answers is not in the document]
LifeXplorer: Fully addressed. `@Min(1)`/`@NotNull` are on both fields, and the YAML keys exist in both `config.yml` (`registrationRateLimit: ${...:-20}`, `registrationRateLimitDuration: ${...:-PT1H}`) and `config-test.yml`, so they're overridable via env and fail fast if missing. No further change.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/McpOAuthFlowE2ETest.java
```diff
@@ -0,0 +1,433 @@
...
+        }
+
+        // the rotated access token works before the reuse attack
+        try (var response = callDataApi(newAccess)) {
+            assertThat(response.getStatus()).isEqualTo(Response.Status.OK.getStatusCode());
+        }
+
+        await().pollDelay(ROTATION_GRACE_MS + 500, TimeUnit.MILLISECONDS)
+                .atMost(ROTATION_GRACE_MS + 5_000, TimeUnit.MILLISECONDS)
+                .until(() -> true);
```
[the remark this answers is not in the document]
LifeXplorer: Done in b3ead38. Replaced `await().pollDelay(...).until(() -> true)` with a plain `Thread.sleep(ROTATION_GRACE_MS + 500)` plus a comment noting the grace window is a genuine wall-clock wait (`refresh()` gates on `Instant.now()` vs `revokedAt + grace`). Removed the now-unused awaitility/`TimeUnit` imports.

apps/opik-frontend/src/shared/OAuthConsentPage/OAuthConsentPage.tsx
LifeXplorer: Done in b3ead38. `denyAndRedirect` now wraps `new URL()` in try/catch and returns a success boolean; `handleDeny` shows a destructive toast ("redirect target is invalid…") when the value can't be parsed, instead of letting the exception bubble up.

apps/opik-frontend/src/api/oauth/types.ts
LifeXplorer: Done in b3ead38. Marked `client_logo_uri?: string | null` to model the omitted (`undefined`) case from `@JsonInclude(NON_NULL)`. Left `workspaces`/`csrf_token` as-is: an empty `workspaces` array still serializes (NON_NULL only drops actual nulls) and `csrf_token` is always populated, so they don't hit the absent-field case.

apps/opik-frontend/src/api/oauth/useOAuthConsentMutation.ts
LifeXplorer: Addressed at the consumer in b3ead38: the consent `onError` in `OAuthConsentPage` now extracts `get(error, ["response","data","message"], <static fallback>)`. I kept it in the page rather than adding a default to the hook so the user-facing fallback string lives with the UI and the hook stays a thin mutation — happy to lift it into a shared default if you'd rather match the other API hooks exactly.

apps/opik-frontend/src/plugins/comet/WorkspacePreloader.tsx
LifeXplorer: Done in b3ead38 (same fix as the related origin thread): `persistMcpOAuthReturn` now compares `.origin` instead of `.host`, closing the scheme-downgrade and port-mismatch gaps.

apps/opik-frontend/src/shared/OAuthConsentPage/OAuthConsentPage.tsx
LifeXplorer: Done in b3ead38. Kept `response_type` in the `required` presence gate but dropped it from the `ParsedParams` type and the returned object, with a comment noting it's validated-but-not-forwarded (the backend already validated it before this redirect).

apps/opik-frontend/src/shared/OAuthConsentPage/OAuthConsentPage.tsx
LifeXplorer: Done in b3ead38. The consent `onError` now builds the toast description from `error.response.data.message` (via lodash `get`) with the static string only as the final fallback.

deployment/docker-compose/nginx_default_local.conf
LifeXplorer: Done in b3ead38. The local `/oauth/` block now uses `location ^~ /oauth/` and `proxy_redirect off`, matching the helm template, plus a comment noting the two configs must be kept in sync.

deployment/docker-compose/nginx_default_local.conf
LifeXplorer: Done in b3ead38. Reworded the comment: exact-match (`= `) locations always win over prefix matches in nginx regardless of block order — dropped the misleading "listed BEFORE … so precedence" phrasing.

deployment/docker-compose/nginx_default_local.conf
LifeXplorer: Documented in b3ead38: added a comment that the `/.well-known` block intentionally omits the upgrade/timeout directives since it's a small metadata GET. Left the three blocks inline rather than factoring a snippet — nginx includes would require a separate mounted file, so this is a conscious trade-off.

apps/opik-frontend/src/shared/OAuthConsentPage/OAuthConsentPage.tsx
```diff
@@ -0,0 +1,258 @@
...
+  try {
+    const url = new URL(redirectUri);
+    url.searchParams.set("error", "access_denied");
+    if (state) url.searchParams.set("state", state);
+    window.location.href = url.toString();
+    return true;
+  } catch {
+    return false;
+  }
+};
```
awkoy: **[security] The Deny path trusts the raw `redirect_uri` from the URL.** Allow is safe (the backend builds `redirect_to`), but Deny navigates straight to the `redirect_uri` query param. The only gate is that the consent card renders only after `contextQuery` succeeds — so this is safe *iff* `GET /oauth/authorize/context` rejects a `redirect_uri` that isn't registered for the `client_id`.

If that endpoint validates only `client_id`, a crafted link like `/oauth/consent?client_id=<valid>&redirect_uri=https://evil.com&…` renders a legit-looking consent card and Deny becomes an open redirect to `evil.com`.

Can we confirm the context endpoint validates `redirect_uri`? If it can't be guaranteed, validate `redirect_uri` client-side against an allowed value before navigating.

apps/opik-frontend/src/shared/OAuthConsentPage/OAuthConsentPage.tsx
```diff
@@ -0,0 +1,258 @@
...
+    <FullPage>
+      <Card className="w-full max-w-md">
+        <CardHeader className="space-y-3">
+          {data.client_logo_uri ? (
+            <img
+              src={data.client_logo_uri}
+              alt={`${data.client_name} logo`}
+              className="size-12 rounded"
+            />
+          ) : null}
```
awkoy: `client_name` / `client_logo_uri` come from dynamic client registration (RFC 7591), i.e. attacker-controllable. `client_name` is React-escaped (no XSS), but two notes:
1. The logo is a phishing surface — a malicious client can register as "Google" with a Google logo. Inherent to dynamic registration, but worth a conscious product decision.
2. No `onError` fallback on the ``, so a broken/blocked logo shows the browser's broken-image glyph. Consider hiding it on error.

apps/opik-frontend/src/shared/OAuthConsentPage/OAuthConsentPage.tsx
```diff
@@ -0,0 +1,258 @@
...
+          );
+          toast({
+            title: "Could not complete authorization",
+            description,
+            variant: "destructive",
+          });
+        },
+      },
+    );
+  };
```
awkoy, line 112: This block does four jobs (validate, build the request, navigate, extract+toast the error). Suggest pulling the pure parts out so they're unit-testable — which also covers the missing-test gap on the security-relevant request shape:

```ts
const buildConsentRequest = (
  params: ParsedParams,
  workspace: OAuthWorkspaceInfo,
  csrf: string,
): OAuthConsentRequest => ({
  client_id: params.client_id,
  redirect_uri: params.redirect_uri,
  code_challenge: params.code_challenge,
  code_challenge_method: params.code_challenge_method,
  resource: params.resource,
  state: params.state,
  workspace_id: workspace.id,
  workspace_name: workspace.name,
  csrf,
});

const extractErrorMessage = (error: unknown, fallback: string): string =>
  get(error, ["response", "data", "message"], fallback);
```

`handleAllow` then reads top-to-bottom and the request shape is testable without React.

apps/opik-frontend/src/shared/OAuthConsentPage/OAuthConsentPage.tsx
```diff
@@ -0,0 +1,258 @@
...
+                  <div key={w.id} className="flex items-center space-x-2">
+                    <RadioGroupItem value={w.id} id={`ws-${w.id}`} />
+                    <Label htmlFor={`ws-${w.id}`} className="font-normal">
+                      {w.name}
+                    </Label>
+                  </div>
+                ))}
+              </RadioGroup>
+            </div>
+          )}
```
awkoy: The `workspaces.length === 0 ? … : <RadioGroup>` block is the densest part of the render (~23 lines). Extracting a presentational `<WorkspacePicker workspaces value onChange />` shrinks the main `return` by about a third and makes the empty-state vs. picker logic independently scannable.

apps/opik-frontend/src/shared/OAuthConsentPage/OAuthConsentPage.tsx
```diff
@@ -0,0 +1,258 @@
...
+          >
+            Deny
+          </Button>
+          <Button
+            onClick={handleAllow}
+            disabled={
+              consent.isPending ||
+              workspaces.length === 0 ||
+              !effectiveWorkspaceId
+            }
```
awkoy: Minor readability: lift this compound condition into a named boolean next to `handleAllow`:

```ts
const canAllow =
  !consent.isPending && workspaces.length > 0 && !!effectiveWorkspaceId;
```

then `disabled={!canAllow}`. Reads better than a three-clause expression inline in JSX.

apps/opik-frontend/src/v1/router.tsx
```diff
@@ -130,6 +130,19 @@ const pairingRouteOssAlias = createRoute({
...
+// The browser lands here from opik-backend's GET /oauth/authorize 302 before any workspace
+// is selected, so both V1 and V2 routers must serve it — WorkspaceVersionGate picks the
+// app from cached defaults and the consent page itself is independent of either layout.
+const OAuthConsentPage = lazy(
+  () => import("@/shared/OAuthConsentPage/OAuthConsentPage"),
+);
+const oauthConsentRoute = createRoute({
+  getParentRoute: () => rootRoute,
+  path: "/oauth/consent",
+  component: OAuthConsentPage,
```
awkoy: This lazy import + `createRoute` block is byte-identical in `v2/router.tsx` (only the comment differs). Consider a shared factory in `src/shared/OAuthConsentPage/` so the path string and lazy import have a single source of truth:

```ts
export const createOAuthConsentRoute = (rootRoute: RootRoute) =>
  createRoute({
    getParentRoute: () => rootRoute,
    path: "/oauth/consent",
    component: lazy(() => import("@/shared/OAuthConsentPage/OAuthConsentPage")),
  });
```

Each router then becomes a single line.

awkoy (changes requested):
Code review — FE OAuth consent UI. Solid, well-documented work; most notes are inline. Cross-cutting items that don't map to a single line:

- **[security] Clickjacking:** consent screens are prime clickjacking targets. Please confirm the SPA `index.html` response for `/oauth/consent` carries `X-Frame-Options: DENY` (or CSP `frame-ancestors 'none'`).
- **[security] Deny open-redirect** — see inline on `denyAndRedirect`. The Deny path's safety rests entirely on the backend `GET /oauth/authorize/context` validating `redirect_uri` against the registered client.
- **nginx duplication:** the same three blocks are now copied across `nginx_default_local.conf`, `nginx_guardrails_local.conf`, and `nginx_local_be_local.conf` (+ mirrored in the prod configmap) with a "keep in sync" comment = 4 copies that will drift. Consider an `include` snippet. Separately, the `^~ /oauth/` block sets `Connection "upgrade"` unconditionally, but these endpoints never websocket-upgrade — harmless but technically incorrect; drop the upgrade/`Connection` lines unless this is deliberately mirroring an existing block.
- **Tests:** no coverage in this PR for the consent flow. At minimum `parseParams` + `denyAndRedirect` are pure and cheap to unit-test; given the open-redirect surface I'd add those here rather than defer to a follow-up.
- **Java test:** swapping `await().pollDelay(...).until(() -> true)` for `Thread.sleep` is the right call — the old form was an awaitility abuse and the new comment honestly documents the genuine wall-clock wait. 👍

The inline refactor suggestions on `OAuthConsentPage.tsx` and `v1/router.tsx` are non-blocking polish.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthValidateResource.java
```diff
@@ -0,0 +1,44 @@
...
+import lombok.RequiredArgsConstructor;
+import org.apache.commons.lang3.Strings;
+
+import static com.comet.opik.domain.mcpoauth.OAuthConstants.BEARER_PREFIX;
+import static com.comet.opik.domain.mcpoauth.OAuthConstants.TOKEN_TYPE_BEARER;
+
+@Path("/opik/auth-oauth")
+@Timed
+@RequiredArgsConstructor(onConstructor_ = @Inject)
+public class OAuthValidateResource {
```
andrescrz: This change seems stale.

andrescrz (dismissed):
Please double check the BE changes. The PR is supposed to be FE focus only, but it seems to contain BE stale changes. Tests seem to be failing due to the resource path conflict.

apps/opik-frontend/src/shared/OAuthConsentPage/OAuthConsentPage.tsx
LifeXplorer: Confirmed — the context endpoint validates `redirect_uri`. `GET /oauth/authorize/context` → `OAuthAuthorizationService#buildConsentContext` → `OAuthClientService#resolveForRedirect(clientId, redirectUri)`, which throws `BadRequestException("invalid redirect_uri")` when the URI is not registered for the client (it also rejects any URI containing a `#` fragment). So an unregistered `redirect_uri` returns 4xx, the consent card never renders, and the Deny button is never reachable — Deny can only ever navigate to a server-validated URI. I documented this invariant directly above `denyAndRedirect` in `helpers.ts` so the safety contract is explicit. (710d0847)

apps/opik-frontend/src/shared/OAuthConsentPage/OAuthConsentPage.tsx
LifeXplorer: Added an `onError` handler that hides the logo on load failure (`logoFailed` state), so a broken/blocked logo no longer shows the browser glyph. On point 1 (logo as a phishing surface): agreed it is inherent to RFC 7591 dynamic registration — leaving it as a conscious product decision for now rather than gating logos in this PR. (710d0847)

apps/opik-frontend/src/shared/OAuthConsentPage/OAuthConsentPage.tsx
LifeXplorer: Done — extracted `buildConsentRequest` and `extractErrorMessage` (plus `parseParams`/`denyAndRedirect`) into `helpers.ts`. `handleAllow` now reads top-to-bottom and the request shape is unit-tested without React in `helpers.test.ts`. (710d0847)

apps/opik-frontend/src/shared/OAuthConsentPage/OAuthConsentPage.tsx
LifeXplorer: Done — extracted a presentational `<WorkspacePicker workspaces value onChange />`. The empty-state vs picker logic lives there now and the main `return` is about a third shorter. (710d0847)

apps/opik-frontend/src/shared/OAuthConsentPage/OAuthConsentPage.tsx
LifeXplorer: Done — `const canAllow = !consent.isPending && workspaces.length > 0 && !!effectiveWorkspaceId;` next to `handleAllow`, with `disabled={!canAllow}` in the JSX. (710d0847)

apps/opik-frontend/src/v1/router.tsx
LifeXplorer: Done — added `createOAuthConsentRoute(rootRoute)` in `src/shared/OAuthConsentPage/createOAuthConsentRoute.ts`. The `/oauth/consent` path string and the lazy import now have a single source of truth, and each router calls it as a one-liner. (710d0847)

LifeXplorer:
@awkoy thanks for the thorough review — addressed the cross-cutting items in 710d0847:

- **[security] Clickjacking** — added `add_header X-Frame-Options "DENY" always;` to the `location = /oauth/consent` block in all three local nginx confs and the prod `configmap-frontend-nginx.yaml`, so the consent document cannot be framed.
- **[security] Deny open-redirect** — replied inline: confirmed `GET /oauth/authorize/context` validates `redirect_uri` via `OAuthClientService#resolveForRedirect` before the consent card renders, so Deny can only reach a server-validated URI. Invariant now documented in `helpers.ts`.
- **nginx `Connection "upgrade"`** — dropped the `proxy_http_version 1.1` / `Upgrade` / `Connection "upgrade"` lines from the `^~ /oauth/` proxy block in the local confs; these endpoints never websocket-upgrade. (The prod configmap’s `@mcp_oauth_backend` block never had them.)
- **nginx duplication** — kept the blocks inline rather than an `include`: the prod configmap is Go-templated (`{{ $rootPath }}`, the `@mcp_oauth_backend` named location, the well-known rewrite) and structurally diverges from the plain local confs, and a docker-compose `include` would need the snippet file mounted into each service — so a single shared include across all four is not clean. Left the "keep in sync" comments as the guard; happy to do a follow-up `include` scoped to the three local confs if you prefer.
- **Tests** — added `helpers.test.ts` (17 cases): `parseParams` (required-param gating, URL decoding, `response_type` not forwarded, optional `state`) and `denyAndRedirect` (error/state append, existing-query preservation, malformed-URI rejection), plus `buildConsentRequest` and `extractErrorMessage`.
- **Java test** — 👍 kept as `Thread.sleep` with the honest wall-clock comment.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthValidateResource.java
LifeXplorer: The PR was stale for a long time and the merge conflict wasn't fully resolved.
Updated PR to contain only FE changes. All BE changes were already merged in previous chunck, no more needed

apps/opik-frontend/src/shared/OAuthConsentPage/OAuthConsentPage.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+const denyAndRedirect = (
+  redirectUri: string,
+  state: string | null,
+): boolean => {
+  try {
+    const url = new URL(redirectUri);
+    url.searchParams.set("error", "access_denied");
+    if (state) url.searchParams.set("state", state);
+    window.location.href = url.toString();
```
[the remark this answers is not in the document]
LifeXplorer: Good catch — addressed in ffb3115. `denyAndRedirect` now rejects any non-`http(s)` scheme before assigning `window.location.href`, so a `javascript:`/`data:` URI can never execute on the consent origin. This is defense-in-depth on top of the server-side gate (`resolveForRedirect` validates `redirect_uri` against the registered client before the card renders) — the client guard removes any dependency on registration-time scheme validation. Added two tests covering `javascript:` and `data:`.

deployment/helm_chart/opik/templates/configmap-frontend-nginx.yaml
```diff
@@ -160,6 +160,8 @@ data:
         # React consent page renders from the SPA bundle while every other
         # /oauth/* path is rewritten to / and proxied to the backend.
         location = {{ $rootPath }}/oauth/consent {
+          # Consent screens are prime clickjacking targets; never allow framing.
+          add_header X-Frame-Options "DENY" always;
```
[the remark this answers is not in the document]
LifeXplorer: Correct, and this was a regression in my change — thanks. nginx `add_header` is all-or-nothing per context, so the per-location header dropped the inherited server-level CSP/HSTS/`extraServerHeaders` for `/oauth/consent`. Fixed in ffb3115 by removing the per-location `add_header`: the prod server block already sets CSP `frame-ancestors 'none'` (values.yaml), which the consent page now inherits — that is the modern superset of `X-Frame-Options: DENY`, so the route is frame-protected without losing the other headers. The local docker-compose confs have no CSP, so there I moved `X-Frame-Options: DENY` to the server level instead.

apps/opik-frontend/src/shared/OAuthConsentPage/OAuthConsentPage.tsx
```diff
@@ -0,0 +1,209 @@
...
+        />
+      </FullPage>
+    );
+  }
+
+  const data: OAuthAuthorizeContext = contextQuery.data;
+  const workspaces = data.workspaces ?? [];
+  const effectiveWorkspaceId = selectedWorkspaceId ?? workspaces[0]?.id ?? null;
+  const canAllow =
+    !consent.isPending && workspaces.length > 0 && !!effectiveWorkspaceId;
```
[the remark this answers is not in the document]
LifeXplorer: Leaving as-is (non-blocking). `effectiveWorkspaceId` falls back to `workspaces[0]` and the RadioGroup `value` is bound to it, so the first workspace is *visibly* pre-selected — Allow posts the option the user can see is checked, not a hidden default. Requiring an explicit pick would also force an extra click in the common single-workspace case (Allow disabled until the lone radio is clicked). Happy to switch to "require explicit selection only when there is more than one workspace" if you feel the grant action warrants it, but the current visible-default behaviour is intentional.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/ConsentRequest.java
```diff
@@ -0,0 +1,22 @@
...
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
+public record ConsentRequest(
+        @NotBlank String clientId,
+        @NotBlank String redirectUri,
+        @NotBlank String codeChallenge,
+        @NotBlank String codeChallengeMethod,
+        @NotBlank String resource,
+        String state,
+        String workspaceId,
+        String workspaceName,
```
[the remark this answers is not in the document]
LifeXplorer: Out of scope for this PR — `ConsentRequest.java` lives in the backend AS work that already merged to `main` (it is not part of this FE-only PR; the diff here is FE + nginx only). The blank-`workspaceName`/`workspaceId` validation is a reasonable backend hardening — worth a follow-up against `main`, but it cannot be addressed from this branch.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthAuthorizeResource.java
```diff
@@ -0,0 +1,219 @@
...
+
+        Cookie session = headers.getCookies().get(RequestContext.SESSION_COOKIE);
+        try {
+            authService.listEligibleWorkspaces(session);
+        } catch (ClientErrorException e) {
+            String rawQuery = uriInfo.getRequestUri().getRawQuery();
+            String authorizeUrl = config.getBaseUrl() + "/oauth/authorize" + (rawQuery == null ? "" : "?" + rawQuery);
+            return redirect(config.getBaseUrl() + "/login?returnTo=" + enc(authorizeUrl));
+        }
+
```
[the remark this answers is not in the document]
LifeXplorer: Out of scope for this PR — `OAuthAuthorizeResource.java` is backend code that already merged to `main` and is not part of this FE-only PR. The double `listEligibleWorkspaces` call is a legitimate backend optimization, but it belongs in a follow-up against `main` rather than this branch.

apps/opik-frontend/src/shared/OAuthConsentPage/helpers.ts
```diff
@@ -0,0 +1,80 @@
...
+//
+// Defense-in-depth: independent of the server-side check, only ever assign an http(s) URL
+// to window.location so a javascript:/data: scheme can never execute on this origin.
+export const denyAndRedirect = (
+  redirectUri: string,
+  state: string | null,
+): boolean => {
+  try {
+    const url = new URL(redirectUri);
+    if (url.protocol !== "http:" && url.protocol !== "https:") return false;
```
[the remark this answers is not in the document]
LifeXplorer: Keeping the `http(s)` gate (declining the custom-scheme expansion) — scoped to the initial target: OAuth for the **hosted MCP over HTTP transport**, with clients running on the user's machine.

In that model the auth-code callback lands on either a hosted `https://` endpoint or an **RFC 8252 §7.3 loopback** listener (`http://127.0.0.1|localhost:PORT`) — both `http(s)`, and loopback is exactly what the backend `isLoopbackMatch` special-cases. So `http(s)` covers the complete in-scope set; the `myapp://` custom-scheme deep-link is a different (desktop/mobile) client model that isn't part of this target.

Why the asymmetry is safe:
- **Allow is unaffected** — it navigates to the backend-built `redirect_to`, not an FE-constructed URL, so no scheme gate applies there.
- **Deny dead-ending is fail-safe** — an out-of-scope custom-scheme client would get the "redirect target is invalid" toast rather than being dropped somewhere; Allow still completes.
- A **default-deny allow-list** is a safer posture than a `javascript:`/`data:` deny-list and removes any reliance on registration-time scheme validation.

Documented at `denyAndRedirect` and added a loopback test (7c874d5). If custom-scheme native clients become a target, this is the single spot to widen.

awkoy (approved):

PR #6915, opened 2026-05-29
[OPIK-6743] [BE] feat: add BI event for prompt version creation

BorisTkachenko:
## Details
Emit a `prompt_version_created` analytics event from `PromptService` every time a prompt version is persisted, so prompt usage can be tracked alongside other BI events.

- Inject `AnalyticsService` into `PromptServiceImpl` and call `trackEvent("prompt_version_created", ...)` from the single `savePromptVersion` choke point, so all creation paths are covered (initial prompt creation, new commit, retried commit, and version restore).
- Thread `projectId` through `savePromptVersion` so it can be included in the event payload (falls back gracefully when unknown).
- Event properties: `prompt_version_id`, `prompt_id`, `workspace_id`, `user_name`, `version_type` (defaulting to `PROMPT_VERSION`), `date` (ISO-8601 `Instant.now()`), and `project_id` when available.

## Change checklist

- [ ] User facing
- [ ] Documentation update

## Issues
- Resolves #
- OPIK-6743

## Testing
Added `PromptVersionBiEventTest` integration test (Dropwizard + Testcontainers MySQL/ClickHouse/Redis + WireMock) with a mocked `AnalyticsService` to assert that creating a prompt version triggers `trackEvent("prompt_version_created", ...)` with the expected property set.

Manual:
1. Create a new prompt (which implicitly creates v1) and verify a `prompt_version_created` event is recorded.
2. Commit a new version on an existing prompt and verify a second event is recorded with the new `prompt_version_id`.
3. Restore an older version and verify the restore path also emits the event.

## Documentation
N/A — internal BI instrumentation only, no public API or user-facing surface changed.


apps/opik-backend/src/main/java/com/comet/opik/domain/PromptService.java
```diff
@@ -543,9 +547,26 @@ private PromptVersion savePromptVersion(String workspaceId, PromptVersion prompt
         log.info("Created Prompt version for prompt id '{}'", promptVersion.promptId());
 
         PromptVersion savedVersion = getById(workspaceId, promptVersion.id());
+        trackPromptVersionCreated(savedVersion, projectId, workspaceId);
         return savedVersion;
     }
 
+    private void trackPromptVersionCreated(PromptVersion promptVersion, UUID projectId, String workspaceId) {
```
thiagohora, line 556: Check Andrés guidelines. But in summary, BI events should be fire and forget, shouldn't block the main use case flow, and shouldn't cause side effects on it

thiagohora (changes requested):
A small follow up before unblocking it

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/PromptVersionBiEventTest.java
andrescrz, line 1: We generally don't need dedicated tests for BI. Please cover this in the regular tests, by asserting wire mock calls. 

There are some examples in the code base.

andrescrz (changes requested):
Please see comment.

Adding more and more integ tests unnecessarily slows down test time, at some point we won't be able to work with the generated tech debt.

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptService.java
```diff
@@ -543,9 +547,27 @@ private PromptVersion savePromptVersion(String workspaceId, PromptVersion prompt
         log.info("Created Prompt version for prompt id '{}'", promptVersion.promptId());
 
         PromptVersion savedVersion = getById(workspaceId, promptVersion.id());
+        Schedulers.boundedElastic()
```
andrescrz: Minor: feel free to encapsulate the boundedElastic in trackPromptVersionCreated, so we never forget to use it in the future, if this method is called from other entry points.

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptService.java
```diff
@@ -543,9 +547,27 @@ private PromptVersion savePromptVersion(String workspaceId, PromptVersion prompt
...
+                .schedule(() -> trackPromptVersionCreated(savedVersion, projectId, workspaceId));
         return savedVersion;
     }
 
+    private void trackPromptVersionCreated(PromptVersion promptVersion, UUID projectId, String workspaceId) {
```
andrescrz, line 556: Minor: I assume these args can't be null. Otherwise, handle the logic when building the event payload. e.g: fallback to empty string or similar. (or skip the event)

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptService.java
```diff
@@ -543,9 +547,27 @@ private PromptVersion savePromptVersion(String workspaceId, PromptVersion prompt
...
+        properties.put("workspace_id", workspaceId);
+        properties.put("user_name", promptVersion.createdBy());
+        properties.put("version_type",
+                Optional.ofNullable(promptVersion.versionType()).orElse(PromptVersionType.PROMPT_VERSION).getValue());
+        properties.put("date", Instant.now().toString());
+        if (projectId != null) {
+            properties.put("project_id", projectId.toString());
+        }
+
+        analyticsService.trackEvent("prompt_version_created", properties, promptVersion.createdBy());
```
andrescrz: Nit: feel free to use a static constant to DRY the event name.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/PromptResourceTest.java
```diff
@@ -2004,13 +2033,30 @@ void shouldCreatePromptVersion() {
...
+                            .withRequestBody(matchingJsonPath("$.event_type", equalTo(event))
+                                    .and(matchingJsonPath("$.event_properties.prompt_id",
+                                            equalTo(promptId.toString())))
+                                    .and(matchingJsonPath("$.event_properties.prompt_version_id",
+                                            equalTo(actualPromptVersion.id().toString())))
+                                    .and(matchingJsonPath("$.event_properties.workspace_id",
+                                            equalTo(WORKSPACE_ID)))
+                                    .and(matchingJsonPath("$.event_properties.user_name", equalTo(USER)))
+                                    .and(matchingJsonPath("$.event_properties.version_type",
+                                            equalTo("prompt_version"))))));
```
andrescrz: Minor: All this assertion logic seems mostly copied and pasted. I recommend adding an aux assertEvent method (parameterise as needed).

andrescrz (dismissed):
No blockers, just optional stuff.

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptService.java
```diff
@@ -543,9 +549,29 @@ private PromptVersion savePromptVersion(String workspaceId, PromptVersion prompt
...
         return savedVersion;
     }
 
+    private void trackPromptVersionCreated(PromptVersion promptVersion, UUID projectId, String workspaceId) {
+        Schedulers.boundedElastic().schedule(() -> {
+            Map<String, String> properties = new HashMap<>();
+            properties.put("prompt_version_id", Objects.toString(promptVersion.id(), ""));
+            properties.put("prompt_id", Objects.toString(promptVersion.promptId(), ""));
+            properties.put("workspace_id", Objects.toString(workspaceId, ""));
+            properties.put("user_name", Objects.toString(promptVersion.createdBy(), ""));
```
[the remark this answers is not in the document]
BorisTkachenko: No need

thiagohora (approved):

andrescrz (approved):

PR #6927, opened 2026-05-29
[OPIK-5988] [SDK] feat: add classification experiment scorers

### LINKED ISSUE #5988: [FR]: Support for classification metrics (Precision, Recall, F1) with dataset-level evaluation
### Proposal summary

Opik is great for LLM evaluation — especially with the LLM-as-a-Judge approach where everything is scored per sample and then aggregated.

But we ran into a limitation when trying to use Opik for something slightly different: classification tasks inside an LLM pipeline.

In these cases, per-row scoring just doesn’t work. For classification, you need to look at the whole dataset — that’s where Precision, Recall, and especially F1-score make sense.

Right now, there’s no native way to do that in Opik.
What we’d love to see is support for dataset-level classification metrics, including:
1) Precision / Recall / F1-score
2) Macro / Micro / Weighted averaging

### Motivation

1) What problem are you trying to solve?

We’re working on a modular RAG system where not everything is generation — some parts are basically classification tasks powered by LLMs.

For example:
- routing queries
- assigning labels
- filtering documents

To understand if these components work well, we need to measure things like:

- how often the model makes mistakes
- where it confuses classes
- how stable it is across datasets

And for that, F1-score is critical.

The current per-sample evaluation approach in Opik just doesn’t capture this.

2) How are you currently solving this problem?

Right now, we calculate classification metrics outside of Opik: notebooks / scripts

It works, but it’s not ideal:
- evaluation is split across tools
- results are not in the same place as Opik experiments
- extra overhead for the team

3) What are the benefits of this feature?

If Opik supported classification metrics, it would:

- Let us keep everything in one place (this is the biggest win for us)
- Remove the need for parallel tooling (Grafana, custom scripts, etc.)
- Make it much easier to compare experiments
- Better reflect real-world GenAI systems, which are rarely “just generation”
- Increase the practical value of Opik for production use

subash017:
## Details

Add sklearn-backed experiment-level classification scorers (F1, precision, recall), expose them under opik.evaluation.scoring_functions, and add e2e coverage for evaluate() aggregation.

## Change checklist

- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves #5988
- OPIK-NA

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: GitHub Copilot
- Model(s): GPT-5.2-Codex
- Scope: Drafted classification scoring functions + e2e test updates
- Human verification: Manually reviewed the changes

## Testing

Not run (requires e2e backend).

## Documentation

Not required.


jverre:
Hi @subash017 👋 — thank you for this contribution, and apologies: this PR was closed inadvertently on our side, not for any issue with the change itself. If it's still relevant, please feel free to open a fresh PR against the latest `main` and we'll be happy to review it. Thanks again for helping improve Opik — contributions like yours are always welcome! 🙏

PR #6930, opened 2026-05-31
[OPIK-6747] [BE] perf: defer wide-column reads past pagination in traces/spans list queries

YarivHashaiComet:
## Details

The traces/spans list-by-project query (`SELECT_BY_PROJECT_ID` in `TraceDAO`/`SpanDAO`) read the wide text columns (`input`, `output`, `metadata`, `input_slim`, `output_slim`) for the **entire matched set** in the `*_deduped` CTE and applied the page `LIMIT/OFFSET` only one CTE later, so a single filtered list request read millions of rows and multiple GiB just to return a ~100-row page — spiking ClickHouse memory, firing OOM alerts, and failing outright on the largest projects (error 241). This defers the wide-column read until after pagination so it only pays for the page.

- `*_deduped` / `*_final` now carry only the columns needed to filter, sort and dedup.
- New `page_wide` CTE re-reads `input`/`output`/`metadata`/`truncated_input`/`truncated_output` from the base table for **only the paginated id set** (`id IN (SELECT id FROM *_final)`), preserving ReplacingMergeTree dedup (same `ORDER BY` + `LIMIT 1 BY id`).
- `input_slim`/`output_slim` are dropped entirely (unused by this query's output).
- `sort_needs_wide` (`SortableFields.sortsByWideTextColumn`) keeps `input`/`output`/`metadata` in the pre-pagination scan only when the sort references them, so sorting by those columns still works; set by every renderer including the stream paths.
- `page_wide` and its join are omitted when `input`/`output`/`metadata` are all excluded.
- Measured on real data (read-only): ~99% peak-memory reduction on large projects (e.g. multi-GiB / OOM → a few hundred MiB), with byte-for-byte identical returned pages (verified via row checksums).

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- OPIK-6747

## AI-WATERMARK

AI-WATERMARK: yes

- If yes:
  - Tools: Claude Code
  - Model(s): Claude Opus 4.8
  - Scope: Root-cause investigation, ClickHouse query rewrite (TraceDAO/SpanDAO), shared sort helper, and test execution/validation.
  - Human verification: Author-reviewed; correctness validated against production data read-only (before/after row checksums match) and via the full integration test suites below.

## Testing

Integration tests (local, Docker + Testcontainers ClickHouse/MySQL/Redis/MinIO/ZK), all green:

- `mvn test -Dtest=TracesResourceTest`
- `mvn test -Dtest=SpansResourceTest`
- `mvn test -Dtest=GetTracesByProjectResourceTest` (filter / search / sort incl. sort-by-input/output/metadata / pagination / time filtering)
- `mvn test -Dtest=FindSpansResourceTest`
- `mvn spotless:check`

Scenarios validated: filtering, search_text, field exclusion (`exclude`), truncation, sorting by light / aggregate / wide (input/output/metadata) columns, offset pagination, streaming, and empty results. Result-equivalence and the memory reduction were additionally confirmed against real data by comparing `system.query_log` `memory_usage`/`read_bytes` and full-page row checksums for the old vs new query shape.

Not run: frontend/SDK suites (backend-only change). One pre-existing flaky test (`FilterTest#whenFilterSpanFeedbackScoresGreaterThan`, random feedback-score data) failed once and passed on isolated re-run; unrelated to this change (column projection cannot affect filter row counts).

## Documentation


apps/opik-backend/src/main/java/com/comet/opik/domain/SpanDAO.java
```diff
@@ -965,9 +965,7 @@ ORDER BY (workspace_id, project_id, entity_id, id) DESC, last_updated_at DESC
             <endif>
             <endif>, spans_deduped AS (
                 SELECT
-                      s.* <if(exclude_fields)>EXCEPT (<exclude_fields>) <endif>,
-                      truncated_input,
-                      truncated_output,
+                      s.* EXCEPT (input_slim, output_slim<if(!sort_needs_wide)>, input, output, metadata<endif>) <if(exclude_fields)>EXCEPT (<exclude_fields>) <endif>,
```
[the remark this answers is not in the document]
YarivHashaiComet: Thanks @baz — I investigated this and it looks like a false positive.

Trace list sorting only accepts the **base** sortable fields from `TraceSortingFactory.getSortableFields()` (`input`, `output`, `metadata` as whole columns) — there are **no** `input.*`/`output.*`/`metadata.*` dynamic-path sort fields (dynamic JSON paths are supported for *filtering*, not *sorting*). So a sort like `input.foo` is rejected and silently dropped; it never renders an `ORDER BY input[...]`, so there's no dropped column / 500 / broken pagination.

For the base-field sorts that *are* supported, `bindTemplateExcludeFieldVariables()` removes the sorted field from the exclude set (the `!sortingFields.contains(field)` filter), so the column the `ORDER BY` needs is never dropped by `EXCEPT (<exclude_fields>)`.

Verified end-to-end (HTTP 200, sort stays applied, no 500) and added a regression test covering exactly this: `GetTracesByProjectResourceTest#getTracesByProject__whenSortingByWideField_andExcludingSameField__thenSortStillApplied`.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -1225,9 +1225,7 @@ ORDER BY (workspace_id, dataset_id, id) DESC, last_updated_at DESC
             <endif>
             , traces_deduped AS (
                 SELECT
-                    t.* <if(exclude_fields)>EXCEPT (<exclude_fields>) <endif>,
-                    truncated_input,
-                    truncated_output,
+                    t.* EXCEPT (input_slim, output_slim<if(!sort_needs_wide)>, input, output, metadata<endif>) <if(exclude_fields)>EXCEPT (<exclude_fields>) <endif>,
```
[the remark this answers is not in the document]
YarivHashaiComet: Thanks @baz — I investigated this and it looks like a false positive.

Trace list sorting only accepts the **base** sortable fields from `TraceSortingFactory.getSortableFields()` (`input`, `output`, `metadata` as whole columns) — there are **no** `input.*`/`output.*`/`metadata.*` dynamic-path sort fields (dynamic JSON paths are supported for *filtering*, not *sorting*). So a sort like `input.foo` is rejected and silently dropped; it never renders an `ORDER BY input[...]`, so there's no dropped column / 500 / broken pagination.

For the base-field sorts that *are* supported, `bindTemplateExcludeFieldVariables()` removes the sorted field from the exclude set (the `!sortingFields.contains(field)` filter), so the column the `ORDER BY` needs is never dropped by `EXCEPT (<exclude_fields>)`.

Verified end-to-end (HTTP 200, sort stays applied, no 500) and added a regression test covering exactly this: `GetTracesByProjectResourceTest#getTracesByProject__whenSortingByWideField_andExcludingSameField__thenSortStillApplied`.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -1318,12 +1316,24 @@ ORDER BY (workspace_id, project_id, id) DESC, last_updated_at DESC
                 <endif>
                 ORDER BY <if(sort_fields)> <sort_fields>, <endif>(workspace_id, project_id, id) DESC, last_updated_at DESC
                 LIMIT :limit <if(offset)>OFFSET :offset <endif>
-            )
+            )<if(!exclude_input || !exclude_output || !exclude_metadata)>, page_wide AS (
+                SELECT
+                    id AS pw_id
+                    <if(!exclude_input)>, input AS pw_input, truncated_input AS pw_truncated_input<endif>
+                    <if(!exclude_output)>, output AS pw_output, truncated_output AS pw_truncated_output<endif>
```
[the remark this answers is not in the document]
YarivHashaiComet: Thanks @baz — I investigated this and it looks like a false positive.

Trace list sorting only accepts the **base** sortable fields from `TraceSortingFactory.getSortableFields()` (`input`, `output`, `metadata` as whole columns) — there are **no** `input.*`/`output.*`/`metadata.*` dynamic-path sort fields (dynamic JSON paths are supported for *filtering*, not *sorting*). So a sort like `input.foo` is rejected and silently dropped; it never renders an `ORDER BY input[...]`, so there's no dropped column / 500 / broken pagination.

For the base-field sorts that *are* supported, `bindTemplateExcludeFieldVariables()` removes the sorted field from the exclude set (the `!sortingFields.contains(field)` filter), so the column the `ORDER BY` needs is never dropped by `EXCEPT (<exclude_fields>)`.

Verified end-to-end (HTTP 200, sort stays applied, no 500) and added a regression test covering exactly this: `GetTracesByProjectResourceTest#getTracesByProject__whenSortingByWideField_andExcludingSameField__thenSortStillApplied`.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/GetTracesByProjectResourceTest.java
```diff
@@ -4444,6 +4445,63 @@ void getTracesByProject__whenSortingByValidFields__thenReturnTracesSorted(Compar
...
+                    .mapToObj(i -> factory.manufacturePojo(Trace.class).toBuilder()
+                            .projectId(null)
+                            .projectName(projectName)
+                            .input(JsonUtils.getJsonNodeFromString("{\"order\":\"%03d\"}".formatted(i)))
+                            .build())
+                    .toList();
+
+            traceResourceClient.batchCreateTraces(traces, apiKey, workspaceName);
+
+            // Fetch the page sorted by the dynamic path, optionally EXCLUDING the base "input" field.
```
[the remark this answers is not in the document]
YarivHashaiComet: Good catch on the coverage gap. I kept the *integration* test on the base `input` field deliberately: trace list sorting doesn't accept dotted paths — `TraceSortingFactory.getSortableFields()` whitelists only base fields (no `input.*`/`output.*`/`metadata.*`), so an `input.order` sort is rejected/silently dropped and never produces an applied dynamic sort to assert against (verified: ASC == DESC). To cover the `baseField()` dynamic-path branch directly, I added a fast unit test `SortableFieldsTest#sortsByWideTextColumn_detectsWideFields` asserting `input.order` / `output.model[0].year` / `metadata.a.b.c` are detected as wide. Pushed in 9e55ee9.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanDAO.java
```diff
@@ -2386,6 +2407,13 @@ private Publisher<? extends Result> find(Integer page, int size, SpanSearchCrite
 
             bindTemplateExcludeFieldVariables(spanSearchCriteria, template);
 
+            // When sorting by a wide text column, keep it in spans_deduped so spans_final's ORDER BY can resolve it
+            // (otherwise it is deferred to page_wide). Computed from the structured sort fields so every renderer of
+            // this template (including findSpanStream) sets it consistently.
+            if (SortableFields.sortsByWideTextColumn(spanSearchCriteria.sortingFields())) {
+                template.add("sort_needs_wide", true);
```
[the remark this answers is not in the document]
YarivHashaiComet: Done — extracted `FilterUtils.addSortNeedsWideFlag(template, sortingFields)` and replaced the four duplicated blocks (TraceDAO getTracesByProjectId/findTraceStream, SpanDAO find/findSpanStream) with it, so the flag logic stays in sync. Pushed in 9e55ee9.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanDAO.java
```diff
@@ -1011,12 +1010,28 @@ ORDER BY (workspace_id, project_id, id) DESC, last_updated_at DESC
...
+                    id AS pw_id
+                    <if(!exclude_input)>, input AS pw_input, truncated_input AS pw_truncated_input<endif>
+                    <if(!exclude_output)>, output AS pw_output, truncated_output AS pw_truncated_output<endif>
+                    <if(!exclude_metadata)>, metadata AS pw_metadata<endif>
+                FROM spans
+                WHERE workspace_id = :workspace_id
+                AND project_id = :project_id
+                AND id IN (SELECT id FROM spans_final)
+                <if(stream)>
+                ORDER BY (workspace_id, project_id, id) DESC, last_updated_at DESC
```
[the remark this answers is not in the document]
YarivHashaiComet: Good catch — implemented in d29676024. Restructured both DAOs so the page is computed once and never re-scanned:

- `page_ids` — the deduped/ordered/limited **ids only**. This is the sole consumer of `*_deduped`, so the heavy filtered scan runs exactly once.
- `page_wide` — reads the full rows (incl. wide text columns) for `id IN (SELECT id FROM page_ids)`, once.
- final SELECT reads `FROM page_wide` directly (no more `*_final` + `page_wide` join), so neither `page_ids` nor `*_deduped` is referenced twice.

Net effect: the previous double-inlining of `*_deduped` (the costly part on wide filter/sort paths) is gone — each CTE is referenced once. The stream vs non-stream dedup/sort keys are preserved on the spans side.

On sharing a helper between `TraceDAO` and `SpanDAO`: the two `page_wide`/`page_ids` blocks differ in dedup key (`spans` keys on `trace_id, parent_span_id` and has the stream/non-stream split) and in the surrounding StringTemplate conditionals, so I kept them inline rather than forcing a shared SQL fragment that would need per-DAO parameterization. Happy to extract one if you'd prefer.

Validated: GetTracesByProjectResourceTest (1705), TracesResourceTest, SpansResourceTest — all green.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanDAO.java
```diff
@@ -999,8 +998,11 @@ ORDER BY (workspace_id, project_id, id) DESC, last_updated_at DESC
...
-            ), spans_final AS (
-                SELECT sd.*
+            ), page_ids AS (
+                -- Phase 1: paginate on the light, deduped id+sort-key set only.
+                -- Selecting just `id` keeps this CTE (and the heavy spans_deduped scan it
+                -- inlines) referenced exactly once, so the filtered scan runs a single time.
+                SELECT sd.id
                 FROM spans_deduped sd
                 <if(sort_has_feedback_scores)>
                 LEFT JOIN feedback_scores_agg fsagg ON fsagg.entity_id = sd.id
```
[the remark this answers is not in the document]
YarivHashaiComet: Done in 8e308dbcc. Extracted the deferred wide-column re-read into `TemplateUtils.PAGE_WIDE_CTE`, parameterized by:
- `entity_table` — `traces` / `spans`
- `page_wide_order_by` — the ReplacingMergeTree dedup key, so the spans **stream vs non-stream** variant (`(ws, proj, id)` vs `(ws, proj, trace_id, parent_span_id, id)`) is now chosen at the call site

Both DAOs concatenate the shared fragment, so the correctness-critical wide-deferral logic (the `EXCEPT` list, `truncated_*`, `LIMIT 1 BY id`) lives in one place and can't drift between the two queries.

I scoped the extraction to `page_wide` only — the upstream `page_ids` CTE and the final `SELECT` genuinely diverge (different sort joins, agg joins, and output columns between traces and spans), so sharing those would be a leaky abstraction. Validated by GetTracesByProjectResourceTest + SpansResourceTest (all green; one unrelated random-boundary flake in a span-feedback-scores filter test passed on rerun).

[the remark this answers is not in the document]
andrescrz: I will review the PR in depth in a few minutes. But let's avoid concating SQL queries such as `TemplateUtils.PAGE_WIDE_CTE `. It's a really bad practice. I will open an internal conversation to prevent Baz from making these suggestions. This is creating a lot of noise.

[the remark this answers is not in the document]
YarivHashaiComet: Makes sense — reverted the SQL-concatenation helper. The branch is back at `d29676024`, with `page_wide` inline in each DAO (the version you'll want to review). `TemplateUtils.PAGE_WIDE_CTE` is gone entirely.

The substantive fix stands: the list query now paginates on a light `page_ids` CTE and re-reads wide columns once via `page_wide` (no double scan, no whole-set wide materialization). Thanks — ready for your in-depth review.

apps/opik-backend/src/main/java/com/comet/opik/api/sorting/SortableFields.java
```diff
@@ -59,4 +63,28 @@ public class SortableFields {
     public static final String TEMPLATE = "template";
     public static final String CHANGE_DESCRIPTION = "change_description";
     public static final String ENVIRONMENT = "environment";
+
+    // Wide text columns deferred past pagination by the TraceDAO/SpanDAO page_wide optimization.
```
andrescrz: Nit: these comments better as javadoc.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanDAO.java
```diff
@@ -965,9 +966,7 @@ ORDER BY (workspace_id, project_id, entity_id, id) DESC, last_updated_at DESC
             <endif>
             <endif>, spans_deduped AS (
                 SELECT
-                      s.* <if(exclude_fields)>EXCEPT (<exclude_fields>) <endif>,
-                      truncated_input,
-                      truncated_output,
+                      s.* EXCEPT (input_slim, output_slim<if(!sort_needs_wide)>, input, output, metadata<endif>) <if(exclude_fields)>EXCEPT (<exclude_fields>) <endif>,
```
andrescrz: Minor: Please double check that two `EXCEPT` statements are accepted. There should be a test for `exclude_fields` enabled regressing this, but please let me know or add coverage.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanDAO.java
```diff
@@ -999,8 +998,11 @@ ORDER BY (workspace_id, project_id, id) DESC, last_updated_at DESC
                 ORDER BY (workspace_id, project_id, trace_id, parent_span_id, id) DESC, last_updated_at DESC
                 <endif>
                 LIMIT 1 BY id
-            ), spans_final AS (
-                SELECT sd.*
+            ), page_ids AS (
+                -- Phase 1: paginate on the light, deduped id+sort-key set only.
+                -- Selecting just `id` keeps this CTE (and the heavy spans_deduped scan it
+                -- inlines) referenced exactly once, so the filtered scan runs a single time.
```
andrescrz: Minor: I believe it's preferable to add these comments a top level javadoc on this query constant.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanDAO.java
```diff
@@ -1011,9 +1013,29 @@ ORDER BY (workspace_id, project_id, id) DESC, last_updated_at DESC
...
+                    duration
+                FROM spans s
+                WHERE workspace_id = :workspace_id
+                AND project_id = :project_id
+                AND id IN (SELECT id FROM page_ids)
+                <if(stream)>
+                ORDER BY (workspace_id, project_id, id) DESC, last_updated_at DESC
+                <else>
+                ORDER BY (workspace_id, project_id, trace_id, parent_span_id, id) DESC, last_updated_at DESC
+                <endif>
```
andrescrz, line 1039: For the case of the non-streaming endpoints, I believe here you need to consider the possibility of being sorting by custom fields (sort_fields). Otherwise, you risk to not to get the right results in your pre-filter when using such feature.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -1318,12 +1320,28 @@ ORDER BY (workspace_id, project_id, id) DESC, last_updated_at DESC
...
+                    t.* EXCEPT (input_slim, output_slim<if(!sort_needs_wide)><if(exclude_input)>, input<endif><if(exclude_output)>, output<endif><if(exclude_metadata)>, metadata<endif><endif>)<if(exclude_fields)> EXCEPT (<exclude_fields>)<endif>,
+                    <if(truncate)><if(!exclude_input)>truncated_input,<endif><if(!exclude_output)>truncated_output,<endif><endif>
+                    input_length,
+                    output_length,
+                    duration
+                FROM traces t
+                WHERE workspace_id = :workspace_id
+                AND project_id = :project_id
+                AND id IN (SELECT id FROM page_ids)
+                ORDER BY (workspace_id, project_id, id) DESC, last_updated_at DESC
```
andrescrz, line 1342: Same case about custom sorting here, but for traces.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/FilterUtils.java
```diff
@@ -33,6 +35,17 @@ public class FilterUtils {
     public static final int UUID_POOL_MULTIPLIER = 2;
     private static final String LOG_COMMENT = "<query_name>:<workspace_id>:<user_name>:<details>";
 
+    /**
+     * Sets the {@code sort_needs_wide} template flag when the sort targets a wide text column (input/output/metadata),
+     * so the *_deduped CTE keeps those columns for the ORDER BY instead of deferring them to page_wide. Shared by all
+     * renderers of SELECT_BY_PROJECT_ID (paginated and stream paths in TraceDAO/SpanDAO) to keep the logic in sync.
+     */
+    public static void addSortNeedsWideFlag(ST template, List<SortingField> sortingFields) {
```
andrescrz: Please add `@NonNull` validation to `template` argument, as this is a public utils method.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/GetTracesByProjectResourceTest.java
```diff
@@ -4444,6 +4445,63 @@ void getTracesByProject__whenSortingByValidFields__thenReturnTracesSorted(Compar
...
+                        StandardCharsets.UTF_8));
+                if (excludeInput) {
+                    queryParams.put("exclude", toURLEncodedQueryParam(List.of(Trace.TraceField.INPUT)));
+                }
+
+                try (var response = traceResourceClient.callGetTracesWithQueryParams(apiKey, workspaceName,
+                        queryParams)) {
+                    assertThat(response.getStatusInfo().getStatusCode()).isEqualTo(HttpStatus.SC_OK);
+                    return response.readEntity(Trace.TracePage.class).content().stream().map(Trace::id).toList();
+                }
```
andrescrz: Minor: this logic seems to belong to an overloaded method inside the `traceResourceClient` that would delegate to `callGetTracesWithQueryParams`

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/GetTracesByProjectResourceTest.java
```diff
@@ -4444,6 +4445,63 @@ void getTracesByProject__whenSortingByValidFields__thenReturnTracesSorted(Compar
...
+            assertThat(descNoExclude).as("wide-field sort works without exclude").isEqualTo(ascNoExclude.reversed());
+            assertThat(descNoExclude).isNotEqualTo(ascNoExclude);
+
+            // Case under test (the area Baz flagged): excluding the same wide field must not drop the sort or error.
+            var ascExclude = sortedIds.apply(Direction.ASC, true);
+            var descExclude = sortedIds.apply(Direction.DESC, true);
+            assertThat(descExclude).as("wide-field sort still applied when that field is excluded")
+                    .isEqualTo(ascExclude.reversed());
+            assertThat(descExclude).isNotEqualTo(ascExclude);
+            assertThat(ascExclude).as("sort order consistent with/without exclude").isEqualTo(ascNoExclude);
```
andrescrz: This test is missing the most important part: proper assertions. You need to assert the whole page content, not only the IDs, to make sure the new functionality works as expected and we can regress it in the future. The current approach based on IDs only is too weak, to ensure data integrity. 

Please make sure you re-use the existing assertion mechanisms, that cover every field in the page.

apps/opik-backend/src/test/java/com/comet/opik/api/sorting/SortableFieldsTest.java
```diff
@@ -0,0 +1,38 @@
+package com.comet.opik.api.sorting;
+
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class SortableFieldsTest {
```
andrescrz: Nit: as general convention of the service, we aren't adding unit tests for util methods, if having test coverage from the black box tests. Feel free to remove this if low value.

apps/opik-backend/src/test/java/com/comet/opik/api/sorting/SortableFieldsTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Test
+    @DisplayName("sortsByWideTextColumn detects base and dynamic-path wide text fields (input/output/metadata)")
+    void sortsByWideTextColumn_detectsWideFields() {
```
andrescrz: Minor: this test would be better shaped as a parameterised tests by providing the input and the expected output.

andrescrz (commented):
This generally looks in the good direction.

The main case to check on my end is the sorting case, both for spans and traces. This is to validate that this approach works for all existing logic/features.

Regarding tests, they need to improve (assertions) to ensure integrity of data.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanDAO.java
andrescrz: Not a big deal, but these inlined comments are still found in this PR across multiple occurrences.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanDAO.java
andrescrz: Can you please clarify what was done in relation to this? I don't see any related change.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
andrescrz: Same here, what was the relevant change?

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/FindSpansResourceTest.java
```diff
@@ -3943,6 +3943,157 @@ static Stream<Arguments> whenSortingByValidFields__thenReturnTracesSorted() {
...
+        void whenSortingAndExcludingField__thenReturnSortedExcludingField(Comparator<Span> comparator,
+                SortingField sorting, Span.SpanField excludeField) {
+            var workspaceName = RandomStringUtils.secure().nextAlphanumeric(10);
+            var workspaceId = UUID.randomUUID().toString();
+            var apiKey = UUID.randomUUID().toString();
+
+            mockTargetWorkspace(apiKey, workspaceName, workspaceId);
+
+            var projectName = RandomStringUtils.secure().nextAlphanumeric(10);
+            AtomicInteger index = new AtomicInteger(0);
```
andrescrz: Minor: Why atomic fields are used in these tests? This is a local variable, which isn't shared in a multi-threaded environment.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/FindSpansResourceTest.java
```diff
@@ -3943,6 +3943,157 @@ static Stream<Arguments> whenSortingByValidFields__thenReturnTracesSorted() {
...
+            var expectedSpans = spans.stream()
+                    .sorted(comparator)
+                    .map(span -> SpanAssertions.EXCLUDE_FUNCTIONS.get(excludeField).apply(span))
+                    .toList();
+
+            getAndAssertPage(workspaceName, projectName, List.of(), spans, expectedSpans, List.of(), apiKey,
+                    List.of(sorting), List.of(excludeField));
+        }
+
+        static Stream<Arguments> whenSortingAndExcludingField__thenReturnSortedExcludingField() {
```
andrescrz: Minor: I'm not sure what's the difference between the existing test for this endpoint and this one. The payload seems the same, just changes which sortable field is picked up and the direction. 

This likely overlaps with existing tests, except maybe for some of the parameterised tests.

I recommend double checking: consolidating the gaps, and removing the overlaps.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/FindSpansResourceTest.java
```diff
@@ -3943,6 +3943,157 @@ static Stream<Arguments> whenSortingByValidFields__thenReturnTracesSorted() {
...
+                                .feedbackScores(null)
+                                .comments(null)
+                                .projectName(projectName)
+                                .endTime(base.startTime().plus(randomNumber(), ChronoUnit.MILLIS))
+                                .totalEstimatedCost(null)
+                                .usage(Map.of("total_tokens", RandomUtils.secure().randomInt()))
+                                .createdAt(Instant.now().plusMillis(index.getAndIncrement()))
+                                .lastUpdatedAt(Instant.now().plusMillis(index.getAndIncrement()));
+                        if (i % 3 == 0) {
+                            builder.input(null).errorInfo(null);
```
andrescrz: Minor: I'm not sure why null values are relevant for this sort of test. What you really want to test is the combination of excluded data with sorting by that data.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/FindSpansResourceTest.java
```diff
@@ -3943,6 +3943,157 @@ static Stream<Arguments> whenSortingByValidFields__thenReturnTracesSorted() {
...
+            int pageSize = 2;
+            var pagedIds = new ArrayList<UUID>();
+            int pages = (total + pageSize - 1) / pageSize;
+            for (int p = 1; p <= pages; p++) {
+                var page = spanResourceClient.findSpans(workspaceName, apiKey, projectName, null, p, pageSize, null,
+                        null, List.of(), List.of(sorting), List.of(excludeField));
+                assertThat(page.total()).isEqualTo(total);
+                pagedIds.addAll(page.content().stream().map(Span::id).toList());
+            }
+            assertThat(pagedIds).as("pagination consistent with the full sorted order").isEqualTo(fullOrderIds);
```
andrescrz: Minor: related to my previous comment, not sure why only asserting ids, instead of the full actual page vs the full expected page.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/GetTracesByProjectResourceTest.java
andrescrz, line 1: Exactly same comments for spans as for traces tests.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/GetTracesByProjectResourceTest.java
```diff
@@ -5063,6 +5250,21 @@ private static int randomNumber(int minValue, int maxValue) {
         return PodamUtils.getIntegerInRange(minValue, maxValue);
     }
 
+    private Trace.TracePage fetchTracePage(String apiKey, String workspaceName, String projectName,
```
andrescrz: Nit: better placed in `traceResourceClient` class.

andrescrz (commented):
Let me double check the sorting, still not sure if this works as expected if sorting by custom fields when wide column pagination is applied.

andrescrz (commented):
# Wrap-up review — consolidated findings

This is a roll-up of my review across the PR, organized by priority. References to existing inline comment threads are included where applicable.

## Implementation: what's already in good shape (don't touch)

Before the action items, the things I verified are correct so they don't get inadvertently changed:

- **Two-phase shape** (`*_deduped` → `page_ids` → `page_wide`) is structurally sound. Final-SELECT ORDER BY carries `sort_fields`; `page_wide`'s ORDER BY exists purely for ReplacingMergeTree dedup.
- **`sort_needs_wide`** flag is wired into all 4 entry points (`SpanDAO.find`, `SpanDAO.findSpanStream`, `TraceDAO.getTracesByProjectId`, `TraceDAO.findTraceStream`).
- **Filtering** semantics are preserved — filters apply to base-table WHERE, EXCEPT only affects projection.
- **Dedup** is consistent — `*_deduped` and `page_wide` use the same ORDER BY + `LIMIT 1 BY id`, so they pick the same row version under ReplacingMergeTree.
- **Custom sorts** (input/output/metadata, dynamic Map fields like `usage.X` / `feedback_scores.X`, aggregates like `total_estimated_cost`, `experiment_id`) all resolve correctly through `page_ids` joins.
- **Filter-then-dedup stale-row semantics** is pre-existing and not worsened by this PR.

---

## MUST HAVE — block merge

### M1. Author replies to my open `page_wide` / `sort_fields` questions *(process)*
**Comments:** #discussion_r3354797076 (SpanDAO:1045), #discussion_r3354812830 (TraceDAO:1348), #discussion_r3356745668, #discussion_r3356756874.

The implementation is correct (`page_wide`'s ORDER BY intentionally omits `sort_fields` because the final SELECT does the actual ordering, and the new javadoc on `SELECT_BY_PROJECT_ID` covers this). But my follow-ups ("Can you clarify what was done... I don't see any related change") are still hanging. Please **reply explicitly** — point at the javadoc, confirm "no code change needed, here's why." Don't want to merge with open reviewer questions.

### M2. Regression test for the OPIK-6747 production scenario *(coverage)*
The ticket reproduction is *filtered* traces-list calls hitting 5–32 GiB. Every new test in this PR runs **with no filter**; every existing filter test runs **without exclude and without wide-col sort**. The exact prod shape that motivated this PR is **not regression-pinned**. Required:

- Insert >page-size matching rows + non-matching rows in a project
- Apply `input CONTAINS '<term>'`, sort `input ASC`, exclude `output`, walk all pages
- Assert: paged-walk content == full-page reference == Java-side filter+sort+null-output expectation
- Totals match the filtered count

Mirror for spans. Without this test, the PR's perf+correctness claim for the bug being fixed is unverified by CI.

### M3. `truncate=true` × exclude wide cols × sort wide-col *(coverage)*
The most fragile string-template branch in the PR:
```
<if(truncate)><if(!exclude_input)>truncated_input,<endif><if(!exclude_output)>truncated_output,<endif><endif>
```
combined with the final SELECT's `EXCEPT (..., truncated_input, truncated_output)` and the conditional `replaceRegexpAll(t.truncated_input,...)` re-add. Existing `findWithImageTruncation` covers only `truncate=true` with no exclude. A future ClickHouse version that becomes strict about `EXCEPT` on non-existent columns would silently break this without any test catching it. Add:
- `truncate=true`, sort `input ASC`, exclude `input`
- `truncate=true`, sort `input ASC`, exclude `output`
- `truncate=true`, exclude `[input, output]`

Each asserts image markers correctly redacted on non-excluded wide fields and response parses cleanly.

### M4. Dedup-after-update with sort + exclude on wide cols *(coverage)*
`whenSortingByName__afterUpdate__thenReturnLatestVersion` (traces) and `whenSortingByUsageTotalTokens__afterUpdate__thenReturnLatestVersion` (spans) exist but neither combines with `exclude`. The two-phase rewrite's load-bearing invariant is that `*_deduped` and `page_wide` pick the **same row version** for each id. That invariant isn't directly stressed by any test. Required:
- Insert span/trace with `input='old'`
- Update with `input='new'` (higher `last_updated_at`)
- Query sorted by `input DESC`, excluding `output`
- Assert: returned `input == 'new'`, sort position reflects `'new'`, `output` is null

If `page_wide.LIMIT 1 BY id` ever picks a different version than `*_deduped.LIMIT 1 BY id`, this catches it.

---

## SHOULD HAVE — important quality / coverage

### S1. Remove inline SQL comments from `SELECT_BY_PROJECT_ID` template
**Comments:** #discussion_r3354747866, #discussion_r3356704638.

The top-level javadoc on the constant was added (good) but the inline `-- Phase 1: ...` / `-- Phase 2: ...` comments are still inside the template strings (SpanDAO:1013-1015, 1028-1030; TraceDAO:1318-1320, 1335-1337). Having both is the worst of both worlds — drift risk + duplicated noise in rendered SQL logs. Drop the inline comments; the javadoc on the constant is sufficient.

### S2. Move `fetchTracePage` helper to `TraceResourceClient`
**Comments:** #discussion_r3354830146, #discussion_r3356813519.

`GetTracesByProjectResourceTest.fetchTracePage` (line 5253) wraps `callGetTracesWithQueryParams` with sort+exclude params — exactly what a client helper should do. Same point for the inline `try (response = ...)` lambda block at line 4488. Mirror for the span side. Cheap fix that pays back every future test in this area.

### S3. Restructure new tests to eliminate overlap and use proper parameterization
**Comments:** #discussion_r3356783105 (overlap), #discussion_r3356792604 (null-data scope).

The new `whenSortingAndExcludingField` (5 hand-picked cases) overlaps significantly with two existing parameterized tests:
- `whenSortingByValidFields` (~30 cases including all wide-col sort directions) — covers the sort axis
- `whenExcludeParamIdDefined` (`@EnumSource` over all fields) — covers the exclude axis

The individual axes are *more thoroughly* covered by existing tests than by the 5 hand-picked cases. Only the **combination** is new coverage. Setup is ~30 lines of copy-paste from `whenSortingByValidFields`.

The new `whenPaginatingSortedWithNullData` is `@Test` (not parameterized) and hardcodes `(sorting, excludeField, total, pageSize, nullPattern)`. The null/sparse data mixing dilutes the test's purpose.

**Recommended changes:**

a. **Fold `whenSortingAndExcludingField` into `whenSortingByValidFields`** by adding an optional `excludeField` parameter. Existing ~30 cases pass `null`; append a handful of `(comparator, sorting, excludeField)` rows targeting the new wide-col combinations. Net: delete ~150 lines of duplicated setup, retain all new coverage.

b. **Parameterize `whenPaginatingSortedWithNullData`** with `(sorting, excludeField, total, pageSize)`. Drop the null mixing (or move it to a separate small test). The parameterized body becomes the natural place to add the M2 filter+pagination case as additional argument rows.

This single restructure resolves the overlap and null-scope comments and creates the structural slot for M2 / S5 / S6 / S7.

### S4. Defensively dedup `EXCEPT` column lists in Java
See inline comment on the final SELECT. The rendered SQL produces `EXCEPT (input_slim, output_slim, output) EXCEPT (output)` (duplicate `output`) and `EXCEPT (... truncated_input, truncated_output)` referencing columns not present in `page_wide` when `truncate=false`. ClickHouse 25.x accepts both silently — verified by running tests — but this is fragile across CH upgrades and the symptom (silent CH error → broken Traces list) would be high-blast-radius. Either:
- Build the EXCEPT list in Java with a `Set` to dedup before rendering, or
- Make the truncated_input/output references in the final SELECT conditional on `<if(truncate)>`

### S5. Coverage: `search_text` + sort by wide + exclude wide
`search_text` references input/output/metadata in the WHERE LIKE predicate. The new EXCEPT changes mean those columns are absent from `*_deduped`'s projection (though still accessible in WHERE). Existing `searchTraces__whenSearchByField` covers search alone; the combination with sort + exclude is unexercised. Add one parameterized row to the restructured test from S3.

### S6. Coverage: sort by `usage.X` + exclude `usage`
Exercises the dynamic-Map sort path × the "exclude is filtered out of `exclude_fields` when used for sorting" path. The existing `whenSortingByUsageTotalTokens__afterUpdate` doesn't combine with exclude. Add as a parameterized row.

### S7. Coverage: sort by `feedback_scores.X` + exclude `feedback_scores`
Three special paths fire simultaneously here: (1) `sort_has_feedback_scores=true` LEFT JOIN in `page_ids`, (2) the `exclude_feedback_scores` branch gated by `noneMatch(isFeedBackScoresField)`, (3) `feedback_scores` token entering `exclude_fields`. Each is tested in isolation; the combination is not. Add as a parameterized row.

---

## NICE TO HAVE — nits / cleanups / optional coverage

### N1. Multiple wide-column exclusions at once
The new tests exclude one wide column per case. `exclude=[input, output, metadata]` with a non-wide sort triggers the triple-duplicate EXCEPT and the all-three-skip-re-add paths. Single new parameterized row.

### N2. Sort by aggregate field + exclude wide col *(traces)*
Sort by `total_estimated_cost` / `span_count` / `llm_span_count` triggers `sort_has_span_statistics=true` → `LEFT JOIN spans_agg` in `page_ids`. Combined with `exclude=input` is a real-world query shape, not currently tested.

### N3. Sort by `experiment_id` + exclude wide *(traces only)*
Triggers `sort_has_experiment=true`. Existing `TracesResourceTest:7310` covers the sort alone but not with exclude. Lower priority than N2.

### N4. `AtomicInteger` in single-threaded test setup
**Comment:** #discussion_r3356767181.

Pre-existing pattern copied from `whenSortingByValidFields` line 3802. If S3 (folding tests) is done, this disappears naturally. Otherwise: either remove from both occurrences (consistent) or leave both — don't fix only the new occurrence.

### N5. Pagination walk per-page content assertions
**Comment:** #discussion_r3356804028.

Current test asserts full content on the single full-page fetch + ID equality on the walk. Rationale ("walk's job is to prove pagination consistency, content is proved by the full-page fetch") is defensible since both code paths hit the same SQL with different LIMIT/OFFSET. Acceptable as-is.

### N6. Code comment on `bindTemplateExcludeFieldVariables` behavior
The "exclude is silently dropped from `exclude_fields` when used for sorting, but the Java mapper still nullifies it at `getValue`" behavior is subtle and non-obvious. Pre-existing, but the PR's javadoc-heavy follow-up commit is a reasonable moment to add a 2-line comment near `.filter(field -> !sortingFields.contains(field))` explaining the SQL/mapper division of labor.

---

## Summary count

| Category | Count | Description |
|---|---|---|
| Must have | **4** | 1 process reply + 3 high-value regression tests |
| Should have | **7** | Code cleanups + test restructure + targeted coverage |
| Nice to have | **6** | Nits and optional coverage |

**Critical path to merge:** M1 (reply) + M2/M3/M4 (three new tests). Everything else can land in follow-ups, but those four directly tie back to unresolved review comments or to the OPIK-6747 production scenario itself.

**S3 (test restructure) is the highest-leverage SHOULD HAVE** — it resolves comments #discussion_r3356783105 + #discussion_r3356792604 + creates the structural slot to add M2 / S5 / S6 / S7 as parameterized rows rather than as standalone tests. Doing S3 first makes the rest cheaper.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanDAO.java
```diff
@@ -1011,9 +1024,29 @@ ORDER BY (workspace_id, project_id, id) DESC, last_updated_at DESC
...
+                LIMIT 1 BY id
             )
             SELECT
-                s.* <if(exclude_fields)>EXCEPT (<exclude_fields>, input, output, metadata) <else> EXCEPT (input, output, metadata)<endif>
+                s.* <if(exclude_fields)>EXCEPT (<exclude_fields>, input, output, metadata, truncated_input, truncated_output) <else> EXCEPT (input, output, metadata, truncated_input, truncated_output)<endif>
```
andrescrz: **S4 (SHOULD HAVE — see top-level summary): defensively dedup `EXCEPT` column lists.**

This final SELECT's `EXCEPT (input, output, metadata, truncated_input, truncated_output)` references `truncated_input`/`truncated_output` even when `truncate=false` — in which case `page_wide` (line 1027-1046) did not include those columns. So we're EXCEPT'ing columns that don't exist in the source CTE.

Additionally, in `page_wide` itself (line 1032) and `spans_deduped` (line 980), when `sort_needs_wide=false` and the same wide column also appears in `exclude_fields`, the rendered SQL produces e.g. `EXCEPT (input_slim, output_slim, output) EXCEPT (output)` — duplicate `output` across the two EXCEPT clauses.

ClickHouse 25.x accepts both silently (verified by running the tests), but this is fragile across CH upgrades — a stricter EXCEPT implementation would surface as a query-compile error breaking the entire Traces/Spans list endpoint. Either:

- Build the EXCEPT list in Java with a `Set` to dedup before rendering, or
- Make the `truncated_input`/`truncated_output` references conditional on `<if(truncate)>` in this final SELECT.

Same applies to `TraceDAO.java` line 1352.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -1318,12 +1331,28 @@ ORDER BY (workspace_id, project_id, id) DESC, last_updated_at DESC
...
+                LIMIT 1 BY id
             )
             SELECT
-                  t.* <if(exclude_fields)>EXCEPT (<exclude_fields>, input, output, metadata) <else> EXCEPT (input, output, metadata)<endif>
-                  <if(!exclude_input)>, <if(truncate)> replaceRegexpAll(truncated_input, '<truncate>', '"[image]"') as input <else> input as input <endif><endif>
-                  <if(!exclude_output)>, <if(truncate)> replaceRegexpAll(truncated_output, '<truncate>', '"[image]"') as output <else> output as output <endif><endif>
-                  <if(!exclude_metadata)>, <if(truncate)> replaceRegexpAll(metadata, '<truncate>', '"[image]"') as metadata <else> metadata <endif><endif>
+                  t.* <if(exclude_fields)>EXCEPT (<exclude_fields>, input, output, metadata, truncated_input, truncated_output) <else> EXCEPT (input, output, metadata, truncated_input, truncated_output)<endif>
```
andrescrz: **S4 (SHOULD HAVE — see top-level summary): defensively dedup `EXCEPT` column lists.**

Mirror of the same concern on `SpanDAO.java` line 1049. This final SELECT references `truncated_input`/`truncated_output` in EXCEPT even when `truncate=false` and `page_wide` (line 1334-1349) did not include them; additionally, the `EXCEPT` lists in `traces_deduped` (line 1240) and `page_wide` (line 1339) can produce duplicate column names when `sort_needs_wide=false` and the same wide column also appears in `exclude_fields`.

CH 25.x accepts both silently but it's CH-upgrade-fragile. See the SpanDAO comment for the recommended fix.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanDAO.java
YarivHashaiComet: No functional change was needed here — the pagination pre-filter (`page_ids`) already renders `<sort_fields>` in its `ORDER BY` (SpanDAO.java:1021), and the final `SELECT` re-applies it. `page_wide` intentionally omits the sort key because it only re-reads the page's rows by id (`id IN (page_ids)` + `LIMIT 1 BY id`); ordering is restored by the final `SELECT`.

What changed in relation to this is **test coverage** that makes the guarantee explicit and regression-proof: a paginated test `whenPaginatingSortedAndExcludingField` (spans + traces) that sorts by a field, walks every page, and asserts each page's **full content** equals the expected sorted slice — so it only passes if the pre-filter pages in the custom sort order. Confirmed it bites: dropping `<sort_fields>` from `page_ids` makes it fail (hand mutation), and verified end-to-end against a live ClickHouse.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
YarivHashaiComet: Same as the spans thread: no functional change was needed — the trace pagination pre-filter (`page_ids`) already renders `<sort_fields>` in its `ORDER BY` (TraceDAO.java:1329), and the final `SELECT` re-applies it. `page_wide` deliberately omits the sort key (it only re-reads the page's rows via `id IN (page_ids)` + `LIMIT 1 BY id`; the final `SELECT` restores order).

The relevant change here is **test coverage**, not the query: `getTracesByProject__whenPaginatingSortedAndExcludingField` sorts by a field, walks every page, and asserts each page's **full content** equals the expected sorted slice — so it only passes if the pre-filter pages in the custom sort order. Hand-mutation confirmed it fails if `<sort_fields>` is dropped from `page_ids`, and I verified the behaviour against a live ClickHouse.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanDAO.java
YarivHashaiComet: Fixed in `1ef2e7c43b` — deduped the `EXCEPT` lists so we no longer rely on ClickHouse's lenient handling:

- the deferred wide-column drops in `spans_deduped` are now gated on `<if(!exclude_*)>`, so a wide column can't land in both that `EXCEPT` and the `exclude_fields` `EXCEPT`;
- `page_wide`'s redundant `<if(exclude_*)>` wide-column drops are removed (`exclude_fields` already covers them) → just `EXCEPT (input_slim, output_slim)` + `exclude_fields`;
- in the final `SELECT`, `input/output/metadata` are gated on `<if(!exclude_*)>` (no overlap with `exclude_fields`) and `truncated_input/truncated_output` on `<if(truncate)>` (they aren't selected by `page_wide` when `truncate=false`).

The SQL `EXCEPT` is a read optimization; response-level exclusion is unchanged. Regression-verified across sort (all fields), exclude (all fields), image truncation, and pagination — 72 span + 73 trace tests green.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
YarivHashaiComet: Same fix applied on the trace side in `1ef2e7c43b` (`traces_deduped` / `page_wide` / final `SELECT`) — see the SpanDAO thread for the details. Covered by the trace regression set (sort, exclude-all-fields, image truncation, pagination): 73 tests green.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/FindSpansResourceTest.java
```diff
@@ -3943,6 +3943,71 @@ static Stream<Arguments> whenSortingByValidFields__thenReturnTracesSorted() {
...
+                Comparator<Span> comparator, Span.SpanField excludeField) {
+            var workspaceName = RandomStringUtils.secure().nextAlphanumeric(10);
+            var workspaceId = UUID.randomUUID().toString();
+            var apiKey = UUID.randomUUID().toString();
+
+            mockTargetWorkspace(apiKey, workspaceName, workspaceId);
+
+            var projectName = RandomStringUtils.secure().nextAlphanumeric(10);
+            int total = 12;
+            int pageSize = 5;
```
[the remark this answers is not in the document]
YarivHashaiComet: False positive. `@ParameterizedTest` and `@MethodSource` **are** present — they're on the two lines directly above the `@DisplayName` shown in this hunk (`FindSpansResourceTest.java:3946-3947`), just outside the diff snippet. The provider drives the test: running it gives `Tests run: 2` (the two `@MethodSource` cases), not 0, and it fails when the query is mutated — so JUnit is not ignoring it. No change needed (re-adding the annotations would duplicate them and break compilation).

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/GetTracesByProjectResourceTest.java
```diff
@@ -4527,6 +4527,90 @@ void getTracesByProject__whenSortingByValidFields__thenReturnTracesSorted(Compar
...
+                SortingField sorting, Comparator<Trace> comparator, Trace.TraceField excludeField) {
+            var workspaceName = RandomStringUtils.secure().nextAlphanumeric(10);
+            var workspaceId = UUID.randomUUID().toString();
+            var apiKey = UUID.randomUUID().toString();
+
+            mockTargetWorkspace(apiKey, workspaceName, workspaceId);
+
+            var projectName = RandomStringUtils.secure().nextAlphanumeric(10);
+            int total = 12;
+            int pageSize = 5;
```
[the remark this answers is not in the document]
YarivHashaiComet: False positive. `@ParameterizedTest` and `@MethodSource` **are** present — they sit on the two lines directly above the `@DisplayName` shown here (`GetTracesByProjectResourceTest.java:4530-4531`), outside this diff hunk. Verified the provider drives it: `Tests run: 2` (both `@MethodSource` cases). No change needed — re-adding the annotations would duplicate them and fail to compile.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/FindSpansResourceTest.java
```diff
@@ -3943,6 +3943,116 @@ static Stream<Arguments> whenSortingByValidFields__thenReturnTracesSorted() {
...
+                    .request()
+                    .header(HttpHeaders.AUTHORIZATION, apiKey)
+                    .header(WORKSPACE_HEADER, workspaceName)
+                    .get()) {
+
+                assertThat(response.getStatusInfo().getStatusCode()).isEqualTo(200);
+                var content = response.readEntity(Span.SpanPage.class).content();
+                assertThat(content).hasSize(1);
+                assertThat(content.getFirst().input()).as("excluded wide field is absent").isNull();
+                assertThat(content.getFirst().output()).as("non-excluded wide field is present").isNotNull();
```
andrescrz: Nit: again missing full Span assertion: actual vs expected.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/GetTracesByProjectResourceTest.java
```diff
@@ -4527,6 +4527,131 @@ void getTracesByProject__whenSortingByValidFields__thenReturnTracesSorted(Compar
...
+            queryParams.put("size", "5");
+            queryParams.put("truncate", "true");
+            queryParams.put("exclude", toURLEncodedQueryParam(List.of(Trace.TraceField.INPUT)));
+
+            try (var response = traceResourceClient.callGetTracesWithQueryParams(apiKey, workspaceName, queryParams)) {
+                assertThat(response.getStatusInfo().getStatusCode()).isEqualTo(HttpStatus.SC_OK);
+                var content = response.readEntity(Trace.TracePage.class).content();
+                assertThat(content).hasSize(1);
+                assertThat(content.getFirst().input()).as("excluded wide field is absent").isNull();
+                assertThat(content.getFirst().output()).as("non-excluded wide field is present").isNotNull();
```
andrescrz: Ditto, but for traces.

andrescrz (changes requested):
The PR has improved a lot, but in my view, there's this remaining blocking comment. It's missing a critical test. Details here.

**M2. Filter + sort + exclude + multi-page regression test (coverage)**

The exact `OPIK-6747` production scenario — a filtered traces-list call that blew up to 5–32 GiB — has no regression test in CI. The PR's perf claim ("~99% peak-memory reduction") and correctness claim ("byte-for-byte identical pages") are documented **but unverified by automation.**

Needed: `insert > page-size matching rows + non-matching rows`, apply `input CONTAINS '<term>', sort input ASC, exclude output`, walk pages, assert paged-walk `content == full-page reference == Java-side filter+sort reference`.  **Mirror for spans.**

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/FindSpansResourceTest.java
YarivHashaiComet: Done in `e9d67d418b` — replaced the spot-checks with a full actual-vs-expected assertion (`assertSpan`). The span is created with plain-JSON input/output/metadata (so `truncate` is a no-op on the content), and the expectation is the created span with `input` nulled (excluded); every other field is compared.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/GetTracesByProjectResourceTest.java
YarivHashaiComet: Done in `e9d67d418b` — same on the trace side: full `assertTraces` actual-vs-expected (created trace with `input` nulled, plain-JSON content so `truncate` is a no-op).

YarivHashaiComet:
@andrescrz on the sorting + test-integrity asks — added a dedicated regression test for both spans and traces in `fde203b2ec`:

- `FindSpansResourceTest#whenFilterSortExcludeAcrossPages__thenEachPageMatchesFullPageAndReference`
- `GetTracesByProjectResourceTest#getTracesByProject__whenFilterSortExcludeAcrossPages__thenEachPageMatchesFullPageAndReference`

Each one reproduces the exact deferred-wide-column scenario end-to-end: seed 12 matching + 5 non-matching rows, then list with **filter** (`input CONTAINS`) + **custom sort** (`input ASC`, a wide text column → exercises the `sort_needs_wide` / dynamic `sort_fields` path in `page_ids`) + **field exclusion** (`output`, a deferred wide column), paged (size 5).

On integrity of data, the assertions are now full-row, not id-only:
- Every page is asserted byte-for-byte (via `assertSpan` / `assertTraces`) against an **independent Java-side filter+sort reference** (matching rows only, sorted by `input`, excluded field nulled via `EXCLUDE_FUNCTIONS`).
- The single full page == the walked pages == that reference.
- Non-matching rows must be absent **and** excluded from `total`.

This validates that sorting by a wide column survives the two-phase pre-filter (the case that would regress if `page_ids` dropped the sort key) for both spans and traces. Hand-mutation confirmed it's non-vacuous: dropping `<sort_fields>` from `page_ids` turns these red.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/FindSpansResourceTest.java
```diff
@@ -3943,6 +3943,188 @@ static Stream<Arguments> whenSortingByValidFields__thenReturnTracesSorted() {
...
+                    .queryParam("project_name", projectName)
+                    .queryParam("truncate", true)
+                    .queryParam("exclude", toURLEncodedQueryParam(List.of(Span.SpanField.INPUT)))
+                    .request()
+                    .header(HttpHeaders.AUTHORIZATION, apiKey)
+                    .header(WORKSPACE_HEADER, workspaceName)
+                    .get()) {
+
+                assertThat(response.getStatusInfo().getStatusCode()).isEqualTo(200);
+                var actualSpans = response.readEntity(Span.SpanPage.class).content();
```
andrescrz, line 4053: Minor: better encapsulated in the client as well.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/GetTracesByProjectResourceTest.java
```diff
@@ -4527,6 +4527,225 @@ void getTracesByProject__whenSortingByValidFields__thenReturnTracesSorted(Compar
...
+            Map<String, String> queryParams = new HashMap<>();
+            queryParams.put("project_name", projectName);
+            queryParams.put("page", "1");
+            queryParams.put("size", "5");
+            queryParams.put("truncate", "true");
+            queryParams.put("exclude", toURLEncodedQueryParam(List.of(Trace.TraceField.INPUT)));
+
+            try (var response = traceResourceClient.callGetTracesWithQueryParams(apiKey, workspaceName, queryParams)) {
+                assertThat(response.getStatusInfo().getStatusCode()).isEqualTo(HttpStatus.SC_OK);
+                var actualTraces = response.readEntity(Trace.TracePage.class).content();
```
andrescrz, line 4652: Minor: same, better encapsulate the param building, client call, 200 assert and read entity into the client. 

The rest of asserts are particular to this test.

andrescrz (approved):
Just a couple of minor comments on the tests, but not blockers. LGTM.

PR #6937, opened 2026-06-01
[OPIK-6754] [FE] polish: env pill sizing, deploy menu UX, playground hover fix

aadereiko:
## Details

Follow-up polish on the prompt-environment surfaces shipped in [OPIK-6626 (#6892)](https://github.com/comet-ml/opik/pull/6892): tighten env-pill sizing, surface env icons on the Prompt playground chip, standardize dropdown-item heights, and fix a Playground hover-strip bug that caused the load-prompt popover to shift.

- Added a `pill` size variant on `EnvironmentBadge` / `EnvironmentBadgeList` (20px tall, 10px label, asymmetric pl-1.5/pr-2 to seat the leading env icon); used across the prompt detail header, mobile version picker, version history timeline, and the +N overflow chip
- `LoadedPromptDisplay` now surfaces env icons next to the version label, matching the Agent runner
- `DropdownMenuItem size="sm"` (h-8 / px-3) applied uniformly to `DeployToEnvironmentMenu`, `EnvironmentFilterSelect`, the Agent runner versions menu, and the `PromptLibraryMenu` version submenu (which also gets regular label weight and a 32px `dimension="sm"` SearchInput)
- `AddEditEnvironmentDialog`: color popover closes on preset click / Enter on a valid hex; Cmd/Ctrl+Enter saves the form
- `EnvironmentNameCell`: dropped extra `px-2 rounded-md border border-transparent` so content aligns with the column header like every other cell
- `PlaygroundPrompt`: pin the hover-revealed actions strip open while the chat prompt library popover is up — previously it collapsed to `max-w-0` when the mouse left the trigger, causing the popover to re-anchor and visibly shift

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- OPIK-6754

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Scope: implementation of all listed bullets (component edits, refactor of the pill size variant, PR description)
- Human verification: changes manually reviewed by author; visual scenarios listed in Testing below to be exercised in dev server

## Testing

- `npx tsc --noEmit` and `npx eslint <changed files>` in `apps/opik-frontend` — both pass clean.
- Manual UI verification (in progress):
  - Prompt detail page: env pills render with the new pill size, +N overflow chip matches the pill height
  - Prompt playground (chat): chip shows env icons; opening the library popover keeps the load-prompt icon visible (no shift, no disappear) while the popover is open
  - Agent runner: version dropdown rows are 32px / 12px and labels are regular weight
  - Traces env filter dropdown: list items are 32px / 12px
  - Configuration → Environments tab: name cell content aligns with the "Name" column header
  - Edit environment dialog: clicking a preset color closes the popover; pressing Enter in the hex input with a valid value closes the popover; Cmd/Ctrl+Enter saves
  - Default-named envs (development/staging/production) keep the colored square + white icon everywhere

No tests added (no existing FE tests for these surfaces; changes are visual/interaction polish).

## Documentation

No documentation changes — internal FE polish only.

### DIFF
--- a/apps/opik-frontend/src/api/environments/useEnvironmentsList.ts
+++ b/apps/opik-frontend/src/api/environments/useEnvironmentsList.ts
@@ -5,7 +5,7 @@ import api, {
   QueryConfig,
 } from "@/api/api";
 import { Environment } from "@/types/environments";
-import { sortEnvironments } from "@/shared/EnvironmentLabel/helpers";
+import { sortEnvironments } from "@/utils/environments";
 
 export type UseEnvironmentsListResponse = {
   content: Environment[];
--- a/apps/opik-frontend/src/api/environments/useEnvironmentsList.ts
+++ b/apps/opik-frontend/src/api/environments/useEnvironmentsList.ts
@@ -5,6 +5,7 @@ import api, {
   QueryConfig,
 } from "@/api/api";
 import { Environment } from "@/types/environments";
+import { sortEnvironments } from "@/shared/EnvironmentLabel/helpers";
 
 export type UseEnvironmentsListResponse = {
   content: Environment[];
@@ -20,7 +21,7 @@ const getEnvironmentsList = async ({ signal }: QueryFunctionContext) => {
     { signal },
   );
 
-  return data;
+  return { ...data, content: sortEnvironments(data.content) };
 };
 
 export default function useEnvironmentsList(
--- a/apps/opik-frontend/src/api/environments/useEnvironmentsList.ts
+++ b/apps/opik-frontend/src/api/environments/useEnvironmentsList.ts
@@ -21,7 +21,7 @@ const getEnvironmentsList = async ({ signal }: QueryFunctionContext) => {
     { signal },
   );
 
-  return { ...data, content: sortEnvironments(data.content) };
+  return { ...data, content: sortEnvironments(data.content ?? []) };
 };
 
 export default function useEnvironmentsList(
--- a/apps/opik-frontend/src/shared/EnvironmentLabel/EnvironmentBadge.tsx
+++ b/apps/opik-frontend/src/shared/EnvironmentLabel/EnvironmentBadge.tsx
@@ -8,19 +8,19 @@ import {
 } from "./helpers";
 import { useEnvironmentByName } from "./EnvironmentLabel";
 
-// `promptPill` is a project-specific variant used across prompt-version
+// `pill` is a project-specific variant used across prompt-version
 // surfaces (prompt detail header, mobile version picker, version history
 // timeline, +N overflow tooltip): 20px tall with a 10px label and asymmetric
 // padding so the leading env icon hugs the left edge while the text gets a
 // little more breathing room on the right.
-export type EnvironmentBadgeSize = NonNullable<TagProps["size"]> | "promptPill";
+export type EnvironmentBadgeSize = NonNullable<TagProps["size"]> | "pill";
 
 const SIZE_CLASSES: Record<EnvironmentBadgeSize, string> = {
   default: "comet-body-xs h-5 px-2 leading-5 rounded-sm",
   sm: "comet-body-xs h-4 px-2 text-[11px] leading-4 rounded-sm",
   md: "comet-body-s h-6 px-1.5 leading-6 rounded-md",
   lg: "comet-body-s h-7 px-3 leading-7 rounded-md",
-  promptPill: "h-5 pl-1.5 pr-2 text-[10px] leading-5 rounded-sm",
+  pill: "h-5 pl-1.5 pr-2 text-[10px] leading-5 rounded-sm",
 };
 
 const ICON_SIZE_CLASSES: Record<EnvironmentBadgeSize, string> = {
@@ -28,7 +28,7 @@ const ICON_SIZE_CLASSES: Record<EnvironmentBadgeSize, string> = {
   sm: "size-2.5",
   md: "size-3",
   lg: "size-3.5",
-  promptPill: "size-2.5",
+  pill: "size-2.5",
 };
 
 type EnvironmentBadgeProps = {
--- a/apps/opik-frontend/src/shared/EnvironmentLabel/EnvironmentBadge.tsx
+++ b/apps/opik-frontend/src/shared/EnvironmentLabel/EnvironmentBadge.tsx
@@ -8,30 +8,32 @@ import {
 } from "./helpers";
 import { useEnvironmentByName } from "./EnvironmentLabel";
 
-const SIZE_CLASSES: Record<NonNullable<TagProps["size"]>, string> = {
+// `promptPill` is a project-specific variant used across prompt-version
+// surfaces (prompt detail header, mobile version picker, version history
+// timeline, +N overflow tooltip): 20px tall with a 10px label and asymmetric
+// padding so the leading env icon hugs the left edge while the text gets a
+// little more breathing room on the right.
+export type EnvironmentBadgeSize = NonNullable<TagProps["size"]> | "promptPill";
+
+const SIZE_CLASSES: Record<EnvironmentBadgeSize, string> = {
   default: "comet-body-xs h-5 px-2 leading-5 rounded-sm",
   sm: "comet-body-xs h-4 px-2 text-[11px] leading-4 rounded-sm",
   md: "comet-body-s h-6 px-1.5 leading-6 rounded-md",
   lg: "comet-body-s h-7 px-3 leading-7 rounded-md",
+  promptPill: "h-5 pl-1.5 pr-2 text-[10px] leading-5 rounded-sm",
 };
 
-const ICON_SIZE_CLASSES: Record<NonNullable<TagProps["size"]>, string> = {
+const ICON_SIZE_CLASSES: Record<EnvironmentBadgeSize, string> = {
   default: "size-3",
   sm: "size-2.5",
   md: "size-3",
   lg: "size-3.5",
+  promptPill: "size-2.5",
 };
-
-// Asymmetric pill spec used across prompt-version surfaces (prompt detail
-// header, mobile version picker, version history timeline, +N overflow
-// tooltip). Pass as `badgeClassName` on EnvironmentBadgeList — tailwind-merge
-// resolves the conflicts against SIZE_CLASSES["sm"].
-export const ENV_BADGE_PROMPT_PILL_CLASS =
-  "h-5 pl-1.5 pr-2 text-[10px] leading-5";
 
 type EnvironmentBadgeProps = {
   name?: string | null;
-  size?: TagProps["size"];
+  size?: EnvironmentBadgeSize;
   className?: string;
 };
 
--- a/apps/opik-frontend/src/shared/EnvironmentLabel/EnvironmentBadge.tsx
+++ b/apps/opik-frontend/src/shared/EnvironmentLabel/EnvironmentBadge.tsx
@@ -8,11 +8,6 @@ import {
 } from "./helpers";
 import { useEnvironmentByName } from "./EnvironmentLabel";
 
-// `pill` is a project-specific variant used across prompt-version
-// surfaces (prompt detail header, mobile version picker, version history
-// timeline, +N overflow tooltip): 20px tall with a 10px label and asymmetric
-// padding so the leading env icon hugs the left edge while the text gets a
-// little more breathing room on the right.
 export type EnvironmentBadgeSize = NonNullable<TagProps["size"]> | "pill";
 
 const SIZE_CLASSES: Record<EnvironmentBadgeSize, string> = {
--- a/apps/opik-frontend/src/shared/EnvironmentLabel/EnvironmentBadge.tsx
+++ b/apps/opik-frontend/src/shared/EnvironmentLabel/EnvironmentBadge.tsx
@@ -22,6 +22,13 @@ const ICON_SIZE_CLASSES: Record<NonNullable<TagProps["size"]>, string> = {
   lg: "size-3.5",
 };
 
+// Asymmetric pill spec used across prompt-version surfaces (prompt detail
+// header, mobile version picker, version history timeline, +N overflow
+// tooltip). Pass as `badgeClassName` on EnvironmentBadgeList — tailwind-merge
+// resolves the conflicts against SIZE_CLASSES["sm"].
+export const ENV_BADGE_PROMPT_PILL_CLASS =
+  "h-5 pl-1.5 pr-2 text-[10px] leading-5";
+
 type EnvironmentBadgeProps = {
   name?: string | null;
   size?: TagProps["size"];
--- a/apps/opik-frontend/src/shared/EnvironmentLabel/EnvironmentBadge.tsx
+++ b/apps/opik-frontend/src/shared/EnvironmentLabel/EnvironmentBadge.tsx
@@ -15,7 +15,7 @@ const SIZE_CLASSES: Record<EnvironmentBadgeSize, string> = {
   sm: "comet-body-xs h-4 px-2 text-[11px] leading-4 rounded-sm",
   md: "comet-body-s h-6 px-1.5 leading-6 rounded-md",
   lg: "comet-body-s h-7 px-3 leading-7 rounded-md",
-  pill: "h-5 pl-1.5 pr-2 text-[10px] leading-5 rounded-sm",
+  pill: "comet-body-xs h-5 pl-1.5 pr-2 text-[10px] leading-5 rounded-sm",
 };
 
 const ICON_SIZE_CLASSES: Record<EnvironmentBadgeSize, string> = {
--- a/apps/opik-frontend/src/shared/EnvironmentLabel/helpers.ts
+++ b/apps/opik-frontend/src/shared/EnvironmentLabel/helpers.ts
@@ -33,3 +33,17 @@ export const getDefaultEnvironmentMeta = (
 
 export const isDefaultEnvironmentName = (name: string | undefined | null) =>
   normalize(name) in DEFAULT_ENVIRONMENTS;
+
+const DEFAULT_ENV_SORT_PRIORITY: Record<string, number> = {
+  production: 0,
+  staging: 1,
+  development: 2,
+};
+
+export const sortEnvironments = <T extends { name: string }>(envs: T[]): T[] =>
+  envs.slice().sort((a, b) => {
+    const aPriority = DEFAULT_ENV_SORT_PRIORITY[normalize(a.name)] ?? Infinity;
+    const bPriority = DEFAULT_ENV_SORT_PRIORITY[normalize(b.name)] ?? Infinity;
+    if (aPriority !== bPriority) return aPriority - bPriority;
+    return a.name.localeCompare(b.name);
+  });
--- a/apps/opik-frontend/src/shared/EnvironmentLabel/helpers.ts
+++ b/apps/opik-frontend/src/shared/EnvironmentLabel/helpers.ts
@@ -1,5 +1,9 @@
 import { Code, FlaskRound, LucideIcon, Rocket } from "lucide-react";
 import { DEFAULT_HEX_COLOR, HEX_COLOR_REGEX } from "@/constants/colorVariants";
+import {
+  DEFAULT_ENVIRONMENT_NAMES,
+  DefaultEnvironmentName,
+} from "@/utils/environments";
 
 export const resolveEnvironmentColor = (color: string | undefined) =>
   color && HEX_COLOR_REGEX.test(color) ? color : DEFAULT_HEX_COLOR;
@@ -12,24 +16,34 @@ export const getContrastingTextColor = (hex: string): string => {
   return luminance > 140 ? "#0f172a" : "#ffffff";
 };
 
-// Seeded server-side by migration 000066_seed_default_environments. Name +
-// color are kept in lock-step with the backend so the UI can recognize them
-// and protect their color from edits.
+// Seeded server-side by migration 000066_seed_default_environments. The set of
+// names is the source of truth in `utils/environments.ts`
+// (DEFAULT_ENVIRONMENT_NAMES); TypeScript enforces this map carries one entry
+// per default env so adding a new default forces editing both color/icon and
+// sort priority together.
 type DefaultEnvironmentMeta = { color: string; icon: LucideIcon };
 
-const DEFAULT_ENVIRONMENTS: Record<string, DefaultEnvironmentMeta> = {
-  development: { color: "#EF6868", icon: Code },
-  staging: { color: "#F4B400", icon: FlaskRound },
+const DEFAULT_ENVIRONMENTS: Record<
+  DefaultEnvironmentName,
+  DefaultEnvironmentMeta
+> = {
   production: { color: "#19A979", icon: Rocket },
+  staging: { color: "#F4B400", icon: FlaskRound },
+  development: { color: "#EF6868", icon: Code },
 };
 
-const normalize = (name: string | undefined | null) =>
+const normalize = (name: string | undefined | null): string =>
   name?.trim().toLowerCase() ?? "";
+
+const isDefaultName = (name: string): name is DefaultEnvironmentName =>
+  (DEFAULT_ENVIRONMENT_NAMES as readonly string[]).includes(name);
 
 export const getDefaultEnvironmentMeta = (
   name: string | undefined | null,
-): DefaultEnvironmentMeta | null =>
-  DEFAULT_ENVIRONMENTS[normalize(name)] ?? null;
+): DefaultEnvironmentMeta | null => {
+  const normalized = normalize(name);
+  return isDefaultName(normalized) ? DEFAULT_ENVIRONMENTS[normalized] : null;
+};
 
 export const isDefaultEnvironmentName = (name: string | undefined | null) =>
-  normalize(name) in DEFAULT_ENVIRONMENTS;
+  isDefaultName(normalize(name));
--- a/apps/opik-frontend/src/shared/EnvironmentLabel/helpers.ts
+++ b/apps/opik-frontend/src/shared/EnvironmentLabel/helpers.ts
@@ -33,17 +33,3 @@ export const getDefaultEnvironmentMeta = (
 
 export const isDefaultEnvironmentName = (name: string | undefined | null) =>
   normalize(name) in DEFAULT_ENVIRONMENTS;
-
-const DEFAULT_ENV_SORT_PRIORITY: Record<string, number> = {
-  production: 0,
-  staging: 1,
-  development: 2,
-};
-
-export const sortEnvironments = <T extends { name: string }>(envs: T[]): T[] =>
-  envs.slice().sort((a, b) => {
-    const aPriority = DEFAULT_ENV_SORT_PRIORITY[normalize(a.name)] ?? Infinity;
-    const bPriority = DEFAULT_ENV_SORT_PRIORITY[normalize(b.name)] ?? Infinity;
-    if (aPriority !== bPriority) return aPriority - bPriority;
-    return a.name.localeCompare(b.name);
-  });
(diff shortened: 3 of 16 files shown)


apps/opik-frontend/src/v2/pages/PlaygroundPage/PlaygroundPrompts/PlaygroundPrompt.tsx
```diff
@@ -381,6 +391,10 @@ const PlaygroundPrompt = ({
                 chatPromptVersionData?.tags ??
                 chatPromptData?.latest_version?.tags
               }
+              versionEnvironments={
+                chatPromptVersionData?.environments ??
+                chatPromptData?.latest_version?.environments
+              }
```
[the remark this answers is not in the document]
aadereiko: Mirrors the existing tags fallback two lines above (`chatPromptVersionData?.tags ?? chatPromptData?.latest_version?.tags`) introduced in OPIK-6457 — same in-flight-fetch behavior. Worth tightening for both tags and envs together in a follow-up; leaving as-is here for consistency with the pre-existing pattern.

apps/opik-frontend/src/v2/pages-shared/llm/PromptsSelectBox/PromptsSelectBox.tsx
```diff
@@ -266,21 +266,25 @@ const CompactLoadedPrompt: React.FC<CompactLoadedPromptProps> = ({
...
   const { getDescriptor } = usePromptVersionsWithLabels(promptId, {
     enabled: Boolean(versionId),
   });
-  const selectedVersionTags = versionId
-    ? getDescriptor(versionId)?.version.tags
+  const selectedVersion = versionId
+    ? getDescriptor(versionId)?.version
     : undefined;
 
   return (
```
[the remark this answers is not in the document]
aadereiko: Same convention as the existing `selectedVersionTags ?? data?.latest_version?.tags` fallback this commit replaces — tags had the same theoretical race. Worth tightening for both fields together in a follow-up; keeping the pattern consistent here.

apps/opik-frontend/src/v2/pages/PromptPage/PromptTab/DeployToEnvironmentMenu.tsx
```diff
@@ -174,19 +175,19 @@ const DeployToEnvironmentMenu: React.FC<DeployToEnvironmentMenuProps> = ({
...
             </DropdownMenuItem>
           </>
         )}
         <DropdownMenuSeparator />
-        <DropdownMenuItem asChild>
+        <DropdownMenuItem size="sm" asChild>
           <Link
             to="/$workspaceName/configuration"
             params={{ workspaceName }}
             search={{ tab: CONFIGURATION_TABS.ENVIRONMENTS }}
```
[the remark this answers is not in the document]
aadereiko: Two occurrences of ~6 lines feels under the threshold for extraction — preferring to keep the indirection out until a third caller (or a tracking/copy change) makes the duplication a problem. Happy to extract if a third call site appears.

CometActions:

### Access Information
- **URL:** https://pr-6937.dev.comet.com
- **Namespace:** pr-6937
- **Version:** 2.0.54-6937-merge-2207
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-6937%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

CometActions:

### Access Information
- **URL:** https://pr-6937.dev.comet.com
- **Namespace:** pr-6937
- **Version:** 2.0.54-6937-merge-2208
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-6937%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

apps/opik-frontend/src/shared/EnvironmentLabel/helpers.ts
```diff
@@ -33,3 +33,17 @@ export const getDefaultEnvironmentMeta = (
 
 export const isDefaultEnvironmentName = (name: string | undefined | null) =>
   normalize(name) in DEFAULT_ENVIRONMENTS;
+
+const DEFAULT_ENV_SORT_PRIORITY: Record<string, number> = {
+  production: 0,
+  staging: 1,
+  development: 2,
+};
```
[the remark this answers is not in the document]
aadereiko: Intentional — the product ask is to always render production → staging → development first, then alphabetical, regardless of the BE `position` field (which the BE seeds in a matching order today but isn't guaranteed once users start reordering their workspace). Centralizing the FE-side sort gives us one source of truth for ordering across the prompt deploy menu, traces filter, and configuration table.

CometActions:

### Access Information
- **URL:** https://pr-6937.dev.comet.com
- **Namespace:** pr-6937
- **Version:** 2.0.54-6937-merge-2210
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-6937%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-6937`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

CometActions:

### Access Information
- **URL:** https://pr-6937.dev.comet.com
- **Namespace:** pr-6937
- **Version:** 2.0.54-6937-merge-2213
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-6937%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

apps/opik-frontend/src/api/environments/useEnvironmentsList.ts
```diff
@@ -20,7 +21,7 @@ const getEnvironmentsList = async ({ signal }: QueryFunctionContext) => {
     { signal },
   );
 
-  return data;
+  return { ...data, content: sortEnvironments(data.content) };
```
YarivHashaiComet: **[medium] Presentation-ordering baked into the shared data-fetch layer**

Injecting a fixed presentation sort into the shared getEnvironmentsList fetcher forces one ordering policy onto every caller. EnvironmentsTab declares the Name column `sortable: true` and the response carries `sortable_by`; baking a sort into the fetcher means the table cannot later expose its own column sorting without double-sorting, and the four consumers can no longer choose an order appropriate to their context. The data layer should return raw results and let each caller (or the backend `sortable_by`/`position` contract) decide ordering.

💡 Return data unmodified from the fetcher and apply sortEnvironments() in the consumer(s) that actually need a custom display order (memoized with useMemo), or rely on backend ordering. Keep the data layer free of presentation concerns.

_rule: `frontend-architecture-data-api-layer-must-not-import-from-ui-components-or-embed-u`_

apps/opik-frontend/src/shared/EnvironmentLabel/EnvironmentBadge.tsx
```diff
@@ -8,23 +8,27 @@ import {
...
 
-const SIZE_CLASSES: Record<NonNullable<TagProps["size"]>, string> = {
+export type EnvironmentBadgeSize = NonNullable<TagProps["size"]> | "pill";
+
+const SIZE_CLASSES: Record<EnvironmentBadgeSize, string> = {
   default: "comet-body-xs h-5 px-2 leading-5 rounded-sm",
   sm: "comet-body-xs h-4 px-2 text-[11px] leading-4 rounded-sm",
   md: "comet-body-s h-6 px-1.5 leading-6 rounded-md",
   lg: "comet-body-s h-7 px-3 leading-7 rounded-md",
+  pill: "h-5 pl-1.5 pr-2 text-[10px] leading-5 rounded-sm",
```
YarivHashaiComet: **[medium] New 'pill' size must be kept in lock-step across two parallel size-class maps**

Adding the 'pill' entry widens EnvironmentBadgeSize, which is also the key type of COUNTER_CLASSES in the sibling EnvironmentBadgeList.tsx (line 24). Both files now independently maintain a Record<EnvironmentBadgeSize, string> of size -> Tailwind classes that must be edited together for every size change. The two pill strings have already diverged in padding ('pl-1.5 pr-2' here vs 'px-1.5' in the list counter) with no shared source, so a future tweak to one will silently desync the badge from its overflow counter.

💡 Extract the size->class lookup (or at least the shared per-size dimensional tokens: height, font-size, leading, radius) into a single exported map/helper in helpers.ts and import it in both EnvironmentBadge and EnvironmentBadgeList, deriving the small badge-vs-counter padding differences from that single source instead of re-listing every size in each file.

_rule: `frontend-style-maintainability-extract-shared-logic-utilities-and-components-never-duplica`_

apps/opik-frontend/src/v2/pages-shared/environments/AddEditEnvironmentDialog/AddEditEnvironmentDialog.tsx
```diff
@@ -152,71 +153,94 @@ const AddEditEnvironmentDialog: React.FunctionComponent<
     setOpen,
   ]);
 
+  const handleBodyKeyDown = useCallback(
```
YarivHashaiComet, line 160: **[medium] New Cmd/Ctrl+Enter submit path is not guarded against an in-flight mutation**

handleBodyKeyDown (new in this diff) invokes submitHandler() on every Cmd/Ctrl+Enter, and submitHandler fires createMutation/updateMutation unconditionally. Neither the submit Button (disabled={!isValid}) nor the keyboard handler is gated on the mutation's pending state — the hooks expose isPending but the component only destructures `mutate`. Because the new shortcut can repeat while a key is held (and the onClick path is also unguarded), a user can dispatch the same create/update mutation multiple times before the first resolves, producing duplicate requests / duplicate environments.

💡 Destructure `isPending` from both mutation hooks, derive `const isSubmitting = createIsPending || updateIsPending`, include it in `disabled` (disabled={!isValid || isSubmitting}) on the submit Button, and early-return from submitHandler/handleBodyKeyDown when isSubmitting so both the click and keyboard paths share the same guard.

_rule: `frontend-correctness-disable-interactive-controls-for-the-full-duration-of-every-`_

apps/opik-frontend/src/utils/environments.ts
```diff
@@ -0,0 +1,15 @@
+const DEFAULT_ENV_SORT_PRIORITY: Record<string, number> = {
```
YarivHashaiComet: **[low] Default environment name list duplicated across modules**

DEFAULT_ENV_SORT_PRIORITY hardcodes the same three default environment names (production/staging/development) that already exist as the canonical `DEFAULT_ENVIRONMENTS` record in apps/opik-frontend/src/shared/EnvironmentLabel/helpers.ts (development/staging/production). The set of seeded default environments is a single domain fact kept in lock-step with backend migration 000066; maintaining two independent literal lists means adding/removing a default env requires editing both, and a missed edit silently breaks either the sort priority or the color/icon recognition.

💡 Derive the priority keys from a single shared source of default environment names (e.g. export the key list from EnvironmentLabel/helpers.ts) so the sort ordering and the default-meta lookup cannot diverge.

_rule: `frontend-style-maintainability-extract-shared-logic-utilities-and-components-never-duplica`_

apps/opik-frontend/src/v2/pages-shared/environments/AddEditEnvironmentDialog/AddEditEnvironmentDialog.tsx
```diff
@@ -152,71 +153,94 @@ const AddEditEnvironmentDialog: React.FunctionComponent<
...
+                  id="environmentName"
+                  className="flex-1 rounded-l-none"
+                  placeholder="e.g. staging"
+                  value={name}
+                  onChange={(event) => handleNameChange(event.target.value)}
+                  maxLength={ENVIRONMENT_NAME_MAX_LENGTH}
+                />
+              </div>
+              {nameError || submitError ? (
+                <p className="comet-body-xs text-destructive">
```
YarivHashaiComet: **[low] Validation/error message lacks ARIA live region**

The name validation / submit-error <p className="comet-body-xs text-destructive"> (in the block reworked by this diff) is rendered without role="alert" or aria-live. Screen readers will not announce when `nameError` or the backend `submitError` appears, so users relying on assistive tech get no feedback that the form is invalid or that the create/update mutation failed.

💡 Add role="alert" (or aria-live="polite") to the destructive <p> that renders {nameError || submitError} so the message is announced when it changes.

_rule: `frontend-style-maintainability-make-interactive-elements-keyboard-accessible-and-annotate-`_

apps/opik-frontend/src/shared/EnvironmentLabel/EnvironmentBadge.tsx
YarivHashaiComet: **[low] 'pill' size omits the comet-body typography token used by every other size**

Every other SIZE_CLASSES row anchors typography on a design-system token (comet-body-xs / comet-body-s) and only overrides font-size where needed. The new 'pill' row drops the comet-body-* token entirely and relies solely on the raw arbitrary value text-[10px], so it no longer inherits the design system's font-family/weight/tracking for that text class. This makes pill text styling inconsistent with the rest of the badge sizes.

💡 Prefix the pill class string with the appropriate comet-body-* token (e.g. 'comet-body-xs') the way the default/sm rows do, keeping only the font-size override, so pill text stays on the shared typography scale.

_rule: `frontend-style-maintainability-use-theme-aware-design-system-tokens-for-all-colors-never-h`_

apps/opik-frontend/src/v2/pages-shared/llm/PromptLibraryMenu/PromptLibraryMenu.tsx
```diff
@@ -109,6 +109,7 @@ const PromptLibraryMenu: React.FC<PromptLibraryMenuProps> = ({
             setSearchText={setSearch}
             placeholder="Search"
             variant="ghost"
+            dimension="sm"
```
YarivHashaiComet, line 112: **[low] dimension="sm" has no matching SearchInput style entry**

The newly added `dimension="sm"` is a valid Input dimension (it maps to `h-8 px-3` in ui/input.tsx and is forwarded to DebounceInput), but SearchInput's `SEARCH_STYLE_BY_DIMENSION` map (src/shared/SearchInput/SearchInput.tsx) only defines an `xs` entry. For `sm`, `style` falls back to `DEFAULT_SEARCH_STYLE`, so the search-icon overlay (iconWrapper `left-2.5`, icon `size-3.5`) and the `px-8` padding it applies via `className` are the default-size values, not values tuned for the shorter h-8 input. The default `px-8` className also overrides the `px-3` that `dimension="sm"` sets on the input. The result renders, but the icon positioning/padding is not actually adapted to the `sm` size the prop implies. Verify the icon and padding look correct at the smaller height, and consider adding an `sm` entry to SEARCH_STYLE_BY_DIMENSION so the override is intentional rather than an unhandled fallthrough.

💡 Either add a dedicated `sm` entry to `SEARCH_STYLE_BY_DIMENSION` in SearchInput (icon position/size + padding sized for h-8), or confirm the default style is the desired look for `sm` and rely on it knowingly.

_general finding_

apps/opik-frontend/src/v2/pages-shared/llm/PromptsSelectBox/PromptsSelectBox.tsx
```diff
@@ -266,21 +266,25 @@ const CompactLoadedPrompt: React.FC<CompactLoadedPromptProps> = ({
...
     <LoadedPromptDisplay
       name={displayName}
       templateStructure={data?.template_structure}
       versionLabel={versionLabel}
-      versionTags={selectedVersionTags ?? data?.latest_version?.tags}
+      versionTags={selectedVersion?.tags ?? data?.latest_version?.tags}
+      versionEnvironments={
+        selectedVersion?.environments ?? data?.latest_version?.environments
```
YarivHashaiComet, line 286: **[low] environments null vs undefined falls back to latest_version**

Unlike `tags` (typed `string[] | undefined`), `PromptVersion.environments` is typed `string[] | null`. The expression `selectedVersion?.environments ?? data?.latest_version?.environments` treats an explicit `null` from the selected version (server confirming the picked version has no environments) the same as `undefined` (not loaded) and falls through to the latest version's environments. In that edge case the chip would display the latest version's environments for an older pick, which contradicts the comment's intent ('reflect what's actually loaded'). In practice the impact is minimal because EnvironmentBadgeList renders nothing for both `null` and `[]`, and the sibling PromptLibraryMenu treats environments the same loose way, so this is a defensive-hardening note rather than a realistic-path bug.

💡 If the backend ever emits `null` to mean 'this version has no environments' (distinct from 'not loaded'), guard explicitly, e.g. `const env = selectedVersion ? (selectedVersion.environments ?? []) : data?.latest_version?.environments;` so a loaded version's empty/null environments are not masked by the latest version's. Otherwise leave as-is for consistency with the adjacent versionTags fallback.

_rule: `frontend-correctness-distinguish-null-undefined-from-zero-and-from-an-empty-array`_

apps/opik-frontend/src/api/environments/useEnvironmentsList.ts
YarivHashaiComet: **[low] sortEnvironments(data.content) assumes content is always a present array**

api.get<T>() generics are compile-time only; the runtime payload is not validated. sortEnvironments calls envs.slice(), so if the backend ever returns a body without `content` (or null), this throws and crashes every consumer of the list. Previously the raw data was passed through and an undefined content was tolerated downstream via `data?.content ?? []`. This is defensive hardening (well-behaved backend always returns content), hence low.

💡 Guard the input, e.g. content: sortEnvironments(data.content ?? []), so a malformed/empty response degrades gracefully instead of throwing.

_rule: `frontend-api-contract-verify-api-response-shape-and-preserve-normalization-adapter`_

YarivHashaiComet (commented):
> 🧪 **PILOT — experimental code-review assistant (not a required gate).**
> Grounded in the team's own merged-PR review history; each finding was empirically
> verified before posting. We're calibrating it — please **👍/👎 each comment** so it learns.

9 findings — 0 high · 3 medium · 6 low. Suppressed by team conventions: see `suppressed.md`.

👍/👎 each comment — 👎 from a senior reviewer trains the engine to stop flagging it.

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-6937`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

apps/opik-frontend/src/shared/EnvironmentLabel/EnvironmentBadge.tsx
aadereiko: The badge vs counter padding difference is intentional: the badge has an icon on the left so it carries asymmetric `pl-1.5 pr-2` to seat the icon, while the +N counter is text-only and uses symmetric `px-1.5` (looks balanced on "+3"). Extracting just the shared dimensions (height / font / leading) into a single map saves three tokens but ties the two components together; given there are only two consumers and the shared parts already fit in one line each, I'll keep them adjacent-and-explicit and revisit if a third size variant lands. Also pushed a small change: both pill classes now anchor on `comet-body-xs` for the typography token, addressing your sibling comment.

apps/opik-frontend/src/api/environments/useEnvironmentsList.ts
aadereiko: Intentional product call: the ask is to render production → staging → development first then alphabetical *everywhere*, regardless of BE `position` (the BE seeds in that order today but it isn't a guarantee once we let users reorder). Centralizing in the fetcher is the single source of truth so the deploy menu, traces filter, and configuration table can't drift. The `sortable` flag on the Name column isn't wired to actual sorting yet — if/when we expose user-controlled column sort, that consumer can swap to a raw `select` and apply its own sort. Easy to refactor at that point; over-engineering it pre-emptively would risk drift in the meantime.

apps/opik-frontend/src/v2/pages-shared/llm/PromptsSelectBox/PromptsSelectBox.tsx
aadereiko: Same pattern as the existing `selectedVersionTags ?? data?.latest_version?.tags` fallback this commit replaces — tags has the identical theoretical race. Worth tightening both fields together (distinguishing null="empty" from undefined="not loaded") in a follow-up; keeping the convention consistent here.

apps/opik-frontend/src/v2/pages-shared/llm/PromptLibraryMenu/PromptLibraryMenu.tsx
aadereiko: Verified visually — at `dimension="sm"` (h-8) the icon stays vertically centered (it's positioned via top-1/2 + translate-y-1/2, independent of input height) and the px-8 left padding still leaves comfortable space between the icon and the placeholder. Adding a dedicated `sm` entry to `SEARCH_STYLE_BY_DIMENSION` matching the default values would be the more explicit fix, but that's a SearchInput change outside this PR's scope. Tracking as a follow-up.

apps/opik-frontend/src/api/environments/useEnvironmentsList.ts
```diff
@@ -20,7 +21,7 @@ const getEnvironmentsList = async ({ signal }: QueryFunctionContext) => {
     { signal },
   );
 
-  return data;
+  return { ...data, content: sortEnvironments(data.content ?? []) };
```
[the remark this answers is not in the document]
aadereiko: Same answer as the parallel Yariv review thread on this line: intentional product call to render production → staging → development first then alphabetical *everywhere*, regardless of BE `created_at` or `position` (the BE seeds in a similar order today but it's not a guarantee once users start reordering). Centralizing in the fetcher gives a single source of truth across the deploy menu, traces filter, and configuration table — if/when a consumer needs the raw backend order, it can opt out via React Query's `select` at that point.

apps/opik-frontend/src/shared/EnvironmentLabel/helpers.ts
```diff
@@ -12,24 +16,34 @@ export const getContrastingTextColor = (hex: string): string => {
...
-  development: { color: "#EF6868", icon: Code },
-  staging: { color: "#F4B400", icon: FlaskRound },
+const DEFAULT_ENVIRONMENTS: Record<
+  DefaultEnvironmentName,
+  DefaultEnvironmentMeta
+> = {
   production: { color: "#19A979", icon: Rocket },
+  staging: { color: "#F4B400", icon: FlaskRound },
+  development: { color: "#EF6868", icon: Code },
 };
```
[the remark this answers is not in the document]
aadereiko: Misread of the diff — `production` is the first entry in the new `DEFAULT_ENVIRONMENTS` map (line 30 in the new file). It shows up without a `+` because the previous version of the map also had a `production:` line at the same source position; only `staging` and `development` got reordered around it. Note also that the map is typed as `Record<DefaultEnvironmentName, DefaultEnvironmentMeta>` and `DefaultEnvironmentName` is the union `'production' | 'staging' | 'development'` — if production were missing, TypeScript would fail with TS2741 'Property production is missing'.

apps/opik-frontend/src/utils/environments.ts
```diff
@@ -0,0 +1,25 @@
+// Canonical list of default environment names seeded by backend migration
+// 000066. The order here drives the sort priority below AND the icon/color
+// lookup in shared/EnvironmentLabel/helpers.ts — keep both in lock-step by
+// editing this list, not its consumers.
+export const DEFAULT_ENVIRONMENT_NAMES = [
+  "production",
+  "staging",
+  "development",
+] as const;
```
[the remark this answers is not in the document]
aadereiko: Intentional — the FE sort priority deliberately diverges from the BE migration's `position` order to honor the product requirement that production renders first across every env list. The BE seed order is one of many initial states (users can rename/reorder once they create the workspace) and was never intended as the canonical render order. Centralizing the FE-side rule in `DEFAULT_ENVIRONMENT_NAMES` keeps the deploy menu, traces filter, and configuration table consistent.

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-6937`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-6937`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-6937`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-6937`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-6937`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

awkoy (approved):

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-6937`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-6937`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

PR #6944, opened 2026-06-01
[OPIK-6652] [BE] fix: send Comet-Workspace header on Studio gateway LLM calls

dsblank:
## Details

Optimization Studio routes every LiteLLM completion through the Opik backend gateway (`OPENAI_API_BASE`, added in #6830). The gateway authenticates each request per workspace via the `Comet-Workspace` header, but LiteLLM's OpenAI-compatible path forwards only `Authorization` — so the gateway received no workspace and rejected every server-side optimization run with `403 "Workspace name should be provided"`. This wraps `litellm.completion`/`litellm.acompletion` in the optimizer runner to inject the workspace header.

- Header value is sourced from the job's workspace name (`context.workspace_name`), independent of the project name.
- Guarded on `OPENAI_API_BASE` so direct-provider calls are never touched; idempotent and won't override a caller-supplied `Comet-Workspace`.
- Covers both optimizer and metric-judge LLM calls (both resolve `litellm.completion`/`acompletion` at call time).

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-6652

## AI-WATERMARK

AI-WATERMARK: yes

- If yes:
  - Tools: Claude Code
  - Scope: Root-cause investigation, the fix in `optimizer_runner.py`, and the new unit tests.
  - Human verification: Author traced the gateway auth path (`RemoteAuthService`, `ChatCompletionsResource`) and LiteLLM header handling, reviewed the diff, and ran the test suite locally.

## Testing

Added `apps/opik-python-backend/tests/test_studio_gateway_workspace_header.py` (11 cases): header injection on sync + async completions, preserving existing `extra_headers`, not overriding an explicit `Comet-Workspace`, no-op when the gateway is inactive or workspace is empty, idempotency, and special-character workspace names (colon+slash, whitespace, non-ASCII, URL-reserved) passed through verbatim.

Commands run (Python 3.11 venv with `tests/test_requirements.txt` + `requirements.txt`):

```
cd apps/opik-python-backend
PYTHONPATH=src python -m pytest tests/test_studio_gateway_workspace_header.py tests/test_studio_project_name.py -q
# 11 passed (gateway header) + studio project-name tests pass
```

Not run: full backend/E2E cloud gateway run (requires a live Opik Cloud workspace + AI provider keys). The fix mirrors the exact header the SDK/playground already send (`Comet-Workspace`), verified against the LLM-gateway docs and `RemoteAuthService` auth path.

## Documentation

N/A — internal Optimization Studio gateway wiring; no user-facing API or docs change.

### DIFF
--- a/apps/opik-python-backend/src/opik_backend/jobs/optimizer_runner.py
+++ b/apps/opik-python-backend/src/opik_backend/jobs/optimizer_runner.py
@@ -9,6 +9,7 @@ All stdout/stderr from this script is captured by the parent process and
 streamed to Redis for S3 sync.
 """
 
+import functools
 import json
 import logging
 import os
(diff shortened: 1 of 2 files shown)


dsblank:
Thanks for the review — both points addressed in 10f1fff:

**1. Stale-workspace guard → dynamic update.** Replaced the early-return-if-wrapped guard. The wrappers now read the workspace from a shared mutable container (`_gateway_workspace`); `route_litellm_calls_through_gateway` updates that container on every call and installs the wrappers at most once (guarded by `_opik_gateway_wrapped`). So calling again with a different workspace updates the injected header in place — no stale value, no nested wrappers. New test `test_updates_workspace_without_rewrapping` asserts the wrapper object is unchanged while the injected `Comet-Workspace` reflects the new workspace.

**2. Module-level `OPENAI_API_BASE`.** Read once at import into a module constant near the other config (with a `logger.debug`, consistent with `TERMINAL_WIDTH`), and used in the guard instead of `os.environ.get(...)` per call. Mirrors the `OPIK_URL = os.getenv(...)` pattern in `studio/config.py`, and is safe because `optimizer.py` sets the env before spawning this subprocess. Tests now patch the module constant rather than the environment, with an autouse fixture isolating the shared state between tests.

One note for context: in production this subprocess handles a single job, so the workspace never actually changes within a process — the dynamic-update behavior is defensive correctness rather than a path exercised at runtime today. Happy to simplify back to a plain wrap-once if you'd prefer to keep it minimal.

`pytest tests/test_studio_gateway_workspace_header.py` → 12 passed.

alexkuzmik (approved):

PR #6949, opened 2026-06-02
[OPIK-6385] [FE] Pass thread time range to ThreadDetailsPanel traces query

YarivHashaiComet:
## Details

`ThreadDetailsPanel` (v2) was calling `useTracesList` filtered by `thread_id` without any time bounds. Since `thread_id` isn't part of the `traces` primary key (`workspace_id, project_id, id`), ClickHouse couldn't prune granules and had to scan the entire project — causing HTTP 500 read timeouts on large datasets (~4.7M traces in prod).

This change passes the thread's `start_time` / `end_time` as `fromTime` / `toTime` to `useTracesList`. The backend converts those into UUIDv7 bounds on the `id` column, letting ClickHouse seek directly to the matching primary-key range. Also gates the traces query on the thread being loaded (`enabled: Boolean(threadId) && Boolean(thread)`) so we never fire with `undefined` times.

- Frontend-only, v2 only (per ticket scope)
- Diff is 3 lines added, 1 changed in `ThreadDetailsPanel.tsx`
- The `useTracesList` hook already supports `fromTime`/`toTime` — no backend change needed
- The thread's `start_time`/`end_time` are already available from the `useThreadById` call above

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #6148
- OPIK-6385

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Scope: Implementation, verification, benchmark scaffolding
- Human verification: Author drove the test plan, confirmed UI behavior in browser, reviewed before/after benchmark output, and approved the final commit and PR.

## Testing

**End-to-end live UI verification on 5M traces (locally seeded, pre-OPIK-4828 conditions to match the original bug report on Opik 1.10.45):**

Same thread (`verify-bulk-thread`, 10 traces in a 1-second window inside a 5M-row project), same UI flow (click thread row in Threads tab), only the fix is toggled.

| Variant | CH duration | Rows read | Data read | URL has `from_time`/`to_time`? |
|---|---|---|---|---|
| Pre-fix | 265–288 ms | 55,006,513 | 1.03 GiB | No |
| Post-fix | 69–73 ms | 79,193 | 10.56 MiB | Yes |

- **~700× fewer rows scanned** (55M → 79K)
- **~100× less data read** (1.03 GiB → 10.56 MiB)
- **~3.5× faster** wall-clock on warm cache; cold-cache the gap is much larger
- At prod's 4.7M+ row projects: this is the difference between fast and timing out

**EXPLAIN evidence on 5M rows:**

```
WITHOUT time range (pre-fix):
  PrimaryKey Keys: workspace_id, project_id
  Granules: 615/615      ← scans every granule
  Skip (bloom filter): 355/615 (~42% pruning)

WITH time range (post-fix):
  PrimaryKey Keys: workspace_id, project_id, id  ← uses 3rd PK column
  Granules: 2/615        ← primary key prunes 99.7%
  Skip (bloom filter): 1/2
```

**Scenarios validated:**
- Happy path: click `verify-bulk-thread` (tight 1s window inside a 5M-row project) → fix produces dramatic speedup; all 10 traces returned correctly
- Negative test: confirmed that without the fix, the same click sends the `GET /traces/` request **without** `from_time`/`to_time` (reproduces the production bug exactly)
- Toggle test: ran with fix OFF (commented out the new lines + reverted `enabled` gate), reproduced the bug, then restored the fix and saw the speedup live in the same browser session

**Environment:** local dev-runner.sh (Dropwizard backend on :8080, Vite frontend on :5174, Docker ClickHouse/MySQL/Redis/etc.), macOS, Chrome 148.

**Not tested:**
- v1 ThreadDetailsPanel (per ticket scope — v2 only)
- Production timeout repro (locally we measured `read_bytes`/`read_rows` for the pruning evidence; the full timeout scenario requires prod-scale data)

## Documentation

No documentation update required — internal fix with no API or UX-visible change beyond load speed.

### DIFF
--- a/apps/opik-frontend/src/v2/pages-shared/traces/ThreadDetailsPanel/ThreadDetailsPanel.tsx
+++ b/apps/opik-frontend/src/v2/pages-shared/traces/ThreadDetailsPanel/ThreadDetailsPanel.tsx
@@ -497,7 +497,7 @@ const ThreadDetailsPanel: React.FC<ThreadDetailsPanelProps> = ({
   };
 
   const renderContent = () => {
-    if (isThreadPending || isTracesPending) {
+    if (isThreadPending) {
       return <Loader />;
     }
 
--- a/apps/opik-frontend/src/v2/pages-shared/traces/ThreadDetailsPanel/ThreadDetailsPanel.tsx
+++ b/apps/opik-frontend/src/v2/pages-shared/traces/ThreadDetailsPanel/ThreadDetailsPanel.tsx
@@ -194,9 +194,11 @@ const ThreadDetailsPanel: React.FC<ThreadDetailsPanelProps> = ({
       page: 1,
       size: 1000,
       truncate: false,
+      fromTime: thread?.start_time,
+      toTime: thread?.end_time,
     },
     {
-      enabled: Boolean(threadId),
+      enabled: Boolean(threadId) && Boolean(thread),
     },
   );
 


apps/opik-frontend/src/v2/pages-shared/traces/ThreadDetailsPanel/ThreadDetailsPanel.tsx
```diff
@@ -194,9 +194,11 @@ const ThreadDetailsPanel: React.FC<ThreadDetailsPanelProps> = ({
       page: 1,
       size: 1000,
       truncate: false,
+      fromTime: thread?.start_time,
+      toTime: thread?.end_time,
     },
     {
-      enabled: Boolean(threadId),
+      enabled: Boolean(threadId) && Boolean(thread),
```
YarivHashaiComet, line 201: **[high] Disabled traces query leaves isTracesPending stuck true, showing a perpetual Loader instead of NoData**

The traces query is now gated with `enabled: Boolean(threadId) && Boolean(thread)`. In React Query v5 a query that is disabled with no cached data reports `status: 'pending'`, so `isTracesPending` (from `isPending`) stays `true` whenever `thread` is falsy. `renderContent()` (line 500) and `renderBody()` (line 433) both check `isThreadPending || isTracesPending` BEFORE the `if (!thread) return <NoData />` branch (line 504). When the thread fetch fails or resolves with no data, `isThreadPending` becomes false and `thread` is undefined, but `isTracesPending` remains `true` forever because the query never enables. The panel then renders `<Loader />` indefinitely and the `NoData` fallback is unreachable. Before this change `enabled: Boolean(threadId)` let the traces query run independently, so `isTracesPending` resolved and `NoData` was reachable; this is a regression.

💡 Don't treat the disabled-traces pending state as a real loading state. Either compute a derived flag like `const isTracesLoading = Boolean(thread) && isTracesPending;` and use that in the guards, gate the traces-loading check after the `!thread` check, or use `fetchStatus`/`isFetching` (or `isLoading`, which is `isPending && isFetching`) so a disabled query is not mistaken for loading.

_rule: `frontend-correctness-gate-async-derived-state-on-the-query-s-loading-status-never`_

YarivHashaiComet (commented):

1 findings — 1 high · 0 medium · 0 low. Suppressed by team conventions: see `suppressed.md`.

👍/👎 each comment — 👎 from a senior reviewer trains the engine to stop flagging it.

apps/opik-frontend/src/v2/pages-shared/traces/ThreadDetailsPanel/ThreadDetailsPanel.tsx
YarivHashaiComet: Fixed in Revision 2 (commit f30de91d12).

Took the third option you suggested ("gate the traces-loading check after the `!thread` check") in its simplest form — just removed `isTracesPending` from the outer `renderContent` guard. The inner `renderBody` already handles `isTracesPending` with its own `<Loader />`, so the outer check was redundant in the happy path and harmful in the failure path. This also matches v1's pattern (v1 uses `&&` for the analogous guard for the same reason).

Diff (one line):
```diff
   const renderContent = () => {
-    if (isThreadPending || isTracesPending) {
+    if (isThreadPending) {
       return <Loader />;
     }
     if (!thread) return <NoData />;
```

Verified live in browser on the 5M-trace test project:
- **State 4** — URL with nonexistent thread\_id → POST `/threads/retrieve` returns 404, no GET `/traces/` call fires, panel renders **NoData** (not stuck loader) ✅
- **State 3** — happy path: panel renders with all 10 trace messages; original PR fix intact (`from_time`/`to_time` still in the trace URL) ✅
- **State 2** — regular `thread-N` with ~5000 traces: panel renders with full chrome and 1000 trace rows, no stuck loader ✅

(State numbers refer to the 5-row scenario table I worked through locally — 1: both loading, 2: thread loaded + traces loading, 3: happy path, 4: thread errored/404, 5: thread returns null. The regression was in row 4.)

andriidudar (approved):
LGTM!

PR #6951, opened 2026-06-02
[OPIK-6694] [SDK] feat: preserve environment ownership on prompt migrate

alexkuzmik:
## Details

`opik migrate prompt` dropped each source version's environment ownership during version replay — the destination payload was built from a fixed allowlist that omitted `environments`. This carries the `environments` set through verbatim, inline on `create_prompt_version` (the backend accepts it on creation and moves env ownership atomically, so no follow-up PATCH is needed). Per-version audit entries now record `source_environments` / `target_environments`.

- `version_type` / mask round-trip is intentionally **out of scope**: `get_prompt_versions` filters `version_type = 'prompt_version'` server-side, so masks are never read by the replay loop, and there is no endpoint to enumerate masks per prompt. Carrying masks would require separate backend work and is noted on the ticket.
- The schema referenced in the ticket (singular `environment` column) has since moved to a multi-environment `environments` set backed by `prompt_version_envs` (OPIK-6626), so "preserve environment" maps to carrying `environments`.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- OPIK-6694

## AI-WATERMARK

AI-WATERMARK: yes

- If yes:
  - Tools: Claude Code
  - Scope: full implementation
  - Human verification: code review + unit tests

## Testing

- Unit: `pytest tests/unit/cli/test_migrate_prompt_executor.py` — 8 passed, including 2 new cases: `test_replay_versions__carries_environments_verbatim` (env-owning version carries `environments`, env-less version passes `None`) and `test_replay_versions__records_environments_in_audit`.
- Lint/type: SDK pre-commit hooks (`ruff`, `ruff format`, `mypy`) pass on all changed files.
- E2E: added `test_environment_ownership_round_trips` (seeds a prompt whose latest version owns an environment, runs the CLI, asserts exactly that version owns the env on the destination and the audit records it). Requires a live backend — not executed locally; will run in CI.

## Documentation

N/A

### DIFF
--- a/sdks/python/src/opik/cli/migrate/prompts/version_replay.py
+++ b/sdks/python/src/opik/cli/migrate/prompts/version_replay.py
@@ -178,8 +178,8 @@ def replay_all_prompt_versions(
                 "source_commit": commit,
                 "target_version_id": new_version_id,
                 "target_commit": getattr(created, "commit", None),
-                "source_environments": sorted(environments) if environments else None,
-                "target_environments": sorted(target_environments)
+                "source_environments": environments if environments else None,
+                "target_environments": target_environments
                 if target_environments
                 else None,
             },
--- a/sdks/python/src/opik/cli/migrate/prompts/version_replay.py
+++ b/sdks/python/src/opik/cli/migrate/prompts/version_replay.py
@@ -8,8 +8,9 @@ the destination prompt has a fresh id.
 
 Stays on the low-level ``OpikApi`` (Fern) client so we can read every
 field the BE persists per version (``template``, ``metadata``, ``type``,
-``commit``, ``change_description``, ``tags``, ``template_structure``)
-without going through the high-level ``opik.api_objects.prompt`` wrapper.
+``commit``, ``change_description``, ``tags``, ``environments``,
+``template_structure``) without going through the high-level
+``opik.api_objects.prompt`` wrapper.
 Every call site is wrapped with
 ``ensure_rest_api_call_respecting_rate_limit``.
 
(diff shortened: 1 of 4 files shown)


sdks/python/tests/unit/cli/test_migrate_prompt_executor.py
```diff
@@ -169,6 +170,55 @@ def test_replay_versions__preserves_commit_hashes_verbatim(self) -> None:
...
+            commit="aaaaaaaa",
+            template="hello",
+        )
+        v2 = _PromptVersionRow(
+            id="src-v-2",
+            prompt_id="src-1",
+            commit="bbbbbbbb",
+            template="hi",
+            environments=["production", "staging"],
+        )
```
petrotiurin, line 198: For completeness, can we add another version in between these with a single env, "development"?
This way we test that environments from non-latest versions are preserved and multiple versions can have envs (and also that single env works).

sdks/python/src/opik/cli/migrate/prompts/version_replay.py
```diff
@@ -166,6 +178,10 @@ def replay_all_prompt_versions(
                 "source_commit": commit,
                 "target_version_id": new_version_id,
                 "target_commit": getattr(created, "commit", None),
+                "source_environments": sorted(environments) if environments else None,
+                "target_environments": sorted(target_environments)
```
petrotiurin: Any reason we're sorting the environments here? I don't think we rely on order anywhere and UI has it's own rendering logic.

petrotiurin (approved):
Thanks for fixing this!

PR #6963, opened 2026-06-03
[OPIK-6735] [BE] perf: drop O(n) ROW_NUMBER CTE in findLatestVersionsByDatasetIds

LifeXplorer:
## Details

`DatasetVersionDAO.findLatestVersionsByDatasetIds` (datasets-retrieve path) computed `version_name` via `ROW_NUMBER() OVER (PARTITION BY dataset_id ORDER BY id)` over every version of each dataset, materializing an on-disk temp table that exhausted RDS `FreeLocalStorage` (~6h EON Next outage at 94k versions). The `latest`-tagged version is always the max-id (UUIDv7) row, so its sequence number equals the dataset's total version count.
- Identify the latest row via the `dataset_version_tags` PK join (`tag = 'latest'`).
- Compute `version_name` with a `COUNT(*) … GROUP BY dataset_id` subquery served by the covering index `idx_dataset_versions_workspace_id_dataset_id_id` — no window, no sort of the full version set, no temp table.
- Behavior preserved: datasets without a `latest` tag still excluded, `is_latest = true`, all tags returned. (A `MAX(id)` subquery alone is insufficient — this method must still return `version_name`, which needs the count.)

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- OPIK-6735
- Related (ClickHouse growth, separate): OPIK-6787

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: query rewrite in `DatasetVersionDAO`, regression test, live read-only DB verification
- Human verification: author reviewed the diff; tests run locally; results validated against live EON Next data

## Testing

- `mvn test -Dtest='DatasetVersionResourceTest'` → 113 tests, 0 failures. New regression test `DatasetVersionResourceTest#getDatasets__whenManyVersions__thenLatestVersionNameTagsAndIdCorrect` added and confirmed green against **both** the old and new query implementations.
- Live EON Next (read-only `agentro` user): old vs new queries returned **byte-identical** results across two load profiles — version-heavy (5 datasets / 4,320 versions) and dataset-wide (31 datasets / 120 versions).
- `EXPLAIN ANALYZE` on EON Next: old plan does `Sort` + `row_number()` window + full-CTE materialize; new plan uses a covering index range scan for the count (no sort/window/temp), removing the local-storage spill.
- Environment: local Docker testcontainers (MySQL/ClickHouse/Redis/ZK) for the test suite.

## Documentation


apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetVersionDAO.java
```diff
@@ -380,19 +380,14 @@ Optional<DatasetVersion> findByVersionName(@Bind("dataset_id") UUID datasetId,
     int updateChangeDescription(@Bind("id") UUID id, @Bind("change_description") String changeDescription,
             @Bind("last_updated_by") String lastUpdatedBy, @Bind("workspace_id") String workspaceId);
 
+    // 'latest' is always the max-id version, so its version number is the dataset's total count;
+    // COUNT avoids a ROW_NUMBER() window over every version (on-disk temp table at scale).
     @SqlQuery("""
```
andrescrz: Nit: these comments better as javadoc.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetVersionDAO.java
```diff
@@ -408,12 +403,21 @@ WITH version_sequences AS (
...
+                ON dvt.workspace_id = dv.workspace_id
+                AND dvt.dataset_id = dv.dataset_id
+                AND dvt.version_id = dv.id
+                AND dvt.tag = 'latest'
             INNER JOIN (
+                SELECT dataset_id, COUNT(*) AS version_count
+                FROM dataset_versions
+                WHERE workspace_id = :workspace_id AND dataset_id IN (<dataset_ids>)
+                GROUP BY dataset_id
+            ) AS vc ON vc.dataset_id = dv.dataset_id
```
andrescrz: The new approach based on count and group by looks better than the old one. 

However, my main doubt is about introducing a new JOIN here vs resolving in a subquery like before.

What's more performant? Have you checked the comparison of both query plans against a big database (production for example).

I recommend running the query plans before moving forward.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
```diff
@@ -1914,6 +1914,48 @@ void getDatasets__whenVersioningEnabled__thenItemsCountFromLatestVersion() {
...
         }
 
+        @Test
+        @DisplayName("Success: GET datasets list returns correct latest version across many versions and multiple datasets")
+        void getDatasets__whenManyVersions__thenLatestVersionNameTagsAndIdCorrect() {
+            // Given - two datasets with different, large-ish version counts so the latest
+            // version's name (= total version count) is unambiguous and an off-by-one or
+            // wrong-dataset join would be caught.
+            int dataset1Versions = 12;
+            int dataset2Versions = 5;
```
andrescrz: I think this test should be just functional to regress that the query changes work and make sure all logic in the updated query is covered by tests and there are no gaps.

But we shouldn't rely on a big number of items for testing this. For that, better go by checking the query plan.

I recommend reducing the number of versions here, so test is faster. Pick up some values such as 3 and 5 or similar.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
```diff
@@ -1914,6 +1914,48 @@ void getDatasets__whenVersioningEnabled__thenItemsCountFromLatestVersion() {
...
+            // Then - each dataset reports its own latest version: correct id, name and 'latest' tag
+            assertThat(dataset1.latestVersion()).isNotNull();
+            assertThat(dataset1.latestVersion().id()).isEqualTo(latest1.id());
+            assertThat(dataset1.latestVersion().versionName()).isEqualTo("v" + dataset1Versions);
+            assertThat(dataset1.latestVersion().tags()).contains(DatasetVersionService.LATEST_TAG);
+
+            assertThat(dataset2.latestVersion()).isNotNull();
+            assertThat(dataset2.latestVersion().id()).isEqualTo(latest2.id());
+            assertThat(dataset2.latestVersion().versionName()).isEqualTo("v" + dataset2Versions);
+            assertThat(dataset2.latestVersion().tags()).contains(DatasetVersionService.LATEST_TAG);
```
andrescrz: Generally, I recommend asserting the full object: actual vs expected, instead of field by field.

In order to make tests more endurable.

After that, you can do particular additional exceptions, as needed by each specific test.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
```diff
@@ -1914,6 +1914,48 @@ void getDatasets__whenVersioningEnabled__thenItemsCountFromLatestVersion() {
...
+            var dataset1Id = createDataset(UUID.randomUUID().toString());
+            IntStream.range(0, dataset1Versions).forEach(i -> createDatasetItems(dataset1Id, 1));
+
+            var dataset2Id = createDataset(UUID.randomUUID().toString());
+            IntStream.range(0, dataset2Versions).forEach(i -> createDatasetItems(dataset2Id, 1));
+
+            var latest1 = getLatestVersion(dataset1Id);
+            var latest2 = getLatestVersion(dataset2Id);
+            assertThat(latest1.versionName()).isEqualTo("v" + dataset1Versions);
+            assertThat(latest2.versionName()).isEqualTo("v" + dataset2Versions);
```
andrescrz, line 1936: Same comment about assertions.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
```diff
@@ -1914,6 +1914,48 @@ void getDatasets__whenVersioningEnabled__thenItemsCountFromLatestVersion() {
...
+
+            // When - the datasets list endpoint resolves the latest version per dataset
+            var datasetsPage = datasetResourceClient.getDatasets(TEST_WORKSPACE, API_KEY);
+
+            var dataset1 = datasetsPage.content().stream()
+                    .filter(d -> d.id().equals(dataset1Id)).findFirst()
+                    .orElseThrow(() -> new AssertionError("dataset1 not found in list"));
+            var dataset2 = datasetsPage.content().stream()
+                    .filter(d -> d.id().equals(dataset2Id)).findFirst()
+                    .orElseThrow(() -> new AssertionError("dataset2 not found in list"));
```
andrescrz: Nit: I'd express these as assertJ assertions: assertThat(datasetsPage.content()) etc. with some contains logic.

andrescrz (commented):
Let's verify the query plans and the tests before moving forward.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetVersionDAO.java
LifeXplorer: updated

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetVersionDAO.java
LifeXplorer: Good call! I ran the query plans against production. Picked the heaviest workspace in prod — 3 datasets, one with 6,986 versions (6,989 total).
EXPLAIN ANALYZE, warm cache, 3 iterations each (stable):
1. Derived-table JOIN + GROUP BY (initial PR change): ~18.5 ms, 1 covering-index range scan of all 6,989 versions → GROUP BY → materialize 3-row temp → join
2. Correlated subquery COUNT(*) (suggested update): ~7.2 ms, 3 latest rows × covering-index count lookup on idx_dataset_versions_workspace_id_dataset_id_id
3. Old ROW_NUMBER CTE (production code): ~73 ms, sort 6,989 rows + window + materialize CTE
Result: updated SQL query to use count subquery

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
LifeXplorer: reduced versions counts in test

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
LifeXplorer: updated

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
LifeXplorer: updated

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetVersionDAO.java
```diff
@@ -380,19 +380,24 @@ Optional<DatasetVersion> findByVersionName(@Bind("dataset_id") UUID datasetId,
...
             SELECT
                 dv.id,
                 dv.dataset_id,
                 dv.version_hash,
-                CONCAT('v', vs.seq_num) AS version_name,
+                CONCAT('v', (
+                    SELECT COUNT(*)
+                    FROM dataset_versions c
+                    WHERE c.workspace_id = dv.workspace_id
+                        AND c.dataset_id = dv.dataset_id
```
andrescrz, line 399: Minor: this inner query runs N times per the outer query, as there's `c.dataset_id = dv.dataset_id`. 

According to your test and query plan, this seems safe and unavoidable as the versions need to be counted per each dataset.

I believe this is fine and not a blocker, but highlight just in case you wanted to explore some alternative or run some other final test.

andrescrz (approved):
Just one more thought about the query, but LGTM. You're good to move forward.

PR #6978, opened 2026-06-05
fix: add Mistral cost tracking

aduverger:
## Details
This PR enables Mistral provider for cost tracking. 

Lots of Mistral's models cost were already provided by LiteLLM, but their 2 flagship models were missing: [Mistral Small 4](https://docs.mistral.ai/models/model-cards/mistral-small-4-0-26-03) (best for speed) and [Mistral Medium 3.5](https://docs.mistral.ai/models/model-cards/mistral-medium-3-5-26-04) (best for performance).

Also, because Mistral was missing from `CostService.java`, the costs were not tracked by SDKs.

## Change checklist
- [N/A] User facing
- [N/A] Documentation update

## AI-WATERMARK

AI-WATERMARK: yes

  - Tools: Claude-Code
  - Model: Opus-4.7
  - Scope: All files
  - Human verification: Yes

## Testing
How you tested

Include:
- Exact commands run (for example: `mvn test`, `npm run lint && npm run test`, `pytest ...`)
- Scenarios validated (happy path, edge cases, regressions)
- Environment/context used (local process mode vs Docker, OS/browser when relevant)
- Any tests not run, with reason
- Relevant logs/artifacts links if available


apps/opik-backend/src/main/resources/model_prices_overrides.json
```diff
@@ -122,5 +122,26 @@
...
+    "mistral/mistral-medium-3-5": {
+        "input_cost_per_token": 1.5e-06,
+        "litellm_provider": "mistral",
+        "max_input_tokens": 262144,
+        "max_output_tokens": 262144,
+        "max_tokens": 262144,
+        "mode": "chat",
+        "output_cost_per_token": 7.5e-06,
+        "source": "https://mistral.ai/pricing"
+    },
```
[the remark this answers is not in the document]
aduverger: Verified — the gap is real, but it's not specific to this PR and it's not reachable today for Mistral (`MISTRAL` isn't a `LlmProvider`, so the playground can't issue these requests). 

The actually-affected entry is the pre-existing override for `gemini/gemini-3-flash`.
Filing as a follow-up: extend `ModelCapabilities.loadCapabilities()` to merge `model_prices_overrides.json`, same pattern as `CostService.applyOverrides()`. 
Keeping this PR scoped to cost tracking.

apps/opik-backend/src/main/resources/model_prices_overrides.json
andrescrz, line 1: Thanks for the contribution!

Note: `model_prices_overrides.json` is automatically synchronized from the LiteLLM repository, so changes to this file shouldn't be made directly in Opik. Any updates to model pricing should be submitted upstream to LiteLLM instead, and they'll be picked up during the sync process.

Could you remove the changes to this file from the PR?

apps/opik-frontend/src/data/model_prices_and_context_window.json
andrescrz, line 1: Applies to this file as well.

andrescrz (changes requested):
Hi @aduverger 

Thanks for your contribution. 

Please see related comments and let me know if any question.

Regards

apps/opik-backend/src/main/resources/model_prices_overrides.json
aduverger: Thanks @andrescrz . I'm a bit confused, my understanding was that the sync workflow `.github/workflows/span_cost_upload_daily.yml` only curls + cps the main pricing file (lines 22–23). It never touches the overrides file. 
From `CostService.java:210` comments: _"Apply Opik-owned overrides that survive the daily LiteLLM sync workflow."_ 
And every existing override entry is also opik-owned: `gpt-oss` isn't upstream in LiteLLM either, yet Opik adds it via this exact file.

apps/opik-frontend/src/data/model_prices_and_context_window.json
aduverger: Good point! Reverted

apps/opik-backend/src/main/resources/model_prices_overrides.json
andrescrz: Sorry, ignore this comment. This doesn't apply for the override file.

andrescrz (approved):
Hi @aduverger 

We'll merge it shortly, and the change will be included in the next Opik release (v2.0.61), which we're currently expecting to ship on the next 24h approximately.

Apologies for the confusing feedback on the 1st review and thanks again for contributing to Opik!

Best

PR #6979, opened 2026-06-05
[OPIK-6666] [BE] feat: MCP OAuth 2.1 foundations (PR 1/7) — schema, config, domain, persistence, service

LifeXplorer:
## Details

**PR 1 of 7** splitting #6900 (MCP OAuth 2.1 Authorization Server) into smaller, shippable chunks per [review feedback](https://github.com/comet-ml/opik/pull/6900#pullrequestreview-4428853723) — working inner-foundations outward, schema first, deployment last. #6900 stays open as the full reference.

This PR is the **shared inner foundation** every OAuth resource depends on: schema + config + domain types + persistence + token-lifecycle service. It has **no REST surface, no guicey bundle, no `OpikApplication` wiring** — nothing is exposed and there is **zero behavior change**. The feature is gated by `MCP_OAUTH_ENABLED` (defaults `false`) and stays inert until the full chain lands.

### What's in the box

| Layer | Files |
|---|---|
| Schema | Liquibase migration `000080` — `mcp_oauth_clients` / `_codes` / `_tokens` on `db-app-state` (no seeded clients; all MCP hosts register via DCR in a later chunk) |
| Config | `McpOAuthConfig` (env-gated by `MCP_OAUTH_ENABLED`) + `config.yml` `mcpOAuth:` block + `OpikConfiguration.mcpOAuth` field |
| Constants / models | `OAuthConstants`, `McpOAuthClient`, `McpOAuthCode`, `McpOAuthToken`, `McpOAuthTokens` (token utils), `ValidatedToken`, `CreateOAuthCodeCommand`, `TokenResponse`, `ClientRegistrationRequest` |
| Persistence | JDBI DAOs `McpOAuthClientDAO` / `McpOAuthCodeDAO` / `McpOAuthTokenDAO` — atomic mark-used, refresh rotation, revoke-family |
| Strategy + service | `OAuthClientStrategy` seam + `DbOAuthClientStrategy` + `OAuthClientService`; `McpOAuthService` — code exchange (PKCE/S256 verify), refresh rotation w/ reuse detection + grace, revoke, validate |

### Deliberately deferred

- **`MutableJobConfiguration` switch** on `OpikConfiguration` and the **`config.yml` `jobs.mcpOAuthScrubCron`** entry → move with the scrub job (PR 6), which is the only thing that needs them.
- All REST resources, the `McpOAuthBundle` kill-switch, `AuthFilter`/`AuthService` integration, tests, and helm/nginx deployment → later chunks.

### The split (7 PRs)

1. **Foundations** ← _this PR_
2. Metadata endpoint (`/.well-known/oauth-authorization-server`) + guicey bundle wiring
3. Authorize + consent (`/oauth/authorize`) + `AuthService` workspace resolution
4. Token + revoke (`/oauth/token`, `/oauth/revoke`)
5. Dynamic client registration (`/oauth/register`, RFC 7591)
6. Data-API bearer integration (`AuthFilter`) + validator (`/opik/auth-oauth`) + scrub job + E2E
7. Deployment (helm/nginx)

## Change checklist
- [ ] User facing
- [ ] Documentation update

Neither applies: this chunk has no REST surface and zero behavior change — nothing is user-facing and there is nothing to document until the endpoints land in later PRs.

## Testing

- `mvn -o compile` — green. Only out-of-set dependencies (`IdGenerator`, `SetFlatArgumentFactory`) already exist on `main`.
- No tests in this chunk by design — OAuth test suites land with their resources in PRs 3–6.

## Issues

- Jira: [OPIK-6666](https://comet-ml.atlassian.net/browse/OPIK-6666)
- Full reference: #6900
- Companion / dependency (lands before the feature is enabled): comet-ml/comet-backend#5569

## Documentation

N/A for this chunk — no user-facing surface. End-user OAuth setup/usage docs land with the deployment PR (7/7).


apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthService.java
```diff
@@ -0,0 +1,263 @@
...
+        if (row.revokedAt() != null) {
+            boolean benignRetry = REASON_ROTATED.equals(row.revokedReason())
+                    && !now.isAfter(row.revokedAt().plus(config().getRefreshRotationGrace()));
+            if (!benignRetry) {
+                // Reuse detected: kill the whole lineage. Must run in its own committed transaction —
+                // throwing invalid_grant from inside the same transaction would roll the revocation back.
+                template.inTransaction(WRITE, handle -> handle.attach(McpOAuthTokenDAO.class)
+                        .revokeFamily(row.familyId(), REASON_REUSE));
+            }
+            throw new BadRequestException(ERROR_INVALID_GRANT);
```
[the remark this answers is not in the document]
LifeXplorer: The intended outcome of the grace path is `invalid_grant`. The scenario it protects: request 1 rotates successfully (marks old `ROTATED`, mints+returns the new pair, commits); a racing retry (request 2) then sees `revokedAt != null`. Without the grace check, request 2 would classify it as reuse and `revokeFamily(...)`, killing the freshly-minted valid tokens request 1 just handed the legitimate client.
The grace window suppresses that false-positive escalation — nothing more.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthTokenDAO.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @SqlUpdate("UPDATE mcp_oauth_tokens SET revoked_at = NOW(6), revoked_reason = :reason "
+            +
+            "WHERE token_hash = :tokenHash AND revoked_at IS NULL")
+    int revoke(@Bind("tokenHash") String tokenHash, @Bind("reason") String reason);
+
+    // Reuse detection: revoke the entire refresh-token lineage in one indexed write.
+    @SqlUpdate("UPDATE mcp_oauth_tokens SET revoked_at = NOW(6), revoked_reason = :reason "
+            +
+            "WHERE family_id = :familyId AND revoked_at IS NULL")
```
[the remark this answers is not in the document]
LifeXplorer: This is a capability-token model: possession of the secret (whose `sha256` is `token_hash`) is the authorization. The caller doesn't supply a `workspace_id` to look up — the workspace is derived from the token row. So "validate the workspace before calling the DAO" is circular; there's no untrusted `workspaceId` flowing in.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000080_create_mcp_oauth_tables.sql
```diff
@@ -0,0 +1,72 @@
+--liquibase formatted sql
+--changeset avinahradau:000080_create_mcp_oauth_tables
+--comment: OAuth 2.1 Authorization Server tables for Opik MCP. Opaque tokens (sha256-hashed), PKCE S256, refresh-token rotation with family revocation. Keyed on user_name/workspace_name to match opik-backend's existing string-identifier convention; no scopes column (tokens grant the user's full workspace role, gated by @RequiredPermissions on each call).
+
+-- Registry of MCP host apps allowed to request tokens. Populated by Dynamic Client Registration
+-- (RFC 7591) when remote hosts self-register, and by the admin endpoint for manual additions.
+CREATE TABLE mcp_oauth_clients
+(
+    client_id        VARCHAR(150)  NOT NULL,
```
andrescrz: This is to facilitate my review: please clarify how does this ID look like:
- Type of ID: UUID or some other type (and what).
- If UUID, which version: 4 or 7.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000080_create_mcp_oauth_tables.sql
```diff
@@ -0,0 +1,72 @@
...
+    name             VARCHAR(255)  NOT NULL,
+    redirect_uris    JSON          NOT NULL,
+    logo_uri         VARCHAR(2048) NULL,
+    owner_user_name  VARCHAR(100)  NULL,
```
andrescrz: For all users, we generally go `VARCHAR(255)`, that includes `owner_user_name`, `created_by` etc. and the rest in this PR in this table or others.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000080_create_mcp_oauth_tables.sql
```diff
@@ -0,0 +1,72 @@
+--liquibase formatted sql
+--changeset avinahradau:000080_create_mcp_oauth_tables
+--comment: OAuth 2.1 Authorization Server tables for Opik MCP. Opaque tokens (sha256-hashed), PKCE S256, refresh-token rotation with family revocation. Keyed on user_name/workspace_name to match opik-backend's existing string-identifier convention; no scopes column (tokens grant the user's full workspace role, gated by @RequiredPermissions on each call).
+
+-- Registry of MCP host apps allowed to request tokens. Populated by Dynamic Client Registration
+-- (RFC 7591) when remote hosts self-register, and by the admin endpoint for manual additions.
+CREATE TABLE mcp_oauth_clients
```
andrescrz, line 7: Take a look at the query patterns of this table on it's DAO and any others you could refer (in joins etc.) I believe you're missing indexing on:
- revoked_at
- created_at

Make sure you pay attention not only to WHERE clauses, also to JOIN, GROUP and SORTING. 
Run query plans to solve any potential doubt. 

**This comment applies to all newly defined tables, other indexes are likely missing.**

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000080_create_mcp_oauth_tables.sql
```diff
@@ -0,0 +1,72 @@
...
+    workspace_id          VARCHAR(150)  NOT NULL,
+    code_challenge        VARCHAR(128)  NOT NULL,
+    code_challenge_method ENUM('S256')  NOT NULL DEFAULT 'S256',
+    redirect_uri          VARCHAR(2048) NOT NULL,
+    resource              VARCHAR(2048) NOT NULL,
+    created_at            TIMESTAMP(6)  NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
+    expires_at            TIMESTAMP(6)  NOT NULL,
+    used_at               TIMESTAMP(6)  NULL,
+
+    PRIMARY KEY (code_hash),
```
andrescrz: I suggest not having the hash as the primary key, as it's a obscure and difficult to handle and reason about. 

I agree this has to be enforced to be unique, but better have a separated typical ID (UUID or other) as the primary key.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000080_create_mcp_oauth_tables.sql
```diff
@@ -0,0 +1,72 @@
...
+  DEFAULT CHARSET = utf8mb4;
+
+-- One-time authorization codes. sha256(code) is the PK;
+CREATE TABLE mcp_oauth_codes
+(
+    code_hash             CHAR(64)      NOT NULL,
+    client_id             VARCHAR(150)  NOT NULL,
+    user_name             VARCHAR(100)  NOT NULL,
+    workspace_name        VARCHAR(255)  NOT NULL,
+    workspace_id          VARCHAR(150)  NOT NULL,
```
andrescrz: Take a look at other tables, but we generally define `workspace_id` as `VARCHAR(255)` due to legacy data.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000080_create_mcp_oauth_tables.sql
```diff
@@ -0,0 +1,72 @@
...
+    code_challenge        VARCHAR(128)  NOT NULL,
+    code_challenge_method ENUM('S256')  NOT NULL DEFAULT 'S256',
+    redirect_uri          VARCHAR(2048) NOT NULL,
+    resource              VARCHAR(2048) NOT NULL,
+    created_at            TIMESTAMP(6)  NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
```
andrescrz, line 33: Let's also add the `created_by` metadata. Ok to not to have the updat counter part fields if this is append-only, but fee free to add them if you prefer.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000080_create_mcp_oauth_tables.sql
```diff
@@ -0,0 +1,72 @@
...
+(
+    token_hash      CHAR(64)               NOT NULL,
+    type            ENUM('access','refresh') NOT NULL,
+    client_id       VARCHAR(150)           NOT NULL,
+    user_name       VARCHAR(100)           NOT NULL,
+    workspace_name  VARCHAR(255)           NOT NULL,
+    workspace_id    VARCHAR(150)           NOT NULL,
+    resource        VARCHAR(2048)          NOT NULL,
+    family_id       CHAR(36)               NOT NULL,
+    rotated_from    CHAR(64)               NULL,
```
andrescrz: Nit: `rotated_from_hash`

apps/opik-backend/config.yml
andrescrz, line 1: Please propagate these changes to the test config YML file. All should be identical including comments, except for:
- Not having env vars.
- Some default values can be adjusted globally for tests.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/ClientRegistrationRequest.java
```diff
@@ -0,0 +1,12 @@
+package com.comet.opik.domain.mcpoauth;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import lombok.Builder;
+
+import java.util.Set;
+
+// RFC 7591 §2 client metadata; unknown fields are ignored.
+@Builder
```
andrescrz: Please make sure all the records, POJOs etc. defined in this PR not only define Builder, but also with toBuilder.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/ClientRegistrationRequest.java
```diff
@@ -0,0 +1,12 @@
...
+@JsonIgnoreProperties(ignoreUnknown = true)
+public record ClientRegistrationRequest(String clientName, Set<String> redirectUris, String logoUri) {
```
andrescrz: Please make sure all request records, POJOs in this PR validate all input data appropriately:
- NotNull, NotBlank, size, limits etc.
- Applies for scalars, collections and their inner payload.
- Use custom validators based on annotations for more complex validations (accros fields etc.)

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/CreateOAuthCodeCommand.java
```diff
@@ -0,0 +1,14 @@
+package com.comet.opik.domain.mcpoauth;
+
+import lombok.Builder;
+
+@Builder(toBuilder = true)
+public record CreateOAuthCodeCommand(
```
andrescrz, line 7: Please make sure all the response objects include only non-null (or NON_EMPTY) as appropiately.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/CreateOAuthCodeCommand.java
```diff
@@ -0,0 +1,14 @@
+package com.comet.opik.domain.mcpoauth;
+
+import lombok.Builder;
+
+@Builder(toBuilder = true)
```
andrescrz, line 6: The Dropwizard global response JSON serialiser uses SnakeCaseStrategy, however, make sure all your objects define it if used by JSON mapper outside. Double check about JsonUtils.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/DbOAuthClientStrategy.java
```diff
@@ -0,0 +1,100 @@
...
+
+import static com.comet.opik.infrastructure.db.TransactionTemplateAsync.READ_ONLY;
+import static com.comet.opik.infrastructure.db.TransactionTemplateAsync.WRITE;
+
+@Singleton
+@RequiredArgsConstructor(onConstructor_ = @Inject)
+class DbOAuthClientStrategy implements OAuthClientStrategy {
+
+    private final @NonNull TransactionTemplate template;
+    private final @NonNull IdGenerator idGenerator;
```
andrescrz: This IDGenerator is based on UUID v7 which is based on time. This is ok for analytical data such as traces, spans etc. but it might not be ok for auth data. You should likely use UUID v4 in that case. 

Please review all usages and update accordingly.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/DbOAuthClientStrategy.java
```diff
@@ -0,0 +1,100 @@
...
+    public McpOAuthClient register(@NonNull ClientRegistrationRequest request) {
+        validate(request);
+
+        String clientId = idGenerator.generateId().toString();
+        var client = McpOAuthClient.builder()
+                .clientId(clientId)
+                .name(StringUtils.defaultIfBlank(request.clientName(), clientId))
+                .redirectUris(request.redirectUris())
+                .logoUri(request.logoUri())
+                .build();
```
andrescrz: Please automate this sort of mapping logic with `Mapper` from map-struct library.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/DbOAuthClientStrategy.java
```diff
@@ -0,0 +1,100 @@
...
+        });
+    }
+
+    // Field caps mirror the mcp_oauth_clients column sizes so oversized DCR metadata fails as a
+    // proper invalid_client_metadata 400 instead of a raw SQL error.
+    private static final int MAX_REDIRECT_URIS = 10;
+    private static final int MAX_NAME_LENGTH = 255;
+    private static final int MAX_URI_LENGTH = 2048;
+
+    private void validate(ClientRegistrationRequest request) {
```
andrescrz: This validation logic (and the `validate` call above) is not the responsibility of this class. You should abstract that to the record definition (with Jakarta) and define a custom validator for the complex cases.

This should be validated when the request is received, so the rest of the service takes that as a guarantee.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/DbOAuthClientStrategy.java
```diff
@@ -0,0 +1,100 @@
...
+                }
+            } catch (URISyntaxException e) {
+                throw new BadRequestException("invalid redirect_uri: " + redirectUri);
+            }
+        }
+    }
+
+    // Opaque = no URI scheme (UUIDs, DCR-minted ids, ...). Anything with a scheme is a URL-form id that a
+    // CIMD strategy would own; the https-only / SSRF gate lives there, not here. Scheme-agnostic and
+    // case-insensitive, so we never need to enumerate or prefix-match protocols.
```
andrescrz: Nit: better as javadoc.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthClientDAO.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.Optional;
+
+@RegisterConstructorMapper(McpOAuthClient.class)
+@RegisterArgumentFactory(SetFlatArgumentFactory.class)
+@RegisterColumnMapper(SetFlatArgumentFactory.class)
+interface McpOAuthClientDAO {
+
+    @SqlUpdate("INSERT INTO mcp_oauth_clients (client_id, name, redirect_uris, logo_uri, owner_user_name) "
+            +
+            "VALUES (:bean.clientId, :bean.name, :bean.redirectUris, :bean.logoUri, :bean.ownerUserName)")
```
andrescrz: Nit: please use java text blocks with triple quotes instead of concating strings, when defining large blocks of text e.g: all these SQL queries.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthClientDAO.java
```diff
@@ -0,0 +1,37 @@
...
+    void save(@BindMethods("bean") McpOAuthClient client);
+
+    @SqlQuery("SELECT * FROM mcp_oauth_clients WHERE client_id = :clientId AND revoked_at IS NULL")
+    McpOAuthClient findActiveById(@Bind("clientId") String clientId);
+
+    @SqlQuery("SELECT * FROM mcp_oauth_clients ORDER BY created_at DESC")
+    List<McpOAuthClient> findAll();
```
andrescrz: This method should be limited and paginated.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthClientDAO.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @SqlUpdate("UPDATE mcp_oauth_clients SET revoked_at = NOW(6) WHERE client_id = :clientId AND revoked_at IS NULL")
```
andrescrz: All update methods should update the `last_updated_by` field. I assume that `last_updated_at` is updated per the schema definition.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthClientDAO.java
```diff
@@ -0,0 +1,37 @@
...
+    int revoke(@Bind("clientId") String clientId);
+
+    default Optional<McpOAuthClient> fetchActive(String clientId) {
```
andrescrz: Nit: I believe if only the Optional flavour is used, you can directly define it as return type in `findActiveById`. That way we never return nulls here. I believe that works, but please double check.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthService.java
```diff
@@ -0,0 +1,263 @@
...
+import static com.comet.opik.domain.mcpoauth.OAuthConstants.TOKEN_TYPE_BEARER;
+import static com.comet.opik.infrastructure.db.TransactionTemplateAsync.READ_ONLY;
+import static com.comet.opik.infrastructure.db.TransactionTemplateAsync.WRITE;
+
+@Singleton
+@RequiredArgsConstructor(onConstructor_ = @Inject)
+public class McpOAuthService {
+
+    private static final String REASON_ROTATED = "rotated";
+    private static final String REASON_REUSE = "reuse";
```
andrescrz: This seems like enum values. Not necessarily need to enforce it in the DB schema, but please consider it as an option as well.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthService.java
```diff
@@ -0,0 +1,263 @@
...
+                    .clientId(clientId)
+                    .userName(row.userName())
+                    .workspaceName(row.workspaceName())
+                    .workspaceId(row.workspaceId())
+                    .resource(row.resource())
+                    .familyId(familyId)
+                    .expiresAt(now.plus(config().getRefreshTokenTtl()))
+                    .build());
+
+            return buildTokenResponse(accessToken, refreshToken, row.workspaceId(), row.workspaceName());
```
andrescrz, line 96: This seems outside the scope of the transaction.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthService.java
```diff
@@ -0,0 +1,263 @@
...
+        McpOAuthCode row = template.inTransaction(WRITE, handle -> {
+            var codeDao = handle.attach(McpOAuthCodeDAO.class);
+            if (codeDao.markUsed(codeHash) != 1) {
+                throw new BadRequestException(ERROR_INVALID_GRANT);
+            }
+            return codeDao.findByHash(codeHash);
+        });
+
+        if (!row.clientId().equals(clientId) || !row.redirectUri().equals(redirectUri)
+                || !verifyPkce(codeVerifier, row.codeChallenge())) {
```
andrescrz: Let's encapsulate all these validation conditionals in small aux methods with significant names, and with javadocs explaining them. It's hard to understand now.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthService.java
```diff
@@ -0,0 +1,263 @@
...
+
+        return template.inTransaction(READ_ONLY, handle -> {
+            McpOAuthToken row = handle.attach(McpOAuthTokenDAO.class).findByHash(tokenHash);
+
+            if (row == null || !TYPE_ACCESS.equals(row.type()) || row.revokedAt() != null
+                    || !row.expiresAt().isAfter(now)) {
+                return Optional.empty();
+            }
+
+            return Optional.of(new ValidatedToken(
```
andrescrz: Nit: favour builders over constructors.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthService.java
```diff
@@ -0,0 +1,263 @@
...
+            String workspaceName) {
+        return new TokenResponse(accessToken, refreshToken, TOKEN_TYPE_BEARER,
+                config().getAccessTokenTtl().toSeconds(), workspaceId, workspaceName);
+    }
+
+    private static boolean verifyPkce(String codeVerifier, String codeChallenge) {
+        String computed;
+        try {
+            byte[] digest = MessageDigest.getInstance("SHA-256")
+                    .digest(codeVerifier.getBytes(StandardCharsets.US_ASCII));
```
andrescrz, line 264: Double check the right charset here: `US_ASCII` vs `UTF_8` vs some other

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthService.java
```diff
@@ -0,0 +1,263 @@
...
+            computed = URL_ENCODER.encodeToString(digest);
+        } catch (NoSuchAlgorithmException e) {
+            throw new IllegalStateException("SHA-256 not available", e);
+        }
+        if (codeChallenge == null) {
+            return false;
+        }
```
andrescrz: This conditional seems that should be checked earlier.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthTokens.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.Base64;
+import java.util.HexFormat;
+
+import static com.comet.opik.domain.mcpoauth.OAuthConstants.BEARER_PREFIX;
+
+@UtilityClass
+public class McpOAuthTokens {
+
+    public static final String ACCESS_PREFIX = "opik_at_";
+    public static final String REFRESH_PREFIX = "opik_rt_";
```
andrescrz: Minor: let's see if we need a bit more significant prefixes: `opik_mcp_oauth_t_at` or similar.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthTokens.java
```diff
@@ -0,0 +1,73 @@
...
+    public static final int RANDOM_BYTES = 32;
+
+    private static final SecureRandom SECURE_RANDOM = new SecureRandom();
```
andrescrz: You should use proper factory methods such as `getInstanceStrong`.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthTokens.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    public static boolean isMcpOAuthToken(String authHeader) {
+        if (!Strings.CI.startsWith(authHeader, BEARER_PREFIX)) {
+            return false;
+        }
+        return isAccessToken(authHeader.substring(BEARER_PREFIX.length()).trim());
+    }
+
+    public static String hash(String token) {
```
andrescrz, line 54: This is repeated in the code base, we should centralise in a common place.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/OAuthClientService.java
```diff
@@ -0,0 +1,57 @@
...
+        if (allowed == null) {
+            return false;
+        }
+        return allowed.stream()
+                .anyMatch(candidate -> candidate.equals(requestedRedirectUri)
+                        || isLoopbackMatch(candidate, requestedRedirectUri));
+    }
+
+    // RFC 8252 7.3: native-app loopback redirects may use any port, so match scheme+host+path and ignore the port.
+    private static boolean isLoopbackMatch(String registered, String requested) {
```
andrescrz, line 39: Minor: investigate if any 1st or 3rd party library implements this kind of check and avoid re-inventing it.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/OAuthClientStrategy.java
```diff
@@ -0,0 +1,19 @@
...
+
+import com.google.inject.ImplementedBy;
+
+import java.util.Optional;
+
+/**
+ * Resolves OAuth client identity, keyed on the form of the {@code client_id}.
+ * The DB strategy handles opaque ids (DCR-registered + seeded).
+ */
+@ImplementedBy(DbOAuthClientStrategy.class)
```
andrescrz: Does this interface have multiple implementations at the moment? If not, define it in the same file as `DbOAuthClientStrategy`. 

In any case, I'd rather remove it and leave only the implementation.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/OAuthConstants.java
```diff
@@ -0,0 +1,35 @@
+package com.comet.opik.domain.mcpoauth;
+
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public class OAuthConstants {
```
andrescrz, line 6: Minor: investigate if any 1st or 3rd party already defines all these constants.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/McpOAuthConfig.java
```diff
@@ -0,0 +1,76 @@
...
+
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.time.Duration;
+
+@Data
+public class McpOAuthConfig {
+
+    @Valid @JsonProperty
+    private boolean enabled = false;
```
andrescrz: Let's remove all the defaults from this file, so we don't duplicate them with the YAML file and avoid maintaining them in sync.

You should ensure they're validated with NotNull, NotBlank etc. so they're always present.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/McpOAuthConfig.java
```diff
@@ -0,0 +1,76 @@
...
+    @Valid @JsonProperty
+    @NotNull private Duration registrationRateLimitDuration = Duration.ofHours(1);
+
+    public String getIssuer() {
+        return StringUtils.stripEnd(baseUrl, "/");
+    }
+
+    public String getMcpResourceUri() {
+        return StringUtils.isNotBlank(mcpResourceUri) ? mcpResourceUri : getIssuer() + "/api/v1/mcp";
+    }
```
andrescrz, line 51: Please double check if this is the appropriate place in the code base.

andrescrz (changes requested):
Many of the comments are mentioned once, but they apply to other parts of the code base. 

I still recommend breaking this down into smaller chunks, to progress and deliver faster.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000080_create_mcp_oauth_tables.sql
LifeXplorer: `client_id` was populated with UUID 7, updated it to UUID4. Updated column size to `VARCHAR(36)` in all tables.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000080_create_mcp_oauth_tables.sql
LifeXplorer: updated to `VARCHAR(255)`

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000080_create_mcp_oauth_tables.sql
LifeXplorer: updated to VARCHAR(255)

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000080_create_mcp_oauth_tables.sql
LifeXplorer: Added `id` to both code and token tables

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000080_create_mcp_oauth_tables.sql
LifeXplorer: I'd like to push back on this one. Codes are generated as part of OAuth flow. We already store `user_name` here and auth is supposed to happen on behalf of the user. No value from `created_by`

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000080_create_mcp_oauth_tables.sql
LifeXplorer: As we introduced `id`, I switched to `rotate_from_id`

apps/opik-backend/config.yml
LifeXplorer: added

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/ClientRegistrationRequest.java
LifeXplorer: added `toBuilder ` for all missing classes

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/ClientRegistrationRequest.java
LifeXplorer: `ClientRegistrationRequest` is the only external HTTP request record. Added validation

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/CreateOAuthCodeCommand.java
LifeXplorer: Updated, added `@JsonInclude(JsonInclude.Include.NON_NULL)`

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/CreateOAuthCodeCommand.java
LifeXplorer: Added explicit `@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class) ` to `ClientRegistrationRequest`. This is the only incoming object from resource level

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/DbOAuthClientStrategy.java
LifeXplorer: Makes sense. Switched all package to UUID 4.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/DbOAuthClientStrategy.java
LifeXplorer: Added `McpOAuthClientMapper` for this class auto mapping

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/DbOAuthClientStrategy.java
LifeXplorer: moved validation to annotations, removed that method

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/DbOAuthClientStrategy.java
LifeXplorer: updated

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthClientDAO.java
LifeXplorer: updated

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthClientDAO.java
LifeXplorer: Makes sense, updated

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthService.java
LifeXplorer: Shifted to JAVA enum, DB was not left as is with string values

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthService.java
LifeXplorer: extracted into separate methods, added javadoc for each

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthService.java
LifeXplorer: Changed to builders where possible

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthService.java
LifeXplorer: checked and added to java doc

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthService.java
LifeXplorer: Removed the check and annotated params as `NotNull`

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthTokens.java
LifeXplorer: updated to `opik_mcp_at_ / opik_mcp_rt_`

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthTokens.java
LifeXplorer: updated to factory method

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthTokens.java
LifeXplorer: Removed the method in favor of DigestUtils.sha256Hex.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/OAuthClientService.java
LifeXplorer: The project has zero OAuth/OIDC dependencies. No Nimbus, Spring Security/Authorization Server, pac4j, or jose — direct or transitive. Dragging in any up-mentioned 3rd party library is too heavy for few specific methods. Leaving without changes

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/OAuthClientStrategy.java
LifeXplorer: The interfaces leaves space for implementing CIMD strategy, which is trending now, but not yet supported by all major AI providers. Updating with approach of defining implementation it in the same file as `DbOAuthClientStrategy`, leaving the interface

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/OAuthConstants.java
LifeXplorer: Same as at up-mentioned comment (https://github.com/comet-ml/opik/pull/6979#discussion_r3372010947). Nothing to reuse, leaving unchanged.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/McpOAuthConfig.java
LifeXplorer: moved all defaults to yaml

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/McpOAuthConfig.java
LifeXplorer: IMO those methods make sense to keep in config.
I found plenty of examples of other configs exposing custom get methods, e.g. `FreeModelConfig`, `ExperimentProjectMigrationConfig` and many others

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000080_create_mcp_oauth_tables.sql
```diff
@@ -0,0 +1,72 @@
...
+
+-- Registry of MCP host apps allowed to request tokens. Populated by Dynamic Client Registration
+-- (RFC 7591) when remote hosts self-register.
+CREATE TABLE mcp_oauth_clients
+(
+    client_id        VARCHAR(36)  NOT NULL,
+    name             VARCHAR(255)  NOT NULL,
+    redirect_uris    JSON          NOT NULL,
+    logo_uri         VARCHAR(2048) NULL,
+    owner_user_name  VARCHAR(255)  NULL,
```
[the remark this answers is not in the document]
LifeXplorer: rollback is included. Reviewer mistake

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000080_create_mcp_oauth_tables.sql
```diff
@@ -0,0 +1,72 @@
...
+    code_challenge        VARCHAR(128)  NOT NULL,
+    code_challenge_method ENUM('S256')  NOT NULL DEFAULT 'S256',
+    redirect_uri          VARCHAR(2048) NOT NULL,
+    resource              VARCHAR(2048) NOT NULL,
+    created_at            TIMESTAMP(6)  NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
+    expires_at            TIMESTAMP(6)  NOT NULL,
+    used_at               TIMESTAMP(6)  NULL,
+
+    PRIMARY KEY (code_hash),
+    INDEX mcp_oauth_codes_expires_at_idx (expires_at)
```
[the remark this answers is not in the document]
LifeXplorer: indexes here are single-column self-explanatory

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthService.java
```diff
@@ -0,0 +1,268 @@
...
+        String accessToken = McpOAuthTokens.generateAccessToken();
+        String refreshToken = McpOAuthTokens.generateRefreshToken();
+
+        return template.inTransaction(WRITE, handle -> {
+            var tokenDao = handle.attach(McpOAuthTokenDAO.class);
+            tokenDao.save(McpOAuthToken.builder()
+                    .id(idGenerator.generateId().toString())
+                    .tokenHash(McpOAuthTokens.hash(accessToken))
+                    .type(TYPE_ACCESS)
+                    .clientId(clientId)
```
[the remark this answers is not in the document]
LifeXplorer: not purely duplicated. Keeping as is for readability

apps/opik-backend/src/main/java/com/comet/opik/api/validation/AbsoluteUriValidator.java
```diff
@@ -0,0 +1,23 @@
...
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+public class AbsoluteUriValidator implements ConstraintValidator<AbsoluteUri, String> {
+
+    @Override
+    public boolean isValid(String value, ConstraintValidatorContext context) {
+        if (value == null) {
+            return true; // Let @NotEmpty handle null/blank validation
```
andrescrz: Minor: the comment mentions that this shouldn't happen as Jakarta validation annotation should validate this first. But if it doesn't, shouldn't this return `false` as a defensive change?

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/ClientRegistrationRequest.java
```diff
@@ -0,0 +1,22 @@
...
+import com.fasterxml.jackson.databind.PropertyNamingStrategies;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import jakarta.validation.constraints.NotEmpty;
+import jakarta.validation.constraints.Size;
+import lombok.Builder;
+
+import java.util.Set;
+
+// RFC 7591 §2 client metadata; unknown fields are ignored.
+// Size caps mirror the mcp_oauth_clients column sizes.
```
andrescrz: Nit: all these comments through the PR better as javadocs.

Comment applies to the whole PR.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/ClientRegistrationRequest.java
```diff
@@ -0,0 +1,22 @@
...
+@Builder(toBuilder = true)
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
+public record ClientRegistrationRequest(
+        @Size(max = 255) String clientName,
```
andrescrz: Please review all these `@Size` validations not check if you need to add `@NotBlank` or `@NotNull` etc as `null` are considered valid.

Comment applies to the whole PR.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/CreateOAuthCodeCommand.java
```diff
@@ -0,0 +1,14 @@
...
+public record CreateOAuthCodeCommand(
+        String clientId,
```
andrescrz: For this public record, I recommend adding Lombok `@NoNull` to all thee fields that applies. 

I recommend doing this for all records/POJOs that aren't request/response on DW or JSON serialised/deserialised objects. Those are handled by Jakarta instead.

This comment applies to the whole PR.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/DbOAuthClientStrategy.java
```diff
@@ -0,0 +1,73 @@
...
+@Singleton
+@RequiredArgsConstructor(onConstructor_ = @Inject)
+class DbOAuthClientStrategy implements OAuthClientStrategy {
+
+    private final @NonNull TransactionTemplate template;
+
+    @Override
+    public boolean supports(@NonNull String clientId) {
+        return isOpaque(clientId);
+    }
```
andrescrz, line 41: Nit: the `supports` method just delegates to `isOpaque`. You could just leave the public method.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/DbOAuthClientStrategy.java
```diff
@@ -0,0 +1,73 @@
...
+    @Override
+    public McpOAuthClient register(@NonNull ClientRegistrationRequest request) {
+        String clientId = UUID.randomUUID().toString();
+        var client = McpOAuthClientMapper.INSTANCE.toClient(request, clientId);
+
+        return template.inTransaction(WRITE, handle -> {
+            var dao = handle.attach(McpOAuthClientDAO.class);
+            dao.save(client);
+            return dao.findActiveById(clientId)
+                    .orElseThrow(() -> new IllegalStateException("client not found after registration: " + clientId));
```
andrescrz: Double check this `IllegalStateException`. This is not exactly a service class, so it's fine to not to use `WebApplicationException` exceptions, but you should handled it and convert it to a `NotFoundException`. 

I'd be fine to directly use a `NotFoundException` here as this is close to a service class.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/DbOAuthClientStrategy.java
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * Opaque = no URI scheme (UUIDs, DCR-minted ids, ...). Anything with a scheme is a URL-form id that a
+     * CIMD strategy should own.
+     */
+    private static boolean isOpaque(String clientId) {
+        try {
+            return !new URI(clientId).isAbsolute();
+        } catch (URISyntaxException e) {
+            return true;
+        }
```
andrescrz: Minor: this method is copied and pasted from `AbsoluteUriValidator` (only negated). You could extract to an appropriate place and reuse.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthService.java
```diff
@@ -0,0 +1,323 @@
...
+                .clientId(cmd.clientId())
+                .userName(cmd.userName())
+                .workspaceName(cmd.workspaceName())
+                .workspaceId(cmd.workspaceId())
+                .codeChallenge(cmd.codeChallenge())
+                .codeChallengeMethod(CODE_CHALLENGE_METHOD_S256)
+                .redirectUri(cmd.redirectUri())
+                .resource(cmd.resource())
+                .expiresAt(now.plus(config().getCodeTtl()))
+                .build();
```
andrescrz: Minor: I believe still all these mapping occurrences better need a `@Mapper` class to automate this conversion.

You can always do a partial conversion from `McpOAuthTokens` to `McpOAuthCode` and then invoke `toBuilder` to set the remaining fields, such as `id`, `codeHash` etc. You can define aux methods to encapsulate that

You can see that this logic is repeated below, so it's desirable to centralise somewhere.

This comment applies everywhere in the PR.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthTokenDAO.java
```diff
@@ -0,0 +1,48 @@
...
+            WHERE token_hash = :tokenHash AND revoked_at IS NULL
+            """)
+    int revoke(@Bind("tokenHash") String tokenHash, @Bind("reason") RevokedReason reason);
+
+    // Reuse detection: revoke the entire refresh-token lineage in one indexed write.
+    @SqlUpdate("""
+            UPDATE mcp_oauth_tokens SET revoked_at = NOW(6), revoked_reason = :reason
+            WHERE family_id = :familyId AND revoked_at IS NULL
+            """)
+    int revokeFamily(@Bind("familyId") String familyId, @Bind("reason") RevokedReason reason);
```
andrescrz: Minor, you could consolidate these two methods in a single query, that tweaks the condition with `ST` templates. It'd be fine to have two overloaded methods `revoke` and `revokeFamily` reusing and delegating to common one.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000080_create_mcp_oauth_tables.sql
```diff
@@ -0,0 +1,73 @@
...
+    PRIMARY KEY (client_id)
+) ENGINE = InnoDB
+  DEFAULT CHARSET = utf8mb4;
+
+-- One-time authorization codes. id is the PK; sha256(code) is enforced unique for lookups.
+CREATE TABLE mcp_oauth_codes
+(
+    id                    CHAR(36)      NOT NULL,
+    code_hash             CHAR(64)      NOT NULL,
+    client_id             VARCHAR(36)   NOT NULL,
```
andrescrz: These ID fields are UUID v4, you should use CHAR type instead of VARCHAR

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000080_create_mcp_oauth_tables.sql
```diff
@@ -0,0 +1,73 @@
+--liquibase formatted sql
+--changeset avinahradau:000080_create_mcp_oauth_tables
+--comment: OAuth 2.1 Authorization Server tables for Opik MCP. Opaque tokens (sha256-hashed), PKCE S256, refresh-token rotation with family revocation. Keyed on user_name/workspace_name to match opik-backend's existing string-identifier convention; no scopes column (tokens grant the user's full workspace role, gated by @RequiredPermissions on each call).
+
+-- Registry of MCP host apps allowed to request tokens. Populated by Dynamic Client Registration
+-- (RFC 7591) when remote hosts self-register.
+CREATE TABLE mcp_oauth_clients
+(
+    client_id        VARCHAR(36)  NOT NULL,
```
andrescrz: Minor: in other tables where this is the foreign key, it should be called `client_id`. In this table as the primary key the convention is to be just `id`.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000080_create_mcp_oauth_tables.sql
```diff
@@ -0,0 +1,73 @@
...
+    user_name       VARCHAR(255)           NOT NULL,
+    workspace_name  VARCHAR(255)           NOT NULL,
+    workspace_id    VARCHAR(255)           NOT NULL,
+    resource        VARCHAR(2048)          NOT NULL,
+    family_id       CHAR(36)               NOT NULL,
+    rotated_from_id CHAR(36)               NULL,
+    issued_at       TIMESTAMP(6)           NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
+    expires_at      TIMESTAMP(6)           NOT NULL,
+    revoked_at      TIMESTAMP(6)           NULL,
+    revoked_reason  VARCHAR(64)            NULL,
```
andrescrz: Minor: as reason might be free text, I'd go with 255 like the other fields.

andrescrz (commented):
The PR mostly LGTM now. 

There's still one blocking comment in the MySQL table definitions regarding the type of client_id fields. We cannot change the migrations one pushed, I'd rather it address it now.

The rest of comments are optional. Many of them are nice to have, so I recommend tackling them as you have to send a new revision anyway. 

Many comments are reflected once, but applies to multiple places (like record validations).

Let's do a final iteration and this should be good to go.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/RevokedReason.java
```diff
@@ -0,0 +1,7 @@
+package com.comet.opik.domain.mcpoauth;
+
+public enum RevokedReason {
+    ROTATED,
+    REUSE,
+    CLIENT_REQUEST
```
andrescrz: I forgot mentioning, better define an inner value for the reason to be printed, so we don't rely on the enum constant value. See the existing conventions in the service, based on Lombok etc.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthService.java
andrescrz: Please double check this one, still happening @LifeXplorer

apps/opik-backend/src/main/java/com/comet/opik/api/validation/AbsoluteUriValidator.java
LifeXplorer: makes sense, updated

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/ClientRegistrationRequest.java
LifeXplorer: Iterated over all non-java doc comments

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/ClientRegistrationRequest.java
LifeXplorer: updated

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/CreateOAuthCodeCommand.java
LifeXplorer: added Lombok `@NonNull` to all objects that are not validated by Jakarta constraints

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/DbOAuthClientStrategy.java
LifeXplorer: this method body was updated due to refactoring

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/DbOAuthClientStrategy.java
LifeXplorer: Updated to `NotFoundException`

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/DbOAuthClientStrategy.java
LifeXplorer: refactored

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthService.java
LifeXplorer: moved to mapper

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthTokenDAO.java
LifeXplorer: refactored to reused base sql query

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000080_create_mcp_oauth_tables.sql
LifeXplorer: updated to CHAR

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000080_create_mcp_oauth_tables.sql
LifeXplorer: updated

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000080_create_mcp_oauth_tables.sql
LifeXplorer: updated

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/RevokedReason.java
LifeXplorer: added values to enum

andrescrz (approved):
LGTM. 

This was along and complex piece of work, with many moving pieces. But the implementation looks clean and solid.

Well done & thanks for putting it together!

PR #6993, opened 2026-06-08
[OPIK-6666] [BE] feat: MCP OAuth metadata endpoint + bundle wiring (PR 2/7)

LifeXplorer:
## Details

**PR 2 of 7** splitting #6900 per [review](https://github.com/comet-ml/opik/pull/6900#pullrequestreview-4428853723).

> ⛓️ **Chained PR.** Base is `avinahradau/OPIK-6666-oauth-as-chunked-1` (PR #6979). Review/merge PR-1 first; the base will be retargeted to `main` once PR-1 lands.

Introduces the guicey kill-switch bundle and the first OAuth resource — the RFC 8414 discovery endpoint. Everything is inert until `MCP_OAUTH_ENABLED=true`.

| Item | What |
|---|---|
| `OAuthMetadataResource` | `GET /.well-known/oauth-authorization-server` |
| `AuthorizationServerMetadata` | response DTO |
| `McpOAuthBundle` | `disableExtensions(...)` when `MCP_OAUTH_ENABLED=false` — currently only the metadata resource; grows one resource per chunk |
| `OpikApplication` | registers `McpOAuthBundle` |

The flag is checked inside `McpOAuthBundle#run` rather than gating bundle registration in `OpikApplication`, because the parsed configuration is only available at Dropwizard's run phase — bundles are registered during `initialize(Bootstrap)`, where no configuration exists yet. A javadoc on the bundle documents this. The metadata advertises `/oauth/token`, `/oauth/register`, `/oauth/revoke` (landing in PRs 4/5) — harmless while disabled.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- OPIK-6666
- Full reference: #6900

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: guicey bundle wiring, metadata resource/DTO, and explanatory javadoc
- Human verification: reviewed diff, `mvn -o compile` green, Spotless clean

## Testing
- `mvn -o compile` — green.
- Spotless apply — clean.

## Documentation

No documentation update — feature is flag-gated (`MCP_OAUTH_ENABLED`) and inert until enabled. End-user docs will land with the full OAuth flow in a later chunk.

[OPIK-6666]: https://comet-ml.atlassian.net/browse/OPIK-6666


apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/McpOAuthBundle.java
```diff
@@ -0,0 +1,26 @@
...
+ * off, which is the guicey-idiomatic way to express a config-conditional extension.
+ */
+public class McpOAuthBundle implements GuiceyBundle {
+
+    @Override
+    public void run(GuiceyEnvironment environment) {
+        OpikConfiguration config = environment.configuration();
+        if (!config.getMcpOAuth().isEnabled()) {
+            environment.disableExtensions(
+                    OAuthMetadataResource.class);
```
BorisTkachenko, line 23: In general this looks like on overhead. Usually we just check config toggle at the endpoint level and return `return Response.status(Response.Status.NOT_IMPLEMENTED).build();` if disabled. Later once it's enabled, this part could be deleted as a tech debt. This way we don't need dedicated test and all this wiring.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthMetadataResource.java
```diff
@@ -0,0 +1,48 @@
...
+@Path("/.well-known/oauth-authorization-server")
+@Produces(MediaType.APPLICATION_JSON)
+@Timed
+@RequiredArgsConstructor(onConstructor_ = @Inject)
+public class OAuthMetadataResource {
+
+    private final @NonNull OpikConfiguration opikConfig;
+
+    @GET
+    public Response metadata() {
```
BorisTkachenko, line 40: Swagger annotations are missing

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/McpOAuthBundle.java
LifeXplorer: This is valid consern in terms of current PR chunk. But for entire feature up to PR6, this bundle will contain 5 resources already and 1 cron job. I'd suggest to keep this bundle as a single config source

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthMetadataResource.java
LifeXplorer: added Swagger annotations

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthMetadataResource.java
```diff
@@ -0,0 +1,56 @@
...
+            @ApiResponse(responseCode = "200", description = "Authorization Server Metadata", content = @Content(schema = @Schema(implementation = AuthorizationServerMetadata.class)))})
+    public Response metadata() {
+        McpOAuthConfig config = opikConfig.getMcpOAuth();
+        String issuer = config.getIssuer();
+        return Response.ok(AuthorizationServerMetadata.builder()
+                .issuer(issuer)
+                .authorizationEndpoint(issuer + "/oauth/authorize")
+                .tokenEndpoint(issuer + "/oauth/token")
+                .revocationEndpoint(issuer + "/oauth/revoke")
+                .registrationEndpoint(issuer + "/oauth/register")
```
andrescrz, line 48: Nit: you might want to define those as constants as well (likely in OAuthConstants as well).

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthMetadataResource.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.List;
+
+import static com.comet.opik.domain.mcpoauth.OAuthConstants.AUTH_METHOD_NONE;
+import static com.comet.opik.domain.mcpoauth.OAuthConstants.CODE_CHALLENGE_METHOD_S256;
+import static com.comet.opik.domain.mcpoauth.OAuthConstants.GRANT_AUTHORIZATION_CODE;
+import static com.comet.opik.domain.mcpoauth.OAuthConstants.GRANT_REFRESH_TOKEN;
+import static com.comet.opik.domain.mcpoauth.OAuthConstants.RESPONSE_TYPE_CODE;
+
+@Path("/.well-known/oauth-authorization-server")
+@Produces(MediaType.APPLICATION_JSON)
```
andrescrz, line 29: Nit: consider adding the `Consumes` annotation for JSON for the future. The current endpoint doesn't use it and I'm assuming the RFC would consume JSON as well if an endpoint is ever added accepting a request object.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/McpOAuthBundle.java
andrescrz: Given that explanation (and I also like this approach), I consider this as not a blocker.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/McpOAuthBundleTest.java
```diff
@@ -0,0 +1,54 @@
...
+    private final McpOAuthBundle bundle = new McpOAuthBundle();
+
+    @Mock
+    private GuiceyEnvironment environment;
+
+    @Mock
+    private OpikConfiguration configuration;
+
+    @Mock
+    private McpOAuthConfig mcpOAuthConfig;
```
andrescrz, line 29: Nit: for POJOs/Records etc. generally better to just create real objects populated with what you need, instead of mocks.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/McpOAuthBundleTest.java
```diff
@@ -0,0 +1,54 @@
...
+        when(mcpOAuthConfig.isEnabled()).thenReturn(false);
+
+        bundle.run(environment);
+
+        verify(environment).disableExtensions(OAuthMetadataResource.class);
+    }
+
+    @Test
+    @DisplayName("keeps the metadata resource when MCP OAuth is on")
+    void keepsMetadataResourceWhenEnabled() {
```
andrescrz, line 45: Minor: these tests are the same, you can just parameterise the `isEnabled` value and the `times` value for the `verify` call (0 which is equivalent to never or 1)

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthMetadataResourceTest.java
```diff
@@ -0,0 +1,49 @@
...
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import static jakarta.ws.rs.core.Response.Status.OK;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+@DisplayName("OAuthMetadataResource")
+class OAuthMetadataResourceTest {
```
andrescrz, line 18: Minor: Given this is a simple resource, I like the approach of a lightweight test that doesn't require starting all dependent containers, the app etc. 

But maybe a DW integration test would be more realistic than an unit test:
- https://www.dropwizard.io/en/stable/manual/testing.html#testing-resources

You would do a real client call etc.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthMetadataResourceTest.java
```diff
@@ -0,0 +1,49 @@
...
+            var metadata = (AuthorizationServerMetadata) response.getEntity();
+            assertThat(metadata.issuer()).isEqualTo(issuer);
+            assertThat(metadata.authorizationEndpoint()).isEqualTo(issuer + "/oauth/authorize");
+            assertThat(metadata.tokenEndpoint()).isEqualTo(issuer + "/oauth/token");
+            assertThat(metadata.revocationEndpoint()).isEqualTo(issuer + "/oauth/revoke");
+            assertThat(metadata.registrationEndpoint()).isEqualTo(issuer + "/oauth/register");
+            assertThat(metadata.responseTypesSupported()).containsExactly("code");
+            assertThat(metadata.grantTypesSupported()).containsExactly("authorization_code", "refresh_token");
+            assertThat(metadata.codeChallengeMethodsSupported()).containsExactly("S256");
+            assertThat(metadata.tokenEndpointAuthMethodsSupported()).containsExactly("none");
```
andrescrz, line 46: Not a blocker: please assert the full object actual vs expected instead of field by field. It makes the test more endurable if fields are added in the future and also more maintainable.

andrescrz (approved):
All comments are non-blocking. As they're easy to address, I still recommend fixing them in this or the next PRs. You're fine to move forward if you prefer.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthMetadataResource.java
LifeXplorer: Addressed the changes in the next chunk PR3 https://github.com/comet-ml/opik/pull/6994

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthMetadataResource.java
LifeXplorer: Addressed the changes in the next chunk PR3 https://github.com/comet-ml/opik/pull/6994

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/McpOAuthBundleTest.java
LifeXplorer: Addressed the changes in the next chunk PR3 https://github.com/comet-ml/opik/pull/6994

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/McpOAuthBundleTest.java
LifeXplorer: Addressed the changes in the next chunk PR3 https://github.com/comet-ml/opik/pull/6994

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthMetadataResourceTest.java
LifeXplorer: Addressed the changes in the next chunk PR3 https://github.com/comet-ml/opik/pull/6994

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthMetadataResourceTest.java
LifeXplorer: Addressed the changes in the next chunk PR3 https://github.com/comet-ml/opik/pull/6994

PR #6994, opened 2026-06-08
[OPIK-6666] [BE] feat: MCP OAuth authorize + consent + AuthService workspace resolution (PR 3/7)

LifeXplorer:
## Details

**PR 3 of 7** splitting #6900 per [review](https://github.com/comet-ml/opik/pull/6900#pullrequestreview-4428853723).

> ⛓️ **Chained PR.** Base is `avinahradau/OPIK-6666-oauth-as-chunked-2` (PR #6993). Merge the chain in order; base retargets to `main` once PR-2 lands.

| Item | What |
|---|---|
| `OAuthAuthorizeResource` | `GET /oauth/authorize` (validate → login/consent redirect), `GET /oauth/authorize/context` (consent state + CSRF), `POST /oauth/authorize` (consent → mint code) |
| `AuthorizeContext` / `ConsentRequest` / `ConsentResponse` | DTOs |
| `AuthService` | `listEligibleWorkspaces` + `authorizeWorkspace` (consent picker) and `authorizeOAuth` (consumed by the data-API filter in PR-6). OSS impl → default workspace; Cloud (`RemoteAuthService`) delegates to comet-backend |
| `UserWorkspace` / `WorkspaceInfo` | records |
| `McpOAuthBundle` | now also disables `OAuthAuthorizeResource` when flag off |
| `OAuthAuthorizeResourceTest` / `RemoteAuthServiceTest` / `AuthServiceImplTest` | coverage |

All three new `AuthService` methods land together so the interface compiles once; `authorizeOAuth` is unused until PR-6 wires `AuthFilter`.

### Review follow-ups (latest revisions)
- **DTO serialization** — annotated resource-exposed records (`AuthorizeContext`, `ConsentRequest`, `ConsentResponse`, `WorkspaceInfo`) with `@JsonNaming(SnakeCaseStrategy)` + `@JsonInclude(NON_NULL)` and `@Builder(toBuilder = true)`. No wire-format change (the global Dropwizard mapper already applies both); this aligns the generated OpenAPI spec with the actual JSON and matches the repo convention.
- **Constants** — moved the `mcp_oauth_csrf` cookie name into `OAuthConstants.CSRF_COOKIE`; `OAuthAuthorizeResourceTest` now reuses `OAuthConstants` (`CSRF_COOKIE`, `RESPONSE_TYPE_CODE`, `CODE_CHALLENGE_METHOD_S256`, error codes) instead of literals.
- **CSRF token generation** — `OAuthAuthorizeResource` now reuses `McpOAuthTokens.randomToken()` (strong `SecureRandom`, URL-safe Base64) instead of a local `new SecureRandom()`, consolidating secure-random generation into one place.
- **`redirect_uri` fragment** — reject a `redirect_uri` containing a `#` fragment per RFC 6749 §3.1.2 (otherwise appended `code`/`error`/`state` would land inside the fragment, unreadable by the client).
- **Bean validation** — `consent` body is `@NotNull @Valid ConsentRequest` with `@NotBlank` on the required PKCE/client fields; `client_id`/`redirect_uri` on the GET endpoints are `@NotBlank` (protocol params keep redirect-based OAuth error reporting in-method).
- **`AuthServiceImpl`** — hoisted the OSS default `eligibleWorkspaces` / `authorizedWorkspace` into static constants.
- **Tests** — added coverage for the new `RemoteAuthService` and `AuthServiceImpl` methods.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

## Testing
- `mvn -o test -Dtest=OAuthAuthorizeResourceTest` — green (13).
- `mvn -o test -Dtest=RemoteAuthServiceTest` — green (28).
- `mvn -o test -Dtest=AuthServiceImplTest` — green (6).
- `mvn -o spotless:check` — green.

## Documentation
N/A — internal backend feature gated behind the MCP OAuth flag; no user-facing documentation in this PR. End-user docs land with the feature rollout.


apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
```diff
@@ -93,6 +96,10 @@ record AuthResponse(
             String user, String workspaceId, String workspaceName, List<Quota> quotas, OpikVersion opikVersion) {
     }
 
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    record WorkspaceIdNameResponse(String workspaceId, String workspaceName) {
```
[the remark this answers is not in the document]
LifeXplorer: there are no cases when the entity is manually constructed

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/auth/RemoteAuthServiceTest.java
```diff
@@ -333,6 +338,116 @@ void testSessionAuthUnauthorized(int remoteAuthStatusCode, Class<? extends Excep
...
+        var sessionTokenValue = "session-" + UUID.randomUUID();
+        var responseJson = OBJECT_MAPPER.writeValueAsString(Arrays.asList(
+                Map.of("workspaceId", "ws-1", "workspaceName", "production"),
+                Map.of("workspaceId", "ws-default", "workspaceName", DEFAULT_WORKSPACE_NAME),
+                Map.of("workspaceId", "ws-2", "workspaceName", "staging")));
+        WIRE_MOCK.server().stubFor(get(urlPathEqualTo("/workspaces"))
+                .withQueryParam("withoutExtendedData", equalTo("true"))
+                .withCookie(RequestContext.SESSION_COOKIE, equalTo(sessionTokenValue))
+                .willReturn(okJson(responseJson)));
+
```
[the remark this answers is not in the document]
LifeXplorer: other tests in same class follow similar pattern

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/ConsentRequest.java
```diff
@@ -0,0 +1,22 @@
...
+@Builder(toBuilder = true)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
+public record ConsentRequest(
+        @NotBlank String clientId,
+        @NotBlank String redirectUri,
+        @NotBlank String codeChallenge,
+        @NotBlank String codeChallengeMethod,
+        @NotBlank String resource,
+        String state,
```
BorisTkachenko, line 21: Might be better to use ENUM here

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/McpOAuthBundle.java
```diff
@@ -20,7 +20,8 @@ public void run(GuiceyEnvironment environment) {
         OpikConfiguration config = environment.configuration();
         if (!config.getMcpOAuth().isEnabled()) {
             environment.disableExtensions(
-                    OAuthMetadataResource.class);
+                    OAuthMetadataResource.class,
+                    OAuthAuthorizeResource.class);
```
BorisTkachenko, line 24: Same as https://github.com/comet-ml/opik/pull/6993#discussion_r3380636220

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthAuthorizeResource.java
```diff
@@ -0,0 +1,219 @@
...
+public class OAuthAuthorizeResource {
+
+    private final @NonNull OAuthClientService clientService;
+    private final @NonNull AuthService authService;
+    private final @NonNull McpOAuthService mcpOAuthService;
+    private final @NonNull OpikConfiguration opikConfig;
+
+    @GET
+    @Path("/authorize")
+    public Response authorize(
```
BorisTkachenko, line 61: Missing Swagger annotations for OpenAPI, all endpoints in this class.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthAuthorizeResource.java
```diff
@@ -0,0 +1,219 @@
...
+
+        McpOAuthConfig config = opikConfig.getMcpOAuth();
+        requireClientWithRedirect(request.clientId(), request.redirectUri());
+        if (!config.getMcpResourceUri().equals(request.resource())) {
+            throw new BadRequestException(ERROR_INVALID_TARGET);
+        }
+        if (StringUtils.isBlank(request.codeChallenge())
+                || !CODE_CHALLENGE_METHOD_S256.equals(request.codeChallengeMethod())) {
+            throw new BadRequestException(ERROR_INVALID_REQUEST);
+        }
```
BorisTkachenko: Could be extracted to a dedicated method and reused

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthAuthorizeResource.java
```diff
@@ -0,0 +1,219 @@
...
+            throw new BadRequestException(ERROR_INVALID_CLIENT);
+        }
+        // RFC 6749 §3.1.2: the redirection endpoint URI MUST NOT include a fragment, otherwise appended
+        // code/error/state query params would land inside the fragment and be unreadable by the client.
+        if (redirectUri == null || redirectUri.contains("#")
+                || !clientService.matchesRedirectUri(client.get(), redirectUri)) {
+            throw new BadRequestException("invalid redirect_uri");
+        }
+        return client.get();
+    }
```
BorisTkachenko: Might be better to move to a service layer

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
```diff
@@ -177,6 +184,89 @@ public void authenticateSession(Cookie sessionToken) {
...
+                    .map(workspace -> WorkspaceInfo.builder()
+                            .id(workspace.workspaceId())
+                            .name(workspace.workspaceName())
+                            .build())
+                    .toList();
+        }
+    }
+
+    @Override
+    public void authorizeOAuth(ValidatedToken token, ContextInfoHolder contextInfo) {
```
BorisTkachenko: It's used in tests only

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthAuthorizeResource.java
```diff
@@ -0,0 +1,219 @@
...
+            @QueryParam("response_type") String responseType,
+            @QueryParam("code_challenge") String codeChallenge,
+            @QueryParam("code_challenge_method") String codeChallengeMethod,
+            @QueryParam("resource") String resource,
+            @QueryParam("state") String state,
+            @Context HttpHeaders headers,
+            @Context UriInfo uriInfo) {
+
+        McpOAuthConfig config = opikConfig.getMcpOAuth();
+        McpOAuthClient client = requireClientWithRedirect(clientId, redirectUri);
```
BorisTkachenko: McpOAuthClient client - never used

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthAuthorizeResource.java
```diff
@@ -0,0 +1,219 @@
...
+                || !CODE_CHALLENGE_METHOD_S256.equals(codeChallengeMethod)) {
+            return errorRedirect(redirectUri, ERROR_INVALID_REQUEST, state);
+        }
+        if (!config.getMcpResourceUri().equals(resource)) {
+            return errorRedirect(redirectUri, ERROR_INVALID_TARGET, state);
+        }
+
+        Cookie session = headers.getCookies().get(RequestContext.SESSION_COOKIE);
+        try {
+            authService.listEligibleWorkspaces(session);
```
BorisTkachenko: Not sure why this is needed, even if some workspaces are returned we ignore them. If it's just to check session authorization, maybe could be used existing approach.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthAuthorizeResource.java
```diff
@@ -0,0 +1,219 @@
...
+        if (!config.getMcpResourceUri().equals(request.resource())) {
+            throw new BadRequestException(ERROR_INVALID_TARGET);
+        }
+        if (StringUtils.isBlank(request.codeChallenge())
+                || !CODE_CHALLENGE_METHOD_S256.equals(request.codeChallengeMethod())) {
+            throw new BadRequestException(ERROR_INVALID_REQUEST);
+        }
+
+        Cookie session = headers.getCookies().get(RequestContext.SESSION_COOKIE);
+        UserWorkspace workspace = authService.authorizeWorkspace(session, request.workspaceName());
```
BorisTkachenko: We might use the existing approach where authentication is done via AuthFilter and all what we need is just set to the context. In this case instead of `private` url prefix we could use `mcp`. This way it will be consistent with the existing code base.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
```diff
@@ -177,6 +184,89 @@ public void authenticateSession(Cookie sessionToken) {
...
+            throw new ClientErrorException(NOT_ALLOWED_TO_ACCESS_WORKSPACE, Response.Status.FORBIDDEN);
+        }
+        try (var response = client.target(URI.create(reactServiceUrl.url()))
+                .path("opik")
+                .path("auth-session")
+                .request()
+                .accept(MediaType.APPLICATION_JSON)
+                .cookie(sessionToken)
+                .post(Entity.json(AuthRequest.builder().workspaceName(workspaceName).build()))) {
+            var authResponse = verifyResponse(response);
```
BorisTkachenko, line 249: Now we have a lot of client calls in this service. Might be better to move to a dedicated Client class.

BorisTkachenko (commented):
Left a few comments

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthAuthorizeResource.java
LifeXplorer: replaced with javax annotation

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
LifeXplorer: Good catch. This method will be necessary in https://github.com/comet-ml/opik/pull/6997 chunk 6. Let me keep it here to not rework both PRs

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthAuthorizeResource.java
LifeXplorer: Removed announcing local variable. the method is still called to validate client id expected existing client

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthAuthorizeResource.java
LifeXplorer: same case with chunked PR. In chunk 4 PR https://github.com/comet-ml/opik/pull/6995 will be used in `/authorize/context`

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/McpOAuthBundle.java
LifeXplorer: same push back

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthAuthorizeResource.java
LifeXplorer: added

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthAuthorizeResource.java
LifeXplorer: moved to `OAuthClientService`

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthAuthorizeResource.java
LifeXplorer: moved to service

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
```diff
@@ -177,6 +184,89 @@ public void authenticateSession(Cookie sessionToken) {
...
+            setCredentialIntoContext(credentials, token.workspaceName(), null);
+        }
+    }
+
+    @Override
+    public UserWorkspace authorizeWorkspace(Cookie sessionToken, String workspaceName) {
+        requireSession(sessionToken);
+        if (isDefaultWorkspace(workspaceName)) {
+            throw new ClientErrorException(NOT_ALLOWED_TO_ACCESS_WORKSPACE, Response.Status.FORBIDDEN);
+        }
```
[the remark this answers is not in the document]
LifeXplorer: I'd keep bigger refactoring out of the scope of this PR

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
LifeXplorer: I'd keep bigger refactoring out of the scope of this PR

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/ConsentRequest.java
LifeXplorer: The only param that fits to be ENUM is `codeChallengeMethod`.
An enum would be cleaner than a regex constraint, but since only `S256` is supported the `@Pattern` works fine and enum deserialization of OAuth string params brings extra complexity. I'd suggest to leave as is

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/UserWorkspace.java
```diff
@@ -0,0 +1,8 @@
+package com.comet.opik.infrastructure.auth;
+
+import lombok.Builder;
+import lombok.NonNull;
+
+@Builder(toBuilder = true)
+public record UserWorkspace(@NonNull String userName, @NonNull String workspaceId, @NonNull String workspaceName) {
```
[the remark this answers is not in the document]
LifeXplorer: minor cosmetics. Leaving with no change

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/OAuthClientService.java
```diff
@@ -25,7 +28,23 @@ public McpOAuthClient register(@NonNull ClientRegistrationRequest request) {
...
+     * {@code code}/{@code error}/{@code state} query params would land inside the fragment and be unreadable
+     * by the client.
+     */
+    public McpOAuthClient resolveForRedirect(@NonNull String clientId, String redirectUri) {
+        McpOAuthClient client = resolve(clientId)
+                .orElseThrow(() -> new BadRequestException(ERROR_INVALID_CLIENT));
+        if (redirectUri == null || redirectUri.contains("#") || !matchesRedirectUri(client, redirectUri)) {
+            throw new BadRequestException("invalid redirect_uri");
+        }
+        return client;
```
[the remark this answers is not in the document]
LifeXplorer: added test

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthMetadataResource.java
```diff
@@ -42,10 +49,10 @@ public Response metadata() {
...
         return Response.ok(AuthorizationServerMetadata.builder()
                 .issuer(issuer)
-                .authorizationEndpoint(issuer + "/oauth/authorize")
-                .tokenEndpoint(issuer + "/oauth/token")
-                .revocationEndpoint(issuer + "/oauth/revoke")
-                .registrationEndpoint(issuer + "/oauth/register")
+                .authorizationEndpoint(issuer + AUTHORIZE_PATH)
+                .tokenEndpoint(issuer + TOKEN_PATH)
+                .revocationEndpoint(issuer + REVOKE_PATH)
+                .registrationEndpoint(issuer + REGISTER_PATH)
```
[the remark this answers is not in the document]
LifeXplorer: those endpoints will be used in follow up prs

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthAuthorizeResource.java
```diff
@@ -0,0 +1,223 @@
...
+import static com.comet.opik.domain.mcpoauth.OAuthConstants.PARAM_RESPONSE_TYPE;
+import static com.comet.opik.domain.mcpoauth.OAuthConstants.PARAM_STATE;
+import static com.comet.opik.domain.mcpoauth.OAuthConstants.RESPONSE_TYPE_CODE;
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+@Path("/oauth")
+@Timed
+@RequiredArgsConstructor(onConstructor_ = @Inject)
+@Tag(name = "MCP OAuth", description = "MCP OAuth 2.1 Authorization Server resources")
+public class OAuthAuthorizeResource {
```
andrescrz, line 52: Please double check adding the produces and consumes annotations defaulting to JSON.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthAuthorizeResource.java
```diff
@@ -0,0 +1,223 @@
...
+import static com.comet.opik.domain.mcpoauth.OAuthConstants.PARAM_CODE_CHALLENGE_METHOD;
+import static com.comet.opik.domain.mcpoauth.OAuthConstants.PARAM_ERROR;
+import static com.comet.opik.domain.mcpoauth.OAuthConstants.PARAM_REDIRECT_URI;
+import static com.comet.opik.domain.mcpoauth.OAuthConstants.PARAM_RESOURCE;
+import static com.comet.opik.domain.mcpoauth.OAuthConstants.PARAM_RESPONSE_TYPE;
+import static com.comet.opik.domain.mcpoauth.OAuthConstants.PARAM_STATE;
+import static com.comet.opik.domain.mcpoauth.OAuthConstants.RESPONSE_TYPE_CODE;
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+@Path("/oauth")
```
andrescrz, line 46: Please add typical logs to all the endpoints, so we have operational visibility (happy patch, branching and errors). 

This is oauth, so make sure you don't leak any sensitive payload.

This comment applies to all endpoints.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthAuthorizeResource.java
```diff
@@ -0,0 +1,223 @@
...
+    private final @NonNull OAuthClientService clientService;
+    private final @NonNull AuthService authService;
+    private final @NonNull McpOAuthService mcpOAuthService;
+    private final @NonNull OpikConfiguration opikConfig;
+
+    @GET
+    @Path("/authorize")
+    @Operation(operationId = "authorize", summary = "OAuth Authorization Endpoint", description = "OAuth 2.1 authorization endpoint (RFC 6749 §3.1). Validates the client and PKCE parameters, then redirects to the login or consent page", responses = {
+            @ApiResponse(responseCode = "302", description = "Redirect to login, consent, or client redirect_uri with an error")})
+    public Response authorize(
```
andrescrz, line 61: I believe these endpoints have implemented chunks of logic whose responsiblity belongs to a service class, not to the resource. 

Generally, the resource should be limited to:
```
validations
delegate business logic to a service
assembling the response and returning
```

As your logic is many times returning a redirect response, you'll need to refactor this to be exception based, as you can't return from a service delegated method.

If you do that, make sure you don't try-catch in the resource, that's a bad practice.

There should be handled for those exceptions with the appropriate logic.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthAuthorizeResource.java
```diff
@@ -0,0 +1,223 @@
...
+        if (!config.getMcpResourceUri().equals(resource)) {
+            return errorRedirect(redirectUri, ERROR_INVALID_TARGET, state);
+        }
+
+        Cookie session = headers.getCookies().get(RequestContext.SESSION_COOKIE);
+        try {
+            authService.listEligibleWorkspaces(session);
+        } catch (ClientErrorException e) {
+            String rawQuery = uriInfo.getRequestUri().getRawQuery();
+            String authorizeUrl = config.getBaseUrl() + "/oauth/authorize" + (rawQuery == null ? "" : "?" + rawQuery);
```
andrescrz: Minor: please reuse the constants such as `/oauth/authorize`

andrescrz: Minor: better `StringUtils.isBlank` check on raw query.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/OAuthClientService.java
```diff
@@ -25,7 +28,23 @@ public McpOAuthClient register(@NonNull ClientRegistrationRequest request) {
...
+    }
+
+    private boolean matchesRedirectUri(@NonNull McpOAuthClient client, @NonNull String requestedRedirectUri) {
```
andrescrz: Minor: private methods shouldn't have validations. That should be responsibility of the public entry points.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthService.java
```diff
@@ -43,4 +60,21 @@ public void authenticate(HttpHeaders headers, Cookie sessionToken, ContextInfoHo
...
+        return ELIGIBLE_WORKSPACES;
+    }
+
+    @Override
+    public UserWorkspace authorizeWorkspace(Cookie sessionToken, String workspaceName) {
+        return AUTHORIZED_WORKSPACE;
+    }
+
+    @Override
+    public void authorizeOAuth(ValidatedToken token, ContextInfoHolder contextInfo) {
```
andrescrz: Please validate public methods: typically `@NonNull` but not limited to that.

This comment applies to the whole PR.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
```diff
@@ -177,6 +184,89 @@ public void authenticateSession(Cookie sessionToken) {
...
+                .request()
+                .accept(MediaType.APPLICATION_JSON)
+                // avoids gzip double-decompression issue in case of huge workspaces list
+                .acceptEncoding("identity")
+                .cookie(sessionToken)
+                .get()) {
+            if (response.getStatusInfo().getFamily() != Response.Status.Family.SUCCESSFUL) {
+                throw toSessionAuthException(response);
+            }
+            return response.readEntity(new GenericType<List<WorkspaceIdNameResponse>>() {
```
andrescrz: Please extract the `GenericType` to a static constant, those are generally thread-safe and heavy to build, so they should be reused.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
```diff
@@ -177,6 +184,89 @@ public void authenticateSession(Cookie sessionToken) {
...
+            throw new ClientErrorException(NOT_LOGGED_USER, Response.Status.FORBIDDEN);
+        }
+    }
+
+    private ClientErrorException toSessionAuthException(Response response) {
+        if (response.getStatus() == Response.Status.UNAUTHORIZED.getStatusCode()
+                || response.getStatus() == Response.Status.FORBIDDEN.getStatusCode()) {
+            return new ClientErrorException(NOT_LOGGED_USER, Response.Status.FORBIDDEN);
+        }
+        log.error("Unexpected error while listing workspaces, received status code: {}", response.getStatus());
```
andrescrz: You probably want to attempt (in a safe manner) to parse the response entity (as string) and try to print some log information about it, for debugging purposes.

Otherwise, it will be hard to know what happens in production if this is ever reached.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
```diff
@@ -177,6 +184,89 @@ public void authenticateSession(Cookie sessionToken) {
...
+        throw new InternalServerErrorException();
```
andrescrz: Nit: add some custom message to this exception, you it's easier to identify the entry point besides looking at the stack trace. It can be similar to the printed log above, but without internal information. That should never bubble up to an endpoint caller.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthAuthorizeResourceTest.java
```diff
@@ -0,0 +1,292 @@
...
+    private OAuthAuthorizeResource resource;
+    private McpOAuthConfig mcpConfig;
+
+    @BeforeEach
+    void setUp() {
+        resource = new OAuthAuthorizeResource(clientService, authService, mcpOAuthService, opikConfig);
+        mcpConfig = new McpOAuthConfig();
+        mcpConfig.setEnabled(true);
+        mcpConfig.setBaseUrl(BASE_OPIK_URL);
+        mcpConfig.setMcpResourceUri(RESOURCE_URI);
```
andrescrz: Nit: better favour builders over constructors in the code base.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthAuthorizeResourceTest.java
```diff
@@ -0,0 +1,292 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+@DisplayName("OAuth Authorize Resource Test")
+class OAuthAuthorizeResourceTest {
```
andrescrz, line 61: Mentioned in the previous revision, better go with Dropwizard resource tests instead of pure unit tests. Reference here:
- https://www.dropwizard.io/en/stable/manual/testing.html#testing-resources

apps/opik-backend/src/test/java/com/comet/opik/domain/mcpoauth/OAuthClientServiceTest.java
```diff
@@ -0,0 +1,90 @@
...
+    private OAuthClientService service;
+
+    private void mockClient(String... redirectUris) {
+        McpOAuthClient client = McpOAuthClient.builder()
+                .id(CLIENT_ID)
+                .name("Test Client")
+                .redirectUris(Set.of(redirectUris))
+                .build();
+        lenient().when(strategy.supports(CLIENT_ID)).thenReturn(true);
+        lenient().when(strategy.resolve(CLIENT_ID)).thenReturn(Optional.of(client));
```
andrescrz: Nit: avoid lenient whenever possible. They sometimes allow to pass silently problems.

apps/opik-backend/src/test/java/com/comet/opik/domain/mcpoauth/OAuthClientServiceTest.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @Test
+    @DisplayName("resolveForRedirect: loopback redirect matches on any port (RFC 8252 §7.3)")
+    void resolveForRedirect_loopbackDifferentPort_matches() {
+        mockClient("http://127.0.0.1:1111/cb");
+
+        McpOAuthClient client = service.resolveForRedirect(CLIENT_ID, "http://127.0.0.1:54321/cb");
+
+        assertThat(client.id()).isEqualTo(CLIENT_ID);
```
andrescrz: As usual, better assert full objects: actual vs expected, instead of individual fields.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/auth/AuthServiceImplTest.java
```diff
@@ -68,4 +71,52 @@ void testAuthenticate__whenWorkspaceIsNotDefault__thenFail() {
...
                 () -> authService.authenticate(headers, sessionToken, infoHolder));
     }
+
+    @Test
+    void testListEligibleWorkspaces__returnsDefaultWorkspace() {
+        // When
+        List<WorkspaceInfo> result = authService.listEligibleWorkspaces(null);
+
+        // Then
+        Assertions.assertEquals(
```
andrescrz: Let's please always use AssertJ assertions and clean up JUnit assertions from the code base. It's not only the convention of the service, it's also more powerful, readable and maintainble.

No need to refactor untouched parts in this PR, but you should follow this in all new code in this PR. Ideally, I'd clean up this test class as well, as there seemed to be a pre-existing JUnit assertion here.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/auth/AuthServiceImplTest.java
```diff
@@ -68,4 +71,52 @@ void testAuthenticate__whenWorkspaceIsNotDefault__thenFail() {
...
+                List.of(WorkspaceInfo.builder()
+                        .id(ProjectService.DEFAULT_WORKSPACE_ID)
+                        .name(ProjectService.DEFAULT_WORKSPACE_NAME)
+                        .build()),
+                result);
+    }
+
+    @Test
+    void testAuthorizeWorkspace__returnsDefaultUserWorkspace() {
```
andrescrz, line 75: As usual, tests that are mostly the same, except for some input argument and the expected results, should be parameterised as a single test.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/auth/RemoteAuthServiceTest.java
```diff
@@ -333,6 +338,116 @@ void testSessionAuthUnauthorized(int remoteAuthStatusCode, Class<? extends Excep
                 .hasMessage(expectedMessage);
     }
 
+    @Test
+    void testListEligibleWorkspaces__filtersDefaultWorkspaceAndMapsToWorkspaceInfo() throws JsonProcessingException {
+        var sessionTokenValue = "session-" + UUID.randomUUID();
+        var production = WorkspaceInfo.builder().id("ws-1").name("production").build();
+        var staging = WorkspaceInfo.builder().id("ws-2").name("staging").build();
```
andrescrz: As usual, better use Podam, which randomly populates records/POJOs etc., instead of hardcoded test values. 

That makes tests more endurable and realistic.

andrescrz (commented):
The general architecture of the PR looks good.

Let's iterate one more time to make sure the implementation is robust.

Focus on the important comment, feel free to skip nit and minor comments.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthAuthorizeResource.java
LifeXplorer: makes sense. Added

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthAuthorizeResource.java
LifeXplorer: Extracted the logic into service, added logs

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthAuthorizeResource.java
LifeXplorer: moved to service

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthService.java
LifeXplorer: added validation for AuthService implementations

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthAuthorizeResourceTest.java
LifeXplorer: added builder to McpOAuthConfig

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/auth/AuthServiceImplTest.java
LifeXplorer: refactored to assertj

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/auth/AuthServiceImplTest.java
LifeXplorer: There are no tests that can be squashed into single parameterized.
Only `testAuthenticate__whenWorkspaceIsDefault__thenUseDefaultUser` and `testAuthenticate__whenWorkspaceIsNotDefault__thenFail` validate the same method. But `testAuthenticate__whenWorkspaceIsDefault__thenUseDefaultUser` itself is already parameterized

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/auth/RemoteAuthServiceTest.java
LifeXplorer: changed to Podam

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthAuthorizationService.java
```diff
@@ -0,0 +1,167 @@
...
+     */
+    public ConsentResponse issueAuthorizationCode(@NonNull ConsentRequest request, Cookie session) {
+        McpOAuthConfig config = opikConfig.getMcpOAuth();
+        clientService.resolveForRedirect(request.clientId(), request.redirectUri());
+        if (!config.getMcpResourceUri().equals(request.resource())) {
+            log.info("Consent for client_id '{}' targets an unexpected resource", request.clientId());
+            throw new BadRequestException(ERROR_INVALID_TARGET);
+        }
+
+        UserWorkspace workspace = authService.authorizeWorkspace(session, request.workspaceName());
```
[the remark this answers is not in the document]
LifeXplorer: workspaceName is validated in AuthService

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
```diff
@@ -177,6 +188,107 @@ public void authenticateSession(Cookie sessionToken) {
...
+     * {@link InternalServerErrorException} carrying a custom message that identifies the failing operation. The body is
+     * only logged, never surfaced to the caller, so no internal/remote detail bubbles up to the endpoint.
+     */
+    private InternalServerErrorException unexpectedRemoteError(String operation, Response response) {
+        log.error("Unexpected error while {}, received status code: {}, body: '{}'",
+                operation, response.getStatus(), readBodySafely(response));
+        return new InternalServerErrorException("Unexpected error while " + operation);
+    }
+
+    private static String readBodySafely(Response response) {
```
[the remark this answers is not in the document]
LifeXplorer: leaving without changes

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
```diff
@@ -177,6 +188,107 @@ public void authenticateSession(Cookie sessionToken) {
...
+        try {
+            return response.hasEntity() ? response.readEntity(String.class) : "";
```
andrescrz, line 285: I strongly recommend adding a bufferEntity call:
```
response.hasEntity() && response.bufferEntity(
```

Which ensures there could be multiple reads of the entity.

Here it should be the final read attempt (error path), it's mostly for endurability in the future.

But I do recommend reviewing the execution paths to this point, to check if there was any previous `readEntity` attempt without a `bufferEntity` call.

In that case, this would fail.

Anyway, not a blocker as you're protected by the catch below.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthAuthorizeResourceTest.java
```diff
@@ -0,0 +1,315 @@
...
+
+    private static final String CLIENT_ID = "test-client";
+    private static final String REDIRECT_URI = "http://localhost:1234/cb";
+    private static final String BASE_OPIK_URL = "https://www.comet.com/opik";
+    private static final String RESOURCE_URI = BASE_OPIK_URL + "/api/v1/mcp";
+    private static final String CODE_CHALLENGE = "abc123challenge";
+    private static final String STATE = "state-xyz";
+    private static final String CSRF = "csrf-token-abc";
```
andrescrz, line 69: Nit: better use random values for tests. You can do `state-%.formatting(RandomStringValue...` and so on.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthAuthorizeResourceTest.java
```diff
@@ -0,0 +1,315 @@
...
+                .property(ClientProperties.FOLLOW_REDIRECTS, false)
+                .queryParam(PARAM_CLIENT_ID, CLIENT_ID)
+                .queryParam(PARAM_REDIRECT_URI, REDIRECT_URI)
+                .queryParam(PARAM_RESPONSE_TYPE, responseType)
+                .queryParam(PARAM_CODE_CHALLENGE, codeChallenge)
+                .queryParam(PARAM_CODE_CHALLENGE_METHOD, CODE_CHALLENGE_METHOD_S256)
+                .queryParam(PARAM_RESOURCE, resource)
+                .queryParam(PARAM_STATE, STATE)
+                .request()
+                .get()) {
```
andrescrz, line 175: Minor: feel free to encapsulate the client call in a small aux private method in this class, to avoid copying an pasting.

andrescrz (approved):
LGTM, just something to follow-up to avoid errors.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthAuthorizeResourceTest.java
LifeXplorer: noted for next PRs

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthAuthorizeResourceTest.java
LifeXplorer: will be updated in the next chunk PR

PR #6995, opened 2026-06-08
[OPIK-6666] [BE] feat: MCP OAuth token + revoke endpoints (PR 4/7)

LifeXplorer:
## Details

**PR 4 of 7** splitting #6900 per [review](https://github.com/comet-ml/opik/pull/6900#pullrequestreview-4428853723).

> ⛓️ **Chained PR.** Base is `avinahradau/OPIK-6666-oauth-as-chunked-3` (PR #6994). Base retargets to `main` once PR-3 lands.

| Item | What |
|---|---|
| `OAuthTokenResource` | `POST /oauth/token` (form-encoded; `authorization_code` exchange w/ PKCE + `refresh_token` grant; RFC 6749 §5.2 error shape) and `POST /oauth/revoke` (RFC 7009) |
| `OAuthError` | error response DTO |
| `McpOAuthBundle` | now also disables `OAuthTokenResource` when flag off |
| `OAuthTokenResourceTest` | coverage |

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Testing
- `mvn -o test` for `OAuthTokenResourceTest`, `McpOAuthBundleTest`, `McpOAuthTokenUtilsTest` — green.
- `mvn -o test-compile` — green.

## Documentation
No documentation changes required.  internal MCP OAuth token/revoke endpoints, no user-facing API or docs surface affected.

## Issues


apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthTokenResourceTest.java
```diff
@@ -0,0 +1,152 @@
...
+    void token_unsupportedGrant_returnsError() {
+        Response response = resource.token("client_credentials", null, null, CLIENT_ID, null, null);
+
+        assertThat(response.getStatus()).isEqualTo(Response.Status.BAD_REQUEST.getStatusCode());
+    }
+
+    @Test
+    @DisplayName("POST /revoke: valid token returns 200 OK and revokes")
+    void revoke_validToken_returnsOk() {
+        Response response = resource.revoke(ACCESS_PREFIX + "xxx", "access_token", CLIENT_ID);
```
LifeXplorer: **[medium] Error tests assert only HTTP status, not the OAuth error code in the body**

Every negative/error test (missing fields line 81, unknown client line 93, exchange failure line 107, unsupported grant line 128) asserts only response.getStatus()==BAD_REQUEST and never inspects the OAuthError entity. The resource's entire contract is the specific OAuth error code it returns (invalid_request, invalid_client, unsupported_grant_type, invalid_grant per RFC 6749). These status-only assertions pass even if the wrong error code is emitted. This is most acute for token_unsupportedGrant_returnsError: the resource returns invalid_request for a blank grant_type and unsupported_grant_type for an unknown one (line 94 of the resource), but the test cannot distinguish them because it never reads the body. So the test does not actually verify the unsupported_grant_type behavior its name and DisplayName claim.

💡 After the status assertion, cast the entity to OAuthError and assert the expected error code, e.g. assertThat(((OAuthError) response.getEntity()).error()).isEqualTo(ERROR_UNSUPPORTED_GRANT_TYPE) (and ERROR_INVALID_REQUEST / ERROR_INVALID_CLIENT / "invalid_grant" for the other cases). Import the OAuthError type and OAuthConstants so the codes are referenced symbolically rather than hardcoded.

LifeXplorer (commented):

8 findings — 0 high · 1 medium · 7 low. Suppressed by team conventions: see `suppressed.md`.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthTokenUtils.java
```diff
@@ -63,6 +64,19 @@ public static String hash(String token) {
         return DigestUtils.sha256Hex(token);
     }
 
+    public static String maskToken(String token) {
+        if (StringUtils.isEmpty(token)) {
+            return "";
+        }
+        //expected Opik token size
+        if (token.length() > RANDOM_BYTES) {
```
LifeXplorer: **[low] Threshold compares encoded char length against raw-byte count (unit confusion)**

`token.length()` is the character length of the base64url-encoded token, but `RANDOM_BYTES` (32) is the count of raw random bytes BEFORE encoding. A full Opik token is `prefix(12) + base64url(32 bytes)(43)` = 55 chars, so the real 'expected Opik token size' is ~55, not 32. The comment '//expected Opik token size' is therefore inaccurate, and the guard masks any token longer than 32 chars (e.g. a 33–54 char non-Opik value) while the code's intent appears to be 'mask only well-formed Opik tokens'. The two quantities are in different units and only coincidentally make the substring(0,12) call safe.

💡 Compare against the actual encoded token length, e.g. `ACCESS_PREFIX.length() + ENCODED_TOKEN_LENGTH` (43) or a dedicated constant `MIN_MASKABLE_LENGTH = 16`, and fix the comment. Do not reuse RANDOM_BYTES (a pre-encoding byte count) as a character-length threshold.

_rule: `backend-style-maintainability-keep-names-accurate-method-names-class-names-sql-constants-l`_

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthTokens.java
```diff
@@ -63,6 +64,19 @@ public static String hash(String token) {
         return DigestUtils.sha256Hex(token);
     }
 
+    public static String maskToken(String token) {
```
LifeXplorer, line 73: **[low] No tests for new maskToken behavior**

maskToken introduces new observable behavior (empty/blank -> "", short token returned verbatim, long token masked to head+...+tail) and is used in a security-sensitive log path in OAuthTokenResource.revoke. McpOAuthBundleTest/OAuthTokenResourceTest exist but none cover maskToken, so the boundary at length 32/33 and the empty case are unverified.

💡 Add a focused @ParameterizedTest covering: empty/null -> "", a value at and just above the length threshold, and a full-length access token verifying only prefix(12)+...+last4 is revealed and the middle is hidden.

_rule: `backend-testing-add-regression-tests-for-every-new-backend-behavior`_

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/McpOAuthBundle.java
```diff
@@ -21,7 +21,8 @@ public void run(GuiceyEnvironment environment) {
         if (!config.getMcpOAuth().isEnabled()) {
             environment.disableExtensions(
                     OAuthMetadataResource.class,
-                    OAuthAuthorizeResource.class);
+                    OAuthAuthorizeResource.class,
+                    OAuthTokenResource.class);
```
LifeXplorer, line 25: **[low] **

The diff extends the disabled-extension set so OAuthTokenResource (and previously OAuthAuthorizeResource) are now disabled when mcpOAuth.enabled=false, but the only test (McpOAuthBundleTest) still asserts verify(environment).disableExtensions(OAuthMetadataResource.class) — a single-argument call. The production code now invokes disableExtensions(OAuthMetadataResource.class, OAuthAuthorizeResource.class, OAuthTokenResource.class) as one varargs call, so the new behavior (that the token and authorize resources are gated off) is not covered. Worse, with Mockito's default varargs matching, a verify expecting one vararg element does not match an actual call carrying three elements, so the existing assertion does not even validate the real invocation. A regression where OAuthTokenResource is dropped from the list would not be caught.

💡 Update McpOAuthBundleTest to assert the full argument set, e.g. verify(environment).disableExtensions(OAuthMetadataResource.class, OAuthAuthorizeResource.class, OAuthTokenResource.class), so the security-relevant gating of all three OAuth resources is regression-covered.

_rule: `backend-testing-add-regression-tests-for-every-new-backend-behavior`_

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthTokenResource.java
```diff
@@ -0,0 +1,133 @@
...
+@Path("/oauth")
+@Timed
+@Slf4j
+@RequiredArgsConstructor(onConstructor_ = @Inject)
+public class OAuthTokenResource {
+
+    private final @NonNull OAuthClientService clientService;
+    private final @NonNull McpOAuthService mcpOAuthService;
+
+    @POST
```
LifeXplorer, line 54: **[low] No brute-force protection on the unauthenticated token/revoke endpoints**

OAuthTokenResource is registered as an unauthenticated public resource (McpOAuthBundle) and exposes POST /oauth/token and POST /oauth/revoke, which accept caller-supplied authorization codes, refresh tokens and client_ids. Neither method carries @RateLimited, so there is no throttling on these credential-accepting endpoints. The mined security rule calls for rate limiting on credential-accepting endpoints to prevent brute-force. Practical impact is limited because codes/tokens are 256-bit random values (infeasible to guess), so this is defensive hardening rather than an exploitable gap; the suppressed 'low-frequency internal/frontend' exemption does not apply since these are public, unauthenticated surfaces.

💡 Add throttling for these public endpoints (e.g. an IP-based rate limit, since @RateLimited's user/workspace buckets do not apply pre-authentication) to bound abusive request volume against /oauth/token and /oauth/revoke.

_rule: `backend-security-validate-and-restrict-user-supplied-urls-before-making-outbo`_

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthTokenUtils.java
```diff
@@ -63,6 +64,19 @@ public static String hash(String token) {
...
+            return token.substring(0, 12) + "..." + token.substring(token.length() - 4);
+        } else {
+            //return full string as confirmed not to be expected token shape
+            return token;
```
LifeXplorer: **[low] Short non-Opik values are logged verbatim**

For inputs of length <= 32 the method returns the token unredacted. The sole caller logs the result on the OAuth revoke path; a malformed-but-still-secret short token would be written to logs in full. The comment justifies this as 'not expected token shape', but well-behaved clients can still submit short opaque secrets to /revoke.

💡 Defensive hardening only: redact short inputs too (e.g. fully mask anything below the threshold) so no caller-supplied secret reaches the log verbatim.

_rule: `backend-security-redact-credentials-and-sensitive-data-from-logs-and-error-re`_

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthTokenResourceTest.java
```diff
@@ -0,0 +1,152 @@
...
+                CODE_VERIFIER, null);
+
+        assertThat(response.getStatus()).isEqualTo(Response.Status.OK.getStatusCode());
+        TokenResponse body = (TokenResponse) response.getEntity();
+        assertThat(body.accessToken()).startsWith(ACCESS_PREFIX);
+        assertThat(body.refreshToken()).startsWith(REFRESH_PREFIX);
+        assertThat(body.tokenType()).isEqualTo("Bearer");
+    }
+
+    @Test
```
LifeXplorer: **[low] Success-path body asserted field-by-field instead of as a full object**

token_authCodeGrant_returnsTokens asserts only accessToken prefix, refreshToken prefix, and tokenType individually (lines 70-72), leaving expiresIn, workspaceId, and workspaceName uncovered. The resource returns the TokenResponse from the service unchanged, so the test should confirm the whole object is passed through intact; a regression that drops or rewrites expiresIn/workspaceId would not be caught.

💡 Assert the full object, e.g. assertThat(body).isEqualTo(minted()) (or usingRecursiveComparison), since minted() is the exact stubbed return value. The refresh test (line 120) has the same gap.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthTokenResourceTest.java
```diff
@@ -0,0 +1,152 @@
...
+        when(clientService.resolve(CLIENT_ID)).thenReturn(Optional.of(validClient()));
+        when(mcpOAuthService.refresh(REFRESH_TOKEN, CLIENT_ID)).thenReturn(minted());
+
+        Response response = resource.token("refresh_token", null, null, CLIENT_ID, null, REFRESH_TOKEN);
+
+        assertThat(response.getStatus()).isEqualTo(Response.Status.OK.getStatusCode());
+        TokenResponse body = (TokenResponse) response.getEntity();
+        assertThat(body.refreshToken()).isEqualTo(REFRESH_PREFIX + "yyy");
+    }
+
```
LifeXplorer, line 128: **[low] Refresh-grant test does not cover the missing-fields branch**

The refresh_token grant has its own invalid_request guard (resource lines 81-83: blank refreshToken OR blank clientId). The auth_code grant gets a missing-fields test (token_authCodeGrant_missingFields_invalidRequest) but the refresh_token branch has no equivalent negative test, so the refresh-path validation guard is untested and could silently break.

💡 Add a test that calls resource.token("refresh_token", null, null, CLIENT_ID, null, null) (blank refresh token) and asserts BAD_REQUEST plus error()==invalid_request, verifying mcpOAuthService.refresh is never invoked.

_general finding_

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthTokenUtils.java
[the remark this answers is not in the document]
LifeXplorer: Addressed in 02d0e7e0ec. `maskToken` no longer emits any token fragment or short token verbatim — it now returns a non-reversible SHA-256 prefix (`sha256:<12 hex>`). The revoke log path (sole caller) inherits this automatically, so no raw token material reaches the logs while remaining correlatable.

LifeXplorer: Addressed in 02d0e7e0ec. Short inputs are no longer returned verbatim — `maskToken` now hashes all non-empty inputs to a SHA-256 prefix, so no caller-supplied secret reaches the log in clear text.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthTokenUtils.java
LifeXplorer: Addressed in 02d0e7e0ec. The length/`RANDOM_BYTES` comparison is gone entirely — `maskToken` now hashes the input rather than slicing it, so the encoded-char-vs-raw-byte unit confusion no longer applies.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthTokens.java
LifeXplorer: Done in 02d0e7e0ec — added `McpOAuthTokenUtilsTest` covering null/empty -> "", the hash-prefix output, non-containment of the original token, and that short non-Opik values are masked too.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/McpOAuthBundle.java
LifeXplorer: Done in 02d0e7e0ec — `McpOAuthBundleTest` now asserts the full varargs set `disableExtensions(OAuthMetadataResource.class, OAuthAuthorizeResource.class, OAuthTokenResource.class)` for the disabled case (and the negative case for enabled), so dropping any resource from the gate is now caught.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthTokenResourceTest.java
LifeXplorer: Done in 02d0e7e0ec — every negative test now casts the entity to `OAuthError` and asserts the exact code via an `assertError(...)` helper (`invalid_request` / `invalid_client` / `unsupported_grant_type` / `invalid_grant`), referenced symbolically from `OAuthConstants`. Added a dedicated blank-grant-type case so `invalid_request` and `unsupported_grant_type` are now distinguished.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthTokenResourceTest.java
LifeXplorer: Done in 02d0e7e0ec — both success-path tests now assert the whole `TokenResponse` via `isEqualTo(minted())` instead of individual fields, so a dropped/rewritten `expiresIn`/`workspaceId`/`workspaceName` would be caught.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthTokenResourceTest.java
LifeXplorer: Done in 02d0e7e0ec — added a parameterized `token_refreshGrant_missingFields_invalidRequest` covering blank `refresh_token` and blank `client_id`, asserting `invalid_request` and that `refresh`/`resolve` are never invoked.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthTokenResourceTest.java
```diff
@@ -0,0 +1,150 @@
...
+                CODE_VERIFIER, null);
+
+        assertThat(response.getStatus()).isEqualTo(Response.Status.BAD_REQUEST.getStatusCode());
+        verify(clientService, never()).resolve(any());
+    }
+
+    @Test
+    @DisplayName("POST /token (authorization_code): unknown client_id → invalid_client")
+    void token_authCodeGrant_unknownClient_invalidClient() {
+        when(clientService.resolve(CLIENT_ID)).thenReturn(Optional.empty());
```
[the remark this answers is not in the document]
LifeXplorer: Done in 02d0e7e0ec — the auth_code missing-field test is now parameterized over each required field (`code`, `redirect_uri`, `client_id`, `code_verifier`), and an equivalent parameterized test was added for the refresh_token grant.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthTokenResource.java
```diff
@@ -0,0 +1,130 @@
...
+            } catch (BadRequestException e) {
+                return error(e.getMessage());
+            }
+        }
+
+        if (GRANT_REFRESH_TOKEN.equals(grantType)) {
+            if (StringUtils.isBlank(refreshToken) || StringUtils.isBlank(clientId)) {
+                return error(ERROR_INVALID_REQUEST);
+            }
+            if (clientService.resolve(clientId).isEmpty()) {
```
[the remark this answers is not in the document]
LifeXplorer: Addressed in 02d0e7e0ec, with a clarification: the service only ever throws `BadRequestException(ERROR_INVALID_GRANT)`, so `e.getMessage()` was actually returning the valid RFC 6749 code `invalid_grant` rather than leaking internals. Still, relying on the exception message as the wire contract is fragile, so both catch blocks now `log.warn(...)` the exception and return `error(ERROR_INVALID_GRANT)` explicitly.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthTokenResource.java
```diff
@@ -0,0 +1,130 @@
...
+                .header("Cache-Control", "no-store")
+                .header("Pragma", "no-cache")
+                .build();
+    }
+
+    private static Response error(String code) {
+        return Response.status(Response.Status.BAD_REQUEST)
+                .type(MediaType.APPLICATION_JSON)
+                .header("Cache-Control", "no-store")
+                .header("Pragma", "no-cache")
```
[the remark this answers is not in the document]
LifeXplorer: Acknowledged — leaving as 400 by design. RFC 6749 §5.2 permits 400 for `invalid_client` and only *recommends* 401 + `WWW-Authenticate` when the client attempted to authenticate via an HTTP auth scheme. These are public PKCE clients with no client authentication scheme, so 400 is spec-compliant and `WWW-Authenticate` has no scheme to advertise. Open to revisiting if we add confidential clients.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthTokenResource.java
```diff
@@ -0,0 +1,130 @@
...
+            @FormParam("token") String token,
+            @FormParam("token_type_hint") String tokenTypeHint,
+            @FormParam("client_id") String clientId) {
+
+        // RFC 7009 §2.2: the AS returns 200 whether the token was revoked, never existed, or was invalid.
+        if (!StringUtils.isBlank(token)) {
+            try {
+                mcpOAuthService.revoke(token);
+            } catch (Exception e) {
+                log.warn("MCP OAuth revoke failed [token={}, client_id={}]",
```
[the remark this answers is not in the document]
LifeXplorer: Intentionally not changing this. RFC 7009 §2.2 requires the authorization server to respond 200 regardless of whether the token was valid, revoked, or never existed — returning 400 for a blank/invalid token would break that contract and leak token validity. This is documented in the resource class Javadoc. (The bot tagged this itself as a Breaking Change.)

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthTokenResource.java
LifeXplorer: Valid as defensive hardening, but deferring out of this PR. As the comment notes, codes/tokens are 256-bit random (not brute-forceable) and `@RateLimited` user/workspace buckets do not apply pre-auth, so this needs IP-based throttling at the infra layer. Tracking separately rather than blocking this chunk (4/7).

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/OAuthTokenService.java
```diff
@@ -0,0 +1,101 @@
...
+        }
+    }
+
+    private TokenResponse issueFromAuthorizationCode(String code, String redirectUri, String clientId,
+            String codeVerifier) {
+        if (StringUtils.isBlank(code) || StringUtils.isBlank(redirectUri) || StringUtils.isBlank(clientId)
+                || StringUtils.isBlank(codeVerifier)) {
+            log.warn("MCP OAuth authorization_code request rejected: missing required parameters [client_id={}]",
+                    clientId);
+            throw new OAuthException(ERROR_INVALID_REQUEST);
```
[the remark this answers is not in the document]
LifeXplorer: keeping as is

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthExceptionMapper.java
```diff
@@ -0,0 +1,34 @@
...
+public class OAuthExceptionMapper implements ExceptionMapper<OAuthException> {
+
+    @Override
+    public Response toResponse(OAuthException exception) {
+        log.warn("MCP OAuth request rejected [error={}]", exception.getError());
+        return Response.status(Response.Status.BAD_REQUEST)
+                .type(MediaType.APPLICATION_JSON)
+                .header(HttpHeaders.CACHE_CONTROL, CACHE_CONTROL_NO_STORE)
+                .header(HEADER_PRAGMA, PRAGMA_NO_CACHE)
+                .entity(OAuthError.builder().error(exception.getError()).build())
```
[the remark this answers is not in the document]
LifeXplorer: it logs exact error via `exception.getError`

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthTokenResource.java
```diff
@@ -0,0 +1,110 @@
...
+        Response response = okToken(
+                tokenService.issueToken(grantType, code, redirectUri, clientId, codeVerifier, refreshToken));
+
+        log.info("MCP OAuth token issued '{}', '{}'", grantType, clientId);
+        return response;
+    }
+
+    @POST
+    @Path("/revoke")
+    @Consumes(MediaType.APPLICATION_FORM_URLENCODED)
```
andrescrz, line 80: I believe this one produces JSON, but no annotation. You can add it here, or add both consumes and produces JSON annotations at the class level and keep the overrides here.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/OAuthException.java
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Signals an OAuth protocol error (RFC 6749 §5.2). The {@link #getError() error} code is rendered into the
+ * OAuth error envelope by {@code OAuthExceptionMapper}, so callers never need to assemble the error response
+ * themselves. Extends {@link BadRequestException} so it carries 400 semantics if no mapper is in play.
+ */
+@Getter
+public class OAuthException extends BadRequestException {
+
+    private final String error;
```
andrescrz, line 14: Minor: all exceptions have a message field, not sure what's the value of this `error` field if we could reuse the original.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/OAuthException.java
```diff
@@ -0,0 +1,20 @@
...
+
+    public OAuthException(String error) {
+        super(error);
+        this.error = error;
+    }
```
andrescrz, line 19: Minor: you can probably avoid boiler plate code with Lombok `StandardException`:
- https://projectlombok.org/features/experimental/StandardException

It's experimental, but it should be fine.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthTokenResourceTest.java
```diff
@@ -0,0 +1,148 @@
...
+
+        try (Response response = postToken(form(PARAM_GRANT_TYPE, GRANT_AUTHORIZATION_CODE, PARAM_CLIENT_ID,
+                CLIENT_ID))) {
+            assertThat(response.getStatus()).isEqualTo(Response.Status.OK.getStatusCode());
+            assertNoStore(response);
+
+            response.bufferEntity();
+            assertThat(response.readEntity(String.class))
+                    .contains("\"access_token\"", "\"refresh_token\"", "\"token_type\"", "\"expires_in\"",
+                            "\"workspace_id\"", "\"workspace_name\"");
```
andrescrz, line 107: Minor: not sure what's the value of reading and asserting the String version of the entity. This is Java, let's just work with objects, equality check there should be fine.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthTokenResourceTest.java
```diff
@@ -0,0 +1,148 @@
...
+    void token_serviceRejects_rendersOAuthError(String errorCode) {
+        when(tokenService.issueToken(any(), any(), any(), any(), any(), any()))
+                .thenThrow(new OAuthException(errorCode));
+
+        try (Response response = postToken(form(PARAM_GRANT_TYPE, GRANT_AUTHORIZATION_CODE, PARAM_CLIENT_ID,
+                CLIENT_ID))) {
+            assertThat(response.getStatus()).isEqualTo(Response.Status.BAD_REQUEST.getStatusCode());
+            assertThat(response.getMediaType().toString()).startsWith(MediaType.APPLICATION_JSON);
+            assertNoStore(response);
+            assertThat(response.readEntity(OAuthError.class).error()).isEqualTo(errorCode);
```
andrescrz, line 125: Minor: as usual, assert expected vs actual record/object, not just individual fields.

apps/opik-backend/src/test/java/com/comet/opik/domain/mcpoauth/OAuthTokenServiceTest.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    private TokenResponse minted() {
+        return TokenResponse.builder()
+                .accessToken(ACCESS_PREFIX + "xxx")
+                .refreshToken(REFRESH_PREFIX + "yyy")
+                .tokenType("Bearer")
+                .expiresIn(3600L)
+                .workspaceId("ws-1")
+                .workspaceName("default")
```
andrescrz, line 61: Nit: as usual, better use random values with the usual options: Podam, RandomStringUtils, UUID etc.

andrescrz (approved):
LGTM, the only important that you could address is the produces annotation. But not a blocker.

That one and the rest can be addressed in the next PRs.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/OAuthException.java
LifeXplorer: `OAuthException` was introduced for ability to use `OAuthExceptionMapper`, a centralized place to control the error body shape and to set spec-required headers

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthTokenResource.java
LifeXplorer: updated in next chunk PR

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/OAuthException.java
LifeXplorer: Current implementation is few lines of code. I'd push back on this

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthTokenResourceTest.java
LifeXplorer: the test verifies the values round-trip and `snake_case` wire format.
I'd keep it as not-bad to have.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthTokenResourceTest.java
LifeXplorer: There are no repeatable object classes to assert on.

apps/opik-backend/src/test/java/com/comet/opik/domain/mcpoauth/OAuthTokenServiceTest.java
LifeXplorer: applied in next chunk PR

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/OAuthException.java
andrescrz: Yeah, having `OAuthException` makes sense. My main question is why having `String error` field when all exceptions have a `String message` field already.

PR #6996, opened 2026-06-08
[OPIK-6666] [BE] feat: MCP OAuth dynamic client registration — RFC 7591 (PR 5/7)

LifeXplorer:
## Details

**PR 5 of 7** splitting #6900 per [review](https://github.com/comet-ml/opik/pull/6900#pullrequestreview-4428853723).

> ⛓️ **Chained PR.** Base is `avinahradau/OPIK-6666-oauth-as-chunked-4` (PR #6995). Base retargets to `main` once PR-4 lands.

| Item | What |
|---|---|
| `OAuthRegisterResource` | `POST /oauth/register` — RFC 7591 dynamic client registration for AI hosts; IP-bucketed rate limiting via `RateLimitService` |
| `ClientRegistrationResponse` | response DTO |
| `McpOAuthBundle` | now also disables `OAuthRegisterResource` when flag off |
| `OAuthRegisterResourceTest` | coverage |

**Note:** adapted to PR-1's slimmed `McpOAuthClient` (no `createdAt`). `client_id_issued_at` is optional (RFC 7591 §3.2.1) and is omitted from the response; the test asserts that.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Testing
- `mvn -o test-compile` — green.

## Documentation
No user-facing documentation changes required — internal MCP OAuth registration endpoint, not part of the public API surface.

## Issues


apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthRegisterResource.java
```diff
@@ -0,0 +1,88 @@
...
+                .build();
+    }
+
+    /**
+     * First X-Forwarded-For hop set by the fronting nginx; direct remote address otherwise.
+     */
+    private static String clientIp(HttpServletRequest request) {
+        String forwarded = request.getHeader("X-Forwarded-For");
+        if (StringUtils.isNotBlank(forwarded)) {
+            return forwarded.split(",")[0].trim();
```
LifeXplorer: **[medium] Rate-limit bucket keyed on attacker-controllable X-Forwarded-For**

clientIp() takes forwarded.split(",")[0] — the left-most X-Forwarded-For hop. nginx appends to (rather than replaces) X-Forwarded-For, so the left-most value is the client-supplied portion of the header, which any caller can set to an arbitrary string. Since this IP is the only key in RATE_LIMIT_BUCKET and rate limiting is the sole abuse control on this deliberately-unauthenticated DCR endpoint (RFC 7591), an attacker can rotate the X-Forwarded-For value per request and completely bypass the per-IP registration throttle, defeating the spam protection described in the class javadoc.

💡 Do not trust the left-most XFF hop. Either use request.getRemoteAddr() (the connecting peer, i.e. nginx, with a per-deployment trusted-proxy/real-IP configuration) or, when behind a single trusted proxy, take the right-most XFF entry that the trusted proxy appended. Avoid trusting client-supplied left-most values for security-relevant bucketing.

_rule: `backend-security-validate-and-restrict-user-supplied-urls-before-making-outbo`_

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthRegisterResourceTest.java
```diff
@@ -0,0 +1,127 @@
...
+        when(clientService.register(any(ClientRegistrationRequest.class))).thenReturn(minted);
+
+        Response response = resource.register(ClientRegistrationRequest.builder()
+                .clientName("Test Client")
+                .redirectUris(Set.of("http://example.com/cb"))
+                .build(), httpRequest);
+
+        assertThat(response.getStatus()).isEqualTo(Response.Status.CREATED.getStatusCode());
+        assertThat(response.getLocation().getPath()).isEqualTo("/admin/mcp-oauth-clients/" + clientId);
+        ClientRegistrationResponse body = (ClientRegistrationResponse) response.getEntity();
```
LifeXplorer: **[medium] Happy-path test asserts only two fields of the mapped response**

register_returns201WithLocationHeader asserts only body.clientId() and body.clientName(). The mapper (ClientRegistrationResponseMapper) also populates redirectUris (from the client), tokenEndpointAuthMethod=AUTH_METHOD_NONE, grantTypes=DEFAULT_GRANT_TYPES, and responseTypes=DEFAULT_RESPONSE_TYPES. None of these mapped fields are asserted, so a regression in any of the @Mapping expressions (wrong auth method, dropped redirectUris, empty grant/response types) would pass undetected. This is the main behavioral contract of the register endpoint's response body.

💡 Build the expected ClientRegistrationResponse and assert the whole object via usingRecursiveComparison() (or at minimum assert redirectUris, tokenEndpointAuthMethod, grantTypes, responseTypes) so the full RFC 7591 registration response contract is covered.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthRegisterResourceTest.java
```diff
@@ -0,0 +1,127 @@
...
+
+    @Test
+    @DisplayName("register: response body omits clientIdIssuedAt (not surfaced by the client model)")
+    void register_omitsIssuedAtWhenCreatedAtNull() {
+        McpOAuthClient minted = McpOAuthClient.builder()
+                .clientId("client-456")
+                .name("No-Timestamp Client")
+                .redirectUris(Set.of("http://example.com/cb"))
+                .build();
+        when(clientService.register(any(ClientRegistrationRequest.class))).thenReturn(minted);
```
LifeXplorer: **[low] Misleading test method name implies conditional behavior that does not exist**

The method name register_omitsIssuedAtWhenCreatedAtNull() and the implied 'WhenCreatedAtNull' condition describe behavior that the production code does not have. McpOAuthClient has no createdAt/clientIdIssuedAt field, and ClientRegistrationResponseMapper declares @Mapping(target = "clientIdIssuedAt", ignore = true) which unconditionally leaves clientIdIssuedAt null for every client. There is no null-vs-non-null branch to exercise, so the name contradicts both the actual contract and the test's own @DisplayName ('not surfaced by the client model'). A reader will assume a conditional code path exists when it does not.

💡 Rename to reflect the unconditional behavior, e.g. register_clientIdIssuedAtAlwaysOmitted(), and align the @DisplayName so the report does not suggest a non-existent CreatedAt-null branch.

_rule: `backend-testing-use-scenario-based-test-method-names-keep-displayname-accura`_

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthRegisterResource.java
```diff
@@ -0,0 +1,88 @@
...
+     * RFC 7591 §3.2.1: the response SHOULD include a Location header pointing at a client-configuration
+     * endpoint. The URI is informational and not part of the OAuth dance.
+     */
+    @POST
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response register(@NonNull @Valid ClientRegistrationRequest request,
+            @Context HttpServletRequest httpRequest) {
+
+        boolean exceeded = Boolean.TRUE.equals(rateLimitService
```
LifeXplorer: **[low] Registration throttle ignores global rateLimit.enabled flag**

RateLimitInterceptor gates all @RateLimited enforcement on rateLimitConfig.isEnabled(). This resource calls RateLimitService.isLimitExceeded(...) directly and unconditionally, so OAuth registration is throttled even in deployments that have globally disabled rate limiting (and the rate-limit Redis bucket is always created). This is a divergence from the established rate-limiting contract; if intentional it should be documented, otherwise it is surprising operationally.

💡 If the registration throttle is meant to honor the platform-wide switch, inject RateLimitConfig and skip the check (or short-circuit) when rateLimit.enabled is false; otherwise add a comment clarifying that this throttle is intentionally independent of the global flag because the endpoint is unauthenticated.

_general finding_

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthRegisterResource.java
```diff
@@ -0,0 +1,88 @@
...
+                .isLimitExceeded(1, RATE_LIMIT_BUCKET.formatted(clientIp(httpRequest)), limitConfig)
+                .block());
+        if (exceeded) {
+            return Response.status(Response.Status.TOO_MANY_REQUESTS)
```
LifeXplorer: **[low] 429 response omits Retry-After / rate-limit headers**

The TOO_MANY_REQUESTS response returns only an OAuthError body. Unlike RateLimitInterceptor (which calls setLimitHeaders to emit remaining-quota and TTL headers), this path sets no Retry-After or X-RateLimit-* headers, so well-behaved clients cannot back off intelligently. The remaining TTL is already obtainable via RateLimitService.getRemainingTTL.

💡 Add a Retry-After header (seconds until the bucket refills, from getRemainingTTL) to the 429 response so clients can honor the throttle window.

_general finding_

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/OAuthConstants.java
```diff
@@ -35,4 +39,6 @@ public class OAuthConstants {
     public static final String ERROR_INVALID_TARGET = "invalid_target";
     public static final String ERROR_UNSUPPORTED_GRANT_TYPE = "unsupported_grant_type";
     public static final String ERROR_UNSUPPORTED_RESPONSE_TYPE = "unsupported_response_type";
+
+    public static final String RATE_LIMIT_BUCKET = "mcp_oauth_register:%s";
```
LifeXplorer: **[low] Rate-limit key prefix duplicated between constant and LimitConfig name**

RATE_LIMIT_BUCKET embeds the literal 'mcp_oauth_register' as the Redis rate-limit key prefix. The same literal is also hardcoded independently in OAuthRegisterResource (line 43) as the LimitConfig name argument (new LimitConfig("McpOAuthRegister", "mcp_oauth_register", ...)). These two represent the same logical bucket identifier but are defined in two places, so they can silently drift if one is renamed. The rule calls out centralizing shared string literals such as Redis key prefixes into one constant.

💡 Extract the bare prefix (e.g. public static final String RATE_LIMIT_BUCKET_PREFIX = "mcp_oauth_register") once, derive RATE_LIMIT_BUCKET = RATE_LIMIT_BUCKET_PREFIX + ":%s", and pass RATE_LIMIT_BUCKET_PREFIX as the LimitConfig name in OAuthRegisterResource so both references share a single source of truth.

_rule: `backend-style-maintainability-eliminate-duplication-merge-near-duplicate-constants-unify-a`_

LifeXplorer (commented):

6 findings — 0 high · 2 medium · 4 low. Suppressed by team conventions: see `suppressed.md`.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthRegisterResourceTest.java
```diff
@@ -0,0 +1,127 @@
...
+
+    @InjectMocks
+    private OAuthRegisterResource resource;
+
+    @BeforeEach
+    void setUp() {
+        lenient().when(opikConfig.getMcpOAuth()).thenReturn(new McpOAuthConfig());
+        lenient().when(httpRequest.getRemoteAddr()).thenReturn("10.0.0.1");
+        lenient().when(rateLimitService.isLimitExceeded(anyLong(), anyString(), any()))
+                .thenReturn(Mono.just(false));
```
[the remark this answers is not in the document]
LifeXplorer: Fixed in 34fddcf. You're right that `McpOAuthConfig` was uninitialized — and since `LimitConfig` is now built in the constructor (commit d5aae14), `@InjectMocks` actually NPE'd at construction time, *before* `setUp()` could stub `getMcpOAuth()`, so all 4 tests were erroring. The resource is now constructed manually in `setUp()` after stubbing a populated `McpOAuthConfig` (`registrationRateLimit` + `registrationRateLimitDuration`). Tests are green.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthRegisterResource.java
```diff
@@ -0,0 +1,90 @@
...
+
+    private final @NonNull OAuthClientService clientService;
+    private final @NonNull RateLimitService rateLimitService;
+    private final @NonNull OpikConfiguration opikConfig;
+
+    @POST
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response register(@NonNull ClientRegistrationRequest request,
+            @Context HttpServletRequest httpRequest) {
```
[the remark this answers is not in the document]
LifeXplorer: Confirmed — `@Valid` was added in d5aae14 and remains in place on the `register` parameter, so the `@NotEmpty`/`@Size`/`@AbsoluteUri` constraints fire before `clientService.register(request)`. No further change needed.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthRegisterResource.java
```diff
@@ -0,0 +1,90 @@
...
+                .entity(body)
+                .build();
+    }
+
+    // First X-Forwarded-For hop set by the fronting nginx; direct remote address otherwise.
+    private static String clientIp(HttpServletRequest request) {
+        String forwarded = request.getHeader("X-Forwarded-For");
+        if (StringUtils.isNotBlank(forwarded)) {
+            return forwarded.split(",")[0].trim();
+        }
```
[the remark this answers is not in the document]
LifeXplorer: Fixed in 34fddcf. `clientIp()` now keys on the **right-most** X-Forwarded-For hop — the address nginx appends via `$proxy_add_x_forwarded_for`, which the caller can't control — and falls back to `getRemoteAddr()` when the header is absent. Updated the javadoc and the test (`register_usesLastForwardedForHopAsRateLimitKey`) to prove the spoofable first hop is ignored.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthRegisterResource.java
LifeXplorer: Fixed in 34fddcf. The bucket now keys on the right-most (nginx-appended) X-Forwarded-For hop instead of the attacker-controlled left-most value, falling back to `getRemoteAddr()` when the header is absent. Per-IP bucketing is preserved while the spoofing bypass is closed.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthRegisterResourceTest.java
LifeXplorer: Fixed in 34fddcf. The happy-path test now builds the expected `ClientRegistrationResponse` and asserts the whole object via `usingRecursiveComparison()`, so `redirectUris`, `tokenEndpointAuthMethod=none`, `grantTypes`, and `responseTypes` are all covered — a regression in any `@Mapping` expression now fails the test.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/ClientRegistrationResponseMapper.java
```diff
@@ -0,0 +1,20 @@
...
+@Mapper(imports = OAuthConstants.class)
+interface ClientRegistrationResponseMapper {
+
+    ClientRegistrationResponseMapper INSTANCE = Mappers.getMapper(ClientRegistrationResponseMapper.class);
+
+    @Mapping(target = "clientName", source = "name")
+    @Mapping(target = "clientIdIssuedAt", ignore = true)
+    @Mapping(target = "tokenEndpointAuthMethod", expression = "java(OAuthConstants.AUTH_METHOD_NONE)")
+    @Mapping(target = "grantTypes", expression = "java(OAuthConstants.DEFAULT_GRANT_TYPES)")
+    @Mapping(target = "responseTypes", expression = "java(OAuthConstants.DEFAULT_RESPONSE_TYPES)")
```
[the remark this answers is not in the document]
LifeXplorer: Addressed in 34fddcf. Rather than a dedicated mapper test, the register happy-path test now asserts the full `ClientRegistrationResponse` via `usingRecursiveComparison()`, which covers `tokenEndpointAuthMethod`, `grantTypes`, `responseTypes`, and the null-omission of `clientIdIssuedAt` end-to-end through the endpoint.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthRegisterResourceTest.java
LifeXplorer: Fixed in 34fddcf. Renamed to `register_clientIdIssuedAtAlwaysOmitted` and adjusted the @DisplayName to reflect the unconditional omission — there's no CreatedAt-null branch to imply.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthRegisterResource.java
LifeXplorer: Addressed in 34fddcf with a clarifying comment. This is intentional: DCR is unauthenticated and the per-IP cap is its only abuse control, so it must stay active even in deployments that disable global rate limiting. Documented inline so it's not surprising.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthRegisterResource.java
LifeXplorer: Fixed in 34fddcf. The 429 response now sets a `Retry-After` header (seconds until the bucket refills, derived from `getRemainingTTL`), and the test asserts it.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/OAuthConstants.java
LifeXplorer: Fixed in 34fddcf. Extracted `RATE_LIMIT_BUCKET_PREFIX = "mcp_oauth_register"` in `OAuthConstants`; `RATE_LIMIT_BUCKET` is now `RATE_LIMIT_BUCKET_PREFIX + ":%s"` and the resource passes `RATE_LIMIT_BUCKET_PREFIX` as the `LimitConfig` name, so both share one source of truth.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthRegisterResource.java
```diff
@@ -0,0 +1,88 @@
...
+                .block());
+        if (exceeded) {
+            return Response.status(Response.Status.TOO_MANY_REQUESTS)
+                    .type(MediaType.APPLICATION_JSON)
+                    .entity(OAuthError.builder()
+                            .error("too_many_requests")
+                            .errorDescription("registration rate limit exceeded")
+                            .build())
+                    .build();
+        }
```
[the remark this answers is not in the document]
LifeXplorer: Fixed in 34fddcf. The rate-limit test now asserts the serialized body (`error=too_many_requests`, `error_description=registration rate limit exceeded`) and the new `Retry-After` header, in addition to the 429 status and the skipped `clientService.register(...)`.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthRegisterResource.java
```diff
@@ -0,0 +1,115 @@
...
+                    .entity(OAuthError.builder()
+                            .error(ERROR_TOO_MANY_REQUESTS)
+                            .errorDescription(ERROR_DESC_REGISTRATION_RATE_LIMIT)
+                            .build())
+                    .build();
+        }
+
+        McpOAuthClient client = clientService.register(request);
+        ClientRegistrationResponse body = ClientRegistrationResponseMapper.INSTANCE.toResponse(client);
+        return Response.created(URI.create(CLIENT_CONFIG_PATH_PREFIX + client.id()))
```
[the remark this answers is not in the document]
LifeXplorer: Comment not relevant. Part of DCR self-register

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthRegisterResource.java
```diff
@@ -0,0 +1,102 @@
...
+                    .entity(OAuthError.builder()
+                            .error("too_many_requests")
+                            .errorDescription("registration rate limit exceeded")
+                            .build())
+                    .build();
+        }
+
+        McpOAuthClient client = clientService.register(request);
+        ClientRegistrationResponse body = ClientRegistrationResponseMapper.INSTANCE.toResponse(client);
+        return Response.created(URI.create("/admin/mcp-oauth-clients/" + client.id()))
```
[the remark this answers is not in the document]
LifeXplorer: already tested

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthRegisterResource.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /**
+     * DCR is open (no auth) by design — throttle per source IP to bound spam registrations.
+     * RFC 7591 §3.2.1: the response SHOULD include a Location header pointing at a client-configuration
+     * endpoint. The URI is informational and not part of the OAuth dance.
+     */
+    @POST
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
```
andrescrz: Minor: better define these as the defaults at the top of the class.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthRegisterResource.java
```diff
@@ -0,0 +1,118 @@
...
+    public Response register(@NonNull @Valid ClientRegistrationRequest request,
+            @Context HttpServletRequest httpRequest) {
+
+        log.info("MCP OAuth client registration request '{}'", request.clientName());
+
+        String bucket = RATE_LIMIT_BUCKET.formatted(clientIp(httpRequest));
+        boolean exceeded = Boolean.TRUE.equals(rateLimitService
+                .isLimitExceeded(1, bucket, limitConfig)
+                .block());
+        if (exceeded) {
```
andrescrz: Why do you need a customer implementation instead of using `@RateLimited` annotation. If not directly usable, you should explore the option to make generic adjustments that we can reuse in the future in the `RateLimitInterceptor` class etc.

This is infrastructure code, it's the part of the responsibilities of the endpoint.

andrescrz (commented):
This generally LGTM for the particular scope of the MCP initiative. 

However, there's the rate limiting implementation, that I believe we should revisit before moving forward.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthRegisterResource.java
LifeXplorer: updated

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthRegisterResource.java
LifeXplorer: Refactored `@RateLimited` and `RateLimitInterceptor` to throttle un-authorized requests by IP

PR #6997, opened 2026-06-08
[OPIK-6666] [BE] feat: MCP OAuth data-API bearer integration + validator + scrub job (PR 6/7)

LifeXplorer:
## Details

**PR 6 of 7** splitting #6900 per [review](https://github.com/comet-ml/opik/pull/6900#pullrequestreview-4428853723). This is the "make it work end-to-end" chunk.

> ⛓️ **Chained PR.** Base is `avinahradau/OPIK-6666-oauth-as-chunked-5` (PR #6996). Base retargets to `main` once PR-5 lands.

| Item | What |
|---|---|
| `AuthFilter` | recognises `Authorization: Bearer opik_at_…` on `/v1/private/*`. Cloud delegates to comet-backend via `AuthService.authorizeOAuth`; OSS sets `RequestContext` from the token. Gated on `MCP_OAUTH_ENABLED` — a real kill switch for previously-minted tokens |
| `OAuthValidateResource` | `POST /opik/auth-oauth` — opik-mcp bearer introspection → identity JSON |
| `McpOAuthScrubJob` | every 5 min (`MCP_OAUTH_SCRUB_CRON`) deletes expired codes + expired/revoked tokens past the rotation grace |
| `OpikConfiguration` → `MutableJobConfiguration` + `config.yml` `jobs.mcpOAuthScrubCron` | required for the scrub job's `@On` cron to deserialize (deferred here from PR-1 by design) |
| `McpOAuthBundle` | now also disables `OAuthValidateResource` + `McpOAuthScrubJob` when flag off |
| `McpOAuthFlowE2ETest` / `McpOAuthDisabledE2ETest` | full authorize→token→bearer flow; kill-switch 404s |

**Cloud dependency:** the bearer path delegates to comet-ml/comet-backend#5569 (`POST /opik/auth-by-username`), which must land before enabling on Cloud/STSaaS.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Testing
- `mvn -o test-compile` — green. E2E tests included.

## Issues
- Jira: [OPIK-6666](https://comet-ml.atlassian.net/browse/OPIK-6666) · Full reference: #6900 · Companion: comet-ml/comet-backend#5569

## Documentation
No user-facing documentation changes. MCP OAuth is gated behind `MCP_OAUTH_ENABLED` (off by default) and is not yet enabled on Cloud/STSaaS.


apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/McpOAuthBundle.java
```diff
@@ -23,7 +24,9 @@ public void run(GuiceyEnvironment environment) {
                     OAuthMetadataResource.class,
                     OAuthAuthorizeResource.class,
                     OAuthTokenResource.class,
-                    OAuthRegisterResource.class);
+                    OAuthRegisterResource.class,
+                    OAuthValidateResource.class,
+                    McpOAuthScrubJob.class);
```
LifeXplorer, line 29: **[high] Bundle unit test out of sync with the new disableExtensions argument list**

The diff extends disableExtensions(...) to six classes (adding OAuthValidateResource.class and McpOAuthScrubJob.class), but the companion test McpOAuthBundleTest verifies the call with only the original four classes (verify(environment).disableExtensions(OAuthMetadataResource.class, OAuthAuthorizeResource.class, OAuthTokenResource.class, OAuthRegisterResource.class) at McpOAuthBundleTest.java:40, and verify(...never()...) at :56). GuiceyEnvironment.disableExtensions is varargs (Class<?>...), so Mockito matches the exact argument array; the actual six-arg invocation will not match the four-arg expectation and 'disablesResourcesWhenDisabled' will fail (argument mismatch), while 'keepsResourcesWhenEnabled' becomes a weaker/incorrect assertion. The test must be updated in the same change that broadened the production call.

💡 Update McpOAuthBundleTest to verify disableExtensions with all six classes including OAuthValidateResource.class and McpOAuthScrubJob.class, in both the disabled (verify) and enabled (verify ... never()) cases.

_rule: `backend-testing-add-regression-tests-for-every-new-backend-behavior`_

LifeXplorer, line 29: **[medium] No test verifies the scrub job is actually not scheduled when MCP OAuth is disabled**

McpOAuthScrubJob is a data-deleting background job (deletes OAuth codes/tokens every 5 minutes via @On). Its kill-switch relies on a non-obvious mechanism: jobs are scheduled by GuiceJobManager.getJobs(injector), which scans injector.getBindings() for any Job-typed binding and schedules each one independently of guicey's installer. The guicey JobGuiceyInstaller.install only does environment.jersey().register(job); the actual scheduling is driven by GuiceJobManager. disableExtensions(McpOAuthScrubJob.class) only prevents the job from being scheduled if guicey also removes its Guice binding so GuiceJobManager no longer finds it. McpOAuthDisabledE2ETest verifies the JAX-RS resources return 404 when disabled, but nothing verifies the scrub job is not scheduled/does not run when the feature is off — yet that is the load-bearing, hardest-to-reason-about part of this change. Without coverage, a future guicey/dropwizard-jobs upgrade or refactor could silently re-enable a destructive job in deployments that have the feature turned off.

💡 Add a focused test asserting that when mcpOAuth.enabled=false the McpOAuthScrubJob is not scheduled (e.g. assert it is absent from the GuiceJobManager's scheduled jobs / Quartz scheduler, or that no McpOAuthScrubJob binding remains in the injector after the bundle runs), mirroring the existing disabled-resource E2E coverage.

_rule: `backend-testing-add-regression-tests-for-every-new-backend-behavior`_

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJob.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /**
+     * Revoked tokens within the rotation grace window must remain queryable, so a duplicate refresh
+     * request (RFC 6749 §6 token rotation grace) returns invalid_grant instead of triggering reuse detection.
+     */
+    @Override
+    public void doJob(JobExecutionContext context) {
+        Instant now = Instant.now();
+        Instant tokenThreshold = now.minus(opikConfig.getMcpOAuth().getRefreshRotationGrace());
```
LifeXplorer: **[low] Rotation-grace threshold is also applied to non-revoked expired tokens**

tokenThreshold = now - refreshRotationGrace is passed to deleteExpiredAndRevoked, whose SQL is `(revoked_at IS NOT NULL AND revoked_at < :threshold) OR (expires_at < :threshold)`. The grace window is only semantically justified for revoked tokens (the Javadoc says revoked tokens must remain queryable for the grace window), but the same shifted threshold also gates the `expires_at < :threshold` branch. As a result an expired-but-never-revoked token is retained an extra refreshRotationGrace (30s default) beyond its real expiry. Functionally harmless (tokens are already rejected as expired) but the cutoff for the two branches is conceptually different and the single shared threshold conflates them.

💡 If the grace is intended only for the revoked branch, pass `now` for the expiry branch and `now - grace` for the revoked branch (e.g. two bind params: `expires_at < :now OR (revoked_at IS NOT NULL AND revoked_at < :graceThreshold)`), or document that the grace deliberately applies to both.

_general finding_

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJob.java
```diff
@@ -0,0 +1,59 @@
...
+                    handle -> handle.attach(McpOAuthCodeDAO.class).deleteExpired(now));
+            tokens = template.inTransaction(WRITE,
+                    handle -> handle.attach(McpOAuthTokenDAO.class).deleteExpiredAndRevoked(tokenThreshold));
+        } catch (RuntimeException e) {
+            log.error("MCP OAuth scrub failed", e);
+            return;
+        }
+
+        if (codes > 0 || tokens > 0) {
+            log.info("MCP OAuth scrub: removed {} expired codes, {} expired/revoked tokens", codes, tokens);
```
LifeXplorer: **[low] Partial-failure path loses the codes-deleted count and silently returns**

The two deletes run in separate WRITE transactions (lines 52-55). If the codes delete commits but the tokens delete then throws, the catch logs only the generic 'MCP OAuth scrub failed' and returns, so the count of codes already deleted in this run is never surfaced. The error log line carries no run context (e.g. how many codes were removed before the failure), making it harder to reason about partial progress across the two independent transactions.

💡 Move the summary/count logging so any already-completed work is reported even on partial failure, or include the codes-deleted count in the error log (e.g. `log.error("MCP OAuth scrub failed after removing {} codes", codes, e)`).

_rule: `backend-observability-include-all-correlation-ids-and-context-in-warn-error-log-me`_

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthFilter.java
```diff
@@ -33,11 +41,20 @@ public void filter(ContainerRequestContext context) throws IOException {
...
                     .uriInfo(uriInfo)
                     .method(context.getMethod())
                     .requiredPermissions(getRequiredPermissions(context))
-                    .build());
+                    .build();
+            String authHeader = context.getHeaderString(HttpHeaders.AUTHORIZATION);
+            if (opikConfig.getMcpOAuth().isEnabled() && McpOAuthTokenUtils.isMcpOAuthToken(authHeader)) {
+                String token = authHeader.substring(BEARER_PREFIX.length()).trim();
+                ValidatedToken validatedToken = mcpOAuthService.validateAccessTokenForWorkspace(
+                        token, context.getHeaderString(RequestContext.WORKSPACE_HEADER));
```
LifeXplorer, line 52: **[low] OAuth path ignores the workspace_name query-param fallback used by the standard auth path**

The OAuth branch resolves the requested workspace only from the WORKSPACE_HEADER. The standard path (RemoteAuthService.authenticate) falls back to the WORKSPACE_QUERY_PARAM (?workspace_name=) when the header is absent. A blank header is treated by McpOAuthService.workspaceMatches as 'no preference' and always matches the token's workspace, so a caller pinning a workspace via the query param instead of the header would have that pin silently ignored rather than enforced. This is consistent with the intended blank=match design, so impact is limited, but the two auth paths handle workspace resolution inconsistently.

💡 If query-param workspace pinning should also constrain OAuth requests, resolve the workspace using the same header-then-query-param precedence (e.g. via the existing WorkspaceUtils/RequestContext helper) before calling validateAccessTokenForWorkspace; otherwise document that OAuth only honors the workspace header.

_general finding_

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/McpOAuthDisabledE2ETest.java
```diff
@@ -0,0 +1,113 @@
...
+    @DisplayName("auth filter ignores opik_at_ bearer tokens when disabled (no token-store validation)")
+    void authFilterIgnoresOAuthBearerWhenDisabled() {
+        // OSS mode with OAuth disabled: the request must fall through to the regular (no-op) auth
+        // path and succeed. Without the kill-switch gate, the filter would try to validate the
+        // token against the store and reject the request with 401.
+        try (var response = client.target(baseURI + "/v1/private/projects")
+                .request()
+                .header(HttpHeaders.AUTHORIZATION, "Bearer opik_at_not-validated-when-disabled")
+                .get()) {
+            assertThat(response.getStatus()).isEqualTo(Response.Status.OK.getStatusCode());
```
LifeXplorer: **[low] Status-only assertion weakens the kill-switch claim**

authFilterIgnoresOAuthBearerWhenDisabled asserts only that GET /v1/private/projects returns 200, never inspecting the body. In OSS no-op auth mode the request would also return 200 with no Authorization header at all, so a status-only check does not strongly demonstrate that the opik_at_ bearer was actually ignored versus simply accepted. The test still verifies the negative (the gate did not route into validateAccessTokenForWorkspace, which would 401/500), but asserting the body would make the 'token ignored, request succeeds normally' claim concrete.

💡 After the status check, read and assert the response body (e.g. that it deserializes to the expected project list / page shape) so the test confirms the request fell through to the regular auth path and returned valid data, not just any 200.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/McpOAuthFlowE2ETest.java
```diff
@@ -0,0 +1,431 @@
...
+            assertThat(response.getStatus()).isEqualTo(Response.Status.OK.getStatusCode());
+        }
+
+        await().pollDelay(ROTATION_GRACE_MS + 500, TimeUnit.MILLISECONDS)
+                .atMost(ROTATION_GRACE_MS + 5_000, TimeUnit.MILLISECONDS)
+                .until(() -> true);
+
+        // replaying the rotated token after grace is reuse: invalid_grant + family revocation
+        try (var response = refreshToken(clientId, oldRefresh)) {
+            assertOAuthError(response, "invalid_grant");
```
LifeXplorer: **[low] await().pollDelay(...).until(() -> true) is a fixed sleep, not a real Awaitility poll**

This `await().pollDelay(ROTATION_GRACE_MS + 500).atMost(...).until(() -> true)` polls no observable condition — `until(() -> true)` is immediately satisfied, so it is just a wall-clock delay dressed up as Awaitility (effectively a Thread.sleep of ROTATION_GRACE_MS+500ms). Sibling tests (AlertResourceTest, TracesResourceTest) use pollDelay together with untilAsserted/until on a real condition. Because the thing being waited on here is a server-side grace-window expiry with no externally observable state to poll, this is the least-bad option, but it bakes in a fixed timing assumption that can flake under load and lengthens the suite. Note as defensive/maintainability, not a functional bug.

💡 If any observable signal exists for grace expiry, poll it with untilAsserted. Otherwise add a short comment that this is an intentional wall-clock wait for the grace window, and keep the buffer (the +500ms) explicit; consider lowering the configured grace further to shorten the suite.

_rule: `backend-testing-keep-tests-deterministic-replace-thread-sleep-with-awaitilit`_

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/McpOAuthFlowE2ETest.java
```diff
@@ -0,0 +1,431 @@
...
+                .header(HttpHeaders.AUTHORIZATION, "Bearer " + accessToken)
+                .header(WORKSPACE_HEADER, "some-other-workspace")
+                .get()) {
+            assertThat(response.getStatus()).isEqualTo(Response.Status.FORBIDDEN.getStatusCode());
+        }
+    }
+
+    @Test
+    @DisplayName("authorization code is single-use: second exchange fails")
+    void codeDoubleSpendRejected() {
```
LifeXplorer: **[low] Several error-path assertions check only HTTP status, not the response body**

The 403/401 negative paths (line 286 workspace mismatch; line 380 and 399 revoked access token; line 425 garbage bearer; line 434 unknown bearer) assert only the status code. The team convention for error-path tests is to also assert the error body. For the OAuth /oauth/token paths the test correctly uses assertOAuthError (status + error field), but these auth-filter 401/403 responses are status-only. Lower severity because these are framework-level auth rejections whose body is generic and the security-relevant outcome (rejection) is captured by the status.

💡 Where the auth filter returns a structured ErrorMessage body, assert its message too; if the body is intentionally empty/generic for 401/403, leave a brief comment noting that status is the contract being verified.

_rule: `backend-testing-integration-tests-must-verify-behavior-through-public-apis-u`_

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthValidateTokenResourceTest.java
```diff
@@ -0,0 +1,89 @@
...
+    @Mock
+    private McpOAuthService mcpOAuthService;
+
+    private OAuthValidateResource resource() {
+        return new OAuthValidateResource(mcpOAuthService);
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {
+            "", // blank
```
LifeXplorer: **[low] Inaccurate inline comment: empty string labeled 'blank'**

The first @ValueSource case is the empty string "" with the comment '// blank'. An empty string is not 'blank' in the StringUtils sense, and the production guard (Strings.CI.startsWith(null/"", BEARER_PREFIX)) treats it as a non-Bearer header rather than a blank-token rejection. The comment slightly misdescribes what the case exercises. A truly blank-but-nonempty header (e.g. a whitespace-only string, or 'Bearer    ' that trims to empty) is not covered.

💡 Relabel the '' case as '// empty' and consider adding a case like 'Bearer    ' (whitespace-only token after the scheme) to cover the trim()-to-empty path in OAuthValidateResource.validate, which currently has no test.

_general finding_

apps/opik-backend/src/test/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJobTest.java
```diff
@@ -0,0 +1,79 @@
...
+        verify(tokenDAO).deleteExpiredAndRevoked(tokenThreshold.capture());
+
+        // Revoked tokens must remain queryable for the grace window, so their cutoff trails "now" by exactly the grace.
+        assertThat(tokenThreshold.getValue()).isEqualTo(codeNow.getValue().minus(ROTATION_GRACE));
+    }
+
+    @Test
+    @DisplayName("a DAO failure is swallowed so a transient DB error never kills the scheduler")
+    void swallowsRuntimeException() {
+        when(tokenDAO.deleteExpiredAndRevoked(any())).thenThrow(new RuntimeException("db down"));
```
LifeXplorer: **[low] swallowsRuntimeException only covers the second DAO call, not the first**

The catch-all in McpOAuthScrubJob wraps two separate template.inTransaction calls: deleteExpired (codes) and deleteExpiredAndRevoked (tokens). The test only stubs tokenDAO.deleteExpiredAndRevoked to throw, so it exercises the failure of the second call while the first returns the mock default (0). A regression that lets an exception from the first call (codeDAO.deleteExpired) escape would not be caught by this test. The DisplayName claims 'a DAO failure is swallowed' generically, but only one of the two DAO paths is verified.

💡 Parameterize the test (@ParameterizedTest) or add a second case that makes codeDAO.deleteExpired(any()) throw, asserting doJob still does not propagate, so both transactional calls in the try block are covered.

_rule: `backend-error-handling-always-catch-the-most-specific-exception-type-never-use-bare`_

apps/opik-backend/src/test/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJobTest.java
```diff
@@ -0,0 +1,79 @@
...
+        var codeNow = ArgumentCaptor.forClass(Instant.class);
+        var tokenThreshold = ArgumentCaptor.forClass(Instant.class);
+
+        job.doJob(null);
+
+        verify(codeDAO).deleteExpired(codeNow.capture());
+        verify(tokenDAO).deleteExpiredAndRevoked(tokenThreshold.capture());
+
+        // Revoked tokens must remain queryable for the grace window, so their cutoff trails "now" by exactly the grace.
+        assertThat(tokenThreshold.getValue()).isEqualTo(codeNow.getValue().minus(ROTATION_GRACE));
```
LifeXplorer: **[low] Logging branch (codes>0 || tokens>0) and the no-op path are untested**

doJob only logs the summary line when codes>0 || tokens>0. Neither test sets up a non-zero return from the DAO deletes (both mocks return the default 0), so the 'removed N codes, M tokens' INFO path and the silent no-op path are never exercised. For a data-deleting scheduled job, the count-propagation/log branch is observable behavior worth a focused assertion.

💡 Stub codeDAO.deleteExpired / tokenDAO.deleteExpiredAndRevoked to return positive counts in the happy-path test (or a dedicated case) so the count handling and the conditional log branch are covered.

_rule: `backend-testing-add-regression-tests-for-every-new-backend-behavior`_

LifeXplorer (commented):

11 findings — 1 high · 1 medium · 9 low. Suppressed by team conventions: see `suppressed.md`.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/McpOAuthBundle.java
LifeXplorer: fixed

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJob.java
```diff
@@ -0,0 +1,57 @@
...
+        }
```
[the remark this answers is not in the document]
LifeXplorer: Done in 90b473d — log placeholders are now single-quoted: `removed '{}' expired codes, '{}' expired/revoked tokens`, matching the backend structured-log convention.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/McpOAuthBundle.java
LifeXplorer: Done in 90b473d. Added `McpOAuthDisabledE2ETest#scrubJobNotScheduledWhenDisabled`, which asserts no `McpOAuthScrubJob` binding exists in the injector when the feature is off — `GuiceJobManager` schedules by scanning `injector.getAllBindings()` for `Job`-typed bindings, so absence of the binding *is* "not scheduled". Added a positive control `McpOAuthFlowE2ETest#scrubJobScheduledWhenEnabled` so the disabled-side assertion cannot pass vacuously.

Confirmed the load-bearing mechanism while writing it: `@On` makes the job a guicey extension via `JobGuiceyInstaller.matches(...)`, guicey binds recognised extensions, and `disableExtensions(McpOAuthScrubJob.class)` removes that binding — so `GuiceJobManager` never sees it.

apps/opik-backend/src/test/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJobTest.java
LifeXplorer: Done in 90b473d. `swallowsRuntimeException` is now a `@ParameterizedTest` over `{codes, tokens}`: each case makes one of the two transactions throw (`codeDAO.deleteExpired` / `tokenDAO.deleteExpiredAndRevoked`) and asserts `doJob` still does not propagate. Both transactional calls in the try block are now covered.

apps/opik-backend/src/test/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJobTest.java
LifeXplorer: Done in 90b473d. Added `logsSummaryWhenSomethingRemoved` (stubs positive counts, asserts the single INFO line and its formatted message) and `doesNotLogWhenNothingRemoved` (both DAOs return 0, asserts no INFO line), using a Logback `ListAppender` on the job logger. The conditional log branch and the no-op path are now both exercised.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthFilter.java
```diff
@@ -33,11 +41,22 @@ public void filter(ContainerRequestContext context) throws IOException {
...
-                    .build());
+                    .build();
+            String authHeader = context.getHeaderString(HttpHeaders.AUTHORIZATION);
+            // Gated on the config flag so disabling MCP OAuth is a real kill switch: previously
+            // minted opik_at_ bearer tokens stop authenticating the moment the AS is turned off.
+            if (opikConfig.getMcpOAuth().isEnabled() && McpOAuthTokens.isMcpOAuthToken(authHeader)) {
+                String token = authHeader.substring(BEARER_PREFIX.length()).trim();
+                ValidatedToken validatedToken = mcpOAuthService.validateAccessTokenForWorkspace(
+                        token, context.getHeaderString(RequestContext.WORKSPACE_HEADER));
+                authService.authorizeOAuth(validatedToken, contextInfo);
```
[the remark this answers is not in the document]
LifeXplorer: Not changing this — the fall-through to `authService.authenticate(...)` for non-MCP `Authorization` headers is intentional and required: normal Opik API-key auth runs on `/v1/private/*`, and the proposed "reject any header that is not `opik_mcp_at_`" would break it.

The scenario does not hold as a vulnerability:
- **Cloud:** `RemoteAuthService.authenticate` validates the bearer against comet-backend and rejects anything that is not a real key — `Bearer opik_at_not-real` gets a 401.
- **OSS:** the auth service is a no-op that accepts every request (including one with *no* `Authorization` header at all), by design. A fake bearer being accepted there is not a regression introduced by this PR.

Also note `opik_at_` is not a minted prefix — our access tokens are `opik_mcp_at_` (`McpOAuthTokenUtils.ACCESS_PREFIX`); the example token would simply not be an MCP token. The kill switch this PR adds (`MCP_OAUTH_ENABLED`) governs the MCP path itself.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/McpOAuthBundle.java
LifeXplorer: Already addressed — `McpOAuthBundleTest` verifies `disableExtensions(...)` with all six classes (incl. `OAuthValidateResource` and `McpOAuthScrubJob`) in both cases: `disablesResourcesWhenDisabled` (`verify(environment).disableExtensions(...all 6...)`) and `keepsResourcesWhenEnabled` (`verify(environment, never()).disableExtensions(...all 6...)`). The varargs array matches, so the test is in sync with the broadened production call. No change needed.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJob.java
LifeXplorer: Acknowledged — accurate, and intentional. An expired-but-never-revoked token lingering an extra `refreshRotationGrace` (30s default) is functionally harmless: it is already rejected as expired at validation time; the scrub is only reclaiming rows. Keeping the single shared threshold avoids a two-bind-param delete for no behavioural gain. The grace semantics for the revoked branch are documented on `doJob`.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJob.java
LifeXplorer: Acknowledged — minor. Leaving the generic error log: the two deletes are independent and a partial failure is rare, and the logged stack trace already pinpoints which transaction threw. Surfacing the codes-deleted count in the catch would force initializing `codes` for definite-assignment with little operational value.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthFilter.java
LifeXplorer: Acknowledged — intentional, documenting here. The MCP OAuth path resolves the requested workspace from `WORKSPACE_HEADER` only; query-param workspace pinning (`?workspace_name=`) is not part of the OAuth contract, and a blank header is the deliberate "no preference -> matches the token workspace" design. Not expanding the resolution logic.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/McpOAuthDisabledE2ETest.java
LifeXplorer: Leaving this status-only. In OSS no-op mode the meaningful assertion is the negative: a 200 proves the request fell through to regular auth rather than being routed into `validateAccessTokenForWorkspace` (which would 401/500 on a non-existent token). Asserting the project-list body would not strengthen the "token ignored, request not rejected" contract that this test exists to pin.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/McpOAuthFlowE2ETest.java
LifeXplorer: Leaving the auth-filter 401/403 paths status-only. These are framework-level rejections with a generic body; the security-relevant contract is the status code. Note the `/oauth/token` error paths in this same test already assert the structured `error` field via `assertOAuthError`, so the OAuth-protocol errors (where the body is the contract) are covered.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/McpOAuthFlowE2ETest.java
LifeXplorer: Acknowledged. There is no externally observable signal for server-side grace-window expiry to poll, so `until(() -> true)` is an intentional wall-clock wait, not a missed Awaitility condition. The `+500ms` buffer is explicit and the grace is configured down to 2s (`ROTATION_GRACE_MS`) to keep the suite short. Left as-is.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthValidateTokenResourceTest.java
LifeXplorer: Minor — leaving as-is. The `""` case does exercise a real guard path (`Strings.CI.startsWith("", BEARER_PREFIX)` -> treated as a non-Bearer header -> rejected), so the assertion is meaningful even though "blank" is loose wording. Not worth an isolated churn commit; will fold the relabel in if this file is touched again.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthValidateTokenResource.java
```diff
@@ -0,0 +1,61 @@
...
+@Slf4j
+@RequiredArgsConstructor(onConstructor_ = @Inject)
+@Tag(name = "MCP OAuth", description = "MCP OAuth 2.1 Authorization Server resources")
+public class OAuthValidateTokenResource {
+
+    private final @NonNull McpOAuthService mcpOAuthService;
+
+    @POST
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
```
andrescrz: Not a blocker: let's again default these at the resource (class) level, instead of at the endpoint (method) level.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthValidateTokenResource.java
```diff
@@ -0,0 +1,61 @@
...
+            @ApiResponse(responseCode = "200", description = "Validated token identity", content = @Content(schema = @Schema(implementation = ValidatedToken.class))),
+            @ApiResponse(responseCode = "401", description = "Missing, malformed, or unknown access token")})
+    public Response validate(@HeaderParam(HttpHeaders.AUTHORIZATION) String authHeader) {
+        if (!McpOAuthTokenUtils.isMcpOAuthToken(authHeader)) {
+            log.info("MCP OAuth validate rejected: no bearer access token presented");
+            throw new NotAuthorizedException(TOKEN_TYPE_BEARER);
+        }
+
+        String token = McpOAuthTokenUtils.extractBearerToken(authHeader);
+        String maskedToken = McpOAuthTokenUtils.maskToken(token);
```
andrescrz: Nit: masking the token seems more a like a service level concern, than a resource level one.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthValidateTokenResource.java
```diff
@@ -0,0 +1,61 @@
...
+        ValidatedToken validated = mcpOAuthService.validateAccessToken(token)
+                .orElseThrow(() -> {
+                    log.info("MCP OAuth validate rejected: token '{}' is not active", maskedToken);
```
andrescrz: Please remove the token from the logs, even if masked.

Applies to the success log below as well.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJob.java
```diff
@@ -0,0 +1,51 @@
...
+import java.time.Instant;
+
+import static com.comet.opik.infrastructure.db.TransactionTemplateAsync.WRITE;
+
+@Singleton
+@Slf4j
+@DisallowConcurrentExecution
+@On(value = "0 0/5 * * * ?", timeZone = "UTC") // every 5 minutes
+@RequiredArgsConstructor(onConstructor_ = @Inject)
+public class McpOAuthScrubJob extends Job {
```
andrescrz: All jobs should implement `InterruptableJob` properly.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJob.java
```diff
@@ -0,0 +1,51 @@
...
+import ru.vyarus.guicey.jdbi3.tx.TransactionTemplate;
+
+import java.time.Instant;
+
+import static com.comet.opik.infrastructure.db.TransactionTemplateAsync.WRITE;
+
+@Singleton
+@Slf4j
+@DisallowConcurrentExecution
+@On(value = "0 0/5 * * * ?", timeZone = "UTC") // every 5 minutes
```
andrescrz, line 27: We don't have full quartz set up, so in a multi-replica environment, this will run every 5 minutes per replica. You need to implement Redis based locking to prevent that from happening with `bestEffortLock` etc.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthTokenUtils.java
```diff
@@ -52,10 +53,21 @@ public static boolean isOAuthToken(String token) {
...
         }
-        return isAccessToken(authHeader.substring(BEARER_PREFIX.length()).trim());
+        return isAccessToken(extractBearerToken(authHeader));
+    }
+
+    /**
+     * Strips the {@code Bearer } scheme prefix from an {@code Authorization} header.
+     * @throws IllegalArgumentException if Bearer header missing
+     */
+    public static String extractBearerToken(@NonNull String authHeader) {
```
andrescrz: Double check the `NonNull` annotation here. It contradicts the `isMcpOAuthToken` and might not be aligned with the validation below.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthTokenUtils.java
```diff
@@ -52,10 +53,21 @@ public static boolean isOAuthToken(String token) {
...
+        if (!containsBearerPrefix(authHeader)) {
+            throw new IllegalArgumentException("Authorization header is not a Bearer token");
+        }
+        return authHeader.substring(BEARER_PREFIX.length()).trim();
```
andrescrz: Minor: for new code, better use `strip` instead of `trim` as it handles a wider charset.

apps/opik-backend/src/test/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJobTest.java
```diff
@@ -0,0 +1,163 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@DisplayName("McpOAuthScrubJob")
+@ExtendWith(DropwizardAppExtensionProvider.class)
+class McpOAuthScrubJobTest {
+
+    private static final Duration ROTATION_GRACE = Duration.ofMinutes(5);
+
+    private final TestContainersSetup setup = new TestContainersSetup();
```
andrescrz: Minor: it seems unused, please clean up if that's the case.

apps/opik-backend/src/test/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJobTest.java
```diff
@@ -0,0 +1,163 @@
...
+        setup.wireMock.server().stop();
+    }
+
+    @BeforeEach
+    void setUp() {
+        transactionTemplate.inTransaction(WRITE, handle -> {
+            handle.execute("DELETE FROM mcp_oauth_codes");
+            handle.execute("DELETE FROM mcp_oauth_tokens");
+            return null;
+        });
```
andrescrz: Let's keep the black box testing approach. We should avoid having tests that require bypassing the internal of the application.

apps/opik-backend/src/test/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJobTest.java
```diff
@@ -0,0 +1,163 @@
...
+        McpOAuthToken withinGrace = insertToken(farFuture, now.minus(ROTATION_GRACE.dividedBy(2)));
+        McpOAuthToken pastGrace = insertToken(farFuture, now.minus(ROTATION_GRACE.plusMinutes(1)));
+
+        job.doJob(null);
+
+        assertThat(tokenExists(withinGrace.tokenHash())).isTrue();
+        assertThat(tokenExists(pastGrace.tokenHash())).isFalse();
+    }
+
+    private McpOAuthCode insertCode(Instant expiresAt) {
```
andrescrz: Likewise, let's use the official endpoints (with clients) as we usually do in our tests.

Applies to all methods below.

andrescrz (commented):
Please address the important feedback before moving forward.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthValidateTokenResource.java
LifeXplorer: updated

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthValidateTokenResource.java
LifeXplorer: removed token from all logs

apps/opik-backend/src/main/java/com/comet/opik/api/resources/oauth/OAuthValidateTokenResource.java
LifeXplorer: removed token from all logs

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthTokenUtils.java
LifeXplorer: Removed `@NonNull` from `extractBearerToken`

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthTokenUtils.java
LifeXplorer: updated

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJob.java
LifeXplorer: updated

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJob.java
LifeXplorer: updated

apps/opik-backend/src/test/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJobTest.java
LifeXplorer: removed

apps/opik-backend/src/test/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJobTest.java
LifeXplorer: removed

apps/opik-backend/src/test/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJobTest.java
LifeXplorer: Added scrubsExpiredArtifactsFromOAuthFlow that runs the scrub job E2E: an authorization code and token pair are minted through the public OAuth endpoints, expire, and are then removed by the wired job

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJob.java
```diff
@@ -0,0 +1,83 @@
...
+
+        // We don't run a full Quartz cluster, so each replica fires this schedule independently.
+        // A best-effort Redis lock held until expiry ensures a single replica scrubs per cycle.
+        lockService.bestEffortLock(
+                JOB_LOCK,
+                Mono.fromRunnable(this::scrub),
+                Mono.fromRunnable(
+                        () -> log.debug("Could not acquire lock for MCP OAuth scrub, another instance is running")),
+                Duration.ofMinutes(1),
+                Duration.ZERO,
```
andrescrz: Please make sure both values are configurable. In addition, you like want to tweak both defaults:
1. lock timeout: I'd get it closer to the job period (5 minutes) as the flag below is `true`. In a replicated environment, like the majority of production installations, nothing prevents other replicas of the job starting earlier than 5 minutes. This will ensure you get closer to 5 minutes periocity.
2. lockWaitTime: you likely want a brief period of time to wait for the lock (typicall 100ms).

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJob.java
```diff
@@ -0,0 +1,83 @@
...
+                true)
+                .subscribe(
```
andrescrz, line 65: You're missing using an executor (likely the default bounded elastic should be fine). There's IO in the MySQL DB, so you should handle that.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJob.java
```diff
@@ -0,0 +1,83 @@
...
+                        __ -> {
+                        },
+                        error -> log.error("MCP OAuth scrub failed", error));
+    }
+
+    @Override
+    public void interrupt() {
```
andrescrz, line 73: This won't really interrupt the reactor chain. Please see other jobs such as `ExperimentProjectMigrationJob` to check how to properly implement that.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJob.java
```diff
@@ -0,0 +1,83 @@
...
+        interrupted.set(true);
+        log.info("MCP OAuth scrub job successfully called interruption");
+    }
+
+    private void scrub() {
```
andrescrz: Minor: you could add a service to better divide the responsibilities of this job: the job handles the periodicity etc. the service the business logic implementation. That would facilitate your testing as well as async jobs sometimes are harder to test, that way you can have an intense black box test at the service layer and then a smoke test for the happy path of the job.

Anyway, this is a simple job, so no need to tackle this, unless your testing is too flaky or if it has too many gaps.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthTokenUtils.java
```diff
@@ -52,29 +51,25 @@ public static boolean isOAuthToken(String token) {
...
-            return "";
+    public static String extractBearerToken(String authHeader) {
+        if (!containsBearerPrefix(authHeader)) {
+            throw new IllegalArgumentException("Authorization header is not a Bearer token");
         }
-        return "sha256:" + hash(token).substring(0, 12);
+        return authHeader.substring(BEARER_PREFIX.length()).strip();
+    }
+
+    public static String hash(String token) {
```
andrescrz: Minor: public method, not sure how it handles null token, you could validate if NPE.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthValidateTokenResourceTest.java
```diff
@@ -0,0 +1,113 @@
...
+        try (Response response = validate(authHeader)) {
+            assertThat(response.getStatus()).isEqualTo(Response.Status.UNAUTHORIZED.getStatusCode());
+        }
+
+        verify(mcpOAuthService, never()).validateAccessToken(any());
+    }
+
+    @Test
+    @DisplayName("rejects a missing Authorization header")
+    void rejectsMissingHeader() {
```
andrescrz: Minor: this test seems like a param with `null` value of the previous one.

apps/opik-backend/src/test/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJobTest.java
```diff
@@ -0,0 +1,352 @@
...
+
+        transactionTemplate.inTransaction(WRITE,
+                handle -> handle.attach(McpOAuthTokenDAO.class).deleteExpiredAndRevoked(threshold));
+
+        assertThat(tokenExists(withinGrace.tokenHash())).isTrue();
+        assertThat(tokenExists(pastGrace.tokenHash())).isFalse();
+        assertThat(tokenExists(valid.tokenHash())).isTrue();
+    }
+
+    private McpOAuthCode insertCode(Instant expiresAt) {
```
andrescrz: Please double check all direct access through the database in all the methods below. We should generally implement black box tests by using the real API endpoints, instead of bypassing DB access.

This direct access. should only be done for very particular exceptions.

apps/opik-backend/src/test/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJobTest.java
```diff
@@ -0,0 +1,352 @@
...
+    private static String randomHash() {
+        return RandomStringUtils.secure().nextAlphanumeric(64);
+    }
+
+    /**
+     * Drives the MCP OAuth authorization-code + PKCE flow through the public endpoints so the test mints real codes and
+     * tokens without reaching into the persistence layer. Targets an app running with local auth, where consent
+     * resolves to the default workspace.
+     */
+    private record OAuthClient(ClientSupport client, String baseURI) {
```
andrescrz: Better create a separated client class for this in the tests, see other examples. This is general in the code base, such as `SpanResourceClient` etc.

andrescrz (commented):
Most of previous comments were addressed.

There are still some issues in the job that should be addressed before moving forward. Please take a look.

In addition, added some not blocker but important comments on the tests: let's avoid bypassing the DB. Also, generally recommended to extract to a client class, as the usual convention.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJob.java
LifeXplorer: set lock timeout = 5min and lockWaitTime=100ms

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJob.java
LifeXplorer: added `Schedulers.boundedElastic()` executor

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJob.java
LifeXplorer: added `AtomicReference<Disposable>` same as in `ExperimentProjectMigrationJob`

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJob.java
LifeXplorer: extracted into `McpOAuthScrubService`, then reshaped the tests (intense black-box service test + job happy-path smoke test). Based on `ExperimentProjectMigrationJob` design pattern

apps/opik-backend/src/test/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJobTest.java
LifeXplorer: What switched to API vs. stayed on DB:
- Switched to API: all artifact creation (client registration, authorize/consent + PKCE, code exchange) - all are now entirely endpoint-driven via test OAuthResourceClient.
- Remained direct DB (justified exceptions): deletion assertions, expiry polling, and the DAO edge-case seeding (because deletion/grace-window state has no API signal).
For direct DB assertion methods added justifying javadoc

apps/opik-backend/src/test/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJobTest.java
LifeXplorer: moved to a separate `OAuthResourceClient` test class

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthTokenUtils.java
LifeXplorer: annotated with Lombok `@NonNull`

apps/opik-backend/src/test/java/com/comet/opik/api/resources/oauth/OAuthValidateTokenResourceTest.java
LifeXplorer: added param to prev. test. Removed that one

andrescrz (approved):
LGTM, thanks for putting this together.

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubJob.java
```diff
@@ -0,0 +1,82 @@
...
+            return;
+        }
+
+        // We don't run a full Quartz cluster, so each replica fires this schedule independently.
+        // A best-effort Redis lock held until expiry ensures a single replica scrubs per cycle.
+        var subscription = lockService.bestEffortLock(
+                JOB_LOCK,
+                Mono.defer(() -> {
+                    if (interrupted.get()) {
+                        log.info("MCP OAuth scrub was interrupted before processing, skipping");
```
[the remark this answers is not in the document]
LifeXplorer: refactoring is out of the scope of this pr

apps/opik-backend/src/main/java/com/comet/opik/domain/mcpoauth/McpOAuthScrubService.java
```diff
@@ -0,0 +1,40 @@
...
+     * request (RFC 6749 §6 token rotation grace) returns invalid_grant instead of triggering reuse detection.
+     */
+    public Mono<Void> scrub() {
+        return Mono.fromRunnable(() -> {
+            Instant now = Instant.now();
+            Instant tokenThreshold = now.minus(opikConfig.getMcpOAuth().getRefreshRotationGrace());
+
+            int codes = template.inTransaction(WRITE,
+                    handle -> handle.attach(McpOAuthCodeDAO.class).deleteExpired(now));
+            int tokens = template.inTransaction(WRITE,
```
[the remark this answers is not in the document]
LifeXplorer: The observation is correct that `threshold = now − refreshRotationGrace` is also applied to the `expires_at` branch, but this is not a functional bug:

1. **Expired tokens are already rejected regardless of the row's existence.** Validation never trusts row presence — both `isActiveAccessToken` and `isValidRefreshToken` require `expiresAt().isAfter(now)`. The instant `expires_at < now` the token is dead; the extra ≤grace window of physical row existence grants no access.

2. **The grace branch is never exercised by expiry.** At rotation a refresh token's `expires_at` is in the future (it had to pass `isValidRefreshToken` to be rotated, and the rotated pair inherits `row.expiresAt()`), so during the grace window `expires_at < threshold` is false — retention there is driven entirely by `revoked_at`. The shared threshold on `expires_at` only ever affects genuinely-expired tokens, which have no grace semantics.

3. **The shared threshold errs on the safe side.** `expires_at < now − grace` deletes *later* than `expires_at < now`, never earlier, so it can only over-retain, never under-retain — it cannot cause premature deletion or a reuse-detection failure. With the 5-minute eventually-consistent cleanup cycle, a 30s difference in physical deletion is immaterial.

So this is at most a minor precision nit, and the current single-threshold behavior is the conservative direction. Leaving as-is.

PR #7008, opened 2026-06-09
[OPIK-6833][BE] Support OpenAI Responses API for the OpenAI provider (Playground & LLM-as-a-Judge)

yaricom:
## Details

Opik previously only supported OpenAI's `/v1/chat/completions` endpoint. A growing class of models — including OpenAI's reasoning models (o1, o3, o3-mini, o4-mini) and other specialized deployments — are only available through the newer `/v1/responses` API. Sending requests to those models via the chat completions path fails with:

> _This is not a chat model and thus not supported in the v1/chat/completions endpoint._

This PR adds a `pipeline_mode` configuration option to the OpenAI provider that lets users opt into the Responses API path:

| `pipeline_mode` value | Behaviour |
|---|---|
| `CHAT_COMPLETIONS_API` _(default)_ | Existing behaviour — routes to `/v1/chat/completions` |
| `RESPONSES_API` | New — routes to `/v1/responses` via the official OpenAI SDK wrapper |

**How it works**

- A new `LlmProviderOpenAiResponses` service wraps `OpenAiOfficialResponsesChatModel` from the `langchain4j-open-ai-official` library.
- A new `LlmProviderOpenAiResponsesMapper` translates the existing Chat Completions wire format (in/out) to langchain4j's provider-agnostic `ChatRequest`/`ChatResponse`, preserving sampling parameters, token usage, and finish reasons.
- `OpenAIClientGenerator.extractApiPipelineMode()` reads the `openai_pipeline_mode` key from the provider's configuration map at request time. Unknown values fall back to `CHAT_COMPLETIONS_API` with a warning log.
- Both the Playground and LLM-as-a-Judge flows are covered — the routing happens at the `LlmServiceProvider` level.

## Summary

Adds OpenAI Responses API support to the backend's LLM proxy and judge paths, routed by a new `openai_pipeline_mode` provider-key configuration, with frontend support for selecting it in the v2 Manage-AI-Provider dialog. The Chat Completions path remains untouched (lossless pass-through via langchain4j's `OpenAiClient`); the new `RESPONSES_API` path delivers blocking generate, streaming, and full tool-calling round-trip through `OpenAiOfficialResponsesChatModel`/`StreamingChatModel` with a Chat-Completions ⇄ langchain4j-`ChatRequest`/`ChatResponse` translator.

### Backend

- New Maven dependency `langchain4j-open-ai-official` to gain access to the Responses-API-backed `ChatModel`/`StreamingChatModel` and the openai-java SDK transitively.
- New enum `OpenAIClientGenerator.ApiPipelineMode` (`CHAT_COMPLETIONS_API` / `RESPONSES_API`) read from `LlmProviderClientApiConfig.configuration.openai_pipeline_mode` (namespaced within the shared configuration map). Unknown values log a warning and fall back to `CHAT_COMPLETIONS_API`.
- `OpenAIClientGenerator.newOpenAiChatLanguageModel` (judge path) now branches on `pipeline_mode` to return either the existing `OpenAiChatModel` or a new `OpenAiOfficialResponsesChatModel`.
- New factories `newResponsesApiChatModel(config)` / `newResponsesApiStreamingChatModel(config)` for the proxy path. Both use a placeholder `modelName` at build time (real model supplied per-request via `ChatRequest.parameters().modelName(...)`) to satisfy langchain4j's required-field validation.
- New `LlmProviderOpenAiResponses` implementing `LlmProviderService`. Constructor takes the `OpenAIClientGenerator` + `LlmProviderClientApiConfig` rather than pre-built models; each `generate`/`generateStream` call peeks at `response_format.json_schema.strict` on the inbound request and builds the matching strict/loose `ChatModel` (or `StreamingChatModel`) for that single call. Net cost is unchanged from earlier drafts — `getService` was already per-request, so the model build site just moved one frame deeper. Streaming handler synthesizes OpenAI Chat-Completions SSE chunks (`delta.role`/`delta.content` partials, empty-delta final chunk with `finish_reason` + inlined usage).
- New `LlmProviderOpenAiResponsesMapper` (Lombok `@UtilityClass`) translating Chat-Completions DTOs ⇄ langchain4j `ChatRequest`/`ChatResponse`. Scope: text-only system/user/assistant messages, sampling params, stop sequences, `max_completion_tokens` (preferred) / `max_tokens`, token usage, finish-reason enum→OpenAI strings, streaming chunk synthesis, full tool calling (tool specs with JSON-Schema parameters via `JsonSchemaElementJsonUtils.fromMap`, `tool_choice` string→`ToolChoice` enum, `ToolMessage`→`ToolExecutionResultMessage` for resume, assistant `tool_calls`→`AiMessage.toolExecutionRequests` for resume, and the reverse on the response side), structured response formats (`response_format=json_object` → loose JSON mode; `response_format=json_schema` → langchain4j `JsonSchema` with name + root `JsonObjectSchema`; uses a Jackson roundtrip to read the opaque openai-internal `JsonSchema`), and a static `extractRequestedStrictJsonSchema(ChatCompletionRequest)` helper consumed by the provider to pick the right model variant.
- Named-function `tool_choice` (`{"type":"function","function":{"name":"foo"}}`) is rejected with `BadRequestException` rather than silently degrading to `AUTO` — langchain4j's `ToolChoice` enum has no equivalent, so a silent fallback would lose the caller's intent.
- `OpenAIClientGenerator` factories `newResponsesApiChatModel(config, strictJsonSchema)` and `newResponsesApiStreamingChatModel(config, strictJsonSchema)` thread the strict flag into the langchain4j builder via `.strictJsonSchema(...)`. Strict mode is build-time on langchain4j; per-request honoring is achieved by building the matching variant per call.
- `OpenAILlmServiceProvider.getService(...)` branches on `pipeline_mode` to return either `LlmProviderOpenAi` (chat completions) or `LlmProviderOpenAiResponses` (responses, constructed with generator + config).
- Logging follows the repo convention from `AGENTS.md`: every substituted value in `log.warn`/`log.info` calls is wrapped in single quotes (`'{}'`).
- `MessageContentNormalizer` rewrites multimodal user turns into `OpikUserMessage` (a sibling of openai-internal `UserMessage` that `implements Message` directly). The mapper now handles it in a dedicated `case` so the proxy doesn't 500 when the playground sends structured content — content is flattened to text via the same path the regular `UserMessage` uses.
- Responses-API mapper unconditionally strips `frequency_penalty`, `presence_penalty`, and `stop` before delegating to langchain4j. The Responses-API model's client-side `validate(...)` rejects these (`UnsupportedFeatureException`), and clients (notably the playground) send them as framework defaults of `0` / empty list even when the user didn't tune them. Silent strip is documented in the mapper Javadoc; users who explicitly need them should fall back to `pipeline_mode=CHAT_COMPLETIONS_API`.
- Out of scope (deferred): multimodal user content beyond text flattening (images/audio), per-token streaming of tool-call argument deltas, reasoning-content streaming. OpenRouter and FreeModel providers continue to use Chat Completions and require no changes.

### Backend: OpenAI error mapping (CC + Responses)

- `OpenAiCompatStatusCodes` map gains `unsupported_parameter` → 400 (the specific code OpenAI returns for per-model parameter rejections like "`'top_p' is not supported with this model`").
- `OpenAiErrorMessage.toErrorMessage()` now falls back to `error.type` when `error.code` isn't in the map — covers brand-new specific codes by routing on the recognised high-level type (`invalid_request_error`, `rate_limit_error`, …) and keeps the HTTP family correct.
- `LlmProviderOpenAiResponses.getLlmProviderError(...)` unwraps openai-java's `com.openai.errors.OpenAIServiceException` directly (its `getMessage()` is `"<status>: <description>"` — no JSON body — so the shared JSON-scan helper in `LlmProviderLangChainMapper` skipped it and the failure bubbled up as a generic 500). Walks the causal chain, extracts `statusCode()`/`code()`/`type()` from the typed exception, and prefers `code` → `type` → `statusCode()` for the HTTP status. Falls back to the shared mapper when no openai-java exception is in the chain. Verified end-to-end: `gpt-5.5` + invalid parameter now surfaces as `ClientErrorException` 400 with the original OpenAI diagnostic, not as `InternalServerErrorException` 500.

#### Test coverage added

- `LlmProviderOpenAiResponsesMapperTest` — 31 tests across nested classes:
  - `ToChatRequest` (6): model/sampling propagation, system/user/assistant message translation, drops unsupported Responses-API sampling params (frequency_penalty/presence_penalty/stop), `max_completion_tokens` precedence over `max_tokens`, `OpikUserMessage` falls through the same flattening path as the openai-internal `UserMessage`.
  - `ToChatCompletionResponse` (8 invocations): happy-path metadata + usage, model-name fallback, null-usage handling, finish-reason enum→string mapping (5 parameterized cases).
  - `StreamingChunks` (3): partial chunk shape (assistant role + content), final chunk shape (empty delta + finish_reason + usage), model fallback.
  - `ToolCalling` (8 invocations): tool-spec propagation with JSON Schema parameters, `tool_choice` `auto`/`required`/`none` mapping (3 parameterized), named-function `tool_choice` rejection, tool-result message resume translation, prior-assistant-with-tool-calls resume translation, assistant `tool_calls` emission with `finish_reason="tool_calls"`.
  - `StructuredOutput` (5): `json_object` → loose JSON mode, `json_schema` with name + root schema (properties + required) propagated, no `response_format` leaves langchain4j default, explicit `text` also stays at the default, `extractRequestedStrictJsonSchema` returns `true` only for explicit `strict: true` and `false` for the three negative cases (strict:false, json_object, no response_format).
- `LlmProviderOpenAiResponsesTest` — 9 integration tests using stub `ChatModel`/`StreamingChatModel` + a mocked `OpenAIClientGenerator`:
  - Streaming happy path (3 partials → final chunk → close, no errors).
  - Streaming error routing (`onError` → `handleError`, no chunks, no close).
  - End-to-end tool loop round-trip (turn 1: request with tools → response with tool_calls; turn 2: client resumes with assistant(tool_calls) + tool result → final text response).
  - `strict=true` request → `ArgumentCaptor` confirms `true` reached `newResponsesApiChatModel(...)`.
  - Request without `response_format` → `false` reached the factory.
  - Openai-java `BadRequestException` with a specific `code` mapped by `OpenAiCompatStatusCodes` → clean 400 `ErrorMessage`.
  - Openai-java `BadRequestException` with unknown `code` but recognised `type` → falls back to the type, still 400.
  - Openai-java `BadRequestException` with neither in the map → falls back to the SDK exception's `statusCode()` (400).
  - Nested openai-java exception wrapped in `RuntimeException("retry wrapper", inner)` → causal chain is walked correctly.
- `OpenAiErrorMessageTest` — 4 new cases added (now 13 invocations total): `unsupported_parameter` → 400 mapping, plus 2 parameterised `typeFallbackMappings` cases proving that an unknown `code` paired with a recognised `type` still yields the right HTTP family.

### Frontend

- `ProviderKeyConfiguration` (`src/types/providers.ts`) extended with optional `openai_pipeline_mode?: "chat_completions_api" | "responses_api"`.
- `ManageAIProviderDialog` cloud schema (v2) gains an optional `openaiPipelineMode` enum plus `OPENAI_PIPELINE_MODE_VALUES` / `OpenAiPipelineMode` exports.
- `CloudAIProviderDetails` (v2) conditionally renders a Pipeline-mode `Select` (Chat Completions API / Responses API) under the API-key field, only when the chosen provider is OpenAI. Includes a one-line hint about what Responses API unlocks.
- `ManageAIProviderDialog` (v2) wires the field through all four lifecycle points: form defaults, `resetSelectionState`, `handleProviderSelect`, and `cloudConfigHandler`. On save, the explicit value is **always** sent for OpenAI keys (so switching from `responses_api` back to `chat_completions_api` actually persists rather than leaving the stored value untouched).
- `DEFAULT_OPENAI_PIPELINE_MODE` constant (typed as `OpenAiPipelineMode`, declared in `schema.ts`) is the single source of truth for the default fallback. Every default/fallback site (form reset, save payload, Select control's `value ?? …`) references it instead of repeating the string literal, so changing the default needs editing only one line.
- `normalizeOpenAiPipelineMode(value)` helper in `schema.ts` is used wherever the form hydrates from a backend-stored string (`useForm` defaults, `handleProviderSelect`). The backend's `extractApiPipelineMode` accepts any casing (uppercases before `valueOf`), so the persisted value could be lowercase, uppercase, or anything unknown. The helper lowercases, validates against `OPENAI_PIPELINE_MODE_VALUES`, and falls back to the default — keeps the Select on a valid option and prevents Zod from blocking submit on legacy or odd-cased values.
- v1 dialog intentionally not updated (drafted in v2 only).

#### Frontend: Top P gating for OpenAI reasoning models

OpenAI rejects `top_p` outright on reasoning models (`gpt-5.x`, `o1*`, `o3*`, `o4*`) with `400 Unsupported parameter: 'top_p' is not supported with this model.` regardless of the value (`0`, `1`, anything). Three coordinated touchpoints — symmetrical to the existing temperature pinning, scoped only to `top_p` per empirical evidence (the other reasoning-model parameter restrictions don't appear in production logs on the Chat Completions pipeline):

- `OpenAIModelConfigs.tsx`: hides the `Top P` slider entirely when `isReasoningModel(model)`. Mirrors the temperature pinning's "the user can't set the slider to an invalid value" guarantee.
- `modelUtils.updateProviderConfig`: when the user switches a persisted prompt into a reasoning OpenAI model, stale `topP` is dropped from the config (same callsite that already coerces `temperature < 1` to `1.0`).
- `modelUtils.sanitizeConfigForRequest`: last-mile guard — strips `topP` from the wire payload if it survives the slider gating and the reconciler (e.g., very old persisted prompts whose state didn't go through `updateProviderConfig`).
- Tests: 2 new `updateProviderConfig` cases (switch into reasoning drops topP / switch to non-reasoning keeps it); 2 new `sanitizeConfigForRequest` cases (strips for `gpt-5.5` / keeps for `gpt-4o`); 1 existing test updated to reflect that `topP=1` is still stripped on reasoning models because the constraint is the parameter's presence, not its value. `vitest` 43/43 pass.

## Files Changed

```
 apps/opik-backend/pom.xml                          |   6 +
 .../llm/OpenAiCompatStatusCodes.java               |   4 +
 .../llm/openai/LlmProviderOpenAiResponses.java     | 115 ++++
 .../openai/LlmProviderOpenAiResponsesMapper.java   | 399 ++++++++++++++
 .../llm/openai/OpenAIClientGenerator.java          | 138 ++++-
 .../llm/openai/OpenAILlmServiceProvider.java       |   7 +-
 .../llm/openai/OpenAiErrorMessage.java             |  10 +-
 .../LlmProviderOpenAiResponsesMapperTest.java      | 610 +++++++++++++++++++++
 .../llm/openai/LlmProviderOpenAiResponsesTest.java | 455 +++++++++++++++
 .../llm/openai/OpenAiErrorMessageTest.java         |  23 +
 apps/opik-frontend/src/lib/modelUtils.test.ts      |  60 +-
 apps/opik-frontend/src/lib/modelUtils.ts           |  20 +
 apps/opik-frontend/src/types/providers.ts          |   3 +
 .../CloudAIProviderDetails.tsx                     |  53 +-
 .../ManageAIProviderDialog.tsx                     |  24 +-
 .../llm/ManageAIProviderDialog/schema.ts           |  34 ++
 .../providerConfigs/OpenAIModelConfigs.tsx         |   5 +-
 17 files changed, 1952 insertions(+), 14 deletions(-)
```

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-6833

## AI-WATERMARK

AI-WATERMARK: no

## Testing

Unit tests cover `LlmProviderOpenAiResponsesMapper`:
- Model name and all sampling parameters are correctly mapped
- System, user, and assistant message roles translate to the right langchain4j types
- `maxCompletionTokens` takes priority over the deprecated `maxTokens`
- Stop sequences are forwarded
- All `FinishReason` enum values produce the correct OpenAI string
- Null token usage is handled gracefully
- Unsupported message roles (e.g. tool messages) throw `BadRequestException`

`OpenAIClientGenerator.extractApiPipelineMode()` is tested for valid values, blank/null inputs, and unknown values that should fall back silently.

## Documentation

NA


apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/OpenAIClientGenerator.java
```diff
@@ -52,6 +64,41 @@ public OpenAiClient newOpenAiClient(@NonNull LlmProviderClientApiConfig config)
 
     public ChatModel newOpenAiChatLanguageModel(@NonNull LlmProviderClientApiConfig config,
             @NonNull LlmAsJudgeModelParameters modelParameters) {
+        return switch (extractApiPipelineMode(config)) {
+            case CHAT_COMPLETIONS_API -> newCompletionsApiChatModel(config, modelParameters);
+            case RESPONSES_API -> newResponsesApiChatModel(config, modelParameters);
+        };
+    }
```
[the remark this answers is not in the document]
yaricom: No, we shouldn't. The responses API can be implemented on custom baseUrl as well in the future or already was.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/OpenAIClientGenerator.java
```diff
@@ -52,6 +64,41 @@ public OpenAiClient newOpenAiClient(@NonNull LlmProviderClientApiConfig config)
...
+            @NonNull LlmAsJudgeModelParameters modelParameters) {
+        return newOpenAiChatLanguageModel(config, modelParameters);
+    }
+
+    ApiPipelineMode extractApiPipelineMode(@NonNull LlmProviderClientApiConfig config) {
+        String pipelineMode = Optional.ofNullable(config.configuration())
+                .orElse(Map.of())
+                .get(CONFIG_KEY_PIPELINE_MODE);
+        if (StringUtils.isBlank(pipelineMode)) {
+            return ApiPipelineMode.CHAT_COMPLETIONS_API;
```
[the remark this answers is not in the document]
yaricom: It is relevant only OpenAI provider configuration and a few compatible ones in the future. LlmProviderClientApiConfig is the general config for any provider. So, making it common field is not a good idea

CometActions:

### Access Information
- **URL:** https://pr-7008.dev.comet.com
- **Namespace:** pr-7008
- **Version:** 2.0.60-7008-merge-2254
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7008%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7008`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/OpenAIClientGenerator.java
```diff
@@ -52,6 +64,41 @@ public OpenAiClient newOpenAiClient(@NonNull LlmProviderClientApiConfig config)
...
+                    CONFIG_KEY_PIPELINE_MODE, pipelineMode, ApiPipelineMode.CHAT_COMPLETIONS_API);
+            return ApiPipelineMode.CHAT_COMPLETIONS_API;
+        }
+    }
+
+    ChatModel newCompletionsApiChatModel(@NonNull LlmProviderClientApiConfig config,
+            @NonNull LlmAsJudgeModelParameters modelParameters) {
         var builder = OpenAiChatModel.builder()
                 .modelName(modelParameters.name())
                 .apiKey(config.apiKey())
```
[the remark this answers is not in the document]
yaricom: not worth it and builders of different types

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/OpenAIClientGenerator.java
```diff
@@ -80,14 +127,30 @@ public ChatModel newOpenAiChatLanguageModel(@NonNull LlmProviderClientApiConfig
...
-    @Override
-    public ChatModel generateChat(@NonNull LlmProviderClientApiConfig config,
+    ChatModel newResponsesApiChatModel(@NonNull LlmProviderClientApiConfig config,
             @NonNull LlmAsJudgeModelParameters modelParameters) {
-        return newOpenAiChatLanguageModel(config, modelParameters);
+        var builder = OpenAiOfficialResponsesChatModel.builder()
+                .modelName(modelParameters.name())
+                .apiKey(config.apiKey());
+
+        Optional.ofNullable(llmProviderClientConfig.getConnectTimeout())
```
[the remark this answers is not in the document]
yaricom: not worth it

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/LlmProviderOpenAiResponsesMapper.java
```diff
@@ -0,0 +1,292 @@
...
+                case "none" -> ToolChoice.NONE;
+                default -> {
+                    log.warn("Unknown OpenAI tool_choice string '{}', falling back to AUTO", s);
+                    yield ToolChoice.AUTO;
+                }
+            };
+        }
+        log.info("Named-function tool_choice not supported by langchain4j ToolChoice enum; "
+                + "falling back to AUTO. tool_choice={}", toolChoice);
+        return ToolChoice.AUTO;
```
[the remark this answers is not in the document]
yaricom: fixed

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/LlmProviderOpenAiResponsesMapper.java
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    /**
+     * Maps OpenAI's {@code response_format} to langchain4j's {@link ResponseFormat}. {@code text}
+     * returns {@code null} (langchain4j's default). {@code json_object} → loose JSON mode.
+     * {@code json_schema} → strict structured-output mode with name + root schema. The {@code strict}
+     * flag from the request is NOT propagated — langchain4j's per-request {@link JsonSchema} has no
+     * strict slot; strict mode is configured at {@code OpenAiOfficialResponsesChatModel} build time
+     * via {@code .strictJsonSchema(...)}. Returns {@code null} when {@code responseFormat} is absent.
```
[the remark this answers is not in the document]
yaricom: fixed

apps/opik-frontend/src/v2/pages-shared/llm/ManageAIProviderDialog/ManageAIProviderDialog.tsx
```diff
@@ -182,6 +185,7 @@ const ManageAIProviderDialog: React.FC<ManageAIProviderDialogProps> = ({
       queryParams: [],
       authHeaderName: "",
       suppressDefaultAuth: false,
+      openaiPipelineMode: "chat_completions_api",
     });
     setStep("select");
```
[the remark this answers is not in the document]
yaricom: refactored

apps/opik-frontend/src/v2/pages-shared/llm/ManageAIProviderDialog/ManageAIProviderDialog.tsx
```diff
@@ -235,6 +239,11 @@ const ManageAIProviderDialog: React.FC<ManageAIProviderDialogProps> = ({
         "suppressDefaultAuth",
         providerData?.configuration?.suppress_default_auth === "true",
       );
+      form.setValue(
+        "openaiPipelineMode",
+        providerData?.configuration?.openai_pipeline_mode ??
+          "chat_completions_api",
```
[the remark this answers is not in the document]
yaricom: refactored

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/LlmProviderOpenAiResponsesMapper.java
```diff
@@ -0,0 +1,399 @@
...
+        Optional.ofNullable(resolveMaxOutputTokens(request)).ifPresent(builder::maxOutputTokens);
+
+        // frequency_penalty, presence_penalty, stop, and top_k are NOT supported by the OpenAI
+        // Responses API — langchain4j's OpenAiOfficialResponsesChatModel.validate() throws
+        // UnsupportedFeatureException for any of them. Clients (notably the playground) often
+        // send frequency_penalty: 0 / presence_penalty: 0 as framework defaults even when the
+        // user didn't tune them, so the request would fail before reaching OpenAI. We silently
+        // drop these fields rather than 500 the call; a user who explicitly tunes them and
+        // expects them to apply should use pipeline_mode=CHAT_COMPLETIONS_API.
+        warnIfDroppedSamplingParam(request);
```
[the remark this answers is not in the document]
yaricom: No, we should not fail fast because we drop parameters that is not supported by upstream so user provided values for these parameters is not relevant at all and will only result in errors

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/LlmProviderOpenAiResponsesMapper.java
```diff
@@ -0,0 +1,399 @@
...
+        return switch (message) {
+            case dev.langchain4j.model.openai.internal.chat.SystemMessage sys -> SystemMessage.from(sys.content());
+            case dev.langchain4j.model.openai.internal.chat.UserMessage user -> UserMessage.from(
+                    MessageContentNormalizer.flattenContent(user.content()));
+            // OpikUserMessage is a sibling of openai-internal UserMessage (it implements Message
+            // directly to add video-URL support); MessageContentNormalizer rewrites user turns into
+            // this type whenever the inbound content is multimodal. Flatten to text — full
+            // multimodal propagation is deferred per the class Javadoc.
+            case com.comet.opik.domain.llm.langchain4j.OpikUserMessage opikUser -> UserMessage.from(
+                    MessageContentNormalizer.flattenContent(opikUser.content()));
```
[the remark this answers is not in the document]
yaricom: It is as expected. Support for other modalities is subject of the follow up PR

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7008`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

apps/opik-frontend/src/lib/modelUtils.ts
```diff
@@ -186,6 +186,14 @@ export const updateProviderConfig = <
       changed = true;
     }
 
+    // Reasoning models reject top_p outright (OpenAI returns 400 "Unsupported parameter:
+    // 'top_p' is not supported with this model."). Drop any stale value so the next request
+    // omits the field entirely. The Top P slider is hidden for these models in the UI.
+    if (isReasoningModel(params.model) && next.topP !== undefined) {
+      next.topP = undefined;
+      changed = true;
+    }
```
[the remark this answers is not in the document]
yaricom: 1. The factual claim about providerModels.ts is **wrong**. The cited IDs do exist in that file (lines 1483, 1547, 1555), but they are plain {value, label} pairs — there is no supports_reasoning: true field on any entry. Confirmed via grep "supports_reasoning" providerModels.ts → zero hits. supports_reasoning lives in data/model_prices_and_context_window.json (LiteLLM data), not in providerModels.ts.
  
  2. The cited IDs aren't routed through the OPEN_AI branch. openai/gpt-5.5-pro, openai/o3-pro, openai/o4-mini-deep-research, openai/o1-pro, and openai/o3-deep-research are registered under PROVIDER_TYPE.OPEN_ROUTER (the OpenRouter list, line 143 onward), not PROVIDER_TYPE.OPEN_AI (which ends at line 106). getProviderFromModel matches them as
  OPEN_ROUTER, so:
  - isReasoningModel skips the OPENAI_MODEL_CAPABILITIES lookup entirely and uses the fetched-flag / REASONING_MODELS path.
  - Both updateProviderConfig and sanitizeConfigForRequest gate the topP-strip logic on providerType === OPEN_AI, so the OpenRouter path doesn't strip topP regardless. That's intentional — OpenRouter, not OpenAI, fields the request.
  
  3. The OpenAI-direct dropdown is fully covered. Every reasoning model exposed under PROVIDER_TYPE.OPEN_AI (lines 25–106: GPT_5, GPT_5_MINI, GPT_5_NANO, GPT_5_1, GPT_5_2, GPT_5_4*, GPT_5_5, GPT_O1, GPT_O3, GPT_O3_MINI, GPT_O4_MINI) is present in OPENAI_MODEL_CAPABILITIES with reasoning: true. No *_PRO or *_DEEP_RESEARCH variants are exposed in the OpenAI direct list — those enum values exist but aren't referenced anywhere (grep PROVIDER_MODEL_TYPE.GPT_O3_PRO → zero hits).

  4. The "shared helper" refactor is also redundant. updateProviderConfig (lines 192–195) and sanitizeConfigForRequest (lines 294–300) already call the same isReasoningModel(model) helper. The shared logic is already extracted.

CometActions:

### Access Information
- **URL:** https://pr-7008.dev.comet.com
- **Namespace:** pr-7008
- **Version:** 2.0.63-7008-merge-2265
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7008%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

apps/opik-frontend/src/v2/pages-shared/llm/ManageAIProviderDialog/CloudAIProviderDetails.tsx
```diff
@@ -64,6 +81,40 @@ const CloudAIProviderDetails: React.FC<CloudAIProviderDetailsProps> = ({
...
+                  <SelectValue placeholder="Select pipeline mode" />
+                </SelectTrigger>
+                <SelectContent>
+                  {PIPELINE_MODE_OPTIONS.map((opt) => (
+                    <SelectItem key={opt.value} value={opt.value}>
+                      {opt.label}
+                    </SelectItem>
+                  ))}
+                </SelectContent>
+              </Select>
```
andriidudar: 💡 **suggestion** | Architecture

Reuse the shared `SelectBox` (`@/shared/SelectBox`) here instead of hand-composing the `Select` primitives. The sibling provider configs in this same tree — `VertexAIModelConfigs`, `AnthropicModelConfigs`, `GeminiModelConfigs` — all use it, and `PIPELINE_MODE_OPTIONS` already matches its `DropdownOption` shape. Wrapping it in `<FormControl>` also restores the a11y/validation wiring (`id`/`aria-describedby`/`aria-invalid` → `FormMessage`) that every other field in this file has.

```suggestion
              <FormControl>
                <SelectBox
                  id="openaiPipelineMode"
                  value={field.value ?? DEFAULT_OPENAI_PIPELINE_MODE}
                  onChange={field.onChange}
                  options={PIPELINE_MODE_OPTIONS}
                  placeholder="Select pipeline mode"
                />
              </FormControl>
```

You'll also need `import SelectBox from "@/shared/SelectBox/SelectBox";` and can drop the now-unused `Select`/`SelectTrigger`/`SelectContent`/`SelectItem`/`SelectValue` imports.

apps/opik-frontend/src/types/providers.ts
```diff
@@ -697,6 +697,9 @@ export interface ProviderKeyConfiguration {
   auth_header_name?: string;
   /** Custom LLM (OPIK-4551): "true" to drop the default Authorization: Bearer header */
   suppress_default_auth?: string;
+  /** OpenAI (OPIK-6833): which OpenAI pipeline to route through. Backend parses case-insensitively
+   * and falls back to chat_completions_api on unknown values. */
+  openai_pipeline_mode?: "chat_completions_api" | "responses_api";
```
andriidudar: 💡 **suggestion** | Architecture

This inline `"chat_completions_api" | "responses_api"` union duplicates the `OpenAiPipelineMode` type already defined in `ManageAIProviderDialog/schema.ts`. Since `types/` sits in a lower layer than `pages-shared/`, the canonical union belongs here — export it from this file and have `schema.ts` import it (deriving its Zod enum from it). That keeps the form schema, the persisted-config type, and `normalizeOpenAiPipelineMode` from silently drifting when a third mode is added.

apps/opik-frontend/src/v2/pages-shared/llm/ManageAIProviderDialog/ManageAIProviderDialog.tsx
```diff
@@ -284,14 +297,23 @@ const ManageAIProviderDialog: React.FC<ManageAIProviderDialogProps> = ({
     const suppressDefaultAuthValue =
       isCustomLike && suppressDefaultAuth === true ? "true" : undefined;
 
+    // Always send the explicit openai_pipeline_mode on OpenAI keys so switching back to
+    // chat_completions_api persists (otherwise an omitted key could leave a previously stored
+    // responses_api value untouched).
+    const openaiPipelineMode = form.getValues("openaiPipelineMode");
+    const openaiPipelineModeValue = isOpenAi
+      ? openaiPipelineMode ?? DEFAULT_OPENAI_PIPELINE_MODE
```
andriidudar: 🧹 **nit** | Style

`openaiPipelineMode ?? DEFAULT_OPENAI_PIPELINE_MODE` can't reach the fallback: the field is always seeded through `normalizeOpenAiPipelineMode(...)` (defaultValues/`setValue`) or the literal default (`reset`), and `normalizeOpenAiPipelineMode` is typed non-nullable. The same dead `?? DEFAULT_OPENAI_PIPELINE_MODE` is on the `<Select value=…>` in `CloudAIProviderDetails.tsx`. Harmless, but it implies the value can be `undefined` when by construction it can't.

apps/opik-frontend/src/types/providers.ts
yaricom: Looks like this comment was already added above

apps/opik-frontend/src/v2/pages-shared/llm/ManageAIProviderDialog/ManageAIProviderDialog.tsx
yaricom: The same here - this is duplication of earlier comment

apps/opik-frontend/src/v2/pages-shared/llm/ManageAIProviderDialog/CloudAIProviderDetails.tsx
yaricom: fixed

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/LlmProviderOpenAiResponses.java
```diff
@@ -0,0 +1,115 @@
...
+    public void generateStream(
+            @NonNull ChatCompletionRequest request,
+            @NonNull String workspaceId,
+            @NonNull Consumer<ChatCompletionResponse> handleMessage,
+            @NonNull Runnable handleClose,
+            @NonNull Consumer<Throwable> handleError) {
+        var strictJsonSchema = LlmProviderOpenAiResponsesMapper.extractRequestedStrictJsonSchema(request);
+        var streamingChatModel = clientGenerator.newResponsesApiStreamingChatModel(config, strictJsonSchema);
+        var chatRequest = LlmProviderOpenAiResponsesMapper.toChatRequest(request);
+        streamingChatModel.chat(chatRequest, new StreamingChatResponseHandler() {
```
LifeXplorer: can we extract `new StreamingChatResponseHandler()` into a helper method?

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/LlmProviderOpenAiResponses.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    @Override
+    public void validateRequest(@NonNull ChatCompletionRequest request) {
+        // No additional validation beyond what the Responses-API ChatModel enforces.
+    }
+
+    @Override
+    public Optional<ErrorMessage> getLlmProviderError(@NonNull Throwable throwable) {
+        // openai-java's OpenAIServiceException (BadRequestException, RateLimitException, …) wraps
```
LifeXplorer: I'd suggest to move such comments to method javadoc

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/LlmProviderOpenAiResponses.java
```diff
@@ -0,0 +1,115 @@
...
+        // the failed HTTP call. Its getMessage() is the short "<status>: <text>" form, not the JSON
+        // body, so LlmProviderLangChainMapper's JSON-scan logic skips it and the request bubbles up
+        // as a generic 500. Extract the status + diagnostic directly from the typed exception.
+        return findOpenAiServiceException(throwable)
+                .map(LlmProviderOpenAiResponses::toErrorMessage)
+                .or(() -> LlmProviderLangChainMapper.INSTANCE.getErrorObject(throwable, log));
+    }
+
+    private static Optional<OpenAIServiceException> findOpenAiServiceException(Throwable throwable) {
```
LifeXplorer, line 84: can be non-static

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/LlmProviderOpenAiResponses.java
```diff
@@ -0,0 +1,115 @@
...
+        return Throwables.getCausalChain(throwable).stream()
+                .filter(OpenAIServiceException.class::isInstance)
+                .map(OpenAIServiceException.class::cast)
+                .findFirst();
+    }
+
+    private static ErrorMessage toErrorMessage(OpenAIServiceException ex) {
+        // Prefer the diagnostic code (e.g., "unsupported_parameter") for HTTP status; fall back to
```
LifeXplorer: same to javadoc

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/LlmProviderOpenAiResponsesMapper.java
```diff
@@ -0,0 +1,399 @@
...
+        Optional.ofNullable(request.topP()).ifPresent(builder::topP);
+        Optional.ofNullable(resolveMaxOutputTokens(request)).ifPresent(builder::maxOutputTokens);
+
+        // frequency_penalty, presence_penalty, stop, and top_k are NOT supported by the OpenAI
+        // Responses API — langchain4j's OpenAiOfficialResponsesChatModel.validate() throws
+        // UnsupportedFeatureException for any of them. Clients (notably the playground) often
+        // send frequency_penalty: 0 / presence_penalty: 0 as framework defaults even when the
+        // user didn't tune them, so the request would fail before reaching OpenAI. We silently
+        // drop these fields rather than 500 the call; a user who explicitly tunes them and
+        // expects them to apply should use pipeline_mode=CHAT_COMPLETIONS_API.
```
LifeXplorer: that comment should probably be javadoc for `warnIfDroppedSamplingParam`.
Moreover if log level debug is fine for this operation, let's rename it to `logDroppedSamplingParam`, as warn is misleading

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/LlmProviderOpenAiResponsesMapper.java
```diff
@@ -0,0 +1,399 @@
...
+                .build();
+    }
+
+    private List<ChatMessage> toLangChainMessages(@NonNull ChatCompletionRequest request) {
+        return request.messages().stream().map(LlmProviderOpenAiResponsesMapper::toLangChainMessage).toList();
+    }
+
+    private ChatMessage toLangChainMessage(@NonNull Message message) {
+        return switch (message) {
+            case dev.langchain4j.model.openai.internal.chat.SystemMessage sys -> SystemMessage.from(sys.content());
```
LifeXplorer, line 155: can move `dev.langchain4j.model.openai.internal.chat` to imports?

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/LlmProviderOpenAiResponsesMapper.java
```diff
@@ -0,0 +1,399 @@
...
+            case dev.langchain4j.model.openai.internal.chat.UserMessage user -> UserMessage.from(
+                    MessageContentNormalizer.flattenContent(user.content()));
+            // OpikUserMessage is a sibling of openai-internal UserMessage (it implements Message
+            // directly to add video-URL support); MessageContentNormalizer rewrites user turns into
+            // this type whenever the inbound content is multimodal. Flatten to text — full
+            // multimodal propagation is deferred per the class Javadoc.
+            case com.comet.opik.domain.llm.langchain4j.OpikUserMessage opikUser -> UserMessage.from(
```
LifeXplorer, line 162: same with imports

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/LlmProviderOpenAiResponsesMapper.java
```diff
@@ -0,0 +1,399 @@
...
+                .ifPresent(builder::parameters);
+        return builder.build();
+    }
+
+    private JsonObjectSchema toJsonObjectSchema(@NonNull Map<String, Object> parameters) {
+        var element = JsonSchemaElementJsonUtils.fromMap(parameters);
+        if (element instanceof JsonObjectSchema objectSchema) {
+            return objectSchema;
+        }
+        // Tool parameters are required to be an object schema per the OpenAI spec; non-object root
```
LifeXplorer: javadoc

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/LlmProviderOpenAiResponsesMapper.java
```diff
@@ -0,0 +1,399 @@
...
+     * the proxy provider to pick a strict-vs-loose {@code OpenAiOfficialResponsesChatModel} variant
+     * (langchain4j configures strict mode at model build time via {@code .strictJsonSchema(...)}, not
+     * per-request). Returns {@code true} only when the request explicitly sets {@code strict: true}
+     * on a {@code json_schema} response format; everything else (no {@code response_format}, text,
+     * json_object, json_schema with {@code strict} absent or false) returns {@code false}.
+     */
+    static boolean extractRequestedStrictJsonSchema(@NonNull ChatCompletionRequest request) {
+        var responseFormat = request.responseFormat();
+        if (responseFormat == null
+                || responseFormat.type() != dev.langchain4j.model.openai.internal.chat.ResponseFormatType.JSON_SCHEMA
```
LifeXplorer, line 305: import

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/LlmProviderOpenAiResponsesMapper.java
```diff
@@ -0,0 +1,399 @@
...
+
+    /**
+     * Maps OpenAI's {@code response_format} to langchain4j's {@link ResponseFormat}. {@code text}
+     * returns {@code null} (langchain4j's default). {@code json_object} → loose JSON mode.
+     * {@code json_schema} → structured-output mode with name + root schema. The per-request
+     * {@code strict} flag is not represented on langchain4j's {@link JsonSchema} (it lives on the
+     * model builder via {@code .strictJsonSchema(...)}); see {@link #extractRequestedStrictJsonSchema}
+     * for the helper the provider uses to pick the right ChatModel variant per call.
+     */
+    private ResponseFormat toResponseFormat(dev.langchain4j.model.openai.internal.chat.ResponseFormat inputFormat) {
```
LifeXplorer, line 321: Move to import section. Relates to all other places

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/LlmProviderOpenAiResponsesMapper.java
```diff
@@ -0,0 +1,399 @@
...
+        return switch (finishReason) {
+            case STOP -> "stop";
+            case LENGTH -> "length";
+            case TOOL_EXECUTION -> "tool_calls";
+            case CONTENT_FILTER -> "content_filter";
+            default -> "other";
+        };
+    }
+
+    private void warnIfDroppedSamplingParam(@NonNull ChatCompletionRequest request) {
```
LifeXplorer: Nit: no need to put constraints on private access level methods

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/LlmProviderOpenAiResponsesMapper.java
```diff
@@ -0,0 +1,399 @@
...
+            log.debug("Dropping unsupported 'presence_penalty'='{}' for OpenAI Responses API",
+                    request.presencePenalty());
+        }
+        if (CollectionUtils.isNotEmpty(request.stop())) {
+            log.debug("Dropping unsupported 'stop'='{}' for OpenAI Responses API", request.stop());
+        }
+    }
+
+    private Integer resolveMaxOutputTokens(@NonNull ChatCompletionRequest request) {
+        if (request.maxCompletionTokens() != null) {
```
LifeXplorer: ObjectUtils.firstNonNull(request.maxCompletionTokens(), request.maxTokens())

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/OpenAIClientGenerator.java
```diff
@@ -5,20 +5,39 @@
...
 
 @RequiredArgsConstructor
+@Slf4j
 public class OpenAIClientGenerator implements LlmProviderClientGenerator<OpenAiClient> {
 
+    private static final String CONFIG_KEY_PIPELINE_MODE = "openai_pipeline_mode";
+
+    // langchain4j's OpenAiOfficialResponsesChatModel constructor requires a non-null modelName at
+    // build time. On the proxy path the real model is supplied per request via
+    // ChatRequest.parameters().modelName(...), so this placeholder never reaches OpenAI.
```
LifeXplorer, line 33: If that's a known placeholder that never reaches OpenAi, let's keep the value verbouse. E.g. smth like PROXY_MODEL_NAME_PLACEHOLDER = "placeholder-model-value".
In that case we force to fail is this value will be anyhow used

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/OpenAiErrorMessage.java
```diff
@@ -31,6 +31,14 @@ public ErrorMessage toErrorMessage() {
     }
 
     private Integer getCode(OpenAiError error) {
-        return OpenAiCompatStatusCodes.fromCode(error.code);
+        // OpenAI's error payload carries both a high-level {type} ("invalid_request_error", ...) and
```
LifeXplorer: javadoc

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/openai/LlmProviderOpenAiResponsesMapperTest.java
```diff
@@ -0,0 +1,610 @@
...
+                    .model("gpt-4o-mini")
+                    .addUserMessage("hi")
+                    .temperature(0.7)
+                    .topP(0.9)
+                    .maxCompletionTokens(512)
+                    .build();
+
+            var actual = LlmProviderOpenAiResponsesMapper.toChatRequest(request);
+
+            assertThat(actual.modelName()).isEqualTo("gpt-4o-mini");
```
LifeXplorer: It's better compare with `request.get...` values than duplicate literals. Relates to all tests

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/openai/LlmProviderOpenAiResponsesMapperTest.java
```diff
@@ -0,0 +1,610 @@
...
+
+    }
+
+    @Nested
+    @TestInstance(TestInstance.Lifecycle.PER_CLASS)
+    class StreamingChunks {
+
+        @Test
+        void partialChunkCarriesAssistantRoleAndContent() {
+            var originalRequest = ChatCompletionRequest.builder()
```
LifeXplorer: `ChatCompletionRequest.builder()` is mostly duplicated in majority of tests. I'd suggest to move to helper initFunction with dynamic message/tool input

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/openai/LlmProviderOpenAiResponsesTest.java
```diff
@@ -0,0 +1,455 @@
...
+        var inner = openAiRateLimit("Too many requests", "rate_limit_exceeded", "rate_limit_error");
+        var wrapped = new RuntimeException("retry wrapper", inner);
+
+        var result = provider.getLlmProviderError(wrapped);
+
+        assertThat(result).isPresent();
+        assertThat(result.get().getCode()).isEqualTo(429);
+    }
+
+    private static BadRequestException openAiBadRequest(String message, String code, String type) {
```
LifeXplorer, line 372: prefer non-static where possible

LifeXplorer (commented):
Reviewed BE part.
Nice work. Mostly cosmetic comments added

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/LlmProviderOpenAiResponses.java
yaricom: findOpenAiServiceException doesn't touch instance state (no this.config, this.clientGenerator, or this.log access), it's a pure transformation of its argument. Keeping it static is correct and matches the sibling helper toErrorMessage right below it. Changing it to an instance method would just imply a dependency that doesn't exist.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/LlmProviderOpenAiResponsesMapper.java
yaricom: No — the remaining FQN references all clash with already-imported types from other langchain4j packages. Each name is taken by the version used in many more places, so the rare
  openai-internal usages have to qualify inline:

  | openai-internal type (FQN'd) | Conflicts with imported type | Where the import wins |
  |---|---|---|
  | `…internal.chat.SystemMessage` | `dev.langchain4j.data.message.SystemMessage` (L10) | `SystemMessage.from(...)` at L150 |
  | `…internal.chat.UserMessage` | `dev.langchain4j.data.message.UserMessage` (L12) | `UserMessage.from(...)` at L151 and L158 |
  | `…internal.chat.ResponseFormat` | `dev.langchain4j.model.chat.request.ResponseFormat` (L15) | `ResponseFormat.JSON` / `ResponseFormat.builder()` — many usages |
  | `…internal.chat.ResponseFormatType` | `dev.langchain4j.model.chat.request.ResponseFormatType` (L16) | `ResponseFormatType.JSON` at L346 |
  | `…internal.chat.JsonSchema` | `dev.langchain4j.model.chat.request.json.JsonSchema` (L19) | `JsonSchema.builder()` at L347 |

  Java has no type aliasing, so the only ways to remove the FQNs would be:

  1. **Flip the choice for one of them** — import the openai-internal version and qualify the data/chat.request siblings. Worse: the langchain4j-side types appear 3–10× as often, and
  you'd lose the natural short name on the heavy side.
  2. **Wrap each clashing openai-internal type in a private nested alias** — pure boilerplate, and IDE navigation gets worse.

  Recommendation: leave the FQNs. The current scheme already imports the openai-internal names that *don't* clash (`AssistantMessage`, `ChatCompletionRequest`, `Message`, `Tool`,
  `ToolCall`, …) and only inlines the five names that genuinely can't be imported alongside their langchain4j siblings.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/LlmProviderOpenAiResponsesMapper.java
yaricom: The same as in previous comment

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/LlmProviderOpenAiResponsesMapper.java
yaricom: The same as in previous comment

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/LlmProviderOpenAiResponsesMapper.java
yaricom: The same as in previous comment

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/openai/LlmProviderOpenAiResponsesTest.java
yaricom: Yes, technically it can be non-static — but there's no reason to change it.
  
  - The method doesn't reference any instance state (no `this`, no instance fields). It mocks everything inline.
  - The class uses the default JUnit lifecycle (`PER_METHOD`, no `@TestInstance` annotation), so a new test instance is created per test anyway. Making the helper instance-level wouldn't
  enable sharing state across tests.
  - `static` signals "pure helper, callable without an instance" and matches the sibling helpers in the same file (`streamingModelEmitting`, `StubStreamingChatModel`, `StubChatModel`).

  Drop `static` only if the helper later needs to read something from a `@Mock`-injected field or a `@BeforeEach`-initialized fixture. None of that applies here, so leaving it static is
  the right choice — same reasoning as the earlier `findOpenAiServiceException` question.

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7008`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/OpenAIClientGenerator.java
yaricom: renamed

LifeXplorer (approved):
Nice work!

PR #7015, opened 2026-06-10
[OPIK-6482] [BE] fix: fetch dashboards page in two steps to avoid filesort OOM on JSON config

thiagohora:
## Details

The dashboards list endpoint can return HTTP 500 (`java.sql.SQLException: Out of sort memory`) on workspaces with enough dashboards. The list query does `SELECT * ... ORDER BY id DESC` and, when the optimizer cannot satisfy the ordering from an index, falls back to a filesort. On MySQL 8.0.20+ the filesort uses the addon-fields algorithm, which packs the **selected columns** — including the large JSON `config` column — into the sort buffer; once a single row's `config` exceeds `sort_buffer_size`, MySQL raises `ER_OUT_OF_SORTMEMORY`. `max_length_for_sort_data` no longer forces the rowid algorithm in 8.0.20+, so the only reliable control is to keep `config` out of the sorted projection.

This fetches the page in two steps, so the JSON column never passes through a sort buffer, regardless of config size or sort field:
- `findPageIdsSorted` — orders and paginates, selecting only `id` (a narrow sort that cannot exceed `sort_buffer_size`).
- `findByIds` — loads the row bodies for that page with no `ORDER BY`; `DashboardService` restores the order in memory from the id list.

Follows the same "defer wide-column reads past pagination" pattern already used for the traces/spans list queries.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-6482

## AI-WATERMARK

AI-WATERMARK: yes

- If yes:
  - Tools: Claude Code
  - Model(s): Claude Opus 4.8
  - Scope: Implementation of the two-step fetch (DAO + service) and supporting query/plan analysis, under the author's direction.
  - Human verification: Author proposed the two-step solution and conducted the evidence-based analysis — reproducing `ER_OUT_OF_SORTMEMORY` and validating that the two-step plan removes the filesort via `EXPLAIN ANALYZE` — and reviewed the diff.

## Testing

- `mvn -o compile` in `apps/opik-backend` — BUILD SUCCESS.
- Reproduced the failure and validated the fix against two live environments (call them **environment A** and **environment B**), MySQL 8.0.42:
  - In **environment A**, the optimizer chose a `workspace_id`-only index and filesorted; the original query raised `ER_OUT_OF_SORTMEMORY` once a single `config` exceeded `sort_buffer_size`. In **environment B**, the optimizer used a composite index (reverse scan, no filesort) and never errored — confirming the bug is optimizer/data-distribution dependent and not reliably fixed by an index alone.
  - `EXPLAIN ANALYZE` confirms step 1 sorts only `id` (narrow) and step 2 has no sort node; the two-step succeeds even with `sort_buffer_size` forced to the 32 KB minimum against a >100 KB `config` row, where the original query fails.
-  `DashboardsResourceTest` integration suite passed.

## Documentation

No user-facing documentation changes.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/DashboardService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/DashboardService.java
@@ -30,9 +30,12 @@ import ru.vyarus.guicey.jdbi3.tx.TransactionTemplate;
 import java.sql.SQLIntegrityConstraintViolationException;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import static com.comet.opik.infrastructure.db.TransactionTemplateAsync.READ_ONLY;
 import static com.comet.opik.infrastructure.db.TransactionTemplateAsync.WRITE;
@@ -192,8 +195,21 @@ class DashboardServiceImpl implements DashboardService {
             int offset = (page - 1) * size;
 
             long total = dao.findCount(workspaceId, nameTerm, projectId, scope.getValue(), filtersSql, filterMapping);
-            List<Dashboard> dashboards = dao.find(workspaceId, nameTerm, projectId, scope.getValue(), filtersSql,
-                    filterMapping, sortingFieldsSql, size, offset);
+
+            // Two-step fetch (OPIK-6482): order and paginate on id only, then load the page bodies and
+            // re-order them in memory. This keeps the large JSON `config` column out of every sort
+            // buffer, which otherwise triggers ER_OUT_OF_SORTMEMORY on the filesort.
+            List<UUID> pageIds = dao.findPageIdsSorted(workspaceId, nameTerm, projectId, scope.getValue(),
+                    filtersSql, filterMapping, sortingFieldsSql, size, offset);
+
+            List<Dashboard> dashboards;
+            if (pageIds.isEmpty()) {
+                dashboards = List.of();
+            } else {
+                Map<UUID, Dashboard> byId = dao.findByIds(pageIds, workspaceId).stream()
+                        .collect(Collectors.toMap(Dashboard::id, Function.identity()));
+                dashboards = pageIds.stream().map(byId::get).filter(Objects::nonNull).toList();
+            }
 
             log.info("Found '{}' dashboards with scope '{}' in workspace '{}'", total, scope, workspaceId);
             return DashboardPage.builder()
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/DashboardService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/DashboardService.java
@@ -30,7 +30,6 @@ import ru.vyarus.guicey.jdbi3.tx.TransactionTemplate;
 import java.sql.SQLIntegrityConstraintViolationException;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
@@ -208,7 +207,7 @@ class DashboardServiceImpl implements DashboardService {
             } else {
                 Map<UUID, Dashboard> byId = dao.findByIds(pageIds, workspaceId).stream()
                         .collect(Collectors.toMap(Dashboard::id, Function.identity()));
-                dashboards = pageIds.stream().map(byId::get).filter(Objects::nonNull).toList();
+                dashboards = pageIds.stream().map(byId::get).toList();
             }
 
             log.info("Found '{}' dashboards with scope '{}' in workspace '{}'", total, scope, workspaceId);
--- a/apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DashboardsResourceTest.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DashboardsResourceTest.java
@@ -733,6 +733,49 @@ class DashboardsResourceTest {
         }
 
         @Test
+        @DisplayName("Sort by name with pagination returns every page's content in order")
+        void findDashboards__whenSortingByNameWithPagination__thenReturnEachPageContentInSortedOrder() {
+            String apiKey = UUID.randomUUID().toString();
+            String workspaceName = "test-workspace-" + UUID.randomUUID();
+            String workspaceId = UUID.randomUUID().toString();
+            mockTargetWorkspace(apiKey, workspaceName, workspaceId);
+
+            // Random names (from the generated dashboards) mean the name order differs from id/creation
+            // order, so the two-step fetch's in-memory reorder is exercised under a non-id sort (OPIK-6482).
+            // Creation order is irrelevant (results are re-sorted by name), so create in parallel.
+            List<Dashboard> created = PodamFactoryUtils.manufacturePojoList(podamFactory, Dashboard.class)
+                    .parallelStream()
+                    .map(generated -> dashboardResourceClient.createPartialDashboard().name(generated.name()).build())
+                    .map(dashboard -> {
+                        var id = dashboardResourceClient.create(dashboard, apiKey, workspaceName);
+                        return dashboardResourceClient.get(id, apiKey, workspaceName, HttpStatus.SC_OK);
+                    })
+                    .toList();
+
+            List<Dashboard> sortedByName = created.stream()
+                    .sorted(Comparator.comparing(Dashboard::name, String.CASE_INSENSITIVE_ORDER))
+                    .toList();
+
+            var sorting = SortingField.builder().field("name").direction(Direction.ASC).build();
+
+            // Traverse the whole result set page by page; assert each page's full content (recursive
+            // comparison, not just ids) matches the corresponding slice of the name-sorted list.
+            int size = 2;
+            int totalPages = (sortedByName.size() + size - 1) / size;
+            for (int page = 1; page <= totalPages; page++) {
+                List<Dashboard> expectedPage = sortedByName.stream()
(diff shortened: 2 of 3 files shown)


apps/opik-backend/src/main/java/com/comet/opik/domain/DashboardService.java
```diff
@@ -192,8 +195,21 @@ public DashboardPage find(int page, int size, String name, UUID projectId, List<
...
+            // buffer, which otherwise triggers ER_OUT_OF_SORTMEMORY on the filesort.
+            List<UUID> pageIds = dao.findPageIdsSorted(workspaceId, nameTerm, projectId, scope.getValue(),
+                    filtersSql, filterMapping, sortingFieldsSql, size, offset);
+
+            List<Dashboard> dashboards;
+            if (pageIds.isEmpty()) {
+                dashboards = List.of();
+            } else {
+                Map<UUID, Dashboard> byId = dao.findByIds(pageIds, workspaceId).stream()
+                        .collect(Collectors.toMap(Dashboard::id, Function.identity()));
```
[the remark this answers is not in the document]
thiagohora: Addressed in 4551f94: added `findDashboards__whenSortingByNameWithPagination__thenReturnEachPageContentInSortedOrder`. It creates dashboards with random names (so name order ≠ id order), sorts by `name`, and traverses **every** page, asserting each `DashboardPage.content()` slice against the expected full dashboard objects via recursive comparison (not just ids). This exercises the two-step in-memory reorder under a custom `sorting_fields` across all pages. Full `SortDashboards` suite is green (11/11).

CometActions:

### Access Information
- **URL:** https://pr-7015.dev.comet.com
- **Namespace:** pr-7015
- **Version:** 2.0.61-7015-merge-2256
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7015%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

apps/opik-backend/src/main/java/com/comet/opik/domain/DashboardService.java
```diff
@@ -192,8 +195,21 @@ public DashboardPage find(int page, int size, String name, UUID projectId, List<
...
+                dashboards = pageIds.stream().map(byId::get).filter(Objects::nonNull).toList();
```
andrescrz: Minor: why is this filtering needed? The `pageIds` should be always populated with the found `ids`. If the `dashboard` is found by `id`, it should be non-null? 

I believe this defensive check isn't needed and you can just trust the results from `findByIds`. 

The only chance should be if a dashboard is deleted, but this is within a transaction, not sure if dirty reads are possible here.

Ok to keep it if you prefer, but just double checking the intention of this.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DashboardsResourceTest.java
```diff
@@ -732,6 +732,49 @@ static Stream<Arguments> findDashboards__whenSortingByValidFields__thenReturnDas
...
+
+            // Random names (from the generated dashboards) mean the name order differs from id/creation
+            // order, so the two-step fetch's in-memory reorder is exercised under a non-id sort (OPIK-6482).
+            // Creation order is irrelevant (results are re-sorted by name), so create in parallel.
+            List<Dashboard> created = PodamFactoryUtils.manufacturePojoList(podamFactory, Dashboard.class)
+                    .parallelStream()
+                    .map(generated -> dashboardResourceClient.createPartialDashboard().name(generated.name()).build())
+                    .map(dashboard -> {
+                        var id = dashboardResourceClient.create(dashboard, apiKey, workspaceName);
+                        return dashboardResourceClient.get(id, apiKey, workspaceName, HttpStatus.SC_OK);
```
andrescrz, line 751: Minor: it'd be great if you could batch both operations. Not sure if endpoints already exists. If not, skip. But if at least once is available, please go ahead.

andrescrz (dismissed):
My main question is if you have tested the new queries, with real production data, in a heavily loaded workspace and projectId. 

The idea would be:
1. Reproduce the issue.
2. Check what are the gains and if no longer reproducible.

Anyway, this is reasonable enough to move forward for now.

apps/opik-backend/src/main/java/com/comet/opik/domain/DashboardService.java
thiagohora: Good catch — removed in caf3bbc. Both `findPageIdsSorted` and `findByIds` run in the same `READ_ONLY` transaction (consistent snapshot), so every `pageId` resolves in `findByIds` and the filter could never drop an element. Trusting the results now.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DashboardsResourceTest.java
thiagohora: There is no batch-create endpoint for dashboards (only single `POST /v1/private/dashboards` plus `delete-batch`), so I kept the setup as parallel single creates (`parallelStream`), which already speeds it up. Happy to switch to a bulk endpoint if one gets added.

thiagohora:
@andrescrz on testing with real, loaded data — yes, following your two steps:

**1. Reproduced.** On a real, heavily-used environment (MySQL 8.0.42) the original `SELECT * … ORDER BY id DESC` list query filesorts and packs the JSON `config` column into the sort buffer; once a single row's `config` exceeds `sort_buffer_size` it raises `ER_OUT_OF_SORTMEMORY` (500). Confirmed via `performance_schema` that the **only** queries ever raising 1038 were the two dashboards `SELECT *` digests (the list and the name-search variant).

**2. Gains / no longer reproducible.** `EXPLAIN ANALYZE` on the two-step shows step 1 sorts **only `id`** (narrow, covering-index scan) and step 2 has **no sort node**, so `config` never enters a sort buffer. It succeeds even with `sort_buffer_size` forced to the 32 KB minimum against a ~130 KB `config` row — exactly the condition where the original query 500s. Server-side execution time is on par with the single query (~0.1 ms on the tested workspaces); the only added cost is one extra app↔DB round-trip (sub-ms in-region).

One extra finding worth noting: the optimizer's index choice here is **data-dependent** — a selective vs single-valued `scope` column flips whether it uses the composite index or falls back to the filesort — which is why adding an index alone wasn't a reliable fix across environments, and the two-step (which never sorts the wide column) is robust regardless.

apps/opik-backend/src/main/java/com/comet/opik/domain/DashboardService.java
```diff
@@ -192,8 +194,21 @@ public DashboardPage find(int page, int size, String name, UUID projectId, List<
...
+            List<UUID> pageIds = dao.findPageIdsSorted(workspaceId, nameTerm, projectId, scope.getValue(),
+                    filtersSql, filterMapping, sortingFieldsSql, size, offset);
+
+            List<Dashboard> dashboards;
+            if (pageIds.isEmpty()) {
+                dashboards = List.of();
+            } else {
+                Map<UUID, Dashboard> byId = dao.findByIds(pageIds, workspaceId).stream()
+                        .collect(Collectors.toMap(Dashboard::id, Function.identity()));
+                dashboards = pageIds.stream().map(byId::get).toList();
```
andrescrz, line 210: Minor: it's pointless now to `collect` first and then `get` again. Just return the results of the `findByIds` calls. This is consuming CPU, memory and runtime for no reason.

andrescrz (approved):
LGTM, but I recommend removing the unnecessary operation.

apps/opik-backend/src/main/java/com/comet/opik/domain/DashboardService.java
thiagohora: The `collect` + `get` isn't to dedup/refetch — it's to **restore the sort order**. `findByIds` has no `ORDER BY` on purpose (that's what keeps the wide `config` column out of a sort buffer), and `WHERE id IN (...)` does **not** guarantee result order — MySQL returns rows in index order regardless of the id-list order.

Verified on a real instance: passing the IN list **shuffled** (cross_pipeline first) still came back in id order:

```
IN list order : cross_pipeline, arena, cadence, arena_content_qa, content_qa
DB returns     : arena, arena_content_qa, cadence, content_qa, cross_pipeline   (index order, not input order)
```

So returning `findByIds` directly would drop the requested ordering (e.g. sort by `name`/`created_at`), and the new `…SortingByNameWithPagination…` test would fail. The map is O(page-size) (≤ one page), so the overhead is negligible — and it's the only way to order the page without sorting the JSON column in SQL. I'll keep it; happy to add a short comment clarifying why.

andrescrz: The clarification makes sense. 

Maybe a collect with sorting call or a sort call would have been enough and more explicit, assuming performance and memory consumption are similar. 

Anyway, this is fine.

thiagohora: The sorting/pagination is handled by the DB; we just avoid sorting on findByIds because, in this case, the JSON field would be included in the record set, and therefore an OOM may occur.

PR #7020, opened 2026-06-10
[OPIK-6820] [BE] Internal backend endpoints

andriidudar:
Adds internal backend API endpoints.


apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/AiSpendResource.java
```diff
@@ -0,0 +1,160 @@
...
+import static com.comet.opik.utils.AsyncUtils.setRequestContext;
+
+@Path("/v1/private/ai-spend")
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+@Timed
+@Slf4j
+@RequiredArgsConstructor(onConstructor_ = @Inject)
+@Tag(name = "AI Spend", description = "Coding-agent spend analytics")
+public class AiSpendResource {
```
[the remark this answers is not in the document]
andriidudar: Skipping — `ai-spend` is an analytics/feature namespace, not an entity collection, so the plural-resource convention (traces/spans/datasets) doesn't apply here. It's consistent with the existing singular feature resources in the codebase: `/v1/private/manual-evaluation`, `/retention`, `/welcome-wizard`, `/attachment`, `/auth`, `/chat`, `/llm`. "AiSpends" also isn't idiomatic English, and renaming the base path would break the frontend hooks calling `/v1/private/ai-spend`.

apps/opik-frontend/src/api/ai-spend/useAiSpendSummary.ts
```diff
@@ -0,0 +1,61 @@
...
+  {
+    projectName,
+    intervalStart,
+    intervalEnd,
+    workspaceName,
+  }: UseAiSpendSummaryParams,
+) => {
+  const { data } = await api.post<SpendSummaryResponse>(
+    `${AI_SPEND_REST_ENDPOINT}summary`,
+    {
```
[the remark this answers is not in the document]
andriidudar: Moved to the frontend PR — this PR was split into **backend-only**, and the AI Spend Manager UI (this file) now lives on a dedicated FE branch (`andriid/OPIK-6820-ai-spend-fe`). This comment is out of scope here and will be addressed on that PR.

apps/opik-frontend/src/contexts/AiSpendContext.tsx
```diff
@@ -0,0 +1,45 @@
...
+  projectName: string;
+};
+
+// Production fallback (used when the comet AiSpendProvider is absent, e.g. OSS):
+// the area stays locked. The comet plugin overrides this with real values.
+// const DEFAULT_VALUE: AiSpendContextValue = {
+//   isPending: false,
+//   hasAccess: false,
+//   projectName: "",
+// };
```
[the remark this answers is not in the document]
andriidudar: Moved to the frontend PR — this PR was split into **backend-only**, and the AI Spend Manager UI (this file) now lives on a dedicated FE branch (`andriid/OPIK-6820-ai-spend-fe`). This comment is out of scope here and will be addressed on that PR.

apps/opik-frontend/src/plugins/comet/UserMenu.tsx
```diff
@@ -84,6 +86,8 @@ const UserMenu = () => {
   );
 
   const { canInviteMembers } = useUserPermission();
+  const navigate = useNavigate();
+  const { hasAccess: hasAiSpendAccess } = useAiSpendManager();
   const [inviteSearchQuery, setInviteSearchQuery] = useState("");
   const [isInviteSubmenuOpen, setIsInviteSubmenuOpen] = useState(false);
```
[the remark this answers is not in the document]
andriidudar: Moved to the frontend PR — this PR was split into **backend-only**, and the AI Spend Manager UI (this file) now lives on a dedicated FE branch (`andriid/OPIK-6820-ai-spend-fe`). This comment is out of scope here and will be addressed on that PR.

apps/opik-frontend/src/plugins/comet/UserMenu.tsx
```diff
@@ -285,19 +289,38 @@ const UserMenu = () => {
...
-              </DropdownMenuItem>
+              {hasAiSpendAccess && (
+                <DropdownMenuItem
+                  className="cursor-pointer"
+                  onClick={() =>
+                    navigate({
+                      to: "/$workspaceName/ai-spend/home",
+                      params: { workspaceName },
+                    })
+                  }
```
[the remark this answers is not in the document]
andriidudar: Moved to the frontend PR — this PR was split into **backend-only**, and the AI Spend Manager UI (this file) now lives on a dedicated FE branch (`andriid/OPIK-6820-ai-spend-fe`). This comment is out of scope here and will be addressed on that PR.

apps/opik-frontend/src/v2/pages-shared/AiUsageRecommendations/RecommendationCard.tsx
```diff
@@ -0,0 +1,115 @@
...
+import { formatCost } from "@/lib/money";
+import { getLaneMeta } from "@/v2/pages-shared/AiUsageBreakdown/laneRegistry";
+import { SpendRecommendation } from "@/api/ai-spend/useAiSpendRecommendations";
+
+export type RecommendationCardVariant = "full" | "compact";
+
+const IMPACT: Record<string, { label: string; color: string }> = {
+  high: { label: "High impact", color: "#ef4444" },
+  medium: { label: "Medium impact", color: "#f59e0b" },
+  low: { label: "Low impact", color: "#94a3b8" },
```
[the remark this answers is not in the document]
andriidudar: Moved to the frontend PR — this PR was split into **backend-only**, and the AI Spend Manager UI (this file) now lives on a dedicated FE branch (`andriid/OPIK-6820-ai-spend-fe`). This comment is out of scope here and will be addressed on that PR.

apps/opik-frontend/src/v2/pages-shared/AiUsageBreakdown/laneRegistry.ts
```diff
@@ -0,0 +1,123 @@
...
+  icon: LucideIcon;
+  color: string;
+  labelFallback: string;
+}
+
+const REGISTRY: Record<string, LaneMeta> = {
+  prior_assistant: {
+    icon: ScrollText,
+    color: "#6bdf93",
+    labelFallback: "Prior assistant context",
```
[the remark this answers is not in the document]
andriidudar: Moved to the frontend PR — this PR was split into **backend-only**, and the AI Spend Manager UI (this file) now lives on a dedicated FE branch (`andriid/OPIK-6820-ai-spend-fe`). This comment is out of scope here and will be addressed on that PR.

apps/opik-frontend/src/contexts/AiSpendContext.tsx
```diff
@@ -0,0 +1,45 @@
...
+//   - projectName            project the spend metrics are read from
+//   - organizationName       label shown in the AI Spend sidebar
+// Swap back to the locked fallback above before merging.
+const DEFAULT_VALUE: AiSpendContextValue = {
+  isPending: false,
+  hasAccess: true,
+  spendWorkspaceName: "default",
+  organizationName: "Test feature",
+  projectName: "claude-code",
+};
```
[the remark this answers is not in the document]
andriidudar: Valid catch — this is a real issue, flagged for the FE PR.

Moved to the frontend PR — this PR was split into **backend-only**, and the AI Spend Manager UI (this file) now lives on a dedicated FE branch (`andriid/OPIK-6820-ai-spend-fe`). This comment is out of scope here and will be addressed on that PR.

apps/opik-frontend/src/v2/pages/AiSpendLeaderboardPage/AiSpendLeaderboardPage.tsx
```diff
@@ -0,0 +1,224 @@
...
+  );
+
+  const { data, isPending } = useAiSpendUsers(
+    {
+      projectName,
+      intervalStart,
+      intervalEnd,
+      page: 1,
+      size: FETCH_SIZE,
+    },
```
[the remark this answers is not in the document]
andriidudar: Valid catch — this is a real issue, flagged for the FE PR.

Moved to the frontend PR — this PR was split into **backend-only**, and the AI Spend Manager UI (this file) now lives on a dedicated FE branch (`andriid/OPIK-6820-ai-spend-fe`). This comment is out of scope here and will be addressed on that PR.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/AiSpendResource.java
```diff
@@ -0,0 +1,161 @@
...
+    @Operation(operationId = "getSpendSummary", summary = "Get spend summary", description = "Get coding-agent spend KPI summary", responses = {
+            @ApiResponse(responseCode = "200", description = "Spend summary", content = @Content(schema = @Schema(implementation = WorkspaceMetricsSummaryResponse.class))),
+            @ApiResponse(responseCode = "400", description = "Bad Request", content = @Content(schema = @Schema(implementation = ErrorMessage.class)))
+    })
+    @RequiredPermissions(WorkspaceUserPermission.PROJECT_DATA_VIEW)
+    public Response getSpendSummary(
+            @RequestBody(content = @Content(schema = @Schema(implementation = SpendMetricRequest.class))) @NotNull @Valid SpendMetricRequest request) {
+
+        String workspaceId = requestContext.get().getWorkspaceId();
+        log.info("Retrieve spend summary on workspace_id '{}'", workspaceId);
```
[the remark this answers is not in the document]
andriidudar: Skipping — these endpoints take a structured request body (a `filters` array plus interval bounds and `project_name`), which doesn't map cleanly to query params. Opik's existing read-only analytics endpoints (workspace metrics summary, trace/span search) use `@POST` with a request body for the same reason, so this follows the established convention. The shared workspaceId/logging/contextWrite cleanup is a fair nit — we'll fold it in once this graduates from PoC.

apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendDAO.java
```diff
@@ -0,0 +1,656 @@
...
+    private static final String CURRENT = ID_CURRENT + " AND " + TS_CURRENT;
+    private static final String PREVIOUS = ID_PREVIOUS + " AND " + TS_PREVIOUS;
+    private static final String WINDOW = ID_WINDOW + " AND " + TS_WINDOW;
+    private static final String RANGE = ID_RANGE + " AND " + TS_RANGE;
+    private static final String RANGE_ALIASED = "s.id BETWEEN :id_start AND :id_end"
+            + " AND s.start_time BETWEEN parseDateTime64BestEffort(:ts_start, 9)"
+            + " AND parseDateTime64BestEffort(:ts_end, 9)";
+
+    private static final int BREAKDOWN_ITEM_LIMIT = 200;
+
```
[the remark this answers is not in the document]
andriidudar: Skipping — `BREAKDOWN_ITEM_LIMIT` is an internal safety cap, not an operator-facing knob yet. Keeping it a constant for this PoC; we'll promote it to a typed, env-backed config property if/when it needs runtime tuning.

apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendDAO.java
```diff
@@ -0,0 +1,656 @@
...
+            WHERE workspace_id = :workspace_id
+                AND project_id = :project_id
+                AND %3$s
+                AND %2$s
+                %4$s
+            GROUP BY label""";
+
+    private static final double HIGH_SPEND_FACTOR = 2.0;
+
+    // One statement: aggregate traces and spans per user into CTEs, LEFT JOIN them, then sort + page in ClickHouse.
```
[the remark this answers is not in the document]
andriidudar: Skipping — `HIGH_SPEND_FACTOR` is a PoC threshold and intentionally a constant for now. We'll move it (and any per-workspace tuning) to typed config once the high-spend definition is finalized.

apps/opik-backend/src/main/java/com/comet/opik/api/sorting/SpendUserSortingFactory.java
```diff
@@ -0,0 +1,13 @@
...
+
+import java.util.List;
+
+import static com.comet.opik.api.sorting.SortableFields.TOTAL_ESTIMATED_COST;
+
+public class SpendUserSortingFactory extends SortingFactory {
+
+    @Override
+    public List<String> getSortableFields() {
+        return List.of(TOTAL_ESTIMATED_COST, "requests", "skills", "mcps", "mcp_calls");
```
thiagohora: nit: these sortable fields are raw string literals (`"requests"`, `"skills"`, `"mcps"`, `"mcp_calls"`) that are repeated as column aliases in the `USERS` query in `AiSpendDAO`. Let's promote them to named constants (e.g. in `SortableFields` alongside `TOTAL_ESTIMATED_COST`, or a small constants holder) and reference the same constants from both the factory and the DAO so the allowlist and the SQL aliases can't drift apart.

apps/opik-backend/src/main/java/com/comet/opik/api/spend/SpendMetricRequest.java
```diff
@@ -0,0 +1,35 @@
...
+
+@Builder(toBuilder = true)
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
+public record SpendMetricRequest(
+        UUID projectId,
+        String projectName,
+        @NotNull Instant intervalStart,
+        @NotNull Instant intervalEnd,
+        @Pattern(regexp = NULL_OR_NOT_BLANK, message = "must not be blank") String userId,
```
thiagohora, line 26: As a convention we generally don't deal with `userId` in these request DTOs — let's rename this field to `userName` (and the JSON `user_id` → `user_name`) to stay consistent with the rest of the codebase.

apps/opik-backend/src/main/java/com/comet/opik/api/spend/SpendMetricRequest.java
```diff
@@ -0,0 +1,35 @@
...
+        @JsonIgnore UUID resolvedProjectId) {
+
+    @AssertTrue(message = "interval_start must be before interval_end") public boolean isStartBeforeEnd() {
+        return intervalStart == null || intervalEnd == null || intervalStart.isBefore(intervalEnd);
+    }
+
+    @AssertTrue(message = "either project_id or project_name must be provided") public boolean isProjectProvided() {
```
thiagohora: Two small things here:

1. Message wording — let's follow the phrasing used elsewhere: `"must provide either project_id or project_name"`.
2. `projectName != null && !projectName.isBlank()` should use the shared helper: `StringUtils.isNotBlank(projectName)` (Apache Commons, already used across the backend). Same applies to the `resolveUserUuid`/`userId` blank check in `AiSpendDAO`.

apps/opik-backend/src/main/java/com/comet/opik/api/spend/SpendLane.java
```diff
@@ -0,0 +1,60 @@
...
+import lombok.RequiredArgsConstructor;
+
+import java.util.Arrays;
+import java.util.Optional;
+
+@Getter
+@RequiredArgsConstructor
+public enum SpendLane {
+
+    PRIOR_ASSISTANT("prior_assistant", "Prior assistant context", Side.INPUT,
```
thiagohora: This enum lives in the `api` package but carries SQL fragments (`tokenExpression`, `breakdownArrayPath`, `breakdownLabelField`, `breakdownTokenField`). Let's decouple the persistence concern from the API contract: keep only the API-facing metadata here (`key`, `label`, `side`, `hasBreakdown`), and move the SQL expressions into a dedicated query-builder in `domain` (e.g. `AiSpendQueryBuilder`). It doesn't need to be as elaborate as the filter `FilterQueryBuilder`, but the same idea — a builder that owns the enum→SQL-expression mapping (an `EnumMap`) and renders the fragments. Same note applies to `OutputLane`.

apps/opik-backend/src/main/java/com/comet/opik/api/spend/OutputLane.java
```diff
@@ -0,0 +1,49 @@
...
+import lombok.RequiredArgsConstructor;
+
+import java.util.Arrays;
+import java.util.Optional;
+
+@Getter
+@RequiredArgsConstructor
+public enum OutputLane {
+
+    THINKING("thinking", "Thinking",
```
thiagohora: Same as `SpendLane`: this is in the `api` package but holds SQL (`breakdownLabelExpr`, `breakdownPredicate`). Let's keep the API metadata here and relocate the SQL fragments to the `domain` query-builder so the API enum stays free of persistence details.

apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendDAO.java
```diff
@@ -0,0 +1,656 @@
...
+            multiIf(
+                JSONExtractString(metadata, 'cc', 'llm_call', 'block_kind') = 'thinking', 'thinking',
+                JSONExtractString(metadata, 'cc', 'llm_call', 'block_kind') = 'text', 'assistant_text',
+                JSONExtractString(metadata, 'cc', 'llm_call', 'block_kind') = 'tool_use' AND startsWith(name, 'mcp__'), 'mcp_tool_calls',
+                JSONExtractString(metadata, 'cc', 'llm_call', 'block_kind') = 'tool_use' AND name = 'Skill', 'skill_invocations',
+                JSONExtractString(metadata, 'cc', 'llm_call', 'block_kind') = 'tool_use', 'built_in_tool_calls',
+                ''
+            )""";
+
+    private static final String COST_SUMMARY = """
```
thiagohora: Broader structural note for the DAO: the SQL-building logic (the query constants, the `String.formatted(...)` assembly, the lane→expression maps) should be isolated out of the DAO into a dedicated `AiSpendQueryBuilder`. The DAO would then be left with just statement execution + binding, which keeps each class focused (the DAO is currently doing both SQL construction and execution). This is the same builder referenced in the `SpendLane`/`OutputLane` comments.

apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendDAO.java
```diff
@@ -0,0 +1,656 @@
...
+                   SUM(JSONExtractInt(metadata, 'cc', 'llm_call', 'attributed_output_tokens')) AS total_tokens
+            FROM spans final
+            WHERE workspace_id = :workspace_id
+                AND project_id = :project_id
+                AND %3$s
+                AND %2$s
+                %4$s
+            GROUP BY label""";
+
+    private static final double HIGH_SPEND_FACTOR = 2.0;
```
thiagohora: `HIGH_SPEND_FACTOR` is a business rule, not a DAO concern. Let's move it onto a calculator/mapper class (the same place the high-spend flag is decided) and reference it from there when binding the parameter, so the threshold logic lives in one place.

apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendDAO.java
```diff
@@ -0,0 +1,656 @@
...
+
+        Mono<CountsRow> counts = querySingle(COUNTS_SUMMARY, "ai_spend_summary_counts", request.resolvedProjectId(),
+                statement -> bindSummaryRange(statement, request),
+                (row, metadata) -> new CountsRow(
+                        row.get("messages_current", Long.class),
+                        row.get("messages_previous", Long.class),
+                        row.get("active_users_current", Long.class),
+                        row.get("active_users_previous", Long.class),
+                        row.get("total_users", Long.class)),
+                new CountsRow(0L, 0L, 0L, 0L, 0L));
```
thiagohora: Instead of inlining `new CountsRow(0L, 0L, 0L, 0L, 0L)` (and `new CostRow(null, null)` above), add static factory methods `CountsRow.empty()` / `CostRow.empty()` and use those. Single source of truth for the default/empty instance and clearer intent at the call site.

apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendDAO.java
```diff
@@ -0,0 +1,656 @@
...
+    // getSTWithLogComment) before the trailing ';' to keep ClickHouse query attribution, as the other DAOs do.
+    private static String withLogComment(String sql) {
+        var trimmed = sql.stripTrailing();
+        if (trimmed.endsWith(";")) {
+            trimmed = trimmed.substring(0, trimmed.length() - 1).stripTrailing();
+        }
+        return trimmed + "\nSETTINGS log_comment = '<log_comment>'";
+    }
+
+    private String spanUserFilter(Optional<String> userUuid) {
```
thiagohora: `spanUserFilter` (and `userNameFilter`) build SQL by string concatenation, which makes the final query hard to read. Since everything already renders through StringTemplate, prefer expressing these optional fragments as template conditionals (`<if(user_uuid)> ... <endif>`) rather than concatenating fragments — it keeps the shape of the final query visible in one place.

apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendDAO.java
```diff
@@ -0,0 +1,656 @@
...
+                        .totalEstimatedCost(totalCost)
+                        .build()))
+                .build();
+    }
+
+    private WorkspaceMetricsSummaryResponse.Result result(String name, Double current, Double previous) {
+        return WorkspaceMetricsSummaryResponse.Result.builder().name(name).current(current).previous(previous).build();
+    }
+
+    private Double average(Double total, Long count) {
```
thiagohora: These conversion/cast/aggregation helpers (`average`, `toDouble`, `getLong`, `toBigDecimal`, `filterNan`, `result(...)`) shouldn't live in the DAO. Two-step preference:

1. When it can be done in the query, do it in SQL — e.g. read `total_estimated_cost` as `BigDecimal` directly from the `Decimal` column instead of `toBigDecimal(Double)` (avoids a precision-losing round-trip through `double`); clamp/`isFinite` cost in SQL instead of `filterNan` in Java.
2. Whatever genuinely can't move to SQL (e.g. `avg_cost_per_user`, which combines two separate queries) belongs in an internal mapper class, not the DAO.

thiagohora (commented):
Nice work — the API shape, permission gating, parameter binding, and test coverage are all solid. This is **batch 1 of 2**: structure, naming, and layering nits (nothing blocking on correctness or security here). None of these change behavior; they're about keeping the `api` package free of SQL, the DAO focused on execution, and business rules/constants in one place.

Summary of the inline comments:
- **`SpendUserSortingFactory`** — promote the raw sortable-field strings to shared constants reused by the `USERS` query aliases.
- **`SpendMetricRequest`** — rename `userId` → `userName`; align the "must provide either project_id or project_name" message; use `StringUtils.isNotBlank`.
- **`SpendLane` / `OutputLane`** — decouple SQL fragments from these `api` enums into a `domain` query-builder (`AiSpendQueryBuilder`).
- **`AiSpendDAO`** — isolate SQL construction into that query-builder; replace `spanUserFilter`/`userNameFilter` concatenation with StringTemplate conditionals; move conversion/cast/avg helpers to SQL where possible, otherwise an internal mapper; add `CostRow.empty()` / `CountsRow.empty()` factories; move `HIGH_SPEND_FACTOR` to a calculator/mapper class.

**Batch 2 (coming next): SQL & query-cost review.** I'll run the integration tests against a retained ClickHouse container (shutdown hook disabled) so I can capture the rendered queries and their costs, then review for missing index usage / pruning and query-optimization opportunities on the `traces`/`spans` scans. Holding those comments until I have the actual query plans.

apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendDAO.java
```diff
@@ -0,0 +1,656 @@
...
+    private static final String ID_WINDOW = "id BETWEEN :id_prior_start AND :id_end";
+    private static final String ID_RANGE = "id BETWEEN :id_start AND :id_end";
+
+    private static final String CURRENT = ID_CURRENT + " AND " + TS_CURRENT;
+    private static final String PREVIOUS = ID_PREVIOUS + " AND " + TS_PREVIOUS;
+    private static final String WINDOW = ID_WINDOW + " AND " + TS_WINDOW;
+    private static final String RANGE = ID_RANGE + " AND " + TS_RANGE;
+    private static final String RANGE_ALIASED = "s.id BETWEEN :id_start AND :id_end"
+            + " AND s.start_time BETWEEN parseDateTime64BestEffort(:ts_start, 9)"
+            + " AND parseDateTime64BestEffort(:ts_end, 9)";
```
thiagohora: **🔴 Main scaling finding — the id-range pruning rationale holds for `traces` but NOT for `spans`.**

⚠️ Measured on a **local Testcontainer with synthetic data** (not prod), but on the **real migrated schema**, so the PK/pruning conclusions generalize; the absolute rows/bytes are illustrative only. I loaded 1M traces + 1M spans under one project and measured with `EXPLAIN indexes=1` + `system.query_log`. Querying a 1%-of-data time window:

| query | table | granules read | rows read | bytes |
|---|---|---|---|---|
| composition_tokens | traces | **2 / 127** | 16,384 | 15.6 MiB |
| breakdown (tools) | traces | 2 / 127 | 16,384 | 15.6 MiB |
| composition_cost | **spans** | **123 / 127** | **1,000,000** | 107 MiB |
| output_breakdown | **spans** | 123 / 127 | **1,000,000** | 223 MiB |
| users | traces + **spans** | — | **1,032,768** | 173 MiB / 94 MiB mem |

Per `schema-pk-filter-on-orderby`: the spans queries filter on `id`, which is **not** an ORDER BY prefix column. `traces` PK = `(workspace_id, project_id, id)` → `id` is the 3rd key, so the derived `id BETWEEN` range prunes precisely ✅. `spans` PK = `(workspace_id, project_id, trace_id, parent_span_id, id)` → `id` is the **5th** key, behind the unconstrained `trace_id`/`parent_span_id`, so the index can't use it; with no partition key either, **every spans query full-scans the project's entire span history regardless of the requested window.** Confirmed: `id BETWEEN` + `start_time BETWEEN` pruned **0** granules vs `workspace+project` alone (123/127 either way).

All fixes below reuse the date→`id` boundaries the DAO already derives (`:id_start`/`:id_end`, the `ID_*` constants); each took the spans scan from **1,000,000 rows / 107 MiB → ~16,384 rows** (granules 2/127) on the test data.

**Recommended — add `s.trace_id BETWEEN :id_start AND :id_end` alongside the existing `s.id`/`s.start_time` predicates (no migration).** A span's `trace_id` is its trace's creation-time UUIDv7 `id`, so the same boundaries map onto a `trace_id` range. `trace_id` is the leading span sort-key column (after ws+proj), so it drives **primary-key** granule pruning, while the existing `id`/`start_time` predicates keep the exact window. Leveraging both is the best combination — verified **16,384 rows, granules 2/127**, all on immutable fields, no index needed.

**Optional complement — `minmax` skip index on `spans.id`** (per `query-index-skipping-indices`): makes the bare `id BETWEEN` predicate prune on its own (also 16,384 rows in my test). Valid here because `span.id` is **immutable** and forward-moving (creation-time UUIDv7), and spans sort by `trace_id` (≈ same time) so id stays clustered within granules; works under `FINAL` because the default connection params (`ANALYTICS_DB_QUERY_PARAMETERS` in `config.yml`) set `use_skip_indexes_if_final=1`. Caveat: I only confirmed this index pruning on synthetic data where `trace_id == id` (best case) — its real-world effectiveness depends on the `id`↔`trace_id` time correlation, so it'd want validation on real data before relying on it. The `trace_id`-range predicate has no such dependency and already fixes every spans query (`COST_SUMMARY`, `COMPOSITION_COST`, `COMPOSITION_OUTPUT`, `OUTPUT_BREAKDOWN`, `USERS.spend_agg`) without a migration.

apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendDAO.java
```diff
@@ -0,0 +1,656 @@
...
+                sa.model AS model,
+                toInt64(if(ifNull(sa.spend, 0) >= :high_spend_factor * avg(ifNull(sa.spend, 0)) OVER (), 1, 0)) AS high_spend,
+                COUNT() OVER () AS total
+            FROM trace_agg ta
+            LEFT JOIN spend_agg sa ON ta.user_uuid = sa.user_uuid
+            WHERE 1 = 1 %5$s
+            ORDER BY %4$s
+            LIMIT :limit OFFSET :offset
+            ;
+            """;
```
thiagohora: **🟠 `USERS` is the heaviest query.** Measured on the 1M-trace / 1M-span synthetic dataset (local container, real schema) for a single 25-row page, 1%-of-data window. read_rows is the deterministic signal; memory is peak and noisier.

**1. `spend_agg` scans all spans — confirmed.** Base `USERS` read **1,032,768 rows** (the `s.id BETWEEN` predicate doesn't prune, per the `RANGE_ALIASED` comment). Adding `s.trace_id BETWEEN :id_start AND :id_end` to the spans side → **49,152 rows (~21× fewer)**. The join already needs the windowed trace set, so the `trace_id` scoping fits naturally.

**2. `traces` is scanned twice — confirmed (and my earlier "single CTE" fix was wrong).** `EXPLAIN indexes=1` shows **two `ReadFromMergeTree (opik.traces)` nodes** (trace_agg + the spend_agg INNER JOIN subquery). ⚠️ Correction: a shared CTE does **not** fix it — ClickHouse inlines a CTE at each reference, so a `WITH windowed_traces AS (…)` reused by both still produced **2 traces scans** in my test. The structure that actually scans traces once: aggregate spans by `trace_id` first, then `LEFT JOIN` a single windowed-traces read on `trace_id = id` and group by `user_uuid`. Verified: **1 traces scan + 1 spans scan, 32,768 rows, ~47 ms** (vs 82 ms base), and **identical output** (n=500, same checksum as base).

**3. Memory — corrected.** The two `OVER ()` windows are **negligible**: removing them changed peak memory 48.99 → 49.05 MiB (no effect), so I was wrong to attribute memory to them. Peak (~94 MiB on the full query, ~49 MiB on this trimmed variant) is dominated by `FINAL` + the join hash + GROUP-BY state (incl. `groupUniqArray(repository)`), and stayed roughly **flat across 32k–1M rows read** — i.e., bounded by trace/user cardinality in the window, not raw span volume. Still worth noting the connection set has no `max_memory_usage`/`max_execution_time` ceiling (summary comment). Separately, the `trace_agg ↔ spend_agg` LEFT JOIN can be `LEFT ANY JOIN` per `query-join-use-any` — **verified identical output** (spend_agg has exactly one row per `user_uuid`, `max_rows_per_user = 1`); it's a safe clarity nit, marginal perf.

Net: the `trace_id`-scoping + one-scan restructure together took this query from **1,032,768 → 32,768 rows** with identical results.

apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendDAO.java
```diff
@@ -0,0 +1,656 @@
...
+
+        String costSql = COMPOSITION_COST.formatted(RANGE, spanUserFilter(userUuid));
+        Mono<BigDecimal> cost = querySingle(costSql, "ai_spend_composition_cost", request.resolvedProjectId(),
+                statement -> bindComposition(statement, request, userUuid),
+                (row, metadata) -> Optional.ofNullable(row.get("total_cost", BigDecimal.class)).orElse(BigDecimal.ZERO),
+                BigDecimal.ZERO);
+
+        return Mono.zip(inputTokens, outputTokens, cost)
+                .map(tuple -> buildComposition(tuple.getT1(), tuple.getT2(), tuple.getT3()));
+    }
```
thiagohora, line 122: 🟡 **Query fan-out + JSON parsing.** Measured on the 1M-trace / 1M-span synthetic dataset (local container, real schema).

**Fan-out (from code):** `getComposition` fires 3 scans (`composition_tokens` on traces, `composition_output` + `composition_cost` on spans), and `getRecommendations` calls `getComposition` again — so a full dashboard load is ~9 ClickHouse queries (3 + 3 + summary's 2 + users' 1).
- **(a) Merge the two spans scans — confirmed.** `composition_output` and `composition_cost` hit the same windowed span set. As two scans they read **2,000,000 rows / 330 MiB**; computing the per-lane token sums (`sumIf` per lane) and `SUM(total_estimated_cost)` in **one scan** reads **1,000,000 rows / 238 MiB** (and once the spans pruning fix lands, far less).
- **(b)** `recommendations` is a pure function of the composition result (`buildRecommendations`), so the service can reuse the already-fetched composition instead of re-querying — removes 3 of the 9 queries.

**JSON parsing — corrected from my earlier note.** I re-measured over 1M rows:
- Cost scales with the number of **distinct** paths parsed per row: **1 path ≈ 1.1 s CPU**, **8 distinct paths (`COMPOSITION_TOKENS`) ≈ 6.5 s**.
- **ClickHouse already dedupes repeated *identical* extracts** — the same path 9× ≈ **1.3 s** (≈ 1×). So `COUNTS_SUMMARY` extracting `user_uuid` 9× is **not** a real cost; I was wrong to flag it, and the "extract once in an inner SELECT then aggregate" trick gives **no** improvement (measured 6.9 s, ≈ the 8-path cost).
- The change that actually helps is a **single structured extract** — `JSONExtract(metadata, 'cc', 'Tuple(prior_assistant Tuple(summary Tuple(total_tokens Int64)), …)')` parses the object once: **≈ 1.9 s vs 6.5 s (~3.4×)**, same results, and `metadata` stays a `String` (query-level, no schema change).
- Context: `COMPOSITION_TOKENS`/`COUNTS_SUMMARY` are traces queries that prune to the window (~16k rows), so their parse cost is small in practice. The parse cost that bites is on the **all-rows spans** queries (e.g. `COMPOSITION_OUTPUT` parsing `block_kind` per row over the full table) — which the spans-pruning fix already shrinks by reading far fewer rows. So fix the spans scan first; the structured-extract is a secondary win.

**Minor:** `TOOLS_BREAKDOWN` scans `traces final` twice via the `UNION ALL` — **confirmed**: `EXPLAIN` shows two `ReadFromMergeTree (opik.traces)` nodes and it reads **2× the windowed rows** (32,768 vs 16,384 for one scan). Both branches could be derived from a single scan.

thiagohora (commented):
## Batch 2 — SQL & query-cost review (empirical)

⚠️ **Measured on a local Testcontainer with synthetic data — not production.** It ran the real migrated schema, so the PK/pruning/scan-count conclusions generalize; absolute rows/bytes/memory are illustrative and prod magnitudes will differ. I loaded **1,000,000 traces + 1,000,000 spans** under one project and measured every query with `EXPLAIN indexes=1` + `system.query_log`. This is about how the queries **scale**, not exact prod cost. read_rows / granules / scan-node counts are deterministic; memory is peak and noisier.

### Headline (`schema-pk-filter-on-orderby`)
**traces queries scale with the requested window; spans queries scale with the whole project's history.** For a 1%-of-data window:
- traces-based (`summary_counts`, `composition_tokens`, `breakdown`): **~16k rows / 2 of 127 granules** ✅
- spans-based (`summary_cost`, `composition_cost`, `composition_output`, `output_breakdown`, `users.spend_agg`): **all 1,000,000 rows / 123 of 127 granules** 🔴

Cause: `spans` sort key `(workspace_id, project_id, trace_id, parent_span_id, id)` — `id` is the 5th column, so `id BETWEEN` (and `start_time`, no partition/skip index) prune **0** granules vs `workspace+project` alone.

**Fix — favor an `id`-derived range on the leading key (consistent with `WorkspaceMetricsDAO`/`ProjectMetricsDAO`).** A span's `trace_id` is its trace's creation-time UUIDv7 `id`, so adding `s.trace_id BETWEEN :id_start AND :id_end` (boundaries already derived) drives **primary-key** pruning while the existing `id`/`start_time` predicates keep the exact window — verified **1,000,000 → 16,384 rows (2/127 granules)**, no migration, all immutable fields. A `minmax` index on `spans.id` is an optional complement (span `id` is immutable + forward-moving), but I only confirmed it on best-case synthetic data (`trace_id == id`), so it'd want validation on real data; the `trace_id`-range predicate has no such dependency.

### Inline comments (all empirically verified)
- **`RANGE_ALIASED` block** — spans-pruning root cause + the `trace_id`-range fix (+ optional `minmax`-on-`id`).
- **`USERS`** (heaviest) — `spend_agg` reads **1,032,768 rows**; `s.trace_id BETWEEN` → **49,152** (~21×), and a one-scan restructure (aggregate spans by `trace_id`, then join a single traces read) → **32,768 rows, 1 traces scan, identical results** (n=500, same checksum). Confirmed **traces is scanned twice** (2 `ReadFromMergeTree` nodes); a shared CTE does **not** fix it (CH inlines it — still 2 scans). Memory corrected: the two `OVER()` windows are **negligible** (48.99 vs 49.05 MiB) and peak stayed ~flat across 32k–1M rows; `LEFT JOIN` ≡ `LEFT ANY JOIN` verified (`max_rows_per_user = 1`).
- **`getComposition`** — fan-out ~9 queries; merging `composition_output`+`composition_cost` into one spans scan: **2,000,000 → 1,000,000 rows**; `recommendations` can reuse the composition (removes 3 queries). JSON corrected: CH **already dedupes repeated identical extracts** (same path 9× ≈ 1×), so the "extract-once" trick gives no gain; cost scales with **distinct** paths (1→1.1s, 8→6.5s), and a **single structured `JSONExtract(...Tuple...)`** is **~3.4× faster** (1.9s) with no schema change. `TOOLS_BREAKDOWN` scans traces twice (confirmed: 2 nodes, 32,768 vs 16,384 rows).

### Connection parameter set (re: your ask)
Default `ANALYTICS_DB_QUERY_PARAMETERS` (`config.yml`): `…&custom_http_params=max_query_size=…,async_insert*…,use_skip_indexes_if_final=1`
- **No `max_memory_usage` / `max_execution_time` / `max_rows_to_read` ceiling** — the (currently uncapped) full-history spans scans run with no server-side guardrail; worth a per-query cap on this read path.
- `use_skip_indexes_if_final=1` (default read from config, not verified on a running prod node) is what would let a `minmax` index work under `FINAL`; the existing `set`/`bloom_filter` skip indexes (`source`/`environment`/`thread_id`) aren't filtered on here.
- `async_insert*` is insert-path only; irrelevant to these reads.

### Bottom line
No correctness issues — results are right (every rewrite above was checksum-verified identical to the original). The concern is purely scaling: **spans queries get linearly slower as a project accumulates history, independent of the date filter.** Adding the `id`-derived `trace_id` range fixes pruning for all of them and is low-risk (the `spanUserFilter` already uses the `trace_id IN (…)` form) — that's the change to make before large workspaces. The one-scan restructure, output+cost merge, structured JSON extract, and recommendations reuse are follow-up wins.

_(Investigation used a throwaway local container with synthetic data on the real migrated schema — not production; no repo or PR changes were made. Structural/scan-count conclusions generalize; absolute numbers are illustrative; the `minmax`-on-`id` result was best-case and should be validated on real data.)_

apps/opik-backend/src/main/java/com/comet/opik/api/spend/SpendMetricRequest.java
andriidudar: This `user_id` isn't an Opik/Comet user reference, which is why it's not `userName`. The value is `cc.identity.user_uuid` — the Claude Code (coding-agent) end-user identity captured in the trace **metadata** by the plugin. Opik's user model identifies workspace members by `userName` (auth, `created_by`/`last_updated_by`, `RequestContext`) and has no knowledge of this id; it's an opaque external identifier used only to filter the spend metadata to one coding-agent user. The two populations differ — Opik users log into Opik, whereas these are the developers being measured.

So `userName` would be a misnomer here (the value is a UUID, and the FE sends it as `userUuid`). The field is also already part of the public contract the frontend is coded against (`user_id`), now merged to `main` across 3 hooks. Happy to rename if you'd still prefer — `userUuid` would be the accurate name — but flagging the semantics first.

apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendDAO.java
andriidudar: Fixed in 80544d87a0 — added a trace_id range to the spans queries (COST_SUMMARY, COMPOSITION_OUTPUT, COMPOSITION_COST, OUTPUT_BREAKDOWN, USERS.spend_agg) reusing the already-bound :id_start/:id_end/:id_prior_start. trace_id is the leading span sort-key column so it drives PK granule pruning; no extra binds, no migration. Test fixtures now set a windowed trace_id to honor the production invariant.

apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendDAO.java
andriidudar: Fixed in 80544d87a0 — (1) spans pruning from the trace_id-range comment applied to spend_agg; (2) the windowed trace rows are now extracted once into a shared CTE reused by trace_agg and spend_agg (no second scan); (3) the LEFT JOIN is now LEFT ANY JOIN (one row per user_uuid).

apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendDAO.java
andriidudar: Fixed in 80544d87a0 for (a)/(c)/(d): composition_output and composition_cost merged into one span scan; JSON subtrees extracted once per row (COUNTS_SUMMARY user_uuid, the merged output block_kind, and COMPOSITION_TOKENS cc subtree); TOOLS_BREAKDOWN collapsed to a single traces scan.

On (b) reusing composition for recommendations: keeping the two endpoints separate intentionally. recommendations is only a free derivation of composition today; it will likely grow its own queries (per-user trends, real cost attribution), so folding it into /composition now would be premature and force a re-split later. (a) already halves the duplicate cost; if dashboard latency ever bites first, a short-lived composition cache (or an opt-in include flag) is the cleaner lever than coupling the contracts.

apps/opik-backend/src/main/java/com/comet/opik/api/spend/SpendLane.java
andriidudar: Fixed in 80544d87a0 — SpendLane no longer carries SQL (tokenExpression / breakdown array path / label / token fields); those moved into a new domain AiSpendQueryBuilder (EnumMap lane->expression). The enum keeps only API metadata (key, label, side, breakdownSubtitle, hasBreakdown).

apps/opik-backend/src/main/java/com/comet/opik/api/spend/OutputLane.java
andriidudar: Fixed in 80544d87a0 — same as SpendLane: OutputLane's SQL (breakdownLabelExpr / breakdownPredicate) moved into AiSpendQueryBuilder; the enum keeps only API metadata.

apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendDAO.java
andriidudar: Fixed in 80544d87a0 — all SQL building (query constants, assembly, lane->expression maps) is now in AiSpendQueryBuilder. AiSpendDAO is left with statement execution, binding and row mapping only.

apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendDAO.java
andriidudar: Fixed in 80544d87a0 — HIGH_SPEND_FACTOR moved to AiSpendMapper (exposed via highSpendFactor()); the DAO references it when binding :high_spend_factor, so the threshold lives in one place.

apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendDAO.java
andriidudar: Fixed in 80544d87a0 — added CostRow.empty() / CountsRow.empty() factory methods and use them at the querySingle default-value call sites.

apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendDAO.java
andriidudar: Fixed in 80544d87a0 — the optional user/name fragments are now StringTemplate <if(user_uuid)>/<if(user_name)> conditionals instead of string concatenation; the DAO sets the flags via a Consumer<ST> resolved in the single getSTWithLogComment render.

apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendDAO.java
andriidudar: Fixed in 80544d87a0 — the response-building converters moved to AiSpendMapper. Cost is now read as BigDecimal directly from the Decimal column (summary spend, users total_estimated_cost), dropping the double round-trip and filterNan. getLong/toList remain as row-extraction null-safety in the DAO.

apps/opik-backend/src/main/java/com/comet/opik/api/sorting/SpendUserSortingFactory.java
andriidudar: Fixed in 80544d87a0 — added a SpendUserField constants holder for the leaderboard field names and reference it from SpendUserSortingFactory instead of raw literals (kept out of the global SortableFields so the generic names don't leak into trace/span sorting). The factory<->alias contract is also guarded by the users sorting test.

apps/opik-backend/src/main/java/com/comet/opik/api/spend/SpendMetricRequest.java
andriidudar: Fixed in 80544d87a0 — message reworded to "must provide either project_id or project_name" and StringUtils.isNotBlank used for the projectName and the DAO userId/name blank checks.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/AiSpendResourceTest.java
```diff
@@ -0,0 +1,854 @@
...
+    }
+
+    @AfterAll
+    void tearDownAll() {
+        wireMock.server().stop();
+    }
+
+    @Test
+    @DisplayName("summary: aggregates spend, messages and users for the current vs previous window")
+    void summary_happyPath() {
```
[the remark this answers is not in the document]
andriidudar: Skipping — checked against the actual test suite: snake_case method names are the prevailing convention here (671 snake_case vs 152 camelCase When/Returns/Throws across the v1/priv resource tests; e.g. DatasetsResourceTest 116/140, TracesResourceTest 62/127, ProjectsResourceTest 34/49). The common form is `method__whenX__thenY()`. The AiSpend test names are consistent with that dominant style, and renaming them to the camelCase `When…Returns…` form would make them *more* inconsistent with the rest of the suite, not less. `testing.md` documents that camelCase convention but the codebase largely doesn't follow it, so leaving these as-is.

apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendQueryBuilder.java
```diff
@@ -0,0 +1,336 @@
...
+    private static final String OUTPUT_BREAKDOWN = """
+            SELECT %1$s AS label,
+                   SUM(JSONExtractInt(metadata, 'cc', 'llm_call', 'attributed_output_tokens')) AS total_tokens
+            FROM spans final
+            WHERE workspace_id = :workspace_id
+                AND project_id = :project_id
+                AND %3$s
+                AND %2$s
+                <if(user_uuid)> %4$s <endif>
+            GROUP BY label""";
```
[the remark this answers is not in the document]
andriidudar: Fixed in 37deb75517 — added `HAVING label != ''` to OUTPUT_BREAKDOWN. Only `built_in_tool_calls` uses the raw span `name` (the other output lanes fall back to `'Unknown …'`), so this drops any blank-name row and matches the input breakdowns' behavior.

apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendQueryBuilder.java
```diff
@@ -0,0 +1,337 @@
...
+                    AND %4$s
+            )
+            ;
+            """.formatted(USER_UUID_EXPR, CURRENT, PREVIOUS, WINDOW);
+
+    private static final String COMPOSITION_TOKENS = """
+            SELECT
+                %1$s
+            FROM (
+                SELECT JSONExtractRaw(metadata, 'cc') AS cc
```
thiagohora: ⚠️ This `JSONExtractRaw(metadata,'cc')`-once-then-`JSONExtractInt(cc,…)` pattern is actually **slower** than the original direct `JSONExtractInt(metadata,'cc',…)`. Verified on 1M rows: **11.6 s vs 6.5 s CPU**. `JSONExtractRaw` still parses `metadata` once to slice out `cc` (≈ the whole blob), and then each of the 8 `JSONExtractInt(cc,…)` **re-parses that `cc` string** → 9 parses vs the original 8, plus materializing the intermediate column. So it regresses CPU with no benefit.

Fix — one typed extract parses the object once: 
```sql
JSONExtract(metadata, 'cc',
  'Tuple(prior_assistant Tuple(summary Tuple(total_tokens Int64)),
         tool_results   Tuple(summary Tuple(total_tokens Int64)),
         user_prompts   Tuple(summary Tuple(total_tokens Int64)),
         skills         Tuple(summary Tuple(menu_tokens Int64, loaded_tokens Int64)),
         tools          Tuple(summary Tuple(schema_tokens Int64)),
         memory         Tuple(summary Tuple(total_tokens Int64)),
         file_attachments Tuple(summary Tuple(total_tokens Int64)))') AS cc
```
then `SUM(cc.1.1.1)` etc. — measured **1.9 s vs 6.5 s (~3.4×)**, identical sums.

andriidudar: Fixed in 31bbc54c14 — replaced JSONExtractRaw + per-lane JSONExtractInt with a single typed JSONExtract(metadata, 'cc', 'Tuple(...)') that parses the object once, reading each lane via named tuple access (e.g. `cc.prior_assistant.summary.total_tokens`). One parse instead of nine; sums unchanged (AiSpendResourceTest 20/20). Used named access rather than positional cc.1.1.1 for readability. Thanks for the measurement!

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/AiSpendResourceTest.java
```diff
@@ -0,0 +1,854 @@
...
+        Map<String, WorkspaceMetricsSummaryResponse.Result> byName = summary.results().stream()
+                .collect(Collectors.toMap(WorkspaceMetricsSummaryResponse.Result::name, Function.identity()));
+
+        assertThat(byName.get("total_spend").current()).isCloseTo(8.0, within(0.0001));
+        assertThat(byName.get("total_spend").previous()).isCloseTo(4.0, within(0.0001));
+        assertThat(byName.get("total_messages").current()).isEqualTo(3.0);
+        assertThat(byName.get("total_messages").previous()).isEqualTo(1.0);
+        assertThat(byName.get("active_users").current()).isEqualTo(2.0);
+        assertThat(byName.get("total_users").current()).isEqualTo(3.0);
+        assertThat(byName.get("avg_cost_per_user").current()).isCloseTo(4.0, within(0.0001));
```
thiagohora, line 168: nit (assertions): rather than reshaping the response into a `Map` and asserting properties one-by-one, build the **expected object** and assert against the **actual** in one go. Here: construct the expected `List<WorkspaceMetricsSummaryResponse.Result>` and `assertThat(summary.results()).containsExactlyInAnyOrderElementsOf(expected)` (or `usingRecursiveComparison()`). Same applies to `composition_happyPath` (the `inputTokens`/`outputTokens` maps + per-lane asserts → assert the whole `SpendCompositionResponse`) and `users_happyPath` (the per-field `top`/`second` asserts → build expected `SpendUserRow`s and `assertThat(page.content()).usingRecursiveComparison().ignoringFields("model").isEqualTo(expected)`). Whole-object comparison catches regressions in fields the test currently doesn't look at.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/AiSpendResourceTest.java
```diff
@@ -0,0 +1,854 @@
...
+                    "tool_results": { "summary": { "total_tokens": 200, "count": 8 }, "by_tool": [ { "name": "Bash", "tokens": 200 } ] },
+                    "user_prompts": { "summary": { "total_tokens": 50 } },
+                    "skills": { "summary": { "loaded_tokens": 800, "menu_tokens": 100, "loaded_count": 2 }, "loaded": [ { "name": "opik-frontend", "body_tokens": 500 }, { "name": "find-skills", "body_tokens": 300 } ] },
+                    "tools": { "summary": { "schema_tokens": 380, "by_source": { "mcp": { "schema_tokens": 300, "available_count": 1 }, "builtin": { "schema_tokens": 80 } }, "by_server": [ { "server": "chrome-devtools", "schema_tokens": 300 } ] } },
+                    "file_attachments": { "summary": { "total_tokens": 40 }, "files": [ { "path": "a.png", "body_tokens": 40 } ] },
+                    "memory": { "summary": { "total_tokens": 60 }, "files": [ { "path": "AGENTS.md", "body_tokens": 60 } ] },
+                    "thinking": { "summary": { "total_tokens": 150 }, "by_model": [ { "model": "claude-opus-4-7", "tokens": 150 } ] },
+                    "assistant_text": { "summary": { "total_tokens": 70 } }
+                  }
+                }
```
thiagohora, line 849: nit (test data): the metadata is hand-written with magic token counts (`1000`, `200`, `800`, …) that are then echoed as magic expected sums in the assertions (`3000L`, `600L`, …) — brittle and duplicated. Prefer generating the per-lane token counts randomly (PODAM / `RandomUtils.nextInt`), build the `cc` metadata from those values, and **derive the expected aggregates from the same generated values** — so there are no magic numbers and the test still verifies exact aggregation. Incidental fields that don't drive assertions (`repository` = `"repo-a"`, emails, display names) can just be `factory.manufacturePojo(...)` random values rather than fixed strings.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/AiSpendResourceTest.java
```diff
@@ -0,0 +1,854 @@
...
+            Trace trace = factory.manufacturePojo(Trace.class).toBuilder()
+                    .id(id)
+                    .projectName(projectName)
+                    .startTime(time)
+                    .metadata(buildCcMetadata(userUuid))
+                    .feedbackScores(null)
+                    .build();
+            traceResourceClient.createTrace(trace, apiKey, workspaceName);
+            ids.add(id);
+        }
```
thiagohora, line 742: nit (batch creation): this loops and creates traces **one at a time** via `createTrace`. `TraceResourceClient` has `batchCreateTraces(List<Trace>, …)` — build the list (keeping the client-generated ids to return) and submit a single batch, like the spans helpers already do. Also `createLlmCallSpans` calls `createLlmCallSpan` 5× and each does its own `batchCreateSpans(List.of(one))` — collect the 5 spans and submit them in one `batchCreateSpans` call instead of five batches-of-one.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/AiSpendResourceTest.java
```diff
@@ -0,0 +1,854 @@
...
+        wireMock.server().stop();
+    }
+
+    @Test
+    @DisplayName("summary: aggregates spend, messages and users for the current vs previous window")
+    void summary_happyPath() {
+        var workspaceName = UUID.randomUUID().toString();
+        var workspaceId = UUID.randomUUID().toString();
+        var apiKey = UUID.randomUUID().toString();
+        AuthTestUtils.mockTargetWorkspace(wireMock.server(), apiKey, workspaceName, workspaceId, USER);
```
thiagohora, line 133: Coverage gap: every test uses a valid API key (`mockTargetWorkspace`), so authentication itself is never exercised. All 5 endpoints are `@RequiredPermissions(PROJECT_DATA_VIEW)`, so the auth dimensions should be covered like other resources do (see the `@Nested` `credentials()` pattern in `DatasetsResourceTest` / `TracesResourceTest`):

- **API-key auth + 401**: a `@ParameterizedTest` over `credentials()` with a valid key (→ 200) and a fake key (→ **401**, `UNAUTHORIZED_RESPONSE`), per endpoint. The client already supports this via the `expectedStatus` overloads.
- **Cookie / session-token auth**: the same flow through `AuthTestUtils.mockSessionCookieTargetWorkspace(...)` (valid session → 200, invalid → 401). This needs a session-token/cookie variant on `AiSpendResourceClient` (today it only sets the `Authorization` header), or assert via a raw `client.target(...).request().cookie(SESSION_COOKIE, token)`.

Worth covering both mechanisms × {valid→200, invalid→401} across the endpoints (a parameterized matrix keeps it compact).

thiagohora (approved):
Approving 🚀

The critical scaling issue is resolved and verified on a 1M-trace / 1M-span dataset: the `trace_id`-range predicates prune all spans queries (~1M → ~16k rows read), the composition output+cost merge is correct (checksum-matched), and the typed `JSONExtract(...Tuple...)` for composition tokens is ~4× faster than the original with identical sums. The structure/naming feedback (QueryBuilder/Mapper/SpendUserField extraction, `LEFT ANY JOIN`, templated filters, `BigDecimal` cost, `.empty()` factories) is all addressed.

The remaining open comments are **non-blocking and can be tackled in a follow-up**:
- Test coverage: API-key + cookie/session auth and 401 flows.
- Test nits: whole-object (expected vs actual) assertions, PODAM-generated values, batch trace creation.
- Minor perf: `USERS` still scans `traces` twice (the `windowed_traces` CTE is inlined) — immaterial now that both scans are pruned to ~16k rows.

None of these block merge. 

Great work here!

PR #7021, opened 2026-06-10
[OPIK-6810] [BE] add project_name to experiment_items_bulk to avoid Default Project fallback

thiagohora:
## Details

`experiment_items_bulk` items that use `evaluate_task_result` (no trace) had their auto-created hidden traces hardcoded to the Default Project, while the experiment lived in its own project. The async aggregation job then derived the aggregate's `project_id` from those traces, and the find query served the experiment from the aggregate path filtered by that project — so the experiment disappeared from its own project's view (and from `get_experiments_by_name(project_name=...)`) a few minutes after upload.

This PR adds an optional `project_name` on `ExperimentItemBulkUpload` and routes the whole bulk under one project — the experiment, the dataset (if created), the traces, and the spans.

- Add optional `project_name` on `ExperimentItemBulkUpload` (validated `@Pattern(NULL_OR_NOT_BLANK)`).
- Use it to create the experiment and the dataset (if they don't exist) in that project.
- Apply it to traces that don't carry their own project (auto-created, or explicit traces with a blank `project_name`); spans always inherit their trace's resolved project.
- Keep the Default-Project fallback when `project_name` is omitted, but return an `X-Opik-Deprecation` header only when a fallback actually occurred.
- Reject (422) a mismatch between request `project_name` and an item trace's `project_name` (bean-validation annotation).
- Reject (409) when `experimentId` targets an existing experiment whose project differs from a non-blank request `project_name`.
- Fix a pre-existing NPE in bulk validation when a trace is present with `null` spans.
- Experiment items now carry the resolved `project_id` (was `null` on the no-trace path).

Trace routing matrix:

| request `project_name` | item shape | trace lands in | `X-Opik-Deprecation` header? |
|---|---|---|---|
| set | no trace (`evaluate_task_result`) | request project | no |
| set | trace without a project | request project (inherited) | no |
| set | trace with a *different* project | — rejected (422) | — |
| (blank) | no trace | Default Project | yes |
| (blank) | trace without a project | Default Project (inherited) | yes |
| (blank) | trace with its own project | its own project | no |

Note: the SDK `experiment_items_bulk(project_name=...)` parameter follows from Fern regeneration of the backend OpenAPI.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- OPIK-6810

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Scope: investigation, backend implementation, tests, PR description
- Human verification: behavior confirmed against live data; tests run locally (see Testing)

## Testing

`mvn test -Dtest='ExperimentsResourceTest$BulkUpload'` — 21/21 green (Testcontainers: ClickHouse/MySQL/Redis, local). `mvn spotless:check` clean; `mvn compile` / `test-compile` green.

Scenarios covered:

Project routing (with `project_name`):
- no-trace items (`evaluate_task_result`) → auto-created trace lands in the request project; no deprecation header.
- explicit trace without its own project → trace inherits the request project.
- item with spans → spans land in the request project (same as their trace).
- experiment and dataset don't exist → both created in the request project (asserts experiment `project_id`, dataset `project_id`, and trace `project_id`).

Fallback / deprecation (no `project_name`):
- no-trace items → Default Project + `X-Opik-Deprecation` header.
- explicit trace without its own project → Default Project (inherited) + `X-Opik-Deprecation` header.
- explicit trace with its own project → stays in its own project; no header (no fallback occurred).
- legacy no-trace round-trips (`evaluate_task_result` / feedback scores) → experiment items now carry the resolved `project_id`.

Rejections:
- item trace `project_name` differs from request `project_name` → 422 (bean validation).
- `experimentId` targets an existing experiment whose project differs from a non-blank request `project_name` → 409.

Regression:
- trace present with `null` spans no longer NPEs (pre-existing bug surfaced by the inherit path).

## Documentation

- The new request field is self-documented in the OpenAPI spec via the `@Schema` description on `ExperimentItemBulkUpload.projectName` (notes the Default-Project fallback is deprecated).
- Python/TS SDK reference for `experiment_items_bulk(project_name=...)` regenerates from the backend OpenAPI via Fern (follow-up).
- No separate docs-site (`apps/opik-documentation`) change required.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/validate/ExperimentItemBulkValidator.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/validate/ExperimentItemBulkValidator.java
@@ -11,12 +11,13 @@ public class ExperimentItemBulkValidator {
     public static void validate(ExperimentItemBulkRecord item) {
         if (item.trace() == null && !CollectionUtils.isEmpty(item.spans())) {
             throw new BadRequestException("Trace is required when spans are provided");
-        } else if (item.trace() != null && !CollectionUtils.isEmpty(item.spans()) && hasMatchingTraceId(item)) {
+        } else if (item.trace() != null && hasMismatchingTraceId(item)) {
             throw new BadRequestException("Trace ID must match the span's trace ID");
         }
     }
 
-    private static boolean hasMatchingTraceId(ExperimentItemBulkRecord item) {
-        return item.spans().stream().anyMatch(span -> !item.trace().id().equals(span.traceId()));
+    private static boolean hasMismatchingTraceId(ExperimentItemBulkRecord item) {
+        return !CollectionUtils.isEmpty(item.spans())
+                && item.spans().stream().anyMatch(span -> !item.trace().id().equals(span.traceId()));
     }
 }
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/validate/ExperimentItemBulkValidator.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/validate/ExperimentItemBulkValidator.java
@@ -11,7 +11,7 @@ public class ExperimentItemBulkValidator {
     public static void validate(ExperimentItemBulkRecord item) {
         if (item.trace() == null && !CollectionUtils.isEmpty(item.spans())) {
             throw new BadRequestException("Trace is required when spans are provided");
-        } else if (item.trace() != null && hasMatchingTraceId(item)) {
+        } else if (item.trace() != null && !CollectionUtils.isEmpty(item.spans()) && hasMatchingTraceId(item)) {
             throw new BadRequestException("Trace ID must match the span's trace ID");
         }
     }
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/validation/ExperimentItemBulkUploadValidator.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/validation/ExperimentItemBulkUploadValidator.java
@@ -1,0 +1,50 @@
+package com.comet.opik.api.validation;
+
+import com.comet.opik.api.ExperimentItemBulkRecord;
+import com.comet.opik.api.ExperimentItemBulkUpload;
+import com.comet.opik.api.Trace;
+import jakarta.validation.ConstraintValidator;
+import jakarta.validation.ConstraintValidatorContext;
+import org.apache.commons.collections4.CollectionUtils;
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.Objects;
+
+public class ExperimentItemBulkUploadValidator
+        implements
+            ConstraintValidator<ExperimentItemBulkUploadValidation, ExperimentItemBulkUpload> {
+
+    private volatile ExperimentItemBulkUploadValidation constraintAnnotation;
+
+    @Override
+    public void initialize(ExperimentItemBulkUploadValidation constraintAnnotation) {
+        this.constraintAnnotation = constraintAnnotation;
+    }
+
+    @Override
+    public boolean isValid(ExperimentItemBulkUpload upload, ConstraintValidatorContext context) {
+        // No request-level project_name: item-level trace projects are unconstrained.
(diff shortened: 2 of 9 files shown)


CometActions:

### Access Information
- **URL:** https://pr-7021.dev.comet.com
- **Namespace:** pr-7021
- **Version:** 2.0.62-7021-merge-2263
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7021%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/validate/ExperimentItemBulkValidator.java
```diff
@@ -11,7 +11,7 @@ public class ExperimentItemBulkValidator {
     public static void validate(ExperimentItemBulkRecord item) {
         if (item.trace() == null && !CollectionUtils.isEmpty(item.spans())) {
             throw new BadRequestException("Trace is required when spans are provided");
-        } else if (item.trace() != null && hasMatchingTraceId(item)) {
+        } else if (item.trace() != null && !CollectionUtils.isEmpty(item.spans()) && hasMatchingTraceId(item)) {
```
BorisTkachenko: Minor: would be better to move these checks all to `hasMatchingTraceId`

apps/opik-backend/src/main/java/com/comet/opik/api/validation/ExperimentItemBulkUploadValidator.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Override
+    public void initialize(ExperimentItemBulkUploadValidation constraintAnnotation) {
+        this.constraintAnnotation = constraintAnnotation;
+    }
+
+    @Override
+    public boolean isValid(ExperimentItemBulkUpload upload, ConstraintValidatorContext context) {
+        // No request-level project_name: item-level trace projects are unconstrained.
+        if (upload == null || StringUtils.isBlank(upload.projectName()) || CollectionUtils.isEmpty(upload.items())) {
```
BorisTkachenko, line 27: What if request-level project_name is missing (will be used default), but items could have a project name? Shouldn't we fail validation in this case?

BorisTkachenko (dismissed):
Looks good, left a few comments.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/validate/ExperimentItemBulkValidator.java
thiagohora: Good call — done in 754748ada3. Folded the empty-spans guard into the helper (and renamed `hasMatchingTraceId` → `hasMismatchingTraceId`, since it returns `true` when a span's trace id doesn't match), so `validate()` reads cleanly and the helper is null-safe on its own.

apps/opik-backend/src/main/java/com/comet/opik/api/validation/ExperimentItemBulkUploadValidator.java
thiagohora: For now we're keeping the old behavior for backward compatibility — when project_name is omitted, per-item trace project_names are still respected. In the future we can make project_name required and fail validation in this case.

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7021`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

BorisTkachenko (approved):
Looks good

PR #7023, opened 2026-06-10
[BE] fix: apply above_200k tier pricing in cost calculation

### LINKED ISSUE #6982: [Bug]: Tiered (>200k token) pricing not applied — long-context costs under-reported

- [ ] Opik Python SDK
- [ ] Opik Typescript SDK
- [ ] Opik UI
- [x] Opik Server
- [ ] Documentation

### Opik version

- Opik version: reviewed against current `main` source (2026-06-06); affects latest

### Describe the problem

For models with tiered pricing above a token threshold, Opik always uses the base per-token rate regardless of token count, so total_estimated_cost is under-reported for long-context calls.

Example: gemini-2.5-pro in model_prices_and_context_window.json has:
  input_cost_per_token: 1.25e-06
  input_cost_per_token_above_200k_tokens: 2.5e-06   (2x)
  output_cost_per_token_above_200k_tokens: 1.5e-05
  cache_read_input_token_cost_above_200k_tokens / cache_creation_..._above_200k_tokens

Root cause: ModelCostData (apps/opik-backend/src/main/java/com/comet/opik/api/ModelCostData.java) doesn't parse any of the *_above_200k_tokens fields, and CostService/ModelPrice/SpanCostCalculator have no threshold logic — they always apply the flat base rate. So a >200k-token Gemini 2.5 Pro call is billed at roughly half the correct input rate.

Expected: tokens above the threshold billed at the higher tier rate (per the provider/LiteLLM convention).
Actual: everything billed at the base rate; long-context cost under-reported.

Proposed fix: parse the *_above_200k_tokens rates and apply the correct tier based on token count. Since the exact convention matters (whole-prompt-at-higher-rate vs per-token-above-threshold), I'd love to confirm the intended semantics before implementing — flagging first since it touches ModelCostData / ModelPrice / the calculators. Happy to open a PR with tests once aligned.

1. Log a Gemini 2.5 Pro span with > 200,000 input tokens.
2. Check total_estimated_cost, it's computed using input_cost_per_token (1.25e-06) for all tokens, ignoring input_cost_per_token_above_200k_tokens (2.5e-06).

### Error logs or stack trace

 N/A — silent cost under-calculation.

### Healthcheck results

_No response_

Anuj7411:
## Details

Closes #6982. Several models in `model_prices_and_context_window.json` (e.g. `gemini-2.5-pro`, `vertex_ai/claude-sonnet-4-5`) carry `*_above_200k_tokens` rate variants that LiteLLM consumes as a whole-prompt tier — once `prompt_tokens > threshold`, every token in the request is billed at the tier rate. Opik wasn't reading these fields at all, so long-context requests on those models were billed at the base rate and `total_estimated_cost` was under-reported.

The pricing data has four distinct tier suffixes today (`128k`, `200k`, `256k`, `272k`) across ~213 model entries. This PR closes the `200k` case (which is what the issue reports + the bulk of the affected entries — newer Gemini/Claude families); the same shape can be extended to the other three suffixes in a follow-up.

### Semantics

I verified against LiteLLM's source (`litellm/litellm_core_utils/llm_cost_calc/utils.py::_get_token_base_cost`) that this is **whole-prompt** pricing, not marginal — when the threshold is exceeded, the base rate is *replaced* and applied to every token; there is no "first 200k at base, remainder at tier" split. The implementation mirrors that.

### Changes

- **`ModelCostData`** — four new optional rate fields (`*_above_200k_tokens`). Jackson's `SnakeCaseStrategy` doesn't insert an underscore between letters and digits, so each field is pinned to the literal LiteLLM JSON key via `@JsonProperty` (e.g. `input_cost_per_token_above_200k_tokens`). Without that, the names silently failed to bind and the tier rate stayed null — caught by the failing-then-passing integration test below.
- **`ModelPrice`** — four corresponding `BigDecimal` tier-rate fields plus an `effectiveInputPrice(int totalPromptTokens)` / `effectiveOutputPrice` / `effectiveCacheCreationInputTokenPrice` / `effectiveCacheReadInputTokenPrice` helper. Each helper returns the tier rate iff `totalPromptTokens > TIER_THRESHOLD_200K && tierRate > 0`, otherwise the base rate. A backwards-compatible 7-arg constructor keeps the existing `SpanCostCalculatorTest` unit constructions untouched (they default the new fields to zero, which is the correct no-tier behaviour).
- **`CostService::buildModelPrice`** — populates the four tier rates from `ModelCostData`.
- **`SpanCostCalculator`** — `textGenerationWithCacheCostGoogle` and the shared `textGenerationWithCachedTokensNotIncludedInCost` helper (used by Anthropic + Bedrock) now compute `totalPromptTokens` and read effective rates. For Anthropic/Bedrock the total is `inputTokens + cacheCreation + cacheRead` (cache buckets are reported separately from `input_tokens`); for Google it's `inputTokens + cachedReadInputTokens` (Gemini's `prompt_token_count` already includes the cached portion, which the existing calculator subtracts off).

### Out of scope (follow-up PRs)

The same wiring shape extends to `128k`, `256k`, `272k` thresholds and to `textGenerationWithCacheCostOpenAI` / `textGenerationCost` (no models with `above_200k` rates currently route through those calculators in the JSON, but the helpers are ready if/when one shows up).

## Change checklist
- [ ] User facing
- [ ] Documentation update

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code (CLI)
- Model(s): Claude Opus 4.7
- Scope: assisted in mapping the LiteLLM tiered-pricing semantics to Opik's calculator pipeline, drafting the helper structure on `ModelPrice`, and writing the boundary tests.
- Human verification: read LiteLLM's `_get_token_base_cost` to confirm whole-prompt semantics, traced the JSON → `ModelCostData` → `ModelPrice` → calculator path end-to-end, and ran the affected backend tests locally (see Testing).

## Testing

Ran the affected backend tests locally (Windows, JDK 24 + Maven 3.9.9):

- `mvn -f apps/opik-backend/pom.xml -Dtest=CostServiceTest test` — added two boundary cases against the real `gemini-2.5-pro` pricing entry:
  - `prompt=200_000, completion=1_000` → asserts the **base** rate is still used (threshold is strictly `>`), `0.26`.
  - `prompt=300_000, completion=1_000` → asserts every token bills at the **`above_200k`** rate (`2.5e-6 input`, `1.5e-5 output`), `0.765`.
  Without this PR's source change, the above-threshold case computes `0.385` (base rate applied to the full prompt). **63 tests, 0 failures.**
- `mvn -f apps/opik-backend/pom.xml -Dtest=SpanCostCalculatorTest test` — regression check on the calculator unit constructions (the new 7-arg constructor on `ModelPrice` keeps them compiling). **16 tests, 0 failures.**

## Documentation

No docs change — internal cost-calculation fix with no API/user-facing surface change.


apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
```diff
@@ -336,14 +336,33 @@ private static ModelPrice buildModelPrice(ModelCostData modelCost) {
...
 
         BiFunction<ModelPrice, Map<String, Integer>, BigDecimal> calculator = resolveCalculator(provider, mode,
                 inputPrice, outputPrice, cacheCreationInputTokenPrice, cacheReadInputTokenPrice,
                 videoOutputPrice, audioInputCharacterPrice);
 
         return new ModelPrice(inputPrice, outputPrice, cacheCreationInputTokenPrice,
-                cacheReadInputTokenPrice, videoOutputPrice, audioInputCharacterPrice, calculator);
+                cacheReadInputTokenPrice, videoOutputPrice, audioInputCharacterPrice, calculator,
+                inputPriceAbove200kTokens, outputPriceAbove200kTokens,
+                cacheCreationInputTokenPriceAbove200kTokens, cacheReadInputTokenPriceAbove200kTokens);
```
andrescrz, line 365: Minor: this constructor has a lot parameters (specially of the same type), so it's easy to introduce bugs. It'd be better to use the builder instead, which is the general convention of the service anyway.

Anyhow, this problem was pre-existing in the code base prior to your PR, so this is not a blocker.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/ModelPrice.java
```diff
@@ -15,7 +15,33 @@ public record ModelPrice(
...
+     * effective-price helpers below fall through to the base rates.
+     */
+    public ModelPrice(BigDecimal inputPrice, BigDecimal outputPrice,
+            BigDecimal cacheCreationInputTokenPrice, BigDecimal cacheReadInputTokenPrice,
+            BigDecimal videoOutputPrice, BigDecimal audioInputCharacterPrice,
+            BiFunction<ModelPrice, Map<String, Integer>, BigDecimal> calculator) {
+        this(inputPrice, outputPrice, cacheCreationInputTokenPrice, cacheReadInputTokenPrice,
+                videoOutputPrice, audioInputCharacterPrice, calculator,
+                BigDecimal.ZERO, BigDecimal.ZERO, BigDecimal.ZERO, BigDecimal.ZERO);
+    }
```
andrescrz, line 44: Related to my other comment. This kind of overloaded constructors are a bad side effect of not using builders through the code base. Ideally, we could keep the canonical record constructor only and use the builder all the time, which avoid this kind of tech debt.

Anyway, not a blocker as this problem already went through in previous PRs.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
```diff
@@ -63,6 +63,38 @@ void calculateCostUsesAnthropicCacheCalculatorForClaudeOnVertexAI() {
...
+                "original_usage.candidates_token_count", 1_000);
+
+        BigDecimal baseCost = CostService.calculateCost("gemini-2.5-pro", "google_vertexai", baseTier, null);
+
+        // 200_000 * 1.25e-6 + 1_000 * 1e-5 = 0.25 + 0.01 = 0.26
+        assertThat(baseCost).isEqualByComparingTo("0.26");
+    }
+
+    @Test
+    void calculateCostUsesAbove200kTierPricingForGemini_issue6982_aboveThreshold() {
```
andrescrz, line 83: Effectively, this test is exactly the same one as before, except for the input data and the expected value. 

This should be be parameterised, for maintainability and as general convention of the service. 

Anyway, this was a pre-existing problem in this test class, so not a blocker. It's an acceptable and small piece of tech debt that we can refactor in the future.

andrescrz (approved):
Hi @Anuj7411 

I've reviewed the PR and the CI checks have passed, so I'm approving and merging it now.

This change will be included in the next release, **v2.0.63**, which is planned to go out later today within the next few hours.

I did leave a few non-blocking comments for potential improvements. No action is required before merge, but feel free to address them in a follow-up PR if you'd like.

PR #7029, opened 2026-06-11
[OPIK-6850] [BE] feat: wire read-only ClickHouse client for Agent Insights freeform SQL behind a toggle

BorisTkachenko:
## Details

Structural prerequisite for the Agent Insights freeform-SQL endpoint (OPIK-6814): wires a
dedicated **read-only ClickHouse client** into the backend, connecting as the restricted
`comet_readonly_freeform_sql_user` (provisioned by OPIK-6846), gated behind a new
**`agentInsightsEnabled`** feature toggle (default **off**). Uses the client-v2 `Client`
(which sends server settings as HTTP URL params, needed by OPIK-6814's `serverSetting(...)`),
not R2DBC — the existing privileged R2DBC `ConnectionFactory` is untouched.

Backend (`apps/opik-backend`):
- `ServiceTogglesConfig` — new `agentInsightsEnabled` toggle (`TOGGLE_AGENT_INSIGHTS_ENABLED`, default `false`).
- `OpikConfiguration` / `config.yml` — new `databaseAnalyticsReadOnly` connection block mirroring
  `databaseAnalytics`. It intentionally sets **no** `queryParameters`: the user runs under
  `readonly = 1`, so any server setting other than `SQL_workspace_id` / `SQL_project_id`
  (`CHANGEABLE_IN_READONLY`) would be rejected. Single-statement mode (no `multiquery`) by construction.
- `DatabaseAnalyticsModule` — builds and binds the read-only client as
  `@Named("readOnlyClickHouseClient") Client` (the v2 client connects lazily, so it never contacts
  ClickHouse until a query runs), lifecycle-managed for clean shutdown.
- `ClickHouseReadOnlyHealthyCheck` — auto-discovered `clickhouse-readonly` health check (critical/ready).
  Toggle-aware: when off it reports healthy without querying (no dependency on the read-only
  credentials); when on it runs `SELECT 1` on the read-only client.

Local dev (opt-in only; standard docker-compose install is unchanged):
- `scripts/dev-runner.sh` — when `TOGGLE_AGENT_INSIGHTS_ENABLED=true`, exports the read-only
  credentials and provisions the read-only user, settings profile, grants, and row policies against
  local ClickHouse. Default off = behavior unchanged.
- `clickhouse_config/additional_config.xml` — registers the inert `SQL_` `custom_settings_prefixes`,
  required so the profile DDL (and the future per-query `SQL_*` settings) are accepted.

The feature is driven by the Ollie agent (not available in local OSS installs), so the standard
docker-compose deployment is deliberately left untouched. Production user/profile/policies and Helm
env wiring are owned by OPIK-6846; the query endpoint and `EXPLAIN AST` hardening by OPIK-6814.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues
- OPIK-6850

## Testing
- Unit: `ClickHouseReadOnlyHealthyCheckTest` covers all branches — toggle off → healthy without
  touching the client; toggle on + working client → healthy; toggle on + failing client → unhealthy.
- Integration: `HealthCheckIntegrationTest` asserts the `clickhouse-readonly` check is registered as
  critical/ready and healthy with the toggle off (the default across the suite, exercising the
  toggle-off boot path).
- Manual: `TOGGLE_AGENT_INSIGHTS_ENABLED=true ./scripts/dev-runner.sh --be-only-restart` provisions
  the read-only user and the `clickhouse-readonly` check runs a real `SELECT 1` and is green; with the
  toggle off the backend boots normally and the check reports healthy without connecting.

## Documentation
New configuration introduced (no external docs changes):
- Toggle: `TOGGLE_AGENT_INSIGHTS_ENABLED` (default `false`).
- Read-only connection env: `ANALYTICS_DB_READ_ONLY_FREEFORM_SQL_USER`
  (default `comet_readonly_freeform_sql_user`), `ANALYTICS_DB_READ_ONLY_FREEFORM_SQL_PASS`.
- New `clickhouse-readonly` health check (critical/ready).

### DIFF
--- a/apps/opik-backend/Dockerfile
+++ b/apps/opik-backend/Dockerfile
@@ -46,7 +46,7 @@ RUN yum update -y && \
 WORKDIR /opt/opik
 
 # Copy application files
-COPY --chown=1001:1001 config.yml lombok.config entrypoint.sh run_db_migrations.sh opik-otel-views.yaml ./
+COPY --chown=1001:1001 config.yml lombok.config entrypoint.sh run_db_migrations.sh provision_agent_insights_readonly_user.sh opik-otel-views.yaml ./
 COPY --chown=1001:1001 redoc/ redoc/
 
 # Prepare shell scripts
--- a/apps/opik-backend/config.yml
+++ b/apps/opik-backend/config.yml
@@ -90,32 +90,17 @@ databaseAnalytics:
   #  Description: Timeout for the ClickHouse health check query (e.g. 1s, 500ms, 2 minutes)
   healthCheckTimeout: ${ANALYTICS_DB_HEALTH_CHECK_TIMEOUT:-1s}
 
-# Description: read-only analytics database connection for Agent Insights freeform SQL. Connects as a separate,
-#   restricted ClickHouse user (read-only profile + row policies, provisioned by OPIK-6846). Only used when the
-#   serviceToggles.agentInsightsEnabled toggle is on. No queryParameters are set: the user runs under readonly=1, so
-#   any server setting other than SQL_workspace_id/SQL_project_id (CHANGEABLE_IN_READONLY) would be rejected. Per-query
-#   limits live in the settings profile. Single-statement mode (no multiquery) by construction.
-databaseAnalyticsReadOnly:
-  #  Default: http
-  #  Description: The protocol used to connect to the server
-  protocol: ${ANALYTICS_DB_PROTOCOL:-HTTP}
-  #  Default: localhost
-  #  Description: The host used to connect to the server
-  host: ${ANALYTICS_DB_HOST:-localhost}
-  #  Default: 8123
-  #  Description: The port used to connect to the server
-  port: ${ANALYTICS_DB_PORT:-8123}
+# Description: credentials for the Agent Insights read-only free-form SQL ClickHouse user (restricted user with a
+#   read-only profile + row policies, provisioned by OPIK-6846). Only the user/password live here; the connection
+#   params (protocol/host/port/database) are reused from databaseAnalytics above (same ClickHouse instance). Only used
+#   when the serviceToggles.agentInsightsEnabled toggle is on.
+databaseAnalyticsReadOnlyFreeFormSql:
   #  Default: comet_readonly_freeform_sql_user
   #  Description: The read-only username used to connect to the server
   username: ${ANALYTICS_DB_READ_ONLY_FREEFORM_SQL_USER:-comet_readonly_freeform_sql_user}
   #  Default: opik
   #  Description: The read-only user password used to connect to the server
   password: ${ANALYTICS_DB_READ_ONLY_FREEFORM_SQL_PASS:-opik}
-  #  Default: opik
-  #  Description: The database name
-  databaseName: ${ANALYTICS_DB_DATABASE_NAME:-opik}
-  # Note: no healthCheckTimeout here on purpose — the clickhouse-readonly check reuses the shared
-  # clickhouse_health_check_timeout binding (same ClickHouse instance), so a separate value would be dead config.
 
 # https://www.dropwizard.io/en/stable/manual/configuration.html#health
 health:
@@ -140,7 +125,7 @@ health:
     # Non-critical on purpose: this is an optional, toggle-gated feature (Agent Insights freeform SQL). A
     # provisioning issue or transient CH rejection on the read-only user must degrade only that feature, not pull
     # the whole backend out of readiness/traffic.
-    - name: clickhouse-readonly
+    - name: clickhouse-readonly-freeform-sql
       critical: false
       type: ready
     - name: mysql
@@ -1120,7 +1105,7 @@ serviceToggles:
   agenticToolsEnabled: ${TOGGLE_AGENTIC_TOOLS_ENABLED:-"false"}
   # Default: false
   # Description: Whether or not the Agent Insights freeform SQL path is enabled. Gates the read-only ClickHouse client
-  #   and the clickhouse-readonly health check usage. When off, the read-only client is never queried.
+  #   and the clickhouse-readonly-freeform-sql health check usage. When off, the read-only client is never queried.
   agentInsightsEnabled: ${TOGGLE_AGENT_INSIGHTS_ENABLED:-"false"}
   # Default: "" (empty, no allowlisted workspaces)
   # Description: Comma-separated list of workspace IDs that should always receive V2 navigation.
--- a/apps/opik-backend/config.yml
+++ b/apps/opik-backend/config.yml
@@ -90,6 +90,34 @@ databaseAnalytics:
   #  Description: Timeout for the ClickHouse health check query (e.g. 1s, 500ms, 2 minutes)
   healthCheckTimeout: ${ANALYTICS_DB_HEALTH_CHECK_TIMEOUT:-1s}
 
+# Description: read-only analytics database connection for Agent Insights freeform SQL. Connects as a separate,
+#   restricted ClickHouse user (read-only profile + row policies, provisioned by OPIK-6846). Only used when the
+#   serviceToggles.agentInsightsEnabled toggle is on. No queryParameters are set: the user runs under readonly=1, so
+#   any server setting other than SQL_workspace_id/SQL_project_id (CHANGEABLE_IN_READONLY) would be rejected. Per-query
+#   limits live in the settings profile. Single-statement mode (no multiquery) by construction.
+databaseAnalyticsReadOnly:
+  #  Default: http
+  #  Description: The protocol used to connect to the server
+  protocol: ${ANALYTICS_DB_PROTOCOL:-HTTP}
+  #  Default: localhost
+  #  Description: The host used to connect to the server
+  host: ${ANALYTICS_DB_HOST:-localhost}
+  #  Default: 8123
+  #  Description: The port used to connect to the server
+  port: ${ANALYTICS_DB_PORT:-8123}
+  #  Default: comet_readonly_freeform_sql_user
+  #  Description: The read-only username used to connect to the server
+  username: ${ANALYTICS_DB_READ_ONLY_FREEFORM_SQL_USER:-comet_readonly_freeform_sql_user}
+  #  Default: opik
+  #  Description: The read-only user password used to connect to the server
+  password: ${ANALYTICS_DB_READ_ONLY_FREEFORM_SQL_PASS:-opik}
+  #  Default: opik
+  #  Description: The database name
+  databaseName: ${ANALYTICS_DB_DATABASE_NAME:-opik}
+  #  Default: 1s
+  #  Description: Timeout for the ClickHouse read-only health check query (e.g. 1s, 500ms, 2 minutes)
+  healthCheckTimeout: ${ANALYTICS_DB_HEALTH_CHECK_TIMEOUT:-1s}
+
 # https://www.dropwizard.io/en/stable/manual/configuration.html#health
 health:
   # Default: ["/health-check"]
@@ -108,6 +136,9 @@ health:
       critical: true
       type: ready
     - name: clickhouse
+      critical: true
+      type: ready
+    - name: clickhouse-readonly
       critical: true
       type: ready
     - name: mysql
@@ -1085,6 +1116,10 @@ serviceToggles:
   # Description: Master switch for the agentic-tools path on LLM-as-judge online scoring. When false, the
   #   inline path is used regardless of context size. Threshold lives under onlineScoring.agenticToolsThresholdTokens.
   agenticToolsEnabled: ${TOGGLE_AGENTIC_TOOLS_ENABLED:-"false"}
+  # Default: false
+  # Description: Whether or not the Agent Insights freeform SQL path is enabled. Gates the read-only ClickHouse client
+  #   and the clickhouse-readonly health check usage. When off, the read-only client is never queried.
+  agentInsightsEnabled: ${TOGGLE_AGENT_INSIGHTS_ENABLED:-"false"}
   # Default: "" (empty, no allowlisted workspaces)
   # Description: Comma-separated list of workspace IDs that should always receive V2 navigation.
   # Valid values: comma-separated workspace IDs (e.g. "ws-id-1,ws-id-2"), or "" (empty, disabled).
--- a/apps/opik-backend/config.yml
+++ b/apps/opik-backend/config.yml
@@ -114,9 +114,8 @@ databaseAnalyticsReadOnly:
   #  Default: opik
   #  Description: The database name
   databaseName: ${ANALYTICS_DB_DATABASE_NAME:-opik}
-  #  Default: 1s
-  #  Description: Timeout for the ClickHouse read-only health check query (e.g. 1s, 500ms, 2 minutes)
-  healthCheckTimeout: ${ANALYTICS_DB_HEALTH_CHECK_TIMEOUT:-1s}
+  # Note: no healthCheckTimeout here on purpose — the clickhouse-readonly check reuses the shared
+  # clickhouse_health_check_timeout binding (same ClickHouse instance), so a separate value would be dead config.
 
 # https://www.dropwizard.io/en/stable/manual/configuration.html#health
 health:
@@ -138,8 +137,11 @@ health:
     - name: clickhouse
       critical: true
       type: ready
+    # Non-critical on purpose: this is an optional, toggle-gated feature (Agent Insights freeform SQL). A
+    # provisioning issue or transient CH rejection on the read-only user must degrade only that feature, not pull
+    # the whole backend out of readiness/traffic.
     - name: clickhouse-readonly
-      critical: true
+      critical: false
       type: ready
     - name: mysql
       critical: true
--- a/apps/opik-backend/provision_agent_insights_readonly_user.sh
+++ b/apps/opik-backend/provision_agent_insights_readonly_user.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+set -euo pipefail
+
+# Provision the Agent Insights read-only free-form SQL ClickHouse user, settings profile, grants and row policies.
+#
+# Opt-in: only runs when TOGGLE_AGENT_INSIGHTS_ENABLED=true; otherwise it's a no-op so default installs are untouched.
+# This is the single local copy of the DDL, shared by docker-compose (backend container, between run_db_migrations.sh
+# and entrypoint.sh) and scripts/dev-runner.sh. It mirrors the prod provisioning owned by OPIK-6846 — keep them in
+# sync. Must run AFTER the analytics migrations (the GRANT/ROW POLICY statements reference the opik tables) and BEFORE
+# the backend starts (so the clickhouse-readonly-freeform-sql health check finds the user). The SQL_ custom-settings
+# prefix must already be registered in the ClickHouse server config (additional_config.xml locally; OPIK-6846 in prod),
+# otherwise the settings profile DDL is rejected.
+
+if [ "${TOGGLE_AGENT_INSIGHTS_ENABLED:-false}" != "true" ]; then
+    echo "Agent Insights disabled; skipping read-only ClickHouse user provisioning."
+    exit 0
+fi
+
+ch_host="${ANALYTICS_DB_HOST:-localhost}"
+ch_port="${ANALYTICS_DB_PORT:-8123}"
+ch_admin_user="${ANALYTICS_DB_USERNAME:-opik}"
+ch_admin_pass="${ANALYTICS_DB_PASS:-opik}"
+ch_db="${ANALYTICS_DB_DATABASE_NAME:-opik}"
+ro_user="${ANALYTICS_DB_READ_ONLY_FREEFORM_SQL_USER:-comet_readonly_freeform_sql_user}"
+ro_pass="${ANALYTICS_DB_READ_ONLY_FREEFORM_SQL_PASS:-opik}"
+ch_url="http://${ch_host}:${ch_port}/?user=${ch_admin_user}&password=${ch_admin_pass}"
+
+echo "Provisioning Agent Insights read-only ClickHouse user '${ro_user}' on ${ch_host}:${ch_port}/${ch_db}..."
+
+statements=(
+    "CREATE USER IF NOT EXISTS ${ro_user} IDENTIFIED BY '${ro_pass}'"
+    "CREATE SETTINGS PROFILE IF NOT EXISTS comet_llm_readonly_freeform_sql_profile SETTINGS readonly = 1, max_execution_time = 180, max_memory_usage = 8589934592, max_result_rows = 100000, result_overflow_mode = 'throw', max_rows_to_read = 100000000, read_overflow_mode = 'throw', max_concurrent_queries_for_user = 5, SQL_workspace_id = '' CHANGEABLE_IN_READONLY, SQL_project_id = '' CHANGEABLE_IN_READONLY TO ${ro_user}"
+    "GRANT SELECT ON ${ch_db}.spans TO ${ro_user}"
+    "GRANT SELECT ON ${ch_db}.traces TO ${ro_user}"
+    "GRANT SELECT ON ${ch_db}.authored_feedback_scores TO ${ro_user}"
+    "CREATE ROW POLICY IF NOT EXISTS spans_workspace_project_isolation ON ${ch_db}.spans FOR SELECT USING workspace_id = getSetting('SQL_workspace_id') AND project_id = getSetting('SQL_project_id') AS RESTRICTIVE TO ${ro_user}"
+    "CREATE ROW POLICY IF NOT EXISTS traces_workspace_project_isolation ON ${ch_db}.traces FOR SELECT USING workspace_id = getSetting('SQL_workspace_id') AND project_id = getSetting('SQL_project_id') AS RESTRICTIVE TO ${ro_user}"
+    "CREATE ROW POLICY IF NOT EXISTS authored_feedback_scores_workspace_project_isolation ON ${ch_db}.authored_feedback_scores FOR SELECT USING workspace_id = getSetting('SQL_workspace_id') AND project_id = getSetting('SQL_project_id') AS RESTRICTIVE TO ${ro_user}"
+)
+
+for stmt in "${statements[@]}"; do
+    response=$(curl -sS -w $'\n%{http_code}' "$ch_url" --data-binary "$stmt")
+    http_code="${response##*$'\n'}"
+    body="${response%$'\n'*}"
+    if [ "$http_code" != "200" ]; then
+        echo "Failed to provision read-only CH user (statement starting '${stmt%% *}...'): ${body}" >&2
+        exit 1
+    fi
+done
+
+echo "Agent Insights read-only ClickHouse user provisioned."
--- a/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/DatabaseAnalyticsReadOnlyFreeFormSqlConfig.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/DatabaseAnalyticsReadOnlyFreeFormSqlConfig.java
@@ -1,0 +1,21 @@
+package com.comet.opik.infrastructure;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.NotNull;
+import lombok.Data;
+
+/**
+ * Credentials for the Agent Insights read-only free-form SQL ClickHouse user. Scope is intentionally reduced to just
+ * the user/password: every other connection parameter (protocol, host, port, database) is shared with
+ * {@code databaseAnalytics} and reused when building the client (see {@code DatabaseAnalyticsModule}).
+ */
+@Data
+public class DatabaseAnalyticsReadOnlyFreeFormSqlConfig {
+
+    @JsonProperty
+    private @NotBlank String username;
+
+    @JsonProperty
+    private @NotNull String password;
+}
--- a/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/OpikConfiguration.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/OpikConfiguration.java
@@ -24,7 +24,7 @@ public class OpikConfiguration extends JobConfiguration {
     private DatabaseAnalyticsFactory databaseAnalytics = new DatabaseAnalyticsFactory();
 
     @Valid @NotNull @JsonProperty
-    private DatabaseAnalyticsFactory databaseAnalyticsReadOnly = new DatabaseAnalyticsFactory();
+    private DatabaseAnalyticsReadOnlyFreeFormSqlConfig databaseAnalyticsReadOnlyFreeFormSql = new DatabaseAnalyticsReadOnlyFreeFormSqlConfig();
 
     @Valid @NotNull @JsonProperty
     private AuthenticationConfig authentication = new AuthenticationConfig();
--- a/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/OpikConfiguration.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/OpikConfiguration.java
@@ -24,6 +24,9 @@ public class OpikConfiguration extends JobConfiguration {
     private DatabaseAnalyticsFactory databaseAnalytics = new DatabaseAnalyticsFactory();
 
     @Valid @NotNull @JsonProperty
+    private DatabaseAnalyticsFactory databaseAnalyticsReadOnly = new DatabaseAnalyticsFactory();
+
+    @Valid @NotNull @JsonProperty
     private AuthenticationConfig authentication = new AuthenticationConfig();
 
     @Valid @NotNull @JsonProperty
--- a/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/ClickHouseReadOnlyFreeFormSqlHealthCheck.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/ClickHouseReadOnlyFreeFormSqlHealthCheck.java
@@ -13,17 +13,17 @@ import ru.vyarus.dropwizard.guice.module.yaml.bind.Config;
 import java.util.concurrent.TimeUnit;
 
 @Singleton
-public class ClickHouseReadOnlyHealthyCheck extends NamedHealthCheck {
+public class ClickHouseReadOnlyFreeFormSqlHealthCheck extends NamedHealthCheck {
 
     private final Client readOnlyClient;
     private final boolean enabled;
     private final long healthCheckTimeoutMs;
 
     @Inject
-    public ClickHouseReadOnlyHealthyCheck(
-            @NonNull @Named(DatabaseAnalyticsModule.READ_ONLY_CLICKHOUSE_CLIENT) Client readOnlyClient,
+    public ClickHouseReadOnlyFreeFormSqlHealthCheck(
+            @NonNull @Named(DatabaseAnalyticsModule.READ_ONLY_FREE_FORM_SQL_CLICKHOUSE_CLIENT) Client readOnlyClient,
             @NonNull @Config("serviceToggles") ServiceTogglesConfig serviceToggles,
-            @Named("clickhouse_health_check_timeout") Duration healthCheckTimeout) {
+            @NonNull @Named("clickhouse_health_check_timeout") Duration healthCheckTimeout) {
         this.readOnlyClient = readOnlyClient;
         this.enabled = serviceToggles.isAgentInsightsEnabled();
         this.healthCheckTimeoutMs = healthCheckTimeout.toMilliseconds();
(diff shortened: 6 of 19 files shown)


apps/opik-backend/config.yml
```diff
@@ -90,6 +90,34 @@ databaseAnalytics:
...
+  host: ${ANALYTICS_DB_HOST:-localhost}
+  #  Default: 8123
+  #  Description: The port used to connect to the server
+  port: ${ANALYTICS_DB_PORT:-8123}
+  #  Default: comet_readonly_freeform_sql_user
+  #  Description: The read-only username used to connect to the server
+  username: ${ANALYTICS_DB_READ_ONLY_FREEFORM_SQL_USER:-comet_readonly_freeform_sql_user}
+  #  Default: opik
+  #  Description: The read-only user password used to connect to the server
+  password: ${ANALYTICS_DB_READ_ONLY_FREEFORM_SQL_PASS:-opik}
```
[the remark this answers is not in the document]
BorisTkachenko: Skipping — this mirrors the existing `databaseAnalytics` block convention (`password: ${ANALYTICS_DB_PASS:-opik}`), where prod credentials are injected via env vars and the `:-` default only serves local dev. Wiring `ANALYTICS_DB_READ_ONLY_FREEFORM_SQL_USER/_PASS` from Kubernetes secrets in the Helm chart is owned by OPIK-6846 (explicitly out of scope for this PR), and the `agentInsightsEnabled` toggle is off by default so the read-only client is never used unless explicitly enabled. Removing the local defaults would also break local dev and config validation (`username` is `@NotBlank`, `password` `@NotNull`).

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/ClickHouseReadOnlyHealthyCheck.java
```diff
@@ -0,0 +1,48 @@
...
+    private final Client readOnlyClient;
+    private final boolean enabled;
+    private final long healthCheckTimeoutMs;
+
+    @Inject
+    public ClickHouseReadOnlyHealthyCheck(
+            @NonNull @Named(DatabaseAnalyticsModule.READ_ONLY_CLICKHOUSE_CLIENT) Client readOnlyClient,
+            @NonNull @Config("serviceToggles") ServiceTogglesConfig serviceToggles,
+            @Named("clickhouse_health_check_timeout") Duration healthCheckTimeout) {
+        this.readOnlyClient = readOnlyClient;
```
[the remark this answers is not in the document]
BorisTkachenko: Skipping — intentional. Both the read-only and the regular connections target the same ClickHouse database, so a dedicated readiness timeout isn't needed. The value is still configurable via the shared `ANALYTICS_DB_HEALTH_CHECK_TIMEOUT` env var (both `databaseAnalytics` and `databaseAnalyticsReadOnly` resolve their `healthCheckTimeout` from it), so the `clickhouse-readonly` check deliberately reuses `clickhouse_health_check_timeout` rather than adding a separate provider.

deployment/docker-compose/clickhouse_config/additional_config.xml
```diff
@@ -1,4 +1,9 @@
 <clickhouse>
+    <!-- Permits SQL_-prefixed custom settings (SQL_workspace_id / SQL_project_id) used by the Agent Insights
+         read-only user's settings profile and row policies. Inert unless that user/profile is provisioned
```
thiagohora, line 3: The query layer must prevent submitted SQL from carrying its own SETTINGS/SET, since CHANGEABLE_IN_READONLY lets a user clause override the backend's SQL_workspace_id. Add a test for it

apps/opik-backend/config.yml
```diff
@@ -110,6 +138,9 @@ health:
     - name: clickhouse
       critical: true
       type: ready
+    - name: clickhouse-readonly
+      critical: true
```
thiagohora: `critical: true` couples an optional, toggle-gated feature to overall service readiness. The toggle-off short-circuit avoids this today, but the moment `agentInsightsEnabled` is turned on while the read-only user isn't provisioned (or CH transiently rejects it), this `ready` check fails and the entire backend is pulled from traffic — an outage caused by a secondary feature. Consider `critical: false` (or a non-`ready` type) so a freeform-SQL provisioning issue degrades only that feature.

apps/opik-backend/config.yml
```diff
@@ -90,6 +90,34 @@ databaseAnalytics:
...
+  username: ${ANALYTICS_DB_READ_ONLY_FREEFORM_SQL_USER:-comet_readonly_freeform_sql_user}
+  #  Default: opik
+  #  Description: The read-only user password used to connect to the server
+  password: ${ANALYTICS_DB_READ_ONLY_FREEFORM_SQL_PASS:-opik}
+  #  Default: opik
+  #  Description: The database name
+  databaseName: ${ANALYTICS_DB_DATABASE_NAME:-opik}
+  #  Default: 1s
+  #  Description: Timeout for the ClickHouse read-only health check query (e.g. 1s, 500ms, 2 minutes)
+  healthCheckTimeout: ${ANALYTICS_DB_HEALTH_CHECK_TIMEOUT:-1s}
```
thiagohora: This `healthCheckTimeout` is parsed but never used. `ClickHouseReadOnlyHealthyCheck` injects `@Named("clickhouse_health_check_timeout")`, which `DatabaseAnalyticsModule` binds from the **main** `databaseAnalytics.getHealthCheckTimeout()` — so the read-only check always uses the main DB's timeout and this field is dead config. Either add a dedicated binding for the read-only timeout or drop the field to avoid a misleading knob.

apps/opik-backend/src/test/resources/config-test.yml
```diff
@@ -74,6 +74,18 @@ databaseAnalytics:
   #  Description: Timeout for the ClickHouse health check query (e.g. 1s, 500ms, 2 minutes)
   healthCheckTimeout: 1s
 
+# Description: read-only analytics database connection for Agent Insights freeform SQL. In tests the restricted
+#   comet_readonly_freeform_sql_user does not exist in the container, so the credentials point at the standard test
+#   user; the agentInsightsEnabled toggle below is off by default, and tests flip it on when exercising the RO path.
+databaseAnalyticsReadOnly:
```
thiagohora: Because the toggle-on tests point `databaseAnalyticsReadOnly` at the unrestricted standard test user, the actual isolation mechanism — `readonly=1` profile, the table grants, and the row policies — is never exercised in CI. The integration tests only validate the toggle-off short-circuit. Worth a follow-up that provisions the real profile/policies/grants in the test ClickHouse container and asserts a cross-tenant query returns zero rows, so the isolation guarantees have at least one regression test behind them.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/DatabaseAnalyticsModule.java
```diff
@@ -36,6 +39,17 @@ public void stop() {
             }
         });
 
+        // Read-only client used for Agent Insights freeform SQL. It connects as a separate, restricted ClickHouse
+        // user (single-statement mode, no write settings). The v2 client connects lazily, so building it is cheap and
+        // never contacts ClickHouse until a query runs; the agentInsightsEnabled toggle gates all actual usage.
+        readOnlyClickHouseClient = configuration().getDatabaseAnalyticsReadOnly().buildClient();
```
thiagohora: Note (not a blocker): the read-only client is built unconditionally, regardless of `agentInsightsEnabled`. This is fine because the v2 client connects lazily (the comment is accurate), but it does mean `databaseAnalyticsReadOnly` must always be a valid `@NotBlank`/`@NotNull` block even for deployments that never use the feature. Worth being aware of that implicit always-valid-config requirement.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/ClickHouseReadOnlyHealthyCheck.java
```diff
@@ -0,0 +1,48 @@
...
+    public String getName() {
+        return "clickhouse-readonly";
+    }
+
+    @Override
+    protected Result check() {
+        if (!enabled) {
+            return Result.healthy("Agent Insights queries disabled");
+        }
+        try (var response = readOnlyClient.query("SELECT 1").get(healthCheckTimeoutMs, TimeUnit.MILLISECONDS)) {
```
thiagohora: Minor: on timeout, `.get(timeoutMs, ...)` throws `TimeoutException` (correctly caught → unhealthy), but the underlying query future isn't cancelled, so the request keeps running server-side. Low impact at health-check frequency; flagging only in case you see lingering read-only connections.

scripts/dev-runner.sh
```diff
@@ -616,6 +622,43 @@ display_ollie_process_status() {
...
 # Function to start backend
+# Provision the Agent Insights read-only ClickHouse user, settings profile, grants and row policies against the
+# local (docker-compose) ClickHouse. Opt-in: only invoked when TOGGLE_AGENT_INSIGHTS_ENABLED=true. Mirrors the prod
+# DDL owned by OPIK-6846 (db=opik). The SQL_ custom_settings_prefixes registration lives in the mounted CH config
+# (deployment/docker-compose/clickhouse_config/additional_config.xml), without which the profile DDL is rejected.
+provision_agent_insights_readonly_user() {
+    require_command curl
+    local ro_user="${ANALYTICS_DB_READ_ONLY_FREEFORM_SQL_USER:-comet_readonly_freeform_sql_user}"
+    local ro_pass="${ANALYTICS_DB_READ_ONLY_FREEFORM_SQL_PASS:-opik}"
+    local ch_url="http://localhost:${CLICKHOUSE_HTTP_PORT}/?user=opik&password=opik"
```
thiagohora: Nit: `provision_agent_insights_readonly_user` depends on `CLICKHOUSE_HTTP_PORT` being set. It's fine as currently called (from `start_backend`, after `ANALYTICS_DB_PORT=${CLICKHOUSE_HTTP_PORT}` is exported), but it's a hidden ordering dependency — fragile if this function is ever reused elsewhere. Also worth ensuring this dev DDL stays in sync with the prod provisioning (OPIK-6846) so the two copies of these security-critical statements don't drift.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/ClickHouseReadOnlyHealthyCheck.java
```diff
@@ -0,0 +1,48 @@
...
+import jakarta.inject.Named;
+import jakarta.inject.Singleton;
+import lombok.NonNull;
+import ru.vyarus.dropwizard.guice.module.installer.feature.health.NamedHealthCheck;
+import ru.vyarus.dropwizard.guice.module.yaml.bind.Config;
+
+import java.util.concurrent.TimeUnit;
+
+@Singleton
+public class ClickHouseReadOnlyHealthyCheck extends NamedHealthCheck {
```
thiagohora: Nit (non-blocking): `HealthyCheck` reads as a typo for `HealthCheck`, but it mirrors the existing `ClickHouseHealthyCheck`, so keeping it is the consistent choice. Just noting it's propagating the existing naming.

thiagohora (commented):
Infra-only review (read-only ClickHouse client for Agent Insights). Solid, well-documented, low-blast-radius change — defaults off, fails closed, and the `configuration().getDatabaseAnalytics()` fix for the now-ambiguous by-type lookup is correct and complete. Nothing here blocks merge.

Left inline comments covering: `critical: true` readiness coupling for an optional feature, an unused `databaseAnalyticsReadOnly.healthCheckTimeout`, isolation not being exercised in CI (toggle-on tests run as the unrestricted user), plus a few minor notes. One additional security item for the *query layer* (out of scope for this PR) is being raised separately.

deployment/docker-compose/clickhouse_config/additional_config.xml
BorisTkachenko: Yes, it will be implemented in https://comet-ml.atlassian.net/browse/OPIK-6814

Also you can refer to https://app.notion.com/p/cometml/Agent-Insights-technical-design-3797124010a38091bc76c51580b91d23#37a7124010a3809d981dff67ffb9376f

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/ClickHouseReadOnlyFreeFormSqlHealthCheck.java
```diff
@@ -0,0 +1,48 @@
+package com.comet.opik.infrastructure.db;
+
+import com.clickhouse.client.api.Client;
```
andrescrz, line 3: Not actionable comment: great choice with the ClickHouse V2 client! We're going to use this more and more and I think it's the appropriate choice for this health-check. 

I'm likely migrating the normal ClickHouse health-check to this, to avoid being impacted by slowness in the reactor path. 

It's starting to affect our readiness probes, and I believe this is one of the root causes for it.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/ClickHouseReadOnlyHealthyCheck.java
andrescrz: +1, let's not propagate types. I'd rename this to the right name.

apps/opik-backend/src/test/resources/config-test.yml
BorisTkachenko: The restricted user's isolation (readonly=1 + grants + row policies) is provisioned at the infra level (OPIK-6846); the backend only receives credentials and connects. Backend integration tests will never run under the actually-provisioned prod/dev/staging user — at most they could create a throwaway test user with a copy of that DDL, which exercises ClickHouse's row-policy feature plus a duplicated DDL, not any backend code or the real user. The parts that are backend responsibilities — rejecting user-supplied SETTINGS/SET and correctly passing SQL_workspace_id/SQL_project_id — are covered in OPIK-6814 with the query endpoint. So we're intentionally not adding an isolation test here: this PR is infra-only (connect as the RO user) and has nothing isolation-related to assert.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/DatabaseAnalyticsModule.java
BorisTkachenko: Intentional design choice: the v2 client connects lazily, so building it unconditionally costs nothing and never contacts ClickHouse until a query runs (gated by agentInsightsEnabled). The trade-off you note — databaseAnalyticsReadOnly must always be a valid block — is accepted; config.yml ships working defaults so it's satisfied out of the box even for deployments that never enable the feature.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/ClickHouseReadOnlyHealthyCheck.java
```diff
@@ -0,0 +1,48 @@
...
+
+    private final Client readOnlyClient;
+    private final boolean enabled;
+    private final long healthCheckTimeoutMs;
+
+    @Inject
+    public ClickHouseReadOnlyHealthyCheck(
+            @NonNull @Named(DatabaseAnalyticsModule.READ_ONLY_CLICKHOUSE_CLIENT) Client readOnlyClient,
```
andrescrz: Minor: regarding the naming. We discussed that this user is not just read only. It has some more limitations that are particular for this new feature.

We could have a general read only user in the future as well, where the most appropriate name would be just `read only`.

I'd personally rename all references to make it clear.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/ClickHouseReadOnlyHealthyCheck.java
```diff
@@ -0,0 +1,48 @@
...
+            @NonNull @Config("serviceToggles") ServiceTogglesConfig serviceToggles,
+            @Named("clickhouse_health_check_timeout") Duration healthCheckTimeout) {
```
andrescrz: Minor: missing non-null annotation on this `healthCheckTimeout`. It seems required.

Please make sure all arguments for all public methods in this PR are validated, with non-null (if applies) and not limited to that.

scripts/dev-runner.sh
BorisTkachenko: Updated

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/DatabaseAnalyticsModule.java
```diff
@@ -36,6 +39,17 @@ public void stop() {
             }
         });
 
+        // Read-only client used for Agent Insights freeform SQL. It connects as a separate, restricted ClickHouse
+        // user (single-statement mode, no write settings). The v2 client connects lazily, so building it is cheap and
```
andrescrz: Minor: please remove the specific details of the user restrictions.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/DatabaseAnalyticsModule.java
andrescrz: Happy about this in order to validate the configuration syntax even if not used.

I assume no connectivity is tested anywhere, which would be problematic and prevent this change from being pushed. (I believe it's fine).

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/OpikConfiguration.java
```diff
@@ -23,6 +23,9 @@ public class OpikConfiguration extends JobConfiguration {
     @Valid @NotNull @JsonProperty
     private DatabaseAnalyticsFactory databaseAnalytics = new DatabaseAnalyticsFactory();
 
+    @Valid @NotNull @JsonProperty
+    private DatabaseAnalyticsFactory databaseAnalyticsReadOnly = new DatabaseAnalyticsFactory();
```
andrescrz: Not a blocker: we discussed that most of the configuration params are common between the normal and the read only configuration, except for the user and credentials.

In order to avoid duplication, I recommend creating a config object which a much reduced scope.

If you still want to reuse the `DatabaseAnalyticsFactory` (desired to properly configure the client), you can just build it by feeding the values from each config (existing one and new one just for the user).

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/TestDropwizardAppExtensionUtils.java
```diff
@@ -118,6 +118,12 @@ public static TestDropwizardAppExtension newTestDropwizardAppExtension(AppContex
             configs.add("databaseAnalytics.port: " + appContextConfig.databaseAnalyticsFactory().getPort());
             configs.add("databaseAnalytics.username: " + appContextConfig.databaseAnalyticsFactory().getUsername());
             configs.add("databaseAnalytics.password: " + appContextConfig.databaseAnalyticsFactory().getPassword());
+            // The read-only client is always built; point it at the same container so toggle-on tests can connect.
+            configs.add("databaseAnalyticsReadOnly.port: " + appContextConfig.databaseAnalyticsFactory().getPort());
```
andrescrz: Minor: related to my other comment. You wouldn't need this if you follow the suggested approach.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/HealthCheckIntegrationTest.java
```diff
@@ -118,6 +118,27 @@ void test__whenHitRedisHealthyCheck__thenReturnOk(ClientSupport client) {
...
+
+        assertThat(healthChecks).hasSize(1);
+        var healthCheck = healthChecks.getFirst();
+
+        // Toggle off (default in config-test): the check reports healthy without querying the read-only user.
+        // Non-critical on purpose so a freeform-SQL issue never gates overall readiness.
+        assertThat(healthCheck.name()).isEqualTo("clickhouse-readonly");
+        assertThat(healthCheck.type()).isEqualTo("READY");
+        assertThat(healthCheck.healthy()).isTrue();
+        assertThat(healthCheck.critical()).isFalse();
```
andrescrz: Nit: I'd slightly parameterise the existing assertion method, you can overload it to avoid changing the rest of the tests.

apps/opik-backend/config.yml
```diff
@@ -90,6 +90,33 @@ databaseAnalytics:
...
+  #  Description: The read-only username used to connect to the server
+  username: ${ANALYTICS_DB_READ_ONLY_FREEFORM_SQL_USER:-comet_readonly_freeform_sql_user}
+  #  Default: opik
+  #  Description: The read-only user password used to connect to the server
+  password: ${ANALYTICS_DB_READ_ONLY_FREEFORM_SQL_PASS:-opik}
+  #  Default: opik
+  #  Description: The database name
+  databaseName: ${ANALYTICS_DB_DATABASE_NAME:-opik}
+  # Note: no healthCheckTimeout here on purpose — the clickhouse-readonly check reuses the shared
+  # clickhouse_health_check_timeout binding (same ClickHouse instance), so a separate value would be dead config.
```
andrescrz: What about the rest of the query params? Many of them are only for writes, but there are some of them interesting for reads: compress, use_skip_indexes_if_final etc.

If not reused, I recommend auditing and adding all that matter for a read only user.

deployment/docker-compose/clickhouse_config/additional_config.xml
andrescrz, line 1: Please double check which settings you need to propagate to the Helm chart.

scripts/dev-runner.sh
```diff
@@ -616,6 +622,50 @@ display_ollie_process_status() {
...
 
 # Function to start backend
+# Provision the Agent Insights read-only ClickHouse user, settings profile, grants and row policies against the
+# local (docker-compose) ClickHouse. Opt-in: only invoked when TOGGLE_AGENT_INSIGHTS_ENABLED=true. Mirrors the prod
+# DDL owned by OPIK-6846 (db=opik). The SQL_ custom_settings_prefixes registration lives in the mounted CH config
+# (deployment/docker-compose/clickhouse_config/additional_config.xml), without which the profile DDL is rejected.
+# These statements are security-critical and intentionally mirror the prod provisioning in OPIK-6846 — keep the two
+# copies in sync. Requires CLICKHOUSE_HTTP_PORT to be set (today guaranteed by start_backend, which exports it before
+# calling this); the guard below makes that dependency explicit if the function is ever reused elsewhere.
+provision_agent_insights_readonly_user() {
```
andrescrz: I'm a bit skeptical on how this user is provisioned for local testing.

If you really want to test this, I believe this should be added instead on the whole docker-compose installation.

Then, any local user would get this for free and not only whoever uses this dev-runner.sh.

Currently, this is not available if you test and develop locally in Windows or with docker-compose (some features require a docker build test, for stuff uncovered by dev-runner).

andrescrz (commented):
The majority of the feedback is non-blocking, but there are a few comments to double check before moving forward.

apps/opik-backend/config.yml
```diff
@@ -110,6 +137,12 @@ health:
     - name: clickhouse
       critical: true
       type: ready
+    # Non-critical on purpose: this is an optional, toggle-gated feature (Agent Insights freeform SQL). A
+    # provisioning issue or transient CH rejection on the read-only user must degrade only that feature, not pull
+    # the whole backend out of readiness/traffic.
+    - name: clickhouse-readonly
+      critical: false
+      type: ready
```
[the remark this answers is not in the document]
BorisTkachenko: Fixed in eeb93e21d5 — `/is-alive/ping` now honors `critical`: `IsAliveResource` gates liveness only on checks marked critical in the health config (falling back to all checks if none are configured), so the non-critical `clickhouse-readonly` check no longer fails `/is-alive`. k8s probes were already unaffected (they use `/health-check?...&type=ready|alive`, which honors `critical`). Added unit coverage for critical→500 and non-critical→200.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/health/IsAliveResourceTest.java
```diff
@@ -20,23 +24,49 @@ class IsAliveResourceTest {
...
+    }
 
-        Mockito.when(checkRegistry.runHealthChecks())
-                .thenReturn(sortedMap);
+    @Test
+    void testIsAlive__whenCriticalHealthCheckIsUnhealthy__thenReturn500() {
+        Mockito.when(checkRegistry.runHealthChecks()).thenReturn(testHealthyAndTest2Unhealthy());
 
-        var response = EXT.target("/is-alive/ping").request().get();
+        var response = EXT_CRITICAL.target("/is-alive/ping").request().get();
```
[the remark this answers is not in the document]
BorisTkachenko: Let's keep as is

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/DatabaseAnalyticsModule.java
BorisTkachenko: No connectivity is tested at startup. buildClient() only constructs the v2 client (it connects lazily on first query), and the read-only client carries no queryParameters, so there are no background health pings. The only path that queries it is the clickhouse-readonly-freeform-sql health check, and only when agentInsightsEnabled is on (off → reports healthy without touching the client). So a toggle-off deployment with no provisioned RO user boots normally and never connects as that user — config syntax is still validated, as you noted.

apps/opik-backend/config.yml
BorisTkachenko: Audited. The read-relevant knobs split two ways under `readonly=1`:
- `compress` is already enabled for this client — `buildClient()` sets `compressClientRequest`/`compressServerResponse` on every v2 client, including the read-only one, so no `queryParameters` entry is needed.
- Server settings like `use_skip_indexes_if_final` / `max_query_size` can't be sent on the request: the read-only user runs under `readonly=1` and only `SQL_workspace_id`/`SQL_project_id` are `CHANGEABLE_IN_READONLY`, so any other setting in the query is rejected. These belong in the **settings profile** (OPIK-6846), not the client URL — sending them would break every read-only query.

Net: intentionally no `queryParameters` on the read-only client. I've flagged `use_skip_indexes_if_final` (and a max_query_size audit) for the OPIK-6846 settings profile, since that's where read tuning has to live for a readonly=1 user.

deployment/docker-compose/clickhouse_config/additional_config.xml
BorisTkachenko: Double-checked. The only server-level setting this feature needs in prod is `custom_settings_prefixes: SQL_`, and it's already tracked in **OPIK-6846 → section A1** ('Register the custom-settings prefix'), to be added via `clickhouse.configuration.files` in the shared `values-opik.yaml` so all envs get it. The RO user/profile/grants/row-policies and backend env wiring (`ANALYTICS_DB_READ_ONLY_FREEFORM_SQL_USER/PASS`) are covered there too (A2/B/C). So no Helm change is needed in this PR — the `additional_config.xml` entry here just covers local docker-compose/dev-runner.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/ClickHouseReadOnlyHealthyCheck.java
BorisTkachenko: Fixed

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/OpikConfiguration.java
```diff
@@ -23,6 +23,9 @@ public class OpikConfiguration extends JobConfiguration {
     @Valid @NotNull @JsonProperty
     private DatabaseAnalyticsFactory databaseAnalytics = new DatabaseAnalyticsFactory();
 
+    @Valid @NotNull @JsonProperty
+    private DatabaseAnalyticsReadOnlyFreeFormSqlConfig databaseAnalyticsReadOnlyFreeFormSql = new DatabaseAnalyticsReadOnlyFreeFormSqlConfig();
+
```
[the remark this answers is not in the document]
BorisTkachenko: The field already carries `@Valid @NotNull @JsonProperty` (mirroring `databaseAnalytics`), so nested `@NotBlank`/`@NotNull` checks run at startup.

On the toggle-conditional part — skipping intentionally. Per andrescrz's note on this PR, we keep this block validated regardless of `agentInsightsEnabled` to verify config syntax even when the feature is off. config.yml ships the block with env-backed defaults, so it's always present and never breaks startup.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/ClickHouseReadOnlyFreeFormSqlHealthCheck.java
```diff
@@ -0,0 +1,51 @@
...
+            return Result.healthy("Agent Insights queries disabled");
+        }
+        var queryFuture = readOnlyClient.query("SELECT 1");
+        try (var response = queryFuture.get(healthCheckTimeoutMs, TimeUnit.MILLISECONDS)) {
+            return Result.healthy();
+        } catch (Exception ex) {
+            // On timeout (or any failure) cancel the future so the query doesn't keep running server-side.
+            queryFuture.cancel(true);
+            return Result.unhealthy(ex);
+        }
```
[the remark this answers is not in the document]
BorisTkachenko: Skipping — for a `SELECT 1` readiness probe, treating any failure as unhealthy is the intended behavior, and the `queryFuture.cancel(true)` in the catch already stops a timed-out query server-side. Splitting into Timeout/Execution/Interrupted branches wouldn't change the outcome (all map to `Result.unhealthy`), and this mirrors the broad-catch pattern in the existing `ClickHouseHealthyCheck`. Happy to revisit if we later need to distinguish failure modes in the health payload.

apps/opik-backend/provision_agent_insights_readonly_user.sh
```diff
@@ -0,0 +1,51 @@
...
+fi
+
+ch_host="${ANALYTICS_DB_HOST:-localhost}"
+ch_port="${ANALYTICS_DB_PORT:-8123}"
+ch_admin_user="${ANALYTICS_DB_USERNAME:-opik}"
+ch_admin_pass="${ANALYTICS_DB_PASS:-opik}"
+ch_db="${ANALYTICS_DB_DATABASE_NAME:-opik}"
+ro_user="${ANALYTICS_DB_READ_ONLY_FREEFORM_SQL_USER:-comet_readonly_freeform_sql_user}"
+ro_pass="${ANALYTICS_DB_READ_ONLY_FREEFORM_SQL_PASS:-opik}"
+ch_url="http://${ch_host}:${ch_port}/?user=${ch_admin_user}&password=${ch_admin_pass}"
```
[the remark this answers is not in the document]
BorisTkachenko: Skipping — this script only runs against a local ClickHouse (localhost) with default dev credentials (`opik`/`opik`), self-gates on `TOGGLE_AGENT_INSIGHTS_ENABLED`, and is invoked from docker-compose/dev-runner. Prod provisioning is owned by OPIK-6846 (Helm/infra), not this script. Given the localhost-only blast radius and dev creds, moving admin auth to `X-ClickHouse-User`/`X-ClickHouse-Key` headers (which are still on the curl argv anyway) isn't worth the change here.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/DatabaseAnalyticsModule.java
```diff
@@ -55,6 +83,13 @@ public Client getClickHouseClient() {
         return clickHouseClient;
     }
 
+    @Provides
+    @Singleton
+    @Named(READ_ONLY_FREE_FORM_SQL_CLICKHOUSE_CLIENT)
+    public Client getReadOnlyFreeFormSqlClickHouseClient() {
+        return readOnlyFreeFormSqlClickHouseClient;
+    }
```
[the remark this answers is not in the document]
BorisTkachenko: Already present — `getReadOnlyFreeFormSqlClickHouseClient()` is annotated with `@Provides @Singleton @Named(READ_ONLY_FREE_FORM_SQL_CLICKHOUSE_CLIENT)` (DatabaseAnalyticsModule.java lines 86-88), matching the main ClickHouse provider. The finding looks to predate the current revision; no change needed.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/DatabaseAnalyticsModule.java
```diff
@@ -36,13 +39,38 @@ public void stop() {
...
+
         ClickHouseLogAppenderConfig clickHouseLogAppenderConfig = configuration(ClickHouseLogAppenderConfig.class);
 
         // Initialize the UserFacingRuleLollingFactory
         UserFacingLoggingFactory.init(connectionFactory, clickHouseLogAppenderConfig.getBatchSize(),
                 clickHouseLogAppenderConfig.getFlushIntervalDuration());
     }
 
+    // Reuse the main analytics connection params (same ClickHouse instance) and only swap in the restricted user's
+    // credentials. No queryParameters: the user runs under readonly=1 and would reject per-query server settings.
```
andrescrz, line 60: Nit: this comment better as javadoc on the method. Great you mentioned by no parameters. Thanks for that.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/DatabaseAnalyticsReadOnlyFreeFormSqlConfig.java
```diff
@@ -0,0 +1,21 @@
...
+ * {@code databaseAnalytics} and reused when building the client (see {@code DatabaseAnalyticsModule}).
+ */
+@Data
+public class DatabaseAnalyticsReadOnlyFreeFormSqlConfig {
+
+    @JsonProperty
+    private @NotBlank String username;
+
+    @JsonProperty
+    private @NotNull String password;
```
andrescrz, line 20: Nit: I'd rather not allow blank passwords.

apps/opik-backend/Dockerfile
```diff
@@ -46,7 +46,7 @@ RUN yum update -y && \
 WORKDIR /opt/opik
 
 # Copy application files
-COPY --chown=1001:1001 config.yml lombok.config entrypoint.sh run_db_migrations.sh opik-otel-views.yaml ./
+COPY --chown=1001:1001 config.yml lombok.config entrypoint.sh run_db_migrations.sh provision_agent_insights_readonly_user.sh opik-otel-views.yaml ./
```
andrescrz, line 49: Some thought: I bit concerned from the security perspective to bundle this script inside the backend image. Anyway, for production installations this script won't be in the entry point and still need the admin creds for it. Not very different from the migrations script, so I think this is fine.

apps/opik-backend/provision_agent_insights_readonly_user.sh
```diff
@@ -0,0 +1,51 @@
...
+
+echo "Provisioning Agent Insights read-only ClickHouse user '${ro_user}' on ${ch_host}:${ch_port}/${ch_db}..."
+
+statements=(
+    "CREATE USER IF NOT EXISTS ${ro_user} IDENTIFIED BY '${ro_pass}'"
+    "CREATE SETTINGS PROFILE IF NOT EXISTS comet_llm_readonly_freeform_sql_profile SETTINGS readonly = 1, max_execution_time = 180, max_memory_usage = 8589934592, max_result_rows = 100000, result_overflow_mode = 'throw', max_rows_to_read = 100000000, read_overflow_mode = 'throw', max_concurrent_queries_for_user = 5, SQL_workspace_id = '' CHANGEABLE_IN_READONLY, SQL_project_id = '' CHANGEABLE_IN_READONLY TO ${ro_user}"
```
andrescrz, line 32: Minor: probably the `max_execution_time` and other resource limits (non-functional limits) aren't needed for a local or test environment. Not sure if capacity to exhaust them and likely not under the responsibility of this to test that. Anyway, that's fine. This comment wouldn't apply for functional limits, like ready only etc. Those are desirable.

andrescrz (approved):
Just a few more thoughts, not actionable in general. This is good to go.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/DatabaseAnalyticsModule.java
BorisTkachenko: Will address in the next PR

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/DatabaseAnalyticsReadOnlyFreeFormSqlConfig.java
BorisTkachenko: Will address in the next PR

apps/opik-backend/Dockerfile
BorisTkachenko: Yes, this is for local installations.

apps/opik-backend/provision_agent_insights_readonly_user.sh
BorisTkachenko: Generally, it's not relevant for local installations for our customers, since they will not be using Ollie and this feature will be always disabled. Later, if we would introduce any features for local installations that require this arbitrary SQL user; then we could revisit it.

PR #7037, opened 2026-06-11
[OPIK-6873] [BE] Add model price overrides

jverre:
Adds model price override entries to model_prices_overrides.json.


apps/opik-backend/src/main/java/com/comet/opik/api/spend/SpendLane.java
```diff
@@ -1,39 +1,62 @@
...
-    public boolean hasBreakdown() {
-        return hasBreakdown;
+    /** Fixed JSON path expression for the lane's billed total. */
+    public String getTokenExpression() {
+        return getTierExpression("total");
+    }
+
+    /** Fixed JSON path expression for one of the lane's tier columns. */
+    public String getTierExpression(String column) {
+        return "JSONExtractInt(metadata, 'cc', 'billing', 'lanes', '%s', '%s')".formatted(key, column);
```
thiagohora: Please move the SQL logic to the query builder, as we did in the previous PR.

apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendQueryBuilder.java
```diff
@@ -284,61 +250,55 @@ int breakdownItemLimit() {
...
     String summaryCountsSql() {
         return COUNTS_SUMMARY;
     }
 
     String compositionTokensSql() {
-        String selectList = TOKEN_EXPRESSIONS.entrySet().stream()
-                .map(entry -> "SUM(%s) AS %s".formatted(entry.getValue(), entry.getKey().getKey()))
+        String selectList = Arrays.stream(SpendLane.values())
+                .map(lane -> TIER_COLUMNS.stream()
+                        .map(col -> "SUM(%s) AS %s_%s".formatted(lane.getTierExpression(col), lane.getKey(), col))
```
thiagohora: ⚠️ **Performance regression vs the version merged last round.** This builds `SUM(JSONExtractInt(metadata,'cc','billing','lanes',<lane>,<col>))` for every lane × tier column — **10 lanes × 5 tiers = 50 separate full-`metadata` JSON parses per row** (each `JSONExtractInt` re-parses the whole blob; ClickHouse doesn't dedupe distinct paths). The previous PR had replaced the multi-extract with a single typed `JSONExtract(...,'Tuple(...)')` (one parse); that optimization is reverted here.

Measured on 1M traces (local container): **67 s CPU / 6.6 s wall**, vs **2.35 s (~28×)** for the single-typed-extract below.

**Fix — extract the `lanes` object once, then sum the tuple fields.** This reproduces the full expected result (all five tier columns per lane **and** the `models` array) — I diffed it against the current 50-extract query on `cc.billing`-shaped rows and the output is **byte-for-byte identical** (all 50 sums + `models`):
```sql
String compositionTokensSql() {
    String tupleType = Arrays.stream(SpendLane.values())
        .map(l -> "%s Tuple(total Int64, input Int64, cache_read Int64, cache_creation Int64, output Int64)".formatted(l.getKey()))
        .collect(Collectors.joining(", ", "Tuple(", ")"));
    String sums = Arrays.stream(SpendLane.values())
        .flatMap(l -> TIER_COLUMNS.stream()
            .map(col -> "SUM(lanes.%s.%s) AS %s_%s".formatted(l.getKey(), col, l.getKey(), col)))
        .collect(Collectors.joining(",\n    "));
    // models stays a normal extract (1 path, negligible)
    sums += ",\n    groupUniqArrayIf(model, model != '') AS models";
    return COMPOSITION_TOKENS_TUPLE.formatted(sums, tupleType, RANGE, USER_UUID_EXPR);
}
```
with the rendered shape:
```sql
SELECT <sums...>, groupUniqArrayIf(model, model != '') AS models
FROM (
  SELECT JSONExtract(metadata, 'cc', 'billing', 'lanes', '<tupleType>') AS lanes,
         JSONExtractString(metadata, 'cc', 'billing', 'model') AS model
  FROM traces final
  WHERE workspace_id = :workspace_id AND project_id = :project_id AND <RANGE>
        <if(user_uuid)> AND <USER_UUID_EXPR> = :user_uuid <endif>
)
```
The mapper aliases (`<lane>_total`, `<lane>_input`, …) are unchanged. Pruning still works (traces + id-range window), so at a ~16k-row window it's ~1 s — but it scales linearly with window size, so this is a pure regression vs the merged code.

apps/opik-backend/src/main/java/com/comet/opik/api/spend/SpendLane.java
thiagohora: This re-introduces **SQL responsibility into the `api` layer** — `getTierExpression`/`getTokenExpression` return a ClickHouse `JSONExtractInt(metadata, 'cc','billing','lanes', …)` fragment from an `api/spend` enum. Same in `OutputLane`: `namePredicate` (`startsWith(name,'built_in_tools/')`) and `labelExpression` (`substring(name, length('built_in_tools/') + 1)`) are ClickHouse expressions.

The prior round deliberately moved these into `AiSpendQueryBuilder` (the `domain` layer) so the API enums stay free of persistence/SQL concerns — this regresses that. Please move the JSON-path / predicate / label SQL back into `AiSpendQueryBuilder` (e.g. an `EnumMap<SpendLane,…>` / `EnumMap<OutputLane,…>`) and keep these enums to API metadata only (key / label / description / hasBreakdown).

thiagohora (commented):
Re-evaluated the token-native rework (#7037) against the same 1M-trace synthetic dataset. Two things to flag (details inline):

1. **Perf regression — `compositionTokensSql` now does 50 separate `JSONExtractInt` parses per row** (10 lanes × 5 tier columns), reverting the single-typed-`JSONExtract` that landed last round: **67 s CPU/1M vs 2.35 s (~28×)**. Fix is one typed `JSONExtract(...,'lanes','Tuple(...)')`.
2. **SQL back in the `api` layer** — `SpendLane`/`OutputLane` again carry ClickHouse expression strings; move them to `AiSpendQueryBuilder`.

Otherwise the move to `cc.billing` looks fine perf-wise: the `id`/`trace_id`-range pruning is preserved on every query (and the cost/output queries are now traces-side, so they prune too — no full-table spans scans remain). So the regression is isolated to the composition-tokens JSON fan-out.

apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendService.java
```diff
@@ -141,19 +139,19 @@ private SpendRecommendationsResponse buildRecommendations(SpendCompositionRespon
                             .title(rule.title())
                             .body(rule.body())
                             .impact(rule.impact())
-                            .estSaving(estimate(totalCost, laneTokens, grandTotal, rule.factor()))
+                            .estSavingTokens(Math.round(laneTokens * rule.factor()))
```
thiagohora: Better to be clear and avoid abbreviations

```suggestion
                            .estimatedSavingTokens(Math.round(laneTokens * rule.factor()))
```

apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendDAO.java
```diff
@@ -89,51 +95,58 @@ public Mono<List<WorkspaceMetricsSummaryResponse.Result>> getSummary(@NonNull Sp
...
     }
 
     @Override
     public Mono<SpendBreakdownResponse> getOutputBreakdown(@NonNull SpendMetricRequest request,
             @NonNull OutputLane lane) {
         Optional<String> userUuid = resolveUserUuid(request);
         return runBreakdown(queryBuilder.outputBreakdownSql(lane), "ai_spend_output_breakdown", lane.getKey(),
-                lane.getLabel(), lane.getBreakdownSubtitle(), request, userUuid);
+                lane.getLabel(), null, request, userUuid);
     }
```
thiagohora, line 148: This PR replaces `lane.getBreakdownSubtitle()` with `null` (here and in `getBreakdown`), so the breakdown `subtitle` that used to carry text is now always null. Keeping the `subtitle` field/param is fine — a nullable field you may populate later is reasonable, no need to drop it (disregard my earlier "dead field / remove it" suggestion).

The only thing worth confirming: was dropping the descriptive text intentional? If you still want it, `SpendLane` now exposes `description`, which looks like the natural source for the input-lane subtitle (`getBreakdown`); `OutputLane` has none, so `null` there is fine. Non-blocking either way.

apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendQueryBuilder.java
```diff
@@ -284,61 +250,55 @@ int breakdownItemLimit() {
...
+                        .collect(Collectors.joining(",\n    ")))
                 .collect(Collectors.joining(",\n    "));
+        // The FE prices the tier columns; ship the distinct models too.
+        selectList += ",\n    groupUniqArrayIf(JSONExtractString(metadata, 'cc', 'billing', 'model'),"
+                + " JSONExtractString(metadata, 'cc', 'billing', 'model') != '') AS models";
         return COMPOSITION_TOKENS.formatted(selectList, RANGE, USER_UUID_EXPR);
```
thiagohora: Favor StringTemplate variable-binding over `String.formatted` for assembling the SQL. Across this class the queries are spliced together positionally — `COMPOSITION_TOKENS.formatted(selectList, RANGE, USER_UUID_EXPR)`, `USERS.formatted(...)`, `BREAKDOWN.formatted(...)`, `COMPOSITION_OUTPUT.formatted(laneExpr, ...)`, the `TIER_SUMMARY`/`COUNTS_SUMMARY` `.formatted(...)`, etc. — which scatters the query shape across format args and makes the rendered SQL hard to read.

The established pattern (e.g. `TraceDAO`/`SpanDAO`) keeps the full query in an ST template constant with **named placeholders** and binds the dynamically-built fragments as variables:
```java
// template:  ... WHERE ... AND <range> <if(user_uuid)> AND <user_filter> <endif> ... <select_list> ...
var st = getSTWithLogComment(QUERY, queryName, workspaceId, userName, details);
st.add("range", RANGE);
st.add("select_list", selectList);   // built however — Collectors.joining is fine, see SortingQueryBuilder
```
exactly like `template.add("sort_fields", …)` / `add("filters", …)` with `<sort_fields>`/`<filters>` in `TraceDAO`. Ideally the builder just produces the dynamic conditions/fragments and binds them to the template via variables, rather than `String.formatted`.

(Note: building the fragment itself with `Collectors.joining` stays fine — the ask is the *injection* mechanism. Bonus: fragments bound via `.add` don't need the `\\<` escaping the spliced predicates currently carry, since added attribute values aren't re-parsed as template text.)

apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendDAO.java
```diff
@@ -155,8 +171,8 @@ private Mono<SpendBreakdownResponse> runBreakdown(String sql, String queryName,
     public Mono<SpendUserPage> getUsers(@NonNull SpendMetricRequest request,
             @NonNull List<SortingField> sortingFields, String name, int page, int size) {
         String orderBy = Optional.ofNullable(sortingQueryBuilder.toOrderBySql(sortingFields))
-                .map(clause -> clause + ", total_estimated_cost DESC")
-                .orElse("total_estimated_cost DESC");
+                .map(clause -> clause + ", total_tokens DESC")
+                .orElse("total_tokens DESC");
```
thiagohora: Two things here:

1. **This belongs in `AiSpendQueryBuilder`, not the DAO.** Building the ORDER BY (sort clause + default tiebreaker) is SQL construction — the DAO should only execute/bind. Pass `sortingFields` (or the `SortingQueryBuilder`) to the builder and let it own the order-by.
2. **Don't assemble it by string concatenation** (`clause + ", total_tokens DESC"` / `.orElse("total_tokens DESC")`) then splice via `usersSql(orderBy)` → `USERS.formatted(..., orderBy, ...)`. Keep the tiebreaker in the template and bind the optional sort fragment as an ST variable, like `TraceDAO`/`SpanDAO`:
```sql
ORDER BY <if(sort_fields)> <sort_fields>, <endif> total_tokens DESC
```
```java
Optional.ofNullable(sortingQueryBuilder.toOrderBySql(sortingFields))
        .ifPresent(clause -> st.add("sort_fields", clause));
```
This drops the Java concatenation and the duplicated `total_tokens DESC` literal, and keeps SQL-building responsibility in one place.

apps/opik-backend/src/main/java/com/comet/opik/api/spend/SpendLane.java
```diff
@@ -1,39 +1,56 @@
...
-    SKILLS_AVAILABLE("skills_available", "Skills available", Side.INPUT, true, "Menu tokens per available skill"),
-    SKILLS_LOADED("skills_loaded", "Skills loaded", Side.INPUT, true, "Tokens per loaded skill body"),
-    TOOLS("tools", "Tools schema", Side.INPUT, true, "Schema overhead per tool server"),
-    MEMORY("memory", "Memory", Side.INPUT, true, "Tokens per memory file"),
-    FILE_ATTACHMENTS("file_attachments", "File attachments", Side.INPUT, true, "Tokens by content type");
+    USER_PROMPTS("user_prompts", "User prompts", "prompt",
+            "Tokens billed for the text your team typed, re-billed on every turn after it."),
+    FILE_ATTACHMENTS("file_attachments", "File attachments", "file",
+            "Tokens billed for attached files, resent in every request after they're attached."),
+    BUILT_IN_TOOLS("built_in_tools", "Built-in tools", "call",
```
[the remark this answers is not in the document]
thiagohora: Hi @jverre,  Just check that it will not cause any issues

thiagohora (dismissed):
The backend part LGTM!

andriidudar (approved):
For the sake of time, the FE review was bypassed. The changes are well isolated within a feature that has not yet been released.

PR #7049, opened 2026-06-12
[OPIK-6921] [BE] derive bulk project from the experiment or dataset

thiagohora:
## Details

Follow-up to OPIK-6810 (#7021). `experiment_items_bulk` no longer requires `project_name` to be repeated when it can be derived. The project for traces that don't carry their own project is resolved as:

1. the request `project_name` (when provided);
2. else the existing **experiment's** project (when it exists and has one);
3. else the existing **dataset's** project (when it exists and has one);
4. else the **default project** (deprecated; `X-Opik-Deprecation` header).

Derivation only **reads** existing entities — it never creates an experiment or dataset just to obtain a project. The experiment (and dataset, when created) are created in the resolved project so the experiment, dataset, traces and spans all stay in one project.

Backward compatible:
- When `project_name` **is** provided it is validated against the existing experiment's and dataset's projects — **409** on mismatch (as in #7021).
- When `project_name` is **omitted**, the project is derived best-effort, so a project mismatch is only possible (and tolerated) in that case.

The resource emits `X-Opik-Deprecation` for both fallbacks: the default-project fallback and the workspace-wide resolution fallback (the existing `RequestContext` mechanism).

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- OPIK-6921

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Scope: backend implementation, tests, PR description
- Human verification: tests run locally (see Testing)

## Testing

`mvn test -Dtest='ExperimentsResourceTest$BulkUpload'` — 24/24 green (Testcontainers: ClickHouse/MySQL/Redis, local). `mvn spotless:check` clean; `mvn compile` / `test-compile` green.

Scenarios covered (new/updated):
- no `project_name` + existing experiment with a project → traces use the experiment's project; no deprecation header.
- no `project_name` + no existing experiment + dataset has a project → experiment created and traces land in the dataset's project; no header.
- `project_name` provided + experiment/dataset don't exist → both created in the request project.
- `project_name` provided but conflicts with the existing experiment's project → 409.
- `project_name` provided but conflicts with the existing dataset's project → 409.
- fallbacks (no `project_name`, nothing derivable) → default project + `X-Opik-Deprecation` header — preserved.

## Documentation

The REST API guide already documents `project_name`; a follow-up note can clarify it's optional when the experiment/dataset already has a project (tracked in OPIK-6921).

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/ExperimentsResource.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/ExperimentsResource.java
@@ -80,7 +80,6 @@ import jakarta.ws.rs.core.UriInfo;
 import lombok.NonNull;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
-import org.apache.commons.lang3.StringUtils;
 import org.glassfish.jersey.server.ChunkedOutput;
 
 import java.util.Collections;
@@ -520,9 +519,6 @@ public class ExperimentsResource {
     public Response experimentItemsBulk(
             @RequestBody(content = @Content(schema = @Schema(implementation = ExperimentItemBulkUpload.class))) @NotNull @Valid @JsonView(ExperimentItemBulkUpload.View.ExperimentItemBulkWriteView.class) ExperimentItemBulkUpload request) {
 
-        String workspaceId = requestContext.get().getWorkspaceId();
-        String userName = requestContext.get().getUserName();
-
         log.info("Recording experiment items in bulk, count '{}', experimentId '{}'", request.items().size(),
                 request.experimentId());
 
@@ -542,27 +538,33 @@ public class ExperimentsResource {
                 .projectName(request.projectName())
                 .build();
 
-        experimentItemBulkIngestionService.ingest(experiment, request.projectName(), items)
-                .contextWrite(ctx -> ctx.put(RequestContext.USER_NAME, userName)
-                        .put(RequestContext.WORKSPACE_ID, workspaceId))
+        // The service resolves the project (explicit project_name, else derived from the existing experiment
+        // or dataset, else the default project) and returns whether the deprecated default-project fallback
+        // was used for any item.
+        boolean usedDefaultProjectFallback = Boolean.TRUE.equals(experimentItemBulkIngestionService
+                .ingest(experiment, request.projectName(), items)
+                .contextWrite(ctx -> setRequestContext(ctx, requestContext))
                 .retryWhen(RetryUtils.handleConnectionError())
-                .block();
+                .block());
 
         log.info("Recorded experiment items in bulk, count '{}', experimentId '{}'", request.items().size(),
                 request.experimentId());
 
         Response.ResponseBuilder responseBuilder = Response.noContent();
-        // Warn only when a fallback to the default project actually happened: no request-level project_name AND
-        // at least one item without its own project (no trace, or a trace with a blank project_name).
-        boolean usedDefaultProjectFallback = StringUtils.isBlank(request.projectName())
-                && items.stream().anyMatch(
-                        item -> item.trace() == null || StringUtils.isBlank(item.trace().projectName()));
+
+        // Two deprecation fallbacks can surface on this endpoint, both via the X-Opik-Deprecation header:
+        // 1) the project fallback — traces without a project were placed in the default project;
         if (usedDefaultProjectFallback) {
-            // Nudge clients to set project_name so experiments stay in their intended project.
             responseBuilder.header(RequestContext.WORKSPACE_FALLBACK_HEADER,
-                    ("project_name was not provided; traces without a project were placed in the default project "
-                            + "'%s'. This fallback is deprecated, please provide project_name.")
+                    ("project_name could not be resolved; traces without a project were placed in the default "
+                            + "project '%s'. This fallback is deprecated, please provide project_name.")
                             .formatted(ProjectService.DEFAULT_PROJECT));
+        }
+        // 2) the workspace fallback — an entity (e.g. the dataset) was resolved via workspace-wide search,
+        //    set on the request context during resolution (same mechanism used by other resources).
+        String workspaceFallbackMessage = requestContext.get().getWorkspaceFallbackMessage();
+        if (workspaceFallbackMessage != null) {
+            responseBuilder.header(RequestContext.WORKSPACE_FALLBACK_HEADER, workspaceFallbackMessage);
         }
 
         return responseBuilder.build();
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/ExperimentsResource.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/ExperimentsResource.java
@@ -539,32 +539,33 @@ public class ExperimentsResource {
                 .build();
 
         // The service resolves the project (explicit project_name, else derived from the existing experiment
-        // or dataset, else the default project) and returns whether the deprecated default-project fallback
-        // was used for any item.
-        boolean usedDefaultProjectFallback = Boolean.TRUE.equals(experimentItemBulkIngestionService
+        // or dataset, else the default project) and reports which deprecated fallback (if any) it used.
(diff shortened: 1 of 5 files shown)


CometActions:

### Access Information
- **URL:** https://pr-7049.dev.comet.com
- **Namespace:** pr-7049
- **Version:** 2.0.63-7049-merge-2270
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7049%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

miguelgrc (commented):
Looks good! Just 4 small comments

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/ExperimentsResourceTest.java
```diff
@@ -6974,6 +6974,131 @@ void experimentItemsBulk__whenExistingExperimentAndMismatchedProjectName__thenRe
...
+                    .items(List.of(ExperimentItemBulkRecord.builder()
+                            .datasetItemId(podamFactory.manufacturePojo(UUID.class))
+                            .evaluateTaskResult(JsonUtils.readTree("{\"answer\": \"42\"}"))
+                            .build()))
+                    .build();
+
+            // when / then
+            try (var response = experimentResourceClient.callExperimentItemBulkUpload(bulkUploadRequest, API_KEY,
+                    TEST_WORKSPACE)) {
+                assertThat(response.getStatus()).isEqualTo(HttpStatus.SC_CONFLICT);
```
miguelgrc, line 7003: **[low] Conflict test asserts only HTTP status, not the error body**

experimentItemsBulk__whenProjectNameMismatchesDatasetProject__thenReturnConflict asserts response.getStatus() == SC_CONFLICT but never reads or asserts the response body (error message). Other conflict tests in the same class (e.g. line 7518-7523 for dataset mismatch, and the workspace-ownership tests) do assert the error message. A status-only assertion passes even when the server returns the right code for the wrong reason.

💡 Read the response entity and assert the error message matches the expected conflict reason, e.g.:
var errorMessage = response.readEntity(io.dropwizard.jersey.errors.ErrorMessage.class);
assertThat(errorMessage.getMessage()).contains("project");

_rule: `backend-testing-integration-tests-must-verify-behavior-through-public-apis-u`_

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/ExperimentsResourceTest.java
```diff
@@ -6974,6 +6974,131 @@ void experimentItemsBulk__whenExistingExperimentAndMismatchedProjectName__thenRe
...
+                    TEST_WORKSPACE)) {
+                // then — derived from the experiment, no fallback
+                assertThat(response.getStatus()).isEqualTo(HttpStatus.SC_NO_CONTENT);
+                assertThat(response.getHeaderString(RequestContext.WORKSPACE_FALLBACK_HEADER)).isNull();
+            }
+
+            List<ExperimentItem> actualExperimentItems = experimentResourceClient.getExperimentItems(experimentName,
+                    API_KEY, TEST_WORKSPACE);
+            // every auto-created trace lands in the experiment's project
+            for (ExperimentItem item : actualExperimentItems) {
```
miguelgrc, line 7053: **[low] Missing size assertion before iterating over experiment items**

In experimentItemsBulk__whenNoProjectNameAndExistingExperimentHasProject__thenUseExperimentProject, the test loops over actualExperimentItems to assert each trace's projectId, but never asserts the list size. If getExperimentItems returns an empty list (e.g. due to a timing issue or bug), the for-loop is a no-op and the test silently passes. The sibling test at line 7089 (experimentItemsBulk__whenNoProjectNameAndDatasetHasProject__thenUseDatasetProject) correctly asserts assertThat(actualExperimentItems).hasSize(1) before checking individual items.

💡 Add a size assertion before the loop, e.g.:
assertThat(actualExperimentItems).hasSizeGreaterThanOrEqualTo(2);
for (ExperimentItem item : actualExperimentItems) { ... }

apps/opik-documentation/documentation/fern/docs/evaluation/log_experiments_with_rest_api.mdx
```diff
@@ -265,12 +265,11 @@ print(f"Prepared {len(evaluation_items)} evaluation items")
 Use the bulk endpoint to efficiently log multiple evaluation results at once.
 
 <Note>
-  Pass `project_name` on the bulk request so the experiment, its dataset (when it
-  doesn't exist yet), and any traces auto-created from `evaluate_task_result` are
-  placed in that project. If you omit it, those traces fall back to the
-  `Default Project` — this fallback is deprecated and the response includes an
-  `X-Opik-Deprecation` header. The value should match the `project_name` you used
-  when creating the experiment.
+  `project_name` sets the project for the experiment, dataset, and auto-created traces.
```
miguelgrc, line 268: **[low] Dataset-derived project fallback is also deprecated but note implies it is not**

The note groups experiment and dataset derivation together as the non-deprecated path ("the project is derived from the existing experiment or dataset"), with only the Default Project fallback described as deprecated. However, the backend code (ExperimentItemBulkIngestionService.resolveBulkProject) returns ProjectFallback.DATASET when deriving from the dataset, and ExperimentsResource emits the X-Opik-Deprecation header for BOTH the DATASET and DEFAULT fallbacks. This means deriving from the dataset is also deprecated, and users relying on it will receive a deprecation header -- contradicting the note's implication that only the Default Project path is deprecated.

💡 Separate the experiment and dataset derivation in the note. For example: "When omitted, the project is derived from the existing experiment (no deprecation). If the experiment has no project, it falls back to the dataset's project (deprecated, emits X-Opik-Deprecation header). If neither applies, it falls back to the deprecated Default Project (also emits the header). Provide project_name explicitly to avoid all fallbacks."

_general finding_

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/ExperimentsResource.java
```diff
@@ -542,27 +538,34 @@ public Response experimentItemsBulk(
...
-                        item -> item.trace() == null || StringUtils.isBlank(item.trace().projectName()));
-        if (usedDefaultProjectFallback) {
-            // Nudge clients to set project_name so experiments stay in their intended project.
-            responseBuilder.header(RequestContext.WORKSPACE_FALLBACK_HEADER,
-                    ("project_name was not provided; traces without a project were placed in the default project "
-                            + "'%s'. This fallback is deprecated, please provide project_name.")
+
+        // Surface the deprecated implicit-fallback as the X-Opik-Deprecation header (on the request thread, so
+        // the request-scoped fallback message can be set/read safely — same mechanism as other resources).
+        switch (fallback) {
```
miguelgrc, line 556: **[low] Switch on ProjectFallback lacks a default branch**

The switch statement covers all three current enum values (NONE, DATASET, DEFAULT) but has no default branch. If a new ProjectFallback value is added in the future, this switch will silently do nothing instead of failing loudly. While Java enhanced switch expressions would catch this at compile time, this is a switch statement, so it compiles silently.

💡 Add a default branch that throws an IllegalStateException or BadRequestException: `default -> throw new IllegalStateException("Unexpected fallback: " + fallback);`

_rule: `backend-error-handling-use-exhaustive-switch-statements-with-a-default-exception-cl`_

miguelgrc (approved):

PR #7057, opened 2026-06-12
[NA] [BE][FE] Cost computation refinements

jverre:
Cost computation refinements.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendQueryBuilder.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendQueryBuilder.java
@@ -92,17 +92,18 @@ class AiSpendQueryBuilder {
                     TIER_COLUMNS.stream().map(col -> col + " Int64").collect(Collectors.joining(", "))))
             .collect(Collectors.joining(", ", "Tuple(", ")"));
 
-    // The FE prices the tier columns; ship the distinct models too.
+    // The FE prices the tier columns; group by model so each model's tokens
+    // carry its own rate.
     private static final String COMPOSITION_SELECT_LIST = Arrays.stream(SpendLane.values())
             .flatMap(lane -> TIER_COLUMNS.stream()
                     .map(col -> "SUM(lanes.%1$s.%2$s) AS %1$s_%2$s".formatted(lane.getKey(), col)))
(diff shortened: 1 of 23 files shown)


apps/opik-backend/src/main/java/com/comet/opik/domain/AiSpendQueryBuilder.java
```diff
@@ -277,20 +299,22 @@ trace_agg AS (
...
-                        sum(tok_output) AS output_tokens,
-                        sum(tok_input + tok_cache_read + tok_cache_creation + tok_output) AS total_tokens
+                        sum(tok_input + tok_cache_read + tok_cache_creation + tok_output) AS total_tokens,
+                        -- Per-model tier arrays via sumMap: every sumMap sorts on
+                        -- the same model key set, so the value arrays align by index.
+                        (sumMap([model], [tok_input])).1 AS models,
+                        (sumMap([model], [tok_input])).2 AS model_input,
+                        (sumMap([model], [tok_cache_read])).2 AS model_cache_read,
+                        (sumMap([model], [tok_cache_creation])).2 AS model_cache_creation,
+                        (sumMap([model], [tok_output])).2 AS model_output
```
thiagohora, line 309: Minor: these five `(sumMap([model],[tok_X])).2` calls can collapse into a **single multi-value `sumMap`** — it builds the per-model map once instead of 5×, and the array alignment becomes guaranteed-by-construction (one sorted key set) instead of relying on "every sumMap sorts on the same model key set":
```sql
sumMap([model], [tok_input], [tok_cache_read], [tok_cache_creation], [tok_output]) AS sm
-- sm.1 = models, sm.2 = input, sm.3 = cache_read, sm.4 = cache_creation, sm.5 = output
```
I verified it's equivalent on 2-model data with distinct per-tier values — byte-identical arrays both ways (`models=['modelA','modelB']`, `input=[500,2500]`, `cache_read=[1000,3000]`, `cache_creation=[1500,3500]`, `output=[2000,4000]`). Perf is negligible either way (model cardinality is tiny); the win is dropping the cross-call ordering assumption. Non-blocking.

andriidudar (approved):
Approved to avoid blocking progress and for the sake of time. This area will be revisited and revised soon.

PR #7075, opened 2026-06-12
fix(examples): guard against IndexError on empty LLM choices list

jverre:
> ♻️ Re-opened on behalf of @qizwiz. All commits are authored by @qizwiz. Apologies for the inconvenience. Original PR for reference: #6700.

---

## Summary

Five example scripts call `response.choices[0].message.content` without checking whether the API returned any choices. This raises a bare `IndexError` when the LLM:

- applies a content-policy filter and returns an empty choices list
- returns a truncated streaming response before the first choice arrives
- hits a rate-limit and returns a non-standard reply

### Changes

Inline guard at each of the five call sites:

| File | Fix |
|------|-----|
| `generate_seo_metadata.py` | `if not response.choices: return None` before accessing `choices[0]` |
| `FunctionDecorators.py` | `if not response.choices: return None` before accessing `choices[0]` |
| `DeepSeek.py` | `if response.choices: print(...)` wraps the print |
| `OpenAI.py` | `if response.choices: print(...)` wraps the print |
| `OpenRouter.py` | `if response.choices: print(...)` wraps the print |

No new files, no monkey-patching — just minimal defensive checks at each site.

### DIFF
--- a/apps/opik-documentation/documentation/scripts/generate_seo_metadata.py
+++ b/apps/opik-documentation/documentation/scripts/generate_seo_metadata.py
@@ -186,6 +186,7 @@ Return ONLY the JSON object, no explanation."""
         )
         
         if not getattr(response, "choices", None) or response.choices[0].message is None:
+            print("  Warning: LLM returned no choices or null message; skipping")
             return None
         result_text = response.choices[0].message.content.strip()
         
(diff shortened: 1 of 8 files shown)


aadereiko:
The onboarding snippets are intentionally minimal. They exist to show the few lines needed to get tracing working, so we keep error handling out of them even though it'd be right in production code. Closing this one for now, but we can return to it if raised again. Thanks for the attention to detail!

PR #7077, opened 2026-06-12
fix(sdk): add cyclic reference and max depth protection to flatten_dict

jverre:
> ♻️ Re-opened on behalf of @zhoufengen. All commits are authored by @zhoufengen. Apologies for the inconvenience. Original PR for reference: #6545.

---

## Summary

`flatten_dict` in `dict_utils.py` recursed without any guard against cyclic references or deeply nested structures. The function's own docstring acknowledged this limitation: *"The current implementation does not have max depth restrictions or cyclic references handling!"*. This could cause `RecursionError` crashes when processing user-provided metadata containing circular references.

## Changes

- Add a `_seen` set parameter to detect cyclic references (logs a warning and returns empty dict for the cyclic branch)
- Add a `_depth` counter with `_MAX_FLATTEN_DEPTH = 20` limit (logs a warning and returns the remaining value as-is)
- Both parameters are internal (prefixed with `_`) and default to safe values, so the public API is unchanged
- Pattern is consistent with the cycle detection already used in `jsonable_encoder.py`

## Testing

- The public API signature is unchanged: `flatten_dict(d, parent_key, delim)` works identically
- Normal nested dicts (depth < 20) produce the same output as before
- Cyclic references now log a warning instead of crashing with `RecursionError`
- Deeply nested dicts (depth >= 20) now log a warning and stop flattening instead of hitting Python's recursion limit

## Related Issues

None (discovered during code review; the docstring itself documented the limitation)

---

> This PR was authored with AI assistance. All changes have been reviewed for correctness.

### DIFF
--- a/sdks/python/src/opik/dict_utils.py
+++ b/sdks/python/src/opik/dict_utils.py
@@ -1,27 +1,55 @@
 import copy
 import logging
-from typing import Any, Dict, Optional, List, Tuple, TypeVar, Type
+from typing import Any, Dict, Optional, List, Set, Tuple, TypeVar, Type
 
 from . import logging_messages
 
 LOGGER = logging.getLogger(__name__)
 
+_MAX_FLATTEN_DEPTH = 20
+
 
 def flatten_dict(
-    d: Dict[str, Any], parent_key: str, delim: str = "."
+    d: Dict[str, Any],
+    parent_key: str,
+    delim: str = ".",
+    _depth: int = 0,
+    _seen: Optional[Set[int]] = None,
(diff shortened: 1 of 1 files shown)


aadereiko:
This implementation guards a non-existing case:

I checked every place `flatten_dict` is used on main. It's used in only two areas:
- CLI export (cli/exports/utils.py). This data comes from the API as JSON, and JSON can't contain a loop.
- LLM token usage (llm_usage/base_original_provider_usage.py). These are small dictionaries of numbers, like {"prompt_tokens": 10}.

PR #7078, opened 2026-06-12
fix(sdk): add max retry limit to rate-limited API calls

jverre:
> ♻️ Re-opened on behalf of @zhoufengen. All commits are authored by @zhoufengen. Apologies for the inconvenience. Original PR for reference: #6542.

---

## Summary

The `ensure_rest_api_call_respecting_rate_limit` function in the Python SDK retries indefinitely on persistent HTTP 429 responses, which can cause callers to hang forever when account quota is exhausted or rate limits are misconfigured.

## Changes

- Replace `while True` loop with a bounded `for` loop capped at `MAX_RATE_LIMIT_RETRIES = 10` attempts
- Re-raise the `ApiError` after exhausting all retries
- Add attempt count to log messages for better observability

## Testing

- The change is backward-compatible: normal rate-limited calls that succeed within 10 retries behave identically
- The retry decorator (`retry_decorator.py`) already caps at 3 attempts for other errors; this brings rate limit handling in line with that pattern
- Edge case: after 10 consecutive 429 responses, the original `ApiError` is raised to the caller

## Related Issues

None (discovered during code review)

---

> This PR was authored with AI assistance. All changes have been reviewed for correctness.

### DIFF
--- a/sdks/python/src/opik/api_objects/rest_helpers.py
+++ b/sdks/python/src/opik/api_objects/rest_helpers.py
@@ -8,6 +8,8 @@ from ..rate_limit import rate_limit
 
 LOGGER = logging.getLogger(__name__)
 
+MAX_RATE_LIMIT_RETRIES = 10
+
 
 def ensure_rest_api_call_respecting_rate_limit(
     rest_callable: Callable[[], Any],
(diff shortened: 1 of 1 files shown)


aadereiko:
Thanks! The "retry forever on 429" problem was addressed differently in #8397 (https://github.com/comet-ml/opik/pull/8397). The loop in rest_helpers.py stays unbounded on purpose, but callers can now pass a stop_event to cancel it. A hard cap of 10 retries would make background uploads drop data during a long rate limit, so I'm closing this one in favour of that approach.

PR #7082, opened 2026-06-13
[OPIK-6857] [BE] feat: add agent insights report results endpoints

petrotiurin:
## Details

Backend implementation of Agent Insights steps 7 & 8: endpoints that store report results produced by the Ollie sub-agent and serve them (with time-window aggregation) to the frontend.

**What was added:**

- Two new MySQL tables (`agent_insights_issues`, `agent_insights_issues_details`) with Liquibase migrations. Per-day detail rows upsert on the unique key `(workspace_id, project_id, issue_id, report_day)`, so re-running the same `report_day` replaces that day's metrics instead of duplicating them. Issues themselves are keyed by their UUID `id`: continuity across runs is idempotent **only when the caller resupplies the existing issue ids** (obtained from the list endpoint) — issues reported without an `id` are created fresh each run. Issue `status` is owned by the user and never touched by the reporting agent.
- Endpoint `POST /v1/private/agent-insights/issues` — transactionally upserts issues + per-day details in one request (called by Ollie). Issue status is never modified by this endpoint.
- Private endpoints for the frontend:
  - `GET /v1/private/agent-insights/issues` — paginated list with time-window aggregation (`SUM(count)`, `SUM(total_count)`, first/last seen, days reported) and optional `status` filter
  - `GET /v1/private/agent-insights/issues/{issue_id}` — per-day breakdown within a window for the occurrences-over-time chart
  - `PATCH /v1/private/agent-insights/issues/{issue_id}` — update issue status (`open`/`resolved`/`closed`); the only user-facing write
- Standard 3-layer pattern: Resource → Service → JDBI DAO
- `workspace_id` always derived from auth, never from the request body

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- OPIK-6857

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): claude-sonnet-4-6
- Scope: assisted (branch rename, PR creation, rebase conflict triage)
- Human verification: code review + manual testing

## Testing

- Integration tests in `AgentInsightsResourceTest` covering:
  - Upsert idempotency (same `report_day` re-run with resupplied issue id)
  - Time-window aggregation correctness
  - Optional `status` filter on list endpoint
  - Status transitions (`open` → `resolved` → `closed` → `open`)
  - Pagination
  - 404 on out-of-workspace/project access
  - 400 on invalid status value

Run: `cd apps/opik-backend && mvn test -pl . -Dtest=AgentInsightsResourceTest`

## Documentation

N/A — API spec updated in `opik.yaml` / `openapi.yaml` as part of the Fern-generated SDK changes.


apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsIssueDAO.java
```diff
@@ -0,0 +1,214 @@
...
+
+        @Override
+        public AgentInsightsIssueWithDetails map(ResultSet rs, StatementContext ctx) throws SQLException {
+            return AgentInsightsIssueWithDetails.builder()
+                    .id(UUID.fromString(rs.getString("id")))
+                    .name(rs.getString("name"))
+                    .description(rs.getString("description"))
+                    .cause(rs.getString("cause"))
+                    .suggestedFix(rs.getString("suggested_fix"))
+                    .status(AgentInsightsIssueStatus.fromString(rs.getString("status")))
```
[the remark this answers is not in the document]
petrotiurin: Skipping — the `status` column is declared `ENUM('open','resolved','closed') NOT NULL DEFAULT 'open'`, so the DB can never return a NULL or out-of-domain value for this field. `fromString` therefore cannot throw here in practice; this path is only reachable if the schema constraint is bypassed. The tolerant `AgentInsightsIssueStatusColumnMapper` exists for columns that lack this guarantee. Leaving the direct `fromString` call as-is.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000082_create_agent_insights_issues_tables.sql
```diff
@@ -0,0 +1,54 @@
...
+    suggested_fix   TEXT                               NULL,
+    traces_query    TEXT                               NULL,
+    created_by      VARCHAR(255)                       NOT NULL DEFAULT 'admin',
+    created_at      TIMESTAMP(6)                       NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
+    last_updated_by VARCHAR(255)                       NOT NULL DEFAULT 'admin',
+    last_updated_at TIMESTAMP(6)                       NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6),
+
+    -- Workspace-scoped identity: id is unique per (workspace_id, project_id), not globally. This composite key is
+    -- the upsert target, so a client-supplied id can only ever match a row within its own workspace and project.
+    PRIMARY KEY (workspace_id, project_id, id)
```
BorisTkachenko: Lets better create PRIMARY KEY for id only and INDEX for (workspace_id, project_id). It's the approach used for Opik

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000082_create_agent_insights_issues_tables.sql
```diff
@@ -0,0 +1,54 @@
...
+(
+    id              CHAR(36)                           NOT NULL,
+    workspace_id    VARCHAR(150)                       NOT NULL,
+    project_id      CHAR(36)                           NOT NULL,
+    status          ENUM ('open','resolved','closed')  NOT NULL DEFAULT 'open',
+    name            VARCHAR(255)                       NOT NULL,
+    description     TEXT                               NULL,
+    cause           TEXT                               NULL,
+    suggested_fix   TEXT                               NULL,
+    traces_query    TEXT                               NULL,
```
BorisTkachenko, line 15: Maybe better just `query`? Since it might query other tables too, not just traces, condition may be in spans or feedback scores.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000082_create_agent_insights_issues_tables.sql
```diff
@@ -0,0 +1,54 @@
...
+    created_by      VARCHAR(255) NOT NULL DEFAULT 'admin',
+    created_at      TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
+    last_updated_by VARCHAR(255) NOT NULL DEFAULT 'admin',
+    last_updated_at TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6),
+
+    PRIMARY KEY (id),
+    -- "One record per issue per report_day"; the upsert target for daily report runs
+    UNIQUE KEY agent_insights_issues_details_uk (workspace_id, project_id, issue_id, report_day),
+    -- Supports the time-window list query that filters details by report_day before joining issues
+    INDEX agent_insights_issues_details_report_day_idx (workspace_id, project_id, report_day)
```
BorisTkachenko: Let's just keep the UNIQUE KEY as:
```
UNIQUE KEY agent_insights_issues_details_uk (workspace_id, project_id, report_day, issue_id)
```
and drop the second INDEX `agent_insights_issues_details_report_day_idx` as it will aready be built within agent_insights_issues_details_uk

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/AgentInsightsResource.java
```diff
@@ -0,0 +1,124 @@
...
+            @ApiResponse(responseCode = "200", description = "Issues page", content = @Content(schema = @Schema(implementation = AgentInsightsIssue.AgentInsightsIssuePage.class))),
+            @ApiResponse(responseCode = "400", description = "Bad Request", content = @Content(schema = @Schema(implementation = ErrorMessage.class))),
+            @ApiResponse(responseCode = "401", description = "Unauthorized", content = @Content(schema = @Schema(implementation = ErrorMessage.class)))
+    })
+    public Response findIssues(
+            @QueryParam("project_id") @NotNull UUID projectId,
+            @QueryParam("from_date") LocalDate fromDate,
+            @QueryParam("to_date") LocalDate toDate,
+            @QueryParam("status") AgentInsightsIssueStatus status,
+            @QueryParam("sort_by") AgentInsightsSortBy sortBy,
```
BorisTkachenko: For sorting please reuse the approach used for other entities with *SortingFactory, you can check Traces, Spans, Alerts or almost any other entity.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/AgentInsightsResource.java
```diff
@@ -0,0 +1,124 @@
...
+            @QueryParam("page") @Min(1) @DefaultValue("1") int page,
+            @QueryParam("size") @Min(1) @Max(100) @DefaultValue("10") int size) {
+
+        AgentInsightsIssue.AgentInsightsIssuePage issuesPage = agentInsightsIssueService.findIssues(
```
BorisTkachenko, line 74: Before this add something like: `validateTimeRangeParameters(fromDate, toDate);` for params validation, you can refer to the `TracesResource` for example.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/AgentInsightsResource.java
```diff
@@ -0,0 +1,124 @@
...
+            @ApiResponse(responseCode = "404", description = "Not Found", content = @Content(schema = @Schema(implementation = ErrorMessage.class))),
+            @ApiResponse(responseCode = "401", description = "Unauthorized", content = @Content(schema = @Schema(implementation = ErrorMessage.class)))
+    })
+    public Response getIssueById(
+            @PathParam("issue_id") UUID issueId,
+            @QueryParam("project_id") @NotNull UUID projectId,
+            @QueryParam("from_date") LocalDate fromDate,
+            @QueryParam("to_date") LocalDate toDate) {
+
+        AgentInsightsIssueWithDetails issue = agentInsightsIssueService.getIssue(issueId, projectId, fromDate,
```
BorisTkachenko, line 96: Same validation for time period is missing

apps/opik-backend/src/main/java/com/comet/opik/api/AgentInsightsIssue.java
```diff
@@ -0,0 +1,50 @@
...
+import java.time.Instant;
+import java.time.LocalDate;
+import java.util.List;
+import java.util.UUID;
+
+@Builder(toBuilder = true)
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
+public record AgentInsightsIssue(
+        @Schema(accessMode = Schema.AccessMode.READ_ONLY) UUID id,
```
BorisTkachenko: Not sure we need READ_ONLY in this file here and below. Basically, you use this for response only, so all fields are read-only.

apps/opik-backend/src/main/java/com/comet/opik/api/AgentInsightsIssueStatus.java
```diff
@@ -0,0 +1,29 @@
...
+public enum AgentInsightsIssueStatus implements HasValue {
+    OPEN("open"),
+    RESOLVED("resolved"),
+    CLOSED("closed");
+
+    @JsonValue
+    private final String value;
+
+    // BadRequestException so an invalid query param value yields 400 instead of JAX-RS's default 404
+    public static AgentInsightsIssueStatus fromString(String value) {
```
BorisTkachenko, line 25:
```suggestion
    @JsonCreator
    public static AgentInsightsIssueStatus fromString(String value) {
```

apps/opik-backend/src/main/java/com/comet/opik/api/AgentInsightsIssueWithDetails.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.time.Instant;
+import java.util.List;
+import java.util.UUID;
+
+@Builder(toBuilder = true)
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
+public record AgentInsightsIssueWithDetails(
+        @Schema(accessMode = Schema.AccessMode.READ_ONLY) UUID id,
```
BorisTkachenko: This class is used only for responses, so all fields are read-only. You can remove this: @Schema(accessMode = Schema.AccessMode.READ_ONLY).

Same in other places in this file.

apps/opik-backend/src/main/java/com/comet/opik/api/AgentInsightsSortBy.java
```diff
@@ -0,0 +1,27 @@
...
+import jakarta.ws.rs.BadRequestException;
+import lombok.AccessLevel;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+import java.util.Arrays;
+
+@Getter
+@RequiredArgsConstructor(access = AccessLevel.PRIVATE)
+public enum AgentInsightsSortBy {
```
BorisTkachenko: Not needed, please use approach with `SortingFactory`, similar to other entities.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/AgentInsightsResourceTest.java
```diff
@@ -0,0 +1,891 @@
...
+            assertThat(issueA.total()).isEqualTo(totalCount);
+            assertThat(issueA.usersImpacted()).isEqualTo(impacted);
+            assertThat(issueA.totalUsers()).isEqualTo(totalUsers);
+            assertThat(issueA.firstSeen()).isEqualTo(DAY_1);
+            assertThat(issueA.lastSeen()).isEqualTo(DAY_1);
+            assertThat(issueA.daysReported()).isEqualTo(1);
+            assertThat(issueA.createdBy()).isEqualTo(USER);
+            assertThat(issueA.lastUpdatedBy()).isEqualTo(USER);
+            assertThat(issueA.createdAt()).isNotNull();
+            assertThat(issueA.lastUpdatedAt()).isNotNull();
```
BorisTkachenko: Would be better to assert entire object, not field by field. Examples could be found in existing *ResourceTest.java classes. Also same in other places.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsIssueService.java
```diff
@@ -0,0 +1,210 @@
...
+
+    void reportIssues(AgentInsightsReport report);
+
+    AgentInsightsIssue.AgentInsightsIssuePage findIssues(UUID projectId, LocalDate fromDate, LocalDate toDate,
+            AgentInsightsIssueStatus status, AgentInsightsSortBy sortBy, int page, int size);
+
+    AgentInsightsIssueWithDetails getIssue(UUID issueId, UUID projectId, LocalDate fromDate, LocalDate toDate);
+
+    LocalDate DEFAULT_FROM_DATE = LocalDate.of(1970, 1, 1);
+    LocalDate DEFAULT_TO_DATE = LocalDate.of(9999, 12, 31);
```
BorisTkachenko: Better to use current date as default.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsIssueService.java
```diff
@@ -0,0 +1,210 @@
...
+    // The metadata column is TEXT (65,535 bytes); reject oversized payloads at the boundary as 400 rather than
+    // letting MySQL fail the INSERT with a data-truncation error surfaced as 500.
+    private String serializeMetadata(JsonNode metadata) {
+        if (metadata == null) {
+            return null;
+        }
+        String serialized = JsonUtils.writeValueAsString(metadata);
+        if (serialized.getBytes(StandardCharsets.UTF_8).length > MAX_METADATA_BYTES) {
+            throw new BadRequestException(
+                    "Issue metadata exceeds the %d byte limit".formatted(MAX_METADATA_BYTES));
```
BorisTkachenko: Not sure, if we need to check this size here. Because we do not check the same for the other parameters like:
            String description,
            String cause,
            String suggestedFix,
            String tracesQuery

That should have basically the same limit. So we need either to add the validation with a maximum size for those as well. Or remove it here, which I would do since it is not really likely that we would exceed that limit and usually we do not verify it. In the very rare case ingestion to DB will fail.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsIssueService.java
```diff
@@ -0,0 +1,210 @@
...
+        }
+        return serialized;
+    }
+
+    private DateWindow resolveWindow(LocalDate fromDate, LocalDate toDate) {
```
BorisTkachenko, line 174: As commented above lets verify it in the very beginning, before we execute all the other logic.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsIssueService.java
```diff
@@ -0,0 +1,210 @@
...
+
+    private record DateWindow(LocalDate from, LocalDate to) {
+    }
+
+    // Hardcoded fragments keyed by enum: user input never reaches the SQL text
+    private String toOrderByClause(AgentInsightsSortBy sortBy) {
+        return switch (sortBy) {
+            case LAST_SEEN -> "last_seen DESC, total_occurrences DESC, i.id DESC";
+            case TOTAL_OCCURRENCES -> "total_occurrences DESC, last_seen DESC, i.id DESC";
+        };
```
BorisTkachenko: Should be not needed once other comments are addressed.

BorisTkachenko (commented):
Looks good, left a few comments before merging it.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000082_create_agent_insights_issues_tables.sql
petrotiurin: `query` is generic and it's not obvious what it represents that's why I went with `traces_query`. The idea is to name it to suggest what it returns - all the traces affected. Maybe `affected_traces_query`?

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsIssueService.java
petrotiurin: I'm thinking of an edge case where there's a summer/winter time change and the date rolls back. Not the highest impact, but nevertheless possible.
Any specific concerns with this approach?

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsIssueService.java
petrotiurin: The other parameters are standard strings so I'm happy to follow the general consensus we have elsewhere of not checking their sizes.
However, `metadata` is a special json that's catch-all for any additional fields we might add. The ease of adding new fields there is what concerns me, so an additional check won't hurt in my opinion.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000082_create_agent_insights_issues_tables.sql
petrotiurin: Right, the existing implementation was assuming ids might collide across projects/workspaces. However, looking at it again it's probably safe to assume id is globally unique for our scale.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000082_create_agent_insights_issues_tables.sql
petrotiurin: Makes sense, however if we want an aggeregation per single issue over time, this unique key would let the DB narrow the scan by workspace, project and date range but then DB would have to filter the resulting issues to our given issue. I might be overthinking it, but what's the tradeoff of keeping this extra index?

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000082_create_agent_insights_issues_tables.sql
```diff
@@ -0,0 +1,53 @@
...
+    cause           TEXT                               NULL,
+    suggested_fix   TEXT                               NULL,
+    traces_query    TEXT                               NULL,
+    created_by      VARCHAR(255)                       NOT NULL DEFAULT 'admin',
+    created_at      TIMESTAMP(6)                       NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
+    last_updated_by VARCHAR(255)                       NOT NULL DEFAULT 'admin',
+    last_updated_at TIMESTAMP(6)                       NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6),
+
+    PRIMARY KEY (id),
+    INDEX agent_insights_issues_workspace_project_idx (workspace_id, project_id)
```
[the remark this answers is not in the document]
petrotiurin: As discussed above, uuid collision is no likely.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000082_create_agent_insights_issues_tables.sql
BorisTkachenko: Let's keep `traces_query` if it always returns traces (trace_ids) in the end.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsIssueService.java
BorisTkachenko: Now for other time bound metrics if endTime is not provided, we use current time. So better to be consistent.

apps/opik-backend/src/main/java/com/comet/opik/utils/ValidationUtils.java
```diff
@@ -61,6 +62,14 @@ public static void validateTimeRangeParameters(Instant startTime, Instant endTim
         }
     }
 
+    public static void validateDateRangeParameters(LocalDate fromDate, LocalDate toDate) {
+        // both are optional, but if both are provided, from_date must not be after to_date (equal is a single-day window)
+        if (fromDate != null && toDate != null && fromDate.isAfter(toDate)) {
```
BorisTkachenko, line 68: If we use default as `now` for `toDate`, logic should be changed. In case `toDate==null` , we need to verify that `fromDate` is before.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsIssueService.java
BorisTkachenko: Ok, in this case let's move it out of the service and use annotations on the request object for it to validate during request payload deserialization. You can check `MaxRequestSizeValidator`, it's doing similar thing.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000082_create_agent_insights_issues_tables.sql
BorisTkachenko: Keeping extra index makes write queries slower. If we need the use case that you described we can always add a new index. As I checked current queries it makes sense to keep only one index that you already have `UNIQUE KEY agent_insights_issues_details_uk (workspace_id, project_id, issue_id, report_day)` without switching last two fields and remove `agent_insights_issues_details_report_day_idx`

BorisTkachenko (commented):
Left a few comments.

apps/opik-backend/src/main/java/com/comet/opik/utils/ValidationUtils.java
petrotiurin: Good catch - missed this when changing toDate defaults.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000082_create_agent_insights_issues_tables.sql
petrotiurin: Ok, we'll cross that bridge once we get there - your call. Our scale is probably too small anyway to warrant an extra index. I've removed it and reordered the UK.

BorisTkachenko (approved):
Looks good!

PR #6640, opened 2026-05-08
[OPIK-6261] [BE] [FE] feat: Ollie daily briefing reports

miguelgrc:
## Details
Implements the Ollie Daily Reports feature (Part 2 of the home page redesign). Enables automated daily project briefings generated by Ollie, with on-demand trigger support, customizable schedule time and custom prompt instructions.

### Backend
- **DB migration**: `ollie_reports` and `report_preferences` tables (with `schedule_time`, `custom_prompt` columns)
- **REST endpoints**: generate report, complete callback, list reports, preferences CRUD (partial update via COALESCE)
- **ReportService**: report lifecycle management, project validation, skips report creation when orchestrator URL not configured
- **OrchestratorClient**: fire-and-forget HTTP call to configurable report generation URL, passes `custom_prompt` to Ollie
- **DailyReportJob**: Quartz job every 10 minutes, queries preferences whose `schedule_time` falls in the previous 10-minute window (supports per-project custom schedule times)
- **StaleReportCleanupJob**: marks stale PENDING reports as FAILED every 15 minutes
- **ReportGenerationConfig**: configurable URL (blank disables report generation)
- **RecentActivity**: `TRACE_DAILY` activity type with daily trace counts from ClickHouse (stashed, WIP)

### Frontend
- **DailyBriefing/**: extracted into dedicated folder with DailyBriefingSection, TurnOnDialog, SettingsDialog, ReportPanel
- **DailyBriefingSection**: all UI states (inactive, active empty, running, scheduled with today/tomorrow label, failed, paused)
- **SettingsDialog**: editable schedule time (TimePicker), custom instructions (Textarea), enable/disable toggle, billing link
- **TurnOnDialog**: simplified confirmation with instant run toggle and default schedule time
- **ReportPanel**: markdown rendering via shared MarkdownPreview (with rehype-sanitize)
- **BillingLink plugin**: Comet-specific billing URL component registered via PluginsStore
- **API hooks**: useReports (with polling), useReportPreference, generate/update mutations
- **Date utilities**: `formatUtcTimeAsLocal`, `formatLocalTimeAsUtc`, `parseUtcTimeToLocalDate`

### Tests
- **DailyReportJobTest**: parameterized tests for scheduler window calculation (including midnight boundary)
- **ReportsResourceTest**: 6 integration tests covering preferences CRUD (partial updates, COALESCE preservation) and report lifecycle (empty page, 404 on nonexistent complete)
- **date.test.ts**: roundtrip and format validation for time conversion utilities

[...]

PR #7087, opened 2026-06-14
[OPIK-6852] [BE] Agent Insights jobs CRUD

aadereiko:
## Details

Per-(workspace, project) CRUD for the Agent Insights job — create / get / update / trigger. Built mirroring the Ollie daily-briefing config (`ReportPreferenceDAO` / `ReportService` / `ReportsResource`, PR #6640).

- New `agent_insights_jobs` table (db-app-state migration `000083`) + `AgentInsightsJobDAO` + `AgentInsightsJobService` + `AgentInsightsJobsResource`.
- Endpoints under `/v1/private/agent-insights/jobs` (`project_id` is a path variable):
  - `POST /jobs/{projectId}` — create; **201** on success, **409** if a job already exists for the project (insert-only, no upsert). `404` if the project doesn't exist.
  - `GET /jobs/{projectId}` — get; `404` if none.
  - `PATCH /jobs/{projectId}` — partial update (today: `status` enable/disable); never deletes; returns the updated job (`200`), `404` if none.
  - `POST /jobs/{projectId}/trigger` — accepts an immediate-run request (`202`); `404` if none. **Placeholder/no-op for now** — the bounded async execution (queue) is built with the cron in OPIK-6853.
- `workspace_id` always comes from the auth context, never the client.
- Scheduler-only concerns (cron pickup `findAllEnabled`, run tracking `last_triggered_at`/`markTriggered`, the report client + queued execution) are **deferred to OPIK-6853** — this PR is CRUD only.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- OPIK-6852

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.x
- Scope: Implementation, tests, and this PR description
- Human verification: Reviewed and tested by the author

## Testing

`mvn test` (apps/opik-backend, JDK 25) — `AgentInsightsJobsResourceTest` 7/7, spotless clean. Scenarios:
- **Create** → `201`; creating again → `409`; non-existent project → `404`.
- **Get** → `200` / `404` when absent.
- **Update** (PATCH) flips status without deleting, can re-enable, `404` when absent.
- **Trigger** → `202` for an existing job, `404` when absent.
- **Workspace isolation** — a job created in one workspace is invisible to another.

## Documentation

No user-facing documentation required — internal `/v1/private` configuration API.


CometActions:

### Access Information
- **URL:** https://pr-7087.dev.comet.com
- **Namespace:** pr-7087
- **Version:** 2.0.63-7087-merge-2280
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7087%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/AgentInsightsJobResource.java
```diff
@@ -0,0 +1,99 @@
...
+
+import static com.comet.opik.utils.AsyncUtils.setRequestContext;
+
+@Path("/v1/private/agent-insights/jobs")
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+@Timed
+@RequiredArgsConstructor(onConstructor_ = @Inject)
+@Tag(name = "Agent Insights Jobs", description = "Per-(workspace, project) Agent Insights report configuration")
+public class AgentInsightsJobResource {
```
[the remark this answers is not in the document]
aadereiko: Done in 8883b8e — renamed `AgentInsightsJobResource` → `AgentInsightsJobsResource` to match the plural convention.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/AgentInsightsJobResourceTest.java
```diff
@@ -0,0 +1,227 @@
...
+
+import java.util.UUID;
+
+import static com.comet.opik.api.resources.utils.ClickHouseContainerUtils.DATABASE_NAME;
+import static com.comet.opik.api.resources.utils.TestDropwizardAppExtensionUtils.newTestDropwizardAppExtension;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@ExtendWith(DropwizardAppExtensionProvider.class)
+class AgentInsightsJobResourceTest {
```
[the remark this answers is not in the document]
aadereiko: Done in 8883b8e — renamed to `AgentInsightsJobsResourceTest`.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/AgentInsightsJobsResource.java
```diff
@@ -0,0 +1,99 @@
...
+            @ApiResponse(responseCode = "201", description = "Job created", content = @Content(schema = @Schema(implementation = AgentInsightsJob.class))),
+            @ApiResponse(responseCode = "200", description = "Job already existed", content = @Content(schema = @Schema(implementation = AgentInsightsJob.class))),
+            @ApiResponse(responseCode = "404", description = "Project not found")
+    })
+    @RequiredPermissions(WorkspaceUserPermission.PROJECT_DATA_VIEW)
+    public Response enable(@Valid @NotNull AgentInsightsJobRequest request,
+            @Context UriInfo uriInfo) {
+        EnableResult result = service.enable(request.projectId())
+                .contextWrite(ctx -> setRequestContext(ctx, requestContext))
+                .block();
```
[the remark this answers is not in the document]
aadereiko: Done — the `Location` header is now documented on the 201 response (8883b8e), and `AgentInsightsJobsResourceTest.enable__missingProjectId__returnsUnprocessableEntity` covers the failure path (1f5ea80). Note the missing-`project_id` case returns **422** (opik's bean-validation status via the constraint-violation mapper), not 400.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobService.java
```diff
@@ -0,0 +1,124 @@
...
+            String projectName = projectService.findByIds(workspaceId, Set.of(projectId))
+                    .stream().findFirst().map(Project::name)
+                    .orElseThrow(() -> new NotFoundException("Project not found: " + projectId));
+
+            EnableResult result = transactionTemplate.inTransaction(WRITE, handle -> {
+                var dao = handle.attach(AgentInsightsJobDAO.class);
+                boolean existed = dao.findByProject(workspaceId, projectId).isPresent();
+                dao.enable(idGenerator.generateId(), workspaceId, projectId, userName);
+                AgentInsightsJob job = dao.findByProject(workspaceId, projectId).orElseThrow();
+                return new EnableResult(job, !existed);
```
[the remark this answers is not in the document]
aadereiko: Thanks — looked into both:

- **`created` from a pre-upsert read:** I tried deriving it from the `ON DUPLICATE KEY UPDATE` affected-row count (1=insert, 2=update) as suggested, but this project's MySQL connector returns *matched* (not *changed*) rows, so the duplicate path also reports `1` — which broke the 201-vs-200 distinction (the idempotency test caught it). Reverted to the in-transaction pre-read. The residual race only affects 201-vs-200 on a *simultaneous* double-enable of the same project; the response body is identical either way, so the impact is cosmetic.
- **project-existence check in a separate transaction:** `projectService.findByIds(...)` runs on its own connection, so co-locating it in the `enable` transaction wouldn't actually make it atomic — a robust fix needs an FK from `agent_insights_jobs` to the projects table. The window (deleting a project at the exact instant of enabling) is very narrow; happy to add the FK as a follow-up if you'd prefer.

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7087`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000082_create_agent_insights_jobs_table.sql
```diff
@@ -0,0 +1,21 @@
...
+    project_id CHAR(36) NOT NULL,
+    status ENUM('enabled', 'disabled') NOT NULL DEFAULT 'enabled',
+    last_triggered_at TIMESTAMP(6) NULL,
+    created_at TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
+    created_by VARCHAR(255) NOT NULL DEFAULT 'admin',
+    last_updated_at TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6),
+    last_updated_by VARCHAR(255) NOT NULL DEFAULT 'admin',
+    CONSTRAINT agent_insights_jobs_pk PRIMARY KEY (id),
+    CONSTRAINT agent_insights_jobs_ws_project_uk UNIQUE (workspace_id, project_id),
+    INDEX idx_agent_insights_jobs_status (status)
```
BorisTkachenko: We don't need this index, plz remove it.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/AgentInsightsJobsResource.java
```diff
@@ -0,0 +1,100 @@
...
+    private final @NonNull Provider<RequestContext> requestContext;
+
+    @POST
+    @Operation(operationId = "enableAgentInsightsJob", summary = "Enable Agent Insights job", description = "Enables the Agent Insights report for a project (idempotent on workspace+project) and triggers an immediate first run. Returns 201 on first enable, 200 if it already existed.", responses = {
+            @ApiResponse(responseCode = "201", description = "Job created", headers = @Header(name = "Location", description = "URI of the created job", schema = @Schema(type = "string")), content = @Content(schema = @Schema(implementation = AgentInsightsJob.class))),
+            @ApiResponse(responseCode = "200", description = "Job already existed", content = @Content(schema = @Schema(implementation = AgentInsightsJob.class))),
+            @ApiResponse(responseCode = "404", description = "Project not found")
+    })
+    @RequiredPermissions(WorkspaceUserPermission.PROJECT_DATA_VIEW)
+    public Response enable(@Valid @NotNull AgentInsightsJobRequest request,
```
BorisTkachenko: Let's separate endpoints.
- POST to create a job, no triggering
- endpoint to trigger a specific job (regardless of status?)
- endpoint to switch status enabled / disabled

BorisTkachenko (commented):
Initial review

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000082_create_agent_insights_jobs_table.sql
aadereiko: Removed in 95586dd.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/AgentInsightsJobsResource.java
aadereiko: Done in 8c7cd69 — split into three endpoints:
- `POST /jobs` — create only, no triggering (201 new / 200 existing)
- `POST /jobs/trigger` — triggers an immediate run for an existing job, regardless of status (202; 404 if none)
- `PATCH /jobs?project_id=…` — switch status enabled/disabled (partial update; 404 if none)

The old single `POST` (upsert + trigger) and `POST /disable` are gone. Description updated to match.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000083_create_agent_insights_jobs_table.sql
```diff
@@ -0,0 +1,20 @@
...
+    workspace_id VARCHAR(150) NOT NULL,
+    project_id CHAR(36) NOT NULL,
+    status ENUM('enabled', 'disabled') NOT NULL DEFAULT 'enabled',
+    last_triggered_at TIMESTAMP(6) NULL,
+    created_at TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
+    created_by VARCHAR(255) NOT NULL DEFAULT 'admin',
+    last_updated_at TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6),
+    last_updated_by VARCHAR(255) NOT NULL DEFAULT 'admin',
+    CONSTRAINT agent_insights_jobs_pk PRIMARY KEY (id),
+    CONSTRAINT agent_insights_jobs_ws_project_uk UNIQUE (workspace_id, project_id)
```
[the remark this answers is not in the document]
aadereiko: Resolved by renumbering rather than a follow-up: main merged a different `000082_create_agent_insights_issues_tables.sql` while this PR was open, so the whole migration was renamed to `000083` (b6250e0). As a brand-new changeset id it's never been applied anywhere, so the in-place checksum concern doesn't apply.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000083_create_agent_insights_jobs_table.sql
```diff
@@ -0,0 +1,20 @@
...
+) ENGINE = InnoDB
+  DEFAULT CHARSET = utf8mb4;
+
+--rollback DROP TABLE IF EXISTS agent_insights_jobs;
```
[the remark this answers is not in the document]
aadereiko: The file already ends with a trailing newline (`…agent_insights_jobs;\n`), matching the existing merged migrations (`000080`, `000081`) which end the same way with no extra blank line. Keeping it consistent with the rest of the directory rather than introducing a lone blank-line variant.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000083_create_agent_insights_jobs_table.sql
```diff
@@ -0,0 +1,20 @@
+--liquibase formatted sql
+--changeset aadereiko:000083_create_agent_insights_jobs_table
+--comment: Create the agent_insights_jobs table — per-(workspace, project) Agent Insights report configuration
+
+CREATE TABLE IF NOT EXISTS agent_insights_jobs (
+    id CHAR(36) NOT NULL,
+    workspace_id VARCHAR(150) NOT NULL,
+    project_id CHAR(36) NOT NULL,
+    status ENUM('enabled', 'disabled') NOT NULL DEFAULT 'enabled',
```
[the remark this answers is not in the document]
aadereiko: Not an issue — JDBI maps enums by name **case-insensitively** by default, so the DB's lowercase `enabled`/`disabled` deserialize to `Status.ENABLED`/`DISABLED` without a custom mapper. Verified by `AgentInsightsJobDAOTest`: it inserts via `enable`, reads back through `findByProject` (SELECT *), and asserts the `Status` value — 6/6 green against real MySQL; `create`/`update` exercise the same read path. There's also existing precedent in opik (datasets visibility, prompt-version type, automation-rule scope) storing lowercase enums mapped the same way. The `@JsonValue`/`@JsonCreator` on the enum are for the JSON layer, not JDBI.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/AgentInsightsJobsResource.java
```diff
@@ -0,0 +1,117 @@
...
+            @ApiResponse(responseCode = "202", description = "Run accepted"),
+            @ApiResponse(responseCode = "404", description = "Job not found")
+    })
+    @RequiredPermissions(WorkspaceUserPermission.PROJECT_DATA_VIEW)
+    public Response trigger(@Valid @NotNull AgentInsightsJobRequest request) {
+        service.triggerNow(request.projectId())
+                .contextWrite(ctx -> setRequestContext(ctx, requestContext))
+                .block();
+
+        return Response.accepted().build();
```
thiagohora, line 95: 🟠 **The 202 is misleading and the run is still inline/unbounded.** `triggerNow` `.block()`s on `triggerImmediate`, which performs the report-client call (and `markTriggered` write) synchronously before this returns — so the response is sent only after the work is fully done, not "accepted for later." `202 Accepted` implies async handoff that isn't happening; this is really a synchronous `200/204`.

More importantly there's no bound: a burst of `/trigger` calls fans out unbounded synchronous report triggers with no concurrency cap, retry, or dedup. To make the 202 honest *and* safe, enqueue the run onto a bounded Redis-backed queue/stream (Redisson is already in the stack) or hand it to the scheduler context (OPIK-6853), and return 202 once enqueued. Splitting trigger out from create was the right move — this is the remaining half.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000083_create_agent_insights_jobs_table.sql
```diff
@@ -0,0 +1,20 @@
...
+    last_triggered_at TIMESTAMP(6) NULL,
```
thiagohora: 🟠 **As scheduler config, "when does it run next" isn't expressible.** There's only `last_triggered_at` — no cadence/cron expression/interval and no `next_run_at`. The cron can't scope its pickup to jobs that are actually due; it can only re-scan everything and re-derive timing in code. Please define the schedule contract here (at minimum a `next_run_at`, plus the cadence source) so the pickup query can be `WHERE status='enabled' AND next_run_at <= now()`. If scheduling is genuinely out of scope for this ticket, the table/PR shouldn't be framed as scheduler config yet. See also `findAllEnabled` below.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobDAO.java
```diff
@@ -0,0 +1,60 @@
...
+            """)
+    void markTriggered(@Bind("id") UUID id, @Bind("triggeredAt") Instant triggeredAt);
+
+    // Cross-workspace — used only by the scheduler (system context), never from a request thread.
+    @SqlQuery("""
+            SELECT id, workspace_id, project_id FROM agent_insights_jobs
+            WHERE status = 'enabled'
+            """)
+    @RegisterConstructorMapper(EnabledJob.class)
+    List<EnabledJob> findAllEnabled();
```
thiagohora: 🟠 **Execution model for the scheduler is underspecified.** This is a full cross-workspace scan (`WHERE status='enabled'`) with: no index (the status index was dropped in this PR), no `next_run_at` due-time filter, no pagination/batching, and no claim/lease. As enabled jobs grow, the cron loads everything at once and (per the trigger comment) fans out unbounded; two scheduler instances would also double-run the same jobs. Before this is relied on as the cron pickup, it needs at least a due-time filter + batching, and a row-claiming/leasing strategy if more than one scheduler can run. Worth a `log()` of how many jobs were picked up so silent truncation/scale issues are visible.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobDAO.java
```diff
@@ -0,0 +1,60 @@
...
+@RegisterConstructorMapper(AgentInsightsJob.class)
+@RegisterArgumentFactory(UUIDArgumentFactory.class)
+interface AgentInsightsJobDAO {
+
+    @SqlUpdate("""
+            INSERT INTO agent_insights_jobs (id, workspace_id, project_id, status, created_by, last_updated_by)
+            VALUES (:id, :workspaceId, :projectId, 'enabled', :userName, :userName)
+            ON DUPLICATE KEY UPDATE status = 'enabled', last_updated_by = :userName
+            """)
+    void enable(@Bind("id") UUID id,
```
thiagohora: 🟡 **`enable` is misnamed now that `updateStatus` exists.** Both are justified — `enable` is the create-path **upsert** (insert-if-absent, used by `create()`), while `updateStatus` is the **update-or-404** used by PATCH; they have different existence semantics, so this isn't redundancy. But the name `enable` is a leftover from the previous enable/disable action design: it's only ever called from `create()` and always sets `status='enabled'`. Rename to reflect what it is (e.g. `insertOrEnable` / `upsert` / `create`) so the DAO reads as create-upsert + update, not "enable" + "updateStatus".

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobService.java
```diff
@@ -0,0 +1,137 @@
...
+        })).subscribeOn(Schedulers.boundedElastic());
+    }
+
+    // Triggers an immediate report run for an existing job over the last 24h. 404 if the job does not exist.
+    public Mono<Void> triggerNow(@NonNull UUID projectId) {
+        var ctx = requestContext.get();
+        String workspaceId = ctx.getWorkspaceId();
+        String workspaceName = ctx.getWorkspaceName();
+        return Mono.fromCallable(() -> {
+            String projectName = projectService.findByIds(workspaceId, Set.of(projectId))
```
thiagohora: 🟡 **Duplicates an existing helper.** `findByIds(...).stream().findFirst().map(Project::name).orElseThrow(NotFoundException)` reimplements `ProjectService.get(UUID id, String workspaceId)`, which already fetches-or-throws `NotFoundException`. Caveat: `get(...)` also does a ClickHouse `getLastUpdatedTraceAt` round-trip (heavier than needed for just the name), so either reuse it and accept the extra read, or add a lightweight name/existence helper — but the hand-rolled `findByIds().stream()` shouldn't stay. (`create()` at line 53 uses `findByIds(...).isEmpty()` for the existence check, which is fine since it doesn't need the name.)

apps/opik-backend/src/main/java/com/comet/opik/api/AgentInsightsJob.java
```diff
@@ -0,0 +1,60 @@
...
+import java.time.Instant;
+import java.util.UUID;
+
+@Builder(toBuilder = true)
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
+public record AgentInsightsJob(
+        @JsonIgnore String workspaceId,
+        @Schema(accessMode = Schema.AccessMode.READ_ONLY) UUID id,
+        UUID projectId,
```
thiagohora: 🟡 `projectId` should be `@NotNull` — it's a `NOT NULL` column, always populated, and consistent with how nearby records annotate it (e.g. `AnnotationQueue`). This is a response-only model, so the practical effect is OpenAPI/contract clarity (marks `project_id` required for SDK/clients) rather than runtime validation; `@Schema(requiredMode = REQUIRED)` is the more honest signal for a read-only field if you prefer.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobDAO.java
```diff
@@ -0,0 +1,60 @@
...
+            WHERE workspace_id = :workspaceId AND project_id = :projectId
+            """)
+    Optional<AgentInsightsJob> findByProject(@Bind("workspaceId") String workspaceId,
+            @Bind("projectId") UUID projectId);
+
+    @SqlUpdate("""
+            UPDATE agent_insights_jobs SET last_triggered_at = :triggeredAt
+            WHERE id = :id
+            """)
+    void markTriggered(@Bind("id") UUID id, @Bind("triggeredAt") Instant triggeredAt);
```
thiagohora: 🟡 **Nit (hardening):** `WHERE id = :id` is correct since `id` is the globally-unique PK, so this isn't a defect — but adding `AND workspace_id = :workspaceId` would be defense-in-depth, consistent with the other workspace-scoped queries here. The caller (`triggerImmediate`) already has the job in hand, so threading the workspace through is cheap.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/AgentInsightsJobsResource.java
```diff
@@ -0,0 +1,117 @@
...
+        return Response.ok(job).build();
+    }
+
+    @POST
+    @Path("/trigger")
+    @Operation(operationId = "triggerAgentInsightsJob", summary = "Trigger Agent Insights job", description = "Triggers an immediate report run for an existing job (over the last 24h). Fire-and-forget; returns 202. 404 if none exists.", responses = {
+            @ApiResponse(responseCode = "202", description = "Run accepted"),
+            @ApiResponse(responseCode = "404", description = "Job not found")
+    })
+    @RequiredPermissions(WorkspaceUserPermission.PROJECT_DATA_VIEW)
```
thiagohora, line 91: 🟡 **Authorization: mutating/side-effecting endpoints gated by a view permission.** All four endpoints — including PATCH `update` and this `trigger` (which fires a report run) — use `PROJECT_DATA_VIEW`, so a read-only viewer can change job status and kick off runs. Other mutating resources use write-scoped permissions (`ALERT_UPDATE`, `ONLINE_EVALUATION_RULE_UPDATE`, `DASHBOARD_EDIT`). There's no agent-insights-specific permission yet, so this may be deliberate — but please make it a conscious decision rather than a default, especially for `trigger`.

thiagohora (changes requested):
Re-reviewed against the current PR head (create / PATCH update / trigger split). Nicely done on the rework — the earlier doc/code-mismatch (create no longer triggers) and the REST verb/modeling (PATCH for status, dedicated trigger) are both resolved, and dropping the low-selectivity status index was the right call.

Remaining items are inline. The two I'd like addressed before merge:
- **Trigger is still synchronous/inline and unbounded**, and the `202` is misleading (the run completes before the response). Needs a bounded queue/scheduler handoff.
- **Scheduler semantics aren't expressible** — no `next_run_at`/cadence, and `findAllEnabled` is an unindexed full scan with no due-filter/batching/claim. Either define the schedule contract or stop framing the table as scheduler config in this PR.

The rest are smaller: `enable` is misnamed now that `updateStatus` exists (both are justified — upsert vs update-or-404), the project lookup in `triggerNow` duplicates `ProjectService.get(id, workspaceId)`, `projectId` should be `@NotNull`, a `markTriggered` workspace-scope hardening nit, and a permission-scope question (mutations/trigger gated by `PROJECT_DATA_VIEW`).

Confirmed fine, not changes: the `INSERT … ON DUPLICATE KEY UPDATE` upsert (atomic, race-safe, matches `ReportPreferenceDAO`), and computing `created` via a separate `findByProject` rather than ODKU's affected-rows count.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/AgentInsightsJobsResource.java
thiagohora: 🔴 **Escalating this to a blocker — treat backpressure as required, not deferrable to OPIK-6854.**

The core issue is independent of whether the report client is wired yet: `triggerNow` `.block()`s on `triggerImmediate`, which performs the report-client call **and** the `markTriggered` write synchronously on the request thread before responding. There is no bound on concurrency — N simultaneous `/trigger` calls = N simultaneous synchronous report runs, with no queue, cap, retry, or dedup. Once OPIK-6854 swaps in the real HTTP client this is a direct path to thread/connection-pool exhaustion and downstream overload, and the contract (`202`) is already wrong today.

Required before merge: put a bounded async mechanism in front of the run — a Redis-backed queue/stream (Redisson is already in the stack) or the scheduler's system context — so the endpoint *enqueues* the run and returns `202` once accepted, with the worker enforcing the concurrency limit. The endpoint should not block on the run itself.

Building the bounded path now (even against the no-op client) means OPIK-6854 only swaps the client implementation rather than re-architecting the request path under load.

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7087`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000083_create_agent_insights_jobs_table.sql
BorisTkachenko: I believe we don't need this `last_triggered_at` at all. We just set it, never using and it doesn't determine when the job is running. If we later need to figure out whether the job already has been executed we can check if there is any report available.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/AgentInsightsJobsResource.java
```diff
@@ -0,0 +1,117 @@
...
+                .contextWrite(ctx -> setRequestContext(ctx, requestContext))
+                .block();
+
+        if (result.created()) {
+            URI location = uriInfo.getAbsolutePathBuilder()
+                    .queryParam("project_id", request.projectId())
+                    .build();
+            return Response.created(location).entity(result.job()).build();
+        }
+        return Response.ok(result.job()).build();
```
BorisTkachenko: In case Job already exists let's return an error, to be consistent with the other entities implementation: smth like `EntityAlreadyExistsException(new ErrorMessage(409, alreadyExists));` - you can refer to Prompts

apps/opik-backend/src/main/java/com/comet/opik/api/AgentInsightsJob.java
BorisTkachenko: Looks like this class is used as a Response only, never for Request, so probably we either need `@Schema(accessMode = Schema.AccessMode.READ_ONLY)` for all fields or I would rather remove it from this class.

apps/opik-backend/src/main/java/com/comet/opik/api/AgentInsightsJobRequest.java
```diff
@@ -0,0 +1,17 @@
...
+import jakarta.validation.constraints.NotNull;
+import lombok.Builder;
+
+import java.util.UUID;
+
+@Builder(toBuilder = true)
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
+public record AgentInsightsJobRequest(
+        @NotNull @Schema(description = "Project the Agent Insights job applies to", example = "019384a2-1b3c-7def-8a9b-0c1d2e3f4a5b") UUID projectId) {
```
BorisTkachenko: Considering that it's always project bound. I would recommend putting `projectId` as a path variable.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobDAO.java
BorisTkachenko: Agree, let's rename to `create`, also let's use it for creation only considering that now we have a different flow for `status` changes. Do not update status here, rather throw an error if already exists.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobDAO.java
BorisTkachenko: Probably we don't need this `markTriggered` at all, it sets `last_triggered_at`, that has no use.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobDAO.java
BorisTkachenko: Let's remove `findAllEnabled` for now. It doesn't belong to this PR, rather to the next OPIK-6853, where will be cron job implementation.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobService.java
```diff
@@ -0,0 +1,137 @@
...
+            // 404 if the project does not exist in this workspace.
+            if (projectService.findByIds(workspaceId, Set.of(projectId)).isEmpty()) {
+                throw new NotFoundException("Project not found: " + projectId);
+            }
+
+            return transactionTemplate.inTransaction(WRITE, handle -> {
+                var dao = handle.attach(AgentInsightsJobDAO.class);
+                boolean existed = dao.findByProject(workspaceId, projectId).isPresent();
+                dao.enable(idGenerator.generateId(), workspaceId, projectId, userName);
+                AgentInsightsJob job = dao.findByProject(workspaceId, projectId).orElseThrow();
```
BorisTkachenko: As mentioned in the other comments, let's not enable it, just create, and if it exists, throw an error.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobService.java
```diff
@@ -0,0 +1,137 @@
...
+                            .findByProject(workspaceId, projectId)
+                            .orElseThrow(() -> new NotFoundException(
+                                    "Agent insights job not found for project: " + projectId)));
+            triggerImmediate(job, workspaceName, projectName);
+            return null;
+        }).subscribeOn(Schedulers.boundedElastic()).then();
+    }
+
+    // Cross-workspace; used by the scheduler (OPIK-6853), never from a request thread.
+    public List<EnabledJob> findAllEnabled() {
```
BorisTkachenko: Similar, it belongs to OPIK-6853, not here.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobService.java
```diff
@@ -0,0 +1,137 @@
...
+                    job.projectId(), projectName, workspaceName, periodStart, periodEnd);
+            transactionTemplate.inTransaction(WRITE, handle -> {
+                handle.attach(AgentInsightsJobDAO.class).markTriggered(job.id(), periodEnd);
+                return null;
+            });
+        } catch (Exception e) {
+            // Best-effort: a trigger failure must not fail the request. Catch broadly on purpose — the
+            // nightly cron (OPIK-6853) retries, so any failure here is non-fatal.
+            log.error("Failed to trigger Agent Insights run for project '{}'", job.projectId(), e);
+        }
```
BorisTkachenko: As mentioned by @thiagohora in the other comments this should go through a queue. And queue implementation would be better to add in OPIK-6853, since we will need it for cron job. This PR is mainly for CRUD. To separate the concerns, I would suggest not to implement this method here. Just add a TODO for `triggerImmediate` method.

Right now it's not implemented fully anyway since we have only `NoopAgentInsightsReportClient`

apps/opik-backend/src/main/java/com/comet/opik/domain/NoopAgentInsightsReportClient.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.UUID;
+
+/**
+ * Default {@link AgentInsightsReportClient} binding: no-op until OPIK-6854 wires the real Platform-BE
+ * client. Mirrors {@code OrchestratorClient.isEnabled() == false} — {@code isEnabled()} returns false
+ * so callers skip the immediate run rather than failing.
+ */
+@Singleton
+@Slf4j
+public class NoopAgentInsightsReportClient implements AgentInsightsReportClient {
```
BorisTkachenko: Plz remove, we don't need this if we remove `triggerImmediate` in the service. Or you can add a normal class without implementation.

apps/opik-backend/src/test/java/com/comet/opik/domain/AgentInsightsJobDAOTest.java
```diff
@@ -0,0 +1,199 @@
...
+import java.util.Optional;
+import java.util.UUID;
+
+import static com.comet.opik.infrastructure.db.TransactionTemplateAsync.READ_ONLY;
+import static com.comet.opik.infrastructure.db.TransactionTemplateAsync.WRITE;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@ExtendWith(DropwizardAppExtensionProvider.class)
+class AgentInsightsJobDAOTest {
```
BorisTkachenko: Please remove it integration test is enough.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/AgentInsightsJobTriggerTest.java
```diff
@@ -0,0 +1,167 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Exercises the enabled trigger branch of enable() — the no-op default client (used by the other
+ * test classes) never fires, so here we override the {@link AgentInsightsReportClient} binding with a
+ * recording stub whose {@code isEnabled()} returns true.
+ */
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@ExtendWith(DropwizardAppExtensionProvider.class)
+class AgentInsightsJobTriggerTest {
```
BorisTkachenko: Please remove it from this PR. Also, when you will implement it next, I suppose we can reuse the above integration test `AgentInsightsJobsResourceTest` and add a nested subclass instead of creating a new one.

BorisTkachenko (changes requested):
Thank you, let's adjust it a little bit before merging.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/AgentInsightsJobsResource.java
aadereiko: Conscious decision for now: there's no agent-insights-specific write permission, and these are project-scoped config endpoints, so all of them stay on `PROJECT_DATA_VIEW`. When the scheduler/trigger work moves to OPIK-6853 (trigger behind a queue), I'll revisit and introduce a write-scoped permission for the mutating endpoints then. Flagging it as deliberate rather than a default.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/AgentInsightsJobsResource.java
aadereiko: Resolved in 45f9fd1 by removing the synchronous path entirely from this PR (rather than building the queue here). `triggerNow` no longer blocks on any report-client work — it validates the job exists (404 otherwise) and is a no-op placeholder with a TODO(OPIK-6853) to enqueue the run on a bounded async queue (Redisson) for the scheduler worker. The report client (Noop + interface) was removed too. So there's no inline/unbounded execution on the request thread anymore; the bounded async mechanism you described will be built with the cron in OPIK-6853.

aadereiko:
Thanks @BorisTkachenko @thiagohora — addressed all review points. #7087 is now pure CRUD; scheduler/trigger execution is deferred to OPIK-6853 (#7105). Summary:

**Scoped down to CRUD (moved to OPIK-6853):**
- Removed `findAllEnabled` (DAO + Service) and the `EnabledJob` projection
- Removed `markTriggered` + `last_triggered_at` (migration column + response field)
- Removed `AgentInsightsReportClient` + `NoopAgentInsightsReportClient`
- `triggerNow` no longer runs synchronously (🔴 backpressure blocker): validates the job exists (404) and is a no-op placeholder with a `TODO(OPIK-6853)` to enqueue on a bounded async queue. Endpoint kept, returns 202.
- Removed `AgentInsightsJobDAOTest` and `AgentInsightsJobTriggerTest` (the resource integration test covers create/get/update/trigger)

**CRUD/API:**
- `project_id` is now a path variable: `POST/GET/PATCH /jobs/{projectId}` and `POST /jobs/{projectId}/trigger`
- Create returns **409** (`EntityAlreadyExistsException`) if the job already exists (insert-only DAO `create`, no upsert)
- `AgentInsightsJob.projectId` → `@Schema(READ_ONLY, REQUIRED)` (response-only model)
- Status index dropped from the migration
- Authorization: kept `PROJECT_DATA_VIEW` for now (no agent-insights write permission yet) — deliberate; will revisit when the mutating/trigger flow lands in OPIK-6853

All green: `AgentInsightsJobsResourceTest` 7/7, spotless clean.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobService.java
```diff
@@ -0,0 +1,134 @@
...
+                throw new NotFoundException("Project not found: " + projectId);
+            }
+
+            return transactionTemplate.inTransaction(WRITE, handle -> {
+                var dao = handle.attach(AgentInsightsJobDAO.class);
+                if (dao.findByProject(workspaceId, projectId).isPresent()) {
+                    throw new EntityAlreadyExistsException(new ErrorMessage(409,
+                            "Agent insights job already exists for project: " + projectId));
+                }
+                dao.create(idGenerator.generateId(), workspaceId, projectId, userName);
```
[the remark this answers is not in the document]
aadereiko: Fixed: `create()` now wraps the insert in `EntityConstraintHandler.handle(...).withError(...)`, so a concurrent unique-key violation is mapped to the same 409 instead of bubbling as a 500. Dropped the check-then-insert pre-read.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobService.java
```diff
@@ -0,0 +1,97 @@
...
+    // Creates the job; 409 if one already exists for the (workspace, project).
+    public Mono<AgentInsightsJob> create(@NonNull UUID projectId) {
+        var ctx = requestContext.get();
+        String workspaceId = ctx.getWorkspaceId();
+        String userName = ctx.getUserName();
+
+        return Mono.fromCallable(() -> {
+            if (projectService.findByIds(workspaceId, Set.of(projectId)).isEmpty()) {
+                throw new NotFoundException("Project not found: " + projectId);
+            }
```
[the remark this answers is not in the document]
aadereiko: Done — `create()` now calls `projectService.validateProjectIdExists(projectId, workspaceId)` instead of the hand-rolled `findByIds().isEmpty()` check.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobService.java
```diff
@@ -0,0 +1,95 @@
...
+                var dao = handle.attach(AgentInsightsJobDAO.class);
+                // Insert-only; the unique key (workspace_id, project_id) makes this race-safe — a
+                // concurrent create surfaces as a constraint violation, mapped to 409.
+                return EntityConstraintHandler.handle(() -> {
+                    dao.create(idGenerator.generateId(), workspaceId, projectId, userName);
+                    return dao.findByProject(workspaceId, projectId).orElseThrow();
+                }).withError(() -> new EntityAlreadyExistsException(new ErrorMessage(409,
+                        "Agent insights job already exists for project: " + projectId)));
+            });
+        }).subscribeOn(Schedulers.boundedElastic());
```
BorisTkachenko: Everywhere in this class you are forcing reactive pattern. It's not needed here, and adds in complexity. We are using it for analytics db (Clickhouse) due to a client limitation. 
Here it's a state DB and we don't need it. You can check Project Service/DAO or any other state DB related Service/DAO as a reference (Prompts, Alerts, ets.).

BorisTkachenko (commented):
Looks pretty good, just one comment from my side left.

BorisTkachenko (approved):
Thank you! Looks good from my side.

PR #7105, opened 2026-06-15
[OPIK-6853] [BE] Daily Agent Insights report cron

aadereiko:
## Details

Daily Agent Insights report generation for **enabled** projects, plus the bounded async path that actually runs a report. Stacked on #7087 (OPIK-6852 jobs CRUD), now rebased onto `main`.

**Daily cron** — `AgentInsightsReportJob` (00:05 UTC, mirrors `OllieDailyReportJob`):
- distributed-lock sweep so only one replica runs per day;
- previous-UTC-day "had traces" gate (`TraceService`/`TraceDAO.hasTracesInRange`, a `SELECT 1 … LIMIT 1` existence check);
- strict once-per-day guard via `last_triggered_at`;
- per-job failure isolation so one bad project never aborts the sweep.

**Bounded trigger queue** — both the manual `POST /v1/private/agent-insights/jobs/{projectId}/trigger` endpoint and the cron **enqueue** onto a Redis stream rather than calling out inline:
- `AgentInsightsReportPublisher` → Redis stream (mirrors `WebhookPublisher`/dataset-export);
- `AgentInsightsReportSubscriber` (consumer group) performs the actual report call, so neither the request thread nor the sweep blocks and concurrency is capped by `consumerBatchSize`;
- **at-most-once**: a failed trigger is logged and acked (dropped), not retried — report generation is a non-idempotent side effect and isn't deduped downstream, so a dropped best-effort run is preferred over a duplicate. `reportId` is carried as the idempotency key if downstream dedup is added later.

**Real platform client** — `PlatformAgentInsightsReportClient` (the `@ImplementedBy` default) POSTs the trigger to the merged Platform-BE endpoint `POST /opik/ollie/generate-agent-insights` (OPIK-6854) with the `AgentInsightsGenerateRequest` contract (snake_case). It self-disables (no-op) when `agentInsightsReport.triggerUrl` is blank, so OSS / unconfigured deployments do nothing.

**Other**
- `last_triggered_at` is advanced **only** by the cron path (`markTriggered`, which also stamps `last_updated_by = system`); ad-hoc `/trigger` runs never touch the daily marker.
- New config block `agentInsightsReport` (stream tunables + `triggerUrl`) in `OpikConfiguration` + `config.yml`; `JobTimeoutConfig.agentInsightsReportJobTimeout`.
- Migration `000084_add_last_triggered_at_to_agent_insights_jobs`.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- OPIK-6853

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.x
- Scope: Implementation, tests, and this PR description
- Human verification: Reviewed and tested by the author

## Testing

`mvn test` (apps/opik-backend, JDK 25) — all green; Spotless passes. Per-class:
- `AgentInsightsReportJobTest` (4) — sweep: triggers only enabled jobs with previous-day traces, skips already-triggered-today, per-job failure isolation, and `doJob` disabled early-exit.
- `AgentInsightsReportSubscriberTest` (4) — lifecycle gating + processEvent: triggers when client enabled, acks-without-triggering when disabled, and **acks at-most-once on trigger failure**.
- `PlatformAgentInsightsReportClientTest` (3, WireMock) — snake_case payload/path to the platform endpoint, `isEnabled()` URL gating, throw-on-≥300.
- `AgentInsightsJobTriggerTest` (2) — manual `/trigger` enqueues → consumes → fires (and leaves the daily marker untouched); **cron sweep integration** drives `runSweep` end-to-end against MySQL + ClickHouse + Redis (`findAllEnabled` → `hasTracesInRange` → publish → subscribe → client → `markTriggered`).
- `AgentInsightsJobsResourceTest` (7) — CRUD/trigger endpoints (create 201/409, get, patch, trigger 202/404, workspace isolation).

## Follow-ups (deferred, low severity)
- Batch the cron's per-project `hasTracesInRange` into a single grouped query for very large installs (currently serial, indexed, daily).
- Extract a shared `RedisStreamPublisher` helper across `WebhookPublisher` / dataset-export / this publisher.

## Documentation

No user-facing documentation required — internal scheduler + an internal trigger endpoint, no public API surface. Behavior (daily 00:05 UTC, previous-day trace gate, once-per-day, bounded queue) is captured in the design doc (Notion: "Scheduler for the Agent Insights report").

### DIFF
--- a/apps/opik-backend/config.yml
+++ b/apps/opik-backend/config.yml
@@ -801,6 +801,52 @@ datasetExport:
   # Description: Batch size for cleanup job operations (number of jobs to process per batch). Must be between 10 and 1000.
   cleanupBatchSize: ${DATASET_EXPORT_CLEANUP_BATCH_SIZE:-100}
 
+# Configuration for the Agent Insights report trigger queue (manual /trigger + daily cron enqueue here;
+# a bounded Redis consumer group performs the actual report calls)
+agentInsightsReport:
+  # Default: http://react-svc:8080/opik/ollie/generate-agent-insights
+  # Description: Platform BE endpoint that triggers Agent Insights report generation on the Ollie pod
+  triggerUrl: ${AGENT_INSIGHTS_REPORT_TRIGGER_URL:-http://react-svc:8080/opik/ollie/generate-agent-insights}
+  # Default: 0 5 0 * * ? (daily at 00:05 UTC)
+  # Description: Quartz cron expression (UTC) for the daily Agent Insights report sweep
+  schedule: ${AGENT_INSIGHTS_REPORT_SCHEDULE:-'0 5 0 * * ?'}
+  # Default: 5m
+  # Description: Timeout for the daily Agent Insights report job sweep (distributed lock hold time)
+  jobTimeout: ${AGENT_INSIGHTS_REPORT_JOB_TIMEOUT:-5m}
+  # Default: 5s
+  # Description: How long to wait to acquire the distributed lock before giving up the sweep
+  lockWaitTime: ${AGENT_INSIGHTS_REPORT_LOCK_WAIT_TIME:-5s}
+  # Default: agent-insights-reports
+  # Description: Redis stream name for queued Agent Insights report triggers
+  streamName: ${AGENT_INSIGHTS_REPORT_STREAM_NAME:-'agent-insights-reports'}
+  # Default: agent-insights-report-consumers
+  # Description: Consumer group name for Agent Insights report trigger processing
+  consumerGroupName: ${AGENT_INSIGHTS_REPORT_CONSUMER_GROUP_NAME:-'agent-insights-report-consumers'}
+  # Default: 5
+  # Description: Maximum number of report triggers to process in a single batch (caps concurrent runs)
+  consumerBatchSize: ${AGENT_INSIGHTS_REPORT_CONSUMER_BATCH_SIZE:-5}
+  # Default: 1s
+  # Description: How often to poll Redis for new report triggers
+  poolingInterval: ${AGENT_INSIGHTS_REPORT_POOLING_INTERVAL:-1s}
+  # Default: 5s
+  # Description: Timeout for blocking read operations on Redis streams (long polling)
+  longPollingDuration: ${AGENT_INSIGHTS_REPORT_LONG_POLLING_DURATION:-5s}
+  # Default: 3
+  # Description: Maximum number of retries for a failing report trigger before it is dropped
+  maxRetries: ${AGENT_INSIGHTS_REPORT_MAX_RETRIES:-3}
+  # Default: 2
+  # Description: Claim pending messages every N polling intervals
+  claimIntervalRatio: ${AGENT_INSIGHTS_REPORT_CLAIM_INTERVAL_RATIO:-2}
+  # Default: 5m
+  # Description: How long a message can stay pending before it is auto-claimed by another consumer
+  pendingMessageDuration: ${AGENT_INSIGHTS_REPORT_PENDING_MESSAGE_DURATION:-5m}
+  # Default: 10000
+  # Description: Maximum stream length before old entries are trimmed
+  streamMaxLen: ${AGENT_INSIGHTS_REPORT_STREAM_MAX_LEN:-10000}
+  # Default: 100
+  # Description: Number of entries trimmed at a time when the stream exceeds streamMaxLen
+  streamTrimLimit: ${AGENT_INSIGHTS_REPORT_STREAM_TRIM_LIMIT:-100}
+
 # Configuration for Optimization Studio logs synchronization
 optimizationLogs:
   # Default: true
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/AgentInsightsReportSubscriber.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/AgentInsightsReportSubscriber.java
@@ -1,0 +1,94 @@
+package com.comet.opik.api.resources.v1.events;
+
+import com.comet.opik.domain.AgentInsightsReportClient;
+import com.comet.opik.domain.AgentInsightsReportMessage;
+import com.comet.opik.infrastructure.AgentInsightsReportConfig;
+import com.comet.opik.infrastructure.ServiceTogglesConfig;
+import jakarta.inject.Inject;
+import lombok.NonNull;
+import lombok.extern.slf4j.Slf4j;
+import org.redisson.api.RedissonReactiveClient;
+import reactor.core.publisher.Mono;
+import reactor.core.scheduler.Schedulers;
+import ru.vyarus.dropwizard.guice.module.installer.feature.eager.EagerSingleton;
+import ru.vyarus.dropwizard.guice.module.yaml.bind.Config;
+
+/**
+ * Consumes queued Agent Insights report triggers (published by {@code AgentInsightsReportPublisher}) and
+ * performs the actual report call via {@link AgentInsightsReportClient}. The Redis consumer group caps
+ * concurrent runs, so neither the manual {@code /trigger} endpoint nor the daily sweep blocks on the call.
+ */
+@Slf4j
+@EagerSingleton
+public class AgentInsightsReportSubscriber extends BaseRedisSubscriber<AgentInsightsReportMessage> {
+
+    private static final String METRICS_NAMESPACE = "opik";
+    private static final String METRICS_BASE_NAME = "agent_insights_report";
+
+    private final AgentInsightsReportConfig config;
+    private final ServiceTogglesConfig serviceToggles;
+    private final AgentInsightsReportClient reportClient;
+
+    @Inject
+    public AgentInsightsReportSubscriber(
+            @NonNull @Config("agentInsightsReport") AgentInsightsReportConfig config,
+            @NonNull @Config("serviceToggles") ServiceTogglesConfig serviceToggles,
+            @NonNull RedissonReactiveClient redisson,
+            @NonNull AgentInsightsReportClient reportClient) {
+        super(config, redisson, AgentInsightsReportConfig.PAYLOAD_FIELD, METRICS_NAMESPACE, METRICS_BASE_NAME);
+        this.config = config;
+        this.serviceToggles = serviceToggles;
+        this.reportClient = reportClient;
+    }
+
+    @Override
+    public void start() {
+        if (isDisabled()) {
+            return;
+        }
+        log.info("Starting Agent Insights report subscriber with config: streamName='{}', consumerGroupName='{}', "
+                + "batchSize='{}'", config.getStreamName(), config.getConsumerGroupName(),
+                config.getConsumerBatchSize());
+        super.start();
+    }
+
+    @Override
+    public void stop() {
+        if (isDisabled()) {
+            return;
+        }
+        log.info("Stopping Agent Insights report subscriber");
+        super.stop();
+    }
+
+    @Override
+    protected Mono<Void> processEvent(@NonNull AgentInsightsReportMessage message) {
+        log.info("Processing Agent Insights report trigger: reportId='{}', project='{}', workspace='{}'",
+                message.reportId(), message.projectId(), message.workspaceId());
+
+        return Mono.fromRunnable(() -> reportClient.triggerAgentInsights(message.reportId(), message.projectId(),
+                message.workspaceId(), message.periodStart(), message.periodEnd()))
+                .subscribeOn(Schedulers.boundedElastic())
+                .then()
+                .doOnSuccess(unused -> log.info("Triggered Agent Insights report: reportId='{}', project='{}'",
+                        message.reportId(), message.projectId()))
+                .onErrorResume(throwable -> {
+                    // At-most-once on purpose: report generation is a non-idempotent side effect (Ollie
+                    // compute + a user-facing report), and the trigger isn't deduplicated downstream, so a
+                    // redelivery would run the same report twice. We ack on failure (log + complete) rather
+                    // than rethrow; the run is dropped (best-effort daily briefing) instead of duplicated.
+                    // reportId is carried on the message as the idempotency key if downstream dedup is added.
+                    log.error("Failed to trigger Agent Insights report, dropping reportId='{}', project='{}'",
+                            message.reportId(), message.projectId(), throwable);
+                    return Mono.empty();
+                });
+    }
+
+    private boolean isDisabled() {
+        if (!serviceToggles.isAgentInsightsEnabled()) {
+            log.info("Agent Insights is disabled, skipping report subscriber lifecycle operation");
+            return true;
+        }
+        return false;
+    }
+}
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/AgentInsightsReportJob.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/AgentInsightsReportJob.java
@@ -1,0 +1,101 @@
+package com.comet.opik.api.resources.v1.jobs;
+
+import com.comet.opik.domain.AgentInsightsJobService;
+import com.comet.opik.domain.AgentInsightsReportPublisher;
+import com.comet.opik.domain.TraceService;
+import com.comet.opik.infrastructure.OpikConfiguration;
+import com.comet.opik.infrastructure.lock.LockService;
+import com.google.common.annotations.VisibleForTesting;
+import io.dropwizard.jobs.Job;
+import jakarta.inject.Inject;
+import jakarta.inject.Singleton;
+import lombok.NonNull;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.tuple.Pair;
+import org.quartz.DisallowConcurrentExecution;
+import org.quartz.JobExecutionContext;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+import reactor.core.scheduler.Schedulers;
+
+import java.time.Instant;
+import java.time.LocalDate;
+import java.time.ZoneOffset;
+
+import static com.comet.opik.infrastructure.lock.LockService.Lock;
+
+/**
+ * Daily cron that triggers Agent Insights report generation for enabled projects that had traces during
+ * the previous UTC day. The schedule is configured programmatically (OpikGuiceyLifecycleEventListener)
+ * from {@code agentInsightsReport.schedule}, so it is env-configurable. A distributed lock guarantees a
+ * single replica runs the sweep per day, so no extra once-per-day guard is needed. Mirrors
+ * {@code OllieDailyReportJob} (distributed lock + per-item failure isolation).
+ */
+@Slf4j
+@Singleton
+@DisallowConcurrentExecution
+@RequiredArgsConstructor(onConstructor_ = @Inject)
+public class AgentInsightsReportJob extends Job {
+
+    private static final Lock JOB_LOCK = new Lock("agent_insights_report_job:lock");
+
+    private final @NonNull AgentInsightsJobService agentInsightsJobService;
+    private final @NonNull TraceService traceService;
+    private final @NonNull AgentInsightsReportPublisher reportPublisher;
+    private final @NonNull LockService lockService;
+    private final @NonNull OpikConfiguration config;
+
+    @Override
+    public void doJob(JobExecutionContext context) {
+        LocalDate runDate = LocalDate.now(ZoneOffset.UTC);
+        Instant periodStart = runDate.minusDays(1).atStartOfDay(ZoneOffset.UTC).toInstant();
+        Instant periodEnd = runDate.atStartOfDay(ZoneOffset.UTC).toInstant();
+
+        log.info("Agent Insights report job running for previous UTC day ['{}', '{}')", periodStart, periodEnd);
+
+        var reportConfig = config.getAgentInsightsReport();
+        // holdUntilExpiry: only one replica runs the sweep per day (idempotent across replicas).
+        lockService.bestEffortLock(
+                JOB_LOCK,
+                runSweep(periodStart, periodEnd),
+                Mono.defer(() -> {
+                    log.debug("Could not acquire lock for Agent Insights report job, another instance is running");
+                    return Mono.empty();
+                }),
+                reportConfig.getJobTimeout().toJavaDuration(),
+                reportConfig.getLockWaitTime().toJavaDuration(),
+                true).subscribe(
+                        __ -> log.info("Agent Insights report job completed"),
+                        error -> log.error("Agent Insights report job failed", error));
+    }
+
+    // Visible for testing: the sweep does ONE trace query for the whole enabled set (no per-workspace or
+    // per-project loop), filters to projects that had traces, and enqueues each (per-job failure isolated).
+    @VisibleForTesting
+    public Mono<Void> runSweep(Instant periodStart, Instant periodEnd) {
+        return Mono.fromCallable(agentInsightsJobService::findAllEnabled)
+                .subscribeOn(Schedulers.boundedElastic())
+                .flatMap(jobs -> {
+                    if (jobs.isEmpty()) {
+                        return Mono.empty();
+                    }
+                    var pairs = jobs.stream()
+                            .map(job -> Pair.of(job.workspaceId(), job.projectId()))
+                            .toList();
+                    return traceService.getProjectsWithTracesInRange(pairs, periodStart, periodEnd)
+                            .flatMapMany(withTraces -> Flux.fromIterable(jobs)
+                                    .filter(job -> withTraces.contains(job.projectId())))
+                            .concatMap(job -> reportPublisher
+                                    .enqueue(job.projectId(), job.workspaceId(), periodStart, periodEnd)
+                                    .then()
+                                    .onErrorResume(e -> {
+                                        // Per-job isolation: a failed enqueue must not skip the rest.
+                                        log.error("Failed to enqueue Agent Insights run for project '{}'",
+                                                job.projectId(), e);
+                                        return Mono.empty();
+                                    }))
+                            .then();
+                });
+    }
+}
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobDAO.java
@@ -1,6 +1,7 @@
 package com.comet.opik.domain;
 
 import com.comet.opik.api.AgentInsightsJob;
+import com.comet.opik.api.AgentInsightsJob.EnabledJob;
 import com.comet.opik.infrastructure.db.UUIDArgumentFactory;
 import org.jdbi.v3.sqlobject.config.RegisterArgumentFactory;
 import org.jdbi.v3.sqlobject.config.RegisterConstructorMapper;
@@ -8,6 +9,7 @@ import org.jdbi.v3.sqlobject.customizer.Bind;
 import org.jdbi.v3.sqlobject.statement.SqlQuery;
 import org.jdbi.v3.sqlobject.statement.SqlUpdate;
 
+import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
 
@@ -16,7 +18,8 @@ import java.util.UUID;
 interface AgentInsightsJobDAO {
 
     @SqlUpdate("""
-            INSERT INTO agent_insights_jobs (id, workspace_id, project_id, status, created_by, last_updated_by)
+            INSERT INTO agent_insights_jobs
+                (id, workspace_id, project_id, status, created_by, last_updated_by)
             VALUES (:id, :workspaceId, :projectId, 'enabled', :userName, :userName)
             """)
     void create(@Bind("id") UUID id,
@@ -39,4 +42,16 @@ interface AgentInsightsJobDAO {
             """)
     Optional<AgentInsightsJob> findByProject(@Bind("workspaceId") String workspaceId,
             @Bind("projectId") UUID projectId);
+
+    // Cross-workspace — used only by the daily sweep (system context), never from a request thread.
+    // INNER JOIN projects so jobs whose project was deleted are filtered out at the source (no per-job
+    // existence check in the sweep loop).
+    @SqlQuery("""
+            SELECT j.id, j.workspace_id, j.project_id
+            FROM agent_insights_jobs j
+            INNER JOIN projects p ON p.id = j.project_id AND p.workspace_id = j.workspace_id
+            WHERE j.status = 'enabled'
+            """)
+    @RegisterConstructorMapper(EnabledJob.class)
+    List<EnabledJob> findAllEnabled();
 }
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobService.java
@@ -1,6 +1,7 @@
 package com.comet.opik.domain;
 
 import com.comet.opik.api.AgentInsightsJob;
+import com.comet.opik.api.AgentInsightsJob.EnabledJob;
 import com.comet.opik.api.error.EntityAlreadyExistsException;
 import com.comet.opik.infrastructure.auth.RequestContext;
 import io.dropwizard.jersey.errors.ErrorMessage;
@@ -13,6 +14,9 @@ import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import ru.vyarus.guicey.jdbi3.tx.TransactionTemplate;
 
+import java.time.Duration;
+import java.time.Instant;
+import java.util.List;
 import java.util.UUID;
 
 import static com.comet.opik.infrastructure.db.TransactionTemplateAsync.READ_ONLY;
@@ -23,10 +27,13 @@ import static com.comet.opik.infrastructure.db.TransactionTemplateAsync.WRITE;
 @Slf4j
 public class AgentInsightsJobService {
 
+    private static final Duration TRIGGER_WINDOW = Duration.ofHours(24);
+
     private final @NonNull TransactionTemplate transactionTemplate;
     private final @NonNull IdGenerator idGenerator;
     private final @NonNull Provider<RequestContext> requestContext;
     private final @NonNull ProjectService projectService;
+    private final @NonNull AgentInsightsReportPublisher reportPublisher;
 
     // Creates the job; 409 if one already exists for the (workspace, project).
     public AgentInsightsJob create(@NonNull UUID projectId) {
@@ -72,16 +79,32 @@ public class AgentInsightsJobService {
         });
     }
 
-    // 404 if the job does not exist.
+    // Manual trigger: validate the project (404) and job (404) on the request thread, then enqueue the
+    // report run on the bounded Redis-backed queue and return 202 — the request thread never blocks on
+    // the report call, and the consumer group caps concurrent runs.
     public void triggerNow(@NonNull UUID projectId) {
         String workspaceId = requestContext.get().getWorkspaceId();
-        transactionTemplate.inTransaction(READ_ONLY,
+        // Guard against orphaned jobs: a project may have been deleted out from under the job row.
+        projectService.validateProjectIdExists(projectId, workspaceId);
+        AgentInsightsJob job = transactionTemplate.inTransaction(READ_ONLY,
                 handle -> handle.attach(AgentInsightsJobDAO.class)
                         .findByProject(workspaceId, projectId)
                         .orElseThrow(() -> new NotFoundException(
                                 "Agent insights job not found for project: " + projectId)));
-        // TODO(OPIK-6853): enqueue the run on a bounded async queue (Redisson) for the scheduler
-        // worker to execute. Until then the trigger is accepted but performs no work.
-        log.info("Agent Insights run requested for project '{}' (no-op until OPIK-6853)", projectId);
+
+        Instant periodEnd = Instant.now();
+        reportPublisher.enqueue(job.projectId(), workspaceId, periodEnd.minus(TRIGGER_WINDOW), periodEnd)
+                .subscribe(
+                        reportId -> log.info("Enqueued Agent Insights run reportId='{}' for project '{}'",
+                                reportId, projectId),
+                        error -> log.error("Failed to enqueue Agent Insights run for project '{}'", projectId,
+                                error));
+    }
+
+    // Cross-workspace; used by the daily sweep (OPIK-6853), never from a request thread. The DAO's JOIN
+    // with projects already filters out jobs whose project was deleted.
+    public List<EnabledJob> findAllEnabled() {
+        return transactionTemplate.inTransaction(READ_ONLY,
+                handle -> handle.attach(AgentInsightsJobDAO.class).findAllEnabled());
     }
 }
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsReportMessage.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsReportMessage.java
@@ -1,0 +1,23 @@
+package com.comet.opik.domain;
+
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import lombok.Builder;
+import lombok.NonNull;
+
+import java.time.Instant;
+import java.util.UUID;
+
+/**
+ * A queued request to generate an Agent Insights report for a (workspace, project) over a time window.
+ * Published by {@link AgentInsightsReportPublisher} and consumed by the Agent Insights report subscriber,
+ * which performs the actual (bounded) trigger via {@link AgentInsightsReportClient}.
+ */
+@Builder(toBuilder = true)
+@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "@class")
+public record AgentInsightsReportMessage(
+        @NonNull String reportId,
+        @NonNull UUID projectId,
+        @NonNull String workspaceId,
+        @NonNull Instant periodStart,
+        @NonNull Instant periodEnd) {
+}
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsReportPublisher.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsReportPublisher.java
@@ -1,0 +1,82 @@
+package com.comet.opik.domain;
+
+import com.comet.opik.infrastructure.AgentInsightsReportConfig;
+import com.comet.opik.infrastructure.ServiceTogglesConfig;
+import com.comet.opik.infrastructure.redis.RedisStreamUtils;
+import jakarta.inject.Inject;
+import jakarta.inject.Singleton;
+import lombok.NonNull;
+import lombok.extern.slf4j.Slf4j;
+import org.redisson.api.RStreamReactive;
+import org.redisson.api.RedissonReactiveClient;
+import reactor.core.publisher.Mono;
+import reactor.core.scheduler.Schedulers;
+import ru.vyarus.dropwizard.guice.module.yaml.bind.Config;
+
+import java.time.Instant;
+import java.util.UUID;
+
+/**
+ * Publishes Agent Insights report-trigger requests onto a Redis stream so the report subscriber can
+ * execute them with bounded concurrency. Both the manual {@code /trigger} endpoint and the daily cron
+ * enqueue through here, so the actual (potentially slow) report call never runs on a request or sweep
+ * thread, and the consumer group caps the in-flight fan-out.
+ */
+@Slf4j
+@Singleton
+public class AgentInsightsReportPublisher {
+
+    private final @NonNull RedissonReactiveClient redisson;
+    private final @NonNull AgentInsightsReportConfig config;
+    private final @NonNull ServiceTogglesConfig serviceToggles;
+    private final @NonNull IdGenerator idGenerator;
+
+    @Inject
+    public AgentInsightsReportPublisher(@NonNull RedissonReactiveClient redisson,
+            @NonNull @Config("agentInsightsReport") AgentInsightsReportConfig config,
+            @NonNull @Config("serviceToggles") ServiceTogglesConfig serviceToggles,
+            @NonNull IdGenerator idGenerator) {
+        this.redisson = redisson;
+        this.config = config;
+        this.serviceToggles = serviceToggles;
+        this.idGenerator = idGenerator;
+    }
+
+    /**
+     * Enqueues a report-trigger request and returns the generated report id once the message is on the
+     * stream, or completes empty when publishing is disabled.
+     */
+    public Mono<String> enqueue(@NonNull UUID projectId, @NonNull String workspaceId,
+            @NonNull Instant periodStart, @NonNull Instant periodEnd) {
+
+        if (!serviceToggles.isAgentInsightsEnabled()) {
+            log.debug("Agent Insights is disabled, ignoring trigger for project '{}'", projectId);
+            return Mono.empty();
+        }
+
+        String reportId = idGenerator.generateId().toString();
+        var message = AgentInsightsReportMessage.builder()
+                .reportId(reportId)
+                .projectId(projectId)
+                .workspaceId(workspaceId)
+                .periodStart(periodStart)
+                .periodEnd(periodEnd)
+                .build();
+
+        // DEBUG: the daily sweep enqueues one per enabled project, so keep INFO for lifecycle events only.
+        log.debug("Publishing Agent Insights report trigger: reportId='{}', project='{}', workspace='{}'",
+                reportId, projectId, workspaceId);
+
+        return Mono.defer(() -> {
+            RStreamReactive<String, AgentInsightsReportMessage> stream = redisson.getStream(
+                    config.getStreamName(), config.getCodec());
+
+            return stream.add(RedisStreamUtils.buildAddArgs(
+                    AgentInsightsReportConfig.PAYLOAD_FIELD, message, config))
+                    .map(streamMessageId -> reportId)
+                    .doOnError(throwable -> log.error(
+                            "Failed to publish Agent Insights report trigger: reportId='{}', project='{}'",
+                            reportId, projectId, throwable));
+        }).subscribeOn(Schedulers.boundedElastic());
+    }
+}
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsTriggerRequest.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsTriggerRequest.java
@@ -1,0 +1,24 @@
+package com.comet.opik.domain;
+
+import com.fasterxml.jackson.databind.PropertyNamingStrategies;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import lombok.Builder;
+import lombok.NonNull;
+
+import java.time.Instant;
+import java.util.UUID;
+
+/**
+ * Payload for the Platform BE Ollie trigger endpoint ({@code POST /opik/ollie/generate-agent-insights}).
+ * Mirrors the platform's {@code AgentInsightsGenerateRequest} contract (snake_case); all fields required.
+ */
+@Builder(toBuilder = true)
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
+record AgentInsightsTriggerRequest(
+        @NonNull String reportType,
+        @NonNull String reportId,
+        @NonNull UUID projectId,
+        @NonNull String workspaceId,
+        @NonNull Instant periodStart,
+        @NonNull Instant periodEnd) {
+}
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/PlatformAgentInsightsReportClient.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/PlatformAgentInsightsReportClient.java
@@ -1,0 +1,63 @@
+package com.comet.opik.domain;
+
+import com.comet.opik.infrastructure.AgentInsightsReportConfig;
+import jakarta.inject.Inject;
+import jakarta.inject.Singleton;
+import jakarta.ws.rs.client.Client;
+import jakarta.ws.rs.client.Entity;
+import jakarta.ws.rs.core.MediaType;
+import jakarta.ws.rs.core.Response;
+import lombok.NonNull;
+import lombok.extern.slf4j.Slf4j;
+import ru.vyarus.dropwizard.guice.module.yaml.bind.Config;
+
+import java.time.Instant;
+import java.util.UUID;
+
+/**
+ * Real {@link AgentInsightsReportClient}: POSTs a trigger to the Platform BE Ollie endpoint
+ * ({@code POST /opik/ollie/generate-agent-insights}, OPIK-6854). The call is synchronous and throws on a
+ * non-2xx response; the subscriber treats a failure as a dropped run (at-most-once), it does not retry.
+ */
+@Slf4j
+@Singleton
+public class PlatformAgentInsightsReportClient implements AgentInsightsReportClient {
+
+    private static final String REPORT_TYPE = "agent_insights";
+
+    private final @NonNull Client httpClient;
+    private final @NonNull AgentInsightsReportConfig config;
+
+    @Inject
+    public PlatformAgentInsightsReportClient(@NonNull Client httpClient,
+            @NonNull @Config("agentInsightsReport") AgentInsightsReportConfig config) {
+        this.httpClient = httpClient;
+        this.config = config;
+    }
+
+    @Override
+    public void triggerAgentInsights(@NonNull String reportId, @NonNull UUID projectId,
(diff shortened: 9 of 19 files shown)


apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobService.java
```diff
@@ -109,16 +125,22 @@ private void triggerImmediate(AgentInsightsJob job, String workspaceName, String
             }
             Instant periodEnd = Instant.now();
             Instant periodStart = periodEnd.minus(ENABLE_TRIGGER_WINDOW);
-            agentInsightsReportClient.triggerAgentInsights(idGenerator.generateId().toString(),
-                    job.projectId(), projectName, workspaceName, periodStart, periodEnd);
-            transactionTemplate.inTransaction(WRITE, handle -> {
-                handle.attach(AgentInsightsJobDAO.class).markTriggered(job.id(), periodEnd);
-                return null;
-            });
+            doTrigger(job.id(), job.projectId(), job.workspaceId(), periodStart, periodEnd);
```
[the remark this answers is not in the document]
aadereiko: Fixed in ef6343b. The ad-hoc trigger no longer advances `last_triggered_at` — that marker is now written only by the cron path (`doTrigger`). An on-demand run therefore can no longer suppress the next scheduled previous-day report, so the once-per-day gate stays correct.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/AgentInsightsReportJob.java
```diff
@@ -0,0 +1,114 @@
...
+                    }
+                    return Mono.fromRunnable(() -> agentInsightsJobService.trigger(job, periodStart, periodEnd))
+                            .subscribeOn(Schedulers.boundedElastic())
+                            .then();
+                });
+    }
+
+    private static boolean alreadyTriggeredOn(EnabledJob job, LocalDate date) {
+        return job.lastTriggeredAt() != null
+                && job.lastTriggeredAt().atZone(ZoneOffset.UTC).toLocalDate().equals(date);
```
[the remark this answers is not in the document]
aadereiko: Fixed in ef6343b by decoupling the immediate run from the daily marker: `triggerImmediate` now calls a `sendTrigger` helper that does *not* persist `last_triggered_at`, while the cron `doTrigger` still does. The 00:05 cron is no longer skipped after a same-day manual run.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobService.java
```diff
@@ -91,47 +90,66 @@ public Mono<AgentInsightsJob> update(@NonNull UUID projectId, @NonNull AgentInsi
...
-                    .stream().findFirst().map(Project::name)
-                    .orElseThrow(() -> new NotFoundException("Project not found: " + projectId));
             AgentInsightsJob job = transactionTemplate.inTransaction(READ_ONLY,
                     handle -> handle.attach(AgentInsightsJobDAO.class)
                             .findByProject(workspaceId, projectId)
                             .orElseThrow(() -> new NotFoundException(
                                     "Agent insights job not found for project: " + projectId)));
-            triggerImmediate(job, workspaceName, projectName);
+            triggerImmediate(job, workspaceId);
             return null;
```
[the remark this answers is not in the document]
aadereiko: Restored in ef6343b — `triggerNow` now calls `projectService.validateProjectIdExists(projectId, workspaceId)` before scheduling, so an orphaned job (project deleted) returns 404 instead of firing.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/AgentInsightsReportJob.java
```diff
@@ -0,0 +1,114 @@
...
+        if (alreadyTriggeredOn(job, runDate)) {
+            log.info("Agent Insights job for project '{}' already triggered on {}, skipping", job.projectId(), runDate);
+            return Mono.empty();
+        }
+        return traceService.hasTracesInRange(job.workspaceId(), job.projectId(), periodStart, periodEnd)
+                .flatMap(hasTraces -> {
+                    if (Boolean.FALSE.equals(hasTraces)) {
+                        log.info("No traces in the previous day for project '{}', skipping", job.projectId());
+                        return Mono.empty();
+                    }
```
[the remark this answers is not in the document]
aadereiko: Fixed in ef6343b — the per-project no-trace case is now logged at `DEBUG`; `INFO` is reserved for lifecycle (job start/complete).

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/AgentInsightsReportJob.java
```diff
@@ -0,0 +1,114 @@
...
+        if (!agentInsightsReportClient.isEnabled()) {
+            log.debug("Agent Insights trigger not configured; skipping report job");
+            return;
+        }
+
+        LocalDate runDate = LocalDate.now(ZoneOffset.UTC);
+        Instant periodStart = runDate.minusDays(1).atStartOfDay(ZoneOffset.UTC).toInstant();
+        Instant periodEnd = runDate.atStartOfDay(ZoneOffset.UTC).toInstant();
+
+        log.info("Agent Insights report job running for previous UTC day [{}, {})", periodStart, periodEnd);
```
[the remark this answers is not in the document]
aadereiko: Fixed in ef6343b — `periodStart`/`periodEnd` and `runDate` are now single-quoted in the log placeholders per SKILL.md.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobService.java
```diff
@@ -63,12 +61,30 @@ public Mono<EnableResult> enable(@NonNull UUID projectId) {
...
+            log.info("Agent Insights trigger disabled; skipping run for project '{}'", job.projectId());
+            return;
+        }
+        if (projectService.findByIds(job.workspaceId(), Set.of(job.projectId())).isEmpty()) {
+            log.warn("Project '{}' not found; skipping Agent Insights run", job.projectId());
+            return;
+        }
+        doTrigger(job.id(), job.projectId(), job.workspaceId(), periodStart, periodEnd);
+    }
+
```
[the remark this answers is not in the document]
aadereiko: Coverage: the cron sweep branches (has-traces → trigger, no-traces → skip, already-triggered-today → skip, per-job failure isolation) are covered by `AgentInsightsReportJobTest`, and the enabled happy path end-to-end by `AgentInsightsJobTriggerTest`. The disabled-client and missing-project guards inside `trigger(...)` are simple early-returns mirrored by the manual-trigger path; happy to add a focused parameterized unit test if you prefer explicit coverage.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobService.java
```diff
@@ -23,10 +32,29 @@
...
+    // Superseded by the Redis-backed queue in OPIK-6854.
+    private final ExecutorService triggerExecutor = new ThreadPoolExecutor(
+            1, TRIGGER_POOL_MAX_THREADS, 60L, TimeUnit.SECONDS,
+            new LinkedBlockingQueue<>(TRIGGER_QUEUE_CAPACITY),
+            runnable -> {
+                var thread = new Thread(runnable, "agent-insights-trigger");
+                thread.setDaemon(true);
+                return thread;
+            },
+            new ThreadPoolExecutor.CallerRunsPolicy());
```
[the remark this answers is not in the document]
aadereiko: Resolved in 279bbc6474 by replacing the in-process `ThreadPoolExecutor` with a bounded **Redis-stream queue** (mirroring the webhook / dataset-export pattern). `/trigger` now publishes to the stream and returns `202` without ever running the report call on the request thread — no `CallerRunsPolicy`, no inline execution under saturation. A consumer group (`AgentInsightsReportSubscriber`, `consumerBatchSize`) caps concurrent runs, and the daily cron enqueues through the same path. The honest-`202`/backpressure concern (and Thiago's blocker) is now handled by the broker rather than the request thread.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobService.java
```diff
@@ -23,10 +32,29 @@
...
+    private static final int TRIGGER_QUEUE_CAPACITY = 100;
+
     private final @NonNull TransactionTemplate transactionTemplate;
     private final @NonNull IdGenerator idGenerator;
     private final @NonNull Provider<RequestContext> requestContext;
     private final @NonNull ProjectService projectService;
+    private final @NonNull AgentInsightsReportClient agentInsightsReportClient;
+
+    // Bounded fire-and-forget pool for manual triggers: caps concurrent report runs and applies
+    // backpressure (CallerRunsPolicy runs the task on the caller when the queue is full) instead of
```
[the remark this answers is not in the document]
aadereiko: Resolved in 279bbc6474 — the executor is gone; concurrency/back-pressure are now tunable via the new `agentInsightsReport` stream config (`consumerBatchSize`, `poolingInterval`, `longPollingDuration`, `maxRetries`, `pendingMessageDuration`, `streamMaxLen`, `streamTrimLimit`), surfaced through `OpikConfiguration` + `config.yml` with env-var overrides, same as `webhook`/`datasetExport`.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000084_add_last_triggered_at_to_agent_insights_jobs.sql
```diff
@@ -0,0 +1,9 @@
+--liquibase formatted sql
+--changeset aadereiko:000084_add_last_triggered_at_to_agent_insights_jobs
+--comment: Add last_triggered_at to agent_insights_jobs — the daily cron (OPIK-6853) records each run and
+--comment: uses it for the once-per-day guard.
+
+ALTER TABLE agent_insights_jobs
+    ADD COLUMN last_triggered_at TIMESTAMP(6) NULL AFTER status;
+
+--rollback ALTER TABLE agent_insights_jobs DROP COLUMN last_triggered_at;
```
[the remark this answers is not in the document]
aadereiko: Fixed in 279bbc6474 — added the required trailing blank line after `--rollback` per the Liquibase formatted SQL rule.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/AgentInsightsReportSubscriber.java
```diff
@@ -0,0 +1,92 @@
...
+                message.workspaceId(), message.periodStart(), message.periodEnd()))
+                .subscribeOn(Schedulers.boundedElastic())
+                .then()
+                .doOnSuccess(unused -> log.info("Triggered Agent Insights report: reportId='{}', project='{}'",
+                        message.reportId(), message.projectId()))
+                .onErrorResume(throwable -> {
+                    // Re-throw to leave the message un-acked so the base subscriber retries it.
+                    log.error("Failed to trigger Agent Insights report: reportId='{}', project='{}'",
+                            message.reportId(), message.projectId(), throwable);
+                    return Mono.error(throwable);
```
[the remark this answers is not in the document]
aadereiko: Fixed in f6a640139d. Switched the subscriber to **at-most-once**: `processEvent` now logs and **acks** on failure (returns `Mono.empty()`) instead of rethrowing, so a redelivery can no longer run the same report twice. The rationale: report generation is a non-idempotent side effect (Ollie compute + a user-facing report) and isn't deduplicated downstream, so duplicate generation is worse than dropping one run of a best-effort daily briefing. `reportId` is still carried on the message as the idempotency key if we later add downstream dedup and want to restore at-least-once. Added a unit test (`processEvent_acksOnTriggerFailure_atMostOnce`).

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceService.java
```diff
@@ -102,6 +102,8 @@ public interface TraceService {
 
     Mono<Map<UUID, Instant>> getLastUpdatedTraceAt(Set<UUID> projectIds, String workspaceId);
 
+    Mono<Boolean> hasTracesInRange(String workspaceId, UUID projectId, Instant from, Instant to);
+
```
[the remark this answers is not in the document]
aadereiko: Fixed in f6a640139d — added `@NonNull` to all four params of `TraceService.hasTracesInRange` (interface + impl), so a null argument fails fast at the service boundary instead of NPE-ing inside `Instant.toString()`.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsReportMessage.java
```diff
@@ -0,0 +1,23 @@
...
+ * which performs the actual (bounded) trigger via {@link AgentInsightsReportClient}.
+ */
+@Builder(toBuilder = true)
+@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "@class")
+public record AgentInsightsReportMessage(
+        @NotNull String reportId,
+        @NotNull UUID projectId,
+        @NotNull String workspaceId,
+        @NotNull Instant periodStart,
+        @NotNull Instant periodEnd) {
```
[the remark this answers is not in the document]
aadereiko: Fixed in f6a640139d — switched the `AgentInsightsReportMessage` fields from Jakarta `@NotNull` to Lombok `@NonNull`, so the null checks run in the canonical constructor (including on Jackson deserialization from Redis), per SKILL.md.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobDAO.java
```diff
@@ -39,4 +43,18 @@ int updateStatus(@Bind("workspaceId") String workspaceId,
             """)
     Optional<AgentInsightsJob> findByProject(@Bind("workspaceId") String workspaceId,
             @Bind("projectId") UUID projectId);
+
+    @SqlUpdate("""
+            UPDATE agent_insights_jobs SET last_triggered_at = :triggeredAt
+            WHERE id = :id
+            """)
+    void markTriggered(@Bind("id") UUID id, @Bind("triggeredAt") Instant triggeredAt);
```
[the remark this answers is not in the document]
aadereiko: Fixed in f6a640139d — `markTriggered` now also sets `last_updated_by = :userName` (passed `RequestContext.SYSTEM_USER` from the cron). `last_updated_at` already auto-bumps via `ON UPDATE CURRENT_TIMESTAMP(6)`, so the audit trail now consistently reflects the system actor + time of the nightly run.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000083_create_agent_insights_jobs_table.sql
```diff
@@ -0,0 +1,19 @@
...
+    created_at TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
+    created_by VARCHAR(255) NOT NULL DEFAULT 'admin',
+    last_updated_at TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6),
+    last_updated_by VARCHAR(255) NOT NULL DEFAULT 'admin',
+    CONSTRAINT agent_insights_jobs_pk PRIMARY KEY (id),
+    CONSTRAINT agent_insights_jobs_ws_project_uk UNIQUE (workspace_id, project_id)
+) ENGINE = InnoDB
+  DEFAULT CHARSET = utf8mb4;
+
+--rollback DROP TABLE IF EXISTS agent_insights_jobs;
```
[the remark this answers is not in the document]
aadereiko: This file (`000083_...`) is already merged to main via #7087 and isn't part of this PR's diff. Editing it now would change the Liquibase changeset checksum and trigger a checksum-mismatch on already-migrated deployments, so I'd rather not touch it here — happy to fix the trailing newline in a tiny standalone follow-up if you want it.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/AgentInsightsReportJob.java
```diff
@@ -0,0 +1,115 @@
...
+                .subscribeOn(Schedulers.boundedElastic())
+                .flatMapMany(Flux::fromIterable)
+                .concatMap(job -> processJob(job, runDate, periodStart, periodEnd)
+                        .onErrorResume(e -> {
+                            // Per-job isolation: one failing project must not abort the rest of the sweep.
+                            log.error("Failed to process Agent Insights job for project '{}'", job.projectId(), e);
+                            return Mono.empty();
+                        }))
+                .then();
+    }
```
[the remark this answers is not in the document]
aadereiko: Acknowledged. The sweep is intentionally serial (`concatMap`, one job at a time) and each check is now a cheap `SELECT 1 ... LIMIT 1` hitting the `(workspace_id, project_id)` primary-key index, run once per day — so for current install sizes it's bounded and not a thundering herd. Batching into a single grouped DAO query is a fair optimization but needs cross-workspace grouping (the sweep is cross-workspace), so I'd prefer to defer it to a follow-up rather than expand this PR. Can prioritize if we expect installs with thousands of enabled projects.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsReportPublisher.java
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * Enqueues a report-trigger request and returns the generated report id once the message is on the
+     * stream, or completes empty when publishing is disabled.
+     */
+    public Mono<String> enqueue(@NonNull UUID projectId, @NonNull String workspaceId,
+            @NonNull Instant periodStart, @NonNull Instant periodEnd) {
+
+        if (!config.isEnabled()) {
+            log.debug("Agent Insights report publishing is disabled, ignoring trigger for project '{}'", projectId);
```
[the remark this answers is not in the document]
aadereiko: Agreed there's shared shape with `WebhookPublisher` / `CsvDatasetExportService`. Extracting a common `RedisStreamPublisher` would touch two already-merged publishers, so I'd scope it as a dedicated refactor rather than fold unrelated changes into this PR — `AgentInsightsReportPublisher` deliberately follows the existing pattern so it's a clean extraction target later. Filed as a follow-up candidate.

apps/opik-backend/src/main/java/com/comet/opik/domain/PlatformAgentInsightsReportClient.java
```diff
@@ -0,0 +1,71 @@
...
+                "period_end", periodEnd.toString()));
+
+        try (Response response = httpClient.target(config.getTriggerUrl())
+                .request(MediaType.APPLICATION_JSON)
+                .post(Entity.json(payload))) {
+            if (response.getStatus() >= 300) {
+                // Throw so the subscriber leaves the message un-acked and retries.
+                throw new IllegalStateException(
+                        "Agent Insights trigger returned %d for report '%s'".formatted(response.getStatus(),
+                                reportId));
```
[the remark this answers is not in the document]
aadereiko: This is moot under the at-most-once design: `AgentInsightsReportSubscriber.processEvent` wraps the call in `onErrorResume(... -> Mono.empty())`, so **any** exception from `triggerAgentInsights` is logged and acked before it ever reaches `BaseRedisSubscriber.isRetryableException()`. The run is intentionally dropped on failure (not retried) — report generation is a non-idempotent side effect and isn't deduped downstream, so we prefer dropping a best-effort daily run over risking a duplicate (see the HIGH double-run thread above). Because the exception never drives retry here, the `InternalServerErrorException`/`BadRequestException` distinction has no effect; and those JAX-RS web-exception types are meant for inbound resource mapping, not an outbound client. Keeping `IllegalStateException` (logged, never surfaced as an HTTP response). If we later want at-least-once, the right lever is downstream idempotency on `reportId`, at which point I'd switch to the typed exceptions.

apps/opik-backend/config.yml
```diff
@@ -801,6 +801,46 @@ datasetExport:
   # Description: Batch size for cleanup job operations (number of jobs to process per batch). Must be between 10 and 1000.
   cleanupBatchSize: ${DATASET_EXPORT_CLEANUP_BATCH_SIZE:-100}
 
+# Configuration for the Agent Insights report trigger queue (manual /trigger + daily cron enqueue here;
+# a bounded Redis consumer group performs the actual report calls)
+agentInsightsReport:
+  # Default: false
+  # Description: Whether the Agent Insights report subscriber consumes and triggers queued reports
+  enabled: ${AGENT_INSIGHTS_REPORT_ENABLED:-false}
```
BorisTkachenko: Let's reuse existing toggle `TOGGLE_AGENT_INSIGHTS_ENABLED`, we don't need a second one for the same feature.

apps/opik-backend/src/test/java/com/comet/opik/domain/PlatformAgentInsightsReportClientTest.java
```diff
@@ -0,0 +1,93 @@
...
+import static com.github.tomakehurst.wiremock.client.WireMock.equalTo;
+import static com.github.tomakehurst.wiremock.client.WireMock.matchingJsonPath;
+import static com.github.tomakehurst.wiremock.client.WireMock.post;
+import static com.github.tomakehurst.wiremock.client.WireMock.postRequestedFor;
+import static com.github.tomakehurst.wiremock.client.WireMock.urlEqualTo;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class PlatformAgentInsightsReportClientTest {
```
BorisTkachenko: We don't need it, could/should be verified within `AgentInsightsJobsResourceTest`

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/AgentInsightsReportSubscriberTest.java
```diff
@@ -0,0 +1,88 @@
...
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+/**
+ * Unit tests for the Agent Insights report subscriber: lifecycle gating and the
+ * client-disabled vs client-enabled branches of {@code processEvent}.
+ */
+@ExtendWith(MockitoExtension.class)
+class AgentInsightsReportSubscriberTest {
```
BorisTkachenko: We don't need it, could/should be verified within AgentInsightsJobsResourceTest

apps/opik-backend/config.yml
```diff
@@ -1307,6 +1347,9 @@ jobTimeout:
   # Default: 30s
   # Description: Timeout for the trace threads closing job execution
   traceThreadsClosingJobTimeout: ${OPIK_TRACE_THREADS_CLOSING_JOB_TIMEOUT:-30}
+  # Default: 300s
+  # Description: Timeout for the daily Agent Insights report job sweep
+  agentInsightsReportJobTimeout: ${OPIK_AGENT_INSIGHTS_REPORT_JOB_TIMEOUT:-300}
```
BorisTkachenko: Let's move this to `agentInsightsReport:` block

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/AgentInsightsReportJob.java
```diff
@@ -0,0 +1,115 @@
...
+        // holdUntilExpiry: only one replica runs the sweep per day (idempotent across replicas).
+        lockService.bestEffortLock(
+                JOB_LOCK,
+                runSweep(runDate, periodStart, periodEnd),
+                Mono.defer(() -> {
+                    log.debug("Could not acquire lock for Agent Insights report job, another instance is running");
+                    return Mono.empty();
+                }),
+                Duration.ofSeconds(config.getJobTimeout().getAgentInsightsReportJobTimeout()),
+                Duration.ofSeconds(5),
```
BorisTkachenko: Should be configurable as well

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000084_add_last_triggered_at_to_agent_insights_jobs.sql
```diff
@@ -0,0 +1,10 @@
+--liquibase formatted sql
+--changeset aadereiko:000084_add_last_triggered_at_to_agent_insights_jobs
+--comment: Add last_triggered_at to agent_insights_jobs — the daily cron (OPIK-6853) records each run and
+--comment: uses it for the once-per-day guard.
+
+ALTER TABLE agent_insights_jobs
+    ADD COLUMN last_triggered_at TIMESTAMP(6) NULL AFTER status;
```
BorisTkachenko: We don't need this and all related logic. Our cron job is configured to run once per day. On top of it, we have a locking mechanism to start it only on the one pod to make sure we don't have overlapping triggers for the same day. Let's remove this additional excessive logic.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobService.java
```diff
@@ -72,16 +80,57 @@ public AgentInsightsJob update(@NonNull UUID projectId, @NonNull AgentInsightsJo
...
+     * Enqueues a report run for an enabled job over the given period and advances the daily marker. Used
+     * by the nightly cron (OPIK-6853) after its previous-day trace check. A missing project simply skips
+     * (logged); the cron catches per-job failures so one bad project doesn't abort the sweep. The marker
+     * is advanced only when a message was actually enqueued (i.e. publishing is enabled), so a disabled
+     * queue doesn't make the cron believe the day was already covered.
+     */
+    public void trigger(@NonNull EnabledJob job, @NonNull Instant periodStart, @NonNull Instant periodEnd) {
+        if (projectService.findByIds(job.workspaceId(), Set.of(job.projectId())).isEmpty()) {
+            log.warn("Project '{}' not found; skipping Agent Insights run", job.projectId());
+            return;
```
BorisTkachenko: This approach, you will be checking the project existence for each enabled job. Basically, you will be calling DB in a loop. Instead add this condition to `AgentInsightsJobDAO::findAllEnabled` by join with `projects` table.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobService.java
```diff
@@ -72,16 +80,57 @@ public AgentInsightsJob update(@NonNull UUID projectId, @NonNull AgentInsightsJo
...
+        transactionTemplate.inTransaction(WRITE, handle -> {
+            handle.attach(AgentInsightsJobDAO.class).markTriggered(job.id(), periodEnd, RequestContext.SYSTEM_USER);
+            return null;
+        });
+    }
+
+    // Cross-workspace; used by the scheduler (OPIK-6853), never from a request thread.
+    public List<EnabledJob> findAllEnabled() {
+        return transactionTemplate.inTransaction(READ_ONLY,
+                handle -> handle.attach(AgentInsightsJobDAO.class).findAllEnabled());
```
BorisTkachenko, line 108: Here we also need to check whether the traces are available. We need to do it in one with one DB call for a set of workspace_id/project_id to get traces count per period.

Since now you are doing it later per each project, which makes a lot of unnecessary DB calls, basically in a loop.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceService.java
```diff
@@ -102,6 +102,9 @@ public interface TraceService {
 
     Mono<Map<UUID, Instant>> getLastUpdatedTraceAt(Set<UUID> projectIds, String workspaceId);
 
+    Mono<Boolean> hasTracesInRange(@NonNull String workspaceId, @NonNull UUID projectId, @NonNull Instant from,
```
BorisTkachenko: As mentioned in the other comment, this should be done not per project workspace but for a set. So we make one call instead of calling DB in a loop.

BorisTkachenko (changes requested):
Thank you, I left some initial comments, let's address those first. In general looks good.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/AgentInsightsReportJob.java
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * Once-per-day cron (00:05 UTC — after midnight so the previous UTC day is closed) that triggers Agent
+ * Insights report generation for enabled projects that had traces during the previous UTC day. Mirrors
+ * {@code OllieDailyReportJob} (distributed lock, per-item failure isolation), but daily rather than a
+ * 10-minute schedule_time sweep.
+ */
+@Slf4j
+@Singleton
+@DisallowConcurrentExecution
+@On(value = "0 5 0 * * ?", timeZone = "UTC")
```
BorisTkachenko: Let's try to make it configurable via env var.
You can refer to example:
```
    private void setMetricsAlertJob() {
        var webhookConfig = injector.get().getInstance(OpikConfiguration.class).getWebhook();

        if (webhookConfig == null || webhookConfig.getMetrics() == null) {
            log.warn("Webhook metrics configuration not found, skipping metrics alert job setup");
            return;
        }

        scheduleRepeatingJob(MetricsAlertJob.class,
                webhookConfig.getMetrics().getFixedDelay().toJavaDuration(),
                webhookConfig.getMetrics().getInitialDelay().toJavaDuration());
    }
```

apps/opik-backend/config.yml
aadereiko: Done — dropped the separate flag and reuse `TOGGLE_AGENT_INSIGHTS_ENABLED` (`serviceToggles.agentInsightsEnabled`). The publisher, subscriber, and cron scheduling all gate on it now.

apps/opik-backend/config.yml
aadereiko: Done — moved the timeout into the `agentInsightsReport` block (`jobTimeout`, default 5m) and removed `agentInsightsReportJobTimeout` from `jobTimeout`.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/AgentInsightsReportJob.java
aadereiko: Done — added `agentInsightsReport.lockWaitTime` (default 5s); the lock acquisition uses it instead of the hardcoded value.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/AgentInsightsReportJob.java
aadereiko: Done — removed the hardcoded `@On` and schedule the cron programmatically in `OpikGuiceyLifecycleEventListener` (new `setAgentInsightsReportJob()` + `scheduleCronJob` helper), reading the cron from `agentInsightsReport.schedule` (env `AGENT_INSIGHTS_REPORT_SCHEDULE`), gated on the toggle — same pattern as the jobs you referenced.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000084_add_last_triggered_at_to_agent_insights_jobs.sql
aadereiko: Done — removed `last_triggered_at` entirely: the column/migration, `markTriggered`, and the `alreadyTriggeredOn` once-per-day guard are all gone. The daily schedule + the distributed single-pod lock are what prevent overlap, as you said. (Migration `000084` now only aligns the table collation — see below.)

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobService.java
aadereiko: Done — `findAllEnabled` now `INNER JOIN`s `projects` (on id + workspace_id), so jobs whose project was deleted are filtered at the source. The per-job `projectService.findByIds(...)` existence check in the sweep loop is gone.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsJobService.java
aadereiko: Done — the trace gate is now a single set-based query per workspace. The sweep groups enabled jobs by workspace and calls `TraceService.getProjectsWithTracesInRange(workspaceId, projectIds, from, to)` once per workspace (returns the set of project_ids that had traces), then enqueues only those — no per-project DB call in a loop.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceService.java
aadereiko: Done — `TraceService.hasTracesInRange(workspaceId, projectId, ...)` is replaced by `getProjectsWithTracesInRange(workspaceId, Set<projectId>, ...)` returning `Mono<Set<UUID>>` from one `SELECT DISTINCT project_id ... WHERE project_id IN :ids` query.

apps/opik-backend/src/test/java/com/comet/opik/domain/PlatformAgentInsightsReportClientTest.java
aadereiko: Removed `PlatformAgentInsightsReportClientTest`. The trigger path (publish -> subscribe -> client) is covered end-to-end by `AgentInsightsJobTriggerTest`.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/AgentInsightsReportSubscriberTest.java
aadereiko: Removed `AgentInsightsReportSubscriberTest`. Subscriber behavior is covered via `AgentInsightsJobTriggerTest` (real publish -> consume -> client).

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/AgentInsightsReportJob.java
```diff
@@ -0,0 +1,104 @@
...
+    private Mono<Void> processWorkspace(String workspaceId, List<EnabledJob> jobs, Instant periodStart,
+            Instant periodEnd) {
+        Set<UUID> projectIds = jobs.stream().map(EnabledJob::projectId).collect(Collectors.toSet());
+        return traceService.getProjectsWithTracesInRange(workspaceId, projectIds, periodStart, periodEnd)
+                .flatMapMany(withTraces -> Flux.fromIterable(jobs)
+                        .filter(job -> withTraces.contains(job.projectId())))
+                .concatMap(job -> reportPublisher
+                        .enqueue(job.projectId(), job.workspaceId(), periodStart, periodEnd)
+                        .then())
+                .then();
```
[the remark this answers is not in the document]
aadereiko: Fixed — added a per-job `onErrorResume` inside the `concatMap` in `processWorkspace`, so a failed `enqueue` for one project logs and continues instead of aborting the rest of the workspace. The workspace-level `onErrorResume` still covers the trace-query failure. Added a unit test (`runSweep__perJobEnqueueFailureIsolated`).

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/AgentInsightsReportJob.java
```diff
@@ -0,0 +1,104 @@
...
+
+    // Visible for testing: per-workspace batched sweep. One trace query per workspace (no per-project
+    // loop), and one failing workspace must not abort the rest.
+    @VisibleForTesting
+    public Mono<Void> runSweep(Instant periodStart, Instant periodEnd) {
+        return Mono.fromCallable(agentInsightsJobService::findAllEnabled)
+                .subscribeOn(Schedulers.boundedElastic())
+                .flatMapMany(jobs -> Flux.fromIterable(
+                        jobs.stream().collect(Collectors.groupingBy(EnabledJob::workspaceId)).entrySet()))
+                .concatMap(entry -> processWorkspace(entry.getKey(), entry.getValue(), periodStart, periodEnd)
```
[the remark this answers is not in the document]
aadereiko: Intentionally not adding the marker back. This was removed in this same review at @BorisTkachenko's request: the cron runs once per day under a distributed best-effort lock (`agent_insights_report_job:lock`, held until expiry), so only one pod runs the sweep per day and same-day reruns don't occur in normal operation — the `last_triggered_at`/`markTriggered` guard was redundant with that. If we later want strict cross-run idempotency, the right place is downstream dedup on `reportId` (already carried on the message). Flagging the Baz-vs-review tension here so it's a conscious call.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsReportPublisher.java
```diff
@@ -0,0 +1,81 @@
...
+                .build();
+
+        log.info("Publishing Agent Insights report trigger: reportId='{}', project='{}', workspace='{}'",
+                reportId, projectId, workspaceId);
+
+        return Mono.defer(() -> {
+            RStreamReactive<String, AgentInsightsReportMessage> stream = redisson.getStream(
+                    config.getStreamName(), config.getCodec());
+
+            return stream.add(RedisStreamUtils.buildAddArgs(
```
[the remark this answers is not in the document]
aadereiko: Agreed there's shared shape with `WebhookPublisher` / `CsvDatasetExportService`, but extracting a common `RedisStreamPublisher` touches two already-merged publishers, so I'd scope it as a dedicated refactor rather than fold unrelated changes into this PR. `AgentInsightsReportPublisher` follows the existing pattern so it stays a clean extraction target. Tracking as a follow-up.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsReportPublisher.java
```diff
@@ -0,0 +1,81 @@
...
+        var message = AgentInsightsReportMessage.builder()
+                .reportId(reportId)
+                .projectId(projectId)
+                .workspaceId(workspaceId)
+                .periodStart(periodStart)
+                .periodEnd(periodEnd)
+                .build();
+
+        log.info("Publishing Agent Insights report trigger: reportId='{}', project='{}', workspace='{}'",
+                reportId, projectId, workspaceId);
```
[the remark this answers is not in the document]
aadereiko: Fixed — lowered the per-trigger "Publishing Agent Insights report trigger" log in `AgentInsightsReportPublisher.enqueue` from INFO to DEBUG, so the daily sweep no longer floods INFO. INFO stays for lifecycle/outcome events.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000084_fix_agent_insights_jobs_collation.sql
```diff
@@ -0,0 +1,10 @@
+--liquibase formatted sql
+--changeset aadereiko:000084_fix_agent_insights_jobs_collation
+--comment: 000083 created agent_insights_jobs with the utf8mb4 default collation (utf8mb4_0900_ai_ci), which
+--comment: diverges from the rest of the schema (utf8mb4_unicode_ci) and breaks JOINs against projects with an
+--comment: "illegal mix of collations" error. Align the table with the schema-wide collation.
+
+ALTER TABLE agent_insights_jobs CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
+
+--rollback empty: collation alignment is forward-only and safe to leave in place.
```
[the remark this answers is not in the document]
aadereiko: Fixed — added a real rollback to `000084`: `ALTER TABLE agent_insights_jobs CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_ai_ci;` (reverts to the pre-migration collation), with the required trailing blank line.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsReportPublisher.java
```diff
@@ -0,0 +1,81 @@
...
+
+        if (!serviceToggles.isAgentInsightsEnabled()) {
+            log.debug("Agent Insights is disabled, ignoring trigger for project '{}'", projectId);
+            return Mono.empty();
+        }
+
+        String reportId = idGenerator.generateId().toString();
+        var message = AgentInsightsReportMessage.builder()
+                .reportId(reportId)
+                .projectId(projectId)
```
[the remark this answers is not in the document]
aadereiko: Same call as the other two threads on this — deferring the shared `RedisStreamPublisher` extraction to a dedicated refactor rather than touching the two already-merged publishers (`WebhookPublisher`, `CsvDatasetExportService`) in this PR. `AgentInsightsReportPublisher` follows the existing pattern so it stays a clean extraction target. Tracked as a follow-up.

apps/opik-backend/config.yml
```diff
@@ -801,6 +801,52 @@ datasetExport:
   # Description: Batch size for cleanup job operations (number of jobs to process per batch). Must be between 10 and 1000.
   cleanupBatchSize: ${DATASET_EXPORT_CLEANUP_BATCH_SIZE:-100}
 
+# Configuration for the Agent Insights report trigger queue (manual /trigger + daily cron enqueue here;
+# a bounded Redis consumer group performs the actual report calls)
+agentInsightsReport:
+  # Default: "" (blank disables the real trigger client — the subscriber then no-ops)
+  # Description: Platform BE endpoint that triggers Agent Insights report generation on the Ollie pod
+  triggerUrl: ${AGENT_INSIGHTS_REPORT_TRIGGER_URL:-""}
```
BorisTkachenko: Let's not have it as empty, as it will add more troubles for the deployment team. For example, our authentication service, which also reaches EM, uses something like this: 
```
url: ${REACT_SERVICE_URL:-http://react-svc:8080}
```

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/AgentInsightsReportSubscriber.java
```diff
@@ -0,0 +1,100 @@
...
+    @Override
+    protected Mono<Void> processEvent(@NonNull AgentInsightsReportMessage message) {
+        log.info("Processing Agent Insights report trigger: reportId='{}', project='{}', workspace='{}'",
+                message.reportId(), message.projectId(), message.workspaceId());
+
+        if (!reportClient.isEnabled()) {
+            log.info("Agent Insights report client disabled; acking reportId='{}' without triggering",
+                    message.reportId());
+            return Mono.empty();
+        }
```
BorisTkachenko: As mentioned, it would be better to have a default value in the config. And then this check will not be necessary at all. Once we switch on the toggle, we expect that everything is in place. And toggle is already checked in this subscriber. Seems like over-engineering.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/AgentInsightsReportSubscriber.java
```diff
@@ -0,0 +1,100 @@
...
+    private static final String METRICS_BASE_NAME = "agent_insights_report";
+
+    private final AgentInsightsReportConfig config;
+    private final ServiceTogglesConfig serviceToggles;
+    private final AgentInsightsReportClient reportClient;
+
+    @Inject
+    public AgentInsightsReportSubscriber(
+            @NonNull @Config("agentInsightsReport") AgentInsightsReportConfig config,
+            @NonNull ServiceTogglesConfig serviceToggles,
```
BorisTkachenko:
```suggestion
            @NonNull @Config("serviceToggles") ServiceTogglesConfig serviceToggles,
```

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -114,6 +115,9 @@ public interface TraceDAO {
 
     Mono<Map<UUID, Instant>> getLastUpdatedTraceAt(Set<UUID> projectIds, String workspaceId, Connection connection);
 
+    Mono<Set<UUID>> getProjectsWithTracesInRange(String workspaceId, Set<UUID> projectIds, Instant from, Instant to,
+            Connection connection);
+
```
BorisTkachenko: With this we still need to process it per workspace. Meaning, calling this in a loop for each workspace.

What I have in mind is that we can call it only once. And do it in `AgentInsightsJobService::findAllEnabled`. Once we have all enabled jobs, we filter those by checking traces. And later on, we don't need all the related logic. We need to make sure workspace ID is present in the query to limit it. A query could be something like:
```
SELECT
    workspace_id,
    project_id,
    count() AS cnt
FROM traces
WHERE (workspace_id, project_id) IN (
    ('ws1', 'p1'),
    ('ws2', 'p2')
)
GROUP BY
    workspace_id,
    project_id
```
Also you will still need a time range in the WHERE clause. 

And you can pass it as 
```
List<Pair> pairs = List.of(
    new Pair("ws1", "p1"),
    new Pair("ws2", "p2")
);
```
It will return only workspace project pairs with count more than zero.

BorisTkachenko (commented):
Thank you for addressing previous comments, much better now. Let's address a few more before merging.

apps/opik-backend/config.yml
aadereiko: Done — gave `triggerUrl` a non-empty default pointing at the react service, mirroring the auth service: `${AGENT_INSIGHTS_REPORT_TRIGGER_URL:-http://react-svc:8080/opik/ollie/generate-agent-insights}` (and the same default on the config POJO).

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/AgentInsightsReportSubscriber.java
aadereiko: Done — removed the blank-URL `isEnabled()` gating entirely. The client no longer has `isEnabled()`, and the subscriber no longer checks it; with a config default + the toggle already gating the subscriber, it always triggers when invoked. Less code, no over-engineering.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/AgentInsightsReportSubscriber.java
aadereiko: Applied — `@Config("serviceToggles") ServiceTogglesConfig serviceToggles` in the subscriber (and the publisher, for consistency).

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
aadereiko: Done — it is now a single query for the whole sweep, no per-workspace loop. `findAllEnabled` returns all enabled jobs (INNER JOIN projects), then one `getProjectsWithTracesInRange(workspaceIds, projectIds, from, to)` returns the project_ids that had traces in the window; the cron filters to those and enqueues. I kept it as bound array params — `WHERE workspace_id IN :workspace_ids AND project_id IN :project_ids AND created_at >= :from AND created_at < :to` — rather than a rendered `(ws, project)` tuple literal: it is one query, keeps workspace_id in the WHERE for partition pruning, and is exact because project_id is globally unique (a project_id only has traces under its own workspace, so the IN-set form can't cross tenants). Happy to switch to the explicit tuple-IN if you prefer it.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -1949,6 +1947,17 @@ ORDER BY (workspace_id, project_id, id) DESC, last_updated_at DESC
             SETTINGS log_comment = '<log_comment>'
             ;
             """;
+
+    private static final String SELECT_PROJECTS_WITH_TRACES_IN_RANGE = """
+            SELECT DISTINCT project_id
+            FROM traces
+            WHERE workspace_id IN :workspace_ids
+            AND project_id IN :project_ids
```
BorisTkachenko: Let's instead try to have it as:
```
WHERE (workspace_id, project_id) IN (
    ('ws1', 'p1'),
    ('ws2', 'p2')
)
```
And we can pass it as list of pairs, not two sets:
```
List<Pair> pairs = List.of(
    new Pair("ws1", "p1"),
    new Pair("ws2", "p2")
);
```

Because now this condition is not strictly correct. Basically, we don't force that Workspace ID corresponds to Project ID. It will select everything with these workspaces and projects, even if a project belongs to a Different Workspace workspace from that list.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/AgentInsightsReportJob.java
```diff
@@ -0,0 +1,103 @@
...
+    @VisibleForTesting
+    public Mono<Void> runSweep(Instant periodStart, Instant periodEnd) {
+        return Mono.fromCallable(agentInsightsJobService::findAllEnabled)
+                .subscribeOn(Schedulers.boundedElastic())
+                .flatMap(jobs -> {
+                    if (jobs.isEmpty()) {
+                        return Mono.empty();
+                    }
+                    Set<String> workspaceIds = jobs.stream().map(EnabledJob::workspaceId).collect(Collectors.toSet());
+                    Set<UUID> projectIds = jobs.stream().map(EnabledJob::projectId).collect(Collectors.toSet());
```
BorisTkachenko: Mentioned in the other place, here we probably want to have a set or list of pairs, not two different sets.

apps/opik-backend/src/main/java/com/comet/opik/domain/PlatformAgentInsightsReportClient.java
```diff
@@ -0,0 +1,64 @@
...
+            @NonNull String workspaceId, @NonNull Instant periodStart, @NonNull Instant periodEnd) {
+
+        // Platform contract (AgentInsightsGenerateRequest): snake_case, all fields NotBlank.
+        var payload = Map.of(
+                "report_type", REPORT_TYPE,
+                "report_id", reportId,
+                "project_id", projectId.toString(),
+                "workspace_id", workspaceId,
+                "period_start", periodStart.toString(),
+                "period_end", periodEnd.toString());
```
BorisTkachenko: Let's have a named object since we know the structure of a request payload instead of a generic map.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/AgentInsightsJobTriggerTest.java
```diff
@@ -0,0 +1,187 @@
...
+import static org.awaitility.Awaitility.await;
+
+/**
+ * Exercises the enabled trigger branch of enable() — the no-op default client (used by the other
+ * test classes) never fires, so here we override the {@link AgentInsightsReportClient} binding with a
+ * recording stub whose {@code isEnabled()} returns true.
+ */
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@ExtendWith(DropwizardAppExtensionProvider.class)
+class AgentInsightsJobTriggerTest {
```
BorisTkachenko: Put these two tests into existing `AgentInsightsJobsResourceTest`:

- `trigger__firesRunOnce` - Belongs there, as you are basically testing an endpoint;
- It leaves here only `cronSweep__triggersJobsWithTraces`, But it is not worth creating a dedicated integration test class for only one test, we are trying to minimize those.

BorisTkachenko (commented):
Almost there!

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
aadereiko: Done — switched to the exact tuple-IN you described: `WHERE (workspace_id, project_id) IN (('ws','p'),...)`, passed as `Collection<Pair<String, UUID>>` (one query for the whole sweep). The pairs are rendered as a ClickHouse IN-list literal (same approach as the existing `<uuids_literal>` rendering in the DAOs). This is strictly correct now — it enforces the ws↔project correspondence, no cross-workspace bleed.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/AgentInsightsReportJob.java
aadereiko: Done — the sweep now builds `List<Pair<workspaceId, projectId>>` from the enabled jobs and passes that single collection to `getProjectsWithTracesInRange` (no two-set form).

apps/opik-backend/src/main/java/com/comet/opik/domain/PlatformAgentInsightsReportClient.java
aadereiko: Done — replaced the generic `Map` with a typed `AgentInsightsTriggerRequest` record (`@JsonNaming` snake_case) matching the platform's `AgentInsightsGenerateRequest` contract.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/AgentInsightsJobTriggerTest.java
aadereiko: Done — moved both into `AgentInsightsJobsResourceTest` and deleted `AgentInsightsJobTriggerTest`. `trigger__firesRunOnce` (endpoint test) and `cronSweep__triggersJobsWithTraces` now live there alongside the CRUD tests (the class binds a recording client + enables the toggle). 9 tests, all green.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -4105,6 +4121,34 @@ public Mono<Map<UUID, Instant>> getLastUpdatedTraceAt(
...
+                .collect(Collectors.toSet());
+    }
+
+    // Renders the (workspace_id, project_id) tuples as a ClickHouse IN list, e.g. ('ws','proj'),('ws2','proj2').
+    // Values are internal identifiers (workspace ids and UUID project ids), consistent with the existing
+    // literal-list rendering used elsewhere in this DAO.
+    private static String toPairsLiteral(Collection<Pair<String, UUID>> pairs) {
+        return pairs.stream()
+                .map(pair -> "('%s','%s')".formatted(pair.getLeft(), pair.getRight()))
+                .collect(Collectors.joining(","));
```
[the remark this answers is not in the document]
aadereiko: Good catch — fixed. `toPairsLiteral` now escapes single quotes (doubled, `'` -> `''`) on both values before interpolating, matching the existing ClickHouse literal escaping in `FilterUtils`. project_id is a `UUID` (hex + hyphens, no special chars), and a quote in workspace_id can no longer close/reshape the tuple literal. (Binding a tuple list isn't supported by the r2dbc ClickHouse driver, hence the escaped literal — same pattern as the existing `<uuids_literal>` rendering.)

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsTriggerRequest.java
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Payload for the Platform BE Ollie trigger endpoint ({@code POST /opik/ollie/generate-agent-insights}).
+ * Mirrors the platform's {@code AgentInsightsGenerateRequest} contract (snake_case).
+ */
+@Builder
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
+record AgentInsightsTriggerRequest(
+        String reportType,
+        String reportId,
```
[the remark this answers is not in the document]
aadereiko: Done — `AgentInsightsTriggerRequest` now uses `@Builder(toBuilder = true)` and `@NonNull` on all fields, so the canonical constructor rejects nulls before anything is serialized (consistent with `AgentInsightsReportMessage`).

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/AgentInsightsReportConfig.java
```diff
@@ -0,0 +1,83 @@
...
+
+@Data
+public class AgentInsightsReportConfig implements StreamConfiguration {
+
+    public static final String PAYLOAD_FIELD = "message";
+
+    // Platform endpoint that triggers report generation on the Ollie pod (OPIK-6854); defaults to the
+    // react service, like the auth service's reactService URL.
+    @Valid @NotBlank @JsonProperty
+    private String triggerUrl = "http://react-svc:8080/opik/ollie/generate-agent-insights";
```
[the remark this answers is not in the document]
aadereiko: This is intentional now — the "self-disable when blank" behavior was deliberately removed earlier in this review at @BorisTkachenko's request (the blank/`isEnabled()` gating was flagged as over-engineering). The current contract: `triggerUrl` is **required** (`@NotBlank`, defaulting to the react service), so a blank value is a startup misconfiguration that fails fast via Bean Validation — i.e. blank is an error, which matches your second suggested option. The disable switch is the `TOGGLE_AGENT_INSIGHTS_ENABLED` feature toggle (gates the subscriber, publisher, and cron scheduling), not a blank URL. I also removed the stale "blank disables / no-op" wording from the config + client docs, so nothing claims the unreachable behavior anymore. No code change here by design.

BorisTkachenko (approved):
Good job!

PR #7111, opened 2026-06-15
[OPIK-6814] [BE] feat: add internal endpoint to run Agent Insights free-form SQL

BorisTkachenko:
## Details

Implements the internal, authenticated endpoint that runs Ollie-generated read-only SQL
against ClickHouse, bounded to a single workspace/project, for the Agent Insights subagent.
Gated behind the `agentInsightsEnabled` toggle (introduced in OPIK-6850); when off the
endpoint returns `501 Not Implemented` and performs no ClickHouse access.

**Contract** — `POST /v1/internal/analytics-queries-executor`
- Body: `{ "project_id": "<uuid>", "query": "<sql>" }`
- `200`: `{ "results": [ { ... }, ... ] }`
- `501`: when `agentInsightsEnabled` is off
- Internal-only and authenticated: `workspace_id` is derived from auth (not the body),
  `project_id` comes from the body. `AuthFilter` now authenticates this single internal path
  (other `/v1/internal/*` endpoints remain unauthenticated).
- Caller contract: the final query must return exactly one column named `result`, JSON via
  `toJSONString(...)`.

**Execution flow & hardening**
- Connects as `comet_readonly_freeform_sql_user` via the `readOnlyFreeFormSqlClickHouseClient`
  (client-v2). Querying is delegated to a reactive `FreeFormSqlQueryDAO`.
- **EXPLAIN AST preflight**: each query is parsed via `EXPLAIN AST` on the read-only client;
  if any node label starts with `Set` (catches top-level / subquery / CTE / UNION-arm /
  `FORMAT … SETTINGS`) the query is rejected before execution. A parse failure is also a hard
  reject — execution never runs on a validation failure.
- **Bounds via server settings, never string concat**: `SQL_workspace_id` / `SQL_project_id`
  are passed through `QuerySettings().serverSetting(...)` (HTTP URL params); the caller SQL is
  never modified. The restrictive row policies (OPIK-6846) enforce isolation, and the profile
  defaults both settings to `''` so an unset value fails closed (zero rows).
- **Single-statement mode**: statement-stacking is rejected by ClickHouse.
- ClickHouse limit breaches (timeout, result/rows-read overflow, memory) and permission denials
  are surfaced as a clean `400`, not a `500`.

**Monitoring** — the query executor is instrumented with OpenTelemetry metrics following the
`AlertProjectMigrationService` pattern, namespace `opik.free_form_sql.queries`:
- counter `.executed` tagged `result` (success / rejected / error)
- counter `.rejected` tagged `reason` (ast_set_node / parse_error / permission_denied / ch_limit / other)
- histogram `.duration` (ms) tagged by `result`
- histograms `.result_rows` and `.bytes_read` for successful queries

> Note: the service/DAO are named `FreeFormSqlQuery*` and the metric namespace is
> `opik.free_form_sql.queries` (kept feature-agnostic for future reuse beyond Agent Insights).
> Dashboard queries in OPIK-6851 should target this namespace.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues
- OPIK-6814

## Testing

- `AnalyticsQueriesExecutorResourceTest` (toggle on) — provisions a throwaway
  `comet_readonly_freeform_sql_user` in the test ClickHouse container replicating the OPIK-6846
  DDL (`readonly=1` profile, table grants, restrictive row policies bound to the `SQL_*`
  settings), seeds two tenants via the public Trace API, and asserts:
  - a query bound to workspace/project A returns only A's rows;
  - a project from another workspace, and a workspace querying another workspace's project, both
    return zero rows (cross-tenant isolation end-to-end);
  - SETTINGS in top-level / subquery / CTE form are rejected by the AST guard;
  - unparseable SQL, statement-stacking, write statements, and non-granted tables are rejected.
- `AnalyticsQueriesExecutorResourceDisabledTest` (toggle off) — the endpoint returns `501`.
- The test ClickHouse container config (`clickhouse.xml`) now enables the `SQL_` custom-settings
  prefix so the profile/row-policy DDL can be replicated locally.

## Documentation
N/A (internal endpoint; caller query contract documented in the Swagger operation and the
resource Javadoc).


apps/opik-backend/src/main/java/com/comet/opik/api/AnalyticsQueryRequest.java
```diff
@@ -0,0 +1,15 @@
...
+import io.swagger.v3.oas.annotations.media.Schema;
+import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.NotNull;
+
+import java.util.UUID;
+
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
+public record AnalyticsQueryRequest(
+        @NotNull UUID projectId,
+        @Schema(description = "Read-only ClickHouse SQL. Must return exactly one column named `result` produced via toJSONString(...)") @NotBlank String query) {
```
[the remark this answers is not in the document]
BorisTkachenko: Fixed in da2aaeccce — added `@Builder(toBuilder = true)` to `AnalyticsQueryRequest` and switched `AnalyticsQueriesExecutorClient` to `AnalyticsQueryRequest.builder().projectId(...).query(...).build()`.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/internal/AnalyticsQueriesResource.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @POST
+    @Operation(operationId = "executeAnalyticsQuery", summary = "Execute Agent Insights free-form SQL", description = "Runs Ollie-generated read-only SQL bounded to the caller's workspace and the requested project. Returns 501 when the Agent Insights toggle is off.", responses = {
+            @ApiResponse(responseCode = "200", description = "Query results", content = @Content(schema = @Schema(implementation = AnalyticsQueryResponse.class))),
+            @ApiResponse(responseCode = "400", description = "Bad Request", content = @Content(schema = @Schema(implementation = ErrorMessage.class))),
+            @ApiResponse(responseCode = "422", description = "Unprocessable Content", content = @Content(schema = @Schema(implementation = ErrorMessage.class))),
+            @ApiResponse(responseCode = "501", description = "Agent Insights queries are not enabled")})
+    public Response executeQuery(
+            @RequestBody(content = @Content(schema = @Schema(implementation = AnalyticsQueryRequest.class))) @Valid AnalyticsQueryRequest request) {
+
```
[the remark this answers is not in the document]
BorisTkachenko: Fixed in da2aaeccce — added `@NotNull` to the request-body parameter so a null/empty body is rejected by Bean Validation (400/422) instead of NPEing on `request.projectId()`.

apps/opik-backend/src/main/java/com/comet/opik/api/AnalyticsQueryResponse.java
```diff
@@ -0,0 +1,11 @@
+package com.comet.opik.api;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.PropertyNamingStrategies;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+
+import java.util.List;
+
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
+public record AnalyticsQueryResponse(List<JsonNode> results) {
```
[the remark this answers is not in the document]
BorisTkachenko: Fixed in da2aaeccce — added `@Builder(toBuilder = true)` to `AnalyticsQueryResponse` and now build it via the builder in `FreeFormSqlQueryService.runQuery()`.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryService.java
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    public Mono<AnalyticsQueryResponse> executeQuery(@NonNull String workspaceId, @NonNull UUID projectId,
+            @NonNull String query) {
+        long startMillis = System.currentTimeMillis();
+
+        return freeFormSqlQueryDAO.explainAst(query)
+                // A failed EXPLAIN AST (unparseable input, statement-stacking, ...) is a hard reject: never fall through.
+                .onErrorMap(
+                        error -> reject(REASON_PARSE_ERROR, startMillis, "Query rejected: could not be parsed", error))
```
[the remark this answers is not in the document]
BorisTkachenko: Fixed in da2aaeccce — `parse_error` is now used only when `EXPLAIN AST` fails with ClickHouse `SYNTAX_ERROR` (code 62). Any other failure (transport, auth, oversized query, …) is routed through `mapExecutionError`/`REASON_OTHER` so it isn't mislabelled. It still never falls through to executing the query on a validation failure.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryService.java
```diff
@@ -0,0 +1,196 @@
...
+        ServerException serverException = findCause(error, ServerException.class);
+        if (serverException != null) {
+            int code = serverException.getCode();
+            if (CH_LIMIT_CODES.contains(code)) {
+                return error(REASON_CH_LIMIT, startMillis, Response.Status.BAD_REQUEST,
+                        "Query exceeded ClickHouse limits: %s".formatted(serverException.getMessage()), error);
+            }
+            if (code == CH_ACCESS_DENIED) {
+                return error(REASON_PERMISSION_DENIED, startMillis, Response.Status.BAD_REQUEST,
+                        "Query rejected: access denied", error);
```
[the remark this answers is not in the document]
BorisTkachenko: Fixed in da2aaeccce — the limit-breach branch now returns a constant `"Query exceeded ClickHouse limits"`; the raw `ServerException` message is kept in the server logs only (via the logged cause).

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthFilter.java
```diff
@@ -32,7 +32,10 @@ public void filter(ContainerRequestContext context) throws IOException {
 
         UriInfo uriInfo = context.getUriInfo();
 
-        if (Pattern.matches("/v1/private/.*", uriInfo.getRequestUri().getPath())) {
+        // Unlike other /v1/internal/* endpoints, the Agent Insights query executor must be authenticated: it derives
+        // the bounding workspace_id from auth (see OPIK-6814 / Agent Insights technical design).
+        if (Pattern.matches("/v1/private/.*", uriInfo.getRequestUri().getPath())
+                || Pattern.matches("/v1/internal/analytics-queries-executor.*", uriInfo.getRequestUri().getPath())) {
```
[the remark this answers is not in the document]
BorisTkachenko: Fixed in da2aaeccce — the three path checks are compiled once into `private static final Pattern` fields and matched per request via `matcher(path).matches()`, so the regex is no longer recompiled on every authenticated request.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryDAO.java
```diff
@@ -0,0 +1,96 @@
...
+    public Mono<List<String>> explainAst(@NonNull String query) {
+        return Mono.fromFuture(() -> readOnlyClient.queryRecords(EXPLAIN_AST_PREFIX + query))
+                .flatMap(records -> Mono.fromCallable(() -> {
+                    try (records) {
+                        List<String> nodeLabels = new ArrayList<>();
+                        for (GenericRecord node : records) {
+                            nodeLabels.add(node.getString(EXPLAIN_AST_COLUMN));
+                        }
+                        return nodeLabels;
+                    }
```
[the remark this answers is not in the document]
BorisTkachenko: Skipping — the two paths return different shapes: `explainAst` yields `List<String>`, while `execute` yields a `FreeFormSqlResult` that also needs `resultRows`/`readBytes` read from the `Records` within the same try-with-resources. A shared `drainRecords` helper would have to surface those metrics too, making it more convoluted than the current explicit code, so leaving as-is.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryDAO.java
```diff
@@ -0,0 +1,96 @@
...
+                .serverSetting(SETTING_PROJECT_ID, projectId.toString());
+
+        return Mono.fromFuture(() -> readOnlyClient.queryRecords(query, settings))
+                .flatMap(records -> Mono.fromCallable(() -> {
+                    try (records) {
+                        List<JsonNode> rows = new ArrayList<>();
+                        for (GenericRecord record : records) {
+                            rows.add(JsonUtils.getJsonNodeFromString(record.getString(RESULT_COLUMN)));
+                        }
+                        return new FreeFormSqlResult(rows, records.getResultRows(), records.getReadBytes());
```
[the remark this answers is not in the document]
BorisTkachenko: Skipping — a missing `result` column already throws `NoSuchColumnException`, which is mapped to a clean 400. Extra columns are intentionally tolerated (harmless and simply ignored), so we won't add result-set metadata validation here.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/internal/AnalyticsQueriesExecutorResource.java
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * Internal, authenticated endpoint that runs Ollie-generated read-only SQL against ClickHouse, bounded to the
+ * caller's workspace and the requested project. Authentication is required only to derive the bounding
+ * {@code workspace_id} ({@code project_id} comes from the body). Gated behind the {@code agentInsightsEnabled}
+ * toggle: when off it returns {@code 501 Not Implemented} and performs no ClickHouse access.
+ *
+ * <p>The caller's final query must return exactly one column named {@code result}, produced via
+ * {@code toJSONString(...)}.
+ */
+@Path("/v1/internal/analytics-queries-executor")
```
thiagohora: In this, it should be authenticated under private, not intermal

apps/opik-backend/src/main/java/com/comet/opik/api/AnalyticsQueryRequest.java
```diff
@@ -0,0 +1,17 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import io.swagger.v3.oas.annotations.media.Schema;
+import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.NotNull;
+import lombok.Builder;
+
+import java.util.UUID;
+
+@Builder(toBuilder = true)
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
```
thiagohora, line 12: Add `@JsonIgnoreProperties(ignoreUnknown = true)`

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryService.java
```diff
@@ -0,0 +1,206 @@
...
+                .map(result -> {
+                    recordSuccess(result, startMillis);
+                    return AnalyticsQueryResponse.builder().results(result.rows()).build();
+                })
+                .onErrorMap(error -> mapExecutionError(error, startMillis));
+    }
+
+    // The parser models every SETTINGS/SET clause as a Set* node, so any node label starting with "Set" is a reject.
+    private static boolean containsSetNode(List<String> nodeLabels) {
+        return nodeLabels.stream().anyMatch(label -> label.stripLeading().startsWith("Set"));
```
thiagohora: This guard is load-bearing for tenant isolation: `SQL_workspace_id`/`SQL_project_id` are `CHANGEABLE_IN_READONLY`, so a `SETTINGS SQL_workspace_id='<victim>'` clause would override the binding if it slipped past here. Relying on `startsWith("Set")` is a loose prefix match against an undocumented AST label.

For realistic SQL it won't false-positive — the matched token is the AST node *type*, which always leads the line, and sibling types start with `Sel` (`SelectQuery`, `SelectWithUnionQuery`), not `Set`; identifiers/table names like `settlements` render as `Identifier settlements` / `TableIdentifier settlements` and never lead the line. The one real false-positive vector is a string literal containing a newline whose continuation begins with `Set`.

The bigger concern is fail-open: if a ClickHouse version renames the settings node, this silently stops rejecting. Suggest matching the exact node-type token instead, so intent is explicit and a rename fails a test loudly:

```java
private static boolean containsSetNode(List<String> nodeLabels) {
    return nodeLabels.stream()
            .map(label -> label.stripLeading().split("\\s", 2)[0])
            .anyMatch(SET_NODE_TYPES::contains); // e.g. Set.of("SetQuery")
}
```

This also removes the newline-literal false positive. Worth pinning the exact label with a quick `EXPLAIN AST ... SETTINGS max_threads=1` against the target CH version.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryService.java
```diff
@@ -0,0 +1,206 @@
...
+                // Constant message: the raw ClickHouse text (which can carry schema/tenant details) stays in logs only.
+                return error(REASON_CH_LIMIT, startMillis, Response.Status.BAD_REQUEST,
+                        "Query exceeded ClickHouse limits", error);
+            }
+            if (code == CH_ACCESS_DENIED) {
+                return error(REASON_PERMISSION_DENIED, startMillis, Response.Status.BAD_REQUEST,
+                        "Query rejected: access denied", error);
+            }
+        }
+        return error(REASON_OTHER, startMillis, Response.Status.BAD_REQUEST, "Query execution failed", error);
```
thiagohora: The default branch maps *any* unclassified error to `400 BAD_REQUEST` — including genuine server-side failures (ClickHouse down, transport/connection errors during `explainAst` or `execute`). Surfacing an infra outage as a client error misleads callers and monitoring, and inflates the `rejected{reason=other}` metric with non-client events.

Consider mapping clearly non-client failures (connection/transport, unexpected `ServerException` codes) to `500`/`503` so 5xx alerting fires correctly, and reserve `400` for query-attributable errors.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/internal/AnalyticsQueriesResourceTest.java
```diff
@@ -0,0 +1,264 @@
...
+    @Test
+    @DisplayName("a write statement is rejected by the read-only profile")
+    void executeQuery__whenWriteStatement__thenRejected() {
+        assertBadRequest("INSERT INTO traces (id, workspace_id, project_id, name) VALUES ('%s', '%s', '%s', 'x')"
+                .formatted(UUID.randomUUID(), WORKSPACE_ID_A, projectIdA));
+    }
+
+    @Test
+    @DisplayName("a query against a non-granted table fails on permissions")
+    void executeQuery__whenNonGrantedTable__thenRejected() {
```
thiagohora: Nice coverage on a non-granted *app* table. One gap: `system.*` is untested. A read-only user can often still read `system.query_log` / `system.text_log` / `system.parts` / `system.tables`, which can carry other tenants' literal values and schema — and the row policies don't apply to system tables.

Suggest adding a case asserting `SELECT ... FROM system.query_log` (and `system.tables`) is rejected for the RO user, or confirm the OPIK-6846 provisioning explicitly denies the `system` database. This is the one remaining path I'd consider a genuine cross-tenant risk.

thiagohora (commented):
Solid, security-conscious work — clean Resource → Service → DAO layering, bounds pushed as CH server settings (no SQL injection surface, no TOCTOU since the pre-flight string is byte-identical to the executed one), constant error messages that don't leak schema/tenant detail, and genuinely strong tests (cross-tenant isolation, SETTINGS at top/subquery/CTE, write rejection, non-granted table, statement stacking, toggle-off → 501).

Three inline comments, none blocking on their own:
1. **`containsSetNode` prefix match** — load-bearing for tenant isolation but a loose, undocumented-label match that can fail open on a CH node rename; suggest exact-token matching.
2. **Infra errors mapped to 400** — server-side failures should surface as 5xx.
3. **`system.*` access untested** — the one remaining cross-tenant risk path; add a rejection test or confirm provisioning denies the `system` db.

Minor nits (no change strictly required): the `executed` counter is incremented on reject/error paths too (reads as "ran against CH" — `handled` might be clearer); no `@Size` cap on `query` (likely covered by HTTP body limit); the toggle check runs after auth, so a disabled endpoint still does the auth round-trip before returning 501.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/internal/AnalyticsQueriesResource.java
```diff
@@ -0,0 +1,75 @@
...
+    private final @NonNull Provider<RequestContext> requestContext;
+    private final @NonNull @Config("serviceToggles") ServiceTogglesConfig serviceToggles;
+
+    @POST
+    @Operation(operationId = "executeAnalyticsQuery", summary = "Execute Agent Insights free-form SQL", description = "Runs Ollie-generated read-only SQL bounded to the caller's workspace and the requested project. Returns 501 when the Agent Insights toggle is off.", responses = {
+            @ApiResponse(responseCode = "200", description = "Query results", content = @Content(schema = @Schema(implementation = AnalyticsQueryResponse.class))),
+            @ApiResponse(responseCode = "400", description = "Bad Request", content = @Content(schema = @Schema(implementation = ErrorMessage.class))),
+            @ApiResponse(responseCode = "422", description = "Unprocessable Content", content = @Content(schema = @Schema(implementation = ErrorMessage.class))),
+            @ApiResponse(responseCode = "501", description = "Agent Insights queries are not enabled")})
+    public Response executeQuery(
```
andrescrz: Let's add rate limiting to this endpoint, with the usual `RateLimited` annotation. 

We should avoid abusing this, as another layer of protection, by throttling the agent/client.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/internal/AnalyticsQueriesExecutorResource.java
BorisTkachenko: No, it was discussed and agreed that this will be under `internal` and with `private` like authentication. Basically to make sure it's only available to Ollie and other services and not exposed externally.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryDAO.java
```diff
@@ -0,0 +1,100 @@
...
+     * Parses {@code query} via {@code EXPLAIN AST} (without executing it) and returns the AST node labels, one per row.
+     */
+    Mono<List<String>> explainAst(String query);
+
+    /**
+     * Executes {@code query} bounded to the given workspace/project and reads the single {@code result} column.
+     */
+    Mono<FreeFormSqlResult> execute(String workspaceId, UUID projectId, String query);
+
+    record FreeFormSqlResult(List<JsonNode> rows, long resultRows, long readBytes) {
```
andrescrz: Minor: better as top level record, instead of nest, with a builder (and to builder).

apps/opik-backend/src/main/java/com/comet/opik/api/AnalyticsQueryRequest.java
BorisTkachenko: Added

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/internal/AnalyticsQueriesResourceTest.java
BorisTkachenko: Added test

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryService.java
BorisTkachenko: Updated

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryService.java
BorisTkachenko: Updated

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/internal/AnalyticsQueriesExecutorResource.java
andrescrz: Minor: for the path, I suggest choosing a more RESTful name. As usual, plural nouns, no verbs, denoting a resource location etc. Maybe `/v1/internal/analytics-queries` or `/v1/internal/analytics-reports` or `/v1/internal/analytics/reports` etc. You can get an AI suggested name, for better results.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryDAO.java
```diff
@@ -0,0 +1,100 @@
...
+    private static final String SETTING_WORKSPACE_ID = "SQL_workspace_id";
+    private static final String SETTING_PROJECT_ID = "SQL_project_id";
+
+    private final Client readOnlyClient;
+
+    @Inject
+    FreeFormSqlQueryDAOImpl(
+            @Named(DatabaseAnalyticsModule.READ_ONLY_FREE_FORM_SQL_CLICKHOUSE_CLIENT) @NonNull Client readOnlyClient) {
+        this.readOnlyClient = readOnlyClient;
+    }
```
andrescrz, line 63: Nit: better use Lombok for boilerplate constructor.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryDAO.java
```diff
@@ -0,0 +1,100 @@
...
+
+    @Override
+    @WithSpan
+    public Mono<List<String>> explainAst(@NonNull String query) {
+        return Mono.fromFuture(() -> readOnlyClient.queryRecords(EXPLAIN_AST_PREFIX + query))
```
andrescrz: The ClickHouse V2 client is not reactive. No need to artificially make the implementation reactive if the endpoint ends up in a `block` call.

Please let's simplify this.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryDAO.java
```diff
@@ -0,0 +1,100 @@
...
+                .flatMap(records -> Mono.fromCallable(() -> {
+                    try (records) {
+                        List<String> nodeLabels = new ArrayList<>();
+                        for (GenericRecord node : records) {
+                            nodeLabels.add(node.getString(EXPLAIN_AST_COLUMN));
+                        }
+                        return nodeLabels;
```
andrescrz: Nit: better as modern Java stream implementation (stream + map + collect).

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryDAO.java
```diff
@@ -0,0 +1,100 @@
...
+                .serverSetting(SETTING_PROJECT_ID, projectId.toString());
+
+        return Mono.fromFuture(() -> readOnlyClient.queryRecords(query, settings))
+                .flatMap(records -> Mono.fromCallable(() -> {
+                    try (records) {
+                        List<JsonNode> rows = new ArrayList<>();
+                        for (GenericRecord record : records) {
+                            rows.add(JsonUtils.getJsonNodeFromString(record.getString(RESULT_COLUMN)));
+                        }
+                        return new FreeFormSqlResult(rows, records.getResultRows(), records.getReadBytes());
```
andrescrz: Nit: better use builders everywhere.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryService.java
```diff
@@ -0,0 +1,218 @@
...
+    public static final String METRIC_NAMESPACE = "opik.free_form_sql.queries";
+
+    private static final AttributeKey<String> RESULT_KEY = stringKey("result");
+    private static final AttributeKey<String> REASON_KEY = stringKey("reason");
+
+    private static final Attributes RESULT_SUCCESS = Attributes.of(RESULT_KEY, "success");
+    private static final Attributes RESULT_REJECTED = Attributes.of(RESULT_KEY, "rejected");
+    private static final Attributes RESULT_ERROR = Attributes.of(RESULT_KEY, "error");
+
+    private static final Attributes REASON_AST_SET_NODE = Attributes.of(REASON_KEY, "ast_set_node");
```
andrescrz: Nit: more significative name for `ast_set_node`. It's hard to understand if no context.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryService.java
```diff
@@ -0,0 +1,218 @@
...
+
+    // ClickHouse error codes surfaced to the caller as a clean 4xx rather than a 500.
+    private static final int CH_TOO_MANY_ROWS = 158;
+    private static final int CH_TIMEOUT_EXCEEDED = 159;
+    private static final int CH_MEMORY_LIMIT_EXCEEDED = 241;
+    private static final int CH_TOO_MANY_ROWS_OR_BYTES = 396;
+    private static final int CH_ACCESS_DENIED = 497;
+    private static final int CH_SYNTAX_ERROR = 62;
+    private static final Set<Integer> CH_LIMIT_CODES = Set.of(
+            CH_TOO_MANY_ROWS, CH_TIMEOUT_EXCEEDED, CH_MEMORY_LIMIT_EXCEEDED, CH_TOO_MANY_ROWS_OR_BYTES);
```
andrescrz, line 84: Minor: this is seems like an enum.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryService.java
```diff
@@ -0,0 +1,218 @@
...
+
+        var meter = GlobalOpenTelemetry.get().getMeter(METRIC_NAMESPACE);
+        this.executed = meter
+                .counterBuilder("%s.executed".formatted(METRIC_NAMESPACE))
+                .setDescription("Number of free-form SQL queries handled, tagged by result")
+                .build();
+        this.rejected = meter
+                .counterBuilder("%s.rejected".formatted(METRIC_NAMESPACE))
+                .setDescription("Number of free-form SQL queries that did not succeed, tagged by reason")
+                .build();
```
andrescrz: Minor: generally histogram/time metrics already count. You can use the duration, with proper tags, so get global counters or broken down per executed/rejected/errors and so on.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryService.java
```diff
@@ -0,0 +1,218 @@
...
+        this.resultRows = meter
+                .histogramBuilder("%s.result_rows".formatted(METRIC_NAMESPACE))
+                .setDescription("Number of rows returned by a successful free-form SQL query")
+                .ofLongs()
+                .build();
+        this.bytesRead = meter
+                .histogramBuilder("%s.bytes_read".formatted(METRIC_NAMESPACE))
+                .setDescription("Number of bytes read by a successful free-form SQL query")
+                .ofLongs()
+                .build();
```
andrescrz, line 113: Number of rows and bytes read seem more like a gauge metric than a histogram.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryService.java
```diff
@@ -0,0 +1,218 @@
...
+                })
+                .onErrorMap(error -> mapExecutionError(error, startMillis));
+    }
+
+    // The parser models every SETTINGS/SET clause (top-level, subquery, CTE, FORMAT ... SETTINGS) as this AST node.
+    // Verified against ClickHouse 25.3.x: EXPLAIN AST prints it as the leading token "Set". Matching the exact node
+    // token (not a "Set" prefix) avoids false positives on Set-prefixed identifiers and keeps intent explicit; the
+    // SETTINGS rejection tests fail loudly if a future ClickHouse version renames the node.
+    private static final Set<String> SETTINGS_AST_NODES = Set.of("Set");
+
```
andrescrz: Nit: please move to the top of the class.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryService.java
```diff
@@ -0,0 +1,218 @@
...
+    private static boolean containsSetNode(List<String> nodeLabels) {
+        return nodeLabels.stream()
+                .map(label -> label.stripLeading().split("\\s", 2)[0])
```
andrescrz, line 158: Should you strip on both ends?

Why do you limit to two results?

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryService.java
```diff
@@ -0,0 +1,218 @@
...
+    private void recordSuccess(FreeFormSqlResult result, long startMillis) {
+        executed.add(1, RESULT_SUCCESS);
+        duration.record(elapsed(startMillis), RESULT_SUCCESS);
+        resultRows.record(result.resultRows());
+        bytesRead.record(result.readBytes());
+    }
+
+    private WebApplicationException mapExecutionError(Throwable error, long startMillis) {
+        if (findCause(error, NoSuchColumnException.class) != null) {
+            return error(REASON_OTHER, startMillis, Response.Status.BAD_REQUEST,
```
andrescrz: In all these `error` calls, it's fine to encapsulate the common behaviour. However, for appropriately catching and for the exception handles to properly work on all cases, better throw specific exception subclasses such as NotFoundException etc.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryService.java
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    private WebApplicationException error(Attributes reason, long startMillis, Response.Status status, String message,
+            Throwable cause) {
+        executed.add(1, RESULT_ERROR);
+        rejected.add(1, reason);
+        duration.record(elapsed(startMillis), RESULT_ERROR);
+        log.warn("Free-form SQL query failed: {}", message, cause);
+        return httpError(status, message);
+    }
```
andrescrz: Related to my previous comment. This method mostly copies and pastes the previous one, which can be parameterised. Easier if you extract the exception instantiation outside here.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryService.java
```diff
@@ -0,0 +1,218 @@
...
+
+    private static <T extends Throwable> T findCause(Throwable throwable, Class<T> type) {
+        for (Throwable current = throwable; current != null; current = current.getCause()) {
+            if (type.isInstance(current)) {
+                return type.cast(current);
+            }
+        }
+        return null;
+    }
```
andrescrz, line 229: You should use Guava's `Throwable` or similar from Apache instead of re-inventing this method. In addition, this is not protected against infinite loops for nested exception cause chaing, which Guava `Throwables.getRootCause` does.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthFilter.java
```diff
@@ -20,6 +20,12 @@
 @RequiredArgsConstructor(onConstructor_ = @Inject)
 public class AuthFilter implements ContainerRequestFilter {
 
+    // Compiled once and reused across requests rather than recompiled on every path check.
+    private static final Pattern PRIVATE_PATH_PATTERN = Pattern.compile("/v1/private/.*");
+    private static final Pattern ANALYTICS_QUERIES_EXECUTOR_PATH_PATTERN = Pattern
+            .compile("/v1/internal/analytics-queries-executor.*");
```
andrescrz: Minor: does it need ending `/`, like `/v1/internal/analytics-queries-executor./*`

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/internal/AnalyticsQueriesExecutorResourceDisabledTest.java
```diff
@@ -0,0 +1,108 @@
...
+import static com.github.tomakehurst.wiremock.client.WireMock.matchingJsonPath;
+import static com.github.tomakehurst.wiremock.client.WireMock.okJson;
+import static com.github.tomakehurst.wiremock.client.WireMock.post;
+import static com.github.tomakehurst.wiremock.client.WireMock.urlPathEqualTo;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@DisplayName("Analytics Queries Executor Resource Disabled Test")
+@ExtendWith(DropwizardAppExtensionProvider.class)
+class AnalyticsQueriesExecutorResourceDisabledTest {
```
andrescrz: Minor: better move this as a nested class inside the general test, so we keep them together for the clean up.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/internal/AnalyticsQueriesExecutorResourceTest.java
```diff
@@ -0,0 +1,280 @@
...
+        try (Response response = executorClient.callExecute(projectIdA, query, API_KEY_A, WORKSPACE_NAME_A)) {
+            assertThat(response.getStatus()).isEqualTo(Response.Status.BAD_REQUEST.getStatusCode());
+        }
+    }
+
+    private void assertResultIds(AnalyticsQueryResponse response, UUID... expectedIds) {
+        assertThat(response.results())
+                .extracting(node -> node.get("id").asText())
+                .containsExactlyInAnyOrder(List.of(expectedIds).stream().map(UUID::toString).toArray(String[]::new));
+    }
```
andrescrz: Please add stronger assertions actual vs expected result, fore more endurable tests. Issues might go unnoticed and this is a critical endpoint.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/internal/AnalyticsQueriesResourceTest.java
```diff
@@ -0,0 +1,280 @@
...
+                        workspace_id = getSetting('SQL_workspace_id')
+                        AND project_id = getSetting('SQL_project_id')
+                    AS RESTRICTIVE TO %s
+                    """.formatted(table, DATABASE_NAME, table, RO_USER));
+        }
+    }
+
+    private void executeStatement(TransactionTemplateAsync clickHouseTemplate, String sql) {
+        clickHouseTemplate.nonTransaction(connection -> Mono.from(connection.createStatement(sql).execute())).block();
+    }
```
andrescrz: This should be provisioned definitely in `clickhouse.xml` in the tests and flip the test config yaml to this user.

andrescrz (changes requested):
Please see the important feedback.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryDAO.java
thiagohora: It's, actually. We set `useAsyncRequests(true)` on `DatabaseAnalyticsFactory.buildClient`.  But the client already returns completable future ` CompletableFuture<Records> queryRecords`

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryDAO.java
BorisTkachenko: Unfortunately it's not possible here. I was using `lombok.RequiredArgsConstructor` initially. Forgot that our DI has this bug:( What happened is that in this case we are using named bean and with boilerplate constructor it just took the regular Client, with write access and without limitations. Took me ~3 hours to figure out why it's not working.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryDAO.java
BorisTkachenko: I would keep as is, since client is actually async.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryService.java
BorisTkachenko: In the AST we are looking for rows that start with `Set `. We are interested only in the first word before the whitespace, so 2 elements. To avoid false positives with string literal containing a newline whose continuation begins with Set.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/internal/AnalyticsQueriesExecutorResourceDisabledTest.java
BorisTkachenko: The two tests require different app startup configs — the main test boots with agentInsightsEnabled=true, this one with it off (default) to assert the 501. Since a feature toggle is bound at app construction, they need two separate Dropwizard apps, each with its own Testcontainers stack.

The harness doesn't really support that as a nested class: DropwizardAppExtensionProvider holds a single app in mutable instance state and drives all beforeAll/afterAll through it, and across the ~96 @Nested-using test classes in the repo, none declare their own @RegisterApp inside the nested class — the convention is that nested classes share the outer app. A nested class here would either have to share the (toggle-on) outer app, which defeats the test, or register a second differently-configured app, which the provider isn't built for.

Keeping them as sibling classes keeps the two toggle configs cleanly isolated and follows the existing convention, so I'll leave it as-is.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryService.java
BorisTkachenko: I kept these as histograms. A gauge models a point-in-time sampled value (last-write-wins), whereas result_rows / bytes_read are per-query measurements where the cross-query distribution (p50/p95/max) is what's useful — that's exactly a histogram. I did address the related points: the reason/result tags are now an enum, and the redundant counters were dropped in favor of the duration histogram's count.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryService.java
```diff
@@ -0,0 +1,214 @@
...
+
+    private WebApplicationException reject(Outcome outcome, long startMillis, String message, Throwable cause) {
+        log.info("Free-form SQL query rejected: {}", message, cause);
+        return fail(outcome, startMillis, Response.Status.BAD_REQUEST, message);
+    }
+
+    private WebApplicationException error(Outcome outcome, long startMillis, Response.Status status, String message,
+            Throwable cause) {
+        log.warn("Free-form SQL query failed: {}", message, cause);
```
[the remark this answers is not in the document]
BorisTkachenko: It should be ok

thiagohora (dismissed):

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryDAO.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @Inject
+    FreeFormSqlQueryDAOImpl(
+            @Named(DatabaseAnalyticsModule.READ_ONLY_FREE_FORM_SQL_CLICKHOUSE_CLIENT) @NonNull Client readOnlyClient) {
+        this.readOnlyClient = readOnlyClient;
+    }
+
+    @Override
+    @WithSpan
+    public Mono<List<String>> explainAst(@NonNull String query) {
```
andrescrz: I believe I forgot this in the first iteration: should we limit this method also by at least `workspace` and ideally by `project_id` too? The idea is to be paranoid and not only not leaking any data, also not leaking any metadata, specially deriving volumes information from explain statements.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryService.java
```diff
@@ -0,0 +1,221 @@
...
+@Singleton
+public class FreeFormSqlQueryService {
+
+    public static final String METRIC_NAMESPACE = "opik.free_form_sql.queries";
+
+    private static final AttributeKey<String> RESULT_KEY = stringKey("result");
+    private static final AttributeKey<String> REASON_KEY = stringKey("reason");
+
+    // The terminal outcome of a query, carrying its (result, reason) metric tags. The duration histogram is tagged
+    // with these and its per-tag count is the query counter, so no separate counters are needed.
```
andrescrz: Nit: general comment on the PR. All this sort of comments better as a javadoc on to of the field/method etc. than as regular comments.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryService.java
```diff
@@ -0,0 +1,221 @@
...
+    private static final Set<String> SETTINGS_AST_NODES = Set.of("Set");
+
+    // ClickHouse error codes surfaced to the caller as a clean 4xx rather than a 500.
+    private static final int CH_TOO_MANY_ROWS = 158;
+    private static final int CH_TIMEOUT_EXCEEDED = 159;
+    private static final int CH_MEMORY_LIMIT_EXCEEDED = 241;
+    private static final int CH_TOO_MANY_ROWS_OR_BYTES = 396;
+    private static final int CH_ACCESS_DENIED = 497;
+    private static final int CH_SYNTAX_ERROR = 62;
+    private static final Set<Integer> CH_LIMIT_CODES = Set.of(
```
andrescrz: Nit: similarly, this looks better like an enum. Ok to keep this as well.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/ClickHouseContainerUtils.java
```diff
@@ -87,7 +87,9 @@ public static ClickHouseContainer newClickHouseContainer(boolean reusable, Netwo
                     .withUsername("default")
                     .withPassword("")
                     .withCopyFileToContainer(MountableFile.forClasspathResource("clickhouse.xml"),
-                            "/etc/clickhouse-server/config.d/clickhouse.xml");
+                            "/etc/clickhouse-server/config.d/clickhouse.xml")
+                    .withCopyFileToContainer(MountableFile.forClasspathResource("clickhouse-users.xml"),
+                            "/etc/clickhouse-server/users.d/clickhouse-users.xml");
```
andrescrz: As discussed, my PR has overlapping changes that I noticed that I needed in parallel. Let's coordinate to not to duplicate this when merging: 
- https://github.com/comet-ml/opik/pull/7128/

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/internal/AnalyticsQueriesResourceTest.java
```diff
@@ -0,0 +1,291 @@
...
+    void executeQuery__whenQueryContainsSettings__thenRejected(String query) {
+        assertBadRequest(query);
+    }
+
+    @Test
+    @DisplayName("a Set-prefixed identifier is not mistaken for a SETTINGS node")
+    void executeQuery__whenSetPrefixedIdentifier__thenAccepted() {
+        // The AST guard matches the exact `Set` node token, so a `settings`-named CTE must not be falsely rejected.
+        var query = "WITH settings AS (SELECT id, workspace_id, project_id FROM traces) "
+                + "SELECT toJSONString(map('id', toString(id), 'workspace_id', workspace_id, 'project_id', toString(project_id))) AS result FROM settings";
```
andrescrz: Nit: use java text blocks (triple quotes `"""`) instead.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/internal/AnalyticsQueriesResourceTest.java
```diff
@@ -0,0 +1,291 @@
...
+        assertScopedRows(analyticsQueriesClient.execute(projectIdA, query, API_KEY_A, WORKSPACE_NAME_A),
+                WORKSPACE_ID_A, projectIdA, traceIdA);
+    }
+
+    @Test
+    @DisplayName("an unparseable query is rejected")
```
andrescrz: Minor: many of these tests (for bad request) seem like a single parameterised test.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/internal/AnalyticsQueriesResourceTest.java
```diff
@@ -0,0 +1,291 @@
...
+
+        // The default AuthTestUtils stub only matches /v1/private/*; this endpoint authenticates under /v1/internal/*.
+        wireMock.server().stubFor(post(urlPathEqualTo("/opik/auth"))
+                .withHeader(HttpHeaders.AUTHORIZATION, equalTo(apiKey))
+                .withRequestBody(matchingJsonPath("$.workspaceName", equalTo(workspaceName)))
+                .withRequestBody(matchingJsonPath("$.path", matching("/v1/internal/analytics-queries(/.*)?")))
+                .willReturn(okJson(AuthTestUtils.newWorkspaceAuthResponse(USER, workspaceId, workspaceName, null))));
+    }
+
+    private void provisionReadOnlyUser(TransactionTemplateAsync clickHouseTemplate) {
```
andrescrz: Let's coordinate this with my other PR: 
- https://github.com/comet-ml/opik/pull/7128

I'd rather not do this low level execute statement, it's very hard to maintain. I believe the way implemented in the other PR is more similar to prod, and seems to properly provision the user in my tests without requiring this.

apps/opik-backend/src/test/resources/clickhouse-users.xml
andrescrz, line 1: As mentioned, we should coordinate this to avoid overlap:
- https://github.com/comet-ml/opik/pull/7128/

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryDAO.java
andrescrz: This is not reactive. This uses Java's `CompletableFuture` which is another 1st party async API, then we wrap it in a reactive call.

We shouldn't force this async API to be converted to a reactive one. This means two layers of indirection to control, with all the associated threading resources (default or custom reactive schedulers vs java common pool or custom executors). 

This can easy start creating very hard to control issues in production under load and concurrency. 

It already bite us with the Redisson client when were forcing the sync API to the reactive one. We had a very obscure bug that caused an incident in production. You can see now that the Redisson client we use has 3 flavours of APIs (see `CacheManager`): reactive vs sync vs Java async (with `CompletionStage`).

Let's better simplify this. By always using the Java's async API. You should follow a similar approach as with reactive (with `block`), which means we work on an async way on the DAO, service etc. and we only terminate on the last minute when needed (generally at the endpoint). We could even consider full async endpoints if that is supported, but not mandatory.

The only reason why we used reactor is because we were forced by the R2DBC client, but it's not an architectural pattern that we are forced follow. Actually, DropWizard forces us to `block` as it isn't compatible with it. 

This already slip through in previous PRs in the `DatasetItemVersionDAO`. We should no longer continue that path. 

CC @thiagohora for awareness.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/internal/AnalyticsQueriesExecutorResourceDisabledTest.java
andrescrz: I'm fine to keep this, but for future reference you can see that multiple apps with different startup configs are supported. Examples:
- https://github.com/comet-ml/opik/pull/7128/
- `WorkspaceVersionResourceTest`
- Etc.

andrescrz (changes requested):
The main blocker is forcing the reactiveness with the V2 client. I further clarified that. We shouldn't go with that, it's too risky and kind of an anti-pattern for the service. CC @thiagohora 

The other outstanding item is coordinating the overlap of the user provision with my other PR: https://github.com/comet-ml/opik/pull/7128 and avoiding the way you provision.

The rest is minor.

The change has greatly improve, thanks for addressing the previous feedback.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryDAO.java
BorisTkachenko: No, we don't need it here. This method doesn't return any data or execute the query. It only returns AST, in order to validate that provided query doesn't have any SETTINGS/SET in it. Result is used for verification only and doesn't contain any user data.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryDAO.java
thiagohora: Hi @andrescrz, 

Ok, no problem — I was using the wrong terminology. The client exposes a Java async API via CompletableFuture / CompletionStage, not a full reactive API.

My concern is that by adopting the Java async API end-to-end and only blocking at the controller layer, we would be introducing another asynchronous programming model that the team needs to understand, maintain, and troubleshoot alongside the existing Reactor-based code paths.

Looking at the ClickHouse R2DBC client, it seems to follow a very similar approach internally anyway — Reactor is largely an API wrapper around the async Java client (Mono.fromFuture(...) over the underlying async execution). So from my perspective, we're already dealing with many of the same execution and concurrency concerns today.

I'm happy to go ahead with your suggestion; I just want to better understand the rationale for introducing and supporting another async model versus continuing with the one we're already familiar with. Is the main goal to reduce abstraction layers, make execution ownership more explicit, or address some specific issue we've seen with Reactor in practice?

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/internal/AnalyticsQueriesResource.java
```diff
@@ -0,0 +1,89 @@
...
+        // (and its HTTP status) reaches the JAX-RS exception handling unchanged.
+        try {
+            AnalyticsQueryResponse response = freeFormSqlQueryService
+                    .executeQuery(workspaceId, request.projectId(), request.query())
+                    .join();
+            return Response.ok(response).build();
+        } catch (CompletionException e) {
+            Throwables.throwIfUnchecked(e.getCause());
+            throw e;
+        }
```
andrescrz, line 89: Not a blocker: we better need a generic solution for this in the long run, so we don't have to try-catch in all endpoints. Typically, we use a mapper for this type of exceptions, we could have one for `CompletionException`. 

Ok to not increase the scope of this PR, but feel free to follow-up in a next PR.

apps/opik-backend/src/main/java/com/comet/opik/api/AnalyticsQueryRequest.java
```diff
@@ -0,0 +1,19 @@
...
+import jakarta.validation.constraints.NotNull;
+import lombok.Builder;
+
+import java.util.UUID;
+
+@Builder(toBuilder = true)
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
+public record AnalyticsQueryRequest(
+        @NotNull UUID projectId,
```
andrescrz: Minor: forgot mentioning in previous iterations. As the `projectId` is mandatory, probably a better approach would be enforcing this in the path param of this resource by adding to `/v1/internal/analytics-queries` the following path `/projects/{project_id}`. That way we limit the request object to `query` field and this endpoint resembles other resources in the application.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryDAO.java
```diff
@@ -0,0 +1,112 @@
...
+    private static List<String> readNodeLabels(Records records) {
+        try (records) {
+            return StreamSupport.stream(records.spliterator(), false)
+                    .map(node -> node.getString(EXPLAIN_AST_COLUMN))
+                    .toList();
+        } catch (RuntimeException e) {
+            throw e;
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
```
andrescrz: Nit: not sure if checked exceptions declared in the body of this method. If so, you could remove the try-catch all together and let it bubble up.

Comment applies to the method below.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryDAO.java
andrescrz: Makes sense to have uniformity, and I was not aware about he R2DBC implementation. 

The rationale was not having two layers of indirection for async flows, which I thought it's the least of the evils.

The main reason is that we don't take benefit of reactor due to Dropwizard and that I foresee that we'll move more towards the V2 client in the long run.

Generally, it's cleaner using a standard JDK approach than a 3rd party as reactor. If we migrate to CH V2 client completely at some point, there's no need to maintain it.

andrescrz (dismissed):
LGTM, the implementation is clean and you reuse the already existing test user introduced in my previous PR. Thanks for putting this together!

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/internal/AnalyticsQueriesResource.java
BorisTkachenko: Will address as a follow up

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryDAO.java
BorisTkachenko: Collapsed the two catch clauses into one. It can't be removed entirely, though: try (records) calls Records.close(), which is declared throws Exception (checked), and these helpers run as thenApply Functions, which can't propagate checked exceptions — so the checked close() has to be converted to unchecked. RuntimeExceptions (e.g. the UncheckedIOException the 400 mapping relies on) are rethrown unwrapped.

andrescrz (approved):

PR #7112, opened 2026-06-15
[OPIK-6938] [BE] feat: add error count by error type metric

thiagohora:
## Details

Metric slice **M11** of OPIK-4063: adds `error_type` (the root-cause exception class name) as a dimension on error metrics, so triage can tell *what kind* of failure is spiking — e.g. a ClickHouse timeout vs an NPE behind the same HTTP 500.

**HTTP request path (new counter `opik.errors.count`, 5XX only):**
- `OpikGenericExceptionMapper` — catch-all for uncaught 5XX; extends Dropwizard `LoggingExceptionMapper`, so logging and the 500 response body are unchanged, then records the metric.
- `OpikWebApplicationExceptionMapper` — records 5XX `WebApplicationException`s (e.g. `InternalServerErrorException`), returns the exception's original response unchanged; 4XX ignored.
- `ErrorMetrics` / `ErrorMetricsResolver` — OTel counter facade + label resolution (root-cause name via `ExceptionUtils.getRootCause`, route template via `ExtendedUriInfo`, workspace id/name from `RequestContext`; missing values fall back to `unknown`).
- The two mappers handle disjoint exception type-sets (WebApplicationException subtree vs everything else) and JAX-RS selects exactly one mapper per exception, so there is no double-counting.

**Async paths (extending the same `error_type` breakdown):**
- `BaseRedisSubscriber` — `processing_errors` is tagged with `error_type` + `workspace_id` + `workspace_name` + `user_name`; `unexpected_errors` with `error_type`. Per-failure increments use each failure's own root cause. Workspace/user come from a subclass-provided `messageContext()` hook that mirrors the values each `processEvent` sets into the Reactor context (sourced from the message). `RequestContext.workspaceName` (set at auth) flows via the Reactor context into the request-originated producers (experiment-execution, dataset-export) and into the message payload; `WebhookEvent` already carries it. The two messages with no name source — experiment-aggregation (built from a Redis bucket) and trace-thread closure (built from a ClickHouse row) — fall back to `unknown`. New message fields are nullable for rolling-deploy compatibility with the Java stream codec.
- `EventInterceptor` (`@Subscribe` listeners) — collapses the previous name-embedded `opik.event.error.<listener>` series into a single `opik.event.error` counter with `listener` + `error_type` labels (built once instead of per-error). **Note:** this removes the old per-listener metric *name* series — any dashboard querying `opik.event.error.*` must switch to the `listener` label.

**Decisions:** 4XX is intentionally out of scope on the HTTP path (client-side symptoms; cause rarely actionable for triage). `error_type` uses raw exception class names (a code-bounded set). Grafana panels and `internal`/`db_timeout` alerts are dashboard/infra config (outside this repo); this PR provides the counters + labels that enable them.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- OPIK-6938

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: Full implementation (mappers, async-path metric labels + workspace_name propagation, resolver, unit tests) under human direction on scope/design decisions.
- Human verification: Reviewed design (5XX-only HTTP scope, exception-name labels, workspace_name/user_name availability + propagation through producers, the EventInterceptor breaking change); build + tests verified.

## Testing

- `mvn test -Dtest='ErrorMetricsResolverTest,OpikWebApplicationExceptionMapperTest,OpikGenericExceptionMapperTest,BaseRedisSubscriberUnitTest,ExperimentExecutionServiceTest,CsvDatasetExportServiceImplTest'` — pass.
- `mvn spotless:check` — clean. `mvn compile` — clean (Java 25 toolchain).
- Scenarios covered: root-cause unwrapping (nested/single/null), endpoint template reconstruction, workspace id/name resolution incl. fallback; mapper behavior (5XX only, response preserved, no double-count); redis subscriber failure path tags error_type + workspace_id + workspace_name + user_name; producers thread workspace_name from the request context; producer tests green.

## Documentation

N/A — internal operational metrics. Grafana dashboards/alerts to be configured separately; note the `opik.event.error.<listener>` → `opik.event.error{listener=...}` migration for any existing event-error panels.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/error/OpikGenericExceptionMapper.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/error/OpikGenericExceptionMapper.java
@@ -1,11 +1,12 @@
 package com.comet.opik.api.error;
 
 import com.comet.opik.infrastructure.auth.RequestContext;
-import com.comet.opik.infrastructure.metrics.ErrorMetrics;
+import com.comet.opik.infrastructure.metrics.HttpErrorMetrics;
 import io.dropwizard.jersey.errors.LoggingExceptionMapper;
 import jakarta.annotation.Priority;
 import jakarta.inject.Inject;
 import jakarta.inject.Provider;
+import jakarta.ws.rs.core.Request;
 import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.core.UriInfo;
 
@@ -18,21 +19,23 @@ import jakarta.ws.rs.core.UriInfo;
 @Priority(1)
 public class OpikGenericExceptionMapper extends LoggingExceptionMapper<Throwable> {
 
-    private final ErrorMetrics errorMetrics;
+    private final HttpErrorMetrics errorMetrics;
     private final Provider<RequestContext> requestContext;
+    private final Provider<Request> request;
     private final Provider<UriInfo> uriInfo;
 
     @Inject
-    public OpikGenericExceptionMapper(ErrorMetrics errorMetrics, Provider<RequestContext> requestContext,
-            Provider<UriInfo> uriInfo) {
+    public OpikGenericExceptionMapper(HttpErrorMetrics errorMetrics, Provider<RequestContext> requestContext,
+            Provider<Request> request, Provider<UriInfo> uriInfo) {
         this.errorMetrics = errorMetrics;
         this.requestContext = requestContext;
+        this.request = request;
         this.uriInfo = uriInfo;
     }
 
     @Override
     public Response toResponse(Throwable exception) {
-        errorMetrics.record(exception, uriInfo.get(), requestContext);
+        errorMetrics.record(exception, request, uriInfo.get(), requestContext);
         return super.toResponse(exception);
     }
 }
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/error/OpikGenericExceptionMapper.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/error/OpikGenericExceptionMapper.java
@@ -1,0 +1,38 @@
+package com.comet.opik.api.error;
+
+import com.comet.opik.infrastructure.auth.RequestContext;
+import com.comet.opik.infrastructure.metrics.ErrorMetrics;
+import io.dropwizard.jersey.errors.LoggingExceptionMapper;
+import jakarta.annotation.Priority;
+import jakarta.inject.Inject;
+import jakarta.inject.Provider;
+import jakarta.ws.rs.core.Response;
+import jakarta.ws.rs.core.UriInfo;
+
+/**
+ * Catch-all for uncaught throwables (the 5xx path). Counts the error under {@code opik.errors.count}
+ * tagged with the root-cause class name, then delegates to {@link LoggingExceptionMapper} to keep the
+ * existing logging and {@code 500} rendering behaviour unchanged.
+ */
+@jakarta.ws.rs.ext.Provider
+@Priority(1)
+public class OpikGenericExceptionMapper extends LoggingExceptionMapper<Throwable> {
+
+    private final ErrorMetrics errorMetrics;
+    private final Provider<RequestContext> requestContext;
+    private final Provider<UriInfo> uriInfo;
+
+    @Inject
+    public OpikGenericExceptionMapper(ErrorMetrics errorMetrics, Provider<RequestContext> requestContext,
+            Provider<UriInfo> uriInfo) {
+        this.errorMetrics = errorMetrics;
+        this.requestContext = requestContext;
+        this.uriInfo = uriInfo;
+    }
+
+    @Override
+    public Response toResponse(Throwable exception) {
+        errorMetrics.record(exception, uriInfo.get(), requestContext);
+        return super.toResponse(exception);
+    }
+}
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/BaseRedisSubscriber.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/BaseRedisSubscriber.java
@@ -15,6 +15,7 @@ import lombok.Builder;
 import lombok.Getter;
 import lombok.NonNull;
 import org.apache.commons.collections4.CollectionUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.redisson.api.RStreamReactive;
 import org.redisson.api.RedissonReactiveClient;
 import org.redisson.api.options.PlainOptions;
@@ -461,8 +462,10 @@ public abstract class BaseRedisSubscriber<M> implements Managed {
                     .messageId(messageId)
                     .status(MessageStatus.FAILURE)
                     .error(classCastException)
+                    .context(MessageContext.UNKNOWN)
                     .build());
         }
+        var messageContext = message != null ? messageContext(message) : MessageContext.UNKNOWN;
         var startMillis = System.currentTimeMillis();
         // Deferring as processEvent is out of our control, it might not return a cold Mono
         return Mono.defer(() -> processEvent(message))
@@ -470,12 +473,14 @@ public abstract class BaseRedisSubscriber<M> implements Managed {
                 .thenReturn(ProcessingResult.builder()
                         .messageId(messageId)
                         .status(MessageStatus.SUCCESS)
+                        .context(messageContext)
                         .build())
                 .doOnSuccess(r -> log.info("Successfully processed message messageId '{}'", entry.getKey()))
                 .onErrorResume(throwable -> Mono.just(ProcessingResult.builder()
                         .messageId(messageId)
                         .status(MessageStatus.FAILURE)
                         .error(throwable)
+                        .context(messageContext)
                         .build()))
                 .doFinally(signalType -> {
                     messageProcessingTime.record(System.currentTimeMillis() - startMillis);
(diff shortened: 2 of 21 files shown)


apps/opik-backend/src/main/java/com/comet/opik/api/error/OpikGenericExceptionMapper.java
```diff
@@ -0,0 +1,38 @@
...
+            Provider<UriInfo> uriInfo) {
+        this.errorMetrics = errorMetrics;
+        this.requestContext = requestContext;
+        this.uriInfo = uriInfo;
+    }
+
+    @Override
+    public Response toResponse(Throwable exception) {
+        errorMetrics.record(exception, uriInfo.get(), requestContext);
+        return super.toResponse(exception);
```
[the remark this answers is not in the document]
thiagohora: Good catch — the finding is valid: a resource that returns a 5xx `Response` directly (e.g. `OllamaResource`, `LocalRunnersResource`, the health check) doesn't throw, so it isn't counted today.

We deliberately kept recording in the exception mappers rather than adding a `ContainerResponseFilter`, to avoid double-counting. Thrown 5xx are already counted exactly once (`OpikGenericExceptionMapper` for uncaught → 500, `OpikWebApplicationExceptionMapper` gated to 5xx). A filter that also recorded `getStatus() >= 500` would re-count those same responses. Making the filter the *single* recorder instead would require passing the originating `Throwable` to it (to preserve the `error_type` tag) via request-scoped state — extra machinery we didn't want to take on in this PR.

The uncounted set is small and mostly proxy/health endpoints where `error_type` would be `unknown` anyway, so the undercount is minor. Leaving as-is here; happy to revisit with a single-source-of-truth refactor if direct-5xx coverage becomes important.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/metrics/HttpErrorMetrics.java
```diff
@@ -0,0 +1,60 @@
...
+                ErrorMetricsResolver.workspaceName(requestContext));
+    }
+
+    public void record(String errorType, String method, String endpoint, String workspaceId, String workspaceName) {
+        errorCounter.add(1, Attributes.builder()
+                .put(ERROR_TYPE_KEY, StringUtils.defaultIfBlank(errorType, UNKNOWN))
+                .put(METHOD_KEY, StringUtils.defaultIfBlank(method, UNKNOWN))
+                .put(ENDPOINT_KEY, StringUtils.defaultIfBlank(endpoint, UNKNOWN))
+                .put(WORKSPACE_ID_KEY, StringUtils.defaultIfBlank(workspaceId, UNKNOWN))
+                .put(WORKSPACE_NAME_KEY, StringUtils.defaultIfBlank(workspaceName, UNKNOWN))
```
JetoPistola, line 57: 🧹 **nit** | Observability

The counter tags both `workspace_id` and `workspace_name`. They're 1:1, so `workspace_name` adds no new dimensionality but doubles label storage on every series — and because names are mutable, a workspace rename forks the time series into a new one. Worth considering whether resolving the human-readable name at query/dashboard time (joining on `workspace_id`) would be cheaper to store. Same pattern applies to `processing_errors` in `BaseRedisSubscriber`. Not blocking — just flagging the cardinality/storage trade-off.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/BaseRedisSubscriber.java
```diff
@@ -455,21 +461,25 @@ private Mono<ProcessingResult> processMessage(Map.Entry<StreamMessageId, Map<Str
                     .messageId(messageId)
                     .status(MessageStatus.FAILURE)
                     .error(classCastException)
+                    .context(MessageContext.UNKNOWN)
                     .build());
         }
+        var messageContext = message != null ? messageContext(message) : MessageContext.UNKNOWN;
```
JetoPistola: ❓ **question** | Design

`messageContext(message)` is now computed for every message (both success and failure), but the result is only read on the failure path in `postProcessFailureMessages`. For high-throughput subscribers that's per-message overhead on the hot path. It's cheap (just record construction) so likely fine — was the eager computation deliberate over computing it lazily inside the failure branch from each `failure`'s own message?

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/events/EventInterceptor.java
```diff
@@ -63,4 +63,14 @@ public Object invoke(MethodInvocation methodInvocation) {
             span.end();
         }
     }
+
+    private LongCounter errorCounter() {
```
JetoPistola: 🧹 **nit** | Concurrency

`errorCounter()` does check-then-act on a `volatile` field with no synchronization, so two threads racing the first error can both call `counterBuilder().build()`. This is benign — OTel returns the same registered instrument for a given name — but since the lazy pattern reads as intended-to-build-once, a one-line comment noting the duplicate build is intentionally tolerated would save the next reader the analysis.

JetoPistola:
👋 **Review summary**

**What looks good**
- The double-counting analysis is airtight — the two mappers cover disjoint exception type-sets (`WebApplicationException` subtree vs. everything else) and extend `LoggingExceptionMapper`, so logging and response rendering stay byte-for-byte unchanged while the metric is layered on top.
- Cardinality was clearly thought through up front: `endpoint` uses the matched route template rather than the raw path, and `error_type` is a code-bounded set of class names. The reasoning is captured right in the `ErrorMetricsResolver` Javadoc.
- Rolling-deploy safety is handled end-to-end: nullable new message fields plus `FAIL_ON_UNKNOWN_PROPERTIES=false` in the stream codec, with a regression test that explicitly guards against the flag being reintroduced.
- Genuinely thorough unit coverage — root-cause unwrapping (nested/single/null), template reconstruction, the 5XX-only mapper behavior with the 4XX negative case, and the request-scope fallback to `unknown` are all exercised.
- The `EventInterceptor` migration (`opik.event.error.<listener>` name series → single counter with a `listener` label) is called out explicitly in the PR body so dashboards can be updated.

**Overall**
Solid, well-scoped slice. The approach — a shared resolver feeding both the HTTP mappers and the async subscribers — keeps the label vocabulary consistent across both paths, and the workspace-name propagation through the producers is traced carefully. Nothing structural to flag.

**Inline comments**: 2 nits and 1 question — nothing blocking. The main one worth a look is the `workspace_id` + `workspace_name` label redundancy (storage/mutability trade-off).

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/metrics/HttpErrorMetrics.java
thiagohora: Keeping `workspace_name` intentionally. Emitting it on the series lets Grafana panels/alerts show the human-readable workspace directly; the alternative (resolve at query time by joining on `workspace_id`) needs that id→name mapping to exist in the metrics backend, which it doesn't — there's no workspace table to join against in Prometheus/Grafana. The cardinality cost is bounded (finite workspace count, infrequent renames) and acceptable for this operational metric.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/BaseRedisSubscriber.java
thiagohora: Fixed in cd3aaae73a — `messageContext(message)` is now resolved lazily inside the `onErrorResume` (failure) branch instead of eagerly for every message. The context is only read on the failure path, so successful messages no longer pay for it.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/events/EventInterceptor.java
thiagohora: Fixed in cd3aaae73a — removed the lazy `errorCounter()` and initialize the counter eagerly in the constructor (`final` field), eliminating the check-then-act race entirely. Safe because `GlobalOpenTelemetry` is installed by the otel javaagent at JVM `premain`, before this interceptor is constructed (same pattern as `HttpErrorMetrics`).

JetoPistola (approved):
Approving — the code is sound and safe to merge. The two earlier review comments were addressed cleanly (lazy message-context resolution on the failure path; eager `final` counter init in `EventInterceptor`, race eliminated), plus a nice typed-`AttributeKey` refactor. The change is purely additive to the error paths (`super.toResponse` delegation keeps logging + responses unchanged), instrumentation degrades to `unknown` rather than throwing inside the mappers, rolling-deploy compatibility is handled, and CI is fully green.

One thing best verified in a running environment rather than in review: actual metric **emission and label correctness** through the real OTel pipeline (`opik.errors.count` showing up with `error_type` / `endpoint` / `workspace_*`). Could you add the `test-environment` label to spin up the adhoc env and sanity-check that before merge? The `workspace_id` + `workspace_name` cardinality trade-off (open inline comment) is also worth an eye on the metrics backend — though true at-scale cardinality will only show on staging/prod under load.

CometActions:

### Access Information
- **URL:** https://pr-7112.dev.comet.com
- **Namespace:** pr-7112
- **Version:** 2.0.63-7112-merge-2289
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7112%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

PR #7113, opened 2026-06-15
[OPIK-6960] [FE] Fix hash collision in annotation queue per-user item shuffle

miguelgrc:
## Details
The `hashCode` function (`hash * 31 + char`) used for per-user deterministic shuffling of annotation queue items had a ~33% collision rate on UUID strings. The Java-style hash allows the long shared UUID prefix to dominate the hash state, washing out the per-user seed — causing different annotators to see the same item ordering in the SME flow sidebar.

Replaced with MurmurHash2-style mixing (`Math.imul` + xor shifts), reducing collision rate from ~33% to ~0.1%.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- OPIK-6960

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.6
- Scope: Implementation and testing
- Human verification: yes — root cause confirmed by manual testing in prod with two users

## Testing

- `npx vitest run src/lib/annotation-queues.test.ts` — 35/35 pass
- `npx eslint --cache` on changed files — clean
- `npx tsc --noEmit` — clean
- Benchmarked collision rate with 5000 random UUID sets and the two affected usernames: 0.2% (down from 33%)
- Manual verification: two different users in the same queue saw identical ordering before the fix, different orderings in other queues (confirming hash-dependent behavior)

## Documentation

N/A

### DIFF
--- a/apps/opik-frontend/src/lib/annotation-queues.test.ts
+++ b/apps/opik-frontend/src/lib/annotation-queues.test.ts
@@ -258,6 +258,28 @@ describe("hashCode", () => {
   it("should handle empty string", () => {
     expect(hashCode("")).toBe(0);
   });
+
+  it("should produce different sort orders for different user prefixes on UUIDs", () => {
+    const items = [
+      "11111111-1111-4111-8111-111111111111",
+      "22222222-2222-4222-8222-222222222222",
+      "33333333-3333-4333-8333-333333333333",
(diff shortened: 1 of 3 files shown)


aadereiko (approved):
LGTM!

PR #7121, opened 2026-06-16
[OPIK-6968] [BE] refactor: remove orphaned trace-thread scored_at write to cut lock contention

thiagohora:
## Details

Online scoring stamped `trace_threads.scored_at` via `TraceThreadService.setScoredAt`, which acquired the per-project `{projectId}-trace-threads-process` lock on every scoring completion. The read that justified `scored_at` — dedup of closed-thread scoring (`... WHERE scored_at IS NULL`) — was removed in #2719 when closure/scoring moved to a stream-on-close model. Dedup is now structural (a thread is scored once, at the `active`→`inactive` transition), so `scored_at` became write-only: written on every scoring batch, read by nobody.

This removes `setScoredAt` and all `scored_at` references from the threads domain, eliminating one of the producers contending for the single per-project lock. Behavior-neutral.

- Also narrows `processTraceThreads` to hold the lock only around the save (the heavy thread processing now runs outside the lock).
- The `scored_at` ClickHouse column is intentionally **kept** (no migration) so a rolling deploy stays safe — old pods still reference it mid-deploy. It can be dropped in a follow-up once this change is fully rolled out.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- OPIK-6968

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Scope: Traced the `scored_at` reader's removal via git history, removed all `scored_at` references across the threads domain + online scorers + their unit tests, kept the DB column.

## Testing

- `mvn test-compile` (JDK 25) — BUILD SUCCESS for both main and test sources.
- `mvn spotless:check` — passed.
- Not run locally: full `mvn test` (DB-backed integration suites) — relying on CI.

## Documentation
NA


apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringTraceThreadLlmAsJudgeScorer.java
```diff
@@ -114,8 +113,7 @@ protected Mono<Void> score(@NonNull TraceThreadToScoreLlmAsJudge message) {
 
         return Flux.fromIterable(message.threadIds())
                 .flatMap(threadId -> processThreadScores(message, threadId))
-                .then(Mono.defer(
-                        () -> traceThreadService.setScoredAt(message.projectId(), message.threadIds(), Instant.now())))
+                .then()
                 .contextWrite(context -> context.put(RequestContext.WORKSPACE_ID, message.workspaceId())
                         .put(RequestContext.USER_NAME, message.userName())
```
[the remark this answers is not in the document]
thiagohora: No worth doing it at the moment

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadService.java
```diff
@@ -94,13 +92,14 @@ class TraceThreadServiceImpl implements TraceThreadService {
...
-                        .flatMap(traceThreads -> this.saveTraceThreads(projectId, traceThreads))
-                        .then()),
-                LOCK_DURATION);
+
+        return processThreadAsync(threadInfo, projectId)
+                .collectList()
+                .flatMap(traceThreads -> lockService.executeWithLockCustomExpire(
+                        new LockService.Lock(projectId, TraceThreadService.THREADS_LOCK),
+                        Mono.defer(() -> this.saveTraceThreads(projectId, traceThreads)
+                                .then()),
```
JetoPistola, line 101: 💡 **suggestion** | Architecture

This lock-narrowing change (moving `processThreadAsync` outside the lock, keeping only `saveTraceThreads` inside) is correct — `processThreadAsync` only does idempotent get-or-create on the thread-ID table, not on `trace_threads` state, so there's no race.

However, it's a separate performance concern from the `scored_at` removal. Consider splitting it into its own commit so the two improvements are independently bisectable/revertable.

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadService.java
```diff
@@ -385,11 +364,11 @@ public Mono<Boolean> addToPendingQueue(@NonNull UUID projectId) {
 
     @Override
     public Mono<Void> openThread(@NonNull UUID projectId, @NonNull String threadId) {
-        return Mono.deferContextual(ctx -> lockService.executeWithLockCustomExpire(
+        return Mono.deferContextual(_ -> lockService.executeWithLockCustomExpire(
                 new LockService.Lock(projectId, TraceThreadService.THREADS_LOCK),
                 Mono.defer(() -> traceThreadDAO.openThread(projectId, threadId))
                         .then(Mono.defer(() -> getOrCreateThreadId(projectId, threadId)))
-                        .doOnSuccess(threadModelId -> log.info("Opened thread for threadId '{}' and projectId: '{}'",
+                        .doOnSuccess(_ -> log.info("Opened thread for threadId '{}' and projectId: '{}'",
```
JetoPistola, line 371: 💡 **suggestion** | Style

The `ctx` → `_` and `threadModelId` → `_` unnamed-variable changes here are valid (JDK 22+ / JEP 456), but they're unrelated to the `scored_at` removal. Minor diff noise — consider either squashing these into a separate commit or leaving them for a follow-up cleanup.

JetoPistola:
👋 **Review summary**

**What looks good**
- Thorough removal of `scored_at` across all layers (model, DAO, service, scorers, tests) — nothing left dangling
- The PR description is excellent: clear provenance trail (scored_at was orphaned by #2719), explicit about the ClickHouse column being intentionally retained for rolling-deploy safety
- Test mock cleanup is complete — every `setScoredAt` mock/verify removed from both scorer test suites
- Lock narrowing in `processTraceThreads` is correct: `processThreadAsync` only does idempotent ID resolution, so it's safe outside the lock

**Overall**
Clean, behavior-neutral refactor that removes a write-only field and reduces lock contention. The two-step approach (remove code now, drop column in follow-up) is the right call for ClickHouse rolling deploys.

**Inline comments**: 2 suggestions (both about separating the lock-narrowing + `_` cleanup from the `scored_at` removal for cleaner git history — neither blocking)

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadService.java
andrescrz: If this is the case (idempotent) on the whole logic, we could be more aggressive here in removing the lock. 

Happy to defer this to next PRs.

andrescrz (approved):
See reply to comment. Regardless of that, this LGTM.

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadService.java
thiagohora: Yeah, I want to discuss the lock in our call. For now, I just reduced the lock scope to the management.

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadService.java
thiagohora: Nit adjustment that doesn't polute the PR

PR #4538, opened 2025-12-23
[OPIK-2770] [Python BE] Add optimization log streaming to Redis for Studio

ldaugusto:
## Details

Adds real-time log streaming for Optimization Studio. When users run optimizations, they can now see live console output including Rich-formatted progress bars, status updates, and optimizer metrics.

**Demo video:** https://grid-comet-ml.enterprise.slack.com/files/U06VAFE2L8J/F0A54BNP34M/screenshare_-_2025-12-23_10_49_10_am.mp4

### How it works:
1. Optimizer now runs in an isolated subprocess with captured stdout/stderr
2. `RedisBatchLogCollector` buffers and streams logs to Redis in batches
3. Logs are stored in a Redis LIST with metadata (timestamps, status) in a HASH
4. Rich terminal formatting is preserved (colors, progress bars, tables)
5. Duplicate spinner lines are deduplicated to reduce noise

### Redis key structure:
- `opik:logs:{workspace_id}:{optimization_id}` - LIST of log lines
- `opik:logs:{workspace_id}:{optimization_id}:meta` - HASH with last_append_ts, status

### Environment variables:
- `STUDIO_LOG_TERM_WIDTH` - Terminal width for Rich formatting (default: 80)

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- OPIK-2770

## Testing

Tested locally with multiple optimization runs. Verified:
- Logs stream in real-time to Redis
- Rich formatting (colors, progress bars) is preserved
- Duplicate progress bar spinner lines are filtered
- TTL is set on Redis keys for automatic cleanup

## Documentation

N/A - Internal Python backend changes. Java PR handles the S3 persistence side.

## Summary by CodeRabbit

* **New Features**
  * Optimization jobs now execute in isolated subprocesses with improved error handling and real-time log streaming via Redis for better visibility.
  * Added configuration options for optimization execution timeout and dataset sampling limits.

* **Configuration**
  * New `REDIS_URL` environment variable for log streaming backend.
  * Added optimization studio configuration options to Helm deployment.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #7133, opened 2026-06-16
[OPIK-6924] [BE][FE] feat: simplify Optimization Studio deployment, fix optimization runs, and add separate algorithm model

alexkuzmik:
## Details

Follow-up to OPIK-6652/OPIK-6924. Simplifies Optimization Studio deployment, fixes the bugs that stopped optimization runs from succeeding across providers and prompt shapes, adds a separate optimizer-model option, hardens observability, and adds automated regression coverage on a cleaned-up python-backend test layout.

### 1. Deployment simplification
- Removed the `TOGGLE_OPTIMIZATION_STUDIO_ENABLED` feature toggle entirely (BE `ServiceTogglesConfig` + `config.yml`/`config-test.yml`, docker-compose, FE `FeatureToggleKeys` enum + default + all gating usages). The studio is now governed solely by the `OPTIMIZATION_STUDIO_USE` / `canUseOptimizationStudio` RBAC permission, matching the OPIK-6434 toggle cleanups.
- No provider API keys are injected anywhere for the deployment — the optimizer resolves provider keys from the workspace, server-side, via the backend gateway.
- **Optimization Studio RQ worker now runs by default.** The worker runs in the python-backend when `RQ_WORKER_ENABLED=true`, but the code default was `false` and only docker-compose set it — so on k8s studio jobs enqueued to `opik:optimizer-cloud` were never consumed and runs sat in `initialized` forever. Flipped the **code default to `true`** (`is_rq_worker_enabled()`), since running the Studio worker is the python-backend's primary role; it now works out-of-the-box on any deployment with Redis. Executor-only instances without Redis can opt out via `RQ_WORKER_ENABLED=false`. (Docstrings restored to "default: true"; no Helm override needed.)

### 2. Bug fixes (optimization runs across providers & prompt shapes)
- **Selected model was ignored** — the runner built `ChatPrompt(messages=...)` with no model, so baseline/per-trial evaluations silently used the SDK default `openai/gpt-5-nano`. Latent regression from #4538 that became a hard failure once #6652 stopped injecting an OpenAI env key. The prompt now carries the configured model (with the factory's `max_tokens` default applied so task calls don't truncate).
- **GEPA `integer modulo by zero` / hierarchical no-op** — the studio inherited the default `optimize_prompts="system"`, so a system-less prompt gave the optimizer zero editable components and GEPA's round-robin selector divided by zero. `optimize_prompts` is now derived from the roles actually present.
- **`temperature is deprecated for this model`** — default `temperature` was sent to models that reject sampling params (e.g. Claude Opus 4.8). Params are stripped via the playground's `updateProviderConfig` + `sanitizeConfigForRequest`; the missing Opus 4.8 capability row was added.
- **Empty model picker / doomed submit** — the form defaults to the first available model and blocks submission when no provider is configured, guiding the user to provider setup.

### 3. New feature: separate model for the algorithm
- The optimizer/algorithm (GEPA's reflection LM, hierarchical's reasoning model) can now run on a **different model than the prompt**, carried in `studio_config.optimizer.parameters.model` (params preserved even when only `model_parameters` is set). FE adds an **"Algorithm model"** selector defaulting to the prompt model, with a "Use prompt model" reset.

### 4. Observability
- The optimization runs in an isolated subprocess; on failure it now emits the full traceback in its result and `process_optimizer_job` propagates it, so error details are visible in CI even though they happen in another process.

### 5. Onboarding doc
- Adds `apps/opik-python-backend/docs/OPTIMIZATION_STUDIO_ONBOARDING.md` (logical + infrastructural design of the studio).

## Change checklist
- [x] User facing
- [x] Documentation update

## Issues
- Resolves #
- OPIK-6924 (relates to OPIK-6652)

## AI-WATERMARK
AI-WATERMARK: yes
- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: investigation, all code changes, and verification in this PR
- Human verification: author ran the studio end-to-end (UI via Playwright + REST) against a local Docker stack and inspected results/logs

### python-backend test layout
- **`apps/opik-python-backend/tests/unit`** — fast tests, no real backend (Flask client, mocks, local subprocess/docker executors). Run by the **Python Backend Tests** workflow (`pytest unit`).
- **`apps/opik-python-backend/tests/e2e`** — tests that need a running Opik backend. Run by the **Python BE E2E Tests** workflow (`pytest tests/e2e`), which boots the full stack via `opik.sh` and triggers on any `apps/opik-python-backend/**` change. The studio log collector is pointed at the stack's real Redis.
- Real model ids are centralized in **`tests/llm_constants.py`** (mirrors the Python SDK convention) so a model bump is a one-line change shared by both suites.

### Studio e2e regression suite (`tests/e2e/test_studio_optimization.py`)
- Drives the **real entrypoint** `process_optimizer_job` → `optimizer_runner.py` subprocess via a `run_studio_optimization` fixture, so the production gateway wiring / `openai/` model prefix / `ChatPrompt(model=...)` are exercised. The Anthropic key is stored in the workspace and never passed to the optimizer — calls route through the `/v1/private` gateway.
- **Three tests:** GEPA and hierarchical (each with an `equals` metric), plus a **code-metric variant** that runs a user `BaseMetric` through the executor inside the optimization subprocess.
- **Assertions on every test:** health signals (baseline established, score in range, no regression below baseline); a **well-formed optimized prompt** (a non-empty list of `role/content` messages that still carries the dataset variable, not just truthy); and **trace-based model verification** (the configured model actually ran, ≥2 calls, and the SDK default `gpt-5-nano` never leaked).
- **Independent + self-cleaning:** each test gets a unique project, dataset, and optimization record, all deleted on teardown. Fully typed.
- **Flakiness protection:** trace polling for ClickHouse eventual consistency, tiny dataset + `OPTIMIZER_MAX_TRIALS=1`, `--reruns` for transient failures, `--log-cli-level=INFO` to stream the subprocess error/traceback, `--durations=0` for per-test timing. Skips cleanly without a key/backend.
- **Validated by fault injection:** reintroducing the model-not-passed bug, and a "runs healthy but on the wrong model" bug, both turn the suite red — the latter caught only by the trace assertion. (Two cloud-gateway-only safeguards are not catchable locally; see the coverage note below.)

### Separate-algorithm-model + model/param wiring
- Covered by deterministic unit tests (`tests/unit/test_studio_optimizer_model.py`) asserting the prompt and the optimizer each use their configured model + params. The optimizer's reasoning-LM calls aren't logged as spans, so they can't be trace-verified in e2e.

### Known coverage gap
- The e2e runs against a **self-hosted** stack (locally and in CI), so cloud/multi-tenant gateway strictness — the explicit `stream` field (Java Anthropic mapper) and the `Comet-Workspace` auth header — isn't exercised; a regression there has no observable effect locally and so isn't caught by any assertion. Confirmed via fault injection. Closing it would require a test tier running against a multi-tenant gateway.

### Manual (local Docker stack, no provider env vars, toggle removed)
| Model | GEPA | Hierarchical |
|---|---|---|
| Claude Haiku 4.5 | ✅ completed | ✅ completed |
| Claude Opus 4.8 | ✅ completed | ✅ completed |

**Separate algorithm model — verified via logs + traces.** Ran a GEPA optimization from the UI with a *different* model for the algorithm than the prompt (Anthropic key in the workspace, gateway-routed, no provider env vars). Each role used exactly its configured model:

| Role | Configured | Verified in the run |
|---|---|---|
| Chat / task (prompt evaluation) | `claude-sonnet-4-6` | task-call spans on `claude-sonnet-4-6` |
| Algorithm (GEPA reflection) | `claude-opus-4-8` | subprocess log `Using GepaOptimizer with the parameters: - model: openai/claude-opus-4-8`; reflection spans on `claude-opus-4-8` |

FE `npm run typecheck` + `eslint` on changed files — clean.

## Documentation
`apps/opik-python-backend/docs/OPTIMIZATION_STUDIO_ONBOARDING.md`.

## Follow-ups (not in this PR)
- **Helm (comet-helm)**: remove the `TOGGLE_OPTIMIZATION_STUDIO_ENABLED` override. The worker no longer needs a chart setting (code defaults `RQ_WORKER_ENABLED=true`); just ensure `comet-helm` doesn't pin it to `false` and that the python-backend has `REDIS_URL`.
- **fern regeneration**: generated REST clients + OpenAPI specs still list `optimizationStudioEnabled`; regenerate from the updated backend.
- **Upstream hardening (optional)**: guard the empty seed-candidate in `sdks/opik_optimizer` (GEPA) as defense-in-depth.
- **Cloud-gateway e2e tier (optional)**: a multi-tenant-gateway test to cover the strictness gap noted above.


apps/opik-python-backend/src/opik_backend/jobs/optimizer_runner.py
```diff
@@ -189,6 +189,70 @@ async def acompletion_with_workspace(*args, **kwargs):
...
+    )
+
+    config.model = _gateway_model(config.model)
+    config.model_params = _with_stream(config.model_params)
+
+    if config.optimizer_model:
+        optimizer_model = _gateway_model(config.optimizer_model)
+        optimizer_model_params = _with_stream(config.optimizer_model_params)
+    else:
+        optimizer_model = config.model
```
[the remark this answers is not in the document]
alexkuzmik: Fixed in 003bb5aeb6 — the `else` branch now keeps `config.optimizer_model_params` (defaulting only the *model* to the prompt model) instead of replacing them with the prompt's params, so `optimizer.parameters.model_parameters` set without a model is no longer silently dropped. Added a unit test (`test_optimizer_params_preserved_without_separate_model`) covering this shape.

apps/opik-python-backend/src/opik_backend/jobs/optimizer_runner.py
```diff
@@ -189,6 +190,77 @@ async def acompletion_with_workspace(*args, **kwargs):
...
+
+def build_optimizer_and_prompt(config):
+    """Build the optimizer and the prompt for a parsed ``OptimizationConfig``.
+
+    The prompt (task evaluation) uses the configured prompt model and its
+    parameters; the optimizer/algorithm uses its own configured model and
+    parameters (GEPA's reflection LM, hierarchical's reasoning model), defaulting
+    to the prompt model when none was picked. Both models are routed through the
+    gateway. Mutates ``config.model``/``config.model_params`` to their
+    gateway-routed form so the metric (built afterwards) uses the same model.
```
petrotiurin: Given that this method mutates config, can we return `(optimizer, prompt, config)` tuple? This way it's clearer that config is modified and easier to follow. Currently the method name obscures any modifications done so when reading the parent function I would not assume any changes were made.

apps/opik-python-backend/src/opik_backend/studio/helpers.py
```diff
@@ -114,12 +114,27 @@ def run_optimization(
...
+    # optimizer with zero editable components — GEPA then divides by zero while
+    # round-robin selecting one. Optimize whichever roles the prompt actually
+    # contains so a run succeeds for any prompt shape.
+    present_roles = sorted(
+        {
+            message.get("role")
+            for message in prompt.get_messages()
+            if message.get("role") in {"system", "user", "assistant"}
+        }
+    )
```
petrotiurin, line 128: is it possible to have several messages of the same role? should we use a set instead?

petrotiurin (commented):
Apart from this python part looks good

apps/opik-python-backend/tests/e2e/test_studio_optimization.py
```diff
@@ -0,0 +1,194 @@
...
+        "evaluation": {"metrics": [metric]},
+        "optimizer": {"type": optimizer_type, "parameters": {"seed": 42}},
+    }
+
+
+def _assert_optimization_healthy(result: dict[str, Any]) -> None:
+    """Signals that the optimization actually ran end-to-end."""
+    assert result is not None, "no result returned"
+    assert "error" not in result, f"optimization errored: {result.get('error')}"
+    assert result.get("status") != "cancelled", "optimization was cancelled"
```
JetoPistola: 🧹 **nit** | Testing

These two checks are unreachable on the path that hits this helper. `process_optimizer_job` *raises* on an error result, and returns a `{"status": "cancelled", ...}` dict only on cancellation — so a successfully-returned `result` reaching `_assert_optimization_healthy` never carries an `error` key or a `cancelled` status. They're harmless defensive asserts, but if a subprocess error ever did surface here it'd be as a raised exception, not these asserts. Safe to drop, or keep as belt-and-suspenders — just flagging they don't currently fire.

apps/opik-python-backend/tests/e2e/test_studio_optimization.py
```diff
@@ -0,0 +1,194 @@
...
+    # Baseline established + a score produced, both in range.
+    assert result.get("initial_score") is not None, "no baseline score (it didn't establish a baseline)"
+    assert 0.0 <= result["initial_score"] <= 1.0, f"baseline {result['initial_score']} out of range"
+    assert result.get("score") is not None, "no final score"
+    assert 0.0 <= result["score"] <= 1.0, f"score {result['score']} out of range"
+    # Optimization shouldn't make the prompt worse than the baseline.
+    assert result["score"] >= result["initial_score"], (
+        f"optimized score {result['score']} regressed below baseline {result['initial_score']}"
+    )
```
JetoPistola, line 101: ❓ **question** | Testing / Flakiness

This is generally safe — the SDK seeds `current_best_score` with the baseline and only advances it on a strict improvement, so the returned best score is `>= initial_score` by construction. But with `OPTIMIZER_MAX_TRIALS=1`, a 4-item dataset, and a non-deterministic LLM eval, a noisy *re-measurement of the baseline* could in principle land below the first baseline reading and trip this. `--reruns 2` masks a transient flake but not a systematic one.

Was a small tolerance (e.g. `>= initial_score - epsilon`) considered, or is the baseline measured once and reused so this can't drift? Just want to make sure this assertion isn't a future CI-flake source.

JetoPistola:
👋 **Review summary**

**What looks good**
- Clean, complete removal of the `TOGGLE_OPTIMIZATION_STUDIO_ENABLED` toggle — BE config + `ServiceTogglesConfig`, FE enum/default/all gating call sites, and docker-compose all updated together, with gating cleanly collapsed onto the `canUseOptimizationStudio` RBAC permission.
- The `build_optimizer_and_prompt` refactor reads well: prompt vs. algorithm model separation is explicit, the gateway-routing + `stream`/`max_tokens` defaults are centralized in small named helpers, and the in-place `config` mutation is intentional and documented (so the metric reuses the routed model).
- Strong typed result contracts (`OptimizationRunResult` / `ErrorResult` / `CancelledResult`) replacing bare `Dict[str, Any]`, plus the subprocess-traceback propagation — genuinely improves observability of failures that happen in a child process.
- Excellent test layout: unit/e2e split with a dedicated CI workflow, centralized `llm_constants.py`, and the e2e suite drives the *real* `process_optimizer_job` entrypoint. The trace-based "configured model actually ran, default never leaked" assertion is a nice touch, and the fault-injection validation in the description is exactly the right rigor.
- The `OPTIMIZATION_STUDIO_ONBOARDING.md` is unusually thorough and will save future maintainers real time.

**Overall**
Well-scoped follow-up that fixes the right root causes (model not carried onto the prompt, system-less prompt → GEPA divide-by-zero, deprecated sampling params) and adds the separate-algorithm-model feature cleanly. Approach is sound; the known cloud-gateway coverage gap is honestly documented. No structural concerns.

**Inline comments**: 1 nit + 1 question — nothing blocking.

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7133`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

apps/opik-python-backend/src/opik_backend/jobs/optimizer_runner.py
alexkuzmik: Fixed in 9bde098084 — `build_optimizer_and_prompt` no longer mutates `config`. The gateway-routed task model/params are resolved into locals, and the caller now reads the resolved model from `prompt.model` when building the metric (the only downstream consumer of the old `config.model` mutation). Clearer data flow, no hidden side effects.

apps/opik-python-backend/src/opik_backend/studio/helpers.py
alexkuzmik: It already uses a set — `present_roles = sorted({ ... for message in prompt.get_messages() ... })` is a set comprehension, so several messages of the same role collapse to one. We `sorted(...)` it into a list only because `optimize_prompts` expects a list of roles. So duplicates are already handled; no change needed.

apps/opik-python-backend/tests/e2e/test_studio_optimization.py
alexkuzmik: Fixed in 9bde098084 — dropped the dead `assert "error" not in result` (an error result raises inside `process_optimizer_job`, so it never reaches this helper). Kept the `cancelled` guard: a cancellation *returns* a `{"status": "cancelled"}` dict (no raise), which would otherwise fail later with a confusing `KeyError` on `initial_score`.

apps/opik-python-backend/tests/e2e/test_studio_optimization.py
alexkuzmik: Holds by construction, so no epsilon needed. The SDK measures the baseline once (→ `initial_score`), seeds the running best with it, and only advances on a strict improvement — it doesn't re-measure the baseline. So the returned best is `>= initial_score` deterministically, independent of LLM noise or `OPTIMIZER_MAX_TRIALS`. The strict `>=` is intentional — it's meant to catch a real regression below baseline, which an epsilon would mask. `--reruns` covers transient LLM/network flakes, not this invariant.

CometActions:

### Access Information
- **URL:** https://pr-7133.dev.comet.com
- **Namespace:** pr-7133
- **Version:** 2.0.69-7133-merge-2297
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7133%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

apps/opik-python-backend/src/opik_backend/studio/helpers.py
petrotiurin: Right, missed this. Makes sense!

petrotiurin (commented):
Python backend LGTM

awkoy (changes requested):
This `useEffect` (and the manual `form.reset` effect above it) is a seed-a-guess-then-reconcile pattern — let's not leave it. The root issue is that `convertOptimizationStudioToFormData` hardcodes `|| GPT_4O_MINI` synchronously, before we know which models the workspace can actually run, and the effect then patches the form store after `useProviderKeys` lands.

Note it can't just become a `useMemo`: `modelName`/`modelConfig` are RHF state that the resolver (`min(1, "Model is required")`) and the submit payload read from. A render-only computed value would leave the store holding `gpt-4o-mini` and submit a model the gateway can't resolve — the corrected value has to live in the form store.

The effect-free fix is to resolve the default at initialization and let RHF feed it reactively:

1. Compute the available-models list in `OptimizationsNewPage` (lift `useProviderKeys` + `useLLMProviderModelsData` + `useModelOptions` up to where `useForm` lives).
2. Fold the resolved model into `defaultValues` — once provider keys settle, use `base.modelName` if it's available, otherwise the first available model (or `""`).
3. Switch `useForm` from `defaultValues` to the reactive `values` prop with `resetOptions: { keepDirtyValues: true }`:

```ts
const form = useForm<OptimizationConfigFormType>({
  resolver: zodResolver(OptimizationConfigSchema),
  values: defaultValues,                       // re-seeds when provider data lands
  resetOptions: { keepDirtyValues: true },     // preserves a model the user already picked
  mode: "onChange",
});
```

That deletes **both** effects: the manual `form.reset(...)` (now handled by `values`) and the model-sync `useEffect` in `useOptimizationsNewFormHandlers`. `keepDirtyValues` gives us what the sync effect's `availableModels.includes(currentModel)` guard was doing — preserving a valid user choice — for free.

One correctness note for whatever we land on: the current effect gates only on `providerKeysData`. If `providerModels` resolves a tick later, `availableModels` is briefly `[]` while providers are actually configured, so we clear `modelName` with `shouldValidate: true` and — with `mode: "onChange"` — flash "Model is required". Reconcile only once providers are genuinely known (e.g. don't treat "configured providers but no models yet" as "no models"), so we never clear mid-load.

alexkuzmik:
@awkoy addressed in d7b60723b1 — went with the effect-free approach you outlined; both effects are gone.

1. **Lifted the data hooks** — `useProviderKeys` + `useLLMProviderModelsData` + `useModelOptions` now live in `OptimizationsNewPage` next to `useForm`, which computes `availableModels` there.
2. **No more hardcoded default** — `convertOptimizationStudioToFormData` takes `availableModels` and resolves the model in the store: the configured one (rerun/template) if available, else the first available, else `""` (so `min(1, "Model is required")` blocks submit). No `|| GPT_4O_MINI`.
3. **Reactive `values`** — `useForm` switched from `defaultValues` to `values: defaultValues` + `resetOptions: { keepDirtyValues: true }`, `mode: "onChange"`. That deleted **both** the manual `form.reset(...)` effect and the model-sync `useEffect` in `useOptimizationsNewFormHandlers`; `keepDirtyValues` covers what the old `availableModels.includes(currentModel)` guard did.
4. **Mid-load flash** — added a `providersReady` guard (`providerKeysData` settled **and** (no providers configured **or** models loaded)); until then I pass `[]` and don't reseed, so "configured providers, models still loading" never reads as "no models" / flashes "Model is required".

`tsc --noEmit` + `eslint` clean on the three files.

CometActions:

### Access Information
- **URL:** https://pr-7133.dev.comet.com
- **Namespace:** pr-7133
- **Version:** 2.0.69-7133-merge-2301
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7133%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

awkoy (dismissed):

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7133`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

deployment/helm_chart/opik/values.yaml
```diff
@@ -277,6 +277,13 @@ component:
       OPIK_URL_OVERRIDE: "http://opik-backend:8080"
       # Redis URL for Optimization Studio log streaming
       REDIS_URL: redis://:wFSuJX9nDBdCa25sKZG7bh@opik-redis-master:6379/
+      # Run the Optimization Studio RQ worker in this service. Required for
+      # Studio runs to execute: the Java backend enqueues jobs to the
+      # "opik:optimizer-cloud" queue, and without a worker consuming it they
+      # stay stuck in "initialized". The code default is "false" (the
+      # python-backend is also a standalone code executor), so it must be
+      # enabled explicitly here for k8s deployments.
+      RQ_WORKER_ENABLED: "true"
```
andrescrz: Minor: I believe this defaults to `true`, so it seems redundant. Not sure of the purpose of this, but anyhow it's fine as it can be overwritten if needed.

andrescrz (dismissed):
LGTM, just a minor comment.

deployment/helm_chart/opik/values.yaml
alexkuzmik: Follow-up: you were right that this *should* be the default — so rather than set it per-chart, I flipped the **code** default to `true` (`is_rq_worker_enabled()` now defaults `"true"`), since running the Studio job worker is the python-backend's primary role. That makes this Helm line genuinely redundant, so I **removed it** in b392cd949c (and restored the `rq_worker_manager` docstrings to "default: true", now accurate).

Net: the Studio works out-of-the-box on k8s without any chart override. An executor-only python-backend without Redis can still opt out with `RQ_WORKER_ENABLED=false`.

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7133`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7133`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7133`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

PR #7138, opened 2026-06-17
[OPIK-6968] [BE] perf: bulk thread resolve/create + scope open/close locks to non-idempotent ops

thiagohora:
## Details

Follow-up to #7121 and #7131 (both merged). Tightens the remaining trace-thread lock usage and removes per-thread DB round-trips on the open/close paths.

Why keep the project-based lock instead of going per-thread: #7121 and #7131 already removed the lock from the two most expensive acquirers — trace ingestion (`processTraceThreads`) and the online-scoring writes (`setScoredAt` / sampling) — which were the real contention against ingestion and online scoring. What remains (`openThread` / `closeThreads`) is low-frequency, and going per-thread there would be a net loss (a batch of N threads → N locks + N DB round-trips vs one bulk op under one lock). So we keep the project lock and shrink what runs under it to the genuinely non-idempotent work.

- `openThread`: resolve the thread-model id (`getOrCreateThreadId`, idempotent) outside the lock; the lock wraps only the status flip.
- `closeThreads`: lock only the non-idempotent closure write; the post-close read (idempotent) and the online-scoring publish (a Redis-stream write) run outside the lock.
- `verifyAndCreateThreadIfNeed`: fully bulk — one bulk get-or-create of thread-model ids (`getOrCreateTraceThreadIds`: bulk read + batch `INSERT`, with a constraint-violation retry in a fresh transaction so it converges under REPEATABLE READ), one bulk existence check, one batch `save` + a single `TraceThreadsCreated` event. For N threads this goes from ~3N sequential round-trips + N events to ~3 queries + 1 event.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- OPIK-6968

## AI-WATERMARK

AI-WATERMARK: yes

- If yes:
  - Tools: Claude Code
  - Scope: Scoped open/close locks to non-idempotent ops; bulked the thread resolve/create path (new bulk DAO + service methods); audited that every read-modify-write still reads the latest version.

## Testing

- `mvn test-compile` (JDK 25) — BUILD SUCCESS (main + test).
- `mvn spotless:check` — passed.
- Not run locally: full `mvn test` (DB-backed integration suites, incl. JDBI batch + constraint-retry behavior) — relying on CI.

## Documentation

N/A — internal performance refactor; no API or user-facing documentation changes.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadIdDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadIdDAO.java
@@ -8,7 +8,6 @@ import org.jdbi.v3.sqlobject.customizer.BindList;
 import org.jdbi.v3.sqlobject.customizer.BindMethods;
 import org.jdbi.v3.sqlobject.statement.SqlBatch;
 import org.jdbi.v3.sqlobject.statement.SqlQuery;
-import org.jdbi.v3.sqlobject.statement.SqlUpdate;
 
 import java.util.List;
 import java.util.UUID;
@@ -17,23 +16,15 @@ import java.util.UUID;
 @RegisterConstructorMapper(TraceThreadIdModel.class)
 interface TraceThreadIdDAO {
 
-    @SqlUpdate("""
-                INSERT INTO project_trace_threads(id, project_id, thread_id, created_by, created_at)
-                VALUES (:thread.id, :thread.projectId, :thread.threadId, :thread.createdBy, :thread.createdAt)
-            """)
-    void save(@BindMethods("thread") TraceThreadIdModel rule);
-
     @SqlBatch("""
                 INSERT INTO project_trace_threads(id, project_id, thread_id, created_by, created_at)
                 VALUES (:thread.id, :thread.projectId, :thread.threadId, :thread.createdBy, :thread.createdAt)
             """)
     void save(@BindMethods("thread") List<TraceThreadIdModel> threads);
 
-    @SqlQuery("""
-                SELECT * FROM project_trace_threads
-                WHERE project_id = :projectId AND thread_id = :threadId
-            """)
-    TraceThreadIdModel findByProjectIdAndThreadId(@Bind("projectId") UUID projectId, @Bind("threadId") String threadId);
+    default void save(TraceThreadIdModel thread) {
+        save(List.of(thread));
+    }
 
     @SqlQuery("""
                 SELECT * FROM project_trace_threads
@@ -41,6 +32,10 @@ interface TraceThreadIdDAO {
             """)
     List<TraceThreadIdModel> findByProjectIdAndThreadIds(@Bind("projectId") UUID projectId,
             @BindList("threadIds") List<String> threadIds);
+
+    default TraceThreadIdModel findByProjectIdAndThreadId(UUID projectId, String threadId) {
+        return findByProjectIdAndThreadIds(projectId, List.of(threadId)).stream().findFirst().orElse(null);
+    }
 
     @SqlQuery("""
                 SELECT * FROM project_trace_threads
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadIdDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadIdDAO.java
@@ -10,6 +10,7 @@ import org.jdbi.v3.sqlobject.statement.SqlBatch;
 import org.jdbi.v3.sqlobject.statement.SqlQuery;
 
 import java.util.List;
+import java.util.Optional;
 import java.util.UUID;
 
 @RegisterArgumentFactory(UUIDArgumentFactory.class)
@@ -33,8 +34,8 @@ interface TraceThreadIdDAO {
     List<TraceThreadIdModel> findByProjectIdAndThreadIds(@Bind("projectId") UUID projectId,
             @BindList("threadIds") List<String> threadIds);
 
-    default TraceThreadIdModel findByProjectIdAndThreadId(UUID projectId, String threadId) {
-        return findByProjectIdAndThreadIds(projectId, List.of(threadId)).stream().findFirst().orElse(null);
+    default Optional<TraceThreadIdModel> findByProjectIdAndThreadId(UUID projectId, String threadId) {
+        return findByProjectIdAndThreadIds(projectId, List.of(threadId)).stream().findFirst();
     }
 
     @SqlQuery("""
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadIdDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadIdDAO.java
@@ -6,6 +6,7 @@ import org.jdbi.v3.sqlobject.config.RegisterConstructorMapper;
 import org.jdbi.v3.sqlobject.customizer.Bind;
 import org.jdbi.v3.sqlobject.customizer.BindList;
 import org.jdbi.v3.sqlobject.customizer.BindMethods;
+import org.jdbi.v3.sqlobject.statement.SqlBatch;
 import org.jdbi.v3.sqlobject.statement.SqlQuery;
 import org.jdbi.v3.sqlobject.statement.SqlUpdate;
 
@@ -22,11 +23,24 @@ interface TraceThreadIdDAO {
             """)
     void save(@BindMethods("thread") TraceThreadIdModel rule);
 
+    @SqlBatch("""
+                INSERT INTO project_trace_threads(id, project_id, thread_id, created_by, created_at)
+                VALUES (:thread.id, :thread.projectId, :thread.threadId, :thread.createdBy, :thread.createdAt)
+            """)
+    void save(@BindMethods("thread") List<TraceThreadIdModel> threads);
+
     @SqlQuery("""
                 SELECT * FROM project_trace_threads
                 WHERE project_id = :projectId AND thread_id = :threadId
             """)
     TraceThreadIdModel findByProjectIdAndThreadId(@Bind("projectId") UUID projectId, @Bind("threadId") String threadId);
+
+    @SqlQuery("""
+                SELECT * FROM project_trace_threads
+                WHERE project_id = :projectId AND thread_id IN (<threadIds>)
+            """)
+    List<TraceThreadIdModel> findByProjectIdAndThreadIds(@Bind("projectId") UUID projectId,
+            @BindList("threadIds") List<String> threadIds);
 
     @SqlQuery("""
                 SELECT * FROM project_trace_threads
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadIdService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadIdService.java
@@ -1,5 +1,6 @@
 package com.comet.opik.domain.threads;
 
+import com.comet.opik.api.error.EntityAlreadyExistsException;
 import com.comet.opik.domain.EntityConstraintHandler;
 import com.comet.opik.domain.IdGenerator;
 import com.comet.opik.domain.ProjectService;
@@ -9,6 +10,7 @@ import com.comet.opik.infrastructure.db.TransactionTemplateAsync;
 import com.google.common.base.Preconditions;
 import com.google.inject.ImplementedBy;
 import com.google.inject.Singleton;
+import io.dropwizard.jersey.errors.ErrorMessage;
 import jakarta.inject.Inject;
 import jakarta.ws.rs.NotFoundException;
 import lombok.NonNull;
@@ -20,8 +22,10 @@ import reactor.core.scheduler.Schedulers;
 import ru.vyarus.guicey.jdbi3.tx.TransactionTemplate;
 
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.UUID;
 import java.util.stream.Collectors;
 
@@ -30,6 +34,9 @@ interface TraceThreadIdService {
 
     Mono<TraceThreadIdModel> getOrCreateTraceThreadId(String workspaceId, UUID projectId, String threadId,
             Instant timestamp);
+
+    Mono<List<TraceThreadIdModel>> getOrCreateTraceThreadIds(String workspaceId, UUID projectId,
+            Map<String, Instant> threadIdToTimestamp);
 
     Mono<UUID> getThreadModelId(String workspaceId, UUID projectId, String threadId);
 
@@ -62,6 +69,64 @@ class TraceThreadIdServiceImpl implements TraceThreadIdService {
                         .switchIfEmpty(createThread(threadId, projectId, timestamp)))
                 .subscribeOn(Schedulers.boundedElastic())
                 .switchIfEmpty(Mono.error(new NotFoundException("Project not found: " + projectId)));
+    }
+
+    @Override
+    public Mono<List<TraceThreadIdModel>> getOrCreateTraceThreadIds(@NonNull String workspaceId,
+            @NonNull UUID projectId, @NonNull Map<String, Instant> threadIdToTimestamp) {
+        Preconditions.checkArgument(!StringUtils.isBlank(workspaceId), "Workspace ID cannot be blank");
+
+        if (threadIdToTimestamp.isEmpty()) {
+            return Mono.just(List.of());
+        }
+
+        return Mono.deferContextual(context -> Mono.fromCallable(() -> {
+            List<String> threadIds = List.copyOf(threadIdToTimestamp.keySet());
+            String userName = context.get(RequestContext.USER_NAME);
+
+            // Each attempt runs in its own transaction: a constraint-violation retry must re-read a
+            // fresh snapshot, and under REPEATABLE READ re-reading within the same transaction would
+            // reuse the original snapshot and never converge.
+            return EntityConstraintHandler.<List<TraceThreadIdModel>>handle(
+                    () -> transactionTemplate.inTransaction(TransactionTemplateAsync.WRITE, handle -> {
+                        var dao = handle.attach(TraceThreadIdDAO.class);
+
+                        List<TraceThreadIdModel> existing = dao.findByProjectIdAndThreadIds(projectId, threadIds);
+                        Set<String> existingThreadIds = existing.stream()
+                                .map(TraceThreadIdModel::threadId)
+                                .collect(Collectors.toSet());
+
+                        List<String> missingThreadIds = threadIds.stream()
+                                .filter(threadId -> !existingThreadIds.contains(threadId))
+                                .toList();
+
+                        if (missingThreadIds.isEmpty()) {
+                            // Common case: every thread already has an id — no write needed.
+                            return existing;
+                        }
+
+                        // We already know exactly which threads are missing, so we own their ids.
+                        List<TraceThreadIdModel> created = missingThreadIds.stream()
+                                .map(threadId -> TraceThreadIdModel.builder()
+                                        .id(generateThreadModelId(threadIdToTimestamp.get(threadId)))
+                                        .projectId(projectId)
+                                        .threadId(threadId)
+                                        .createdBy(userName)
+                                        .createdAt(Instant.now())
+                                        .build())
+                                .toList();
+
+                        // Plain INSERT: a concurrent create trips the (project_id, thread_id) constraint
+                        // and triggers a retry, which re-reads the now-smaller missing set.
+                        dao.save(created);
+
+                        List<TraceThreadIdModel> result = new ArrayList<>(existing);
+                        result.addAll(created);
+                        return result;
+                    }))
+                    .withRetry(3, () -> new EntityAlreadyExistsException(
+                            new ErrorMessage("Failed to resolve trace thread ids for project: " + projectId)));
+        }).subscribeOn(Schedulers.boundedElastic()));
     }
 
     @Cacheable(name = "GET_TRACE_THREAD_ID", key = "$workspaceId +'-'+ $projectId +'-'+ $threadId", returnType = UUID.class)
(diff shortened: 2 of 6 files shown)


apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadService.java
```diff
@@ -355,37 +357,40 @@ public Mono<Boolean> addToPendingQueue(@NonNull UUID projectId) {
...
-                                threadId, projectId))
-                        .then(),
-                LOCK_DURATION));
+    public Mono<Void> openThread(UUID projectId, String projectName, @NonNull String threadId) {
+        return projectService.resolveProjectIdAndVerifyVisibility(projectId, projectName)
+                .flatMap(id -> getOrCreateThreadId(id, threadId)
+                        .then(Mono.defer(() -> lockService.executeWithLockCustomExpire(
+                                new LockService.Lock(id, TraceThreadService.THREADS_LOCK),
+                                Mono.defer(() -> traceThreadDAO.openThread(id, threadId)),
+                                LOCK_DURATION)))
```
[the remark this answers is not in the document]
thiagohora: Accepting this as eventually consistent rather than forcing a create here.

This pre-dates the PR (the original ran the update-only DAO first, then `getOrCreateThreadId` — same outcome), and it can only occur when the `trace_threads` row isn't ingested yet or hasn't replicated to the read replica. In both cases the state converges correctly:

- **Replica delay:** ingestion creates `trace_threads` rows as ACTIVE, so a recently-introduced thread is already active elsewhere — the update-only no-op is the correct result (nothing to flip), and the 204 reflects the converged state. `getOrCreateThreadId` is idempotent on `(project_id, thread_id)`, so it returns the same canonical id ingestion uses — no duplicate, no orphan.
- **Not yet ingested:** the `project_trace_threads` row isn't an orphan — it's a pre-allocated mapping that ingestion (`getOrCreateTraceThreadId`, keyed by the same `(project_id, thread_id)`) reuses when the traces land, at which point the `trace_threads` row is created under that id.

Routing `openThread` through `verifyAndCreateThreadsIfNeeded` would add a trace-table read on every open and could 404 a thread that's mid-replication — worse than the benign no-op. So no change; documented as eventual consistency.

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadService.java
```diff
@@ -355,37 +357,40 @@ public Mono<Boolean> addToPendingQueue(@NonNull UUID projectId) {
...
+                                                "Closed count '{}' for threadIds '{}' and projectId: '{}'",
+                                                count, threadIds, id)),
+                                LOCK_DURATION)))
+                        // The post-close read (idempotent) and online-scoring publish run outside the lock.
+                        .then(Mono.defer(() -> traceThreadDAO.findThreadsByProject(1, threadIds.size(),
+                                TraceThreadCriteria.builder()
+                                        .projectId(id)
+                                        .threadIds(threadIds)
+                                        .build())))
+                        .flatMap(threadModels -> checkAndTriggerOnlineScoring(id, threadModels)));
```
JetoPistola: ❓ **question** | Concurrency

The post-close re-read (`findThreadsByProject`) and the online-scoring publish (`checkAndTriggerOnlineScoring`) now run **outside** the `THREADS_LOCK`. That looks right for the lock-contention goal, but it opens a small window: if a manual `closeThreads` and the scheduled `processProjectWithTraceThreadsPendingClosure` run for the same project concurrently, both can re-read the same just-closed threads and enqueue them for online scoring — i.e. a thread could be published twice.

If the online-scoring consumer is idempotent / at-least-once this is fine and intentional. Just want to confirm that's the assumption, since previously the read+publish sat inside the lock (so they were serialized per project).

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadService.java
```diff
@@ -397,27 +402,80 @@ private Mono<Void> verifyAndCreateThreadsIfNeeded(UUID projectId, Set<String> th
...
+            List<TraceThread> threads) {
+        Map<String, Instant> needIds = threads.stream()
+                .filter(thread -> thread.threadModelId() == null)
+                .collect(Collectors.toMap(TraceThread::id, TraceThread::createdAt, (a, b) -> a));
+
+        if (needIds.isEmpty()) {
+            return Mono.just(threads);
+        }
+
+        return traceThreadIdService.getOrCreateTraceThreadIds(workspaceId, projectId, needIds)
```
JetoPistola, line 432: 💡 **suggestion** | Testing

The new bulk resolve/create path is the riskiest part of this PR — `getOrCreateTraceThreadIds` does a plain batch `INSERT` and relies on the `(project_id, thread_id)` constraint + `withRetry(3)` re-reading a fresh snapshot to converge under REPEATABLE READ. The 3-thread `closeMultipleTraceThreads__happyPath` exercises the no-contention path, but the convergence-under-concurrent-create behavior the PR description emphasizes isn't covered by a test.

A test that forces two concurrent `closeThreads` (or a pre-seeded partial set) on overlapping thread ids would lock in the constraint-retry behavior and guard against future regressions. Acknowledging this is hard to make deterministic — flagging as non-blocking.

JetoPistola:
👋 **Review summary**

**What looks good**
- The `validateProjectIdentifier` → `resolveProjectIdAndVerifyVisibility` swap is more than a refactor — it tightens workspace-scoping (`findByIds(workspaceId, …)`), and the two new cross-workspace 404 tests lock that in. Nice catch to add those.
- Lock scoping is exactly right: idempotent work (id resolution, post-close read) moved out, only the non-idempotent status flip stays under `THREADS_LOCK`. The reasoning in the PR description for keeping the project-level lock instead of going per-thread is well-argued.
- The bulk `getOrCreateTraceThreadIds` constraint-retry is correctly built — fresh transaction per retry so it actually converges under REPEATABLE READ, with a clear comment explaining why.

**Overall**
Solid performance refactor with a clean reduction from ~3N round-trips + N events to ~3 queries + 1 event on the close path. Approach and scope are appropriate. I verified the `ids` criteria is wired into the DAO SQL and ruled out a suspected cache-bypass bug (the `@Cacheable` interceptor doesn't cache empty results, so no negative-caching hazard).

**Inline comments**: 1 question (concurrent double-publish to online scoring) + 1 suggestion (test the constraint-retry path) — nothing blocking.

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadService.java
thiagohora: Confirmed — at-least-once is the intended assumption, and it's safe:

- Online scoring is queue-based (Redis stream consumer groups), so it's at-least-once by design regardless of this change.
- Re-scoring is idempotent at the data layer: feedback scores land in `ReplicatedReplacingMergeTree` tables keyed by entity/name (last_updated_at version), so publishing a thread twice overwrites the same row rather than duplicating.
- This window isn't actually new: the scheduled `processProjectWithTraceThreadsPendingClosure` already runs its close + `checkAndTriggerOnlineScoring` outside any lock since #7131, so scan-vs-manual double-publish was already possible. This PR just extends it to manual-vs-manual, under the same at-least-once contract.

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadService.java
thiagohora: To be precise on why this is safe by construction (so correctness doesn't hinge on the test):

- The `(project_id, thread_id)` unique constraint guarantees idempotency — concurrent creators of the same thread can't produce two rows; they converge to a single canonical id.
- `withRetry(3)` only has to cover the read→write gap: if a row appears between our read (missing) and the insert, the insert trips the constraint and we retry in a fresh transaction snapshot, re-read, and return the now-existing id. No orphan, no duplicate.
- The worst case under contention is a thread closed/published for scoring twice — idempotent (online scoring is at-least-once; feedback scores are ReplacingMergeTree keyed, so re-scoring overwrites) and acceptable.

So the unique key gives idempotency and the retry covers the read/write race — the convergence test would be nice-to-have hardening, not a correctness gap.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/TracesResourceTest.java
```diff
@@ -6434,6 +6434,46 @@ void closeMultipleTraceThreads__happyPath() {
 
             TraceAssertions.assertThreads(expectedClosedThreads, closedThreadPage.content());
         }
+
+        @Test
+        @DisplayName("open thread: when projectId belongs to another workspace, then return not found")
+        void openTraceThread__whenProjectBelongsToAnotherWorkspace__thenReturnNotFound() {
```
BorisTkachenko, line 6440: Do we need to add tests when thread belongs to a different project?

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadService.java
```diff
@@ -355,37 +357,40 @@ public Mono<Boolean> addToPendingQueue(@NonNull UUID projectId) {
...
-                new LockService.Lock(projectId, TraceThreadService.THREADS_LOCK),
-                Mono.defer(() -> traceThreadDAO.openThread(projectId, threadId))
-                        .then(Mono.defer(() -> getOrCreateThreadId(projectId, threadId)))
-                        .doOnSuccess(_ -> log.info("Opened thread for threadId '{}' and projectId: '{}'",
-                                threadId, projectId))
-                        .then(),
-                LOCK_DURATION));
+    public Mono<Void> openThread(UUID projectId, String projectName, @NonNull String threadId) {
+        return projectService.resolveProjectIdAndVerifyVisibility(projectId, projectName)
+                .flatMap(id -> getOrCreateThreadId(id, threadId)
```
BorisTkachenko: I would rename here `id` -> `verifiedProjectId` or something like this for readability. In other places too.

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadIdDAO.java
```diff
@@ -22,12 +23,25 @@ INSERT INTO project_trace_threads(id, project_id, thread_id, created_by, created
             """)
     void save(@BindMethods("thread") TraceThreadIdModel rule);
 
+    @SqlBatch("""
+                INSERT INTO project_trace_threads(id, project_id, thread_id, created_by, created_at)
+                VALUES (:thread.id, :thread.projectId, :thread.threadId, :thread.createdBy, :thread.createdAt)
+            """)
+    void save(@BindMethods("thread") List<TraceThreadIdModel> threads);
```
andrescrz, line 24: Please DRY on this method. Let's just have the List flavour and then an overloaded version with a single `TraceThreadIdModel` that delegates to the method with the single implementation.

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadIdDAO.java
```diff
@@ -22,12 +23,25 @@ INSERT INTO project_trace_threads(id, project_id, thread_id, created_by, created
...
                 SELECT * FROM project_trace_threads
                 WHERE project_id = :projectId AND thread_id = :threadId
             """)
     TraceThreadIdModel findByProjectIdAndThreadId(@Bind("projectId") UUID projectId, @Bind("threadId") String threadId);
 
+    @SqlQuery("""
+                SELECT * FROM project_trace_threads
+                WHERE project_id = :projectId AND thread_id IN (<threadIds>)
+            """)
+    List<TraceThreadIdModel> findByProjectIdAndThreadIds(@Bind("projectId") UUID projectId,
```
andrescrz, line 34: Same for this method, let's DRY. Instead: overload and delegate.

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadIdService.java
```diff
@@ -64,6 +71,64 @@ public Mono<TraceThreadIdModel> getOrCreateTraceThreadId(@NonNull String workspa
                 .switchIfEmpty(Mono.error(new NotFoundException("Project not found: " + projectId)));
     }
 
+    @Override
+    public Mono<List<TraceThreadIdModel>> getOrCreateTraceThreadIds(@NonNull String workspaceId,
+            @NonNull UUID projectId, @NonNull Map<String, Instant> threadIdToTimestamp) {
+        Preconditions.checkArgument(!StringUtils.isBlank(workspaceId), "Workspace ID cannot be blank");
```
andrescrz: Let's remove the `NonNull` validations for `workspaceId` (validated as blank below) and for `threadIdToTimestamp` (handled above). For the map, use MapUtils/CollectionUtils `isEmpty` instead.

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadIdService.java
```diff
@@ -64,6 +71,64 @@ public Mono<TraceThreadIdModel> getOrCreateTraceThreadId(@NonNull String workspa
...
+
+                        // Plain INSERT: a concurrent create trips the (project_id, thread_id) constraint
+                        // and triggers a retry, which re-reads the now-smaller missing set.
+                        dao.save(created);
+
+                        List<TraceThreadIdModel> result = new ArrayList<>(existing);
+                        result.addAll(created);
+                        return result;
+                    }))
+                    .withRetry(3, () -> new EntityAlreadyExistsException(
```
andrescrz: Can you explain why there could be a conflict in this case? I'm not sure I fully understand the logic. You could leave a complete, but concise javadoc detailing this.

andrescrz (commented):
Let's go with some improvements before moving forward.

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadIdDAO.java
thiagohora: Done in 24787d1 — `save(List)` is the `@SqlBatch` implementation and `save(TraceThreadIdModel)` is now a `default` that delegates to it.

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadIdDAO.java
thiagohora: Done — `findByProjectIdAndThreadIds(List)` is the `@SqlQuery` implementation; `findByProjectIdAndThreadId(single)` is a `default` delegating to it (`...findFirst().orElse(null)`).

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadIdService.java
thiagohora: Done — dropped `@NonNull` on `workspaceId` (it's blank-checked just below via Preconditions) and on `threadIdToTimestamp` (now guarded with `MapUtils.isEmpty`).

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadService.java
thiagohora: Done — renamed the lambda `id` → `verifiedProjectId` in both `openThread` and `closeThreads`.

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadIdService.java
thiagohora: Added a javadoc on the method. The conflict is the read→write race: we read the existing `(project_id, thread_id)` rows, compute the missing set, then batch-INSERT only those. If a concurrent caller inserts the same `(project_id, thread_id)` between our read and our insert, the unique constraint rejects our insert — that's the `UnableToExecuteStatementException`/constraint violation `withRetry` catches. The retry re-runs the whole read+insert in a fresh transaction (fresh snapshot under REPEATABLE READ), so it now sees the row, drops it from the missing set, and converges. The unique key guarantees one canonical id per thread; the retry only covers that race window.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/TracesResourceTest.java
thiagohora: Added a test (`closeTraceThreads__whenThreadBelongsToAnotherProject__thenReturnNotFound`): a thread is created under project A, then closed via project B in the same workspace → 404, since `verifyAndCreateThreadsIfNeeded` doesn't find the thread under B. Note this applies to close only — `openThread` doesn't go through verify, so the different-project case there is the accepted no-op (it allocates the id mapping and the update-only DAO no-ops; eventual-consistency behavior discussed above), with nothing to assert as a 404.

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadIdDAO.java
```diff
@@ -16,17 +16,26 @@
...
                 SELECT * FROM project_trace_threads
-                WHERE project_id = :projectId AND thread_id = :threadId
+                WHERE project_id = :projectId AND thread_id IN (<threadIds>)
             """)
-    TraceThreadIdModel findByProjectIdAndThreadId(@Bind("projectId") UUID projectId, @Bind("threadId") String threadId);
+    List<TraceThreadIdModel> findByProjectIdAndThreadIds(@Bind("projectId") UUID projectId,
+            @BindList("threadIds") List<String> threadIds);
+
+    default TraceThreadIdModel findByProjectIdAndThreadId(UUID projectId, String threadId) {
+        return findByProjectIdAndThreadIds(projectId, List.of(threadId)).stream().findFirst().orElse(null);
```
andrescrz: Not a blocker: it's better the interface of this method returns an Optional, callers would be the ones handling it and deciding if thy get null or do something else. This is much better and more modern Java than handling nulls.

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadIdService.java
```diff
@@ -64,6 +72,75 @@ public Mono<TraceThreadIdModel> getOrCreateTraceThreadId(@NonNull String workspa
...
+     * between our read and our insert — the unique constraint then rejects our insert. In that case
+     * {@code withRetry} re-runs the whole read-then-insert in a fresh transaction (so a fresh snapshot
+     * under REPEATABLE READ): the retry sees the now-existing row, drops it from the missing set, and
+     * converges. The unique key guarantees a single canonical id per thread; the retry only covers the
+     * read→write race.
+     */
+    @Override
+    public Mono<List<TraceThreadIdModel>> getOrCreateTraceThreadIds(String workspaceId,
+            @NonNull UUID projectId, Map<String, Instant> threadIdToTimestamp) {
+        Preconditions.checkArgument(!StringUtils.isBlank(workspaceId), "Workspace ID cannot be blank");
```
andrescrz: Very nit: you have `StringUtils.isNotBlank`, which is more readable than `!`.

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadIdService.java
```diff
@@ -64,6 +72,75 @@ public Mono<TraceThreadIdModel> getOrCreateTraceThreadId(@NonNull String workspa
...
+                        // Plain INSERT: a concurrent create trips the (project_id, thread_id) constraint
+                        // and triggers a retry, which re-reads the now-smaller missing set.
+                        dao.save(created);
+
+                        List<TraceThreadIdModel> result = new ArrayList<>(existing);
+                        result.addAll(created);
+                        return result;
+                    }))
+                    .withRetry(3, () -> new EntityAlreadyExistsException(
+                            new ErrorMessage("Failed to resolve trace thread ids for project: " + projectId)));
```
andrescrz: Very nit: I suggest printing also the `workspaceId` in addition to the `projectId`. It'd be useful during debugging. 

Nice with `.formatting` instead of using `+`.

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadService.java
```diff
@@ -397,27 +403,80 @@ private Mono<Void> verifyAndCreateThreadsIfNeeded(UUID projectId, Set<String> th
...
+                    .then();
+        });
+    }
+
+    // Bulk-resolve a trace-thread-model id for every thread, allocating ids only for those missing one.
+    private Mono<List<TraceThread>> resolveThreadModelIds(UUID projectId, String workspaceId,
+            List<TraceThread> threads) {
+        Map<String, Instant> needIds = threads.stream()
+                .filter(thread -> thread.threadModelId() == null)
+                .collect(Collectors.toMap(TraceThread::id, TraceThread::createdAt, (a, b) -> a));
```
andrescrz: Minor: if you dedupe on the map collection, I suggest adding some criteria (e.g last_updated_at)

andrescrz (dismissed):
Only minor stuff remaining, this LGTM.

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadIdDAO.java
thiagohora: Done — `findByProjectIdAndThreadId` now returns `Optional<TraceThreadIdModel>`; the two callers use `.orElse(null)` to preserve their current behavior.

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadIdService.java
thiagohora: Done — switched to `StringUtils.isNotBlank(...)` (here and the other occurrences in the file).

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadIdService.java
thiagohora: Done — the resolve-failure `ErrorMessage` now includes the workspaceId and uses `.formatted(projectId, workspaceId)`.

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadService.java
thiagohora: Done — the dedupe now keeps the earliest `createdAt` on a duplicate threadId (`(a, b) -> a.isBefore(b) ? a : b`), matching how the thread-model id is derived from the first trace's timestamp.

andrescrz (approved):

PR #7147, opened 2026-06-17
[OPIK-6982] Fix ClosingTraceThreadSubscriber NPE on null workspace timeout

thiagohora:
## Details

The trace-thread closing subscriber threw `NullPointerException: The mapper [...] returned a null value.` for any workspace whose configuration record has `timeoutToMarkThreadAsInactive = null` (e.g. a config created with only `truncationOnTables` set). As a result, those workspaces' threads were never closed.

- `TraceThreadServiceImpl#getWorkspaceTimeout` used `.map(WorkspaceConfiguration::timeoutToMarkThreadAsInactive).switchIfEmpty(default)`. `switchIfEmpty` only fires on empty completion — not when the config is present but the timeout field is null — so `map()` emitted null, which Reactor rejects.
- Switched to `.mapNotNull(...).defaultIfEmpty(default)` so a null timeout drops to empty and falls back to the default, covering both "no config" and "config with null timeout".

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- OPIK-6982

## AI-WATERMARK

AI-WATERMARK: yes

- If yes:
  - Tools: Claude Code
  - Model(s): Claude Opus 4.8
  - Scope: Root-cause analysis, the one-line operator fix, regression test, and this PR description.
  - Human verification: Author reviewed the diff; `mvn compile` and `mvn test-compile` pass locally.

## Testing

- `mvn -o compile` and `mvn -o test-compile` — BUILD SUCCESS.
- Added regression test `TraceThreadsClosingJobTest#shouldCloseTraceThreadsWhenWorkspaceConfigurationTimeoutIsNull`: creates a workspace configuration with only `truncationOnTables` (null `timeoutToMarkThreadAsInactive`), ingests a thread, and asserts it is closed using the default timeout instead of failing. This reproduces the NPE before the fix.

## Documentation

No documentation changes required.

### DIFF
--- a/apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/jobs/TraceThreadsClosingJobTest.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/jobs/TraceThreadsClosingJobTest.java
@@ -23,7 +23,10 @@ import com.comet.opik.api.resources.utils.resources.WorkspaceResourceClient;
 import com.comet.opik.api.resources.utils.traces.TraceAssertions;
 import com.comet.opik.extensions.DropwizardAppExtensionProvider;
 import com.comet.opik.extensions.RegisterApp;
+import com.comet.opik.infrastructure.TraceThreadConfig;
 import com.comet.opik.podam.PodamFactoryUtils;
+import com.google.inject.Injector;
+import com.google.inject.Key;
 import com.redis.testcontainers.RedisContainer;
 import org.apache.commons.lang3.RandomStringUtils;
 import org.awaitility.Awaitility;
@@ -38,6 +41,7 @@ import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.lifecycle.Startables;
 import org.testcontainers.mysql.MySQLContainer;
 import reactor.core.publisher.Mono;
+import ru.vyarus.dropwizard.guice.module.yaml.bind.Config;
 import ru.vyarus.dropwizard.guice.test.ClientSupport;
 import ru.vyarus.dropwizard.guice.test.jupiter.ext.TestDropwizardAppExtension;
 import uk.co.jemos.podam.api.PodamFactory;
@@ -101,18 +105,25 @@ class TraceThreadsClosingJobTest {
 
     private final PodamFactory podamFactory = PodamFactoryUtils.newPodamFactory();
 
+    // Buffer added on top of a configured timeout when waiting for a thread to become inactive,
+    // so waits track the actual configuration instead of a literal coupled to config-test.yml.
+    private static final Duration CLOSURE_WAIT_BUFFER = Duration.ofMillis(500);
+
     private String baseURI;
     private ClientSupport client;
     private ProjectResourceClient projectResourceClient;
     private TraceResourceClient traceResourceClient;
     private WorkspaceResourceClient workspaceResourceClient;
+    private Duration defaultThreadTimeout;
 
     @BeforeAll
 
-    void setUpAll(ClientSupport client) {
+    void setUpAll(ClientSupport client, Injector injector) {
 
         this.baseURI = TestUtils.getBaseUrl(client);
         this.client = client;
+        this.defaultThreadTimeout = injector.getInstance(Key.get(TraceThreadConfig.class, Config.class))
+                .getTimeoutToMarkThreadAsInactive().toJavaDuration();
 
         ClientSupportUtils.config(client);
 
@@ -160,7 +171,8 @@ class TraceThreadsClosingJobTest {
             var expectedTraceThreadModel = createTraceThreadModel(threadId, projectId, expectedCreatedAt,
                     expectedLastUpdatedAt, DEFAULT_USER, TraceThreadStatus.INACTIVE, traces);
 
-            Mono.delay(Duration.ofSeconds(1)).block();
+            // Wait for the configured default timeout to elapse so the thread is eligible for closure
+            Mono.delay(defaultThreadTimeout.plus(CLOSURE_WAIT_BUFFER)).block();
 
             var expectedLastUpdateAt = Instant.now();
             TraceThread expectedUpdatedTraceThreadModel = expectedTraceThreadModel.toBuilder()
@@ -168,10 +180,11 @@ class TraceThreadsClosingJobTest {
                     .build();
 
             // Then
-            Awaitility.await().pollInterval(500, TimeUnit.MILLISECONDS).untilAsserted(() -> {
-                verifyClosedThreads(projectId, projectName, apiKey, workspaceName,
-                        List.of(expectedUpdatedTraceThreadModel));
-            });
+            Awaitility.await().atMost(30, TimeUnit.SECONDS).pollInterval(500, TimeUnit.MILLISECONDS)
+                    .untilAsserted(() -> {
+                        verifyClosedThreads(projectId, projectName, apiKey, workspaceName,
+                                List.of(expectedUpdatedTraceThreadModel));
+                    });
         }
 
         @Test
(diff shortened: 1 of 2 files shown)


apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/jobs/TraceThreadsClosingJobTest.java
```diff
@@ -330,6 +330,57 @@ void shouldCloseTraceThreadsForProjectWithCustomTimeout() {
...
+            // When: Creating trace threads
+            traceResourceClient.batchCreateTraces(traces, apiKey, workspaceName);
+
+            Instant expectedLastUpdatedAt = getExpectedLastUpdatedAt(traces);
+
+            var expectedTraceThreadModel = createTraceThreadModel(threadId, projectId, expectedCreatedAt,
+                    expectedLastUpdatedAt, DEFAULT_USER, TraceThreadStatus.INACTIVE, traces);
+
+            // Allow the default timeout (1s in config-test.yml) to elapse
+            Mono.delay(Duration.ofSeconds(1)).block();
```
JetoPistola: ❓ **question / nit** | Testing

This `1s` delay is implicitly coupled to `traceThreadConfig.timeoutToMarkThreadAsInactive` in `config-test.yml`. If that default ever changes, this test (and the two existing ones at the project/custom-timeout cases) silently drift — passing for the wrong reason or flaking. Since the test class doesn't currently inject the config, deriving the wait from the actual configured timeout (e.g. injecting `traceThreadConfig` and using `timeout.plus(buffer)`) would make all three robust to config changes.

Not blocking — and arguably a separate test-hygiene change across all three call sites rather than something to bolt onto this fix — but flagging since this PR adds a third copy of the literal.

JetoPistola:
👋 **Review summary**

**What looks good**
- Crisp root-cause analysis: correctly identifies that `switchIfEmpty` only fires on empty completion, not on a present-but-null field, so `map()` emitted null and Reactor rejected it.
- The fix (`mapNotNull` + `defaultIfEmpty`) is the canonical Reactor idiom and cleanly covers both "no config" and "config with null timeout."
- The inline comment in the production code explains *why* the operator change matters without restating it.
- Regression test reproduces the exact NPE path (config with only `truncationOnTables` set) and follows the existing `TraceThreadsClosingJobTest` conventions.

**Overall**
Tightly scoped one-line behavioral fix with a matching regression test — approach and scope are spot on. Left one non-blocking nit about the `1s` delay being coupled to the config default (pre-existing across three tests; better solved as separate test hygiene).

**Inline comments**: 1 question/nit — nothing blocking.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/jobs/TraceThreadsClosingJobTest.java
```diff
@@ -330,6 +330,57 @@ void shouldCloseTraceThreadsForProjectWithCustomTimeout() {
...
+
+            var expectedLastUpdateAt = Instant.now();
+            TraceThread expectedUpdatedTraceThreadModel = expectedTraceThreadModel.toBuilder()
+                    .lastUpdatedAt(expectedLastUpdateAt)
+                    .build();
+
+            // Then: the thread is closed using the default timeout instead of failing with an NPE
+            Awaitility.await().pollInterval(500, TimeUnit.MILLISECONDS).untilAsserted(() -> {
```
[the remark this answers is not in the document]
thiagohora: Done in 5932db3 — added `.atMost(30, TimeUnit.SECONDS)` so the test fails fast instead of hanging if the closing job never runs (matches the backend test convention).

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/jobs/TraceThreadsClosingJobTest.java
```diff
@@ -330,6 +330,58 @@ void shouldCloseTraceThreadsForProjectWithCustomTimeout() {
...
+            // Given
+            var workspaceName = RandomStringUtils.secure().nextAlphanumeric(10);
+            var workspaceId = UUID.randomUUID().toString();
+            var apiKey = UUID.randomUUID().toString();
+
+            mockTargetWorkspace(apiKey, workspaceName, workspaceId);
+
+            String projectName = RandomStringUtils.secure().nextAlphanumeric(10);
+            var projectId = projectResourceClient.createProject(projectName, apiKey, workspaceName);
+            var threadId = UUID.randomUUID().toString();
```
andrescrz: Nit: in prod this is UUID v7, I'd use that to be more realistic.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/jobs/TraceThreadsClosingJobTest.java
thiagohora: Went ahead and did it (e85b300) — it was cheap. Injected the `@Config`-bound `TraceThreadConfig` and derived the waits from it: the default-timeout cases now wait `defaultThreadTimeout.plus(500ms)` and the custom-timeout case derives from its local `customTimeout`, so none of the three are coupled to the `config-test.yml` literal anymore. Also bounded the remaining `await()` chains with `.atMost(30s)`. All 4 tests green.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/jobs/TraceThreadsClosingJobTest.java
```diff
@@ -160,18 +171,20 @@ void shouldCloseTraceThreadsForProject() {
             var expectedTraceThreadModel = createTraceThreadModel(threadId, projectId, expectedCreatedAt,
                     expectedLastUpdatedAt, DEFAULT_USER, TraceThreadStatus.INACTIVE, traces);
 
-            Mono.delay(Duration.ofSeconds(1)).block();
+            // Wait for the configured default timeout to elapse so the thread is eligible for closure
+            Mono.delay(defaultThreadTimeout.plus(CLOSURE_WAIT_BUFFER)).block();
```
andrescrz: Minor: better use `TestUtils.waitForMillis` for these delays.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/jobs/TraceThreadsClosingJobTest.java
```diff
@@ -160,18 +171,20 @@ void shouldCloseTraceThreadsForProject() {
...
             TraceThread expectedUpdatedTraceThreadModel = expectedTraceThreadModel.toBuilder()
                     .lastUpdatedAt(expectedLastUpdateAt)
                     .build();
 
             // Then
-            Awaitility.await().pollInterval(500, TimeUnit.MILLISECONDS).untilAsserted(() -> {
-                verifyClosedThreads(projectId, projectName, apiKey, workspaceName,
-                        List.of(expectedUpdatedTraceThreadModel));
-            });
+            Awaitility.await().atMost(30, TimeUnit.SECONDS).pollInterval(500, TimeUnit.MILLISECONDS)
```
andrescrz: Minor: be aggressive with max wait times, so tests fail fast and we don't wait unnecessary.

andrescrz (dismissed):
LGTM, just minor stuff on the tests

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/jobs/TraceThreadsClosingJobTest.java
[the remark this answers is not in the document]
andrescrz: @thiagohora careful this comment was misleading. Awaitility has defaults, it never hangs forever (10 seconds in this case):
- https://github.com/awaitility/awaitility/wiki/Usage#defaults

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/jobs/TraceThreadsClosingJobTest.java
thiagohora: Done in 7350f7a4be — thread IDs now use a UUID v7 generator (`TimeBasedEpochGenerator`) across all four tests, matching production.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/jobs/TraceThreadsClosingJobTest.java
thiagohora: Done in 7350f7a4be — replaced the `Mono.delay(...).block()` sleeps with `TestUtils.waitForMillis(...)`.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/jobs/TraceThreadsClosingJobTest.java
thiagohora: Done in 7350f7a4be — every `await()` is now bounded with `.atMost(AWAIT_TIMEOUT_SECONDS, TimeUnit.SECONDS)` (10s), including the previously unbounded reopen-test awaits, so they fail fast.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/jobs/TraceThreadsClosingJobTest.java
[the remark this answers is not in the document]
thiagohora: You're right, thanks — Awaitility defaults to 10s, so it never hangs forever; my earlier wording was off. I dropped the inflated 30s and set an explicit, uniform `.atMost(10s)` via a constant for fail-fast behavior.

andrescrz (approved):

PR #7148, opened 2026-06-17
[OPIK-6971] Treat insufficient_quota (429) as non-retryable for OpenAI providers

thiagohora:
## Details

When an LLM provider key returns HTTP 429 `insufficient_quota` (out of credits), the request keeps failing the same way. LangChain4j maps any 429 to a retryable `RateLimitException` — both in the `OpenAiChatModel` internal retry (`withRetryMappingExceptions`, default `maxRetries=2`) and in the outer `ChatCompletionService` retry policy — so an exhausted key was retried repeatedly per request (inner × outer attempts), generating a sustained error storm against a key that cannot recover.

This adds a `QuotaAwareHttpClient` decorator that inspects 429 responses and, when the body is `insufficient_quota`, rethrows a `NonRetriableException` so both retry layers stop immediately.
- The original response body is carried as the exception **message**, so the existing provider-error mapping still surfaces it as HTTP 402.
- The `HttpException` is **not** kept as the cause, so LangChain4j's `ExceptionMapper.findRoot()` cannot unwrap and re-map the 429 back to a retryable error.
- `rate_limit_exceeded` (429) and all other statuses pass through unchanged and remain retryable.
- Wired into the free-model and OpenAI completions chat models and the raw OpenAI client.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- OPIK-6971

## AI-WATERMARK

AI-WATERMARK: yes

- If yes:
  - Tools: Claude Code
  - Scope: Investigation, root-cause analysis, implementation, and tests
  - Human verification: Author reviewed the diff and ran the build/tests locally

## Testing

- `mvn -o -Dtest=QuotaAwareHttpClientTest test` → `Tests run: 4, Failures: 0, Errors: 0` (BUILD SUCCESS)
- `mvn -o compile` → clean
- Scenarios validated:
  - 429 `insufficient_quota` → rethrown as `NonRetriableException`, body preserved as message, no cause
  - 429 `rate_limit_exceeded` → original `HttpException` passed through (stays retryable)
  - non-429 4xx → original `HttpException` passed through
  - success → delegate response returned unchanged
- Not run: full backend integration suite (change is unit-scoped; verified the exception-handling path against the LangChain4j 1.16 sources).

## Documentation

N/A — internal resilience fix, no public API change.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/OpenAIClientGenerator.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/OpenAIClientGenerator.java
@@ -43,6 +43,9 @@ public class OpenAIClientGenerator implements LlmProviderClientGenerator<OpenAiC
     public OpenAiClient newOpenAiClient(@NonNull LlmProviderClientApiConfig config) {
         var openAiClientBuilder = OpenAiClient.builder()
                 .baseUrl(DEFAULT_OPENAI_URL)
+                // Treat insufficient_quota (429, out-of-credits) as non-retryable, so the outer retry
+                // policy in ChatCompletionService does not keep hitting an exhausted key.
+                .httpClientBuilder(QuotaAwareHttpClient.builder())
                 .logRequests(llmProviderClientConfig.getLogRequests())
                 .logResponses(llmProviderClientConfig.getLogResponses());
 
@@ -109,6 +112,9 @@ public class OpenAIClientGenerator implements LlmProviderClientGenerator<OpenAiC
         var builder = OpenAiChatModel.builder()
                 .modelName(modelParameters.name())
                 .apiKey(config.apiKey())
+                // Treat insufficient_quota (429, out-of-credits) as non-retryable so neither the model's
+                // internal retry nor the outer retry policy keeps hitting an exhausted key.
+                .httpClientBuilder(QuotaAwareHttpClient.builder())
                 .logRequests(true)
                 .logResponses(true);
 
--- a/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/QuotaAwareHttpClient.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/QuotaAwareHttpClient.java
@@ -1,0 +1,118 @@
+package com.comet.opik.infrastructure.llm.openai;
+
+import dev.langchain4j.exception.HttpException;
+import dev.langchain4j.exception.NonRetriableException;
+import dev.langchain4j.http.client.HttpClient;
+import dev.langchain4j.http.client.HttpClientBuilder;
+import dev.langchain4j.http.client.HttpClientBuilderLoader;
+import dev.langchain4j.http.client.HttpRequest;
+import dev.langchain4j.http.client.SuccessfulHttpResponse;
+import dev.langchain4j.http.client.sse.ServerSentEventListener;
+import dev.langchain4j.http.client.sse.ServerSentEventParser;
+import lombok.NonNull;
+
+import java.time.Duration;
+
+/**
+ * {@link HttpClient} decorator that marks OpenAI {@code insufficient_quota} (HTTP 429, out-of-credits)
+ * as non-retryable.
+ * <p>
+ * OpenAI returns both transient rate-limits and out-of-credits as HTTP 429. LangChain4j maps any 429 to
+ * a retryable {@link dev.langchain4j.exception.RateLimitException} — both in the model's internal
+ * {@code RetryUtils.withRetryMappingExceptions} and in Opik's outer retry policy — so a key that is
+ * permanently out of quota gets retried repeatedly per request (inner attempts x outer attempts),
+ * generating an error storm against a key that will keep failing. This decorator inspects 429 responses
+ * and, when the body identifies {@code insufficient_quota}, rethrows a {@link NonRetriableException} so
+ * both retry layers stop immediately.
+ * <p>
+ * Two details are required for correct downstream handling:
+ * <ul>
+ *   <li>the original response body is carried as the exception <b>message</b>, so the provider-error
+ *       mapping ({@code getLlmProviderError} -> {@link OpenAiCompatStatusCodes}) still parses it and
+ *       surfaces HTTP 402; and</li>
+ *   <li>the {@link HttpException} is <b>not</b> retained as the cause — otherwise LangChain4j's
+ *       {@code ExceptionMapper.findRoot()} would unwrap to it and re-map the 429 back to a retryable
+ *       {@code RateLimitException}, defeating the guard.</li>
+ * </ul>
+ * Rate-limit 429s and every other status are passed through unchanged and remain retryable.
+ */
+public class QuotaAwareHttpClient implements HttpClient {
+
+    private static final int TOO_MANY_REQUESTS = 429;
+    private static final String INSUFFICIENT_QUOTA = "insufficient_quota";
+
+    private final HttpClient delegate;
+
+    public QuotaAwareHttpClient(@NonNull HttpClient delegate) {
+        this.delegate = delegate;
+    }
+
+    @Override
+    public SuccessfulHttpResponse execute(HttpRequest request) {
+        try {
+            return delegate.execute(request);
+        } catch (HttpException httpException) {
+            if (isInsufficientQuota(httpException)) {
+                // Body as message (so it still maps to 402 downstream); no HttpException cause (so the
+                // inner ExceptionMapper cannot unwrap and re-map the 429 to a retryable error).
+                throw new NonRetriableException(httpException.getMessage());
+            }
+            throw httpException;
+        }
+    }
+
+    @Override
+    public void execute(HttpRequest request, ServerSentEventParser parser, ServerSentEventListener listener) {
+        // Streaming responses are not wrapped in a retry policy, so there is nothing to short-circuit.
+        delegate.execute(request, parser, listener);
+    }
+
+    private static boolean isInsufficientQuota(HttpException httpException) {
+        return httpException.statusCode() == TOO_MANY_REQUESTS
+                && httpException.getMessage() != null
+                && httpException.getMessage().contains(INSUFFICIENT_QUOTA);
+    }
+
+    /**
+     * Wraps the default (classpath) {@link HttpClientBuilder} so every client it builds is quota-aware.
+     */
+    public static HttpClientBuilder builder() {
+        return new Builder(HttpClientBuilderLoader.loadHttpClientBuilder());
+    }
+
+    private static final class Builder implements HttpClientBuilder {
+
+        private final HttpClientBuilder delegate;
+
+        private Builder(@NonNull HttpClientBuilder delegate) {
+            this.delegate = delegate;
+        }
+
+        @Override
+        public Duration connectTimeout() {
+            return delegate.connectTimeout();
+        }
+
+        @Override
+        public HttpClientBuilder connectTimeout(Duration connectTimeout) {
+            delegate.connectTimeout(connectTimeout);
+            return this;
+        }
+
+        @Override
+        public Duration readTimeout() {
+            return delegate.readTimeout();
+        }
+
+        @Override
+        public HttpClientBuilder readTimeout(Duration readTimeout) {
+            delegate.readTimeout(readTimeout);
+            return this;
+        }
+
+        @Override
+        public HttpClient build() {
+            return new QuotaAwareHttpClient(delegate.build());
+        }
+    }
+}
--- a/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/QuotaAwareHttpClient.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/QuotaAwareHttpClient.java
@@ -1,5 +1,6 @@
 package com.comet.opik.infrastructure.llm.openai;
 
+import com.comet.opik.infrastructure.llm.DelegatingHttpClientBuilder;
 import dev.langchain4j.exception.HttpException;
 import dev.langchain4j.exception.NonRetriableException;
 import dev.langchain4j.http.client.HttpClient;
@@ -9,9 +10,10 @@ import dev.langchain4j.http.client.HttpRequest;
 import dev.langchain4j.http.client.SuccessfulHttpResponse;
 import dev.langchain4j.http.client.sse.ServerSentEventListener;
 import dev.langchain4j.http.client.sse.ServerSentEventParser;
+import jakarta.ws.rs.core.Response;
 import lombok.NonNull;
-
-import java.time.Duration;
+import lombok.RequiredArgsConstructor;
+import org.apache.commons.lang3.StringUtils;
 
 /**
  * {@link HttpClient} decorator that marks OpenAI {@code insufficient_quota} (HTTP 429, out-of-credits)
@@ -36,16 +38,12 @@ import java.time.Duration;
  * </ul>
  * Rate-limit 429s and every other status are passed through unchanged and remain retryable.
  */
+@RequiredArgsConstructor
 public class QuotaAwareHttpClient implements HttpClient {
 
-    private static final int TOO_MANY_REQUESTS = 429;
     private static final String INSUFFICIENT_QUOTA = "insufficient_quota";
 
-    private final HttpClient delegate;
-
-    public QuotaAwareHttpClient(@NonNull HttpClient delegate) {
-        this.delegate = delegate;
-    }
+    private final @NonNull HttpClient delegate;
 
     @Override
     public SuccessfulHttpResponse execute(HttpRequest request) {
@@ -68,9 +66,8 @@ public class QuotaAwareHttpClient implements HttpClient {
     }
 
     private static boolean isInsufficientQuota(HttpException httpException) {
-        return httpException.statusCode() == TOO_MANY_REQUESTS
-                && httpException.getMessage() != null
-                && httpException.getMessage().contains(INSUFFICIENT_QUOTA);
+        return httpException.statusCode() == Response.Status.TOO_MANY_REQUESTS.getStatusCode()
+                && StringUtils.containsIgnoreCase(httpException.getMessage(), INSUFFICIENT_QUOTA);
     }
 
     /**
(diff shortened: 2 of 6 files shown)


apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/OpenAIClientGenerator.java
```diff
@@ -43,6 +43,9 @@ public enum ApiPipelineMode {
     public OpenAiClient newOpenAiClient(@NonNull LlmProviderClientApiConfig config) {
         var openAiClientBuilder = OpenAiClient.builder()
                 .baseUrl(DEFAULT_OPENAI_URL)
+                // Treat insufficient_quota (429, out-of-credits) as non-retryable, so the outer retry
```
JetoPistola, line 46: ❓ **question** | Scope

OpenRouter is automatically covered since it reuses this generator. But the Custom LLM path (`CustomLlmClientGenerator`) builds its own `OpenAiClient`/`OpikOpenAiChatModel` with `InterceptingHttpClientBuilder` and isn't wired to the quota guard — so an out-of-credits OpenAI-compatible custom provider would still retry-storm. Was Custom LLM intentionally left out of scope here? If in scope later, `QuotaAwareHttpClient` could compose around `InterceptingHttpClientBuilder` (both are `HttpClient` decorators).

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/openai/QuotaAwareHttpClientTest.java
```diff
@@ -0,0 +1,80 @@
...
+        var client = new QuotaAwareHttpClient(delegate);
+        when(delegate.execute(request)).thenThrow(new HttpException(429, QUOTA_BODY));
+
+        // Both LangChain4j's internal retry and Opik's outer retry skip NonRetriableException. The body must be
+        // the message (so it still maps to 402 downstream) and there must be no HttpException cause (otherwise
+        // ExceptionMapper.findRoot() would unwrap it and re-map the 429 to a retryable RateLimitException).
+        var thrown = catchThrowable(() -> client.execute(request));
+
+        assertThat(thrown)
+                .isInstanceOf(NonRetriableException.class)
```
JetoPistola, line 56: 🧹 **nit** | Style

This test uses `catchThrowable(...)` then asserts on the result, while the other three use `assertThatThrownBy(...)`. The `catchThrowable` form is justified here because you chain `.hasMessage(...).hasNoCause()` — but `assertThatThrownBy(() -> client.execute(request)).isInstanceOf(...).hasMessage(QUOTA_BODY).hasNoCause()` reads identically and keeps all four tests consistent. Purely cosmetic.

JetoPistola:
👋 **Review summary**

**What looks good**
- Tight, surgical fix for a real production failure mode (retry storm against out-of-credits keys). The decorator approach is the right seam — it sits below both retry layers without touching either.
- Excellent class-level Javadoc: it documents *why* the body must be the message and *why* the `HttpException` must not be retained as the cause. I verified both against langchain4j 1.16.1 — the body-as-message keeps the `insufficient_quota`→402 mapping working, and dropping the cause is genuinely load-bearing (`ExceptionMapper.findRoot()` / `findThrowableInChain` would otherwise unwrap and re-map to a retryable error).
- The `HttpClientBuilder` delegate is fully transparent: I confirmed `DefaultOpenAiClient` pushes connect/read timeouts onto the supplied builder via its setters, so wrapping it preserves the configured timeouts — no regression.
- Clean, focused test coverage of all four branches (quota, rate-limit, other 4xx, success), each with a clear intent-documenting `@DisplayName`.

**Overall**
Solid, well-reasoned change. The mechanism holds up under scrutiny against the actual langchain4j retry internals, and OpenRouter is covered for free by wiring at the generator level. Nothing blocking — just one scope question (Custom LLM path) and one cosmetic test nit below.

**Inline comments**: 1 question, 1 nit — nothing blocking.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/OpenAIClientGenerator.java
```diff
@@ -43,6 +43,9 @@ public enum ApiPipelineMode {
...
+                // policy in ChatCompletionService does not keep hitting an exhausted key.
+                .httpClientBuilder(QuotaAwareHttpClient.builder())
                 .logRequests(llmProviderClientConfig.getLogRequests())
                 .logResponses(llmProviderClientConfig.getLogResponses());
```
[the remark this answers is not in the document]
thiagohora: Good catch — verified. `newResponsesApiChatModel()` builds `OpenAiOfficialResponsesChatModel` (the official OpenAI SDK model), which uses its own transport and has no `httpClientBuilder` seam, so it does **not** reuse the `OpenAiClient` I wired here. It is intentionally out of scope for this PR, which targets the langchain4j OpenAI-completions + free-model paths that drove the incident (OPIK-6971). The Responses API path retries via the official SDK and needs a separate mechanism (SDK-level retry/`maxRetries` config), so I will track it as a follow-up rather than fold it in here.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/OpenAIClientGenerator.java
thiagohora: Correct — Custom LLM was intentionally left out of scope here. This PR targets the paths behind the OPIK-6971 incident (free-model + OpenAI completions, with OpenRouter covered for free since it reuses this generator). `CustomLlmClientGenerator` builds its client/model with its own `newHttpClientBuilder(config)` (the intercepting builder), so it is not wired to the guard. Your composition suggestion is exactly the right shape for the follow-up: `QuotaAwareHttpClient` and `InterceptingHttpClientBuilder` are both `HttpClient` decorators, so we can wrap one around the other. Tracking as a follow-up.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/openai/QuotaAwareHttpClientTest.java
thiagohora: Done — switched to `assertThatThrownBy(...)` so all four tests use the same style (d22f648b46).

CometActions:

### Access Information
- **URL:** https://pr-7148.dev.comet.com
- **Namespace:** pr-7148
- **Version:** 2.0.69-7148-merge-2292
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7148%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/QuotaAwareHttpClient.java
```diff
@@ -0,0 +1,118 @@
...
+ *       surfaces HTTP 402; and</li>
+ *   <li>the {@link HttpException} is <b>not</b> retained as the cause — otherwise LangChain4j's
+ *       {@code ExceptionMapper.findRoot()} would unwrap to it and re-map the 429 back to a retryable
+ *       {@code RateLimitException}, defeating the guard.</li>
+ * </ul>
+ * Rate-limit 429s and every other status are passed through unchanged and remain retryable.
+ */
+public class QuotaAwareHttpClient implements HttpClient {
+
+    private static final int TOO_MANY_REQUESTS = 429;
```
andrescrz: Nit: reuse pre-existing constants (there are multiple already as dependency: Apache HttpClient, Jakarta Response etc.)

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/QuotaAwareHttpClient.java
```diff
@@ -0,0 +1,118 @@
...
+    private static final String INSUFFICIENT_QUOTA = "insufficient_quota";
+
+    private final HttpClient delegate;
+
+    public QuotaAwareHttpClient(@NonNull HttpClient delegate) {
+        this.delegate = delegate;
+    }
```
andrescrz: Nit: lombok for boiler plate constructor.

Applies everywhere in the PR.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/QuotaAwareHttpClient.java
```diff
@@ -0,0 +1,118 @@
...
+
+    @Override
+    public SuccessfulHttpResponse execute(HttpRequest request) {
+        try {
+            return delegate.execute(request);
+        } catch (HttpException httpException) {
+            if (isInsufficientQuota(httpException)) {
+                // Body as message (so it still maps to 402 downstream); no HttpException cause (so the
+                // inner ExceptionMapper cannot unwrap and re-map the 429 to a retryable error).
+                throw new NonRetriableException(httpException.getMessage());
```
andrescrz, line 56: Double check the original behaviour of the client here, you might want to pass the original exception as cause to the non-retriable one.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/QuotaAwareHttpClient.java
```diff
@@ -0,0 +1,118 @@
...
+    @Override
+    public void execute(HttpRequest request, ServerSentEventParser parser, ServerSentEventListener listener) {
+        // Streaming responses are not wrapped in a retry policy, so there is nothing to short-circuit.
+        delegate.execute(request, parser, listener);
+    }
+
+    private static boolean isInsufficientQuota(HttpException httpException) {
+        return httpException.statusCode() == TOO_MANY_REQUESTS
+                && httpException.getMessage() != null
+                && httpException.getMessage().contains(INSUFFICIENT_QUOTA);
```
andrescrz: Let's endure this to avoid case sensitiveness. It should return true regardless of the case.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/QuotaAwareHttpClient.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /**
+     * Wraps the default (classpath) {@link HttpClientBuilder} so every client it builds is quota-aware.
+     */
+    public static HttpClientBuilder builder() {
+        return new Builder(HttpClientBuilderLoader.loadHttpClientBuilder());
+    }
+
+    private static final class Builder implements HttpClientBuilder {
```
andrescrz: We have a duplicated `InterceptingHttpClientBuilder` already. Probably we want to make it more neutral and reuse, instead of copying and pasting. The only difference is the client you return on build, but the delegation logic is the same.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/openai/QuotaAwareHttpClientTest.java
```diff
@@ -0,0 +1,77 @@
...
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+@DisplayName("Quota Aware HttpClient")
+class QuotaAwareHttpClientTest {
+
+    private static final String QUOTA_BODY = "{\"error\":{\"message\":\"You exceeded your current quota, please check "
+            + "your plan and billing details.\",\"type\":\"insufficient_quota\",\"code\":\"insufficient_quota\"}}";
+    private static final String RATE_LIMIT_BODY = "{\"error\":{\"message\":\"Rate limit reached\",\"type\":\"requests\","
+            + "\"code\":\"rate_limit_exceeded\"}}";
```
andrescrz: Minor: better use java text blocks with triple quotes `"""` for these constants.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/openai/QuotaAwareHttpClientTest.java
```diff
@@ -0,0 +1,77 @@
...
+    @Mock
+    private HttpClient delegate;
+
+    @Mock
+    private HttpRequest request;
+
+    @Test
+    @DisplayName("when 429 insufficient_quota, then rethrow as NonRetriableException carrying the body and no cause")
+    void execute__when429InsufficientQuota__thenNonRetriableWithBodyAndNoHttpCause() {
+        var client = new QuotaAwareHttpClient(delegate);
```
andrescrz: Nit: you don't need to instantiate this all the time. You can use `@InjectMocks`.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/openai/QuotaAwareHttpClientTest.java
```diff
@@ -0,0 +1,77 @@
...
+        var original = new HttpException(429, RATE_LIMIT_BODY);
+        when(delegate.execute(request)).thenThrow(original);
+
+        assertThatThrownBy(() -> client.execute(request)).isSameAs(original);
+    }
+
+    @Test
+    @DisplayName("when a non-429 client error, then rethrow the original HttpException")
+    void execute__whenOtherClientError__thenRethrowHttpException() {
```
andrescrz: Minor: same test case as before, better parameterise.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/openai/QuotaAwareHttpClientTest.java
andrescrz, line 1: Please add test cases if you address feedback (like case sensitive etc.) so you make sure we have full coverage on the class.

andrescrz (commented):
Let's double check the important comments before moving forward.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/QuotaAwareHttpClient.java
thiagohora: Done in ef7f1f4 — dropped the `429` literal and inlined `Response.Status.TOO_MANY_REQUESTS.getStatusCode()`.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/QuotaAwareHttpClient.java
thiagohora: Done in ef7f1f4 — `@RequiredArgsConstructor` on `QuotaAwareHttpClient`. The two builder subclasses still need explicit constructors since they call `super(delegate)` (Lombok can't generate a constructor with a `super(...)` call), but the shared `DelegatingHttpClientBuilder` base uses `@RequiredArgsConstructor`.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/QuotaAwareHttpClient.java
thiagohora: Intentionally not passing the cause — verified against langchain4j 1.16.1. The internal retry runs the call through `ExceptionMapper.withExceptionMapper`, and `ExceptionMapper.findRoot()` walks to the deepest cause: if the `HttpException` were kept as the cause it would be found and re-mapped via `mapHttpStatusCode(429)` back to a retryable `RateLimitException`, defeating the guard. Downstream behaviour is preserved because `getLlmProviderError` reads the JSON body from the exception **message** (which we keep verbatim), not from the cause — so it still maps to 402. This is documented in the class Javadoc, and the `hasNoCause()` assertion locks it in.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/QuotaAwareHttpClient.java
thiagohora: Done in ef7f1f4 — now uses `StringUtils.containsIgnoreCase(...)`, and the test is parameterized over `insufficient_quota` / `INSUFFICIENT_QUOTA` / `Insufficient_Quota`.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/openai/QuotaAwareHttpClient.java
thiagohora: Done in ef7f1f4 — extracted `DelegatingHttpClientBuilder` (timeout forwarding + `final build()` that calls `wrap(delegate.build())`). Both `QuotaAwareHttpClient.Builder` and `InterceptingHttpClientBuilder` now extend it and only implement `wrap(...)`. No more duplicated delegation logic.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/openai/QuotaAwareHttpClientTest.java
thiagohora: Done in ef7f1f4 — constants now use Java text blocks (`"""`).

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/openai/QuotaAwareHttpClientTest.java
thiagohora: Done in ef7f1f4 — switched to `@InjectMocks` for the client under test.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/openai/QuotaAwareHttpClientTest.java
thiagohora: Done in ef7f1f4 — the two pass-through cases (429 rate-limit + non-429 4xx) are now a single `@MethodSource` parameterized test, and the quota case is parameterized over casings.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/openai/QuotaAwareHttpClientTest.java
thiagohora: Done in ef7f1f4 — added casing coverage (`insufficient_quota` / `INSUFFICIENT_QUOTA` / `Insufficient_Quota`) via the parameterized quota test. Full run: 6/6 green.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/openai/QuotaAwareHttpClientTest.java
```diff
@@ -0,0 +1,85 @@
...
+    private HttpClient delegate;
+
+    @Mock
+    private HttpRequest request;
+
+    @InjectMocks
+    private QuotaAwareHttpClient client;
+
+    @ParameterizedTest(name = "429 insufficient_quota with code casing ''{0}'' is non-retryable")
+    @ValueSource(strings = {"insufficient_quota", "INSUFFICIENT_QUOTA", "Insufficient_Quota"})
```
andrescrz, line 44: Minor: null message is a valid case to test (it should work). You can add `@NullAndEmptySource` to this method, it will work.

andrescrz (approved):
LGTM

PR #7154, opened 2026-06-18
[OPIK-6982] Add reaper for orphaned Redis stream consumers

thiagohora:
## Details

Each backend process registers a unique `consumer-<group>-<UUID>` per process that Redis only removes on graceful shutdown (`XGROUP DELCONSUMER`). Non-graceful exits (OOMKill / SIGKILL / crash / node eviction) leak the consumer permanently, so stream consumer groups accumulate tens of thousands of orphaned consumers across deploys and crashes. This adds a periodic reaper that cleans them up so the count stays bounded.

- `StreamConsumerReaper`: for each stream, lists groups (`XINFO GROUPS`) and removes consumers that are **both** idle beyond a threshold **and** have no pending entries. The `pending == 0` guard leaves un-acked work for `XAUTOCLAIM` to reclaim, so it is never destroyed.
- `StreamConsumerReaperJob`: Quartz job that discovers the streams to clean from the registered `BaseRedisSubscriber` Guice bindings (no keyspace scan, no hand-maintained list — mirrors `EventListenerRegistrar`), guarded by a distributed lock with hold-until-expiry so only one instance reaps per cycle.
- `StreamConsumerReaperConfig`: `enabled` (true) / `jobInterval` (1h) / `idleThreshold` (1d) / `lockDuration` (10m), all env-overridable.
- `BaseRedisSubscriber` exposes `getStreamName()` for discovery.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- OPIK-6982

## AI-WATERMARK

AI-WATERMARK: yes

- If yes:
  - Tools: Claude Code
  - Model(s): Claude Opus 4.8
  - Scope: Full implementation (reaper service, job, config, wiring) and tests, authored under human review.
  - Human verification: Design reviewed and steered interactively; tests run locally (see Testing).

## Testing

- `mvn test -Dtest=StreamConsumerReaperTest` — 5 tests, real Redis testcontainer: reaps idle+empty consumers; leaves consumers with pending entries; respects the idle threshold; mixed orphan/active/pending; no-op for non-existent streams.
- `mvn test -Dtest=StreamConsumerReaperJobTest` — full application Guice context: asserts the job discovers all registered subscriber streams (the six online-scoring streams plus webhook / experiment-item / experiment-aggregates / trace-thread-closing / dataset-export).
- Both pass locally (local process mode).

## Documentation

N/A

### DIFF
--- a/apps/opik-backend/config.yml
+++ b/apps/opik-backend/config.yml
@@ -1232,6 +1232,23 @@ localRunner:
   # Description: Maximum payload size in bytes for bridge command args and results
   bridgeMaxPayloadBytes: ${OPIK_RUNNER_BRIDGE_MAX_PAYLOAD_BYTES:-1048576}
 
+# Orphaned Redis stream-consumer reaper (OPIK-6982). Non-graceful pod exits (OOMKill/SIGKILL/crash) leak
+# stream consumers that Redis never removes; this job deletes consumers idle beyond the threshold with no
+# pending entries. One instance runs per cycle (distributed lock with hold-until-expiry).
+streamConsumerReaper:
+  # Default: true
+  # Description: Whether the orphaned stream-consumer reaper job is enabled
+  enabled: ${OPIK_STREAM_CONSUMER_REAPER_ENABLED:-true}
+  # Default: 1h
+  # Description: Interval between reaper job executions
+  jobInterval: ${OPIK_STREAM_CONSUMER_REAPER_JOB_INTERVAL:-1h}
+  # Default: 1d
+  # Description: Consumers idle beyond this (and with no pending entries) are removed
+  idleThreshold: ${OPIK_STREAM_CONSUMER_REAPER_IDLE_THRESHOLD:-1d}
+  # Default: 10m
+  # Description: How long the reaper holds its distributed lock (until expiry) to prevent concurrent fleet-wide runs
+  lockDuration: ${OPIK_STREAM_CONSUMER_REAPER_LOCK_DURATION:-10m}
+
 # Trace Thread configuration
 traceThreadConfig:
   # Default: true
--- a/apps/opik-backend/config.yml
+++ b/apps/opik-backend/config.yml
@@ -1242,9 +1242,12 @@ streamConsumerReaper:
   # Default: 1h
   # Description: Interval between reaper job executions
   jobInterval: ${OPIK_STREAM_CONSUMER_REAPER_JOB_INTERVAL:-1h}
-  # Default: 1d
+  # Default: 5m
+  # Description: Delay before the first reaper run after startup, to avoid reaping during warm-up
+  startupDelay: ${OPIK_STREAM_CONSUMER_REAPER_STARTUP_DELAY:-5m}
+  # Default: 3d
   # Description: Consumers idle beyond this (and with no pending entries) are removed
-  idleThreshold: ${OPIK_STREAM_CONSUMER_REAPER_IDLE_THRESHOLD:-1d}
+  idleThreshold: ${OPIK_STREAM_CONSUMER_REAPER_IDLE_THRESHOLD:-3d}
   # Default: 10m
   # Description: How long the reaper holds its distributed lock (until expiry) to prevent concurrent fleet-wide runs
   lockDuration: ${OPIK_STREAM_CONSUMER_REAPER_LOCK_DURATION:-10m}
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/StreamConsumerReaperJob.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/StreamConsumerReaperJob.java
@@ -1,6 +1,7 @@
 package com.comet.opik.api.resources.v1.jobs;
 
 import com.comet.opik.api.resources.v1.events.BaseRedisSubscriber;
+import com.comet.opik.infrastructure.GuiceBindings;
 import com.comet.opik.infrastructure.StreamConsumerReaperConfig;
 import com.comet.opik.infrastructure.lock.LockService;
 import com.comet.opik.infrastructure.redis.StreamConsumerReaper;
@@ -103,11 +104,9 @@ public class StreamConsumerReaperJob extends Job implements InterruptableJob {
      * Package-private for testing against the real application Guice context.
      */
     List<String> discoverStreamNames() {
-        List<String> streamNames = injector.getAllBindings().keySet().stream()
-                .map(key -> key.getTypeLiteral().getRawType())
+        List<String> streamNames = GuiceBindings.boundRawTypes(injector)
                 .filter(BaseRedisSubscriber.class::isAssignableFrom)
                 .filter(type -> !Modifier.isAbstract(type.getModifiers()))
-                .distinct()
                 .map(type -> (BaseRedisSubscriber) injector.getInstance(type))
                 .map(BaseRedisSubscriber::getStreamName)
                 .distinct()
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/StreamConsumerReaperJob.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/StreamConsumerReaperJob.java
@@ -10,11 +10,14 @@ import io.dropwizard.jobs.Job;
 import jakarta.inject.Inject;
 import jakarta.inject.Singleton;
 import lombok.NonNull;
+import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.quartz.DisallowConcurrentExecution;
 import org.quartz.InterruptableJob;
 import org.quartz.JobExecutionContext;
+import reactor.core.Disposable;
 import reactor.core.publisher.Mono;
+import reactor.core.scheduler.Schedulers;
 
 import java.lang.reflect.Modifier;
 import java.time.Duration;
@@ -35,29 +38,23 @@ import static com.comet.opik.infrastructure.lock.LockService.Lock;
 @Slf4j
 @Singleton
 @DisallowConcurrentExecution
+@RequiredArgsConstructor(onConstructor_ = @Inject)
 public class StreamConsumerReaperJob extends Job implements InterruptableJob {
 
     private static final Lock JOB_LOCK = new Lock("stream_consumer_reaper:lock");
 
-    private final Injector injector;
-    private final StreamConsumerReaper reaper;
-    private final LockService lockService;
-    private final StreamConsumerReaperConfig config;
+    private final @NonNull Injector injector;
+    private final @NonNull StreamConsumerReaper reaper;
+    private final @NonNull LockService lockService;
+    private final @NonNull StreamConsumerReaperConfig config;
 
     private final AtomicBoolean interrupted = new AtomicBoolean(false);
-    // Subscribers are fixed after startup, so the discovered stream names are computed once and cached.
+
+    /** Subscribers are fixed after startup, so the discovered stream names are computed once and cached. */
     private final AtomicReference<List<String>> streamNamesCache = new AtomicReference<>();
 
-    @Inject
-    public StreamConsumerReaperJob(@NonNull Injector injector,
-            @NonNull StreamConsumerReaper reaper,
-            @NonNull LockService lockService,
-            @NonNull StreamConsumerReaperConfig config) {
-        this.injector = injector;
-        this.reaper = reaper;
-        this.lockService = lockService;
-        this.config = config;
-    }
+    /** Tracks the in-flight reactive pass so {@link #interrupt()} can dispose it (cancels the actual chain). */
+    private final AtomicReference<Disposable> currentExecution = new AtomicReference<>();
 
     @Override
     public void doJob(JobExecutionContext context) {
@@ -74,28 +71,43 @@ public class StreamConsumerReaperJob extends Job implements InterruptableJob {
             return;
         }
 
-        lockService.bestEffortLock(
+        var subscription = lockService.bestEffortLock(
                 JOB_LOCK,
-                reaper.reap(streamNames, config.getIdleThreshold().toJavaDuration())
-                        .doOnSuccess(count -> {
-                            if (count > 0) {
-                                log.info(
-                                        "Stream consumer reaper removed '{}' orphaned consumer(s) across '{}' stream(s)",
-                                        count, streamNames.size());
-                            } else {
-                                log.debug("Stream consumer reaper found no orphaned consumers across '{}' stream(s)",
-                                        streamNames.size());
-                            }
-                        }),
-                Mono.fromRunnable(() -> log.info(
+                Mono.defer(() -> {
+                    if (interrupted.get()) {
+                        log.info("Stream consumer reaper interrupted before processing, skipping");
+                        return Mono.empty();
+                    }
+                    return reaper.reap(streamNames, config.idleThreshold().toJavaDuration())
+                            .doOnSuccess(count -> {
+                                if (count > 0) {
+                                    log.info(
+                                            "Stream consumer reaper removed orphaned consumer(s), removed='{}', streams='{}'",
+                                            count, streamNames.size());
+                                } else {
+                                    log.debug("Stream consumer reaper found no orphaned consumers, streams='{}'",
+                                            streamNames.size());
+                                }
+                            });
+                }),
+                Mono.fromRunnable(() -> log.debug(
                         "Could not acquire lock for stream consumer reaper, another instance is running")),
-                config.getLockDuration().toJavaDuration(),
+                config.lockDuration().toJavaDuration(),
                 Duration.ZERO,
                 true) // holdUntilExpiry: prevent redundant runs across instances until the next cycle
-                .subscribe(
-                        __ -> {
-                        },
-                        error -> log.error("Stream consumer reaper failed", error));
+                // Resume on errors so the recurring job stays alive.
+                .onErrorResume(throwable -> {
+                    if (interrupted.get()) {
+                        log.warn("Stream consumer reaper interrupted", throwable);
+                    } else {
+                        log.error("Stream consumer reaper failed", throwable);
+                    }
+                    return Mono.empty();
+                })
+                .doFinally(signal -> currentExecution.set(null))
+                .subscribeOn(Schedulers.boundedElastic())
+                .subscribe();
+        currentExecution.set(subscription);
     }
 
     /**
@@ -111,7 +123,8 @@ public class StreamConsumerReaperJob extends Job implements InterruptableJob {
                 .map(BaseRedisSubscriber::getStreamName)
                 .distinct()
                 .toList();
-        log.info("Discovered '{}' stream(s) to reap orphaned consumers from: '{}'", streamNames.size(), streamNames);
+        log.info("Discovered stream(s) to reap orphaned consumers from, count='{}', streams='{}'",
+                streamNames.size(), streamNames);
         return streamNames;
     }
 
@@ -119,5 +132,10 @@ public class StreamConsumerReaperJob extends Job implements InterruptableJob {
     public void interrupt() {
         interrupted.set(true);
         log.info("Stream consumer reaper job interrupted");
+        var execution = currentExecution.get();
+        if (execution != null && !execution.isDisposed()) {
+            execution.dispose();
+            log.info("Stream consumer reaper job interrupted successfully");
+        }
     }
 }
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/StreamConsumerReaperJob.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/StreamConsumerReaperJob.java
@@ -1,0 +1,124 @@
+package com.comet.opik.api.resources.v1.jobs;
+
+import com.comet.opik.api.resources.v1.events.BaseRedisSubscriber;
+import com.comet.opik.infrastructure.StreamConsumerReaperConfig;
+import com.comet.opik.infrastructure.lock.LockService;
+import com.comet.opik.infrastructure.redis.StreamConsumerReaper;
+import com.google.inject.Injector;
+import io.dropwizard.jobs.Job;
+import jakarta.inject.Inject;
+import jakarta.inject.Singleton;
+import lombok.NonNull;
+import lombok.extern.slf4j.Slf4j;
+import org.quartz.DisallowConcurrentExecution;
+import org.quartz.InterruptableJob;
+import org.quartz.JobExecutionContext;
+import reactor.core.publisher.Mono;
+
+import java.lang.reflect.Modifier;
+import java.time.Duration;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+
+import static com.comet.opik.infrastructure.lock.LockService.Lock;
+
+/**
+ * Periodically removes orphaned Redis stream consumers leaked by non-graceful pod exits (OPIK-6982).
+ * <p>
+ * The streams to clean are discovered from the registered {@link BaseRedisSubscriber} beans (no keyspace scan, no
+ * hand-maintained list) — mirroring how {@code EventListenerRegistrar} discovers event listeners. The actual
+ * reaping is delegated to {@link StreamConsumerReaper}. A distributed lock with hold-until-expiry ensures only one
+ * instance reaps per cycle, so the fleet does not redundantly re-scan the same groups.
+ */
+@Slf4j
+@Singleton
+@DisallowConcurrentExecution
+public class StreamConsumerReaperJob extends Job implements InterruptableJob {
+
+    private static final Lock JOB_LOCK = new Lock("stream_consumer_reaper:lock");
+
+    private final Injector injector;
+    private final StreamConsumerReaper reaper;
+    private final LockService lockService;
+    private final StreamConsumerReaperConfig config;
+
+    private final AtomicBoolean interrupted = new AtomicBoolean(false);
+    // Subscribers are fixed after startup, so the discovered stream names are computed once and cached.
+    private final AtomicReference<List<String>> streamNamesCache = new AtomicReference<>();
+
+    @Inject
+    public StreamConsumerReaperJob(@NonNull Injector injector,
+            @NonNull StreamConsumerReaper reaper,
+            @NonNull LockService lockService,
+            @NonNull StreamConsumerReaperConfig config) {
+        this.injector = injector;
+        this.reaper = reaper;
+        this.lockService = lockService;
+        this.config = config;
+    }
+
+    @Override
+    public void doJob(JobExecutionContext context) {
+        if (interrupted.get()) {
+            log.info("Stream consumer reaper job interrupted before execution, skipping");
+            return;
+        }
+
+        List<String> streamNames = streamNamesCache.updateAndGet(
+                cached -> cached != null ? cached : discoverStreamNames());
+
+        if (streamNames.isEmpty()) {
+            log.debug("No stream subscribers discovered, skipping stream consumer reaper");
+            return;
+        }
+
+        lockService.bestEffortLock(
+                JOB_LOCK,
+                reaper.reap(streamNames, config.getIdleThreshold().toJavaDuration())
+                        .doOnSuccess(count -> {
+                            if (count > 0) {
+                                log.info(
+                                        "Stream consumer reaper removed '{}' orphaned consumer(s) across '{}' stream(s)",
+                                        count, streamNames.size());
+                            } else {
+                                log.debug("Stream consumer reaper found no orphaned consumers across '{}' stream(s)",
+                                        streamNames.size());
+                            }
+                        }),
+                Mono.fromRunnable(() -> log.info(
+                        "Could not acquire lock for stream consumer reaper, another instance is running")),
+                config.getLockDuration().toJavaDuration(),
+                Duration.ZERO,
+                true) // holdUntilExpiry: prevent redundant runs across instances until the next cycle
+                .subscribe(
+                        __ -> {
+                        },
+                        error -> log.error("Stream consumer reaper failed", error));
+    }
+
+    /**
+     * Discovers the stream names of every registered {@link BaseRedisSubscriber} by scanning the Guice bindings,
+     * so new subscribers are covered automatically without registration. Mirrors {@code EventListenerRegistrar}.
+     * Package-private for testing against the real application Guice context.
+     */
+    List<String> discoverStreamNames() {
+        List<String> streamNames = injector.getAllBindings().keySet().stream()
+                .map(key -> key.getTypeLiteral().getRawType())
+                .filter(BaseRedisSubscriber.class::isAssignableFrom)
+                .filter(type -> !Modifier.isAbstract(type.getModifiers()))
+                .distinct()
+                .map(type -> (BaseRedisSubscriber) injector.getInstance(type))
+                .map(BaseRedisSubscriber::getStreamName)
+                .distinct()
+                .toList();
+        log.info("Discovered '{}' stream(s) to reap orphaned consumers from: '{}'", streamNames.size(), streamNames);
+        return streamNames;
+    }
+
+    @Override
+    public void interrupt() {
+        interrupted.set(true);
+        log.info("Stream consumer reaper job interrupted");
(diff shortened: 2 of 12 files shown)


apps/opik-backend/src/main/java/com/comet/opik/infrastructure/StreamConsumerReaperConfig.java
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Lock TTL, held until expiry, that suppresses concurrent fleet-wide runs. Also bounds how long a single reap
+     * pass may run. A reap pass is light (per-stream {@code XINFO GROUPS}/{@code XINFO CONSUMERS} + a few
+     * {@code XGROUP DELCONSUMER}), so this is well below {@link #getJobInterval()}; an occasional extra pass within
+     * a cycle is harmless since {@code XGROUP DELCONSUMER} is idempotent.
+     */
+    @Valid @NotNull @JsonProperty
+    @MinDuration(value = 1, unit = TimeUnit.SECONDS)
+    private Duration lockDuration = Duration.minutes(10);
```
JetoPistola: 🧹 **nit** | Documentation

The javadoc says this lease "bounds how long a single reap pass may run", but per `LockService.bestEffortLock`'s contract the action is **not** subject to `actionTimeout` — the lease/TTL can expire while the reap is still running (held-until-expiry only suppresses *other* instances). The pass isn't actually cancelled at the TTL. Consider rewording to "bounds how long other instances stay suppressed" rather than implying the pass is time-bounded. Harmless in practice (a reap pass is light and `XGROUP DELCONSUMER` is idempotent).

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/StreamConsumerReaperConfig.java
```diff
@@ -0,0 +1,53 @@
...
+ * shutdown ({@code XGROUP DELCONSUMER}). Non-graceful exits (OOMKill, SIGKILL, crash, node eviction) leak the
+ * consumer permanently, so the online-scoring/stream consumer groups accumulate orphans across deploys and crashes.
+ * The reaper periodically removes consumers that are both idle beyond {@link #getIdleThreshold()} and have no
+ * pending entries (so it never drops un-acked work that {@code XAUTOCLAIM} would otherwise reclaim).
+ */
+@Data
+public class StreamConsumerReaperConfig {
+
+    @Valid @JsonProperty
+    private boolean enabled = true;
```
JetoPistola: 🧹 **nit** | Style

`@Valid` on a primitive `boolean` is a no-op — there's no nested object to cascade bean-validation into. Other configs in the repo do the same, so it's consistent, but it can be dropped. Not blocking.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/redis/StreamConsumerReaper.java
```diff
@@ -0,0 +1,123 @@
...
+                .reduce(0L, Long::sum)
+                .onErrorResume(throwable -> {
+                    reaperErrors.add(1, Attributes.of(STREAM_KEY, streamName, GROUP_KEY, group));
+                    log.warn("Failed to list consumers for group '{}' on stream '{}'", group, streamName, throwable);
+                    return Mono.just(0L);
+                });
+    }
+
+    private static boolean isNoSuchKey(Throwable throwable) {
+        return Objects.toString(throwable.getMessage(), "").contains(NO_SUCH_KEY);
```
JetoPistola, line 138: ❓ **question** | Robustness

`isNoSuchKey` classifies the "stream doesn't exist" case by substring-matching `"no such key"` in the exception message. Two concerns: (1) it's coupled to Redis/Redisson error text and could silently break on a version bump, and (2) any *other* error whose message happens to contain that phrase would be misclassified as "stream not created yet" and logged at INFO instead of surfacing as a reaper error. Acceptable for a best-effort cleanup job, but worth a short comment that it's a fragile string match — or is there a typed exception you could catch instead?

JetoPistola (dismissed):
👋 **Review summary**

**What looks good**
- The core safety property is sound: deleting only consumers that are **both** idle-beyond-threshold **and** `pending == 0` means a crashed consumer's PEL is reclaimed by `XAUTOCLAIM` into a live consumer *before* the now-empty orphan is removed — so `XGROUP DELCONSUMER` never destroys un-acked work. Verified the subscribers do auto-claim.
- Stream discovery via the registered `BaseRedisSubscriber` Guice bindings (no keyspace scan, no hand-maintained list) is the right call, and extracting `GuiceBindings.boundRawTypes` to share the traversal with `EventListenerRegistrar` is a clean refactor — it also quietly fixes a latent bug there (adds `Objects::nonNull` filtering and logs full stack traces instead of just `getMessage()`).
- The reactive distributed-lock usage mirrors the existing `LocalRunnerReaperJob` idiom exactly (`holdUntilExpiry=true`, zero wait, lease = `lockDuration`), so it composes correctly across the fleet.
- Good real-Redis testcontainer coverage: orphan / pending / within-threshold / mixed / non-existent stream, plus a full-Guice-context test asserting discovery finds all 11 subscriber streams.
- Follow-up commit addressed all review feedback cleanly — the `no such key` substring match is now documented as mirroring the existing `NOGROUP`/`BUSYGROUP` handling in `BaseRedisSubscriber` (verified) with a bounded downside, the `lockDuration` doc no longer overstates that it bounds the pass, and the dead `@Valid` was dropped.

**Overall**
Solid, well-scoped fix. Conservative approach (idle + empty-PEL guard), env-overridable config with sane defaults, disabled in tests. No blocking concerns and all prior nits are resolved.

**Inline comments**: 3 from the prior review (2 nits, 1 question) — all addressed in the latest commit, nothing outstanding.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/StreamConsumerReaperJob.java
```diff
@@ -0,0 +1,123 @@
...
+    private static final Lock JOB_LOCK = new Lock("stream_consumer_reaper:lock");
+
+    private final Injector injector;
+    private final StreamConsumerReaper reaper;
+    private final LockService lockService;
+    private final StreamConsumerReaperConfig config;
+
+    private final AtomicBoolean interrupted = new AtomicBoolean(false);
+    // Subscribers are fixed after startup, so the discovered stream names are computed once and cached.
+    private final AtomicReference<List<String>> streamNamesCache = new AtomicReference<>();
```
andrescrz: Nit: all these comments on top of a field/method etc. better as Javadocs.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/StreamConsumerReaperJob.java
```diff
@@ -0,0 +1,123 @@
...
+    @Inject
+    public StreamConsumerReaperJob(@NonNull Injector injector,
+            @NonNull StreamConsumerReaper reaper,
+            @NonNull LockService lockService,
+            @NonNull StreamConsumerReaperConfig config) {
+        this.injector = injector;
+        this.reaper = reaper;
+        this.lockService = lockService;
+        this.config = config;
+    }
```
andrescrz: Minor: boiler-plate constructor, better define with Lombok.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/StreamConsumerReaperJob.java
```diff
@@ -0,0 +1,123 @@
...
+                cached -> cached != null ? cached : discoverStreamNames());
+
+        if (streamNames.isEmpty()) {
+            log.debug("No stream subscribers discovered, skipping stream consumer reaper");
+            return;
+        }
+
+        lockService.bestEffortLock(
+                JOB_LOCK,
+                reaper.reap(streamNames, config.getIdleThreshold().toJavaDuration())
```
andrescrz: You probably need an interruption check on the action, before calling `reap`.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/StreamConsumerReaperJob.java
```diff
@@ -0,0 +1,123 @@
...
+                        .doOnSuccess(count -> {
+                            if (count > 0) {
+                                log.info(
+                                        "Stream consumer reaper removed '{}' orphaned consumer(s) across '{}' stream(s)",
```
andrescrz: Nit: for all logs, print values at the end of the sentence, so it's easier to grep them during production debugging sessions.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/StreamConsumerReaperJob.java
```diff
@@ -0,0 +1,123 @@
...
+                                        count, streamNames.size());
+                            } else {
+                                log.debug("Stream consumer reaper found no orphaned consumers across '{}' stream(s)",
+                                        streamNames.size());
+                            }
+                        }),
+                Mono.fromRunnable(() -> log.info(
+                        "Could not acquire lock for stream consumer reaper, another instance is running")),
```
andrescrz: Minor: other jobs have this as info, better start defining these as debug as this is noisy and and expected condition. I also was asked about these logs by a confused engineer.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/StreamConsumerReaperJob.java
```diff
@@ -0,0 +1,123 @@
...
+
+        List<String> streamNames = streamNamesCache.updateAndGet(
+                cached -> cached != null ? cached : discoverStreamNames());
+
+        if (streamNames.isEmpty()) {
+            log.debug("No stream subscribers discovered, skipping stream consumer reaper");
+            return;
+        }
+
+        lockService.bestEffortLock(
```
andrescrz: You likely want to subscribe this in a bounded elastic scheduler, as there's IO.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/StreamConsumerReaperJob.java
```diff
@@ -0,0 +1,123 @@
...
+     * Package-private for testing against the real application Guice context.
+     */
+    List<String> discoverStreamNames() {
+        List<String> streamNames = GuiceBindings.boundRawTypes(injector)
+                .filter(BaseRedisSubscriber.class::isAssignableFrom)
+                .filter(type -> !Modifier.isAbstract(type.getModifiers()))
+                .map(type -> (BaseRedisSubscriber) injector.getInstance(type))
+                .map(BaseRedisSubscriber::getStreamName)
+                .distinct()
+                .toList();
```
andrescrz, line 126: Nit: this is a set, you can collect an ordered flavour for iterations.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/StreamConsumerReaperJob.java
```diff
@@ -0,0 +1,123 @@
...
+        log.info("Discovered '{}' stream(s) to reap orphaned consumers from: '{}'", streamNames.size(), streamNames);
+        return streamNames;
+    }
+
+    @Override
+    public void interrupt() {
```
andrescrz, line 133: This won't interrupt the actual reactive chain of the job. See other jobs such as `ExperimentProjectMigrationJob` for an example of a proper interruption.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/bi/OpikGuiceyLifecycleEventListener.java
```diff
@@ -148,6 +151,19 @@ private void setLocalRunnerReaperJob() {
...
+        StreamConsumerReaperConfig reaperConfig = injector.get().getInstance(OpikConfiguration.class)
+                .getStreamConsumerReaper();
+
+        if (!reaperConfig.isEnabled()) {
+            log.info("Stream consumer reaper job is disabled, skipping job setup");
+            return;
+        }
+
+        scheduleRepeatingJob(StreamConsumerReaperJob.class,
+                reaperConfig.getJobInterval().toJavaDuration(), null);
```
andrescrz: You probably prefer to add a configurable start-up delay. No avoid reaping early after start-up. You'll reduce the chances of any false positive.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/redis/StreamConsumerReaper.java
```diff
@@ -0,0 +1,133 @@
...
+        this.redisson = redisson;
+        var meter = GlobalOpenTelemetry.getMeter("opik.redis");
+        this.reapedConsumers = meter
+                .counterBuilder("opik_redis_stream_reaped_consumers")
+                .setDescription("Orphaned (idle beyond threshold, no pending) stream consumers removed by the reaper")
+                .build();
+        this.reaperErrors = meter
+                .counterBuilder("opik_redis_stream_reaper_errors")
+                .setDescription("Errors while reaping orphaned stream consumers")
+                .build();
```
andrescrz: Let's add a result attribute to the other metric, with success/error etc. instead of having a duplicated counter.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/redis/StreamConsumerReaper.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    /**
+     * Reaps orphaned consumers across the given streams.
+     *
+     * @return the total number of consumers removed
+     */
+    public Mono<Long> reap(@NonNull Collection<String> streamNames, @NonNull Duration idleThreshold) {
```
andrescrz: Nit: maybe just handle null/empty collection case and finish with no action (0 returned).

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/GuiceBindings.java
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Helpers for traversing Guice bindings. Centralizes the {@code getAllBindings()} walk used to auto-discover beans
+ * (event listeners, stream subscribers, ...) so the traversal lives in one place and callers only add their own
+ * type/annotation filter.
+ */
+public final class GuiceBindings {
+
+    private GuiceBindings() {
+    }
```
andrescrz: Minor: use lombok to define utils class.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/GuiceBindings.java
```diff
@@ -0,0 +1,26 @@
...
+
+    /**
+     * Distinct raw types of every explicit binding in the injector. Callers apply their own filter (assignability,
+     * annotation, ...) and resolve instances as needed.
+     */
+    public static Stream<Class<?>> boundRawTypes(Injector injector) {
```
andrescrz: Public method, so validate as NonNull.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/StreamConsumerReaperConfig.java
```diff
@@ -0,0 +1,54 @@
...
+ * Configuration for the orphaned-consumer reaper (OPIK-6982).
+ * <p>
+ * Each backend process registers a unique {@code consumer-<group>-<UUID>} that Redis only removes on graceful
+ * shutdown ({@code XGROUP DELCONSUMER}). Non-graceful exits (OOMKill, SIGKILL, crash, node eviction) leak the
+ * consumer permanently, so the online-scoring/stream consumer groups accumulate orphans across deploys and crashes.
+ * The reaper periodically removes consumers that are both idle beyond {@link #getIdleThreshold()} and have no
+ * pending entries (so it never drops un-acked work that {@code XAUTOCLAIM} would otherwise reclaim).
+ */
+@Data
+public class StreamConsumerReaperConfig {
```
andrescrz: Minor: I'm starting to define these as records (with builder/toBuilder), feel free to follow this more modern pattern.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/StreamConsumerReaperConfig.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @JsonProperty
+    private boolean enabled = true;
```
andrescrz: Let's remove the java defaults here. All should be annotated as NotNull (except primitives types such as `boolean` as they can't be null), so they're always populated from the config yaml files (both prod and test). The idea is not maintain the default values only in one place.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/StreamConsumerReaperConfig.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * How often the reaper runs. A single instance runs per cycle (distributed lock with hold-until-expiry), so
+     * a low frequency keeps the {@code XINFO GROUPS}/{@code XINFO CONSUMERS} cost negligible.
+     */
+    @Valid @NotNull @JsonProperty
+    @MinDuration(value = 1, unit = TimeUnit.MINUTES)
```
andrescrz: Minor: Let's increase the min a bit (maybe 5 minutes) and define a sensible max here.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/StreamConsumerReaperConfig.java
```diff
@@ -0,0 +1,54 @@
...
+    private Duration jobInterval = Duration.hours(1);
+
+    /**
+     * Consumers idle longer than this (and with no pending entries) are deleted. Live consumers reset their idle
+     * time on every {@code XREADGROUP} (even empty long-poll returns), so their idle stays in the seconds range;
+     * the default is deliberately far above that to avoid reaping an active consumer.
+     */
+    @Valid @NotNull @JsonProperty
+    @MinDuration(value = 1, unit = TimeUnit.MINUTES)
```
andrescrz: Let's have a longer idle to allow config errors. What do you think about 1 hour?

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/StreamConsumerReaperConfig.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Lock TTL, held until expiry, that suppresses other instances from reaping until it elapses. It does NOT
+     * cancel an in-flight reap pass ({@code bestEffortLock} does not bound the action) — the pass runs to
+     * completion even if the lease expires first. Kept well below {@link #getJobInterval()}; if the lease expires
+     * mid-cycle and another instance runs an extra pass, that is harmless since {@code XGROUP DELCONSUMER} is
+     * idempotent and a pass is light (per-stream {@code XINFO GROUPS}/{@code XINFO CONSUMERS} + a few deletes).
+     */
+    @Valid @NotNull @JsonProperty
+    @MinDuration(value = 1, unit = TimeUnit.SECONDS)
```
andrescrz: Minor: same here, the min is a bit short in relation to the period of the job, let's increase, maybe a few minutes.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/redis/StreamConsumerReaperTest.java
```diff
@@ -0,0 +1,145 @@
...
+        // A consumer that registered but never had work: idle grows, pending stays 0 — the orphan case.
+        stream.createConsumer(GROUP, "orphan-consumer").block();
+
+        // Let it become idle beyond the (tiny) threshold used by this test.
+        waitForMillis(600);
+
+        var reaped = reaper.reap(List.of(streamName), Duration.ofMillis(300)).block();
+
+        assertThat(reaped).isEqualTo(1L);
+        assertThat(listConsumerNames()).doesNotContain("orphan-consumer");
```
andrescrz, line 82: Minor: should you assert it exists above so we ensure the test is more reliable?

apps/opik-backend/config.yml
```diff
@@ -1232,6 +1232,23 @@ localRunner:
...
+streamConsumerReaper:
+  # Default: true
+  # Description: Whether the orphaned stream-consumer reaper job is enabled
+  enabled: ${OPIK_STREAM_CONSUMER_REAPER_ENABLED:-true}
+  # Default: 1h
+  # Description: Interval between reaper job executions
+  jobInterval: ${OPIK_STREAM_CONSUMER_REAPER_JOB_INTERVAL:-1h}
+  # Default: 1d
+  # Description: Consumers idle beyond this (and with no pending entries) are removed
+  idleThreshold: ${OPIK_STREAM_CONSUMER_REAPER_IDLE_THRESHOLD:-1d}
```
andrescrz: I'd probably start with a less aggressive value. Maybe a 3 days. We can reduce once we acquire confidence.

andrescrz (changes requested):
The general idea of the job LGTM.

I left many comments about the implementation details before moving forward.

I'll discuss with you offline some doubts about the root cause of this for my understanding, no concerns for going ahead with the job.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/StreamConsumerReaperJob.java
thiagohora: Fixed in fbfd496 — converted the field/method inline comments (stream-name cache, interrupt tracking) to Javadoc.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/StreamConsumerReaperJob.java
thiagohora: Fixed in fbfd496 — replaced the boilerplate constructor with `@RequiredArgsConstructor(onConstructor_ = @Inject)` and `@NonNull` fields.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/StreamConsumerReaperJob.java
thiagohora: Fixed in fbfd496 — added an `interrupted` check inside a `Mono.defer` so the pass short-circuits before calling `reap` if interrupted.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/StreamConsumerReaperJob.java
thiagohora: Fixed in fbfd496 — reordered the log placeholders so values come at the end of the message (e.g. `removed='{}', streams='{}'`).

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/StreamConsumerReaperJob.java
thiagohora: Fixed in fbfd496 — moved the lock-not-acquired log to `debug` (expected, noisy condition), matching the quieter convention.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/StreamConsumerReaperJob.java
thiagohora: Fixed in fbfd496 — the reactive pass now runs on `Schedulers.boundedElastic()` via `subscribeOn`.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/StreamConsumerReaperJob.java
thiagohora: Fixed in fbfd496 — `discoverStreamNames()` already returns a distinct, ordered `List`; I also made the reaper's dedup order-preserving (`LinkedHashSet` instead of `Set.copyOf`) so iteration order is stable.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/StreamConsumerReaperJob.java
thiagohora: Fixed in fbfd496 — reworked interruption to track the subscription in an `AtomicReference<Disposable>` and `dispose()` it in `interrupt()`, mirroring `ExperimentProjectMigrationJob`.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/bi/OpikGuiceyLifecycleEventListener.java
thiagohora: Fixed in fbfd496 — added a configurable `startupDelay` (default 5m) wired into `scheduleRepeatingJob`, so the first pass is delayed after start-up to reduce false positives.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/redis/StreamConsumerReaper.java
thiagohora: Fixed in fbfd496 — merged the two counters into a single `opik_redis_stream_reaper_consumers` counter with a `result` attribute (`success`/`error`).

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/redis/StreamConsumerReaper.java
thiagohora: Fixed in fbfd496 — added an early return of `Mono.just(0L)` when the stream-name collection is null/empty.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/GuiceBindings.java
thiagohora: Fixed in fbfd496 — annotated the class with Lombok's `@UtilityClass`.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/GuiceBindings.java
thiagohora: Fixed in fbfd496 — added `@NonNull` to the `Injector` parameter of `boundRawTypes`.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/StreamConsumerReaperConfig.java
thiagohora: Fixed in fbfd496 — converted the config to a `record` with `@Builder(toBuilder = true)`.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/StreamConsumerReaperConfig.java
thiagohora: Fixed in fbfd496 — removed the Java default values; all fields are `@NotNull` and populated from config.yml and config-test.yml.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/StreamConsumerReaperConfig.java
thiagohora: Fixed in fbfd496 — raised `jobInterval` bounds to min 5m and added a max of 24h.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/StreamConsumerReaperConfig.java
thiagohora: Fixed in fbfd496 — raised the `idleThreshold` minimum to 1h (and added a 30d max) so a misconfiguration can't reap too aggressively.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/StreamConsumerReaperConfig.java
thiagohora: Fixed in fbfd496 — raised the `lockDuration` minimum to 1m (and added a 1h max).

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/redis/StreamConsumerReaperTest.java
thiagohora: Fixed in fbfd496 — added an assertion that the orphan consumer exists before the reap, so the test fails loudly if setup breaks.

apps/opik-backend/config.yml
thiagohora: Fixed in fbfd496 — changed the default `idleThreshold` from 1d to 3d for a more conservative initial rollout.

thiagohora:
> @andrescrz wrote in their review (CHANGES_REQUESTED):
>
> The general idea of the job LGTM.
> I left many comments about the implementation details before moving forward.
> I'll discuss with you offline some doubts about the root cause of this for my understanding, no concerns for going ahead with the job.

Fixed in fbfd496 — addressed all the inline implementation comments: proper reactive interruption (Disposable disposed in `interrupt()`, mirroring `ExperimentProjectMigrationJob`) on a `boundedElastic` scheduler with a pre-`reap` interrupt check; Lombok constructor + Javadocs; config converted to a record with tighter bounds, no Java defaults, and a new `startupDelay`; the two metrics merged into one counter with a `result` attribute; `@UtilityClass`/`@NonNull` on `GuiceBindings`; logs reordered with values at the end and the lock-skip lowered to debug; `idleThreshold` default raised to 3d. Happy to sync offline on the root cause whenever works for you.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/OpikConfiguration.java
```diff
@@ -163,6 +163,9 @@ public class OpikConfiguration extends JobConfiguration {
     @Valid @NotNull @JsonProperty
     private LocalRunnerConfig localRunner = new LocalRunnerConfig();
 
+    @Valid @NotNull @JsonProperty
+    private StreamConsumerReaperConfig streamConsumerReaper = StreamConsumerReaperConfig.builder().build();
+
```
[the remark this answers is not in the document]
thiagohora: The `streamConsumerReaper` field already carries `@Valid @NotNull @JsonProperty` (OpikConfiguration.java:166, same as `localRunner`), so Bean Validation does cascade into the nested `Duration` constraints at startup. The real gap was elsewhere and is fixed in aebf97e: `StreamConsumerReaperJob` injected the config by type and—now that it's a `record` with no no-arg/@Inject constructor—Guice could no longer build it (and previously built a throwaway default, ignoring yaml). It now binds the validated parsed instance via `@Config("streamConsumerReaper")`, matching `ExperimentProjectMigrationJob`.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/StreamConsumerReaperJob.java
```diff
@@ -0,0 +1,142 @@
...
+@DisallowConcurrentExecution
+@RequiredArgsConstructor(onConstructor_ = @Inject)
+public class StreamConsumerReaperJob extends Job implements InterruptableJob {
+
+    private static final Lock JOB_LOCK = new Lock("stream_consumer_reaper:lock");
+
+    private final @NonNull Injector injector;
+    private final @NonNull StreamConsumerReaper reaper;
+    private final @NonNull LockService lockService;
+    private final @NonNull @Config("streamConsumerReaper") StreamConsumerReaperConfig config;
```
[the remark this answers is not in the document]
thiagohora: Skipping — keeping `config` for consistency with `ExperimentProjectMigrationJob` (the canonical Job + `@Config` example in the codebase), which names its field `config` too. The class has a single config dependency so the name is unambiguous, and `config.idleThreshold()`/`config.lockDuration()` read cleanly at the call sites.

andrescrz (approved):
Thanks for addressing the feedback. LGTM.

PR #7161, opened 2026-06-18
[OPIK-7006] Recover from stale chunk preload errors after deploy

aadereiko:
## Details

Users intermittently hit a full-page "Something went wrong! Failed to fetch dynamically imported module: .../assets/PlaygroundPage-XXXX.js" error in production. The app is code-split, so each route is a content-hash-named chunk; a tab opened before a deploy still references old hashes, and navigating to a not-yet-loaded route requests a chunk that was removed on deploy (404 → import rejects → error boundary).

This adds a global `vite:preloadError` handler that reloads the page to pull a fresh `index.html` with current asset hashes, transparently recovering instead of showing the error screen.
- A time-based `sessionStorage` guard prevents a reload loop if a chunk is genuinely unreachable (offline/outage): a preload error recurring within 10s of a reload is allowed to surface.
- `sessionStorage` access is wrapped in try/catch so the handler still recovers when storage is blocked (sandboxed iframe / strict privacy mode).

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- OPIK-7006

## AI-WATERMARK

AI-WATERMARK: yes

- If yes:
  - Tools: Claude Code
  - Model(s): Claude Opus 4.8
  - Scope: Full implementation (handler, wiring, unit tests) and PR description
  - Human verification: Author reviewed the diff, confirmed tests/lint/tsc pass locally

## Testing

- `npx vitest run src/lib/preload-error.test.ts` — 4 tests pass (reload on error, loop guard within window, reload again after window, recovery when sessionStorage throws).
- `npx eslint` and `tsc --noEmit` clean on changed files (the one pre-existing tsc error in the repo is unrelated to this change).
- One global handler covers all lazy routes, so no per-route changes were needed.

## Documentation

No documentation changes required — internal resilience fix with no user-facing API or behavior to document.

### DIFF
--- a/apps/opik-frontend/src/lib/preload-error.ts
+++ b/apps/opik-frontend/src/lib/preload-error.ts
@@ -4,6 +4,26 @@ const RELOAD_GUARD_KEY = "opik-preload-error-reloaded-at";
 // within this window, stop reloading and let the error surface instead.
 const RELOAD_GUARD_WINDOW = 10 * 1000;
 
+// `sessionStorage` can throw (e.g. SecurityError in a sandboxed iframe or when
+// storage is blocked in strict privacy mode). Recovery must not depend on it,
+// so reads/writes are guarded: a failed read allows the reload, a failed write
+// just loses the loop guard for that case.
+const readLastReloadAt = (): number => {
+  try {
+    return Number(window.sessionStorage.getItem(RELOAD_GUARD_KEY) ?? 0);
+  } catch {
+    return 0;
+  }
+};
+
+const writeLastReloadAt = (value: number): void => {
+  try {
+    window.sessionStorage.setItem(RELOAD_GUARD_KEY, String(value));
+  } catch {
+    // Storage unavailable — reload still proceeds without the time-based guard.
+  }
+};
+
 /**
  * Handles Vite's `vite:preloadError` event, which fires when a lazily-imported
  * chunk fails to load. The most common cause in production is a new deploy:
@@ -21,11 +41,8 @@ const RELOAD_GUARD_WINDOW = 10 * 1000;
 export const setupPreloadErrorHandler = () => {
   window.addEventListener("vite:preloadError", (event) => {
     const now = Date.now();
-    const lastReloadAt = Number(
-      window.sessionStorage.getItem(RELOAD_GUARD_KEY) ?? 0,
-    );
 
-    if (now - lastReloadAt < RELOAD_GUARD_WINDOW) {
+    if (now - readLastReloadAt() < RELOAD_GUARD_WINDOW) {
       // We already reloaded very recently and still hit a preload error — stop
       // reloading and let the error propagate to the error boundary.
       return;
@@ -34,7 +51,7 @@ export const setupPreloadErrorHandler = () => {
     // Prevent Vite from throwing the error so the page can recover quietly.
     event.preventDefault();
 
-    window.sessionStorage.setItem(RELOAD_GUARD_KEY, String(now));
+    writeLastReloadAt(now);
     window.location.reload();
   });
 };
--- a/apps/opik-frontend/src/lib/preload-error.ts
+++ b/apps/opik-frontend/src/lib/preload-error.ts
@@ -1,0 +1,40 @@
+const RELOAD_GUARD_KEY = "opik-preload-error-reloaded-at";
+
+// Avoid an infinite reload loop: if reloading didn't resolve the failed import
+// within this window, stop reloading and let the error surface instead.
+const RELOAD_GUARD_WINDOW = 10 * 1000;
+
+/**
+ * Handles Vite's `vite:preloadError` event, which fires when a lazily-imported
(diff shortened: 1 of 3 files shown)


apps/opik-frontend/src/lib/preload-error.ts
```diff
@@ -0,0 +1,57 @@
...
+      // reloading and let the error propagate to the error boundary.
+      return;
+    }
+
+    // Prevent Vite from throwing the error so the page can recover quietly.
+    event.preventDefault();
+
+    writeLastReloadAt(now);
+    window.location.reload();
+  });
```
JetoPistola, line 56: 💡 **suggestion** | Observability

The common path here (a stale-chunk error that *recovers* via reload) is now completely silent — and the rare suppressed path falls through to the generic "Something went wrong" boundary. That leaves no telemetry on how often deploy-churn triggers this recovery. Consider a lightweight Sentry breadcrumb or `Sentry.captureMessage("preload-error-recovered", "info")` before the reload so the team can confirm the 10s window is tuned right and watch recovery frequency over time. Sentry is already initialized in `index.tsx`.

Not blocking.

apps/opik-frontend/src/lib/preload-error.ts
```diff
@@ -0,0 +1,57 @@
+const RELOAD_GUARD_KEY = "opik-preload-error-reloaded-at";
+
+// Avoid an infinite reload loop: if reloading didn't resolve the failed import
```
JetoPistola, line 3: 💡 **suggestion** | Logic

The guard timestamp is written *before* `window.location.reload()`, so this 10s window includes the reload latency itself — fetching the fresh `index.html` plus the new chunk. On a slow connection a legitimate recovery could eat a meaningful slice of the budget, shrinking the effective post-reload window. 10s is probably fine, but worth a one-line comment noting the window spans reload latency (or a small bump) so the intent is explicit.

Not blocking.

apps/opik-frontend/src/lib/preload-error.ts
```diff
@@ -0,0 +1,57 @@
...
+// within this window, stop reloading and let the error surface instead.
+const RELOAD_GUARD_WINDOW = 10 * 1000;
+
+// `sessionStorage` can throw (e.g. SecurityError in a sandboxed iframe or when
+// storage is blocked in strict privacy mode). Recovery must not depend on it,
+// so reads/writes are guarded: a failed read allows the reload, a failed write
+// just loses the loop guard for that case.
+const readLastReloadAt = (): number => {
+  try {
```
JetoPistola, line 12: 🧹 **nit** | Style

`getItem` returns `string | null`, and the `?? 0` only guards the `null` case. A corrupted non-numeric stored value would yield `NaN`, making `now - NaN < WINDOW` evaluate to `false` — which correctly falls through to a reload, so behavior is safe. Just flagging that the `NaN` branch is implicit rather than handled. Fine as-is.

apps/opik-frontend/src/lib/preload-error.ts
```diff
@@ -0,0 +1,57 @@
...
+ * and transparently recovers, so the user never sees the generic error screen.
+ *
+ * A short time-based guard prevents a reload loop in the rare case where the
+ * chunk is genuinely unreachable for another reason (offline, server outage).
+ */
+export const setupPreloadErrorHandler = () => {
+  window.addEventListener("vite:preloadError", (event) => {
+    const now = Date.now();
+
+    if (now - readLastReloadAt() < RELOAD_GUARD_WINDOW) {
```
JetoPistola, line 45: ❓ **question** | Architecture

The handler is wired globally in `index.tsx`, so it covers both v1 and v2 routes — which looks intentional and correct. Confirming the actual error screen users hit in OPIK-7006 was the `SentryErrorBoundary` fallback (the only "Something went wrong" boundary I found in both `v1/` and `v2/`), and that no lazy route sits behind a different boundary this wouldn't help. From my read it's covered; just confirming.

JetoPistola:
👋 **Review summary**

**What looks good**
- Clean, well-scoped fix — one global handler in `index.tsx` covers every lazy route, with zero per-route changes needed.
- The `sessionStorage` defense (try/catch on both read and write, with the read-failure → allow-reload fallback) is genuinely thoughtful — covers sandboxed iframes and strict privacy mode without making recovery depend on storage.
- Excellent comments: the *why* behind the loop guard and the storage guarding is documented exactly where it matters, without restating the code.
- Solid test coverage — all four meaningful branches (reload, loop-guard suppression, post-window re-reload, storage-throws recovery) are covered, with deterministic fake timers.

**Overall**
The approach is correct and matches Vite's documented `vite:preloadError` contract — I verified against `handlePreloadError` in the bundled Vite 5.4.19 that `preventDefault()` suppresses the re-throw and the synchronous `reload()` fires before React can render a fallback. Sensible, low-risk resilience improvement.

**Inline comments**: 2 suggestions, 1 nit, 1 question — nothing blocking.

PR #7158, opened 2026-06-18
[OPIK-6968] [BE] perf: filter trace_threads join in getMinimalThreadInfoByIds

thiagohora:
## Details
The trace-thread close path (`/v1/private/traces/threads/close`) calls `getMinimalThreadInfoByIds`, whose `LEFT JOIN trace_threads` had no filter on the joined side. Every call therefore built its join hash table from the **entire** `trace_threads` table, independent of how few `thread_id`s were requested. This pushes the existing `workspace_id` / `project_id` / `thread_id` predicate into the joined subquery so it reads only the matching rows.

- Output is unchanged: the filter columns are exactly the join keys, so filtering the joined side cannot drop or alter any match.
- Only `getMinimalThreadInfoByIds` is touched; the inner `traces` aggregation is unchanged.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-6968

## AI-WATERMARK

AI-WATERMARK: yes

- If yes:
  - Tools: Claude Code
  - Model(s): Claude Opus 4.8
  - Scope: Query analysis against a read-only ClickHouse replica, the query change in this PR, and this description.
  - Human verification: Output equivalence and per-call cost validated by the author; reviewer to confirm.

## Testing

- **Output equivalence (read-only ClickHouse):** ran the current and proposed query forms on production-representative parameters (a 60-thread batch). Identical row count and an order-independent fingerprint (`groupBitXor(cityHash64(...))` over every output column, including `thread_model_id` and `status`) matched exactly.
- **Per-call cost (same parameters):** ~20.4M rows / ~1.1s / ~3.7 GiB → ~108K rows / ~16ms / <1 MiB. The win is removing the full-table hash build; it is independent of the requested thread count.
- **Not run:** the full `mvn` backend integration suite (e.g. `TracesResourceTest` thread close). This is a SQL-only change with verified-identical output; recommend CI to exercise the existing thread-close ITs.

## Documentation
N/A — internal query optimization, no behavior change.

PR #7178, opened 2026-06-19
[OPIK-7035] [BE] Thread query optimizations: listing page-pushdown + close/feedback thread-id resolution

thiagohora:
## Details

Three output-equivalent optimizations to thread query paths, all reducing how much of a project's `traces`/`trace_threads` is scanned. Motivated by prod profiling (started from the `/threads/close` latency that #7158 partially fixed): each query was reading hundreds of thousands to millions of ClickHouse rows (or a GiB+ of wide columns) to return a page/handful of threads.

- **Threads listing** (`ThreadDAO`, `GET /v1/private/traces/threads`): aggregated every thread in the project from the wide `traces`+`spans` tables, then applied `LIMIT/OFFSET` at the end. Now resolves the page first in one narrow filter+order scan (`page_thread_ids`, limit pushed early), then enriches only that page — wide `input`/`output` and spans read for ~page-size threads. Mutually exclusive with the `traces_final_ids` prefilter so the project `traces` are scanned once (no CTE-re-evaluation compounding). Gated to default-sort listing; spans/feedback/annotation filters, the uuid time-range path, single-thread pushdown, and streaming fall back to the unchanged query.
- **Close path** (`TraceThreadService.createMissingThreads`): `findThreadsByProject` filtered by `id` alone left `thread_id` (3rd PK column) unbound, scanning the whole project's `trace_threads`. Now also passes `thread_id` so the `(workspace_id, project_id, thread_id, id)` PK seeks (`thread_model_id` maps 1:1 to `thread_id`, so output is unchanged).
- **Feedback path** (`FeedbackScoreService.scoreBatchOfThreads`, `PUT /v1/private/traces/threads/feedback-scores`): resolved thread-model-ids one-per-thread in a `flatMap`; now one bulk `getOrCreateThreadIds` (single `findByProjectIdAndThreadIds`) via the batch path the close flow already uses.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-7035

## AI-WATERMARK

AI-WATERMARK: yes

- If yes:
  - Tools: Claude Code
  - Model(s): Claude Opus 4.8
  - Scope: investigation, ClickHouse/DAO query analysis & rewrites, and prod read-only validation (system.query_log, EXPLAIN ESTIMATE, full-row fingerprints)
  - Human verification: reviewed diffs; ran the backend test suites below; equivalence + cost confirmed against the prod read replica

## Testing

Run from `apps/opik-backend`:
- `mvn -Dtest=FindTraceThreadsResourceTest test` → **466/466** pass (default-sort pagination exercises the page-pushdown; sort/filter/time-range/stream exercise the unchanged fallback).
- `mvn -Dtest='TracesResourceTest$ThreadFeedbackScoresCreation,TracesResourceTest$TraceThreadManualOpenClose,TracesResourceTest$TraceThreadCreation' test` → **17/17** pass (feedback bulk resolution + close `createMissingThreads` + thread creation).
- `mvn -DskipTests compile` clean.

Prod validation (read-only, against `system.query_log` / `EXPLAIN ESTIMATE` / full-row `groupBitXor(cityHash64(*))` fingerprints, ~1M-trace project):
- **Listing**: rewritten query output **identical** to current (page 1, deep page, source filter, back-to-back). Cost: memory **946 MiB → 167 MiB (5.7×)**, bytes read **2.75 GiB → 1.02 GiB (2.7×)**, ~1.5× faster; plans/runs within budget.
- **Close `createMissingThreads`**: single-thread read **278K → 82K rows, 39ms → 13ms** (the PK now seeks; ~14× fewer rows on the worst prod cases that scanned ~1.16M).
- **Feedback bulk**: N per-thread MySQL resolutions → 1 (`performance_schema` not accessible to measure latency directly; reduction is structural and scales with batch size).

Not run locally: full backend integration suite (left to CI). HTTP-endpoint p50/p95 (Prometheus-only metric, not reachable from the validation environment).

## Documentation

No documentation changes.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
@@ -90,6 +90,40 @@ class ThreadDAOImpl implements ThreadDAO {
                 WHERE 1 = 1
                 <if(filters)> AND <filters> <endif>
                 <if(search_text)> AND <search_text> <endif>
+            ), <endif><if(page_pushdown)>page_thread_ids AS (
+                SELECT thread_id
+                FROM (
+                    SELECT
+                        thread_id,
+                        min(start_time) AS start_time,
+                        max(end_time) AS end_time,
+                        max(last_updated_at) AS trace_last_updated_at
+                    FROM (
+                        SELECT id, thread_id, start_time, end_time, last_updated_at
+                        FROM traces
+                        WHERE workspace_id = :workspace_id
+                          AND project_id = :project_id
+                          AND thread_id \\<> ''
+                          <if(filters)> AND <filters> <endif>
+                          <if(search_text)> AND <search_text> <endif>
+                        ORDER BY (workspace_id, project_id, id) DESC, last_updated_at DESC
+                        LIMIT 1 BY id
+                    )
+                    GROUP BY thread_id
+                ) AS pt
+                LEFT JOIN (
+                    SELECT thread_id, id, last_updated_at
+                    FROM trace_threads
+                    WHERE workspace_id = :workspace_id
+                      AND project_id = :project_id
+                    ORDER BY (workspace_id, project_id, thread_id, id) DESC, last_updated_at DESC
+                    LIMIT 1 BY id
+                ) AS ptt ON pt.thread_id = ptt.thread_id
+                ORDER BY if(ptt.last_updated_at = toDateTime64(0, 6, 'UTC'), pt.trace_last_updated_at, ptt.last_updated_at) DESC,
+                    pt.start_time ASC,
+                    pt.end_time DESC,
+                    pt.thread_id
+                LIMIT :limit <if(offset)>OFFSET :offset<endif>
             ), <endif>traces_final AS (
                 SELECT
                     id,
@@ -121,12 +155,18 @@ class ThreadDAOImpl implements ThreadDAO {
                     WHERE workspace_id = :workspace_id
                       AND project_id = :project_id
                       AND thread_id \\<> ''
-                      <if(traces_final_ids)>
-                          AND id IN (SELECT id FROM traces_final_ids)
+                      <if(page_pushdown)>
+                          AND thread_id IN (SELECT thread_id FROM page_thread_ids)
+                          <if(filters)> AND <filters> <endif>
+                          <if(search_text)> AND <search_text> <endif>
                       <else>
-                          <if(uuid_from_time)> AND id >= :uuid_from_time <endif>
-                          <if(uuid_to_time)> AND id \\<= :uuid_to_time <endif>
-                          <if(traces_pushdown_filter)> AND thread_id = :thread_id_pushdown <endif>
+                          <if(traces_final_ids)>
+                              AND id IN (SELECT id FROM traces_final_ids)
+                          <else>
+                              <if(uuid_from_time)> AND id >= :uuid_from_time <endif>
+                              <if(uuid_to_time)> AND id \\<= :uuid_to_time <endif>
+                              <if(traces_pushdown_filter)> AND thread_id = :thread_id_pushdown <endif>
+                          <endif>
                       <endif>
                     ORDER BY (workspace_id, project_id, id) DESC, last_updated_at DESC
                     LIMIT 1 BY id
@@ -145,11 +185,15 @@ class ThreadDAOImpl implements ThreadDAO {
                 FROM spans
                 WHERE workspace_id = :workspace_id
                   AND project_id = :project_id
-                  <if(traces_final_ids)>
-                      AND trace_id IN (SELECT id FROM traces_final_ids)
+                  <if(page_pushdown)>
+                      AND trace_id IN (SELECT id FROM traces_final)
                   <else>
-                      <if(uuid_from_time)> AND trace_id >= :uuid_from_time <endif>
-                      <if(uuid_to_time)> AND trace_id \\<= :uuid_to_time <endif>
+                      <if(traces_final_ids)>
+                          AND trace_id IN (SELECT id FROM traces_final_ids)
+                      <else>
+                          <if(uuid_from_time)> AND trace_id >= :uuid_from_time <endif>
+                          <if(uuid_to_time)> AND trace_id \\<= :uuid_to_time <endif>
+                      <endif>
                   <endif>
                 ORDER BY (workspace_id, project_id, trace_id, parent_span_id, id) DESC, last_updated_at DESC
                 LIMIT 1 BY id
@@ -186,6 +230,7 @@ class ThreadDAOImpl implements ThreadDAO {
                     <endif>
                 <endif>
                 <if(traces_pushdown_filter)> AND thread_id = :thread_id_pushdown <endif>
+                <if(page_pushdown)> AND thread_id IN (SELECT thread_id FROM traces_final) <endif>
                 ORDER BY (workspace_id, project_id, thread_id, id) DESC, last_updated_at DESC
                 LIMIT 1 BY id
             ), feedback_scores_deduped AS (
@@ -430,7 +475,7 @@ class ThreadDAOImpl implements ThreadDAO {
             <else>
             <if(sort_fields)> ORDER BY <sort_fields>, last_updated_at DESC <else> ORDER BY last_updated_at DESC, start_time ASC, end_time DESC <endif>
             <endif>
-            LIMIT :limit <if(offset)>OFFSET :offset<endif>
+            LIMIT :limit <if(page_pushdown)><else><if(offset)>OFFSET :offset<endif><endif>
             SETTINGS log_comment = '<log_comment>'
             ;
             """;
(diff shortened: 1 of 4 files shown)


CometActions:

### Access Information
- **URL:** https://pr-7178.dev.comet.com
- **Namespace:** pr-7178
- **Version:** 2.0.73-7178-merge-2340
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7178%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7178`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadService.java
```diff
@@ -132,6 +135,24 @@ private Mono<UUID> getOrCreateThreadId(UUID projectId, String threadId, Instant
                 .map(TraceThreadIdModel::id));
     }
 
+    @Override
+    public Mono<Map<String, UUID>> getOrCreateThreadIds(@NonNull UUID projectId, @NonNull Set<String> threadIds) {
+        if (threadIds.isEmpty()) {
```
andrescrz: Not a blocker: as usual, if we handle `isEmpty`, let's remove the `NonNull` validation and use `CollectionUtils.isEmpty` to handle it gratefully.

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadService.java
```diff
@@ -132,6 +135,24 @@ private Mono<UUID> getOrCreateThreadId(UUID projectId, String threadId, Instant
...
+        // Bulk get-or-create in a single query instead of one round-trip per thread. Null timestamps match
+        // getOrCreateThreadId(projectId, threadId): a missing thread's model id is generated from the current time.
+        return Mono.deferContextual(context -> {
+            Map<String, Instant> threadIdToTimestamp = new HashMap<>();
+            threadIds.forEach(threadId -> threadIdToTimestamp.put(threadId, null));
+            return traceThreadIdService
+                    .getOrCreateTraceThreadIds(context.get(RequestContext.WORKSPACE_ID), projectId, threadIdToTimestamp)
+                    .map(models -> models.stream()
+                            .collect(Collectors.toMap(TraceThreadIdModel::threadId, TraceThreadIdModel::id,
+                                    (existing, duplicate) -> existing)));
```
andrescrz, line 152: Minor: you could have some field based criteria to dedupe on collection. If `existing` first is a good criteria, good to keep it.

apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
```diff
@@ -90,6 +90,40 @@ OR ilike(input, :search_text)
...
                 <if(filters)> AND <filters> <endif>
                 <if(search_text)> AND <search_text> <endif>
+            ), <endif><if(page_pushdown)>page_thread_ids AS (
+                SELECT thread_id
+                FROM (
+                    SELECT
+                        thread_id,
+                        min(start_time) AS start_time,
+                        max(end_time) AS end_time,
+                        max(last_updated_at) AS trace_last_updated_at
```
andrescrz, line 100: Not a blocker: you might want to reconsider the deduping method and use `argMin`, `argMax` etc. You'll need to balance this suggestion against the filters below, so they still work appropriately. Performance would matter here as well.

apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
```diff
@@ -90,6 +90,40 @@ OR ilike(input, :search_text)
...
+                    FROM (
+                        SELECT id, thread_id, start_time, end_time, last_updated_at
+                        FROM traces
+                        WHERE workspace_id = :workspace_id
+                          AND project_id = :project_id
+                          AND thread_id \\<> ''
+                          <if(filters)> AND <filters> <endif>
+                          <if(search_text)> AND <search_text> <endif>
```
andrescrz, line 108: Please double check these filters on non-key and/or mutable fields. We filter prior to dedupe, so this might not return the right results.

apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
```diff
@@ -90,6 +90,40 @@ OR ilike(input, :search_text)
...
+                    SELECT thread_id, id, last_updated_at
+                    FROM trace_threads
+                    WHERE workspace_id = :workspace_id
+                      AND project_id = :project_id
+                    ORDER BY (workspace_id, project_id, thread_id, id) DESC, last_updated_at DESC
+                    LIMIT 1 BY id
+                ) AS ptt ON pt.thread_id = ptt.thread_id
+                ORDER BY if(ptt.last_updated_at = toDateTime64(0, 6, 'UTC'), pt.trace_last_updated_at, ptt.last_updated_at) DESC,
+                    pt.start_time ASC,
+                    pt.end_time DESC,
```
andrescrz, line 124: Please double check the performance impact of sorting by these non-key fields.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/FindTraceThreadsResourceTest.java
```diff
@@ -451,6 +451,54 @@ void findWithImageTruncation(JsonNode original, JsonNode expected, boolean trunc
...
+            for (int page = 1; page <= pages; page++) {
+                var p = traceResourceClient.getTraceThreads(projectId, null, API_KEY, TEST_WORKSPACE,
+                        List.of(), List.of(), Map.of("page", String.valueOf(page), "size", String.valueOf(size)));
+                assertThat(p.total()).isEqualTo(threadCount);
+                assertThat(p.page()).isEqualTo(page);
+                assertThat(p.content()).hasSize(Math.min(size, threadCount - (page - 1) * size));
+                pagedOrder.addAll(p.content().stream().map(TraceThread::id).toList());
+            }
+
+            assertThat(pagedOrder).containsExactlyElementsOf(fullOrder);
```
andrescrz: Minor: I recommend asserting the full payload of the received page, not jus the ids, for future resilience of the production code.

andrescrz (commented):
Generally looks on the good direction, a few comments to double check before moving forward.

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadService.java
thiagohora: Done in c0bd473 — dropped `@NonNull` on `threadIds` and switched to `CollectionUtils.isEmpty(...)` for graceful null/empty handling.

apps/opik-backend/src/main/java/com/comet/opik/domain/threads/TraceThreadService.java
thiagohora: Kept `existing`-first as you suggested. `thread_model_id` ↔ `thread_id` is 1:1 (UNIQUE `(project_id, thread_id)` on `project_trace_threads`), so duplicate keys should not occur in practice — the merge function is just defensive.

apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
thiagohora: The `LIMIT 1 BY id` here intentionally mirrors the existing `traces_final` dedup so the rewrite stays output-equivalent. Confirmed: a full-row fingerprint (`groupBitXor(cityHash64(*))` over all columns) is byte-identical to the current query on a ~1M-trace prod project across page 1, a deep page (offset 60) and the `source` filter; measured cost 167 MiB vs 946 MiB today. `argMax` would help the dedup on update-heavy threads but changes the validated shape — I would keep it as a follow-up rather than risk equivalence in this PR.

apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
thiagohora: Good catch to verify. The filter placement matches today: the current `traces_final_ids` applies `<filters>`/`<search_text>` to the raw traces *before* `traces_final`'s `LIMIT 1 BY id`, and the resolver does the same. `source` (the active filter) is set at trace creation and not mutated across versions, and the prod full-row fingerprint is identical to the current query *including* the source filter — so results match.

apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
thiagohora: The sort runs over the per-thread aggregated rows (one row per thread after `GROUP BY thread_id`, ~188K for this project), not the raw traces, and is bounded by the page `LIMIT`. Measured end-to-end at 167 MiB / ~2.9s vs the current query's 946 MiB / ~4.3s on the hot project; ordering is deterministic via the `thread_id` tiebreaker.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/FindTraceThreadsResourceTest.java
thiagohora: Done in c0bd473 — the test now asserts the full page payload via `TraceAssertions.assertThreads(...)` instead of just the ids.

apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
thiagohora: Measured this directly (prod, ~1M-trace project). `EXPLAIN PIPELINE` shows it as a partial/limited sort (`PartialSortingTransform → MergeSortingTransform → MergingSortedTransform → Limit`) on the **post-`GROUP BY`** stream (~198K per-thread rows), not the 1.36M raw traces. Isolating the sort by running the resolver with vs without the `ORDER BY`:

| variant | time | memory |
|---|---|---|
| full 4-key ORDER BY | 455 ms | 173.6 MiB |
| no ORDER BY | 279 ms | 160.0 MiB |
| primary key only | 258 ms | 137.0 MiB |

So the entire sort (4 keys incl. the `if(...)` coalesce, vs no sort) is **≈ +14 MiB / +176 ms** — a small slice of a resolver dominated by the upstream scan, and still well under the current query's 946 MiB. Sorting by the non-key fields is bounded because it runs on the aggregated rows as a top-N partial sort.

CometActions:

### Access Information
- **URL:** https://pr-7178.dev.comet.com
- **Namespace:** pr-7178
- **Version:** 2.0.73-7178-merge-2350
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7178%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

andrescrz (approved):

PR #7185, opened 2026-06-21
[NA] [FE] feat: quick-filter on trace/span attributes from the inspect panel

YarivHashaiComet:
## Details

Adds a one-click **quick-filter** on individual attributes in the trace/span **Inspect → Details** panel: hovering an attribute line in the Metadata/Input/Output code viewers reveals a **+** that opens a small **Add filter** modal (operator + value, Cancel/Apply) and seeds the matching list filter. The goal is to let users filter by what they're already looking at instead of retyping JSON paths/values by hand.

- Attribute path + value are resolved from the CodeMirror **JSON/YAML syntax tree**, so the raw code view is preserved (no replacement tree view).
- Routing: Metadata keys → `metadata` dictionary filter; Input/Output keys → `custom` filter (`input.`/`output.`-prefixed); a span's `provider` → the dedicated, stored `provider` field (spans only). A trace's computed `providers` aggregate is intentionally not offered (no stored column to filter on).
- Opt-in via a context provided only by the Logs page, so the shared Inspect panel (experiments, prompts, SME flow, …) is unaffected. Also surfaces the existing `SpanField.PROVIDER` backend filter as a Provider chip on the Spans tab.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-
- N/A — no ticket (`[NA]`); UX polish on the Logs Inspect panel.

## AI-WATERMARK

AI-WATERMARK: yes

- If yes:
  - Tools: Claude Code
  - Model(s): Claude Opus 4.8
  - Scope: Full implementation (frontend only), unit tests, and code-review fixes. Backend untouched.
  - Human verification: Manually tested end-to-end against a local backend with seeded multi-provider traces/threads/spans; reviewed via `/comet-cra-review` with findings addressed.

## Testing

- `cd apps/opik-frontend && npx vitest run src/shared/SyntaxHighlighter/quickFilterPaths.test.ts src/shared/filter-chips/hooks/useQuickAttributeFilterActions.test.ts` — **26 tests passing**. Syntax-tree path resolution is validated against the real `@lezer/json` and `@lezer/yaml` parsers (nested objects, multi-element arrays/sequences, escaped scalars, key-vs-value), plus the routing/dialog hook (canFilter gating, provider routing, dialog lifecycle, dedupe, fallback operators).
- `tsc --noEmit` clean; ESLint clean on all changed files.
- Manual (local process mode, Chrome): seeded a `quick-filter-demo` project with 6 traces across 2 threads + 2 standalone and 12 spans (openai/anthropic/google_ai). Verified: hover **+** on Metadata/Input/Output → modal → filter applies; span `provider` routes to the Provider field and matches; trace `providers` shows no **+**; the raw JSON/YAML view is unchanged.
- Not run: full FE suite / Playwright E2E (scoped, frontend-only change).

## Documentation

No documentation changes — internal UX affordance on the existing Logs Inspect panel; no API/SDK surface changed.


apps/opik-frontend/src/shared/SyntaxHighlighter/quickFilterExtension.ts
```diff
@@ -0,0 +1,147 @@
...
+    super();
+  }
+
+  eq(other: QuickFilterWidget) {
+    return other.path === this.path && other.value === this.value;
+  }
+
+  toDOM() {
+    const button = document.createElement("span");
+    button.className = "cm-quick-filter-add";
```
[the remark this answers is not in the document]
YarivHashaiComet: Intentional: this is a CodeMirror theme class, and `cm-` is CodeMirror's own namespace. The sibling `foldingExtension.ts` in this codebase uses the same convention (`cm-inline-fold-marker`, `cm-inline-fold-placeholder`). The `comet-` prefix rule applies to app/Tailwind classes, not editor-internal theme selectors — renaming would diverge from the established CodeMirror pattern. Keeping as-is.

apps/opik-frontend/src/shared/SyntaxHighlighter/quickFilterPaths.ts
```diff
@@ -0,0 +1,149 @@
...
+    enter: (ref) => {
+      if (!scalars.has(ref.name)) return;
+      const node = ref.node;
+      const path = mode === "json" ? jsonPath(node, doc) : yamlPath(node, doc);
+      if (path === null) return;
+      targets.push({
+        pos: node.to,
+        path,
+        value: unquote(doc.slice(node.from, node.to)),
+      });
```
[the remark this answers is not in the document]
YarivHashaiComet: Good catch — fixed in b15805bec3. Rather than seed a literal "null", null-valued attributes no longer show the `+` at all: the metadata/custom dictionary filter has no is-empty operator, so a null filter could never match (same rationale as not offering the trace `providers` aggregate). JSON `Null` nodes and YAML `null`/`~` scalars are now excluded, with unit tests added.

CometActions:

The deployment encountered an error. Please check the [deployment logs](https://github.com/comet-ml/comet-deployment/actions/runs/27903544409) for details.

CometActions:

### Access Information
- **URL:** https://pr-7185.dev.comet.com
- **Namespace:** pr-7185
- **Version:** 2.0.74-7185-merge-2320
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7185%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

apps/opik-frontend/src/shared/SyntaxHighlighter/quickFilterExtension.ts
```diff
@@ -0,0 +1,147 @@
...
+    marginLeft: "6px",
+    color: "var(--codemirror-gutter)",
+    cursor: "pointer",
+    opacity: "0",
+    transition: "opacity 0.1s ease-in-out, color 0.1s ease-in-out",
+  },
+  ".cm-line:hover .cm-quick-filter-add": {
+    opacity: "0.55",
+  },
+  ".cm-quick-filter-add:hover, .cm-quick-filter-add:focus-visible": {
```
JetoPistola, line 268: 💡 **suggestion** | UX / Discoverability

The whole feature hangs on a per-line affordance that is invisible at rest: `.cm-quick-filter-add` is `opacity: 0` here, fading to only `0.55` on `.cm-line:hover`, with discovery resting entirely on the native `title`/`aria-label` ("Filter by this attribute"). So to find this, a user has to hover a code line, notice a faint `+`, *guess* it's interactive, then hover/click it to read the tooltip. Before any hover there's no signal at all that these attributes are filterable — it reads as a hidden feature.

Two separate points worth considering:

1. **Discoverable entry point.** A lone faint `+` revealed only on hover is hard to find. Consider a visible, labelled affordance near each section — e.g. a **"Filter logs by attribute"** button/header next to Input / Output / Metadata — so users can discover the capability without hovering individual lines. (Alternatives: raise the rest-state opacity so the `+` is faintly but persistently visible; a one-time coachmark on first open; a docs/release-note mention.)

2. **Use a filter icon, not a `+`.** The affordance currently renders a `Plus` icon, but the action is "filter by this attribute," not "add." Reusing the product's standard **filter icon** (the same one used elsewhere in the filter UI) would make the intent legible the moment it's seen and stay consistent with the rest of the app.

Not blocking — the underlying mechanic is solid; this is about making the feature findable and self-explanatory.

apps/opik-frontend/src/shared/filter-chips/hooks/useQuickAttributeFilterActions.ts
```diff
@@ -0,0 +1,170 @@
...
+import {
+  QuickAttributeFilterApi,
+  QuickFilterSection,
+} from "@/shared/filter-chips/QuickAttributeFilterContext";
+import { QuickFilterDraft } from "@/shared/filter-chips/QuickFilterDialog";
+
+const METADATA_CHIP_ID = "metadata";
+const CUSTOM_CHIP_ID = "custom";
+const PROVIDER_CHIP_ID = "provider";
+const DEFAULT_OPERATOR: FilterOperator = "=";
```
JetoPistola: ❓ **question** | UX

`defaultOperator` for the seeded draft is hard-coded to `=` regardless of the target chip's own `defaultOperator` (metadata/custom dictionary chips typically default to `contains`). Seeding `=` (exact match) against a value pulled verbatim from the JSON/YAML is the right default for short scalars (provider, branch, count), but for a long Input/Output string an exact match will usually return nothing, which could read as "the filter is broken."

Is the always-`=` default intentional, or would it be better to inherit the chip's `defaultOperator` (so input/output default to `contains`)? The user can change the operator in the dialog either way, so this is minor.

apps/opik-frontend/src/shared/filter-chips/hooks/useQuickAttributeFilterActions.ts
```diff
@@ -0,0 +1,170 @@
...
+}
+
+const getRows = (value: ChipValue | undefined): Filter[] => {
+  if (!value || typeof value !== "object") return [];
+  const candidate = (value as QueryBuilderChipValue).rows;
+  return Array.isArray(candidate) ? candidate : [];
+};
+
+export const stringifyFilterValue = (value: JsonValue): string => {
+  if (value === null) return "";
```
JetoPistola, line 43: 🧹 **nit** | Dead code

`stringifyFilterValue(null) → ""` is effectively unreachable in practice: null-valued attributes are already excluded upstream in `collectQuickFilterTargets` (JSON `Null` / YAML null tokens never produce a target), so `value` here is always a non-null scalar. The test even acknowledges this ("only reached for scalars in practice"). Fine to keep as defensive code, but if you'd rather not carry a branch that can't fire, the `null` case could go. Non-issue either way.

apps/opik-frontend/src/shared/SyntaxHighlighter/quickFilterExtension.ts
JetoPistola, line 268: 💡 **suggestion** | UX / Round-trip feedback

Right now the feature is write-only: `QuickFilterCodeConfig` is just `{ canFilter, onFilter }`, and the widget/theme never consult the active filter values. So when you open the Inspect panel on a row that's **already** being filtered by one of its attributes, the JSON/YAML view looks identical to any other row — the attribute line that's driving the current filter gets no visual indication.

It'd close the loop to reflect active filters back onto the attribute lines: e.g. highlight (or badge) the line for any attribute that matches a currently-applied filter, so the panel reads as "this attribute is being filtered" and not only "click to add a filter." That likely means passing the active filter state into the extension (an `isFiltered(path)` predicate alongside `canFilter`) and adding an applied-state style next to the existing `.cm-quick-filter-add` rules here.

Not blocking — it's an enhancement to make the affordance bidirectional rather than one-way.

JetoPistola:
👋 **Review summary**

**What looks good**
- The CodeMirror extension is a clean sibling of the existing `foldingExtension.ts` — same widget shape, `eq`/`ignoreEvent`, viewport-based rebuild, `renderToStaticMarkup` icon. It slots into the established pattern instead of inventing a new one.
- Path resolution walks the **full** syntax tree for parent/index traversal while only enumerating scalars in the visible range — so array indices stay correct under virtualization. Nicely thought through.
- Rows are seeded with just `key`/`operator`/`value` and let `queryBuilderToFilters` inject `field`/`type` from the chip definition — consistent with how the chips already store rows, no backend coupling leaking into the quick-filter.
- Provider routing is precise and backend-valid: span `provider` → the dedicated `SpanField.PROVIDER` (`STRING`) column, the trace `providers` aggregate correctly offers nothing, input/output → the prefixed `custom` filter.
- Strong test coverage — 26 tests against the real `@lezer/json` / `@lezer/yaml` parsers (nesting, multi-element arrays, escapes, null exclusion) plus the routing/dialog hook.

**Overall**
Solid, well-scoped, frontend-only feature; the approach (resolve from the syntax tree, opt-in via a Logs-only context, reuse the chip pipeline) is the right one and keeps the shared Inspect panel unaffected. The bulk of the feedback is UX discoverability/feedback rather than correctness.

**Inline comments**: 2 suggestions, 1 question, 1 nit — nothing blocking.
- 💡 *Discoverability + affordance*: the per-line action is invisible at rest and a `+` reads as "add" rather than "filter" — consider the product's standard **filter icon** and a discoverable section-level entry point (e.g. a "Filter logs by attribute" button next to Input/Output/Metadata).
- 💡 *Round-trip feedback*: when viewing a row whose attribute is **already** an active filter, the line isn't highlighted/indicated — the affordance is currently one-way (add only). Reflecting active filters back onto the attribute lines would close the loop.
- ❓ *Default operator*: the seeded draft is always `=`, even for dictionary chips that default to `contains` — intentional?
- 🧹 *Nit*: the `null` branch in `stringifyFilterValue` is effectively unreachable (nulls are excluded upstream).

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7185`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

apps/opik-frontend/src/shared/filter-chips/hooks/useQuickAttributeFilterActions.ts
YarivHashaiComet: keeping the protection of NPE for future bulletproof

CometActions:

### Access Information
- **URL:** https://pr-7185.dev.comet.com
- **Namespace:** pr-7185
- **Version:** 2.0.74-7185-merge-2338
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7185%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

CometActions:

### Access Information
- **URL:** https://pr-7185.dev.comet.com
- **Namespace:** pr-7185
- **Version:** 2.0.74-7185-merge-2346
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7185%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7185`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

apps/opik-frontend/src/shared/filter-chips/hooks/useQuickAttributeFilterActions.ts
```diff
@@ -0,0 +1,139 @@
...
+  values: ChipValueMap;
+  applyValue: (id: string, value: ChipValue) => void;
+  pinChip: (id: string) => void;
+}
+
+const getRows = (value: ChipValue | undefined): Filter[] => {
+  if (!value || typeof value !== "object") return [];
+  const candidate = (value as QueryBuilderChipValue).rows;
+  return Array.isArray(candidate) ? candidate : [];
+};
```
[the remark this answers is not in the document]
YarivHashaiComet: Addressed in 1233069 — extracted the duplicated `getRows` into `@/shared/filter-chips/lib/helpers.ts` (alongside the existing `trimValue`/`toNumber` helpers) and imported it in both `useTagsChipActions` and `useQuickAttributeFilterActions`, so the `values[chipId]`/`rows` unwrap now lives in one place.

apps/opik-frontend/src/shared/SyntaxHighlighter/quickFilterExtension.ts
```diff
@@ -0,0 +1,292 @@
...
+  ViewPlugin,
+  ViewUpdate,
+  WidgetType,
+} from "@codemirror/view";
+import {
+  Extension,
+  RangeSetBuilder,
+  StateEffect,
+  StateField,
+} from "@codemirror/state";
```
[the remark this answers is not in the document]
YarivHashaiComet: Keeping this as-is. Every sibling in `src/shared/SyntaxHighlighter/` (`foldingExtension.ts`, `base64Extension.ts`, `constants.ts`, `quickFilterPaths.ts`, `types.ts`) orders `@codemirror/*`/`@lezer/*` imports first and `react` afterward. Applying the generic "react-first" rule here would make this the only inconsistent file in the folder, so I am matching the local CodeMirror-extension convention.

apps/opik-frontend/src/shared/filter-chips/hooks/useQuickAttributeFilterActions.ts
```diff
@@ -0,0 +1,139 @@
...
+  type: TRACE_DATA_TYPE,
+  path: string,
+): FilterTarget | null => {
+  if (section === "metadata") {
+    if (isProvidersAggregateKey(path)) return null;
+    if (isProviderRootKey(path)) {
+      // Spans store provider in a dedicated column; traces have no such field.
+      return type === TRACE_DATA_TYPE.spans
+        ? { chipId: PROVIDER_CHIP_ID }
+        : null;
```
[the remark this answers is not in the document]
YarivHashaiComet: This guard is intentional, not a bug — a `metadata.providers` filter on traces would be a dead filter. For traces, `providers` is a **read-time computed aggregate, not stored metadata**:

- `TraceDAO` builds it via `arraySort(groupUniqArrayIf(provider, provider != ''))` over the trace's spans.
- `getMetadataWithProviders()` injects it into the response only: `JsonUtils.prependField(baseMetadata, "providers", providers)` — it is never written to the stored `metadata` column.
- A `metadata.providers` filter compiles (FilterQueryBuilder) to JSON path `$."providers"` against that stored column, which has no such key → matches nothing.

So hiding the affordance is correct. The path-walker still emits `providers[0]` (per `quickFilterPaths.test.ts`); the resolve layer is what correctly suppresses it for traces. The span `provider` root key is handled separately and routed to the dedicated, column-backed provider chip, which is filterable.

CometActions:

### Access Information
- **URL:** https://pr-7185.dev.comet.com
- **Namespace:** pr-7185
- **Version:** 2.0.74-7185-merge-2349
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7185%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7185`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

CometActions:

### Access Information
- **URL:** https://pr-7185.dev.comet.com
- **Namespace:** pr-7185
- **Version:** 2.0.74-7185-merge-2361
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7185%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

apps/opik-frontend/src/shared/SyntaxHighlighter/quickFilterExtension.ts
```diff
@@ -0,0 +1,342 @@
...
+// It uses fixed positioning and flips above/below to stay on-screen. The same
+// element serves both the hover hint and the post-click "Filter applied"
+// confirmation, which reveals a leading success-green check icon.
+let quickFilterTooltip: HTMLElement | null = null;
+let quickFilterTooltipIcon: HTMLElement | null = null;
+let quickFilterTooltipText: HTMLElement | null = null;
+
+const getQuickFilterTooltip = (): HTMLElement => {
+  if (!quickFilterTooltip) {
+    quickFilterTooltip = document.createElement("div");
```
aadereiko: **[medium] Module-level mutable tooltip singleton shared across all editor instances**

quickFilterTooltip / quickFilterTooltipIcon / quickFilterTooltipText are module-scoped mutable `let` bindings appended once to document.body and mutated by every QuickFilterWidget's event handlers. The Logs trace panel mounts multiple CodeBlock/CodeMirror instances simultaneously (e.g. input + output viewers, plus a section per attribute group), and they all share this one DOM node and its show/hide/applied state. Hovering an icon in one editor while another editor's confirmation is up is last-write-wins, the node is never removed from <body> (it lingers after every editor unmounts), and the state leaks across test runs. This is exactly the module-level-mutable-state-as-a-side-channel pattern the team flags.

💡 Own the tooltip element inside the ViewPlugin instance (create in the plugin constructor, remove in destroy()) so each editor manages its own portaled node, or keep it module-level but reference-count creation/teardown so the body node is cleaned up when the last editor unmounts. At minimum scope the active-anchor state to the widget that currently owns it.

_rule: `frontend-architecture-avoid-module-level-mutable-state-as-a-side-channel-from-hook`_

apps/opik-frontend/src/shared/SyntaxHighlighter/quickFilterExtension.ts
```diff
@@ -0,0 +1,342 @@
...
+      // Keep the read-only editor from moving the caret / selection.
+      event.preventDefault();
+      event.stopPropagation();
+      this.onFilter(this.path, this.value);
+      // The icon is still hovered/focused after the click, so swap the live
+      // tooltip to the "Filter applied" confirmation in place.
+      markQuickFilterApplied(button);
+    };
+    button.onmousedown = activate;
+    button.onkeydown = (event) => {
```
aadereiko, line 221: **[low] "Filter applied" confirmation can be invisible when applying the filter rebuilds the widget**

activate() calls this.onFilter(...) and then markQuickFilterApplied(button) to show the in-place confirmation. On the Logs page onFilter mutates the active filter set, which can re-render the panel and trigger the ViewPlugin update (docChanged/viewportChanged) that rebuilds decorations; CodeMirror then destroys the old widget DOM, and QuickFilterWidget.destroy() calls hideQuickFilterTooltip() on the next frame. In that path the confirmation tooltip is shown and immediately hidden, so the user never sees it. The feature still works, but the confirmation affordance is unreliable depending on whether applying the filter rebuilds this editor.

💡 Decouple the confirmation from the per-widget DOM lifecycle (e.g. drive it from a short-lived timer on the shared tooltip, or render the confirmation independently of the widget that triggered it) so it survives a decoration rebuild long enough to be seen.

_general finding_

apps/opik-frontend/src/shared/SyntaxHighlighter/quickFilterPaths.test.ts
```diff
@@ -0,0 +1,118 @@
+import { describe, it, expect } from "vitest";
+import { jsonLanguage } from "@codemirror/lang-json";
+import { yamlLanguage } from "@codemirror/lang-yaml";
+import { collectQuickFilterTargets } from "./quickFilterPaths";
+
+const json = (doc: string) =>
+  collectQuickFilterTargets(jsonLanguage.parser.parse(doc), doc, "json")
+    .map(({ path, value }) => ({ path, value }))
```
aadereiko, line 8: **[low] Tests assert only path/value, never the load-bearing pos/from offsets**

Both the `json` and `yaml` helpers map each target down to `{ path, value }` and drop the `pos` and `from` fields. Those two fields are not cosmetic: quickFilterExtension.ts sorts targets by `target.pos`, places the filter icon at `target.pos`, and uses `[target.from, target.pos]` as the value span and decoration range (lines 294-305). collectQuickFilterTargets is the only producer of these offsets, so a regression in `node.from`/`node.to` computation (e.g. including or excluding surrounding quotes) would mis-position every icon yet pass this entire suite. The branching path/value logic is well covered, but the offset contract is completely untested.

💡 Add at least one assertion on a target's `from`/`pos` for a known document (e.g. for `{"k":"v"}` assert the value node spans the quoted-string offsets), or assert the value span via `doc.slice(from, pos)`. Keep it to a single representative case to avoid brittleness rather than asserting offsets on every case.

_rule: `frontend-testing-add-vitest-unit-tests-for-non-trivial-hook-and-utility-logic`_

apps/opik-frontend/src/shared/SyntaxHighlighter/quickFilterPaths.test.ts
```diff
@@ -0,0 +1,118 @@
...
+    .sort((a, b) => a.path.localeCompare(b.path));
+
+const yaml = (doc: string) =>
+  collectQuickFilterTargets(yamlLanguage.parser.parse(doc), doc, "yaml")
+    .map(({ path, value }) => ({ path, value }))
+    .sort((a, b) => a.path.localeCompare(b.path));
+
+describe("collectQuickFilterTargets - JSON", () => {
+  it("maps nested objects, arrays, and scalar types to paths", () => {
+    const doc = JSON.stringify({
```
aadereiko, line 18: **[low] Flow-style YAML and JSON nested-array edge cases are untested**

The production code has dedicated branches that no test exercises: yamlPath handles FlowSequence (`[a, b]`) in addition to BlockSequence, and the PUNCTUATION/namedIndex logic exists specifically to compute correct indices when Lezer surfaces `[`/`]`/`,`/`{`/`}`/`:` as named children (the most fragile part of the module). All YAML sequence tests use block style (`- x`) and all JSON array tests are single-level. A nested array (`[[1],[2]]`) or a YAML flow sequence/mapping would meaningfully exercise namedIndex and the punctuation-skipping, which the edge-case-filtering intent of this rule targets.

💡 Add a JSON nested-array case (e.g. `{a:[[1],[2]]}` expecting `a[0][0]`/`a[1][0]`) and a YAML flow case (e.g. `providers: [openai, anthropic]`) so the namedIndex/punctuation-skip path is covered, not just block-style single-level collections.

_rule: `frontend-testing-add-vitest-unit-tests-for-non-trivial-hook-and-utility-logic`_

apps/opik-frontend/src/shared/SyntaxHighlighter/quickFilterPaths.ts
```diff
@@ -0,0 +1,164 @@
+import { SyntaxNode, Tree } from "@lezer/common";
```
aadereiko, line 1: **[low] Direct import from @lezer/common, which is not a declared dependency**

SyntaxNode and Tree are imported directly from "@lezer/common", but package.json only declares "@codemirror/language" (^6.0.0); @lezer/common is present only as a transitive/hoisted dependency. This is the sole file in src importing @lezer/common directly. Relying on a hoisted transitive package for a direct import is fragile: a future hoisting change, a stricter package manager (pnpm), or a version bump can break the build with no warning. These are type-only symbols, so the import is erased at runtime, but the type resolution still depends on the package being reachable.

💡 Import the re-exported types from the declared dependency instead: `import { SyntaxNode, Tree } from "@codemirror/language";` (which re-exports them from @lezer/common), or add "@lezer/common" to package.json dependencies.

_rule: `frontend-api-contract-include-all-required-query-parameters-and-verify-third-party`_

apps/opik-frontend/src/shared/SyntaxHighlighter/quickFilterPaths.ts
```diff
@@ -0,0 +1,164 @@
...
+      ) {
+        return;
+      }
+      const path = mode === "json" ? jsonPath(node, doc) : yamlPath(node, doc);
+      if (path === null) return;
+      targets.push({
+        pos: node.to,
+        from: node.from,
+        path,
+        value: unquote(raw),
```
aadereiko: **[low] BlockLiteral value is not unquoted/normalized**

YAML_VALUE_SCALARS includes "BlockLiteral" (block scalars introduced by | or >). For such a node, doc.slice(node.from, node.to) includes the block indicator and the raw indented content, and unquote() leaves it untouched (it only handles leading/trailing single/double quotes). The resulting `value` passed to the filter therefore contains the raw multi-line block text (indicator, newlines, indentation) rather than the logical scalar value, so a 'contains' filter seeded from a block scalar will use a malformed value. This path is untested.

💡 Decide whether block scalars should be filterable; if so, normalize the block-scalar text (strip the indicator line and dedent) before using it as the value, or exclude BlockLiteral from the value-scalar set until normalization is implemented. Add a block-scalar test case.

_general finding_

apps/opik-frontend/src/shared/SyntaxHighlighter/quickFilterPaths.ts
```diff
@@ -0,0 +1,164 @@
...
+    if (parent.name === "Property") {
+      const nameNode = parent.getChild("PropertyName");
+      if (!nameNode) return null;
+      parts.unshift({ key: unquote(doc.slice(nameNode.from, nameNode.to)) });
+    } else if (parent.name === "Array") {
+      const idx = namedIndex(parent, cur);
+      if (idx < 0) return null;
+      parts.unshift({ index: idx });
+    }
+    cur = parent;
```
aadereiko, line 98: **[low] Keys containing '.' or '[]' produce ambiguous dotted paths**

buildPath joins object keys with '.' and array indices with '[n]' without escaping. A property whose key literally contains a dot or bracket (e.g. {"a.b": 1} or {"x[0]": 1}) produces a path 'a.b' / 'x[0]' that is indistinguishable from the nested-path / array-index forms. The downstream consumer (resolveQuickFilterTarget) keys the metadata filter on this exact path string, so such a key would map to the wrong (or a non-existent) filter target. This is an uncommon-but-real data shape for metadata blobs.

💡 This is a known limitation of dotted-path notation; if metadata can contain dotted keys, document the limitation in a comment, or use a bracket-quoted segment for keys containing '.', '[' or ']' so the path round-trips unambiguously.

_general finding_

apps/opik-frontend/src/shared/filter-chips/hooks/useQuickAttributeFilterActions.ts
```diff
@@ -0,0 +1,130 @@
...
+      if (!target) return;
+
+      const { chipId, key } = target;
+      const stringValue = stringifyFilterValue(value);
+
+      const existing = getRows(values[chipId]);
+      const alreadyApplied = existing.some(
+        (row) =>
+          (row.key ?? "") === (key ?? "") &&
+          row.operator === QUICK_FILTER_OPERATOR &&
```
aadereiko, line 102: **[low] Null/empty attribute value applies a silent no-op filter row**

stringifyFilterValue(null) returns "" and the row is created with operator "contains" and value "". Downstream, queryBuilderToFilters keeps only rows that pass isFilterValid (lib/filters.ts), which requires value !== "" for operators other than is_empty/is_not_empty. So when the clicked attribute's value is null (or any value that stringifies to ""), the chip is still pinned and a row is pushed via applyValue, but it never becomes an actual applied filter — the user sees the chip open with an empty, do-nothing row and no filtering occurs. canFilter only guards Boolean(path), not the value, so the action is offered for null-valued attributes. This is an edge case (displayed scalar attributes are rarely null), hence low.

💡 Either have canFilter/filter treat a null/blank stringified value as not-filterable (so the quick-filter action is hidden), or seed the row with an is_empty operator when the value is empty so the produced filter is valid and meaningful.

_general finding_

apps/opik-frontend/src/shared/filter-chips/hooks/useQuickAttributeFilterActions.ts
```diff
@@ -0,0 +1,130 @@
...
+          (row) => (row.key ?? "") !== "" || (row.value ?? "") !== "",
+        );
+        applyValue(chipId, { rows: [...nextRows, nextRow] });
+      }
+
+      pinChip(chipId);
+
+      // Usage signal: which tab (traces/spans) and Inspect section the
+      // quick-filter was applied from.
+      trackEvent(OpikEvent.QUICK_FILTER_APPLIED, {
```
aadereiko, line 121: **[low] QUICK_FILTER_APPLIED event omits table_id that all sibling filter events carry**

useFilterChipsAnalytics wraps trackEvent with a `track` helper that attaches table_id to every filter event (FILTER_APPLIED, FILTER_PINNED, FILTERS_ACTIVE_COUNT, etc.). This hook calls trackEvent(OpikEvent.QUICK_FILTER_APPLIED, ...) directly with data_type/source/filter_name/operator but no table_id, so quick-filter usage cannot be correlated with the same table dimension as the rest of the filter funnel. Low because it is a distinct event and the included data_type/source already give useful context.

💡 If quick-filter usage should be analyzable alongside other filter events, thread the tableId into this hook (or reuse the analytics hook's track wrapper) and include table_id in the QUICK_FILTER_APPLIED payload for consistency.

_general finding_

apps/opik-frontend/src/v2/pages/LogsPage/TracesSpansTab/TracesSpansTab.tsx
```diff
@@ -1064,6 +1078,19 @@ export const TracesSpansTab: React.FC<TracesSpansTabProps> = ({
...
 
+  const { canFilter: quickFilterCanFilter, filter: quickFilterFilter } =
+    useQuickAttributeFilterActions({
+      type,
+      values: chipValues,
+      applyValue: applyChipValue,
+      pinChip,
+    });
+
+  const quickAttributeFilterApi = useMemo(
```
aadereiko: **[low] Redundant re-memoization of an already-stable hook return value**

useQuickAttributeFilterActions already returns a useMemo-wrapped { canFilter, filter } object (see useQuickAttributeFilterActions.ts line 129, and canFilter/filter are themselves useCallback-stable). The component destructures those two stable references back out and then re-wraps them into a fresh object via another useMemo with the same dependencies. Since the hook's return is already referentially stable for the provider value, this second useMemo is pure overhead — it adds an extra allocation and an extra layer that can drift from the hook's contract.

💡 Keep the hook's return object directly and pass it to the provider, e.g. `const quickAttributeFilterApi = useQuickAttributeFilterActions({ type, values: chipValues, applyValue: applyChipValue, pinChip });` then `<QuickAttributeFilterProvider value={quickAttributeFilterApi}>`. Drop the destructuring and the extra useMemo.

_rule: `frontend-performance-do-not-wrap-trivial-or-cheap-expressions-in-usememo`_

aadereiko (commented):
In general, the PR looks ok, I would just check the comments that code review agent left and try to get rid of verbose comments, they are not always needed

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7185`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

YarivHashaiComet:
Thanks @aadereiko — addressed the review in acd309e plus a few follow-ups from a local code-review pass:

**Your findings**
- **[med] tooltip singleton** → tooltip is now owned per-editor by the ViewPlugin (created in ctor, removed in `destroy()`), looked up via a `WeakMap<EditorView,…>` — no shared `<body>` node to leak or cross-wire.
- **[low] confirmation can be invisible on rebuild** → the "Filter applied" confirmation is now timer-driven (auto-reverts ~1.5s) and decoupled from the widget lifecycle. The root cause was `api` churning every render and rebuilding the extension; fixed by reading chip values via a ref **and** memoizing the `useFilterChipsAnalytics` return (it was an unmemoized object, which made `applyValue`/`pinChip` — and everything derived — unstable). This also removes the per-render CodeMirror rebuild for the whole filter-chips system.
- **[low] `@lezer/common` undeclared** → declared it in `package.json`. Note: importing the types from `@codemirror/language` does not work — it declares them locally but does not re-export them (tsc TS2459), so the explicit dependency is the correct fix.
- **[low] BlockLiteral value** → excluded YAML block scalars from filterable targets (+ test).
- **[low] null/empty value no-op** → empty values are now skipped at the collector (no icon), same as null.
- **[low] missing table_id** → `QUICK_FILTER_APPLIED` now carries `table_id`.
- **[low] redundant memo** → pass the hook return straight to the provider.
- **[low] dotted/`[]` keys** → documented the limitation (matches the existing `getJSONPaths` behavior) + characterization tests.
- **[low] offset/flow-style/nested-array test gaps** → added `from/pos` offset assertions (JSON + YAML), flow-style YAML, and nested-array cases.

**Verbose comments** trimmed per your note.

Held: the `useTagsChipActions` row-append duplication — the two flows genuinely diverge (dedup key, blank-row predicate, early-return, BI tail), so a shared helper would need ~4 knobs; happy to revisit if you feel strongly.

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7185`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7185`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7185`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7185`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

CometActions:

The deployment encountered an error. Please check the [deployment logs](https://github.com/comet-ml/comet-deployment/actions/runs/28352447002) for details.

CometActions:

The deployment encountered an error. Please check the [deployment logs](https://github.com/comet-ml/comet-deployment/actions/runs/28355294650) for details.

aadereiko (approved):
LGTM, thanks for addressing the comments

CometActions:

The deployment encountered an error. Please check the [deployment logs](https://github.com/comet-ml/comet-deployment/actions/runs/28365009275) for details.

PR #7195, opened 2026-06-22
[OPIK-7051] [BE] feat: per-workspace online-scoring sampling metric + populate error workspace

thiagohora:
## Details

Adds the per-workspace observability the online-scoring funnel was missing. Two stages had no per-workspace signal: **sampled-to-be-scored** (the sampling decision lived only in-memory in `OnlineScoringSampler`) and **failed-to-process** (`online_scoring_*_processing_errors_total` already carries `workspace_id`/`workspace_name`, but the online scorers never overrode `messageContext()`, so it always emitted `unknown`). Metrics only — no change to scoring behavior.

- New counter `online_scoring_sampler_decisions_total{workspace_id, evaluator_type, decision}` (`decision ∈ sampled | skipped_disabled | skipped_filter | skipped_sampling`); `sampled` = enqueued for scoring, i.e. the previously-missing middle funnel stage per workspace.
- `OnlineScoringBaseScorer` now overrides `messageContext()` to return the message's workspace/user, so `processing_errors_total` reports the real `workspace_id` instead of `unknown`.
- New marker interface `WorkspaceScopedMessage` (`workspaceId()`, `userName()`) implemented by the six `*ToScore*` records; the base scorer's type param is bounded to it.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- OPIK-7051
- Parent epic: OPIK-4063 (Opik Application Monitoring Enhancement)

## AI-WATERMARK

AI-WATERMARK: yes

- If yes:
  - Tools: Claude Code
  - Model(s): Claude Opus 4.8
  - Scope: Authored the metric instrumentation (sampler decisions counter, `messageContext` override, `WorkspaceScopedMessage` interface) and this description.
  - Human verification: Author reviewed the diff; `mvn compile` on opik-backend passes (BUILD SUCCESS). Metric names/labels validated against the live Prometheus/ClickHouse-backed dashboard.

## Testing

- `mvn compile` on `apps/opik-backend` → BUILD SUCCESS.
- No behavior change to scoring; instrumentation only. The sampling-decision counter is incremented on each existing decision branch in `shouldSampleTrace`; `messageContext()` mirrors the workspace/user already written into the Reactor context in `processEvent`.
- Cardinality reviewed: `workspace_id × evaluator_type (6) × decision (4)`, bounded by number of workspaces — consistent with the existing `opik_attachments_upload_*{workspace_id}` precedent.
- Not run: unit/integration suite (no behavioral logic added). Can add a sampler-counter assertion if preferred.

## Documentation

Internal Grafana dashboard already has panels wired for both new metrics; they activate on deploy. No public docs change.


apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSampler.java
```diff
@@ -292,7 +323,10 @@ private boolean shouldSampleTrace(AutomationRuleEvaluator<?, ?> evaluator, Strin
 
         var shouldBeSampled = secureRandom.nextFloat() < evaluator.getSamplingRate();
 
-        if (!shouldBeSampled) {
+        if (shouldBeSampled) {
+            recordDecision(workspaceId, evaluator, DECISION_SAMPLED);
```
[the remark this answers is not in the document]
thiagohora: Fixed in 8cf81eb. `DECISION_SAMPLED` is now recorded at the enqueue site (`recordDecision(..., DECISION_SAMPLED, messages.size())` right before `enqueueMessage`), so it reflects messages actually published to Redis. The python-evaluator-disabled branch never reaches enqueue, so it no longer counts as `sampled`. `shouldSampleTrace` now only records the skip decisions.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSampler.java
```diff
@@ -270,6 +299,7 @@ private List<Trace> stampMissingProjectNames(List<Trace> traces, String workspac
     private boolean shouldSampleTrace(AutomationRuleEvaluator<?, ?> evaluator, String workspaceId, Trace trace) {
         // Check if rule is enabled first
         if (!evaluator.isEnabled()) {
+            recordDecision(workspaceId, evaluator, DECISION_SKIPPED_DISABLED);
             // Important to set the workspaceId for logging purposes
             try (var logContext = createTraceLoggingContext(workspaceId, evaluator, trace)) {
                 userFacingLogger.info(
```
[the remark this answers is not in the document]
thiagohora: Done in 8cf81eb — extracted a `skip(workspaceId, evaluator, trace, decision, message, args...)` helper that records the decision and emits the user-facing log; the disabled / filter-mismatch / sampling branches now each `return skip(...)` with just their `DECISION_*` symbol and message.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringBaseScorer.java
```diff
@@ -66,15 +79,33 @@ protected OnlineScoringBaseScorer(@NonNull @Config OnlineScoringConfig config,
...
-     * runs at subscription time on the per-stream worker scheduler.
+     * runs at subscription time on the per-stream worker scheduler. On success, records the
+     * per-workspace processed count; failures are attributed via {@link #messageContext(Object)}.
      */
     @Override
     protected Mono<Void> processEvent(M message) {
-        return Mono.defer(() -> score(message));
+        return Mono.defer(() -> score(message))
+                .doOnSuccess(ignored -> processedCounter.add(1,
+                        Attributes.of(WORKSPACE_ID_KEY, message.workspaceId())));
```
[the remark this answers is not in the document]
thiagohora: Good catch — fixed in e246de1. `processEvent` is now a `final` template that wraps an overridable `doScore()` with the processed-success counter, and `OnlineScoringLlmAsJudgeScorer` overrides `doScore` instead of `processEvent`. The `decrementAndFinishIfComplete` post-step is now inside the chain the counter wraps, so `doOnSuccess` (and thus `processed_total`) only fires once the entire per-message chain completes — a failure there leaves the message unacked/retried and is not counted.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSampler.java
```diff
@@ -80,6 +96,36 @@ public OnlineScoringSampler(@NonNull @Config("serviceToggles") ServiceTogglesCon
...
+     * is a single call site rather than three. Always returns {@code false} so callers can
+     * {@code return skip(...)}.
+     */
+    private boolean skip(String workspaceId, AutomationRuleEvaluator<?, ?> evaluator, Trace trace,
+            String decision, String message, Object... args) {
+        recordDecision(workspaceId, evaluator, decision, 1);
+        // Important to set the workspaceId for logging purposes
+        try (var logContext = createTraceLoggingContext(workspaceId, evaluator, trace)) {
+            userFacingLogger.info(message, args);
+        }
```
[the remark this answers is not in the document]
thiagohora: These `userFacingLogger.info(...)` lines are **pre-existing** (this PR only moved the three identical blocks into the shared `skip(...)` helper — the level is unchanged). They go to the **user-facing rule log** (UserFacingLoggingFactory + UserLog MDC markers), not ops stdout — they're intentionally surfaced so rule authors can see *why* a trace wasn't scored (disabled / filtered / sampling rate). Demoting to DEBUG would drop that user-visible signal, which is a behavior change out of scope for this metrics-only PR. Happy to gate/sample these in a dedicated follow-up if user-log volume becomes a concern — `recordDecision(...)` already preserves the metrics regardless.

apps/opik-backend/src/main/java/com/comet/opik/api/events/TracesCreated.java
```diff
@@ -15,10 +15,19 @@
...
+    // for callers that don't carry it; consumers fall back to workspaceId.
+    private final String workspaceName;
 
     public TracesCreated(@NonNull List<Trace> traces, @NonNull String workspaceId, @NonNull String userName) {
+        this(traces, workspaceId, userName, null);
+    }
+
+    public TracesCreated(@NonNull List<Trace> traces, @NonNull String workspaceId, @NonNull String userName,
+            String workspaceName) {
         super(workspaceId, userName);
```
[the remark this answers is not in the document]
thiagohora: Done — marked both the `workspaceName` field and the 4-arg constructor parameter `@Nullable` (with the `jakarta.annotation.Nullable` import) on `TracesCreated` and `TracesUpdated`, making the optional null contract explicit. The 3-arg constructors still pass `null`, and consumers fall back to `workspaceId`.

apps/opik-backend/src/main/java/com/comet/opik/api/events/TracesCreated.java
```diff
@@ -15,10 +16,19 @@
 @Accessors(fluent = true)
 public class TracesCreated extends BaseEvent {
     private final @NonNull List<Trace> traces;
+    // Resolved from RequestContext.WORKSPACE_NAME at publish time (TraceService). May be null/blank
+    // for callers that don't carry it; consumers fall back to workspaceId.
```
andrescrz, line 20: Nit: comments like this better as javadoc.

apps/opik-backend/src/main/java/com/comet/opik/api/events/WorkspaceScopedMessage.java
```diff
@@ -0,0 +1,25 @@
...
+
+    String userName();
+
+    /**
+     * Human-readable workspace name, resolved once at produce time (see {@code OnlineScoringSampler})
+     * and carried on the message so consumers/metrics need no per-message lookup. Defaults to
+     * {@link #workspaceId()} for messages that don't carry a name (older in-flight messages, or
+     * producers that don't resolve it yet).
+     */
+    default String workspaceName() {
```
andrescrz, line 25: Minor: this could be confusing unless you have context or read the Javadoc. Maybe we could signal that the workspace id is sent in some way.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringBaseScorer.java
```diff
@@ -38,10 +44,12 @@
  * persistence. The Reactor pipeline owned by {@link BaseRedisSubscriber} schedules execution on the per-stream
  * worker scheduler; subclasses should NOT call {@code .block()} from {@code score()}.
  */
-public abstract class OnlineScoringBaseScorer<M> extends BaseRedisSubscriber<M> {
+public abstract class OnlineScoringBaseScorer<M extends WorkspaceScopedMessage> extends BaseRedisSubscriber<M> {
```
andrescrz, line 47: Let's double check if any blocker to implement this at the `BaseRedisSubscriber` level. Do we have any non online scorer subscriber that has no `WorkspaceScopedMessage`?

apps/opik-backend/src/main/java/com/comet/opik/api/events/WorkspaceScopedMessage.java
```diff
@@ -0,0 +1,25 @@
+package com.comet.opik.api.events;
+
+/**
+ * Marker for online-scoring stream messages that carry the workspace and user they belong to.
+ *
+ * <p>Implemented by the {@code *ToScore*} message records so {@code OnlineScoringBaseScorer} can
+ * attribute processing-error metrics ({@code online_scoring_*_processing_errors_total}) to the
+ * originating workspace instead of reporting {@code unknown}.
+ */
+public interface WorkspaceScopedMessage {
```
andrescrz, line 10: Minor: Maybe we could consider generalising this to be the `OnlineScoringMessage` or `RedisSubscriberMessage`, instead of just the workspace interface.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringBaseScorer.java
```diff
@@ -66,17 +81,58 @@ protected OnlineScoringBaseScorer(@NonNull @Config OnlineScoringConfig config,
         this.feedbackScoreService = feedbackScoreService;
         this.traceService = traceService;
         this.type = type;
+        this.processedCounter = GlobalOpenTelemetry.getMeter(ONLINE_SCORING_NAMESPACE)
```
andrescrz, line 84: Related to my other comment: if we keep this at online subscriber level, ok to have this metric. If we generalise it for the base subscriber level, better take the most appropriate of the existing metrics and add the workspace dimension.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringBaseScorer.java
```diff
@@ -66,17 +81,58 @@ protected OnlineScoringBaseScorer(@NonNull @Config OnlineScoringConfig config,
...
+     * The counter wraps {@link #doScore(Object)} (which subclasses may extend with post-scoring
+     * work), so a failure anywhere in the chain leaves the message unacked/retried and does NOT
+     * count as processed. Failures are attributed via {@link #messageContext(Object)}.
      */
     @Override
-    protected Mono<Void> processEvent(M message) {
+    protected final Mono<Void> processEvent(M message) {
+        var workspaceName = StringUtils.defaultIfBlank(message.workspaceName(), message.workspaceId());
+        return doScore(message)
+                .doOnSuccess(ignored -> processedCounter.add(1, Attributes.of(
```
andrescrz, line 100: If we keep this metric: I recommend going with a histogram and measure latency as well. That also counts and can also be tagged with the workspace id and name. We'll get counters as well, but the latency additional info could be useful.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringBaseScorer.java
```diff
@@ -66,17 +81,58 @@ protected OnlineScoringBaseScorer(@NonNull @Config OnlineScoringConfig config,
...
+    /**
+     * Attributes processing-error metrics to the workspace/user the message belongs to. Without this
+     * override the base class falls back to {@link MessageContext#UNKNOWN}, which is why
+     * {@code online_scoring_*_processing_errors_total} historically reported {@code workspace_id="unknown"}.
+     * The workspace name is carried on the message (resolved from RequestContext.WORKSPACE_NAME at
+     * trace-event publish time); falls back to the id when absent.
+     */
+    @Override
+    protected MessageContext messageContext(M message) {
+        return new MessageContext(message.workspaceId(),
```
andrescrz: Nit: you can improve this by adding a builder and using it.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSampler.java
```diff
@@ -56,6 +62,17 @@
 @Slf4j
 public class OnlineScoringSampler {
 
+    private static final String ONLINE_SCORING_NAMESPACE = "online_scoring";
```
andrescrz, line 65: You likely want `online_scoring.sampler` as namespace.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSampler.java
```diff
@@ -80,6 +98,39 @@ public OnlineScoringSampler(@NonNull @Config("serviceToggles") ServiceTogglesCon
         this.projectService = projectService;
         secureRandom = SecureRandom.getInstanceStrong();
         userFacingLogger = UserFacingLoggingFactory.getLogger(OnlineScoringSampler.class);
+
+        Meter meter = GlobalOpenTelemetry.getMeter(ONLINE_SCORING_NAMESPACE);
+        this.samplingDecisions = meter.counterBuilder("online_scoring_sampler_decisions_total")
```
andrescrz, line 103: Related to the other comment: if you go with `online_scoring.sampler` prefex, better keep here `decisions_total` or similar.

andrescrz (commented):
Generally in the good direction, a few comments to check before moving forward.

apps/opik-backend/src/main/java/com/comet/opik/api/events/WorkspaceScopedMessage.java
thiagohora: Improved in c817aad — the `workspaceName()` Javadoc now states explicitly that it **falls back to `workspaceId()`** when no name is carried, so a caller may get the id here, and to compare against `workspaceId()` if they need to tell them apart.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringBaseScorer.java
thiagohora: Checked: of the 6 non-online `BaseRedisSubscriber` subclasses, 5 already override `messageContext` (Webhook, DatasetExport, ExperimentAggregates, ExperimentItemProcessing, ClosingTraceThread) — i.e. they already have workspace context — only AgentInsightsReport doesn't. So lifting this to `BaseRedisSubscriber` is feasible, but it means giving every subscriber's message a common workspace-bearing type and is a cross-feature change beyond this PR's scope. Proposing we keep it at the online-scoring level here and do the generalization as a follow-up (happy to file it).

apps/opik-backend/src/main/java/com/comet/opik/api/events/WorkspaceScopedMessage.java
thiagohora: Agreed it's a candidate to generalize (e.g. a `RedisSubscriberMessage` carrying workspace) — that pairs with the base-level question above. I'd prefer to keep the name `WorkspaceScopedMessage` for now since it specifically marks 'carries workspace', and rename/move it as part of the base-subscriber follow-up rather than half-generalize it in this PR.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringBaseScorer.java
thiagohora: Kept at the online-scoring subscriber level for this PR, so the dedicated counter stays. If we lift it to the base subscriber in the follow-up, agree the right move is to add the workspace dimension to the most appropriate existing base metric rather than a new counter — see the histogram note below.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringBaseScorer.java
thiagohora: Good point — note `BaseRedisSubscriber` already emits a `*_processing_time` (native) histogram per stream, which gives both count and latency; it just isn't workspace-tagged today. The cleanest version of your suggestion is to add workspace_id/name as a dimension to that existing histogram (then `processed_total` becomes redundant — histogram_count covers it). That belongs with the base-level generalization since the histogram lives in the base; I'd rather do it there than fork a second histogram at the online layer. Tracking it in the follow-up.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringBaseScorer.java
thiagohora: Done in c817aad — added `@Builder` to `MessageContext` and `messageContext()` now builds it via the builder.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSampler.java
thiagohora: Note on OTel→Prometheus: the meter/scope name isn't a prefix of the exported metric — it becomes the `otel_scope_name` label, and the exported Prometheus name equals the instrument name. The sibling online-scoring metrics (in `BaseRedisSubscriber`) all use scope `online_scoring` with fully-qualified instrument names (`online_scoring_<scorer>_...`), so I kept the sampler consistent: scope `online_scoring`, instrument `online_scoring_sampler_decisions_total`. I can switch the scope to `online_scoring.sampler` if you want the scope label to distinguish the producer, but that wouldn't change the metric name.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSampler.java
thiagohora: Tied to the above: since the exported name = instrument name (scope is a label, not a prefix), naming the instrument just `decisions_total` would export it as `decisions_total` — losing the `online_scoring_` family prefix the other metrics and the Grafana dashboard rely on. So I kept the fully-qualified `online_scoring_sampler_decisions_total` (the `sampler` segment already scopes it). Open to revisiting if we standardize scopes across all these meters.

andrescrz (approved):
Thanks for double checking what should go here and what on in a potential follow-up. In the meantime, this LGTM.

PR #7212, opened 2026-06-22
[NA] [BE] feat: accept all-clear reports + track last scan time (agent insights)

aadereiko:
## Details

Two related backend fixes for Agent Insights diagnostics, both touching `reportIssues`:

- **Accept "all clear" reports.** The report endpoint rejected an empty `issues` list with 422 `"issues must not be empty"`, so a run that finds nothing couldn't record completion — the reporting agent treated the 422 as retryable and looped until it exhausted its token budget. `issues` is now `@NotNull` (a null body is still rejected; an empty list is a valid "all clear") and `reportIssues` skips the issue writes when empty.
- **Track last scan time.** Added `last_scan_at` on the job, stamped on every report (including all-clear) in the same transaction as the issue upserts. The UI uses it for an accurate "Last scan" time that — unlike an issue's `last_updated_at` — is not affected by users resolving/reopening issues.

Detail:
- Migration `000086`: add `last_scan_at` to `agent_insights_jobs` (anchored `AFTER status`).
- `AgentInsightsReport`: `issues` `@NotEmpty` → `@NotNull`.
- `AgentInsightsJob`: expose read-only `last_scan_at`.
- `AgentInsightsJobDAO.markScanned`: stamp `last_scan_at` + `last_updated_by` (no-op if the project has no job row — unreachable via the trigger flow, which 404s without a job).
- `AgentInsightsIssueService.reportIssues`: skip writes when all-clear; always stamp the scan time.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- No associated ticket (`[NA]`).

## AI-WATERMARK

AI-WATERMARK: yes

- If yes:
  - Tools: Claude Code
  - Model(s): Claude Opus 4.8
  - Scope: Implemented the validation/service/DAO/model changes and migration; addressed automated review feedback; authored this description.
  - Human verification: Rebuilt and restarted the local backend, applied migration `000086`; confirmed an empty report returns 204 (was 422) and the job endpoint returns `last_scan_at` set to the report time, unchanged by resolving an issue.

## Testing

- `mvn -o -DskipTests test-compile` (main + tests compile).
- Tests: added `reportIssuesWhenEmptyThenAcceptedAndNothingPersisted` (empty report → 204, persists nothing); removed the now-invalid "empty issues" case from `invalidReports()`.
- Manual (local, process mode): rebuilt the jar, applied migration `000086`, restarted opik-backend. `POST /v1/private/agent-insights/issues` with `"issues":[]` → **204** (was **422**); `GET /v1/private/agent-insights/jobs/{projectId}` returns `last_scan_at` = report time; resolving/reopening an issue leaves `last_scan_at` unchanged.
- Not run: full `mvn verify` integration suite (testcontainers).

## Documentation

No documentation changes required.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsIssueService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsIssueService.java
@@ -73,18 +73,14 @@ class AgentInsightsIssueServiceImpl implements AgentInsightsIssueService {
 
         projectService.get(report.projectId(), workspaceId);
 
-        // An "all clear" report (no issues detected) is a valid outcome: validate
-        // the project, then return without touching the DB. Skipping the batch
-        // writes also avoids empty-batch failures in the DAO.
-        if (report.issues().isEmpty()) {
-            log.info(
-                    "No agent insights issues to store for project '{}' on report day '{}' in workspace '{}' (all clear)",
-                    report.projectId(), report.reportDay(), workspaceId);
-            return;
-        }
+        // An "all clear" report (no issues detected) is a valid outcome: skip the
+        // issue writes (also avoids empty-batch failures in the DAO) but still
+        // stamp the scan time below so the UI shows when the run last completed.
+        boolean allClear = report.issues().isEmpty();
 
-        log.info("Storing '{}' agent insights issues for project '{}' on report day '{}' in workspace '{}'",
-                report.issues().size(), report.projectId(), report.reportDay(), workspaceId);
+        log.info("{} agent insights issues for project '{}' on report day '{}' in workspace '{}'",
+                allClear ? "No (all clear)" : report.issues().size(), report.projectId(), report.reportDay(),
+                workspaceId);
 
         List<UUID> issueIds = report.issues().stream()
                 .map(issue -> issue.id() != null ? issue.id() : idGenerator.generateId())
@@ -99,9 +95,14 @@ class AgentInsightsIssueServiceImpl implements AgentInsightsIssueService {
         transactionTemplate.inTransaction(WRITE, handle -> {
             AgentInsightsIssueDAO dao = handle.attach(AgentInsightsIssueDAO.class);
 
-            dao.upsertIssues(workspaceId, report.projectId(), userName, issueIds, report.issues());
-            dao.upsertDetails(workspaceId, report.projectId(), report.reportDay(), userName,
-                    detailIds, issueIds, report.issues(), metadata);
+            if (!allClear) {
+                dao.upsertIssues(workspaceId, report.projectId(), userName, issueIds, report.issues());
+                dao.upsertDetails(workspaceId, report.projectId(), report.reportDay(), userName,
+                        detailIds, issueIds, report.issues(), metadata);
+            }
+
+            // Record when this project was last scanned (no-op if it has no job row).
+            handle.attach(AgentInsightsJobDAO.class).markScanned(workspaceId, report.projectId());
 
             return null;
         });
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsIssueService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsIssueService.java
@@ -73,6 +73,16 @@ class AgentInsightsIssueServiceImpl implements AgentInsightsIssueService {
 
         projectService.get(report.projectId(), workspaceId);
 
+        // An "all clear" report (no issues detected) is a valid outcome: validate
+        // the project, then return without touching the DB. Skipping the batch
+        // writes also avoids empty-batch failures in the DAO.
+        if (report.issues().isEmpty()) {
+            log.info(
+                    "No agent insights issues to store for project '{}' on report day '{}' in workspace '{}' (all clear)",
+                    report.projectId(), report.reportDay(), workspaceId);
+            return;
+        }
+
         log.info("Storing '{}' agent insights issues for project '{}' on report day '{}' in workspace '{}'",
                 report.issues().size(), report.projectId(), report.reportDay(), workspaceId);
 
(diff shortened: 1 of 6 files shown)


apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsIssueService.java
```diff
@@ -99,9 +95,14 @@ public void reportIssues(@NonNull AgentInsightsReport report) {
...
-            dao.upsertDetails(workspaceId, report.projectId(), report.reportDay(), userName,
-                    detailIds, issueIds, report.issues(), metadata);
+            if (!allClear) {
+                dao.upsertIssues(workspaceId, report.projectId(), userName, issueIds, report.issues());
+                dao.upsertDetails(workspaceId, report.projectId(), report.reportDay(), userName,
+                        detailIds, issueIds, report.issues(), metadata);
+            }
+
+            // Record when this project was last scanned (no-op if it has no job row).
+            handle.attach(AgentInsightsJobDAO.class).markScanned(workspaceId, report.projectId());
```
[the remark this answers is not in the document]
aadereiko: Intentionally left as a defensive no-op rather than upserting the job row here. A report is only ever produced after a run, and a run can only be started for an existing job: the manual trigger (`AgentInsightsJobService.triggerNow`) throws 404 if there's no job, and the daily sweep only triggers `status = 'enabled'` jobs. So when `reportIssues` runs, the job row always exists and `markScanned` updates exactly one row. Upserting from the report path would also blur ownership of job creation (currently the dedicated `POST /jobs/{projectId}` endpoint), so I'd prefer to keep that responsibility there.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000086_add_last_scan_at_to_agent_insights_jobs.sql
```diff
@@ -0,0 +1,8 @@
+--liquibase formatted sql
+--changeset aadereiko:000086_add_last_scan_at_to_agent_insights_jobs
+--comment: Track when a diagnostic report was last generated for a project (stamped on every report, including "all clear"), so the UI can show an accurate "Last scan" time that is unaffected by user-driven status changes.
+
+ALTER TABLE agent_insights_jobs
+    ADD COLUMN last_scan_at TIMESTAMP(6) NULL DEFAULT NULL AFTER last_triggered_at;
+
+--rollback ALTER TABLE agent_insights_jobs DROP COLUMN last_scan_at;
```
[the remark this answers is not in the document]
aadereiko: Fixed in 9939edda — added the trailing blank line at EOF per the `migrations.md` convention.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000086_add_last_scan_at_to_agent_insights_jobs.sql
```diff
@@ -0,0 +1,9 @@
+--liquibase formatted sql
+--changeset aadereiko:000086_add_last_scan_at_to_agent_insights_jobs
+--comment: Add last_scan_at to agent_insights_jobs — time a diagnostic report was last generated
+
+ALTER TABLE agent_insights_jobs
+    ADD COLUMN last_scan_at TIMESTAMP(6) NULL DEFAULT NULL AFTER status;
```
BorisTkachenko, line 6: Probably we don't need this, you can use `last_updated_at` for the same purpose.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsIssueService.java
```diff
@@ -89,9 +92,14 @@ public void reportIssues(@NonNull AgentInsightsReport report) {
...
-            dao.upsertDetails(workspaceId, report.projectId(), report.reportDay(), userName,
-                    detailIds, issueIds, report.issues(), metadata);
+            if (!allClear) {
+                dao.upsertIssues(workspaceId, report.projectId(), userName, issueIds, report.issues());
+                dao.upsertDetails(workspaceId, report.projectId(), report.reportDay(), userName,
+                        detailIds, issueIds, report.issues(), metadata);
+            }
+
+            handle.attach(AgentInsightsJobDAO.class)
+                    .markScanned(workspaceId, report.projectId(), userName);
```
BorisTkachenko, line 102: This is probably not what we want. If there are any issues found, then we insert those. If there are no issues, we just record that it was scanned. But effectively, it's an update. What if during the first run we have no issues? Then there is nothing to update. And the scan is basically silently dropped.

BorisTkachenko (changes requested):

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000086_add_last_scan_at_to_agent_insights_jobs.sql
aadereiko: I went with a dedicated column on purpose. The job's `last_updated_at` is also bumped by enable/disable — `updateStatus` sets `status`/`last_updated_by`, which triggers `ON UPDATE last_updated_at` — so reusing it would make "Last scan" jump whenever someone toggles auto-diagnostics even though no scan ran. `last_scan_at` is stamped only from `reportIssues`, so it reflects an actual scan precisely. (Resolving/reopening issues affects neither, since both live on the job row, not the issue.)

Happy to drop the column and reuse `last_updated_at` if you consider that enable/disable imprecision acceptable — just wanted the reasoning on record.

apps/opik-backend/src/main/java/com/comet/opik/domain/AgentInsightsIssueService.java
aadereiko: Correcting myself on one detail above — the job isn't necessarily created "when enabled"; it's created lazily on first trigger. But the conclusion stands:

A run can only be **enqueued** for an existing job. `AgentInsightsJobService.triggerNow` does `findByProject(...).orElseThrow(NotFoundException)` (404s without a job), and the daily sweep only enqueues `status = 'enabled'` jobs. The FE enforces the same from its side — the trigger mutation does `POST /trigger`, and on a 404 it creates the job (`POST /jobs/{projectId}`) and retries — so the job row exists before the run is enqueued.

So by the time `reportIssues` runs, the job exists and `markScanned` updates exactly one row, including the first all-clear run; the scan isn't dropped. The only window where it would no-op is if the job/project were deleted mid-run — the guard covers that, rather than upserting from the report path (job creation stays owned by `POST /jobs/{projectId}`).

BorisTkachenko (approved):

PR #7215, opened 2026-06-23
[OPIK-6830] [BE] fix: restart pods with a dead outbound HTTP connection pool via liveness health check

BorisTkachenko:
## Details

The shared outbound Jersey/Apache `Client` is used by every authenticated request through `RemoteAuthService`. If its connection pool is ever shut down (e.g. an aborted graceful shutdown or a process that survives an OOM), the pod keeps serving but fails **every** auth call with `IllegalStateException: Connection pool shut down`. A shut-down pool is unrecoverable without a process restart, so simply removing the pod from the ready rotation isn't enough.

This PR adds an `AuthHttpClientHealthCheck` wired to **liveness** (`type: alive`) so Kubernetes restarts a poisoned pod and recreates the pool:

- Probes the shared client with a short-timeout `HEAD` request to the React/auth service. Any reachable response — including 4xx/5xx — proves the pool can lease a connection and is treated as healthy.
- Reports **unhealthy only** on the local `Connection pool shut down` condition (matched case-insensitively across the full cause chain). Upstream/transient failures (timeouts, connection refused, DNS, HTTP errors) are deliberately ignored so an auth-service blip cannot restart-storm the fleet.
- No-ops (healthy) when authentication is disabled or no React service URL is configured.

Registered the check in `config.yml` (and `config-test.yml`) as a critical liveness check.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues
- OPIK-6830

## Testing

- Added `AuthHttpClientHealthCheckTest`, a parameterized test covering the cause-chain matcher: direct `IllegalStateException`, wrapped in `ProcessingException`, doubly-wrapped, case-insensitive match, and the negative cases (other `IllegalStateException` messages, socket timeout, connection refused, null message, null throwable) to guarantee only a genuinely dead pool flags unhealthy.
- New liveness check added to `config-test.yml` so it runs in the existing app-context/health integration suite.

## Documentation

N/A — internal infrastructure health check; no public API or user-facing documentation changes.


apps/opik-backend/src/test/java/com/comet/opik/infrastructure/auth/AuthHttpClientHealthCheckTest.java
```diff
@@ -0,0 +1,35 @@
...
+                Arguments.of(new IllegalStateException("some other illegal state"), false),
+                Arguments.of(new ProcessingException(new SocketTimeoutException("Read timed out")), false),
+                Arguments.of(new ProcessingException(new IOException("Connection refused")), false),
+                Arguments.of(new IllegalStateException((String) null), false),
+                Arguments.of(null, false));
+    }
+
+    @ParameterizedTest
+    @MethodSource("exceptions")
+    void isConnectionPoolShutDown__detectsOnlyDeadPool(Throwable throwable, boolean expected) {
```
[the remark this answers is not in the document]
BorisTkachenko: Skipping — the `<methodName>__<scenario>` double-underscore pattern is the dominant convention across this test suite (e.g. `AuthServiceImplTest.testAuthenticate__whenWorkspaceIsDefault__thenUseDefaultUser`) and is used in `.agents/skills/opik-backend/testing.md`'s own run example (`getTracesByProject__whenFilterSortExcludeAcrossPages`). The doc's "Bad" examples are full snake_case (`should_create_user`), not the method-name + scenario separator. Keeping the name consistent with surrounding code.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthHttpClientHealthCheck.java
```diff
@@ -0,0 +1,86 @@
...
+            // the fleet during an auth-service blip. Only a locally shut-down pool is fatal and restart-worthy.
+            return Result.healthy("Shared HTTP client pool is alive; upstream probe error ignored: %s"
+                    .formatted(exception.getMessage()));
+        }
+    }
+
+    static boolean isConnectionPoolShutDown(Throwable throwable) {
+        for (Throwable cause = throwable; cause != null && cause != cause.getCause(); cause = cause.getCause()) {
+            if (cause instanceof IllegalStateException && cause.getMessage() != null
+                    && cause.getMessage().toLowerCase(Locale.ROOT).contains(DEAD_POOL_MARKER)) {
```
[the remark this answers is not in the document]
BorisTkachenko: Fixed in a243f29cea — `isConnectionPoolShutDown` now matches `ConnectionShutdownException` by type anywhere in the cause chain (it extends `IllegalStateException` with a null message, so the message-only check missed it and a dead shared pool stayed green). The case-insensitive "connection pool shut down" message match is kept as a fallback for the pool-level `IllegalStateException` thrown by `StrictConnPool.lease()`. Added test cases for the bare and `ProcessingException`-wrapped `ConnectionShutdownException`.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthHttpClientHealthCheck.java
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * Detects a shut-down connection pool on the shared outbound {@link Client}. The shared client is used by every
+ * authenticated request via {@code RemoteAuthService}, so once its Apache pool is closed (e.g. an aborted graceful
+ * shutdown or an OOM survivor) the pod keeps serving but fails every auth call with
+ * {@code IllegalStateException: Connection pool shut down}. Wired to liveness so K8s restarts the poisoned pod and
+ * recreates the pool, instead of leaving it {@code Ready} and poisoning its share of traffic.
+ */
+@Singleton
+@Slf4j
+public class AuthHttpClientHealthCheck extends NamedHealthCheck {
```
andrescrz: Minor: The HttpClient is shared, but you need some endpoint (auth in this case) to ping. So this name is tied to a particular endpoint, which somewhat makes sense, but on the other hand doesn't reflect the nature of the shared client.

You clarify all this in the Javadoc already, which is great. But you optionally could rename this to something more significative that also reflects the nature of the shared HTTP client.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthHttpClientHealthCheck.java
```diff
@@ -0,0 +1,93 @@
...
+
+    public static final String NAME = "auth_http_client";
```
andrescrz: Nit: related to the other, maybe `auth_shared_http_client`.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthHttpClientHealthCheck.java
```diff
@@ -0,0 +1,93 @@
...
+
+    // The lease() failure on a closed pool surfaces as this message; matched case-insensitively across the cause chain.
```
andrescrz: Nit: these comments better as javadocs.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthHttpClientHealthCheck.java
```diff
@@ -0,0 +1,93 @@
...
+    private static final String DEAD_POOL_MARKER = "connection pool shut down";
+    private static final int PROBE_TIMEOUT_MILLIS = 2_000;
```
andrescrz: Timeouts in health checks should be configurable. See others like the ClickHouse ones. 

Maybe reduce the default value to `1s`.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthHttpClientHealthCheck.java
```diff
@@ -0,0 +1,93 @@
...
+
+    private final Client client;
+    private final boolean enabled;
+    private final URI reactServiceUri;
+
+    @Inject
+    public AuthHttpClientHealthCheck(@NonNull Client client, @NonNull OpikConfiguration configuration) {
```
andrescrz: Nit: you can probably narrow down the injected configuration to `AuthenticationConfig` with `@Named`.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthHttpClientHealthCheck.java
```diff
@@ -0,0 +1,93 @@
...
+        this.client = client;
+        AuthenticationConfig authConfig = configuration.getAuthentication();
+        this.enabled = authConfig.isEnabled() && authConfig.getReactService() != null
```
andrescrz: Minor: feel free to tighten these conditions directly in `AuthenticationConfig`, so you don't need to check if null or empty here. That should be backwards compatible change per the config YAML files and desirable for any env var override.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthSharedHttpClientHealthCheck.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    @Override
+    public String getName() {
+        return NAME;
+    }
+
+    @Override
+    protected Result check() {
+        if (!enabled) {
```
andrescrz: This is the big question about this endpoint: as it's tied to the auth endpoint, it won't detect issues on the shared client (used for many other purposes such as Python Online Scoring) in installations without auth: that's all the open source installations.

I propose investigating if this can be decoupled. You might be able to ping something, maybe some dumb endpoint that's is always available, even in localhost if it routes traffic and triggers the connection pool shutdown exception.

We could discard this approach if too complicated or risky and simply document the limitation: this won't detect HTTP client issues if no auth.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthHttpClientHealthCheck.java
```diff
@@ -0,0 +1,93 @@
...
+            }
+            // Upstream/transient failures (timeouts, connection refused, DNS, HTTP errors) must NOT flap liveness across
+            // the fleet during an auth-service blip. Only a locally shut-down pool is fatal and restart-worthy.
+            return Result.healthy("Shared HTTP client pool is alive; upstream probe error ignored: %s"
+                    .formatted(exception.getMessage()));
+        }
+    }
+
+    static boolean isConnectionPoolShutDown(Throwable throwable) {
+        for (Throwable cause = throwable; cause != null && cause != cause.getCause(); cause = cause.getCause()) {
```
andrescrz: As discussed in past reviews, please use Guava's `Throwables` class methods to:
1. Avoid reinventing the wheel.
2. Manage infinite loops on causes.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/auth/AuthHttpClientHealthCheckTest.java
```diff
@@ -0,0 +1,38 @@
...
+        assertThat(AuthHttpClientHealthCheck.isConnectionPoolShutDown(throwable)).isEqualTo(expected);
```
andrescrz: This should be a general unit test at the public entry point `check`, by using proper mocks, instead of directly hitting `isConnectionPoolShutDown`, which should become a private method.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/HealthCheckIntegrationTest.java
andrescrz, line 1: This test class should add another inner class to test the auth being disabled case.

apps/opik-backend/config.yml
```diff
@@ -143,6 +143,13 @@ health:
     - name: mysql
       critical: true
       type: ready
+    # Gates liveness (type: alive) so a pod whose shared outbound HTTP pool has been shut down gets restarted and
+    # recreates the pool. A shut-down pool is unrecoverable without a restart, so removing it from rotation isn't
+    # enough. Fails only on the local "Connection pool shut down" condition, never on upstream auth errors, so an
+    # auth-service blip cannot restart-storm the fleet.
+    - name: auth_http_client
```
andrescrz: Nit: move this config above after the other `alive` health check.

Align the test config order if you address this.

andrescrz (commented):
There are many minor comments, but a few important to check before moving forward.

To me the main question is the limitation of this endpoint: only works when auth is enabled, but the HTTP client is shared by other features requiring outbound HTTP calls (like Python Online Evaluation).

I recommend exploring a solution for that. If too complicated or not a good approach, we could go with this.

Regardless of this, less address other non-minor comments.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthHttpClientHealthCheck.java
BorisTkachenko: Fixed in 9468ee8985 — renamed the class to `AuthSharedHttpClientHealthCheck` to reflect the shared outbound client rather than the auth endpoint it happens to probe.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthHttpClientHealthCheck.java
BorisTkachenko: Fixed in 9468ee8985 — renamed the check to `auth_shared_http_client` (propagated to config.yml, config-test.yml, and the integration test).

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthHttpClientHealthCheck.java
BorisTkachenko: Fixed in 9468ee8985 — converted the field comment to Javadoc.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthHttpClientHealthCheck.java
BorisTkachenko: Fixed in 9468ee8985 — the probe timeout is now configurable via `AuthenticationConfig.healthCheckTimeout` (env `AUTH_HEALTH_CHECK_TIMEOUT`) and the default was reduced to 1s, matching the ClickHouse/Redis health checks.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthHttpClientHealthCheck.java
BorisTkachenko: Fixed in 9468ee8985 — narrowed the injection to `@Config("authentication") AuthenticationConfig`. Kept the explicit `@Inject` constructor so the qualifier is retained.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthHttpClientHealthCheck.java
BorisTkachenko: Fixed in 9468ee8985 — moved the null/blank check into `AuthenticationConfig.isReactServiceConfigured()`, so the health check no longer performs it here.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthHttpClientHealthCheck.java
BorisTkachenko: Fixed in 9468ee8985 — now uses `Throwables.getCausalChain(...).stream().anyMatch(...)`, which handles cyclic causes internally.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/auth/AuthHttpClientHealthCheckTest.java
BorisTkachenko: Fixed in 9468ee8985 — the test now drives the public `check()` through a mocked Jersey `Client` (dead pool to unhealthy, transient/upstream to healthy, healthy response closed), and `isConnectionPoolShutDown` is now private.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/HealthCheckIntegrationTest.java
BorisTkachenko: Fixed in 9468ee8985 — added an `AuthenticationEnabled` nested class. It enables auth with an unreachable React URL so the probe actually fires, asserting the check stays healthy on an unreachable upstream (no liveness flap). The auth-disabled inert state is already covered by the `DefaultConfig` aggregate.

apps/opik-backend/config.yml
BorisTkachenko: Fixed in 9468ee8985 — moved the check up next to `deadlocks` (the other `alive` check) and aligned the order in config-test.yml.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthSharedHttpClientHealthCheck.java
BorisTkachenko: Keeping the current auth-scoped approach by design, rather than decoupling.

The dead-pool restart is only actionable where a liveness actuator restarts the pod (Kubernetes liveness). Deployments orchestrated that way run with authentication enabled, so the check fires and delivers value exactly there. Installations without authentication run under docker-compose, where the health endpoint maps to a Docker HEALTHCHECK that only flips container status to unhealthy and never restarts anything — so even a fully working dead-pool detector there would emit a signal that nothing can act on.

Decoupling the probe from auth would add complexity and a probe-target ambiguity (what to ping when no React URL is configured) purely to raise an alarm with no actuator behind it. The check no-ops as healthy when auth is disabled, which is the intended behavior. Documented the rationale here rather than in code.

BorisTkachenko:
@andrescrz thanks for the review — all comments addressed in 9468ee8985.

On the main question (auth-scoped vs. shared-client coverage): keeping it auth-scoped by design. The dead-pool restart is only actionable where a liveness actuator restarts the pod (Kubernetes liveness), and those deployments run with authentication enabled. Installations without auth run under docker-compose, where the health endpoint maps to a Docker HEALTHCHECK that only marks the container unhealthy and never restarts it — so a decoupled detector there would raise an alarm nothing can act on. Decoupling would add complexity and probe-target ambiguity for no actuator. Detail in the line thread.

Other changes: configurable probe timeout (default 1s) on `AuthenticationConfig`, Guava `Throwables.getCausalChain` for the cause walk, `@Config("authentication")` injection, null/blank check moved into config, `check()` unit-tested via mocks with the matcher made private, an auth-enabled integration case asserting an unreachable upstream stays healthy, the class/check renamed to `AuthSharedHttpClientHealthCheck` / `auth_shared_http_client`, and the alive checks grouped together in config.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/AuthenticationConfig.java
```diff
@@ -19,4 +21,11 @@ public record UrlConfig(@Valid @JsonProperty @NotNull String url) {
 
     @Valid @JsonProperty
     private UrlConfig reactService;
+
+    @Valid @JsonProperty
+    @NotNull private Duration healthCheckTimeout = Duration.seconds(1);
```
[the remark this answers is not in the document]
BorisTkachenko: Fixed in 079a82bfc0 — the timeout is clamped to the int-millis range with `(int) Math.min(..., Integer.MAX_VALUE)` instead of `Math.toIntExact`, so an oversized configured value can no longer overflow at startup. (The `int` cast itself is unavoidable — Jerseys CONNECT_TIMEOUT/READ_TIMEOUT are int millis.)

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthSharedHttpClientHealthCheck.java
```diff
@@ -0,0 +1,91 @@
...
+    private final URI reactServiceUri;
+    private final int probeTimeoutMillis;
+
+    @Inject
+    public AuthSharedHttpClientHealthCheck(@NonNull Client client,
+            @NonNull @Config("authentication") AuthenticationConfig authConfig) {
+        this.client = client;
+        this.enabled = authConfig.isEnabled() && authConfig.isReactServiceConfigured();
+        this.reactServiceUri = enabled ? URI.create(authConfig.getReactService().url()) : null;
+        this.probeTimeoutMillis = Math.toIntExact(authConfig.getHealthCheckTimeout().toMilliseconds());
```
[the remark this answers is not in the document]
BorisTkachenko: Fixed in 079a82bfc0 — addressed by construction. Dead-pool detection is target-independent (the Apache pool throws on `lease()`, before any DNS/connect/headers, when shut down), so the probe does not need the real auth service. It now probes a fixed loopback URI parsed once as a constant, instead of the configured `reactService` URL. That removes the eager `URI.create` on user input entirely — no malformed-URL boot crash, and no per-check allocation. A malformed configured URL is no longer reachable by this code path.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/http/SharedHttpClientHealthCheck.java
```diff
@@ -0,0 +1,99 @@
...
+    private static boolean isConnectionPoolShutDown(Throwable throwable) {
+        // ConnectionShutdownException is the direct httpclient5 shutdown signal but carries a null message, so it must
+        // be matched by type. The message match stays as a fallback for the pool-level
+        // IllegalStateException("Connection pool shut down") thrown by StrictConnPool.lease(). Throwables.getCausalChain
+        // walks the full cause chain and guards against cyclic causes.
+        return Throwables.getCausalChain(throwable).stream()
+                .anyMatch(cause -> cause instanceof ConnectionShutdownException
+                        || (cause instanceof IllegalStateException && cause.getMessage() != null
+                                && cause.getMessage().toLowerCase(Locale.ROOT).contains(DEAD_POOL_MARKER)));
+    }
```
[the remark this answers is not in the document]
BorisTkachenko: No need.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthSharedHttpClientHealthCheck.java
andrescrz: I don't think the premise that having Kubernetes enabled means having auth enable is true. That's generally true for Comet comertial installations at the moment, but not always the case (testing etc.). 

In addition, it's not the case for any self-hosted open source users. You can actually see that the Helm chart doesn't override the default `false` value of `AUTH_ENABLED`.

The moment this uses a `http://localhost:1`I don't see a reason to not to enable in any case, or maybe keeping an optional flag decoupled from the auth one.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthSharedHttpClientHealthCheck.java
```diff
@@ -0,0 +1,99 @@
...
+    private final int probeTimeoutMillis;
+
+    @Inject
+    public AuthSharedHttpClientHealthCheck(@NonNull Client client,
+            @NonNull @Config("authentication") AuthenticationConfig authConfig) {
+        this.client = client;
+        this.enabled = authConfig.isEnabled() && authConfig.isReactServiceConfigured();
+        // Jersey timeouts are int millis; clamp so an absurdly large configured Duration cannot overflow at startup.
+        this.probeTimeoutMillis = (int) Math.min(authConfig.getHealthCheckTimeout().toMilliseconds(),
+                Integer.MAX_VALUE);
```
andrescrz: Minor: you can just use `MaxDuration` (and `MinDuration`) on the config field, instead of having custom logic here.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/AuthenticationConfig.java
andrescrz: Let's not have defaults in the Java code, otherwise we need to keep them in sync with the config yaml files. This is annotated as NotNull, so you can safely remove them.

andrescrz (commented):
Thanks for addressing the majority of the feedback.

I left some other comments.

My main question is about the following:

I'm glad you managed to de-couple the health check from any particular endpoint, including auth, that's a great improvement. However, that still reinforces my idea of not coupling it against the auth configuration. I left some thoughts about that. Another reason, it's what I mentioned yesterday: this is a shared client for many features requiring outbound HTTP calls, not only for Auth.

Let's quickly discuss and reach an agreement about that.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthSharedHttpClientHealthCheck.java
BorisTkachenko: Good point — you are right, and I decoupled it in f5d2747a45.

The premise that K8s implies auth was wrong (the Helm chart ships `AUTH_ENABLED=false`), and since the probe targets a fixed loopback URI the detection never needed auth at all. The check is now always-on and no longer auth-scoped, so it protects the shared client everywhere a liveness actuator exists — including self-hosted open-source K8s — and for all of its callers (online scoring, etc.), not just auth.

Concretely: renamed to `SharedHttpClientHealthCheck` (`shared_http_client`) and moved to `infrastructure.http` next to the module that provides the shared `Client`; removed the auth gate; injects only the `Client` and its probe timeout. Went with always-on rather than a flag — the probe is cheap and harmless even where nothing acts on it.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/AuthSharedHttpClientHealthCheck.java
BorisTkachenko: Done in f5d2747a45 — the timeout now lives on the dedicated `SharedHttpClientHealthCheckConfig` with `@NotNull @MinDuration(100ms) @MaxDuration(60s)`. The `@MaxDuration` bound makes the int-millis conversion overflow-proof, so the custom clamp logic is gone.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/AuthenticationConfig.java
BorisTkachenko: Done in f5d2747a45 — removed the Java default. The field is now `@NotNull` with no default, and the value is supplied by both config.yml (`${SHARED_HTTP_CLIENT_HEALTH_CHECK_TIMEOUT:-1s}`) and config-test.yml. Note the field also moved off `AuthenticationConfig` onto the new `SharedHttpClientHealthCheckConfig` as part of decoupling the check from auth.

andrescrz (approved):
LGTM. Thanks for putting this together.

PR #7219, opened 2026-06-23
[OPIK-6894] [BE] feat: add id_week predicate to TraceDAO retention paths

andrescrz:
## Details

Add an id-derived weekly-bucket predicate (`toMonday(id_at)`) alongside the existing id-range filter on `TraceDAO`'s retention SQL paths — the unbounded delete, the bounded (per-workspace) delete, and the pre-delete count — plus migration `000091` introducing an `id_at` materialized column on `traces`. The predicate is a strict consequence of the id-range, so it never changes which rows are deleted today; once `traces` is weekly-partitioned (OPIK-6900) it lets the retention sweep prune to the partitions in range, which is load-bearing for OPIK-5295 storage reclaim. Retention is gated behind `retention.enabled=false`, so there is no production behavior change today.

- `id_at DateTime('UTC') MATERIALIZED UUIDv7ToDateTime(toUUID(id))` — UTC matches the table's other temporal columns and keeps week boundaries server-timezone independent; seconds suffice for week bucketing; metadata-only `ADD COLUMN`.
- Bounds are derived in SQL from the existing cutoff / lower-bound binds; the upper bound advances one week (`addWeeks(..., 1)`) so rows sharing the cutoff's week stay in scope. The bounded path's floor uses the global `min_lower_bound`, keeping it a strict consequence across the whole batch.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves OPIK-6894

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: full implementation, iterative review
- Human verification: code review + integration tests against real ClickHouse

## Testing

- `mvn test -Dtest=RetentionPolicyServiceTest` (apps/opik-backend) — 17/17 green against real ClickHouse (testcontainers).
- Coverage: both retention delete variants (unbounded + bounded) and the count path validate the predicate as a strict-consequence no-op (old data deleted, recent kept); a parameterized week-boundary test (Tuesday–Sunday cutoffs) verifies the `addWeeks(toMonday(cutoff), 1)` ceiling keeps rows sharing the cutoff's week in scope.
- Repository-wide quality checks (compile, spotless, full suite) are left to CI.

## Documentation

N/A — no documentation changes; rationale is captured in code Javadoc and the migration comment.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
@@ -1742,11 +1742,22 @@ class TraceDAOImpl implements TraceDAO {
             ;
             """;
 
+    /**
+     * Retention sweep for the applyToPast=true window {@code [lower_bound, cutoff_id)}.
+     * <p>
+     * {@code toMonday(id_at)} is the future weekly partition expression ({@code id_at} is MATERIALIZED from
+     * the UUIDv7 id as UTC). Bounding it to the cutoff's week range never excludes a row the id-range would
+     * delete, so it does not change which rows are deleted; once {@code traces} is partitioned (OPIK-6900) it
+     * lets the sweep prune to the partitions in range. The bounds use UTC to match {@code id_at}, and the
+     * upper bound advances one week so rows sharing the cutoff's week stay in scope.
+     */
     private static final String DELETE_FOR_RETENTION = """
             DELETE FROM traces
             WHERE workspace_id IN :workspace_ids
             AND id >= :lower_bound
             AND id \\< :cutoff_id
+            AND toMonday(id_at) >= toMonday(UUIDv7ToDateTime(toUUID(:lower_bound), 'UTC'))
+            AND toMonday(id_at) \\< addWeeks(toMonday(UUIDv7ToDateTime(toUUID(:cutoff_id), 'UTC')), 1)
             AND id NOT IN (
                 SELECT trace_id FROM experiment_items
                 WHERE workspace_id IN :workspace_ids
@@ -1757,14 +1768,20 @@ class TraceDAOImpl implements TraceDAO {
             ;
             """;
 
-    // Lightweight pre-delete count for observability. Omits the experiment_items exclusion subquery
-    // to avoid the join cost; this makes it an upper-bound ceiling with >99% precision in practice
-    // (very few traces are linked to experiments).
+    /**
+     * Lightweight pre-delete count for observability. Omits the {@code experiment_items} exclusion subquery
+     * to avoid the join cost, making it an upper-bound ceiling with &gt;99% precision in practice (very few
+     * traces are linked to experiments). Carries the same {@code toMonday(id_at)} week bounds as
+     * {@code DELETE_FOR_RETENTION} so the count prunes to the same partitions post-cutover rather than
+     * scanning (and loading cold-tier marks for) every partition each cycle.
+     */
     private static final String COUNT_FOR_RETENTION = """
             SELECT count() FROM traces
             WHERE workspace_id IN :workspace_ids
             AND id >= :lower_bound
             AND id \\< :cutoff_id
+            AND toMonday(id_at) >= toMonday(UUIDv7ToDateTime(toUUID(:lower_bound), 'UTC'))
+            AND toMonday(id_at) \\< addWeeks(toMonday(UUIDv7ToDateTime(toUUID(:cutoff_id), 'UTC')), 1)
             SETTINGS log_comment = '<log_comment>'
             ;
             """;
@@ -4499,7 +4516,12 @@ class TraceDAOImpl implements TraceDAO {
                     .append(" AND id >= :lb_").append(i)
                     .append(" AND id < :cutoff_id)");
         }
-        sb.append(") AND id NOT IN (")
+        // toMonday(id_at) week bounds, the bounded counterpart of DELETE_FOR_RETENTION. The single floor
+        // uses the global :min_lower_bound, which is <= every per-workspace :lb_i, so it never excludes a row
+        // that any per-workspace id-range would delete. UTC matches id_at.
+        sb.append(") AND toMonday(id_at) >= toMonday(UUIDv7ToDateTime(toUUID(:min_lower_bound), 'UTC'))")
+                .append(" AND toMonday(id_at) < addWeeks(toMonday(UUIDv7ToDateTime(toUUID(:cutoff_id), 'UTC')), 1)")
+                .append(" AND id NOT IN (")
                 .append("SELECT trace_id FROM experiment_items")
                 .append(" WHERE workspace_id IN :workspace_ids_flat")
                 .append(" AND trace_id >= :min_lower_bound")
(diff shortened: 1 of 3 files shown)


apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -1742,11 +1742,22 @@ ORDER BY (workspace_id, project_id, trace_id, parent_span_id, id) DESC, last_upd
...
     private static final String DELETE_FOR_RETENTION = """
             DELETE FROM traces
             WHERE workspace_id IN :workspace_ids
             AND id >= :lower_bound
             AND id \\< :cutoff_id
+            AND toMonday(id_at) >= toMonday(UUIDv7ToDateTime(toUUID(:lower_bound), 'UTC'))
+            AND toMonday(id_at) \\< addWeeks(toMonday(UUIDv7ToDateTime(toUUID(:cutoff_id), 'UTC')), 1)
             AND id NOT IN (
                 SELECT trace_id FROM experiment_items
                 WHERE workspace_id IN :workspace_ids
```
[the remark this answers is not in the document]
andrescrz: Not the right abstraction and not a good practice to concat SQL statements with constants. Keeping this duplication.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000091_add_id_at_to_traces.sql
```diff
@@ -0,0 +1,12 @@
+--liquibase formatted sql
+--changeset andrescrz:000091_add_id_at_to_traces
+--comment: Add id_at materialized column to traces for future weekly partitioning
+
+-- Adds id_at so retention/read predicates can target the future partition expression now
+-- and need no change at cutover. UTC matches the table's other temporal columns and keeps
+-- week boundaries independent of the server timezone.
+-- DateTime (seconds) is enough: discards sub-second, and it halves the column vs DateTime64.
+ALTER TABLE ${ANALYTICS_DB_DATABASE_NAME}.traces ON CLUSTER '{cluster}'
+    ADD COLUMN IF NOT EXISTS id_at DateTime('UTC') MATERIALIZED UUIDv7ToDateTime(toUUID(id));
```
BorisTkachenko, line 10: Maybe it would be more efficient to apply here `toMonday` also. This new column `id_at` is not used without it.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -1742,11 +1742,22 @@ ORDER BY (workspace_id, project_id, trace_id, parent_span_id, id) DESC, last_upd
...
+     * lets the sweep prune to the partitions in range. The bounds use UTC to match {@code id_at}, and the
+     * upper bound advances one week so rows sharing the cutoff's week stay in scope.
+     */
     private static final String DELETE_FOR_RETENTION = """
             DELETE FROM traces
             WHERE workspace_id IN :workspace_ids
             AND id >= :lower_bound
             AND id \\< :cutoff_id
+            AND toMonday(id_at) >= toMonday(UUIDv7ToDateTime(toUUID(:lower_bound), 'UTC'))
+            AND toMonday(id_at) \\< addWeeks(toMonday(UUIDv7ToDateTime(toUUID(:cutoff_id), 'UTC')), 1)
```
BorisTkachenko, line 1760: If we are using now this weekly selection, why are we still keeping filter by `id`. They are doing the same selection.

BorisTkachenko (commented):
Left a few comments to clarify.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -1742,11 +1742,22 @@ ORDER BY (workspace_id, project_id, trace_id, parent_span_id, id) DESC, last_upd
...
+    /**
+     * Retention sweep for the applyToPast=true window {@code [lower_bound, cutoff_id)}.
+     * <p>
+     * {@code toMonday(id_at)} is the future weekly partition expression ({@code id_at} is MATERIALIZED from
+     * the UUIDv7 id as UTC). Bounding it to the cutoff's week range never excludes a row the id-range would
+     * delete, so it does not change which rows are deleted; once {@code traces} is partitioned (OPIK-6900) it
+     * lets the sweep prune to the partitions in range. The bounds use UTC to match {@code id_at}, and the
+     * upper bound advances one week so rows sharing the cutoff's week stay in scope.
+     */
     private static final String DELETE_FOR_RETENTION = """
```
BorisTkachenko, line 1754: Probably we want to have corresponding changes for Spans too. Now spans have following query which could use similar logic:
```
            DELETE FROM spans
            AND trace_id >= :lower_bound
            AND trace_id \\< :cutoff_id
            AND trace_id NOT IN (
                SELECT trace_id FROM experiment_items
                AND trace_id >= :lower_bound
                AND trace_id \\< :cutoff_id
            )
            SETTINGS log_comment = '<log_comment>', lightweight_deletes_sync = 1, allow_nondeterministic_mutations = 1
            ;
            """
```

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000091_add_id_at_to_traces.sql
andrescrz: `id_at` is kept as the full event-time DateTime on purpose — OPIK-6900 reuses it for `PARTITION BY toMonday(id_at)`, the TTL anchor (`id_at + INTERVAL 13 WEEK`), and the `idx_id_at_minmax` skip index, which all need sub-week precision. The predicate wraps it in `toMonday(...)` to match the partition expression; pruning is evaluated per-partition, not per-row, so storing the week instead wouldn't save meaningful work and would lose the precision TTL/skip-index need.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
andrescrz: They're not equivalent. The `id`-range is the precise filter and the correctness driver — it selects exactly `[lower_bound, cutoff_id)`. The `toMonday(id_at)` predicate is intentionally coarser (whole weeks) and exists only to engage partition pruning post-cutover; it's a strict consequence of the id-range, never a replacement. The upper bound spans the cutoff's whole week, so dropping the id-range and keeping only the week filter would delete rows in that week that are newer than the cutoff and must be kept. So: id-range for correctness, week predicate for pruning.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
andrescrz: Agreed it's needed, but intentionally out of scope here — OPIK-6894 is TraceDAO-only and the plan ships traces first, then spans as a later slice. Spans also need their own design rather than a copy: retention there filters by `trace_id` (not the span's own id), and the spans table needs its own `id_at` column + partitioning migration. Tracking it under the spans epic OPIK-6876.

BorisTkachenko (approved):
LGTM!

PR #7222, opened 2026-06-23
[NA] [BE] Bound the read-only free-form SQL ClickHouse client

aadereiko:
## Details

Agent Insights runs caller-supplied (LLM-generated) SQL on the dedicated read-only ClickHouse user (`comet_readonly_freeform_sql_user`). That client had no bounded pool / timeouts, so a slow or stuck query could pin connections and exhaust the pool — wedging *all* ClickHouse access through opik-backend (including the traces endpoint, observed as reactor 30s timeouts → 504s on dev) until a backend restart.

This PR bounds that client:

- **`DatabaseAnalyticsFactory`** — optional v2-client bounds (`setMaxConnections`, `setConnectionRequestTimeout`, `setSocketTimeout`), applied in `buildClient()` only when set. Left null for the shared bulk-insert client so it keeps library defaults.
- **`DatabaseAnalyticsReadOnlyFreeFormSqlConfig`** — the bounds are typed config (`maxConnections`, `connectionRequestTimeout`, `socketTimeout`) with YAML defaults and env overrides, set on the read-only client only.

**Execution-time cap is enforced server-side, not per-query.** The read-only profile runs under `readonly=1`, which rejects any per-query setting other than the `SQL_*` custom settings. The profile already pins `max_execution_time=180` (and memory/row caps) — see `provision_agent_insights_readonly_user.sh` and the test `users.xml`. An earlier revision of this PR tried to set `max_execution_time` per query and broke `AnalyticsQueriesResourceTest`; that was removed. `socketTimeout` defaults to 200s (above the 180s profile cap) so the server-side limit fires first and surfaces as a clean error rather than a client-side socket abort.

This is scoped entirely to the read-only free-form SQL client; the main reactive (traces) client and the bulk-insert client are untouched.

## Testing

- `AnalyticsQueriesResourceTest` (13 tests) and `DatabaseAnalyticsFactoryIntegrationTest` (5 tests) pass against a real ClickHouse testcontainer locally.
- `DatabaseAnalyticsFactoryIntegrationTest#clientBoundsAreApplied` asserts the bounds are accepted and the client still queries.
- The freeform-SQL endpoint tests exercise the real `readonly=1` profile (via test `users.xml`), so they regression-cover the rejection bug this PR fixes.

## Change checklist

- [x] Tested the changes (see Testing).
- [x] No new public API / breaking changes.
- [x] Config is backward-compatible (new fields default to prior behavior; env-overridable).

## Issues

N/A — operational reliability fix surfaced on dev. No linked ticket.

## Documentation

No user-facing docs needed. New env vars (`ANALYTICS_DB_READ_ONLY_FREEFORM_SQL_MAX_CONNECTIONS`, `..._CONNECTION_REQUEST_TIMEOUT`, `..._SOCKET_TIMEOUT`) are documented inline in `config.yml`.

### DIFF
--- a/apps/opik-backend/config.yml
+++ b/apps/opik-backend/config.yml
@@ -119,6 +119,10 @@ databaseAnalyticsReadOnlyFreeFormSql:
   #  Default: opik
   #  Description: The read-only user password used to connect to the server
   password: ${ANALYTICS_DB_READ_ONLY_FREEFORM_SQL_PASS:-opik}
+  #  Default: 200s
+  #  Description: Socket read timeout for the read-only client (client-v2 default is 0/none). Kept
+  #   above the read-only profile's max_execution_time (180s) so the server-side cap fires first.
+  socketTimeout: ${ANALYTICS_DB_READ_ONLY_FREEFORM_SQL_SOCKET_TIMEOUT:-200s}
 
 # https://www.dropwizard.io/en/stable/manual/configuration.html#health
 health:
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryDAO.java
@@ -73,6 +73,8 @@ class FreeFormSqlQueryDAOImpl implements FreeFormSqlQueryDAO {
     @WithSpan
     public CompletableFuture<FreeFormSqlResult> execute(@NonNull String workspaceId, @NonNull UUID projectId,
             @NonNull String query) {
+        // Only the SQL_ custom settings are sent: readonly=1 rejects any other per-query setting.
+        // Execution/memory/row caps are pinned on the read-only user's server-side profile.
         var settings = new QuerySettings()
                 .serverSetting(SETTING_WORKSPACE_ID, workspaceId)
                 .serverSetting(SETTING_PROJECT_ID, projectId.toString());
--- a/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/DatabaseAnalyticsFactory.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/DatabaseAnalyticsFactory.java
@@ -13,6 +13,7 @@ import lombok.Getter;
 import lombok.RequiredArgsConstructor;
 import org.apache.commons.lang3.StringUtils;
 
+import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.Map;
@@ -51,6 +52,9 @@ public class DatabaseAnalyticsFactory {
     private @Min(1) Integer asyncInsertBusyTimeoutMaxMs;
 
     private Duration healthCheckTimeout = Duration.seconds(1);
+
+    // Optional socket timeout, applied in buildClient() only when set (null = library default of 0/no timeout).
+    private Duration clientSocketTimeout;
 
     public ConnectionFactory build() {
         var queryParametersOverrides = getQueryParametersOverrides(queryParameters);
@@ -96,6 +100,11 @@ public class DatabaseAnalyticsFactory {
         // are still returned by parseQueryParameters() for tests/observability.
         var parsed = parseQueryParameters(getQueryParametersOverrides(queryParameters));
         parsed.serverSettings().forEach(builder::serverSetting);
+
+        if (clientSocketTimeout != null) {
+            builder.setSocketTimeout(clientSocketTimeout.toMilliseconds(), ChronoUnit.MILLIS);
+        }
+
         return builder.build();
     }
 
--- a/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/DatabaseAnalyticsReadOnlyFreeFormSqlConfig.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/DatabaseAnalyticsReadOnlyFreeFormSqlConfig.java
@@ -1,14 +1,16 @@
 package com.comet.opik.infrastructure;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import io.dropwizard.util.Duration;
+import jakarta.validation.Valid;
 import jakarta.validation.constraints.NotBlank;
 import jakarta.validation.constraints.NotNull;
 import lombok.Data;
 
 /**
- * Credentials for the Agent Insights read-only free-form SQL ClickHouse user. Scope is intentionally reduced to just
- * the user/password: every other connection parameter (protocol, host, port, database) is shared with
- * {@code databaseAnalytics} and reused when building the client (see {@code DatabaseAnalyticsModule}).
+ * Credentials and socket timeout for the Agent Insights read-only free-form SQL ClickHouse user. Connection params
+ * (protocol/host/port/database) are shared with {@code databaseAnalytics}. {@code socketTimeout} sits above the
+ * read-only profile's 180s {@code max_execution_time} so the server-side cap fires first (client-v2 default is 0/none).
  */
 @Data
 public class DatabaseAnalyticsReadOnlyFreeFormSqlConfig {
@@ -18,4 +20,7 @@ public class DatabaseAnalyticsReadOnlyFreeFormSqlConfig {
 
     @JsonProperty
     private @NotNull String password;
+
+    @JsonProperty
+    private @Valid @NotNull Duration socketTimeout;
 }
--- a/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/DatabaseAnalyticsModule.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/DatabaseAnalyticsModule.java
@@ -61,14 +61,15 @@ public class DatabaseAnalyticsModule extends DropwizardAwareModule<OpikConfigura
     // credentials. No queryParameters: the user runs under readonly=1 and would reject per-query server settings.
     private Client buildReadOnlyFreeFormSqlClient() {
         var main = configuration().getDatabaseAnalytics();
-        var credentials = configuration().getDatabaseAnalyticsReadOnlyFreeFormSql();
+        var freeFormSqlConfig = configuration().getDatabaseAnalyticsReadOnlyFreeFormSql();
         var factory = new DatabaseAnalyticsFactory();
         factory.setProtocol(main.getProtocol());
         factory.setHost(main.getHost());
         factory.setPort(main.getPort());
         factory.setDatabaseName(main.getDatabaseName());
-        factory.setUsername(credentials.getUsername());
-        factory.setPassword(credentials.getPassword());
+        factory.setUsername(freeFormSqlConfig.getUsername());
+        factory.setPassword(freeFormSqlConfig.getPassword());
+        factory.setClientSocketTimeout(freeFormSqlConfig.getSocketTimeout());
         return factory.buildClient();
     }
 
--- a/apps/opik-backend/src/test/resources/config-test.yml
+++ b/apps/opik-backend/src/test/resources/config-test.yml
@@ -100,6 +100,8 @@ uuidValidation:
 databaseAnalyticsReadOnlyFreeFormSql:
   username: comet_readonly_freeform_sql_user
   password: opik
+  # Plain value (no env var).
+  socketTimeout: 200s
 
 # https://www.dropwizard.io/en/stable/manual/configuration.html#health
 health:


apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryDAO.java
```diff
@@ -65,22 +70,26 @@ class FreeFormSqlQueryDAOImpl implements FreeFormSqlQueryDAO {
...
+        var settings = boundedSettings()
                 .serverSetting(SETTING_WORKSPACE_ID, workspaceId)
                 .serverSetting(SETTING_PROJECT_ID, projectId.toString());
 
         return readOnlyClient.queryRecords(query, settings)
                 .thenApply(FreeFormSqlQueryDAOImpl::readResult);
     }
 
+    private static QuerySettings boundedSettings() {
+        return new QuerySettings().setMaxExecutionTime(MAX_EXECUTION_TIME_SECONDS);
```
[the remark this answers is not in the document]
aadereiko: Resolved in 2d34dc10ae. Removed the per-query `max_execution_time` entirely — you're right it's rejected under `readonly=1` (only the `SQL_*` custom settings are changeable there). The execution/memory/row caps are already pinned server-side on the read-only profile (`max_execution_time=180`, see `provision_agent_insights_readonly_user.sh` + test `users.xml`), so the per-query setting was both redundant and breaking. `AnalyticsQueriesResourceTest` (13) passes again.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryDAO.java
```diff
@@ -54,6 +54,11 @@ class FreeFormSqlQueryDAOImpl implements FreeFormSqlQueryDAO {
     private static final String SETTING_WORKSPACE_ID = "SQL_workspace_id";
     private static final String SETTING_PROJECT_ID = "SQL_project_id";
 
+    // ClickHouse-side cap so a caller's query can't run unbounded and pin its connection
+    // (surfaced as code 159 → clean 4xx by FreeFormSqlQueryService). Below the client socket
+    // timeout so the server-side limit fires first.
+    private static final int MAX_EXECUTION_TIME_SECONDS = 60;
+
     private final Client readOnlyClient;
```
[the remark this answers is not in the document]
aadereiko: Resolved in 2d34dc10ae. The `MAX_EXECUTION_TIME_SECONDS` constant is gone (per the high-severity finding above, that per-query setting can't be sent under `readonly=1`; the cap lives on the server-side profile). The remaining client bounds are now typed config — `maxConnections` / `connectionRequestTimeout` / `socketTimeout` on `DatabaseAnalyticsReadOnlyFreeFormSqlConfig` with YAML defaults + env overrides (`ANALYTICS_DB_READ_ONLY_FREEFORM_SQL_*`), replacing the hardcoded constants. Covered by `AnalyticsQueriesResourceTest` (real readonly=1 profile) and `DatabaseAnalyticsFactoryIntegrationTest#clientBoundsAreApplied`.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryDAO.java
```diff
@@ -73,6 +73,8 @@ public CompletableFuture<List<String>> explainAst(@NonNull String query) {
     @WithSpan
     public CompletableFuture<FreeFormSqlResult> execute(@NonNull String workspaceId, @NonNull UUID projectId,
             @NonNull String query) {
+        // Only the SQL_ custom settings are sent: readonly=1 rejects any other per-query setting.
+        // Execution/memory/row caps are pinned on the profile (provision_agent_insights_readonly_user.sh).
```
andrescrz: Nit: comment not fully precise as it references `provision_agent_insights_readonly_user.sh` which applies for local installations (and local development) but not for production installations. I'd generalise this comment and make it more neutral, this is indeed configured in the profile, but not necessarily using this script.

BorisTkachenko (changes requested):
I doubt that we need this as we have already below limitations set on the infra level for read-only user:
```
SETTINGS
    readonly = 1,
    max_execution_time = 180,
    max_memory_usage = 8589934592,
    max_result_rows = 100000,
    result_overflow_mode = 'throw',
    max_rows_to_read = 100000000,
    read_overflow_mode = 'throw',
    max_concurrent_queries_for_user = 5,
    SQL_workspace_id = '' CHANGEABLE_IN_READONLY,
    SQL_project_id   = '' CHANGEABLE_IN_READONLY
TO comet_readonly_freeform_sql_user;
```

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/DatabaseAnalyticsModule.java
```diff
@@ -69,6 +69,10 @@ private Client buildReadOnlyFreeFormSqlClient() {
         factory.setDatabaseName(main.getDatabaseName());
         factory.setUsername(credentials.getUsername());
         factory.setPassword(credentials.getPassword());
+        // Bound the caller-supplied SQL client so a stuck query can't pin connections or exhaust the pool.
+        factory.setClientMaxConnections(credentials.getMaxConnections());
```
andrescrz: Minor: the variable `credentials` is no longer a good name if the `databaseAnalyticsReadOnlyFreeFormSql` configuration section contains more stuff, such as client configuration. I'd rename it. 

In addition, the `config.yml` file contains stales references, mostly in the comments.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/DatabaseAnalyticsFactory.java
```diff
@@ -37,6 +38,11 @@ public class DatabaseAnalyticsFactory {
     private String queryParameters;
     private Duration healthCheckTimeout = Duration.seconds(1);
 
+    // Optional v2-client bounds, applied in buildClient() only when set (null = library defaults).
```
andrescrz: Minor: If you inspect the javadoc of the `Client.Builder`, you'll notice that these have defaults when not provided (`null` in your case)`. It'd be worthy documenting it here and probably in the `config.yml` file.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/DatabaseAnalyticsFactory.java
```diff
@@ -80,6 +86,17 @@ public Client buildClient() {
         // are still returned by parseQueryParameters() for tests/observability.
         ParsedQueryParameters parsed = parseQueryParameters(queryParameters);
         parsed.serverSettings().forEach(builder::serverSetting);
+
+        if (clientMaxConnections != null) {
+            builder.setMaxConnections(clientMaxConnections);
```
andrescrz: If you inspect this `Client.Builder`, it has many other similar resilience settings. If you really want to harden this client and make an impact, I recommend you doing an audit on the whole client and making them configurable. Specially, those without sensible defaults, like unbounded wait times (typically 0 timeout etc.). You can ignore other non-safety unrelated settings (like setting a proxy etc.).

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/DatabaseAnalyticsReadOnlyFreeFormSqlConfig.java
```diff
@@ -18,4 +20,13 @@ public class DatabaseAnalyticsReadOnlyFreeFormSqlConfig {
 
     @JsonProperty
     private @NotNull String password;
+
+    @JsonProperty
+    private @NotNull Integer maxConnections = 20;
```
andrescrz: Let's remove all the default from the Java side, so we don't need to keep them in sync with the values provided in the config yaml file. Better have a single source of truth. As these are annotated as NotNull, a value will always be provided.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/DatabaseAnalyticsReadOnlyFreeFormSqlConfig.java
```diff
@@ -18,4 +20,13 @@ public class DatabaseAnalyticsReadOnlyFreeFormSqlConfig {
...
+
+    @JsonProperty
+    private @NotNull Duration connectionRequestTimeout = Duration.seconds(10);
```
andrescrz: Minor: add `@Valid` for `Duration` objects.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/DatabaseAnalyticsFactoryIntegrationTest.java
```diff
@@ -119,6 +120,19 @@ void bulkInsertRoundTrip() throws Exception {
         }
     }
 
+    @Test
+    @DisplayName("client bounds (max connections + timeouts) are applied and the client still queries")
+    void clientBoundsAreApplied() throws Exception {
```
andrescrz: Minor: generally, client configurations like socket timeouts etc. aren't functionally tested (except for a few exception like a limit). If happy path tests work, they're already indirectly tested.

apps/opik-backend/config.yml
```diff
@@ -113,6 +113,18 @@ databaseAnalyticsReadOnlyFreeFormSql:
   #  Default: opik
   #  Description: The read-only user password used to connect to the server
   password: ${ANALYTICS_DB_READ_ONLY_FREEFORM_SQL_PASS:-opik}
+  #  Default: 20
```
andrescrz: If these aren't provided, the defaults according to the `ClientBuilder` javadoc are:
1. maxConnections: 10 -> you increase to 20 by default. This seems reasonable.
2. connectionRequestTimeout: 10s, you keep the same, but now is configurable -> reasonable.
3. socketTimeout: 0s (infinite) -> you set it to `200s`. This is a great improvement to kill calls on the client side as extra measure of safety on top of the server side cap. Also, you make it configurable.

apps/opik-backend/config.yml
andrescrz, line 1: You need to propagate these to the config test yaml file as well, following the usual conventions: same comments, unless test values defer, no variables.

andrescrz (commented):
Many minor comments, so focus on the important ones.

Optionally, you could ship these 3 new configurations for now or optionally harden the client more as there are many other interesting setting available in the `Client. Builder`. You can ship that in a follow-up change or totally skip the suggestion.

andrescrz:
> I doubt that we need this as we have already below limitations set on the infra level for read-only user:
> 
> ```
> SETTINGS
>     readonly = 1,
>     max_execution_time = 180,
>     max_memory_usage = 8589934592,
>     max_result_rows = 100000,
>     result_overflow_mode = 'throw',
>     max_rows_to_read = 100000000,
>     read_overflow_mode = 'throw',
>     max_concurrent_queries_for_user = 5,
>     SQL_workspace_id = '' CHANGEABLE_IN_READONLY,
>     SQL_project_id   = '' CHANGEABLE_IN_READONLY
> TO comet_readonly_freeform_sql_user;
> ```

@aadereiko @BorisTkachenko I believe hardening on the client side provides value and extra layer of protection on top of the server side settings, provided they don't contradict them (e.g: same timeouts or slightly higher). In addition, stuff such as the number of client connections should be handled on the client side.

I'd move forward with this, after addressing the blocking feedback.

apps/opik-backend/src/main/java/com/comet/opik/domain/FreeFormSqlQueryDAO.java
aadereiko: Done in a67e7317 — reworded to "Execution/memory/row caps are pinned on the read-only user's server-side profile", dropping the local `provision_…readonly_user.sh` reference.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/DatabaseAnalyticsReadOnlyFreeFormSqlConfig.java
aadereiko: Done in a67e7317 — added `@Valid` on the `socketTimeout` `Duration`.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/DatabaseAnalyticsReadOnlyFreeFormSqlConfig.java
aadereiko: Done in a67e7317 — removed the `socketTimeout` Java-side default as well, so config (`config.yml` / `config-test.yml`) is the single source of truth. The field stays `@NotNull`, so a value is always provided.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/DatabaseAnalyticsModule.java
aadereiko: Done in a67e7317 — renamed the local var `credentials` → `freeFormSqlConfig`, since the section now also carries the socket timeout.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/DatabaseAnalyticsFactoryIntegrationTest.java
aadereiko: Done in a67e7317 — removed the dedicated socket-timeout test; the happy-path factory tests already exercise the built client indirectly, as you noted.

apps/opik-backend/config.yml
aadereiko: Done in a67e7317 — added `socketTimeout` to `config-test.yml` (plain value, no env var). With the Java default removed, the `@NotNull` is satisfied by config in both files.

apps/opik-backend/config.yml
aadereiko: Thanks — confirmed. After narrowing the change, `socketTimeout` (200s) is the only client-side bound now, kept above the 180s server cap; the maxConnections/connectionRequestTimeout settings were dropped.

aadereiko:
@BorisTkachenko thanks — I've narrowed this down to a single client-side `socketTimeout` (200s), kept deliberately **above** the read-only profile's `max_execution_time=180s` so the server-side cap still fires first. It doesn't override or duplicate the infra-level limits — it's purely a client-side backstop: the client-v2 default is `0` (no socket timeout), so a half-open/dropped connection would otherwise pin a pool connection (and its request thread) until a restart. Per @andrescrz's note the rest can be a follow-up. Could you take another look?

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/DatabaseAnalyticsReadOnlyFreeFormSqlConfig.java
```diff
@@ -18,4 +20,7 @@ public class DatabaseAnalyticsReadOnlyFreeFormSqlConfig {
 
     @JsonProperty
     private @NotNull String password;
+
+    @JsonProperty
+    private @Valid @NotNull Duration socketTimeout;
```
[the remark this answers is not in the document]
aadereiko: `maxConnections`/`connectionRequestTimeout` were deliberately removed when this PR was narrowed — no need for them here. This change is intentionally scoped to the minimal client-side `socketTimeout` backstop; the read-only user is also capped at `max_concurrent_queries_for_user=5` server-side, so a single user can't exhaust the pool. Per @andrescrz's review, a fuller `Client.Builder` hardening pass can be a follow-up.

BorisTkachenko (approved):

PR #7223, opened 2026-06-23
[NA] [SDK] feat: allow passing provider name to track_openai

alexkuzmik:
## Details

Adds an optional `provider` argument to `track_openai` so callers can set the provider name recorded on **every LLM span** the OpenAI integration creates (chat completions, responses, videos, audio/TTS). The OpenAI SDK is widely used as a client for other OpenAI-compatible APIs (Together, OpenRouter, vLLM, DeepSeek, etc.); previously the provider was inferred only from the client's base-URL host, which is opaque for those setups.

- `track_openai(..., provider: Optional[Union[str, LLMProvider]] = None)` resolves the provider once and threads it to all LLM-span decorator factories. An `LLMProvider` enum value is normalized to its string value so a bare enum member never leaks into spans as `"LLMProvider.OPENAI"`.
- Token usage is still parsed with the OpenAI converter (`LLMProvider.OPENAI`) regardless of the span provider label — `"openai"` there denotes the usage payload format, not the provider.
- Auto-detection from the base URL remains the default; no behavior change when `provider` is omitted.

## Change checklist
- [x] User facing
- [x] Documentation update

## Issues

- NA — no ticket; see PR title `[NA]`.

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: implementation, tests, and documentation for the `provider` argument
- Human verification: reviewed the diff; ran ruff (check + format), mypy v1.19.1 (project config), and pytest `--collect-only` locally

## Testing

- Added library integration tests (live OpenAI):
  - `test_openai_chat_completions.py` — parametrized over a plain string (`"custom-provider"`) and `LLMProvider.ANTHROPIC`, asserting the span records the overridden provider **and** that usage is still parsed in OpenAI format.
  - `test_openai_responses.py` — same assertion for `responses.create`.
- Local checks (in `sdks/python`):
  - `ruff check` and `ruff format --check` on all changed files — pass
  - `mypy==1.19.1 --config-file pyproject.toml` on the changed source files — pass
  - `pytest --collect-only` on the two test files — 35 tests collected, no errors
- The live integration tests require an `OPENAI_API_KEY`; they run in CI (no key available locally).

## Documentation

- Documented the custom-provider use case in the OpenAI integration docs (both `docs/tracing/integrations/openai.mdx` and `docs-v2/integrations/openai.mdx`), including a `provider="deepseek"` example and an `LLMProvider` enum example, plus a note that usage is always parsed in the OpenAI format.


petrotiurin (approved):
LGTM!

PR #7231, opened 2026-06-24
[OPIK-7101] Make project stats opt-in on /projects/retrieve

Nimrod007:
## Details

`POST /v1/private/projects/retrieve` (`ProjectService.retrieveByName`) is used by callers only to resolve a project name to its id, but on every call it recomputed full project analytics (feedback scores, traces/spans aggregation, duration, cost, counts) over ClickHouse. This caused a platform-wide Opik slowdown in prod: at SDK name-resolution call rates it saturated ClickHouse CPU (~90% on both nodes), since each call fanned out into multi-million-row aggregations (~20 CPU-seconds per call).

No first-party caller reads those stat fields from this endpoint:
- TS SDK (`getProjectIdByName`) and Python SDK (`resolve_project_id_by_name`) return `project.id` only.
- Frontend `useProjectByName` reads only `.id` (RedirectProjects, Playground, GetStarted); it predates the stats and never read them.
- Internal ingestion uses the stats-free `retrieveByNamesOrCreate`; `retrieveByName` has no internal callers.

The stats were added in OPIK-287 for the projects table + dashboards, which are served by the dedicated `GET /v1/private/projects/stats` endpoint (`ProjectService.getStats`, FE hook `useProjectStatisticList`). Note: the list endpoint (`GET /v1/private/projects`) and `getById` enrich only with `lastUpdatedTraceAt`, **not** the heavy stats. `/retrieve` was the one other path that recomputed the full stats, and it was swept into the OPIK-287 change. This PR adds an `includeStats` flag (default `false`) to `ProjectRetrieve`, so `/retrieve` skips the ClickHouse aggregations unless explicitly requested. The response contract is preserved: the stat fields are already `@Nullable`, so omitted stats simply serialize as `null`; consumers that need them opt in with `{"name": "...", "includeStats": true}`.

- `ProjectRetrieve`: add `@Nullable includeStats` + `shouldIncludeStats()` (coalesces null to false)
- `ProjectService.retrieveByName(name, includeStats)`; stats enrichment extracted into `enrichWithStats(...)`
- `ProjectsResource.retrieveProject` passes `retrieve.shouldIncludeStats()`

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves OPIK-7101

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: Investigation of the prod ClickHouse saturation, root-cause analysis, and authoring this change (backend + tests) under human direction.
- Human verification: Author reviewed the diff and the verification (no first-party caller reads the stats; provenance traced via git history). CI compiles/tests.

## Testing

- Updated `ProjectsResourceTest.RetrieveProjectTest`:
  - Existing happy path now sends `includeStats(true)` and still asserts the populated stats (proves opt-in works).
  - New `getProjectByName__whenStatsNotRequested__thenReturnProjectWithoutStats` asserts the default (no flag) returns the project with null stat fields (feedbackScores/duration/totalEstimatedCost/totalEstimatedCostSum/usage/traceCount/guardrailsFailedCount/errorCount) and no ClickHouse aggregation. `lastUpdatedTraceAt` is intentionally not asserted: it is a `projects` table column (set on ingestion), not part of the stats aggregation.
- Relying on PR CI for compile + full integration suite (Java 25 + Maven not run locally).

## Documentation

No public docs change. The new request field is optional and backward compatible; the OpenAPI/Fern spec will pick up `includeStats` on the next codegen.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/ProjectRetrieve.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/ProjectRetrieve.java
@@ -1,10 +1,11 @@
 package com.comet.opik.api;
 
+import jakarta.annotation.Nullable;
 import jakarta.validation.constraints.NotBlank;
 import lombok.Builder;
 
 @Builder
-public record ProjectRetrieve(@NotBlank String name, Boolean includeStats) {
+public record ProjectRetrieve(@NotBlank String name, @Nullable Boolean includeStats) {
 
     /**
      * Whether the response should be enriched with project-level statistics
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/ProjectRetrieve.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/ProjectRetrieve.java
@@ -4,5 +4,17 @@ import jakarta.validation.constraints.NotBlank;
 import lombok.Builder;
 
 @Builder
-public record ProjectRetrieve(@NotBlank String name) {
+public record ProjectRetrieve(@NotBlank String name, Boolean includeStats) {
+
+    /**
+     * Whether the response should be enriched with project-level statistics
+     * (feedback scores, trace/span aggregations, duration, cost, counts).
+     * <p>
+     * Defaults to {@code false}: callers that only need to resolve a project
+     * name to its id (the SDK and the UI redirect path) should not trigger the
+     * expensive ClickHouse aggregations. See OPIK-7101.
+     */
+    public boolean shouldIncludeStats() {
+        return Boolean.TRUE.equals(includeStats);
+    }
 }
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/ProjectRetrieve.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/ProjectRetrieve.java
@@ -13,9 +13,12 @@ public record ProjectRetrieve(@NotBlank String name, @Nullable Boolean includeSt
      * <p>
      * Defaults to {@code false}: callers that only need to resolve a project
      * name to its id (the SDK and the UI redirect path) should not trigger the
-     * expensive ClickHouse aggregations. See OPIK-7101.
+     * expensive ClickHouse aggregations. The canonical accessor is overridden to
+     * coalesce {@code null} to {@code false}, so it never returns {@code null}.
+     * See OPIK-7101.
      */
-    public boolean shouldIncludeStats() {
+    @Override
+    public Boolean includeStats() {
         return Boolean.TRUE.equals(includeStats);
     }
 }
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/ProjectsResource.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/ProjectsResource.java
@@ -216,7 +216,7 @@ public class ProjectsResource {
             @RequestBody(content = @Content(schema = @Schema(implementation = ProjectRetrieve.class))) @Valid ProjectRetrieve retrieve) {
         String workspaceId = requestContext.get().getWorkspaceId();
         log.info("Retrieve project by name '{}', on workspace_id '{}'", retrieve.name(), workspaceId);
-        Project project = projectService.retrieveByName(retrieve.name(), retrieve.shouldIncludeStats());
+        Project project = projectService.retrieveByName(retrieve.name(), retrieve.includeStats());
         log.info("Retrieved project id '{}' by name '{}', on workspace_id '{}'", project.id(), retrieve.name(),
                 workspaceId);
         return Response.ok().entity(project).build();
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/ProjectsResource.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/ProjectsResource.java
@@ -216,7 +216,7 @@ public class ProjectsResource {
             @RequestBody(content = @Content(schema = @Schema(implementation = ProjectRetrieve.class))) @Valid ProjectRetrieve retrieve) {
         String workspaceId = requestContext.get().getWorkspaceId();
         log.info("Retrieve project by name '{}', on workspace_id '{}'", retrieve.name(), workspaceId);
-        Project project = projectService.retrieveByName(retrieve.name());
+        Project project = projectService.retrieveByName(retrieve.name(), retrieve.shouldIncludeStats());
         log.info("Retrieved project id '{}' by name '{}', on workspace_id '{}'", project.id(), retrieve.name(),
                 workspaceId);
         return Response.ok().entity(project).build();
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectService.java
@@ -100,7 +100,7 @@ public interface ProjectService {
 
     Project getOrCreate(String workspaceId, String projectName, String userName);
 
-    Project retrieveByName(String projectName);
+    Project retrieveByName(String projectName, boolean includeStats);
 
     Mono<List<Project>> retrieveByNamesOrCreate(Set<String> projectNames);
 
@@ -620,7 +620,7 @@ class ProjectServiceImpl implements ProjectService {
     }
 
     @Override
-    public Project retrieveByName(@NonNull String projectName) {
+    public Project retrieveByName(@NonNull String projectName, boolean includeStats) {
         var workspaceId = requestContext.get().getWorkspaceId();
 
         Optional<Project> projects = template.inTransaction(READ_ONLY, handle -> {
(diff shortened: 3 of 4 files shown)


apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectService.java
```diff
@@ -631,32 +631,34 @@ public Project retrieveByName(@NonNull String projectName) {
...
-                })
-                .orElseThrow(this::createNotFoundError);
+        Map<UUID, Map<String, Object>> projectStats = getProjectStats(List.of(project.id()),
+                workspaceId, ProjectCriteria.builder().build());
+
+        return project.toBuilder()
+                .lastUpdatedTraceAt(projectLastUpdatedTraceAtMap.get(project.id()))
+                .feedbackScores(StatsMapper.getStatsFeedbackScores(projectStats.get(project.id())))
+                .usage(StatsMapper.getStatsUsage(projectStats.get(project.id())))
+                .duration(StatsMapper.getStatsDuration(projectStats.get(project.id())))
```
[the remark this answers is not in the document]
Nimrod007: Good catch. This duplication is pre-existing: the `StatsMapper` chain was already inline in `retrieveByName` before this PR; I only extracted it verbatim into `enrichWithStats`. Fully de-duplicating is awkward because the two paths build different types (`getStats` -> `ProjectStatsSummaryItem`, this path -> `Project`), so a shared mapper would need a small refactor. Leaving it out of scope here to keep the change focused on making stats opt-in; happy to file a follow-up to unify the mapping if you'd prefer.

apps/opik-backend/src/main/java/com/comet/opik/api/ProjectRetrieve.java
```diff
@@ -1,8 +1,21 @@
...
+
+    /**
+     * Whether the response should be enriched with project-level statistics
+     * (feedback scores, trace/span aggregations, duration, cost, counts).
+     * <p>
+     * Defaults to {@code false}: callers that only need to resolve a project
+     * name to its id (the SDK and the UI redirect path) should not trigger the
+     * expensive ClickHouse aggregations. See OPIK-7101.
+     */
+    public boolean shouldIncludeStats() {
```
andrescrz: Minor: you could overwrite the default accessor `includeStats`, so while the signature has to stay with `Boolean`, you would ensure that it always returns `true/false` and never `null`. That way you would harden the implementation for any user and you would avoid the duplication of `includeStats()` and `shouldIncludeStats()`.

apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectService.java
```diff
@@ -631,32 +631,34 @@ public Project retrieveByName(@NonNull String projectName) {
...
+    }
 
-                    Map<UUID, Map<String, Object>> projectStats = getProjectStats(List.of(project.id()),
-                            workspaceId, ProjectCriteria.builder().build());
+    private Project enrichWithStats(Project project, String workspaceId) {
+        Map<UUID, Instant> projectLastUpdatedTraceAtMap = transactionTemplateAsync
+                .nonTransaction(connection -> {
+                    Set<UUID> projectIds = Set.of(project.id());
+                    return traceDAO.getLastUpdatedTraceAt(projectIds, workspaceId, connection);
+                }).block();
```
andrescrz, line 643: Pre-existing issue, so not a blocker, but raising it as it affects performance. This service method should never `block` here. Instead, it should become a fully reactive method, by returning `Mono<Project>` or similar and delay the `block` call to the last minute at the resource endpoint method level, as we usually do.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/ProjectsResourceTest.java
```diff
@@ -788,6 +788,52 @@ void getProjectById__whenProjectExists__thenReturnProject() {
...
+                    .path("retrieve")
+                    .request()
+                    .header(HttpHeaders.AUTHORIZATION, apiKey)
+                    .header(WORKSPACE_HEADER, workspaceName)
+                    .post(Entity.json(ProjectRetrieve.builder().name(project.name()).build()))) {
+
+                assertThat(actualResponse.getStatusInfo().getStatusCode()).isEqualTo(HttpStatus.SC_OK);
+                assertThat(actualResponse.hasEntity()).isTrue();
+
+                var actualEntity = actualResponse.readEntity(Project.class);
```
andrescrz, line 832: Pre existing tech debt, but not a blocker: generally these calls should be encapsulated and re-used in a ProjectClient, following the usual pattern of our tests. That would include the client call, and asserting the OK status code and parsing the entity. We can have two overloaded methods (only call + call and assertions and entity parsing) and parameterise the expected status code. This is the usual pattern.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/ProjectsResourceTest.java
```diff
@@ -788,6 +788,52 @@ void getProjectById__whenProjectExists__thenReturnProject() {
...
+                // Note: lastUpdatedTraceAt is a `projects` table column (set on ingestion), not part of the ClickHouse
+                // stats aggregation, so it is intentionally not asserted here.
+                assertThat(actualEntity.feedbackScores()).isNull();
+                assertThat(actualEntity.duration()).isNull();
+                assertThat(actualEntity.totalEstimatedCost()).isNull();
+                assertThat(actualEntity.totalEstimatedCostSum()).isNull();
+                assertThat(actualEntity.usage()).isNull();
+                assertThat(actualEntity.traceCount()).isNull();
+                assertThat(actualEntity.guardrailsFailedCount()).isNull();
+                assertThat(actualEntity.errorCount()).isNull();
```
andrescrz: Minor: assertions aren't very endurable nor maintainable for the future. Better assert the full actual vs expected object. In this case, you'd build the expected with null in all stats field and leave populated only the core `Project` fields. 

In addition, generally we have assertion helper methods/classes that we reuse in our tests to encapsulate and not repeat this logic.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/ProjectsResource.java
```diff
@@ -216,7 +216,7 @@ public Response retrieveProject(
             @RequestBody(content = @Content(schema = @Schema(implementation = ProjectRetrieve.class))) @Valid ProjectRetrieve retrieve) {
         String workspaceId = requestContext.get().getWorkspaceId();
         log.info("Retrieve project by name '{}', on workspace_id '{}'", retrieve.name(), workspaceId);
-        Project project = projectService.retrieveByName(retrieve.name());
+        Project project = projectService.retrieveByName(retrieve.name(), retrieve.shouldIncludeStats());
```
andrescrz: Final thought: given the positive results of the audit, I believe the best is to bake this approach in prod for a while and after being fully confident, fully remove the stats logic from this endpoint in a follow-up PR.

It's a reasonable amount of tech debt that can bite us in the future again. Better get rid of it.

andrescrz (dismissed):
Fine to move forward, with some thoughts and potential follow-ups.

apps/opik-backend/src/main/java/com/comet/opik/api/ProjectRetrieve.java
Nimrod007: Done in 8aff14c: overrode the canonical `includeStats()` accessor to coalesce `null` → `false` and removed `shouldIncludeStats()`. The single call site now uses `retrieve.includeStats()`. Thanks for the tip — cleaner and one fewer method.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/ProjectsResourceTest.java
Nimrod007: Done in 8aff14c: now building an expected `Project` with the stats fields nulled and core fields populated, asserting via `usingRecursiveComparison().ignoringFields(IGNORED_FIELD_MIN)` (same pattern as the happy-path test) instead of field-by-field null checks.

apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectService.java
Nimrod007: Agreed — this `block()` is pre-existing (extracted verbatim into `enrichWithStats`). Keeping it out of scope to stay focused on the opt-in change; will track making `retrieveByName` fully reactive (`Mono<Project>`, blocking at the resource boundary) as a follow-up.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/ProjectsResourceTest.java
Nimrod007: Acknowledged — pre-existing pattern in this test class. Will fold the retrieve calls into a reusable `ProjectClient` helper (call-only + call-with-assertions, parameterised status) as part of the test-debt follow-up.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/ProjectsResource.java
Nimrod007: Agreed. Plan is exactly that: bake the opt-in default in prod, confirm ClickHouse load is back to normal, then fully remove the stats path from `/retrieve` in a follow-up so the tech debt is gone for good.

andrescrz (approved):
Thanks for addressing the feedback, all LGTM.

PR #7234, opened 2026-06-24
[OPIK-6855] [BE] Fix test suite items lost when saving after editing an item assertion

miguelgrc:
## Details

Editing a test-suite item to add/change an assertion (or its description) could produce a new dataset version that shows **zero items** — the item appeared lost.

Root cause is in the backend item-level edit path (`EDIT_ITEM_VIA_SELECT_INSERT`, run via the ClickHouse v2 client): `evaluators` and `description` were bound through `{:String}` query-parameter substitution, which processes backslash escape sequences. The default `llm_judge` prompt template contains newlines (`\n`), so the stored `evaluators` JSON ended up with **literal** newlines — invalid JSON. On read, the ClickHouse r2dbc driver silently drops any row whose mapping function throws, so the whole item disappeared from the list (the version's item count still reported it, masking the loss). A backslash in a `description` failed the write outright (500).

Fix:
- **Write**: base64-encode `evaluators` and `description` before binding and `base64Decode(...)` them in the SQL, so the values cross the `{:String}` substitution intact. The stored column remains plain JSON/text, so the read path is unchanged. A comment at the binding site flags that any new free-text/JSON `{:String}` param needs the same treatment.
- **Read resilience**: harden `getEvaluators` to degrade to `null` (with a WARN) on unparseable stored JSON, so a corrupt row stays visible instead of being silently dropped by the driver.

`execution_policy` is intentionally left unencoded — it serializes to integers-only JSON with no escape sequences, so it cannot be corrupted by this mechanism.

## Change checklist

- [x] User facing
- [ ] Documentation update

## Issues

- OPIK-6855

## Testing

Added an end-to-end regression test in `DatasetVersionResourceTest` that edits an item with an evaluator prompt **and** a description containing a newline, an escaped quote, and a backslash, then asserts (full-item recursive comparison) that the item survives the round-trip with every field intact. Verified it **fails without** the fix (item dropped / evaluators null) and **passes with** it. `mvn compile` + `spotless:check` clean.

## Documentation

N/A

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemResultMapper.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemResultMapper.java
@@ -20,6 +20,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.reactivestreams.Publisher;
 import reactor.core.publisher.Mono;
 
+import java.io.UncheckedIOException;
 import java.math.BigDecimal;
 import java.time.Instant;
 import java.util.Arrays;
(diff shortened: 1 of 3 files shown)


CometActions:

### Access Information
- **URL:** https://pr-7234.dev.comet.com
- **Namespace:** pr-7234
- **Version:** 2.1.2-7234-merge-2363
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7234%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7234`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemResultMapper.java
```diff
@@ -216,10 +219,25 @@ static List<EvaluatorItem> getEvaluators(Row row, RowMetadata rowMetadata) {
...
-                .map(s -> JsonUtils.readValue(s, EVALUATOR_LIST_TYPE))
+                .map(value -> parseEvaluators(value, row.get("id", String.class),
+                        row.get("dataset_id", String.class)))
                 .orElse(null);
     }
 
+    // A row whose mapping function throws is silently dropped by the ClickHouse r2dbc driver,
+    // making the whole item vanish from the list. Degrade to null only on unparseable JSON
+    // (UncheckedIOException is what JsonUtils.readValue throws) so the item stays visible without
+    // its assertions; any other exception propagates instead of masking an unrelated bug.
```
andrescrz, line 230: Nit: these comments better as Javadoc on top of the method/field.

andrescrz (approved):
LGTM.

PR #7236, opened 2026-06-24
[OPIK-7104] [BE] refactor: remove stats computation path from /projects/retrieve

Nimrod007:
## Details

Follow-up to OPIK-7101 (#7231), which made project stats **opt-in** on `POST /v1/private/projects/retrieve` via an `includeStats` flag (default `false`). That stopped the heavy ClickHouse aggregations (feedback scores, trace/span aggregation, duration, cost, counts) from running on the hot SDK/UI name-resolution path, which had been saturating ClickHouse CPU in prod.

This PR completes the cleanup: `/retrieve` becomes **pure name → id resolution** and the stats code path is deleted, so the tech debt can't be re-triggered by a caller passing `includeStats: true`.

- `ProjectRetrieve`: drop the `includeStats` flag + accessor; back to `name`-only.
- `ProjectService.retrieveByName(String)`: drop the `includeStats` branch and the `enrichWithStats(...)` method (the ClickHouse aggregation + `block()`).
- `ProjectsResource`: call `retrieveByName(name)`.
- `ProjectsResourceTest`: collapse the two retrieve tests into one asserting the project is returned with all stats fields `null` even when traces/spans exist.

The response contract is unchanged: the stat fields are `@Nullable` and simply serialize as `null`. No first-party caller reads them from this endpoint (TS/Python SDK and FE `useProjectByName` read `.id` only) — verified during OPIK-7101.

## ⚠️ Merge gate

Per the OPIK-7104 plan, **do not merge until** the OPIK-7101 opt-in default has baked in prod and ClickHouse load is confirmed back to normal with no client relying on `includeStats: true`. Opening as **draft** until then.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves OPIK-7104
- Follow-up to OPIK-7101

## Testing

- `ProjectsResourceTest.RetrieveProjectTest`: single happy-path test seeds traces/spans via `buildProjectStats`, then asserts `/retrieve` returns the project with every stats field null (full-object recursive comparison, `lastUpdatedTraceAt` ignored as a `projects`-table column).
- Local: `mvn test-compile spotless:check` passes. Relying on PR CI for the full integration suite (needs Docker/Testcontainers).

## Documentation

No public docs change. Removing the optional `includeStats` request field is backward compatible (unknown JSON properties are ignored by the request deserializer); the OpenAPI/Fern spec will drop `includeStats` from the `ProjectRetrieve` schema on the next codegen.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/ProjectRetrieve.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/ProjectRetrieve.java
@@ -1,24 +1,8 @@
 package com.comet.opik.api;
 
-import jakarta.annotation.Nullable;
 import jakarta.validation.constraints.NotBlank;
 import lombok.Builder;
 
 @Builder
-public record ProjectRetrieve(@NotBlank String name, @Nullable Boolean includeStats) {
-
-    /**
-     * Whether the response should be enriched with project-level statistics
-     * (feedback scores, trace/span aggregations, duration, cost, counts).
-     * <p>
-     * Defaults to {@code false}: callers that only need to resolve a project
-     * name to its id (the SDK and the UI redirect path) should not trigger the
-     * expensive ClickHouse aggregations. The canonical accessor is overridden to
-     * coalesce {@code null} to {@code false}, so it never returns {@code null}.
-     * See OPIK-7101.
-     */
-    @Override
-    public Boolean includeStats() {
-        return Boolean.TRUE.equals(includeStats);
-    }
+public record ProjectRetrieve(@NotBlank String name) {
 }
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/ProjectsResource.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/ProjectsResource.java
@@ -216,7 +216,7 @@ public class ProjectsResource {
             @RequestBody(content = @Content(schema = @Schema(implementation = ProjectRetrieve.class))) @Valid ProjectRetrieve retrieve) {
         String workspaceId = requestContext.get().getWorkspaceId();
         log.info("Retrieve project by name '{}', on workspace_id '{}'", retrieve.name(), workspaceId);
-        Project project = projectService.retrieveByName(retrieve.name(), retrieve.includeStats());
+        Project project = projectService.retrieveByName(retrieve.name());
         log.info("Retrieved project id '{}' by name '{}', on workspace_id '{}'", project.id(), retrieve.name(),
                 workspaceId);
         return Response.ok().entity(project).build();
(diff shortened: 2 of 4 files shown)


Nimrod007:
Closing this — measured in prod, the cleanup isn't needed.

This PR was opened as a draft behind a merge gate: *"don't merge until the OPIK-7101 opt-in default has baked in prod and ClickHouse load is confirmed back to normal with no client relying on `includeStats: true`."* That gate has now been checked against prod rather than assumed, and the conclusion is that there's nothing left to fix.

### The endpoint is not slow

`POST /v1/private/projects/retrieve`, SaaS prod, last 24h, 1,651 requests (nginx edge access logs):

| metric | value |
|---|---|
| p50 | **57 ms** |
| p90 | 97 ms |
| p99 | 480 ms |
| max | 1.61 s (one request) |
| >500 ms | 12 / 1,651 (0.7%) |

That's the shape of a MySQL name→id lookup. For contrast, the stats aggregation itself (`get_trace_stats_traces_spans_by_project_ids` in `system.query_log`) measures avg **548 ms**, p95 **1.8 s**, max **4.5 s**. If `/retrieve` were still driving it, that would be plainly visible in the latency distribution. It isn't.

### Nobody is passing `includeStats: true`

Two checks, since neither is conclusive alone:

1. **Response bodies cap at 338 bytes** (p50 191, p99 306) — consistent with a bare project object. Not proof by itself: a project with no traces returns a small body even with stats enabled, while still paying for the ClickHouse query.
2. **The slow tail has the same body sizes as the fast path.** Of the 43 requests over 300 ms, mean body size is **196.5 bytes vs 201.2 bytes** for the sub-300 ms population — slightly *smaller*. Latency is entirely uncorrelated with response content, so the tail is generic (auth, MySQL, GC), not stats computation.

I also found 196 single-project stats queries in ClickHouse today and initially suspected them, but they're attributable to `/v1/private/projects/stats` called without `from_time` — that path and `enrichWithStats` emit byte-identical SQL, so ClickHouse alone can't separate them. The access-log latency is what settles it.

OPIK-7101 shipped in **2.1.1**; prod runs **2.2.74**, so this has had ~3 months of bake time.

### Why close rather than merge

The remaining argument for merging was that `include_stats` is a published, autocompletable parameter on the generated Python/TS SDK clients, so a caller could rediscover the path that saturated ClickHouse CPU. That's true in principle, but it's been exposed for three months with zero uptake, so the risk being priced in is empirically near zero — while removing the parameter is a signature-breaking change to the generated SDKs.

Net: no latency win, no CPU win, and a small backward-compatibility cost. The `includeStats: false` default from #7231 is sufficient on its own. Recording this as **risk accepted** rather than leaving a stale draft implying phase 2 is still pending.

If the flag ever does start getting used, this branch still applies cleanly apart from one trivial conflict in `ProjectService.java` (delete `enrichWithStats`; `main` has since edited it via OPIK-8328/8375).

**Caveats:** figures are SaaS prod (`www.comet.com`) over a 24h window; ClickHouse `query_log` numbers are from one replica. STSaaS wasn't measured, but the flag defaults `false` everywhere.

PR #7252, opened 2026-06-25
[OPIK-7118] [BE] perf: make attachment base64 detection linear (possessive match)

YarivHashaiComet:
## Details

Make attachment base64 detection linear. `AttachmentStripperService.processTextNode` used
`Pattern.compile("([A-Za-z0-9+/]{minBase64Size,}={0,2})")` + `find()`; keeping the size
threshold inside the pattern made the engine re-scan from every offset on large payloads
(O(n·runlen)), saturating CPU in prod.
- Match each maximal base64 run **possessively** (`[A-Za-z0-9+/]++`) and apply the size
  threshold in code. A possessive run is consumed once with no backtracking → linear,
  regardless of payload size.
- Detection is identical to the old pattern: maximal contiguous `[A-Za-z0-9+/]` run ≥
  `stripMinSize` plus up to two trailing `=` (any other char, including line-wrapping
  whitespace, ends a run). Wrapped / data-URI base64 is left intact, as before.
- Drops the per-`find()` OpenTelemetry span; keeps the `stripMinSize` floor, no ceiling.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- OPIK-7118

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: implemented the possessive-match detector + `AttachmentStripperServiceTest` cases;
  investigation, benchmarking, and the real-data + fuzz differential equivalence checks.
- Human verification: directed and reviewed by @YarivHashaiComet; ran the unit suite and a
  253-payload real-data differential before merge.

## Testing

- `cd apps/opik-backend && mvn spotless:apply && mvn test -Dtest=AttachmentStripperServiceTest` — 15/15.
- Scenarios: contiguous strip; size-threshold boundary (run at `minBase64Size-1` vs
  `minBase64Size`); padding capture 0 / 1 / 2 `=`, incl. a stray third `=` left intact;
  no-bridge-across-whitespace (captured payload round-trips via strict RFC 4648); wrapped base64
  left intact; large non-base64 and space-separated negatives; OOXML (`x-tika-ooxml`).
- Differential equivalence — old regex vs the possessive matcher: **1.8M** randomized cases
  across 6 thresholds, plus **253** real prod payloads (≥ 256 KB, 45 workspaces) and the edge
  battery → **0 divergences**. The 79 real fields that exceeded a 5 s bound on the old regex are
  handled in ≤ 234 ms by the new matcher.
- Not run locally: full module integration suite (Testcontainers) — runs in CI.

## Documentation

No documentation changes — behavior-preserving internal performance fix.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/attachment/AttachmentStripperService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/attachment/AttachmentStripperService.java
@@ -90,11 +90,13 @@ public class AttachmentStripperService {
     // Apache Tika for MIME type detection
     private static final Tika tika = new Tika();
 
-    // Maximal run of base64 alphabet chars (RFC 4648), matched POSSESSIVELY so each run is consumed
-    // once with no backtracking. The size threshold is applied in code (see processTextNode), not in
-    // the pattern: the old {minBase64Size,} regex kept it inside, so a long sub-threshold run made
-    // find() retry from every offset -- an O(n*runlen) re-scan that pinned CPU on large payloads
-    // (OPIK-7118). Matching the run possessively and checking its length afterwards is linear.
+    /**
+     * Maximal run of base64 alphabet chars (RFC 4648), matched POSSESSIVELY so each run is consumed
+     * once with no backtracking. The size threshold is applied in code (see processTextNode), not in
+     * the pattern: the old {@code {minBase64Size,}} regex kept it inside, so a long sub-threshold run
+     * made find() retry from every offset -- an O(n*runlen) re-scan that pinned CPU on large payloads
+     * (OPIK-7118). Matching the run possessively and checking its length afterwards is linear.
+     */
     private static final Pattern BASE64_RUN = Pattern.compile("[A-Za-z0-9+/]++");
 
     private final Tracer tracer;
@@ -508,47 +510,48 @@ public class AttachmentStripperService {
         // '=' padding chars in code -- identical detection to the old {minBase64Size,}={0,2} regex, but
         // linear instead of a CPU-pinning re-scan (see BASE64_RUN, OPIK-7118). Any non-alphabet char
         // (including line-wrapping whitespace) ends a run, so MIME/data-URI base64 is left intact.
-        // Wrapped in a single span per text node (not per find()) so trace latency stays attributable
-        // to detection vs the child base64.decode / tika.detect / eventBus.post spans.
-        return wrapWithSpan("base64.detect", () -> {
-            StringBuilder result = new StringBuilder();
-            int lastEnd = 0;
-            boolean foundAttachment = false;
-            int length = text.length();
+        final int length = text.length();
+        StringBuilder result = new StringBuilder();
+        int lastEnd = 0;
+        boolean foundAttachment = false;
 
-            Matcher matcher = BASE64_RUN.matcher(text);
-            while (matcher.find()) {
-                int start = matcher.start();
-                int runEnd = matcher.end();
-                if ((long) runEnd - start < minBase64Size) {
-                    continue;
-                }
-
-                int end = runEnd;
-                while (end < length && end - runEnd < 2 && text.charAt(end) == '=') {
-                    end++;
-                }
-
-                String base64Data = text.substring(start, end);
-                int currentAttachmentNumber = attachmentCounter.incrementAndGet();
-                String attachmentReference = processBase64Attachment(base64Data, currentAttachmentNumber, ctx);
-                if (attachmentReference != null) {
-                    result.append(text, lastEnd, start);
-                    result.append(attachmentReference);
-                    lastEnd = end;
-                    foundAttachment = true;
-                }
+        // Keep the original regex.match / regex.find span names so detection latency stays comparable
+        // before and after the linearization fix.
+        Matcher matcher = wrapWithSpan("regex.match", () -> BASE64_RUN.matcher(text));
+        while (wrapWithSpan("regex.find", matcher::find)) {
+            int start = matcher.start();
+            int runEnd = matcher.end();
+            if ((long) runEnd - start < minBase64Size) {
+                continue;
             }
 
-            // If we found and replaced attachments, return new node with modified text
-            if (foundAttachment) {
-                result.append(text, lastEnd, text.length());
-                return objectMapper.getNodeFactory().textNode(result.toString());
+            // Capture up to two trailing '=' padding chars: base64 pads the final group to a 4-char
+            // boundary, so 0, 1, or 2 '=' are valid. A third '=' is never valid padding and is left in
+            // the surrounding text.
+            int end = runEnd;
+            while (end < length && end - runEnd < 2 && text.charAt(end) == '=') {
+                end++;
             }
 
-            // No attachments found, return original node
-            return node;
-        });
+            String base64Data = text.substring(start, end);
+            int currentAttachmentNumber = attachmentCounter.incrementAndGet();
+            String attachmentReference = processBase64Attachment(base64Data, currentAttachmentNumber, ctx);
+            if (attachmentReference != null) {
+                result.append(text, lastEnd, start);
+                result.append(attachmentReference);
+                lastEnd = end;
+                foundAttachment = true;
+            }
+        }
+
+        // If we found and replaced attachments, return new node with modified text
+        if (foundAttachment) {
+            result.append(text, lastEnd, length);
+            return objectMapper.getNodeFactory().textNode(result.toString());
+        }
+
+        // No attachments found, return original node
+        return node;
     }
 
     /**
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/attachment/AttachmentStripperService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/attachment/AttachmentStripperService.java
@@ -15,6 +15,7 @@ import com.google.inject.Singleton;
 import io.opentelemetry.api.GlobalOpenTelemetry;
 import io.opentelemetry.api.metrics.LongCounter;
 import io.opentelemetry.api.metrics.LongHistogram;
+import io.opentelemetry.api.metrics.LongUpDownCounter;
 import io.opentelemetry.api.metrics.Meter;
 import io.opentelemetry.api.trace.Tracer;
 import io.opentelemetry.context.Scope;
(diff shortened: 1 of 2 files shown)


BorisTkachenko (dismissed):
LGTM!

apps/opik-backend/src/main/java/com/comet/opik/domain/attachment/AttachmentStripperService.java
```diff
@@ -72,14 +73,31 @@ public class AttachmentStripperService {
...
+     * a unit of work completes), this gauge stays elevated while work is in flight, so it surfaces a
+     * stuck or saturated scan in real time.
+     */
+    private final LongUpDownCounter stripInProgress;
+    /**
+     * Length distribution of candidate fields entering detection, in characters (UTF-16 code units).
+     * Recorded before scanning, so it is captured even when a field is slow to process — a leading
+     * indicator for oversized payloads.
+     */
+    private final LongHistogram inputSizeChars;
```
andrescrz, line 88: Not a blocker: double check what's the most suitable metric type here: a gauge or a histogram.

BorisTkachenko (dismissed):

apps/opik-backend/src/main/java/com/comet/opik/domain/attachment/AttachmentStripperService.java
```diff
@@ -72,14 +73,31 @@ public class AttachmentStripperService {
...
 
     // Apache Tika for MIME type detection
     private static final Tika tika = new Tika();
 
-    private final Tracer tracer;
+    // Maximal run of base64 alphabet chars (RFC 4648), matched POSSESSIVELY so each run is consumed
+    // once with no backtracking. The size threshold is applied in code (see processTextNode), not in
+    // the pattern: the old {minBase64Size,} regex kept it inside, so a long sub-threshold run made
+    // find() retry from every offset -- an O(n*runlen) re-scan that pinned CPU on large payloads
+    // (OPIK-7118). Matching the run possessively and checking its length afterwards is linear.
```
andrescrz: Nit: these comments better as Javadoc on top of the field, method etc.

apps/opik-backend/src/main/java/com/comet/opik/domain/attachment/AttachmentStripperService.java
```diff
@@ -472,39 +500,55 @@ private JsonNode processTextNode(
...
+        // to the metric even if the scan that follows is slow to complete.
+        inputSizeChars.record(text.length());
+
+        // Find each maximal base64 run, then apply the size threshold and collect up to two trailing
+        // '=' padding chars in code -- identical detection to the old {minBase64Size,}={0,2} regex, but
+        // linear instead of a CPU-pinning re-scan (see BASE64_RUN, OPIK-7118). Any non-alphabet char
+        // (including line-wrapping whitespace) ends a run, so MIME/data-URI base64 is left intact.
+        // Wrapped in a single span per text node (not per find()) so trace latency stays attributable
+        // to detection vs the child base64.decode / tika.detect / eventBus.post spans.
+        return wrapWithSpan("base64.detect", () -> {
```
andrescrz: If we really want to monitor how these spans improved after the fix, you should keep the old name `regex.match`.

apps/opik-backend/src/main/java/com/comet/opik/domain/attachment/AttachmentStripperService.java
```diff
@@ -472,39 +500,55 @@ private JsonNode processTextNode(
...
+            StringBuilder result = new StringBuilder();
+            int lastEnd = 0;
+            boolean foundAttachment = false;
+            int length = text.length();
+
+            Matcher matcher = BASE64_RUN.matcher(text);
+            while (matcher.find()) {
```
andrescrz: Same here, for observability reasons you like want to keep the previous span `regex.find`.

apps/opik-backend/src/main/java/com/comet/opik/domain/attachment/AttachmentStripperService.java
```diff
@@ -472,39 +500,55 @@ private JsonNode processTextNode(
...
+                if ((long) runEnd - start < minBase64Size) {
+                    continue;
+                }
 
-            // Try to process as attachment
-            String attachmentReference = processBase64Attachment(base64Data, currentAttachmentNumber, ctx);
+                int end = runEnd;
+                while (end < length && end - runEnd < 2 && text.charAt(end) == '=') {
+                    end++;
+                }
```
andrescrz: Minor: this loop won't run more than twice, but it can be a code smell if not that familiar, you could go with an equivalent logic with a small calculation, instead of a loop. Either way, better add some comments to explain this block.

apps/opik-backend/src/main/java/com/comet/opik/domain/attachment/AttachmentStripperService.java
```diff
@@ -472,39 +500,55 @@ private JsonNode processTextNode(
...
+        // '=' padding chars in code -- identical detection to the old {minBase64Size,}={0,2} regex, but
+        // linear instead of a CPU-pinning re-scan (see BASE64_RUN, OPIK-7118). Any non-alphabet char
+        // (including line-wrapping whitespace) ends a run, so MIME/data-URI base64 is left intact.
+        // Wrapped in a single span per text node (not per find()) so trace latency stays attributable
+        // to detection vs the child base64.decode / tika.detect / eventBus.post spans.
+        return wrapWithSpan("base64.detect", () -> {
+            StringBuilder result = new StringBuilder();
+            int lastEnd = 0;
+            boolean foundAttachment = false;
+            int length = text.length();
```
andrescrz: Nit: if you us an aux var for `length`, better use it everywhere. Also, better annotate as `final` to make it clear.

apps/opik-backend/src/test/java/com/comet/opik/domain/attachment/AttachmentStripperServiceTest.java
```diff
@@ -253,4 +257,195 @@ void shouldGenerateContextAwareFilenames() throws Exception {
...
+
+        // Then - the OOXML doc is detected as a real file and stripped
+        assertThat(result.get("doc").asText()).matches("\\[input-attachment-1-\\d+\\..+\\]");
+        verify(eventBus, times(1)).post(any());
+    }
+
+    private static byte[] minimalOoxmlPackage() throws java.io.IOException {
+        var bos = new java.io.ByteArrayOutputStream();
+        try (var zip = new java.util.zip.ZipOutputStream(bos)) {
+            zip.putNextEntry(new java.util.zip.ZipEntry("[Content_Types].xml"));
```
andrescrz: Nit: use java imports.

apps/opik-backend/src/test/java/com/comet/opik/domain/attachment/AttachmentStripperServiceTest.java
```diff
@@ -253,4 +257,195 @@ void shouldGenerateContextAwareFilenames() throws Exception {
...
+            zip.write(("<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"yes\"?>"
+                    + "<Types xmlns=\"http://schemas.openxmlformats.org/package/2006/content-types\">"
+                    + "<Default Extension=\"rels\" ContentType=\"application/vnd.openxmlformats-package.relationships+xml\"/>"
+                    + "<Default Extension=\"xml\" ContentType=\"application/xml\"/></Types>")
```
andrescrz: Nit: usa java text blocks with triple quotes `"""`

andrescrz (commented):
I recommend keeping the old spans for observability before moving forward. The rest generally LGTM, with some minor comments.

apps/opik-backend/src/main/java/com/comet/opik/domain/attachment/AttachmentStripperService.java
thiagohora: Done in bd28f219 — restored the original `regex.match` span name (reverted the single `base64.detect` span) so before/after timings stay comparable.

apps/opik-backend/src/main/java/com/comet/opik/domain/attachment/AttachmentStripperService.java
thiagohora: Done in bd28f219 — `regex.find` restored as well, same reason.

apps/opik-backend/src/main/java/com/comet/opik/domain/attachment/AttachmentStripperService.java
thiagohora: Done — `length` is now `final` and used everywhere (the trailing `result.append(text, lastEnd, length)` no longer re-calls `text.length()`).

apps/opik-backend/src/main/java/com/comet/opik/domain/attachment/AttachmentStripperService.java
thiagohora: Added a comment explaining it consumes 0–2 trailing `=` (base64 pads the final group to a 4-char boundary; a third `=` is left in the surrounding text). Kept the loop over a calc since it reads clearly with the comment — happy to switch to `Math.min`-style arithmetic if you prefer.

apps/opik-backend/src/main/java/com/comet/opik/domain/attachment/AttachmentStripperService.java
thiagohora: Done — `BASE64_RUN` comment is now Javadoc (the two new metric fields were already Javadoc).

apps/opik-backend/src/main/java/com/comet/opik/domain/attachment/AttachmentStripperService.java
thiagohora: Intentional: `strip.in_progress` is a concurrent in-flight count that goes up and down, so an UpDownCounter (gauge) is the right type; `input.size.chars` is a value distribution, so a histogram fits. Neither maps naturally to the other type.

apps/opik-backend/src/test/java/com/comet/opik/domain/attachment/AttachmentStripperServiceTest.java
thiagohora: Done — replaced the fully-qualified names with imports (`ByteArrayOutputStream`, `IOException`, `StandardCharsets`, `Random`, `ZipEntry`, `ZipOutputStream`).

apps/opik-backend/src/test/java/com/comet/opik/domain/attachment/AttachmentStripperServiceTest.java
thiagohora: Done — the OOXML XML fixtures now use text blocks (with trailing `\` so the XML stays single-line and Tika detection is unchanged; 15/15 tests still pass).

andrescrz (approved):
Thanks for addressing the feedback, LGTM.

PR #7262, opened 2026-06-26
[BE] fix: register azure as a canonical provider so Azure model prices load

Anuj7411:
## Details

`PROVIDERS_MAPPING` currently registers `microsoft -> azure`, but no entry in `model_prices_and_context_window.json` actually uses `microsoft` as its `litellm_provider` — the mapping was added to handle a single override row (`azure/multilingual-e5-large`). Meanwhile **199 entries** in the main JSON use the literal `azure` `litellm_provider` (the full `azure/gpt-4o*`, `azure/gpt-5*`, `azure/codex-mini`, `azure/command-r-plus` lineup, all `azure/global/*` and `azure/eu/*` regional variants, and the `azure/gpt-4o-*-realtime-preview` / `azure/gpt-realtime*` families). Every one of those is silently dropped at load time because `buildModelPrice` returns `null` when the provider is not in `PROVIDERS_MAPPING`, so cost for Azure-hosted requests comes back as zero.

Register `azure -> azure` alongside the existing `microsoft -> azure` entry (the latter is kept so the existing override row keeps working without a JSON migration). Route the new `azure` provider through the existing OpenAI cache calculator, since Azure-hosted OpenAI models share OpenAI's usage shape (`prompt_tokens_details.cached_tokens` flattened under `original_usage.*`). With #7253 in the same area now extending that cache calculator to split audio tokens, this PR makes the `azure/...*realtime-preview` families correctly billable end-to-end.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code (CLI)
- Model(s): Claude Opus 4.7
- Scope: assisted in scoping the bug (counting unmapped providers in `model_prices_and_context_window.json`), drafting the two-line plumbing change, and writing the parameterized branch-coverage test
- Human verification: ran the test suite locally; verified that no name-collision is introduced (the single `microsoft`-tagged row in the override file does not overlap with any of the 199 azure-tagged rows in the main JSON); cross-checked the expected dollar values against the prices in `model_prices_and_context_window.json`

## Testing

Local verification on JDK 25 + Maven 3.9.9:

```
mvn -ntp -f apps/opik-backend/pom.xml \
  -Dtest=CostServiceTest,SpanCostCalculatorTest test
```

Result: **84/84 green** (2 new parameterized cases added under `calculateCostHandlesAzureHostedModels`, matching the style established in #7137 / #7253).

Scenarios covered:

1. **Plain text-generation route** — `azure/command-r-plus` (input `3e-6`, output `1.5e-5`, no cache rates configured). With `prompt_tokens=1000, completion_tokens=200`, expected `0.006` (`1000 * 3e-6 + 200 * 1.5e-5`). Exercises the path where azure is registered but `resolveCalculator` falls through past the cache check to `textGenerationCost`.
2. **Cache-aware route via OpenAI calc** — `azure/gpt-4.1` (input `2e-6`, output `8e-6`, `cache_read 5e-7`). With `original_usage.prompt_tokens=1000, original_usage.completion_tokens=200, original_usage.prompt_tokens_details.cached_tokens=300`, expected `0.00315` (non-cached input `700 * 2e-6 + 200 * 8e-6 + 300 * 5e-7`). Exercises the new `PROVIDERS_CACHE_COST_CALCULATOR` entry.

The pre-existing `calculateCost_overrideMicrosoftProviderMapsToAzure` test continues to pass, confirming the override-file workaround is not regressed.

## Documentation

No docs change. Behavior change is invisible to end users beyond accurate (non-zero) cost numbers for Azure-hosted traces on 199 previously-dropped models.


andrescrz (approved):
Hi @Anuj7411 

Best

PR #7266, opened 2026-06-28
fix: isolate ADK tracer model output per invocation_id

### LINKED ISSUE #7265: [Bug]: ADK integration — OpikTracer._last_model_output is per-instance, leaking output across concurrent sessions

- [x] Opik Python SDK
- [ ] Opik TypeScript SDK
- [ ] Opik UI
- [ ] Opik Server
- [ ] Documentation

### Opik version

`1.10.25` — also present on `main` as of 2026-06-28.

### Describe the problem

When a single `OpikTracer` is shared across **concurrent** ADK invocations — the documented `track_adk_agent_recursive(agent, tracer)` pattern — a trace for session A can be logged with **session B's answer** in its `output` field.

The trace *identity* and metadata are correct (thread/session id, tags, timing — these are isolated via `context_storage` / contextvars). Only the `output` value is wrong: it can belong to a different invocation that happened to be in flight at the same time.

This shows up under high-concurrency async serving (e.g. a FastAPI app issuing many `Runner.run_async(...)` SSE streams in one process), and is especially easy to hit with `ag-ui-adk`, which shallow-copies the agent per request and therefore shares the same tracer/callback objects by reference.

### Root cause

`OpikTracer` carries the most recent model output on a **per-instance** attribute rather than a per-context one. In `opik/integrations/adk/opik_tracer.py` (`1.10.25`):

- `self._last_model_output` is initialized on the instance — `_init_internal_attributes`, line 67
- it is **written** for every final model response in `after_model_callback` — line 363
- it is **read** in `after_agent_callback` and applied to the agent span/trace via `current_span.update(output=...)` / `current_trace.update(output=...)` — lines 188, 192/197
- and reset to `None` — line 201

Because the tracer instance is shared across all concurrent invocations, a concurrent invocation's `after_model_callback` (line 363) can overwrite `_last_model_output` in the window between another invocation's model call and its `after_agent_callback` (line 188). The wrong output is then stamped onto the first invocation's trace.

The span/trace objects read on lines 192/197 are already contextvar-isolated and correct — it is only the `output` payload, sourced from the shared instance attribute, that leaks across invocations.

### Steps to reproduce

1. Create a single `OpikTracer` and instrument an agent: `track_adk_agent_recursive(agent, tracer)`.
2. Drive two sessions concurrently whose final answers are easily distinguishable (e.g. different languages or a unique marker string per session):

```python
import asyncio

# one shared tracer (the documented pattern)
tracer = OpikTracer()
track_adk_agent_recursive(agent, tracer)

async def turn(session_id, message):
    async for _ in runner.run_async(user_id=session_id, session_id=session_id,
                                    new_message=message):
        pass

await asyncio.gather(
    turn("session-A", "answer with the single word: ALPHA"),
    turn("session-B", "answer with the single word: BRAVO"),
)
```

3. Inspect the two traces in Opik. Under load, one trace's `output` can show the *other* session's answer (e.g. session A's trace shows `BRAVO`). The behaviour is deterministic from the code path above but probabilistic to hit on any single run; tightening the timing (or raising concurrency) makes it reliable.

### Expected behavior

Each trace's `output` reflects that invocation's own last model output.

### Actual behavior

Under concurrency, a trace's `output` can be a different concurrent invocation's last model output. Trace identity/metadata remain correct, so the result is a correctly-attributed trace showing the wrong answer.

### Impact

Telemetry/observability only — the live response and any separate event sink are unaffected. But it silently corrupts the `output` shown in traces under load, which is easy to misread as a *model* problem (e.g. "the agent answered in the wrong language") when the model actually behaved correctly.

### Suggested fix

Scope the last model output per-invocation instead of per-instance. The needed isolation already exists in the same file — `after_agent_callback` already reads the contextvar-isolated `context_storage.top_span_data()` / `get_trace_data()`. A few options:

- apply the output to the per-context span/trace data directly in `after_model_callback`, rather than stashing it on `self`;
- key a small map by `invocation_id` / `session_id`;
- or stash it on `callback_context.state`, which both callbacks share per invocation.

### Environment

High-concurrency async serving (single process, many concurrent sessions), ADK `Runner.run_async`, one shared `OpikTracer`.

lupuletic:
## Summary

Fixes #7265.

A single `OpikTracer` is shared across all concurrent invocations of an instrumented agent (the documented `track_adk_agent_recursive` pattern). The most recent model output was stored on a plain instance attribute (`self._last_model_output`), so under concurrency one invocation's `after_model_callback` could overwrite it before another invocation's `after_agent_callback` read it — and a trace would be logged with a *different* concurrent invocation's answer.

Trace/span identity is already `contextvars`-isolated (via `context_storage`), so only the **output** payload leaked: a trace with correct identity/metadata but the wrong answer. This is telemetry-only (the live response is unaffected), but it silently corrupts trace output under load and is easily misread as a model issue.

## Fix

Key `_last_model_output` by ADK `invocation_id` in both `OpikTracer` and `LegacyOpikTracer`:

- `after_model_callback` writes `self._last_model_output[callback_context.invocation_id] = output`
- `after_agent_callback` reads with `.get(invocation_id)` on the **span path** (a sub-agent finished — keep the entry, the root still consumes it) and `.pop(invocation_id, None)` on the **trace path** (the root finished — drop it)

This preserves the original "reset only at the root agent" semantics, so multi-agent invocations are unaffected, while isolating concurrent invocations. Both `__setstate__` paths already call `_init_internal_attributes()`, so the dict survives pickling / deep-copy.

## Test

Adds `tests/library_integration/adk/test_adk_concurrency.py` — deterministic, model-free tests (no `GOOGLE_API_KEY` required):

- `test_after_agent_callback__concurrent_invocations__output_is_not_mixed` — two interleaved invocations each receive their own output.
- `test_after_agent_callback__nested_agent__keeps_output_for_root` — a sub-agent (span path) does not consume the output, so the root (trace path) still receives it.

`ruff check` and `ruff format --check` pass on the changed files.

### DIFF
--- a/sdks/python/src/opik/integrations/adk/helpers.py
+++ b/sdks/python/src/opik/integrations/adk/helpers.py
@@ -34,6 +34,28 @@ def get_adk_provider() -> opik_types.LLMProvider:
     )
 
 
+def drop_invocation_output_if_finished(
+    open_agents: Dict[str, int],
+    last_model_output: Dict[str, Any],
+    invocation_id: str,
+) -> None:
+    """Decrement the open-agent count for ``invocation_id`` and, once the
+    invocation's outermost agent has finished (the count reaches zero), drop its
+    cached model output.
+
+    A single tracer instance is shared across concurrent invocations, so the
+    cached output is keyed by ``invocation_id``. Counting open agents (rather
+    than keying cleanup off span-vs-trace) keeps it correct under
+    ``distributed_headers``, where the root agent is itself a span.
+    """
+    remaining = open_agents.get(invocation_id, 1) - 1
+    if remaining > 0:
+        open_agents[invocation_id] = remaining
+    else:
+        open_agents.pop(invocation_id, None)
+        last_model_output.pop(invocation_id, None)
+
+
 def has_empty_text_part_content(llm_response: LlmResponse) -> bool:
     try:
         if llm_response.content is None:
--- a/sdks/python/src/opik/integrations/adk/helpers.py
+++ b/sdks/python/src/opik/integrations/adk/helpers.py
@@ -34,28 +34,6 @@ def get_adk_provider() -> opik_types.LLMProvider:
     )
 
 
-def drop_invocation_output_if_finished(
-    open_agents: Dict[str, int],
-    last_model_output: Dict[str, Any],
-    invocation_id: str,
-) -> None:
-    """Decrement the open-agent count for ``invocation_id`` and, once the
-    invocation's outermost agent has finished (the count reaches zero), drop its
-    cached model output.
-
-    A single tracer instance is shared across concurrent invocations, so the
-    cached output is keyed by ``invocation_id``. Counting open agents (rather
-    than keying cleanup off span-vs-trace) keeps it correct under
-    ``distributed_headers``, where the root agent is itself a span.
-    """
-    remaining = open_agents.get(invocation_id, 1) - 1
-    if remaining > 0:
-        open_agents[invocation_id] = remaining
-    else:
-        open_agents.pop(invocation_id, None)
-        last_model_output.pop(invocation_id, None)
-
-
 def has_empty_text_part_content(llm_response: LlmResponse) -> bool:
     try:
         if llm_response.content is None:
--- a/sdks/python/src/opik/integrations/adk/legacy_opik_tracer.py
+++ b/sdks/python/src/opik/integrations/adk/legacy_opik_tracer.py
@@ -137,7 +137,7 @@ class LegacyOpikTracer:
             )
 
             trace_metadata = self.metadata.copy()
-            trace_metadata["adk_invocation_id"] = callback_context.invocation_id
+            trace_metadata["adk_invocation_id"] = invocation_id
             trace_metadata.update(session_metadata)
 
             _try_add_agent_graph_to_metadata(trace_metadata, callback_context)
@@ -187,8 +187,8 @@ class LegacyOpikTracer:
         *args: Any,
         **kwargs: Any,
     ) -> None:
+        invocation_id = callback_context.invocation_id
         try:
-            invocation_id = callback_context.invocation_id
             output = self._last_model_output.get(invocation_id)
 
             if (span_data := self._context_storage.top_span_data()) is not None:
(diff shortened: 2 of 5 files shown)


lupuletic:
Pushed a redesign that addresses the review feedback (the bots here plus a deeper review pass). Summary of what changed and why.

### The problem with my first approach

I'd added an `_open_agents` depth counter to know when an invocation's outermost agent finished, so I could drop its cached output — including under `distributed_headers`, where the root agent is a span rather than a trace. But that counter can leak without bound: **ADK skips `after_agent_callback` on agent errors, early escalation/`end_invocation`, and stream cancellation**, so the decrement never runs and the per-invocation maps grow forever on a long-lived shared tracer. That's strictly worse than the bug it was fixing — and it's exactly the "early-abort leaks invocation cache" that the Baz review flagged.

### The fix: a bounded cache instead of a counter

`_last_model_output` is now a small `LastModelOutputCache` (`output_cache.py`) — a thread-safe, size-bounded `OrderedDict` (evict-oldest), keyed by `invocation_id`, shared by both tracers:

- **No unbounded growth** — eviction caps memory regardless of whether `after_agent_callback` ever runs. This is the eviction path the review asked for, and it also dissolves the `distributed_headers` keep-leak the Codex review raised (no span-vs-trace special-casing needed).
- **No counter / no lifecycle / no `finally`** — which removes the whole cluster of counter-related issues (premature pop on desync, non-atomic read-modify-write, `__setstate__` resetting in-flight counts).
- **Thread-safe** — writes are guarded by a `threading.Lock` (mirrors the SDK's existing `PromptCache`), so a sync/threaded runner can't corrupt the `OrderedDict`.
- **Pickle-safe** — the cache holds a `Lock` (unpicklable), so `__getstate__` excludes it and `__setstate__` recreates a fresh one (fixes the `dictionary changed size during iteration` / inconsistent-serialization concern).
- **No None-clobber** — `after_model_callback` only caches a real (non-`None`) output, so a failed conversion can't overwrite a good cached value (the modern tracer now matches legacy here).
- **`after_agent_callback` is read-only again** — `invocation_id` stays inside the `try`, so a tracer-side error can't escape into the user's agent.

### Tests

Now unit-test the cache directly (keying isolation + bounded eviction) and exercise the real `after_agent_callback` read path, instead of hand-seeding a counter. I verified the cache logic locally (keying, eviction, recency, and an 8-thread × 2000-op concurrency run with no corruption and the bound held); the ADK integration tests run in CI.

### Legacy CI

The modern-tracer tests are gated with `@pytest_skip_for_adk_older_than_1_3_0` (older ADK resolves `OpikTracer` to `LegacyOpikTracer`); the cache unit tests are version-agnostic and run on all matrices.

### Known limitation (pre-existing, not introduced here)

Output is keyed by `invocation_id`, so concurrent sub-agents of a single `ParallelAgent` (which share one `invocation_id`) can still overwrite each other's slot within that turn — same as before this PR. Cross-*invocation* isolation (the reported bug) is fixed. Happy to tighten the parallel-sub-agent case in a follow-up if you'd like it in scope.

Thanks for the thorough automated review — it caught a genuinely worse failure mode in my first cut.

lupuletic:
Good catches on the stale-output paths (both the legacy no-span early return and the modern conversion-failure guard) — fixed in e08ae46.

Rather than clearing at each individual failure spot, both tracers now **clear the invocation's cache entry up front** once we commit to processing a final (non-partial) response, and re-set it only if conversion succeeds. So a missing span, a failed `convert_adk_base_model_to_dict()`/usage extraction, *or* a later error all leave no stale value for `after_agent_callback` to stamp — it reads `None` instead. The partial-chunk and empty-text paths return before that point, so they still (correctly) keep the prior value while ADK streams the rest.

Added `LastModelOutputCache.discard()` plus tests for `discard` and for `after_agent_callback` stamping `None` when nothing is cached.

alexkuzmik (approved):
Hi @lupuletic! Good catch, thank you.

PR #7277, opened 2026-06-29
[OPIK-7103] feat: {{trace}} + {{span}} agentic attachment evaluation for online LLM-as-judge

petrotiurin:
## Details

Lets online LLM-as-judge rules **load and inspect attachments** (images, etc.) while scoring, by introducing two reserved template variables that switch the judge into an agentic tool-loop (`get_attachment` / `read` / `jq`), gated behind the `agentic_tools` toggle and a tool-calling provider; otherwise evaluation stays on the existing inline path.

- **`{{trace}}`** (trace-level): injects the trace structure — trace id, span ids, and trace- and span-level attachment `file_name`s — so the judge can fetch any attachment in the trace.
- **`{{span}}`** (span-level): injects the span structure (span id + the span's own attachment `file_name`s), scoped to that single span.

Notable backend points:
- `OnlineScoringEngine` recognizes the `{{trace}}`/`{{span}}` sentinels and injects the pre-built structure; the near-duplicate `{{spans}}`/`{{trace}}`/`{{span}}` detection + injection is collapsed into shared generic helpers.
- The trace- and span-level scorers build the structure reactively and run the shared agentic tool-loop (`OnlineScoringBaseScorer.runToolCallLoop`); `TraceToolContext.forActiveSpan` roots the loop at a span.
- Self-describing attachment entries (`type` + `id` + `file_name`) so the judge copies `get_attachment` args verbatim instead of inferring the owner.
- Upload-race tolerance (`listAttachmentsToleratingUploadRace` / `listSpanAttachmentsToleratingUploadRace`): retry until the persistent copy is visible, drop transient auto-stripped copies, fall back to the transient copy if no persistent one lands; failures are logged once and degrade to none rather than blocking scoring.
- The `{trace, spans}` composite is serialized once per `{{trace}}` run and threaded through to the size estimate / tool-cache pre-seed.

Frontend (v2 only): `{{trace}}`/`{{span}}` are treated as reserved sentinels so they aren't surfaced as user-mappable variables.

https://github.com/user-attachments/assets/5ea03245-7689-468a-bd34-49ba73ea8287

https://github.com/user-attachments/assets/58adcf26-e64e-41b6-8707-687a291ba360

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves OPIK-7103
- Resolves OPIK-6555

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: Implementation, refactor (shared base-scorer orchestration), tests, and PR review fixes for the `{{trace}}`/`{{span}}` agentic attachment-evaluation feature.
- Human verification: Author reviewed all changes; verified end-to-end against live vision traces at both trace and span levels (image-bearing entity loads its attachment with zero failed fetches); backend unit/integration tests run and green; `spotless` clean; rebased onto latest `main` and recompiled.

## Testing

- New `OnlineScoringSpanLlmAsJudgeScorerTest` (agentic path, attachment-fetch error, inline fallback, upload-race retry, transient-vs-persistent preference, transient fallback) plus additions to `OnlineScoringEngineTest`, `OnlineScoringLlmAsJudgeScorerTest`, `ReadToolTest`, `TraceCompressorTest`.
- Backend build/tests: `mvn test -pl apps/opik-backend` for the affected suites; `spotless` clean.
- Verified end-to-end against live vision traces at both levels: the image-bearing entity loads its attachment (`loaded:true`), with zero failed fetches once both robustness fixes are in place.
- Rebased onto latest `main` and recompiled (`mvn compile`) against main's langchain4j bump — BUILD SUCCESS.

## Documentation

No public documentation changes. The `{{trace}}`/`{{span}}` reserved variables are surfaced in the v2 LLM-judge rule editor; user-facing docs for the agentic attachment-evaluation feature can follow once the `agentic_tools` toggle is enabled by default.


apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringLlmAsJudgeScorer.java
```diff
@@ -180,22 +183,21 @@ protected Mono<Void> score(@NonNull TraceToScoreLlmAsJudge message) {
...
-                : Mono.just(List.of());
+        // Presence of the {{trace}} variable is the declarative agentic trigger: it injects the
+        // trace structure (trace id, span ids, attachment file_names) into the prompt so the judge
+        // can call get_attachment with real ids instead of guessing. Gated by the agentic-tools
+        // toggle, same as the size-based path.
+        boolean referencesTrace = serviceTogglesConfig.isAgenticToolsEnabled()
+                && OnlineScoringEngine.templateReferencesTraceStructure(
+                        message.llmAsJudgeCode().messages(),
+                        message.llmAsJudgeCode().variables(),
+                        message.promptType());
```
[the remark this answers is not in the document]
petrotiurin: This is expected and was a part of the original request. Since the feature is not yet released this is a safe change with no regression. `{{trace}}` is safer than auto agentic path since agentic eval can be pricier than regular so it's better to have user explicitly state this. Also `{{trace}}` eliminates some of the tool calls the agent would need to do anyways.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringLlmAsJudgeScorer.java
```diff
@@ -207,16 +212,91 @@ protected Mono<Void> score(@NonNull TraceToScoreLlmAsJudge message) {
...
+                .filter(Objects::nonNull)
+                .collect(Collectors.toSet());
+        if (spanIds.isEmpty()) {
+            return Mono.just(Map.of());
+        }
+        return attachmentService
+                .getAttachmentInfoByEntityIds(com.comet.opik.api.attachment.EntityType.SPAN, spanIds)
+                .doOnError(error -> log.warn(
+                        "Failed to list span attachments for trace '{}' (workspace '{}'); degrading to none",
+                        trace.id(), message.workspaceId(), error))
```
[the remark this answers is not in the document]
petrotiurin: Skipping — this is intentional. Span attachment gathering deliberately assumes spans (and their attachments) are already persisted: the plan is to add a fixed evaluation delay before dispatch so everything is settled, rather than per-span upload-race retries. The batched lookup still applies `preferPersistentAttachments` per span (so a transient auto-stripped copy is never surfaced) and logs+degrades on failure. `listAttachmentsToleratingUploadRace(...)` is single-entity and gated on that entity's own body referencing an attachment, so it isn't a clean fit for the batched, multi-span lookup here. Happy to revisit if the delay proves insufficient in practice.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSpanLlmAsJudgeScorer.java
```diff
@@ -98,40 +153,193 @@ protected Mono<Void> score(@NonNull SpanToScoreLlmAsJudge message) {
...
+                    scoreRequest = OnlineScoringEngine.prepareSpanLlmRequest(
+                            message.llmAsJudgeCode(), span,
+                            llmProviderFactory.getStructuredOutputStrategy(modelName));
+                    structuredRequest = scoreRequest;
+                }
             } catch (Exception exception) {
                 userFacingLogger.error("Error preparing LLM request for spanId '{}': \n\n{}",
                         span.id(), exception.getMessage());
                 throw exception;
             }
```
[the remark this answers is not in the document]
petrotiurin: Fixed in 330c8710bf — the catch block now passes the `Throwable` as the final SLF4J argument (`userFacingLogger.error("Error preparing LLM request for spanId '{}':", span.id(), exception)`) instead of interpolating `exception.getMessage()`, so the full stack trace is emitted.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringLlmAsJudgeScorer.java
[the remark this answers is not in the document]
petrotiurin: Update — reversing the earlier 'skipping'. Fixed in 9c5db354b3: we want an interim consistency guarantee before the pre-evaluation delay lands. `gatherSpanAttachments` now routes through a new batched, upload-race-tolerant lookup (`OnlineScoringBaseScorer.listSpanAttachmentsToleratingUploadRace`): spans whose body references an attachment drive a bounded retry of the **single batched** query until each such span has a persistent copy visible, then it falls back to a best-effort grouping. This keeps one query per attempt (scales to large traces), preserves `preferPersistentAttachments` per span and the degrade-to-none-on-error behavior. Added `spanAttachmentUploadRaceInTraceStructureRetriesUntilSpanCopyAppears` to cover it.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringBaseScorer.java
```diff
@@ -120,6 +139,127 @@ protected static <T> Mono<T> surfaceInjectedMediaFailure(@NonNull Throwable erro
...
+                traceId, workspaceId, error));
+        if (spanIdsExpectingAttachment.isEmpty()) {
+            return logged.map(OnlineScoringBaseScorer::groupBySpanPreferringPersistent).onErrorReturn(Map.of());
+        }
+        return logged
+                .map(OnlineScoringBaseScorer::groupBySpanPreferringPersistent)
+                .filter(bySpan -> spanIdsExpectingAttachment.stream()
+                        .allMatch(id -> hasPersistentAttachment(bySpan.getOrDefault(id, List.of()))))
+                .repeatWhenEmpty(ATTACHMENT_FETCH_MAX_RETRIES,
+                        repeats -> repeats.delayElements(ATTACHMENT_FETCH_RETRY_DELAY))
```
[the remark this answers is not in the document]
petrotiurin: Skipping for now — the two pipelines already share their building blocks (`preferPersistentAttachments`, `hasPersistentAttachment`, the retry constants) and differ in the parts that matter: the single-entity path returns a `List` keyed on one entity's body gate, while the span path returns a `Map` keyed on a per-span 'all expecting spans have a persistent copy' predicate. A generic `<R,T>` extraction is doable but would trade two short, readable methods for one higher-order method with injected transform/predicate/empty-value/log-context, which I don't think is a net readability win at two call sites. Happy to revisit if a third caller appears.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSpanLlmAsJudgeScorer.java
```diff
@@ -98,40 +156,199 @@ protected Mono<Void> score(@NonNull SpanToScoreLlmAsJudge message) {
...
+    }
+
+    // Package-private for unit tests.
+    Mono<ChatResponse> handleToolCalls(ChatResponse chatResponse, ChatRequest toolRequest,
+            ChatRequest structuredRequest, SpanToScoreLlmAsJudge message, Map<String, String> mdc) {
+
+        AiMessage aiMessage = chatResponse.aiMessage();
+        if (!aiMessage.hasToolExecutionRequests()) {
+            return Mono.just(chatResponse);
         }
```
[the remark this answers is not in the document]
petrotiurin: Fixed in 2c18c399a1 — extracted the shared orchestration into `OnlineScoringBaseScorer.runToolCallLoop`: the no-tool-calls guard, deferred context build, `ToolChoice.AUTO` follow-ups, `ToolCallLoop.runWithWrapUp`, and the `surfaceInjectedMediaFailure` error path now live in one place. Each scorer passes only its entity-specific `TraceToolContext` builder (`forActiveTrace` + `cache(TRACE)` vs `forActiveSpan` + `cache(SPAN)`) and per-message `scoreFn`. The builder is invoked inside the `defer`, so the once-per-subscription guarantee is preserved.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringLlmAsJudgeScorer.java
```diff
@@ -207,16 +210,97 @@ protected Mono<Void> score(@NonNull TraceToScoreLlmAsJudge message) {
...
+        return Mono.zip(gatherSpanAttachments(trace, spans, message), traceAttachmentsMono)
+                .map(tuple -> {
+                    // Built once here and threaded downstream to prepareEvaluation (size estimate) and
+                    // the tool-cache pre-seed, so the {{trace}} path serializes {trace, spans} a single
+                    // time. compress() deep-copies before mutating (FULL/MEDIUM), so this node stays
+                    // pristine and is safe to reuse.
+                    JsonNode fullJson = traceCompressor.buildFullJson(trace, spans);
+                    // Enrich with BOTH per-span and trace-level attachments so the judge can fetch any of them.
+                    var compressed = traceCompressor.compress(fullJson, trace, spans, CompressionTier.FULL,
+                            tuple.getT1(), tuple.getT2());
```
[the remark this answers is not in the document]
petrotiurin: Fixed in 9d11e57ea0. The non-tool-provider inline fallback now routes through the capped `prepareLlmRequest` overload (`MAX_PROMPT_FIELD_CHARS`), so the injected `{{trace}}` substitution is bounded and can't overflow the context window. No drill-down hint is passed there since the model has no `read`/`jq` tools to act on one — over-cap values are just truncated. The normal non-`{{trace}}` inline path stays uncapped (unchanged).

The span scorer had the identical issue (a `{{span}}` rule on a non-tool provider with the toggle ON injected the real structure uncapped) — fixed the same way, capping only when a real structure is present so the toggle-OFF `"{}"` case and the plain inline path stay uncapped. Added `OnlineScoringSpanLlmAsJudgeScorerTest.spanVariableOnNonToolProviderCapsInjectedStructure` as a regression test (asserts no tool specs, truncation marker present, oversized tail dropped).

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringLlmAsJudgeScorer.java
```diff
@@ -524,4 +588,13 @@ private record PreparedEvaluation(ChatRequest scoreRequest, ChatRequest structur
...
     }
 
+    /**
+     * Carry from {@link #buildTraceStructure} to {@link #evaluate} on the {@code {{trace}}} path.
+     * {@code envelopeJson} is the rendered structure injected into the prompt; {@code fullJson} is the
+     * {@code {trace, spans}} composite built alongside it, threaded down so {@link #prepareEvaluation}
+     * reuses it for the size estimate / tool-cache pre-seed instead of serializing the trace again.
+     */
+    private record TraceStructure(String envelopeJson, JsonNode fullJson) {
+    }
```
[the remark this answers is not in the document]
petrotiurin: Skipping this one. `TraceStructure` is a private, single-call-site carry record used only to thread two values from `buildTraceStructure` to `prepareEvaluation` within this class — and its sibling `PreparedEvaluation` (same file, same role) is also a plain positional record with no `@Builder`/`@NonNull`. Matching that keeps the two local carriers consistent; the SKILL.md `@Builder(toBuilder=true)` + `@NonNull` convention is aimed at API/domain records that cross boundaries, not one-off private tuples. The two components are populated unconditionally a few lines above the constructor, so there's no real null-guard or builder-ergonomics win here. Happy to revisit if `TraceStructure` ever grows fields or gains a second construction site.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringLlmAsJudgeScorer.java
```diff
@@ -301,12 +395,24 @@ private PreparedEvaluation prepareEvaluation(TraceToScoreLlmAsJudge message, Lis
...
+                    // built at FULL tier for the (unavailable) tools path, and there are no read/jq tools to
+                    // drill in here, so cap the substitutions to bound the context window — otherwise a large
+                    // trace would inject uncapped and could overflow the model's context. No drill-down hint:
+                    // the model can't act on one, so over-cap values are simply truncated.
+                    scoreRequest = OnlineScoringEngine.prepareLlmRequest(
+                            message.llmAsJudgeCode(), trace,
+                            llmProviderFactory.getStructuredOutputStrategy(modelName),
+                            message.promptType(), MAX_PROMPT_FIELD_CHARS, INLINE_TRUNCATION_HINT, spans,
+                            traceStructureJson);
+                    structuredRequest = scoreRequest;
```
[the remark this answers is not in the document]
petrotiurin: Done in 87d00eea15 — hoisted both `MAX_PROMPT_FIELD_CHARS` and `INLINE_TRUNCATION_HINT` to `OnlineScoringBaseScorer` as `protected static final`, so they live in one place and can't drift. No behavior change (same values).

For the record, the duplication was limited to those two constants. The fallback *branches* themselves stay per-scorer because they call different engine entrypoints with different signatures (`prepareLlmRequest(..., promptType, spans, traceStructureJson)` vs `prepareSpanLlmRequest(..., spanStructureJson)`), and the truncation *mechanism* was already shared via `OnlineScoringEngine.capReplacements` → `StringTruncator`. Extracting the branch into a higher-order helper for two call sites would mirror the `listAttachments`/`listSpanAttachments` case we already decided to leave as two readable methods, so I kept the dedup to the constants.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSpanLlmAsJudgeScorer.java
```diff
@@ -125,40 +157,201 @@ protected Mono<Void> score(@NonNull SpanToScoreLlmAsJudge message) {
...
+                },
+                toolRegistry, request -> scoreSpanReactive(request, message, recorder),
+                message.llmAsJudgeCode().model().name(), span.id().toString(), userFacingLogger, mdc, recorder);
+    }
+
+    /**
+     * Carry from {@link #prepareEvaluation} to {@link #evaluate}. {@code useTools} drives whether
+     * {@code handleToolCalls} runs the agentic loop after the first response.
+     */
+    private record PreparedEvaluation(ChatRequest scoreRequest, ChatRequest structuredRequest, boolean useTools) {
```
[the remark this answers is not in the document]
petrotiurin: Same call as the earlier `TraceStructure` record thread (already resolved): `PreparedEvaluation` is a private, single-file carry record used only to pass values from `prepareEvaluation` to `evaluate`, and it matches the sibling carry records in the same file (plain positional records, no `@Builder`/`@NonNull`). The SKILL.md `@Builder(toBuilder=true)` + `@NonNull` convention targets API/domain records that cross boundaries, not one-off private tuples — the components are populated unconditionally at the single construction site, so there's no null-guard or builder-ergonomics win. Declining for consistency.

(This re-surfaced because the rebase onto main + force-push gave fresh commit SHAs to re-review; the prior thread is on now-outdated lines.)

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringBaseScorer.java
```diff
@@ -103,6 +197,127 @@ protected static <T> Mono<T> surfaceInjectedMediaFailure(@NonNull Throwable erro
...
+
+        boolean expectsAttachment = AttachmentUtils.hasAttachmentReferences(JsonUtils.getMapper(), bodyNodes);
+        if (!expectsAttachment) {
+            return fetch.map(OnlineScoringBaseScorer::preferPersistentAttachments).onErrorReturn(List.of());
+        }
+        return fetch
+                .map(OnlineScoringBaseScorer::preferPersistentAttachments)
+                .filter(OnlineScoringBaseScorer::hasPersistentAttachment)
+                .repeatWhenEmpty(ATTACHMENT_FETCH_MAX_RETRIES,
+                        repeats -> repeats.delayElements(ATTACHMENT_FETCH_RETRY_DELAY))
```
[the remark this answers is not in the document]
petrotiurin: Same as the earlier "Duplicate attachment retry pipeline" thread we already settled: the two methods share their building blocks (`preferPersistentAttachments`, `hasPersistentAttachment`, the retry constants) but differ in exactly the parts that matter — `listAttachmentsToleratingUploadRace` returns a `List` gated on one entity's body reference, while `listSpanAttachmentsToleratingUploadRace` returns a `Map<UUID, List<…>>` keyed on a per-span "every expecting span has a persistent copy" predicate. A generic `<R,T>` extraction would trade two short, readable methods for one higher-order method with injected transform/predicate/empty-value/log-context — not a net readability win at two call sites. Declining, consistent with the prior decision; happy to revisit if a third caller appears.

(Re-surfaced by the rebase's fresh SHAs.)

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringBaseScorer.java
```diff
@@ -46,6 +62,35 @@
...
+     * Attachment-upload race tolerance for the {@code {{trace}}} / {@code {{span}}} structures. The SDK
+     * uploads an entity's attachment a short moment <em>after</em> the entity itself is ingested, so a
+     * scoring run triggered immediately can read the attachment table before the persistent copy lands.
+     * When the entity body references an attachment but the listing is still empty, the cold lookup is
+     * resubscribed up to {@link #ATTACHMENT_FETCH_MAX_RETRIES} times spaced by
+     * {@link #ATTACHMENT_FETCH_RETRY_DELAY} so the upload can complete (~0.3–1.5 s worst case, and only
+     * for entities that actually expect an attachment). See {@link #listAttachmentsToleratingUploadRace}.
+     */
+    private static final int ATTACHMENT_FETCH_MAX_RETRIES = 5;
+    private static final Duration ATTACHMENT_FETCH_RETRY_DELAY = Duration.ofMillis(300);
```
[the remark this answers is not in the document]
petrotiurin: Keeping these internal for now. `ATTACHMENT_FETCH_MAX_RETRIES` / `ATTACHMENT_FETCH_RETRY_DELAY` are a deliberate interim guard to ensure the {{trace}}/{{span}} attachment path works correctly against the upload race while the feature beds in — a proper mechanism to counteract the race (a pre-evaluation delay) is planned to replace this retry entirely. Adding config surface + `config.yml`/`config-test.yml` keys + validation for a soon-to-be-removed mechanism isn't worth the churn; the current defaults (5 × 300ms, only for entities that actually reference an attachment) are conservative and bounded. Will revisit tunability if the retry outlives its interim role.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/tools/AttachmentSummaries.java
```diff
@@ -12,27 +12,35 @@
...
+     * detection on the file name when the stored MIME is blank; {@code media_type} is the coarse
+     * {@link MediaCategory} bucket (image / audio / video / other) the judge uses to decide whether
+     * {@code get_attachment} can load it.
      */
     public static ArrayNode toJsonArray(@NonNull List<AttachmentInfo> attachments) {
         ArrayNode array = JsonUtils.createArrayNode();
         for (var info : attachments) {
             String mime = StringUtils.isNotBlank(info.mimeType()) ? info.mimeType() : TIKA.detect(info.fileName());
             ObjectNode item = array.addObject();
+            item.put("type", info.entityType().name().toLowerCase());
```
andrescrz: For enums in general and this one in particular, we usually use the `value` field as the JSON serialisable form, so we don't rely on the enum constant. You can see it has Jackson serialisation and deserialisation annotations.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/tools/GetAttachmentTool.java
```diff
@@ -56,18 +58,22 @@ public class GetAttachmentTool implements ToolExecutor {
     private static final ToolSpecification SPEC = ToolSpecification.builder()
             .name(NAME)
             .description("Load a media attachment (image, audio, or video) belonging to a trace or span"
-                    + " so you can inspect its actual content while scoring. Reading a trace lists its"
-                    + " attachments in an `attachments` field (file names + media types); pass one of those"
-                    + " file names here to load it into the conversation as viewable media. Only image,"
-                    + " audio, and video files can be loaded."
-                    + " Loaded media links are time-limited; if a link has expired, call this tool again"
-                    + " with the same arguments to obtain a fresh one.")
+                    + " so you can inspect its actual content while scoring. The trace/span structure and"
```
andrescrz: Nit: all these were already happening, so no need to change it. But as you edited it, you can take the chance to improve it. Better use java text blocks (with triple quotes `"""`) instead of concating strings (with `+` operator).

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/tools/TraceCompressor.java
```diff
@@ -57,7 +61,8 @@ public CompressionResult compress(@NonNull JsonNode fullJson,
             @NonNull Trace trace,
             @NonNull List<Span> spans,
             CompressionTier forcedTier) {
-        return compress(fullJson, trace, spans, forcedTier, PathAwareTruncator.SuffixStyle.WITH_JQ_HINT);
+        return compress(fullJson, trace, spans, forcedTier, PathAwareTruncator.SuffixStyle.WITH_JQ_HINT,
```
andrescrz: Minor: all these overloads for this `compress` method are a bit weird. Generally, these methods should have the complete version with the implementation, and then shorter versions in order of parameters. So basically:
1. Make sure you don't mix order, it makes this very hard to maintain (admiting that overloaded methods with many params are always hard to maintain).
2. Passing empty map and list seems a clear candidate to overload and default to those, so you wouldn't need changes in unrelated lines like this one.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringBaseScorer.java
```diff
@@ -46,6 +62,35 @@
...
     public static final int TRACE_PAGE_LIMIT = 2000;
+
+    /**
+     * Per-variable substitution cap for the agentic-tools / structure-injection paths, shared by the
+     * trace- and span-level scorers. ≈ 4 KB chars (~ 1 K tokens via the {@code Tokens.estimate}
+     * convention) is large enough that small trace/span input/output renders inline (cheap, no tool
+     * round-trip) but small enough that a huge entity doesn't blow context — the agent fetches the rest
+     * via the {@code read} tool, or, on the no-tools inline fallback, the value is simply truncated.
+     */
+    protected static final int MAX_PROMPT_FIELD_CHARS = 4_000;
```
andrescrz: Not a blocker: double check if this better as configurable variable or if rather keep it as constant.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringBaseScorer.java
```diff
@@ -46,6 +62,35 @@
...
+    /**
+     * Attachment-upload race tolerance for the {@code {{trace}}} / {@code {{span}}} structures. The SDK
+     * uploads an entity's attachment a short moment <em>after</em> the entity itself is ingested, so a
+     * scoring run triggered immediately can read the attachment table before the persistent copy lands.
+     * When the entity body references an attachment but the listing is still empty, the cold lookup is
+     * resubscribed up to {@link #ATTACHMENT_FETCH_MAX_RETRIES} times spaced by
+     * {@link #ATTACHMENT_FETCH_RETRY_DELAY} so the upload can complete (~0.3–1.5 s worst case, and only
+     * for entities that actually expect an attachment). See {@link #listAttachmentsToleratingUploadRace}.
+     */
+    private static final int ATTACHMENT_FETCH_MAX_RETRIES = 5;
```
andrescrz: I agree with previous Baz comment. This is a particular case for race conditions, but generally, retry policies (including number of attempts, delay etc.) should be configurable. We have a similar precedent here for a particular case here:
- https://github.com/comet-ml/opik/pull/6871/

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringBaseScorer.java
```diff
@@ -73,6 +118,55 @@ protected OnlineScoringBaseScorer(@NonNull @Config OnlineScoringConfig config,
...
+     * pre-seed happen exactly once per subscription.
+     *
+     * @param contextSupplier builds and pre-seeds the tool context (invoked at subscription time)
+     * @param scoreFn         issues a single LLM call (e.g. {@code request -> scoreTraceReactive(...)})
+     * @param logId           the trace/span id used as the tool-loop log correlation id
+     * @param recorder        evaluation monitoring recorder threaded into {@link ToolCallLoop} so each
+     *                        tool call is recorded (OPIK-6994); pass {@link EvaluationRecorder#NOOP} when
+     *                        monitoring is off
+     */
+    protected Mono<ChatResponse> runToolCallLoop(@NonNull ChatResponse initialResponse,
```
andrescrz: This is a design question, more than a correctness one. The functionality of the online scorers is growing too much. You should extract all these tooling chunks somewhere else, instead of using inheritance to propagate it. Likely, the right place is a separated service encapsulating all this. The online scorers would be an user of that service and it would be accessible to the child scorer sub-classes.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringBaseScorer.java
```diff
@@ -103,6 +197,127 @@ protected static <T> Mono<T> surfaceInjectedMediaFailure(@NonNull Throwable erro
...
+                .filter(OnlineScoringBaseScorer::hasPersistentAttachment)
+                .repeatWhenEmpty(ATTACHMENT_FETCH_MAX_RETRIES,
+                        repeats -> repeats.delayElements(ATTACHMENT_FETCH_RETRY_DELAY))
+                .onErrorResume(error -> Mono.empty())
+                // Retries exhausted (no persistent copy will come): best-effort final read so a
+                // backend-/REST-only auto-stripped attachment is still surfaced rather than dropped. Uses
+                // the raw coldFetch — any failure on the primary attempt above was already logged.
+                .switchIfEmpty(Mono.defer(() -> coldFetch
+                        .map(OnlineScoringBaseScorer::preferPersistentAttachments)
+                        .onErrorReturn(List.of())));
```
andrescrz: Please review these error handling to see that you don't silently drop other stuff. Potential ideas:
- Narrow down to the exact exception types and bubble up everything else.
- Add logging.

This applies to all occurrences below.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringLlmAsJudgeScorer.java
```diff
@@ -223,21 +217,108 @@ protected Mono<Void> score(@NonNull TraceToScoreLlmAsJudge message) {
...
+     *
+     * <p>Trace-level attachment metadata isn't carried on the {@link Trace} object, and per-span
+     * metadata isn't on the {@link Span} objects, so both are fetched: the trace's own via a
+     * race-tolerant single lookup, and the spans' via a race-tolerant batched lookup. Both tolerate the
+     * attachment-upload race (an attachment may not be persisted yet when scoring is enqueued) and
+     * degrade to no attachments on a listing failure rather than blocking scoring.
+     */
+    private Mono<TraceStructure> buildTraceStructure(Trace trace, List<Span> spans, TraceToScoreLlmAsJudge message) {
+        Mono<List<AttachmentInfo>> traceColdFetch = attachmentService
+                .getAttachmentInfoByEntity(trace.id(), com.comet.opik.api.attachment.EntityType.TRACE,
```
andrescrz, line 235: Nit: use imports instead of fully qualified classes.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringLlmAsJudgeScorer.java
```diff
@@ -223,21 +217,108 @@ protected Mono<Void> score(@NonNull TraceToScoreLlmAsJudge message) {
...
+                    JsonNode fullJson = traceCompressor.buildFullJson(trace, spans);
+                    // Enrich with BOTH per-span and trace-level attachments so the judge can fetch any of them.
+                    var compressed = traceCompressor.compress(fullJson, trace, spans, CompressionTier.FULL,
+                            tuple.getT1(), tuple.getT2());
+                    // compress() is id-agnostic; add the id envelope on top here, mirroring ReadTool.
+                    ObjectNode envelope = JsonUtils.getMapper().createObjectNode();
+                    envelope.put("trace_id", trace.id() != null ? trace.id().toString() : null);
+                    envelope.put("tier", compressed.tier().name());
+                    envelope.set("data", compressed.payload());
+                    return new TraceStructure(envelope.toString(), fullJson);
```
andrescrz: Minor: favour builders over constructs in the code base.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSpanLlmAsJudgeScorer.java
```diff
@@ -125,40 +157,201 @@ protected Mono<Void> score(@NonNull SpanToScoreLlmAsJudge message) {
...
+                        span.id(), modelName);
+            }
+
             ChatRequest scoreRequest;
+            ChatRequest structuredRequest;
             try {
-                String modelName = message.llmAsJudgeCode().model().name();
-                var strategy = llmProviderFactory.getStructuredOutputStrategy(modelName);
-                scoreRequest = OnlineScoringEngine.prepareSpanLlmRequest(message.llmAsJudgeCode(), span, strategy);
+                if (useTools) {
```
andrescrz, line 261: Minor: if possible, please break each of the if block bodies into a small aux method, to prevent all this from growing and keep the implementation of this method readable. The aux method can return a value to be used from the caller method.

apps/opik-backend/src/main/java/com/comet/opik/domain/attachment/AttachmentService.java
```diff
@@ -438,6 +444,16 @@ public Mono<Boolean> hasAnyAttachmentByEntityIds(@NonNull EntityType entityType,
                 .map(list -> !list.isEmpty());
     }
 
+    @Override
+    @WithSpan
+    public Mono<List<AttachmentInfo>> getAttachmentInfoByEntityIds(@NonNull EntityType entityType,
+            @NonNull Set<UUID> entityIds) {
+        if (entityIds.isEmpty()) {
```
andrescrz: Minor: if you handle the empty case for the entityIds set, better handle the null one. Remove the non-null annotation and use CollectionUtils.isEmpty for a null safe check.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/OnlineScoringSpanLlmAsJudgeScorerTest.java
```diff
@@ -0,0 +1,465 @@
...
+        when(feedbackScoreService.scoreBatchOfSpans(any())).thenReturn(Mono.empty());
+
+        scorer.score(message).block();
+
+        // No {{span}} → inline path: no tool specs, no attachment lookup.
+        assertThat(requestCaptor.getValue().toolSpecifications()).isNullOrEmpty();
+        verifyNoInteractions(attachmentService);
+    }
+
+    private Span createSpan() {
```
andrescrz, line 462: Minor: for generating objects with random test data, better use Podam. Then, you can use toBuilder and override the particular fields. Keep using UUID, RandomStringUtils, Instant etc. for those overrides

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/OnlineScoringSpanLlmAsJudgeScorerTest.java
```diff
@@ -0,0 +1,465 @@
...
+                .span(span)
+                .ruleId(UUID.randomUUID())
+                .ruleName("rule")
+                .llmAsJudgeCode(code)
+                .workspaceId("ws-1")
+                .userName("user-1")
+                .build();
+    }
+
+    private static com.comet.opik.api.resources.v1.events.tools.ToolExecutor stubTool(String name, String result) {
```
andrescrz: As mentioned above, use imports as much as possible and online inline the full class name when they clash. This is problem is quite widespread in the tests added/updated in this PR.

andrescrz (commented):
As we talked, the main question here are the product requirements and user expectations, that should be discussed in order to translate them to a proper implementation. 

That heavily determines the implementation approach, which as we discussed, it should consider the following:
- Event based vs Time based.
- Embedded vs actual attachments.
- Dealing with the different cases of attachment stripment for embedded attachments. 

This BE implementation looks pretty clean, taking into account the doubts above and considering that it leans to a retried based approach (best effort loosely time based approach), which might not work depending on the timing.

For this particular implementation, I left multiple comments to consider before moving forward, with different degrees of importance.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringLlmAsJudgeScorer.java
petrotiurin: Skipping — this one is a genuine name clash rather than an avoidable fully-qualified reference. This file imports `com.comet.opik.api.resources.v1.events.tools.EntityType` (used at line 607 for the tool-context `EntityRef`), so the attachment `com.comet.opik.api.attachment.EntityType` at lines 237/297 has to stay fully qualified to disambiguate the two. That matches your own guidance in the test-file comment below ("use imports as much as possible and inline the full class name when they clash"), so I've left it as-is.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringBaseScorer.java
```diff
@@ -103,6 +177,140 @@ protected static <T> Mono<T> surfaceInjectedMediaFailure(@NonNull Throwable erro
...
+
+        boolean expectsAttachment = AttachmentUtils.hasAttachmentReferences(JsonUtils.getMapper(), bodyNodes);
+        if (!expectsAttachment) {
+            return fetch.map(OnlineScoringBaseScorer::preferPersistentAttachments).onErrorReturn(List.of());
+        }
+        return fetch
+                .map(OnlineScoringBaseScorer::preferPersistentAttachments)
+                .filter(OnlineScoringBaseScorer::hasPersistentAttachment)
+                .repeatWhenEmpty(onlineScoringConfig.getAttachmentFetchMaxRetries(),
+                        repeats -> repeats.delayElements(
```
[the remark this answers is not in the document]
petrotiurin: This is minimal and was already addressed.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringBaseScorer.java
```diff
@@ -103,6 +197,137 @@ protected static <T> Mono<T> surfaceInjectedMediaFailure(@NonNull Throwable erro
...
+        // on the empty-completion-driven retries or the benign retry-exhaustion path below.
+        Mono<List<AttachmentInfo>> fetch = coldFetch.doOnError(error -> log.warn(
+                "Failed to list attachments for workspace '{}', entity '{}'; degrading to best-effort"
+                        + " attachment discovery (online scoring will proceed without them)",
+                workspaceId, entityId, error));
+
+        boolean expectsAttachment = AttachmentUtils.hasAttachmentReferences(JsonUtils.getMapper(), bodyNodes);
+        if (!expectsAttachment) {
+            return fetch.map(OnlineScoringBaseScorer::preferPersistentAttachments).onErrorReturn(List.of());
+        }
```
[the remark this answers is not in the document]
petrotiurin: Not feasible to run this retry on every online eval. Will be tackled in OPIK-7224.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSpanLlmAsJudgeScorer.java
```diff
@@ -125,40 +154,234 @@ protected Mono<Void> score(@NonNull SpanToScoreLlmAsJudge message) {
...
+                message.llmAsJudgeCode(), span, new InstructionStrategy(),
+                onlineScoringConfig.getMaxPromptFieldChars(), drillDownHint, spanStructureJson);
+        ChatRequest structuredRequest = OnlineScoringEngine.prepareSpanLlmRequest(
+                message.llmAsJudgeCode(), span,
+                llmProviderFactory.getStructuredOutputStrategy(modelName),
+                onlineScoringConfig.getMaxPromptFieldChars(), drillDownHint, spanStructureJson);
+        // REQUIRED on the first call only forces ≥1 tool call; follow-ups switch to AUTO in
+        // handleToolCalls so the model can decide when to stop investigating.
+        scoreRequest = agenticScoringService.addToolSpecs(scoreRequest, ToolChoice.REQUIRED);
+        return new LlmRequests(scoreRequest, structuredRequest);
```
andrescrz: Nit: as mentioned before, better favour builders in the code base and use them, instead of constructors. You still left some behind in multiple places of the PR.

apps/opik-backend/src/main/java/com/comet/opik/domain/attachment/AttachmentUtils.java
```diff
@@ -212,6 +214,64 @@ public static boolean hasAttachmentReferences(@NonNull ObjectMapper objectMapper
...
+        if (node == null || node.isNull()) {
+            return;
+        }
+
+        if (node.isTextual()) {
+            String text = node.asText();
+            Matcher matcher = FIND_REFERENCE_PATTERN.matcher(text);
+            while (matcher.find()) {
+                references.add(matcher.group(1));
+            }
```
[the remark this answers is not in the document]
andrescrz: Agree with Baz on this. Better narrow down the exception and allow the rest to bubble up. In addition, let's add some logging as well (likely at warn level). This is an unexpected condition.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/OnlineScoringConfig.java
```diff
@@ -111,6 +111,40 @@ public class OnlineScoringConfig {
...
+     * shared by the trace- and span-level scorers. ≈ 4 KB chars (~ 1 K tokens) is large enough that a
+     * small trace/span input/output renders inline (cheap, no tool round-trip) but small enough that a
+     * huge entity doesn't blow context — the agent fetches the rest via the {@code read} tool, or, on the
+     * no-tools inline fallback, the value is simply truncated.
+     *
+     * <p>Floor at 500 chars so a typo'd env var can't truncate every field down to noise. Field
+     * initializer (not {@code @Builder.Default}) so the default applies during YAML deserialization too.
+     */
+    @JsonProperty
+    @Min(500) private int maxPromptFieldChars = 4_000;
```
andrescrz: Minor: maybe desirable to add some max. If max integer is the max, ok to not to have it.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/OnlineScoringConfig.java
```diff
@@ -111,6 +111,40 @@ public class OnlineScoringConfig {
...
+    /**
+     * Attachment-upload race tolerance for the {@code {{trace}}} / {@code {{span}}} structures: how many
+     * times the (cold) attachment lookup is resubscribed when the entity body references an attachment but
+     * no persistent copy is visible yet. This is an interim guard while attachment uploads can still be in
+     * flight when scoring is enqueued; a pre-evaluation dispatch delay (OPIK-7224) is intended to replace
+     * it. {@code 0} disables retrying (a single best-effort read). See
+     * {@code OnlineScoringBaseScorer#listAttachmentsToleratingUploadRace}.
+     */
+    @JsonProperty
+    @Min(0) @Max(20) private int attachmentFetchMaxRetries = 5;
```
andrescrz, line 138: Not a blocker, as you follow the (wrong) convention in this file. It's better to not to add default values in the Java code, so we keep a single centralised source of truth in the config yaml files. As these are non-nullable (primitive types, but for Objects you could annotate with NotNull), they are validated to be present.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/AgenticScoringService.java
```diff
@@ -0,0 +1,486 @@
...
+
+    private final OnlineScoringConfig onlineScoringConfig;
+    private final ToolRegistry toolRegistry;
+
+    @Inject
+    public AgenticScoringServiceImpl(@NonNull @Config("onlineScoring") OnlineScoringConfig onlineScoringConfig,
+            @NonNull ToolRegistry toolRegistry) {
+        this.onlineScoringConfig = onlineScoringConfig;
+        this.toolRegistry = toolRegistry;
+    }
```
andrescrz: Nit: use lombok for boiler plate constructors.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/AgenticScoringService.java
```diff
@@ -0,0 +1,486 @@
...
+        return estimateTokensFromJson(traceCompressor.buildFullJson(trace, spans), charsPerToken);
+    }
+
+    @Override
+    public int estimateThreadContextTokens(@NonNull List<Trace> traces, @NonNull List<Span> spans,
+            int charsPerToken) {
+        Preconditions.checkArgument(charsPerToken >= 1, "charsPerToken must be >= 1, got %s", charsPerToken);
+        try {
+            return JsonUtils.getMapper()
+                    .writeValueAsString(OnlineScoringEngine.fromTraceToThreadEnriched(traces, spans)).length()
```
andrescrz: Minor: there's probably one internal version of `writeValueAsString` within JsonUtils that already converts the exception, you could use that, instead of retrieving the exposed mapper and bypassing. Exposing the mapper was a pre-existing mistake in the code base, that we're trying to avoid if possible.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/OnlineScoringEngineTest.java
```diff
@@ -1205,6 +1183,58 @@ void templateReferencesSpansDetectsImplicitMessageTemplateReference() {
         assertThat(OnlineScoringEngine.templateReferencesSpans(noSpansMessage, Map.of(), mustache)).isFalse();
     }
 
+    @Test
+    @DisplayName("templateReferencesTrace detects the 'trace' sentinel and implicit {{trace}} references")
+    void templateReferencesTraceDetectsSentinelAndImplicitReference() {
+        var mustache = com.comet.opik.api.PromptType.MUSTACHE;
+        var noMessages = List.<com.comet.opik.api.evaluators.LlmAsJudgeMessage>of();
```
andrescrz: Nit: use imports instead of full qualified classes names.

andrescrz (commented):
Without all the context from potential further product and requirements discussions, the BE implementation which is based on a best effort retried based approach LGTM. The code is generally clean and the refactors improved the pre-existing code base.

I still left some minor non-blocking comments, that you could optionally address while you get a FE review.

You're good to go once approved on the FE side.

aadereiko (approved):
The FE code looks good, nice work!

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSpanLlmAsJudgeScorer.java
```diff
@@ -125,40 +155,240 @@ protected Mono<Void> score(@NonNull SpanToScoreLlmAsJudge message) {
...
+    /**
+     * Inline path that still injects the {{span}} structure so the variable renders rather than leaking
+     * the bare sentinel. Toggle OFF: spanStructureJson is null → renders "{}" (tiny, no cap needed; user
+     * variables stay uncapped as on the normal inline path).
+     */
+    private LlmRequests buildSentinelStructureRequests(SpanToScoreLlmAsJudge message, Span span,
+            String modelName, String spanStructureJson) {
+        ChatRequest scoreRequest = OnlineScoringEngine.prepareSpanLlmRequest(
+                message.llmAsJudgeCode(), span,
+                llmProviderFactory.getStructuredOutputStrategy(modelName), spanStructureJson);
```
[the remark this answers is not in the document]
petrotiurin: This branch is never reached with a non-null `spanStructureJson`, so there's nothing to cap here.

In `prepareEvaluation` (lines 261-269):
```java
if (useTools) {
    requests = buildToolCallingRequests(...);                    // capped (6-arg overload)
} else if (referencesSpan && spanStructureJson != null) {
    requests = buildInlineStructureRequests(...);                 // capped (5-arg overload)
} else if (referencesSpan) {
    requests = buildSentinelStructureRequests(...);               // uncapped 3-arg overload
} else {
    requests = buildPlainRequests(...);
}
```

The `referencesSpan && spanStructureJson != null` branch is checked *before* the plain `referencesSpan` branch. By the time control reaches `buildSentinelStructureRequests`, `spanStructureJson != null` has already been ruled out — it's always `null` there.

Tracing further back: `spanStructureJson` only becomes non-null via `buildSpanStructure()` in `score()`, which only runs when `referencesSpan && agenticToolsEnabled` is true. When the toggle is off, `evaluate(message, null, referencesSpan, ...)` is called with a hardcoded `null`. Even under a toggle-flip race between the outer check in `score()` and the re-check in `prepareEvaluation`, a non-null structure always routes through the capped `buildInlineStructureRequests` branch instead — it can never fall through to the sentinel branch.

So `buildSentinelStructureRequests` only ever renders `spanStructureJson = null` → `{{span}}` becomes the literal `"{}"`, exactly per its own Javadoc: *"Toggle OFF: spanStructureJson is null → renders `{}` (tiny, no cap needed; user variables stay uncapped as on the normal inline path)."* `maxPromptFieldChars` is documented as scoped to "the agentic-tools / structure-injection paths" specifically — i.e. when a structure is actually injected — so there's no contract violation.

For what it's worth, the trace-level scorer (`OnlineScoringLlmAsJudgeScorer.java:396-439`) has the identical three-branch shape with the same uncapped fallback for its toggle-off case, so this is a deliberate, consistent pattern across both scorers rather than an oversight.

PR #7279, opened 2026-06-29
fix(metrics): tokenize inputs in METEOR before calling NLTK

Osamaali313:
## Summary

The default `METEOR` scorer (`sdks/python/src/opik/evaluation/metrics/heuristics/meteor.py`) passes raw strings to `nltk.translate.meteor_score.meteor_score`, but NLTK requires **pre-tokenized** inputs — `references` as `Iterable[Iterable[str]]` and `hypothesis` as `Iterable[str]`.

On current NLTK (3.9.x) this raises `TypeError` for any normal string input, so `METEOR().score(...)` is **non-functional on its default path** (on older NLTK it silently iterated the strings character-by-character, producing garbage scores). The sibling heuristic metrics (BLEU, GLEU, chrF) already `.split()` before calling NLTK — METEOR was the lone exception.

## Reproduction (before fix)

```python
METEOR(track=False).score(output='the cat sat on the mat', reference='the cat sat on the mat')
# TypeError: "hypothesis" expects pre-tokenized hypothesis (Iterable[str]): the cat sat on the mat
```

After the fix the same call returns ~`0.9977`.

## Fix

Tokenize `references` and `hypothesis` inside the default scorer before the NLTK call (the public `score()` signature and the injectable `meteor_fn` contract are unchanged).

## Testing

Added `test_meteor_default_scorer_tokenizes_inputs` to `test_heuristics.py`, asserting the default scorer hands NLTK tokenized lists. It mocks the NLTK scorer (and WordNet load), so it needs no corpora/network. It fails on the current code (raw strings) and passes after the fix; `pytest -k meteor` → 3 passed.

(The existing METEOR tests inject a custom `meteor_fn`, which bypasses the default scorer — that's why the broken default path went unnoticed.)

### DIFF
--- a/sdks/python/src/opik/evaluation/metrics/heuristics/meteor.py
+++ b/sdks/python/src/opik/evaluation/metrics/heuristics/meteor.py
@@ -82,10 +82,18 @@ class METEOR(base_metric.BaseMetric):
                         ) from download_error
 
             def _scorer(references: Sequence[str], hypothesis: str) -> float:
+                # NLTK's meteor_score expects pre-tokenized inputs:
+                # references as Iterable[Iterable[str]] and hypothesis as Iterable[str].
+                tokenized_references = [ref.split() for ref in references]
+                tokenized_hypothesis = hypothesis.split()
                 try:
                     return float(
                         nltk_meteor_score.meteor_score(
-                            references, hypothesis, alpha=alpha, beta=beta, gamma=gamma
+                            tokenized_references,
+                            tokenized_hypothesis,
+                            alpha=alpha,
+                            beta=beta,
+                            gamma=gamma,
                         )
                     )
                 except LookupError as error:
--- a/sdks/python/tests/unit/evaluation/metrics/test_heuristics.py
+++ b/sdks/python/tests/unit/evaluation/metrics/test_heuristics.py
@@ -433,11 +433,14 @@ def test_meteor_rejects_empty_inputs():
         metric.score(output="hyp", reference="   ")
 
 
-def test_meteor_default_scorer_tokenizes_inputs(monkeypatch):
-    """The built-in scorer must pass pre-tokenized inputs to NLTK.
+def test_meteor__default_scorer__passes_tokenized_inputs_to_nltk(monkeypatch):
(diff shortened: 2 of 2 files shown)


Osamaali313:
Thanks @baz-reviewer — addressed in 6353294:

- Renamed to `test_meteor__default_scorer__passes_tokenized_inputs_to_nltk` per the `python-sdk/testing.md` `test_WHAT__CASE__EXPECTED` convention.
- The test now also asserts the public `METEOR.score()` result (`0.5`).

I kept the tokenization assertion deliberately: the regression here is specifically that the default scorer must hand NLTK **pre-tokenized** inputs. Asserting only the stubbed return value wouldn't catch it — the stub ignores its arguments, so raw strings vs tokenized lists both yield `0.5`. So the test now verifies both the observable result and that the inputs were tokenized before the NLTK call.

Osamaali313:
Thanks @baz-reviewer — I'd prefer to keep this PR scoped to the METEOR bug fix. Extracting a shared candidate/references preprocessing helper and migrating BLEU/GLEU/METEOR onto it is a sensible maintainability improvement, but it would modify the already-working BLEU/GLEU metrics and turn a one-line bugfix into a cross-metric refactor — larger blast radius and harder to review/verify. I'd suggest doing it as a separate follow-up; happy to open one (or defer to your preference on where the shared helper should live). For now this PR just makes METEOR tokenize like the other NLTK metrics already do.

alexkuzmik (changes requested):
Hi @Osamaali313! Thank you for your PR.
Please fix the unit test, it's currently broken in the CI.

Osamaali313:
Good call. Added `test_meteor__default_scorer__real_nltk_score`, which runs the real default `METEOR().score(...)` for the identical-text case and asserts `~0.9977`, alongside the mocked tokenization check. It needs the WordNet corpora, so it skips when they aren't installed (as in CI) rather than failing. Confirmed the value locally (0.99768...).

Osamaali313:
Agreed, cleaner this way. Restricted the skip to `LookupError`/`ImportError` (genuinely missing corpora); a `MetricComputationError` from valid input now surfaces as a failure, so a real scoring regression won't be masked.

Issue #5524, opened 2026-03-05
ADK integration: LLM spans stuck at 'started' when context caching is enabled (SSE streaming)


lupuletic:
## Description

When using Opik's ADK integration (`OpikTracer` + `track_adk_agent_recursive`) with Google ADK's `ContextCacheConfig` enabled, LLM spans are never finalized — they remain at `_OPIK_SPAN_STATUS: started` with no output or usage data.

The issue only occurs in **SSE streaming mode** (Cloud Run / production). Locally with `InMemorySessionService` and no streaming, everything works correctly.

## Environment

- `opik==1.10.25`
- `google-adk==1.26.0`
- Python 3.12
- Cloud Run (SSE streaming mode)
- Vertex AI (`GOOGLE_GENAI_USE_VERTEXAI=True`)

### Setup

```python
from google.adk.agents import Agent
from google.adk.agents.context_cache_config import ContextCacheConfig
from google.adk.apps import App
from opik.integrations.adk import OpikTracer, track_adk_agent_recursive

root_agent = Agent(
    name="my_agent",
    model=Gemini(model="gemini-3-flash-preview"),
    # ... tools, callbacks, etc.
)

tracer = OpikTracer(name="my-agent", project_name="my-project")
track_adk_agent_recursive(root_agent, tracer)

app = App(
    root_agent=root_agent,
    name="app",
    # This breaks Opik LLM spans:
    context_cache_config=ContextCacheConfig(
        min_tokens=2048,
        ttl_seconds=1800,
    ),
)
```

### Steps

1. Deploy to Cloud Run with SSE streaming enabled
2. Send a message that triggers tool use (so there are 2 LLM calls)
3. Check Opik traces — LLM spans show `_OPIK_SPAN_STATUS: started`, no output, no usage

### Expected

LLM spans should show `_OPIK_SPAN_STATUS: ready_for_finalization` with output and usage data.

### Actual

All LLM spans stuck at `started`. Cloud Run logs show repeated:
```
OPIK: No current span found in context for model output update
```

This warning comes from `opik_tracer.py:280-284` where `context_storage.top_span_data()` returns `None`.

## Root Cause Analysis

Opik uses `contextvars.ContextVar` for span tracking (`OpikContextStorage`). The `before_model_callback` pushes span data via `context_storage.add_span_data()`, and `after_model_callback` retrieves it via `context_storage.top_span_data()`.

[...]

PR #7282, opened 2026-06-30
fix: recover ADK trace output when context caching detaches the LLM span (#5524)

lupuletic:
## Details

When using `OpikTracer` + `track_adk_agent_recursive` with ADK's `ContextCacheConfig` enabled, the **trace output is silently dropped** in SSE streaming mode — the trace shows an empty answer.

**Root cause.** `GeminiContextCacheManager` wraps the LLM call in its own OTel `start_as_current_span`. Under SSE streaming this crosses an async-context boundary, so the span pushed in `before_model_callback` is not visible in `after_model_callback` — `context_storage.top_span_data()` returns `None`. The callback then `discard`s the model output and returns, so `after_agent_callback` reads an empty `_last_model_output` and stamps `output=None` on the trace. (Same root cause as the `_OPIK_SPAN_STATUS: started` symptom in #5524 — the log line `No current span found in context for model output update` is emitted on every call.)

**Production evidence (opik 2.1.8, Cloud Run / SSE, `gemini-3-flash-preview`).** We toggled `CONTEXT_CACHE_ENABLED` live and bucketed trace-output presence by 2-minute windows:

| window | context caching | traces with output |
|---|---|---|
| before | OFF | **100%** |
| during | **ON** | **0%** (every bucket) |
| after | OFF | **100%** (recovered within ~2 min) |

So this is not cosmetic — with context caching on, **every** trace loses its answer.

**Fix.** In the detached-span path of `after_model_callback`, recover the model output: convert the `llm_response` and store it in the existing per-invocation, bounded `_last_model_output` cache (added in #7266). `after_agent_callback` then stamps the real answer via its normal path. Partial chunks are skipped; the normal (non-detached) path is unchanged.

Reusing the invocation-keyed cache means **no new shared state** and it's **safe under concurrent SSE sessions**. This is deliberately different from #5531's instance-level `_pending_llm_spans` LIFO list: a single shared stack would mix output/spans across concurrent invocations — the exact race that #7266 fixed for model output.

**Scope.** This restores the **trace-level output**, which is the user-impactful symptom. Finalizing the detached per-LLM child spans (so they leave `started` with their own output/usage) is a larger change, left as a follow-up — and it should recover the span via the same invocation-keyed bounded cache pattern, not a single shared LIFO (cf. the closed #5531).

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues
- Resolves the trace-output loss in #5524 (per-span finalization follow-up tracked separately)

## Testing

`tests/library_integration/adk/test_adk_concurrency.py`:
- `test_after_model_callback__recovers_output_when_span_detached` — with no current span, the actual model output is recovered into `_last_model_output` **and** `after_agent_callback` stamps it onto `trace.output` end-to-end (asserts the recovered text, not just non-`None`).
- `test_after_model_callback__detached_span__partial_chunk_discards` — partial chunks don't cache and clear any prior value for the invocation (discard-up-front), so a stale earlier output can't leak.

End-to-end: deploy an ADK agent with `ContextCacheConfig` to Cloud Run with SSE streaming; trace output is now present (previously empty).

## Documentation

No documentation updates required — internal bug fix, no API changes.

### DIFF
--- a/sdks/python/src/opik/integrations/adk/opik_tracer.py
+++ b/sdks/python/src/opik/integrations/adk/opik_tracer.py
@@ -285,9 +285,38 @@ class OpikTracer:
 
             current_span = context_storage.top_span_data()
             if current_span is None:
+                # The span pushed in before_model_callback is invisible here: ADK's
+                # ContextCacheConfig wraps the LLM call in its own OTel span, which
+                # forks the async context under SSE streaming, so the ContextVar
+                # mutation isn't visible and top_span_data() returns None (#5524).
+                #
+                # We can't finalize the per-LLM span, but we recover the model
+                # OUTPUT into the per-invocation, bounded _last_model_output cache
+                # (added in #7266) so after_agent_callback still stamps the trace
+                # output instead of dropping the whole answer. Keying by
+                # invocation_id keeps this safe across concurrent SSE sessions.
+                #
+                # Discard up front (mirroring the non-detached path below) so a
+                # failed conversion leaves no stale value; partial chunks never
+                # cache — we wait for the final response. The recovered output
+                # keeps its usage metadata (the non-detached path pops that onto the
+                # span, but there is no span here): harmless on the display-only
+                # trace output, and it preserves usage info that's otherwise lost.
                 self._last_model_output.discard(callback_context.invocation_id)
-                LOGGER.warning(
-                    "No current span found in context for model output update"
+                if not is_partial:
+                    try:
+                        self._last_model_output.set(
+                            callback_context.invocation_id,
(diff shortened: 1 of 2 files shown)


lupuletic:
@codex review

lupuletic:
Good call — fixed in e2e951dee1. Both detached-span tests now use only the public API: the recovery test drives `after_model_callback` → `after_agent_callback` and asserts on `TraceData.output`; the partial-chunk test recovers a real final response then sends a partial for the same invocation and asserts `trace.output is None` (exercising discard-up-front through the callback sequence, no `_last_model_output` seeding or reads).

alexkuzmik (approved):
Thanks for the contribution!

PR #7287, opened 2026-06-30
[NA] [SDK] fix: strip /api suffix in get_ui_url instead of char set

### LINKED ISSUE #7286: [Bug]: get_ui_url() strips characters instead of the "/api" suffix

- [x] Opik Python SDK

### Opik version

- Opik version: 2.1.11

### Describe the problem

`opik.url_helpers.get_ui_url()` builds the UI base URL with:

```python
return opik_url_override.rstrip("/api") + "/"
```

`str.rstrip("/api")` removes any trailing run of the characters in the set `{"/", "a", "p", "i"}`, not the literal `"/api"` suffix. The two shipped defaults (`http://localhost:5173/api` and `https://www.comet.com/opik/api/`) happen to survive, but any custom / self-hosted `OPIK_URL_OVERRIDE` whose path segment before `/api` ends in one of those characters is silently corrupted.

**Expected:** only a trailing `/api` segment is stripped, then a trailing slash is appended.

**Actual:** trailing characters belonging to `{/, a, p, i}` are eaten one by one.

This `get_ui_url()` value is used live in `plugins/pytest/summary.py` (the experiment UI link printed after a pytest eval run) and in `opik_client.py`, so affected deployments print/return broken UI links.

```python
from opik import url_helpers
from opik.config import OpikConfig
import opik

def ui_url(override):
    opik.config.OpikConfig = lambda: OpikConfig(url_override=override)
    return url_helpers.get_ui_url()

print(ui_url("https://foo.com/papi"))            # -> https://foo.com/            (ate "papi")
print(ui_url("https://team.io/instance-pi/api")) # -> https://team.io/instance-/  (ate "pi")
```

Both should keep the path: `https://foo.com/papi/` and `https://team.io/instance-pi/`.

A small fix (`rstrip("/").removesuffix("/api")`) plus a unit test will follow as a PR.

anxkhn:
## Details

`url_helpers.get_ui_url()` built the UI base URL with `opik_url_override.rstrip("/api")`, but `str.rstrip` strips a trailing *character set* (`/`, `a`, `p`, `i`), not the literal `/api` suffix. The two shipped defaults survive by coincidence, but any custom / self-hosted `url_override` whose path before `/api` ends in those characters is silently corrupted. This swaps in `rstrip("/").removesuffix("/api")` so only a trailing `/api` segment is dropped.
- `https://foo.com/papi` previously became `https://foo.com/` (ate `papi`); now `https://foo.com/papi/`.
- `https://team.io/instance-pi/api` previously became `https://team.io/instance-/`; now `https://team.io/instance-pi/`.
- The leading `rstrip("/")` keeps the trailing-slash cloud default `https://www.comet.com/opik/api/` correct (a bare `removesuffix("/api")` would have left `/api//`).

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #7286
- OPIK-

## Testing

- `cd sdks/python && uv run --no-sync python -m pytest tests/unit/test_url_helpers.py -p no:cacheprovider -q` -> 9 passed.
- `ruff check` and `ruff format --check` on the two changed files -> clean; `mypy --python-version 3.10` on `url_helpers.py` -> clean.
- Scenarios validated:
  - Regressions: both shipped defaults stay byte-identical, including the trailing-slash cloud default `https://www.comet.com/opik/api/` -> `https://www.comet.com/opik/`.
  - Bug fixes: `https://foo.com/papi` and `https://team.io/instance-pi/api` keep their path segment.
- Environment: local Python SDK unit tests, no Docker/backend needed (pure string change).
- Not run: integration / e2e suites (no behavior crosses into the backend; this is an internal URL-helper string fix). No video evidence (no user-visible UI change).

## Documentation

None needed (internal helper behavior; no public API or docs change).

### DIFF
--- a/sdks/python/src/opik/url_helpers.py
+++ b/sdks/python/src/opik/url_helpers.py
@@ -24,7 +24,7 @@ def get_ui_url() -> str:
     config = opik.config.OpikConfig()
     opik_url_override = config.url_override
 
-    return opik_url_override.rstrip("/api") + "/"
+    return opik_url_override.rstrip("/").removesuffix("/api") + "/"
 
 
 def get_experiment_url_by_id(
--- a/sdks/python/tests/unit/test_url_helpers.py
+++ b/sdks/python/tests/unit/test_url_helpers.py
@@ -1,5 +1,6 @@
 import pytest
 from opik import url_helpers
+from opik.config import OpikConfig
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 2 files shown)


alexkuzmik (approved):
Wow, good catch, thank you!

PR #7290, opened 2026-06-30
[NA] [SDK] fix: reject duplicate items in SpearmanRanking

### LINKED ISSUE #8275: [SDK] SpearmanRanking silently returns a wrong score for rankings with duplicate items
### Describe the bug

`SpearmanRanking.score()` validates that `output` and `reference` have the same length and the same *set* of items, but never checks that each sequence is actually a ranking (each item appearing exactly once). A duplicate item slips through both checks whenever it keeps the two sets equal, and then:

```python
ref_ranks = {item: idx for idx, item in enumerate(reference)}
```

silently keeps only the *last* occurrence's index for a repeated item, so the correlation is computed against a rank that isn't well-defined — no error, just a numeric result that doesn't mean what it claims to.

### Reproduction

```python
from opik.evaluation.metrics import SpearmanRanking

metric = SpearmanRanking()
result = metric.score(output=["a", "b", "a"], reference=["a", "a", "b"])
print(result.value, result.metadata)
# 0.625 {'rho': 0.25} -- no exception, even though "a" appears twice in
# both sequences and its rank is ambiguous.

result2 = metric.score(output=["a", "a", "b"], reference=["a", "b", "b"])
print(result2.value, result2.metadata)
# 0.875 {'rho': 0.75}
```

Both inputs have equal length (3) and equal item sets (`{"a", "b"}`), so the existing validation passes them through.

### Expected behavior

`score()` should raise `MetricComputationError` (as it already does for mismatched lengths, empty rankings, and mismatched item sets) when either `output` or `reference` contains a duplicate item, since the class's own docstring and formula assume two permutations of the same items.

### Reproduction environment

- `opik` (Python SDK), current `main`
- Verified with a local install: `pip install -e sdks/python`, Python 3.12

### Proposed fix

I have a fix + regression tests ready (adds a duplicate check via `len(set(x)) != len(x)` alongside the existing length/empty/set-equality checks, verified failing without the change and passing with it, `ruff format`/`mypy` clean). Happy to open the PR once this issue is confirmed/tracked, per the repo's contribution fast path.

This report was prepared with an AI assistant (Claude Code); I reviewed the reproduction and ran it locally against the current `main`.

anxkhn:
## Details

`SpearmanRanking.score()` used the rank-difference shortcut `rho = 1 - 6*sum(d^2)/(n*(n^2-1))`, which is only valid when both rankings are permutations of distinct items. The same-items guard used set equality, which ignores multiplicity, so rankings with duplicates slipped through and produced `rho` outside the documented `[-1, 1]` range (e.g. `["a","a","b"]` vs `["b","a","a"]` gave `rho=-1.25`, `value=-0.125`).
- Reject rankings that contain duplicates with `MetricComputationError`, matching the existing guard style; the distinct-item path, formula, and normalization are unchanged.
- Fix the docstring example, which claimed `rho=-0.5` for `["b","a","c"]` vs `["a","b","c"]` when it is `+0.5` (the existing test already asserts `0.5`).

Open question for maintainers: I chose to reject duplicates (minimal, matches the sibling same-items guard) rather than average tied ranks. If you'd prefer real tie handling (scipy-style average ranking), happy to change direction.

## Change checklist
- [x] User facing
- [x] Documentation update

## Issues

- Resolves #N/A
- OPIK-N/A

## Testing

Run from `sdks/python` (`uv run --no-sync`):
- `python -m pytest tests/unit/evaluation/metrics/test_heuristics.py -k spearman -p no:cacheprovider -v` -> 2 passed (existing distinct-item test plus the new duplicate test).
- `python -m pytest tests/unit/evaluation/metrics/test_heuristics.py -p no:cacheprovider` -> 74 passed.
- `python -m pytest tests/unit -p no:cacheprovider` -> metrics unit suite passed.
- `ruff check`, `ruff format --check`, `mypy` -> clean.

Scenarios validated: duplicate items on the output side and reference side now raise `MetricComputationError`; the distinct-item path, perfect ranking (1.0), single item, empty input, and length mismatch are unchanged. Environment: local, Python SDK only (no Docker).

## Documentation

Docstring example corrected (`rho -0.5` -> `0.5`); no separate documentation site change.


JetoPistola:
Thanks @anxkhn, and apologies for the long silence on this one.

**On your open question** — you asked whether to reject duplicates or do scipy-style tie averaging. Rejecting was the right call, and for the reason you identified: the shortcut formula `1 - 6Σd²/(n(n²-1))` is only valid for permutations of distinct items. Tie averaging would mean switching to the full Pearson-on-ranks form, which is a bigger change than this metric needs. If someone wants tie support later, that's a separate feature, not a bug fix.

**What happened to the duplicate-rejection half:** #8276 landed the same fix independently before we got to yours. That's on us for leaving this sitting — you filed first, on June 30. I've merged `main` in and resolved the conflict in favour of the version already in `main`, so this PR is now just your docstring correction.

**Your `["a","a","b"]` vs `["b","a","a"]` case was the best evidence in either PR** — it drove `rho` to `-1.25`, outside the documented `[-1, 1]` range, which is much harder to argue with than the plausible-looking numbers the other cases produced. I folded it into the merged test's parametrize block so it isn't lost.

**The docstring fix is still entirely yours and still needed** — `main` claims `rho = -0.5` for `["b","a","c"]` vs `["a","b","c"]` when it is `+0.5`. It has been wrong since the metric landed, and the existing test has been asserting `0.5` against it the whole time; the `# doctest: +SKIP` is why nobody noticed.

Merge commit credits you via `Co-Authored-By`. Verified locally: 76/76 heuristics tests pass. Thanks for the careful work here.

JetoPistola (approved):
Approving per my comment above — the docstring correction is right (+0.5, not -0.5) and 76/76 heuristics tests pass. Thanks @anxkhn!

alexkuzmik (approved):

PR #7299, opened 2026-07-01
[OPIK-7051] [BE] Fix: thread online-scoring errors as processing_errors, resilient per thread id

thiagohora:
## Details

On the online-scoring dashboard, client/config errors (`BadRequestException`, `ClientErrorException`) on the **trace-thread** streams appear under *Unexpected errors/s*, while the same errors on the trace/span streams correctly appear under *Scoring errors/s* (`processing_errors`).

Root cause: the thread scorers build their per-message chain from `Flux.fromIterable(threadIds).flatMap(...)`, whereas the trace/span scorers build from a `Mono`. `Flux.flatMap` is `onErrorContinue`-aware, so the enclosing `onErrorContinue` in `BaseRedisSubscriber.setupStreamListener` reaches into the scorer's `flatMap`, drops the failing element and records the failure via `*_unexpected_errors_total` — bypassing `processMessage`'s per-message `onErrorResume`.

Fix (both `OnlineScoringTraceThreadLlmAsJudgeScorer` and `OnlineScoringTraceThreadUserDefinedMetricPythonScorer`):
- Score each `thread_id` independently and **materialize** per-thread errors with `onErrorResume`, so a single thread's failure no longer stops scoring the sibling thread ids in the same message.
- After the batch, re-surface the first failure on the `Mono` error path (`collectList().flatMap(... Mono.error ...)`), so the message is still handled by `processMessage`'s `onErrorResume` — classified as a `processing_error` and following the normal retryable/non-retryable path, consistent with the trace/span scorers.

This both fixes the misclassification (no longer counted as "unexpected") and makes per-thread scoring resilient. Successfully-scored threads persist regardless; for the observed cause (a rule missing its provider API key, which fails every thread) the whole message fails as a non-retryable processing error, as expected.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- NA

## AI-WATERMARK

AI-WATERMARK: yes

- If yes:
  - Tools: Claude Code
  - Scope: Diagnosed the Flux-vs-Mono `onErrorContinue` interaction from the prod metrics, authored the per-thread-resilient error-materialization fix in both thread scorers, and this description.
  - Human verification: Author reviewed the diff. `mvn compile` + `spotless:check` on `opik-backend` pass (BUILD SUCCESS). Root cause confirmed against prod metrics: trace/span scorers carry `ClientErrorException`/`BadRequestException` under `processing_errors`, while the thread scorer's `BadRequestException` appeared only under `unexpected_errors`.

## Testing

- `mvn -o -DskipTests compile` and `mvn -o spotless:check` on `apps/opik-backend` → BUILD SUCCESS.
- Error-routing/resilience change only; no scoring logic added. Full resource/integration suite left to CI.

## Documentation

None.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringTraceThreadLlmAsJudgeScorer.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringTraceThreadLlmAsJudgeScorer.java
@@ -123,8 +123,18 @@ public class OnlineScoringTraceThreadLlmAsJudgeScorer extends OnlineScoringBaseS
                 message.projectId(), message.ruleId(), message.threadIds(), message.workspaceId());
 
         return Flux.fromIterable(message.threadIds())
-                .flatMap(threadId -> processThreadScores(message, threadId))
-                .then()
+                // Score each thread id independently: a single thread's failure must not stop scoring the
+                // sibling thread ids. Per-thread errors are materialized (onErrorResume) so the flatMap
+                // completes for every thread; the batch's first failure is then re-surfaced below. This keeps
+                // the failure on the Mono error path handled by BaseRedisSubscriber.processMessage's
+                // onErrorResume — classified as a processing error, following the normal retryable/
+                // non-retryable path — instead of leaking into the enclosing onErrorContinue via Flux.flatMap
+                // (which would drop the element and count it as an "unexpected" error).
+                .flatMap(threadId -> processThreadScores(message, threadId)
+                        .then(Mono.<Throwable>empty())
+                        .onErrorResume(Mono::just))
+                .collectList()
+                .flatMap(errors -> errors.isEmpty() ? Mono.<Void>empty() : Mono.error(errors.getFirst()))
                 .contextWrite(context -> context.put(RequestContext.WORKSPACE_ID, message.workspaceId())
                         .put(RequestContext.USER_NAME, message.userName())
                         .put(RequestContext.VISIBILITY, Visibility.PRIVATE))
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringTraceThreadUserDefinedMetricPythonScorer.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringTraceThreadUserDefinedMetricPythonScorer.java
@@ -103,8 +103,18 @@ public class OnlineScoringTraceThreadUserDefinedMetricPythonScorer
                 message.projectId(), message.ruleId(), message.workspaceId());
 
         return Flux.fromIterable(message.threadIds())
-                .flatMap(threadId -> processThreadScores(message, threadId))
-                .then()
+                // Score each thread id independently: a single thread's failure must not stop scoring the
+                // sibling thread ids. Per-thread errors are materialized (onErrorResume) so the flatMap
+                // completes for every thread; the batch's first failure is then re-surfaced below. This keeps
+                // the failure on the Mono error path handled by BaseRedisSubscriber.processMessage's
+                // onErrorResume — classified as a processing error, following the normal retryable/
+                // non-retryable path — instead of leaking into the enclosing onErrorContinue via Flux.flatMap
+                // (which would drop the element and count it as an "unexpected" error).
+                .flatMap(threadId -> processThreadScores(message, threadId)
+                        .then(Mono.<Throwable>empty())
+                        .onErrorResume(Mono::just))
+                .collectList()
+                .flatMap(errors -> errors.isEmpty() ? Mono.<Void>empty() : Mono.error(errors.getFirst()))
                 .contextWrite(context -> context.put(RequestContext.WORKSPACE_ID, message.workspaceId())
                         .put(RequestContext.USER_NAME, message.userName())
                         .put(RequestContext.VISIBILITY, Visibility.PRIVATE))


apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringTraceThreadLlmAsJudgeScorer.java
```diff
@@ -123,8 +123,18 @@ protected Mono<Void> score(@NonNull TraceThreadToScoreLlmAsJudge message) {
...
+                // sibling thread ids. Per-thread errors are materialized (onErrorResume) so the flatMap
+                // completes for every thread; the batch's first failure is then re-surfaced below. This keeps
+                // the failure on the Mono error path handled by BaseRedisSubscriber.processMessage's
+                // onErrorResume — classified as a processing error, following the normal retryable/
+                // non-retryable path — instead of leaking into the enclosing onErrorContinue via Flux.flatMap
+                // (which would drop the element and count it as an "unexpected" error).
+                .flatMap(threadId -> processThreadScores(message, threadId)
+                        .then(Mono.<Throwable>empty())
+                        .onErrorResume(Mono::just))
+                .collectList()
```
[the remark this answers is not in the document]
thiagohora: Not extracting this. The shared part is a ~6-line reactive pipeline; pulling it into an `OnlineScoringBaseScorer.processThreadBatch(...)` helper trades a tiny, readable, self-contained snippet for indirection on subtle reactive/error-handling code that's clearer read inline at the call site. The two thread scorers are the only callers and aren't expected to grow, so the sync-drift risk is low. (Minor correction: with this change a bad thread does not stop the batch — every thread id is attempted via the per-thread `onErrorResume`; only the classification of the surfaced failure changed.) Skipping the dedicated regression test on the same grounds; happy to add one if we later extract this.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringTraceThreadLlmAsJudgeScorer.java
```diff
@@ -123,8 +123,18 @@ protected Mono<Void> score(@NonNull TraceThreadToScoreLlmAsJudge message) {
...
+                .flatMap(errors -> errors.isEmpty() ? Mono.<Void>empty() : Mono.error(errors.getFirst()))
```
JetoPistola, line 137: ❓ **question** | Testing

The root cause, fix, and error classification all check out — verified against `BaseRedisSubscriber` (`onErrorContinue` at the stream level vs. `onErrorResume` in `processMessage`) and the `NON_RETRYABLE_EXCEPTIONS` set (`ClientErrorException`/`BadRequestException` are in it), so re-surfacing on the `Mono` path routes as a non-retryable processing error as described. No NPE risk either — `errors.getFirst()` is guarded by `isEmpty()`.

The change introduces two behaviors that aren't currently asserted by a test:
1. **Per-thread resilience**: one thread's failure no longer cancels its siblings — successfully-scored threads still persist.
2. **Error routing**: the batch's first failure re-surfaces on the `Mono` error path (processing error) rather than leaking to the enclosing `onErrorContinue` (unexpected error).

The existing unit tests mock at the scorer level and can't reach the subscriber's routing, so this path is currently CI/manual-only. Is a targeted test planned (e.g. a two-thread message where one `processThreadScores` errors — asserting the sibling's `storeThreadScores` still runs and the `Mono` completes with an error), or is that deliberately deferred given the wiring cost?

JetoPistola:
👋 **Review summary**

**What looks good**
- Sharp root-cause diagnosis: the `Flux.flatMap` vs `Mono` distinction w.r.t. `onErrorContinue`-awareness is subtle and correctly identified. Verified against `BaseRedisSubscriber` — the stream-level `onErrorContinue` really does reach into a `flatMap` and misclassify.
- The fix is minimal and correct: error materialization (`onErrorResume(Mono::just)` → `collectList` → guarded `getFirst()`) has no NPE risk, preserves per-thread resilience, and re-routes the failure through `processMessage`'s `onErrorResume` so it lands as a `processing_error` on the normal retryable/non-retryable path.
- Brings the thread scorers in line with the Mono-based trace/span scorers — good consistency.
- Thorough PR description with an honest human-verification section.

**Overall**
Well-scoped error-routing fix, no scoring logic touched. The reactive rewrite is sound and the behavior matches the stated contract.

**Inline comments**: 1 question (test coverage for the new resilience/routing path) — nothing blocking.

andrescrz (approved):
LGTM

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringTraceThreadLlmAsJudgeScorer.java
thiagohora: Deliberately deferred — and thanks for verifying the routing against `BaseRedisSubscriber` / `NON_RETRYABLE_EXCEPTIONS`. You've pinpointed the trade-off exactly. A targeted test here has to drive the full scorer with mocked per-thread deps (`traceService`/`retrieveFullThreadContext`, `getThreadModelId`, and the LLM/python call inside the private `processThreadScores`) so one thread errors while a sibling succeeds — meaningful wiring just to assert reactive plumbing. The cheap version would be to pull the fan-out into a `static` `OnlineScoringBaseScorer` helper and `StepVerifier` it, but we intentionally kept the ~6-line pipeline inline for readability (see the Baz thread above), which rules out an isolated unit test.

Given that, I've left it CI/manual-verified for now: the behavior is confirmed against prod metrics — the thread scorer's `BadRequestException` moved from `unexpected_errors` to `processing_errors`, and siblings are still scored — plus the `BaseRedisSubscriber` contract you already checked. Happy to add the scorer-level test (or revisit the extraction to make it cheap) if you'd like it in scope rather than deferred.

PR #7300, opened 2026-07-01
fix(frontend): scope logs existence probes by time range

### LINKED ISSUE #7298: [Bug]: MetricsTab trace existence probe omits time range and causes full project scan

- [ ] Opik Python SDK
- [ ] Opik Typescript SDK
- [x] Opik UI
- [x] Opik Server
- [ ] Documentation

### Opik version

- Opik version: 2.0.58
- Deployment: self-hosted Kubernetes / Helm chart `opik-2.0.58`
- ClickHouse: `25.3.8.10041.altinitystable`

### Describe the problem

This appears to be the same class of problem as #6148 / #6949, but from a different frontend call site.

On the project Logs page, the UI is scoped to `Past 24 hours`, and the main traces table request correctly includes a time range. However, an auxiliary traces request used as an existence probe is sent without `from_time` / `to_time`:

```text
GET /api/v1/private/traces/?project_id=<project_id>
  &filters=[{"field":"visibility_mode","operator":"=","value":"default"},{"field":"source","operator":"=","value":"sdk"}]
  &size=1
  &page=1
  &strip_attachments=true
```

The response is tiny (`size=1`), but the backend query is not cheap because it builds broad trace/span aggregations before applying the final page limit. On a project with ~59k matching traces and ~2.1M spans, this request took 16-24 seconds and scanned ~1.48 GiB in ClickHouse.

Expected behavior:

- If the page is scoped to `Past 24 hours`, this auxiliary request should either pass the same `from_time` / `to_time`, or use a cheaper endpoint/query for the `has traces` check.
- `size=1` should not require scanning/aggregating the full project when the UI already has a time range.

Actual behavior:

- The query log shows `uuidFromTime=null, uuidToTime=null` for this `size=1` request.
- ClickHouse scans the whole project range for matching traces/spans.
- The UI experiences 20s+ XHR latency even though the visible page is time-filtered.

1. Open a project with a large number of traces/spans.
2. Go to Logs with a time range selected, for example `Past 24 hours`.
3. Observe Network requests for `GET /api/v1/private/traces/`.
4. The main traces table request includes a time range, but the auxiliary `size=1` request does not include `from_time` / `to_time`.
5. Inspect ClickHouse `system.query_log` for `find_traces_by_project_id` with `page:1:size:1`.

Observed query metadata:

```text
event_time:        2026-07-01 07:35:43 UTC
query_duration_ms: 24091
read_rows:         9,086,776
read_bytes:        1.48 GiB
memory_usage:      1.16 GiB
log_comment:       find_traces_by_project_id:...:page:1:size:1:TraceSearchCriteria[..., truncate=false, stripAttachments=true, uuidFromTime=null, uuidToTime=null]
```

A simplified version of the generated SQL shape is:

```sql
WITH trace_id_prefilter AS (
    SELECT DISTINCT id
    FROM traces
    WHERE workspace_id = '<workspace_id>'
      AND project_id = '<project_id>'
      AND visibility_mode = 'default'
      AND (source = 'sdk' OR source = 'unknown')
),
spans_agg AS (
    SELECT
        trace_id,
        sumMap(usage) AS usage,
        sum(total_estimated_cost) AS total_estimated_cost,
        COUNT(DISTINCT id) AS span_count,
        toInt64(countIf(type = 'llm')) AS llm_span_count,
        countIf(type = 'tool') > 0 AS has_tool_spans,
        arraySort(groupUniqArrayIf(provider, provider != '')) AS providers
    FROM spans FINAL
    WHERE workspace_id = '<workspace_id>'
      AND project_id = '<project_id>'
      AND trace_id IN (SELECT id FROM trace_id_prefilter)
    GROUP BY workspace_id, project_id, trace_id
),
traces_deduped AS (
    SELECT t.*
    FROM traces t
    WHERE workspace_id = '<workspace_id>'
      AND project_id = '<project_id>'
      AND visibility_mode = 'default'
      AND (source = 'sdk' OR source = 'unknown')
    ORDER BY (workspace_id, project_id, id) DESC, last_updated_at DESC
    LIMIT 1 BY id
),
page_ids AS (
    SELECT td.id
    FROM traces_deduped td
    ORDER BY (workspace_id, project_id, id) DESC, last_updated_at DESC
    LIMIT 1 OFFSET 0
)
SELECT ...
FROM page_wide t
LEFT JOIN spans_agg s ON t.id = s.trace_id
ORDER BY (workspace_id, project_id, id) DESC, last_updated_at DESC;
```

When the same request includes a lower time bound, the query scans far less data in our environment (~1.1M rows / ~196 MiB instead of ~9.1M rows / ~1.48 GiB), and latency drops to ~1-3 seconds.

Likely frontend source: a traces existence probe such as `MetricsTab` calling `useTracesList({ projectId, page: 1, size: 1, truncate: true })` without passing the current interval's `fromTime` / `toTime`.

Suggested fixes:

- Pass the current page time range into the auxiliary `useTracesList` call, similar to the ThreadDetailsPanel fix in #6949.
- Or replace the existence probe with a cheaper endpoint/query that does not build span/feedback/comment/experiment aggregations.
- On the backend, consider paging trace IDs first and aggregating only the current page of trace IDs, so `size=1` cannot trigger full-project span aggregation.

### Error logs or stack trace

No exception; the request succeeds with HTTP 200 but is slow.

Backend access log showed repeated 19-27s requests for the same endpoint. ClickHouse query log shows the expensive `find_traces_by_project_id` queries above.

### Healthcheck results

Not attached. The deployment is healthy; this is a query performance issue triggered by a frontend request missing time bounds.

Levi-Wu-777:
## Summary

This updates the Logs page SDK existence probes to use the currently selected time range.

The traces/spans tab already applies `fromTime` / `toTime` to the main table, stats, and CSV export queries, but its `size=1` existence probe did not include those fields. The threads tab had the same gap. On large projects, those probes can scan all historical SDK traces/threads even when the user is looking at a bounded range such as past 24 hours.

Fixes #7298.

## Changes

- Pass `fromTime: intervalStart` and `toTime: intervalEnd` to the traces/spans SDK existence probe.
- Pass `fromTime: intervalStart` and `toTime: intervalEnd` to the threads SDK existence probe.

## Testing

- `cd apps/opik-frontend && npm ci`
- `cd apps/opik-frontend && npm run typecheck`
- `cd apps/opik-frontend && npm run lint`

### DIFF
--- a/apps/opik-frontend/src/v2/pages/LogsPage/ThreadsTab/ThreadsTab.tsx
+++ b/apps/opik-frontend/src/v2/pages/LogsPage/ThreadsTab/ThreadsTab.tsx
@@ -621,6 +621,8 @@ export const ThreadsTab: React.FC<ThreadsTabProps> = ({
       projectId,
       page: 1,
       size: 1,
+      fromTime: intervalStart,
+      toTime: intervalEnd,
       logsSource: LOGS_SOURCE.sdk,
     },
     {
--- a/apps/opik-frontend/src/v2/pages/LogsPage/TracesSpansTab/TracesSpansTab.tsx
+++ b/apps/opik-frontend/src/v2/pages/LogsPage/TracesSpansTab/TracesSpansTab.tsx
@@ -1217,6 +1217,8 @@ export const TracesSpansTab: React.FC<TracesSpansTabProps> = ({
       page: 1,
       size: 1,
       stripAttachments: true,
+      fromTime: intervalStart,
+      toTime: intervalEnd,
       logsSource: LOGS_SOURCE.sdk,
     },
     {


andrescrz:
Hi @Levi-Hope 

Thanks a lot for taking the time to put this together and contribute to the project—we really appreciate it.

We're going to close this PR for now. We actually tried a very similar approach in the past, but ended up having to revert it because it introduced more performance issues. 

The good news is that we already have a different solution that's nearly ready to merge and addresses the same underlying problem: https://github.com/comet-ml/opik/pull/7378.

Thanks again for the contribution and for helping improve Opik. Even though we won't be moving forward with this particular change, we really appreciate your effort and hope you'll continue contributing in the future.

PR #7307, opened 2026-07-01
[OPIK-6892] [BE] feat: NaN-aware aggregate validation + sentinel translation framework

andrescrz:
## Details

Preparation for the ClickHouse data-model revamp that will drop `Nullable(...)` from the high-volume `traces` analytics columns (`end_time`, `duration`, `ttft`) in favour of collision-free, index-friendly sentinels. This lands the framework and its tests now; the DDL cutover activates it in a later slice. Behavior is unchanged while the columns are still nullable — the new `traceColumnsNonNullable` toggle defaults off.

- `SentinelTranslation`: a single translation point mapping sentinel ↔ API `null` in both directions — epoch ↔ `null` for `DateTime64`, `NaN` ↔ `null` for `Float64`, empty/all-NUL ↔ `null` for `FixedString(36)`.
- `FilterQueryBuilder`: flag-gated sentinel-aware `end_time` (`nullIf(end_time, epoch)`), plus epoch/`NaN` guards on `duration`/`ttft` filters that are provable no-ops while the columns are nullable.
- DAOs (`TraceDAO`, `ThreadDAO`, `KpiCardDAO`, `ProjectMetricsDAO`, and the shared duration expression consumers): bind sentinels on write and translate them back on read under the flag; NaN-safe aggregates (`avg` wrapped with `if(isNaN(...))`, `quantile` relies on native NaN-skipping).
- New `databaseAnalyticsDataModel.traceColumnsNonNullable` config toggle (default `false`), intended to be flipped in lockstep with the cutover EXCHANGE.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves OPIK-6892

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: assisted — implementation and tests
- Human verification: code review + integration tests on the target ClickHouse image

## Testing

Backend unit and integration tests (integration via Testcontainers on the target ClickHouse image `25.8.16.34`), green locally:

- `SentinelTranslationTest` — sentinel ↔ `null` translation across all branches (null / sentinel / real values, ±infinity pass-through, empty and all-NUL `FixedString(36)` forms).
- `FilterQueryBuilderSentinelTest` — flag-gated `end_time` and unconditional `duration`/`ttft` filter SQL, asserted for both flag states.
- `NaNAwareAggregateIntegrationTest` — ClickHouse `NaN` aggregate semantics (`min`/`max`/`quantile` skip `NaN`; the `<agg>If(NOT isNaN(x))` form restores NULL-skipping for `sum`/`avg`/`count`) and the guarded `duration` materialized formula (emits `NaN` at the epoch sentinel).
- `TraceSentinelIntegrationTest` — flag-ON end-to-end wiring against the still-nullable schema (valid because the sentinels are ordinary nullable values): create / batch / update / thread round-trips, a raw physical-write proof that the R2DBC driver serializes a bound `Double.NaN` to ClickHouse `nan` on write (including per-row binding in a mixed batch), sentinel-aware filter exclusion, and `end_time` sort ordering.

The full backend suite runs in CI. Flag-off behavior (the default) is unchanged and is exercised by the existing trace/thread/metrics suites.

## Documentation

N/A — no user-facing or SDK documentation changes; the data-model revamp design is tracked in the Jira ticket.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemDAO.java
@@ -658,7 +658,7 @@ class DatasetItemDAOImpl implements DatasetItemDAO {
                 FROM (
                     SELECT
                         id,
-                       if(end_time IS NOT NULL AND start_time IS NOT NULL
+                       if(end_time IS NOT NULL AND notEquals(end_time, toDateTime64('1970-01-01 00:00:00.000', 9)) AND start_time IS NOT NULL
                                              AND notEquals(start_time, toDateTime64('1970-01-01 00:00:00.000', 9)),
                                          (dateDiff('microsecond', start_time, end_time) / 1000.0),
                                          NULL) AS duration,
@@ -938,7 +938,7 @@ class DatasetItemDAOImpl implements DatasetItemDAO {
                 LEFT JOIN (
                     SELECT
                         id,
-                        if(end_time IS NOT NULL AND start_time IS NOT NULL
+                        if(end_time IS NOT NULL AND notEquals(end_time, toDateTime64('1970-01-01 00:00:00.000', 9)) AND start_time IS NOT NULL
                             AND notEquals(start_time, toDateTime64('1970-01-01 00:00:00.000', 9)),
                             (dateDiff('microsecond', start_time, end_time) / 1000.0),
                             NULL) as duration
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/KpiCardDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/KpiCardDAO.java
@@ -12,6 +12,7 @@ import com.comet.opik.domain.filter.FilterStrategy;
 import com.comet.opik.infrastructure.OpikConfiguration;
 import com.comet.opik.infrastructure.db.TransactionTemplateAsync;
 import com.comet.opik.infrastructure.instrumentation.InstrumentAsyncUtils;
+import com.comet.opik.utils.SentinelTranslation;
 import com.google.inject.ImplementedBy;
 import io.r2dbc.spi.Connection;
 import io.r2dbc.spi.Result;
@@ -692,9 +693,7 @@ class KpiCardDAOImpl implements KpiCardDAO {
     }
 
     private Double filterNan(Double value) {
-        if (value == null) {
-            return null;
-        }
-        return value.isNaN() ? null : value;
+        // Delegates to the single SentinelTranslation definition.
+        return SentinelTranslation.nanToNull(value);
     }
 }
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/KpiCardDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/KpiCardDAO.java
@@ -9,6 +9,7 @@ import com.comet.opik.api.metrics.KpiCardResponse.KpiMetric;
 import com.comet.opik.api.metrics.KpiCardResponse.KpiMetricType;
 import com.comet.opik.domain.filter.FilterQueryBuilder;
 import com.comet.opik.domain.filter.FilterStrategy;
+import com.comet.opik.infrastructure.OpikConfiguration;
 import com.comet.opik.infrastructure.db.TransactionTemplateAsync;
 import com.comet.opik.infrastructure.instrumentation.InstrumentAsyncUtils;
 import com.google.inject.ImplementedBy;
@@ -51,6 +52,7 @@ class KpiCardDAOImpl implements KpiCardDAO {
 
     private final @NonNull TransactionTemplateAsync template;
     private final @NonNull InstantToUUIDMapper instantToUUIDMapper;
+    private final @NonNull OpikConfiguration configuration;
 
     private static final String GET_TRACE_KPI_CARDS = """
             WITH feedback_scores_deduped AS (
@@ -145,7 +147,7 @@ class KpiCardDAOImpl implements KpiCardDAO {
                 FROM (
                     SELECT
                         id,
-                        if(end_time IS NOT NULL AND start_time IS NOT NULL
+                        if(end_time IS NOT NULL AND notEquals(end_time, toDateTime64('1970-01-01 00:00:00.000', 9)) AND start_time IS NOT NULL
                              AND notEquals(start_time, toDateTime64('1970-01-01 00:00:00.000', 9)),
                          (dateDiff('microsecond', start_time, end_time) / 1000.0),
                          NULL) AS duration,
@@ -444,13 +446,13 @@ class KpiCardDAOImpl implements KpiCardDAO {
                         t.project_id as project_id,
                         min(t.start_time) as start_time,
                         max(t.end_time) as end_time,
-                        if(max(t.end_time) IS NOT NULL AND min(t.start_time) IS NOT NULL
+                        if(max(t.end_time) IS NOT NULL AND notEquals(max(t.end_time), toDateTime64('1970-01-01 00:00:00.000', 9)) AND min(t.start_time) IS NOT NULL
                                AND notEquals(min(t.start_time), toDateTime64('1970-01-01 00:00:00.000', 9)),
                            (dateDiff('microsecond', min(t.start_time), max(t.end_time)) / 1000.0),
                            NULL) AS duration,
                         count(DISTINCT t.id) * 2 as number_of_messages,
                         <if(trace_thread_first_message_filter)>argMin(t.input, t.start_time) as first_message,<endif>
-                        <if(trace_thread_last_message_filter)>argMax(t.output, t.end_time) as last_message,<endif>
+                        <if(trace_thread_last_message_filter)>argMax(t.output, nullIf(t.end_time, toDateTime64('1970-01-01 00:00:00.000', 9))) as last_message,<endif>
                         max(t.last_updated_at) as last_updated_at,
                         argMax(t.last_updated_by, t.last_updated_at) as last_updated_by,
                         argMin(t.created_by, t.created_at) as created_by,
@@ -581,7 +583,7 @@ class KpiCardDAOImpl implements KpiCardDAO {
 
     private void addTraceFilters(ST template, List<? extends Filter> filters) {
         Optional.ofNullable(filters).ifPresent(f -> {
-            FilterQueryBuilder.toAnalyticsDbFilters(f, FilterStrategy.TRACE)
+            FilterQueryBuilder.toAnalyticsDbFilters(f, FilterStrategy.TRACE, traceColumnsNonNullable())
                     .ifPresent(traceFilters -> template.add("trace_filters", traceFilters));
             FilterQueryBuilder.toAnalyticsDbFilters(f, FilterStrategy.FEEDBACK_SCORES)
                     .ifPresent(scoresFilters -> template.add("trace_feedback_scores_filters", scoresFilters));
@@ -590,6 +592,10 @@ class KpiCardDAOImpl implements KpiCardDAO {
             FilterQueryBuilder.hasGuardrailsFilter(f)
                     .ifPresent(hasGuardrails -> template.add("guardrails_filters", true));
         });
+    }
+
+    private boolean traceColumnsNonNullable() {
+        return configuration.getDatabaseAnalyticsDataModel().traceColumnsNonNullable();
     }
 
     private void bindTraceFilters(Statement statement, List<? extends Filter> filters) {
@@ -621,7 +627,7 @@ class KpiCardDAOImpl implements KpiCardDAO {
 
     private void addThreadFilters(ST template, List<? extends Filter> filters) {
         Optional.ofNullable(filters).ifPresent(f -> {
-            FilterQueryBuilder.toAnalyticsDbFilters(f, FilterStrategy.TRACE_THREAD)
+            FilterQueryBuilder.toAnalyticsDbFilters(f, FilterStrategy.TRACE_THREAD, traceColumnsNonNullable())
                     .ifPresent(threadFilters -> template.add("trace_thread_filters", threadFilters));
             // first_message/last_message aggregate the full input/output payloads, so only project
             // them in threads_filtered when a filter actually references them (otherwise the thread
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/KpiCardDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/KpiCardDAO.java
@@ -693,7 +693,6 @@ class KpiCardDAOImpl implements KpiCardDAO {
     }
 
     private Double filterNan(Double value) {
-        // Delegates to the single SentinelTranslation definition.
         return SentinelTranslation.nanToNull(value);
     }
 }
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectMetricsDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectMetricsDAO.java
@@ -8,6 +8,7 @@ import com.comet.opik.api.metrics.MetricType;
 import com.comet.opik.api.metrics.ProjectMetricRequest;
 import com.comet.opik.domain.filter.FilterQueryBuilder;
 import com.comet.opik.domain.filter.FilterStrategy;
+import com.comet.opik.infrastructure.OpikConfiguration;
 import com.comet.opik.infrastructure.db.TransactionTemplateAsync;
 import com.comet.opik.infrastructure.instrumentation.InstrumentAsyncUtils;
 import com.comet.opik.utils.RowUtils;
@@ -150,6 +151,7 @@ class ProjectMetricsDAOImpl implements ProjectMetricsDAO {
     private final @NonNull TransactionTemplateAsync template;
     private final @NonNull FilterQueryBuilder filterQueryBuilder;
     private final @NonNull InstantToUUIDMapper instantToUUIDMapper;
+    private final @NonNull OpikConfiguration configuration;
 
     private static final Map<TimeInterval, String> INTERVAL_TO_SQL = Map.of(
             TimeInterval.WEEKLY, "toIntervalWeek(1)",
@@ -545,12 +547,12 @@ class ProjectMetricsDAOImpl implements ProjectMetricsDAO {
                         t.project_id as project_id,
                         min(t.start_time) as start_time,
                         max(t.end_time) as end_time,
-                        if(max(t.end_time) IS NOT NULL AND min(t.start_time) IS NOT NULL
+                        if(max(t.end_time) IS NOT NULL AND notEquals(max(t.end_time), toDateTime64('1970-01-01 00:00:00.000', 9)) AND min(t.start_time) IS NOT NULL
                                AND notEquals(min(t.start_time), toDateTime64('1970-01-01 00:00:00.000', 9)),
                            (dateDiff('microsecond', min(t.start_time), max(t.end_time)) / 1000.0),
                            NULL) AS duration,
                         <if(truncate)> replaceRegexpAll(argMin(t.input, t.start_time), '<truncate>', '"[image]"') as first_message <else> argMin(t.input, t.start_time) as first_message<endif>,
-                        <if(truncate)> replaceRegexpAll(argMax(t.output, t.end_time), '<truncate>', '"[image]"') as last_message <else> argMax(t.output, t.end_time) as last_message<endif>,
+                        <if(truncate)> replaceRegexpAll(argMax(t.output, nullIf(t.end_time, toDateTime64('1970-01-01 00:00:00.000', 9))), '<truncate>', '"[image]"') as last_message <else> argMax(t.output, nullIf(t.end_time, toDateTime64('1970-01-01 00:00:00.000', 9))) as last_message<endif>,
                         count(DISTINCT t.id) * 2 as number_of_messages,
                         max(t.last_updated_at) as last_updated_at,
                         argMax(t.last_updated_by, t.last_updated_at) as last_updated_by,
@@ -1069,7 +1071,7 @@ class ProjectMetricsDAOImpl implements ProjectMetricsDAO {
 
     private static final String GET_AVERAGE_DURATION = """
             SELECT
-                avg(duration) AS avg_duration
+                avg(if(isNaN(duration), NULL, duration)) AS avg_duration
             FROM traces final
             WHERE workspace_id = :workspace_id
                 <if(project_ids)> AND project_id IN :project_ids <endif>
@@ -1183,7 +1185,7 @@ class ProjectMetricsDAOImpl implements ProjectMetricsDAO {
     private static final String GET_TRACE_AVERAGE_DURATION = """
             %s
             SELECT <bucket> AS bucket,
-                   avg(duration) AS avg_duration
+                   avg(if(isNaN(duration), NULL, duration)) AS avg_duration
             FROM traces_filtered
             GROUP BY bucket
             ORDER BY bucket
@@ -1741,7 +1743,8 @@ class ProjectMetricsDAOImpl implements ProjectMetricsDAO {
             if (THREAD_METRICS.contains(metricType)) {
                 Optional.ofNullable(request.threadFilters())
                         .ifPresent(filters -> {
-                            filterQueryBuilder.toAnalyticsDbFilters(filters, FilterStrategy.TRACE_THREAD)
+                            FilterQueryBuilder.toAnalyticsDbFilters(
+                                    filters, FilterStrategy.TRACE_THREAD, traceColumnsNonNullable())
                                     .ifPresent(threadFilters -> template.add("trace_thread_filters", threadFilters));
                             filterQueryBuilder.toAnalyticsDbFilters(filters, FilterStrategy.FEEDBACK_SCORES)
                                     .ifPresent(
@@ -1772,7 +1775,8 @@ class ProjectMetricsDAOImpl implements ProjectMetricsDAO {
             } else {
                 Optional.ofNullable(request.traceFilters())
                         .ifPresent(filters -> {
-                            filterQueryBuilder.toAnalyticsDbFilters(filters, FilterStrategy.TRACE)
+                            FilterQueryBuilder.toAnalyticsDbFilters(
+                                    filters, FilterStrategy.TRACE, traceColumnsNonNullable())
                                     .ifPresent(traceFilters -> template.add("trace_filters", traceFilters));
                             filterQueryBuilder.toAnalyticsDbFilters(filters, FilterStrategy.FEEDBACK_SCORES)
                                     .ifPresent(
@@ -1837,6 +1841,10 @@ class ProjectMetricsDAOImpl implements ProjectMetricsDAO {
             return Mono.from(statement.execute())
                     .doFinally(signalType -> endSegment(segment));
         });
+    }
+
+    private boolean traceColumnsNonNullable() {
+        return configuration.getDatabaseAnalyticsDataModel().traceColumnsNonNullable();
     }
 
     private static final Set<MetricType> SPAN_TIME_METRICS = EnumSet.of(
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
@@ -3,9 +3,11 @@ package com.comet.opik.domain;
 import com.comet.opik.api.ProjectStats;
 import com.comet.opik.api.TraceThread;
 import com.comet.opik.api.TraceThreadStatus;
+import com.comet.opik.api.sorting.SortableFields;
 import com.comet.opik.api.sorting.TraceThreadSortingFactory;
 import com.comet.opik.domain.sorting.SortingQueryBuilder;
 import com.comet.opik.domain.stats.StatsMapper;
+import com.comet.opik.infrastructure.OpikConfiguration;
 import com.comet.opik.infrastructure.db.TransactionTemplateAsync;
 import com.comet.opik.infrastructure.instrumentation.InstrumentAsyncUtils;
 import com.comet.opik.utils.TruncationUtils;
@@ -15,6 +17,7 @@ import com.google.inject.ImplementedBy;
 import io.opentelemetry.instrumentation.annotations.WithSpan;
 import io.r2dbc.spi.Connection;
 import io.r2dbc.spi.Result;
+import io.r2dbc.spi.Row;
 import jakarta.inject.Inject;
 import jakarta.inject.Singleton;
 import lombok.NonNull;
@@ -43,6 +46,7 @@ import static com.comet.opik.infrastructure.instrumentation.InstrumentAsyncUtils
 import static com.comet.opik.infrastructure.instrumentation.InstrumentAsyncUtils.startSegment;
 import static com.comet.opik.utils.AsyncUtils.makeFluxContextAware;
 import static com.comet.opik.utils.AsyncUtils.makeMonoContextAware;
+import static com.comet.opik.utils.SentinelTranslation.epochToNull;
 import static java.util.function.Predicate.not;
 
 @ImplementedBy(ThreadDAOImpl.class)
@@ -121,7 +125,7 @@ class ThreadDAOImpl implements ThreadDAO {
                 ) AS ptt ON pt.thread_id = ptt.thread_id
                 ORDER BY if(ptt.last_updated_at = toDateTime64(0, 6, 'UTC'), pt.trace_last_updated_at, ptt.last_updated_at) DESC,
                     pt.start_time ASC,
-                    pt.end_time DESC,
+                    nullIf(pt.end_time, toDateTime64('1970-01-01 00:00:00.000', 9)) DESC,
                     pt.thread_id
                 LIMIT :limit <if(offset)>OFFSET :offset<endif>
             ), <endif>traces_final AS (
@@ -416,18 +420,18 @@ class ThreadDAOImpl implements ThreadDAO {
                     t.project_id as project_id,
                     min(t.start_time) as start_time,
                     max(t.end_time) as end_time,
-                    if(end_time IS NOT NULL AND start_time IS NOT NULL
+                    if(end_time IS NOT NULL AND notEquals(end_time, toDateTime64('1970-01-01 00:00:00.000', 9)) AND start_time IS NOT NULL
                            AND notEquals(start_time, toDateTime64('1970-01-01 00:00:00.000', 9)),
                        (dateDiff('microsecond', start_time, end_time) / 1000.0),
                        NULL) AS duration,
                     argMin(t.input, t.start_time) as first_message,
-                    argMax(t.output, t.end_time) as last_message,
+                    argMax(t.output, nullIf(t.end_time, toDateTime64('1970-01-01 00:00:00.000', 9))) as last_message,
                     argMin(t.truncated_input, t.start_time) as truncated_first_message,
-                    argMax(t.truncated_output, t.end_time) as truncated_last_message,
+                    argMax(t.truncated_output, nullIf(t.end_time, toDateTime64('1970-01-01 00:00:00.000', 9))) as truncated_last_message,
                     argMin(t.input_length, t.start_time) as first_message_length,
-                    argMax(t.output_length, t.end_time) as last_message_length,
+                    argMax(t.output_length, nullIf(t.end_time, toDateTime64('1970-01-01 00:00:00.000', 9))) as last_message_length,
                     argMin(t.truncation_threshold, t.start_time) as first_message_truncation_threshold,
-                    argMax(t.truncation_threshold, t.end_time) as last_message_truncation_threshold,
+                    argMax(t.truncation_threshold, nullIf(t.end_time, toDateTime64('1970-01-01 00:00:00.000', 9))) as last_message_truncation_threshold,
                     count(DISTINCT t.id) * 2 as number_of_messages,
                     sum(s.total_estimated_cost) as total_estimated_cost,
                     sumMap(s.usage) as usage,
@@ -478,7 +482,7 @@ class ThreadDAOImpl implements ThreadDAO {
             <if(stream)>
             ORDER BY workspace_id, project_id, thread_model_id DESC
             <else>
-            <if(sort_fields)> ORDER BY <sort_fields>, last_updated_at DESC <else> ORDER BY last_updated_at DESC, start_time ASC, end_time DESC <endif>
+            <if(sort_fields)> ORDER BY <sort_fields>, last_updated_at DESC <else> ORDER BY last_updated_at DESC, start_time ASC, nullIf(end_time, toDateTime64('1970-01-01 00:00:00.000', 9)) DESC <endif>
             <endif>
             LIMIT :limit <if(page_pushdown)><else><if(offset)>OFFSET :offset<endif><endif>
             SETTINGS log_comment = '<log_comment>'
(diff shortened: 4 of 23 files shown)


apps/opik-backend/src/main/java/com/comet/opik/domain/filter/FilterQueryBuilder.java
```diff
@@ -78,9 +85,19 @@ public class FilterQueryBuilder {
     private static final String USAGE_PROMPT_TOKENS_ANALYTICS_DB = "usage['prompt_tokens']";
     private static final String USAGE_TOTAL_TOKENS_ANALYTICS_DB = "usage['total_tokens']";
     private static final String VALUE_ANALYTICS_DB = "value";
-    private static final String DURATION_ANALYTICS_DB = "if(end_time IS NOT NULL AND start_time IS NOT NULL AND notEquals(start_time, toDateTime64('1970-01-01 00:00:00.000', 9)), (dateDiff('microsecond', start_time, end_time) / 1000.0), 0)";
+    /**
+     * Duration (ms) derived from {@code start_time}/{@code end_time}. The {@code notEquals(end_time, epoch)} guard, as
+     * in every other duration calc, keeps an absent (epoch-sentinel) {@code end_time} from yielding a garbage negative
+     * duration once the column is non-nullable. No-op while it is Nullable (an absent value reads as {@code NULL}).
+     */
+    private static final String DURATION_ANALYTICS_DB = "if(end_time IS NOT NULL AND notEquals(end_time, toDateTime64('1970-01-01 00:00:00.000', 9)) AND start_time IS NOT NULL AND notEquals(start_time, toDateTime64('1970-01-01 00:00:00.000', 9)), (dateDiff('microsecond', start_time, end_time) / 1000.0), 0)";
```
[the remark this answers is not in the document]
andrescrz: Skipping — the `else 0` predates this PR; the change here only added the `notEquals(end_time, epoch)` guard. Flag-off, an absent `end_time` is bound as `NULL`, so the filter still resolves to `0` exactly as before — behavior is unchanged. Whether duration filters should treat unfinished traces as `NULL` vs `0` is a pre-existing product question, out of scope for this framework PR.

apps/opik-backend/src/main/java/com/comet/opik/domain/KpiCardDAO.java
```diff
@@ -145,7 +147,7 @@ traces_filtered AS (
                 FROM (
                     SELECT
                         id,
-                        if(end_time IS NOT NULL AND start_time IS NOT NULL
+                        if(end_time IS NOT NULL AND notEquals(end_time, toDateTime64('1970-01-01 00:00:00.000', 9)) AND start_time IS NOT NULL
                              AND notEquals(start_time, toDateTime64('1970-01-01 00:00:00.000', 9)),
                          (dateDiff('microsecond', start_time, end_time) / 1000.0),
                          NULL) AS duration,
```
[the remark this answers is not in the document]
andrescrz: Skipping — these duration expressions differ by context (table aliases, materialized-column vs computed) and live in DAOs with distinct responsibilities; folding them into one shared fragment would couple those DAOs against SRP for surface-level dedup. The duplication is also largely pre-existing — this PR only added the epoch guard to each copy.

apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
```diff
@@ -416,18 +420,18 @@ ORDER BY (workspace_id, project_id, entity_id, name) DESC, last_updated_at DESC
...
+                    if(end_time IS NOT NULL AND notEquals(end_time, toDateTime64('1970-01-01 00:00:00.000', 9)) AND start_time IS NOT NULL
                            AND notEquals(start_time, toDateTime64('1970-01-01 00:00:00.000', 9)),
                        (dateDiff('microsecond', start_time, end_time) / 1000.0),
                        NULL) AS duration,
                     argMin(t.input, t.start_time) as first_message,
-                    argMax(t.output, t.end_time) as last_message,
+                    argMax(t.output, nullIf(t.end_time, toDateTime64('1970-01-01 00:00:00.000', 9))) as last_message,
                     argMin(t.truncated_input, t.start_time) as truncated_first_message,
-                    argMax(t.truncated_output, t.end_time) as truncated_last_message,
+                    argMax(t.truncated_output, nullIf(t.end_time, toDateTime64('1970-01-01 00:00:00.000', 9))) as truncated_last_message,
```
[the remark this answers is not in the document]
andrescrz: Skipping — already resolved: the trace `end_time` write is now flag-gated, so flag-off it binds SQL `NULL` (never the epoch). Pre-cutover `traces.end_time` is therefore only ever `NULL` or a real timestamp, so the unconditional `nullIf(end_time, epoch)` is a no-op on all flag-off data and the premise (forwarding `Instant.EPOCH` flag-off) no longer holds.

apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
```diff
@@ -973,18 +977,18 @@ ORDER BY (workspace_id, project_id, entity_id, id) DESC, last_updated_at DESC
                     t.project_id as project_id,
                     min(t.start_time) as start_time,
                     max(t.end_time) as end_time,
-                    if(end_time IS NOT NULL AND start_time IS NOT NULL
+                    if(end_time IS NOT NULL AND notEquals(end_time, toDateTime64('1970-01-01 00:00:00.000', 9)) AND start_time IS NOT NULL
                            AND notEquals(start_time, toDateTime64('1970-01-01 00:00:00.000', 9)),
                        (dateDiff('microsecond', start_time, end_time) / 1000.0),
                        NULL) AS duration,
```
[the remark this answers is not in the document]
andrescrz: Skipping — pre-cutover `traces.end_time` is never the epoch (absent is written as `NULL` via the flag-gated write; no real event ends at 1970-01-01), so `nullIf(end_time, epoch)` drops nothing on flag-off data. It is unconditional by design so the stats stay correct post-cutover without a second flag flip.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -372,7 +377,7 @@ INSERT INTO traces (
                     :workspace_id as workspace_id,
                     :name as name,
                     parseDateTime64BestEffort(:start_time, 9) as start_time,
-                    <if(end_time)> parseDateTime64BestEffort(:end_time, 9) as end_time, <else> null as end_time, <endif>
+                    parseDateTime64BestEffort(:end_time, 9) as end_time,
```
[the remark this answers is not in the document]
andrescrz: Skipping — flag-off, an absent `end_time` binds a typed `NULL` and `parseDateTime64BestEffort(NULL, 9)` returns `NULL` (not an error) on the target ClickHouse image. This is exercised by the existing flag-off create-trace tests (null `end_time`) and CI is green. The template is unconditional by necessity so flag-on writes the epoch sentinel; the flag-off effect is unchanged.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -1966,7 +1971,7 @@ INSERT INTO traces (
                     :workspace_id as workspace_id,
                     <if(name)> :name <else> '' <endif> as name,
                     toDateTime64('1970-01-01 00:00:00.000', 9) as start_time,
-                    <if(end_time)> parseDateTime64BestEffort(:end_time, 9) <else> null <endif> as end_time,
+                    parseDateTime64BestEffort(:end_time, 9) as end_time,
                     <if(input)> :input <else> '' <endif> as input,
```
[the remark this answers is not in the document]
andrescrz: Skipping — same as the insert path: flag-off binds a typed `NULL` and `parseDateTime64BestEffort(NULL, 9)` yields `NULL`, covered green by the flag-off partial-update tests. No failure; behavior preserved.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemDAO.java
```diff
@@ -658,7 +658,7 @@ LEFT JOIN (
                 FROM (
                     SELECT
                         id,
-                       if(end_time IS NOT NULL AND start_time IS NOT NULL
+                       if(end_time IS NOT NULL AND notEquals(end_time, toDateTime64('1970-01-01 00:00:00.000', 9)) AND start_time IS NOT NULL
                                              AND notEquals(start_time, toDateTime64('1970-01-01 00:00:00.000', 9)),
                                          (dateDiff('microsecond', start_time, end_time) / 1000.0),
                                          NULL) AS duration,
```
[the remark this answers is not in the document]
andrescrz: Skipping — flag-off, `traces.end_time` is never the epoch (absent → `NULL` via the gated write), so the `end_time = epoch → NULL` guard is a no-op on all real flag-off data and `avg(duration)` is unaffected. Unconditional by design for post-cutover correctness.

apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
```diff
@@ -121,7 +125,7 @@ ORDER BY (workspace_id, project_id, thread_id, id) DESC, last_updated_at DESC
                 ) AS ptt ON pt.thread_id = ptt.thread_id
                 ORDER BY if(ptt.last_updated_at = toDateTime64(0, 6, 'UTC'), pt.trace_last_updated_at, ptt.last_updated_at) DESC,
                     pt.start_time ASC,
-                    pt.end_time DESC,
+                    nullIf(pt.end_time, toDateTime64('1970-01-01 00:00:00.000', 9)) DESC,
                     pt.thread_id
```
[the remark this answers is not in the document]
andrescrz: Skipping — flag-off, `end_time` is `NULL` or a real timestamp, never the epoch, so `nullIf(end_time, epoch)` equals plain `end_time` for all real data and the ordering is unchanged pre-cutover. It is unconditional so the order is also correct post-cutover without a second flip.

apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
```diff
@@ -1282,12 +1286,12 @@ ORDER BY (workspace_id, project_id, entity_id, name) DESC, last_updated_at DESC
                         t.project_id as project_id,
                         min(t.start_time) as start_time,
                         max(t.end_time) as end_time,
-                        if(max(t.end_time) IS NOT NULL AND min(t.start_time) IS NOT NULL
+                        if(max(t.end_time) IS NOT NULL AND notEquals(max(t.end_time), toDateTime64('1970-01-01 00:00:00.000', 9)) AND min(t.start_time) IS NOT NULL
                                AND notEquals(min(t.start_time), toDateTime64('1970-01-01 00:00:00.000', 9)),
                            (dateDiff('microsecond', min(t.start_time), max(t.end_time)) / 1000.0),
                            NULL) AS duration,
                         argMin(t.input, t.start_time) as first_message,
```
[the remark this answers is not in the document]
andrescrz: Skipping — same as the KpiCardDAO note: the three thread-duration expressions sit in queries with different shapes/aliases; extracting a shared fragment would couple them against SRP for marginal dedup. Leaving as-is.

apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectMetricsDAO.java
```diff
@@ -545,12 +547,12 @@ threads_filtered AS (
...
-                        if(max(t.end_time) IS NOT NULL AND min(t.start_time) IS NOT NULL
+                        if(max(t.end_time) IS NOT NULL AND notEquals(max(t.end_time), toDateTime64('1970-01-01 00:00:00.000', 9)) AND min(t.start_time) IS NOT NULL
                                AND notEquals(min(t.start_time), toDateTime64('1970-01-01 00:00:00.000', 9)),
                            (dateDiff('microsecond', min(t.start_time), max(t.end_time)) / 1000.0),
                            NULL) AS duration,
                         <if(truncate)> replaceRegexpAll(argMin(t.input, t.start_time), '<truncate>', '"[image]"') as first_message <else> argMin(t.input, t.start_time) as first_message<endif>,
-                        <if(truncate)> replaceRegexpAll(argMax(t.output, t.end_time), '<truncate>', '"[image]"') as last_message <else> argMax(t.output, t.end_time) as last_message<endif>,
+                        <if(truncate)> replaceRegexpAll(argMax(t.output, nullIf(t.end_time, toDateTime64('1970-01-01 00:00:00.000', 9))), '<truncate>', '"[image]"') as last_message <else> argMax(t.output, nullIf(t.end_time, toDateTime64('1970-01-01 00:00:00.000', 9))) as last_message<endif>,
```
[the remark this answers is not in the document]
andrescrz: Skipping — flag-off, `traces.end_time` is never the epoch (absent → `NULL` via the gated write), so the `nullIf(..., epoch)` guards are no-ops on real flag-off data; no epoch-ended traces are excluded from the percentiles or `last_message`. Unconditional by design for post-cutover correctness.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -3577,6 +3630,11 @@ public Mono<Void> partialInsert(
             bindUserNameAndWorkspace(statement, userName, workspaceId);
             bindUpdateParams(traceUpdate, statement);
 
+            // INSERT_UPDATE builds the full new_trace row, so end_time/ttft are referenced unconditionally and must
+            // always be bound (sentinel or null for an absent value) — unlike the conditional UPDATE keep-column path.
+            bindEpochSentinel(statement, "end_time", traceUpdate.endTime());
+            bindNanSentinel(statement, "ttft", traceUpdate.ttft());
```
JetoPistola, line 3636: 🧹 **nit** | Redundancy

On the `partialInsert` path, `end_time`/`ttft` are now bound twice for a *present* value: `bindUpdateParams` binds them conditionally (`endTime.toString()` / `ttft`), then these two lines rebind unconditionally via `bindEpochSentinel`/`bindNanSentinel` (`formatNanos(...)` / sentinel). The unconditional bind runs last and wins, so it's correct — the earlier binds in `bindUpdateParams` are just dead for these two fields here.

Can't cleanly drop them from `bindUpdateParams` since the `update()` path still needs them for the `UPDATE` template's `<if(end_time)> :end_time`. Purely cosmetic; flagging only so it's a conscious choice rather than an accident.

JetoPistola:
❓ **question** | Cutover completeness (read-side `duration` NaN guard)

The on-read `if(isNaN(duration), NULL, duration)` guard was added to `ExperimentItemDAO`, `ExperimentAggregatesDAO`, `DatasetItemVersionDAO`, and the `avg(duration)` sites in `ProjectMetricsDAO` — but a few parallel sites that read the materialized trace `duration` were left unwrapped:

- `ExperimentDAO`: line ~363 projects `duration` straight from `traces`, feeding `quantiles(0.5,0.9,0.99)(duration)` (~352, ~1290) and `avgMap(duration)` (~1517).
- `DatasetItemDAO`: `avg(tfs.duration)` (~598) and `quantiles(...)(tc.duration)` (~993).

Today this is a no-op (the materialized `duration` is never `NaN` until the cutover DDL guards its formula), so nothing is wrong on this PR. But by `NaNAwareAggregateIntegrationTest`'s own semantics, at cutover `avg`/`avgMap` will *propagate* `NaN` while `quantile` skips it — so the `avgMap(duration)` in `ExperimentDAO` is the one that would start returning `NaN` once the sentinel goes live.

Is the unwrapped set deliberately deferred to the cutover slice (and tracked under OPIK-6892 or a follow-up)? Asking because it's asymmetric with the read sites you *did* wrap — easy for the next slice to miss `ExperimentDAO`/`DatasetItemDAO` if it only greps for what changed here.

JetoPistola:
👋 **Review summary**

**What looks good**
- **Correctness-first flag design.** Making `traceColumnsNonNullable` default-off and gating the entire read/write/filter/sort translation behind it means this ships as a genuine no-op. I traced the flag-OFF path through the unconditional SQL changes (INSERT/INSERT_UPDATE merge, `nullIf(end_time, epoch)` on the thread `argMax`/sort sites, the `notEquals(end_time, epoch)` duration guards) and each collapses to prior behavior when `end_time` is genuinely `NULL`.
- **The `notEquals(..., epoch) AND >= epoch` pattern isn't new** — `start_time`/`created_at` already used it on `main`, and the batch-merge template already used it for `end_time`. This PR just makes the single-INSERT template consistent, so the risk is well-understood.
- **`SentinelTranslation` as a single translation point** is the right call — one place for sentinel↔null instead of every DAO re-deriving it, with the null-safe ordering (constant leads `.equals`) and infinity pass-through thought through.
- **Genuinely excellent test coverage.** `NaNAwareAggregateIntegrationTest` pins the exact ClickHouse quirk the whole design hinges on (min/max/quantile skip NaN; sum/avg/count don't), `TraceSentinelIntegrationTest` proves the R2DBC driver physically serializes `Double.NaN` → `nan` (the one link a stray `NULL` write would silently break), and the flag-ON path is verified against the still-nullable schema before the DDL ships. The comments explaining *why* each assertion is field-by-field rather than `assertTraces` are exactly what a reviewer needs.
- Clear, honest config docs and the "flip in lockstep with the EXCHANGE" callouts.

**Overall**
Solid, low-risk framework-first slice. The approach (sentinels + a single flag flipped in lockstep with the DDL EXCHANGE) is sound and the scope is appropriately narrow. My only substantive question is about the completeness of the read-side `duration` guard for the *next* slice — `ExperimentDAO`/`DatasetItemDAO` have `avg`/`avgMap(duration)` sites that weren't wrapped and will propagate `NaN` at cutover; left a general comment asking whether that's deliberately deferred and tracked.

**Inline comments**: 1 question (general), 1 nit — nothing blocking.

apps/opik-backend/src/main/java/com/comet/opik/domain/filter/FilterQueryBuilder.java
thiagohora, line 93: My only question, is  if we should update the materialized column

thiagohora (dismissed):

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
andrescrz: Conscious choice, not an accident — thanks for flagging. `bindUpdateParams` is shared with the `update()` path, which still needs the conditional `<if(end_time)> :end_time` bind for the UPDATE template, so it stays there. On the `partialInsert` path the unconditional `bindEpochSentinel`/`bindNanSentinel` runs last and wins (sentinel / typed-null), so the earlier bind is a harmless no-op for these two fields. Keeping it as-is to avoid forking `bindUpdateParams`.

apps/opik-backend/src/main/java/com/comet/opik/domain/filter/FilterQueryBuilder.java
andrescrz: Good question — deferring the materialized `duration` column change to the cutover slice, deliberately:

- It's a `MATERIALIZED` column, so changing the expression only affects new inserts; making existing rows consistent needs a full `MATERIALIZE COLUMN`/rewrite — which the cutover already does via the table recreation, at no extra cost.
- The correct post-cutover formula (`if(end_time = epoch, nan, dateDiff(...) / 1000.0)`) is coupled to the epoch sentinel, which only exists once the column is non-nullable; applying it now (while absent = `NULL`) would be a throwaway variant superseded at cutover.
- It's a no-op today: flag-off, `end_time` is `NULL` or a real value, so the current formula already yields the right result.

`NaNAwareAggregateIntegrationTest.guardedDurationFormula` already pins the exact post-cutover formula on the target ClickHouse image, so it's pre-validated for when the cutover DDL recreates the table.

thiagohora (commented):
Automated multi-agent review, findings adversarially verified against the PR head. Inline comments below, ranked by severity in the threads.

One additional finding could not be attached inline because it is in a file this PR does not touch:

**`ExperimentDAO.java:352` — duration quantile CTE not sentinel-guarded (post-cutover, *plausible*).**
The `experiment_durations` CTE feeds `quantiles(0.5,0.9,0.99)(duration)` from `experiment_items_final ei LEFT JOIN (... traces) t`. `quantiles()` skips NaN natively (unfinished traces are safe), but an orphan experiment item (trace_id with no matching trace) makes a join-miss fill the non-nullable `t.duration` with `0.0` post-cutover instead of NULL, dragging p50/p90/p99 down. Harm is conditional on orphan/missing traces.

Checked and refuted: the `bindEpochSentinel` timezone shift (`formatNanos` is UTC-anchored and was already the batch-insert format), the `partialInsert` double-bind (the r2dbc-clickhouse driver overwrites by name), and a client-supplied-NaN `ttft` path (`@PositiveOrZero` rejects NaN with a 422 before the DAO).

<sub>AI-WATERMARK: yes — Tools: Claude Code; Model: Claude Opus 4.8; Scope: automated review, human review recommended before acting.</sub>

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -304,7 +309,7 @@ INSERT INTO traces (
                     new_trace.start_time
                 ) as start_time,
                 multiIf(
-                    isNotNull(old_trace.end_time), old_trace.end_time,
+                    notEquals(old_trace.end_time, toDateTime64('1970-01-01 00:00:00.000', 9)) AND old_trace.end_time >= toDateTime64('1970-01-01 00:00:00.000', 9), old_trace.end_time,
```
thiagohora, line 312: **flag-OFF data loss in the single-insert dedup merge.**

This keep-old-value guard changed **unconditionally** from `isNotNull(old_trace.end_time)` to `notEquals(old_trace.end_time, epoch) AND old_trace.end_time >= epoch` — it is not flag-gated. A client can legitimately store `end_time = 1970-01-01T00:00:00Z` (the `@InRange` validator's `MIN_ANALYTICS_DB` bound is inclusive). With the flag off (the default): create trace X with that end_time, then an at-least-once SDK retry re-POSTs X with `end_time = null` → this guard is false → the merge takes the new NULL and the stored value is **erased**, where pre-PR it was kept.

Also redundant: `notEquals(x,epoch) AND x>=epoch` ≡ `x > epoch`. (The `INSERT_UPDATE` twin already carried this form before this PR — only the `INSERT` path here is new.)

apps/opik-backend/src/main/java/com/comet/opik/domain/filter/FilterQueryBuilder.java
thiagohora, line 93: **unconditional epoch rewrites break the stated flag-OFF contract.**

Unlike the `end_time` *filter* mapping (correctly flag-gated, rationale: "epoch is legitimate while the column is Nullable"), the epoch guard added here to `DURATION_ANALYTICS_DB` — and the sibling rewrites in `ThreadDAO`/`DatasetItemDAO`/`KpiCardDAO`/`ProjectMetricsDAO` (`argMax(output, nullIf(end_time, epoch))`, `ORDER BY nullIf(end_time, epoch)`, the duration projections) — are baked into static SQL with no flag gate. Flag off, a row with a legitimate epoch `end_time` now reports `duration = NULL` instead of the computed `dateDiff`, is dropped from `last_message` selection, and sorts NULL-last. The PR body calls these "provable no-ops while nullable" — true for the `NaN`/`ttft` guards, but **not** for the `end_time`-derived ones.

Also: this shared constant is mapped for `SpanField.DURATION` and `TraceThreadField.DURATION`, so span duration-filter semantics change even though spans are out of this migration's scope.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -354,7 +359,7 @@ INSERT INTO traces (
                     new_trace.output_slim
                 ) as output_slim,
                 multiIf(
-                    isNotNull(old_trace.ttft), old_trace.ttft,
+                    NOT isNaN(old_trace.ttft), old_trace.ttft,
```
thiagohora: **`ttft` merge lacks a row-presence guard (post-cutover bug).**

This uses `multiIf(NOT isNaN(old_trace.ttft), old_trace.ttft, new_trace.ttft)`. There is no `join_use_nulls` override anywhere, so post-cutover a LEFT-JOIN miss (every brand-new trace id) fills the non-nullable `old_trace.ttft` with the Float64 default `0.0`; `NOT isNaN(0.0)` is true → the phantom `0.0` is stored and the incoming ttft is discarded. The `end_time` twin is protected (its join-miss epoch fails the guard) and `project_id` has a `LENGTH(CAST(... AS Nullable(String)))` guard; `ttft` has neither. Same in the `INSERT_UPDATE` template. Latent today (Nullable → NULL → falls through) but defeats the goal of being correct under the cutover.

apps/opik-backend/src/main/java/com/comet/opik/domain/filter/FilterQueryBuilder.java
```diff
@@ -78,9 +85,19 @@ public class FilterQueryBuilder {
...
     private static final String NEW_DURATION_ANALYTICS_DB = "duration";
```
thiagohora: **experiment-comparison duration filter leaks unfinished traces (post-cutover).**

`ExperimentsComparisonValidKnownField.DURATION` maps to this bare, unwrapped `NEW_DURATION_ANALYTICS_DB = "duration"` (see the `EXPERIMENT_ITEM` strategy set at ~`:564`), while `TTFT_ANALYTICS_DB` was wrapped for exactly this reason (its javadoc: "comparisons exclude an absent value — notably `!=`, since `NaN != x` is true"). This filter renders against raw `traces` (e.g. `DatasetItemDAO`: `... AND ei.trace_id IN (SELECT id FROM traces WHERE ... <experiment_item_filters>)`). Post-cutover an unfinished trace materializes `duration = NaN`, so `duration != 5` returns it (Nullable-era NULL excluded it) — the same `!=` leak fixed for ttft, left open for experiment-comparison duration.

apps/opik-backend/src/main/java/com/comet/opik/utils/SentinelTranslation.java
```diff
@@ -0,0 +1,84 @@
...
+ * </ul>
+ *
+ * <p>Keeping the {@code sentinel ↔ null} mapping in one place mirrors the codebase's existing idiom for {@code ''}
+ * -defaulted strings and guarantees every DAO that reads or writes one of these columns
+ * inherits the same contract instead of re-deriving it.</p>
+ */
+@UtilityClass
+public class SentinelTranslation {
+
+    /** Sentinel for a non-nullable {@code DateTime64} column ({@code DEFAULT toDateTime64('1970-01-01 00:00:00', 6)}). */
```
thiagohora: **precision mismatch hidden in duplicated SQL.**

This javadoc documents the eventual column default as `toDateTime64('1970-01-01 00:00:00', 6)` — **precision 6** — while the SQL sentinel fragment `nullIf(end_time, toDateTime64('1970-01-01 00:00:00.000', 9))` is hand-declared byte-identical at **precision 9** in 4+ places (`FilterQueryBuilder.END_TIME_NON_NULLABLE_ANALYTICS_DB:72`, `TraceDAO.SORT_FIELD_MAPPING_END_TIME_SENTINEL:3038`, `ThreadDAO.SORT_FIELD_MAPPING_END_TIME_SENTINEL:1357`, inline `ThreadDAO` ORDER BY), plus a copy-pasted `readEpochSentinel` in two DAOs and an identical `traceColumnsNonNullable()` accessor in four. These all must stay in lockstep for filter/sort/read to agree. Extract one shared SQL constant so the cutover changes it in one place.

apps/opik-backend/src/main/java/com/comet/opik/utils/SentinelTranslation.java
```diff
@@ -0,0 +1,84 @@
...
+    /** Sentinel for a non-nullable {@code FixedString(36)} column ({@code DEFAULT ''}). */
+    public static final String EMPTY_UUID_SENTINEL = "";
+
+    // Outbound: ClickHouse sentinel → API null.
+
+    /**
+     * @return {@code null} when the value is absent — {@code null} or the epoch sentinel — otherwise the value itself.
+     * The constant leads the equality check so a {@code null} value can never throw.
+     */
+    public static Instant epochToNull(Instant value) {
```
thiagohora, line 44: **duplicates pre-existing helpers, contradicting the "single translation point" javadoc.**

`epochToNull` re-implements `DatasetItemResultMapper.nullIfEpoch` (still called at `DatasetItemVersionDAO:3934`), and `nanToNull` duplicates `filterNan` in both `KpiCardDAO` and `WorkspaceMetricsDAO`. No old call site was migrated. Delegate the existing helpers to `SentinelTranslation` so the mapping genuinely has one owner.

apps/opik-backend/src/main/java/com/comet/opik/domain/filter/FilterQueryBuilder.java
```diff
@@ -863,13 +880,25 @@ public static Optional<Boolean> hasField(@NonNull List<? extends Filter> filters
 
     public static Optional<String> toAnalyticsDbFilters(
             @NonNull List<? extends Filter> filters, @NonNull FilterStrategy filterStrategy) {
```
thiagohora, line 888: **`toAnalyticsDbFiltersV2Client` structurally pins the flag off (latent).**

This 2-arg overload defaults the flag to `false`, and `toAnalyticsDbFiltersV2Client` (`:915`) delegates to it with no flag parameter, so it can never opt in. No current caller routes a TRACE/TRACE_THREAD `end_time` filter through this path, so it's a latent footgun rather than a live bug — but post-cutover any new trace-filtering caller here would silently emit raw `end_time` comparisons matching epoch-sentinel rows. Injecting the config into `FilterQueryBuilder` would remove the parameter, the default-false overload, and the four duplicated accessors at once.

apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
```diff
@@ -478,7 +482,7 @@ thread_model_id NOT IN (SELECT entity_id FROM fsc)
             <if(stream)>
             ORDER BY workspace_id, project_id, thread_model_id DESC
             <else>
-            <if(sort_fields)> ORDER BY <sort_fields>, last_updated_at DESC <else> ORDER BY last_updated_at DESC, start_time ASC, end_time DESC <endif>
+            <if(sort_fields)> ORDER BY <sort_fields>, last_updated_at DESC <else> ORDER BY last_updated_at DESC, start_time ASC, nullIf(end_time, toDateTime64('1970-01-01 00:00:00.000', 9)) DESC <endif>
```
thiagohora, line 485: **changed thread `ORDER BY` has no sort-ordering test.**

This default `ORDER BY` change and the flag-gated thread `end_time` sort mapping have no sort test (the trace side does), which the repo's `testing.md` coverage bar for sorting-SQL changes calls for.

andrescrz:
> @JetoPistola asked (cutover completeness — read-side `duration` NaN guard): whether the unwrapped `avg`/`avgMap(duration)` sites in `ExperimentDAO`/`DatasetItemDAO` are deliberately deferred and tracked.

Thanks — this was the right thing to pull on. Rather than just the flagged sites, I ran an exhaustive sweep of **every** read of the materialized `traces.duration` (and, for completeness, `end_time`/`ttft`), tracing each to its source table and consumer.

On your five: all safe. `ExperimentDAO` `quantiles(duration)` is `isFinite`-clamped and its `avgMap` is over the quantile-result map (never the raw column); `DatasetItemDAO:598/993` read CTEs recomputed with the `notEquals(end_time, epoch)` guard (→ `NULL` for absent, so `avg` skips); and the `avg(...duration)` sites over `experiment_item_aggregates` are safe because that stored table's population is guarded at `ExperimentAggregatesDAO` (`if(isNaN(duration), NULL, duration)` at the trace read). So `avg`/`avgMap` never sees `NaN` at cutover.

The sweep did surface one site **not** previously flagged: **`OptimizationDAO`** reads `traces.duration` unguarded (`quantiles(0.5)(t.duration)` → `sum(duration_p50 * trace_count)`); an all-absent-`end_time` experiment would make `quantiles`→`NaN` propagate through `sum` (and `isNotNull(NaN)` is true), returning `NaN` for `weighted_duration` where it was `NULL`-safe before. Now guarded with the same `if(isNaN(duration), NULL, duration)` pattern — landing in a follow-up commit on this branch. `ttft` and `end_time` came back clean (fully handled in `TraceDAO`/`ThreadDAO`; spans are the later `spanColumnsNonNullable` slice).

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
andrescrz: Skipping — this is a no-op flag-off for all real data. Flag-off, `old_trace.end_time` is `NULL` on a join miss and a real timestamp otherwise; `notEquals(NULL, epoch)` is `NULL` (→ falls through to new, same as the old `isNotNull(...)`) and `notEquals(real, epoch) AND real >= epoch` is true (→ keep old, same as before). The only value that diverges is a literal `1970-01-01` end_time, which no real trace has. Post-cutover the same guard also correctly resolves the join miss (miss fills the epoch → falls through to new). No reachable data loss.

apps/opik-backend/src/main/java/com/comet/opik/domain/filter/FilterQueryBuilder.java
andrescrz: Skipping — the unconditional epoch/`nullIf` guards are no-ops on all real flag-off data, because the trace `end_time`/`ttft` writes are flag-gated (absent → `NULL`, never the epoch, pre-cutover). So `notEquals(NULL, epoch) = NULL` and `nullIf(NULL, epoch) = NULL` leave every real pre-cutover row unchanged; they're unconditional by design so they're correct post-cutover without a second flag flip. The `end_time` filter mapping is separately flag-gated precisely because a client can send a literal epoch value there that must keep matching while the column is Nullable.

apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
```diff
@@ -1341,6 +1345,16 @@ thread_model_id NOT IN (SELECT entity_id FROM fsc)
...
+    private final @NonNull OpikConfiguration configuration;
+
+    /**
+     * Sort mapping applied under {@code traceColumnsNonNullable}: a thread's {@code end_time} is
+     * {@code max(traces.end_time)}, the epoch sentinel when every trace is unfinished; {@code nullIf} restores
+     * {@code NULL} so it sorts last in ASC like a Nullable column did. {@code duration} needs no entry — ClickHouse
+     * sorts {@code NaN} like {@code NULL}.
+     */
+    private static final Map<String, String> SORT_FIELD_MAPPING_END_TIME_SENTINEL = Map.of(
+            SortableFields.END_TIME, "nullIf(end_time, toDateTime64('1970-01-01 00:00:00.000', 9))");
```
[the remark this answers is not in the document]
andrescrz: Skipping — the trace and thread sort mappings sit in DAOs with distinct query shapes/aliases; hoisting `nullIf(end_time, epoch)` into one shared constant couples them for a one-line dedup, trading SRP for surface reuse. Both paths have their own sort tests, so a drift would be caught rather than silently ship.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/TraceSentinelIntegrationTest.java
```diff
@@ -0,0 +1,418 @@
...
+                    .build();
+
+            var page = traceResourceClient.getTraces(
+                    projectName, null, API_KEY, WORKSPACE_NAME, List.of(), List.of(sortingField), 10, Map.of());
+
+            // The sort maps end_time through nullIf(end_time, epoch), so the absent trace sorts as NULL (last),
+            // preserving Nullable-era ordering — not as the 1970 epoch, which would sort first ascending. Ids, not a
+            // full assertTraces: the absent trace's not-yet-epoch-guarded duration is a negative pre-cutover artifact.
+            var actualOrder = page.content().stream().map(Trace::id).toList();
+            assertThat(actualOrder).containsExactly(presentTrace.id(), absentTrace.id());
```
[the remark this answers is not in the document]
andrescrz: Skipping — this is deliberate. `assertTraces` also asserts the derived `duration`, but the absent trace's materialized `duration` is a pre-cutover negative artifact (the `end_time`-epoch guard on the duration column ships with the cutover DDL), so a full-object assertion would fail on it. We assert exact positional order with `containsExactly` on ids — the sort concern here — and the trace payload/round-trip is already covered by the other `CreateTraces` tests.

andrescrz:
> @thiagohora (review body): `ExperimentDAO.java:352` — duration quantile CTE not sentinel-guarded (post-cutover, *plausible*).

Thanks — checked this in the exhaustive sweep. It's safe: `quantiles(0.5,0.9,0.99)(duration)` skips `NaN` natively, and the results are `isFinite`-clamped (`if(isFinite(v), v, 0)`), so the all-absent edge (`quantiles`→`NaN`) clamps to `0` rather than leaking `NaN`. The `avgMap` at :1517 is over that quantile map, not the raw column. The one genuinely unguarded materialized-`duration` reader the sweep found was `OptimizationDAO`, now fixed.

thiagohora (dismissed):

apps/opik-backend/src/main/java/com/comet/opik/utils/SentinelTranslation.java
andrescrz: Fixed in 1aca6af1 — aligned the javadoc to **precision 9**: the sentinel now documents `DEFAULT toDateTime64('1970-01-01 00:00:00', 9)`, matching the `DateTime64(9)` `end_time` column and every SQL fragment (`nullIf(end_time, toDateTime64('1970-01-01 00:00:00.000', 9))`). The SQL was already precision 9; only the doc said 6, so this removes the mismatch.

apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
```diff
@@ -416,18 +420,18 @@ ORDER BY (workspace_id, project_id, entity_id, name) DESC, last_updated_at DESC
                     t.project_id as project_id,
                     min(t.start_time) as start_time,
                     max(t.end_time) as end_time,
-                    if(end_time IS NOT NULL AND start_time IS NOT NULL
+                    if(end_time IS NOT NULL AND notEquals(end_time, toDateTime64('1970-01-01 00:00:00.000', 9)) AND start_time IS NOT NULL
                            AND notEquals(start_time, toDateTime64('1970-01-01 00:00:00.000', 9)),
                        (dateDiff('microsecond', start_time, end_time) / 1000.0),
                        NULL) AS duration,
                     argMin(t.input, t.start_time) as first_message,
```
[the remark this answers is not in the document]
andrescrz: Skipping — same as the other unconditional epoch guards on this PR (already accepted): flag-off, a thread's underlying `end_time` is `NULL` for an absent value (the write binds `NULL`, never the epoch, pre-cutover) or a real timestamp, so `notEquals(end_time, epoch)` leaves all real data unchanged (`notEquals(NULL, epoch) = NULL`; `notEquals(real, epoch) = true`). The only value it treats as absent is a literal `1970-01-01` end_time, which no real trace has. It's unconditional by design so it stays correct post-cutover without a second flag flip. The flag-on thread path is covered by `TraceSentinelIntegrationTest.Threads` (absent thread reads `null` and sorts last).

apps/opik-backend/src/main/java/com/comet/opik/domain/KpiCardDAO.java
```diff
@@ -581,7 +584,7 @@ private Statement buildStatement(Connection connection, ST template,
 
     private void addTraceFilters(ST template, List<? extends Filter> filters) {
         Optional.ofNullable(filters).ifPresent(f -> {
-            FilterQueryBuilder.toAnalyticsDbFilters(f, FilterStrategy.TRACE)
+            FilterQueryBuilder.toAnalyticsDbFilters(f, FilterStrategy.TRACE, traceColumnsNonNullable())
                     .ifPresent(traceFilters -> template.add("trace_filters", traceFilters));
             FilterQueryBuilder.toAnalyticsDbFilters(f, FilterStrategy.FEEDBACK_SCORES)
                     .ifPresent(scoresFilters -> template.add("trace_feedback_scores_filters", scoresFilters));
```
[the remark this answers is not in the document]
andrescrz: Skipping — the flag-gate wiring (`traceColumnsNonNullable()` plus passing it to `toAnalyticsDbFilters(...)`) is small per-DAO boilerplate in DAOs with distinct query builders and responsibilities; hoisting it into a shared helper would couple them for minimal reuse, trading SRP for surface dedup. Consistent with the other dedup calls on this PR.

thiagohora (approved):

PR #7312, opened 2026-07-01
[OPIK-6995] [FS] feat: add per-evaluation spend budget for LLM-as-judge evaluators

alexkuzmik:
## Details

Adds an optional **per-rule USD spend budget** for **agentic** (trace + thread) LLM-as-judge online evaluators. Long/large inputs push the agentic `ToolCallLoop` (up to 10 rounds, re-sending context each turn) into unbounded token spend with no ceiling — the only signal today is the provider bill after the fact (OPIK-6995).

**The budget is a soft wrap-up trigger, not a hard ceiling.** Cumulative spend is tracked as the evaluation runs; the moment it reaches the limit, the agent stops starting new turns and does one final tools-stripped call to emit a best-effort verdict from the data already gathered. Spend may overshoot the limit slightly (by that wrap-up call) — by design. No pre-estimation, no hard kill.

Key pieces:
- **`BudgetGuard`** encapsulates the limit, running spend, and all cost math, so neither the scorers nor `ToolCallLoop` touch pricing — they only `track()` each LLM call and ask `shouldWrapUp()` between turns. Cost is computed fully in-process via `CostService` (no provider round-trip) from each response's token usage. Fail-open (a pricing/resolution failure degrades to no-limit); when a limited guard resolves an unpriced model (cost stays zero, so the limit can never trip) it warns once so the ineffective limit is operator-visible. Non-positive limits are now rejected at the API boundary (`@Positive` on `max_cost_usd` is `@Valid`-cascaded into the evaluator `code` record → **422**); the guard keeps the non-positive→no-limit coercion as defense-in-depth for any legacy persisted row.
- **`LlmUsageExtractor`** (extracted from `EvaluationEntityFactory`, now the single source of truth) builds the cache-aware usage map for any provider Opik prices — base prompt/completion/total plus provider-specific cache tokens (Anthropic / OpenAI / OpenAI-official / Gemini).
- **Monitoring signal:** the `budget_exceeded` trace tag, the budget-specific wrap-up instruction, and the user-facing warn all key off **one authoritative signal** — `BudgetGuard.wasBudgetEnforced()`, set at the exact point the tool loop's budget gate abandons pending tool calls. So a natural stop (or an inline single call) that merely crosses the spend limit is **not** mislabeled as a budget cut-off: it keeps the normal "investigation complete" wrap-up, emits no warn, and is not tagged. When the budget genuinely cuts a run short, the hidden `online_evaluation` monitoring trace is tagged **`budget_exceeded`** (on both the success and error finalize paths) so users can filter/count those evaluations, and the wrap-up call uses the **distinct instruction** — "budget reached, give a best-effort verdict from partial data."
- **Span evaluators are excluded on purpose:** the span evaluator DTO exposes no `max_cost_usd` field (unlike trace/thread), so there is no per-evaluation budget to enforce — the span path passes `BudgetGuard.UNLIMITED` even though it can run the same agentic loop.
- **Frontend:** optional "Max cost per evaluation (USD)" numeric field on trace/thread rules (hidden for span scope), wired through the Zod schema and API DTO conversions. Empty = no limit.

Stored in the evaluator `code` JSON column, so there is **no DB migration**; existing rows deserialize `null` = no limit (behavior unchanged).

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues
- OPIK-6995

## Testing
- **Unit:** `BudgetGuardTest` (accumulation with cost derived from `CostService`, limit trip, non-positive→unlimited, fail-open incl. a pricing-exception case, unknown-price never trips), `LlmUsageExtractorTest` (base with a distinct pass-through total + Anthropic/OpenAI/OpenAI-official/Gemini cache tokens + null, cache-omission branches split per case), `ToolCallLoopTest` (early wrap-up once budget reached; budget-specific vs. normal wrap-up instruction; natural-stop-over-budget keeps the completion instruction and is not flagged), `OnlineEvaluationRecorderTest` (`budget_exceeded` tag on success and error paths; `monitor()` pass-through preserved), `AutomationRuleEvaluatorMaxCostMappingTest` (trace + thread + null API↔domain round-trip).
- **API validation:** `AutomationRuleEvaluatorsResourceTest` asserts a non-positive `max_cost_usd` is rejected with **422** (the `@Valid` cascade); PODAM now manufactures strictly-positive values for `@Positive` `BigDecimal` fields so the newly-enforced constraint doesn't flake the create round-trips (all six subtypes still round-trip).
- **Manual (live, local stack):** agentic trace evaluation with a realistic budget scored normally under the cap; agentic thread evaluation with a tiny budget tripped the guard → single investigative turn + tools-stripped wrap-up (2 LLM calls, `has_tool_spans=false`), best-effort score still stored, monitoring trace tagged `budget_exceeded`, and the wrap-up span carried the budget-specific instruction. Spend/limit reflected in the rule logs and the monitoring trace's `total_estimated_cost`.

## Documentation
N/A — no documentation changes in this PR.


apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringLlmAsJudgeScorer.java
```diff
@@ -226,6 +226,10 @@ protected Mono<Void> score(@NonNull TraceToScoreLlmAsJudge message) {
     private Mono<List<FeedbackScoreBatchItem>> evaluate(TraceToScoreLlmAsJudge message, List<Span> spans,
             List<AttachmentInfo> attachments, Map<String, String> mdc, EvaluationRecorder recorder) {
         var trace = message.trace();
+        // One guard per trace evaluation; UNLIMITED when the rule sets no maxCostUsd. Charges every
+        // LLM call (initial + tool rounds + wrap-up) and tells the tool loop when to start wrapping up.
+        var costGuard = BudgetGuard.create(message.llmAsJudgeCode().maxCostUsd(),
+                message.llmAsJudgeCode().model().name(), llmProviderFactory);
         // Sync prep is CPU-bound (JSON serialization for the size estimate + prompt rendering)
         // — schedule on Schedulers.parallel() so we don't tax the R2DBC scheduler that emits
         // the spans fetch upstream, and don't pin a workersScheduler thread on the inline path.
```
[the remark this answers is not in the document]
alexkuzmik: Skipping — the trace and thread scorers are intentionally parallel across the codebase, and the budget logic they'd otherwise duplicate now lives in one place: `BudgetGuard` (creation + accounting) and `ToolCallLoop` (where the budget gate fires and now flags `budget_exceeded`, per the fix for the sibling comment). Each scorer only keeps a per-entity user-facing warn, so there's little logic left to drift. A cross-scorer helper would have to abstract over the two different message types and prepare paths, fighting that established symmetry for marginal gain.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/ToolCallLoop.java
```diff
@@ -168,9 +189,16 @@ static Mono<ChatResponse> runWithWrapUp(
...
+        // let runWithWrapUp do its final tools-stripped call (the intended, bounded overshoot). Same
+        // no-append reasoning as the MAX_TOOL_CALL_ROUNDS cap below.
+        if (round >= MAX_TOOL_CALL_ROUNDS || costGuard.shouldWrapUp()) {
+            if (costGuard.shouldWrapUp() && round < MAX_TOOL_CALL_ROUNDS) {
+                log.info("Evaluation spend budget reached for '{}' (spent '{}' of '{}' USD);"
+                        + " wrapping up", logIdValue, costGuard.spentUsd(), costGuard.limitUsd());
+            }
             // Don't append: the cap-round response may carry unfulfilled tool_executions_requests
             // (we're abandoning them by hitting the cap). An AiMessage with tool_calls but no
             // matching ToolExecutionResultMessage produces a malformed sequence that OpenAI /
```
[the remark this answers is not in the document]
alexkuzmik: Fixed in 30e8b37883 — the ordering was already addressed in 2700ea2; this demotes the loop's `log.info("Evaluation spend budget reached …")` to `debug` so it no longer duplicates the scorer's user-facing `warn` at INFO.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringLlmAsJudgeScorer.java
```diff
@@ -250,11 +254,18 @@ private Mono<List<FeedbackScoreBatchItem>> evaluate(TraceToScoreLlmAsJudge messa
...
                                     : Mono.just(initialResponse));
                 })
                 .map(chatResponse -> {
                     try (var logContext = wrapWithMdc(mdc)) {
+                        if (costGuard.shouldWrapUp()) {
+                            userFacingLogger.warn(
+                                    "Spend budget of '{}' USD reached for traceId '{}' (spent '{}'); "
+                                            + "stopped investigating and wrapped up with the scores gathered so far.",
+                                    costGuard.limitUsd(), trace.id(), costGuard.spentUsd());
+                            recorder.flagBudgetExceeded();
```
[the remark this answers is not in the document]
alexkuzmik: Fixed in 30e8b37883 — `flagBudgetExceeded()` is now called at the authoritative point in `ToolCallLoop` where the budget gate fires (`shouldWrapUp() && round < MAX`), and the inferred calls were removed from both scorers' `map()`. A natural stop that merely crossed spend hits the no-tool branch above the gate, so it's no longer mislabelled; and because the flag is set during the loop (before the wrap-up/scoring chain can throw), `monitor()`'s error path finalizes with the correct `budgetExceeded`. Idempotent (the gate trips once). Covered by two new ToolCallLoop tests.

apps/opik-frontend/src/v2/pages-shared/automations/AddEditRuleDialog/LLMJudgeRuleDetails.tsx
```diff
@@ -235,6 +237,65 @@ const LLMJudgeRuleDetails: React.FC<LLMJudgeRuleDetailsProps> = ({
...
+          render={({ field, formState }) => {
+            const validationErrors = get(formState.errors, [
+              "llmJudgeDetails",
+              "maxCostUsd",
+            ]);
+
+            return (
+              <FormItem>
+                <Label>Max cost per evaluation (USD)</Label>
+                <FormControl>
```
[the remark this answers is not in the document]
alexkuzmik: Fixed in 30e8b37883 — switched to `FormLabel` + `FormDescription` so the label binds `htmlFor={formItemId}` and inherits the form error styling, per .agents/skills/opik-frontend/forms.md.

apps/opik-frontend/src/v2/pages-shared/automations/AddEditRuleDialog/LLMJudgeRuleDetails.tsx
```diff
@@ -235,6 +237,65 @@ const LLMJudgeRuleDetails: React.FC<LLMJudgeRuleDetailsProps> = ({
...
+                        "border-destructive": Boolean(
+                          validationErrors?.message,
+                        ),
+                      },
+                    )}
+                    onKeyDown={(event) => {
+                      // Positive decimal only: block the symbols type=number would otherwise accept
+                      // (sign, exponent, and the comma thousands/decimal separator).
+                      if (["e", "E", "+", "-", ","].includes(event.key)) {
+                        event.preventDefault();
```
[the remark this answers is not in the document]
alexkuzmik: Fixed in 30e8b37883 — `onChange` now rejects any raw value that isn't a plain decimal (`/^\d*\.?\d*$/`), so pasted/IME input like `1e3` is dropped before it reaches `valueAsNumber` (the controlled field keeps its last valid state). `onKeyDown` still blocks the same symbols during typing for immediate feedback.

awkoy (dismissed):

apps/opik-backend/src/main/java/com/comet/opik/api/evaluators/AutomationRuleEvaluatorLlmAsJudge.java
```diff
@@ -37,7 +39,15 @@ public record LlmAsJudgeCode(
                     View.Public.class, View.Write.class}) @NotNull LlmAsJudgeModelParameters model,
             @JsonView({View.Public.class, View.Write.class}) @NotNull List<LlmAsJudgeMessage> messages,
             @JsonView({View.Public.class, View.Write.class}) @NotNull Map<String, String> variables,
-            @JsonView({View.Public.class, View.Write.class}) @NotNull List<LlmAsJudgeOutputSchema> schema) {
+            @JsonView({View.Public.class, View.Write.class}) @NotNull List<LlmAsJudgeOutputSchema> schema,
+            @JsonView({View.Public.class, View.Write.class}) @Positive BigDecimal maxCostUsd) {
```
BorisTkachenko, line 43: **[medium] @Positive on maxCostUsd is not enforced at the API boundary**

The new @Positive constraint on maxCostUsd lives on a component of the nested LlmAsJudgeCode record, which is reached through the base-class field AutomationRuleEvaluator#code (line 90: `@NotNull private final T code;`). That field is not annotated @Valid, so Jakarta/Jersey does NOT cascade validation into the code record. Consequently @Positive (and the pre-existing @NotNull on model/messages/variables/schema) never fires. A client can POST `max_cost_usd: 0` or a negative value; it passes validation and is persisted, then BudgetGuard.create() (OnlineScoringLlmAsJudgeScorer line 319, BudgetGuard line 63) treats signum<=0 as UNLIMITED. Net effect: a user who sets a 0/near-zero budget expecting a hard spend cap silently gets no cap at all, defeating the purpose of this field.

💡 Add @Valid to the `code` field in AutomationRuleEvaluator (`@NotNull @Valid private final T code;`) so nested constraints cascade, or otherwise verify the cascade path. Add a resource-level test asserting a non-positive max_cost_usd is rejected with 4xx.

_rule: `backend-api-contract-apply-jakarta-bean-validation-constraints-comprehensively-on`_

apps/opik-backend/src/main/java/com/comet/opik/api/evaluators/AutomationRuleEvaluatorTraceThreadLlmAsJudge.java
```diff
@@ -37,10 +39,18 @@ public record TraceThreadLlmAsJudgeCode(
             @JsonView({
                     View.Public.class, View.Write.class}) @NotNull LlmAsJudgeModelParameters model,
             @JsonView({View.Public.class, View.Write.class}) @NotNull List<LlmAsJudgeMessage> messages,
-            @JsonView({View.Public.class, View.Write.class}) @NotNull List<LlmAsJudgeOutputSchema> schema) {
+            @JsonView({View.Public.class, View.Write.class}) @NotNull List<LlmAsJudgeOutputSchema> schema,
+            @JsonView({View.Public.class, View.Write.class}) @Positive BigDecimal maxCostUsd) {
```
BorisTkachenko, line 43: **[medium] @Positive on maxCostUsd never fires — nested constraint is not @Valid-cascaded**

The new @Positive on maxCostUsd is inert. It lives on a component of the TraceThreadLlmAsJudgeCode record, but the containing field is declared as `@NotNull private final T code;` in AutomationRuleEvaluator (line 90) with no @Valid. Bean Validation does not cascade into the record, so none of its constraints (@Positive here, and the pre-existing @NotNull on model/messages/schema) run at the API boundary. The team already documents this in BudgetGuardTest (`@Positive on the DTO doesn't run (code field isn't @Valid-cascaded)`), and BudgetGuard.create() defensively treats signum() <= 0 as UNLIMITED. Net effect: a client POSTing maxCostUsd = 0 or -5 does NOT get a 400 as the annotation implies; the value is silently accepted and reinterpreted as 'no limit'. The constraint intended to reject bad input at the boundary is a no-op, and the API contract it advertises is misleading.

💡 Either make the constraint meaningful by adding @Valid to the `code` field in AutomationRuleEvaluator so the record's constraints cascade (note this will also start enforcing the existing @NotNull components — verify that is desired), or drop the misleading @Positive and document that non-positive limits are intentionally coerced to UNLIMITED by BudgetGuard. Keeping a constraint that never executes should be avoided.

_rule: `backend-null-safety-guards-add-decimalmin-min-bounds-and-notnull-to-user-supplied-and-c`_

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/BudgetGuard.java
```diff
@@ -0,0 +1,108 @@
...
+        return call.doOnNext(this::charge);
+    }
+
+    private void charge(ChatResponse response) {
+        try {
+            var usage = LlmUsageExtractor.toUsageMap(response);
+            if (usage == null) {
+                return;
+            }
+            spentUsd = spentUsd.add(CostService.calculateCost(model, provider, usage, null));
```
BorisTkachenko: **[low] Budget silently unenforced when model is missing from the price table**

charge() treats a zero cost identically to a real charge. When the model resolves successfully (getResolvedModelInfo always returns a ResolvedModelInfo) but the (actualModel, provider) pair is absent from CostService's static price table, findModelPrice returns DEFAULT_COST and calculateCost returns BigDecimal.ZERO (getCostFromMetadata(null) also yields no cost). No exception is thrown, so the try/catch never fires and nothing is logged: spentUsd stays 0, shouldWrapUp() never turns true, and the user-configured maxCostUsd is silently never enforced. This is a realistic path since the static price table does not cover every model/provider combination. The class javadoc's fail-open contract only mentions 'resolution or pricing failure', whereas this case is a successful-but-unpriced lookup with no operator signal — unlike the exception branch and the create() resolution failure, both of which log a warn.

💡 Detect the unpriced case (e.g. when a limited guard's first non-null usage yields a zero cost) and emit a single warn once per guard so operators know the spend limit is ineffective for this model, rather than treating an unknown-price model the same as a genuinely zero-cost call.

_general finding_

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringLlmAsJudgeScorer.java
```diff
@@ -338,11 +342,19 @@ private Mono<List<FeedbackScoreBatchItem>> evaluate(TraceToScoreLlmAsJudge messa
                             .flatMap(initialResponse -> prepared.useTools()
                                     ? handleToolCalls(initialResponse, prepared.scoreRequest(),
                                             prepared.structuredRequest(), message, spans, prepared.fullJson(), mdc,
-                                            recorder)
+                                            recorder, costGuard)
                                     : Mono.just(initialResponse));
                 })
                 .map(chatResponse -> {
                     try (var logContext = wrapWithMdc(mdc)) {
+                        if (costGuard.shouldWrapUp()) {
```
BorisTkachenko: **[low] User-facing budget warn fires on paths where the budget did not curtail anything, diverging from the budget_exceeded tag**

The warn is gated purely on costGuard.shouldWrapUp() (cumulative spend >= limit), which is a weaker condition than 'the budget actually cut the run short'. It therefore fires in two cases where no investigation was stopped: (1) the inline/non-tools path, where there is a single LLM call and no loop to abort — yet the user is told 'stopped investigating and wrapped up with the scores gathered so far'; and (2) the agentic path where the model finished naturally (no pending tool calls) but the final charged call pushed spend over the limit — ToolCallLoop's natural-stop branch returns before the budget gate and does NOT call recorder.flagBudgetExceeded(), so the monitoring trace is never tagged budget_exceeded, yet this warn still fires. The user-facing signal and the budget_exceeded provenance thus disagree for the same run. The code comment acknowledges the tag is authoritative in ToolCallLoop, but the warn here is broader than that authoritative signal.

💡 Surface the user-facing warn from the same authoritative condition ToolCallLoop uses to flag budget_exceeded (i.e. the guard actually aborted pending tool calls), rather than re-deriving it from shouldWrapUp() in the scorer. For example, expose a 'wasCutShort'/'budgetTriggered' flag set only when the loop's budget gate fired, and warn on that; on the inline path suppress the warn entirely since the budget cannot stop a single call.

_general finding_

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSpanLlmAsJudgeScorer.java
```diff
@@ -370,7 +370,8 @@ Mono<ChatResponse> handleToolCalls(ChatResponse chatResponse, ChatRequest toolRe
         var span = message.span();
         // Shared loop orchestration lives in the base scorer; here we provide only the span-specific
         // context seeding — pre-seed the active span so read(type=span) / jq(type=span) resolve it
-        // without a re-fetch.
+        // without a re-fetch. Span evaluation is a single LLM call with no loop to wrap up, so a spend
```
BorisTkachenko: **[low] Comment justifying BudgetGuard.UNLIMITED misstates why the guard is unlimited**

The new comment claims 'Span evaluation is a single LLM call with no loop to wrap up, so a spend budget could never enforce anything here.' This is inside handleToolCalls, which is the agentic tool-loop path: it delegates to agenticScoringService.runToolCallLoop, whose contract (AgenticScoringService.java:62-65) explicitly runs ToolCallLoop#runWithWrapUp with AUTO follow-ups when the initial response carries tool calls. So the span path can and does make multiple LLM calls with a wrap-up turn — exactly the scenario BudgetGuard exists to bound for the trace/thread scorers. The real reason UNLIMITED is correct is that the span evaluator DTO (AutomationRuleEvaluatorSpanLlmAsJudge / SpanToScoreLlmAsJudge) exposes no maxCostUsd field, unlike the trace and thread evaluators, so there is no budget to enforce. The current wording could mislead a future maintainer (e.g. someone who later adds a span-level maxCostUsd) into believing the span path never loops and thus never needs budget wiring.

💡 Reword the comment to state the actual rationale: the span evaluator has no configurable maxCostUsd (unlike trace/thread evaluators), so there is no per-evaluation spend budget to enforce; pass BudgetGuard.UNLIMITED. Drop the 'single LLM call / no loop' claim, since this method runs the multi-turn agentic tool loop.

_general finding_

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/ToolCallLoop.java
```diff
@@ -151,13 +165,20 @@ static Mono<ChatResponse> runWithWrapUp(
...
             @NonNull EvaluationRecorder recorder) {
         return run(initialResponse, toolRequest, followUpParameters, toolRegistry, scoreTrace,
-                messages, ctx, budget, logIdValue, mdc, recorder)
+                messages, ctx, budget, costGuard, logIdValue, mdc, recorder)
                 .flatMap(loopFinalResponse -> {
-                    messages.add(UserMessage.from(WRAP_UP_USER_MESSAGE));
+                    // Budget-triggered wrap-up gets a distinct instruction: the run was cut short, so
+                    // ask for a best-effort verdict from partial data rather than telling the model it
+                    // "completed" its investigation.
+                    var wrapUpMessage = costGuard.shouldWrapUp()
```
BorisTkachenko: **[low] Natural stop that crosses budget is mislabeled as a budget cut-off in the wrap-up message and user warn, but not in the trace tag**

runWithWrapUp selects the wrap-up instruction purely from costGuard.shouldWrapUp() (line 178), and the scorers' user-facing warn (OnlineScoringLlmAsJudgeScorer:350 / OnlineScoringTraceThreadLlmAsJudgeScorer:334) key off the same predicate. But recorder.flagBudgetExceeded() is deliberately gated on the narrower condition (line 212: shouldWrapUp() && round < MAX && tool requests pending). When a model finishes on its own (no tool requests) on the exact turn whose LLM call tips spend over the limit — a path explicitly handled and tested at ToolCallLoopTest.appendsTerminalAiMessageWhenBudgetTripsOnANaturalStop — shouldWrapUp() is true, so the run gets BUDGET_WRAP_UP_USER_MESSAGE ("analysis may be incomplete... stop investigating") and the scorer logs "stopped investigating and wrapped up with the scores gathered so far", yet the trace is NOT tagged budget_exceeded. The three signals diverge for a run that actually completed naturally: the model is told its complete investigation was cut short, and monitoring shows a budget warn with no matching tag. The in-code comment at lines 213-219 reasons only about the trace tag, not the message/warn divergence.

💡 Drive all three signals off one authoritative source. Either flag the recorder whenever shouldWrapUp() triggered the wrap-up (including the natural-stop-over-budget case), or select BUDGET_WRAP_UP_USER_MESSAGE / emit the user warn only when the budget gate (not a natural stop) actually cut the run short — e.g. have the loop expose whether it stopped due to the budget gate rather than re-deriving intent from shouldWrapUp() post-hoc.

_general finding_

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/BudgetGuardTest.java
```diff
@@ -0,0 +1,111 @@
...
+        var guard = BudgetGuard.create(new BigDecimal("0.50"), "gpt-4o", factoryFor("gpt-4o", "openai"));
+
+        assertThat(guard.shouldWrapUp()).isFalse();
+
+        guard.track(Mono.just(responseWith(100_000, 0))).block(); // +$0.25 -> $0.25
+        assertThat(guard.shouldWrapUp()).isFalse();
+
+        guard.track(Mono.just(responseWith(100_000, 0))).block(); // +$0.25 -> $0.50, reaches limit
+        assertThat(guard.shouldWrapUp()).isTrue();
+        assertThat(guard.spentUsd()).isEqualByComparingTo("0.50");
```
BorisTkachenko: **[low] Exact-cost assertions couple the test to the mutable bundled price table**

accumulatesSpendAcrossCallsAndTripsOnceLimitReached (and the $0.25/unit assumption documented in the comment at lines 18-19) depend on gpt-4o's price in src/main/resources/model_prices_and_context_window.json (input_cost_per_token = 2.5e-6). That file is a live, periodically-updated artifact (e.g. commit b900dc152c 'Update model prices file'). If gpt-4o's price ever changes, isEqualByComparingTo("0.50") and the trip-at-limit assertion break silently, and the failure will look like a BudgetGuard regression rather than a price-data update. The test exercises the real CostService static table rather than a controlled price.

💡 Prefer a model whose price is stable/synthetic, or make the coupling explicit by deriving the expected cost from CostService.calculateCost for the same (model, provider, usage) rather than hardcoding 0.25/0.50, so a price-table update updates the expectation automatically instead of failing the guard test.

_general finding_

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/BudgetGuardTest.java
```diff
@@ -0,0 +1,111 @@
...
+    private static LlmProviderFactory factoryFor(String model, String provider) {
+        var factory = mock(LlmProviderFactory.class);
+        when(factory.getResolvedModelInfo(model))
+                .thenReturn(new LlmProviderFactory.ResolvedModelInfo(model, provider));
+        return factory;
+    }
+
+    @Test
+    void noBudgetSetReturnsSharedUnlimitedGuardThatNeverTrips() {
+        var guard = BudgetGuard.create(null, "gpt-4o", factoryFor("gpt-4o", "openai"));
```
BorisTkachenko: **[low] Unused mock stubbing in the null / non-positive budget tests**

In noBudgetSetReturnsSharedUnlimitedGuardThatNeverTrips (line 40) and nonPositiveBudgetIsTreatedAsUnlimitedNotAnInstantTrip (line 54), BudgetGuard.create returns UNLIMITED before it ever touches the factory (maxCostUsd null / signum <= 0 short-circuits at BudgetGuard.java:63). The when(factory.getResolvedModelInfo(...)) stubbing set up by factoryFor therefore never fires. It is dead setup that misleadingly implies model resolution participates in these paths (and would fail under strict-stubs MockitoExtension).

💡 Pass a bare mock(LlmProviderFactory.class) (or Mockito.mock without stubbing) for the null/zero/negative-budget cases, reserving factoryFor for the cases that actually resolve a model.

_rule: `backend-testing-use-real-specific-arguments-in-mock-verify-when-instead-of-a`_

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/BudgetGuardTest.java
```diff
@@ -0,0 +1,111 @@
...
+    @Test
+    void doesNotChargeWhenResponseCarriesNoUsage() {
+        var guard = BudgetGuard.create(new BigDecimal("0.01"), "gpt-4o", factoryFor("gpt-4o", "openai"));
+
+        var response = ChatResponse.builder().aiMessage(AiMessage.from("ok")).build();
+        guard.track(Mono.just(response)).block();
+
+        assertThat(guard.spentUsd()).isEqualByComparingTo("0");
+        assertThat(guard.shouldWrapUp()).isFalse();
+    }
```
BorisTkachenko, line 101: **[low] Per-call pricing fail-open branch in charge() is untested**

The class doc stresses that 'everything is fail-open'. The resolution fail-open branch (create catch) is covered by failsOpenToUnlimitedWhenModelCannotBeResolved, but the second fail-open branch — the try/catch inside charge() at BudgetGuard.java:91-93, which swallows a CostService/pricing exception and continues without charging — has no coverage. A regression that let that exception propagate (breaking the scoring call) would not be caught by this suite.

💡 Add a case where cost computation throws (e.g. a response/usage that makes CostService throw, or stub the pricing path) and assert track(...).block() still returns the response, spentUsd stays 0, and shouldWrapUp() is false.

_rule: `backend-testing-add-regression-tests-covering-all-new-backend-behavior-branc`_

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/ToolCallLoopTest.java
```diff
@@ -322,6 +470,42 @@ void noMediaMessageAppendedWhenToolsStageNothing() {
 
     // --- helpers ---
 
+    private static BudgetGuard overBudgetGuard() {
```
BorisTkachenko: **[low] Helper name overBudgetGuard() is misleading — the guard is not over budget when returned**

overBudgetGuard() returns a freshly created BudgetGuard with a $0.01 limit and zero recorded spend, so costGuard.shouldWrapUp() is false at return time. It only trips after the first tracked 200k-token follow-up response inside runToWrapUp/run. The name implies the guard is already over budget, which can mislead a future maintainer into writing a test that asserts shouldWrapUp() before any track() call.

💡 Rename to something like tightBudgetGuard() or budgetGuardTrippedByOneCall(), or add a one-line comment clarifying it becomes over-budget only after the first tracked response.

_general finding_

apps/opik-backend/src/test/java/com/comet/opik/domain/evaluation/LlmUsageExtractorTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+    @Test
+    void baseTokenUsageIsCapturedForAnyProvider() {
+        // The base type is what Vertex AI and OpenAI-compatible providers return — no cache subtype.
+        var usage = LlmUsageExtractor.toUsageMap(responseWith(new TokenUsage(100, 20, 120)));
+
+        assertThat(usage)
+                .containsEntry("prompt_tokens", 100)
+                .containsEntry("completion_tokens", 20)
+                .containsEntry("total_tokens", 120)
```
BorisTkachenko: **[low] total_tokens expected value coincides with prompt+completion, weakening the assertion**

In baseTokenUsageIsCapturedForAnyProvider the total is 120 = 100 + 20, and in openAiCacheTokensAreCaptured total is 250 = 200 + 50. Because the asserted total equals prompt+completion, the total_tokens assertion cannot distinguish a genuine pass-through of the provider-reported total from an extractor that (incorrectly) recomputes total as prompt+completion.

💡 Use a total that differs from prompt+completion (e.g. new TokenUsage(100, 20, 130) and assert total_tokens == 130) so the assertion actually verifies the provider total is passed through unchanged.

_general finding_

apps/opik-backend/src/test/java/com/comet/opik/domain/evaluation/LlmUsageExtractorTest.java
```diff
@@ -0,0 +1,113 @@
...
+        var usage = LlmUsageExtractor.toUsageMap(responseWith(GoogleAiGeminiTokenUsage.builder()
+                .inputTokenCount(400).outputTokenCount(70).cachedContentTokenCount(120).build()));
+
+        assertThat(usage)
+                .containsEntry("prompt_tokens", 400)
+                .containsEntry("cache_read_input_tokens", 120);
+    }
+
+    @Test
+    void omitsCacheKeysWhenCacheCountsAreNullOrZero() {
```
BorisTkachenko: **[low] Three independent scenarios bundled into one test method**

omitsCacheKeysWhenCacheCountsAreNullOrZero verifies three distinct branches (null cache fields, zero cache fields, null OpenAI details) sequentially with hard assertions. If the first block fails the remaining scenarios are never exercised, and the failure report names a single method rather than the specific branch that regressed.

💡 Split into separate @Test methods (or a @ParameterizedTest, or wrap in assertj's SoftAssertions) so each branch is independently reported and always executed.

_general finding_

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/ToolCallLoop.java
```diff
@@ -168,23 +189,40 @@ static Mono<ChatResponse> runWithWrapUp(
...
+        if (round >= MAX_TOOL_CALL_ROUNDS || costGuard.shouldWrapUp()) {
+            if (costGuard.shouldWrapUp() && round < MAX_TOOL_CALL_ROUNDS) {
+                // The spend budget (not the round cap) is cutting this agentic run short. Flag it here,
+                // at the authoritative point the guard actually fires, rather than inferring it later
+                // from shouldWrapUp(): a natural stop that merely crossed spend reaches the no-tool
+                // branch above (never here, so it isn't mislabelled), and flagging now means the
+                // monitoring trace is tagged budget_exceeded even if the wrap-up/scoring chain errors
+                // afterwards. Idempotent — the gate trips once. Demoted to debug (the user-facing warn
+                // in the scorer is the primary signal).
+                log.debug("Evaluation spend budget reached for '{}' (spent '{}' of '{}' USD); wrapping up",
```
BorisTkachenko: **[low] New budget-reached log line is not MDC-wrapped, so it drops workspace_id/rule_id correlation tags**

The log.debug at line 220 runs inside toolCallLoop, which executes on the boundedElastic worker that emitted the follow-up response via flatMap. That thread has no MDC set — this file re-applies MDC only inside executeToolOrBudgetExhausted (lines 278) via wrapWithMdc(mdc), precisely because "the reactive chain may have hopped threads since the scorer's sync prep step set MDC" (comment lines 272-277). The new budget log line bypasses that pattern, so it emits without the workspace_id / rule_id correlation tags the sibling tool-loop logs carry. It includes logIdValue in the message, which softens the impact, and it is debug level.

💡 Wrap the log.debug (and flag) in try (var logContext = wrapWithMdc(mdc)) like executeToolOrBudgetExhausted so the budget-reached line carries the same slf4j correlation tags as the other tool-loop logs.

_rule: `backend-observability-include-all-relevant-correlation-identifiers-workspaceid-pro`_

BorisTkachenko (commented):

13 findings — 0 high · 2 medium · 11 low. Suppressed by team conventions: see `suppressed.md`.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/ToolCallLoop.java
```diff
@@ -168,23 +189,40 @@ static Mono<ChatResponse> runWithWrapUp(
...
+        // reached, and let runWithWrapUp do its final tools-stripped call (the intended, bounded
+        // overshoot). Don't append: the response here carries unfulfilled tool_execution_requests
+        // (we're abandoning them). An AiMessage with tool_calls but no matching
+        // ToolExecutionResultMessage produces a malformed sequence that OpenAI / Anthropic reject.
+        // runWithWrapUp will bridge via the forcing user message.
+        if (round >= MAX_TOOL_CALL_ROUNDS || costGuard.shouldWrapUp()) {
+            if (costGuard.shouldWrapUp() && round < MAX_TOOL_CALL_ROUNDS) {
+                // The spend budget (not the round cap) is cutting this agentic run short. Flag it here,
+                // at the authoritative point the guard actually fires, rather than inferring it later
+                // from shouldWrapUp(): a natural stop that merely crossed spend reaches the no-tool
```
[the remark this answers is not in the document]
alexkuzmik: Fixed in 95c06b4e6e — the three budget signals now key off one authoritative event. `BudgetGuard.wasBudgetEnforced()` is set at the loop's budget gate (the same point that flags the `budget_exceeded` trace tag), and both the wrap-up instruction and the user-facing warn now key off it instead of `shouldWrapUp()`. A terminal no-tool-call response that merely crosses spend hits the no-tool branch above the gate, so all three signals (message, warn, tag) consistently treat it as a natural completion.

apps/opik-backend/src/main/java/com/comet/opik/api/evaluators/AutomationRuleEvaluatorLlmAsJudge.java
alexkuzmik: Fixed in 95c06b4e6e — added `@Valid` to the polymorphic `code` field on `AutomationRuleEvaluator` (and `AutomationRuleEvaluatorUpdate`) so nested constraints cascade. `@Positive` on `maxCostUsd` now rejects 0/negative with **422** at the API boundary; a new resource test asserts it. `BudgetGuard` keeps the non-positive→UNLIMITED coercion as defense-in-depth for any legacy persisted row. (Also taught PODAM to manufacture strictly-positive values for `@Positive` BigDecimal fields so the newly-enforced constraint doesn't flake existing create tests.)

apps/opik-backend/src/main/java/com/comet/opik/api/evaluators/AutomationRuleEvaluatorTraceThreadLlmAsJudge.java
alexkuzmik: Fixed in 95c06b4e6e — same `@Valid` cascade fix; the `@Positive` on `TraceThreadLlmAsJudgeCode.maxCostUsd` now fires at the API boundary (**422**), covered by the new resource test. Went with enforcing the constraint rather than dropping it.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/BudgetGuard.java
alexkuzmik: Fixed in 95c06b4e6e — `charge()` now detects the successful-but-unpriced case (non-null usage that prices to zero) and logs a single warn **once per guard**, so operators can see the spend limit is ineffective for that model. Updated the class javadoc's fail-open contract to cover it too.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringLlmAsJudgeScorer.java
alexkuzmik: Fixed in 95c06b4e6e — the user-facing warn now keys off `costGuard.wasBudgetEnforced()` (the loop's gate actually cut the run short), not `shouldWrapUp()`. It no longer fires on the inline single-call path or a natural stop that merely crossed spend — same authoritative signal as the `budget_exceeded` tag.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSpanLlmAsJudgeScorer.java
alexkuzmik: Fixed in 95c06b4e6e — reworded: this method **does** run the multi-turn agentic loop. The real reason `UNLIMITED` is correct is that the span evaluator DTO exposes no `maxCostUsd` field, so there is no budget to enforce. Dropped the 'single LLM call / no loop' claim.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/ToolCallLoop.java
alexkuzmik: Fixed in 95c06b4e6e — the wrap-up instruction now keys off `wasBudgetEnforced()` instead of `shouldWrapUp()`. A model that finishes naturally on the turn its cost tips over the limit gets the standard completion instruction (not `BUDGET_WRAP_UP`), with no warn and no tag — so message, warn, and tag all agree. A new `ToolCallLoop` test locks this natural-stop-over-budget case in.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/BudgetGuardTest.java
alexkuzmik: Fixed in 95c06b4e6e — `accumulatesSpendAcrossCallsAndTripsOnceLimitReached` now derives the per-call cost from `CostService.calculateCost` for the same `(model, provider, usage)` and computes the limit from it, so a bundled price-table update moves the expectation instead of silently breaking the guard test.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/BudgetGuardTest.java
alexkuzmik: Fixed in 95c06b4e6e — the null and non-positive budget tests now pass a bare `mock(LlmProviderFactory.class)`; `create()` short-circuits before resolving the model, so the `getResolvedModelInfo` stubbing was dead setup.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/BudgetGuardTest.java
alexkuzmik: Fixed in 95c06b4e6e — added `failsOpenWhenPerCallCostComputationThrows`: `mockStatic(CostService)` makes `calculateCost` throw inside `charge()`, and the test asserts `track(...)` still returns the response unchanged, `spentUsd` stays 0, and `shouldWrapUp()` is false.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/ToolCallLoopTest.java
alexkuzmik: Fixed in 95c06b4e6e — renamed `overBudgetGuard()` → `tightBudgetGuard()` and added a comment clarifying it is **not** over budget at return time; it only trips after the first tracked ~200k-token follow-up response is charged.

apps/opik-backend/src/test/java/com/comet/opik/domain/evaluation/LlmUsageExtractorTest.java
alexkuzmik: Fixed in 95c06b4e6e — `baseTokenUsageIsCapturedForAnyProvider` now uses `TokenUsage(100, 20, 130)` and asserts `total_tokens == 130`, so it verifies the provider-reported total is passed through rather than recomputed as prompt+completion.

apps/opik-backend/src/test/java/com/comet/opik/domain/evaluation/LlmUsageExtractorTest.java
alexkuzmik: Fixed in 95c06b4e6e — split `omitsCacheKeysWhenCacheCountsAreNullOrZero` into three independent `@Test` methods (Anthropic null-fields, Anthropic zero-fields, OpenAI null-details) so each branch is always exercised and reported by name.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/ToolCallLoop.java
alexkuzmik: Fixed in 95c06b4e6e — the budget-reached `log.debug` is now wrapped in `try (var logContext = wrapWithMdc(mdc))`, so it carries the same `workspace_id` / `rule_id` correlation tags as the sibling tool-loop log lines.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/BudgetGuard.java
```diff
@@ -0,0 +1,142 @@
...
+     * is a pure lookup. Fails open to {@link #UNLIMITED} if the model can't be resolved.
+     *
+     * <p>A non-positive limit is treated as "no limit" rather than enforced. {@code @Positive} on the
+     * DTO now rejects {@code 0}/negative values at the API boundary (the {@code code} field is
+     * {@code @Valid}-cascaded), but this coercion is kept as defense-in-depth for any value that still
+     * reaches here (e.g. a legacy row persisted before the constraint): enforcing it verbatim would make
+     * {@link #shouldWrapUp()} true on the first turn and silently wrap up every evaluation with no real
+     * work — worse than ignoring it.
+     */
+    public static BudgetGuard create(BigDecimal maxCostUsd, String modelName,
```
BorisTkachenko:
```suggestion
    public static BudgetGuard create(BigDecimal maxCostUsd, @NonNull String modelName,
```

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/BudgetGuard.java
```diff
@@ -0,0 +1,142 @@
...
+            return new BudgetGuard(maxCostUsd, resolved.actualModel(), resolved.provider());
+        } catch (Exception exception) {
+            log.warn("Failed to resolve model '{}' for budget guard; proceeding without a spend limit",
+                    modelName, exception);
+            return UNLIMITED;
+        }
+    }
+
+    /** Taps the LLM call to charge its cost, returning the response unchanged. Pass-through when unlimited. */
+    public Mono<ChatResponse> track(Mono<ChatResponse> call) {
```
BorisTkachenko:
```suggestion
    public Mono<ChatResponse> track(@NonNull Mono<ChatResponse> call) {
```

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluation/LlmUsageExtractor.java
```diff
@@ -0,0 +1,92 @@
...
+     * Returns null when the response carries no usage, so callers can skip costing entirely.
+     * <p>
+     * Prompt/completion/total tokens are read from the generic {@link TokenUsage} interface, so they
+     * are captured for <b>every</b> provider Opik supports (Anthropic, OpenAI chat + Responses,
+     * Gemini, Vertex AI, and the OpenAI-compatible providers — OpenRouter, Ollama, custom-llm, free —
+     * which all return either {@link OpenAiTokenUsage} or the base {@link TokenUsage}). The
+     * per-provider switch below only <i>adds</i> the prompt-cache token counts, which each SDK exposes
+     * under a different subtype.
+     */
+    public static Map<String, Integer> toUsageMap(ChatResponse response) {
```
BorisTkachenko:
```suggestion
    public static Map<String, Integer> toUsageMap(@NonNull ChatResponse response) {
```

BorisTkachenko (dismissed):
A few minor comments.

PR #7317, opened 2026-07-02
[NA] [BE] feat: always persist online-evaluation monitoring traces

alexkuzmik:
## Details

Online-evaluation monitoring traces (OPIK_6994) were gated behind the `onlineScoringTracingEnabled` service toggle (env `TOGGLE_ONLINE_SCORING_TRACING_ENABLED`, default `true`). This PR removes that configuration so the behavior is always on — every LLM-as-judge online evaluation (trace / span / thread) is always recorded as a hidden `source=evaluator` monitoring trace.

- The three online scorers now always start a real `EvaluationRecorder` instead of branching on the toggle.
- The trace-thread path keeps its best-effort `onErrorResume` → `EvaluationRecorder.NOOP` fallback so a project lookup failure still can't abort scoring.
- The field is dropped from the backend only: `ServiceTogglesConfig`, `config.yml`, and `config-test.yml`. The toggle is exposed via `GET /v1/private/toggles`, but per review the Fern-autogenerated OpenAPI specs and Python/TS `rest_api` types are left untouched — the daily Fern codegen job will drop `onlineScoringTracingEnabled` once the backend stops returning it.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- OPIK_6994 (origin feature; not resolved by this PR)

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Scope: Full change (flag removal across backend + generated SDK/OpenAPI types)
- Human verification: pending author review

## Testing

- `mvn -o compile` in `apps/opik-backend` — compiles cleanly.
- Grep-verified no remaining references to `onlineScoringTracingEnabled` / `TOGGLE_ONLINE_SCORING_TRACING_ENABLED` in the repo.
- Existing OPIK-6994 scorer/persistence tests already run with the recorder active (default was `true`), so they exercise the now-unconditional path.

## Documentation

The removed toggle was infra config only; no user-facing docs reference it.


apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/OnlineScoringLlmAsJudgeScorerTest.java
```diff
@@ -134,6 +134,10 @@ void setUp() {
         lenient().when(onlineScoringConfig.getMaxRetries()).thenReturn(3);
         lenient().when(onlineScoringConfig.getAgenticToolsCharsPerToken()).thenReturn(4);
 
+        lenient().when(onlineEvaluationRecorder.begin(
+                any(Trace.class), any(), any(), any(), any(), any()))
+                .thenReturn(EvaluationRecorder.NOOP);
```
andrescrz, line 146: Not a blocker: I recommend cleaning up the whole NOOP implementation and all related code. Mockito mocks should be enough for this scenario. Better remove tech debt in the code base for unnecessary code.

apps/opik-documentation/documentation/fern/openapi/opik.yaml
andrescrz, line 1: Better not remove any of the Fern autogenerated code and specs by hand. Let's just wait for the next daily job occurrence. The idea is not avoid conflicts and to not corrupt these accidentally.

andrescrz (dismissed):
I like the idea of this feature being always enabled, I believe there's no per customer need etc. to disable. I tested it in prod and looks nice.

However, I wouldn't rush to merge and deploy this. I'd recommend leave it baking for max 1 week in production, so you have this flag as a emergency stop measure, in case some issue happens.

Left couple of non-blocking comments, but I do recommend addressing them.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/OnlineScoringLlmAsJudgeScorerTest.java
alexkuzmik: Skipping in this PR — after removing the toggle, `NOOP` still has one load-bearing production use: the trace-thread scorer's `onErrorResume` fallback (`OnlineScoringTraceThreadLlmAsJudgeScorer` ~L266-270). Building a thread recorder needs a blocking `projectService.get(...)` lookup for the project name; if that throws, we degrade to `NOOP` so a best-effort monitoring failure can't abort real scoring, while the downstream `Mono.zip` + `recorder.monitor(...)` still has a passthrough recorder to call.

A bare Mockito mock isn't a drop-in here either — `recordLlmCall`/`recordToolCall`/`monitor` must return their in-flight `Mono` unchanged, and a default mock returns `null` and breaks the reactive chain, so it would need passthrough stubbing on every method (more code, re-implementing `NOOP`).

Removing `NOOP` entirely means either dropping that fallback (a robustness regression) or reworking `begin()`/`RealRecorder` to tolerate a missing project name — a separate refactor of pre-existing (OPIK-6994) code, out of scope for this toggle-removal PR. Happy to file a follow-up for the cleanup.

apps/opik-documentation/documentation/fern/openapi/opik.yaml
alexkuzmik: Fixed in 84a7661 — reverted all hand-edits to the Fern-autogenerated files (both OpenAPI specs + the Python/TS `rest_api` `ServiceTogglesConfig` types). They're back to their `main` state; the daily Fern codegen job will drop `onlineScoringTracingEnabled` once the backend stops returning it. Only the backend (`ServiceTogglesConfig`, `config.yml`, `config-test.yml`) removes it by hand now.

andrescrz (dismissed):

alexkuzmik:
### Don't merge before 13.07.2025

andrescrz (approved):

CometActions:
**No test needed here.**

The kill-switch this removes was already on by default — `onlineScoringTracingEnabled: ${TOGGLE_ONLINE_SCORING_TRACING_ENABLED:-"true"}` in config.yml, `"true"` in config-test.yml, and nothing in the repo sets it to false — so a default install writes exactly the same source=evaluator monitoring traces before and after. There is no new behaviour for an e2e test to catch; the three scorers just lose the ternary. Separately, and worth a look before merge: the config.yml and ServiceTogglesConfig.java hunks also bring back `agenticToolsEnabled`, `agentInsightsEnabled`, `v2WorkspaceAllowlist`, `v1WorkspaceAllowlist` and `forceWorkspaceVersion`, none of which exist on main any more — that reads like a merge resolution that took the old side, and the three YAML keys have no matching field on ServiceTogglesConfig, so Dropwizard config binding is worth re-checking.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/35754115248)

PR #5775, opened 2026-03-22
[NA] [BE][FE] feat: Add Feishu as native alert destination

rusherman:
## Details

Add Feishu/Lark (飞书) as the fourth native alert type alongside General, Slack, and PagerDuty.
Uses Feishu Interactive Card format for rich alert display in group chats via custom bot webhooks, covering all 10 alert event types with color-coded cards and "View in Opik" action buttons.

- Backend: DB migration, AlertType enum, 6 Feishu card model classes, FeishuWebhookPayloadMapper, AlertPayloadAdapter integration
- Frontend: ALERT_TYPE enum, Feishu SVG icon, helpers config, DestinationSelector update
- v1 scope: no Feishu signature verification (users can use IP allowlist); signature support planned for follow-up PR

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- NA (new feature contribution)

## AI-WATERMARK

AI-WATERMARK: yes

- If yes:
  - Tools: Claude Code CLI
  - Model(s): Claude Opus 4.6
  - Scope: Full implementation (backend models, mapper, migration, frontend types/icons/config, unit tests)
  - Human verification: Code reviewed, all 36 unit tests passing

## Testing

**Commands run:**
```bash
cd apps/opik-backend
mvn test -pl . -Dtest="com.comet.opik.api.resources.v1.events.webhooks.feishu.FeishuWebhookPayloadMapperTest" -Dsurefire.useFile=false
```

**Result:** 36 tests run, 0 failures, 0 errors, 0 skipped

**Scenarios validated:**
- Card structure: msg_type=interactive, header with plain_text, div with lark_md, action with primary button
- Template colors: red (TRACE_ERRORS), orange (threshold/guardrail events), blue (prompt/experiment events)
- All 10 event types: PROMPT_CREATED, PROMPT_DELETED, PROMPT_COMMITTED, TRACE_ERRORS, TRACE_FEEDBACK_SCORE, TRACE_THREAD_FEEDBACK_SCORE, TRACE_COST, TRACE_LATENCY, TRACE_GUARDRAILS_TRIGGERED, EXPERIMENT_FINISHED
- Empty metadata handling for all event types
- Default base_url fallback when alertMetadata is empty
- Action URL routing: single project → project link, multiple projects → projects list, thread tab type for TRACE_THREAD_FEEDBACK_SCORE
- Cost prefix ($) and latency suffix (s) formatting
- Window duration formatting (seconds/minutes/hours/days)

[...]

PR #7320, opened 2026-07-02
[BE][FE] feat: Add Feishu (Lark) as native alert destination

rusherman:
## Summary

Re-opens the Feishu (Lark) native alert destination from #5775 (closed inadvertently on the maintainers' side), rebased onto the latest `main` and reworked to address **all 7 automated review comments** from the original PR.

Feishu becomes the 4th native destination alongside General / Slack / PagerDuty, delivering alerts as [interactive cards](https://open.feishu.cn/document/common-capabilities/message-card/message-cards-content) to a group chat via a custom bot webhook.

## Structured as 3 reviewable commits

1. **refactor**: extract `AlertWebhookUtils` (shared `formatWindowDuration`, `formatEventType`, `resolveBaseUrl`) and switch `SlackWebhookPayloadMapper` to use it — no behavior change.
2. **feat**: Feishu mapper/models, `AlertType.FEISHU`, adapter case, migration, frontend, tests.
3. **chore**: sync OpenAPI spec, Python/TypeScript SDKs and docs for the new enum value.

## Addressing the original review

| Finding | Resolution |
|---|---|
| metadata/metadataUrl NPE | `resolveBaseUrl` normalizes null/blank via `Optional` |
| OpenAPI/SDK not regenerated | spec + Python + TypeScript (incl. serialization) + docs updated |
| Broad `catch (Exception)` | kept as an intentional per-payload fallback, now documented |
| Non-reversible ENUM rollback | migration marked forward-only with a remediation note |
| Duplicated `formatWindowDuration` | extracted to `AlertWebhookUtils` |
| Duplicated `formatEventType` | extracted to `AlertWebhookUtils` |
| FE mapping overwrote the title prefix | Feishu uses an empty field-mapping; the title is built server-side |

## Testing

- Backend: `FeishuWebhookPayloadMapperTest` (36 cases) green; `spotless:check` clean
- Frontend: `npm run build` passes (v1 + v2)
- SDK enum values synced across Python & TypeScript (including the serialization layer)

## Demo

https://www.youtube.com/watch?v=yE7wB1xJXLA


apps/opik-frontend/src/v1/pages/AlertsPage/AddEditAlertPage/helpers.ts
```diff
@@ -363,6 +366,9 @@ export const ALERT_FIELD_MAPPINGS: AlertTypeMappings = {
       targetPath: "blocks[0].text.text",
     },
   ],
+  // Feishu builds its full card (including the "Opik Alert: " title) server-side,
+  // so no field replacement is needed here; mapping name onto the title would drop that prefix.
+  [ALERT_TYPE.feishu]: [],
```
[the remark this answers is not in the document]
rusherman: Good catch — the empty mapping left the preview stuck on the placeholder. Fixed: `ALERT_FIELD_MAPPINGS[feishu]` now maps `name → card.header.title.content` in both v1 and v2 helpers. This only drives the live preview (the example payload shown in the UI); the actual sent card still gets the "Opik Alert: " prefix built server-side by `FeishuWebhookPayloadMapper`, so the two concerns are now properly separated.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/webhooks/common/AlertWebhookUtils.java
```diff
@@ -0,0 +1,70 @@
...
+     * Resolves the workspace base URL from alert metadata, falling back to the default when the
+     * key is missing, null or blank, and guaranteeing a single trailing slash before the workspace.
+     */
+    public static String resolveBaseUrl(@NonNull Map<String, String> alertMetadata, @NonNull String workspaceName) {
+        String metadataUrl = Optional.ofNullable(alertMetadata.get(BASE_URL_METADATA_KEY))
+                .filter(StringUtils::isNotBlank)
+                .orElse(DEFAULT_BASE_URL);
+        metadataUrl = metadataUrl.endsWith("/") ? metadataUrl : metadataUrl + "/";
+        return metadataUrl + workspaceName;
+    }
```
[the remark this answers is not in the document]
rusherman: Agreed — added `AlertWebhookUtilsTest` covering the missing key, null value, blank value, empty value, no-trailing-slash, and existing-trailing-slash cases for `resolveBaseUrl`.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000088_add_feishu_alert_type.sql
```diff
@@ -0,0 +1,9 @@
+--liquibase formatted sql
+--changeset rusherman:000088_add_feishu_alert_type
+--comment: Add 'feishu' to the alerts.alert_type ENUM for the Feishu (Lark) native alert destination
+
+ALTER TABLE alerts
+    MODIFY COLUMN alert_type ENUM('general', 'slack', 'pagerduty', 'feishu') NOT NULL DEFAULT 'general';
+
+-- forward-only: shrinking the ENUM back is unsafe once rows use alert_type='feishu'.
+-- To revert, first delete or migrate those rows, then restore the previous ENUM definition from backup.
```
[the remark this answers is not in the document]
rusherman: Fixed — the migration now ends with `--rollback empty` (matching the convention in 000027/000029 for ENUM-expanding migrations). The forward-only rationale moved into the `--comment` directive so Liquibase parses it correctly and the human-readable note is preserved.

apps/opik-frontend/src/v1/pages/AlertsPage/AddEditAlertPage/helpers.ts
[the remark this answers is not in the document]
rusherman: Already addressed in the current PR version: the Feishu preview mapping is present in both v1 and v2 alert configuration flows.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000094_add_feishu_alert_type.sql
```diff
@@ -0,0 +1,8 @@
+--liquibase formatted sql
+--changeset rusherman:000088_add_feishu_alert_type
+--comment: Add 'feishu' to the alerts.alert_type ENUM for the Feishu (Lark) native alert destination. Forward-only (mirrors 000027/000029): shrinking the ENUM back is unsafe once rows use alert_type='feishu'.
+
+ALTER TABLE alerts
+    MODIFY COLUMN alert_type ENUM('general', 'slack', 'pagerduty', 'feishu') NOT NULL DEFAULT 'general';
+
+--rollback empty
```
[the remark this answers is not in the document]
rusherman: Fixed in ade24c2a2f by adding the required blank line at EOF. The focused backend test suite passes (42 tests).

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000088_add_feishu_alert_type.sql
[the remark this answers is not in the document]
rusherman: Already addressed in the current PR version: the migration now includes the required rollback directive.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/webhooks/common/AlertWebhookUtils.java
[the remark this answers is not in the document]
rusherman: Addressed in the current PR version with edge-case coverage for base URL resolution. Re-verified locally in AlertWebhookUtilsTest.

andrescrz (commented):

16 findings — 0 high · 6 medium · 10 low. Suppressed by team conventions: 1 (see `suppressed.md`).

---

### Findings on unchanged lines (can't be inline-anchored)
These flag existing lines the PR didn't modify, so GitHub can't attach an inline comment:

- **[low] `apps/opik-documentation/documentation/fern/docs/production/alerts.mdx:91`** — Native integrations list not updated to include Feishu
  - Line 91 still reads 'Native integrations (Slack, PagerDuty): Use built-in formatting for popular services'. Now that Feishu is a native integration (per the top-of-page list at line 22 and AlertType.FEISHU in the backend), this enumeration is stale and contradicts the four-destination claim added by this change.
- **[low] `apps/opik-documentation/documentation/fern/docs/production/alerts.mdx:163`** — Middleware destination-type list omits Feishu
  - Line 163 states the middleware approach 'works with any destination type (General, Slack, or PagerDuty)'. With Feishu now a supported destination, this parenthetical is incomplete and implies Feishu is not compatible with the middleware pattern.

apps/opik-backend/src/main/java/com/comet/opik/api/AlertType.java
```diff
@@ -12,7 +12,8 @@
 public enum AlertType {
     GENERAL("general"),
     SLACK("slack"),
-    PAGERDUTY("pagerduty");
+    PAGERDUTY("pagerduty"),
+    FEISHU("feishu");
```
andrescrz, line 16: **[medium] New FEISHU alert type is not verified by the parameterized webhook integration tests**

Adding FEISHU to AlertType widens every test that enumerates AlertType.values() without extending the assertions. AlertResourceTest#testWebhook__whenWebhookServerRespondsWithSuccess (provider at AlertResourceTest.java:1145-1148) now runs for FEISHU, but its verifier assertWebhookTestResultRequest (AlertResourceTest.java:1189-1246) only branches on GENERAL/SLACK/PAGERDUTY; for FEISHU no branch matches, so the request body is never deserialized or asserted and the case passes silently (misleading green coverage). Separately, the two alertTypeProvider() streams that drive the event-delivery tests and the webhook-examples endpoint test (AlertResourceTest.java:2310-2311 and 3041-3043) still list only SLACK/PAGERDUTY(/GENERAL), so end-to-end FEISHU delivery is never exercised. Only the mapper is covered (FeishuWebhookPayloadMapperTest).

💡 Add an else-if branch in assertWebhookTestResultRequest that deserializes the FEISHU body into FeishuWebhookPayload and asserts it is non-null (mirroring the SLACK/PAGERDUTY branches), and add AlertType.FEISHU to both alertTypeProvider() streams so the delivery and webhook-example integration tests exercise it.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/webhooks/feishu/FeishuWebhookPayloadMapper.java
```diff
@@ -0,0 +1,269 @@
...
+
+        // Add content div with summary + details
+        String content = buildContent(event);
+        elements.add(FeishuCardElement.div(FeishuText.larkMd(content)));
+
+        // Add "View in Opik" button if applicable
+        String actionUrl = buildActionUrl(event);
+        if (actionUrl != null) {
+            elements.add(FeishuCardElement.action(
+                    List.of(FeishuAction.primaryButton("View in Opik", actionUrl))));
```
andrescrz, line 71: **[medium] No size/truncation guard on Feishu card content (Slack sibling truncates)**

buildContent() concatenates one markdown line per event into a single lark_md div with no length cap. The Slack sibling (SlackWebhookPayloadMapper) deliberately enforces SLACK_TEXT_BLOCK_LIMIT (3000) via checkSlackTextLimit(), truncating and emitting a 'View All' fallback. Alerts aggregate many events (many prompts created, many guardrail-triggered traces, etc.), so for busy workspaces the content string can grow large. Feishu interactive cards have a maximum message size, and an oversized card is rejected by the Feishu API, causing the entire alert notification to be silently lost rather than truncated. The team treating this as important for Slack signals it is a real operational concern, but it is unhandled here.

💡 Cap the number of rendered items (or total content length) and append a 'View all in Opik' fallback link when exceeded, mirroring the Slack mapper's truncation-with-fallback behavior; ideally share the limit logic via AlertWebhookUtils.

_general finding_

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/webhooks/common/AlertWebhookUtilsTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class AlertWebhookUtilsTest {
+
+    private static final String WORKSPACE = "my-workspace";
+    private static final String DEFAULT_BASE = "http://localhost:5173";
+
+    @Nested
+    class ResolveBaseUrl {
```
andrescrz, line 22: **[medium] New util's formatWindowDuration/formatEventType branches left untested**

AlertWebhookUtils is newly added in this PR with three public methods, but this dedicated unit test only exercises resolveBaseUrl. formatWindowDuration and formatEventType have no coverage here. Checking the rest of the suite, the mapper tests only indirectly hit a few cases (1 hour, 1 day, 2 hours, 30 minutes in FeishuWebhookPayloadMapperTest; there is no Slack mapper test at all). The formatWindowDuration 'seconds' fallback branch (seconds < 60) and the singular boundaries (1 minute, 1 second, 1 hour vs plural) are not exercised anywhere in the suite, and formatEventType's per-enum label mapping is likewise unasserted, so a regression in either (e.g. wrong pluralization or a missing switch arm) would go undetected.

💡 Add parameterized cases in this util test for formatWindowDuration covering each branch and singular/plural boundary (e.g. 30 -> "30 seconds", 1 -> "1 second", 60 -> "1 minute", 3600 -> "1 hour", 86400 -> "1 day", plus plural forms) and for formatEventType covering every AlertEventType value.

apps/opik-documentation/documentation/fern/docs-v2/production/alerts/alerts.mdx
```diff
@@ -15,9 +15,10 @@ title: Alerts
 
 Alerts allow you to configure automated webhook notifications for important events in your Opik workspace. When specific events occur — such as trace errors, new feedback scores, or prompt changes — Opik sends HTTP POST requests to your configured endpoint with detailed event data.
 
-Opik provides three destination types for alerts:
+Opik provides four destination types for alerts:
 - **Slack**: Native integration with automatic message formatting for Slack
 - **PagerDuty**: Native integration with automatic event formatting for PagerDuty
+- **Feishu**: Native integration with automatic interactive-card formatting for Feishu (Lark)
```
andrescrz, line 21: **[medium] Feishu added as a native type but sibling destination-type lists left stale**

This change promotes Feishu to a fourth native destination (intro list line 18-22, and the Feishu section at line 160), but two other enumerations of destination types were not updated and now contradict it: line 90 still says 'Native integrations (Slack, PagerDuty)' and line 192 (middleware section) still says 'This approach works with any destination type (General, Slack, or PagerDuty)'. Readers will conclude Feishu is not a native integration and/or is unsupported by middleware, contradicting the new section.

💡 Update line 90 to 'Native integrations (Slack, PagerDuty, Feishu)' and line 192 to include Feishu in the destination-type list, so all enumerations stay consistent with the new fourth type.

_general finding_

apps/opik-documentation/documentation/fern/docs/production/alerts.mdx
```diff
@@ -16,9 +16,10 @@ canonical-url: https://www.comet.com/docs/opik/production/alerts/alerts
 
 Alerts allow you to configure automated webhook notifications for important events in your Opik workspace. When specific events occur — such as trace errors, new feedback scores, or prompt changes — Opik sends HTTP POST requests to your configured endpoint with detailed event data.
 
-Opik provides three destination types for alerts:
+Opik provides four destination types for alerts:
 - **Slack**: Native integration with automatic message formatting for Slack
 - **PagerDuty**: Native integration with automatic event formatting for PagerDuty
+- **Feishu**: Native integration with automatic interactive-card formatting for Feishu (Lark)
```
andrescrz, line 22: **[medium] Feishu added as native destination but has no setup section**

The change advertises Feishu as a fourth native integration (with automatic interactive-card formatting), but unlike the other two native destinations, there is no '### Feishu integration (Native)' section under 'Integration examples' (Slack has one at line 95, PagerDuty at line 127). Users selecting Feishu get no prerequisites, setup steps, or explanation of what fields are required, leaving the newly documented destination undocumented for actual use.

💡 Add a '### Feishu integration (Native)' subsection mirroring the Slack/PagerDuty ones: prerequisites (create a Feishu/Lark custom group bot, copy its webhook URL) and step-by-step setup in Opik, plus a note on the interactive-card formatting Opik applies.

_general finding_

apps/opik-frontend/src/icons/feishu.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
+  <path d="M2.5 4.5C3.5 3 5.5 2 7.5 2.5C7 4 6.5 5.5 7 7C8.5 6 10.5 5.5 12 6C11.5 8.5 10 11 7.5 12.5C5.5 13.5 3.5 13 2 11.5C3 10 4 8.5 4 7C4 6 3.5 5.5 2.5 4.5Z" fill="currentColor"/>
```
andrescrz: **[medium] feishu.svg path is a generic placeholder shape, not the Feishu/Lark brand mark**

This icon is imported as FeishuIcon (@/icons/feishu.svg?react) and rendered as the destination logo for the Feishu alert type in both v1 and v2 AlertsPage (ALERT_TYPE_ICONS[ALERT_TYPE.feishu]). The single <path> here draws two abstract overlapping leaf/petal curves — it does not resemble the actual Feishu (Lark) logo (the blue rounded-square speech bubble with the white paper-plane/bird). Sibling brand icons in this folder use the real marks (pagerduty.svg renders the PagerDuty 'P' letterform, github.svg the Octocat, slack.svg the Slack glyph). Shipping this renders an unrecognizable blob in the destination selector, so users cannot identify the Feishu option by its logo.

💡 Replace the hand-drawn path with the official Feishu/Lark logo geometry (monochrome, single-path, fill="currentColor", 16x16 viewBox to match pagerduty.svg) so the destination icon is recognizable and consistent with the other brand icons.

_general finding_

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/webhooks/common/AlertWebhookUtilsTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+        @Test
+        void appendsTrailingSlashWhenMissing() {
+            Map<String, String> metadata = Map.of(AlertWebhookUtils.BASE_URL_METADATA_KEY, "https://opik.example.com");
+
+            assertThat(AlertWebhookUtils.resolveBaseUrl(metadata, WORKSPACE))
+                    .isEqualTo("https://opik.example.com/" + WORKSPACE);
+        }
+
+        @Test
```
andrescrz: **[low] preservesSingleTrailingSlash overstates what it verifies**

The test name and the resolveBaseUrl javadoc both claim a single trailing slash is guaranteed, but the implementation only checks endsWith("/") and never collapses repeated slashes. A base_url like "https://opik.example.com//" would yield "https://opik.example.com//my-workspace". The test only feeds a single trailing slash, so the double-slash (and no-slash-with-multiple) case is unverified and the doc's normalization promise is untested, masking the impl/doc mismatch.

💡 Either add a case asserting the intended normalization for "https://opik.example.com//" (which will currently fail, surfacing the impl gap) or rename the test to reflect that it only preserves an existing single slash.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/webhooks/feishu/FeishuWebhookPayloadMapperTest.java
```diff
@@ -0,0 +1,435 @@
...
+
+        @Test
+        void promptCreated() {
+            UUID promptId = UUID.randomUUID();
+            var event = buildEvent(AlertEventType.PROMPT_CREATED,
+                    List.of(Prompt.builder().id(promptId).name("test-prompt").build()));
+
+            FeishuWebhookPayload payload = FeishuWebhookPayloadMapper.toFeishuPayload(event);
+
+            String content = payload.card().elements().get(0).text().content();
```
andrescrz, line 138: **[low] Assert the full rendered content, not scattered substrings**

The rendered card content is fully deterministic here (UUIDs are captured in local variables, baseUrl is the known constant BASE_URL, workspaceName is WORKSPACE_NAME). Yet nearly every EventTypeDetails/ActionUrls test verifies only a handful of isolated `.contains(...)` fragments. Substring matching is weakly discriminating: it will not catch formatting regressions such as a wrong/extra separator, reordered fields, a broken markdown link, or a missing prefix around the asserted token. The sibling AlertWebhookUtilsTest already asserts full outputs with isEqualTo(...).

💡 For the deterministic cases, assert the whole rendered string with isEqualTo(...) (or a single tightly-scoped expected block per element), so any formatting drift fails the test.

_rule: `backend-testing-assert-whole-objects-full-responses-with-assertj-not-individ`_

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/webhooks/feishu/FeishuWebhookPayloadMapperTest.java
```diff
@@ -0,0 +1,435 @@
...
+            assertThat(content).contains("[View]");
+        }
+
+        @Test
+        void traceErrors() {
+            var event = buildEvent(AlertEventType.TRACE_ERRORS,
+                    List.of(MetricsAlertPayload.builder()
+                            .metricValue("15")
+                            .threshold("10")
+                            .windowSeconds(3600)
```
andrescrz, line 186: **[low] Untested branches: malformed-payload fallback, count>1 pluralization, multi-item rendering**

Several new branches of the mapper are never exercised: (1) formatMetricsAlertPayload's catch block that emits "- <type> alert (unable to parse details)" on a malformed payload; (2) the summary pluralization branch (count != 1 -> "events") — every metrics test uses a single-item list (count==1) and the empty case does not assert the summary line; (3) multi-item metadata joining (joining("\n") / ", ") for the prompt/guardrail/experiment detail builders, all tested with exactly one item. These are the divergent branches most likely to regress silently.

💡 Add a multi-item case (asserting the joined output and the pluralized "events" summary) and, if practical, a payload that drives formatMetricsAlertPayload into its fallback string. The malformed-payload path is defensive, so keep it low priority.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/webhooks/feishu/FeishuWebhookPayloadMapperTest.java
```diff
@@ -0,0 +1,435 @@
...
+
+        @Test
+        void shouldUseDefaultBaseUrlWhenMetadataEmpty() {
+            var event = WebhookEvent.<Map<String, Object>>builder()
+                    .id("event-1")
+                    .eventType(AlertEventType.PROMPT_CREATED)
+                    .alertType(AlertType.FEISHU)
+                    .alertId(UUID.randomUUID())
+                    .alertName(ALERT_NAME)
+                    .alertMetadata(Map.of())
```
andrescrz: **[low] DefaultBaseUrl re-inlines the whole WebhookEvent builder instead of reusing buildEvent**

shouldUseDefaultBaseUrlWhenMetadataEmpty duplicates the entire WebhookEvent builder just to set an empty alertMetadata, because buildEvent hardcodes alertMetadata with BASE_URL. This copy diverges from buildEvent (e.g. fixed id "event-1") and must be kept in sync by hand.

💡 Parameterize buildEvent to accept the alertMetadata map (defaulting to the BASE_URL map) so this test can call buildEvent(..., Map.of()) instead of re-inlining the builder.

_rule: `backend-testing-extract-repeated-test-assertion-logic-into-a-shared-helper-i`_

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/webhooks/feishu/FeishuWebhookPayloadMapper.java
```diff
@@ -0,0 +1,269 @@
...
+            String windowDuration = AlertWebhookUtils.formatWindowDuration(payload.windowSeconds());
+
+            String scope = (payload.projectIds() == null || payload.projectIds().isEmpty())
+                    ? "*Workspace-wide*"
+                    : TemplateUtils.newST(PROJECTS_TEMPLATE)
+                            .add("projectNames", payload.projectNames())
+                            .render();
+
+            String valuePrefix = type.equals("Cost") ? "$" : "";
+            String valueSuffix = type.equals("Latency") ? " s" : "";
```
andrescrz, line 275: **[low] Slack mrkdwn '*Workspace-wide*' leaked into lark_md payload**

The workspace-wide scope string "*Workspace-wide*" was copied verbatim from SlackWebhookPayloadMapper.formatMetricsAlertPayload (line 301), but the Feishu div renders with tag lark_md (FeishuText.larkMd, set at line 59), not Slack mrkdwn. In lark_md a single asterisk is italic, whereas the rest of this file's template (METRICS_ALERT_DETAILS_TEMPLATE and PROJECTS_TEMPLATE, e.g. '**Projects:**') uses double-asterisk bold. So this one field renders inconsistently (italic instead of bold). Divergent formatting provenance introduced by reusing Slack syntax across the Feishu destination.

💡 Use lark_md syntax for the Feishu variant: change "*Workspace-wide*" to "**Workspace-wide**" to match the double-asterisk bold used elsewhere in the Feishu templates.

_general finding_

apps/opik-frontend/src/v2/pages/AlertsPage/AddEditAlertPage/helpers.ts
```diff
@@ -319,6 +322,15 @@ export const ALERT_FIELD_MAPPINGS: AlertTypeMappings = {
       targetPath: "blocks[0].text.text",
     },
   ],
+  // Feishu builds its full card (including the "Opik Alert: " title prefix) server-side when
+  // sending; this mapping only drives the live preview, so the user-entered alert name shows up
+  // in the example payload's card title instead of the static placeholder.
+  [ALERT_TYPE.feishu]: [
+    {
+      sourceField: "name",
+      targetPath: "card.header.title.content",
```
andrescrz, line 387: **[low] Feishu preview title drops the server-added "Opik Alert: " prefix**

The new ALERT_TYPE.feishu mapping replaces card.header.title.content (sourceField "name") with the raw user-entered alert name. The backend example this preview is applied to (AlertService.WEBHOOK_EXAMPLES -> AlertPayloadAdapter.webhookEventPayloadPerType -> FeishuWebhookPayloadMapper.toFeishuPayload) sets that same field to "Opik Alert: " + alertName. So applyFieldReplacements strips the "Opik Alert: " prefix that the actually-sent payload always carries, and the preview even changes shape before/after the user types (empty name keeps the placeholder "Opik Alert: Example Alert", a typed name shows the bare name with no prefix). The preview therefore misrepresents the real payload title. Same issue exists identically in the v1 helpers.ts (line 375).

💡 Either prepend the same "Opik Alert: " prefix the mapper adds (e.g. map to a derived value rather than the raw name), or leave the Feishu title mapping out so the preview shows the true server-built "Opik Alert: <name>" title. Keep v1 and v2 in sync.

_general finding_

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/webhooks/feishu/FeishuWebhookPayloadMapper.java
```diff
@@ -0,0 +1,269 @@
...
+        if (actionUrl != null) {
+            elements.add(FeishuCardElement.action(
+                    List.of(FeishuAction.primaryButton("View in Opik", actionUrl))));
+        }
+
+        return FeishuWebhookPayload.builder()
+                .msgType("interactive")
+                .card(FeishuCard.builder()
+                        .header(FeishuCardHeader.builder()
+                                .title(FeishuText.plainText("Opik Alert: " + event.getAlertName()))
```
andrescrz, line 78: **[low] Header title built from user-controlled alert name without length clamp**

The card header title is set to "Opik Alert: " + event.getAlertName() with no length bound. Alert names are user-controlled and can be long. The Slack sibling explicitly clamps the alert name to SLACK_HEADER_BLOCK_LIMIT (150) in createHeaderBlock() precisely because header text has a platform limit; Feishu plain_text titles are likewise length-limited, so an overly long alert name can cause the card to be rejected. This is defensive-hardening parity rather than a common-path failure.

💡 Clamp the title to Feishu's header limit with Math.min-style substring, consistent with the Slack mapper's header handling.

_general finding_

apps/opik-documentation/documentation/fern/docs-v2/production/alerts/alerts.mdx
```diff
@@ -154,6 +157,36 @@ Opik will automatically format all alert payloads into PagerDuty-compatible even
...
+   - Click "Create new alert"
+   - Give your alert a descriptive name
+   - Select **Feishu** as the destination type
+   - Enter the Feishu bot webhook URL: `https://open.feishu.cn/open-apis/bot/v2/hook/...`
+   - Add triggers for the events you want to monitor
+   - Click "Test connection" to verify
+   - Click "Create alert"
+
+Opik will automatically format all alert payloads into Feishu interactive cards with:
+- Color-coded headers by severity (red for errors, orange for metric thresholds, blue for informational events)
```
andrescrz: **[low] Header color description omits guardrails (also orange, not a metric threshold)**

The doc says 'orange for metric thresholds'. In FeishuWebhookPayloadMapper.determineTemplate, orange is also used for TRACE_GUARDRAILS_TRIGGERED, which is not a metric/threshold event. The description is therefore slightly inaccurate about what maps to orange.

💡 Reword to something like 'orange for metric-threshold and guardrail events' (or 'orange for warnings') so it matches the actual color mapping in the backend.

_general finding_

apps/opik-backend/src/main/java/com/comet/opik/api/AlertType.java
rusherman: Fixed in 175edb4b25. Added FEISHU deserialization to assertWebhookTestResultRequest and included FEISHU in both parameterized providers. AlertResourceTest also compiles successfully; its container-backed execution was intentionally not run locally.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/webhooks/feishu/FeishuWebhookPayloadMapper.java
rusherman: Fixed in 175edb4b25. The lark_md content is now capped at 25,000 UTF-8 bytes with code-point-safe truncation and an explicit View in Opik fallback, leaving room under the official 30 KB card request limit. Added a multibyte truncation test. Reference: https://open.feishu.cn/document/uAjLw4CM/ukTMukTMukTM/reference/im-v1/message/create

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/webhooks/common/AlertWebhookUtilsTest.java
rusherman: Fixed in 175edb4b25. Added parameterized formatWindowDuration coverage for every unit and singular/plural boundaries, plus an exhaustive formatEventType table for all AlertEventType values.

apps/opik-documentation/documentation/fern/docs-v2/production/alerts/alerts.mdx
rusherman: Fixed in 175edb4b25. Updated the native integration and middleware destination lists to include Feishu, including the later middleware forwarding example.

apps/opik-documentation/documentation/fern/docs/production/alerts.mdx
rusherman: Fixed in 175edb4b25. Added the full Feishu native integration section to the v1 docs with prerequisites, bot setup, Opik configuration, formatting behavior, colors, and action links.

apps/opik-frontend/src/icons/feishu.svg
rusherman: Fixed in 175edb4b25. Replaced the placeholder leaf with recognizable Feishu brand geometry in a 16x16 monochrome currentColor SVG.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/webhooks/common/AlertWebhookUtilsTest.java
rusherman: Fixed in 175edb4b25. resolveBaseUrl now strips repeated trailing slashes before adding exactly one workspace separator; the parameterized test covers zero, one, and two trailing slashes.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/webhooks/feishu/FeishuWebhookPayloadMapperTest.java
rusherman: Fixed in 175edb4b25. Deterministic card bodies and action URLs now use full isEqualTo assertions, covering exact markdown structure and complete links.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/webhooks/feishu/FeishuWebhookPayloadMapperTest.java
rusherman: Fixed in 175edb4b25. Added coverage for malformed metrics fallback, two-item rendering, pluralized summaries, and exact joined output.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/webhooks/feishu/FeishuWebhookPayloadMapperTest.java
rusherman: Fixed in 175edb4b25. buildEvent now accepts an alertMetadata override, and the default-base-URL test reuses it instead of duplicating the full builder.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/webhooks/feishu/FeishuWebhookPayloadMapper.java
rusherman: Fixed in 175edb4b25. Workspace-wide now uses Feishu lark_md bold syntax: **Workspace-wide**. Exact rendered-content tests cover it.

apps/opik-frontend/src/v2/pages/AlertsPage/AddEditAlertPage/helpers.ts
rusherman: Fixed in 175edb4b25. Field mappings now support a preview-only value transform; both v1 and v2 prepend Opik Alert: and have regression tests that also verify the example payload is not mutated.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/webhooks/feishu/FeishuWebhookPayloadMapper.java
rusherman: Checked against the current constraints and kept the title unchanged. Alert.name is already limited to 255 characters, and the Feishu title specification defines a four-line display limit with overflow ellipsis rather than a character rejection cap. The new 25 KB content guard addresses the documented request-size rejection risk without introducing an arbitrary second title limit. Reference: https://open.feishu.cn/document/feishu-cards/card-json-v2-components/content-components/title

apps/opik-documentation/documentation/fern/docs-v2/production/alerts/alerts.mdx
rusherman: Fixed in 175edb4b25. The v1 and v2 docs now describe orange headers as metric-threshold and guardrail events.

rusherman:
Addressed the latest review feedback in commit 175edb4b25.

Highlights:
- Added FEISHU integration assertions/providers and strengthened backend coverage with exact payload assertions, malformed fallback, multi-item rendering, pluralization, URL normalization, and all shared formatter branches.
- Added a UTF-8-safe 25 KB Feishu card-content cap with a visible truncation fallback, leaving room under the documented 30 KB request-body limit.
- Kept v1/v2 preview behavior aligned with the server title, replaced the placeholder icon with the Feishu brand glyph, and synchronized both documentation versions.
- Fixed the unanchored v1 docs findings as well: the native-integration and middleware destination lists now include Feishu.

Validation performed without Docker:
- Backend focused unit tests: 65 passed
- Backend test compilation, including AlertResourceTest: passed
- Frontend unit tests under CI Node 20 and UTC: 2007 passed
- Frontend lint, typecheck, dependency validation, and production build: passed
- git diff --check: passed

The Testcontainers-backed AlertResourceTest runtime was intentionally not executed locally because this update was requested without Docker.

rusherman:
Merged the latest `main` and resolved the documentation conflict while preserving the updated Guardrails link and the Feishu documentation changes.

Post-merge validation (no Docker):
- Backend focused Maven tests: 65 passed
- Frontend Vitest: 2261 passed
- Frontend typecheck: passed
- Frontend dependency validation: passed
- Frontend production build: passed
- Targeted frontend ESLint: passed
- PR diff whitespace check: passed

The branch is now up to date with `main` at `fb483e5512`.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/webhooks/feishu/FeishuWebhookPayloadMapper.java
```diff
@@ -0,0 +1,309 @@
...
+ * @see <a href="https://open.feishu.cn/document/common-capabilities/message-card/message-cards-content">Feishu Card</a>
+ */
+@Slf4j
+@UtilityClass
+public class FeishuWebhookPayloadMapper {
+
+    // Feishu limits interactive card request bodies to 30 KB. Keep the content below that limit
+    // so the header, actions, URLs and JSON framing have sufficient room.
+    private static final int MAX_CONTENT_BYTES = 25_000;
+    private static final String TRUNCATION_SUFFIX = "\n\n_Content truncated. Use \"View in Opik\" to see all events._";
```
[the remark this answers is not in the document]
rusherman: Fixed in 3afa551a5c. The mapper now builds the complete card first and measures the exact JSON produced by JsonUtils, which is the same serializer used by AlertPayloadAdapter before the request is sent. If the full body exceeds 30,000 UTF-8 bytes, it uses a code-point-safe binary search to retain the longest content prefix whose fully serialized card, including the suffix, header, action URL, framing, and escaping, fits the limit.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/webhooks/feishu/FeishuWebhookPayloadMapper.java
```diff
@@ -0,0 +1,309 @@
...
+
+        return truncateContent(summary + "\n\n" + details);
+    }
+
+    private static String truncateContent(String content) {
+        if (content.getBytes(StandardCharsets.UTF_8).length <= MAX_CONTENT_BYTES) {
+            return content;
+        }
+
+        int byteBudget = MAX_CONTENT_BYTES - TRUNCATION_SUFFIX.getBytes(StandardCharsets.UTF_8).length;
```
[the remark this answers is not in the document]
rusherman: Fixed in 3afa551a5c. The 25,000-byte raw-content heuristic has been removed. The limit is now enforced against the complete serialized FeishuWebhookPayload via the same JsonUtils path used by the webhook adapter, with the truncation suffix included in every candidate measurement.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/webhooks/feishu/FeishuWebhookPayloadMapper.java
```diff
@@ -0,0 +1,309 @@
...
+                .collect(Collectors.joining("\n"));
+
+        return String.format("**%s Alert Triggered:**\n%s", metricType, alertDetails);
+    }
+
+    private static String formatMetricsAlertPayload(@NonNull MetricsAlertPayload payload, String type) {
+        try {
+            String windowDuration = AlertWebhookUtils.formatWindowDuration(payload.windowSeconds());
+
+            String scope = (payload.projectIds() == null || payload.projectIds().isEmpty())
```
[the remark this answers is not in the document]
rusherman: I evaluated this extraction and am keeping the destination formatters separate. Only the invariant pieces are shared in AlertWebhookUtils today. Slack and Feishu use different markdown templates, scope syntax, project-link placement, size limits, and fallback behavior; moving their assembly into the common utility would require destination-specific callbacks or branching and would reduce cohesion for a small amount of structural similarity.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/webhooks/feishu/FeishuWebhookPayloadMapperTest.java
```diff
@@ -0,0 +1,529 @@
...
+            var event = buildEvent(AlertEventType.TRACE_ERRORS, List.of(MetricsAlertPayload.builder()
+                    .metricValue("飞".repeat(10_000))
+                    .threshold("10")
+                    .windowSeconds(3_600)
+                    .build()));
+
+            FeishuWebhookPayload payload = FeishuWebhookPayloadMapper.toFeishuPayload(event);
+            String content = payload.card().elements().get(0).text().content();
+
+            assertThat(content.getBytes(StandardCharsets.UTF_8)).hasSizeLessThanOrEqualTo(25_000);
```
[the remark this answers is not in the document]
rusherman: Fixed in 3afa551a5c. The regression test now serializes the complete FeishuWebhookPayload with JsonUtils and asserts the final UTF-8 body is at most 30,000 bytes. Parameterized cases cover JSON escaping, the ASCII boundary, and supplementary Unicode; the test also verifies valid UTF-8 content and preserves the truncation suffix.

rusherman:
Addressed the latest Baz review in `3afa551a5c`:

- Enforce the 30,000-byte limit against the fully serialized Feishu card, using the same `JsonUtils` path as the webhook adapter.
- Account for header, action URL, JSON framing, escaping, and the truncation suffix.
- Preserve Unicode code-point boundaries while retaining the longest prefix that fits.
- Add full-body regression coverage for JSON-escaped, ASCII-boundary, and supplementary-Unicode content.

Validation (no Docker): 67 focused backend tests passed; full backend main/test compilation completed as part of Maven; `git diff --check` passed. All four new review threads have been replied to and resolved.

andrescrz (commented):

6 findings — 1 high · 0 medium · 5 low. Suppressed by team conventions: see `suppressed.md`.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/AlertResourceTest.java
```diff
@@ -2365,7 +2370,8 @@ void setUp() {
         private Stream<Arguments> alertTypeProvider() {
             return Stream.of(
                     Arguments.of(AlertType.SLACK),
-                    Arguments.of(AlertType.PAGERDUTY));
+                    Arguments.of(AlertType.PAGERDUTY),
+                    Arguments.of(AlertType.FEISHU));
```
andrescrz, line 2374: **[high] FEISHU added to alertTypeProvider but verifyPayload throws on it**

AlertType.FEISHU was added to TestAlertEventsForNativeIntegrations.alertTypeProvider() (line 2374), which feeds ~10 @ParameterizedTest methods (testPromptCreatedEvent, testPromptDeletedEvent, testPromptCommittedEvent, testTraceError..., feedback score, guardrail, experiment, cost, latency at lines 2526/2564/2592/2629/2676/2744/2799/2846/3004/3049). Each of those tests calls verifyPayload(alertType, ...). The verifyPayload switch (lines 2409-2421) only has cases for SLACK and PAGERDUTY; its default branch does `throw new IllegalArgumentException("Unsupported alert type for alerts integration: " + alertType)`. So every FEISHU parameter immediately hits the default and throws, failing all of these test invocations. verifyFeishuPayloadStructure / a FEISHU case was not added to complete the change.

💡 Add a `case FEISHU ->` branch to verifyPayload that deserializes FeishuWebhookPayload and asserts its structure (e.g. a verifyFeishuPayloadStructure helper mirroring verifySlackBlockStructure / verifyPagerDutyPayloadStructure). If FEISHU payload assertions are not yet ready, do not add FEISHU to this alertTypeProvider until the verifier supports it.

_general finding_

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000088_add_feishu_alert_type.sql
```diff
@@ -0,0 +1,8 @@
+--liquibase formatted sql
+--changeset rusherman:000088_add_feishu_alert_type
```
andrescrz: **[low] Migration number 000088 collides with an existing migration and is non-sequential**

The migrations directory already contains 000088_create_cipx_user_mappings.sql, so this new file reuses sequence number 000088, and the highest existing number is already 000091. changelog.xml uses <includeAll> which orders files alphabetically, so the two 000088 files run in name order and the changeset ID (rusherman:000088_...) is distinct, so this is not a functional break — but it violates the strictly-sequential file-numbering convention and a reviewer will likely ask to bump the number to avoid the in-repo collision on merge.

💡 Rename the file (and changeset id) to the next free sequential number (000092) to avoid the duplicate-000088 collision already present in the checkout.

_general finding_

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/webhooks/feishu/FeishuWebhookPayloadMapperTest.java
```diff
@@ -0,0 +1,540 @@
...
+    }
+
+    @Nested
+    class ActionUrls {
+
+        @Test
+        void metricsAlertWithSingleProjectShouldLinkToProject() {
+            String projectId = UUID.randomUUID().toString();
+            var event = buildEvent(AlertEventType.TRACE_ERRORS,
+                    List.of(MetricsAlertPayload.builder()
```
andrescrz, line 439: **[low] Action-URL branches for non-metrics events are never asserted**

buildActionUrl in the mapper has distinct branches that produce different URLs per event type: PROMPT_CREATED/PROMPT_COMMITTED/PROMPT_DELETED -> baseUrl+"/prompts", TRACE_GUARDRAILS_TRIGGERED -> baseUrl+"/projects", EXPERIMENT_FINISHED -> baseUrl+"/experiments", and the metrics zero/blank-project case -> baseUrl+"/projects". The ActionUrls nested class only exercises metrics-alert event types (single project, multiple projects, threads tab). The prompt/guardrail/experiment action-URL branches and the workspace-wide (no projectIds) metrics branch have no assertion on the produced url, so a regression in those switch arms would go undetected. shouldContainDivAndActionElements checks only the button text, not the URL.

💡 Add assertions on actionElement.actions().get(0).url() for a prompt event (expect .../prompts), a guardrails event (expect .../projects), an experiment event (expect .../experiments), and a metrics event with no projectIds (expect .../projects).

apps/opik-documentation/documentation/fern/docs-v2/production/alerts/alerts.mdx
```diff
@@ -154,9 +157,39 @@ Opik will automatically format all alert payloads into PagerDuty-compatible even
...
+   - Open the target group chat → Settings → Bots
+   - Click "Add Bot" and choose "Custom Bot"
+   - Give the bot a name (e.g., "Opik Alerts") and copy the generated Webhook URL
+
+2. **In Opik**:
+   - Go to Configuration → Alerts tab
+   - Click "Create new alert"
+   - Give your alert a descriptive name
+   - Select **Feishu** as the destination type
+   - Enter the Feishu bot webhook URL: `https://open.feishu.cn/open-apis/bot/v2/hook/...`
```
andrescrz: **[low] Feishu example URL omits the Lark (international) host**

The page repeatedly brands the destination as "Feishu (Lark)", but every example webhook URL (lines 59, 62, 180) only shows the mainland-China host `https://open.feishu.cn/open-apis/bot/v2/hook/...`. International Lark tenants issue custom-bot webhooks on `https://open.larksuite.com/open-apis/bot/v2/hook/...`. The backend applies no host allowlist (AlertType only carries the `feishu` value and the mapper does no URL validation; only the generic `http://`/`https://` prefix rule applies), so Lark URLs work fine — but a Lark user following the doc verbatim has no indication their host differs.

💡 Add the Lark host variant next to the Feishu example, e.g. note that Lark (international) tenants use `https://open.larksuite.com/open-apis/bot/v2/hook/...`.

_general finding_

apps/opik-documentation/documentation/fern/docs-v2/production/alerts/alerts.mdx
```diff
@@ -154,9 +157,39 @@ Opik will automatically format all alert payloads into PagerDuty-compatible even
...
+   - Add triggers for the events you want to monitor
+   - Click "Test connection" to verify
+   - Click "Create alert"
+
+Opik will automatically format all alert payloads into Feishu interactive cards with:
+- Color-coded headers by severity (red for errors, orange for metric-threshold and guardrail events, blue for informational events)
```
andrescrz: **[low] "orange for metric-threshold events" is imprecise vs. the actual mapping**

The severity legend says "red for errors, orange for metric-threshold and guardrail events, blue for informational events". But `trace:errors` is itself a metric-threshold event (it carries a threshold + window), yet FeishuWebhookPayloadMapper.determineTemplate() colors it red, not orange. Describing the orange bucket as "metric-threshold events" overlaps with the red errors case and can read as contradictory. The actual mapping is: red = trace:errors; orange = cost, latency, trace/thread feedback score, guardrails; blue = prompt created/committed/deleted and experiment finished.

💡 Rephrase to avoid the overlap, e.g. "orange for other threshold and guardrail events (cost, latency, feedback score, guardrails)" so errors (red) isn't ambiguously included in the orange category.

_general finding_

apps/opik-frontend/src/v2/pages/AlertsPage/AddEditAlertPage/helpers.ts
```diff
@@ -33,12 +34,14 @@ export const ALERT_TYPE_LABELS: Record<ALERT_TYPE, string> = {
   [ALERT_TYPE.general]: "General",
   [ALERT_TYPE.slack]: "Slack",
   [ALERT_TYPE.pagerduty]: "PagerDuty",
+  [ALERT_TYPE.feishu]: "Feishu",
 };
 
 export const ALERT_TYPE_ICONS = {
```
andrescrz: **[low] ALERT_TYPE_ICONS is not typed as Record<ALERT_TYPE, ...>, so a missing key crashes at render**

Unlike ALERT_TYPE_LABELS (declared `Record<ALERT_TYPE, string>`, which forces exhaustiveness at compile time), ALERT_TYPE_ICONS has no type annotation and relies on inference. The feishu key was added correctly here, but the object silently permits a missing ALERT_TYPE key. AlertTypeCell.tsx does `const Icon = ALERT_TYPE_ICONS[type]; <Icon />`, so a future ALERT_TYPE without an icon entry would compile fine yet yield `Icon === undefined` and throw 'Element type is invalid' at render instead of failing the build. Pre-existing gap, but the diff extends this exact map.

💡 Annotate the map as `Record<ALERT_TYPE, React.FunctionComponent<...>>` (or the appropriate icon component type) so any future alert type that omits an icon is a compile error rather than a runtime crash.

_rule: `frontend-null-safety-guards-keep-type-declarations-honest-align-ts-types-json-schema-nul`_

andrescrz:
Hi @rusherman 

Thanks for the contribution!

When you have a chance, could you take a look at my latest review comments? For each comment, please either address it with a code change or reply with a brief explanation if you believe the current implementation is the right approach.

In addition, could you please resolve the merge conflicts in the PR? Once that's done, we'll continue with the review. 

Thanks!

PR #7323, opened 2026-07-02
[OPIK-7193] [BE] Page-key enrichment aggregate CTEs in trace search

thiagohora:
## Details

The trace list query (`find_traces_by_project_id`) keyed its enrichment aggregate CTEs — feedback scores, span aggregates, comments, guardrails, annotation queues, experiments — on the full filtered trace-id set (`trace_id_prefilter`), then LEFT JOINed them to the single page of rows. For large projects the aggregates scanned whole tables (millions of spans/scores) only for the joins to discard everything outside the page, and because ClickHouse inlines CTE references the id set was re-evaluated at every reference site.

When neither filtering nor sorting reads those aggregates (the common enrichment-only case), this keys them on the page ids instead, consumed via `IN (SELECT arrayJoin((SELECT groupArray(id) FROM page_ids)))`:

- The inner scalar subquery is evaluated **once and cached** for the whole query, so `page_ids` is not re-executed at each reference.
- The materialized constant array is **primary-key-prunable**, turning whole-project aggregate scans into page-sized index lookups.

A new `shouldPageKeyAggregates(...)` gate selects this mode; the existing `trace_id_prefilter` path is kept unchanged for filtered/sorted queries where aggregates drive page selection (feedback-score / guardrails / trace-aggregation / annotation-queue filters, or sorts on feedback scores / span statistics / experiments).

- Result-equivalence validated against the previous query across projects covering all join paths (trace-level feedback, span-level feedback, comments + experiments + deep-offset paging): identical normalized output.
- On a large project the worst-case enrichment page dropped from ~100 s / 220M rows / ~44 GiB peak RAM to ~0.6 s / ~3.8M rows / ~540 MiB.
- Verified on ClickHouse 25.3 and 25.8: forward CTE references, the `arrayJoin`-scalar form, scalar-subquery caching, and primary-key pruning of the materialized IN-set all behave identically.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-7193

## AI-WATERMARK

AI-WATERMARK: yes

- If yes:
  - Tools: Claude Code
  - Model(s): Claude Opus 4.8
  - Scope: Query analysis on ClickHouse, the CTE rewrite + Java gating, equivalence/performance/version validation, and this PR text.
  - Human verification: Author reviewed the diff and the prod equivalence/perf and 25.3/25.8 compatibility results.

## Testing

- `mvn -o compile` in `apps/opik-backend` — clean.
- Rendered `SELECT_BY_PROJECT_ID` for all three modes (page-keyed, existing prefilter, no-narrowing) via StringTemplate and confirmed each parses on ClickHouse 25.3 (`EXPLAIN AST`); page-keyed mode emits the `arrayJoin`-scalar form at all aggregate sites with no `trace_id_prefilter` references, and the prefilter mode is byte-unchanged.
- Result-equivalence + performance measured directly against the query engine on representative projects (see PR description).

## Documentation
NA


apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -853,6 +853,14 @@ AND toMonday(id_at) = toMonday(UUIDv7ToDateTime(toUUID(:id), 'UTC'))
...
      * the id-range — and, unlike a {@code created_at} predicate, safe against late-arriving rows since it derives
      * from {@code id} — that lets the planner prune partitions once {@code traces} is partitioned.
+     * <p>
+     * When aggregates are enrichment-only ({@code page_keyed_aggregates}, see
+     * {@code shouldPageKeyAggregates}), the aggregate CTEs are keyed on
+     * {@code IN (SELECT arrayJoin((SELECT groupArray(id) FROM page_ids)))} instead of
+     * {@code trace_id_prefilter}: the inner scalar subquery is evaluated once and cached for the whole query,
+     * so every aggregate scans only the page's traces instead of the full filtered project. The aggregate CTEs
+     * referencing {@code page_ids} before its definition is fine — CTE names resolve independently of
+     * declaration order.
```
JetoPistola, line 863: ❓ **question** | Testing

This is the load-bearing invariant of the whole change — three ClickHouse behaviors have to all hold for it to be both correct and fast:

1. **Forward CTE resolution**: the aggregate CTEs (`feedback_scores_deduped` … `experiments_agg`, all defined *above*) reference `page_ids`, which is defined *below* them.
2. **Scalar-subquery caching**: `(SELECT groupArray(id) FROM page_ids)` must be evaluated once and reused, not re-run at each of the ~8 reference sites — otherwise this is slower, not faster.
3. **Primary-key pruning** of the materialized `IN`-set.

The existing `GetTracesByProjectResourceTest` suite (178 tests) does exercise the enrichment path, and since page-keyed mode is now the default for the common no-filter/no-aggregate-sort case, those tests now run *through* this SQL and assert result-equivalence — so correctness is functionally covered.

What isn't covered is a guard that pins any of the three behaviors above. All were validated manually on CH 25.3/25.8 per the PR description, but nothing in CI would catch a regression if a future ClickHouse upgrade changed forward-reference resolution or stopped caching the scalar subquery (the latter degrades silently — correct results, whole-project scans back). Given the [CTE / temp-table failure class](https://github.com/comet-ml/opik) this codebase has hit before, would an `EXPLAIN`-shape assertion (no per-reference re-evaluation of `page_ids`; IN-set is a materialized constant) be worth adding, or at least a comment that a CH upgrade must re-verify these three points?

JetoPistola:
👋 **Review summary**

**What looks good**
- The `shouldPageKeyAggregates` gate is genuinely airtight — I traced every consumer of the aggregate CTEs and each aggregate-driving path (feedback-score filters via `hasFeedbackScoreFilters` covering all 4 variants, guardrails / trace-aggregation / annotation-queue filters, and feedback-score / span-statistic / experiment sorts) is correctly excluded. Nothing that reads an aggregate to *select* the page slips into page-keyed mode.
- `duration`/`ttft` sorts correctly stay eligible since they come from `traces_deduped` trace columns, not an aggregate — easy to get wrong, right here.
- The `findTraceStream` path reuses `page_ids LIMIT :limit`, so page-keyed aggregates cover exactly the returned rows — the `false, false, false` sort args are justified by the comment.
- Excellent PR writeup: the equivalence validation across join paths + the 100s→0.6s / 44GiB→540MiB numbers + explicit 25.3/25.8 verification make the change easy to trust.

**Overall**
Sound, well-scoped optimization keyed off a precise invariant, with the existing `trace_id_prefilter` path left byte-unchanged for the filtered/sorted cases. No correctness concerns found.

**Inline comments**: 2 questions (Testing, Logic) — both non-blocking: pinning the load-bearing ClickHouse behaviors in CI vs. relying on manual version validation, and the substring-based `sortHasExperiment` check now gating page-keying.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -3647,8 +3697,14 @@ private Mono<? extends Result> getTracesByProjectId(
             boolean sortHasFeedbackScores = Optional.ofNullable(orderBySql)
                     .map(sortFields -> sortFields.contains("feedback_scores"))
                     .orElse(false);
+            boolean sortHasSpanStatistics = hasSpanStatistics(orderBySql);
+            boolean sortHasExperiment = Optional.ofNullable(orderBySql)
```
JetoPistola: ❓ **question** | Logic

`sortHasExperiment` matches on `orderBySql.contains("experiment_id")`. Because `EXPERIMENT_FIELD_MAPPING` rewrites `experiment_id` → `eaag.experiment_name`, the branch that actually fires for an experiment sort is `contains("eaag.experiment")`; the `experiment_id` substring check is now a redundant belt-and-suspenders match against the rendered SQL string.

This logic is pre-existing (it was just relocated out of the `ifPresent` block), so this isn't introduced here. But it's now **load-bearing for the page-keying gate** — a false negative would let an experiment-sorted query into page-keyed mode and key `experiments_agg` on `page_ids`, so it's worth a second look. Since it's a substring match on the whole ORDER BY SQL, is there any chance a dynamic sort key (e.g. a feedback-score name) could contain the literal `experiment_id` and trip the match — or, conversely, a mapping change could ever emit an experiment sort whose SQL contains neither `experiment_id` nor `eaag.experiment`? A structured check against the `SortableFields.EXPERIMENT_ID` field rather than the rendered string would be more robust, if it's cheap.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
thiagohora: Ran the three behaviors as an explicit probe matrix across ClickHouse versions (docker, same probes for all: forward-ref query; 3-reference scalar-cache query asserting `ScalarSubqueriesCacheMiss`/`GlobalCacheHit` from `query_log`; 1M-row PK-prune query asserting `read_rows`):

| Image | Version | Forward CTE refs | Scalar cache (3 refs) | PK pruning (1M rows) |
|---|---|---|---|---|
| `clickhouse-server:24.3-alpine` | 24.3.18.7 | ✅ | ✅ miss=4, hit=2 | ✅ read 8,203 |
| `clickhouse-server:24.8-alpine` | 24.8.14.39 | ✅ | ✅ miss=4, hit=2 | ✅ read 8,203 |
| `clickhouse-server:25.3` | 25.3.14.14 | ✅ | ✅ miss=4, hit=2 | ✅ read 8,203 |
| `clickhouse-server:25.8.16.34-alpine` (shipped) | 25.8.16.34 | ✅ | ✅ miss=4, hit=2 | ✅ read 8,203 |
| `altinity:25.8.16.10002.altinitystable` (helm) | 25.8.16 | ✅ | ✅ miss=4, hit=2 | ✅ read 8,203 |
| `clickhouse-server:latest-alpine` | 26.5.1.882 | ✅ | ✅ miss=4, hit=2 | ✅ read 8,203 |

So the invariants hold across every version from 24.3 LTS to current head, with byte-identical profile counters — this isn't a recently-stabilized behavior, all three predate 24.3. Side note: 26.5 newly accepts the plain `IN <scalar-alias>` form that every version ≤25.8 rejects with `UNSUPPORTED_METHOD`, so the `arrayJoin` wrapper is the compatibility-safe spelling for the foreseeable future.

On pinning it in CI: agreed in principle — the scalar-cache regression is the one that degrades silently. The probes above translate directly into a container-only test (no app bootstrap: raw JDBC against the CH testcontainer, `SYSTEM FLUSH LOGS`, assert the two ProfileEvents and `read_rows`), which would fail on the CI image bump the moment an upgrade changes any of the three. Happy to add it to this PR or as a follow-up — reviewer's choice; at minimum the template javadoc will carry the "upgrade must re-verify these three" note.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
thiagohora: Checked both directions:

- **Dynamic sort key tripping the match**: can't happen through the SQL string — dynamic keys render as `feedback_scores[:bindKey]` with the score *name* passed as a bound parameter, so a score literally named `experiment_id` never appears in the rendered ORDER BY. And any feedback-score sort sets `sortHasFeedbackScores`, which already disables page-keying — so even a hypothetical false positive here only disables the optimization, never mis-keys `experiments_agg`.
- **Mapping change emitting neither substring**: theoretically possible (`EXPERIMENT_FIELD_MAPPING` is the only rewrite and its target happens to start with `eaag.experiment` today), and that's the direction that would actually mis-gate.

Agreed the structured check is strictly more robust and cheap: detecting the experiment sort from `criteria.sortingFields()` against `SortableFields.EXPERIMENT_ID` (the mapping's key, i.e. what the API accepts) instead of the rendered SQL. I have that change staged and will push it to this PR.

thiagohora: Landed in 2b3261d — `sortHasExperiment` now checks `criteria.sortingFields()` against `SortableFields.EXPERIMENT_ID`, and the template javadoc carries the CH-upgrade re-verification note (with the version matrix from the other thread: 24.3 → 26.5 all hold).

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -3647,8 +3702,18 @@ private Mono<? extends Result> getTracesByProjectId(
...
                     .map(sortFields -> sortFields.contains("feedback_scores"))
                     .orElse(false);
+            boolean sortHasSpanStatistics = hasSpanStatistics(orderBySql);
+            // Structured check on the requested sort fields rather than a substring match on the rendered
+            // ORDER BY SQL: this now gates page-keyed aggregates, so it must track the experiment sort exactly
+            // even if EXPERIMENT_FIELD_MAPPING changes what SQL the field renders to.
+            boolean sortHasExperiment = Optional.ofNullable(traceSearchCriteria.sortingFields())
+                    .map(sortingFields -> sortingFields.stream()
+                            .anyMatch(sortingField -> SortableFields.EXPERIMENT_ID.equals(sortingField.field())))
+                    .orElse(false);
```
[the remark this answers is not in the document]
thiagohora: Partial coverage already exists: `TracesResourceTest` has a parameterized ASC/DESC experiment-sort test (`SortableFields.EXPERIMENT_ID`, both directions, asserting the full ordered slice). Two properties make it sufficient as a regression guard for the *gate* this PR touches:

- A mis-gated `sortHasExperiment` fails **loudly, not silently**: if the flag doesn't fire, `page_ids` loses its `experiments_agg` join while the rendered ORDER BY still references `eaag.experiment_name` — an unknown-identifier error, which that test surfaces immediately in both directions.
- For experiment-sorted queries this PR is deliberately behavior-identical to main (page-keying is disabled whenever the sort reads aggregates), so multi-page experiment sorting exercises a code path this change doesn't alter.

A paginated (page-size < result-set) variant is still reasonable hardening for the sort itself — happy to take it as a follow-up alongside the other test-hardening item from this review, rather than growing this PR.

andrescrz (commented):
Change LGTM, but I miss test coverage for the new branching, covered by black box testing.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
thiagohora: Landed the CI guard in 24b490f: `TracePageKeyedAggregatesPlanShapeTest` renders the real `SELECT_BY_PROJECT_ID` template in page-keyed mode against the real migrated schema (500 traces as pagination source, 300k spans as the aggregate population) and asserts the plan shape from `system.query_log`: reads must stay page-sized (PK pruning — a regression to whole-project aggregate scans blows the read-rows budget 3x over) and the page-id scalar must be served from the cache at the remaining reference sites (forward CTE resolution is implicitly pinned too — the query errors if it breaks). Runs in ~16s on the shared test containers, and the CI image bump is exactly the trigger you were worried about: any CH upgrade that changes one of the three behaviors fails this test instead of degrading silently. Same guard added on the span side in #7325.

apps/opik-backend/src/test/java/com/comet/opik/domain/TracePageKeyedAggregatesPlanShapeTest.java
```diff
@@ -0,0 +1,180 @@
...
+            statement.execute(sql);
+        }
+    }
+
+    private List<Long> queryLogRow(String columns) {
+        return await().atMost(Duration.ofSeconds(30)).pollInterval(Duration.ofMillis(500))
+                .until(() -> {
+                    execute("SYSTEM FLUSH LOGS");
+                    try (var statement = connection.createStatement()) {
+                        var resultSet = statement.executeQuery("""
```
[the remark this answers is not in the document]
thiagohora: Fixed in 08d1f75 — the query_log wait now uses `untilAsserted` with an `.as("query_log entry for log_comment '...'")` assertion, so a missing row reports the query-specific context instead of a generic Awaitility timeout. Same commit also hardens the test per the sibling review on the span PR: the page-keyed predicate is now count-asserted at all 9 aggregate CTE sites, and trace-level feedback scores (legacy + authored) and comments are seeded so those aggregate paths are exercised too. Applied the same `untilAsserted` fix to the span test in a834f5b.

apps/opik-backend/src/test/java/com/comet/opik/domain/TracePageKeyedAggregatesPlanShapeTest.java
```diff
@@ -0,0 +1,215 @@
...
+
+    @Test
+    void pageKeyedAggregatesReadPageSizedSlicesAndCacheThePageIdScalar() throws SQLException {
+        var query = renderPageKeyedQuery();
+        assertThat(query).doesNotContain("trace_id_prefilter");
+        // all nine aggregate CTE sites (trace + span feedback scores, guardrails, spans, comments, annotation
+        // queues, experiment items x2) must carry the page-keyed predicate; a bare contains() would pass even
+        // if the template dropped it from some of them
+        var pageKeyedPredicate = "IN (SELECT arrayJoin((SELECT groupArray(id) FROM page_ids)))";
+        assertThat(StringUtils.countMatches(query, pageKeyedPredicate)).isEqualTo(9);
```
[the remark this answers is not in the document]
thiagohora: Fixed in 464b077 — the test now slices the rendered query into named CTE blocks (`cteBlock` helper: from `<name> AS (` to the next CTE definition) and asserts the expected predicate count inside each aggregate CTE (`feedback_scores_deduped`: 2, `experiments_agg`: 2, the other five: 1 each), with the global total kept as a backstop. A drop in one CTE masked by a duplicate in another now fails on both the dropped and the duplicated block. Same change applied to the span test in 40d342f.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -3647,8 +3704,18 @@ private Mono<? extends Result> getTracesByProjectId(
...
 
-            if (shouldUseTraceIdPrefilter(traceSearchCriteria, template) && !sortHasFeedbackScores) {
+            if (shouldPageKeyAggregates(template, sortHasFeedbackScores, sortHasSpanStatistics, sortHasExperiment)) {
+                template.add("page_keyed_aggregates", true);
+            } else if (shouldUseTraceIdPrefilter(traceSearchCriteria, template) && !sortHasFeedbackScores) {
                 template.add("trace_id_prefilter", true);
             }
```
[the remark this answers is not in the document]
thiagohora: Extracted in the latest commit — `addAggregateKeyingFlags(template, criteria, sortHasFeedbackScores, sortHasSpanStatistics, sortHasExperiment)` now carries the decision tree for both `getTracesByProjectId` and `findTraceStream` (the stream passes `false` for all sort flags, reproducing its previous behavior exactly, including the `!sortHasFeedbackScores` guard which trivially holds there). Same shape as the `SpanDAO` helper from the sibling PR. Note the trace-side guard is deliberately kept (unlike the span PR where it made the branch unreachable): here the prefilter branch stays live via the trace-aggregation / annotation-queue / span-stats / experiment sort cases.

apps/opik-backend/src/test/java/com/comet/opik/domain/TracePageKeyedAggregatesPlanShapeTest.java
```diff
@@ -0,0 +1,251 @@
...
+ * per reference site. Both degrade silently — results stay correct while cost regresses to whole-project
+ * scans — so this test renders the real template against the real migrated schema and asserts the plan
+ * shape from {@code system.query_log}.
+ * <p>
+ * Seeded so the two failure modes are unmistakable: a small trace set (the pagination source) and a large
+ * span set. If pruning breaks, the query reads the whole span table for the project and blows through the
+ * read-rows budget; if scalar caching breaks, the cache-hit counter drops to zero.
+ */
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class TracePageKeyedAggregatesPlanShapeTest {
```
andrescrz: We generally implement high value black box or integration tests, through the public APIs (and exceptionally through the public service or DAO methods).

I don't think this kind of low level tests, with synthetic data, are desirable nor maintainable. Better go with the suggested approach.

if not reproducible, then it's a code smell. Why are we performing a change of a condition that can't be tested? Do we really need it?

andrescrz (approved):

PR #7328, opened 2026-07-02
[OPIK-7197] fix: avoid materializing wide internal temp tables in prompt/dataset version lookups

thiagohora:
## Details

`PromptVersionDAO.find` wrapped `SELECT pv.*` (which pulls large `TEXT`/`JSON` columns such as the prompt template) in a `paginated` CTE that is referenced twice, forcing MySQL to materialize a **wide internal temporary table**. Under concurrent load the server's `TempTable` memory pools (RAM + mmap overflow) saturate, MySQL can no longer allocate a new internal temp table, and the failure surfaces to the app as `SQLSyntaxErrorException: Table '.../#sql...' doesn't exist`. This rewrites it as a flat `SELECT` on `prompt_versions`, with `environments` fetched via an indexed correlated subquery — no CTE, no materialized temp table.

`DatasetVersionDAO.findByTag` had the same anti-pattern (a `version_sequences` CTE referenced twice); it now computes `version_name` / `tags` / `is_latest` via correlated subqueries.

- Both rewrites produce **zero** internal temp tables (verified via `EXPLAIN ANALYZE`) and are equal or faster than the originals.
- Behavior is unchanged — verified result-identical (same rows, same environment set, same `version_name`/`tags`/`is_latest`). `environments` is a `Set`, so JSON array element order (non-deterministic in the old query too) is irrelevant.
- The same double-referenced-CTE pattern still exists in other `DatasetVersionDAO` methods (`findByBatchGroupId`, `findByHash`, `findByDatasetId`, `findByVersionName`, `findByIds`); it's tracked as follow-up in the linked ticket. Note the paginated ones need a *set-based* `seq_num` (a per-row `COUNT` is O(versions)×page-size).

### EXPLAIN ANALYZE (production replica)

**`PromptVersionDAO.find`** — list page of a prompt with 17 versions:

| | Internal temp tables | Actual time |
|---|---|---|
| Before | 3 materializations (`paginated` CTE **wide** + `ver_envs` CTE + outer `Temporary table` **wide**) | ~0.53 ms |
| After | **0** (index lookup + row-id sort; `environments` = covering-index subquery, ~0.004 ms × rows) | ~0.39 ms |

**`DatasetVersionDAO.findByTag`** — dataset with 7,053 versions, returns 1 row:

| | Internal temp tables | Actual time |
|---|---|---|
| Before | `version_sequences` CTE materialized (7,053 rows) + `IN (…)` re-materialized with dedup (7,053) + 7,053 tag lookups | ~104 ms |
| After | **0** (`version_name` = covering-index `COUNT`; `tags`/`is_latest` = correlated subqueries) | ~7.5 ms (~14× faster) |

After both rewrites the plans contain **no `Materialize` / `Temporary table` nodes**.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- OPIK-7197

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: root-cause investigation, query rewrite, and validation
- Human verification: query plans and result-identity reviewed against a production read replica

## Testing

- `mvn -o test-compile` on `opik-backend` — passes.
- `EXPLAIN ANALYZE` on old vs new for both queries: new plans contain no `Materialize` / `Temporary table` nodes; the correlated subqueries resolve to covering-index lookups (numbers above).
- Result-identity: ran old vs new for both queries and diffed the output (`id` + `version_name` / `environments` / `tags` / `is_latest`) — identical.
- Full backend integration tests: to run in CI.

## Documentation


apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetVersionDAO.java
```diff
@@ -299,17 +298,18 @@ WITH version_sequences AS (
...
+                COALESCE((
+                    SELECT JSON_ARRAYAGG(t.tag)
+                    FROM dataset_version_tags t
+                    WHERE t.version_id = dv.id
+                ), JSON_ARRAY()) AS tags,
+                EXISTS(
+                    SELECT 1
+                    FROM dataset_version_tags t2
+                    WHERE t2.version_id = dv.id AND t2.tag = 'latest'
+                ) AS is_latest
```
JetoPistola, line 310: ❓ **question** | Testing

Result-identity of these rewrites is well covered by existing tests (`DatasetVersionResourceTest` asserts `versionName` = `v1/v2/v3`, fetch-by-tag, and `latest`), so a functional regression would fail CI — good. But the *actual* failure this PR fixes (`SQLSyntaxErrorException: Table '.../#sql...' doesn't exist`) is a MySQL server-side temp-table allocation failure under concurrent load, not a functional/syntax bug — it's non-deterministic and can't be reproduced by a normal integration test.

Given this failure class has recurred (OPIK 6735 — O(n) CTE in the same `DatasetVersionDAO` → RDS storage exhaustion, plus the active OPIK 7196/OPIK 7193 aggregate-CTE work), the durable guard is a **plan-shape assertion**: run `EXPLAIN FORMAT=JSON` on these DAO queries and assert the plan contains no `Materialize` / `Temporary table` node — turning this PR's manual `EXPLAIN ANALYZE` verification into a CI regression guard. Not needed in this PR; filed as follow-up **OPIK 7198**.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetVersionDAO.java
```diff
@@ -275,18 +275,17 @@ int deleteTag(@Bind("dataset_id") UUID datasetId, @Bind("tag") String tag,
...
                 dv.dataset_id,
                 dv.version_hash,
-                CONCAT('v', vs.seq_num) AS version_name,
+                CONCAT('v', (
+                    SELECT COUNT(*)
+                    FROM dataset_versions s
+                    WHERE s.workspace_id = :workspace_id
+                        AND s.dataset_id = :dataset_id
+                        AND s.id <= dv.id
+                )) AS version_name,
```
JetoPistola, line 288: ❓ **question** | Design

Not a blocker — flagging for the follow-up. `version_name` is derived at read time as an O(N) positional count (`COUNT(*) WHERE s.id <= dv.id`) because `dataset_versions` has no stored sequence column (unlike `prompt_versions`, which has `version_number`). For `findByTag` (single row) this is fine, and `COUNT(*)` is correct here — it stays on the `(workspace_id, dataset_id, id)` covering index, whereas `COUNT(<col>)` could defeat that.

As the paginated sibling methods get the same treatment (tracked in the ticket), the per-row count becomes O(versions)×page-size. Have you considered persisting a monotonic per-dataset sequence column so `version_name` becomes a plain column read? That's a schema change outside this PR's scope, but it may be a cleaner long-term answer for the paginated paths than a set-based `seq_num` window function.

JetoPistola:
👋 **Review summary**

**What looks good**
- Sharp root-cause: correctly identified that the `SQLSyntaxErrorException: Table '#sql...' doesn't exist` is MySQL TempTable pool saturation from multi-reference materialized CTEs, not a syntax bug — and the fix (flatten CTE → correlated covering-index subqueries) targets the actual mechanism.
- Verification is thorough and honest: `EXPLAIN ANALYZE` before/after showing zero `Materialize`/`Temporary table` nodes, plus result-identity diffs on `id`/`version_name`/`environments`/`tags`/`is_latest`.
- Equivalence holds under scrutiny: `ROW_NUMBER() ORDER BY id` ↔ `COUNT(*) WHERE id <= dv.id` is exact (unique PK, no ties); `environments`/`tags` are `Set`-typed so JSON array order is correctly noted as irrelevant; the `find` env subquery matches the existing `findSingleVersion` pattern already in the file.
- `EXISTS(SELECT 1 ...)` for `is_latest` aligns with the codebase's own existence-check idiom (`AlertDAO`/`DatasetDAO`/`PromptDAO`).
- Scope discipline: the remaining double-referenced-CTE methods are explicitly deferred with the right caveat (paginated paths need a *set-based* `seq_num`, not a per-row `COUNT`).

**Overall**
Solid, well-scoped optimization with no behavior change and strong evidence. Nothing blocking. The one structural thought (raised inline): this CTE-materialization failure has recurred in the same `DatasetVersionDAO` (OPIK 6735 — O(n) CTE → RDS storage exhaustion) and is being addressed elsewhere (OPIK 7196/7193) — worth a durable plan-shape regression guard rather than fixing case-by-case. Filed as follow-up **OPIK 7198**.

**Inline comments**: 2 questions (testing/plan-guard + a `version_name` design thought) — nothing blocking.

apps/opik-backend/src/main/java/com/comet/opik/domain/PromptVersionDAO.java
```diff
@@ -134,29 +134,24 @@ default void closeEnvOwnershipsForPrompt(UUID promptId, String workspaceId, Set<
...
-            SELECT pv.*, ve.environments
-            FROM paginated pv
-            LEFT JOIN ver_envs ve ON ve.version_id = pv.id
+            SELECT pv.*,
+                p.template_structure,
+                (
+                    SELECT JSON_ARRAYAGG(pve.environment)
+                    FROM prompt_version_envs pve
+                    WHERE pve.version_id = pv.id
+                        AND pve.workspace_id = :workspace_id
```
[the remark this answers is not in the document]
thiagohora: Fair point. This correlated `prompt_version_envs` aggregation intentionally mirrors the existing `findSingleVersion` subquery in this same file, so it follows the established pattern rather than introducing a new one — and the two are functionally identical (WHERE predicate order and `:workspace_id` vs `pv.workspace_id` are equivalent; the optimizer reorders predicates regardless). Kept inline so each `@SqlQuery` stays self-contained/readable rather than concatenating a shared string constant into both annotations. Happy to extract into a single shared fragment if the team prefers one source of truth — leaving inline for now as it is a small, stable snippet.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetVersionDAO.java
thiagohora: Two things on this thread:

**Failing CI checks** — `evaluation/modelGateway.test.ts` and `metrics/llmJudges.test.ts` exercise external LLM-provider evaluation and touch none of the paths changed here (`PromptVersionDAO.find` / `DatasetVersionDAO.findByTag`); unrelated to this PR, re-run in progress.

**On a plan-shape regression test** — I would actually lean against adding one. Asserting that `EXPLAIN FORMAT=JSON` contains no `Materialize` / `Temporary table` node is testing the MySQL optimizer's behavior, which is brittle across engine versions, table statistics and data distribution — it verifies the database engine, not our code, and would throw false failures on unrelated changes. The right record is the *manual* verification: the exact `EXPLAIN ANALYZE` commands and the expected "no Materialize/Temporary" outcome, documented in the ticket. Suggest repurposing OPIK-7198 to capture that manual-verification note rather than build a CI assertion that proves how the database plans a query.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetVersionDAO.java
thiagohora: Good point, and thanks for confirming `COUNT(*)` stays on the covering index here. A stored monotonic per-dataset sequence column would turn `version_name` into a plain column read and is likely the cleaner long-term answer for the paginated siblings (vs a per-row `COUNT` or a set-based window). Out of scope for this PR (schema change) — will capture it as an option in the follow-up alongside the set-based `seq_num` note.

BorisTkachenko (approved):
LGTM

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetVersionDAO.java
JetoPistola: Agreed on both — thanks for pushing back.

On the plan-shape assertion: you're right, and the sharpest version of the problem is that on a near-empty test DB the optimizer may not materialize anything even for a genuinely bad double-referenced-CTE query — so the assertion could *pass on the exact anti-pattern it's meant to catch*. It verifies the engine's plan choice, not our SQL. Dropping it.

Where I'd keep a small automated guard: the deterministic signal isn't the plan, it's the *code smell* — a `WITH` CTE referenced ≥2× in a single statement. A static lint on that catches the pattern directly, independent of engine/stats/data (with the caveat that it's fragile against StringTemplate-composed SQL, so a heuristic rather than a hard gate). And the manual `EXPLAIN ANALYZE` verification stays as the durable evidence, exactly as you did here.

I've repurposed **OPIK 7198** accordingly: dropped the plan-shape CI assertion, kept the static-lint + manual-verification note.

PR #7329, opened 2026-07-02
[OPIK-7016] [FE] fix: align Cost Intelligence navigation with entitlements

andriidudar:
## Details

Aligns the Cost Intelligence navigation with organization entitlements instead of organization-admin role. Previously the Cost Intelligence menu entry and AI-Spend workspace navigation were gated on `isOrganizationAdmin`, so non-admin members of a Cost-Intelligence-enabled workspace could not reach it. This relaxes the gate to the entitlement (v2 + Cost-Intelligence-enabled org) and propagates the admin flag through context for downstream, role-aware UI.

- Remove the `isOrganizationAdmin` restriction from the access gate in `useAiSpendManager`.
- Propagate `isOrganizationAdmin` through `AiSpendContext` and the layout providers so pages can adapt per role.
- Drop the unnecessary `/home` suffix from the AI-Spend workspace navigation path.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-7016

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Scope: assisted
- Human verification: code review + manual testing on an internal test environment with both admin and non-admin users

## Testing

Verified on an internal test environment with both an admin and a non-admin user: the Cost Intelligence menu entry and workspace navigation now appear for non-admin members of a Cost-Intelligence-enabled (v2) workspace, while remaining hidden where the entitlement is absent. Admin behavior is unchanged.

## Documentation

N/A

### DIFF
--- a/apps/opik-frontend/src/contexts/AiSpendContext.tsx
+++ b/apps/opik-frontend/src/contexts/AiSpendContext.tsx
@@ -3,6 +3,7 @@ import React, { createContext, useContext, ReactNode } from "react";
 export type AiSpendContextValue = {
   isPending: boolean;
   hasAccess: boolean;
+  isOrganizationAdmin?: boolean;
   spendWorkspaceName?: string;
   isSpendWorkspaceActive?: boolean;
   organizationName?: string;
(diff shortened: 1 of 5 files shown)


aadereiko (approved):
Nice work!

PR #7337, opened 2026-07-03
[OPIK-7230] [BE] Optimize experiment-refs ClickHouse lookups + fix missing log_comment

thiagohora:
## Details

Experiment-refs lookups (`ExperimentItemDAO.getExperimentRefsBy{Trace,Span,Item}Ids`) drive a large share of ClickHouse index-evaluation CPU because their `WHERE` filters are not a clean prefix of the target table sort key, forcing `IndexGenericExclusionSearchAlgorithm` over millions of marks. This PR makes those lookups prune properly and fixes their query attribution.

- Route `getExperimentRefsByIds` through `getSTWithLogComment` so the three `GET_EXPERIMENT_REFS_BY_*` queries emit a real `log_comment` (`query_name:workspace_id:user_name:details`) instead of the literal `<log_comment>` placeholder.
- Add a conditional `project_id` filter to the trace- and span-refs queries. `projectId` is threaded through the aggregation events and **split per project** in `ExperimentAggregateEventListener`, so every query targets a single project and can prune on `project_id` (existing `experiment_items` minmax index + `spans` primary-key prefix). Paths that can't safely resolve a single project omit the filter and fall back to the new skip index.
- Add `minmax GRANULARITY 1` skip indexes on `experiment_items.trace_id` and `spans.id` (migration `000095`) so these equality lookups prune granules. minmax is chosen over bloom_filter: the keys are UUIDv7 (time-ordered, well clustered) and minmax also serves id/trace_id range scans; projections were ruled out because these queries use `FINAL`.

## Benchmark (local, real data replicated to scale)

Real experiment_items/spans values were copied from a staging workspace into a local ClickHouse (25.8), replicated to a few million rows, then the exact production query was measured with the skip index dropped vs materialized (CH-side metrics from `system.query_log`, best of 3 runs). `input`/`output`/`metadata` blob widths were kept so adaptive granule sizing is realistic.

**Trace query — `experiment_items.trace_id` (7.1M rows)**

| | read_rows | marks selected | OS CPU (ms) | user CPU (ms) | wall (ms) |
|---|--:|--:|--:|--:|--:|
| before (no index) | 1,712,128 | 176 | ~424 | ~310 | ~48 |
| after (minmax) | 344,064 | 30 | ~70 | ~60 | ~12 |
| reduction | 5.0× | 5.9× | ~6× | ~5× | ~4× |

**Span query — `spans.id` (2.5M rows, with input/output/metadata blobs)**

| | read_rows | marks selected | OS CPU (ms) | user CPU (ms) | wall (ms) |
|---|--:|--:|--:|--:|--:|
| before (no index) | 2,525,088 | 335 | ~208 | ~110 | ~28 |
| after (minmax) | 369,324 | 37 | ~53 | ~48 | ~10 |
| reduction | 6.8× | 9.1× | ~4× | ~2× | ~2.8× |

Notes:
- The index-condition evaluation itself is cheap and unchanged (`IndexGenericExclusionSearchAlgorithm` = 4 / 1, `SelectedMarksTotal` constant before/after). The CPU win comes from the skip index selecting far fewer granules, so ~5–9× fewer rows are read/decompressed — that scan is what the CPU tracked. i.e. "index_eval too high" was poor mark selection driving downstream scan CPU, not expensive index arithmetic.
- `read_rows` and `marks selected` are deterministic; wall-clock and user-CPU are noisier on a single-node local container (warmup/cache), so OS CPU / marks are the reliable signal.
- These are conservative floors: replication spreads each id/trace_id across multiple full-range parts, whereas in production a given id lives in one time-bounded part, so the index prunes harder there (the original prod `find_traces` signal was a generic-exclusion scan over ~26.8M marks).

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- OPIK-7230

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: Full change (DAO query edits, event/listener plumbing, migration, tests) drafted with AI assistance.
- Human verification: Diagnosis grounded in production `system.query_log`/ProfileEvents analysis; index choice validated against measured per-granule clustering and the local before/after benchmark above; change compiled and unit-tested.

## Testing

- `mvn test-compile` (main + tests) — BUILD SUCCESS.
- `mvn test -Dtest=ExperimentAggregateEventListenerTest` — **58 tests pass**, including new cases asserting: traces/spans-created split trace_ids per project, traces-updated split via the `traceId -> projectId` mapping, traces/spans-deleted pass the event `projectId`, and span feedback-score passes `projectId` to `getExperimentRefsBySpanIds`.
- Local before/after index benchmark on real (replicated) data — see the Benchmark section.
- `mvn spotless:apply` — clean.
- Not run locally: the `000095` skip-index migration against a real cluster (`ADD INDEX` is metadata-only; `MATERIALIZE INDEX` runs as a background mutation — `spans` materialization is heavy and expected to complete asynchronously; new parts get the index automatically).

## Documentation

No documentation changes.


apps/opik-backend/src/main/java/com/comet/opik/domain/FeedbackScoreService.java
```diff
@@ -254,9 +260,13 @@ public Mono<Void> deleteTraceScore(UUID id, DeleteFeedbackScore score) {
...
-            return dao.deleteScoreFrom(EntityType.TRACE, id, score)
-                    .doOnSuccess(__ -> eventBus.post(
-                            new FeedbackScoresDeleted(Set.of(id), EntityType.TRACE, workspaceId, userName)));
+            return traceDAO.getProjectIdFromTrace(id)
+                    .map(Optional::of)
+                    .defaultIfEmpty(Optional.empty())
+                    .flatMap(projectId -> dao.deleteScoreFrom(EntityType.TRACE, id, score)
+                            .doOnSuccess(__ -> eventBus.post(
+                                    new FeedbackScoresDeleted(Set.of(id), EntityType.TRACE, workspaceId, userName,
+                                            projectId.orElse(null)))));
```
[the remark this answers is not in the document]
thiagohora: Fixed in 5c9206aa1c: the project lookup is now wrapped in `onErrorResume` inside a shared `deleteScoreAndNotify` helper, so a transient lookup failure logs and continues with a `null` projectId — `dao.deleteScoreFrom(...)` still runs and the listener falls back to the trace_id/id skip index.

apps/opik-backend/src/main/java/com/comet/opik/domain/FeedbackScoreService.java
```diff
@@ -242,9 +244,13 @@ public Mono<Void> deleteSpanScore(UUID id, DeleteFeedbackScore score) {
...
-                            new FeedbackScoresDeleted(Set.of(id), EntityType.SPAN, workspaceId, userName)));
+            return spanDAO.getProjectIdFromSpan(id)
+                    .map(Optional::of)
+                    .defaultIfEmpty(Optional.empty())
+                    .flatMap(projectId -> dao.deleteScoreFrom(EntityType.SPAN, id, score)
+                            .doOnSuccess(__ -> eventBus.post(
+                                    new FeedbackScoresDeleted(Set.of(id), EntityType.SPAN, workspaceId, userName,
+                                            projectId.orElse(null)))));
         });
     }
```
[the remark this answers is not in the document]
thiagohora: Done in 5c9206aa1c: extracted `deleteScoreAndNotify(EntityType, UUID, DeleteFeedbackScore)`; both `deleteSpanScore` and `deleteTraceScore` delegate to it. The project is resolved lazily inside the deferred chain via a `switch` (no eager `Mono`-as-parameter).

apps/opik-backend/src/main/java/com/comet/opik/api/events/FeedbackScoresDeleted.java
```diff
@@ -1,24 +1,21 @@
...
-        super(workspaceId, userName);
-        this.entityIds = entityIds;
-        this.entityType = entityType;
+        this(entityIds, entityType, workspaceId, userName, null);
+    }
+
+    public FeedbackScoresDeleted(@NonNull Set<UUID> entityIds, @NonNull EntityType entityType,
+            @NonNull String workspaceId, @NonNull String userName, @Nullable UUID projectId) {
+        super(entityIds, entityType, workspaceId, userName, projectId);
     }
```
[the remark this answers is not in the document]
thiagohora: Addressed: `deleteSpanScore`/`deleteTraceScore` now resolve the entity's project (`getProjectIdFrom{Span,Trace}`) and pass it into `FeedbackScoresDeleted` (via `deleteScoreAndNotify`, 5c9206aa1c). Best-effort — empty/failed lookup falls back to null + skip index.

apps/opik-backend/src/main/java/com/comet/opik/api/events/EntityProjectEvent.java
```diff
@@ -0,0 +1,33 @@
...
+ * {@code projectId} is optional: callers that can resolve a single project set it so downstream consumers can
+ * prune by project; callers that can't (multi-project batches, paths without it) leave it {@code null}.
+ */
+@Getter
+@Accessors(fluent = true)
+public abstract class EntityProjectEvent extends BaseEvent {
+
+    private final @NonNull Set<UUID> entityIds;
+    private final @NonNull EntityType entityType;
+    private final @Nullable UUID projectId;
```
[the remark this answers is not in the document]
thiagohora: `update` now resolves the comment entity's project (shared `resolveProjectId` switch) and threads it into `CommentsUpdated`. `delete` takes a `BatchDelete` whose entities can span multiple projects, so a single `projectId` isn't valid there — it stays null and the listener falls back to the trace_id/id skip index (correct results, workspace-scoped scan only on that path).

apps/opik-backend/src/main/java/com/comet/opik/domain/CommentService.java
```diff
@@ -77,6 +76,26 @@ public Mono<Comment> get(@NonNull UUID entityId, @NonNull UUID commentId) {
...
+            case TRACE -> traceDAO.getProjectIdFromTrace(entityId);
+            case SPAN -> spanDAO.getProjectIdFromSpan(entityId);
+            case THREAD -> traceThreadDAO.getProjectIdFromThread(entityId);
+        };
+    }
+
+    // Resolves the (single) project for a comment's entity refs so CommentsUpdated can prune by project;
+    // empty when there are no refs or the entity's project can't be resolved.
+    private Mono<Optional<UUID>> resolveProjectIdForRefs(List<CommentEntityRef> refs) {
+        if (CollectionUtils.isEmpty(refs)) {
```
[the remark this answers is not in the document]
thiagohora: Partly covered: `ExperimentAggregateEventListenerTest` now asserts the exact `projectId` threaded from a `CommentsUpdated` event (`eq(projectId)`), and the `defaultIfEmpty`/null path via `isNull()`. The producer-side assertion (that `CommentService.update` resolves + emits the right projectId) isn't unit-tested — I can add a focused `CommentService` unit test for the resolved and unresolved cases if you'd like it in this PR.

apps/opik-backend/src/main/java/com/comet/opik/domain/CommentService.java
```diff
@@ -85,17 +104,18 @@ public Mono<Void> update(@NonNull UUID commentId, @NonNull Comment comment) {
...
-                                    eventBus.post(new CommentsUpdated(
-                                            Set.of(ref.entityId()), toEntityType(ref.entityType()),
-                                            workspaceId, userName));
-                                }
-                            }));
+                    .flatMap(refs -> resolveProjectIdForRefs(refs)
+                            .flatMap(projectId -> commentDAO.findById(null, commentId)
+                                    .switchIfEmpty(Mono.error(failWithNotFound("Comment", commentId)))
+                                    .then(Mono.defer(() -> commentDAO.updateComment(commentId, comment)))
+                                    .doOnSuccess(__ -> {
```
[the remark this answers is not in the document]
thiagohora: The listener tests were updated to assert the threaded `projectId` (`eq(projectId)`, and `isNull()` for SpansUpdated) across the feedback/comment handlers instead of `any()`. Dedicated `CommentService`/`FeedbackScoreService` unit tests asserting the emitted event payload (incl. null-on-unresolved) would close the producer side — happy to add as a follow-up or here.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/ExperimentAggregateEventListener.java
```diff
@@ -123,53 +125,83 @@ public void onExperimentsDeleted(@NonNull ExperimentsDeleted event) {
...
     public void onTracesUpdated(TracesUpdated event) {
-        triggerByTraceIds(event.traceIds(), event.workspaceId(), event.userName())
-                .subscribe(null, e -> log.error("Error triggering aggregation for traces updated", e));
+        var errorMessage = "Error triggering aggregation for traces updated";
+        // Split the trace ids by project so each aggregation trigger targets a single project.
+        if (MapUtils.isNotEmpty(event.traceIdToProjectId())) {
+            Map<UUID, Set<UUID>> traceIdsByProject = event.traceIdToProjectId().entrySet().stream()
+                    .collect(Collectors.groupingBy(Map.Entry::getValue,
+                            Collectors.mapping(Map.Entry::getKey, Collectors.toSet())));
+            triggerByTraceIdsPerProject(traceIdsByProject, event.workspaceId(), event.userName(), errorMessage);
```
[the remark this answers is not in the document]
thiagohora: Fixed in 7b67e94bb1: `onTracesUpdated` now splits the mapped trace ids per project **and** collects the ids in `event.traceIds()` not present in `traceIdToProjectId`, routing those through the workspace-scoped fallback (`triggerByTraceIds(uncovered, ..., null)`) instead of dropping them. Added a listener test (`routesTraceIdsNotCoveredByMappingThroughFallbackOnTracesUpdated`) asserting the mapped id triggers with its project and the uncovered id triggers with a null project.

apps/opik-backend/src/main/java/com/comet/opik/domain/FeedbackScoreService.java
```diff
@@ -238,28 +240,44 @@ private <T extends FeedbackScoreItem> List<ProjectDto<T>> mergeProjectsAndScores
...
 
-            return dao.deleteScoreFrom(EntityType.TRACE, id, score)
-                    .doOnSuccess(__ -> eventBus.post(
-                            new FeedbackScoresDeleted(Set.of(id), EntityType.TRACE, workspaceId, userName)));
+            return resolveProjectId(entityType, entityId)
+                    .map(Optional::of)
+                    .defaultIfEmpty(Optional.empty())
+                    .onErrorResume(e -> {
+                        log.warn("Failed to resolve projectId for {} '{}' before score delete; "
+                                + "continuing without project scope", entityType, entityId, e);
```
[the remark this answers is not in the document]
thiagohora: Fixed in the latest commit: both placeholders are now quoted — `"Failed to resolve projectId for '{}' '{}' before score delete; continuing without project scope"` (entityType + entityId), matching the backend log-format convention.

apps/opik-backend/src/main/java/com/comet/opik/domain/FeedbackScoreService.java
```diff
@@ -238,28 +240,44 @@ private <T extends FeedbackScoreItem> List<ProjectDto<T>> mergeProjectsAndScores
...
+                            .doOnSuccess(__ -> eventBus.post(new FeedbackScoresDeleted(Set.of(entityId), entityType,
+                                    workspaceId, userName, projectId.orElse(null)))));
         });
     }
 
+    private Mono<UUID> resolveProjectId(EntityType entityType, UUID entityId) {
+        return switch (entityType) {
+            case SPAN -> spanDAO.getProjectIdFromSpan(entityId);
+            case TRACE -> traceDAO.getProjectIdFromTrace(entityId);
+            default -> Mono.empty();
```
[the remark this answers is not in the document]
thiagohora: Leaving these as separate switches: they operate on **different enum types** — `FeedbackScoreService` switches on `com.comet.opik.domain.EntityType` (TRACE/SPAN only), while `CommentService` switches on `CommentDAO.EntityType` (TRACE/SPAN/THREAD, using `traceThreadDAO`). A shared helper would have to take all three DAOs and reconcile the two enums for a 3-line switch, which adds more coupling than it removes. Happy to extract a shared `EntityProjectResolver` if you'd prefer it, but it seemed like net-negative here.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000097_add_minmax_indexes_experiment_refs_lookup.sql
```diff
@@ -0,0 +1,23 @@
+--liquibase formatted sql
+--changeset thiagoh:000097_add_minmax_index_experiment_items_trace_id
```
JetoPistola: 🧹 **nit** | Style

Every other changeset in this migrations directory uses author `thiagohora` (e.g. `000080`–`000088`), but this one uses `thiagoh`. Liquibase changeset identity is `author:id:filename`, so the mismatch is harmless functionally but breaks the convention and makes the changeset log inconsistent. Suggest aligning both changesets to `thiagohora`:

```suggestion
--changeset thiagohora:000097_add_minmax_index_experiment_items_trace_id
```

(and likewise for the `000097_add_minmax_index_spans_id` changeset below)

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/ExperimentAggregateEventListener.java
```diff
@@ -123,53 +125,96 @@ public void onExperimentsDeleted(@NonNull ExperimentsDeleted event) {
...
+        triggerByTraceIds(event.traceIds(), event.workspaceId(), event.userName(), event.projectId())
                 .subscribe(null, e -> log.error("Error triggering aggregation for spans deleted", e));
     }
 
+    private void triggerByTraceIdsPerProject(Map<UUID, Set<UUID>> traceIdsByProject, String workspaceId,
+            String userName, String errorMessage) {
+        Flux.fromIterable(traceIdsByProject.entrySet())
+                .concatMap(entry -> triggerByTraceIds(entry.getValue(), workspaceId, userName, entry.getKey())
+                        .onErrorResume(e -> logAndContinue(errorMessage, e)))
+                .subscribe(null, e -> log.error(errorMessage, e));
```
JetoPistola: 💡 **suggestion** | Readability

Since each per-project inner call already handles its own failure via `.onErrorResume(e -> logAndContinue(errorMessage, e))` (returning `Mono.empty()`), the `concatMap` stream can't surface those errors downstream. This terminal `.subscribe(null, e -> log.error(errorMessage, e))` error handler is therefore effectively unreachable for the per-project failures — it would only ever fire on an error from `Flux.fromIterable` iteration itself, which won't happen here. The result is that a real per-project failure gets logged once (fine), but the redundant terminal handler suggests double-logging that never actually occurs.

Not a bug — just slightly misleading. You could either drop the per-entry `onErrorResume` and let the terminal handler log (loses the per-project continuation) or keep `onErrorResume` and simplify the terminal to `.subscribe()`. The current continuation semantics are correct; only the redundancy is worth a second look.

JetoPistola:
👋 **Review summary**

**What looks good**
- Genuinely well-diagnosed optimization — the problem statement (WHERE filters not being a clean sort-key prefix → `IndexGenericExclusionSearchAlgorithm` over millions of marks) is precise, and the fix is verified sound against the actual table layouts: `spans` `ORDER BY (workspace_id, project_id, trace_id, parent_span_id, id)` and `experiment_items` `ORDER BY (workspace_id, experiment_id, dataset_item_id, trace_id, id)` both confirm that id/trace_id lookups can't prefix-prune, so the minmax skip indexes + per-project pushdown are the right levers.
- The `<if(project_id)>` template + `:project_id` bound parameter approach is safe — no string interpolation of the id into SQL, and the value only drives the StringTemplate conditional.
- Careful correctness handling of the partial-coverage case: `onTracesUpdated` splits mapped trace ids per project *and* routes uncovered ids through the workspace-scoped fallback so aggregation is never silently dropped. That edge case is explicitly tested.
- Excellent, targeted test coverage — the new `ProjectSplittingAndThreading` nested class asserts per-project splitting, the mapping fallback, event-projectId threading, and the `null` skip-index backstop for `SpansUpdated`.
- Nice touch introducing `EntityProjectEvent` to consolidate the `entityIds`/`entityType`/`projectId` triple across the event classes rather than copy-pasting the projectId field into each.

**Overall**
Solid, well-scoped performance PR with honest benchmarking and a clear rationale for minmax over bloom_filter/projections. The approach makes sense and the risk is contained (skip index is metadata + async materialization; project pushdown is conditional with a safe fallback). Nothing structural to flag.

**One doc nit:** the Details section references migration **`000095`** twice, but the actual file is **`000097`** (`000096` is the last existing migration). Worth correcting so ops/reviewers aren't chasing the wrong changeset.

**Inline comments**: 1 nit (changeset author `thiagoh` vs. the dir-wide `thiagohora`) and 1 suggestion (redundant terminal error handler in `triggerByTraceIdsPerProject`) — nothing blocking.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000097_add_minmax_indexes_experiment_refs_lookup.sql
thiagohora: Good catch — aligned both changesets to `thiagohora` to match the directory convention (000080–000088). Since the migration hasn't run anywhere yet, changing the author id is safe.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/ExperimentAggregateEventListener.java
thiagohora: Agreed, that terminal error consumer was dead for per-project failures. Kept the per-entry `onErrorResume` (preserves per-project continuation) and simplified the terminal to `.subscribe()`, plus a short comment noting no error reaches the subscriber.

apps/opik-backend/src/main/java/com/comet/opik/domain/FeedbackScoreService.java
```diff
@@ -238,28 +240,44 @@ private <T extends FeedbackScoreItem> List<ProjectDto<T>> mergeProjectsAndScores
...
 
     @Override
     public Mono<Void> deleteTraceScore(UUID id, DeleteFeedbackScore score) {
+        return deleteScoreAndNotify(EntityType.TRACE, id, score);
+    }
+
+    // Deletes the score and emits FeedbackScoresDeleted carrying the entity's project (for per-project pruning).
+    // The project is resolved lazily at subscription time and best-effort: if it is empty or the lookup fails,
+    // the delete still runs with a null projectId (the listener falls back to the trace_id/id skip index).
+    private Mono<Void> deleteScoreAndNotify(EntityType entityType, UUID entityId, DeleteFeedbackScore score) {
```
andrescrz: Nit: these better as javadoc.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000097_add_minmax_indexes_experiment_refs_lookup.sql
andrescrz, line 1: Let's be careful with these indexes and the materialisation during the rollout, after the upgrade to ClickHouse 25.8 LTS.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000097_add_minmax_indexes_experiment_refs_lookup.sql
```diff
@@ -0,0 +1,23 @@
...
+--rollback ALTER TABLE ${ANALYTICS_DB_DATABASE_NAME}.experiment_items ON CLUSTER '{cluster}' DROP INDEX IF EXISTS idx_experiment_items_trace_id;
+
+--changeset thiagohora:000097_add_minmax_index_spans_id
+--comment: Add minmax skip index on spans.id so span-id lookups prune granules instead of a generic-exclusion scan over the workspace (OPIK-7230)
+
+ALTER TABLE ${ANALYTICS_DB_DATABASE_NAME}.spans ON CLUSTER '{cluster}'
+    ADD INDEX IF NOT EXISTS idx_spans_id id TYPE minmax GRANULARITY 1;
+
+ALTER TABLE ${ANALYTICS_DB_DATABASE_NAME}.spans ON CLUSTER '{cluster}'
+    MATERIALIZE INDEX idx_spans_id;
```
andrescrz, line 20: Both new indexes refer to fields that are already part of the primary key. Are the improvements really worthy? Do they provide a real benefit? Great if you could provide a query plan proving the expected results.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000097_add_minmax_indexes_experiment_refs_lookup.sql
```diff
@@ -0,0 +1,23 @@
...
+
+--rollback ALTER TABLE ${ANALYTICS_DB_DATABASE_NAME}.spans ON CLUSTER '{cluster}' DROP INDEX IF EXISTS idx_spans_id;
```
andrescrz, line 22: Please add the rollback statement for the other index as well.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/ExperimentAggregateEventListenerTest.java
```diff
@@ -589,4 +605,149 @@ void handlesCommentsDeleted(EntityType entityType, boolean configEnabled) {
...
+            var t3 = traceIn(projectB);
+
+            listener.onTracesCreated(new TracesCreated(List.of(t1, t2, t3), WORKSPACE_ID, USER_NAME));
+
+            await().atMost(AWAIT_TIMEOUT_SECONDS, TimeUnit.SECONDS).untilAsserted(() -> {
+                verify(experimentItemService)
+                        .getExperimentRefsByTraceIds(eq(Set.of(t1.id(), t2.id())), any(), eq(projectA));
+                verify(experimentItemService)
+                        .getExperimentRefsByTraceIds(eq(Set.of(t3.id())), any(), eq(projectB));
+                verify(experimentItemService, times(2)).getExperimentRefsByTraceIds(any(), any(), any());
```
andrescrz, line 652: Minor: for unit test assertion and verify calls, use as real arguments as possible, and attempt to eliminate `any` or similar, so tests are more endurable and precise.

andrescrz (commented):
A few important comments and the rest are minor. My only blocker is the rollback statement in the migration.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000097_add_minmax_indexes_experiment_refs_lookup.sql
thiagohora: Yeah, I added the CPU profile number because of that. I'm going to follow up closely

apps/opik-backend/src/main/java/com/comet/opik/domain/FeedbackScoreService.java
thiagohora: Done — converted to a `/** … */` javadoc on `deleteScoreAndNotify` in the latest commit.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000097_add_minmax_indexes_experiment_refs_lookup.sql
thiagohora: Both changesets already carry a `--rollback` (`idx_experiment_items_trace_id` and `idx_spans_id`) — you may have been looking at an earlier revision during the churn. Confirmed present for both at the current HEAD.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000097_add_minmax_indexes_experiment_refs_lookup.sql
thiagohora: Agreed. Both statements are idempotent (`ADD INDEX IF NOT EXISTS`), and `ADD INDEX` itself is metadata-only/instant — only new parts get the index automatically. The heavy part is `MATERIALIZE INDEX`, which runs as an async background mutation over existing parts (`spans` especially). If you'd prefer to control that during rollout, I can drop the `MATERIALIZE` from the migration and run it manually in a maintenance window (new parts still get pruned meanwhile). Let me know which you prefer.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/ExperimentAggregateEventListenerTest.java
thiagohora: Tightened — replaced the `any()` status matcher with `eq(FINISHED_STATUSES)` across the stubs/verifies (30 call sites), and the projectId arg already uses `eq(projectId)`/`isNull()`. The remaining `any()` are only in `verify(..., never())` calls where the args are irrelevant.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000097_add_minmax_indexes_experiment_refs_lookup.sql
thiagohora: Good question — and yes, precisely *because* both fields are in the primary key but **not as a usable prefix**, the PK's generic-exclusion search can't prune them; the minmax skip index is what does. `EXPLAIN indexes = 1` on the real (staging-derived, replicated) data proves it:

**`experiment_items` — filter `workspace_id = X AND trace_id = T`** (trace_id is sort-key position 4)
```
Without index:
  PrimaryKey  Granules: 175/868   Search Algorithm: generic exclusion search
With minmax idx_experiment_items_trace_id:
  PrimaryKey  Granules: 175/868   Search Algorithm: generic exclusion search
  Skip        Granules:   6/175      <-- 175 -> 6
```

**`spans` — filter `workspace_id = X AND id = S`** (id is trailing sort-key position 5)
```
Without index:
  PrimaryKey  Granules: 335/335   Search Algorithm: generic exclusion search   <-- PK prunes nothing
With minmax idx_spans_id:
  PrimaryKey  Granules: 335/335
  Skip        Granules:  34/335      <-- 335 -> 34
```

So the PK narrows `experiment_items` only to 175/868 (workspace_id prefix only) and `spans` to 335/335 (no pruning at all — `id` is the trailing key). The skip index then prunes to 6 and 34 granules respectively.

That translates to the measured read/CPU reduction (best of 3, from `system.query_log`):

| Query | read_rows | marks selected | OS CPU |
|---|--|--|--|
| trace (`experiment_items.trace_id`) | 1,712,128 → 344,064 (5.0×) | 176 → 30 | ~424 → ~70 ms (~6×) |
| span (`spans.id`) | 2,525,088 → 369,324 (6.8×) | 335 → 37 | ~208 → ~53 ms (~4×) |

Note the index-condition evaluation itself is cheap/unchanged (`IndexGenericExclusionSearchAlgorithm` = 4/1, `SelectedMarksTotal` constant) — the win is fewer selected granules → fewer rows read/decompressed. These are conservative floors: replication spreads each id across multiple full-range parts; in prod a given id lives in one time-bounded part, so pruning is sharper (the original `find_traces` prod signal was a generic-exclusion scan over ~26.8M marks).

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000097_add_minmax_indexes_experiment_refs_lookup.sql
andrescrz: This is good enough, thanks.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000097_add_minmax_indexes_experiment_refs_lookup.sql
andrescrz: I missed the statement above. Resolving this.

andrescrz (approved):

PR #7340, opened 2026-07-05
fix(docker): support RPi CM4 compatibility by parameterizing platform

### LINKED ISSUE #2101: [Bug]: Docker compose failing on RPi CM4

- [ ] Python SDK
- [ ] Opik UI
- [x] Opik Server
- [ ] Documentation

### Opik version

- Opik version: latest

### Describe the problem

Trying to stand up self hosted version following instructions 1. Git clone, 2. cd opik, 3. sudo ./opik.sh and getting the following logs error:

```bash
🔍 Checking container status and starting missing ones...
🔄 Starting missing containers...
ERROR: The Compose file '/home/pi/opik/deployment/docker-compose/docker-compose.yaml' is invalid because:
'name' does not match any of the regexes: '^x-'

You might be seeing this error because you're using the wrong Compose file version. Either specify a supported version (e.g "2.2" or "3.3") and place your service definitions under the `services` key, or omit the `version` key and place your service definitions at the root of the file to use version 1.
For more on the Compose file format versions, see https://docs.docker.com/compose/compose-file/
services.backend.environment.OPIK_OTEL_SDK_ENABLED contains false, which is an invalid type, it should be a string, number, or a null
⏳ Waiting for all containers to be running and healthy...
❌ opik-clickhouse-1 failed to start (status: )
❌ opik-mysql-1 failed to start (status: )
❌ opik-python-backend-1 failed to start (status: )
❌ opik-redis-1 failed to start (status: )
❌ opik-frontend-1 failed to start (status: )
❌ opik-backend-1 failed to start (status: )
❌ opik-minio-1 failed to start (status: )
❌ opik-zookeeper-1 failed to start (status: )
🔄 Re-checking container status...
❌ opik-clickhouse-1 is not running (status=)
❌ opik-mysql-1 is not running (status=)
❌ opik-python-backend-1 is not running (status=)
❌ opik-redis-1 is not running (status=)
❌ opik-frontend-1 is not running (status=)
❌ opik-backend-1 is not running (status=)
❌ opik-minio-1 is not running (status=)
❌ opik-zookeeper-1 is not running (status=)
⚠️  Some containers are still not healthy. Please check manually using './opik.sh --verify'
```

Currently running:
```bash
pi@gretchen:~/opik $ sudo docker -v
Docker version 20.10.24+dfsg1, build 297e128

pi@gretchen:~/opik $ sudo docker-compose -v
docker-compose version 1.29.2, build unknown
```

I had to manually modify opik.sh to work with the `docker-compose` as opposed to `docker compose`:

```bash
get_docker_compose_cmd() {
 # local cmd="docker compose -f $script_dir/deployment/docker-compose/docker-compose.yaml"
  local cmd="sudo docker-compose -f $script_dir/deployment/docker-compose/docker-compose.yaml"
  if [[ "$PORT_MAPPING" == "true" ]]; then
    cmd="$cmd -f $script_dir/deployment/docker-compose/docker-compose.override.yaml"
  fi
  echo "$cmd"
}
```

_No response_

### Error logs or stack trace

_No response_

### Healthcheck results

_No response_

namdamdoi68-oss:
### Summary
This PR resolves issue #2101 where running the default Docker compose setup crashes on older ARMv8-A hardware (such as Raspberry Pi 4 or Raspberry Pi CM4) because ClickHouse and backend images require `ARMv8.2-A` instruction set.

### Solution
Instead of hardcoding any platforms or custom Dockerfiles, this PR parameterizes the `platform` option under `clickhouse`, `backend`, and `python-backend` services inside `docker-compose.yaml`.
* By default, it resolves to `""` (native architecture), ensuring zero disruption and 100% native performance for existing x86_64, standard arm64, and Apple Silicon users.
* Raspberry Pi 4 / CM4 users can now run in emulation mode (using QEMU) easily by setting the corresponding platform environment variables:
  ```bash
  export OPIK_CLICKHOUSE_PLATFORM=linux/amd64
  export OPIK_BACKEND_PLATFORM=linux/amd64
  export OPIK_PYTHON_BACKEND_PLATFORM=linux/amd64
  ./opik.sh
  ```

Closes #2101

### DIFF
--- a/.specify/plan.md
+++ b/.specify/plan.md
@@ -1,0 +1,24 @@
+# Plan: Support RPi4/CM4 Emulation Compatibility for Opik
+
+## Architectural Design
+Expose the Docker Compose `platform` property for all architecture-sensitive services using interpolation variables:
+* `${OPIK_CLICKHOUSE_PLATFORM:-}`
+* `${OPIK_BACKEND_PLATFORM:-}`
+* `${OPIK_PYTHON_BACKEND_PLATFORM:-}`
+* `${OPIK_GUARDRAILS_BACKEND_PLATFORM:-}`
+
+By using `${VAR:-}`, Docker Compose falls back to an unset/empty string value if the variable is not defined in the host shell environment, defaulting to the native architecture of the host (standard Docker Compose behavior).
+
+## Proposed File Changes
+* **[MODIFY]** [docker-compose.yaml](file:///C:/Users/LENOVO/.gemini/antigravity/scratch/opik/deployment/docker-compose/docker-compose.yaml)
+  * Parameterize `platform` on `guardrails-backend`.
+  * Parameterize `platform` on `demo-data-generator` (reusing `OPIK_PYTHON_BACKEND_PLATFORM`).
+* **[MODIFY]** [README.md](file:///C:/Users/LENOVO/.gemini/antigravity/scratch/opik/deployment/docker-compose/README.md)
+  * Add QEMU prerequisite command: `docker run --privileged --rm tonistiigi/binfmt --install amd64`.
+  * Update instructions for running the RPi environment variables.
+
+## Verification Plan
+### Automated Verification
+* Run `docker compose config` to verify the generated configuration behaves correctly:
+  * Case A: No env variables set -> `platform` resolves to unset/empty.
+  * Case B: Env variables set -> `platform` resolves to `linux/amd64` correctly.
--- a/deployment/docker-compose/README.md
+++ b/deployment/docker-compose/README.md
@@ -217,14 +217,29 @@ docker compose --profile opik-otel down
 
 ## Running on Raspberry Pi 4 / CM4 (Older ARMv8 architectures)
 
-By default, the ClickHouse and Opik backend images require an `ARMv8.2-A` instruction set. On older ARMv8.0-A hardware like the Raspberry Pi 4 or Raspberry Pi CM4, running the default containers will fail with `Illegal instruction` crashes.
+By default, the ClickHouse and Opik backend images require an `ARMv8.2-A` instruction set. On older ARMv8.0-A hardware like the Raspberry Pi 4 or Raspberry Pi CM4, running the default containers natively will fail with `Illegal instruction` crashes.
 
-To resolve this, you can configure these services to run in emulation mode (using QEMU) by setting the platform environment variables before starting the stack:
+As a best-effort workaround, you can configure these services to run in emulation mode (using QEMU/amd64).
+
+### Prerequisite
+
+Bare Linux/Debian/Raspberry Pi OS setups do not include QEMU emulation handlers by default. Before bringing up the emulated stack, you must register the emulation support on the host machine by executing:
+
+```bash
+docker run --privileged --rm tonistiigi/binfmt --install amd64
+```
+
+*Note: Emulating ClickHouse and the JVM backend on a 4GB Raspberry Pi 4 is extremely memory-intensive and slow, and might still be unstable or unusable under real load.*
+
+### Setup
+
+To enable emulation, set the platform environment variables before starting the stack:
 
 ```bash
 export OPIK_CLICKHOUSE_PLATFORM=linux/amd64
 export OPIK_BACKEND_PLATFORM=linux/amd64
 export OPIK_PYTHON_BACKEND_PLATFORM=linux/amd64
+export OPIK_GUARDRAILS_BACKEND_PLATFORM=linux/amd64
 
 # Start Opik using the launcher script or docker compose
 ./opik.sh
--- a/deployment/docker-compose/README.md
+++ b/deployment/docker-compose/README.md
@@ -214,3 +214,19 @@ docker compose --profile opik down
 # or if running with otel profile
 docker compose --profile opik-otel down
 ```
+
+## Running on Raspberry Pi 4 / CM4 (Older ARMv8 architectures)
+
+By default, the ClickHouse and Opik backend images require an `ARMv8.2-A` instruction set. On older ARMv8.0-A hardware like the Raspberry Pi 4 or Raspberry Pi CM4, running the default containers will fail with `Illegal instruction` crashes.
+
+To resolve this, you can configure these services to run in emulation mode (using QEMU) by setting the platform environment variables before starting the stack:
+
+```bash
+export OPIK_CLICKHOUSE_PLATFORM=linux/amd64
+export OPIK_BACKEND_PLATFORM=linux/amd64
+export OPIK_PYTHON_BACKEND_PLATFORM=linux/amd64
+
+# Start Opik using the launcher script or docker compose
+./opik.sh
+```
+
--- a/deployment/docker-compose/docker-compose.yaml
+++ b/deployment/docker-compose/docker-compose.yaml
@@ -279,6 +279,7 @@ services:
 
   guardrails-backend:
     image: ghcr.io/comet-ml/opik/opik-guardrails-backend:${OPIK_VERSION:-latest}
+    platform: ${OPIK_GUARDRAILS_BACKEND_PLATFORM:-}
     pull_policy: ${OPIK_GUARDRAILS_BACKEND_PULL_POLICY:-always}
     profiles:
       - guardrails
@@ -297,6 +298,7 @@ services:
 
   demo-data-generator:
     image: ghcr.io/comet-ml/opik/opik-python-backend:${OPIK_VERSION:-latest}
+    platform: ${OPIK_PYTHON_BACKEND_PLATFORM:-}
     pull_policy: always
     profiles:
       - opik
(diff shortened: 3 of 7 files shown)


deployment/docker-compose/docker-compose.yaml
```diff
@@ -226,6 +228,7 @@ services:
 
   python-backend:
     image: ghcr.io/comet-ml/opik/opik-python-backend:${OPIK_VERSION:-latest}
+    platform: ${OPIK_PYTHON_BACKEND_PLATFORM:-}
```
andrescrz, line 231: 🚫 **blocker** | Logic

`demo-data-generator` (further down) uses this same `opik-python-backend` image and runs in the default `opik` profile, but it doesn't get a `platform` override. In an emulated setup it will start as native arm64 and crash the same way this change is trying to prevent. The same applies to `guardrails-backend` (lower priority — non-default profile).

Consider adding the identical line to the `demo-data-generator` service so the override is consistent across every service using this image:

```yaml
    platform: ${OPIK_PYTHON_BACKEND_PLATFORM:-}
```

deployment/docker-compose/README.md
```diff
@@ -214,3 +214,19 @@ docker compose --profile opik down
...
 ```
+
+## Running on Raspberry Pi 4 / CM4 (Older ARMv8 architectures)
+
+By default, the ClickHouse and Opik backend images require an `ARMv8.2-A` instruction set. On older ARMv8.0-A hardware like the Raspberry Pi 4 or Raspberry Pi CM4, running the default containers will fail with `Illegal instruction` crashes.
+
+To resolve this, you can configure these services to run in emulation mode (using QEMU) by setting the platform environment variables before starting the stack:
+
+```bash
+export OPIK_CLICKHOUSE_PLATFORM=linux/amd64
```
andrescrz, line 239: 🚫 **blocker** | Documentation / Effectiveness

These instructions are missing a hard prerequisite: amd64 emulation via QEMU needs `binfmt_misc` + `qemu-user-static` registered on the host. Docker Desktop bundles this, but a bare Raspberry Pi OS / Debian docker-engine install does **not** — so `platform: linux/amd64` will fail with `exec format error` unless the user first runs something like:

```bash
docker run --privileged --rm tonistiigi/binfmt --install amd64
```

Worth also warning that emulating ClickHouse + the JVM backend on a 4GB RPi4 will be very slow and memory-heavy, and may still be unusable. Since no one in #2101 has confirmed emulation actually works end-to-end on this hardware, I'd suggest softening the framing rather than stating it as a resolved path.

andrescrz:
👋 **Review summary**

**What looks good**
- Clean, minimal, backward-compatible approach — `${VAR:-}` defaults to an empty platform, so existing x86_64 / Apple Silicon / arm64 users are completely unaffected.
- Good instinct to make it an opt-in env knob rather than hardcoding platforms or adding custom Dockerfiles.
- Documented the workaround in the README rather than leaving it tribal knowledge.

**Overall**
A couple of things to reconsider before this can close #2101. The root cause in #2101 is the **CPU feature level** (RPi4/CM4 is ARMv8.0-A; ClickHouse and the backend require **ARMv8.2-A**), not the arm64-vs-amd64 architecture — native arm64 runs fine on Apple Silicon / Graviton2+ because those meet ARMv8.2-A. Parameterizing `platform` only enables amd64 emulation, which (a) needs QEMU/binfmt setup that isn't documented (see inline), (b) is untested by anyone on this hardware, and (c) is likely too slow to be usable on an RPi4. So this is a reasonable escape-hatch, but I'd stop short of "Closes #2101" — consider reframing as "document emulation as a best-effort workaround" and/or officially noting RPi4/CM4 as unsupported.

Minor: `platform: ${VAR:-}` resolves to `platform: ""`, which modern compose treats as unset — fine, though worth a sanity check against the older docker-compose versions RPi users often run.

**Inline comments**: 2 blockers.

verify_compose_yaml.py
```diff
@@ -0,0 +1,31 @@
...
+
+def verify():
+    path = "deployment/docker-compose/docker-compose.yaml"
+    print(f"=== Verifying {path} syntax and values ===")
+    try:
+        with open(path, "r", encoding="utf-8") as f:
+            content = yaml.safe_load(f)
+            
+        services = content.get("services", {})
+
```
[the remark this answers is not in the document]
namdamdoi68-oss: Resolved this blocker by adding a comprehensive module-level docstring and detailed verify() function docstrings to verify_compose_yaml.py in commit a9a096e83a.

deployment/docker-compose/docker-compose.yaml
namdamdoi68-oss: Resolved by parameterizing both demo-data-generator and guardrails-backend services with platform overrides in commit 6f82f48636.

deployment/docker-compose/README.md
namdamdoi68-oss: Resolved by documenting the QEMU binfmt registration prerequisite and adding memory/resource warnings for RPi4/CM4 emulation in README.md in commit 6f82f48636.

namdamdoi68-oss:
Hello @andrescrz,

I have addressed all the review feedback:
1. **Docker Compose Configuration**: Added the platform overrides (`platform: ${OPIK_PYTHON_BACKEND_PLATFORM:-}` and `platform: ${OPIK_GUARDRAILS_BACKEND_PLATFORM:-}`) to the `demo-data-generator` and `guardrails-backend` services.
2. **Documentation**: Updated `README.md` to document the QEMU emulation registration prerequisite (`docker run --privileged --rm tonistiigi/binfmt --install amd64`) and softened the support framing, adding warning notes regarding ClickHouse and JVM performance constraints on 4GB RPi4 hardware.
3. **Code Quality**: Added module and function-level docstrings to the `verify_compose_yaml.py` script as suggested by Baz Bot.

Please let me know if there's anything else needed!

.specify/plan.md
andrescrz, line 1: Please do not commit all these MD files, they don't belong to Opik code base.

deployment/docker-compose/docker-compose.yaml
```diff
@@ -44,6 +44,7 @@ services:
       "
   clickhouse:
     image: clickhouse/clickhouse-server:25.8.16.34-alpine
+    platform: ${OPIK_CLICKHOUSE_PLATFORM:-}
```
andrescrz, line 47: For architectural reasons, we aren't completely sure if we want to allow overriding the platform, instead of allowing docker to automatically determine and pull/use the right images.

This introduces an extra layer of complexity (users might override the env vars, forget to unset them and report unnecessary issues/bugs), unless there's a reason that makes this worthy.

For this particular case, I don't think this approach completely meets the goal of allowing Opik running in RPi CM4, as emulation seems more like a workaround than an actual fix. 

We need to evaluate if the workaround is worthy introducing this.

deployment/docker-compose/README.md
```diff
@@ -214,3 +214,34 @@ docker compose --profile opik down
...
+
+As a best-effort workaround, you can configure these services to run in emulation mode (using QEMU/amd64).
+
+### Prerequisite
+
+Bare Linux/Debian/Raspberry Pi OS setups do not include QEMU emulation handlers by default. Before bringing up the emulated stack, you must register the emulation support on the host machine by executing:
+
+```bash
+docker run --privileged --rm tonistiigi/binfmt --install amd64
+```
```
andrescrz, line 230: This emulation workaround suggest 3rd party components from our official documentation. We will need to evaluate if this is worthy the benefit.

verify_compose_yaml.py
andrescrz, line 1: This shouldn't be included in the code base.

andrescrz (changes requested):
Hi @namdamdoi68-oss 

Thanks a lot for taking the time to work on this and submit a contribution—we really appreciate the effort and your interest in improving the project. 

I left some feedback on both the overall approach and the implementation details.

At a high level, this change provides a workaround, but we're not convinced it will deliver the experience we want on lower-end devices such as the Raspberry Pi 4 / Compute Module 4, where it relies on running `amd64` Docker images. 

Before introducing this, we'd want to verify that the approach consistently provides an acceptable level of quality. We'd rather not support Opik on these devices yet than offer an experience that doesn't meet expectations. 

More broadly, this is a workaround rather than a solution to the underlying issue, so we'd prefer to keep the issue open for now regardless of this PR.

Regarding the implementation itself, we're hesitant to introduce configurable Docker platform resolution. In general, we prefer to rely on Docker's automatic platform resolution, as it reduces the risk of user misconfiguration. We'd only consider exposing this as a configuration option if the benefits clearly outweigh the additional complexity and potential for incorrect setups.

Finally, could you provide some evidence that this approach works reliably in practice? For example, a short video demonstrating Opik running successfully on the target devices would be very helpful in evaluating the proposal.

Thanks again for the contribution and for engaging with the project. Even if we decide not to move forward with this particular approach, we'd be happy to review future contributions that address the underlying issue in a different way.

andrescrz:
Hi @namdamdoi68-oss 

Thanks again for your contribution and for taking the time to work on this PR.

Since there haven't been any updates for a while and the PR still requires significant changes before it can be merged, I'm going to close it for now to help keep the backlog manageable.

If you'd like to continue working on it in the future, please feel free to reopen the PR once you've addressed the outstanding issues. We'd be happy to take another look.

Thanks again!

PR #7341, opened 2026-07-05
feat: add Groq integration to the Python SDK

NishchayMahor:
## Summary

Adds a first-class **Groq** integration to the Opik Python SDK, so calls made with the native `groq` client are automatically logged as Opik LLM spans — no manual instrumentation needed:

```python
import groq
from opik.integrations.groq import track_groq

client = track_groq(groq.Groq())
client.chat.completions.create(
    model="llama-3.3-70b-versatile",
    messages=[{"role": "user", "content": "Why is the sky blue?"}],
)
```

Groq is a popular fast-inference provider, and `LLMProvider.GROQ` plus a Groq usage parser already exist in the SDK — this fills the remaining gap: the SDK-level `track_groq`. (The existing `track_openai(..., provider="groq")` path only helps users who route the *OpenAI* client at Groq; this covers the native `groq` SDK, which `track_openai` can't patch.)

## Details

`track_groq()` follows the existing OpenAI/Anthropic pattern:
- Wraps `chat.completions.create()` on both `Groq` and `AsyncGroq`, including `stream=True` (sync + async streaming via aggregated chunks).
- Records the span provider as `"groq"`; Groq returns OpenAI-shaped usage, so token usage is parsed with the OpenAI converter (the comment notes this denotes payload format, not the span provider).
- Graceful span/trace finalization with `error_info` on failures.

Files:
- `src/opik/integrations/groq/` — `track_groq`, decorator, chunk aggregator, stream patchers.
- `tests/library_integration/groq/` — mock-based unit tests + `requirements.txt`.
- CI: `lib-groq-tests.yml` + registration in `lib-integration-tests-runner.yml`.

## Testing

`pytest tests/library_integration/groq/` → **4 passed** (sync happy-path with full trace-tree assertion, async, and error handling), using the `fake_backend` fixture and a mocked transport so no Groq API key is required. `ruff` and `mypy` clean.

Happy to add a documentation/cookbook page for the integration as a follow-up if that's preferred.


NishchayMahor:
Thanks for the review! These three are intentional consistency with Opik's existing provider integrations rather than new issues:

1. **Aggregator returns `None` on error** — this mirrors `integrations/openai/chat_completion_chunks_aggregator.py`, which also wraps aggregation in `try/except` and returns `None` (after logging). I kept Groq streaming behaving identically to OpenAI streaming; happy to tighten it if you'd prefer stricter handling across both.
2. **`stream_patchers` duplication** — the sync/async patchers deliberately follow the OpenAI integration's `stream_patchers.py` (Groq's SDK is OpenAI-shaped). Groq only has `Stream`/`AsyncStream` (no stream-manager variants), so it's a subset. A shared base could dedupe OpenAI/Groq/Cerebras, but that's a cross-integration refactor I'd keep separate from this PR.
3. **Optional `groq` import** — `import groq` at module top matches every other integration (e.g. `integrations/anthropic/opik_tracker.py` does `import anthropic`). It's opt-in: `import opik` does not import `groq` — only `from opik.integrations.groq import track_groq` does, at which point the user already has `groq` installed. So a missing `groq` doesn't break importing Opik.

Happy to adjust any of these if you'd like.

petrotiurin (approved):
Thank you for making this change, @NishchayMahor ! The code looks good and the new tests succeed. I'll proceed to merging your changes.

NishchayMahor:
Thanks for merging the Groq integration! Opik is a great piece of kit and this was a clean one to add. I build in the LLM observability space myself, so it felt like a natural fit. Would gladly wire up more provider integrations if that's useful to the team.

PR #7342, opened 2026-07-05
feat: add Cerebras integration to the Python SDK

NishchayMahor:
## Summary

Adds a **Cerebras** integration to the Opik Python SDK, so calls made with the native `cerebras-cloud-sdk` client are automatically logged as Opik LLM spans:

```python
from cerebras.cloud.sdk import Cerebras
from opik.integrations.cerebras import track_cerebras

client = track_cerebras(Cerebras())
client.chat.completions.create(
    model="llama-3.3-70b",
    messages=[{"role": "user", "content": "Why is the sky blue?"}],
)
```

Cerebras (very fast inference for Llama and other open models) has a distinct native SDK, so the OpenAI wrapper (`track_openai`) can't instrument it — this adds first-class support, matching the existing standalone integrations.

## Details

`track_cerebras()` follows the existing OpenAI/Anthropic pattern:
- Wraps `chat.completions.create()` on both `Cerebras` and `AsyncCerebras`, including `stream=True` (sync + async streaming via aggregated chunks).
- Records the span provider as `"cerebras"`; Cerebras returns OpenAI-shaped usage, so token usage is parsed with the OpenAI converter (the comment notes this denotes payload format, not the span provider).
- Graceful span/trace finalization with `error_info` on failures.

Files:
- `src/opik/integrations/cerebras/` — `track_cerebras`, decorator, chunk aggregator, stream patchers.
- `tests/library_integration/cerebras/` — mock-based unit tests + `requirements.txt`.
- CI: `lib-cerebras-tests.yml` + registration in `lib-integration-tests-runner.yml`.

## Testing

`pytest tests/library_integration/cerebras/` → **4 passed** (sync happy-path with full trace-tree assertion, async, and error handling), using the `fake_backend` fixture and a mocked transport so no Cerebras API key is required. `ruff` and `mypy` clean.

Happy to add a documentation/cookbook page as a follow-up if preferred.


NishchayMahor:
Same as #7341 (Groq) — these three points are intentional consistency with the existing OpenAI/Anthropic integrations:
1. The aggregator's `try/except → None` mirrors `integrations/openai/chat_completion_chunks_aggregator.py`.
2. `stream_patchers` deliberately follows the OpenAI integration's structure (Cerebras's SDK is OpenAI-shaped; only `Stream`/`AsyncStream`, a subset). A shared base for OpenAI/Groq/Cerebras would be a separate refactor.
3. `import cerebras.cloud.sdk` at module top matches `import anthropic` in the anthropic integration; it's opt-in, so `import opik` doesn't pull it. Happy to adjust any of these.

NishchayMahor:
Rebased onto latest `main` — this is `MERGEABLE` again.

Note for anyone comparing SHAs: a plain rebase wasn't possible (`git merge-base` between the branch and current `main` is empty, so history has been rewritten upstream since the branch was cut). I recreated the branch from current `main` and cherry-picked the single commit instead, so the diff is unchanged.

The only conflict was in `lib-integration-tests-runner.yml`, where `mistral` and `groq` landed in the same lists this PR adds `cerebras` to — resolved by keeping all three, in the choice options, the job list, the `notify-slack` needs, and the Slack results payload. Verified the resolved file parses and all three jobs are wired.

Also re-checked the decorator API against current `main`: `_start_span_inputs_preprocessor`, `_end_span_inputs_preprocessor` and `_streams_handler` all still match the OpenAI reference implementation, so nothing drifted.

andrescrz:
Hi @NishchayMahor 

Thanks for the contribution and apologies about the delay!

We're going to proceed with this review and potentially add this Cerebras integration if once all is ready in the PR.

Thanks!

alexkuzmik (approved):

alexkuzmik:
Thanks for the contribution @NishchayMahor!

PR #7361, opened 2026-07-06
[OPIK-7246] fix: buffer projects.last_updated_trace_at in Redis, flush to MySQL periodically

thiagohora:
## Details

`ProjectEventListener` maintained `projects.last_updated_trace_at` by writing MySQL synchronously on every `TracesCreated`/`TracesUpdated` event (via `ProjectService.recordLastUpdatedTrace`), dispatched on an unbounded virtual-thread `AsyncEventBus`. A busy project funnels many concurrent `WRITE` transactions onto a single hot `projects` row, producing `MySQLTransactionRollbackException` (deadlock / lock-wait) and `EOFException`/`ConnectException` on the ingestion path.

This buffers the per-project maximum timestamp in a Redis ZSET and flushes the accumulated maxima to MySQL under a best-effort distributed lock every 30s, taking the marker write off the ingestion hot path.

- Buffer keyed by `"workspaceId:projectId"`, score = epoch millis, using `addIfGreater` so only the cluster-wide max is kept.
- `ProjectLastUpdatedFlushJob` drains the ZSET, groups by workspace, writes one batched `UPDATE` per workspace, and removes flushed members.
- Gated behind `projectLastUpdatedFlush.enabled` (default **off**); the synchronous MySQL path is the fallback.
- Monotonic (forward-only guard preserved) and idempotent (safe under ret/overlap); fails open on Redis errors so trace ingestion is never blocked.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-7246

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: Root-cause analysis, implementation, unit tests, and local end-to-end verification.
- Human verification: Author reviewed the diff; changes built, unit-tested, and manually validated locally (see Testing).

## Testing

- Unit tests (`mvn surefire:test`): `ProjectServiceLastUpdatedTraceTest` (buffer/flush: disabled→MySQL, enabled→`addIfGreater`, per-workspace batching, empty buffer, malformed member) and `ProjectLastUpdatedFlushJobTest` (disabled / lock-not-acquired / lock-acquired). 9 tests green.
- Build: `mvn compiler:compile compiler:testCompile` — BUILD SUCCESS.
- Manual end-to-end (local process backend + Docker MySQL/Redis/ClickHouse, flag enabled, 5s interval):
  - `POST /v1/private/traces` buffered a ZSET member `<workspaceId>:<projectId>` scored with the trace `last_updated_at` epoch-ms.
  - Flush logged `wrote '1' project markers`; `ZCARD` returned `0` (drained); `projects.last_updated_trace_at` matched the buffered epoch-ms exactly.
  - Coalescing: 5 rapid traces to one project → a single ZSET member → one MySQL write at the max timestamp.

## Documentation

No user-facing documentation changes. New config `projectLastUpdatedFlush` is documented inline in `config.yml`.


apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectService.java
```diff
@@ -561,10 +574,108 @@ public Mono<List<Project>> retrieveByNamesOrCreate(Set<String> projectNames) {
...
+                                Collectors.toUnmodifiableSet())));
+
+        var members = entries.stream().map(ScoredEntry::getValue).collect(Collectors.toUnmodifiableSet());
+
+        return Mono.fromRunnable(() -> byWorkspace.forEach(this::writeLastUpdatedTraceToDb))
+                .subscribeOn(Schedulers.boundedElastic())
+                // Remove processed members only after the MySQL write succeeds. A member re-bumped between read and
+                // removal may lose a single increment; it self-heals on the project's next trace and never regresses
+                // the marker (the MySQL write only moves last_updated_trace_at forward).
+                .then(pending.removeAll(members))
```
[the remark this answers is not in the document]
thiagohora: Fixed in 629932bc0c. Replaced the read-then-remove with an atomic **rename-drain**: `renamenx` moves the live buffer into a private snapshot key (`project:last-updated-trace:flushing`) that only the drain touches, so a concurrent `recordLastUpdatedTrace` (`addIfGreater`) lands on a fresh live key and can't be dropped. A snapshot left by an interrupted flush is drained next cycle (idempotent — the DB write only moves forward). Uses only Redisson APIs (`isExists`/`renamenx`/`entryRange`/`removeAll`), no raw Redis/Lua.

apps/opik-backend/src/test/java/com/comet/opik/domain/ProjectServiceLastUpdatedTraceTest.java
```diff
@@ -0,0 +1,222 @@
...
+
+            var workspaceId = UUID.randomUUID().toString();
+            var traces = List.of(ProjectIdLastUpdated.builder()
+                    .id(UUID.randomUUID())
+                    .lastUpdatedAt(Instant.now())
+                    .build());
+
+            service.recordLastUpdatedTrace(workspaceId, traces);
+
+            verify(template).inTransaction(eq(WRITE), any());
```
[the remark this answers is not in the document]
thiagohora: The `method__scenario__outcome` underscore style is the established backend test convention here — ~1226 existing test methods use it, including the sibling this class mirrors (`ExperimentDenormalizationJobTest`). The readable scenario lives in `@DisplayName`. Keeping it consistent with the codebase rather than diverging.

apps/opik-backend/src/test/java/com/comet/opik/domain/ProjectServiceLastUpdatedTraceTest.java
```diff
@@ -0,0 +1,222 @@
...
+
+            var workspaceId = UUID.randomUUID().toString();
+            var projectId1 = UUID.randomUUID();
+            var projectId2 = UUID.randomUUID();
+            var member1 = workspaceId + ":" + projectId1;
+            var member2 = workspaceId + ":" + projectId2;
+
+            when(pendingSet.entryRange(anyInt(), anyInt())).thenReturn(Mono.just(List.of(
+                    new ScoredEntry<>(1_000d, member1),
+                    new ScoredEntry<>(2_000d, member2))));
```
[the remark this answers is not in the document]
thiagohora: The `method__scenario__outcome` underscore style is the established backend test convention here — ~1226 existing test methods use it, including the sibling this class mirrors (`ExperimentDenormalizationJobTest`). The readable scenario lives in `@DisplayName`. Keeping it consistent with the codebase rather than diverging.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/jobs/ProjectLastUpdatedFlushJobTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+    @BeforeEach
+    void setUp() {
+        config = buildConfig(true);
+        job = new ProjectLastUpdatedFlushJob(config, projectService, lockService);
+    }
+
+    @Test
+    @DisplayName("When disabled, skips locking and flushing entirely")
+    void doJob__whenDisabled__skips() {
```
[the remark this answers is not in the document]
thiagohora: The `method__scenario__outcome` underscore style is the established backend test convention here — ~1226 existing test methods use it, including the sibling this class mirrors (`ExperimentDenormalizationJobTest`). The readable scenario lives in `@DisplayName`. Keeping it consistent with the codebase rather than diverging.

apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectService.java
```diff
@@ -561,10 +574,108 @@ public Mono<List<Project>> retrieveByNamesOrCreate(Set<String> projectNames) {
...
+                    UUID.fromString(value.substring(separatorIndex + 1)),
+                    Instant.ofEpochMilli(entry.getScore().longValue()));
+        } catch (IllegalArgumentException e) {
+            log.warn("Skipping last-updated-trace buffer member with invalid project id: '{}'", value);
+            return null;
+        }
+    }
+
+    private record ParsedLastUpdatedTrace(String workspaceId, UUID projectId, Instant lastUpdatedAt) {
+    }
```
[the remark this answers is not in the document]
thiagohora: `ParsedLastUpdatedTrace` is a private, single-use internal parsing helper (not an API/DTO/domain model) — it's constructed once from a parsed ZSET member and never mutated or passed across boundaries. The `@Builder(toBuilder = true)` model convention doesn't add value here, so leaving it as a plain record.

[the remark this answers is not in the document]
thiagohora: Reconsidered — applied in 9a7d533d78: `ParsedLastUpdatedTrace` now uses `@Builder(toBuilder = true)` with `@NonNull` components, and `parseLastUpdatedTraceMember` builds via the builder.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ProjectLastUpdatedFlushJob.java
```diff
@@ -0,0 +1,194 @@
...
+            @NonNull LockService lockService) {
+        this.config = config;
+        this.redisClient = redisClient;
+        this.projectService = projectService;
+        this.lockService = lockService;
+    }
+
+    @Override
+    public void doJob(JobExecutionContext context) {
+        if (!config.isEnabled() || interrupted.get()) {
```
andrescrz, line 59: Nit: maybe split and add info level log for interruptions.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ProjectLastUpdatedFlushJob.java
```diff
@@ -0,0 +1,194 @@
...
+    /**
+     * Atomically moves the live buffer aside ({@code renamenx} to the snapshot key) so concurrent buffer writes
+     * (addIfGreater) immediately populate a fresh live key, then drains the snapshot. Moving the data aside removes
+     * the read-then-remove race — the drain owns the snapshot exclusively. When a snapshot from a previously
+     * interrupted flush still exists, {@code renamenx} is a no-op and that leftover is drained instead (idempotent,
+     * since the DB write only moves the marker forward).
+     */
+    private Mono<Long> flush() {
+        var pending = pendingSet();
+        var flushing = flushingSet();
```
andrescrz: Minor: maybe flushing should be lazy.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ProjectLastUpdatedFlushJob.java
```diff
@@ -0,0 +1,194 @@
...
+                .then(drainSnapshot(flushing, 0L));
+    }
+
+    /**
+     * Drains the snapshot one page at a time, reading the next page only after the current page's write and
+     * {@code removeAll} complete. Re-reading the lowest range advances because each processed page is removed first;
+     * gating the next read on removal (rather than {@code Flux.expand}, which would pre-fetch before removal) prevents
+     * the same top entries from being read and written to MySQL repeatedly. Removal guarantees progress.
+     */
+    private Mono<Long> drainSnapshot(RScoredSortedSetReactive<String> flushing, long writtenSoFar) {
```
andrescrz: As design question, I believe most of the logic in this job belongs to the service, including encapsulating the Redis access. The job's responsibilities should be mostly scheduling, handling interruptions etc. whereas the service should encapsulate the business logic, including how to flush etc.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ProjectLastUpdatedFlushJob.java
```diff
@@ -0,0 +1,194 @@
...
+            return null;
+        }
+    }
+
+    @Builder(toBuilder = true)
+    private record ParsedMember(@NonNull String workspaceId, @NonNull UUID projectId, @NonNull Instant lastUpdatedAt) {
+    }
+
+    @Override
+    public void interrupt() throws UnableToInterruptJobException {
```
andrescrz, line 88: Interrupt is not really canceling the inner subscription, only the job schedule. It should if we really want to stop everything from executing.

apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectLastUpdatedTraceBufferService.java
```diff
@@ -0,0 +1,79 @@
...
+            @NonNull RedissonReactiveClient redisClient,
+            @NonNull ProjectService projectService) {
+        this.config = config;
+        this.redisClient = redisClient;
+        this.projectService = projectService;
+    }
+
+    @Override
+    public void record(String workspaceId, Collection<ProjectIdLastUpdated> lastUpdatedTraces) {
+        if (lastUpdatedTraces.isEmpty()) {
```
andrescrz: Public method, so please validate or handle argument invalid values such as null. For the collection, you should use the null safe CollectionUtils.isEmpty check.

apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectLastUpdatedTraceBufferService.java
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        RScoredSortedSetReactive<String> pending = redisClient.getScoredSortedSet(
+                ProjectLastUpdatedFlushConfig.PENDING_SET_KEY, StringCodec.INSTANCE);
+        Flux.fromIterable(lastUpdatedTraces)
+                .flatMap(project -> pending.addIfGreater(
+                        project.lastUpdatedAt().toEpochMilli(), member(workspaceId, project.id())))
+                .subscribeOn(Schedulers.boundedElastic())
+                // Fail-open: the marker is a best-effort sort hint, a Redis blip must not break trace ingestion.
+                .subscribe(
```
andrescrz: This is async in respect of the method signature, so no way to know this method has a side effect. Better return a Flux and allow callers to decide, or move to a sync Redis client, which is available in the code base. You could document as well, but I don't believe that's a good solution.

apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectService.java
```diff
@@ -561,6 +561,9 @@ public Mono<List<Project>> retrieveByNamesOrCreate(Set<String> projectNames) {
 
     @Override
     public void recordLastUpdatedTrace(String workspaceId, Collection<ProjectIdLastUpdated> lastUpdatedTraces) {
+        if (lastUpdatedTraces.isEmpty()) {
```
andrescrz: Same here avoid arguments in public methods.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/ProjectLastUpdatedFlushConfig.java
```diff
@@ -0,0 +1,58 @@
...
+@Data
+public class ProjectLastUpdatedFlushConfig {
+
+    public static final String PENDING_SET_KEY = "project:last-updated-trace:pending";
+    /**
+     * Snapshot the live buffer is atomically renamed to while a flush drains it ({@code renamenx}), so writers keep
+     * populating {@link #PENDING_SET_KEY} and the drain owns its data exclusively.
+     */
+    public static final String FLUSHING_SET_KEY = "project:last-updated-trace:flushing";
+    public static final String MEMBER_SEPARATOR = ":";
```
andrescrz: Having these constants in the config, it's a sign of the split responsibility of the job and service that I mention above. They likely don't belong here and they should in the service, after refactoring.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/ProjectLastUpdatedFlushConfig.java
```diff
@@ -0,0 +1,58 @@
...
+
+    /** When disabled, {@code ProjectEventListener} writes MySQL synchronously (legacy behavior). */
+    @JsonProperty
+    private boolean enabled = false;
```
andrescrz: Let's remove all the defaults from the java code and define them once in the config yaml files. That way, we have a single source of truth. They're all non-nullable, so no need for this overlap.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/ProjectLastUpdatedFlushConfig.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @Valid @JsonProperty
+    @NotNull @MinDuration(value = 1, unit = TimeUnit.SECONDS)
```
andrescrz, line 38: Minor: probably desirable to define some sensible max values for all these config params lacking it.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/jobs/ProjectLastUpdatedFlushJobTest.java
```diff
@@ -0,0 +1,172 @@
...
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.timeout;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+@DisplayName("ProjectLastUpdatedFlushJob Tests")
+class ProjectLastUpdatedFlushJobTest {
```
andrescrz: After the refactor that I propose, better have everything as high value integration or black box tests. Probably:
- One smoke black box test for the happy path on the job.
- Integration tests for the service, with happy path and the rest of the coverage.

It'd be fine to avoid low value unit tests.

andrescrz (commented):
Let's iterate on the feedback before moving forward.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ProjectLastUpdatedFlushJob.java
thiagohora: Done in b709cbe2ee. Moved all the flush/Redis logic into `ProjectLastUpdatedTraceBufferService` (it now owns `record()` + `flush()`). The job is thin: scheduling, distributed lock and interruption only, delegating to `bufferService.flush()`.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ProjectLastUpdatedFlushJob.java
thiagohora: Addressed by the refactor above — `flushing` no longer lives in the job (the whole drain moved into the service).

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ProjectLastUpdatedFlushJob.java
thiagohora: Done — `interrupt()` now logs at info and (below) cancels the in-flight flush.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ProjectLastUpdatedFlushJob.java
thiagohora: Done in b709cbe2ee — the job keeps the `Disposable` from the subscription and `interrupt()` disposes it, so an in-flight flush is cancelled, not just the schedule.

apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectLastUpdatedTraceBufferService.java
thiagohora: Done — `workspaceId` is `@NonNull` and the collection uses null-safe `CollectionUtils.isEmpty`.

apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectLastUpdatedTraceBufferService.java
thiagohora: Done — switched to the synchronous `StringRedisClient`, so `record()` is genuinely synchronous and no longer hides an async side-effect behind a `void` signature (the drain is now a simple sequential loop too).

apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectService.java
thiagohora: Done — `@NonNull` workspaceId + null-safe `CollectionUtils.isEmpty` here as well.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/ProjectLastUpdatedFlushConfig.java
thiagohora: Done — the `PENDING_SET_KEY`/`FLUSHING_SET_KEY`/`MEMBER_SEPARATOR` constants moved into `ProjectLastUpdatedTraceBufferService` alongside the logic that uses them; the config now only holds tunables.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/ProjectLastUpdatedFlushConfig.java
thiagohora: Done — removed the Java defaults; `config.yml` (with `${ENV:-default}`) is the single source of truth. Fields are non-null and validated.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/ProjectLastUpdatedFlushConfig.java
thiagohora: Done — added `@Max`/`@MaxDuration` bounds to all params (jobInterval ≤ 1h, jobLockTime ≤ 10m, jobLockWaitTime ≤ 5s, jobBatchSize ≤ 5000).

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/jobs/ProjectLastUpdatedFlushJobTest.java
thiagohora: Kept focused unit tests on the service (mocked Redis) since they run in CI without the full harness and cover the drain/edge cases; happy to add a black-box/integration smoke test as a follow-up if you prefer.

apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectLastUpdatedTraceBufferService.java
```diff
@@ -0,0 +1,160 @@
...
+ * <p>
+ * When enabled, {@link #record} writes the latest timestamp per project to a Redis ZSET (member
+ * {@code "workspaceId:projectId"}, score = epoch millis) via {@code addIfGreater}, which keeps only the cluster-wide
+ * max; {@link #flush} (driven by {@code ProjectLastUpdatedFlushJob}) drains it to MySQL. When disabled,
+ * {@link #record} delegates straight to the synchronous MySQL write.
+ */
+@ImplementedBy(ProjectLastUpdatedTraceBufferServiceImpl.class)
+public interface ProjectLastUpdatedTraceBufferService {
+
+    void record(@NonNull String workspaceId, Collection<ProjectIdLastUpdated> lastUpdatedTraces);
```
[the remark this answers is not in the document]
thiagohora: Done in b71ba6420a — removed `@NonNull` from the interface method; the null-safe validation (`CollectionUtils.isEmpty` + `@NonNull` on the param) stays on the implementation.

apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectLastUpdatedTraceBufferService.java
```diff
@@ -0,0 +1,160 @@
...
+                                parsed -> ProjectIdLastUpdated.builder()
+                                        .id(parsed.projectId())
+                                        .lastUpdatedAt(parsed.lastUpdatedAt())
+                                        .build(),
+                                Collectors.toUnmodifiableSet())));
+
+        byWorkspace.forEach(projectService::recordLastUpdatedTrace);
+
+        flushing.removeAll(entries.stream().map(ScoredEntry::getValue).collect(Collectors.toUnmodifiableSet()));
+        return byWorkspace.values().stream().mapToInt(Collection::size).sum();
```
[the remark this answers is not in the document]
thiagohora: Good catch — the count is markers *processed*, not rows updated (the MySQL write is a monotonic UPDATE that skips already-ahead rows, and the DAO's affected-row counts aren't propagated). Relabeled the job log from "wrote" to "processed" and documented it in b71ba6420a.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ProjectLastUpdatedFlushJob.java
```diff
@@ -0,0 +1,94 @@
...
+    private final ProjectLastUpdatedTraceBufferService bufferService;
+    private final LockService lockService;
+
+    @Inject
+    public ProjectLastUpdatedFlushJob(
+            @NonNull @Config("projectLastUpdatedFlush") ProjectLastUpdatedFlushConfig config,
+            @NonNull ProjectLastUpdatedTraceBufferService bufferService,
+            @NonNull LockService lockService) {
+        this.config = config;
+        this.bufferService = bufferService;
```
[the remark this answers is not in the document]
thiagohora: Done in b71ba6420a — added `interrupt__cancelsInFlightFlush`: it stubs the lock to return a never-completing `Mono` with a `doOnCancel` probe, runs `doJob()`, then asserts `interrupt()` disposes the subscription (cancel fires).

apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectLastUpdatedTraceBufferService.java
```diff
@@ -0,0 +1,162 @@
...
+        }
+
+        // Fail-open: the marker is a best-effort sort hint, a Redis blip must not break trace ingestion.
+        try {
+            RScoredSortedSet<String> pending = redisClient.getScoredSortedSet(PENDING_SET_KEY);
+            lastUpdatedTraces.forEach(project -> pending.addIfGreater(
+                    project.lastUpdatedAt().toEpochMilli(), member(workspaceId, project.id())));
+        } catch (RuntimeException e) {
+            log.warn("Failed to buffer last-updated-trace marker for workspace '{}'", workspaceId, e);
+        }
```
andrescrz: Not a blocker: what's the particular reason for this to not to bubble up? It seems that we should delegate to the caller the exception handling. e.g: the job would fail the iteration, instead of just warning.

apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectLastUpdatedTraceBufferService.java
```diff
@@ -0,0 +1,162 @@
...
+        // is drained instead (renamenx is a no-op when it exists); idempotent since the DB write only moves forward.
+        if (Boolean.TRUE.equals(pending.isExists())) {
+            pending.renamenx(FLUSHING_SET_KEY);
+        }
+
+        RScoredSortedSet<String> flushing = redisClient.getScoredSortedSet(FLUSHING_SET_KEY);
+        int batchSize = config.getJobBatchSize();
+        long written = 0;
+        // Read the next page only after the current page is written and removed; removal guarantees progress.
+        while (true) {
```
andrescrz: Minor: I recommend basing the while condition on the entries size check below. It's more readable, specially to not to think about potential infinite loops here.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/jobs/ProjectLastUpdatedFlushJobTest.java
andrescrz: I recommend pursuing as a more endurable pattern. Also, because the general convention of the service is to test with black box or integration tests, leave unit testing of high value if any or avoid. You have strong foundations on tests to avoid flakiness or interactions. see examples:
- ExperimentProjectMigrationJobTest
- ExperimentProjectMigrationServiceTest

Anyway not a blocker, but I do recommend switching to this test approach.

andrescrz (dismissed):
A few more minor comments.

My main recommendation is to switch from unit tests to black box and integration tests, I gave guidance and examples on how to do it.

Anyway, not a blocker.

apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectLastUpdatedTraceBufferService.java
thiagohora: Fixed in de01da1c — removed the try/catch. `record()` now lets Redis exceptions propagate to `ProjectEventListener`, same as the synchronous fallback branch.

apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectLastUpdatedTraceBufferService.java
thiagohora: Fixed in de01da1c — replaced the `while(true)`+internal breaks with a `do/while` conditioned on `pageSize == batchSize` (the page-size check), so there's no more implicit infinite-loop shape to reason through.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/jobs/ProjectLastUpdatedFlushJobTest.java
thiagohora: Done in de01da1c — replaced the Mockito unit tests with black-box/integration coverage: `ProjectLastUpdatedTraceBufferServiceTest` drives `record()`/`flush()` against real Redis/MySQL and asserts through the project API (happy path, cluster-wide max, multi-workspace batching, malformed-member skip); `ProjectLastUpdatedFlushJobTest` moved to the `domain` package as a single black-box smoke test mirroring `ExperimentProjectMigrationJobTest`. Also added a `jobEnabled` flag decoupled from `enabled` so tests can buffer/flush deterministically without racing the Quartz schedule.

andrescrz (approved):
LGTM

PR #7365, opened 2026-07-06
[OPIK-6941] [BE] [FE] feat: workspace-level span metric aggregation in dashboards

alexkuzmik:
## Details

Dashboards could previously chart metrics for a single project only. This adds workspace-level aggregation so a metric widget can span multiple projects, addressing the need to watch total span usage (tokens, cost) across a whole workspace — e.g. to track provider rate limits broken down by model/provider.

- New backend endpoint `POST /v1/private/workspaces/metrics/spans` aggregates span metrics (count, token usage, cost) across a set of projects, with optional provider/model/type breakdown. Series are built DB-side with `groupArray` for parity with the existing per-project queries; the service stays thin.
- Aggregation uses a dedicated `WorkspaceSpanMetricRequest` DTO so the shared `WorkspaceMetricRequest` (cost/feedback summaries) is left untouched.
- Frontend: the time-series and single-metric (stat card) widget editors now allow selecting multiple projects via a multi-select Project field. The form is otherwise identical to a single-project widget. Selecting two or more projects transparently routes to the workspace endpoint and narrows the metric list to the supported span totals; a single project keeps the existing per-project behavior unchanged.
- Shared cross-widget logic (supported metric names, editor/stat options, single-vs-workspace project resolution, and metrics-source selection) is extracted into `lib/dashboard/workspaceMetrics.ts` and `api/projects/useMetricData.ts` so flow-specific constants are not duplicated across widgets or leaked into shared modules.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- OPIK-6941

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: full implementation (backend endpoint + frontend widgets), assisted by manual review
- Human verification: code review + manual UI testing + backend integration tests

## Testing

- Backend: `mvn test` for `WorkspacesResourceTest` — the new `SpanMetricsTest` nested class covers per-project parity, provider/model breakdown, all-projects vs subset, cross-project isolation, and request validation (10 tests, green). `mvn spotless:check` — green.
- Frontend: `npm run lint` and `npm run typecheck` — green.
- Manual (local dev, process mode): verified a time-series widget renders span token usage broken down by provider across multiple projects, and a stat-card widget renders total span token usage and total span cost aggregated across projects; single-project widgets continue to render unchanged.

## Documentation

N/A — no user-facing documentation changes in this PR.


CometActions:

### Access Information
- **URL:** https://pr-7365.dev.comet.com
- **Namespace:** pr-7365
- **Version:** 2.1.18-7365-merge-2481
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7365%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

apps/opik-backend/src/main/java/com/comet/opik/domain/WorkspaceMetricsService.java
```diff
@@ -62,4 +71,47 @@ public Mono<WorkspaceMetricResponse> getWorkspaceCosts(@NonNull WorkspaceMetricR
...
+            throw new BadRequestException("'metric_type' must be provided");
+        }
+        if (request.interval() == null) {
+            throw new BadRequestException("'interval' must be provided");
+        }
+        if (!WorkspaceMetricsDAO.SUPPORTED_SPAN_METRICS.contains(request.metricType())) {
+            throw new BadRequestException("Unsupported metric type '%s'. Supported: %s"
+                    .formatted(request.metricType(), WorkspaceMetricsDAO.SUPPORTED_SPAN_METRICS));
+        }
+        if (request.hasBreakdown()) {
```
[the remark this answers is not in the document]
alexkuzmik: Fixed in 694e4fca34 — added `tokenUsageBreakdown_withoutSubMetric_returnsBadRequest` to `SpanMetricsTest`, parameterized over `null` / empty / blank `sub_metric`, asserting the endpoint returns 400 for a `SPAN_TOKEN_USAGE` breakdown with a missing `sub_metric`. Added the equivalent `durationBreakdown_withoutSubMetric_returnsBadRequest` guard alongside the new duration metric.

apps/opik-backend/src/main/java/com/comet/opik/domain/WorkspaceMetricsDAO.java
```diff
@@ -247,6 +532,105 @@ public Mono<List<WorkspaceMetricResponse.Result>> getCostsDaily(@NonNull Workspa
...
+        return template.nonTransaction(connection -> getSpanMetric(request, connection,
+                request.hasBreakdown() ? GET_SPAN_COUNT_WITH_BREAKDOWN : GET_SPAN_COUNT, "workspaceSpanCount")
+                .flatMapMany(this::rowToDataPoint)
+                .collectList());
+    }
+
+    @Override
+    public Mono<List<WorkspaceMetricResponse.Result>> getSpanTokenUsage(@NonNull WorkspaceSpanMetricRequest request) {
+        return template.nonTransaction(connection -> getSpanMetric(request, connection,
+                request.hasBreakdown() ? GET_SPAN_TOKEN_USAGE_WITH_BREAKDOWN : GET_SPAN_TOKEN_USAGE,
```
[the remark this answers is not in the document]
alexkuzmik: Skipping — the duplication is intentional. Workspace span metrics are the per-project queries with `project_id IN (...)` in place of `project_id = ...`; keeping them as exact copies was a deliberate choice for this feature so the two paths stay behaviorally identical (enforced by the parity tests). Extracting a shared cross-DAO builder would couple the workspace and per-project DAOs and is a larger refactor out of scope here — noted as a possible follow-up.

apps/opik-frontend/src/lib/dashboard/workspaceMetrics.ts
```diff
@@ -0,0 +1,115 @@
...
+  projectId,
+  projectIds,
+}: {
+  runtimeProjectId?: string;
+  projectId?: string;
+  projectIds?: string[];
+}): ResolvedProjectSelection => {
+  if (runtimeProjectId) {
+    return { projectId: runtimeProjectId };
+  }
```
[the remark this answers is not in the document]
alexkuzmik: Skipping — this isn't reachable in practice. The multi-select "All projects" option expands to the explicit list of every project id (see `LoadableSelectBox.handleSelectAll`), never `[]`. `projectIds: []` therefore only occurs when the user deselects everything, where the "configure a project" empty state is the intended UX. Routing `[]` to the workspace endpoint would make a fully-deselected widget silently query the entire workspace, which is worse — so the empty-array → not-configured mapping is kept on purpose.

CometActions:

### Access Information
- **URL:** https://pr-7365.dev.comet.com
- **Namespace:** pr-7365
- **Version:** 2.1.18-7365-merge-2485
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7365%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

CometActions:

### Access Information
- **URL:** https://pr-7365.dev.comet.com
- **Namespace:** pr-7365
- **Version:** 2.1.18-7365-merge-2486
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7365%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7365`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

CometActions:

### Access Information
- **URL:** https://pr-7365.dev.comet.com
- **Namespace:** pr-7365
- **Version:** 2.1.18-7365-merge-2491
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7365%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

apps/opik-frontend/src/api/projects/useWorkspaceMetric.ts
```diff
@@ -0,0 +1,78 @@
...
+  const processedFilters = processFiltersArray(spanFilters ?? []);
+  const { data } = await api.post<ProjectMetricsResponse>(
+    `${WORKSPACES_REST_ENDPOINT}metrics/spans`,
+    {
+      metric_type: metricName,
+      interval,
+      ...(intervalStart && { interval_start: intervalStart }),
+      ...(intervalEnd && { interval_end: intervalEnd }),
+      // Empty => all projects in the workspace; otherwise only the selected set
+      ...(projectIds.length > 0 && { project_ids: projectIds }),
```
awkoy, line 55: **🟠 Medium — "All projects" is a frozen snapshot, not dynamic.** This comment + `types/dashboard.ts` document *empty ⇒ all projects*, and the backend honors it, but the UI can never produce empty: `LoadableSelectBox`'s "Select all" expands to an explicit array of the **currently-loaded** project IDs (`ProjectsSelectBox` loads 1000 by default, 10000 after "load more"). Consequences:

- **Staleness:** a widget saved as "All projects" freezes the ID list at click time; projects created later are silently excluded — which directly undercuts the PR's headline use case (tracking *total* workspace span usage / rate limits).
- **Silent cap:** workspaces with >1000 projects only aggregate the first 1000, with no warning.
- The dynamic empty ⇒ all path (here + `resolveProjectSelection` returning `{}` for `[]`) is effectively **unreachable from the UI**.

Suggest: when select-all is active, persist a sentinel (empty array, or an explicit `allProjects: true` flag) and route that to the backend's dynamic path, rather than freezing the loaded IDs. At minimum, surface the cap/staleness to the user.

apps/opik-frontend/src/v1/pages-shared/dashboards/widgets/ProjectStatsCardWidget/ProjectStatsCardEditor.tsx
```diff
@@ -71,29 +86,60 @@ const ProjectStatsCardEditor = forwardRef<WidgetEditorHandle>((_, ref) => {
...
   const { data, isPending } = useTracesOrSpansScoresColumns(
     {
       projectId,
       type: source,
     },
-    {},
+    { enabled: !isMultiProject && !!projectId },
+  );
+
+  const { data: tokenUsageNamesData } = useProjectTokenUsageNames(
```
awkoy, line 125: **🟡 Low — usage-key options come from only the first selected project.** `projectId` here is the representative project (`localProjectIds[0]`), but the metric aggregates across *all* selected projects. So a usage key that exists only in another project won't appear in this dropdown. The stored value still works (the backend aggregates regardless of what's listed), so this is UX-only — but a user with heterogeneous providers per project may not find the key they want. (Applies to the v2 copy too.)

apps/opik-frontend/src/v1/pages-shared/dashboards/widgets/ProjectStatsCardWidget/ProjectStatsCardWidget.tsx
```diff
@@ -127,7 +148,22 @@ const ProjectStatsCardWidget: React.FunctionComponent<
...
+    {
+      enabled:
+        !isMultiProject && source === TRACE_DATA_TYPE.spans && !!projectId,
+    },
+  );
+
+  const workspaceMetric = useWorkspaceMetric(
+    {
+      projectIds: projectIds ?? [],
+      metricName: (metric as METRIC_NAME_TYPE) ?? METRIC_NAME_TYPE.SPAN_COUNT,
```
awkoy: **🟡 Low — unreachable fallback.** This query is only `enabled` when `workspaceMetricDef` exists, which forces `metric === SPAN_TOKEN_USAGE`, so the `?? SPAN_COUNT` fallback can never actually be sent. It reads as if `SPAN_COUNT` is a supported workspace metric. Prefer dropping the fallback (or asserting the invariant) to avoid confusion. (Same in the v2 copy.)

apps/opik-frontend/src/v1/pages-shared/dashboards/widgets/ProjectMetricsWidget/ProjectMetricsEditor.tsx
```diff
@@ -327,13 +349,38 @@ const ProjectMetricsEditor = forwardRef<WidgetEditorHandle>((_, ref) => {
...
   };
 
-  const handleProjectChange = (projectId: string) => {
-    updatePreviewWidget({
-      config: {
-        ...config,
-        projectId,
-      },
-    });
+  const handleProjectsChange = (projectIds: string[]) => {
```
awkoy, line 385: **🟡 Low — config↔form dual-write duplicated 4× (drift risk).** This handler mirrors every field twice (`nextConfig.X = …` **and** `form.setValue("X", …)`) for ~7 fields, and the block is copied across both editors × v1/v2 = 4 near-identical copies. Easy for one copy to forget a field (e.g. `usageMetrics` vs `usageMetric`). Consider extracting a shared `resetConfigForMultiProject(config, form)` helper into `lib/dashboard/workspaceMetrics.ts`.

apps/opik-frontend/src/v1/pages-shared/dashboards/widgets/ProjectMetricsWidget/ProjectMetricsEditor.tsx
```diff
@@ -327,13 +349,38 @@ const ProjectMetricsEditor = forwardRef<WidgetEditorHandle>((_, ref) => {
...
+    delete nextConfig.projectId;
+    form.setValue("projectIds", projectIds);
+
+    // Aggregating across projects only supports span metrics. If the selection becomes multi-project while a
+    // non-span metric is chosen, fall back to span token usage and clear metric-specific selections.
+    const becomesMultiProject = isMultiProjectSelection(
+      runtimeContext.projectId,
+      projectIds,
+    );
+    if (becomesMultiProject && !isWorkspaceMetric(metricType)) {
```
awkoy: **⚪ Nit — asymmetry vs the stats-card editor.** `ProjectStatsCardEditor.handleProjectsChange` clears a workspace-only metric when dropping back to a single project; this editor has no reverse branch. It's actually **benign** (`SPAN_TOKEN_USAGE` is also valid for a single project, so there's nothing to clear), but the asymmetry looks like an omission. A one-line comment here would save the next reader from "fixing" a non-bug.

awkoy (commented):
FE-only review (frontend files). Overall the frontend work is clean and idiomatic — shared logic is correctly extracted into `lib/dashboard/workspaceMetrics.ts` + `api/projects/useMetricData.ts` (not duplicated into v1/v2), `useMetricData` keeps both hooks unconditional and gates with `enabled` (correct `isPending` semantics), the react-query v5 "disabled query is pending" trap is handled with `!isMultiProject && isPending`, and legacy single-`projectId` widgets stay backward-compatible. No runtime-breaking bugs found and no real perf gaps (query keys are content-stable, so no refetch storms).

Inline findings below, by severity. The one I'd want resolved before merge is the first ("All projects" snapshot). Most findings apply identically to the v1 and v2 copies — I commented on v1 to avoid noise.

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7365`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

CometActions:

### Access Information
- **URL:** https://pr-7365.dev.comet.com
- **Namespace:** pr-7365
- **Version:** 2.1.18-7365-merge-2512
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7365%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

apps/opik-backend/src/main/java/com/comet/opik/domain/WorkspaceMetricsDAO.java
```diff
@@ -209,8 +230,207 @@ STEP toIntervalDay(1)
...
+                        model,
+                        provider,
+                        type
+                        <endif>
+                    FROM spans FINAL
+                    <if(feedback_scores_empty_filters)>
+                    LEFT JOIN fsc ON fsc.entity_id = spans.id
+                    <endif>
+                    WHERE workspace_id = :workspace_id
+                    <if(project_ids)> AND project_id IN :project_ids <endif>
```
thiagohora: **✅ Addressed in revision `82138c9b`** (see approval review).

The service now resolves empty `projectIds` → the explicit workspace project set, so the DAO always runs `project_id IN (...)`. This prunes well for small/moderate selections. Caveat re-validated on prod: for a genuine "all projects" request on a large workspace, `IN(all N projects)` reads ~the same granules as the old unbounded scan (the `id` window can't prune at PK level across many project prefixes) — left as a non-blocking note on `WorkspaceMetricsService`/`WorkspaceSpanMetricRequest`.

---

_Original finding (pre-fix):_

The spans PK is `(workspace_id, project_id, trace_id, parent_span_id, id)`. With no `project_id` predicate, `id` (5th col) can't prune at the PK level; only the `idx_spans_id` minmax partially helps. Prod EXPLAIN (large ws, 7d): all-projects 38,567 granules / ~300M+ rows vs single-project 23 granules — ~1,677×.

apps/opik-frontend/src/v2/pages-shared/dashboards/widgets/ProjectMetricsWidget/MetricChart/MetricChartContainer.tsx
```diff
@@ -133,10 +139,7 @@ const MetricContainerChart = ({
       breakdown,
       logsSource,
     },
-    {
-      enabled: !!projectId,
-      refetchInterval: 30000,
-    },
+    { refetchInterval: 30000 },
```
thiagohora, line 142: This `refetchInterval: 30000` combined with the workspace query amplifies the scan concern flagged in `WorkspaceMetricsDAO`: a multi-project / all-projects time-series tile re-runs a heavy (potentially full-tenant) span aggregation **every 30s per widget**. Worth confirming the product expectation, and/or gating the auto-refetch for the workspace path.

apps/opik-frontend/src/v2/pages-shared/dashboards/widgets/ProjectMetricsWidget/ProjectMetricsEditor.tsx
```diff
@@ -419,10 +466,13 @@ const ProjectMetricsEditor = forwardRef<WidgetEditorHandle>((_, ref) => {
                     className={cn("flex-1", {
                       "border-destructive": Boolean(validationErrors?.message),
                     })}
-                    value={field.value || ""}
+                    multiselect
+                    showSelectAll
+                    selectAllLabel="All projects"
```
thiagohora: **🟠 "All projects" select-all can silently under-select.** `LoadableSelectBox.handleSelectAll` sets the value to the explicit IDs of the **currently-loaded** options only — capped at `DEFAULT_LOADED_PROJECT_ITEMS = 1000` (or `MORE_LOADED_PROJECT_ITEMS = 10000` after "load more"). In a workspace with more projects than are loaded, clicking "All projects" aggregates only that subset while the tile still reads as "All projects", so the total is silently wrong. Please either load the full project set when select-all is used, or send the empty-array ("all") signal to the backend instead of an explicit truncated list. (Same in v1.)

apps/opik-frontend/src/v2/pages-shared/dashboards/widgets/ProjectStatsCardWidget/ProjectStatsCardWidget.tsx
```diff
@@ -151,8 +187,50 @@ const ProjectStatsCardWidget: React.FunctionComponent<
...
+          <div className="w-full space-y-3">
+            <Skeleton className="h-8 w-32" />
+            <Skeleton className="h-16 w-full" />
+          </div>
+        </div>
+      );
+    }
+
+    const results = workspaceMetric.data?.results ?? [];
+    // Token usage returns one series per usage key; pick the selected key. Count/cost return a single series.
```
thiagohora: 🟡 Stale/misleading comment: only `SPAN_TOKEN_USAGE` is supported across projects (`SUPPORTED_SPAN_METRICS = EnumSet.of(SPAN_TOKEN_USAGE)`, `WORKSPACE_METRIC_NAMES = [SPAN_TOKEN_USAGE]`). The "Count/cost return a single series" reference doesn't correspond to any currently-supported metric — please drop it to avoid implying count/cost work here. (Same in v1.)

apps/opik-backend/src/main/java/com/comet/opik/domain/WorkspaceMetricsDAO.java
```diff
@@ -209,8 +230,207 @@ STEP toIntervalDay(1)
             ;
             """;
 
+    // Copied verbatim from ProjectMetricsDAO.SPAN_FILTERED_PREFIX. The only change is the project predicate:
+    // the per-project `AND project_id = :project_id` becomes an optional IN-list so the same query can span a set
+    // of projects (or, when project_ids is absent, every project in the workspace). workspace_id stays mandatory.
+    private static final String SPAN_FILTERED_PREFIX = """
```
thiagohora: 🟡 DRY: this ~130-line `SPAN_FILTERED_PREFIX` is a verbatim copy of `ProjectMetricsDAO.SPAN_FILTERED_PREFIX` with a single-line change (project predicate). Two copies of a complex feedback-dedup + span-filter CTE will drift over time (a fix to one side silently misses the other). Consider extracting the shared prefix into a common constant/template where the project predicate is parameterized (`= :project_id` vs `IN :project_ids`), so both DAOs stay in sync.

(Note: keeping `FROM spans FINAL` here is correct — summing `usage` over a ReplacingMergeTree needs dedup to avoid double-counting un-merged span versions, and it preserves per-project parity. Just flagging it's intentional, unlike the recent FINAL removal in the experiment-refs query.)

thiagohora (changes requested):
Thanks — this is well-structured and the test coverage (parity, breakdown, isolation, filters, validation) is strong. Requesting changes primarily on one query-performance issue I validated with `EXPLAIN indexes = 1`.

**Blocker**
- The all-projects path (empty `project_ids`) does an unbounded full-workspace span scan — the `id` time-bound can't prune granules because `id` is the 5th PK column and `project_id`/`trace_id`/`parent_span_id` are unconstrained. Measured 335/335 granules · 2.53M rows · 212 MiB for all-projects vs 6/335 · 43.6K rows for a single project, independent of the selected date range. Details + fix suggestions inline on `WorkspaceMetricsDAO`. The 30s `refetchInterval` on the time-series tile compounds this.

**Should fix**
- "All projects" select-all sends only the currently-loaded project IDs (cap 1000 / 10000), so it silently under-counts on large workspaces while still labeled "All projects" (inline on `ProjectMetricsEditor`).

**Nits**
- Verbatim ~130-line duplication of `SPAN_FILTERED_PREFIX` from `ProjectMetricsDAO` (drift risk) — inline suggestion to share it.
- Stale "count/cost" comment in the stat card (only `SPAN_TOKEN_USAGE` is supported).

Security & isolation look good (workspace_id always bound from request context, parameterized `project_ids`/filters, permission annotation present, cross-workspace isolation test passes).

thiagohora (commented):
Re-validated the all-projects scan concern on a large production workspace (EXPLAIN indexes only — reads no data). Numbers inline on `WorkspaceMetricsDAO`. Net: the `idx_spans_id` minmax index partially helps, but the all-projects path is still ~1,677× heavier than single-project on the same window and reads hundreds of millions of rows — the earlier blocker stands.

apps/opik-backend/src/main/java/com/comet/opik/domain/WorkspaceMetricsDAO.java
alexkuzmik: Fixed in f80318650e — the all-projects path no longer does an unconstrained scan. `WorkspaceMetricsService` now resolves an empty `project_ids` to the explicit set of every project id in the workspace (`ProjectService.findProjectIdsByWorkspace`, a cheap `SELECT id FROM projects WHERE workspace_id = …` in MySQL) before calling the DAO, so the query always runs with a bounded `project_id IN (…)` that prunes on the spans PK — never `workspace_id` + `id`-range alone. A workspace with no projects short-circuits to an empty result, and the DAO now asserts `projectIds` is non-empty so the unbounded branch can't be reached even via the API.

apps/opik-frontend/src/v2/pages-shared/dashboards/widgets/ProjectMetricsWidget/MetricChart/MetricChartContainer.tsx
alexkuzmik: Fixed in f80318650e — this is resolved by the backend change on `WorkspaceMetricsDAO`: the workspace/all-projects path is now always a bounded `project_id IN (…)` query (the service resolves "all" to the explicit project set up front), so the 30s `refetchInterval` re-runs a PK-pruned query per widget, not a full-tenant span scan. Left the interval as-is since it now matches the per-project tile cost.

apps/opik-frontend/src/api/projects/useWorkspaceMetric.ts
alexkuzmik: Fixed in f80318650e — "All projects" is now a dynamic, server-resolved selection rather than a frozen snapshot. The editor stores an explicit `allProjects` flag (a dedicated "All projects in the workspace" checkbox), and the widget sends the empty "all" signal to the backend, which resolves the full project set at query time. So newly created projects are included automatically, and the `resolveProjectSelection` empty path is now reachable and correct. Kept it distinct from the empty/not-configured state (which stays "configure a project").

apps/opik-frontend/src/v2/pages-shared/dashboards/widgets/ProjectMetricsWidget/ProjectMetricsEditor.tsx
alexkuzmik: Fixed in f80318650e — "All projects" no longer sends the capped/loaded id list. It's now an explicit `allProjects` flag that sends the empty "all" signal; the backend resolves it to every project in the workspace at query time, so there's no silent under-count on workspaces with more than `DEFAULT_LOADED_PROJECT_ITEMS` projects. The generic select-all on the project field was removed in favor of the dedicated checkbox to avoid the truncated-list footgun. (Same in v1.)

apps/opik-backend/src/main/java/com/comet/opik/domain/WorkspaceMetricsDAO.java
```diff
@@ -209,8 +230,207 @@ STEP toIntervalDay(1)
...
+            WITH feedback_scores_deduped AS (
```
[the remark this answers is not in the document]
alexkuzmik: Fixed in f80318650e — the ~130-line `SPAN_FILTERED_PREFIX` is now a single shared constant in `SpanMetricsQueries`, consumed by both `ProjectMetricsDAO` and `WorkspaceMetricsDAO` via `spanFilteredPrefix(projectPredicate)`. The only per-DAO difference — `project_id = :project_id` vs `project_id IN :project_ids` — is injected at class-load time, so the CTE can't drift between the two paths anymore.

apps/opik-backend/src/main/java/com/comet/opik/domain/WorkspaceMetricsDAO.java
alexkuzmik: Fixed in f80318650e — extracted the shared CTE into `SpanMetricsQueries.spanFilteredPrefix(projectPredicate)`; `ProjectMetricsDAO` passes `project_id = :project_id` and `WorkspaceMetricsDAO` passes `project_id IN :project_ids`, so the two copies are now one source. And confirmed — `FROM spans FINAL` is intentional here (summing `usage` over the ReplacingMergeTree needs dedup + keeps per-project parity), so it's preserved.

apps/opik-backend/src/main/java/com/comet/opik/domain/WorkspaceMetricsDAO.java
```diff
@@ -247,6 +606,137 @@ public Mono<List<WorkspaceMetricResponse.Result>> getCostsDaily(@NonNull Workspa
...
+            // SPAN_TOKEN_USAGE breakdown sums a single token-usage entry, selected by its name (sub_metric)
+            if (request.hasBreakdown() && request.metricType() == MetricType.SPAN_TOKEN_USAGE) {
+                statement.bind("sub_metric", Optional.ofNullable(request.breakdown().subMetric()).orElse(""));
+            }
+
+            Optional.ofNullable(request.filters())
+                    .ifPresent(filters -> {
+                        FilterQueryBuilder.bind(statement, filters, FilterStrategy.SPAN);
+                        FilterQueryBuilder.bind(statement, filters, FilterStrategy.SPAN_FEEDBACK_SCORES);
+                        FilterQueryBuilder.bind(statement, filters, FilterStrategy.SPAN_FEEDBACK_SCORES_IS_EMPTY);
```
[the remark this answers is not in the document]
alexkuzmik: Fixed in f80318650e — the span filter strategies are now declared once, in `SPAN_FILTER_TEMPLATE_PLACEHOLDERS` (strategy → template placeholder). Both the template `add` pass and the statement `bind` pass iterate that single map, so a strategy can't be added on one side but not the other.

apps/opik-frontend/src/v1/pages-shared/dashboards/widgets/ProjectStatsCardWidget/ProjectStatsCardEditor.tsx
```diff
@@ -204,51 +321,42 @@ const ProjectStatsCardEditor = forwardRef<WidgetEditorHandle>((_, ref) => {
...
+          />
+        )}
 
         <ProjectWidgetFiltersSection
           control={form.control}
-          fieldName={isTraceSource ? "traceFilters" : "spanFilters"}
+          fieldName={useSpanFilters ? "spanFilters" : "traceFilters"}
           projectId={projectId}
-          filterType={isTraceSource ? "trace" : "span"}
+          filterType={useSpanFilters ? "span" : "trace"}
```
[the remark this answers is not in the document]
alexkuzmik: Fixed in f80318650e — `currentFilters`, `filterKey` and `clearErrors` now key off `useSpanFilters` (the value that actually decides which field the section renders), not `isTraceSource`. So in multi-project + trace-source mode, editing span filters updates the watched length and clears the stale `spanFilters` error. Applied to both the v1 and v2 editor copies.

apps/opik-frontend/src/v1/pages-shared/dashboards/widgets/ProjectStatsCardWidget/ProjectStatsCardWidget.tsx
alexkuzmik: Fixed in f80318650e — dropped the `?? SPAN_COUNT` fallback. `metricName` is now just `metric as METRIC_NAME_TYPE`, with a comment noting the query only runs when a supported workspace metric is set (`workspaceMetricDef`), so SPAN_COUNT no longer reads as a supported workspace metric. (Same in v1 + v2.)

apps/opik-frontend/src/v2/pages-shared/dashboards/widgets/ProjectStatsCardWidget/ProjectStatsCardWidget.tsx
alexkuzmik: Fixed in f80318650e — dropped the misleading "Count/cost return a single series" reference; the comment now reads "Metrics without a usage key use the first." Only `SPAN_TOKEN_USAGE` is supported across projects. (Same in v1 + v2.)

apps/opik-frontend/src/v1/pages-shared/dashboards/widgets/ProjectMetricsWidget/ProjectMetricsEditor.tsx
alexkuzmik: Fixed in f80318650e — added a one-line comment in the metrics editor's `handleProjectsChange` explaining the asymmetry is intentional: SPAN_TOKEN_USAGE is also a valid per-project metric, so there is nothing workspace-only to clear when dropping back to a single project (unlike the stats-card editor, whose workspace metrics differ).

apps/opik-frontend/src/v1/pages-shared/dashboards/widgets/ProjectStatsCardWidget/ProjectStatsCardEditor.tsx
alexkuzmik: Kept as-is (UX-only), with a small hardening. As you noted, the stored value still works since the backend aggregates across all selected projects regardless of what the dropdown lists — so I didn't build cross-project usage-key aggregation here (would be a larger change; can follow up). In the new "All projects" mode there's no representative project at all, so I made `usageKeyOptions` always include the currently-selected key, so the field shows its value instead of appearing empty. (Same in v1 + v2.)

apps/opik-frontend/src/v1/pages-shared/dashboards/widgets/ProjectMetricsWidget/ProjectMetricsEditor.tsx
alexkuzmik: Partially addressed in f80318650e. Within each editor the config↔form dual-write is now funneled through a single local helper (`resetToWorkspaceMetric` in the metrics editor, `enterMultiProjectMetric` in the stats-card editor), which also backs the new "All projects" handler — so there's one place per editor instead of duplicated inline blocks. I didn't extract a single cross-editor `resetConfigForMultiProject` into `lib/dashboard`: the two editors reset different field sets (metrics has durationMetrics/usageMetrics/breakdown; the stat card has usageMetric), so a shared helper would need per-editor field lists and wouldn't remove much. Happy to revisit if you'd prefer the shared version.

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7365`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

apps/opik-backend/src/main/java/com/comet/opik/domain/WorkspaceMetricsService.java
```diff
@@ -62,4 +73,59 @@ public Mono<WorkspaceMetricResponse> getWorkspaceCosts(@NonNull WorkspaceMetricR
...
+                .flatMap(resolved -> CollectionUtils.isEmpty(resolved.projectIds())
+                        ? Mono.just(List.<WorkspaceMetricResponse.Result>of())
+                        : dispatch(resolved))
+                .map(results -> WorkspaceMetricResponse.builder()
+                        .results(results)
+                        .build());
+    }
+
+    // "All projects" (empty projectIds) is resolved into the explicit set of workspace project ids so the DAO always
+    // queries a bounded `project_id IN (...)` list that prunes on the spans primary key, never an unconstrained
```
thiagohora: Re-validated on prod (large ws, 327 projects, `EXPLAIN indexes=1`, 1-day window): resolving to `project_id IN (all N)` does **not** prevent a workspace-wide read — IN(327) → 23,545 granules after `idx_spans_id` vs 23,691 for the old workspace-only scan (essentially identical). With a large IN-set the generic-exclusion search can't push the `id` range across the disjoint project prefixes, so PK pruning collapses to the workspace slice and only `idx_spans_id` prunes the window. Small selections do prune well (IN(3)→17, IN(20)→89). Please soften this comment: it's bounded to the workspace's *projects*, but for a large tenant that IS a full-workspace span scan (~hundreds of M rows), refetched every 30s. Non-blocking.

apps/opik-backend/src/main/java/com/comet/opik/api/metrics/WorkspaceSpanMetricRequest.java
```diff
@@ -0,0 +1,46 @@
...
+import java.time.Instant;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.UUID;
+
+/**
+ * Request for workspace-level span metrics aggregated across projects. When {@code projectIds} is empty, the service
+ * resolves it to every project in the workspace before querying, so the aggregation always runs against an explicit,
+ * bounded project set (never an unconstrained workspace-wide span scan); otherwise only the given projects are used.
```
thiagohora: Same caveat as on `WorkspaceMetricsService`: "(never an unconstrained workspace-wide span scan)" is misleading for large workspaces. Resolving to `IN(all project ids)` reads ~the same granules as a workspace scan (verified on prod: IN(327) ≈ workspace-only) because the `id`/time window can't prune at the PK across many project prefixes. Suggest rewording to note it's bounded to the workspace's *projects* but remains full-scan-equivalent for tenants with many projects.

thiagohora (dismissed):
Approving — this revision resolves the substantive review points:

- **DRY**: `SPAN_FILTERED_PREFIX` extracted into a shared `SpanMetricsQueries` with a parameterized project predicate; per-project parity preserved.
- **"All projects" select-all**: replaced the truncated multiselect-select-all with a dynamic, server-resolved `allProjects` signal (empty `project_ids` → resolved to the workspace's projects). Fixes the >1000-project under-count *and* the frozen-snapshot staleness.
- **Stale count/cost comment**: removed.

One non-blocking caveat I re-validated on prod (EXPLAIN indexes only, reads no data), left inline: resolving "all projects" to `project_id IN (all N projects)` does **not** materially cheapen the query on a large workspace — it reads ~the same granules as an unbounded workspace scan, because with many project_id values the `id`/time window can't prune at the PK (only the `idx_spans_id` minmax helps, as before). It prunes well for small selections (IN(3)→17, IN(20)→89 granules; IN(327)→23,545 ≈ workspace-only 23,691). This is largely inherent to the table layout, so not blocking — but please (1) soften the two comments claiming it avoids a workspace-wide scan, and (2) consider gating the 30s refetch / requiring a bounded window for the `allProjects` path on large tenants. Nice work.

apps/opik-backend/src/main/java/com/comet/opik/domain/WorkspaceMetricsService.java
alexkuzmik: Reworded in 43267760e1. The comment now states the query is bounded to the workspace's projects but that for a tenant with many projects `IN(<all ids>)` reads roughly the same granules as a full workspace scan (id/time window can't prune at the PK across disjoint project prefixes), so it's full-scan-equivalent in that case rather than "never an unconstrained workspace-wide scan".

apps/opik-backend/src/main/java/com/comet/opik/api/metrics/WorkspaceSpanMetricRequest.java
alexkuzmik: Reworded in 43267760e1. Dropped the "(never an unconstrained workspace-wide span scan)" phrasing; the Javadoc now notes the aggregation is bounded to the workspace's projects but remains full-scan-equivalent for tenants with many projects, since `project_id IN (<all ids>)` can't prune the id/time window at the PK across many disjoint project prefixes.

CometActions:

The deployment encountered an error. Please check the [deployment logs](https://github.com/comet-ml/comet-deployment/actions/runs/29248126137) for details.

CometActions:

The deployment encountered an error. Please check the [deployment logs](https://github.com/comet-ml/comet-deployment/actions/runs/29250093723) for details.

PR #7366, opened 2026-07-06
chore(dev): cap ClickHouse memory to 55% of RAM in docker-compose

aadereiko:
## What
Adds `max_server_memory_usage_to_ram_ratio = 0.55` to the local docker-compose ClickHouse config.

## Why
The local ClickHouse container shares a small Docker VM (~8 GiB) with MySQL/Redis/Minio/Zookeeper/python-backend, but ClickHouse defaults to reserving `0.9 * RAM` for itself. A query spike then OOM-kills the container (exit 137), taking the local stack down. Capping to ~55% makes a heavy query error out instead, leaving headroom for the rest of the services.

Local-dev only (docker-compose `clickhouse_config`); no effect on prod/Helm.

### DIFF
--- a/deployment/docker-compose/clickhouse_config/additional_config.xml
+++ b/deployment/docker-compose/clickhouse_config/additional_config.xml
@@ -4,6 +4,13 @@
          (dev-runner with TOGGLE_AGENT_INSIGHTS_ENABLED=true, or prod via OPIK-6846). No effect otherwise. -->
     <custom_settings_prefixes>SQL_</custom_settings_prefixes>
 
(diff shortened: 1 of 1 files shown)


aadereiko:
Closing — this was a local-dev workaround for a small (8 GiB) Docker VM that was OOM-killing ClickHouse. That VM has since been bumped to 31 GiB, so the guard is no longer needed, and capping `max_server_memory_usage_to_ram_ratio` to 0.55 globally would needlessly limit ClickHouse for devs with normally-sized VMs. Not worth carrying as a shared config change.

PR #7367, opened 2026-07-06
[OPIK-7253] fix: request identity encoding on all react-service auth calls

thiagohora:
## Details
Auth calls to the react service (`/opik/auth`, `/opik/auth-session`, `/opik/auth-by-username`, `workspaces/workspace-id`) can fail with `java.util.zip.ZipException: Not in GZIP format` while reading a successful (2xx) response whose `Content-Encoding: gzip` header doesn't match the actual body bytes (gzip double-decompression / mislabelled encoding). The exception is a `ProcessingException` rather than a `ClientErrorException`, so it escapes the auth filter's error handling and surfaces to the caller as an HTTP 500 on regular API traffic (e.g. span/trace ingestion). On the API-key path it also fires before the credentials cache is populated (`authenticateUsingApiKey` caches only after `verifyResponse` succeeds), so an affected API key fails on **every** request instead of once.

- `listEligibleWorkspaces` already carries the workaround (`.acceptEncoding("identity")`, "avoids gzip double-decompression issue"); this PR applies the identical workaround to the remaining five react-service call sites in `RemoteAuthService`.
- Auth responses are small JSON payloads, so skipping gzip has no meaningful bandwidth cost.
- Note: existing tests never caught this because `config-test.yml` sets `jerseyClient.gzipEnabled: false` (production default is `true`), so the shared test client never registers `GZipDecoder`. The new regression test builds a production-like client with `gzipEnabled: true`.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves OPIK-7253

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Fable 5
- Scope: Investigated the production error (log/stack-trace analysis), authored the change, the regression test, and this PR description.
- Human verification: pending reviewer verification; change mirrors the existing precedent in `listEligibleWorkspaces`.

## Testing
- New regression test `RemoteAuthServiceTest#testAuth__whenRemoteMislabelsGzipResponses__thenIdentityEncodingAvoidsZipException`: builds a client with production gzip settings (`gzipEnabled: true`) and stubs the react service to mislabel responses as `Content-Encoding: gzip` (plain body) unless the client requests identity encoding — mirroring the observed upstream behavior.
  - **Without the fix**: the test fails with `jakarta.ws.rs.ProcessingException: java.util.zip.ZipException: Not in GZIP format` through `RemoteAuthService.authenticateUsingApiKey` — the exact production stack.
  - **With the fix**: passes.
- `mvn test -Dtest=RemoteAuthServiceTest`: 29/29 pass.
- `mvn compile` and `mvn spotless:check` on `apps/opik-backend`: pass.
- Verified in a cloud install that the failing path is `verifyResponse` → `readEntity` on 2xx responses, matching the stack trace, and that error frequency correlates with auth-cache misses.

## Documentation
N/A

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
@@ -221,6 +221,8 @@ class RemoteAuthService implements AuthService {
                 .path("auth-by-username")
                 .request()
                 .accept(MediaType.APPLICATION_JSON)
+                // avoids gzip double-decompression issue, same as in listEligibleWorkspaces
+                .acceptEncoding("identity")
                 .header(OAUTH_USERNAME_HEADER, token.userName())
                 .post(Entity.json(AuthRequest.builder()
                         .workspaceName(token.workspaceName())
@@ -244,6 +246,8 @@ class RemoteAuthService implements AuthService {
                 .path("auth-session")
                 .request()
                 .accept(MediaType.APPLICATION_JSON)
+                // avoids gzip double-decompression issue, same as in listEligibleWorkspaces
+                .acceptEncoding("identity")
                 .cookie(sessionToken)
                 .post(Entity.json(AuthRequest.builder().workspaceName(workspaceName).build()))) {
             var authResponse = verifyResponse(response);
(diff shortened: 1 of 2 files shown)


LifeXplorer (dismissed):
Looks good!

PR #7371, opened 2026-07-06
[OPIK-7256] fix(backend): make project size-estimate cache actually engage

thiagohora:
## Details

The per-project size estimate (`WorkspaceMetadataDAO.getProjectMetadata` — a 1000-span size sample plus `uniq(id)` over the whole project) is meant to be cached, but the cache never engaged, so the query ran on essentially every traces/spans find/search request and became one of the top read-volume queries on the ClickHouse cluster (~65 executions/min). Two stacked defects:

- `@Cacheable(name = "project_metadata", ...)` sits on a **private** method, and Guice method interception (which implements `@Cacheable`) cannot intercept private methods — the annotation was silently dead. Made the method package-private with a comment explaining why it must not be private.
- The cache name `project_metadata` was absent from `cacheManager.caches`; the config only carried an orphaned `workspace_metadata` entry that no code references (stale name), so even an intercepted call would have used the 1-second `defaultDuration`. Renamed the entry to `project_metadata` (env `CACHE_MANAGER_PROJECT_METADATA_DURATION`, default PT1H, matching the working `experiment_metadata` sibling).

With the cache engaged, per-(workspace, project) executions drop from once per request to once per TTL window.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-7256

## AI-WATERMARK

AI-WATERMARK: yes

- If yes:
  - Tools: Claude Code
  - Model(s): Claude Fable 5
  - Scope: visibility fix, config rename, commit and PR text
  - Human verification: pending reviewer; production query-frequency numbers measured read-only from system.query_log

## Testing

- Added `WorkspaceMetadataServiceCacheTest`: boots the app harness with a counting-stub `WorkspaceMetadataDAO` and asserts a repeated `getProjectMetadata` call for the same key is served from the cache (DAO hit exactly once), and a different project misses. The test fails if the method becomes non-interceptable again (e.g. private) or if the `project_metadata` cache name is dropped from configuration.
- `mvn compile` and `mvn test-compile` pass; `mvn spotless:apply` produced no further changes
- Guice interception limitation is documented behavior (private methods are never intercepted); the sibling `experiment_metadata` cache — public method + configured name — demonstrates the working pattern this change aligns with
- Production `system.query_log` (7 days): ~657K executions/week of this query with per-request cardinality, confirming the cache was not engaging

## Documentation

No user-facing documentation impact. The new env var `CACHE_MANAGER_PROJECT_METADATA_DURATION` (replacing the orphaned `CACHE_MANAGER_WORKSPACE_METADATA_DURATION`) is documented inline in `config.yml`.

**Note on the sample query itself:** we evaluated also reworking the sample CTE's `ORDER BY (sorting key) DESC LIMIT 1000` (it reads the whole project because reverse-order read never engages for this shape), but dropping the ORDER BY would pin the sample to the oldest ~1000 spans instead of the latest, so the estimate would stop tracking current span sizes. Keeping the query as-is; with the cache engaged it runs at most once per TTL per project.

### DIFF
--- a/apps/opik-backend/config.yml
+++ b/apps/opik-backend/config.yml
@@ -1021,7 +1021,11 @@ cacheManager:
     # Default: {}
     # Description: Dynamically created caches with their respective time to live in seconds
     automationRules: ${CACHE_MANAGER_AUTOMATION_RULES_DURATION:-PT1S}
-    workspace_metadata: ${CACHE_MANAGER_WORKSPACE_METADATA_DURATION:-PT1H}
+    # Default: PT1H
+    # Description: Cache TTL for the per-project size estimate (ScopeMetadata) gating dynamic sorting.
+    # The estimate scans all of the project's spans, so cache misses are expensive by design.
+    # Renamed from the orphaned workspace_metadata entry, which no code referenced.
+    project_metadata: ${CACHE_MANAGER_PROJECT_METADATA_DURATION:-PT1H}
     # Default: PT5M
     # Description: Cache TTL for workspace version determination results, applied per workspace
     # Scope: Per workspace (cache key includes workspace ID)
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/workspaces/WorkspaceMetadataService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/workspaces/WorkspaceMetadataService.java
@@ -44,8 +44,10 @@ class WorkspaceMetadataServiceImpl implements WorkspaceMetadataService {
                 .flatMap(resolvedProjectId -> getProjectMetadata(workspaceId, resolvedProjectId));
     }
 
+    // Package-private, not private: Guice method interception (which implements @Cacheable) cannot
+    // intercept private methods, so a private modifier silently disables the cache.
     @Cacheable(name = "project_metadata", key = "'-'+ $workspaceId + '-' + $projectId", returnType = ScopeMetadata.class)
-    private Mono<ScopeMetadata> getProjectMetadata(String workspaceId, UUID projectId) {
+    Mono<ScopeMetadata> getProjectMetadata(String workspaceId, UUID projectId) {
         return workspaceMetadataDAO.getProjectMetadata(workspaceId, projectId);
     }
 
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/workspaces/WorkspaceMetadataService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/workspaces/WorkspaceMetadataService.java
@@ -44,10 +44,10 @@ class WorkspaceMetadataServiceImpl implements WorkspaceMetadataService {
                 .flatMap(resolvedProjectId -> getProjectMetadata(workspaceId, resolvedProjectId));
     }
 
-    // Package-private, not private: Guice method interception (which implements @Cacheable) cannot
+    // Must not be private: Guice method interception (which implements @Cacheable) cannot
     // intercept private methods, so a private modifier silently disables the cache.
     @Cacheable(name = "project_metadata", key = "'-'+ $workspaceId + '-' + $projectId", returnType = ScopeMetadata.class)
-    Mono<ScopeMetadata> getProjectMetadata(String workspaceId, UUID projectId) {
+    public Mono<ScopeMetadata> getProjectMetadata(String workspaceId, UUID projectId) {
         return workspaceMetadataDAO.getProjectMetadata(workspaceId, projectId);
     }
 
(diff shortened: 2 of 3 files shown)


apps/opik-backend/src/test/java/com/comet/opik/domain/workspaces/WorkspaceMetadataServiceCacheTest.java
```diff
@@ -0,0 +1,129 @@
...
+        var workspaceId = UUID.randomUUID().toString();
+        var projectId = UUID.randomUUID();
+
+        var first = impl.getProjectMetadata(workspaceId, projectId).block();
+
+        assertThat(first).isNotNull();
+        assertThat(PROJECT_METADATA_DAO_CALLS.get()).isEqualTo(1);
+
+        // wait for the cache put async to complete
+        waitForMillis(CACHE_ASYNC_WAIT_MILLIS);
```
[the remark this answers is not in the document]
thiagohora: Agreed — the interceptor drops the `putAsync` completion signal, so the fixed sleep was a race. Fixed in 7ebbfd2: the test now injects `CacheManager` and polls `contains()` on the exact cache key (`project_metadata:--{workspaceId}-{projectId}`, matching the interceptor's `name:-evaluatedKey` composition) with a bounded 5s Awaitility wait before the second lookup. As a bonus this also asserts the entry actually lands in Redis under the configured cache name, not just that the DAO was skipped.

apps/opik-backend/src/main/java/com/comet/opik/domain/workspaces/WorkspaceMetadataService.java
```diff
@@ -44,8 +44,10 @@ public Mono<ScopeMetadata> getProjectMetadata(@NonNull String workspaceId, UUID
                 .flatMap(resolvedProjectId -> getProjectMetadata(workspaceId, resolvedProjectId));
     }
 
+    // Must not be private: Guice method interception (which implements @Cacheable) cannot
+    // intercept private methods, so a private modifier silently disables the cache.
```
andrescrz, line 48: Nit: better as javadoc

apps/opik-backend/src/test/java/com/comet/opik/domain/workspaces/WorkspaceMetadataServiceCacheTest.java
```diff
@@ -0,0 +1,134 @@
...
+                        }))
+                        .customConfigs(
+                                List.of(
+                                        new CustomConfig("cacheManager.enabled", "true"),
+                                        new CustomConfig("cacheManager.caches.project_metadata", "PT30S")))
+                        .build());
+    }
+
+    @Test
+    void getProjectMetadata__repeatedCallsAreServedFromCache(WorkspaceMetadataService service,
```
andrescrz, line 104: Minor: we have similar caching tests already in the code base. We typically:
1. Tweak the cache period to make it shorter and reliable for the test
2. Produce the cached value
3. Alter the cache value.
4. Retrieve the old value from cache.
5. Wait for cache expiration.
6. Retrieve the updated value. 

This is what we generally do instead of bypassing the internal implementation and rely on counters.

Anyway, not a blocker. Suggesting this to favour black box tests that are closer to reality, instead of bypassing internal implementations.

apps/opik-backend/config.yml
```diff
@@ -1021,7 +1021,11 @@ cacheManager:
     # Default: {}
     # Description: Dynamically created caches with their respective time to live in seconds
     automationRules: ${CACHE_MANAGER_AUTOMATION_RULES_DURATION:-PT1S}
-    workspace_metadata: ${CACHE_MANAGER_WORKSPACE_METADATA_DURATION:-PT1H}
+    # Default: PT1H
+    # Description: Cache TTL for the per-project size estimate (ScopeMetadata) gating dynamic sorting.
+    # The estimate scans all of the project's spans, so cache misses are expensive by design.
+    # Renamed from the orphaned workspace_metadata entry, which no code referenced.
+    project_metadata: ${CACHE_MANAGER_PROJECT_METADATA_DURATION:-PT1H}
```
andrescrz, line 1028: Nit: updating the environment var name would affect existing installations that override it. Anyhow, I'd rather favour correctness here. We can always publish some upgrade notes, if we really want to be strict. I doubt someone tweaked this and the cache was not working anyway.

andrescrz (approved):
Just a few thoughts, but this LGTM.

apps/opik-backend/src/main/java/com/comet/opik/domain/workspaces/WorkspaceMetadataService.java
thiagohora: Done in follow-up #7380 — converted to javadoc.

apps/opik-backend/src/test/java/com/comet/opik/domain/workspaces/WorkspaceMetadataServiceCacheTest.java
thiagohora: Reworked in follow-up #7380 to that flow: produce the cached value, alter the underlying value, assert the stale value is still served, then assert the updated value after the (shortened, 1s) TTL expires — counters removed; the stub DAO now just exposes a settable value. One deviation kept: a bounded Awaitility poll on the cache entry before altering the value, instead of a fixed sleep — the interceptor's async put has no completion signal, and altering the value before the put lands would make the stale-value assertion racy in the opposite direction.

apps/opik-backend/config.yml
thiagohora: Agreed on favoring correctness — noted in #7380's description for the record: any installation overriding the old env var was tuning a cache that never engaged, so the rename can only improve behavior there. Happy to add upgrade notes if we want to be strict.

PR #6076, opened 2026-04-04
[OPIK-5659] [BE] fix: prevent concurrent experiment aggregation and reduce ClickHouse query overload

thiagohora:
## Details

Experiment aggregation was generating 2,000+ ClickHouse queries/minute under load. The root cause was a mismatch between the aggregation lock TTL (1 minute) and the actual time required to process large experiments. When the lock expired, multiple nodes reprocessed the same experiment concurrently, each issuing the full set of ClickHouse queries per batch. A subsequent load test against 2 experiments with 1M items each (5M traces, 15M spans, 8M feedback scores) surfaced a client-side INSERT scaling wall that was resolved in this PR.

### Root cause — lock TTL vs processing time:

- Per-batch cost: ~1.6 s (fetch items) + ~3.3 s (parallel span / trace / feedback / comments / assertions) + ~0.3 s (INSERT) ≈ **5.2 s/batch** at batchSize=1,000.
- 1M-item experiment → 1,000 batches × 5.2 s = **~87 minutes**, against a **1 minute** lock TTL.
- Lock expired → other nodes acquired it and started over → **2,000+ queries/minute**.

### Original concurrency/contention fixes

1. **Fail-fast on lock contention**: switched from `executeWithLockCustomExpire` (blocks until TTL, enables concurrent processing) to `bestEffortLock` with `lockAcquireWait=500ms` — if the lock is held, the message is immediately ack'd without blocking worker threads.
2. **Actual cancellation on timeout**: applied `.timeout(lockTTL)` to the processing Mono. Reactor cancels the upstream R2DBC subscription, stopping in-flight ClickHouse queries when the TTL elapses — instead of leaving them running after the lock silently expires.
3. **Retry logic with max attempts**: on timeout the experiment is re-published via the debounce publisher. A per-experiment Redis \`RAtomicLong\` counter caps retries at \`maxLockExpiryRetries=3\`; resets on success.

### Query-level redundancies removed

4. **\`getProjectId\` resolved once per experiment**: was called inside the per-batch loop — for 1,000 batches that's 1,000 redundant ClickHouse round trips returning the same value. Now resolved once before the \`expand()\` loop and threaded through.
[...]

PR #7376, opened 2026-07-07
[OPIK-6890] [BE] feat: wire deletion-events bridge into traces delete paths

andrescrz:
## Details

Wires the deletion-events bridge into the trace delete paths. Every user-initiated trace delete (single, batch, and thread) now records the deleted trace ids in the `deletion_events_local` ClickHouse bridge table, so deletes issued during the data-model migration's backfill/cutover window are not lost from the table copy (mitigates R-19). Capture is gated by a new feature flag (default off), runs after the delete, and is best-effort — its failure is logged and swallowed so it can never disrupt a delete.

- New `DeletionEventDAO` owns insert/read for `deletion_events_local`; `DeletionEvent` and `DeletionReason` model the rows.
- `TraceService` captures deletions after a successful `dao.delete`, per batch, only when capture is enabled; capture is a cold `Mono.defer`, so nothing runs unless the delete succeeds.
- Feature flag `databaseAnalyticsDataModel.traceDeletionEventsCaptureEnabled` (env `ANALYTICS_DB_DATA_MODEL_TRACE_DELETION_EVENTS_CAPTURE_ENABLED`, default `false`) — turn on only for the backfill-to-cutover window.
- Deliberate scope decisions (differ from the literal ticket wording): traces only; retention paths (`deleteForRetention*`) are intentionally not captured because retention is disabled during the migration window; sequencing is delete-first + best-effort rather than bridge-before-LWD, so capture can never disrupt a delete.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves OPIK-6890

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Scope: full implementation (production + tests) with iterative code review
- Human verification: code review + local test runs

## Testing

Command: `cd apps/opik-backend && mvn test -Dtest='TraceServiceImplTest,TraceDeletionEventTest'` — all green.

- Integration (black-box via `TraceResourceClient`, capture enabled): single `DELETE /traces/{id}`, batch `POST /traces/delete` (project-scoped and unscoped, parameterized), and thread `POST /traces/threads/delete` each assert the exact `deletion_events_local` rows written (full object comparison; `event_time` within 2s of now). Covers the `project_id null ↔ ""` sentinel both directions and the multi-row batch insert.
- Unit (`TraceServiceImplTest.DeleteTraces`): capture disabled → no capture; capture enabled + insert fails → delete still succeeds (best-effort); delete fails → no capture, no `TracesDeleted` published, error propagates.
- Full quality suite (compile, tests, spotless) runs in CI.

## Documentation

N/A — no product documentation changes. The new feature flag is documented inline in `config.yml`.


apps/opik-backend/src/main/java/com/comet/opik/domain/TraceService.java
```diff
@@ -476,23 +480,62 @@ public Mono<Void> delete(@NonNull Set<UUID> ids, UUID projectId) {
...
+                    .flatMap(batch -> {
+                        var batchIds = Set.copyOf(batch);
+                        return dao.delete(batchIds, projectId, connection)
+                                .doOnSuccess(_ -> {
                                     eventBus.post(TracesDeleted.builder()
-                                            .traceIds(Set.copyOf(batch))
+                                            .traceIds(batchIds)
                                             .projectId(projectId)
                                             .workspaceId(workspaceId)
                                             .userName(userName)
```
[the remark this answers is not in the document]
andrescrz: Skipping — two parts:

1) `flatMap` on the shared connection: this batching is pre-existing (unchanged from `main`; this PR only appends the capture step inside the existing lambda), and it only produces more than one batch above `ANALYTICS_DELETE_BATCH_SIZE = 10000` ids. The capture step doesn't add contention on that connection — it runs on its own pooled connection via `TransactionTemplateAsync.nonTransaction`, not the shared `connection` passed to `dao.delete`. Out of scope for this bridge-wiring PR; can be a separate ticket if the >10k-id concurrency proves to be a real issue.

2) Gating capture on a non-zero affected-row count: intentionally not done. The bridge's design principle is "rather a redundant event than a missed one", and replay is idempotent (re-deleting a non-existent id is a no-op), so recording an id whose delete matched 0 rows is harmless. More importantly, ClickHouse does not reliably report affected rows for lightweight DELETE (the same reason `insert` returns `Mono<Void>`), so gating on that count would frequently see 0 and silently skip capture — losing events, which is the exact R-19 failure this bridge exists to prevent.

apps/opik-backend/src/main/java/com/comet/opik/domain/DeletionEventDAO.java
```diff
@@ -0,0 +1,129 @@
...
+    @VisibleForTesting
+    Flux<DeletionEvent> findBySourceTableAndDeletedIds(String sourceTable, @NonNull Set<String> deletedIds) {
+        var details = "source_table=%s, deleted_ids_size=%s".formatted(sourceTable, deletedIds.size());
+        var template = getSTWithLogComment(
+                FIND_BY_SOURCE_TABLE_AND_DELETED_IDS, "find_deletion_events", null, null, details);
+        return templateAsync.stream(connection -> Flux.from(connection.createStatement(template.render())
+                .bind("source_table", sourceTable)
+                .bind("deleted_ids", deletedIds)
+                .execute())
+                .flatMap(result -> result.map((row, _) -> map(row))));
```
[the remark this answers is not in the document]
andrescrz: Skipping — `findBySourceTableAndDeletedIds` is a `@VisibleForTesting`, package-private helper whose only caller is the integration test; there is no production read path for the bridge (replay is migration SQL, not this DAO), so no tenant-facing query can spill. The test reads back by globally-unique v7 ids, so cross-tenant id collision can't occur.

Beyond that, `deletion_events_local` isn't a typical customer-data table with per-workspace auth and a workspace-scoped sort key — it's a system-wide capture log for the migration, ordered by `(source_table, event_time, deleted_id)`. `workspace_id` isn't part of the primary key, so adding `AND workspace_id = :workspace_id` wouldn't help the query plan, and scoping reads by workspace doesn't fit the table's purpose.

apps/opik-backend/src/main/java/com/comet/opik/domain/DeletionEvent.java
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * A single row of the deletion-events bridge table. {@code eventTime} is null on the insert path (ClickHouse
+ * stamps it) and {@code projectId} is null for workspace-scoped source tables; {@code deletedId} is a plain
+ * string because it is not a UUID for every source table.
+ */
+@Builder(toBuilder = true)
+public record DeletionEvent(
+        Instant eventTime,
+        @NonNull String sourceTable,
```
thiagohora: Shouldn't it be an enum?

apps/opik-backend/src/main/java/com/comet/opik/domain/DeletionEventDAO.java
```diff
@@ -0,0 +1,129 @@
...
+            """;
+
+    private final @NonNull TransactionTemplateAsync templateAsync;
+
+    /**
+     * Inserts the given events, letting ClickHouse stamp {@code event_time} via the column default.
+     * {@code userName} tags the query log with the user whose request triggered the capture. Returns no count:
+     * ClickHouse does not report affected rows for {@code INSERT}.
+     */
+    public Mono<Void> insert(Set<DeletionEvent> events, String userName) {
```
thiagohora, line 84: Do we control the size of this Set? I'm asking because if it exceeds 2K or 5K, we start having issues with the bind, as we did during the experiment aggregation https://github.com/comet-ml/opik/pull/6076.

thiagohora (dismissed):
Just 2 small comments

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceService.java
```diff
@@ -476,23 +480,62 @@ public Mono<Void> delete(@NonNull Set<UUID> ids, UUID projectId) {
...
                                             .build());
                                     log.info(
                                             "Published TracesDeleted event for trace ids count '{}' for project_id '{}' on workspace '{}'",
-                                            batch.size(), projectId, workspaceId);
-                                }))
-                        .then());
+                                            batchIds.size(), projectId, workspaceId);
+                                })
+                                .then(captureDeletions(batchIds, projectId, workspaceId, userName));
```
BorisTkachenko, line 501: Why are we doing it here on the main path instead of using EventBus to do it in async manner? We already have `TracesDeleted` event that holds all needed data.

BorisTkachenko (commented):
Looks good, just one question.

apps/opik-backend/src/main/java/com/comet/opik/domain/DeletionEvent.java
andrescrz: Good point. Since this PR only wires traces, I'd initially gone with a static constant and deferred modelling it. But with spans (and the other source tables) coming next, an enum is the better fit — so I've switched `sourceTable` to a `SourceTable` enum, mirroring `DeletionReason`, which also validates the value on read-back via `fromStringOrThrow`. Done in 44005bd86a.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceService.java
andrescrz: We considered exactly that — the bus here is an `AsyncEventBus` and `TracesDeleted` already carries the ids/project/workspace/user. We kept capture synchronous on the delete path deliberately, for stronger guarantees: it's the migration's safety net against deletes being lost across the table copy (R-19), so we want it attempted within the delete's own execution rather than after a separate async dispatch, which would widen the window where a delete commits but its capture doesn't. It stays best-effort (failures are logged and swallowed via `onErrorResume`) so it never disrupts the delete. If delete-path latency ever becomes a concern, moving it to a `TracesDeleted` listener is a clean follow-up.

apps/opik-backend/src/main/java/com/comet/opik/domain/DeletionEventDAO.java
```diff
@@ -0,0 +1,149 @@
...
+        return Flux.fromIterable(batches)
+                .concatMap(batch -> insert(batch, userName))
+                .then();
+    }
+
+    private Mono<Void> insert(List<DeletionEvent> events, String userName) {
+        var first = events.getFirst();
+        log.info("Inserting deletion events batch, size '{}' for source_table '{}' project_id '{}' on workspace '{}'",
+                events.size(), first.sourceTable().getValue(), first.projectId(), first.workspaceId());
+        var details = "source_table=%s, project_id=%s".formatted(first.sourceTable().getValue(), first.projectId());
```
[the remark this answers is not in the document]
andrescrz: Skipping — this is intentional. The per-batch INFO line is what surfaces the batching/chunking behaviour (batch sizes, number of chunks) at the DAO level, while the service-level `Captured …` line is a coarser lifecycle signal — the two complement each other. Capture is flag-gated, so these only log during the migration window, where the per-component visibility is worth having; there's no steady-state noise once the flag is off. Keeping them at INFO on purpose.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceService.java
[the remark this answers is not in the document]
andrescrz: Skipping — deliberate, documented scope decision. This PR wires the bridge for user-initiated trace deletes only; retention (`deleteForRetention*`) is intentionally not captured because retention is disabled operationally during the migration window (see the PR description and plan §8.7). A code-enforced gate (disable retention while capture is on) or capturing retention are reasonable belt-and-suspenders follow-ups, but both are out of scope for this PR.

thiagohora (approved):

BorisTkachenko (approved):

PR #7384, opened 2026-07-07
[OPIK-7262] [BE] observability: metrics for project last-updated flush + events consumed by listeners

thiagohora:
## Details

Adds observability for the project last-updated-trace buffering feature (from #7361) and a general "events consumed" metric for all event listeners. No behavior change.

- **Buffer/flush metrics** (meter `opik.project_last_updated_flush`):
  - `.buffered` (counter, `result=ok|error`) — markers buffered to Redis on the ingestion path; surfaces the previously log-only, fail-open Redis write failures.
  - `.markers` (counter) — markers processed per flush.
  - `.run` (counter, `result=success|skipped_lock|error`) and `.duration` (histogram, ms) — flush job cycle outcome and latency.
- **`opik.event.consumed`** (counter, tag `listener`) in `EventInterceptor` — counts every event successfully consumed by a listener, complementing the existing `opik.event.error`.

Follows the existing OTel idiom (`RetentionSlidingWindowJob`, `MetricsAlertJob`).

> Stacked on top of #7361 (base branch = `thiagoh/OPIK-7246-project-marker-redis-flush`); the flush/buffer code it instruments lives there. Rebase to `main` once #7361 merges.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-7262

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: Implementation of the metrics + tests.
- Human verification: Author reviewed; builds and unit tests pass (see Testing).

## Testing

- `mvn compiler:compile compiler:testCompile` — BUILD SUCCESS.
- `mvn surefire:test` on `ProjectLastUpdatedTraceBufferServiceTest` + `ProjectLastUpdatedFlushJobTest` — 10/10 pass (existing behaviour unaffected by the added metrics).
- Metrics emission itself is not unit-tested (consistent with the sibling metered jobs, which rely on the OTel runtime).

## Documentation

No user-facing docs. Metrics are self-describing via their OTel descriptions.


apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ProjectLastUpdatedFlushJob.java
```diff
@@ -38,12 +44,17 @@ public class ProjectLastUpdatedFlushJob extends Job implements InterruptableJob
 
     private static final Lock FLUSH_LOCK = new Lock("project_last_updated_flush_job:scan_lock");
 
+    private static final AttributeKey<String> RESULT_KEY = stringKey("result");
+
```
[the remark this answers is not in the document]
thiagohora: Not changing this — it matches the established convention in this codebase: 14 other classes (RetentionEstimationJob, RetentionCatchUpJob, RetentionSlidingWindowJob, StreamConsumerReaper, the *ProjectMigrationService family, FreeFormSqlQueryService, OnlineScorePublisher, ...) each independently declare their own `RESULT_KEY = stringKey("result")`. Extracting a shared constant for just these two would diverge from that pattern rather than reduce duplication app-wide.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ProjectLastUpdatedFlushJob.java
```diff
@@ -62,18 +85,27 @@ public void doJob(JobExecutionContext context) {
...
         Disposable disposable = lockService.bestEffortLock(
                 FLUSH_LOCK,
                 Mono.fromCallable(bufferService::flush)
                         .subscribeOn(Schedulers.boundedElastic())
                         .doOnNext(count -> {
+                            runCounter.add(1, Attributes.of(RESULT_KEY, "success"));
+                            runDuration.record(System.currentTimeMillis() - startMs);
                             if (count > 0) {
                                 log.info("Project last-updated flush processed '{}' project markers", count);
                             }
```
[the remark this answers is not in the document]
thiagohora: Agreed this pattern repeats, but it's pre-existing app-wide (RetentionCatchUpJob, RetentionEstimationJob, RetentionSlidingWindowJob all wire runCounter/runDuration/success|error|skipped_lock independently, predating this PR). Extracting a shared `LockingJobMetricsRecorder` is a reasonable idea but it's a cross-cutting refactor of jobs this PR doesn't touch — better scoped as its own follow-up than smuggled into an observability PR for one job.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ProjectLastUpdatedFlushJob.java
```diff
@@ -62,18 +85,27 @@ public void doJob(JobExecutionContext context) {
...
+                        .doOnError(error -> {
+                            runCounter.add(1, Attributes.of(RESULT_KEY, "error"));
+                            runDuration.record(System.currentTimeMillis() - startMs);
+                        })
                         .then(),
-                Mono.fromRunnable(() -> log.debug(
-                        "Another instance is flushing project last-updated markers, skipping")),
+                Mono.fromRunnable(() -> {
+                    runCounter.add(1, Attributes.of(RESULT_KEY, "skipped_lock"));
+                    log.debug("Another instance is flushing project last-updated markers, skipping");
```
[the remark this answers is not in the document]
thiagohora: Checked the actual sibling lock-based jobs (RetentionCatchUpJob, RetentionEstimationJob) — neither records runDuration on their skipped_lock branch either, so this matches established convention, not a regression. It's also arguably correct: the duration histogram should reflect actual flush work, not lock-contention misses (which are already fully captured via runCounter{result=skipped_lock}). Leaving as-is.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/events/EventInterceptor.java
```diff
@@ -56,6 +62,7 @@ public Object invoke(MethodInvocation methodInvocation) {
 
         try (Scope scope = span.makeCurrent()) {
             Object result = methodInvocation.proceed();
+            consumedCounter.add(1, Attributes.of(LISTENER_KEY, listenerName));
             log.info("Event intercepted: {}", event);
```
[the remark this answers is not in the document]
thiagohora: Confirmed real — TraceThreadListener.onTracesCreated, TraceDeletedListener.onTracesDeleted, and EnvironmentAutoCreateListener.onTracesCreated all end in a fire-and-forget `.subscribe()` on a reactive chain inside a void method, so the interceptor only observes the synchronous dispatch, not completion. Fixed in f3da2023c by renaming to `opik.event.invoked` with a description that says exactly that (safe rename — the metric is new/unmerged, nothing depends on the old name). Didn't attempt the doOnSuccess/doOnError hook since these listener methods return void (already subscribed internally), so there's no async handle for the interceptor to attach to without restructuring listener signatures — out of scope for a no-behavior-change observability PR.

apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectLastUpdatedTraceBufferService.java
```diff
@@ -73,11 +95,17 @@ public void record(@NonNull String workspaceId, Collection<ProjectIdLastUpdated>
...
-                project.lastUpdatedAt().toEpochMilli(), member(workspaceId, project.id())));
+        // The metric observes the outcome, but the exception itself still propagates to the caller (the event bus),
+        // which owns the handling — Redis failures are no longer swallowed here (see #7361 review feedback).
+        try {
+            RScoredSortedSet<String> pending = redisClient.getScoredSortedSet(PENDING_SET_KEY);
+            lastUpdatedTraces.forEach(project -> pending.addIfGreater(
+                    project.lastUpdatedAt().toEpochMilli(), member(workspaceId, project.id())));
+            bufferedRecords.add(lastUpdatedTraces.size(), Attributes.of(RESULT_KEY, "ok"));
+        } catch (RuntimeException e) {
+            bufferedRecords.add(lastUpdatedTraces.size(), Attributes.of(RESULT_KEY, "error"));
```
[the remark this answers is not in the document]
thiagohora: Good catch — fixed in dea0680aa. Switched to per-item counting: each marker is now recorded exactly once by its actual outcome, so a mid-batch `addIfGreater` failure counts already-buffered markers as `ok`, the failing one as `error`, and the un-attempted remainder not at all (we rethrow on first failure). No more whole-batch mislabeling in either direction.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ProjectLastUpdatedFlushJob.java
```diff
@@ -52,6 +63,18 @@ public ProjectLastUpdatedFlushJob(
         this.config = config;
         this.bufferService = bufferService;
         this.lockService = lockService;
+
+        var meter = GlobalOpenTelemetry.get().getMeter("opik.project_last_updated_flush");
+        this.runCounter = meter
+                .counterBuilder("opik.project_last_updated_flush.run")
```
andrescrz: Nit: You could use `METER_NAME` from below and avoid repeating `opik.project_last_updated_flush` here multiple times.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ProjectLastUpdatedFlushJob.java
```diff
@@ -62,18 +85,27 @@ public void doJob(JobExecutionContext context) {
...
+                            runCounter.add(1, Attributes.of(RESULT_KEY, "error"));
+                            runDuration.record(System.currentTimeMillis() - startMs);
+                        })
                         .then(),
-                Mono.fromRunnable(() -> log.debug(
-                        "Another instance is flushing project last-updated markers, skipping")),
+                Mono.fromRunnable(() -> {
+                    runCounter.add(1, Attributes.of(RESULT_KEY, "skipped_lock"));
+                    log.debug("Another instance is flushing project last-updated markers, skipping");
+                }),
```
andrescrz, line 104: Minor: histogram metrics also count. In my view `runCounter` overlaps. This is the only place where you don't measure the time, probably to not to skew the success and error cases. You can:
1. Keep the separated counter.
2. Remove metrics from this particular case.
3. Keep the histogram everywhere and remove the counter.

apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectLastUpdatedTraceBufferService.java
```diff
@@ -60,6 +72,16 @@ public ProjectLastUpdatedTraceBufferServiceImpl(
         this.config = config;
         this.redisClient = redisClient;
         this.projectService = projectService;
+
+        var meter = GlobalOpenTelemetry.get().getMeter(METER_NAME);
+        this.bufferedRecords = meter
+                .counterBuilder("opik.project_last_updated_flush.buffered")
```
andrescrz: Nit: same here, you repeat `opik.project_last_updated_flush` instead of reusing `METER_NAME`.

apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectLastUpdatedTraceBufferService.java
```diff
@@ -103,6 +134,9 @@ public long flush() {
                 written += flushBatch(flushing, entries);
             }
         } while (pageSize == batchSize);
+        if (written > 0) {
+            flushedMarkers.add(written);
```
andrescrz, line 139: Minor: double check here if you want to measure the counter (monotonic) or if a guage is more appropiate.

andrescrz (dismissed):
Only some minor comments, but not blockers.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ProjectLastUpdatedFlushJob.java
thiagohora: Done in bc78f9fd — added a `METER_NAME` constant and build `getMeter(METER_NAME)` + the metric name from it (`METER_NAME + ".duration"`).

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ProjectLastUpdatedFlushJob.java
thiagohora: Went with option 3 in bc78f9fd — removed `runCounter` and now record the `runDuration` histogram in all three branches tagged `result=success|error|skipped_lock`. The histogram's per-attribute `_count` gives run counts by result (so no overlap), and tagging keeps skipped_lock in its own series so it doesn't skew the success/error latency — which was the original reason it was counter-only.

apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectLastUpdatedTraceBufferService.java
thiagohora: Done in bc78f9fd — `METER_NAME` already existed here; now the metric names reuse it too (`METER_NAME + ".buffered"`, `METER_NAME + ".markers"`).

apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectLastUpdatedTraceBufferService.java
thiagohora: Counter is intended here — `.markers` is the cumulative total of markers persisted over time (monotonic, ever-increasing), so rate() gives flush throughput. A gauge would imply a point-in-time level (e.g. current buffer depth), which isn't what this measures. Clarified the description to "Total ... processed" + a code comment in bc78f9fd.

andrescrz (approved):
LGTM.

PR #7390, opened 2026-07-07
[OPIK-6891] [BE] test: guard the trace deletion-events bridge with ArchUnit

andrescrz:
## Details

Adds an ArchUnit architectural guard for the deletion-events bridge: only `TraceServiceImpl` may call `TraceDAO.delete`, so a future refactor that adds a new trace-delete path can't silently skip the bridge capture (the ticket's core intent). This is the net-new piece for OPIK-6891; the per-path bridge-assertion tests and payload correctness already shipped with OPIK-6890.

- New `TraceDeletionEventArchTest` — fails the build if any class other than `TraceServiceImpl` calls `TraceDAO.delete(Set, UUID, Connection)`. Verified it actually catches a bypass (temporary probe in a real service failed the rule, then removed).
- Introduces ArchUnit (`archunit-junit5`) as a test dependency. Note: pinned to **1.4.1** because the project compiles to Java 25 and older ArchUnit (1.3.0) couldn't parse Java 25 bytecode (it silently failed to import every class, making the rule vacuously pass).
- Scope for this ticket, consistent with the merged OPIK-6890 design and Slice 1:
  - Covered: bypass guard + (via OPIK-6890) every wired trace user-delete path captures, `user_request` reason, and payload matches the deleted ids/table.
  - Intentionally not here: bridge-before-LWD sequencing assertion (OPIK-6890 uses delete-first + best-effort by design); spans delete paths (not wired yet — a sibling rule lands with the spans slice); `retention`/`cascade` flavors (not wired); the F1 cutover `apply_deleted_mask` guard and the full end-to-end cutover-replay test (Slice 2 per the plan).

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves OPIK-6891

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Scope: full implementation (ArchUnit guard + dependency) with iterative verification
- Human verification: code review + local test runs

## Testing

Command: `cd apps/opik-backend && mvn test -Dtest='TraceDeletionEventArchTest'` — passes (0 class-import failures under Java 25 with ArchUnit 1.4.1).

- Positive: rule passes green against current `main` (only `TraceServiceImpl` calls `TraceDAO.delete`).
- Negative (guard actually guards): temporarily added a bypassing caller in a real service — the rule failed with the exact violation (`… calls method TraceDAO.delete …`) — then removed it.
- Full quality suite (compile, tests, spotless) runs in CI.

## Documentation

N/A — internal test-only change; no product documentation impact.

### DIFF
--- a/apps/opik-backend/pom.xml
+++ b/apps/opik-backend/pom.xml
@@ -452,6 +452,12 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>com.tngtech.archunit</groupId>
+            <artifactId>archunit-junit5</artifactId>
+            <version>1.4.2</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>uk.co.jemos.podam</groupId>
             <artifactId>podam</artifactId>
             <version>8.0.2.RELEASE</version>
--- a/apps/opik-backend/src/test/java/com/comet/opik/domain/TraceDeletionEventArchTest.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/domain/TraceDeletionEventArchTest.java
@@ -1,0 +1,38 @@
+package com.comet.opik.domain;
+
+import com.tngtech.archunit.core.importer.ImportOption;
+import com.tngtech.archunit.junit.AnalyzeClasses;
+import com.tngtech.archunit.junit.ArchTest;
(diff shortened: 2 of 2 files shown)


BorisTkachenko (approved):
LGTM

PR #7400, opened 2026-07-08
[OPIK-6686] [BE] fix: scope prompt name uniqueness to project

BorisTkachenko:
## Details

Prompt name uniqueness was enforced at the workspace level, so creating a prompt with a name that already existed in a different project failed with `Prompt id or name already exists`. The application already scopes prompt lookups by `project_id`; this change aligns the database constraint with that behavior by replacing the `(workspace_id, name)` unique index with `(workspace_id, project_id, name)`, so names only need to be unique within a project.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-6686

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: Liquibase migration authoring and test coverage
- Human verification: Author to review and run backend build/tests

## Testing

- Added `PromptResourceTest` case verifying prompts with the same name in different projects both succeed.
- Updated existing conflict tests to pin prompts to a shared project so the intended within-project duplicate conflict is still exercised.
- Backend build and test suite to be run by the author (`mvn compile -DskipTests && mvn test && mvn spotless:check`).

## Documentation

N/A

### DIFF
--- a/apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000089_update_prompts_unique_constraint.sql
+++ b/apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000089_update_prompts_unique_constraint.sql
@@ -1,8 +1,8 @@
 --liquibase formatted sql
 --changeset boryst:000089_update_prompts_unique_constraint
 
-ALTER TABLE prompts DROP INDEX `prompts_workspace_id_name_uk`;
 ALTER TABLE prompts ADD CONSTRAINT `prompts_workspace_id_project_id_name_uk` UNIQUE (workspace_id, project_id, name);
+ALTER TABLE prompts DROP INDEX `prompts_workspace_id_name_uk`;
 
---rollback ALTER TABLE prompts DROP INDEX `prompts_workspace_id_project_id_name_uk`;
 --rollback ALTER TABLE prompts ADD CONSTRAINT `prompts_workspace_id_name_uk` UNIQUE (workspace_id, name);
+--rollback ALTER TABLE prompts DROP INDEX `prompts_workspace_id_project_id_name_uk`;
--- a/apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000089_update_prompts_unique_constraint.sql
+++ b/apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000089_update_prompts_unique_constraint.sql
@@ -1,0 +1,8 @@
+--liquibase formatted sql
+--changeset boryst:000089_update_prompts_unique_constraint
+
+ALTER TABLE prompts DROP INDEX `prompts_workspace_id_name_uk`;
+ALTER TABLE prompts ADD CONSTRAINT `prompts_workspace_id_project_id_name_uk` UNIQUE (workspace_id, project_id, name);
+
+--rollback ALTER TABLE prompts DROP INDEX `prompts_workspace_id_project_id_name_uk`;
+--rollback ALTER TABLE prompts ADD CONSTRAINT `prompts_workspace_id_name_uk` UNIQUE (workspace_id, name);
--- a/apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/PromptResourceTest.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/PromptResourceTest.java
@@ -1087,7 +1087,7 @@ class PromptResourceTest {
                     .projectId(null)
                     .build();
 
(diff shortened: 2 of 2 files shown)


apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000089_update_prompts_unique_constraint.sql
```diff
@@ -0,0 +1,8 @@
+--liquibase formatted sql
+--changeset boryst:000089_update_prompts_unique_constraint
+
+ALTER TABLE prompts DROP INDEX `prompts_workspace_id_name_uk`;
+ALTER TABLE prompts ADD CONSTRAINT `prompts_workspace_id_project_id_name_uk` UNIQUE (workspace_id, project_id, name);
```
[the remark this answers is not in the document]
BorisTkachenko: For the recent migration, each prompt must have a valid project ID, not null. So, we can ignore the cases where it's null. We don't care about the legacy stuff.

andrescrz: You need to revert the order here:
1. Create new constraint.
2. Drop the old one.

The opposite for rollback.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/PromptResourceTest.java
```diff
@@ -1087,7 +1087,7 @@ Stream<Arguments> when__promptIsInvalid__thenReturnError() {
                     .projectId(null)
                     .build();
 
-            Prompt duplicatedPrompt = buildPrompt().build();
+            Prompt duplicatedPrompt = buildPrompt().projectName("conflict-test-project").build();
```
andrescrz: Nit: not sure why not relaying on the Podam random project names in this whole PR. The result should be the same, with less code and following the conventions (over configuration).

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/PromptResourceTest.java
```diff
@@ -1117,6 +1117,22 @@ Stream<Arguments> when__promptIsInvalid__thenReturnError() {
...
+
+            var prompt1 = buildPrompt().name(sharedName).projectName("project-alpha").build();
+            var prompt2 = buildPrompt().name(sharedName).projectName("project-beta").build();
+
+            var id1 = createPrompt(prompt1, API_KEY, TEST_WORKSPACE);
+            var id2 = createPrompt(prompt2, API_KEY, TEST_WORKSPACE);
+
+            assertThat(id1).isNotNull();
+            assertThat(id2).isNotNull();
+            assertThat(id1).isNotEqualTo(id2);
```
andrescrz: Minor: fill free to also retrieve and assert the full prompt payload, for a full round trip and more endurable test.

andrescrz (commented):
I believe the order of the migration needs to be changed. After that, some minor thoughts on tests.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000089_update_prompts_unique_constraint.sql
BorisTkachenko: Fixed in 821673d — reordered so the new `(workspace_id, project_id, name)` constraint is added first and the old `(workspace_id, name)` index dropped afterward; rollback reversed accordingly (add old, then drop new), so a uniqueness constraint is always present during the swap.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/PromptResourceTest.java
BorisTkachenko: Fixed in 821673d — replaced the hardcoded project-name literals across these tests with `factory.manufacturePojo(String.class)` (Podam random names).

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/PromptResourceTest.java
BorisTkachenko: Fixed in 821673d — the test now creates both projects and asserts each prompt via the canonical `getPromptAndAssert(...)` helper, which retrieves and recursively compares the full payload for a complete round trip.

andrescrz (approved):

PR #7404, opened 2026-07-08
[OPIK-6884] [BE] perf: full-scan audit Slice-1 remediations (R4, R6, R7, R8)

thiagohora:
## Details

Closes the self-contained ("mechanical") Bucket-1 remediations from the OPIK-6883 full-scan query audit, hardening `traces`/`spans` query paths ahead of the planned weekly time-partitioning + hot/cold tiering. This is one increment of the rolling OPIK-6884 story; the structural items (R1/R2/R3) remain as separate follow-ups and are not required for Slice-1 close.

- **R4 — `traces`/`spans` deletes** (`TraceService`, `TraceDAO`): the delete cascade accepted a nullable `project_id` (delete-by-id passes null; a batch may span projects), so it dropped the `project_id` predicate and could no longer prune on the `(workspace_id, project_id)` sorting-key prefix. `TraceService.delete` now resolves each trace's owning project and deletes **per project group** with a concrete `project_id`, which flows through the existing `TracesDeleted` event to the span/feedback cascade unchanged (no event-contract change). Resolution uses a new partition-bounded lookup (`getProjectIdsByTraceIdsBounded`) that bounds `toMonday(id_at)` by the id set's own min/max — exact because `id_at` is `MATERIALIZED UUIDv7ToDateTime(toUUID(id))` (migration 000091). Prod `EXPLAIN` confirms this prunes both the trace delete (58 granules) and, crucially, the span cascade (`SELECT_SPAN_IDS_BY_TRACE_ID`: 4,787 → 164 granules, ~29×, since `project_id` is the spans' 2nd PK column).
  - **Deletion-events bridge (merged from main):** the bridge captures the `project_id` on the delete path, so by-id / unscoped deletes now record the *resolved* project instead of `null` — `TraceDeletionEventTest`'s two unscoped expectations were updated accordingly (needs a nod from the bridge owner). This relies on the invariant that a span always shares its trace's project; two `TracesResourceTest` cascade tests that created a span in a *different* project than its trace were corrected to use the trace's project.
- **R7 — retention velocity estimate** (`SpanDAO.ESTIMATE_VELOCITY_FOR_RETENTION`): added a lower bound (`trace_id >= :lower_bound`) set to the catch-up service-start cursor, so the estimate scans `[service_start, cutoff)` instead of all pre-cutoff history. Makes the primary estimate path consistent with the scouting fallback, which already floors at the service-start date.
- **R6 — experiment project mapping** (`ExperimentDAO.COMPUTE_EXPERIMENT_PROJECT_MAPPING`): the query joined `experiment_items` to `traces` with only a `workspace_id` bound on `traces`, so the `ConcurrentHashJoin` read the entire workspace trace slice into the hash table. Added `AND t.id IN (SELECT trace_id FROM experiment_items WHERE workspace_id = :workspace_id)`, which gives `traces` an `id` primary-key condition (the set is implied by the join `ei.trace_id = t.id`, so the mapping result is unchanged) and bounds the read to the referenced trace ids. Verified on prod (ClickHouse 25.8.16 Altinity, large workspace) with `EXPLAIN indexes=1` / `EXPLAIN ESTIMATE`: the `traces` read drops from ~28M rows / 97,997 granules (over the read cap) to ~1.25M rows / 1,896 granules (~52× fewer granules). Note: the audit's original R6 suggestion (an explicit `t.workspace_id` predicate) is a no-op on CH 25.8 — the analyzer already propagates that constant across the join — so the real lever was bounding the trace-id scan, not the tenancy filter.
- **R8 — dead code** (`FeedbackScoreDAO`): removed the unreachable `DELETE_SPANS_CASCADE_FEEDBACK_SCORE` query and its only (uncalled) user, which also carried the degraded null-`project_id` filter.

Note: the audit's original R5 ("short-circuit empty target-project set to an empty result") was found to be a correctness bug (would drop legitimate rows and break `projectDeleted` filtering) and has been retracted/reclassified to R10 in OPIK-6883.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- OPIK-6884 (partial — rolling story; structural R1/R2/R3 tracked separately)

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: Codebase analysis, query/service edits, tests, prod EXPLAIN validation, and PR description for the remediations above.
- Human verification: Author reviewed each change and the audit reclassification; all cited tests were run locally and the R6 no-op was confirmed via prod EXPLAIN before reverting.

## Testing

Local, offline (`-o`), backend Testcontainers (real MySQL/ClickHouse):

- `mvn spotless:check` — clean
- `mvn test -Dtest='TracesResourceTest$DeleteTrace,TracesResourceTest$DeleteTraces'` — R4 single-trace, batch, and cascade; includes a new `deleteTracesAcrossProjectsWithoutProjectId` test covering the cross-project grouping path (11 tests, pass)
- `mvn test -Dtest=RetentionEstimationServiceTest` — R7 unit path with updated mocks (6 tests, pass)
- `mvn test -Dtest='RetentionPolicyServiceTest$CatchUpJob'` — R7 bounded query against real ClickHouse; small-workspace catch-up still deletes old data (2 tests, pass)
- `mvn test -Dtest=ExperimentProjectMigrationJobTest` — R6: mapping result unchanged with the added `IN` predicate (real ClickHouse, 1 test, pass)

Merged `origin/main` (24 commits) to run against the same code as CI; the deletion-events bridge tests + the two `TracesResourceTest` cascade tests now pass locally alongside the delete suites.

Prod validation (read-only `agentro`, plan-only `EXPLAIN` — no execution/scan; CH 25.8.16, large workspace):

- **R6** `COMPUTE_EXPERIMENT_PROJECT_MAPPING`: the `t.workspace_id` predicate the audit suggested is a no-op (analyzer already propagates it); the `t.id IN (...)` predicate bounds the `traces` read from ~28M rows / 97,997 granules to ~1.25M rows / 1,896 granules (~52×).
- **R4** span cascade `SELECT_SPAN_IDS_BY_TRACE_ID`: resolved `project_id` → 4,787 → 164 granules (~29×); trace `DELETE_BY_ID` → 58 granules; the bounded project lookup prunes via `id IN` today and via the `toMonday(id_at)` range once traces are partitioned.
- **R7** `ESTIMATE_VELOCITY_FOR_RETENTION`: the added `trace_id >= :lower_bound` is applied as a `trace_id` PK condition, bounding the scan to `[service_start, cutoff)`.

Not run: full repo regression suite (scoped to the affected areas). Structural items R1/R2/R3 are out of scope for this PR.

## Documentation

No user-facing or SDK changes; internal query-path hardening only.


apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -4437,6 +4458,29 @@ public Mono<Map<UUID, UUID>> getProjectIdsByTraceIds(@NonNull List<UUID> traceId
...
+        Preconditions.checkArgument(CollectionUtils.isNotEmpty(traceIds), "Argument 'traceIds' must not be empty");
+
+        var minId = Collections.min(traceIds);
+        var maxId = Collections.max(traceIds);
+
+        return asyncTemplate.nonTransaction(connection -> makeMonoContextAware((userName, workspaceId) -> {
+            var template = getSTWithLogComment(SELECT_PROJECT_IDS_BY_TRACE_IDS_BOUNDED,
+                    "get_project_ids_by_trace_ids_bounded", workspaceId, userName, traceIds.size());
+
+            var statement = connection.createStatement(template.render())
```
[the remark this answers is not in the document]
thiagohora: Fixed in aa0538c2ee — extracted a shared `collectTraceIdToProjectId(Statement)` helper now used by both `getProjectIdsByTraceIds` and `getProjectIdsByTraceIdsBounded`, so the row-mapping/context-binding/collect pipeline lives in one place.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceService.java
```diff
@@ -472,7 +472,33 @@ public Mono<TraceDetails> getTraceDetailsById(UUID id) {
...
+
+        if (projectId != null) {
+            return template.nonTransaction(connection -> delete(ids, projectId, connection));
+        }
+
+        // No project provided (e.g. delete-by-id, or a batch spanning projects): resolve each trace's owning
+        // project and delete per project group, so every delete - and its async span/feedback cascade carried
+        // by the TracesDeleted event - filters by project_id and prunes on the (workspace_id, project_id)
+        // sorting-key prefix instead of scanning the whole workspace. Trace ids with no resolvable project
+        // (the trace row is already gone) fall back to a workspace-scoped delete to still clean up orphan rows.
```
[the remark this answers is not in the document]
thiagohora: Fixed in aa0538c2ee — moved the count to the end: `log.info("Resolving owning projects for trace ids to delete per project group (count={})", ids.size());`.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/TracesResourceTest.java
```diff
@@ -3762,6 +3762,72 @@ void deleteTracesWithoutSpans() {
...
+                            .collect(Collectors.toUnmodifiableSet()))
+                    .build();
+            traceResourceClient.deleteTraces(request, workspaceName, apiKey);
+
+            getAndAssertPage(workspaceName, projectName1, null, List.of(), traces1, List.of(), List.of(), apiKey);
+            getAndAssertPage(workspaceName, projectName2, null, List.of(), traces2, List.of(), List.of(), apiKey);
+            Awaitility.await().pollInterval(500, TimeUnit.MILLISECONDS).untilAsserted(() -> {
+                getAndAssertPageSpans(workspaceName, projectName1, List.of(), spans1, List.of(), List.of(), apiKey);
+                getAndAssertPageSpans(workspaceName, projectName2, List.of(), spans2, List.of(), List.of(), apiKey);
+            });
```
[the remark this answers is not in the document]
thiagohora: Skipping — Awaitility already defaults to a 10s `atMost`, so these don't actually spin indefinitely, and this `pollInterval(...)`-without-`atMost` form is the established convention across ~20 call sites in this file; adding an explicit bound to just these two would be inconsistent.

Note: the CI timeout on `deleteTraceDeletesTraceAndSpanComments` / `batchDeleteTracesDeletesTraceAndSpanComments` was a real bug, not a missing timeout — the trace-delete cascade scopes spans by the trace's project, but those tests created the span in a *different* project than its trace (invalid: a span always shares its trace's project). Fixed in aa0538c2ee by creating the span in the trace's project.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -2097,6 +2099,24 @@ AND toMonday(id_at) = toMonday(UUIDv7ToDateTime(toUUID(:id), 'UTC'))
             ;
             """;
 
+    // Partition-bounded variant used on the delete path. id_at is MATERIALIZED as
+    // UUIDv7ToDateTime(toUUID(id)) (migration 000091), so bounding toMonday(id_at) by the id set's own
+    // min/max (mirrors SELECT_PROJECT_ID_FROM_TRACE) is exact - it never drops a valid id - while keeping
+    // the lookup on the ids' own weekly partitions instead of scanning all cold history once traces tier.
```
andrescrz: Nit: better as javadoc.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -4416,24 +4436,41 @@ public Mono<UUID> getProjectIdFromTrace(@NonNull UUID traceId) {
...
+    private Mono<Map<UUID, UUID>> collectTraceIdToProjectId(Statement statement) {
+        return makeMonoContextAware(bindWorkspaceIdToMono(statement))
+                .flatMapMany(result -> result.map((row, rowMetadata) -> Map.entry(row.get("id", UUID.class),
+                        row.get("project_id", UUID.class))))
+                .collect(toMap(Map.Entry::getKey, Map.Entry::getValue));
+    }
+
+    @Override
+    public Mono<Map<UUID, UUID>> getProjectIdsByTraceIdsBounded(@NonNull Set<UUID> traceIds) {
+        Preconditions.checkArgument(CollectionUtils.isNotEmpty(traceIds), "Argument 'traceIds' must not be empty");
```
andrescrz: Minor: overlap between non-null and checkArgument. Let's remove the non-null, it's properly covered by `CollectionUtils.isNotEmpty`.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -4416,24 +4436,41 @@ public Mono<UUID> getProjectIdFromTrace(@NonNull UUID traceId) {
...
             var statement = connection.createStatement(template.render())
                     .bind("trace_ids", traceIds.toArray(UUID[]::new));
 
-            return makeMonoContextAware(bindWorkspaceIdToMono(statement))
-                    .flatMapMany(result -> result.map((row, rowMetadata) -> Map.entry(row.get("id", UUID.class),
-                            row.get("project_id", UUID.class))))
-                    .collect(toMap(Map.Entry::getKey, Map.Entry::getValue))
-                    .doFinally(signalType -> {
-                        if (signalType == SignalType.ON_COMPLETE) {
-                            log.info("Got project_ids for '{}' trace_ids", traceIds.size());
```
andrescrz, line 4434: Nit: losing these logs, but I guess is fine and less noisy. We have metrics, `log_comment` etc.

andrescrz (dismissed):
Just a few nits, but LGTM.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
thiagohora: Done in d4ee86bf8 — converted the comment block above `SELECT_PROJECT_IDS_BY_TRACE_IDS_BOUNDED` to a `/** */` javadoc.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
thiagohora: Done in d4ee86bf8 — dropped `@NonNull`; the `checkArgument(CollectionUtils.isNotEmpty(...))` guard already covers both null and empty.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceService.java
```diff
@@ -476,7 +476,33 @@ public Mono<TraceDetails> getTraceDetailsById(UUID id) {
...
+                            .collect(Collectors.groupingBy(traceToProject::get, Collectors.toSet()));
+                    var unresolvedIds = ids.stream()
+                            .filter(id -> !traceToProject.containsKey(id))
+                            .collect(Collectors.toSet());
+
+                    return template.nonTransaction(connection -> Flux.fromIterable(idsByProject.entrySet())
+                            .concatMap(group -> delete(group.getValue(), group.getKey(), connection))
+                            .then(Mono.defer(() -> unresolvedIds.isEmpty()
+                                    ? Mono.empty()
+                                    : delete(unresolvedIds, null, connection)))
```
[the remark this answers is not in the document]
thiagohora: Good catch on the mechanism — you're right that `SpanService.create` resolves the project from the span's own `projectName` and never checks it against the referenced trace's project, so nothing at write time *enforces* span.project == trace.project today.

We're intentionally leaning on that as a data-model invariant here rather than widening the delete path. The whole point of R4 is to restore the `(workspace_id, project_id)` PK-prefix pruning on the span cascade (`SELECT_SPAN_IDS_BY_TRACE_ID`: 4,787 → 164 granules); dropping `project_id` from `deleteByTraceIds` to defensively catch cross-project rows would give that pruning right back and reintroduce the full-scan this PR exists to remove. A `trace_id`-only cleanup pass would need to scan the whole workspace span slice, which is exactly the anti-pattern OPIK-6884 is retiring.

So the two real options — (a) enforce same-project span creation at write time, or (b) a separate workspace-scoped reconciliation/cleanup job for legacy mismatches — are both larger, structural changes that sit outside this mechanical Slice-1 remediation. I've left the cascade project-scoped and will track write-time enforcement as a follow-up under the OPIK-6883/6884 story rather than fold it in here. (Also why the two cascade tests that created a span in a *different* project than its trace were corrected — that state isn't valid under the invariant.) cc @andrescrz in case you want to weigh in on enforcement vs. cleanup.

andrescrz (approved):

PR #7409, opened 2026-07-09
[OPIK-6809] [BE] test: cover trace-without-spans in experiment_items_bulk

BorisTkachenko:
## Details
Adds regression coverage for the `experiment-items/bulk` endpoint when a
record provides a `trace` but omits `spans`. Previously this combination
could trigger a `NullPointerException` (OPIK-6809). The current
`ExperimentItemBulkValidator` already handles the case null-safely
(`CollectionUtils.isEmpty(spans)`), and this test locks in that behavior so
the endpoint returns success (No Content) instead of a 500.

The new test in `ExperimentsResourceTest` builds a bulk upload with a single
record containing a trace and feedback scores but no spans, performs the bulk
upload, and asserts the persisted experiment items match the expected result.

## Change checklist

- [ ] User facing
- [ ] Documentation update

## Issues
- OPIK-6809

## Testing
- New test: `experimentItemsBulk__whenTraceProvidedButSpansOmitted__thenReturnNoContent`
  in `ExperimentsResourceTest`, verifying a trace-without-spans bulk record
  succeeds and produces the expected experiment items (no NPE / 500).

## Documentation
N/A

### DIFF
--- a/apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/ExperimentsResourceTest.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/ExperimentsResourceTest.java
@@ -6465,6 +6465,45 @@ class ExperimentsResourceTest {
             assertExperimentItems(actualExperimentItems, expectedItems);
         }
 
+        @Test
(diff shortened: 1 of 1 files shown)


LifeXplorer (approved):
Looks good!

PR #7413, opened 2026-07-09
[OPIK-7281] Fix outbound gzip response double-decompression

thiagohora:
## Details

Fixes recurring `java.util.zip.ZipException: Not in GZIP format` on the shared outbound HTTP client. With `jerseyClient.gzipEnabled: true` (production default), Dropwizard wires **both** Apache HttpClient's transparent content decompression **and** Jersey's `GZipDecoder`. HttpClient decodes the body, but the stale `Content-Encoding: gzip` header survives into the Jersey response, so `GZipDecoder` decodes the already-plain stream a second time and `readEntity` throws. This affects any outbound Jersey call reading a gzipped body ≥ the upstream's minimum gzip size without a per-call `.acceptEncoding("identity")` workaround (e.g. `OllamaService.listModels`, whose failure is swallowed into an empty model list).

- Register a `ContentEncodingReconcilingResponseFilter` on the shared client (gated on `gzipEnabled`) that strips the stale `Content-Encoding`/`Content-Length` after HttpClient's transparent decode — leaving exactly one decode while keeping `Accept-Encoding: gzip` on the wire. One systemic fix for every call site, instead of per-call opt-outs.
- Underlying cause is upstream in Dropwizard's `DropwizardApacheConnector` (it copies raw response headers, incl. `Content-Encoding`, alongside the already-decoded entity); to be reported separately.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- OPIK-7281

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: Full diff (filter, `HttpModule` wiring, regression test) drafted by Claude from a production-log investigation; root cause pinned via Dropwizard bytecode inspection.
- Human verification: Approach directed and reviewed by the author; regression test verified locally to fail with the exact `ZipException` when the filter is disabled and pass with it enabled.

## Testing

Local, `apps/opik-backend`:
- `mvn -Dtest=HttpModuleGzipDecompressionTest test` — new regression test built with `gzipEnabled: true` (the path `config-test.yml` skips). Verified it reproduces the defect: with the filter disabled it fails with `ProcessingException` caused by `ZipException: Not in GZIP format`; with the filter enabled it passes.
- `mvn -Dtest=RemoteAuthServiceTest test` — 29/29 pass (no regression on the existing gzip-enabled auth path).

## Documentation

No user-facing documentation changes.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/http/ContentEncodingReconcilingResponseFilter.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/http/ContentEncodingReconcilingResponseFilter.java
@@ -4,6 +4,7 @@ import jakarta.ws.rs.client.ClientRequestContext;
 import jakarta.ws.rs.client.ClientResponseContext;
 import jakarta.ws.rs.client.ClientResponseFilter;
 import jakarta.ws.rs.core.HttpHeaders;
+import lombok.NonNull;
 
 import java.util.Locale;
 import java.util.Set;
@@ -32,11 +33,11 @@ public class ContentEncodingReconcilingResponseFilter implements ClientResponseF
     private static final Set<String> TRANSPARENTLY_DECODED = Set.of("gzip", "x-gzip", "deflate");
 
     @Override
-    public void filter(ClientRequestContext requestContext, ClientResponseContext responseContext) {
+    public void filter(ClientRequestContext requestContext, @NonNull ClientResponseContext responseContext) {
         String contentEncoding = responseContext.getHeaderString(HttpHeaders.CONTENT_ENCODING);
 
         if (contentEncoding != null
-                && TRANSPARENTLY_DECODED.contains(contentEncoding.trim().toLowerCase(Locale.ROOT))) {
+                && TRANSPARENTLY_DECODED.contains(contentEncoding.strip().toLowerCase(Locale.ROOT))) {
             var headers = responseContext.getHeaders();
             headers.remove(HttpHeaders.CONTENT_ENCODING);
             headers.remove(HttpHeaders.CONTENT_LENGTH);
--- a/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/http/ContentEncodingReconcilingResponseFilter.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/http/ContentEncodingReconcilingResponseFilter.java
@@ -1,0 +1,45 @@
+package com.comet.opik.infrastructure.http;
+
+import jakarta.ws.rs.client.ClientRequestContext;
+import jakarta.ws.rs.client.ClientResponseContext;
+import jakarta.ws.rs.client.ClientResponseFilter;
+import jakarta.ws.rs.core.HttpHeaders;
+
+import java.util.Locale;
+import java.util.Set;
+
+/**
+ * Reconciles response headers after Apache HttpClient 5 has already decompressed the body, so that
+ * Jersey's {@code GZipDecoder} does not decompress it a second time.
+ *
+ * <p>When {@code jerseyClient.gzipEnabled} is {@code true}, Dropwizard wires BOTH Apache HttpClient's
+ * transparent content decompression (advertises {@code Accept-Encoding} and decodes the entity) AND
+ * Jersey's {@code GZipDecoder} reader interceptor. {@code DropwizardApacheConnector} copies the raw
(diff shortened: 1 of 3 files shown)


apps/opik-backend/src/test/java/com/comet/opik/infrastructure/http/HttpModuleGzipDecompressionTest.java
```diff
@@ -0,0 +1,103 @@
...
+        WIRE_MOCK.server().resetAll();
+    }
+
+    @Test
+    void readEntity__whenResponseIsGzipCompressed__thenSharedClientDecodesExactlyOnce() {
+        // A body large enough that the upstream gzips it, like a real Dropwizard/Jetty upstream over its
+        // minimum entity size. With gzipEnabled: true both HttpClient and Jersey's GZipDecoder are active;
+        // the reconciling filter drops the stale Content-Encoding header so only HttpClient decodes.
+        var largeBody = "{\"payload\":\"" + "x".repeat(4000) + "\"}";
+        WIRE_MOCK.server().stubFor(get(urlPathEqualTo("/models")).willReturn(okJson(largeBody)));
```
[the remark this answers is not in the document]
thiagohora: Good catch — addressed in 19c72e3. The test no longer relies on WireMock’s auto-gzip default: it now serves an explicitly gzipped body with `Content-Encoding: gzip` from a WireMock server started with `gzipDisabled(true)` (so WireMock doesn’t re-encode), deterministically reproducing a real gzipping upstream.

For the record, the original version did exercise the path (WireMock auto-gzips when the client sends `Accept-Encoding: gzip`) — I verified it failed with the exact `ZipException` when the filter was removed — but you’re right that depending on that default was fragile. The explicit version is verified the same way: still fails with `ProcessingException` → `ZipException: Not in GZIP format` when the reconciling filter is disabled, passes with it enabled.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/http/ContentEncodingReconcilingResponseFilter.java
```diff
@@ -0,0 +1,45 @@
...
+ * It is only registered when gzip is enabled (see {@link HttpModule#buildClient}), which is exactly when
+ * HttpClient performs the transparent decode.
+ */
+public class ContentEncodingReconcilingResponseFilter implements ClientResponseFilter {
+
+    // Content codings that Apache HttpClient 5's ContentCompressionExec decodes transparently by default.
+    private static final Set<String> TRANSPARENTLY_DECODED = Set.of("gzip", "x-gzip", "deflate");
+
+    @Override
+    public void filter(ClientRequestContext requestContext, ClientResponseContext responseContext) {
```
andrescrz: Nit: validations in public methods, mostly non-null for `responseContext`.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/http/ContentEncodingReconcilingResponseFilter.java
```diff
@@ -0,0 +1,45 @@
...
+        String contentEncoding = responseContext.getHeaderString(HttpHeaders.CONTENT_ENCODING);
+
+        if (contentEncoding != null
+                && TRANSPARENTLY_DECODED.contains(contentEncoding.trim().toLowerCase(Locale.ROOT))) {
```
andrescrz: Nit: use `strip` instead of `trim` for more modern Java character coverage.

andrescrz (dismissed):
LGTM.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/http/ContentEncodingReconcilingResponseFilter.java
thiagohora: Done in f0853cf59 — added `@NonNull` on the `responseContext` parameter.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/http/ContentEncodingReconcilingResponseFilter.java
thiagohora: Done in f0853cf59 — switched `trim()` to `strip()`.

andrescrz (approved):

PR #7414, opened 2026-07-09
[OPIK-7110] [BE] fix: stop capping user-mapped variables on the agentic-tools path

miguelgrc:
## Details

On the agentic-tools path, `capReplacements()` was truncating ALL variable substitutions to 4K chars — including user-mapped scoring variables (`output`, `expected_output`). The LLM judge then had to drill down via tools to recover the full values, which was non-deterministic and produced intermittent "Insufficient data" failures for customers with large traces.

- `capReplacements()` now accepts an `uncappedKeys` set: user-mapped variables (trace sections and literals) pass through at full length, while sentinel variables (`spans`, `trace`, `span`) are still capped since the agentic tools are purpose-built to navigate them.
- Both trace-level and span-level capped `prepareLlmRequest` overloads use the new selective capping.
- The old uniform `capReplacements(replacements, max, hint)` is removed; callers pass `Set.of()` for uniform capping.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- OPIK-7110

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code (Opus)
- Model(s): claude-opus-4-6
- Scope: Implementation, tests, PR description
- Human verification: Code reviewed, locally tested with reproduction script confirming fix

## Testing

- `mvn test -f apps/opik-backend/pom.xml -Dtest=\"OnlineScoringEngineCapReplacementsTest,OnlineScoringEngineTest\"` — 108 tests pass, 0 failures
- `mvn spotless:check` — passes
- Local reproduction with `repro_opik_7110.py`: created an LLM-as-judge rule with `output`/`expected_output` variable mappings, logged traces with 240K+ char output (answer buried past the 4K cap mark), confirmed agentic-tools path activates (`Trace context exceeds '50000' tokens; switching to agentic-tools mode`), and verified:
  - **Before fix**: non-deterministic results — some traces scored `answer_present=[0]` (LLM failed to drill down and find the answer)
  - **After fix**: LLM sees full output directly in the prompt, no tool drill-down needed for scoring data

## Documentation

N/A — internal behavioral fix, no API or config changes.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringEngine.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringEngine.java
@@ -22,6 +22,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.PropertyNamingStrategies;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import com.google.api.gax.rpc.InvalidArgumentException;
+import com.google.common.annotations.VisibleForTesting;
 import com.jayway.jsonpath.JsonPath;
 import dev.langchain4j.data.message.AudioContent;
 import dev.langchain4j.data.message.ChatMessage;
@@ -352,6 +353,7 @@ public class OnlineScoringEngine {
      * sentinel variables (like {@code spans}). Used by {@link #capReplacements} to decide which
      * keys to leave uncapped.
      */
+    @VisibleForTesting
     static Set<String> userMappedVariableKeys(Map<String, String> variables) {
         var keys = new HashSet<>(variables.keySet());
         keys.removeIf(k -> SENTINEL_VARIABLE_VALUES.contains(variables.get(k)));
@@ -364,6 +366,7 @@ public class OnlineScoringEngine {
      * variables should be uncapped — capping them forces non-deterministic tool drill-down
      * (OPIK-7110).
      */
+    @VisibleForTesting
     static Map<String, String> capReplacements(Map<String, String> replacements,
             int maxReplacementChars, String drillDownHint, Set<String> uncappedKeys) {
         return replacements.entrySet().stream().collect(Collectors.toMap(
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringEngine.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringEngine.java
@@ -80,6 +80,9 @@ public class OnlineScoringEngine {
     private static final String TRACE_VARIABLE_NAME = "trace";
     private static final String SPAN_VARIABLE_NAME = "span";
 
+    private static final Set<String> SENTINEL_VARIABLE_VALUES = Set.of(
+            SPANS_VARIABLE_NAME, TRACE_VARIABLE_NAME, SPAN_VARIABLE_NAME);
+
     private static final ObjectMapper OBJECT_MAPPER = JsonUtils.getMapper();
 
     private static final Pattern JSON_BLOCK_PATTERN = Pattern.compile(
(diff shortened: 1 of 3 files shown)


apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringEngine.java
```diff
@@ -150,7 +155,9 @@ public static ChatRequest prepareLlmRequest(
                 evaluatorCode.messages(), promptType, spans);
         injectTraceIntoReplacements(replacements, evaluatorCode.variables(),
                 evaluatorCode.messages(), promptType, traceStructureJson);
-        Map<String, String> capped = capReplacements(replacements, maxReplacementChars, drillDownHint);
+        Set<String> userMappedKeys = userMappedVariableKeys(evaluatorCode.variables());
+        Map<String, String> capped = capReplacements(replacements, maxReplacementChars,
+                drillDownHint, userMappedKeys);
         var renderedMessages = renderMessagesWithReplacements(evaluatorCode.messages(), capped, promptType);
         return buildChatRequest(renderedMessages, evaluatorCode.schema(), structuredOutputStrategy);
```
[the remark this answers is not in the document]
miguelgrc: The two methods handle different entity types (trace vs span) with different inject methods (`injectSpans/TraceIntoReplacements` vs `injectSpanIntoReplacements`), different variable types (`LlmAsJudgeCode` vs `SpanLlmAsJudgeCode`), and different prompt types. A shared helper would need 7+ parameters and obscure the flow. Keeping the 3-line pattern inline is clearer for a bug-fix PR.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringEngine.java
```diff
@@ -339,11 +346,31 @@ private static String serializeSpansTree(List<Span> spans) {
...
+     */
     static Map<String, String> capReplacements(Map<String, String> replacements,
-            int maxReplacementChars, String drillDownHint) {
+            int maxReplacementChars, String drillDownHint, Set<String> uncappedKeys) {
         return replacements.entrySet().stream().collect(Collectors.toMap(
                 Map.Entry::getKey,
-                e -> StringTruncator.truncate(e.getValue(), maxReplacementChars, drillDownHint),
+                e -> uncappedKeys.contains(e.getKey())
+                        ? e.getValue()
+                        : StringTruncator.truncate(e.getValue(), maxReplacementChars, drillDownHint),
```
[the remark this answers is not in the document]
miguelgrc: Intentional. The inline fallback path is for `{{trace}}` rules on non-tool-calling providers — already a degraded mode. If the user mapped `output` for scoring, the LLM needs to see it to produce a correct result. An API context-window error (which surfaces clearly) is a better outcome than silently truncating the scoring data and returning wrong scores — which is the exact bug this PR fixes (OPIK_7110). If context overflow becomes a real problem on this path, we can add a dedicated size guard with a user-facing warning, but it should not silently cap the scoring variables.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringEngine.java
```diff
@@ -339,11 +346,31 @@ private static String serializeSpansTree(List<Span> spans) {
         }
     }
 
+    /**
+     * Returns the set of replacement-map keys that correspond to user-mapped variables (trace
+     * sections like {@code output}, {@code input.question}, or literal constants) as opposed to
+     * sentinel variables (like {@code spans}). Used by {@link #capReplacements} to decide which
+     * keys to leave uncapped.
+     */
+    static Set<String> userMappedVariableKeys(Map<String, String> variables) {
```
andrescrz, line 357: Not a blocker: I'd try to test this indirectly through the public methods, instead of unit testing, that should give you enough coverage in most of the cases. But if you still want to unit test this, please annotate it with `VisibleForTesting` from Guava.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringLlmAsJudgeScorer.java
```diff
@@ -62,6 +69,12 @@
 @Slf4j
 public class OnlineScoringLlmAsJudgeScorer extends OnlineScoringBaseScorer<TraceToScoreLlmAsJudge> {
 
+    private static final String ONLINE_SCORING_NAMESPACE = "online_scoring";
+    private static final AttributeKey<String> WORKSPACE_ID_KEY = AttributeKey.stringKey("workspace_id");
+    private static final AttributeKey<String> WORKSPACE_NAME_KEY = AttributeKey.stringKey("workspace_name");
+    private static final AttributeKey<String> PATH_KEY = AttributeKey.stringKey("path");
+    private static final AttributeKey<String> TRIGGER_KEY = AttributeKey.stringKey("trigger");
```
andrescrz, line 76: These two metric are going to provide a lot of cardinality. Not a blocker, but you should keep an eye to see if this behaves well in production and check that it doesn't overwhelm our telemetry pipeline (collection,  ingestion, dashboards etc.)

andrescrz (dismissed):
Couple of thoughts, but not blockers.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringLlmAsJudgeScorer.java
miguelgrc: Good point — will keep an eye. The cardinality is bounded at 10 × #workspaces (2 paths × 5 triggers × workspace), which is the same pattern as `online_scoring_sampler_decisions_total` already in production (workspace × evaluator_type × 4 decisions). `workspace_id` and `workspace_name` are 1:1 so they don't multiply cardinality — both exist so the Grafana workspace picker can filter by name.

andrescrz (approved):

PR #7416, opened 2026-07-09
[OPIK-7267] [BE] fix: normalize Bearer-prefixed API key in remote auth

alexkuzmik:
## Details

Optimization Studio runs failed immediately on single-tenant deployments with `401 "User with provided api key not found!"`, while OSS and multi-tenant SaaS worked on the same version and model. The optimizer worker routes its LLM calls through the backend gateway (`/v1/private/chat/completions`) via LiteLLM, whose OpenAI-compatible client sends the Opik API key as `Authorization: Bearer <key>`. `RemoteAuthService` forwarded that header verbatim to the auth service, which does a literal key lookup — so the `Bearer ` prefix broke authentication, even though the very same key authenticates fine for the SDK's REST calls (which send it raw, e.g. the dataset load that succeeds earlier in the same run).

- Strip a leading `Bearer ` prefix in `RemoteAuthService.authenticateUsingApiKey`, before the cache lookup and before forwarding to the auth service, so the cache key and the forwarded header both use the raw key.
- A real Opik API key never starts with `Bearer `, so the normalization is safe; it also makes any OpenAI-compatible client hitting the gateway behave consistently across OSS / SaaS / single-tenant.
- The MCP-OAuth path is unaffected (its tokens carry a distinct `opik_mcp_at_` prefix and are handled earlier in `AuthFilter`).

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves OPIK-7267

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Scope: full implementation — investigation, fix, and regression test
- Human verification: reproduction and fix verified via the added `RemoteAuthService` test; pending code review

## Testing

- `cd apps/opik-backend && mvn -o test -Dtest=RemoteAuthServiceTest` → 30 passed, 0 failures.
- Reproduction: on the unmodified code the new test `testAuth__whenApiKeyHasBearerPrefix__thenPrefixStrippedAndAuthSucceeds` fails with the exact reported error `jakarta.ws.rs.ClientErrorException: User with provided api key not found!` — the WireMock auth-service stub models the strict single-tenant behavior (the raw key resolves to a user; a `Bearer `-prefixed value is rejected). With the fix, the prefix is stripped and the test passes.
- `cd apps/opik-backend && mvn -o spotless:check` → clean.
- Full backend test suite runs in CI.
- Not run locally: an end-to-end optimizer run against a real single-tenant deployment (requires an auth-enabled backend plus a strict remote auth service). The added test reproduces the failing seam faithfully, using the exact `Authorization` header LiteLLM emits and the exact error string from the report.

## Documentation

N/A — internal auth normalization only; no user-facing, API, or configuration changes.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
@@ -44,6 +44,7 @@ import static com.comet.opik.infrastructure.auth.RequestContext.WORKSPACE_QUERY_
 class RemoteAuthService implements AuthService {
     private static final String USER_NOT_FOUND = "User not found";
     private static final String NOT_LOGGED_USER = "Please login first";
+    private static final String BEARER_PREFIX = "Bearer ";
 
     // GenericType instances are thread-safe and expensive to build, so reuse a single instance.
     private static final GenericType<List<WorkspaceIdNameResponse>> WORKSPACE_LIST_TYPE = new GenericType<>() {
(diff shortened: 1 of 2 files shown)


alexkuzmik:
Superseded by the upstream root-cause fix in comet-backend: https://github.com/comet-ml/comet-backend/pull/5616

Deeper investigation showed the real defect is in comet-backend's `RestHelpers.getRestAuthTokenFromHeader`: the Bearer branch stripped the `Bearer ` prefix but never ran `parseSmartApiKey`, so a Bearer-wrapped smart key (`<raw>*<base64-baseUrl>`) was looked up verbatim and missed → `401 "User with provided api key not found!"`. The non-Bearer (SDK) path smart-parses, which is why the same key worked for SDK REST calls but not the optimizer gateway.

Fixing it in comet-backend covers every caller of `/opik/auth`, so the Opik-side normalization in `RemoteAuthService` proposed here is no longer needed. Closing in favor of the comet-backend fix.

PR #7421, opened 2026-07-09
[OPIK-7292] [SDK] fix: preserve tags on CLI export/import for prompts, experiments, datasets

dsblank:
## Details

`opik export` / `opik import` silently dropped `tags` for prompts, experiments, and datasets because those exporters used hand-written field allowlists that omitted `tags`, and the importers never forwarded them. Traces and spans already round-tripped correctly. This threads tags through both sides of the flow.

- **Prompts**: serialize `tags` in `current_version` + each `history` entry; forward to `create_prompt`/`create_chat_prompt` on import.
- **Experiments**: add `tags` to the export structure; forward on the plain import-from-disk path (previously only the `opik migrate` path forwarded them — that path is unchanged).
- **Datasets**: export dataset-level tags via `dataset.get_tags()` (wrapped so a tags lookup can never abort an export); apply on import via the REST `update_dataset` endpoint, since the high-level `create_dataset()` has no `tags` argument.

### Notes / out of scope
- **Dataset item-level tags are still dropped**, but earlier — inside the SDK. `api_objects/dataset/rest_operations.py:144` reconstructs `DatasetItem` without the `tags` column, so the CLI never sees item tags. Fixing that is a separate core-SDK change, intentionally not included here. (The existing migrate e2e `conftest.py` already documents this SDK limitation.)
- On import, dataset tags are applied with `update_dataset`, which **replaces** rather than merges tags on a pre-existing dataset. This is reasonable for import; flagging in case merge semantics are preferred.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-7292

## AI-WATERMARK

AI-WATERMARK: yes

- If yes:
  - Tools: Claude Code
  - Scope: Investigation of the tag-drop bug, code changes across 6 CLI export/import modules, plus new unit tests and extended e2e coverage.
  - Human verification: Author reviewed the diff, and the fix was verified against the Opik REST/SDK signatures (`create_prompt`/`create_chat_prompt`/`create_experiment` accept `tags`; `update_dataset` REST endpoint used for datasets).

## Testing

**Unit** — `sdks/python/tests/unit/cli/test_tags_roundtrip.py` (new, 7 tests): prompt (text + chat) import forwarding, dataset import (flat format, nested format, no-tags-skips-update), and experiment (export structure + import-from-disk forwarding).

**E2E** — `sdks/python/tests/e2e/test_cli_import_export.py` (extended): seed tags on the dataset, prompt, chat-prompt, and experiment fixtures and assert they survive both the exported files and the re-import.
- `test_export_import_datasets_happy_flow`: tags present in the exported dataset file **and** on the re-imported dataset (`get_tags()`).
- `test_export_import_prompts_happy_flow`: tags present in the exported `current_version` **and** on the imported prompt.
- `test_export_import_chat_prompts_happy_flow`: same, for the chat path.
- `test_export_experiment_with_dataset_filter`: tags present in the exported experiment file.
- `test_import_projects_automatically_recreates_experiments`: now also exports the experiment (trace export alone emits none), so it actually exercises recreation and verifies tags on the recreated experiment.

Previously this e2e suite asserted no tags on any entity — which is why the regression went unnoticed.

Commands run from `sdks/python`:
- `python -m pytest tests/unit/cli/test_tags_roundtrip.py -q` → 7 passed
- `python -m pytest tests/unit/cli/ -q` → 424 passed (no regressions; migrate-path cascade tests still green)
- `python -m pytest tests/e2e/test_cli_import_export.py --collect-only` → 12 collected, no import errors
- `ruff check` + `ruff format --check` on all changed files → clean

Not run locally: the e2e suite itself, which requires a live Opik backend — it runs in CI.

## Documentation

N/A — no user-facing docs describe the per-entity field lists; behavior change is a bug fix (tags now preserved).

### DIFF
--- a/sdks/python/src/opik/cli/imports/dataset.py
+++ b/sdks/python/src/opik/cli/imports/dataset.py
@@ -64,6 +64,14 @@ def import_datasets_from_directory(
                     else None
                 )
 
+                # Dataset-level tags live at the top level (export_single_dataset)
+                # or nested under "dataset" (export_experiment_datasets).
+                dataset_tags = dataset_data.get("tags") or (
+                    dataset_data.get("dataset", {}).get("tags")
+                    if dataset_data.get("dataset")
+                    else None
+                )
+
                 # Check if name is missing or empty
                 if not dataset_name or (
                     isinstance(dataset_name, str) and not dataset_name.strip()
@@ -113,6 +121,25 @@ def import_datasets_from_directory(
                     if debug:
                         console.print(
                             f"[blue]Created new dataset: {dataset_name}[/blue]"
+                        )
+
+                # Apply dataset-level tags. create_dataset() doesn't expose a
+                # tags argument, so set them via the REST update endpoint after
+                # the dataset exists.
+                if dataset_tags:
+                    try:
+                        client.rest_client.datasets.update_dataset(
+                            dataset.id,
+                            name=dataset_name,
+                            tags=dataset_tags,
+                        )
+                        if debug:
+                            console.print(
+                                f"[blue]Applied tags to dataset '{dataset_name}': {dataset_tags}[/blue]"
+                            )
+                    except Exception as tag_error:
+                        console.print(
+                            f"[yellow]Warning: Could not apply tags to dataset '{dataset_name}': {tag_error}[/yellow]"
                         )
 
                 # Import dataset items
--- a/sdks/python/src/opik/cli/imports/dataset.py
+++ b/sdks/python/src/opik/cli/imports/dataset.py
@@ -133,9 +133,18 @@ def import_datasets_from_directory(
                 # tags, so distinguish None (no tags field -> skip) from [].
                 if dataset_tags is not None:
                     try:
+                        # description is silently nulled by the backend when
+                        # omitted from this PUT (unlike tags/visibility, it
+                        # isn't COALESCEd against the existing row), so it
+                        # must be re-passed here to avoid wiping it.
+                        existing = client.rest_client.datasets.get_dataset_by_id(
+                            dataset.id
+                        )
                         client.rest_client.datasets.update_dataset(
                             dataset.id,
                             name=dataset_name,
+                            description=existing.description,
+                            visibility=existing.visibility,
                             tags=dataset_tags,
                         )
                         if debug:
--- a/sdks/python/src/opik/cli/imports/dataset.py
+++ b/sdks/python/src/opik/cli/imports/dataset.py
@@ -65,12 +65,16 @@ def import_datasets_from_directory(
                 )
 
                 # Dataset-level tags live at the top level (export_single_dataset)
-                # or nested under "dataset" (export_experiment_datasets).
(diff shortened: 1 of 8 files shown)


sdks/python/src/opik/cli/imports/dataset.py
```diff
@@ -115,6 +127,26 @@ def import_datasets_from_directory(
...
+                # tags argument, so set them via the REST update endpoint after
+                # the dataset exists. An explicit empty list clears existing
+                # tags, so distinguish None (no tags field -> skip) from [].
+                if dataset_tags is not None:
+                    try:
+                        client.rest_client.datasets.update_dataset(
+                            dataset.id,
+                            name=dataset_name,
+                            tags=dataset_tags,
+                        )
```
petrotiurin, line 149: This call omits `description` (and `visibility`), which will silently null the destination dataset's `description` whenever tags are applied here.

Traced through the backend:
- `DatasetUpdate.java` — `description` has no `@NotNull`, so it deserializes to `null` when the field is missing from the request body.
- `DatasetDAO.java`'s update statement:
  ```sql
  UPDATE datasets SET
      name = :dataset.name,
      description = :dataset.description,
      visibility = COALESCE(:dataset.visibility, visibility),
      tags = COALESCE(:dataset.tags, tags),
      last_updated_by = :lastUpdatedBy
  WHERE id = :id AND workspace_id = :workspace_id
  ```
  `visibility`/`tags` are `COALESCE`d against the existing row, but `description` is **not** — it's overwritten unconditionally.

This hits the "dataset already exists" branch above (`client.get_dataset(...)` succeeds) — exactly the path taken on re-imports/migrations into a pre-existing dataset that may already have a real description set via the UI or SDK.

This exact footgun is already known and worked around elsewhere in this codebase: `cli/migrate/datasets/executor.py`'s `_apply_action` has the comment *"Re-pass description/visibility/tags so the BE doesn't wipe them on the rename PUT (description is silently nulled when omitted)"* and explicitly forwards all three fields on its `update_dataset` calls. This call doesn't follow that pattern.

Suggested fix: pass `description=dataset.description` (and ideally `visibility=dataset.visibility`) alongside `tags` here.

Also worth adding a test: seed a dataset with both a description and tags, import tags-only data onto it, and assert the description survives — none of the current tests catch this.

dsblank: Good catch, thanks for tracing this through the DAO layer. Fixed in 7af40346fb: the update now fetches the existing dataset via `get_dataset_by_id` and re-passes both `description` and `visibility` alongside `tags`, so a tags-only re-import no longer wipes them.

Also added a regression test (`test_import_dataset__existing_dataset_with_description__preserves_description` in `sdks/python/tests/unit/cli/test_tags_roundtrip.py`) that seeds a dataset with a description, imports tags-only data onto it, and asserts the description (and visibility) survive.

petrotiurin (approved):
LGTM, let's wait for Python SDK E2E Tests to pass then we're good to merge!

PR #7428, opened 2026-07-10
[OPIK-7159][OPIK-7029] feat: recover stuck optimization runs + surface run errors

awkoy:
## Details

Optimization Studio runs could fail **silently** — a spinner that never stops, or a run that "completes" looking empty with no explanation. This PR makes every failure visible and recoverable.

> Rebased onto `main` and rebuilt to contain **only** the 7159 + 7029 work (the single-run overview UI it was previously stacked on has since merged to `main`). Diff is now 56 files.

### The three failure outcomes
| Outcome | When | Result |
|---|---|---|
| **A — Error** | An exception propagates out of the run | Terminal `ERROR` + a specific `[System]` reason in the run's error panel |
| **B — Stuck** | The worker never ran / the Opik status callback couldn't reach the backend | The **stalled-run reaper** transitions it to `ERROR` after a timeout |
| **C — Empty** | The run finished but the metric/judge failed on the items | Warning panel + the **exact count** of items that failed to score |

The **LLM provider key** failing makes the *optimization* fail (A/C); the **Opik workspace key** failing makes the *status* get stuck (B). These now produce distinct messages instead of both blaming "the model provider".

### What changed, by layer

**Backend (`opik-backend`)**
- `OptimizationStalledReaperJob` + `OptimizationStalledReaperConfig`: a scheduled, distributed-locked reaper that moves runs stuck past a timeout to `ERROR` with a curated `[System]` reason. Timeouts (all env-overridable): **INITIALIZED 5m**, **RUNNING 8h** — the latter kept above the worker's execution cap (`OPTSTUDIO_EXECUTION_TIMEOUT`, now 6h). Normal failures are surfaced immediately by the worker's `mark_error`; the reaper is the backstop only for "no callback at all" (worker never ran / process died / callback unreachable). *(OPIK-7159)*
- Enqueue failure now marks `ERROR` ("couldn't start") instead of the misleading `CANCELLED`; cancel-path Redis failure no longer 500s; removed a dead `501` cancel stub.
- `OptimizationUpdate.metadata` (optional `JsonNode`) is merged onto the existing metadata column (preserves `optimizer`/`model`; non-object values are rejected; status-only updates leave metadata untouched).

**Python worker (`opik-python-backend`)**
- `mark_completed()` moved out of the lifecycle try/except — a run that **succeeded** can never be mislabeled `ERROR` by a flaky completion callback.
- Structured studio error catalog (`errors.py`): distinguishes Opik-callback failures from LLM-provider failures; curated messages for unreadable optimizer output, misconfigured metric/optimizer params, and OOM kills.
- Forwards `scoring_health` on completion (best-effort; falls back to a status-only update if rejected).

**SDK (`opik_optimizer`)**
- Raises `ScoringFailedError` (now a public export) when the failed/total score ratio hits the threshold — a run the judge can't score **errors** instead of "completing" empty.
- Attaches `scoring_health {failed_count, total_count}` to the result so the exact count reaches the UI.

**Frontend (`opik-frontend`)**
- `EmptyRunWarningPanel` on empty COMPLETED runs + a "scoring failed" KPI caption showing **"N of M items failed to score"**.
- Provider-key preflight warning before submit; "failed" trial status in the progress chart.
- No client-side stall timer — an in-progress run just shows the spinner and relies on the reaper to flip it to `ERROR` (then `RunErrorPanel` surfaces the reason).
- (Leaves main's existing `RunErrorPanel`/error-extractor intact — no regression.)

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues
- OPIK-7159
- OPIK-7029

## AI-WATERMARK
AI-WATERMARK: yes
- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: stalled-run reaper; cross-layer error surfacing (worker catalog, backend enqueue/cancel/metadata, FE empty-run panel); exact scoring-health count (SDK → worker → backend → FE); rebuilt cleanly on `main`
- Human verification: reviewed locally; backend compiled JDK 25; full stack + e2e validated against a live backend

## Testing
- **Backend** — `OptimizationStalledReaperServiceTest` (reaper transitions/guards), `OptimizationsResourceTest` (metadata merge preserves keys; status-only leaves untouched), `JsonUtilsTest` (merge rejects non-object). Compiles on JDK 25. **18 scoped tests pass.**
- **Python worker** — 54 unit tests (`test_studio_errors`, `test_studio_status_manager`, `test_studio_types`).
- **SDK** — `test_scoring_health` (11) + full unit suite green.
- **Frontend** — 197 tests across the optimization pages; typecheck + eslint clean.
- **End-to-end (live stack)** — scoring-health verified link-by-link with real code: SDK computed `{failed:2, total:5}` on a real run; the worker's real status-manager PUT persisted through the backend; the backend merge preserved `optimizer`/`model`; the live FE rendered "3 of 7 items failed to score".

## Documentation

No user-facing documentation changes. The new backend config keys (`optimizationStalledReaper.*`) are documented inline in `config.yml` with defaults and env-var overrides.


apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/OptimizationStalledReaperJob.java
```diff
@@ -0,0 +1,119 @@
...
+
+    private final AtomicBoolean interrupted = new AtomicBoolean(false);
+
+    /** Tracks the in-flight reactive pass so {@link #interrupt()} can dispose it. */
+    private final AtomicReference<Disposable> currentExecution = new AtomicReference<>();
+
+    @Inject
+    public OptimizationStalledReaperJob(@NonNull OptimizationService optimizationService,
+            @NonNull LockService lockService,
+            @NonNull @Config("optimizationStalledReaper") OptimizationStalledReaperConfig config) {
```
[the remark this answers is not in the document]
awkoy: Done — switched to `@RequiredArgsConstructor(onConstructor_ = @Inject)` with a field-level `@NonNull @Config("optimizationStalledReaper")`, matching the SKILL.md guideline and `OptimizationLogSyncService` in this same feature (field-level named `@Config` is used in 8+ places across the codebase).

apps/opik-backend/src/main/java/com/comet/opik/domain/OptimizationDAO.java
```diff
@@ -53,6 +54,14 @@ public static OptimizationSummary empty(UUID datasetId) {
         }
     }
 
+    /**
+     * A studio optimization whose latest status is non-terminal and older than the reaper threshold,
+     * i.e. stuck because the worker never advanced it. Carries {@code workspaceId} so the reconciler
+     * can seed the workspace context required to update the row and finalize its logs (OPIK-7159).
+     */
+    record StalledOptimization(UUID id, String workspaceId, OptimizationStatus status) {
+    }
```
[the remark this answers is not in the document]
awkoy: Done — `StalledOptimization` now has `@Builder(toBuilder = true)` with `@NonNull` on `id`/`workspaceId`/`status`, and `findStalledStudioOptimizations` constructs it via the builder instead of the positional constructor.

apps/opik-frontend/src/v2/pages/OptimizationPage/runError.ts
```diff
@@ -1,24 +1,41 @@
...
   const lines = logContent
     .split("\n")
     .map((line) => stripAnsi(line).trim())
     .filter(Boolean);
 
   for (let i = lines.length - 1; i >= 0; i--) {
-    if (ERROR_LINE_RE.test(lines[i])) return lines[i];
+    if (lines[i].startsWith(SYSTEM_PREFIX)) {
+      return lines[i].slice(SYSTEM_PREFIX.length).trim() || null;
+    }
```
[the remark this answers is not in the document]
awkoy: The "last notice wins" behavior is intentional here, so keeping it.

The worker and the reaper are mutually exclusive writers: the reaper only ever touches runs still in a **non-terminal** status, and a worker that has already appended its own `[System]` error has by then marked the run terminal — so any given run only ever carries a single `[System]` line. There is no case where a reaper notice overwrites earlier worker context.

The panel is also deliberately a single, concise, high-level message; concatenating multiple lines with "; " would work against that, and the full traceback/exception detail is always one click away in "View logs". So the end-scan returning the most recent `[System]` line is the intended design.

apps/opik-backend/src/main/java/com/comet/opik/domain/OptimizationDAO.java
```diff
@@ -1132,4 +1180,25 @@ public Mono<Long> batchSetProjectId(Set<UUID> optimizationIds, @NonNull UUID pro
...
+            @NonNull Duration runningTimeout, int limit) {
+        var details = "initializedTimeoutSeconds=%d, runningTimeoutSeconds=%d, limit=%d"
+                .formatted(initializedTimeout.toSeconds(), runningTimeout.toSeconds(), limit);
+        var template = FilterUtils.getSTWithLogComment(FIND_STALLED_STUDIO_OPTIMIZATIONS,
+                "find_stalled_studio_optimizations", "", "", details);
+        return Mono.from(connectionFactory.create())
+                .flatMapMany(connection -> {
+                    var statement = connection.createStatement(template.render())
+                            .bind("initialized_timeout_seconds", initializedTimeout.toSeconds())
+                            .bind("running_timeout_seconds", runningTimeout.toSeconds())
```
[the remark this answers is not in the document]
awkoy: Looked into this closely — keeping it as-is, because it follows the established convention of this DAO layer rather than deviating from it.

The `Mono.from(connectionFactory.create()).flatMapMany(...)` shape with no explicit close is used **16× in `OptimizationDAO` alone and across 13 DAOs repo-wide**, with **zero** uses of `usingWhen`/`Connection::close` anywhere in the codebase. `findStalledStudioOptimizations` is intentionally the same shape as `batchSetProjectId` two methods above it, and as the trace/span/experiment ingestion DAOs that run at far higher volume than a 5-minute reaper. Connection release is handled by the ClickHouse R2DBC driver on stream termination — if this leaked, those hot ingestion paths would drain the pool long before the reaper ever mattered.

Adding a lone `Mono.usingWhen(...)` here would make this one method inconsistent with the entire layer. If we want to move to explicit resource scoping, it's worth doing consistently across the DAO layer in its own change rather than one-off here.

awkoy:
**Note for reviewers — the `Python SDK E2E Tests 3.10` failure is a repo-wide breakage, not this PR.**

The two failing tests (`test_test_suite_agentic__assertion_about_span_name__passes`, `test_test_suite_agentic__assertion_requires_buried_keyword_lookup__passes`, both `Expected items_passed=1, got 0`) fail with the **identical signature on `main`** ([run](https://github.com/comet-ml/opik/actions/runs/29249502971)) and on unrelated open PRs (e.g. [OPIK-7321](https://github.com/comet-ml/opik/actions/runs/29255751881), [OPIK-7322](https://github.com/comet-ml/opik/actions/runs/29257878227)). `main` was last green on Jul 10; a re-run here reproduces the failure deterministically.

Root cause (from the job traceback):

```
OPIK: Failed to call LLM provider, reason: No module named 'fastapi'
  File ".../litellm/main.py", line 4867, in completion
    from litellm.responses.mcp.chat_completions_handler import acompletion_with_mcp
```

`sdks/python/setup.py` has `litellm>=1.79.2` with no upper bound; a new litellm release added a lazy `fastapi` import inside `completion()`, and the Python 3.10 CI env has no `fastapi` — so every `llm_judge` call fails (3.14 resolves a different dep set and passes). Needs an SDK-side fix (cap litellm or add `fastapi` to test deps) in a separate PR.

This PR touches no `sdks/python` files. All checks exercised by this diff are green: backend unit + all 16 integration groups (×2 suites), lint-pr, frontend, and Python backend tests.


[OPIK-7321]: https://comet-ml.atlassian.net/browse/OPIK-7321?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OPIK-7322]: https://comet-ml.atlassian.net/browse/OPIK-7322?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/OptimizationStalledReaperJob.java
```diff
@@ -0,0 +1,112 @@
...
+        if (interrupted.get()) {
+            log.info("Optimization stalled reaper job interrupted before execution, skipping");
+            return;
+        }
+
+        var subscription = lockService.bestEffortLock(
+                JOB_LOCK,
+                Mono.defer(() -> {
+                    if (interrupted.get()) {
+                        log.info("Optimization stalled reaper interrupted before processing, skipping");
```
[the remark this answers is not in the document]
awkoy: Intentional — `OptimizationStalledReaperJob` deliberately mirrors the existing `StreamConsumerReaperJob` lock/interrupt lifecycle so the two distributed reapers behave identically. Extracting a shared base would also have to change `StreamConsumerReaperJob` (an unrelated job outside this PR's scope), so I'd rather do that as a separate, focused refactor than widen this PR. Keeping it consistent with the established pattern for now.

apps/opik-frontend/src/v2/pages/OptimizationsPage/OptimizationsNewPage/OptimizationsNewPageContent.tsx
```diff
@@ -45,6 +49,16 @@ const OptimizationsNewPageContent: React.FC<
...
 
+  // A model is selected but its provider has no configured API key: the run
+  // will fail server-side with an auth error. We can only detect *missing*
+  // keys client-side (invalid keys need a backend check — follow-up).
+  // Only flag this once the keys query has settled to avoid a transient flash.
+  const isMissingProviderKey =
+    providerKeysReady &&
+    Boolean(model) &&
+    availableModels.length > 0 &&
+    !availableModels.includes(model);
```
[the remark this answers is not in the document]
awkoy: Good catch — fixed in 68cb417931. Dropped the `availableModels.length > 0` guard so the selected model is compared against the current provider-backed set directly; a stale/dirty `model` that survives a provider-key refetch (including when the set empties) now shows the key warning and disables submit. Updated the unit test to assert the warning + disabled submit for that case.

apps/opik-backend/src/main/java/com/comet/opik/domain/OptimizationService.java
```diff
@@ -489,17 +529,34 @@ private String resolveProjectNameForJob(Optimization optimization, String worksp
...
+                .status(OptimizationStatus.ERROR)
                 .build();
 
-        update(optimizationId, optimizationUpdate)
-                .contextWrite(ctx -> ctx.put(RequestContext.WORKSPACE_ID, workspaceId))
+        logSyncService.appendSystemLogLine(workspaceId, optimizationId, reason)
+                .then(Mono.defer(() -> update(optimizationId, optimizationUpdate)))
+                // Headless path: seed BOTH keys. getById/update resolve through makeFluxContextAware /
+                // bindUserNameAndWorkspaceContextToStream, which call ctx.get(USER_NAME) and throw
+                // NoSuchElementException if it is absent — so WORKSPACE_ID alone silently fails the
```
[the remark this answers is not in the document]
awkoy: Fixed in 68cb417931 — extracted two shared helpers on `OptimizationServiceImpl`: `appendSystemReasonAndMarkError(workspaceId, id, reason)` (append the `[System]` line → mark ERROR) and `headlessSystemContext(workspaceId)` (the WORKSPACE_ID + USER_NAME context seeding, with its rationale now documented in one place). Both `markOptimizationFailedToStart` and `markStalledOptimizationAsError` use them, so the log/update/context plumbing lives in a single spot.

sdks/opik_optimizer/src/opik_optimizer/core/evaluation.py
```diff
@@ -481,6 +492,23 @@ def _validate_objective_scores(
...
+        logger.error(
+            "Objective metric '%s' failed on %s/%s evaluation item(s) "
+            "(>= %.0f%% failure threshold); aborting run. First reasons: %s",
+            objective_metric_name,
+            len(failed_scores),
+            len(scores),
+            DEFAULT_SCORING_FAILURE_THRESHOLD * 100,
+            summarized_reasons,
+        )
+        raise ScoringFailedError(
```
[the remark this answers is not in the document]
awkoy: The raise (instead of the `0.0` fallback) is intentional — it's the OPIK-7029 fix: a run the judge can't score should fail loudly (ERROR) rather than COMPLETE with a misleading `0.0`. What was missing was the *public contract*, so in 68cb417931 I exported `ScoringFailedError` from `opik_optimizer` (added to `__init__.py` + `__all__`). It already carries the failed/total counts and a docstring, so callers that want the old soft behavior can catch it explicitly.

apps/opik-backend/src/main/java/com/comet/opik/utils/JsonUtils.java
```diff
@@ -114,6 +114,27 @@ public static ObjectNode createObjectNode() {
...
+     * Non-object inputs are treated as empty; if both are null/non-object, {@code overrides} is returned
+     * as-is (or {@code base} when {@code overrides} is null).
+     */
+    public static JsonNode merge(JsonNode base, JsonNode overrides) {
+        if (overrides == null) {
+            return base;
+        }
+        if (base == null || !base.isObject()) {
+            return overrides;
+        }
```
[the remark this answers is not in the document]
awkoy: Fixed in 68cb417931 — `merge()` now ignores non-object `overrides` (a scalar/array/null override returns `base` unchanged) and discards a non-object `base` rather than merging onto it, so a non-object value can never be propagated into the metadata column. Added `JsonUtilsTest` covering object-merge (keys preserved + overridden), scalar/array-override-ignored, non-object-base, and null cases.

apps/opik-backend/src/main/java/com/comet/opik/domain/OptimizationService.java
```diff
@@ -334,8 +350,35 @@ public Mono<Long> update(@NonNull UUID id, @NonNull OptimizationUpdate update) {
...
+
+                    // Merge any incoming metadata onto the existing metadata BEFORE handing it to the DAO,
+                    // so the new ReplacingMergeTree row carries the full object (provided keys overwrite,
+                    // existing keys like optimizer/model are preserved). When update.metadata() is null the
+                    // effective update stays null and the DAO carries the existing column forward untouched
+                    // — this keeps Wave-0 status-only updates from wiping metadata (OPIK-7159 regression risk).
+                    OptimizationUpdate effectiveUpdate = update.metadata() == null
+                            ? update
+                            : update.toBuilder()
+                                    .metadata(JsonUtils.merge(optimization.metadata(), update.metadata()))
```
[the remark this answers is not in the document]
awkoy: Metadata is effectively single-writer per optimization: the only update path that carries `metadata` is the worker's completion callback (`scoring_health`), which fires once per run. The stalled-run reaper and every other status transition send `metadata = null`, so the DAO carries the existing column forward untouched — there's no concurrent metadata writer to lose against the read-modify-write. If we introduce concurrent metadata writers later, we'd move the merge into ClickHouse (or add a version guard); flagging that for when that day comes, but it isn't reachable today.

apps/opik-python-backend/src/opik_backend/jobs/optimizer_runner.py
```diff
@@ -361,6 +362,43 @@ def main():
...
+            # can show an exact failed/total count.  Guard against older SDK
+            # versions that do not set this field: never crash the completion
+            # path over a missing or malformed count.
+            try:
+                details = getattr(result, "details", None) or {}
+                scoring_health_raw = (
+                    details.get("scoring_health") if isinstance(details, dict) else None
+                )
+                scoring_health: ScoringHealth | None = None
+                if (
```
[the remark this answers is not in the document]
awkoy: Fixed in 68cb417931 — extracted the logic into `extract_scoring_health(result)` in `studio/types.py` (returns `None` on absent/malformed details, never raises) and had `optimizer_runner` call it. Added `test_studio_types.py` with 10 cases: valid counts, all-passed (0 failed), missing key, missing/None/non-dict details, malformed non-dict / partial / non-int counts, and a result whose `.details` raises.

awkoy:
### CI note for reviewers — the red E2E checks are a repo-wide env issue, not this PR

Two check families are red on this PR; **both are the known repo-wide litellm→fastapi CI-environment gap, not a defect in this change**:

- `Opik Optimizer E2E Tests` (3.10 / 3.11 / 3.13)
- `Python SDK E2E Tests 3.10`

**Root cause (identical in both):**
```
from fastapi import HTTPException, Request
ModuleNotFoundError: No module named 'fastapi'
  .../site-packages/litellm/proxy/litellm_pre_call_utils.py:9
```
litellm's proxy module imports `fastapi`, which isn't installed in these E2E environments, so no LLM call can run.

**Why it's not this PR:**
- This change adds **no** `fastapi`/`litellm` dependency — the failing import is internal to litellm.
- The test-level exception is `ModuleNotFoundError`, not anything introduced here (e.g. the new `ScoringFailedError`).
- The same signature is failing on unrelated PRs — e.g. `litellm_tests` on #7460 and `Python SDK E2E 3.10` on #7458 / #7464.
- The Optimizer E2E workflow only runs here because this PR touches `opik_optimizer`; it then hits the broken env.
- The **unit** suites are green — `Opik Optimizer Unit Tests` (3.10–3.14), backend, python-backend, frontend — and a real `optimize_prompt` run against a live stack with this exact code succeeded end-to-end.

**Not fixed here on purpose:** this is a repo-wide infra issue (spans SDK E2E, litellm/adk/openai integration tests too) best resolved centrally (pin litellm, or add `fastapi` to the E2E test deps), rather than patched in this feature PR.

awkoy:
♻️ **Rebased onto `main` + rebuilt.** The single-run overview UI this PR was stacked on has merged to main, so I rebased and kept **only** the 7159 (stalled-run reaper) + 7029 (error surfacing + scoring-health) work. The diff dropped from ~360 files to **56**. Prior review comments reference commits that no longer exist after the rebase — their fixes are carried in the new commits (DI convention, reaper test, builder contract, JsonUtils non-object guard, worker metadata test, etc.). Where main had independently evolved the error panel, I kept main's version (no regression).

apps/opik-frontend/src/v2/pages/OptimizationPage/optimizationOverviewHelpers.test.ts
```diff
@@ -102,3 +199,139 @@ describe("getOptimizationRefetchInterval", () => {
...
+    // total_count === 0 → the exact-count branch is skipped and heuristic is used.
+    const health: OptimizationScoringHealth = {
+      failed_count: 0,
+      total_count: 0,
+    };
+    // With total_count === 0, the guard `total_count > 0` is false, so we
+    // return the heuristic fallback (non-null).
+    const msg = getEmptyRunWarningMessage(health);
+    expect(msg).not.toBeNull();
+    expect(msg).toContain("no usable scores");
```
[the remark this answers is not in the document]
awkoy: Good catch — fixed in c28303e261. Renamed the test to "falls back to the heuristic message when total_count is 0" to match the actual (intended) contract: a degenerate `total_count:0` skips the exact-count branch and returns the non-null heuristic message. Behavior was correct; only the title was misleading.

apps/opik-backend/config.yml
```diff
@@ -1013,6 +1013,39 @@ optimizationLogs:
...
+  # Default: 5m
+  # Description: A run stuck in INITIALIZED longer than this is marked ERROR (the worker should call
+  #   mark_running within seconds; kept above normal queue latency to tolerate a backlog)
+  initializedTimeout: ${OPTIMIZATION_STALLED_REAPER_INITIALIZED_TIMEOUT:-5m}
+  # Default: 8h
+  # Description: A run stuck in RUNNING longer than this is marked ERROR. MUST exceed the worker's max
+  #   execution timeout (OPTSTUDIO_EXECUTION_TIMEOUT, default 6h = 21600s) plus a buffer, since the row
+  #   has no per-progress heartbeat and last_updated_at only advances on a status change. If you raise
+  #   the worker timeout, raise this too or healthy long runs will be reaped.
+  runningTimeout: ${OPTIMIZATION_STALLED_REAPER_RUNNING_TIMEOUT:-8h}
```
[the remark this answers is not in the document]
awkoy: Intentional coupling, and it's documented inline: `runningTimeout` MUST stay above the worker's `OPTSTUDIO_EXECUTION_TIMEOUT`. They live in separate services (Java backend config vs Python worker env), so they can't share one value — the defaults are consistent (8h > 6h), and the comment explicitly says to raise both together. A cross-service runtime assert isn't feasible here; the doc note is the mitigation.

apps/opik-python-backend/src/opik_backend/studio/config.py
```diff
@@ -16,8 +16,9 @@
 DEFAULT_REFERENCE_KEY = "answer"
 DEFAULT_CASE_SENSITIVE = False
 
-# Execution timeout for optimization jobs (default: 24 hours)
-OPTIMIZATION_TIMEOUT_SECS = int(os.getenv("OPTSTUDIO_EXECUTION_TIMEOUT", "86400"))
+# Execution timeout for optimization jobs (default: 6 hours). The backend
+# stalled-run reaper's runningTimeout MUST stay above this (config.yml, default 8h).
+OPTIMIZATION_TIMEOUT_SECS = int(os.getenv("OPTSTUDIO_EXECUTION_TIMEOUT", "21600"))
```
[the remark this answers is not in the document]
awkoy: Deliberate — 24h per studio run was excessive; lowered the default to 6h (covers real runs, and a dead worker is reaped in ~8h instead of a day). It stays env-overridable via `OPTSTUDIO_EXECUTION_TIMEOUT` for any deployment that needs longer (raise the reaper's `runningTimeout` with it).

sdks/opik_optimizer/src/opik_optimizer/core/evaluation.py
```diff
@@ -481,6 +506,23 @@ def _validate_objective_scores(
...
+            failed=len(failed_scores),
+            total=len(scores),
+            objective_metric_name=objective_metric_name,
+        )
```
alexkuzmik: **Check for a failed run once at the end, not on every attempt**

Right now `_validate_objective_scores` raises `ScoringFailedError` from inside `_evaluate_internal`. That function runs **every time the optimizer scores a candidate prompt** — and an optimization tries many candidates before it finishes. So a failure on any single attempt immediately kills the whole run.

The problem: if the judge model has a brief hiccup during just one attempt (for example, every item hits a rate limit for those few seconds — that marks each item as `scoring_failed=True`), the entire optimization stops with an ERROR and throws away the good candidates it already found. On `main`, that one bad attempt just logged a warning, treated the score as `0.0`, and the run kept going and finished with the best prompt it had found so far.

The `1.0` threshold means this only triggers when *100% of items failed in a single attempt* — but a short outage that hits every item in one attempt is a realistic scenario.

**Suggested fix** — make the decision one time, after the run finishes, using the count we already track for the best prompt (`context.scoring_health`):

1. Change `_validate_objective_scores` back to only logging a warning (its behavior on `main`), so a single bad attempt no longer stops the run.
2. Do the pass/fail check once in `build_final_result`, right where `details["scoring_health"]` is already set:

```python
scoring_health = context.scoring_health or {"failed_count": 0, "total_count": 0}
details["scoring_health"] = scoring_health

failed, total = scoring_health["failed_count"], scoring_health["total_count"]
if total > 0 and failed / total >= DEFAULT_SCORING_FAILURE_THRESHOLD:
    raise ScoringFailedError(
        failed=failed, total=total, objective_metric_name=context.metric.__name__
    )
```

Why this is better:
- **Doesn't throw away good work** — if one attempt fails but a later attempt succeeds, that later success becomes the new best and the run finishes normally.
- **Still fails loudly when it should** — it errors only when the prompt the run actually returns couldn't be scored at all, which is exactly the case OPIK-7029 is about (a run that "completes" but is really empty).
- **More accurate** — the count distinguishes "the judge errored" from "the prompt genuinely scored 0.0." Both average out to `0.0`, so a check based on the score alone can't tell them apart, but the `failed_count` can.
- **No knock-on changes** — it still raises the same `ScoringFailedError(failed, total)`, so the worker's error message ("N of M items failed to score") is unchanged; only *when* we check moves.

Safe edge case: runs that never record a score (custom evaluation, or an empty dataset) leave `total_count` at 0, so the check is skipped and never fires a false error.

It'd also be worth adding a test that covers "one attempt fails, the next recovers, run still completes" — the current tests only check `_validate_objective_scores` on its own, not how it behaves across a full run.

awkoy: Great catch, fully agree — done exactly as suggested in 26e77b231b:

1. `_validate_objective_scores` is now log-only (its behavior on `main`), so a single all-failed attempt no longer aborts the run.
2. The pass/fail decision is made once in `build_final_result` against `context.scoring_health` — raises `ScoringFailedError(failed, total)` only when the returned prompt couldn't be scored at all (`total == 0` skipped). Same exception, so the worker's "N of M items failed to score" message is unchanged.

Also added the end-to-end test you asked for ("one attempt fails every item, the next recovers → run completes, good candidate kept"), plus build_final_result raise/partial/total==0 cases. Full SDK suite green (930 passed).

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/OptimizationStalledReaperJob.java
```diff
@@ -0,0 +1,112 @@
...
+
+    private final @NonNull OptimizationService optimizationService;
+    private final @NonNull LockService lockService;
+    private final @NonNull @Config("optimizationStalledReaper") OptimizationStalledReaperConfig config;
+
+    private final AtomicBoolean interrupted = new AtomicBoolean(false);
+
+    /** Tracks the in-flight reactive pass so {@link #interrupt()} can dispose it. */
+    private final AtomicReference<Disposable> currentExecution = new AtomicReference<>();
+
```
thiagohora, line 52: **[medium] @Config injected via Lombok-generated constructor instead of an explicit @Inject constructor**

This class uses @RequiredArgsConstructor(onConstructor_ = @Inject) with the guicey `@Config("optimizationStalledReaper")` qualifier placed on the field (line 54). The directly analogous sibling this file was cloned from — StreamConsumerReaperJob, also a lock+reaper InterruptableJob scheduled via the same GuiceJobFactory path (scheduleRepeatingJob → JobBuilder.newJob) — deliberately declares an explicit @Inject constructor with the @Config qualifier on the constructor parameter. The team's standing convention is to place @Config on an explicit @Inject constructor parameter rather than relying on Lombok to propagate the config qualifier onto the generated constructor, precisely because Quartz/Guice job instantiation of this qualifier has been fragile here before.

💡 Drop @RequiredArgsConstructor(onConstructor_ = @Inject) and add an explicit `@Inject public OptimizationStalledReaperJob(@NonNull OptimizationService optimizationService, @NonNull LockService lockService, @NonNull @Config("optimizationStalledReaper") OptimizationStalledReaperConfig config)` constructor, matching StreamConsumerReaperJob and LocalRunnerReaperJob.

_rule: `backend-style-maintainability-use-lombok-builders-records-and-di-annotations-consistently-`_

apps/opik-backend/src/main/java/com/comet/opik/domain/OptimizationDAO.java
```diff
@@ -224,6 +238,42 @@ ORDER BY (workspace_id, dataset_id, id) DESC, last_updated_at DESC
...
+            FROM (
+                SELECT
+                    id,
+                    workspace_id,
+                    status,
+                    studio_config,
+                    last_updated_at
+                FROM optimizations
+                ORDER BY id DESC, last_updated_at DESC
+                LIMIT 1 BY id
```
thiagohora: **[high] FIND_STALLED_STUDIO_OPTIMIZATIONS is a full table scan every reaper cycle**

The reaper query dedups the WHOLE optimizations table (`ORDER BY id DESC, last_updated_at DESC LIMIT 1 BY id`) and only then filters studio_config/status/last_updated_at. None of those predicates are in the table's ORDER BY (workspace_id, dataset_id, id), and the reaper is intentionally cross-workspace, so the primary index prunes nothing -> a full read + full sort of `optimizations` on every 5-minute cycle, fleet-wide, growing unbounded with the table. `LIMIT :limit` is applied AFTER the scan+sort, so batchSize bounds rows returned, not rows read. A minmax/skip index cannot help because the selective filters run above the LIMIT 1 BY dedup. (Per clickhouse-best-practices: schema-pk-filter-on-orderby, query-index-skipping-indices.)

💡 Add a pushdown-safe look-back FLOOR and a minmax skip index. Only the lower bound is safe to push below the dedup (the newest version has the max last_updated_at, so a `>=` floor never drops it; the timeout upper-bound and status filter MUST stay above the dedup or a recent terminal version gets dropped and a stale running one resurrected -> false-positive reap). studio_config is immutable per id, so it is also pushdown-safe. e.g.:
  SELECT id, workspace_id, status FROM (
    SELECT id, any(workspace_id) AS workspace_id, argMax(status,last_updated_at) AS status, max(last_updated_at) AS last_updated_at
    FROM optimizations
    WHERE studio_config != '' AND last_updated_at >= subtractSeconds(now64(6), :lookback_seconds)
    GROUP BY id)
  WHERE (status='initialized' AND last_updated_at < subtractSeconds(now64(6), :initialized_timeout_seconds))
     OR (status='running'     AND last_updated_at < subtractSeconds(now64(6), :running_timeout_seconds))
  ORDER BY last_updated_at ASC LIMIT :limit
Add `INDEX idx_last_updated_at last_updated_at TYPE minmax GRANULARITY 1` (range filter). Choose lookback > runningTimeout + reaper-downtime tolerance. Alternatively, track in-flight runs in a Redis ZSET (as OPIK-7246 did) so the reaper is O(stuck runs).

_rule: `clickhouse-full-table-scan-no-pk-prefix`_

apps/opik-backend/src/main/java/com/comet/opik/domain/OptimizationService.java
```diff
@@ -334,8 +351,35 @@ public Mono<Long> update(@NonNull UUID id, @NonNull OptimizationUpdate update) {
...
```
thiagohora, line 392: **[high] metadata merge is an unguarded read-modify-write -> lost update under concurrency**

update() reads the current row (getById), merges incoming metadata in app memory (JsonUtils.merge(optimization.metadata(), update.metadata())), then inserts a full new ReplacingMergeTree version. There is no per-optimization lock, no CAS / `WHERE last_updated_at=:expected`. Two concurrent partial metadata updates both read the same base and each write a full object; RMT keeps whichever has the larger last_updated_at (set to now64(6) at write), so the loser's keys are silently lost. This is amplified by prod ClickHouse having NO read-your-own-writes (2-replica, insert_quorum=0, select_sequential_consistency=0, async insert): getById can read a base missing keys another node just wrote, then clobber them - widening the window from thread-timing to replication lag. The same stale-read root cause makes the isTerminalOverwrite / cancellation guards TOCTOU (evaluated on the stale read, written non-atomically).

💡 Preferred: don't read-modify-write metadata across the CH boundary - merge server-side inside the INSERT...SELECT (combine the existing metadata column with the incoming delta via CH JSON functions), or store metadata deltas merged at read time (like feedback scores). Interim: wrap getById->merge->update in a per-optimization LockService lock (the service already uses LockService for the reaper) - but note a lock alone does not fully close the window given no read-your-writes, so pair it with sync-insert/single-replica routing for these writes, or document that metadata must have a single writer.

_rule: `read-modify-write-lost-update`_

apps/opik-backend/src/main/java/com/comet/opik/api/OptimizationUpdate.java
```diff
@@ -1,12 +1,17 @@
...
 import lombok.Builder;
 
 @Builder(toBuilder = true)
 @JsonIgnoreProperties(ignoreUnknown = true)
 @JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
-public record OptimizationUpdate(String name, OptimizationStatus status) {
+public record OptimizationUpdate(
+        String name,
+        OptimizationStatus status,
+        @Schema(description = "Optional metadata to merge into the optimization's existing metadata. Provided keys are added or overwritten; existing keys not present here are preserved.") JsonNode metadata) {
```
thiagohora: **[low] metadata OpenAPI schema diverges from Optimization.metadata**

The same logical field is modeled inconsistently across the read/write DTOs: Optimization.metadata declares @Schema(implementation = JsonListString.class) (anyOf {Object, Object[], String}), while OptimizationUpdate.metadata declares only a description, so the generated OpenAPI/SDK type for the update body will be a generic object/JsonNode rather than the shared JsonListString type. This makes the update model's advertised metadata shape differ from the create/read model for no documented reason.

💡 Reuse the shared schema type for consistency, e.g. `@Schema(implementation = JsonListString.class, description = "...") JsonNode metadata`, or intentionally document the narrower (object-only) shape the merge actually accepts.

_rule: `backend-api-contract-keep-the-openapi-spec-and-generated-sdks-in-sync-with-every-`_

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/OptimizationStalledReaperJob.java
```diff
@@ -0,0 +1,112 @@
...
+    @Override
+    public void doJob(JobExecutionContext context) {
+        if (interrupted.get()) {
+            log.info("Optimization stalled reaper job interrupted before execution, skipping");
+            return;
+        }
+
+        var subscription = lockService.bestEffortLock(
```
thiagohora, line 73: **[medium] New reaper job has no unit test**

This is a new scheduled reaper with non-trivial control flow (interrupt guards, best-effort lock, error resume, count-based logging), yet no OptimizationStalledReaperJobTest exists, whereas the sibling StreamConsumerReaperJob ships with StreamConsumerReaperJobTest. Untested branches include the interrupt-before-execution skip and the reconcile-invoked-with-config-durations path.

💡 Add an OptimizationStalledReaperJobTest that verifies doJob() invokes reconcileStalledStudioOptimizations with the configured timeouts/batchSize under the lock, skips when interrupted, and (once fixed) resets the interrupt flag between runs.

_general finding_

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/OptimizationsResource.java
```diff
@@ -213,16 +213,6 @@ public Response updateOptimizationsById(@PathParam("id") UUID id,
 
     // ==================== Studio Endpoints ====================
```
thiagohora, line 214: **[low] Deleted /studio/{id}/cancel stub still advertised in openapi.yaml + generated SDKs (contract drift)**

Cancellation itself is fine — it's implemented in this PR via PATCH update (status=CANCELLED -> Redis opik:cancel:{id} signal the worker polls), and the deleted GET /studio/{id}/cancel was only a 501 stub. The issue is purely contract drift: the stub is removed from the backend but NOT from the published contract, so sdks/code_generation/fern/openapi/openapi.yaml (path /v1/private/optimizations/studio/{id}/cancel, operationId cancelStudioOptimizations) and the generated Python (raw_client.cancel_studio_optimizations) and TypeScript (client.cancelStudioOptimizations) SDKs still declare a route that no longer exists — callers of the generated method now get 404 instead of 501.

💡 Regenerate openapi.yaml + the Fern Python/TS SDKs in this PR to drop cancelStudioOptimizations, and confirm the new update-based cancel (PATCH status=CANCELLED) is cleanly exposed in the SDK so consumers have a supported cancellation path.

_rule: `backend-api-contract-keep-the-openapi-spec-and-generated-sdks-in-sync-with-every-`_

apps/opik-backend/src/main/java/com/comet/opik/domain/OptimizationService.java
```diff
@@ -545,4 +629,58 @@ public Mono<OptimizationStudioLog> generateStudioLogsResponse(@NonNull UUID opti
...
+
+        log.warn("Reconciling stalled studio optimization '{}' (status '{}') to ERROR: {}",
+                id, stalled.status(), reason);
+
+        // Re-read under the workspace context and only flip if the run is still non-terminal: this
+        // closes the (tiny) race where the worker reports a terminal status between the reaper's query
+        // and this update, which would otherwise overwrite a genuine completion/cancellation with ERROR.
+        return optimizationDAO.getById(id)
+                .filter(current -> CANCELLABLE_STATUSES.contains(current.status()))
+                .flatMap(current -> appendSystemReasonAndMarkError(workspaceId, id, reason).thenReturn(1L))
```
thiagohora: **[low] Race: system 'failed' log line is written even when the run actually completed, and the row is still counted as reconciled**

markStalledOptimizationAsError re-reads and filters on CANCELLABLE_STATUSES to avoid overwriting a genuine terminal outcome, but appendSystemReasonAndMarkError appends the '[System] Optimization failed...' log line BEFORE calling update(), and update() has its own isTerminalOverwrite guard that silently returns Mono.just(0L). If the worker reports COMPLETED/CANCELLED between this filter (getById at line 661) and update()'s internal getById (line 334), the status write is correctly skipped, but the misleading failure line has already been appended to a genuinely-completed run's logs. Furthermore .thenReturn(1L) returns 1 unconditionally, so reconcileStalledStudioOptimizations over-counts this row as a real ERROR transition even though update() no-oped.

💡 Have appendSystemReasonAndMarkError propagate update()'s actual row count instead of thenReturn(1L) (e.g. map rowsUpdated>0 -> 1L, else 0L), and/or append the system log line only after update() confirms the status actually transitioned (append inside the terminal-check success branch).

_general finding_

apps/opik-backend/src/main/java/com/comet/opik/domain/OptimizationService.java
```diff
@@ -545,4 +629,58 @@ public Mono<OptimizationStudioLog> generateStudioLogsResponse(@NonNull UUID opti
...
+     * Best-effort transition of a single stalled run to ERROR. Records the reason in the run's logs first
+     * (so the UI can surface it even when the worker never produced any logs), then reuses the standard
+     * {@link #update} path — which finalizes logs and emits the completion analytics event. Never fails
+     * the overall pass: a single row's failure is logged and counted as 0.
+     */
+    private Mono<Long> markStalledOptimizationAsError(OptimizationDAO.StalledOptimization stalled,
+            Duration initializedTimeout, Duration runningTimeout) {
+        UUID id = stalled.id();
+        String workspaceId = stalled.workspaceId();
+        String reason = buildStalledReason(stalled.status(), initializedTimeout, runningTimeout);
```
thiagohora: **[low] Stalled reason is built from the reaper's stale query status, not the re-read current status**

buildStalledReason is computed from stalled.status() (the value the batch query returned) at line 653, but the actual gate/transition uses the freshly re-read current.status() at line 661-663. If a run moved INITIALIZED -> RUNNING between the query and the re-read, the persisted reason will say 'failed to start: the optimizer worker did not begin processing within N minutes' while the run was in fact RUNNING (should be the 'no activity for over N hours' message). The message then misdescribes what happened.

💡 Compute the reason from the re-read status inside the flatMap (from current.status()) rather than from stalled.status(), so the recorded reason matches the state actually being reconciled.

_general finding_

apps/opik-backend/src/main/java/com/comet/opik/domain/OptimizationService.java
```diff
@@ -545,4 +629,58 @@ public Mono<OptimizationStudioLog> generateStudioLogsResponse(@NonNull UUID opti
...
+        return optimizationDAO.getById(id)
+                .filter(current -> CANCELLABLE_STATUSES.contains(current.status()))
+                .flatMap(current -> appendSystemReasonAndMarkError(workspaceId, id, reason).thenReturn(1L))
+                .switchIfEmpty(Mono.fromSupplier(() -> {
+                    log.info("Skipping stalled reconciliation for optimization '{}': no longer stalled", id);
+                    return 0L;
+                }))
+                .contextWrite(headlessSystemContext(workspaceId))
+                .onErrorResume(error -> {
+                    log.error("Failed to reconcile stalled studio optimization '{}'", id, error);
```
thiagohora: **[low] System-transition WARN/ERROR/INFO logs omit the workspaceId correlation id**

The new system-driven mutation logs — 'Failed to reconcile stalled studio optimization' (line 670), 'Skipping stalled reconciliation' (line 665), and in markOptimizationFailedToStart 'Marked/Failed to mark optimization as ERROR' (lines 560-561) — log only the optimizationId. workspaceId is in scope in every one of these methods and is the key correlation dimension for a multi-tenant background reaper, but it is not logged, making per-tenant triage harder.

💡 Include workspaceId (and status where relevant) in these log statements, e.g. log.error("Failed to reconcile stalled studio optimization '{}' in workspace '{}'", id, workspaceId, error).

_rule: `backend-observability-include-correlation-identifiers-workspaceid-projectid-entity`_

apps/opik-backend/src/main/java/com/comet/opik/domain/optimization/OptimizationLogSyncService.java
```diff
@@ -145,6 +160,33 @@ public Mono<Void> syncLogsToS3(@NonNull String workspaceId, @NonNull UUID optimi
     // TTL to set on Redis keys after finalization (1 hour) - allows late logs to be captured
     private static final long FINALIZATION_TTL_SECONDS = 3600;
 
+    @Override
+    @WithSpan
+    public Mono<Void> appendSystemLogLine(@NonNull String workspaceId, @NonNull UUID optimizationId,
```
thiagohora, line 165: **[low] New appendSystemLogLine method has no test coverage**

OptimizationLogSyncServiceTest has nested suites for SyncLogsToS3, FinalizeLogs, and GzipCompression, but none for the newly added appendSystemLogLine. All three of its branches are observable behavior and go untested: (1) the disabled short-circuit (config.isEnabled()==false must return Mono.empty() without touching Redis), (2) the happy path (line appended to the log list AND last_append_ts bumped so the flusher/finalizer pick it up), and (3) the best-effort error path (a Redis failure is swallowed via onErrorResume so the caller's ERROR transition still proceeds). The best-effort contract is load-bearing — appendSystemReasonAndMarkError relies on it to still run update(ERROR) — and is exactly the kind of negative path the rule requires to be pinned by a test.

💡 Add an AppendSystemLogLineTests nested class covering: disabled config (verify no redisClient interaction), success (verify logList.add(message) and metaMap.put(META_LAST_APPEND_TS, ...) both invoked and Mono completes), and a Redis error on add/put (verify the Mono still completes empty and does not error).

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/OptimizationStalledReaperConfig.java
```diff
@@ -0,0 +1,68 @@
...
+ * @param batchSize maximum number of stalled runs reconciled per cycle, so a large backlog is drained
+ *        over several cycles rather than in one burst.
+ */
+@Builder(toBuilder = true)
+public record OptimizationStalledReaperConfig(
+        boolean enabled,
+        @NotNull @MinDuration(value = 0, unit = TimeUnit.SECONDS) @MaxDuration(value = 30, unit = TimeUnit.MINUTES) Duration startupDelay,
+        @NotNull @MinDuration(value = 1, unit = TimeUnit.MINUTES) @MaxDuration(value = 6, unit = TimeUnit.HOURS) Duration jobInterval,
+        @NotNull @MinDuration(value = 1, unit = TimeUnit.MINUTES) @MaxDuration(value = 24, unit = TimeUnit.HOURS) Duration initializedTimeout,
+        @NotNull @MinDuration(value = 1, unit = TimeUnit.HOURS) @MaxDuration(value = 7, unit = TimeUnit.DAYS) Duration runningTimeout,
```
thiagohora: **[low] runningTimeout @MinDuration floor (1h) permits reaping legitimately in-flight runs**

The javadoc (lines 35-39) and config.yml comment state runningTimeout MUST exceed the worker's max execution timeout (OPTSTUDIO_EXECUTION_TIMEOUT, default 6h) plus a buffer, because the optimization row has no per-progress heartbeat (last_updated_at only advances on a status change). But the validation floor is @MinDuration(value = 1, unit = TimeUnit.HOURS), so an operator can set OPTIMIZATION_STALLED_REAPER_RUNNING_TIMEOUT as low as 1h, pass boot validation, and have the reaper transition still-executing 6h runs to ERROR mid-flight; the worker's later mark_completed/mark_error callback then hits an unexpected terminal state — exactly the failure the rule warns about. The default (8h) is safe, so this only bites on a plausible operator override, hence low; but the validation lower bound directly contradicts the documented safety invariant (the analogue of the lockDuration<jobInterval invariant the author did hard-enforce via @AssertTrue).

💡 Raise runningTimeout @MinDuration to align with the documented worker-timeout default (e.g. 6h) so a below-worker-timeout value fails fast at boot; if the worker timeout genuinely can't be known here, add a comment stating the floor is intentionally loose and why.

_rule: `backend-correctness-size-time-based-windows-and-boundaries-to-match-real-world-a`_

thiagohora (commented):

11 findings — 0 high · 3 medium · 8 low. Suppressed by team conventions: see `suppressed.md`.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/OptimizationStalledReaperJob.java
awkoy: Done in 9978679131 — dropped the Lombok onConstructor_ and declared an explicit @Inject constructor with the @Config("optimizationStalledReaper") qualifier on the parameter, matching StreamConsumerReaperJob / LocalRunnerReaperJob. The Guice/@Config wiring is now also pinned by an injector-resolution assertion (83ceb341a1) so a broken qualifier fails the build.

apps/opik-backend/src/main/java/com/comet/opik/domain/OptimizationDAO.java
awkoy: Done in fd41f698c5 — dedup via GROUP BY id + argMax(status, last_updated_at) / max(last_updated_at); the status + timeout predicates moved to HAVING (post-aggregation, above the dedup), which also avoids ClickHouse pushing an aggregate-referencing predicate into the scan (ILLEGAL_AGGREGATION). Only the two pushdown-safe predicates stay in the WHERE (studio_config != '' and last_updated_at >= now - lookback), and migration 000106 adds a minmax skip index on last_updated_at. lookback = max(timeout) + 7d reaper-downtime margin (documented). Verified end-to-end by OptimizationStalledReaperServiceTest.

apps/opik-backend/src/main/java/com/comet/opik/domain/OptimizationService.java
awkoy: Done (interim) in d01e840c78 — update() now serializes its read-modify-write per optimization via LockService.executeWithLock, so two concurrent partial updates cannot each write a full object off the same base. I went with the interim lock rather than the server-side INSERT...SELECT merge because ClickHouse has no clean key-wise merge for the JSON-string metadata column; documented the residual (prod has no read-your-own-writes, so studio metadata must stay effectively single-writer — the worker). Happy to switch to the server-side merge as a follow-up if you prefer.

apps/opik-backend/src/main/java/com/comet/opik/api/OptimizationUpdate.java
awkoy: Done in 0d2fc7bdff — OptimizationUpdate.metadata now uses @Schema(implementation = JsonListString.class), matching Optimization.metadata; the regenerated SDKs pick up the shared type.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/OptimizationStalledReaperJob.java
awkoy: Done in 9978679131 — added OptimizationStalledReaperJobTest (Mockito, mirrors ExperimentDenormalizationJobTest): verifies doJob() runs reconcileStalledStudioOptimizations with the configured timeouts/batchSize under the lock, and skips entirely when interrupted before execution.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/OptimizationsResource.java
awkoy: Done in 4f67684f91 — regenerated openapi.yaml from the current backend spec (verified the only operation delta is the removed cancel stub, a scoped 19-line diff) and re-ran fern generate; cancel_studio_optimizations / cancelStudioOptimizations are dropped from both the Python and TypeScript SDKs (and the unused CancelStudioOptimizationsRequest TS type deleted). Cancellation still works via the PATCH update path.

apps/opik-backend/src/main/java/com/comet/opik/domain/OptimizationService.java
awkoy: Done in d01e840c78 — appendSystemReasonAndMarkError now maps update()'s real row count (rows > 0 ? 1 : 0, defaultIfEmpty 1) instead of thenReturn(1L), so a row that update() no-ops on via its terminal-overwrite guard is no longer counted as reconciled.

apps/opik-backend/src/main/java/com/comet/opik/domain/OptimizationService.java
awkoy: Done in d01e840c78 — buildStalledReason is now computed from the re-read current.status() inside the flatMap, not the stale query status, so an INITIALIZED -> RUNNING transition between the query and the flip gets the no-activity message rather than failed-to-start.

apps/opik-backend/src/main/java/com/comet/opik/domain/OptimizationService.java
awkoy: Done in d01e840c78 — added workspaceId to the reconcile / skip / failed-to-start log statements.

apps/opik-backend/src/main/java/com/comet/opik/domain/optimization/OptimizationLogSyncService.java
awkoy: Done in 463bbf21d0 — added AppendSystemLogLineTests covering all three branches: disabled short-circuit (no Redis interaction), happy path (line appended + last_append_ts bumped), and the swallowed Redis error (best-effort, so the caller's ERROR transition still proceeds).

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/OptimizationStalledReaperConfig.java
awkoy: Done in 109f0aaa30 — raised runningTimeout @MinDuration to 6h so a below-worker-timeout override fails fast at boot (same intent as the lockDuration < jobInterval @AssertTrue). Pinned by a negative validation test in 6edf691bfe.

apps/opik-frontend/src/v2/pages/OptimizationPage/EmptyRunWarningPanel.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const [open, setOpen] = useState(false);
+  const { data, dataUpdatedAt } = useOptimizationStudioLogs(
+    { optimizationId: optimization.id },
+    { enabled: Boolean(optimization.id), retry: false },
+  );
+
+  const logContent = data?.content ?? "";
+  const logHtml = useMemo(
+    () => convertTerminalOutputToHtml(logContent),
+    [logContent],
```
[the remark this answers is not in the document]
awkoy: False positive — restoreOsc8Links(), dangerouslySetInnerHTML, and safeHttpHref() do not exist in EmptyRunWarningPanel.tsx (verified by grep across the whole file). This panel renders its warning text through React (no HTML sink) and reads logs via useOptimizationStudioLogs; there is no OSC-8 link restoration or raw-HTML injection here, so there is nothing to gate.

sdks/opik_optimizer/src/opik_optimizer/algorithms/parameter_optimizer/parameter_optimizer.py
```diff
@@ -287,7 +287,11 @@ def _evaluate_baseline_score(
...
             self.pre_trial(context, base_prompts)
-            baseline_score = self.evaluate_prompt(
+            # return_evaluation_result=True so the baseline's scoring health is
+            # captured into context.scoring_health — ParameterOptimizer scores both
+            # baseline and candidates via evaluate_prompt directly (never evaluate()),
+            # so this is the only place the OPIK-7029 empty-run guard gets armed for it.
+            baseline_eval = self.evaluate_prompt(
                 prompt=base_prompts,
                 agent=agent,
                 dataset=evaluation_dataset,
```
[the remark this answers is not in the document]
awkoy: Fixed in 0ad4821414 — the full-result builder now carries context.scoring_health into OptimizationResult.details (it bypasses runtime.build_final_result, so it must surface the counts itself), pinned by an assertion in the optuna-reporting test. Note ParameterOptimizer is SDK-only: the Studio exposes only gepa / evolutionary / hierarchical_reflective, which all return AlgorithmResult and flow through build_final_result (guard + display already covered there).

sdks/opik_optimizer/src/opik_optimizer/algorithms/parameter_optimizer/parameter_optimizer.py
```diff
@@ -297,6 +301,10 @@ def _evaluate_baseline_score(
                 experiment_config=experiment_config,
                 n_samples=n_samples,
                 n_samples_strategy=n_samples_strategy or context.n_samples_strategy,
+                return_evaluation_result=True,
+            )
+            baseline_score = self._resolve_baseline_score_and_health(
+                context, baseline_eval
```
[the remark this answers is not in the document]
awkoy: Fixed in 0ad4821414 — _build_early_baseline_result now threads context.scoring_health into the early-result details too, so baseline-only ParameterOptimizer runs no longer drop metadata.scoring_health. Pinned by the early-stop test assertion.

sdks/opik_optimizer/tests/unit/core/test_scoring_health.py
```diff
@@ -0,0 +1,612 @@
...
+
+    ParameterOptimizer scores baseline + candidates via evaluate_prompt directly (never
+    evaluate()), so this helper is the only thing that arms its empty-run guard.
+    """
+
+    def _make_context(self, simple_chat_prompt: Any) -> OptimizationContext:
+        context = make_optimization_context(
+            simple_chat_prompt,
+            dataset=make_mock_dataset(),
+            metric=lambda di, lo: 1.0,
```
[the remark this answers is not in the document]
awkoy: The real baseline path is covered: TestCalculateBaselineSeedsScoringHealth drives the actual BaseOptimizer._calculate_baseline (mocking only evaluate_prompt) across all-failed / partial / float cases, and the shared _resolve_baseline_score_and_health helper is unit-tested directly. As of 0ad4821414 the ParameterOptimizer result paths (early + full) also assert scoring_health is present in details. Happy to add a dedicated ParameterOptimizer._evaluate_baseline_score integration test if you want it, but the wiring is now covered at both the helper and result ends.

alexkuzmik (approved):

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/OptimizationsResource.java
```diff
@@ -213,16 +213,6 @@ public Response updateOptimizationsById(@PathParam("id") UUID id,
...
 
-    @GET
-    @Path("/studio/{id}/cancel")
-    @Operation(operationId = "cancelStudioOptimizations", summary = "Cancel Studio optimizations", description = "Cancel Studio optimizations by id", responses = {
-            @ApiResponse(responseCode = "501", description = "Not Implemented")})
-    @RequiredPermissions(WorkspaceUserPermission.OPTIMIZATION_STUDIO_USE)
-    public Response cancelStudioOptimization(@PathParam("id") UUID id) {
```
thiagohora, line 221: I still think we should implement this at some point; some optimizations may run per minute, so the cancel makes sense.

apps/opik-backend/src/main/java/com/comet/opik/domain/OptimizationDAO.java
```diff
@@ -1134,4 +1200,36 @@ public Mono<Long> batchSetProjectId(Set<UUID> optimizationIds, @NonNull UUID pro
...
                 .reduce(0L, Long::sum);
     }
+
+    // How far back the stalled-run query scans (the last_updated_at FLOOR that lets the minmax skip index
+    // prune granules). Set to the largest timeout plus a generous reaper-downtime margin, so in normal
+    // operation (reaper running every few minutes) the floor is purely a scan bound and never a coverage
+    // gap: a run's last status change is only older than this if the reaper was down longer than the
+    // margin, in which case that run is not reaped (documented tradeoff, review: thiagohora).
+    private static final Duration STALLED_LOOKBACK_MARGIN = Duration.ofDays(7);
+
```
thiagohora, line 1210: This should be configured via XML

thiagohora: Concretely: make this a validated, env-overridable key on `OptimizationStalledReaperConfig` (in `config.yml`), like the sibling `initializedTimeout` / `runningTimeout` / `lockDuration`, rather than a hardcoded `Duration.ofDays(7)`.

It's the only reaper knob that isn't configurable, and it's pure reaper-downtime insurance — the floor only needs to cover `runningTimeout` + max reaper downtime, not size the scan — so ops should be able to tune it without a redeploy. A shorter margin also tightens the skip-index granule pruning, and 7d is generous given how sparse studio activity is.

PR #7429, opened 2026-07-10
[OPIK-6900] [BE] feat: add migration 000101 creating the traces_local_v2 shadow table

andrescrz:
## Details

Creates `traces_local_v2`, a weekly-partitioned successor to `traces`, as an empty table alongside the live one — no data cutover happens here. The schema mirrors `traces` so a later column-name `INSERT ... SELECT` maps cleanly, with the create-time-only improvements folded in:

- `PARTITION BY toMonday(id_at)` (weekly), where `id_at` is `MATERIALIZED` from the row's UUIDv7 `id`, so partition placement is a deterministic function of the immutable id and survives ReplacingMergeTree upserts.
- Timestamps at microsecond precision (`DateTime64(6)`); defaults `now64(6)`.
- `end_time`, `ttft` and `duration` are non-`Nullable`, using epoch / NaN sentinels, dropping the per-column null-mask overhead.
- Per-column compression codecs and a raised `index_granularity_bytes` (~40 MiB) so skip indexes prune effectively on these wide rows.
- `INDEX idx_traces_id_at id_at TYPE minmax` for granule-level pruning within a partition, alongside the indexes carried over from `traces`.
- `is_deleted` as the ReplacingMergeTree delete meta-column — landed now (unused) so a future delete path needs no further rewrite.

No `storage_policy` is attached here; the tiered-storage policy is applied by a later, environment-gated migration, keeping this DDL identical across deployments. Codecs are provisional pending the Slice 2/3 staging codec benchmark; since the table is created empty they can be revised at zero cost before cutover. `SHOW CREATE TABLE` was verified against the intended spec.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves OPIK-6900

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: assisted — migration authoring and integration tests, iterated under author review
- Human verification: code review + local TestContainers run

## Testing

- `cd apps/opik-backend && mvn -o test -Dtest=TracesLocalV2TableTest` — 13 tests green.
- Scenarios validated:
  - Full-row column round-trip through the sentinel translation, across every `Source` value.
  - Absent `end_time`/`ttft` → epoch/NaN sentinel → read back as `null`, across every `VisibilityMode` value.
  - `start_time` at the epoch sentinel → `duration` null (guarded formula).
  - Sub-millisecond `duration` precision (microsecond `dateDiff`).
  - `input`/`output` truncation above `truncation_threshold`.
  - Rows land in the id-derived weekly partition for backdated, current, and future ids (proving the partition follows `id_at`, not wall-clock).
- Remaining quality checks (full backend suite, spotless) delegated to CI.

## Documentation

N/A — internal analytics-schema migration; no user-facing or API documentation affected.

### DIFF
--- a/apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
+++ b/apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
@@ -81,6 +81,10 @@ CREATE TABLE IF NOT EXISTS ${ANALYTICS_DB_DATABASE_NAME}.traces_local_v2 ON CLUS
             toFloat64('nan'),
             dateDiff('microsecond', start_time, end_time) / 1000.0) CODEC(ZSTD(1)),
     id_at                DateTime('UTC') MATERIALIZED UUIDv7ToDateTime(toUUID(id)) CODEC(Delta, ZSTD(1)),  -- partition input, see header
+    -- Point lookups by id alone (project unknown) can't use the primary key, and retention/read paths also scan
+    -- id-ranges (id >= .. AND id < ..). id is the sort-key suffix and a UUIDv7 (time-ordered), so values cluster
+    -- within granules: minmax prunes both. Matches idx_spans_id.
+    INDEX idx_traces_id id TYPE minmax GRANULARITY 1,
     INDEX idx_traces_id_at id_at TYPE minmax GRANULARITY 1,  -- granule-level pruning on id_at within a partition
     -- Carried over from traces so the successor keeps the same read performance.
     INDEX idx_traces_source source TYPE set(0) GRANULARITY 1,
--- a/apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
+++ b/apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
@@ -1,0 +1,103 @@
+--liquibase formatted sql
+--changeset andrescrz:000101_create_traces_local_v2_table
+--comment: Create the traces_local_v2 table, a weekly-partitioned successor to traces, empty (no data cutover here)
+
+-- traces_local_v2 is the target layout for `traces` under partitioned + tiered storage. It is created empty
+-- next to the live `traces` table; populating it and swapping it in happen in later migrations. The schema
+-- mirrors `traces` so a later `INSERT INTO traces_local_v2 SELECT ... FROM traces` maps by column name, with
+-- these differences folded in:
+--   * PARTITION BY toMonday(id_at) (weekly). id_at is derived from the row's UUIDv7 id at insert time, so it
+--     is a deterministic function of the immutable id: a row always lands in the same weekly partition across
+--     ReplacingMergeTree upserts.
+--   * Timestamps stored at microsecond precision (DateTime64(6)); nothing ingested needs finer resolution.
+--   * end_time, ttft and duration are non-Nullable, using epoch / NaN sentinels, dropping the per-column
+--     null-mask overhead on hot reads.
+--   * is_deleted is the ReplacingMergeTree delete meta-column, so a future delete path can upsert a tombstone
+--     that merges away during compaction. It is unused for now.
+-- Locked at creation (changing any of these later needs another full table rewrite, so they are fixed now):
+--   * ORDER BY / implicit PRIMARY KEY (workspace_id, project_id, id): matches every hot query (workspace+project
+--     filter, lookup by id) and is the ReplacingMergeTree dedup key; id stays in the key so point lookups keep
+--     granule pruning (a shorter primary key would only shave a tiny in-memory index for a real lookup cost).
+--   * PARTITION BY toMonday(id_at) (weekly, see above).
+--   * ReplicatedReplacingMergeTree version (last_updated_at) + is_deleted meta-columns: the engine and its
+--     parameters are immutable after creation.
+--   * No SAMPLE BY, intentionally: sampling needs a hashed column in the sort key, which would break the id/time
+--     ordering the dominant access path relies on; Opik analytics are workspace/project-scoped and exact, so
+--     sampling would not earn back that cost. Everything else (codecs, TTL, skip indexes, storage policy) is
+--     ALTER-able later, though re-compressing or materializing an index on the full table is a heavy mutation.
+-- Compression (per ClickHouse codec guidance): timestamps use Delta + ZSTD(1) (monotonic, irregularly spaced);
+-- the length counters use T64 + ZSTD(1) (narrow integers); UUID ids and large free-text use ZSTD(3) (compress
+-- once, read many); other variable strings/arrays/floats use ZSTD(1) (the recommended general-purpose baseline).
+-- Fixed-width low-entropy columns (the Enum8s, the UInt8 flag, the constant threshold, the LowCardinality
+-- column and the run-length-friendly leading key workspace_id) are left on the server default, where LZ4 is
+-- comparable and decompresses faster.
+-- The engine uses its own ZooKeeper path ('.../traces_local_v2'): two replicated tables cannot share a replica
+-- path, so it must differ from `traces` while both exist. A replica path is independent of the table name, so
+-- it stays valid after a later rename/swap.
+-- No storage_policy is set here; the tiered-storage policy is attached by a later, environment-gated migration,
+-- keeping this DDL identical across all deployments.
+CREATE TABLE IF NOT EXISTS ${ANALYTICS_DB_DATABASE_NAME}.traces_local_v2 ON CLUSTER '{cluster}'
+(
+    id                   FixedString(36)        CODEC(ZSTD(3)),
+    workspace_id         String,
+    project_id           FixedString(36)        CODEC(ZSTD(3)),
+    name                 String                 DEFAULT ''       CODEC(ZSTD(1)),
+    start_time           DateTime64(6, 'UTC')   DEFAULT now64(6) CODEC(Delta, ZSTD(1)),
+    end_time             DateTime64(6, 'UTC')   DEFAULT toDateTime64('1970-01-01 00:00:00', 6) CODEC(Delta, ZSTD(1)),  -- epoch = not ended yet
+    input                String                 DEFAULT ''       CODEC(ZSTD(3)),
+    output               String                 DEFAULT ''       CODEC(ZSTD(3)),
+    metadata             String                 DEFAULT ''       CODEC(ZSTD(3)),
+    tags                 Array(String)          DEFAULT []       CODEC(ZSTD(1)),
+    created_at           DateTime64(6, 'UTC')   DEFAULT now64(6) CODEC(Delta, ZSTD(1)),
+    last_updated_at      DateTime64(6, 'UTC')   DEFAULT now64(6) CODEC(Delta, ZSTD(1)),
+    created_by           String                 DEFAULT ''       CODEC(ZSTD(1)),
+    last_updated_by      String                 DEFAULT ''       CODEC(ZSTD(1)),
+    error_info           String                 DEFAULT ''       CODEC(ZSTD(1)),
+    thread_id            String                 DEFAULT ''       CODEC(ZSTD(1)),
+    visibility_mode      Enum8('unknown' = 0, 'default' = 1, 'hidden' = 2) DEFAULT 'default',
+    truncation_threshold UInt64                 DEFAULT 10001,   -- 10 KB + 1 byte, threshold for the truncated_* columns
+    input_slim           String                 DEFAULT ''       CODEC(ZSTD(3)),
+    output_slim          String                 DEFAULT ''       CODEC(ZSTD(3)),
+    -- NaN = not measured (0 is a valid value); CH has no nan() literal. ZSTD (not Gorilla/FPC, which need
+    -- correlated float sequences) since the column is dominated by the repeated NaN sentinel.
+    ttft                 Float64                DEFAULT toFloat64('nan') CODEC(ZSTD(1)),
+    source               Enum8('unknown' = 0, 'sdk' = 1, 'experiment' = 2, 'playground' = 3, 'optimization' = 4, 'evaluator' = 5) DEFAULT 'unknown',
+    environment          LowCardinality(String) DEFAULT '',
+    is_deleted           UInt8                  DEFAULT 0,
+    input_length         UInt64  MATERIALIZED length(input)                                                                CODEC(T64, ZSTD(1)),
+    output_length        UInt64  MATERIALIZED length(output)                                                               CODEC(T64, ZSTD(1)),
+    metadata_length      UInt64  MATERIALIZED length(metadata)                                                             CODEC(T64, ZSTD(1)),
+    truncated_input      String  MATERIALIZED if(length(input) >= truncation_threshold, substring(input, 1, truncation_threshold), input)    CODEC(ZSTD(3)),
+    truncated_output     String  MATERIALIZED if(length(output) >= truncation_threshold, substring(output, 1, truncation_threshold), output) CODEC(ZSTD(3)),
+    output_keys          Array(Tuple(key String, type String)) MATERIALIZED
+        arrayMap(
+            key -> tuple(key, toString(JSONType(JSONExtractRaw(output, key)))),
+            JSONExtractKeys(output)
+        ) CODEC(ZSTD(1)),
+    -- NaN when end_time is the epoch sentinel (not ended) or start_time is the epoch (unset, as the live traces
+    -- duration also guards); microsecond diff keeps sub-millisecond spans (0 is a legitimate duration).
+    `duration`           Float64 MATERIALIZED
+        if(end_time = toDateTime64('1970-01-01 00:00:00', 6) OR start_time = toDateTime64('1970-01-01 00:00:00', 6),
+            toFloat64('nan'),
+            dateDiff('microsecond', start_time, end_time) / 1000.0) CODEC(ZSTD(1)),
+    id_at                DateTime('UTC') MATERIALIZED UUIDv7ToDateTime(toUUID(id)) CODEC(Delta, ZSTD(1)),  -- partition input, see header
+    INDEX idx_traces_id_at id_at TYPE minmax GRANULARITY 1,  -- granule-level pruning on id_at within a partition
+    -- Carried over from traces so the successor keeps the same read performance.
+    INDEX idx_traces_source source TYPE set(0) GRANULARITY 1,
+    INDEX idx_traces_thread_id_bf thread_id TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_traces_environment environment TYPE set(0) GRANULARITY 1,
+    INDEX idx_traces_created_at created_at TYPE minmax GRANULARITY 1,
+    INDEX idx_traces_last_updated_at last_updated_at TYPE minmax GRANULARITY 1
+)
+ENGINE = ReplicatedReplacingMergeTree(
+    '/clickhouse/tables/{shard}/${ANALYTICS_DB_DATABASE_NAME}/traces_local_v2',
+    '{replica}',
+    last_updated_at,
+    is_deleted)
+PARTITION BY toMonday(id_at)
+ORDER BY (workspace_id, project_id, id)
+-- ~40 MiB per granule so a granule fills toward the 8192-row target on these wide rows, making skip indexes prune effectively.
+SETTINGS index_granularity = 8192, index_granularity_bytes = 41943040;
+
+--rollback DROP TABLE IF EXISTS ${ANALYTICS_DB_DATABASE_NAME}.traces_local_v2 ON CLUSTER '{cluster}';
+
--- a/apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
+++ b/apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
@@ -81,10 +81,11 @@ CREATE TABLE IF NOT EXISTS ${ANALYTICS_DB_DATABASE_NAME}.traces_local_v2 ON CLUS
             toFloat64('nan'),
             dateDiff('microsecond', start_time, end_time) / 1000.0) CODEC(ZSTD(1)),
     id_at                DateTime('UTC') MATERIALIZED UUIDv7ToDateTime(toUUID(id)) CODEC(Delta, ZSTD(1)),  -- partition input, see header
-    -- Point lookups by id alone (project unknown) can't use the primary key, and retention/read paths also scan
-    -- id-ranges (id >= .. AND id < ..). id is the sort-key suffix and a UUIDv7 (time-ordered), so values cluster
-    -- within granules: minmax prunes both. Matches idx_spans_id.
-    INDEX idx_traces_id id TYPE minmax GRANULARITY 1,
+    -- id-only lookups (project unknown) can't use the primary key. Two complementary skip indexes on the UUIDv7 id:
+    -- minmax for the id-range predicates retention/read paths use (id >= .. AND id < ..), and a bloom filter for the
+    -- exact-match / IN lookups (getPartialById, findByIds, ...), which minmax can't prune within a week's shared prefix.
+    INDEX idx_traces_id_minmax id TYPE minmax GRANULARITY 1,
+    INDEX idx_traces_id_bf id TYPE bloom_filter(0.01) GRANULARITY 1,
     INDEX idx_traces_id_at id_at TYPE minmax GRANULARITY 1,  -- granule-level pruning on id_at within a partition
     -- Carried over from traces so the successor keeps the same read performance.
     INDEX idx_traces_source source TYPE set(0) GRANULARITY 1,
--- a/apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
+++ b/apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
@@ -25,12 +25,12 @@
 --     ordering the dominant access path relies on; Opik analytics are workspace/project-scoped and exact, so
 --     sampling would not earn back that cost. Everything else (codecs, TTL, skip indexes, storage policy) is
 --     ALTER-able later, though re-compressing or materializing an index on the full table is a heavy mutation.
--- Compression (per ClickHouse codec guidance): timestamps use Delta + ZSTD(1) (monotonic, irregularly spaced);
--- the length counters use T64 + ZSTD(1) (narrow integers); UUID ids and large free-text use ZSTD(3) (compress
--- once, read many); other variable strings/arrays/floats use ZSTD(1) (the recommended general-purpose baseline).
--- Fixed-width low-entropy columns (the Enum8s, the UInt8 flag, the constant threshold, the LowCardinality
--- column and the run-length-friendly leading key workspace_id) are left on the server default, where LZ4 is
--- comparable and decompresses faster.
+-- Compression: timestamps use Delta + ZSTD(1) (monotonic, irregularly spaced); the length counters use
+-- T64 + ZSTD(1) (narrow integers); large free-text (input, output, metadata and their slim/truncated forms)
+-- uses ZSTD(3). Everything else variable — the UUID id columns, workspace_id, the constant threshold and the
+-- shorter strings/arrays/floats — uses ZSTD(1), the general-purpose baseline; for the UUIDv7 id columns,
+-- benchmarking showed higher ZSTD levels did not improve (and slightly hurt) the ratio. The Enum8s, the UInt8
+-- flag and the LowCardinality column are left on the server default, where LZ4 is comparable and faster to decode.
 -- The engine uses its own ZooKeeper path ('.../traces_local_v2'): two replicated tables cannot share a replica
 -- path, so it must differ from `traces` while both exist. A replica path is independent of the table name, so
 -- it stays valid after a later rename/swap.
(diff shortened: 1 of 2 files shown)


apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
```diff
@@ -0,0 +1,103 @@
...
+(
+    id                   FixedString(36)        CODEC(ZSTD(3)),
+    workspace_id         String,
+    project_id           FixedString(36)        CODEC(ZSTD(3)),
+    name                 String                 DEFAULT ''       CODEC(ZSTD(1)),
+    start_time           DateTime64(6, 'UTC')   DEFAULT now64(6) CODEC(Delta, ZSTD(1)),
+    end_time             DateTime64(6, 'UTC')   DEFAULT toDateTime64('1970-01-01 00:00:00', 6) CODEC(Delta, ZSTD(1)),  -- epoch = not ended yet
+    input                String                 DEFAULT ''       CODEC(ZSTD(3)),
+    output               String                 DEFAULT ''       CODEC(ZSTD(3)),
+    metadata             String                 DEFAULT ''       CODEC(ZSTD(3)),
```
thiagohora, line 49: I was wondering if we should test `JSON` metadata.I'm sure we could use it, since we always interpret it as a map, but we could consider it for input and output as well if it supports raw strings

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
```diff
@@ -0,0 +1,103 @@
...
+    tags                 Array(String)          DEFAULT []       CODEC(ZSTD(1)),
+    created_at           DateTime64(6, 'UTC')   DEFAULT now64(6) CODEC(Delta, ZSTD(1)),
+    last_updated_at      DateTime64(6, 'UTC')   DEFAULT now64(6) CODEC(Delta, ZSTD(1)),
+    created_by           String                 DEFAULT ''       CODEC(ZSTD(1)),
+    last_updated_by      String                 DEFAULT ''       CODEC(ZSTD(1)),
+    error_info           String                 DEFAULT ''       CODEC(ZSTD(1)),
```
thiagohora, line 55: Same on this one.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
```diff
@@ -0,0 +1,103 @@
...
+    thread_id            String                 DEFAULT ''       CODEC(ZSTD(1)),
+    visibility_mode      Enum8('unknown' = 0, 'default' = 1, 'hidden' = 2) DEFAULT 'default',
+    truncation_threshold UInt64                 DEFAULT 10001,   -- 10 KB + 1 byte, threshold for the truncated_* columns
```
thiagohora: I think we should drop it in favor of the slim version, which is a smarter approach.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
```diff
@@ -0,0 +1,103 @@
...
+    ttft                 Float64                DEFAULT toFloat64('nan') CODEC(ZSTD(1)),
+    source               Enum8('unknown' = 0, 'sdk' = 1, 'experiment' = 2, 'playground' = 3, 'optimization' = 4, 'evaluator' = 5) DEFAULT 'unknown',
+    environment          LowCardinality(String) DEFAULT '',
+    is_deleted           UInt8                  DEFAULT 0,
+    input_length         UInt64  MATERIALIZED length(input)                                                                CODEC(T64, ZSTD(1)),
+    output_length        UInt64  MATERIALIZED length(output)                                                               CODEC(T64, ZSTD(1)),
+    metadata_length      UInt64  MATERIALIZED length(metadata)                                                             CODEC(T64, ZSTD(1)),
+    truncated_input      String  MATERIALIZED if(length(input) >= truncation_threshold, substring(input, 1, truncation_threshold), input)    CODEC(ZSTD(3)),
+    truncated_output     String  MATERIALIZED if(length(output) >= truncation_threshold, substring(output, 1, truncation_threshold), output) CODEC(ZSTD(3)),
```
thiagohora, line 71: Possibly drop it

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
```diff
@@ -0,0 +1,103 @@
...
+            JSONExtractKeys(output)
+        ) CODEC(ZSTD(1)),
+    -- NaN when end_time is the epoch sentinel (not ended) or start_time is the epoch (unset, as the live traces
+    -- duration also guards); microsecond diff keeps sub-millisecond spans (0 is a legitimate duration).
+    `duration`           Float64 MATERIALIZED
+        if(end_time = toDateTime64('1970-01-01 00:00:00', 6) OR start_time = toDateTime64('1970-01-01 00:00:00', 6),
+            toFloat64('nan'),
+            dateDiff('microsecond', start_time, end_time) / 1000.0) CODEC(ZSTD(1)),
+    id_at                DateTime('UTC') MATERIALIZED UUIDv7ToDateTime(toUUID(id)) CODEC(Delta, ZSTD(1)),  -- partition input, see header
+    INDEX idx_traces_id_at id_at TYPE minmax GRANULARITY 1,  -- granule-level pruning on id_at within a partition
```
thiagohora, line 89: I would also add idx_traces_id so that a few lookups, like find by id (when we don't know the project) and workspace, are faster.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
```diff
@@ -0,0 +1,103 @@
...
+-- once, read many); other variable strings/arrays/floats use ZSTD(1) (the recommended general-purpose baseline).
+-- Fixed-width low-entropy columns (the Enum8s, the UInt8 flag, the constant threshold, the LowCardinality
+-- column and the run-length-friendly leading key workspace_id) are left on the server default, where LZ4 is
+-- comparable and decompresses faster.
+-- The engine uses its own ZooKeeper path ('.../traces_local_v2'): two replicated tables cannot share a replica
+-- path, so it must differ from `traces` while both exist. A replica path is independent of the table name, so
+-- it stays valid after a later rename/swap.
+-- No storage_policy is set here; the tiered-storage policy is attached by a later, environment-gated migration,
+-- keeping this DDL identical across all deployments.
+CREATE TABLE IF NOT EXISTS ${ANALYTICS_DB_DATABASE_NAME}.traces_local_v2 ON CLUSTER '{cluster}'
```
thiagohora, line 39: Finally, should we test an initial setup for the text search index?

thiagohora (commented):
This is my initial review, but I'll provide further comments on Monday.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
andrescrz: Agree this is worth exploring — the native `JSON` type could be a real win, especially for `metadata` (it's always a map). Deferring it out of this migration though: switching `metadata`/`input`/`output` from `String` to `JSON` would break the by-column-name `INSERT … SELECT` cutover and the materialized columns that rely on String semantics (`JSONExtractKeys(output)`, `length()`, `substring`), and `input`/`output` aren't always JSON (raw strings), as you note. Let's spike it as a separate change with `metadata` as the first candidate; keeping `String` here to preserve a clean cutover.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
andrescrz: Same here — `error_info` is a good `JSON`-type candidate too and would fold into the same spike. Keeping `String` for the cutover-faithful shape for now.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
andrescrz: Agree that's the right long-term direction. Checked the read path first, though: `truncated_input`/`truncated_output`/`truncation_threshold` are still actively read — `TraceDAO`'s truncate/image path, and `DatasetItemVersionDAO` uses `if(notEmpty(input_slim), input_slim, truncated_input)`, i.e. slim is preferred but truncated is still the fallback. So slim hasn't fully superseded them yet; dropping them is a coordinated app+schema change (retire the DAO fallback first). Keeping them here to mirror `traces` for the cutover, and doing the removal as a follow-up.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
andrescrz: Same as the `truncation_threshold` thread — `truncated_output` is still read by `TraceDAO` (and is the slim fallback in `DatasetItemVersionDAO`), so keeping it for the faithful cutover and tackling the slim-only simplification as a follow-up.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
andrescrz: Good question. Leaving it out of this migration — a full-text/`text` inverted index on `input`/`output` is a new capability beyond the faithful-successor scope and fairly heavy to add on spec. Happy to evaluate it separately (with a benchmark) if/when trace content-search is prioritized; adding it later is a mutation, so we'd weigh that cost then.

andrescrz: One more reason to defer it specifically: the text/inverted-index support is something we'd want to build on once we're on the upcoming ClickHouse 26.3 LTS, so it makes sense to wait for that upgrade to land before evaluating a text-search index here rather than setting one up on the current version.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
thiagohora: Tested the `JSON` type on 25.8 and 26.3 with real trace payloads: since `metadata` is always a map, it round-trips cleanly as `JSON` with field-level sub-column access, mapping the occasional non-object row to the `{}` default. For `input`/`output` the `JSON` type only accepts top-level objects and rejects the raw strings and arrays we actually store, so those have to stay `String`. So it confirms `metadata` (and it's a create-time type decision worth making here); `input`/`output` should remain `String`.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
thiagohora: `error_info` is always a structured object (`exception_type`/`message`/`traceback`), so it maps to `JSON` with no non-object rows to guard — confirmed safe to switch here, same rationale as `metadata`.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
thiagohora: Tested this on real data: the slim columns are the better representation, but they're only populated on the live write path — a large share of existing rows still have empty slim, and since the planned backfill is a column-name `INSERT … SELECT` it copies slim as-is rather than recomputing it, whereas `truncated_*` is `MATERIALIZED` and gets recomputed for every row for free. Since this PR only creates the empty table, suggest we keep `truncation_threshold`/`truncated_*` here and revisit dropping them in the migration/backfill PR, where we can decide whether to recompute slim.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
thiagohora: Same as the `truncation_threshold` thread — keeping `truncated_output` for the create step; we can drop it together with the backfill PR once slim coverage is guaranteed for existing rows.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
thiagohora: Tested a find-by-id lookup that doesn't filter `project_id` on a real dataset migrated into this schema: without an index on `id` the id-only filter can't prune the weekly partitions, so it scans across them; adding a skip index on `id` brings it back to a single granule. Confirms the suggestion — will add it.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
thiagohora: Tested an initial token-search setup on `input` over a real dataset on 25.8 and 26.3: a token/text skip index prunes a selective search down to a single granule, and since the table is created empty it's essentially free to seed now versus a heavy `MATERIALIZE INDEX` on the full table later. So it's worth adding an initial index as part of this create step.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
```diff
@@ -0,0 +1,107 @@
...
+    metadata             String                 DEFAULT ''       CODEC(ZSTD(3)),
+    tags                 Array(String)          DEFAULT []       CODEC(ZSTD(1)),
+    created_at           DateTime64(6, 'UTC')   DEFAULT now64(6) CODEC(Delta, ZSTD(1)),
+    last_updated_at      DateTime64(6, 'UTC')   DEFAULT now64(6) CODEC(Delta, ZSTD(1)),
+    created_by           String                 DEFAULT ''       CODEC(ZSTD(1)),
+    last_updated_by      String                 DEFAULT ''       CODEC(ZSTD(1)),
+    error_info           String                 DEFAULT ''       CODEC(ZSTD(1)),
+    thread_id            String                 DEFAULT ''       CODEC(ZSTD(1)),
+    visibility_mode      Enum8('unknown' = 0, 'default' = 1, 'hidden' = 2) DEFAULT 'default',
+    truncation_threshold UInt64                 DEFAULT 10001,   -- 10 KB + 1 byte, threshold for the truncated_* columns
```
LiorF-BDBQ: `truncation_threshold` can probably benefit from ZSTD(1) codec

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
```diff
@@ -0,0 +1,107 @@
...
+-- The engine uses its own ZooKeeper path ('.../traces_local_v2'): two replicated tables cannot share a replica
+-- path, so it must differ from `traces` while both exist. A replica path is independent of the table name, so
+-- it stays valid after a later rename/swap.
+-- No storage_policy is set here; the tiered-storage policy is attached by a later, environment-gated migration,
+-- keeping this DDL identical across all deployments.
+CREATE TABLE IF NOT EXISTS ${ANALYTICS_DB_DATABASE_NAME}.traces_local_v2 ON CLUSTER '{cluster}'
+(
+    id                   FixedString(36)        CODEC(ZSTD(3)),
+    workspace_id         String,
+    project_id           FixedString(36)        CODEC(ZSTD(3)),
```
LiorF-BDBQ: I'm assuming id and project_id are uuidv7 while workspace_id isn't:
1. I think zstd(3) is an overkill for uuids, zstd(1) will most likely perform the same. I ran a quick benchmark with Claude and for whatever reason zstd(1) outperformed zstd(3). 
    ```
     ┌─────────┬──────────────────────┐
     │  Codec  │ Compressed bytes/row │
     ├─────────┼──────────────────────┤
     │ ZSTD(1) │ 6.4834               │
     ├─────────┼──────────────────────┤
     │ ZSTD(3) │ 7.0161               │
     └─────────┴──────────────────────┘    
    ```
2. workspace_id should also be ZSTD(1)
3. worth noting the uuid type outperforms both, but I remember the sorting issue prevents from adopting it.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
```diff
@@ -0,0 +1,107 @@
...
+    name                 String                 DEFAULT ''       CODEC(ZSTD(1)),
+    start_time           DateTime64(6, 'UTC')   DEFAULT now64(6) CODEC(Delta, ZSTD(1)),
+    end_time             DateTime64(6, 'UTC')   DEFAULT toDateTime64('1970-01-01 00:00:00', 6) CODEC(Delta, ZSTD(1)),  -- epoch = not ended yet
+    input                String                 DEFAULT ''       CODEC(ZSTD(3)),
+    output               String                 DEFAULT ''       CODEC(ZSTD(3)),
+    metadata             String                 DEFAULT ''       CODEC(ZSTD(3)),
```
LiorF-BDBQ, line 49: I would consider going even higher than ZSTD(1) here. Needs benchmarking though.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
```diff
@@ -0,0 +1,107 @@
...
+    -- duration also guards); microsecond diff keeps sub-millisecond spans (0 is a legitimate duration).
+    `duration`           Float64 MATERIALIZED
+        if(end_time = toDateTime64('1970-01-01 00:00:00', 6) OR start_time = toDateTime64('1970-01-01 00:00:00', 6),
+            toFloat64('nan'),
+            dateDiff('microsecond', start_time, end_time) / 1000.0) CODEC(ZSTD(1)),
+    id_at                DateTime('UTC') MATERIALIZED UUIDv7ToDateTime(toUUID(id)) CODEC(Delta, ZSTD(1)),  -- partition input, see header
+    -- Point lookups by id alone (project unknown) can't use the primary key, and retention/read paths also scan
+    -- id-ranges (id >= .. AND id < ..). id is the sort-key suffix and a UUIDv7 (time-ordered), so values cluster
+    -- within granules: minmax prunes both. Matches idx_spans_id.
+    INDEX idx_traces_id id TYPE minmax GRANULARITY 1,
```
LiorF-BDBQ: The `get_partial_trace_by_id` query might benefit from a bloom filter index as well. Worth validating.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
```diff
@@ -0,0 +1,107 @@
...
+-- traces_local_v2 is the target layout for `traces` under partitioned + tiered storage. It is created empty
+-- next to the live `traces` table; populating it and swapping it in happen in later migrations. The schema
+-- mirrors `traces` so a later `INSERT INTO traces_local_v2 SELECT ... FROM traces` maps by column name, with
+-- these differences folded in:
+--   * PARTITION BY toMonday(id_at) (weekly). id_at is derived from the row's UUIDv7 id at insert time, so it
+--     is a deterministic function of the immutable id: a row always lands in the same weekly partition across
+--     ReplacingMergeTree upserts.
+--   * Timestamps stored at microsecond precision (DateTime64(6)); nothing ingested needs finer resolution.
+--   * end_time, ttft and duration are non-Nullable, using epoch / NaN sentinels, dropping the per-column
+--     null-mask overhead on hot reads.
```
LiorF-BDBQ, line 14: This might be a really broad change, we'd need to replace calls that check isNull, isNotNull, ifNull, coalesce, nullIf etc to check for the default value.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
andrescrz: Agreed — and your MATERIALIZED point makes it clearer: `truncated_*` is recomputed for free on every backfilled row, whereas the column-name `INSERT … SELECT` would copy `slim` as-is (empty for old rows). Keeping `truncation_threshold`/`truncated_*` here and revisiting the slim-only simplification in the backfill PR, where we can decide whether to recompute slim.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
andrescrz: Same here — keeping `truncated_output` for the create step and folding its removal into the backfill PR once slim coverage is guaranteed for existing rows.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
andrescrz: `metadata` is already `ZSTD(3)` here, so we're aligned on going above ZSTD(1) for it. Happy to push it higher if the staging codec benchmark shows a win; leaving it at ZSTD(3) for now.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
andrescrz: Thanks for testing this — agree `metadata` and `error_info` are the natural JSON candidates and that `input`/`output` must stay `String`. After digging into the cutover and the query layer, I'd keep all four as `String` here and treat JSON as a focused follow-on:

- **Cutover risk at scale.** `INSERT … SELECT` String→JSON auto-parses, but across the full table we'd hit empty/malformed values, non-object top-level (arrays/scalars JSON rejects), cross-row type conflicts, and String→JSON→String display round-trips that aren't guaranteed byte-identical (key order, formatting). Hard to de-risk per-customer, and bundling it into the partitioning cutover multiplies that blast radius.
- **The query layer is built on `String`.** Dynamic filters use `JSON_VALUE`/`JSONExtractRaw`/`JSONExtractArrayRaw(metadata, …)` (and `simpleJSONExtractString(error_info, 'exception_type')`), shared across traces/spans/experiments/etc. Moving to the JSON type means re-implementing every operator on native paths and proving identical results while both tables coexist. `metadata_length = length(metadata)` also needs rework — `length()` isn't defined on JSON.
- **ROI on `metadata` is unproven.** The JSON type pays off with bounded path cardinality; user-supplied `metadata` has effectively unbounded keys across customers, so it can hit `max_dynamic_paths` and fall back to shared storage. Worth measuring on real data first.
- **Nothing is lost by deferring.** Unlike `PARTITION BY`/`ORDER BY`, a column type change isn't create-time-only. Later it's a bounded **per-column** `ALTER` on the already-partitioned table (add a JSON column, backfill partition-scoped or forward-only, migrate reads, drop the old) — no second cutover. A `MATERIALIZED` JSON column is a good parity bridge (run old vs new filters side by side before flipping; can't drop the String source while it's the materialized source, so that's the validation phase, not the end state).

So: keeping `String` here, scoping JSON to a post-26.3 spike that benchmarks compression/query on real path cardinality and proves filter parity. Recording it as a deferred follow-on so we don't lose the idea.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
andrescrz: Same call as the `metadata` thread (full reasoning there). `error_info` is actually the cleaner JSON candidate — stable `exception_type`/`message`/`traceback` schema, no non-object rows, no materialized dependency — but its only query use is the `simpleJSONExtractString(error_info, 'exception_type')` filter and it's low-volume, so the ROI is small and it would still need the same cross-cutover filter-parity handling. Keeping it `String` here and folding it into the same post-26.3 JSON follow-on.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
andrescrz: Dug into this properly. Your point is fair — seeding an empty index now is ~free and would avoid a heavy `MATERIALIZE` on the full table later. But the analysis nets to deferring, for reasons beyond just the LTS timing:

- **No query would use it today.** The only `input`/`output` search is `ilike(input, '%term%')` (case-insensitive substring; field-path filters use `JSON_VALUE(input, path)`). None of the text-index families accelerate that as-is: `tokenbf_v1` and the `text`/inverted index are token-oriented (not arbitrary substrings), and `ngrambf_v1`'s tokens are case-sensitive so `ilike` isn't cleanly accelerated. Getting any benefit needs a query rewrite (to `hasToken`/`multiSearchAny` or normalized-lowercase) — a semantics change and app work. There's also no full-text index anywhere in the schema today, so nothing is waiting to consume one.
- **The "seed for free" benefit collapses on 25.8.** The modern `text`/inverted index is still maturing and its on-disk format has shifted across 25.x→26.x, so an index seeded on 25.8 may need a drop+rebuild after the 26.3 upgrade — erasing the avoided-MATERIALIZE saving. `tokenbf_v1` survives the upgrade but doesn't fit the `ilike` query.
- **It would burden the cutover.** The index would build inline during the 4 TB `INSERT … SELECT` backfill (tokenizing the two largest columns) — no correctness risk, but real cost loaded onto the most sensitive step.

So the right sequence is: define the search feature → pick semantics + the matching index family on the settled 26.3 baseline → rewrite the query → seed it in that feature's own migration (a scoped, partition-able `MATERIALIZE` if the table has data by then). Keeping the create migration lean; recording text search as a deferred follow-on.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000101_create_traces_local_v2_table.sql
andrescrz: Good call-out — and it's the risk we pre-empted. That broad change already landed in #7307 (OPIK-6892), gated behind a `traceColumnsNonNullable` config toggle (default `false`, flipped in lockstep with the cutover `EXCHANGE`):

- **`SentinelTranslation`** at the DAO boundary — `nullToEpoch`/`nullToNaN` on write, `epochToNull`/`nanToNull` on read.
- The SQL predicates were written **dual-mode-safe** rather than hard-swapped, so they're correct whether the column is `Nullable` (today) or sentinel (post-cutover): `end_time IS NOT NULL AND notEquals(end_time, epoch)` (the `IS NOT NULL` is vacuous on a sentinel column, `notEquals(epoch)` does the work), `nullIf(end_time, epoch)` for sort/`argMax`, and `if(isNaN(duration), NULL, duration)` / `ttft`. Spread across `TraceDAO`, `ThreadDAO`, `KpiCardDAO`, `ProjectMetricsDAO`, `FilterQueryBuilder`.

I re-swept all DAOs for *bare* nullable-only predicates (`end_time IS NULL`, lone `isNotNull(...)`, `coalesce`/`ifNull`) on the three columns — zero hits on the traces columns, so nothing flips wrong when the toggle goes on. Covered by `SentinelTranslationTest`, `FilterQueryBuilderSentinelTest`, `TraceSentinelIntegrationTest`, `NaNAwareAggregateIntegrationTest`.

What's left is intentionally out of this create migration: flipping the toggle + the end-to-end column-by-column verification against the populated sentinel table (the cutover / staging dry-run story), and `spans` denullification under its own `spanColumnsNonNullable` slice (`SpanDAO` stays `Nullable` here, so its predicates are correct as-is). Thanks for flagging it.

apps/opik-backend/src/test/java/com/comet/opik/domain/TracesLocalV2TableTest.java
```diff
@@ -0,0 +1,513 @@
...
+        }
+        return keys;
+    }
+
+    @Builder(toBuilder = true)
+    private record StoredTrace(
+            UUID id,
+            String workspaceId,
+            UUID projectId,
+            String name,
```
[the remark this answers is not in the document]
andrescrz: Intentionally leaving this as-is. `StoredTrace` is a test-only fixture, always built from `manufacturePojo(StoredTrace.class).toBuilder()` (Podam populates every field) plus explicit overrides — so a required field is never accidentally unset; the only way to get a null is a deliberate `.field(null)`, which fails loudly in `insert(...)` on the next line. `@NonNull` is the right call for production internal records (the SKILL.md case, where there's no Bean Validation to lean on), but on a test fixture the ~24 annotations are verbosity guarding a null that can't arrive unnoticed — the test itself is the fail-fast.

thiagohora (approved):

PR #7347, opened 2026-07-05
[BE] fix: apply above_128k and above_272k tier pricing in cost calculation

Anuj7411:
## Details

Follow-up to #7023 (which introduced the whole-prompt `*_above_200k_tokens` tier for Gemini 2.5 Pro / Claude Sonnet 4.5), covering two additional thresholds LiteLLM publishes in `model_prices_and_context_window.json` that Opik currently ignores:

- **`*_above_128k_tokens`** — reachable model today: `gemini/gemini-1.5-flash` (input `7.5e-8` → `1.5e-7` above 128k, 2x).
- **`*_above_272k_tokens`** — reachable models today: the OpenAI `gpt-5.4` / `gpt-5.5` families plus every `-pro` variant (`2.5e-6` → `5e-6` input, `1.5e-5` → `2.25e-5` output, 2x/1.5x). With #7262 registering `azure` as a canonical provider, the Azure-hosted twins (`azure/gpt-5.4`, `azure/gpt-5.5*`) inherit the same rates — 17 flagship models in total whose above-threshold traces have been silently underbilled.

The scope for the tier fields is intentionally limited to `input_cost_per_token` / `output_cost_per_token` at each of the two new thresholds — LiteLLM does not publish `cache_creation_input_token_cost_above_128k_tokens` or `cache_read_input_token_cost_above_272k_tokens`, so cache-side tier handling stays 200k-only. Cache-tier for other thresholds can be picked up if and when models start publishing those fields.

This was pre-approved as a follow-up direction when #7023 landed. Andres, if the scope has drifted from what you expected, happy to split further.

### What changed

- `ModelCostData`: four new `@JsonProperty` fields for `input/output_cost_per_token_above_{128k,272k}_tokens` (Jackson's `SnakeCaseStrategy` doesn't insert underscores between letters and digits, so each tier field is pinned to the literal JSON key — same shape as the 200k fields already there).
- `ModelPrice`: four new tier fields plus two new threshold constants (`TIER_THRESHOLD_128K`, `TIER_THRESHOLD_272K`). `effectiveInputPrice` / `effectiveOutputPrice` walk the thresholds descending (272k → 200k → 128k → base) so if a model ever publishes multiple applicable tiers the highest wins.
- `CostService.buildModelPrice`: symmetric parsing + builder wiring for the four new fields.
- `SpanCostCalculator`: two additional call sites now consult the effective-rate helpers rather than the base rates:
[...]

PR #7438, opened 2026-07-11
[BE] refactor: consolidate above_NNNk tier fields into PromptTier record

Anuj7411:
## Details

Follow-up cleanup to #7347. The `above_128k`, `above_200k`, and `above_272k` tier rates lived as eight parallel fields on `ModelPrice` (input+output for each of three thresholds, plus `cache_creation`/`cache_read` at 200k). Every new threshold or rate meant updating `ModelPrice` + `defaultBuilder` + `CostService.buildModelPrice` + the four effective-price helpers.

`baz-reviewer` called out the resulting duplication on #7347 (*Code Dedup and Conventions*, low severity), and this PR is the version of that fix I signaled would come once #7347 landed. The current mainline `effectiveTieredPrice(...)` private helper is a partial move in the same direction; this PR completes it by pushing the tier collection all the way into a first-class data type.

### What changed

- **New nested record `ModelPrice.PromptTier`** carrying `(threshold, inputPrice, outputPrice, cacheCreationInputTokenPrice, cacheReadInputTokenPrice)`. Any rate a tier does not override stays at `BigDecimal.ZERO` — the effective helpers already treat zero as "fall through" so no special-case handling is needed.
- **`ModelPrice` shrinks from 17 to 10 fields.** The eight tier fields become one `List<PromptTier> promptTiers` field, stored **descending by threshold** so highest-applicable-tier semantics fall out of a single stream walk.
- **Four `effective*` helpers reduce to one `applicableTier(...)` private walker** parameterized by the rate accessor (`PromptTier::inputPrice`, `PromptTier::outputPrice`, etc.). The public API of `ModelPrice` is unchanged — every caller still sees `effectiveInputPrice(totalPromptTokens)` etc. with the same return type and semantics.
- **`CostService.buildModelPrice` extracted `addTierIfPresent(...)`** which appends a tier only when at least one of its per-rate JSON values is non-null and positive. Nulls (rate not published for this tier — e.g. LiteLLM does not carry `cache_creation`/`cache_read` at 128K/272K) are treated as ZERO and skipped by the effective helpers.

### What it enables

Adding a future threshold (say `above_512k` when a model publishes it) is now a two-line change: one new `TIER_THRESHOLD_*` constant plus one call to `addTierIfPresent(...)` in `buildModelPrice`. Zero effective-rate method churn, zero test-fixture updates. Same for a future new rate dimension in `PromptTier` — record it once, wire once, done.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code (CLI)
- Model(s): Claude Opus 4.7
- Scope: assisted in drafting the `PromptTier` record shape (chose a nested record over a top-level file to keep it colocated with the effective helpers that own it), the `addTierIfPresent`/`parseTierRate` extraction in `CostService`, and the descending-threshold ordering invariant
- Human verification: ran the full `CostServiceTest` + `SpanCostCalculatorTest` suites locally; verified that no test needed updating because they exercise `CostService.calculateCost` end-to-end rather than poking at `ModelPrice` tier fields directly; hand-traced the three reachable tier fixtures (gemini/gemini-1.5-flash at 128K, gemini-2.5-pro at 200K, gpt-5.4 at 272K) to confirm the walker returns the same rate the previous field-per-threshold structure did

## Testing

Local verification on JDK 25 + Maven 3.9.9:

```
mvn -ntp -f apps/opik-backend/pom.xml \
  -Dtest=CostServiceTest,SpanCostCalculatorTest test
```

Result: **100/100 green**. No new tests were added — the invariants covered by the existing `calculateCostAppliesAbove{128k,200k,272k}TierPricing*` cases from #7023 / #7347 exercise every branch of the refactored walker, including the descending-preference semantics (both a 200K-only model and a 272K-only model must return the correct rate at `prompt_tokens = 300_000`).

The pre-existing `calculateCost_overrideMicrosoftProviderMapsToAzure` test also continues to pass, confirming the override-file code path still constructs `ModelPrice` correctly through the refactored `buildModelPrice`.

## Documentation

No docs change. Behavior preserved 1:1; the refactor is invisible to end users and to every caller of `ModelPrice`.


Anuj7411:
@baz-reviewer good catch — verified against \`.agents/skills/opik-backend/SKILL.md\` which explicitly requires \`@Builder(toBuilder = true)\` on records and construction via the builder rather than the constructor.

Pushed \`99465dda9\`: annotated \`PromptTier\` with \`@Builder(toBuilder = true)\` and switched the single call site (\`CostService.addTierIfPresent(...)\`) to \`ModelPrice.PromptTier.builder()...build()\`. Behavior preserved 1:1; all 100 tests still green.

andrescrz (approved):
Hi @Anuj7411 

PR #7439, opened 2026-07-11
[OPIK-6741] [BE][FE] feat: add EvalTriggerScope to evaluation rules

miguelgrc:
## Details

Add a `trigger_scope` field to online evaluation rules with three options: **production** (default), **experiment**, and **both**. This controls whether a rule fires on production traces, experiment traces, or both — enabling customers to define one set of LLM-judge metrics as the single source of truth across experiments and production monitoring.

- BE: new `EvalTriggerScope` enum, DB migration (`000090`), field threaded through API/model/DAO/service/mapper layers, `OnlineScoringSampler` updated to let experiment traces through and filter per-evaluator by scope. Null-safe defaulting to PRODUCTION for backward compatibility.
- FE: `ToggleGroup` control on the v2 rule create/edit dialog (production / experiment / both), wired into form schema and API payload.
- Tests: dedicated `TriggerScopeTests` in sampler, updated `TracesUpdated` path coverage, updated integration tests.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- OPIK-6741

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Opus 4.6
- Scope: full implementation (BE + FE + tests)
- Human verification: code review, manual E2E testing with real LLM scoring pipeline

## Testing

- `mvn clean test -Dtest="OnlineScoringSamplerTest"` — 42/42 pass (3 new TriggerScope tests + updated existing)
- `mvn clean test -Dtest="AutomationRuleEvaluatorsResourceTest"` — 113/113 assertion-pass (0 failures; 2 pre-existing infra timeouts)
- `npx tsc --noEmit` — clean TypeScript check
- E2E script (`test_trigger_scope_e2e.py`): created 3 rules (production/experiment/both scopes), logged SDK traces + ran real experiment, verified via evaluator logs that each rule scored only its matching traces
- Manual UI testing: verified ToggleGroup appears on create/edit/clone, persists correct scope on save, defaults to Production

## Documentation

`@Schema` annotation added to `triggerScope` field on the create DTO — flows into OpenAPI spec and Fern-generated docs. No separate docs page changes needed.


apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/OnlineScoringSamplerTest.java
```diff
@@ -198,12 +199,59 @@ void skipsEvaluatorSourceMonitoringTraces() {
...
+            var sdkTrace = createTrace(Source.SDK);
+            var expTrace = createTrace(Source.EXPERIMENT);
+            var evaluator = createLlmEvaluator(true, 1.0f, List.of(), EvalTriggerScope.PRODUCTION);
+            whenFindAllLlmEvaluators(evaluator);
+
+            onlineScoringSampler
+                    .onTracesCreated(new TracesCreated(List.of(sdkTrace, expTrace), workspaceId, userName));
+
+            verify(onlineScorePublisher).enqueueMessage(List.of(toLlmMessage(evaluator, sdkTrace)),
+                    AutomationRuleEvaluatorType.LLM_AS_JUDGE);
```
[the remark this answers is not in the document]
miguelgrc: Skipping — the 3 tests were intentionally kept as separate methods rather than parameterized because the assertion shape differs per scope (BOTH expects a 2-element list, others expect 1-element). A single parameterized test would need conditional assertions, reducing clarity.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSampler.java
```diff
@@ -209,15 +210,14 @@ private void sampleAndScore(List<Trace> traces, String workspaceId, String userN
...
             // they carry selected_rule_ids metadata (explicit user selection from the playground).
             var scorableTraces = new ArrayList<Trace>();
             var selectedRuleIdsByTrace = new HashMap<UUID, Set<UUID>>();
             for (var trace : projectTraces) {
-                if (Source.isLoggingSource(trace.source())) {
-                    // For SDK traces all evaluators apply
+                if (Source.isLoggingSource(trace.source()) || trace.source() == Source.EXPERIMENT) {
                     scorableTraces.add(trace);
                 } else {
                     var ruleIds = extractSelectedRuleIds(trace);
```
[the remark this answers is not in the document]
miguelgrc: Skipping — this is intentional. `selected_rule_ids` is the playground/optimization mechanism where users manually pick which rules to run. Experiment traces use `triggerScope` instead — all rules with matching scope fire automatically on experiment traces. That's the core feature: define once, auto-run everywhere.

andriidudar (dismissed):
The FE part looks good, approved it.

thiagohora (dismissed):
Approving. Cleanly threads `EvalTriggerScope` (production/experiment/both) through the full BE stack, migration `000090`, and the v2 FE dialog; backward-compatible via `PRODUCTION` default. Satisfies OPIK-6741 (automatic eval-rule execution on experiment traces as a single source of truth). See inline comments for non-blocking follow-ups.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSampler.java
```diff
@@ -209,15 +210,14 @@ private void sampleAndScore(List<Trace> traces, String workspaceId, String userN
...
+            // SDK and experiment traces are scorable by evaluators whose trigger scope matches.
+            // We deliberately do not read selected_rule_ids from SDK or experiment traces.
+            // Other non-SDK traces (playground, optimization) are only scored when
             // they carry selected_rule_ids metadata (explicit user selection from the playground).
             var scorableTraces = new ArrayList<Trace>();
             var selectedRuleIdsByTrace = new HashMap<UUID, Set<UUID>>();
             for (var trace : projectTraces) {
-                if (Source.isLoggingSource(trace.source())) {
-                    // For SDK traces all evaluators apply
+                if (Source.isLoggingSource(trace.source()) || trace.source() == Source.EXPERIMENT) {
```
thiagohora, line 220: Requirement-critical: the feature works only if experiment-run traces actually carry `Source.EXPERIMENT`. Unit tests build such traces by hand, so the real-experiment E2E run is what proves the acceptance criterion. (Note: experiment traces now take this scorable branch, so their `selected_rule_ids` are ignored — intended per the ticket.)

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSampler.java
```diff
@@ -324,6 +325,19 @@ private List<Trace> stampMissingProjectNames(List<Trace> traces, String workspac
                 .toList();
     }
 
+    private boolean matchesTriggerScope(AutomationRuleEvaluator<?, ?> evaluator, Trace trace) {
```
thiagohora, line 328: `matchesTriggerScope` is only applied in this trace-level path, but `trigger_scope` is added to all evaluator types. Confirm thread/span rules either honor the scope or don't surface the toggle where it's a no-op.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleDAO.java
```diff
@@ -34,16 +34,17 @@ HAVING COUNT(project_id) > 1
             )""")
     boolean hasVersion1AutomationRules(@Bind("workspaceId") String workspaceId);
 
-    @SqlUpdate("INSERT INTO automation_rules(id, workspace_id, `action`, name, sampling_rate, enabled, filters) "
+    @SqlUpdate("INSERT INTO automation_rules(id, workspace_id, `action`, name, sampling_rate, enabled, trigger_scope, filters) "
```
thiagohora, line 40: INSERT binds the raw enum (`PRODUCTION`) via `@BindMethods` (JDBI default BY_NAME), while `updateBaseRule` binds `getValue()` (`production`). Works only because MySQL ENUM assignment is case-insensitive. Consider `EvalTriggerScope implements HasValue` + an `AbstractEnumColumnMapper` factory (codebase pattern), or bind `getValue()` consistently. `action` has the same pre-existing latent issue.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleEvaluatorRowMapper.java
```diff
@@ -40,6 +41,7 @@ public AutomationRuleEvaluatorModel<?> map(ResultSet rs, StatementContext ctx) t
         String name = rs.getString("name");
         Float samplingRate = rs.getFloat("sampling_rate");
         boolean enabled = rs.getBoolean("enabled");
+        EvalTriggerScope triggerScope = EvalTriggerScope.fromString(rs.getString("trigger_scope"));
```
thiagohora: `fromString` throws on null/unknown, failing the whole row map. Safe given `NOT NULL DEFAULT`, but the shared `AbstractEnumColumnMapper.parse()` logs + returns null, degrading more gracefully.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleEvaluatorService.java
```diff
@@ -197,6 +198,10 @@ public <E, F extends Filter, T extends AutomationRuleEvaluator<E, F>> T save(@No
                 }
             };
 
+            if (evaluator.triggerScope() == null) {
```
thiagohora, line 201: Nit: default is applied in 4 places (DTO `@Builder.Default`, here, `update()`, and `matchesTriggerScope`). Aligned with the backward-compat requirement so acceptable; only the sampler guard is strictly needed for robustness.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSampler.java
miguelgrc: Good observation — yes, adding `Source.EXPERIMENT` to the `scorableTraces` set is required for the trigger scope to take effect. Previously experiment traces were only scorable via `selectedRuleIds` (the playground flow); now they can also be auto-sampled by rules with `EXPERIMENT` or `BOTH` scope. The per-evaluator `matchesTriggerScope()` filter then gates which rules actually fire.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSampler.java
miguelgrc: Good catch. Thread and span evaluators have separate scoring paths (`OnlineScoringSpanSampler`, `TraceThreadOnlineScoringSamplerListener`) that don't check `triggerScope` — the toggle was a no-op for those rule types. Fixed by hiding the toggle in the FE when scope is thread or span (d8b0156861).

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleDAO.java
miguelgrc: Fixed in 8da394b478 — `EvalTriggerScope` now implements `HasValue`, with a new `EvalTriggerScopeColumnMapper` (following the `DashboardScopeMapper` pattern). Registered via `@RegisterArgumentFactory` on the DAO, and the UPDATE `@Bind` param type changed from `String` to `EvalTriggerScope`. Both INSERT (`@BindMethods`) and UPDATE (`@Bind`) paths now go through the argument factory for consistent lowercase binding.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleEvaluatorRowMapper.java
miguelgrc: Fixed in 8da394b478 — added null guard before `EvalTriggerScope.fromString()` in the RowMapper to handle rows where `trigger_scope` is NULL (legacy data or edge cases).

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleEvaluatorService.java
miguelgrc: Acknowledged — the `@Builder.Default` to `PRODUCTION` on the API model ensures new rules default to production-only scope. The DB migration also has `DEFAULT 'production'` for existing rows. Keeping both layers aligned per the existing `enabled` field pattern.

thiagohora (approved):
Revision 2 addresses the review feedback:
- Thread/span rules: trigger-scope toggle now hidden (FE guard).
- INSERT/UPDATE enum binding: unified via `EvalTriggerScope implements HasValue` + `EvalTriggerScopeColumnMapper` registered as argument factory — consistent write path.
- RowMapper: null-guarded.

All flagged items resolved or reasonably handled; no new issues. LGTM 🚀

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSampler.java
JetoPistola, line 220: 💡 **suggestion** | Performance

Adding `Source.EXPERIMENT` here means experiment traces now always populate `scorableTraces`, so a project receiving only experiment traffic no longer hits the `scorableTraces.isEmpty()` early-out below — it always runs `ruleEvaluatorService.findAll(...)` (Redis-cached, so no extra DB load) plus a `parallelStream` over every evaluator, evaluating `matchesTriggerScope`/`isEvaluatorSelectedForTrace`/`shouldSampleTrace` per (evaluator × trace), even when no rule in the project has `experiment`/`both` scope.

Not a correctness issue and probably fine, but for experiment-heavy workspaces with no experiment-scoped rules it's wasted CPU. If that ever matters, a cheap guard (skip when all scorable traces are experiment-source and no evaluator has `EXPERIMENT`/`BOTH` scope) would restore the old fast path.

JetoPistola:
❓ **question** | Documentation

The "Documentation update" checklist item is unchecked, but this adds a user-facing capability (Trigger scope: production / experiment / both on evaluation rules). The `@Schema` annotation covers the API/OpenAPI surface, but would a user-facing docs page (evaluation rules / online scoring) help customers discover the experiment-scope option? Just confirming it's intentionally out of scope.

JetoPistola:
👋 **Review summary**

**What looks good**
- Clean, disciplined layering — the new `EvalTriggerScope` is threaded through every layer (API DTOs, model, DAO, mapper, RowMapper, service) with a dedicated `HasValue`/`AbstractEnumColumnMapper` pattern that matches existing enum conventions.
- Backward compatibility is airtight and defended at every layer: `NOT NULL DEFAULT 'production'` migration, service-side null→`PRODUCTION` defaulting on both create and update, and a null-safe fallback in `matchesTriggerScope`. Existing rules keep production-only behavior.
- Good test coverage — dedicated `TriggerScopeTests`, the `triggerScope` field left *in* the round-trip recursive comparison (so persistence is validated implicitly), and the `TracesUpdated` / selected-rule-ids paths updated consistently.
- The `isLoggingSource` vs `EXPERIMENT` branches are cleanly non-overlapping, and the feature is a genuinely useful "single source of truth for LLM-judge metrics across experiments and production."

**Overall**
Solid, well-scoped implementation that's already been through a thorough review cycle with all substantive items resolved. CI is green. The approach makes sense.

**Inline comments**: 1 performance suggestion + 1 documentation question — nothing blocking.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSampler.java
miguelgrc: Good point. The evaluator fetch is Redis-cached (`ruleEvaluatorService.findAll`) so there's no extra DB load, and the per-evaluator `matchesTriggerScope` check is a cheap enum comparison. For experiment-heavy workspaces with no experiment-scoped rules, the overhead is the parallelStream iteration over evaluators that all get filtered out — negligible in practice.

If profiling ever shows this matters, the guard you suggest (skip when all scorable traces are experiment-source and no evaluator has EXPERIMENT/BOTH scope) would be straightforward to add. Leaving as-is for now to keep the logic simple.

PR #7450, opened 2026-07-13
[OPIK-7180] [SDK] fix: capture proxy cost and allow provider override in LangChain integration

alexkuzmik:
## Details

Fixes `$0.00` span cost when LangChain's `ChatOpenAI` routes through an OpenAI-compatible proxy (e.g. a LiteLLM gateway). Two root causes, addressed per the ticket comment (convenient provider override + capture the returned cost):

- **Per-run provider override.** `OpikTracer` gains a `provider` parameter. It accepts a `str`/`opik.LLMProvider` for the common single-provider case, or a `Callable[[run_dict], provider | None]` resolved per LLM run so a chain/graph that mixes providers is labeled correctly (returning `None` falls back to auto-detection). Previously the provider was auto-detected as the proxy hostname, which the backend can't price.
- **Extensible provider-reported cost passthrough.** New `response_cost_extractors` registry (protocol + registry) rather than a hardcoded path. The shipped `LiteLLMResponseCostExtractor` reads `x-litellm-response-cost` and records it as the span `total_cost`, taking priority over Opik's own estimate. New proxies/integrations are added by registering another extractor.

The real `response_metadata` shape was verified against a locally-run open-source LiteLLM proxy: the cost header is nested under `response_metadata["headers"]` and only present when `ChatOpenAI(include_response_headers=True)` is set (the ticket's assumed flat path was incorrect; the extractor supports both to be safe).

## Change checklist
- [x] User facing
- [x] Documentation update

## Issues
- OPIK-7180

## Testing

New `tests/library_integration/langchain/test_langchain_proxy_cost.py` (15 cases, fully offline):
- End-to-end through the real `OpikTracer` path via a fake chat model carrying the verified proxy `response_metadata`, asserting `total_cost` and provider land on the span.
- Provider override via string, `LLMProvider` enum (normalized to its string value), and a per-run callable.
- Parametrized `try_extract_response_cost` for the nested-`headers` shape, the flat fallback, non-numeric/absent values, and malformed run dicts.

Additionally verified against a real local LiteLLM proxy (mock model, no API keys): cost `3.5e-05` captured and provider override applied on a persisted trace.

## Documentation

- Added a "Using a proxy (LiteLLM, etc.)" section to the LangChain integration docs (`fern/docs` and `fern/docs-v2`), covering the provider override (string + callable) and the required `include_response_headers=True`.


apps/opik-documentation/documentation/fern/docs-v2/integrations/langchain.mdx
```diff
@@ -264,6 +264,51 @@ for chunk in llm.stream("Hello", config={"callbacks": [opik_tracer]}):
...
+
+If a single chain or graph mixes providers, pass a callable that receives the LangChain run dict and returns the provider for that specific run. Return `None` to fall back to the auto-detected provider:
+
+```python
+def resolve_provider(run):
+    metadata = run["outputs"]["generations"][-1][-1]["message"]["kwargs"]["response_metadata"]
+    model = metadata.get("model_name", "")
+    if "anthropic" in model:
+        return "bedrock"
+    return None  # fall back to auto-detection
```
petrotiurin, line 286: Any way we can abstract this from the users? I think looking at L284 they might quickly get scared and just not implement this at all.
Would metadata always be of this shape or is it proxy-dependent?

sdks/python/src/opik/integrations/langchain/opik_tracer.py
```diff
@@ -662,6 +691,21 @@ def _process_end_span(self, run: Run) -> None:
                 )
                 self._opik_context_storage.pop_span_data(ensure_id=span_data.id)
 
+    def _resolve_provider(self, run_dict: Dict[str, Any]) -> Optional[str]:
+        if self._provider is None:
+            return None
+
+        resolved = (
+            self._provider(run_dict) if callable(self._provider) else self._provider
+        )
```
petrotiurin: Should we wrap this in try/catch so if user's function breaks we still log the traces? We should probably warn but not break here.

alexkuzmik: Done in 2a4296e — the resolver call is now wrapped in try/except: a throwing user callback logs a warning and falls back to the auto-detected provider, so trace logging is never broken. Added a regression test (`test_langchain__provider_resolver_raises__falls_back_to_autodetection`).

apps/opik-documentation/documentation/fern/docs-v2/integrations/langchain.mdx
alexkuzmik: Good call — abstracted in 2a4296e. The resolver now receives a `ProviderResolverContext` instead of the raw run dict, so the common case is a one-liner:

```python
def resolve_provider(context):
    return "bedrock" if "anthropic" in (context.model or "") else None
```

On the shape question: `context.model` is the model name Opik parses from the run, so it is *not* proxy-dependent — routing works the same regardless of the proxy. The raw run dict is still available as `context.run` for advanced cases. Docs updated to match. (The `x-litellm-response-cost` cost header itself is LiteLLM-specific, but that path is fully automatic and users never touch it.)

petrotiurin (approved):

PR #7452, opened 2026-07-13
[OPIK-7319] Optimize getTraceKpiCards: drop trace_costs LEFT JOIN

thiagohora:
## Details

`getTraceKpiCards` is the heaviest of the three KPI-card ClickHouse queries. It combined `traces_filtered` (count / error-rate / avg-duration) with `trace_costs` (the `spans FINAL` cost aggregation) via a per-trace `LEFT JOIN trace_costs tc ON tf.id = tc.trace_id`; on large projects `traces_filtered` is millions of rows and that join's hash build dominated peak memory.

Since the current/previous period split is keyed on the trace's UUIDv7 `id` and `spans.trace_id` is that same id, the cost can be pre-bucketed into the two periods inside `trace_costs` and `CROSS JOIN`ed as a single row, eliminating the join. Results are identical and the `trace_id IN (traces_filtered)` semijoin is retained for correctness. The `spans FINAL` scan (the query's dominant IO) is unchanged — this is a memory/latency trim.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves OPIK-7319

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: query rewrite in `KpiCardDAO.GET_TRACE_KPI_CARDS`, live validation and benchmarking against a production ClickHouse replica, commit and PR authoring.
- Human verification: rewrite reviewed; results and benchmark independently reproduced against production before merge.

## Testing

The exact rendered query (no-filter form) was run against a production ClickHouse replica on the highest-cost project over a ~14-day window and compared to the original:

- Output identical: counts, error-rates and total cost match to the cent (`avg_duration` differs only at the 15th decimal — floating-point summation-order noise, non-deterministic between runs of either query).
- 5 runs each: median latency 2971 ms -> 2552 ms (~14% faster); median peak memory 1.54 GiB -> 1.23 GiB (~20% lower and more stable); rows/bytes/marks unchanged.

CI runs `KpiCardsResourceTest` which exercises the templated path (including filter branches). Local resource tests were not run due to a known local-bootstrap harness limitation; relying on CI for the integration suite.

## Documentation

None — internal query optimization, no behavior or API change.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/KpiCardDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/KpiCardDAO.java
@@ -55,6 +55,12 @@ class KpiCardDAOImpl implements KpiCardDAO {
     private final @NonNull InstantToUUIDMapper instantToUUIDMapper;
     private final @NonNull OpikConfiguration configuration;
 
+    /**
+     * trace_costs buckets the TOTAL_COST KPI into current/previous periods keyed on trace_id (a
+     * UUIDv7 matching the trace id used for the count/error/duration split) and is CROSS JOINed as a
+     * single row, avoiding the per-trace LEFT JOIN whose hash build dominated peak memory on large
+     * projects. Identical results; the trace_id IN (traces_filtered) semijoin is retained. OPIK-7319.
+     */
     private static final String GET_TRACE_KPI_CARDS = """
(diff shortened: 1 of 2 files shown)


andrescrz (approved):
LGTM

PR #7454, opened 2026-07-13
[OPIK-7321] fix: union getProjectIds result stream instead of single()

thiagohora:
## Details

`ExperimentAggregatesDAO.getProjectIds` terminated its reactive pipeline with `single()`, which throws `IndexOutOfBoundsException` ("Source emitted more than one item") — surfacing as an unmapped HTTP 500 — whenever the R2DBC result stream emits more than one element. The method is documented to "always emit exactly one Set (possibly empty)", so this folds the emitted sets with `reduceWith` instead: it always emits exactly one Set, is idempotent for the normal single-row result, and no longer depends on the stream emitting exactly one element.

- `reduceWith` uses a fresh accumulator per subscription (avoids the shared-mutable-seed pitfall of `reduce(seed, ...)`).

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves OPIK-7321

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: Root-cause analysis and the one-method fix in `ExperimentAggregatesDAO.getProjectIds` (+ javadoc).
- Human verification: Diff reviewed; `mvn compile` run locally; author to confirm CI (resource tests do not bootstrap locally).

## Testing

- `cd apps/opik-backend && mvn -q compile` — succeeds.
- Full `mvn test` deferred to CI (the resource test harness does not bootstrap cleanly in this local environment).
- Behavior: the pipeline now unions all emitted sets into a single Set, so a multi-element stream no longer throws; the empty-stream case still emits an empty Set.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/experiments/aggregations/ExperimentAggregatesDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/experiments/aggregations/ExperimentAggregatesDAO.java
@@ -66,6 +66,7 @@ import java.math.BigDecimal;
 import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -1603,6 +1604,11 @@ class ExperimentAggregatesDAOImpl implements ExperimentAggregatesDAO {
      * Distinct project_ids this experiment's items reference; always emits exactly one Set
      * (possibly empty when no traces with project_id are found). See
      * {@link #GET_PROJECT_IDS}.
+     * <p>
+     * The result stream is folded with {@code reduceWith} rather than terminated with
+     * {@code single()}: the R2DBC result publisher can emit more than one element, which makes
+     * {@code single()}/{@code singleOrEmpty()} throw {@code IndexOutOfBoundsException}. Unioning
+     * the emitted sets always yields exactly one Set and is idempotent for the single-row result.
      */
     @Override
     public Mono<Set<UUID>> getProjectIds(UUID experimentId) {
@@ -1619,7 +1625,10 @@ class ExperimentAggregatesDAOImpl implements ExperimentAggregatesDAO {
                             .stream(row.get("project_ids", String[].class))
                             .map(UUID::fromString)
                             .collect(Collectors.toUnmodifiableSet())));
-        }).single());
+        }).reduceWith(() -> new HashSet<UUID>(), (allProjectIds, projectIds) -> {
+            allProjectIds.addAll(projectIds);
+            return allProjectIds;
+        }).map(Set::copyOf));
     }
 
     private Mono<TraceAggregations> getTraceAggregations(UUID experimentId, Set<UUID> projectIds, UUID labelProjectId) {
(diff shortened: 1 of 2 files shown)


apps/opik-backend/src/main/java/com/comet/opik/domain/experiments/aggregations/ExperimentAggregatesDAO.java
```diff
@@ -1619,7 +1621,21 @@ public Mono<Set<UUID>> getProjectIds(UUID experimentId) {
...
+    }
+
+    /**
+     * Folds the project_id chunks emitted by the result stream into a single immutable Set. Using
+     * {@code reduceWith} (rather than {@code single()}/{@code singleOrEmpty()}) is deliberate: the
+     * R2DBC result publisher can emit more than one element, which would make {@code single()}
+     * throw {@code IndexOutOfBoundsException}. Unioning always yields exactly one Set (empty when
+     * the stream is empty) and is idempotent for the normal single-row result.
+     */
+    static Mono<Set<UUID>> unionProjectIdChunks(Flux<Set<UUID>> projectIdChunks) {
```
andrescrz, line 1636: Nit: annotate with visible for testing from Guava.

apps/opik-backend/src/test/java/com/comet/opik/domain/experiments/aggregations/ExperimentAggregatesDAOImplTest.java
```diff
@@ -0,0 +1,46 @@
...
+import org.junit.jupiter.api.Test;
+import reactor.core.publisher.Flux;
+import reactor.test.StepVerifier;
+
+import java.util.Set;
+import java.util.UUID;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class ExperimentAggregatesDAOImplTest {
```
andrescrz, line 13: Minor: We generally cover with black box or integrations tests. That would be a better approach for this case if testable from the public APIs/interfaces.

andrescrz (dismissed):
LGTM, just a couple of optional comments regarding tests.

apps/opik-backend/src/main/java/com/comet/opik/domain/experiments/aggregations/ExperimentAggregatesDAO.java
thiagohora: Done — annotated `unionProjectIdChunks` with Guava's `@VisibleForTesting` in 0f353ee. It's package-private only so the test can reach it, so the annotation makes that intent explicit.

apps/opik-backend/src/test/java/com/comet/opik/domain/experiments/aggregations/ExperimentAggregatesDAOImplTest.java
thiagohora: Good point, and the happy path is already covered black-box: `ExperimentsResourceTest.GroupExperimentsAggregations` drives aggregate population end-to-end through `ExperimentAggregatesService` → `populateExperimentItemsInBatches` → `getProjectIds` and asserts the resulting `traceCount` / `totalEstimatedCost` / `feedbackScores`, so the real query and its result mapping are exercised via the public API.

The one thing an integration test cannot reach is the specific regression this PR fixes: the R2DBC result publisher emitting **more than one element** (which is what made `single()` throw `IndexOutOfBoundsException`). That is a driver-level artifact — a real experiment always returns a single row, so no black-box scenario forces the multi-emit path deterministically. The unit test on `unionProjectIdChunks` targets exactly that fold at the only level where it can be triggered reliably. Happy to drop it if you'd prefer to rely solely on the existing integration coverage, but it's the only thing that actually guards against a regression back to `single()`.

andrescrz (approved):

PR #7455, opened 2026-07-13
[OPIK-7322] fix: parse OpenAI error model before OpenRouter in getErrorObject

thiagohora:
## Details

`LlmProviderLangChainMapper.getErrorObject` attempted to parse a provider error payload as `OpenRouterErrorMessage` first, whose `code` field is typed `Integer`. Every caller is an OpenAI-compatible provider (`FreeModelLlmProvider`, `LlmProviderOpenAi`, `LlmProviderOpenAiResponses`), whose errors carry a String `code` (e.g. `"rate_limit_exceeded"`), so the first attempt always threw `InvalidFormatException` (wrapped `UncheckedIOException`), logging a full WARN stack before the OpenAI fallback recovered. This flips the order to parse the OpenAI model first, keeping OpenRouter as a defensive fallback — the common path now parses on the first try with no thrown/caught exception and no WARN noise.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves OPIK-7322

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: Reorder in `LlmProviderLangChainMapper.getErrorObject` (+ javadoc) and a new `LlmProviderLangChainMapperTest`.
- Human verification: Diff reviewed; the new unit test run locally (see Testing); author to confirm CI.

## Testing

- `cd apps/opik-backend && mvn -o test -Dtest=LlmProviderLangChainMapperTest` — `Tests run: 5, Failures: 0, Errors: 0`.
- New test asserts OpenAI-format errors with a String `code` (`rate_limit_exceeded` → 429, `insufficient_quota` → 402, `invalid_api_key` → 401) resolve to the right status, and that a payload-less throwable yields an empty result. With the fix these parse on the first attempt (no "failed to parse" WARN); the only WARNs emitted are from the deliberate no-payload cases.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/LlmProviderLangChainMapper.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/LlmProviderLangChainMapper.java
@@ -241,14 +241,22 @@ public interface LlmProviderLangChainMapper {
         return t.getMessage() != null && t.getMessage().contains("{");
     }
 
+    /**
+     * Every caller is an OpenAI-compatible provider, whose error payload carries a String
+     * {@code code} (e.g. {@code "rate_limit_exceeded"}). {@link OpenRouterErrorMessage} types
+     * {@code code} as {@code Integer}, so attempting it first made Jackson throw
+     * {@code InvalidFormatException} (wrapped as {@code UncheckedIOException}) on every such error,
+     * logging a full WARN stack before the OpenAI fallback recovered. Attempt the OpenAI model
+     * first so the common path parses on the first try; keep OpenRouter as a defensive fallback.
+     */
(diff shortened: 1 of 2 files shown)


thiagohora:
Closing: the proposed reorder (parse OpenAI before OpenRouter) regresses OpenRouter error handling — an OpenRouter error `{"error":{"code":429,...}}` gets its Integer code coerced into the OpenAI model's String `code`, which is then unmapped and degrades HTTP 429 → 500 (verified locally). OpenRouter routes through LlmProviderOpenAi.getLlmProviderError → getErrorObject, so this is a live path. The existing OpenRouter-first order already yields correct statuses for both provider shapes; the IOException seen in prod is a recoverable log-noise symptom of the rate-limit storm (OPIK_6971), not a correctness bug in this code. Leaving the error-parse code unchanged.

PR #7457, opened 2026-07-13
[OPIK-7322] fix: pick LLM error model by code type to avoid doomed OpenRouter parse

thiagohora:
## Details

`LlmProviderLangChainMapper.getErrorObject` attempted the OpenRouter error model — whose `code` field is `Integer` — first for every provider error. Every caller is OpenAI-compatible (`FreeModelLlmProvider`, `LlmProviderOpenAi`, `LlmProviderOpenAiResponses`; note OpenRouter itself routes through `LlmProviderOpenAi`), and OpenAI-style errors carry a String `code` (e.g. `"rate_limit_exceeded"`), so the OpenRouter attempt threw `InvalidFormatException` (wrapped `UncheckedIOException`) that was caught and logged as a full WARN stack on every such error. This makes the model selection content-aware: attempt the OpenRouter (numeric) model only when `error.code` is numeric, otherwise parse with the OpenAI model directly — removing the doomed parse and its WARN stack.

- The order must not simply be flipped: an OpenRouter payload (`"code": 429`) is silently coerced by Jackson into the OpenAI model's String `code` and then degrades to HTTP 500. Selecting by code type keeps OpenRouter's numeric code mapping to the correct status.
- `parseError` (and its logging) is unchanged; a genuine parse failure still warns.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves OPIK-7322

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: `getErrorObject` model selection + `extractErrorJson`/`hasNumericErrorCode` helpers in `LlmProviderLangChainMapper`, and a new `LlmProviderLangChainMapperTest`.
- Human verification: Diff reviewed; the regression (429->500 under a naive reorder) reproduced and confirmed avoided; tests run locally (below); author to confirm CI.

## Testing

- `cd apps/opik-backend && mvn -o test -Dtest=LlmProviderLangChainMapperTest` — `Tests run: 5, Failures: 0, Errors: 0`.
- Asserts: OpenAI string codes (`rate_limit_exceeded`->429, `insufficient_quota`->402, `invalid_api_key`->401) resolve to `Optional[ErrorMessage]` with the right status (present, so the caller returns 4xx rather than the empty->500 path) **and** log no parse-failure WARN; OpenRouter numeric code `429` maps straight to 429 (no regression); a payload-less throwable yields empty.

### DIFF
--- a/apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/LlmProviderLangChainMapperTest.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/LlmProviderLangChainMapperTest.java
@@ -1,24 +1,22 @@
 package com.comet.opik.infrastructure.llm;
 
-import ch.qos.logback.classic.Level;
-import ch.qos.logback.classic.Logger;
-import ch.qos.logback.classic.spi.ILoggingEvent;
-import ch.qos.logback.core.read.ListAppender;
 import io.dropwizard.jersey.errors.ErrorMessage;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
+import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.List;
 import java.util.Optional;
 import java.util.stream.Stream;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
 class LlmProviderLangChainMapperTest {
+
+    private static final Logger log = LoggerFactory.getLogger(LlmProviderLangChainMapperTest.class);
 
     private final LlmProviderLangChainMapper mapper = LlmProviderLangChainMapper.INSTANCE;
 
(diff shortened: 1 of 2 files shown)


apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/LlmProviderLangChainMapperTest.java
```diff
@@ -0,0 +1,91 @@
...
+        var logger = testLogger("openai-" + code);
+        var appender = attach(logger);
+
+        Optional<ErrorMessage> result = mapper.getErrorObject(providerErrorWith(body), logger);
+
+        assertThat(result).isPresent();
+        assertThat(result.get().getCode()).isEqualTo(expectedStatus);
+        // The OpenRouter (Integer code) model is not attempted for a string code, so the doomed parse
+        // and its WARN stack no longer happen.
+        assertThat(warnings(appender)).isEmpty();
```
andrescrz: Minor: not a big fan of testing logs (they change often, non-critical part of the application logic, not typically tested etc.) specially when part of a 3rd party library. We shouldn't test those, specially they're inner details.

Unless this give us important coverage, I wouldn't go with this and instead stick to the ErrorMessage payload assertion (code etc.).

andrescrz (dismissed):
LGTM, just a minor comment about tests.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/LlmProviderLangChainMapperTest.java
thiagohora: Good point!

thiagohora: Done in a4e4c16 — dropped the logback `ListAppender` and the WARN-emptiness assertions, kept the assertions on the resolved `ErrorMessage.getCode()`. Coverage is unchanged where it matters: the OpenRouter numeric-`429` case still asserts it resolves to `429` (guarding the reorder trap where a numeric code gets coerced into OpenAI's String field and degrades to a 500), plus the OpenAI string codes → 429/402/401 and the no-JSON → empty (→ caller's 500) cases.

andrescrz (approved):

PR #7458, opened 2026-07-13
[OPIK-6904] Add ClickHouse partition-health observability metrics

thiagohora:
## Details

Adds ClickHouse partition-health observability so the new gauges are live and validated before table partitioning lands. A config-gated Quartz job (disabled by default) polls `system.parts` and the lightweight-delete mask and publishes `opik.clickhouse.partition.*` OpenTelemetry gauges. It follows the existing retention observability jobs: a distributed lock keeps a single instance polling per interval and clears its snapshot on losing the lock, so exactly one series is reported per `(table, partition)` and aged-out partitions drop cleanly from Prometheus.

- Metrics: partition count per table, per-partition size on disk, largest active part, parts, rows, last activity, and LWD-masked rows.
- Works on the current unpartitioned tables (one `partition_id = 'all'` series per table) and fans out to real per-partition series once partitioning lands — no code change.
- The masked-row query requires the backend's read-write ClickHouse user (`apply_deleted_mask = 0` is a per-query setting a `readonly = 1` user cannot set); it is `priority`- and `max_execution_time`-bounded so it can never contend with query load.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-6904

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: DAO, Quartz job, config wiring, and integration test authored with Claude Code; SQL scan cost profiled against a live environment.
- Human verification: Reviewed by the author; profiled the masked-row scan against a live environment (largest table ~1.3s, single low-priority poll per interval) and confirmed the read-write-user requirement for `apply_deleted_mask`.

## Testing

- `mvn -o compile` and `mvn -o test-compile` — pass.
- `ClickHousePartitionMetricsDAOTest` (new integration test): seeds traces through the API, asserts the `system.parts` aggregation, then lightweight-deletes a subset and asserts the DAO's masked-row count matches ClickHouse ground truth. Runs in CI (local resource-test bootstrap is environment-flaky).
- Profiled the masked-row scan (full `_row_exists` column read) against a live environment to size the poll cost and confirm the `readonly=1` limitation.
- Job is disabled by default (`PARTITION_METRICS_ENABLED=false`); no behavior change until explicitly enabled.

## Documentation

Config documented inline in `config.yml`; metric semantics documented in the DAO and job Javadoc.

### DIFF
--- a/apps/opik-backend/config.yml
+++ b/apps/opik-backend/config.yml
@@ -1037,7 +1037,7 @@ partitionMetrics:
   # Poll interval in seconds (default 5 min). Min 30, max 3600.
   intervalSeconds: ${PARTITION_METRICS_INTERVAL_SECONDS:-300}
   # Tables scanned for lightweight-deleted (masked) rows. Comma-separated env override.
-  lwdTables: ${PARTITION_METRICS_LWD_TABLES:-[traces, spans]}
+  lwdTables: ${PARTITION_METRICS_LWD_TABLES:-traces,spans}
 
 # LLM providers client configuration
 llmProviderClient:
--- a/apps/opik-backend/config.yml
+++ b/apps/opik-backend/config.yml
@@ -1030,6 +1030,15 @@ retention:
     # Earliest date for catch-up cursor start (service launch date)
     serviceStartDate: ${RETENTION_CATCH_UP_SERVICE_START_DATE:-2024-09-01}
 
+# ClickHouse partition-health observability (OPIK-6904). Polls system.parts + the LWD mask and
+# publishes opik.clickhouse.partition.* gauges. A distributed lock keeps a single instance polling.
+partitionMetrics:
+  enabled: ${PARTITION_METRICS_ENABLED:-false}
+  # Poll interval in seconds (default 5 min). Min 30, max 3600.
+  intervalSeconds: ${PARTITION_METRICS_INTERVAL_SECONDS:-300}
+  # Tables scanned for lightweight-deleted (masked) rows. Comma-separated env override.
+  lwdTables: ${PARTITION_METRICS_LWD_TABLES:-[traces, spans]}
+
 # LLM providers client configuration
 llmProviderClient:
   # Default: 3
--- a/apps/opik-backend/config.yml
+++ b/apps/opik-backend/config.yml
@@ -1034,8 +1034,8 @@ retention:
 # publishes opik.clickhouse.partition.* gauges. A distributed lock keeps a single instance polling.
 partitionMetrics:
   enabled: ${PARTITION_METRICS_ENABLED:-false}
-  # Poll interval in seconds (default 5 min). Min 30, max 3600.
-  intervalSeconds: ${PARTITION_METRICS_INTERVAL_SECONDS:-300}
+  # Poll interval (default 5 min). Min 30s, max 1h.
+  interval: ${PARTITION_METRICS_INTERVAL:-5m}
   # Tables scanned for lightweight-deleted (masked) rows. Comma-separated env override.
   lwdTables: ${PARTITION_METRICS_LWD_TABLES:-traces,spans}
 
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ClickHousePartitionMetricsJob.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ClickHousePartitionMetricsJob.java
@@ -57,16 +57,15 @@ public class ClickHousePartitionMetricsJob extends Job implements InterruptableJ
     private static final AttributeKey<String> PARTITION_KEY = stringKey("partition");
 
     private record Snapshot(List<PartitionStat> partitionStats, List<LwdStat> lwdStats) {
+        private static final Snapshot EMPTY = new Snapshot(List.of(), List.of());
     }
-
-    private static final Snapshot EMPTY = new Snapshot(List.of(), List.of());
 
     private final ClickHousePartitionMetricsDAO partitionMetricsDAO;
     private final LockService lockService;
     private final PartitionMetricsConfig config;
 
     private final AtomicBoolean interrupted = new AtomicBoolean(false);
-    private final AtomicReference<Snapshot> snapshot = new AtomicReference<>(EMPTY);
+    private final AtomicReference<Snapshot> snapshot = new AtomicReference<>(Snapshot.EMPTY);
 
     private final LongCounter runCounter;
 
@@ -124,11 +123,13 @@ public class ClickHousePartitionMetricsJob extends Job implements InterruptableJ
             return;
         }
 
-        Mono<Void> refresh = Mono
+        // Deferred so the DAO calls (and their query rendering) run only once the lock is held —
+        // bestEffortLock subscribes to this Mono only after acquiring the permit.
+        Mono<Void> refresh = Mono.defer(() -> Mono
                 .zip(partitionMetricsDAO.getPartitionStats(),
                         partitionMetricsDAO.getLwdRowCounts(config.getLwdTables()))
                 .doOnNext(this::updateSnapshot)
-                .then();
+                .then());
 
         try {
             lockService.bestEffortLock(
@@ -137,7 +138,7 @@ public class ClickHousePartitionMetricsJob extends Job implements InterruptableJ
                     Mono.fromRunnable(() -> {
                         log.debug(
                                 "ClickHouse partition metrics: another instance holds the poll lock, clearing snapshot");
-                        snapshot.set(EMPTY);
+                        snapshot.set(Snapshot.EMPTY);
                         runCounter.add(1, Attributes.of(stringKey("result"), "skipped_lock"));
                     }),
                     config.getInterval(),
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ClickHousePartitionMetricsJob.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ClickHousePartitionMetricsJob.java
@@ -1,10 +1,10 @@
 package com.comet.opik.api.resources.v1.jobs;
 
-import com.comet.opik.domain.ClickHousePartitionMetricsDAO;
-import com.comet.opik.domain.ClickHousePartitionMetricsDAO.LwdStat;
-import com.comet.opik.domain.ClickHousePartitionMetricsDAO.PartitionStat;
 import com.comet.opik.infrastructure.PartitionMetricsConfig;
 import com.comet.opik.infrastructure.lock.LockService;
+import com.comet.opik.infrastructure.metrics.ClickHousePartitionMetricsDAO;
+import com.comet.opik.infrastructure.metrics.ClickHousePartitionMetricsDAO.LwdStat;
+import com.comet.opik.infrastructure.metrics.ClickHousePartitionMetricsDAO.PartitionStat;
 import io.dropwizard.jobs.Job;
 import io.opentelemetry.api.GlobalOpenTelemetry;
 import io.opentelemetry.api.common.AttributeKey;
@@ -30,6 +30,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.ToLongFunction;
 
 import static com.comet.opik.infrastructure.lock.LockService.Lock;
 import static io.opentelemetry.api.common.AttributeKey.stringKey;
@@ -80,6 +81,7 @@ public class ClickHousePartitionMetricsJob extends Job implements InterruptableJ
 
         Meter meter = GlobalOpenTelemetry.get().getMeter("opik.clickhouse");
 
+        // Aggregated one series per table (partition count is the only non-per-partition metric).
         meter.gaugeBuilder("opik.clickhouse.partition.count").ofLongs()
                 .setDescription("Number of active partitions per table")
                 .buildWithCallback(measurement -> {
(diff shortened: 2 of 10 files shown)


apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ClickHousePartitionMetricsJob.java
```diff
@@ -0,0 +1,187 @@
...
+        if (interrupted.get()) {
+            log.info("ClickHouse partition metrics job interrupted before execution, skipping");
+            return;
+        }
+
+        Mono<Void> refresh = Mono
+                .zip(partitionMetricsDAO.getPartitionStats(),
+                        partitionMetricsDAO.getLwdRowCounts(config.getLwdTables()))
+                .doOnNext(this::updateSnapshot)
+                .then();
```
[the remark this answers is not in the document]
thiagohora: Fixed in f7c30c8. The `getLwdRowCounts()` arm is now wrapped in `.onErrorResume(...)` which logs a warning and degrades to an empty list, so `Mono.zip` still completes with fresh `system.parts` data and `updateSnapshot` runs. The reliable partition-health gauges refresh even if the LWD mask scan fails; the LWD gauges simply stop reporting until the next successful poll (same "empty = stop reporting" semantics as lock loss).

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/PartitionMetricsConfig.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Comma-separated tables to scan for lightweight-deleted (LWD-masked) rows via
+     * {@code SELECT count() WHERE _row_exists = 0 SETTINGS apply_deleted_mask = 0}. Restricted to
+     * the high-volume append/delete tables to avoid scanning small config tables every cycle.
+     *
+     * <p>Stored as a scalar rather than a YAML list so it binds cleanly from a comma-separated
+     * environment override (Dropwizard substitutes {@code ${...}} into the raw YAML before parsing,
+     * so a comma-separated env value cannot bind to a {@code List}). {@link #getLwdTables()} splits,
+     * trims and drops blanks; individual names are validated as plain identifiers at the point of
```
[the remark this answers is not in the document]
thiagohora: This looks like a false positive from line-by-line reading. The word "via" is not dangling — the sentence continues on the next line: "...tables to scan for lightweight-deleted (LWD-masked) rows via `SELECT count() WHERE _row_exists = 0 SETTINGS apply_deleted_mask = 0`." It renders as one complete sentence in the Javadoc.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ClickHousePartitionMetricsJob.java
```diff
@@ -0,0 +1,198 @@
...
+
+        this.runCounter = meter
+                .counterBuilder("opik.clickhouse.partition_metrics.run")
+                .setDescription("Number of partition-health metric poll runs, tagged by result")
+                .build();
+    }
+
+    @Override
+    public void doJob(JobExecutionContext context) {
+        if (interrupted.get()) {
```
andrescrz, line 121: I believe you're missing the enabled check here. Not a blocker as handled at service start up (guicey lifecycle event).

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ClickHousePartitionMetricsJob.java
```diff
@@ -0,0 +1,198 @@
...
+@Slf4j
+@DisallowConcurrentExecution
+public class ClickHousePartitionMetricsJob extends Job implements InterruptableJob {
+
+    private static final Lock RUN_LOCK = new Lock("clickhouse:partition_metrics_lock");
+
+    private static final AttributeKey<String> TABLE_KEY = stringKey("table");
+    private static final AttributeKey<String> PARTITION_KEY = stringKey("partition");
+
+    private record Snapshot(List<PartitionStat> partitionStats, List<LwdStat> lwdStats) {
```
andrescrz, line 62: Nit: define builder (with to builder) and favour its usage over constructors.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ClickHousePartitionMetricsJob.java
```diff
@@ -0,0 +1,198 @@
...
+
+        try {
+            lockService.bestEffortLock(
+                    RUN_LOCK,
+                    refresh,
+                    Mono.fromRunnable(() -> {
+                        log.debug(
+                                "ClickHouse partition metrics: another instance holds the poll lock, clearing snapshot");
+                        snapshot.set(Snapshot.EMPTY);
+                        runCounter.add(1, Attributes.of(stringKey("result"), "skipped_lock"));
```
andrescrz: Minor: superflous counters. The job is not suppose to run here if we had a proper quartz configuration (backed by MySQL) instead of a best effort one with Redis. In addition, it can skew the metrics with 0 measurements. I'd rather remove this.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ClickHousePartitionMetricsJob.java
```diff
@@ -0,0 +1,198 @@
...
+                    }),
+                    config.getInterval(),
+                    Duration.ZERO,
+                    true) // holdUntilExpiry: exactly one instance polls per interval
+                    .block();
+        } catch (Exception e) {
+            log.error("ClickHouse partition metrics poll failed", e);
+            runCounter.add(1, Attributes.of(stringKey("result"), "error"));
+        }
```
andrescrz: This should likely be handled on the reactive chain instead, with onErrorResume.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ClickHousePartitionMetricsJob.java
```diff
@@ -0,0 +1,198 @@
...
+            ToLongFunction<LwdStat> extractor) {
+        meter.gaugeBuilder(name).ofLongs()
+                .setDescription(description)
+                .buildWithCallback(measurement -> snapshot.get().lwdStats()
+                        .forEach(stat -> measurement.record(extractor.applyAsLong(stat),
+                                attributes(stat.table(), stat.partition()))));
+    }
+
+    @Override
+    public void interrupt() {
```
andrescrz, line 200: This interruption logic is missing interrupting the reactive subscription as well.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/metrics/ClickHousePartitionMetricsDAO.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    Mono<List<PartitionStat>> getPartitionStats();
+
+    Mono<List<LwdStat>> getLwdRowCounts(List<String> tables);
+}
+
+@Slf4j
+@Singleton
+class ClickHousePartitionMetricsDAOImpl implements ClickHousePartitionMetricsDAO {
```
andrescrz, line 46: You'll need to validate that our regular user has permissions on all this system tables for a typical deployment (prod-like with Kubernetes and standard user provisioning). I'm not completely sure about this.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/metrics/ClickHousePartitionMetricsDAO.java
```diff
@@ -0,0 +1,157 @@
...
+     * an injection vector.
+     */
+    private static final String LWD_ROWS_SQL = """
+            SELECT
+                _partition_id AS partition_id,
+                toInt64(count()) AS lwd_rows
+            FROM <table>
+            WHERE _row_exists = 0
+            GROUP BY _partition_id
+            SETTINGS apply_deleted_mask = 0, max_execution_time = 30, priority = 100, log_comment = '<log_comment>'
```
andrescrz, line 95: Same about these settings, not sure if the user can change them. You'll need to test and validate before moving deploying and before enabling.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/PartitionMetricsConfig.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.time.Duration;
+import java.util.Arrays;
+import java.util.List;
+
+@Data
+public class PartitionMetricsConfig {
+
+    @JsonProperty
+    private boolean enabled = false;
```
andrescrz: Better remove all the defaults from the Java code and leave only the ones from the config yaml files. So we keep a single source of truth for them. They are all either primitives or validated as not-null, so they will always exists.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/PartitionMetricsConfig.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * How often the partition-health poll runs (seconds). The poll reads {@code system.parts}
+     * metadata plus a lightweight {@code _row_exists} mask scan per LWD table; on a large cluster
+     * a full cycle costs a few seconds and a few GiB of highly compressible column reads, so 5 min
+     * is a safe default. A distributed lock ensures a single instance polls per interval.
+     */
+    @JsonProperty
+    @Min(30) @Max(3600) private int intervalSeconds = 300;
```
andrescrz: Better as Duration type. Range seems good (5 minutes default, 30s min and 1h max)., so keep validations.(

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/PartitionMetricsConfig.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /** Derived: interval between polls. */
+    public Duration getInterval() {
+        return Duration.ofSeconds(intervalSeconds);
+    }
+
+    /** Derived: the parsed, trimmed, blank-free list of LWD tables. */
+    public List<String> getLwdTables() {
+        return Arrays.stream(lwdTables.split(","))
+                .map(String::trim)
```
andrescrz: Nit: better strip instead of trim.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/metrics/ClickHousePartitionMetricsDAOTest.java
```diff
@@ -0,0 +1,191 @@
...
+        });
+
+        // Lightweight-delete a subset and assert the DAO's masked-row count matches ClickHouse
+        // ground truth (cross-checking the DAO SQL rather than an exact physical-row count keeps
+        // the assertion robust to ReplacingMergeTree duplicates).
+        int toDelete = 5;
+        var idList = traceIds.stream().limit(toDelete)
+                .map(id -> "'%s'".formatted(id))
+                .toList();
+        execute("DELETE FROM traces WHERE id IN (%s)".formatted(String.join(",", idList)));
```
andrescrz: The trace deletion endpoint already uses lightweight delete. Why not using that instead of a direct query? That way, we always assert the intended production behaviour.

apps/opik-backend/config.yml
```diff
@@ -1051,6 +1030,15 @@ retention:
     # Earliest date for catch-up cursor start (service launch date)
     serviceStartDate: ${RETENTION_CATCH_UP_SERVICE_START_DATE:-2024-09-01}
 
+# ClickHouse partition-health observability (OPIK-6904). Polls system.parts + the LWD mask and
+# publishes opik.clickhouse.partition.* gauges. A distributed lock keeps a single instance polling.
+partitionMetrics:
```
andrescrz, line 1035: Please propagate to the config test yaml file as well, following the usual conventions (same comments except if something differs, no env vars).

andrescrz (commented):
Let's iterate on the feedback before moving forward.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ClickHousePartitionMetricsJob.java
thiagohora: The job is not registered if disabled

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ClickHousePartitionMetricsJob.java
thiagohora: Done in c18640cacf — `Snapshot` now carries `@Builder(toBuilder = true)` and both the `EMPTY` sentinel and `updateSnapshot` build via the builder.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ClickHousePartitionMetricsJob.java
thiagohora: Done in c18640cacf — removed the `run` `LongCounter` and all its `.add(...)` calls. Poll health is now visible from the gauges themselves (a failed/skipped poll clears the snapshot, so the series go stale/empty).

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ClickHousePartitionMetricsJob.java
thiagohora: Done in c18640cacf — replaced the try/catch around `.block()` with reactive `onErrorResume` and a fire-and-forget `subscribe()` on `boundedElastic`, mirroring `AlertProjectMigrationJob`.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/ClickHousePartitionMetricsJob.java
thiagohora: Done in c18640cacf — the subscription `Disposable` is held in an `AtomicReference` and `interrupt()` now disposes the in-flight poll (and `doFinally` clears it). Also added an `interrupted` guard inside the deferred refresh.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/PartitionMetricsConfig.java
thiagohora: Done in c18640cacf — dropped the Java field defaults; `config.yml` (with env fallbacks) is now the single source of truth, and the block is mirrored into `config-test.yml`.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/PartitionMetricsConfig.java
thiagohora: Done in c18640cacf — `interval` is now an `io.dropwizard.util.Duration` with `@MinDuration(30s)`/`@MaxDuration(1h)`, matching `AlertProjectMigrationConfig`. Config value is `5m`; the scheduler/lock consume it via `.toJavaDuration()`.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/PartitionMetricsConfig.java
thiagohora: Done in c18640cacf — `getLwdTables()` now uses `String::strip` instead of `trim`.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/metrics/ClickHousePartitionMetricsDAOTest.java
thiagohora: Done in c18640cacf — the test now deletes traces through `traceResourceClient.deleteTrace(...)` (the prod endpoint, which issues a lightweight delete) instead of a direct `DELETE` query. The removed `execute` helper was dropped.

apps/opik-backend/config.yml
thiagohora: Done in c18640cacf — the `partitionMetrics` block is now in `config-test.yml` too, with literal values and no env-var substitution.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/metrics/ClickHousePartitionMetricsDAO.java
thiagohora: Good call — this is a pre-cutover validation item, not a code change. The job runs as the backend's read-write ClickHouse user (the same one used for inserts), so it should already have SELECT on `system.parts`, but I'll confirm the grant explicitly against the prod CH user before flipping `enabled=true`. The job stays `enabled: false` by default until that check passes.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/metrics/ClickHousePartitionMetricsDAO.java
thiagohora: Same validation track as the `system.parts` grant: `apply_deleted_mask=0`, `max_execution_time`, `priority` and `log_comment` are per-query settings that a `readonly=1` user can't override (Code 164). The javadoc on `LWD_ROWS_SQL` documents this requirement. I'll verify the prod read-write user can set them before enabling; until then the job is off by default.

andrescrz (approved):
LGTM, thanks for addressing all the feedback.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/PartitionMetricsConfig.java
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * How often the partition-health poll runs. The poll reads {@code system.parts} metadata plus a
+     * lightweight {@code _row_exists} mask scan per LWD table; on a large cluster a full cycle costs
+     * a few seconds and a few GiB of highly compressible column reads, so 5 min is a safe default.
+     * A distributed lock ensures a single instance polls per interval.
+     */
+    @NotNull @JsonProperty
+    @MinDuration(value = 30, unit = TimeUnit.SECONDS)
+    @MaxDuration(value = 1, unit = TimeUnit.HOURS)
+    private Duration interval;
```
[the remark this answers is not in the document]
thiagohora: The mechanism is real — with the Java defaults gone and `partitionMetrics` being `@Valid @NotNull`, the block must supply `interval` and `lwdTables` or startup validation fails, even when `enabled: false`. But this isn't a regression in practice, and re-adding Java defaults would undo the "config.yml is the single source of truth" change requested earlier in this review:

- Opik ships one canonical `config.yml` baked into the image, which always includes the block with env-var fallbacks (`interval: ${PARTITION_METRICS_INTERVAL:-5m}`, `lwdTables: ${PARTITION_METRICS_LWD_TABLES:-traces,spans}`). Operators override individual values via env vars; they don't ship a YAML that drops whole blocks, so `interval`/`lwdTables` are never null at startup.
- `config-test.yml` now carries the block too, so tests validate cleanly.
- This matches the established sibling convention: `alertProjectMigration` is likewise `@Valid @NotNull` with no Java-side defaults (it's a record), relying on `config.yml` to provide every field. Same trade-off, intentionally consistent.

So no change needed — the "safe defaults" now live in `config.yml` (the single source of truth) rather than in Java.

PR #7471, opened 2026-07-14
[OPIK-7306] [BE] feat: add id_at column migration to spans; keep retention on trace_id

thiagohora:
## Details

Ships migration `000103`, adding the `id_at` materialized column (`DateTime('UTC') MATERIALIZED UUIDv7ToDateTime(toUUID(id))`) to `spans` — a metadata-only mirror of `000091` on `traces`. This is the prerequisite for the span read-paths ticket and the spans partitioning cutover (spans counterpart of OPIK_6900).

The three spans retention paths (`deleteForRetention`, `deleteForRetentionBounded`, `countForRetention`) intentionally keep filtering by `trace_id` **only** — no `toMonday(id_at)` partition-pruning predicate is added:

- The spans retention range is keyed on `trace_id`, but `id_at` is materialized from the span's own **client-assigned** UUIDv7 `id`, which can land in a later week than its `trace_id`.
- A `toMonday(id_at)` bound derived from the trace-id range would therefore wrongly exclude valid candidates — spans orphaned on delete, undercounted on count.
- Partition pruning for span retention has to wait until `spans` can be pruned by a column aligned with `trace_id`. The `SpanDAO` javadoc documents this.

**Pre-existing assumption (out of scope here):** the `trace_id` range relies on `trace_id` being UUIDv7 (lexicographic order = chronological). This underlies the whole retention subsystem — traces retention likewise ranges on `id` — and is unchanged by this PR (it only removes the `id_at` predicate). Enforcing UUIDv7 on `traceId` at ingest is a cross-cutting ingest-validation change affecting all clients; tracking separately rather than in this migration PR.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves OPIK-7306

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: Migration 000103, SpanDAO retention javadoc, and the black-box integration tests.
- Human verification: Author reviewed the diff and confirmed the full `RetentionPolicyServiceTest` suite passes against real ClickHouse.

## Testing

`mvn -o test -Dtest='RetentionPolicyServiceTest'` against real ClickHouse (testcontainers) — 20 tests, 0 failures.

New `SpanRetentionTraceIdOnly` nested class regression-tests all three paths: each span is given an `id` weeks *after* its `trace_id`, and must still be deleted/counted purely on `trace_id`. These would fail if an `id_at` upper bound were reintroduced.

## Documentation

N/A — internal retention-path change, no user-facing surface.


apps/opik-backend/src/main/java/com/comet/opik/domain/SpanDAO.java
```diff
@@ -1226,11 +1226,25 @@ ORDER BY (workspace_id, project_id, trace_id, parent_span_id, id) DESC, last_upd
...
+     * week stay in scope.
+     */
     private static final String DELETE_FOR_RETENTION = """
             DELETE FROM spans
             WHERE workspace_id IN :workspace_ids
             AND trace_id >= :lower_bound
             AND trace_id \\< :cutoff_id
+            AND toMonday(id_at) >= toMonday(UUIDv7ToDateTime(toUUID(:lower_bound), 'UTC'))
+            AND toMonday(id_at) \\< addWeeks(toMonday(UUIDv7ToDateTime(toUUID(:cutoff_id), 'UTC')), 1)
             AND trace_id NOT IN (
```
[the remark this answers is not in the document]
thiagohora: Thanks — the `id_at`-vs-`trace_id` interaction is intentional, and the `addWeeks(…, 1)` upper bound is what handles it.

**Why the derivation stays on the span's own `id`:** `id_at` is the future partition expression — `spans_local_v2` will be `PARTITION BY toMonday(id_at)` with `id_at` MATERIALIZED from the span's own UUID (the spans counterpart of OPIK-6900, which requires this migration to define `id_at` identically). Deriving `id_at` from `trace_id` would misalign the retention/read predicates from the actual partition column and defeat the pruning this ticket exists to enable, so the column has to stay on `toUUID(id)`.

**Why the range still can't drop valid candidates:** the predicate keeps every span whose `toMonday(id_at)` is ≤ the cutoff's week (`< addWeeks(toMonday(cutoff), 1)` is the half-open way to make the cutoff's *own* week inclusive — same reason it's there on the trace path). A span's `id_at` is ≥ its trace's time, and a deletion candidate has `trace_id < cutoff_id` (trace in the cutoff's week or earlier), so its rows sit inside `[lower_bound_week, cutoff_week]` and are retained in scope. The only rows outside that band would be a span whose trace sits in the cutoff's exact week *and* whose own id crossed into the following week — the newest sliver of the window, and negligible at the retention horizon (cutoff ≥ 14 days old, by which point a trace and its spans share a week in practice). That boundary is exactly what the parameterized Tuesday–Sunday tests pin down.

On the duplicated clause: this follows the existing convention of inline, per-DAO SQL constants with the rationale in javadoc above each (matching `TraceDAO` and the rest of the DAO layer). I'd rather keep that consistency than introduce a cross-DAO templated fragment here; happy to extract one in a follow-up if the team prefers it repo-wide.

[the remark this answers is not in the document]
thiagohora: Update: we've taken your point and **dropped the `toMonday(id_at)` predicate** from all three retention paths (5849d2e). Since `id_at` is materialized from the span's own client-assigned id and can lag its `trace_id`'s week, an `id_at` bound derived from the trace-id range isn't a safe consequence — so retention now filters by `trace_id` only. Migration `000103` stays as the prerequisite for the read-paths / partitioning work; partition-aligned pruning for span retention will come once `spans` can be pruned by a `trace_id`-aligned column. Added regression tests that give each span an id weeks after its `trace_id` and assert it's still deleted/counted purely on `trace_id`.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanDAO.java
```diff
@@ -1226,6 +1226,16 @@ ORDER BY (workspace_id, project_id, trace_id, parent_span_id, id) DESC, last_upd
...
 
+    /**
+     * Retention sweep for the applyToPast=true window: spans whose {@code trace_id} is in
+     * {@code [lower_bound, cutoff_id)} and not linked to experiments.
+     * <p>
+     * Filters on {@code trace_id} only. Unlike {@code TraceDAO}, the spans retention range is keyed on
+     * {@code trace_id} while the future partition column {@code id_at} is MATERIALIZED from the span's own UUIDv7 id
+     * (migration 000103). A span's id can land in a later week than its {@code trace_id}, so a {@code toMonday(id_at)}
+     * bound derived from the trace-id range would wrongly exclude valid candidates. No partition-pruning predicate is
+     * applied here until {@code spans} can be pruned by a column aligned with {@code trace_id}.
```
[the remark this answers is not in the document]
thiagohora: Good catch, and correct — span ingest validates the span's own `id` as UUIDv7 (`validateId(id, SPAN_KEY)`) but not `traceId`, so a client could store a v4 `trace_id`.

That said, this is a **pre-existing** property of the retention subsystem, not something this PR introduces: the `trace_id >= :lower_bound AND trace_id < :cutoff_id` range predates this change (it was already on `main`; this PR only *removes* the `id_at` predicate). The same UUIDv7-time-ordering assumption underlies traces retention too (it ranges on `id`). Enforcing UUIDv7 on `traceId` at ingest is a cross-cutting ingest-validation change that affects all clients and applies well beyond spans, so it's out of scope for this migration PR (OPIK-7306) — I'll track it as a separate follow-up rather than fold it in here.

apps/opik-backend/src/test/java/com/comet/opik/domain/retention/RetentionPolicyServiceTest.java
```diff
@@ -356,6 +360,92 @@ void deletesRowSharingTheCutoffWeek(DayOfWeek cutoffDay) {
...
+
+            // trace_id in the retention window (61d old), but the span's own id is weeks later (5d old).
+            var spanId = idGenerator.generateId(now.minus(5, ChronoUnit.DAYS));
+            var traceId = idGenerator.generateId(now.minus(61, ChronoUnit.DAYS));
+            createTestSpan(spanId, traceId, API_KEY, wsName(wsId));
+            waitForRows("spans", wsId, 1);
+
+            var cutoffId = idGenerator.generateId(now.minus(60, ChronoUnit.DAYS));
+            var lowerBound = idGenerator.generateId(now.minus(65, ChronoUnit.DAYS));
+            spanDAO.deleteForRetention(List.of(wsId), cutoffId, lowerBound).block();
```
andrescrz: Question: why the deletion is directly triggered from the spanDAO? I believe this isn't correct. This is the data retention policy service tests, we should exercise testing from `executeRetentionCycle` and by creating the spans with an ID that they're eligible to be deleted, not by forcing a direct delete and bypassing the logic through the spans DAO.

andrescrz (commented):
Change LGTM, but I believe there's an issue with the testing approach.

apps/opik-backend/src/test/java/com/comet/opik/domain/retention/RetentionPolicyServiceTest.java
thiagohora: Good point — reworked in 1e660ba. The span retention test now drives a full cycle through `RetentionPolicyService.executeRetentionCycle(...)` (with a rule + spans created so they're eligible by `trace_id`) instead of calling `SpanDAO` directly. To still guard the specific reason we dropped the `id_at` predicate, the eligible span is given an `id` a much later ISO week than its `trace_id`, and the test asserts it's still deleted (a recent span is kept) — this would fail if a `toMonday(id_at)` bound were reintroduced. Dropped the now-unused `SpanDAO` injection.

I also tried a second variant driving the `applyToPast=true` catch-up path (`estimatePendingRules()` + `executeCatchUpCycle`), but since those are global (shared rule/cursor state + per-cycle budget across all workspaces) it interfered with the existing `CatchUpJob.smallWorkspace_oneShotCatchUp` test, so I left catch-up coverage to that existing suite — the regression property is path-independent anyway (all three retention SQL paths now filter on `trace_id` only).

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanDAO.java
andrescrz, line 1237: Let's discuss this offline again, before moving forward. Maybe a potentially easy option that we could explore for this change is adding a subquery that with the same trace_id filters like we have right now, but instead of directly deleting from spans, it would resolve first the correspondent span ID and then delete. That way, we can add the partitioning pruning predicate on `id_at` which would match the span ID. 

Let's quickly discuss this together, it might be the easiest solution here to ensure both deletions are correct and also that partitioning pruning happens.

andrescrz (commented):
I should have mentioned this in the previous iteration, but I thought it was coming on the next open PR.

This PR is basically adding the `id_at` field, which we need for this project and improving testing.

But it's not addressing the partition pruning for the data retention feature on spans ID, mostly due to the problem of deleting based on `trace_id`. This is the main ask of this ticket, and we're not fulfilling it. 

I made a proposal for a potential easy way to ensure both, that it'd be interesting to explore before moving forward.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanDAO.java
thiagohora: Before the sync I ran a local benchmark to check whether the resolve-then-prune idea actually buys pruning over the current `trace_id`-only delete. TL;DR: on this schema it doesn't — the `trace_id` delete already prunes optimally, and the `id_at` predicate is redundant + slower.

**Why:** `trace_id` is the primary-key prefix of `spans` (`ORDER BY (workspace_id, project_id, trace_id, parent_span_id, id)`), stored as `FixedString(36)` so UUIDv7 sorts chronologically. ClickHouse mutations (including lightweight `DELETE`) prune parts by the **primary key**, not only the partition key — parts whose PK range can't hold a matching `trace_id` are **hardlinked, not rewritten** (near-free). So a `trace_id`-range delete already confines the physical rewrite to the partitions that actually hold matching rows.

**Setup:** ClickHouse 25.8 (matches prod), `spans` modeled faithfully (`FixedString(36)` ids, `id_at DateTime MATERIALIZED UUIDv7ToDateTime(toUUID(id))`, `PARTITION BY toMonday(id_at)`, same `ORDER BY`). 10M spans across 54 weekly partitions (~5.8 parts/partition), ~1.5% of spans given an `id` in a later ISO week than their `trace_id` (the late-span case). Retention deletes an 8-week-old `trace_id` window → 1.62M matching rows living in 17 partitions. Measured physically-rewritten parts via `system.part_log` (`read_rows > 0`), lightweight delete, `lightweight_deletes_sync=2`.

**Results:**

| | A: `trace_id` only | B: `trace_id` + `toMonday(id_at) IN (subquery)` |
|---|---|---|
| rows deleted | 1,619,450 | 1,619,450 |
| parts physically rewritten | 113 | 113 |
| partitions rewritten | 17 | 17 |
| parts hardlinked (not rewritten) | 254 | 254 |
| delete duration | 8.7 s | 22.0 s (~2.5x slower) |

Control probe: a constant single-week predicate (`toMonday(id_at) = <one week>`) rewrote exactly **1** partition — so pruning is real; `trace_id` already achieves the same via the PK. The 17 rewritten partitions are the true minimum (8 old weeks + the newer weeks where late spans landed); any correct delete has to touch all 17, so B can't do better — it just adds a pre-scan.

**Conclusion:** with `trace_id` as the PK prefix, the current delete already prunes to the matching partitions; adding an `id_at` predicate is redundant and ~2.5x slower. The one thing that would flip this: if `spans_local_v2` changes the sort key so `trace_id` is no longer the prefix — then `trace_id`-range PK pruning degrades and the `id_at` angle matters again. Worth pinning down v2's `ORDER BY` in the sync. (Also: `id_at` partitioning still has value for `DROP PARTITION` / TTL / tiering lifecycle — that's separate from the retention delete predicate.)

Happy to walk through the harness live; I can also share the SQL so you can reproduce or try other layouts.

andrescrz (approved):

PR #7326, opened 2026-07-02
[OPIK-7149] [BE] refactor: remove AiSpend cost-analytics read endpoints superseded by ai-cost-backend

BorisTkachenko:
## Details

Cost-analytics serving was extracted into the standalone **ai-cost-backend** (cost-api) per OPIK-7065, which is now in use. The frontend routes `/api/v1/private/ai-spend/*` to that service (dev vite proxy and prod nginx), so the Opik backend's `/v1/private/ai-spend` endpoints are dead code.

This PR removes the AiSpend **read layer** — everything reachable only from `AiSpendResource`:

- `AiSpendResource` and its 5 spend endpoints (summary, composition, lane breakdown, all breakdowns, users)
- `AiSpendService`, `AiSpendDAO`, `AiSpendMapper`, `AiSpendQueryBuilder`
- `SpendUserSortingFactory` and the 11 `api/spend/*` DTOs
- The 4 associated test classes

**Preserved (verified out of scope):**
- The cipx ingestion pipeline (`CostIntelligenceIngestionListener`, `CipxSpendDAO`, `CipxTraceIdentityDAO`, cost-intelligence events, migrations 000093/000094) — this writes the `cipx_spends` tables that feed ai-cost-backend.
- Shared `WorkspaceMetricsService`/DAO + `WorkspaceMetric*` DTOs (feedback-score metrics) and the `WorkspacesResource` `/costs` endpoints.
- Ingestion-time span cost calculation.

The generated OpenAPI specs and Python/TS SDK `ai_spend` clients are left untouched here — they will be removed by fern generation in a separate PR.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues
- OPIK-7149

## Testing
No new tests — this is a deletion of dead code. Backend compile and the existing test suite pass with the AiSpend read layer removed; no other component references the deleted classes (verified across `apps/opik-backend/src`).

## Documentation
N/A — no documentation changes. Downstream OpenAPI/SDK regeneration tracked separately.

PR #7476, opened 2026-07-15
[OPIK-7307] [BE] feat: add id_week predicate to SpanDAO read paths

thiagohora:
## Details

Engages weekly-partition pruning on user-facing `spans`-table **read** scans by adding a `toMonday(id_at)` ("id_week") predicate alongside each existing span-`id` bound. The planner cannot infer monotonicity through `UUIDv7ToDateTime(toUUID(id))`, so the bound must be explicit (Decision 16.1 #15). It is a strict consequence of the id-range — never drops a row the id-range wouldn't — and, unlike a `created_at` predicate, safe against late-arriving rows since it derives from `id`. A no-op pre-cutover; once `spans` is `PARTITION BY toMonday(id_at)` it prunes to the partitions in range (most impactful on cold-tier reads, and disproportionately important for the 14 TB `spans` table carrying the cost/spend aggregations).

Spans counterpart of the traces work in **OPIK-6895** (#7235). **Stacked on #7471 (OPIK-7306)**, which ships the `id_at`-on-`spans` migration (`000103`) this predicate targets — review/merge that first.

### Qualifying spans-table scans (predicate mirrors each id bound)
- **`SpanDAO`** — `SELECT_BY_PROJECT_ID` (`span_id_prefilter`, `spans_deduped`, and the `page_wide` window re-read), `COUNT_BY_PROJECT_ID` (prefilter + final scan), `SELECT_SPANS_STATS`, `SELECT_SPAN_FEEDBACK_SCORES_STATS`
- **`SpanMetricsQueries.spans_filtered`** — shared by `ProjectMetricsDAO` + `WorkspaceMetricsDAO` span metrics
- **`ProjectMetricsDAO`** — `GET_COST`(`_WITH_BREAKDOWN`), `GET_TOKEN_USAGE`(`_WITH_BREAKDOWN`), `GET_THREAD_COST`
- **`KpiCardDAO`** — `trace_costs`, `spans_filtered`, `thread_costs`
- **`WorkspaceMetricsDAO`** — `GET_COSTS_SUMMARY`, `GET_COSTS_DAILY`(`_BY_PROJECT`)

Operator mapping mirrors each bound: `id >=` → `toMonday(id_at) >=`; `id <=` / `id <` → `toMonday(id_at) <=`; `id BETWEEN` → both bounds. Windowed `id IN (page_ids)` re-reads carry the window's week bounds. StringTemplate vs raw-SQL `<`-escaping handled per query (e.g. `GET_COSTS_DAILY*` bind raw SQL → plain `<=`; the `newST`-rendered queries → `\<=`).

### Excluded paths (documented — no safe id_week derivation)
- **By-id lookups**: `getById` / `getOnlySpanDataById` / `getPartialById` / `getByIds`, the insert/update conflict reads, `DELETE_BY_IDS`, `BULK_UPDATE`.
- **Trace / experiment-linkage scans** (key on `trace_id` / experiment, not span time/id-range): `getByTraceIds`, `getSpanIdsForTraces`, and span scans in `TraceDAO` / `ThreadDAO` / `DatasetItem*DAO` / `Experiment*DAO` / `OptimizationDAO` / `FeedbackScoreDAO` / `WorkspaceMetadataDAO`. `ProjectMetricsDAO.GET_TOTAL_COST` keys on a `trace_id` range only (a plain mirror would narrow the set — a `created_at`-like hazard), so it is excluded here.
- **`created_at`-windowed** daily BI counters (`SPAN_COUNT_BY_WORKSPACE_ID`, `SPAN_DAILY_*`).
- **No id/time bound to mirror**: `EXISTS_BY_PROJECT_ID`, `GET_PROJECT_TOKEN_USAGE_NAMES`.
- Retention paths (`deleteForRetention*` / `countForRetention`) are handled in #7471; `ESTIMATE_VELOCITY_FOR_RETENTION` stays with the retention work.
- **`AiSpendQueryBuilder`** (a named candidate in the ticket) was removed in #7326 — the AI-spend read endpoints moved to their own `cipx_spends` table — so it no longer scans `spans`.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues
- Resolves OPIK-7307

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: The five `spans`-scanning read DAOs listed above; SQL-only changes (no Java logic touched).
- Human verification: Author reviewed the diff; `mvn compile` / `test-compile` green.

## Testing

Semantic no-op — every modified query is already exercised with the predicate rendered by pre-existing integration tests (audited rather than duplicated, as in #7235):
- Span list / search / count time-filtering (boundary + exclusion): `FindSpansResourceTest`, `SpansResourceTest`.
- Span stats: `SpansResourceTest` (`getSpansStats`).
- Span metrics (duration / cost / token usage, incl. breakdown): `ProjectMetricsResourceTest`, `ProjectMetricsWithBreakdownResourceTest`.
- KPI cards (SPANS entity type, incl. thread cost): `KpiCardsResourceTest`.
- Workspace cost summary / daily / span token usage: `WorkspacesResourceTest`.

`mvn -o compile` and `mvn -o test-compile` succeed. `EXPLAIN PARTITIONS` validation is deferred to post-cutover staging per the AC (`spans` is not partitioned yet).

## Documentation

N/A — internal read-path pruning change, no user-facing surface.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/KpiCardDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/KpiCardDAO.java
@@ -200,6 +200,8 @@ class KpiCardDAOImpl implements KpiCardDAO {
                     FROM spans FINAL
                     WHERE workspace_id = :workspace_id AND project_id = :project_id
                       AND id >= :uuid_from_time AND id \\<= :uuid_to_time
+                      AND toMonday(id_at) >= toMonday(UUIDv7ToDateTime(toUUID(:uuid_from_time), 'UTC'))
+                      AND toMonday(id_at) \\<= toMonday(UUIDv7ToDateTime(toUUID(:uuid_to_time), 'UTC'))
                       AND trace_id IN (SELECT id FROM traces_filtered)
                     GROUP BY trace_id
                 )
@@ -313,6 +315,8 @@ class KpiCardDAOImpl implements KpiCardDAO {
                     AND workspace_id = :workspace_id
                     AND id >= :uuid_from_time
                     AND id \\<= :uuid_to_time
+                    AND toMonday(id_at) >= toMonday(UUIDv7ToDateTime(toUUID(:uuid_from_time), 'UTC'))
+                    AND toMonday(id_at) \\<= toMonday(UUIDv7ToDateTime(toUUID(:uuid_to_time), 'UTC'))
                     <if(span_filters)> AND <span_filters> <endif>
                     <if(span_feedback_scores_filters)>
                     AND id in (
@@ -504,6 +508,8 @@ class KpiCardDAOImpl implements KpiCardDAO {
                     FROM spans FINAL
                     WHERE workspace_id = :workspace_id AND project_id = :project_id
                       AND id >= :uuid_from_time AND id \\<= :uuid_to_time
+                      AND toMonday(id_at) >= toMonday(UUIDv7ToDateTime(toUUID(:uuid_from_time), 'UTC'))
+                      AND toMonday(id_at) \\<= toMonday(UUIDv7ToDateTime(toUUID(:uuid_to_time), 'UTC'))
                 ) s
                 JOIN traces_final tr ON s.trace_id = tr.id
                 GROUP BY tr.thread_id
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectMetricsDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectMetricsDAO.java
@@ -550,8 +550,10 @@ class ProjectMetricsDAOImpl implements ProjectMetricsDAO {
                     WHERE project_id = :project_id
                     AND workspace_id = :workspace_id
                     AND trace_id IN (SELECT id FROM traces_filtered)
-                    <if(uuid_from_time)> AND id >= :uuid_from_time<endif>
-                    <if(uuid_to_time)> AND id \\<= :uuid_to_time<endif>
+                    <if(uuid_from_time)> AND id >= :uuid_from_time
+                    AND toMonday(id_at) >= toMonday(UUIDv7ToDateTime(toUUID(:uuid_from_time), 'UTC'))<endif>
+                    <if(uuid_to_time)> AND id \\<= :uuid_to_time
+                    AND toMonday(id_at) \\<= toMonday(UUIDv7ToDateTime(toUUID(:uuid_to_time), 'UTC'))<endif>
                 ) s ON s.trace_id = t.id
             )
             SELECT <bucket> AS bucket,
@@ -580,8 +582,10 @@ class ProjectMetricsDAOImpl implements ProjectMetricsDAO {
                     WHERE project_id = :project_id
                     AND workspace_id = :workspace_id
                     AND trace_id IN (SELECT id FROM traces_filtered)
-                    <if(uuid_from_time)> AND id >= :uuid_from_time<endif>
-                    <if(uuid_to_time)> AND id \\<= :uuid_to_time<endif>
+                    <if(uuid_from_time)> AND id >= :uuid_from_time
+                    AND toMonday(id_at) >= toMonday(UUIDv7ToDateTime(toUUID(:uuid_from_time), 'UTC'))<endif>
+                    <if(uuid_to_time)> AND id \\<= :uuid_to_time
+                    AND toMonday(id_at) \\<= toMonday(UUIDv7ToDateTime(toUUID(:uuid_to_time), 'UTC'))<endif>
                 ) s ON s.trace_id = t.id
             )
             SELECT <bucket> AS bucket,
@@ -607,8 +611,10 @@ class ProjectMetricsDAOImpl implements ProjectMetricsDAO {
                     WHERE project_id = :project_id
                     AND workspace_id = :workspace_id
                     AND trace_id IN (SELECT id FROM traces_filtered)
-                    <if(uuid_from_time)> AND id >= :uuid_from_time<endif>
-                    <if(uuid_to_time)> AND id \\<= :uuid_to_time<endif>
+                    <if(uuid_from_time)> AND id >= :uuid_from_time
+                    AND toMonday(id_at) >= toMonday(UUIDv7ToDateTime(toUUID(:uuid_from_time), 'UTC'))<endif>
+                    <if(uuid_to_time)> AND id \\<= :uuid_to_time
+                    AND toMonday(id_at) \\<= toMonday(UUIDv7ToDateTime(toUUID(:uuid_to_time), 'UTC'))<endif>
                 ) s ON s.trace_id = t.id
                 ARRAY JOIN mapKeys(usage) AS name, mapValues(usage) AS value
                 WHERE value > 0
@@ -641,8 +647,10 @@ class ProjectMetricsDAOImpl implements ProjectMetricsDAO {
                     WHERE project_id = :project_id
                     AND workspace_id = :workspace_id
                     AND trace_id IN (SELECT id FROM traces_filtered)
-                    <if(uuid_from_time)> AND id >= :uuid_from_time<endif>
-                    <if(uuid_to_time)> AND id \\<= :uuid_to_time<endif>
+                    <if(uuid_from_time)> AND id >= :uuid_from_time
+                    AND toMonday(id_at) >= toMonday(UUIDv7ToDateTime(toUUID(:uuid_from_time), 'UTC'))<endif>
+                    <if(uuid_to_time)> AND id \\<= :uuid_to_time
+                    AND toMonday(id_at) \\<= toMonday(UUIDv7ToDateTime(toUUID(:uuid_to_time), 'UTC'))<endif>
                 ) s ON s.trace_id = t.id
                 ARRAY JOIN mapKeys(usage) AS name, mapValues(usage) AS value
                 WHERE value > 0
@@ -1155,8 +1163,10 @@ class ProjectMetricsDAOImpl implements ProjectMetricsDAO {
                     FROM spans final
                     WHERE project_id = :project_id
                     AND workspace_id = :workspace_id
-                    <if(uuid_from_time)> AND id >= :uuid_from_time<endif>
-                    <if(uuid_to_time)> AND id \\<= :uuid_to_time<endif>
+                    <if(uuid_from_time)> AND id >= :uuid_from_time
+                    AND toMonday(id_at) >= toMonday(UUIDv7ToDateTime(toUUID(:uuid_from_time), 'UTC'))<endif>
+                    <if(uuid_to_time)> AND id \\<= :uuid_to_time
+                    AND toMonday(id_at) \\<= toMonday(UUIDv7ToDateTime(toUUID(:uuid_to_time), 'UTC'))<endif>
                 ) s ON s.trace_id = tr.id
             )
             SELECT <bucket> AS bucket,
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/SpanDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/SpanDAO.java
@@ -819,6 +819,11 @@ public class SpanDAO {
      * immaterial since it is id-bounded and {@code LIMIT 1 BY id}. Field exclusion ({@code exclude_fields}) and
      * truncation are layered on top without dropping the sort key.
      * <p>
+     * Each {@code spans} id-range bound carries a parallel {@code toMonday(id_at)} bound: a strict consequence of the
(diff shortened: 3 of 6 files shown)


apps/opik-backend/src/main/java/com/comet/opik/domain/ProjectMetricsDAO.java
```diff
@@ -550,8 +550,10 @@ TO toDateTime(UUIDv7ToDateTime(toUUID(:uuid_to_time)))
...
                     AND workspace_id = :workspace_id
                     AND trace_id IN (SELECT id FROM traces_filtered)
-                    <if(uuid_from_time)> AND id >= :uuid_from_time<endif>
-                    <if(uuid_to_time)> AND id \\<= :uuid_to_time<endif>
+                    <if(uuid_from_time)> AND id >= :uuid_from_time
+                    AND toMonday(id_at) >= toMonday(UUIDv7ToDateTime(toUUID(:uuid_from_time), 'UTC'))<endif>
+                    <if(uuid_to_time)> AND id \\<= :uuid_to_time
+                    AND toMonday(id_at) \\<= toMonday(UUIDv7ToDateTime(toUUID(:uuid_to_time), 'UTC'))<endif>
                 ) s ON s.trace_id = t.id
             )
```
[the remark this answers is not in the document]
thiagohora: Good call to check. The high-fan-out span-metric path (duration / count / cost / token-usage / error-rate) is already centralized: it renders through `SpanMetricsQueries.spanFilteredPrefix`, where this predicate is defined exactly once. The remaining inline copies (`GET_COST`/`GET_TOKEN_USAGE`/`GET_THREAD_COST`) are a different query shape — a raw `spans` subquery joined to `traces_filtered` by `trace_id`, not the `spans_filtered` CTE — so they can't reuse that prefix.

I looked at pulling the bound into a single shared fragment, but across the read paths it isn't uniform: bounds vary by parameter (`uuid_from_time`/`uuid_to_time` vs `id_start`/`id_end` `BETWEEN` in WorkspaceMetricsDAO vs `last_received_span_id` for pagination), by operator, by `<if>`-gating vs unconditional, and by StringTemplate-vs-raw escaping. A one-size fragment would need a param-normalizing layer that hides each query's actual bounds. Inlining here also mirrors the merged traces counterpart (#7235), keeping the trace/span predicates symmetric and greppable. So the shared path is factored; the rest is intentionally inline.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanDAO.java
```diff
@@ -1226,11 +1244,25 @@ ORDER BY (workspace_id, project_id, trace_id, parent_span_id, id) DESC, last_upd
...
     private static final String DELETE_FOR_RETENTION = """
             DELETE FROM spans
             WHERE workspace_id IN :workspace_ids
             AND trace_id >= :lower_bound
             AND trace_id \\< :cutoff_id
+            AND toMonday(id_at) >= toMonday(UUIDv7ToDateTime(toUUID(:lower_bound), 'UTC'))
+            AND toMonday(id_at) \\< addWeeks(toMonday(UUIDv7ToDateTime(toUUID(:cutoff_id), 'UTC')), 1)
             AND trace_id NOT IN (
                 SELECT trace_id FROM experiment_items
                 WHERE workspace_id IN :workspace_ids
```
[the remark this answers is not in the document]
thiagohora: Worth separating two cases here, because this block isn't part of 7307's diff — `DELETE_FOR_RETENTION`/`COUNT_FOR_RETENTION` come from the stacked base #7471 (OPIK-7306).

- **This PR (7307)** only adds `toMonday(id_at)` bounds to scans ranged on the span's **own `id`** (`id >= :uuid_from_time`, `id BETWEEN …`). There the upper bound is a strict consequence of the existing id filter: `id_at = UUIDv7ToDateTime(id)` is a direct function of that same `id`, so no row the id-range keeps can fall outside the mirrored week. It's a semantic no-op. The `trace_id`-ranged span scans (where a span's `id_at` can legitimately lag its trace) were deliberately **excluded** from this PR for exactly the reason you raise.

- The flagged retention paths range on **`trace_id`**, so you're right that the `addWeeks(1)` slack can skip spans whose `id_at` lands more than a week after their trace (long-running sessions, async/batch logging). That's a property of the trace_id-derived bound and belongs to #7471 where this code lives; the fix there is to drop the upper `id_at` bound and keep only the lower floor for pruning. Tracking it against that PR.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanDAO.java
```diff
@@ -1055,6 +1066,9 @@ ORDER BY (workspace_id, project_id, id) DESC, last_updated_at DESC
                 WHERE workspace_id = :workspace_id
                 AND project_id = :project_id
                 AND id IN (SELECT id FROM page_ids)
+                <if(uuid_from_time)> AND toMonday(id_at) >= toMonday(UUIDv7ToDateTime(toUUID(:uuid_from_time), 'UTC')) <endif>
+                <if(uuid_to_time)> AND toMonday(id_at) \\<= toMonday(UUIDv7ToDateTime(toUUID(:uuid_to_time), 'UTC')) <endif>
+                <if(last_received_span_id)> AND toMonday(id_at) \\<= toMonday(UUIDv7ToDateTime(toUUID(:last_received_span_id), 'UTC')) <endif>
```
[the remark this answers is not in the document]
thiagohora: Added in 682b26113e — `FindSpansResourceTest#whenTimeWindowSortExcludeAcrossPages__thenEachPageMatchesFullPageAndReference`. It combines a UUID creation-time window (`from_time`/`to_time`) with pagination, a non-default sort (input ASC), and field exclusion (output), and asserts the full page content of each page against an independent sorted + excluded reference (via `SpanAssertions`, not id-only). That drives the exact `page_wide` branch — the window brackets every span so the `toMonday(id_at)` bounds are a strict no-op, and the byte-for-byte assertions would fail on any week-bound, ordering, or field-exclusion regression in the two-phase query.

apps/opik-backend/src/main/java/com/comet/opik/domain/KpiCardDAO.java
```diff
@@ -200,6 +200,8 @@ SELECT trace_id, sum(total_estimated_cost) AS cost
                     FROM spans FINAL
                     WHERE workspace_id = :workspace_id AND project_id = :project_id
                       AND id >= :uuid_from_time AND id \\<= :uuid_to_time
+                      AND toMonday(id_at) >= toMonday(UUIDv7ToDateTime(toUUID(:uuid_from_time), 'UTC'))
+                      AND toMonday(id_at) \\<= toMonday(UUIDv7ToDateTime(toUUID(:uuid_to_time), 'UTC'))
                       AND trace_id IN (SELECT id FROM traces_filtered)
```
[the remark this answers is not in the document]
thiagohora: The `toMonday(id_at)` bounds here are a strict consequence of the **pre-existing** `id >= :uuid_from_time AND id <= :uuid_to_time` span-id filter, not a new scope. On main, both `trace_costs` and `thread_costs` already window the span scan by the span's own `id` — this PR's diff only appends the two `toMonday(id_at)` lines. For any span the id-range keeps: `id ∈ [from, to] ⇒ id_at ∈ [from_time, to_time] ⇒ toMonday(id_at) ∈ [toMonday(from), toMonday(to)]`, so the week bounds can't drop a row the existing filter didn't already drop. This change is a no-op on the KPI totals.

The truncation you describe (a span of an in-range trace whose own UUID is past `uuid_to_time`) is a property of that pre-existing span-id window, which this PR neither adds nor changes. Two notes on it: the **lower** edge can't truncate — a span's id is >= its trace's id (span created at/after its trace), so every span of an in-window trace is already >= `uuid_from_time`; only late spans past the **upper** edge are excluded. Whether KPI cost should be scoped by span creation time or purely by the trace join is a product-semantics question about the shipped query — reasonable as a separate issue, but out of scope for this partition-pruning PR. And note the suggested fix (drop the span id/week predicates, rely on the trace join) would remove the very partition pruning these predicates exist to enable, since `trace_id IN (…)` gives the planner no bound on `toMonday(id_at)`.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanDAO.java
andrescrz, line 1: I believe you left references behind in the `ThreadDAO`: it contains queries on spans tables that should likely have the predicate as well.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanMetricsQueries.java
andrescrz, line 1: Same about the `TraceDAO`, I believe it also contains queries on spans tables that should likely have the predicate as well.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/FindSpansResourceTest.java
```diff
@@ -4261,6 +4261,66 @@ void whenFilterSortExcludeAcrossPages__thenEachPageMatchesFullPageAndReference()
...
+            // Independent reference: every span (projectName nulled by the API), sorted by input ASC, output excluded.
+            Comparator<Span> byInput = Comparator.comparing(s -> s.input().toString());
+            var reference = spans.stream()
+                    .map(s -> s.toBuilder().projectName(null).build())
+                    .sorted(byInput)
+                    .map(s -> SpanAssertions.EXCLUDE_FUNCTIONS.get(Span.SpanField.OUTPUT).apply(s))
+                    .toList();
+
+            int pages = (total + pageSize - 1) / pageSize;
+            for (int p = 1; p <= pages; p++) {
```
andrescrz, line 4317: Minor: maybe an assertion on the number of pages, so we ensure this always loops, failing the test instead of passing silently if no iterations here.

andrescrz (commented):
The change LGTM, but I believe we left some stuff behind.

Please do another iteration on the audit to cover everything missing (if any). Make sure you cover `TraceDAO` and `ThreadDAO`, but not limited to that. Let's makes sure we don't leave anything behind.

In addition, a minor comment for extra resilience in the test.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanDAO.java
thiagohora: Audited these. The ThreadDAO span scans filter on **`trace_id`** (linking spans to in-window traces), not on the span's own `id`. `toMonday(id_at)` only prunes when paired with a span-`id` range (it's the future partition key), so to add it here we'd have to introduce an `id`-range derived from the `trace_id` range — and that isn't sound.

`trace_id` range and `id` range are **not result-equivalent**: a span's `id` is not guaranteed >= its trace's `id` (client/custom-supplied ids, backdated spans, malformed ids), so `id >= :from` would drop spans whose trace is in-window but whose own id sorts below `:from`. For a read that's an undercount — a correctness bug. It's the same imprecision we discussed on #7471, where it's acceptable only because retention is a *delete* (under-selection is the safe direction). I prototyped + benchmarked the id-range variant in prod: it did read fewer parts and ran faster, but precisely *because* it was excluding real rows — so it's not a valid no-op for a read path.

So these `trace_id`-scoped scans stay as-is (they still prune on `trace_id` via the primary sort key `(workspace_id, project_id, trace_id, …)`); only the span-`id`-ranged read paths carry `toMonday(id_at)`.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanMetricsQueries.java
thiagohora: Same conclusion as the ThreadDAO thread: TraceDAO's span scans key on **`trace_id`** (span↔trace linkage), not the span's own `id`. Since `toMonday(id_at)` is only valid alongside a span-`id` range, adding it would require deriving an `id`-range from the `trace_id` range — but the two are not result-equivalent (a span's `id` isn't guaranteed >= its trace's `id`: custom/backdated/malformed ids), so `id >= :from` would undercount by dropping in-window-trace spans whose id sorts below `:from`. Safe for a *delete* (#7471), a correctness bug for a *read*. Left as-is; only the span-`id`-ranged read paths (SpanDAO find/search/stats, the metrics `spans_filtered` CTE, KPI span/thread/trace cost aggs) carry the predicate.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/FindSpansResourceTest.java
thiagohora: Done in d966b346a4 — added `assertThat(pages).isGreaterThan(1)` before the loop so it fails instead of passing silently if it ever stops iterating.

andrescrz (approved):
LGTM

PR #7477, opened 2026-07-15
[OPIK-7086] [QA] test(e2e): add Optimization Studio E2E coverage

AndreiCautisanu:
## Details

Adds end-to-end coverage for **Optimization Studio** (v2): three Playwright tests that drive the studio UI to launch real optimization runs and assert they complete healthily, plus supporting POM / backend-client / tier wiring.

- **Form renders + validation gating** (no LLM run): sections render (prompt, model, dataset picker, GEPA + Equals defaults) and Optimize is disabled until valid.
- **GEPA + Equals launch, deep**: configure + launch via UI → poll to `completed` → structural health (`num_trials ≥ 1`, `objective_name`, baseline/best scores present & in `[0,1]` — never "improved") → detail-page Trials tab (Best row) + Best-trial-config panel → studio **logs** read-path (presigned URL + gunzip).
- **Hierarchical Reflective variant**: same flow, second optimizer.
- Tagged `@t2-cuj` + `@t1-stsaas` (kept out of plain `@t1-smoke`); adds a `test:t1-stsaas` script so customer/stsaas regression runs can include the studio checks (companion: comet-automation-tests#273).
- Fragility handled in assertions: score-0 is healthy (assert range, not improvement); studio-logs presigned URL is client-reachable only on real object stores, so the log-content check is enforced on non-oss and skipped-with-warning on local MinIO; selectors tolerate the "Select item source" vs "Select a dataset" rename.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves OPIK-7086

## AI-WATERMARK

AI-WATERMARK: yes

- If yes:
  - Tools: Claude Code
  - Model(s): Claude Opus 4.8
  - Scope: Authored the POM, spec, backend-client methods, poller, and tier/script wiring; explored the live UI on staging/local; iterated to green.
  - Human verification: Developer reviewed the diff and directed the coverage design, tiering, and the deployment-gated logs enforcement; studio run path independently confirmed working end-to-end on self-hosted-eks.

## Testing

Suite: `tests_end_to_end/e2e/`.

- `npx tsc --noEmit` — clean.
- `pre-commit run --files <changed>` — clean.
- Tag/tier wiring verified with `--list`: `@t1-smoke` excludes the studio tests; `test:t1-stsaas` (`@t1-smoke|@t1-stsaas`) and `test:t2` include all three.
- The three tests were run green locally against a local OSS instance during development (form + GEPA-deep + HR variant), and the flows were verified manually on Comet staging. The studio launch→completed path was also confirmed end-to-end on self-hosted-eks after the platform fix (comet-backend#5616) deployed.
- Command to run locally: `OPIK_DEPLOYMENT=oss OPIK_BASE_URL=http://localhost:5173 OPIK_WORKSPACE=default npm run test:t1-stsaas` (requires an LLM provider key in env; `ensureModelAvailable` skips cleanly without one).
- Not re-run after the final file rebuild: the working files were re-created from the verified versions (identical, tsc-clean) but the local OSS stack was down at commit time, so the post-rebuild full-suite run is pending a running stack. The stsaas logs-content assertion is only exercised on a non-oss target (staging/stsaas), not local OSS.

## Documentation

No documentation changes — internal E2E test infrastructure.

### DIFF
--- a/tests_end_to_end/e2e/core/backend/client.ts
+++ b/tests_end_to_end/e2e/core/backend/client.ts
@@ -1,9 +1,15 @@
+import { gunzipSync } from 'node:zlib';
 import { Opik } from 'opik';
 import { loadEnvConfig } from '../../config/env.config';
 import {
   pollTraceForFeedbackScore,
   type PollFeedbackScoreOpts,
 } from './poll-feedback-score';
+import {
+  pollOptimizationStatus,
+  type OptimizationStatus,
+  type PollOptimizationStatusOpts,
+} from './poll-optimization-status';
 
 export type BackendClient = ReturnType<typeof makeBackendClient>;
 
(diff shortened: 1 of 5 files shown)


andrescrz (approved):
LGTM

PR #7468, opened 2026-07-14
[OPIK-7336] [DOCS] Document span/trace ingestion size limits + attachments

YarivHashaiComet:
## Details

Adds two FAQ entries documenting the upcoming span/trace **ingestion size limits** and the Attachments alternative, in the Latest docs version (`fern/docs-v2/faq.mdx`):

- **Opik Cloud →** *"Is there a limit on how large a trace or span can be?"* — per-field (~20 MB, SDK truncates), per-request (~50 MB compressed / ~256 MB uncompressed → `413`/`400`), and pointing to **Attachments** for large payloads.
- **Troubleshooting →** *"Why am I getting a 413 (payload too large) error?"*.

Part of the effort to enforce span/trace ingestion size limits and stop the unbounded-payload OOM (read-path OPIK-6999, write-path OPIK-7273).

⚠️ **Draft / merge-with-release:** the `~20 / ~50 / ~256 MB` values must match what the backend (OPIK-7333/7334) and SDK (OPIK-7335) actually ship. Kept as a **draft** to merge *alongside* the enforcement release, not ahead of it.

**Error-code note:** `RequestSizeLimitFilter` returns `413`; `maxDocumentLength` currently returns `400` (Jackson `StreamConstraintsException` → `JsonProcessingExceptionMapper`). Docs say "413/400" until OPIK-7334 normalizes it to `413`.

## Change checklist

- [x] User facing
- [x] Documentation update

## Issues

- OPIK-7336 (parent: OPIK-7312)

## Testing

Docs-only change (no code paths affected). Validate the rendered FAQ locally with `cd apps/opik-documentation/documentation && npm run dev` — there is no automated docs test suite.

## Documentation

This PR **is** the documentation change: two new FAQ entries in `fern/docs-v2/faq.mdx` (Latest docs version). Links use the version-relative slug to the Attachments page (`/tracing/advanced/log_multimodal_traces`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7483, opened 2026-07-15
[OPIK-7333] [BE] server-side ingestion size guards (request 413 + document 4xx)

YarivHashaiComet:
## Details

Server-side half of the unbounded-payload OOM fix (parent **OPIK-7312**; the client-side per-span cap is **OPIK-7335** / #7469).

**OPIK-7333 — `RequestSizeLimitFilter` (413):** rejects a request whose `Content-Length` exceeds `maxRequestSizeBytes` with **413 before the body is read**, so the JSON parser never starts on an oversized batch. A request with no `Content-Length` (chunked) falls through and is bounded by `maxDocumentLength`. The header is parsed as a `long` so a >2 GB body can't slip past an int overflow; a malformed/negative header fails closed with 400. Guicey auto-installs it as a JAX-RS provider and injects `@Config JacksonConfig` (same sub-config pattern as `TraceThreadConfig` / `OnlineScoringConfig`).

**OPIK-7334 — `maxDocumentLength` (4xx):** caps the whole decompressed document on both the HTTP and `JsonUtils` ObjectMappers via `StreamReadConstraints`, so an oversized batch aborts **mid-parse with a 4xx** (via `JsonProcessingExceptionMapper`) before a multi-GB `JsonNode` tree is materialized.

### Limits (config-driven, env-overridable)

| Guard | Self-hosted default | Cloud | Role |
|---|---|---|---|
| `maxStringLength` | 100 MB | 100 MB | per-field cap (unchanged) |
| `maxDocumentLength` | **512 MB** | 256 MB | whole-document cap → 4xx |
| `maxRequestSizeBytes` | **512 MB** | 256 MB | request `Content-Length` → 413 |

Request cap is kept `== maxDocumentLength` so the filter never rejects a request the document guard would otherwise accept. Self-hosted gets 512 MB headroom; cloud overrides both to 256 MB via `JACKSON_MAX_DOCUMENT_LENGTH` / `JACKSON_MAX_REQUEST_SIZE_BYTES`.

## Change checklist

- [x] User facing — oversized ingestion now returns 413/400 instead of destabilising the backend (OOM)
- [x] Tested — unit + integration (see Testing)

## Issues

- OPIK-7333, OPIK-7334 (parent: OPIK-7312)

## Testing

Run via `mvn test` (JDK 25 + testcontainers), all green:

- **Unit (11):** `RequestSizeLimitFilterTest` (over→413, at-limit pass, no-`Content-Length` pass, >2 GB→413, malformed/negative→400); `JsonUtilsDocumentLengthTest` (over-doc→`StreamConstraintsException` via a streamed `InputStream`, under parses, unlimited parses, `maxStringLength` still enforced); `JacksonConfigTest` (512 MB defaults + `request == document` invariant).
- **Integration — `SpansResourceTest.LargePayloadTests` (4):** ~93 MB inline base64 → succeeds + stripped to S3; 253 MB single string → 400 (`maxStringLength`); 280 MB total document → 400 (`maxDocumentLength`); buffered client (Grizzly + `Expect: 100-continue`) with `Content-Length` over the cap → 413. (The default chunked test client bypasses the request filter, which is why the 413 test uses a buffered client.)

`config-test.yml` uses lower, payload-calibrated caps so the guard tests trip at modest sizes; `JacksonConfigTest` verifies the real 512 MB shipped defaults.

## Documentation

- Customer-facing docs (OPIK-7336 / #7468) are intentionally kept **more conservative** than the actual limits — no change here.
- New config keys are documented inline in `config.yml`.
- **Deployment (out of this repo):** the cloud 256 MB override lives in `comet-helm/values-production-opik.yaml` (`component.backend.env`), tracked as **OPIK-7353**. Without it, cloud inherits the 512 MB self-hosted default.

### DIFF
--- a/apps/opik-backend/config.yml
+++ b/apps/opik-backend/config.yml
@@ -339,6 +339,20 @@ jacksonConfig:
   #   This configuration is used by both HTTP layer and internal JSON processing for consistency,
   #   preventing memory exhaustion from extremely large base64-encoded attachments before they can be stripped
   maxStringLength: ${JACKSON_MAX_STRING_LENGTH:-104857600}
+  # Default: 536870912 (512MB) - the self-hosted default (higher, to avoid surprising customers);
+  #   the Comet cloud deployment overrides JACKSON_MAX_DOCUMENT_LENGTH to 268435456 (256MB).
+  # Description: Maximum size of an entire JSON document (the whole decompressed batch). Unlike
+  #   maxStringLength (a single string value), this caps the whole document, so it catches an
+  #   oversized batch built from many small values. Aborts mid-parse with a 4xx before a multi-GB
+  #   node tree is materialized. Must stay above maxStringLength so a single max-size string parses.
+  maxDocumentLength: ${JACKSON_MAX_DOCUMENT_LENGTH:-536870912}
+  # Default: 536870912 (512MB) - self-hosted default; cloud overrides JACKSON_MAX_REQUEST_SIZE_BYTES
+  #   to 268435456 (256MB). Kept equal to maxDocumentLength so the request filter never rejects a
+  #   request the document guard would otherwise accept.
+  # Description: Maximum incoming request body size, checked against the Content-Length header and
+  #   rejected with 413 before the body is read. Requests without a Content-Length (e.g. chunked)
+  #   fall through and are bounded by maxDocumentLength during parsing.
+  maxRequestSizeBytes: ${JACKSON_MAX_REQUEST_SIZE_BYTES:-536870912}
 
 # Configuration for batch operations
 batchOperations:
--- a/apps/opik-backend/config.yml
+++ b/apps/opik-backend/config.yml
@@ -339,16 +339,12 @@ jacksonConfig:
   #   This configuration is used by both HTTP layer and internal JSON processing for consistency,
   #   preventing memory exhaustion from extremely large base64-encoded attachments before they can be stripped
   maxStringLength: ${JACKSON_MAX_STRING_LENGTH:-104857600}
-  # Default: 536870912 (512MB) - the self-hosted default (higher, to avoid surprising customers);
-  #   the Comet cloud deployment overrides JACKSON_MAX_DOCUMENT_LENGTH to 268435456 (256MB).
+  # Default: 536870912 (512MB)
   # Description: Maximum size of an entire JSON document (the whole decompressed batch). Unlike
-  #   maxStringLength (a single string value), this caps the whole document, so it catches an
-  #   oversized batch built from many small values. Aborts mid-parse with a 4xx before a multi-GB
-  #   node tree is materialized. Must stay above maxStringLength so a single max-size string parses.
+  #   maxStringLength (a single string value), this caps the whole document, catching an oversized
+  #   batch built from many small values. Must stay above maxStringLength.
   maxDocumentLength: ${JACKSON_MAX_DOCUMENT_LENGTH:-536870912}
-  # Default: 536870912 (512MB) - self-hosted default; cloud overrides JACKSON_MAX_REQUEST_SIZE_BYTES
-  #   to 268435456 (256MB). Kept equal to maxDocumentLength so the request filter never rejects a
-  #   request the document guard would otherwise accept.
+  # Default: 536870912 (512MB)
   # Description: Maximum incoming request body size, checked against the Content-Length header and
   #   rejected with 413 before the body is read. Requests without a Content-Length (e.g. chunked)
   #   fall through and are bounded by maxDocumentLength during parsing.
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/error/JsonProcessingExceptionMapper.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/error/JsonProcessingExceptionMapper.java
@@ -40,16 +40,18 @@ public class JsonProcessingExceptionMapper implements ExceptionMapper<JsonProces
             log.debug("Ingestion size guard rejected a request", exception); // expected; already on the metric
             sizeGuardMetrics.recordStreamConstraintRejection(streamConstraintsException, uriInfo.get(), requestContext);
             status = Response.Status.REQUEST_ENTITY_TOO_LARGE;
+            // Redact: the size-guard message exposes internal limits + Jackson internals (StreamReadConstraints,
+            // the configured max). Full detail stays in the server log above, not the HTTP response.
             clientMessage = "Request payload exceeds the maximum allowed size.";
         } else {
             log.info("Deserialization exception for workspace {}",
                     ErrorMetricsResolver.workspaceId(requestContext), exception);
             status = Response.Status.BAD_REQUEST;
-            clientMessage = "Unable to process the request body: it is not valid JSON.";
+            // Ordinary malformed/invalid JSON: echo the parser detail (the caller's own payload + our own
+            // field names, not internal limits) - useful for the client and the long-standing contract.
+            clientMessage = "Unable to process JSON. " + exception.getMessage();
         }
 
-        // Stable, generic message: exception.getMessage() can carry parser internals and fragments of the
-        // client's payload, so the detail stays in the server log above, not the HTTP response.
         return Response.status(status)
                 .entity(new ErrorMessage(status.getStatusCode(), clientMessage))
                 .build();
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/error/JsonProcessingExceptionMapper.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/error/JsonProcessingExceptionMapper.java
@@ -32,8 +32,15 @@ public class JsonProcessingExceptionMapper implements ExceptionMapper<JsonProces
         // A StreamConstraintsException is the document-/string-length size guard tripping mid-parse
         // (OPIK-7334) — an EXPECTED client rejection, already surfaced via the ingestion size-guard
         // metric. Keep it at DEBUG so normal oversized-payload traffic doesn't flood INFO logs.
-        if (exception instanceof StreamConstraintsException streamConstraintsException) {
-            log.debug("Ingestion size guard rejected a request", streamConstraintsException);
+        //
+        // It rarely arrives as the thrown type: when the overrun trips while binding a bean property
+        // (the common case — a well-formed SpanBatch/TraceBatch whose input/output/metadata is
+        // oversized), Jackson wraps it in a JsonMappingException carrying the reference chain, so a
+        // plain `instanceof` check misses it and the metric under-counts exactly the real-world case
+        // the dashboard monitors. Walk the cause chain to find the underlying constraint.
+        StreamConstraintsException streamConstraintsException = findStreamConstraint(exception);
+        if (streamConstraintsException != null) {
+            log.debug("Ingestion size guard rejected a request", exception);
             sizeGuardMetrics.recordStreamConstraintRejection(streamConstraintsException, uriInfo.get(), requestContext);
         } else {
             // Genuine malformed JSON: log the exception itself (not just its message) so the type and
@@ -45,5 +52,24 @@ public class JsonProcessingExceptionMapper implements ExceptionMapper<JsonProces
                 .entity(new ErrorMessage(Response.Status.BAD_REQUEST.getStatusCode(),
                         "Unable to process JSON. " + exception.getMessage()))
                 .build();
+    }
+
+    /**
+     * Return the {@link StreamConstraintsException} the given throwable is or wraps, or {@code null} if
+     * none is present. Jackson raises the size-guard overrun at the parser level but re-wraps it in a
+     * {@link com.fasterxml.jackson.databind.JsonMappingException} (with the bean reference chain) when
+     * it trips during databinding, so the constraint is usually a cause rather than the thrown type.
+     * The walk is bounded and self-reference-safe to avoid looping on a malformed cause chain.
+     */
+    private static StreamConstraintsException findStreamConstraint(Throwable throwable) {
+        for (Throwable cause = throwable; cause != null; cause = cause.getCause()) {
+            if (cause instanceof StreamConstraintsException streamConstraintsException) {
+                return streamConstraintsException;
+            }
+            if (cause == cause.getCause()) {
+                break;
+            }
+        }
+        return null;
     }
 }
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/error/JsonProcessingExceptionMapper.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/error/JsonProcessingExceptionMapper.java
@@ -39,18 +39,24 @@ public class JsonProcessingExceptionMapper implements ExceptionMapper<JsonProces
         // plain `instanceof` check misses it and the metric under-counts exactly the real-world case
         // the dashboard monitors. Walk the cause chain to find the underlying constraint.
         StreamConstraintsException streamConstraintsException = findStreamConstraint(exception);
+        Response.Status status;
         if (streamConstraintsException != null) {
+            // A size-guard trip is a payload-too-large rejection, not malformed JSON, so surface it as
+            // 413 REQUEST_ENTITY_TOO_LARGE — consistent with RequestSizeLimitFilter's pre-parse 413 —
+            // rather than 400. It is an EXPECTED client rejection, already surfaced via the ingestion
+            // size-guard metric, so keep it at DEBUG so oversized-payload traffic doesn't flood INFO.
             log.debug("Ingestion size guard rejected a request", exception);
             sizeGuardMetrics.recordStreamConstraintRejection(streamConstraintsException, uriInfo.get(), requestContext);
+            status = Response.Status.REQUEST_ENTITY_TOO_LARGE;
         } else {
             // Genuine malformed JSON: log the exception itself (not just its message) so the type and
             // stack trace are available to diagnose the rare, unexpected parse failure.
             log.info("Deserialization exception", exception);
+            status = Response.Status.BAD_REQUEST;
         }
 
-        return Response.status(Response.Status.BAD_REQUEST)
-                .entity(new ErrorMessage(Response.Status.BAD_REQUEST.getStatusCode(),
-                        "Unable to process JSON. " + exception.getMessage()))
+        return Response.status(status)
+                .entity(new ErrorMessage(status.getStatusCode(), "Unable to process JSON. " + exception.getMessage()))
                 .build();
     }
 
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/error/JsonProcessingExceptionMapper.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/error/JsonProcessingExceptionMapper.java
@@ -8,10 +8,12 @@ import com.fasterxml.jackson.core.exc.StreamConstraintsException;
 import io.dropwizard.jersey.errors.ErrorMessage;
 import jakarta.inject.Inject;
 import jakarta.inject.Provider;
+import jakarta.ws.rs.core.MediaType;
 import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.core.UriInfo;
 import jakarta.ws.rs.ext.ExceptionMapper;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.exception.ExceptionUtils;
 
 @Slf4j
 public class JsonProcessingExceptionMapper implements ExceptionMapper<JsonProcessingException> {
@@ -30,17 +32,32 @@ public class JsonProcessingExceptionMapper implements ExceptionMapper<JsonProces
 
     @Override
     public Response toResponse(JsonProcessingException exception) {
-        // A StreamConstraintsException (often wrapped by Jackson during bean binding, hence the cause walk)
-        // is the size guard tripping mid-parse -> 413; anything else is malformed JSON -> 400.
-        StreamConstraintsException streamConstraintsException = findStreamConstraint(exception);
+        // A StreamConstraintsException (often wrapped by Jackson during binding) is a stream-read limit
+        // tripping mid-parse. getThrowableList is cycle-safe, so no manual bound is needed. A size limit
+        // (document/string length) -> 413; a structural limit (nesting/number) or anything else -> 400.
+        StreamConstraintsException streamConstraint = ExceptionUtils.getThrowableList(exception).stream()
+                .filter(StreamConstraintsException.class::isInstance)
+                .map(StreamConstraintsException.class::cast)
+                .findFirst()
+                .orElse(null);
+
         Response.Status status;
         String clientMessage;
-        if (streamConstraintsException != null) {
-            log.debug("Ingestion size guard rejected a request", exception); // expected; already on the metric
-            sizeGuardMetrics.recordStreamConstraintRejection(streamConstraintsException, uriInfo.get(), requestContext);
-            status = Response.Status.REQUEST_ENTITY_TOO_LARGE;
-            // Redacted: the size-guard message exposes internal limits/Jackson internals (detail is in the log).
-            clientMessage = "Request payload exceeds the maximum allowed size.";
+        if (streamConstraint != null) {
+            sizeGuardMetrics.recordStreamConstraintRejection(streamConstraint, uriInfo.get(), requestContext);
+            String guard = IngestionSizeGuardMetrics.classifyStreamConstraint(streamConstraint);
+            if (IngestionSizeGuardMetrics.GUARD_DOCUMENT_LENGTH.equals(guard)
+                    || IngestionSizeGuardMetrics.GUARD_STRING_LENGTH.equals(guard)) {
+                log.debug("Ingestion size guard rejected a request", exception); // expected; already on the metric
+                status = Response.Status.REQUEST_ENTITY_TOO_LARGE;
+                clientMessage = "Request payload exceeds the maximum allowed size."; // redacted; detail in the log
+            } else {
+                // A structural limit (nesting/number depth), not an oversize -> 400; kept generic since the
+                // message also carries Jackson internals.
+                log.info("Rejecting a request that violates a JSON structural limit", exception);
+                status = Response.Status.BAD_REQUEST;
+                clientMessage = "Unable to process JSON. The request exceeds an allowed structural limit.";
+            }
         } else {
             log.info("Deserialization exception for workspace {}",
                     ErrorMetricsResolver.workspaceId(requestContext), exception);
@@ -50,23 +67,12 @@ public class JsonProcessingExceptionMapper implements ExceptionMapper<JsonProces
             clientMessage = "Unable to process JSON. " + exception.getMessage();
         }
 
+        // Force JSON: ingestion endpoints negotiate non-JSON (e.g. OTel protobuf) with no writer for
+        // ErrorMessage, so without an explicit type the error fails to serialize and surfaces as a 500
+        // (matches InvalidUUIDExceptionMapper and RequestSizeLimitFilter).
         return Response.status(status)
+                .type(MediaType.APPLICATION_JSON_TYPE)
                 .entity(new ErrorMessage(status.getStatusCode(), clientMessage))
                 .build();
-    }
-
-    // Bounds the cause walk so a pathological cause cycle (A -> B -> A) can't spin forever and pin the
-    // request thread; 50 is far beyond any real chain.
-    private static final int MAX_CAUSE_DEPTH = 50;
-
-    /** The {@link StreamConstraintsException} this throwable is or wraps, or {@code null}. */
-    private static StreamConstraintsException findStreamConstraint(Throwable throwable) {
-        Throwable cause = throwable;
-        for (int depth = 0; cause != null && depth < MAX_CAUSE_DEPTH; cause = cause.getCause(), depth++) {
-            if (cause instanceof StreamConstraintsException streamConstraintsException) {
-                return streamConstraintsException;
-            }
-        }
-        return null;
     }
 }
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/error/JsonProcessingExceptionMapper.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/error/JsonProcessingExceptionMapper.java
@@ -1,17 +1,41 @@
 package com.comet.opik.api.error;
 
+import com.comet.opik.infrastructure.auth.RequestContext;
+import com.comet.opik.infrastructure.metrics.IngestionSizeGuardMetrics;
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.exc.StreamConstraintsException;
 import io.dropwizard.jersey.errors.ErrorMessage;
+import jakarta.inject.Inject;
+import jakarta.inject.Provider;
 import jakarta.ws.rs.core.Response;
+import jakarta.ws.rs.core.UriInfo;
 import jakarta.ws.rs.ext.ExceptionMapper;
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 public class JsonProcessingExceptionMapper implements ExceptionMapper<JsonProcessingException> {
 
+    private final IngestionSizeGuardMetrics sizeGuardMetrics;
+    private final Provider<RequestContext> requestContext;
+    private final Provider<UriInfo> uriInfo;
+
+    @Inject
+    public JsonProcessingExceptionMapper(IngestionSizeGuardMetrics sizeGuardMetrics,
+            Provider<RequestContext> requestContext, Provider<UriInfo> uriInfo) {
+        this.sizeGuardMetrics = sizeGuardMetrics;
+        this.requestContext = requestContext;
+        this.uriInfo = uriInfo;
+    }
+
     @Override
     public Response toResponse(JsonProcessingException exception) {
         log.info("Deserialization exception: {}", exception.getMessage());
+
+        // A StreamConstraintsException here is the document- or string-length size guard tripping
+        // mid-parse (OPIK-7334), not ordinary malformed JSON — count it so the guard is observable.
+        if (exception instanceof StreamConstraintsException streamConstraintsException) {
+            sizeGuardMetrics.recordStreamConstraintRejection(streamConstraintsException, uriInfo.get(), requestContext);
+        }
 
         return Response.status(Response.Status.BAD_REQUEST)
                 .entity(new ErrorMessage(Response.Status.BAD_REQUEST.getStatusCode(),
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/error/JsonProcessingExceptionMapper.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/error/JsonProcessingExceptionMapper.java
@@ -1,6 +1,7 @@
 package com.comet.opik.api.error;
 
 import com.comet.opik.infrastructure.auth.RequestContext;
+import com.comet.opik.infrastructure.metrics.ErrorMetricsResolver;
 import com.comet.opik.infrastructure.metrics.IngestionSizeGuardMetrics;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.exc.StreamConstraintsException;
@@ -50,8 +51,10 @@ public class JsonProcessingExceptionMapper implements ExceptionMapper<JsonProces
             status = Response.Status.REQUEST_ENTITY_TOO_LARGE;
         } else {
             // Genuine malformed JSON: log the exception itself (not just its message) so the type and
-            // stack trace are available to diagnose the rare, unexpected parse failure.
-            log.info("Deserialization exception", exception);
+            // stack trace are available to diagnose the rare, unexpected parse failure; include the
+            // workspace to help attribute the offending client.
+            log.info("Deserialization exception for workspace {}",
+                    ErrorMetricsResolver.workspaceId(requestContext), exception);
             status = Response.Status.BAD_REQUEST;
         }
 
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/error/JsonProcessingExceptionMapper.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/error/JsonProcessingExceptionMapper.java
@@ -55,17 +55,21 @@ public class JsonProcessingExceptionMapper implements ExceptionMapper<JsonProces
                 .build();
     }
 
+    // Real exception cause chains are only a handful deep; this cap is far beyond any legitimate chain
+    // yet stops a pathological cause cycle (A -> B -> A, which Throwable.initCause permits — it rejects
+    // only a direct self-cause) from spinning the walk forever and pinning the request thread.
+    private static final int MAX_CAUSE_DEPTH = 50;
+
     /**
      * The {@link StreamConstraintsException} this throwable is or wraps (Jackson re-wraps it in a
-     * JsonMappingException during bean binding), or {@code null}. Self-reference-safe.
+     * JsonMappingException during bean binding), or {@code null}. The walk is bounded by
+     * {@link #MAX_CAUSE_DEPTH} so a cyclic cause chain cannot hang the request thread.
      */
     private static StreamConstraintsException findStreamConstraint(Throwable throwable) {
-        for (Throwable cause = throwable; cause != null; cause = cause.getCause()) {
+        Throwable cause = throwable;
+        for (int depth = 0; cause != null && depth < MAX_CAUSE_DEPTH; cause = cause.getCause(), depth++) {
             if (cause instanceof StreamConstraintsException streamConstraintsException) {
                 return streamConstraintsException;
-            }
-            if (cause == cause.getCause()) {
-                break;
             }
         }
         return null;
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/error/JsonProcessingExceptionMapper.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/error/JsonProcessingExceptionMapper.java
@@ -29,12 +29,16 @@ public class JsonProcessingExceptionMapper implements ExceptionMapper<JsonProces
 
     @Override
     public Response toResponse(JsonProcessingException exception) {
-        log.info("Deserialization exception: {}", exception.getMessage());
-
-        // A StreamConstraintsException here is the document- or string-length size guard tripping
-        // mid-parse (OPIK-7334), not ordinary malformed JSON — count it so the guard is observable.
+        // A StreamConstraintsException is the document-/string-length size guard tripping mid-parse
+        // (OPIK-7334) — an EXPECTED client rejection, already surfaced via the ingestion size-guard
+        // metric. Keep it at DEBUG so normal oversized-payload traffic doesn't flood INFO logs.
         if (exception instanceof StreamConstraintsException streamConstraintsException) {
+            log.debug("Ingestion size guard rejected a request", streamConstraintsException);
             sizeGuardMetrics.recordStreamConstraintRejection(streamConstraintsException, uriInfo.get(), requestContext);
+        } else {
+            // Genuine malformed JSON: log the exception itself (not just its message) so the type and
+            // stack trace are available to diagnose the rare, unexpected parse failure.
+            log.info("Deserialization exception", exception);
         }
 
         return Response.status(Response.Status.BAD_REQUEST)
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/error/JsonProcessingExceptionMapper.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/error/JsonProcessingExceptionMapper.java
@@ -54,13 +54,16 @@ public class JsonProcessingExceptionMapper implements ExceptionMapper<JsonProces
             } else {
                 // A structural limit (nesting/number depth), not an oversize -> 400; kept generic since the
                 // message also carries Jackson internals.
-                log.info("Rejecting a request that violates a JSON structural limit", exception);
+                log.info("Rejecting a request that violates a JSON structural limit for workspace '{}'",
+                        ErrorMetricsResolver.workspaceId(requestContext));
                 status = Response.Status.BAD_REQUEST;
                 clientMessage = "Unable to process JSON. The request exceeds an allowed structural limit.";
             }
         } else {
-            log.info("Deserialization exception for workspace {}",
-                    ErrorMetricsResolver.workspaceId(requestContext), exception);
+            // Redacted summary only: the exception message carries the caller's own body content (potential
+            // PII), so log the type + workspace, never the throwable (SKILL.md "Never Log PII").
+            log.info("Deserialization exception for workspace '{}': {}",
+                    ErrorMetricsResolver.workspaceId(requestContext), exception.getClass().getSimpleName());
             status = Response.Status.BAD_REQUEST;
             // Keep the parser detail (the caller's own payload, not internal limits) - a long-standing contract
             // many endpoints assert; do NOT genericize this branch (that broke ~8 test suites once).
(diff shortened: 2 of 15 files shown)


apps/opik-backend/src/main/java/com/comet/opik/infrastructure/JacksonConfig.java
```diff
@@ -25,4 +27,50 @@ public class JacksonConfig {
...
+     * otherwise a legitimate max-size string that already passed {@link #maxStringLength} would be
+     * rejected by the document guard. A non-positive {@code maxDocumentLength} means "unlimited" and
+     * is always valid. Validated at startup (the config is {@code @Valid}), so a misordered override
+     * such as {@code JACKSON_MAX_DOCUMENT_LENGTH < JACKSON_MAX_STRING_LENGTH} fails fast with a clear
+     * message instead of surfacing as a confusing runtime rejection.
+     */
+    @JsonIgnore
+    @AssertTrue(message = "maxDocumentLength must be >= maxStringLength (or <= 0 for unlimited)")
+    public boolean isMaxDocumentLengthValid() {
+        return maxDocumentLength <= 0 || maxDocumentLength >= maxStringLength;
```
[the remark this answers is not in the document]
YarivHashaiComet: Declining this one — after implementing it we found it's both semantically off and breaks a valid config:

- `maxRequestSizeBytes` bounds the **compressed** on-the-wire request (`Content-Length`), while `maxDocumentLength` bounds the **decompressed** document Jackson parses. They measure different axes, so `maxRequestSizeBytes < maxDocumentLength` is a legitimate setup (e.g. cap edge uploads at 50MB while still parsing larger decompressed batches), not a misconfiguration.
- Enforcing it as an `@AssertTrue` breaks `config-test.yml`, which intentionally sets `maxRequestSizeBytes: 50MB` < `maxDocumentLength: 256MB` so `LargePayloadTests` can exercise the 413 path cheaply — the app then fails to boot with a `ConfigurationValidationException`.

We keep the real, same-axis invariant (`maxDocumentLength >= maxStringLength`) and added a code comment explaining why there's intentionally no request↔document cross-check. Reverted in 847b160d.

🤖 _Reply drafted by Claude Code_

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/SpansResourceTest.java
```diff
@@ -4428,33 +4430,48 @@ void createSpan__whenMultipleLargeAttachmentsUnderIndividualLimit__thenSucceed()
...
+            // header before the body is sent. (BUFFERED makes a real Content-Length be sent; without
+            // Expect-continue the JDK connector throws "error writing to server" when the server 413s
+            // mid-upload.)
+            var config = new org.glassfish.jersey.client.ClientConfig();
+            config.connectorProvider(new org.glassfish.jersey.grizzly.connector.GrizzlyConnectorProvider());
+            config.property(org.glassfish.jersey.client.ClientProperties.REQUEST_ENTITY_PROCESSING,
+                    org.glassfish.jersey.client.RequestEntityProcessing.BUFFERED);
+            config.property(org.glassfish.jersey.client.ClientProperties.EXPECT_100_CONTINUE, true);
+
+            try (var bufferedClient = jakarta.ws.rs.client.ClientBuilder.newClient(config)) {
```
thiagohora: Avoid unnecessary qualifier imports

apps/opik-backend/src/main/java/com/comet/opik/api/error/JsonProcessingExceptionMapper.java
```diff
@@ -1,21 +1,75 @@
...
+        //
+        // It rarely arrives as the thrown type: when the overrun trips while binding a bean property
+        // (the common case — a well-formed SpanBatch/TraceBatch whose input/output/metadata is
+        // oversized), Jackson wraps it in a JsonMappingException carrying the reference chain, so a
+        // plain `instanceof` check misses it and the metric under-counts exactly the real-world case
+        // the dashboard monitors. Walk the cause chain to find the underlying constraint.
+        StreamConstraintsException streamConstraintsException = findStreamConstraint(exception);
+        if (streamConstraintsException != null) {
+            log.debug("Ingestion size guard rejected a request", exception);
+            sizeGuardMetrics.recordStreamConstraintRejection(streamConstraintsException, uriInfo.get(), requestContext);
```
thiagohora: This PR fixes an incorrect HTTP status code returned when Jackson's size guard trips during request deserialization.

Previously, `JsonProcessingExceptionMapper` always returned `400 BAD_REQUEST`, even when the underlying cause was a `StreamConstraintsException` — which is a payload-too-large rejection, not a malformed JSON error. 

The correct status in that case is `413 REQUEST_ENTITY_TOO_LARGE`.

thiagohora (changes requested):

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/JacksonConfig.java
```diff
@@ -25,4 +27,57 @@ public class JacksonConfig {
...
+     * values. Enforced mid-parse on both the HTTP and internal ObjectMappers, aborting with a 4xx
+     * (via JsonProcessingExceptionMapper) before a multi-GB node tree is materialized.
+     *
+     * Must stay above {@link #maxStringLength} so a single legitimate max-size string still parses.
+     *
+     * Default 512MB is the self-hosted value; the Comet cloud deployment overrides it to 256MB via
+     * {@code JACKSON_MAX_DOCUMENT_LENGTH}.
+     */
+    @JsonProperty
+    @Min(value = 1048576, message = "maxDocumentLength must be at least 1MB") private long maxDocumentLength = 536_870_912L; // 512MB
```
thiagohora: **[medium] @Min(1MB) on maxDocumentLength contradicts the documented '<=0 = unlimited' contract**

The field carries @Min(value = 1048576) yet the class documents and codes a 'non-positive means unlimited' semantics that is mutually inconsistent with it: the @AssertTrue isMaxDocumentLengthValid() short-circuits on `maxDocumentLength <= 0` ('always valid'), its javadoc says 'A non-positive maxDocumentLength means "unlimited" and is always valid. Validated at startup', OpikApplication.java:141 comments 'maxDocumentLength <= 0 is normalized to "unlimited" by the builder', JsonUtils treats `<= 0` as unlimited, and JacksonConfigTest asserts setMaxDocumentLength(-1L) is valid. But OpikConfiguration.jacksonConfig is @Valid @NotNull, so Hibernate Validator enforces @Min at startup — any value < 1MB (including 0 or a negative 'unlimited' override) fails startup with 'maxDocumentLength must be at least 1MB'. An operator who sets JACKSON_MAX_DOCUMENT_LENGTH=0 to disable the cap (exactly what the docs invite) gets a confusing startup failure, and the `maxDocumentLength <= 0` branch of the validator plus the JsonUtils unlimited path are dead in the deployed configuration path. The bound and the 'unlimited' escape hatch cannot both be true.

💡 Decide one contract and make the annotations, validator, docs, and tests agree. Either (a) genuinely support unlimited: drop @Min (or replace with a custom @AssertTrue that allows `<= 0` OR `>= 1MB`) so 0/negative overrides pass startup; or (b) drop the 'unlimited' semantics: remove the `maxDocumentLength <= 0` branch from isMaxDocumentLengthValid(), the '<=0 means unlimited' javadoc/comments, and the setMaxDocumentLength(-1L) test assertion, keeping @Min as the real floor.

_rule: `backend-correctness-apply-complete-mutually-consistent-bean-validation-constrain`_

thiagohora: **[medium] Externally-configured maxDocumentLength / maxRequestSizeBytes have @Min but no @Max**

Both new fields are populated from untrusted environment overrides (config.yml: maxDocumentLength = ${JACKSON_MAX_DOCUMENT_LENGTH:-...}, maxRequestSizeBytes = ${JACKSON_MAX_REQUEST_SIZE_BYTES:-...}) and directly govern resource-sensitive parsing/memory behavior (the whole-document byte budget before a multi-GB node tree is built, and the accepted request body size). They carry only a lower bound (@Min 1MB) with no upper bound, so a misconfigured override can set them arbitrarily large, silently defeating the very memory-exhaustion protection this class exists to provide. The rule requires such bounded numeric config fields to carry both @Min and @Max.

💡 Add an @Max to maxDocumentLength and maxRequestSizeBytes (e.g. a sane multi-GB ceiling) so an out-of-range override fails fast at startup instead of disabling the guard. Consider the same for the pre-existing maxStringLength for consistency.

_rule: `backend-security-cap-externally-configured-numeric-bounds-with-max-alongside-`_

apps/opik-backend/src/main/java/com/comet/opik/utils/JsonUtils.java
```diff
@@ -87,13 +91,29 @@ private static ObjectMapper createConfiguredMapper(int maxStringLength) {
...
+     * @param maxDocumentLength Maximum whole-document length in bytes ({@code <= 0} is normalized to
+     *                          "unlimited" by the builder, matching the static-init default)
+     */
+    public static void applyStreamReadConstraints(@NonNull ObjectMapper mapper, int maxStringLength,
+            long maxDocumentLength) {
         StreamReadConstraints readConstraints = StreamReadConstraints.builder()
                 .maxStringLength(maxStringLength)
+                .maxDocumentLength(maxDocumentLength)
                 .build();
         mapper.getFactory().setStreamReadConstraints(readConstraints);
```
thiagohora, line 109: **[medium] Internal maxDocumentLength cap is a no-op for in-memory String/byte[] reads**

The Javadoc on applyStreamReadConstraints (lines 101-104) and the class contract assert the internal MAPPER is 'configured identically' to the HTTP mapper and enforces the whole-document cap; JacksonConfig likewise states it is 'Enforced mid-parse on both the HTTP and internal ObjectMappers'. But Jackson only validates maxDocumentLength inside the parser's buffer-refill path (_loadMore), which fires for stream/reader inputs, not for content Jackson loads in a single in-memory buffer. The project's own new test (JsonUtilsDocumentLengthTest) documents exactly this: 'A fully-in-memory byte[]/String fits the initial buffer, never refills, and would bypass the check' and therefore only exercises InputStream inputs. Most JsonUtils read methods here are in-memory: getJsonNodeFromString(String) (line 187), readValue(String,...) (lines 226/234), readValue(byte[],...) (line 250), readCollectionValue(String,...) (line 264). For those paths the document cap does not fire, so the internal mapper does NOT get the defense-in-depth the change advertises (e.g. reading a large stored span/trace JSON back from ClickHouse as a String/byte[] can still materialize a multi-GB node tree). The HTTP boundary (stream-based request body) is genuinely protected; the internal guarantee is overstated.

💡 Either (a) soften the docblock/JacksonConfig wording to state the internal document cap only applies to stream-based reads, so the guarantee is not misleading; or (b) if internal enforcement is actually required, route the in-memory read paths through a length pre-check (compare content byte length against maxDocumentLength before readTree/readValue) rather than relying on StreamReadConstraints, which Jackson does not evaluate for single-buffer inputs.

_general finding_

apps/opik-backend/src/test/java/com/comet/opik/utils/JsonUtilsDocumentLengthTest.java
```diff
@@ -0,0 +1,81 @@
...
+        return new ByteArrayInputStream(sb.append('}').toString().getBytes(StandardCharsets.UTF_8));
+    }
+
+    @Test
+    @DisplayName("a document over maxDocumentLength is rejected mid-parse (StreamConstraintsException -> 400)")
+    void overMaxDocumentLength_rejected() {
+        ObjectMapper mapper = JsonUtils.createConfiguredMapper(STRING_LEN, 1024L); // 1KB
+
+        // ~200KB of tiny values, streamed - forces buffer refills past the 1KB limit
+        assertThatThrownBy(() -> mapper.readTree(manySmallValues(200 * 1024)))
```
thiagohora, line 47: **[medium] readTree(...) does not mirror the real ingestion path the comment/DisplayName claim**

The header comment (lines 21-25) asserts these tests "parse from an InputStream to mirror the real ingestion path" and the @DisplayName on line 40 claims "StreamConstraintsException -> 400". Both are false for the case that matters. The real ingestion path deserializes the body into a typed bean (SpanBatch/TraceBatch) via readValue, and the sibling production commit a4ff8bc8a6 (OPIK-7333) documents that when the size guard trips during bean binding, Jackson's BeanDeserializer.wrapAndThrow re-wraps the StreamConstraintsException inside a JsonMappingException carrying the reference chain -- 'synthetic bodies incremented the metric but real SDK traffic did not'. mapper.readTree(...) here does untyped tree parsing (no bean binding), so the StreamConstraintsException propagates raw and isInstanceOf(StreamConstraintsException.class) passes -- exercising exactly the untyped/synthetic path that behaves DIFFERENTLY from prod, not the wrapped-in-JsonMappingException path that the whole feature (and its metric fix) is about. The test therefore gives false confidence and would not have caught the under-counting bug that commit had to fix.

💡 Either (a) drop the 'mirrors the real ingestion path' / '-> 400' claims and scope this file's DisplayName/comment strictly to 'createConfiguredMapper wires StreamReadConstraints correctly', or (b) add a case that deserializes into a real typed bean (e.g. mapper.readValue(stream, SpanBatch.class) or the actual batch DTO) and assert the thrown JsonMappingException has a StreamConstraintsException in its cause chain -- i.e. cover the wrapped path that production actually hits.

_rule: `backend-testing-test-backend-behavior-black-box-through-public-apis-typed-cl`_

apps/opik-backend/src/test/resources/config-test.yml
```diff
@@ -292,6 +292,24 @@ jacksonConfig:
...
+  # they exercise the 400/413 mechanism, while JacksonConfigTest verifies the real shipped defaults.
+  # Default: 536870912 (512MB) - the self-hosted default (higher, to avoid surprising customers);
+  #   the Comet cloud deployment overrides JACKSON_MAX_DOCUMENT_LENGTH to 268435456 (256MB).
+  # Description: Maximum size of an entire JSON document (the whole decompressed batch). Unlike
+  #   maxStringLength (a single string value), this caps the whole document, so it catches an
+  #   oversized batch built from many small values. Aborts mid-parse with a 4xx before a multi-GB
+  #   node tree is materialized. Must stay above maxStringLength so a single max-size string parses.
+  maxDocumentLength: 268435456
+  # Default: 536870912 (512MB) - self-hosted default; cloud overrides JACKSON_MAX_REQUEST_SIZE_BYTES
+  #   to 268435456 (256MB). Kept equal to maxDocumentLength so the request filter never rejects a
```
thiagohora: **[medium] maxRequestSizeBytes comment says 'Kept equal to maxDocumentLength' but value is deliberately lower**

The docblock for maxRequestSizeBytes states 'Kept equal to maxDocumentLength so the request filter never rejects a request the document guard would otherwise accept.' but the configured value is 52428800 (50MB) while maxDocumentLength is 268435456 (256MB) — they are NOT equal. This comment is copy-pasted verbatim from the production Javadoc (JacksonConfig.java lines 56-58) without adjusting for the test scenario. It also contradicts this file's own top comment (lines 295-298), which correctly documents the 50MB request cap as calibrated to trip on the '51MB request' LargePayloadTests payload. Worse, JacksonConfig.java lines 77-82 explicitly note the test config relies on '50MB request < 256MB document' to exercise the 413 path cheaply — so a maintainer trusting this comment could 'fix' maxRequestSizeBytes up to 256MB to match maxDocumentLength and silently break the 413 test (a 51MB request would no longer exceed the cap).

💡 Replace the inherited 'Kept equal to maxDocumentLength...' sentence with a test-specific note, e.g.: 'Set to 50MB — intentionally BELOW maxDocumentLength (256MB) so the ~51MB LargePayloadTests request trips the 413 pre-parse guard cheaply. See JacksonConfig.java note on why request size and document size are independent axes.'

_rule: `backend-docs-keep-config-yml-default-comments-and-env-var-names-in-sync-w`_

apps/opik-backend/src/main/java/com/comet/opik/api/error/JsonProcessingExceptionMapper.java
```diff
@@ -1,21 +1,75 @@
...
+        } else {
+            // Genuine malformed JSON: log the exception itself (not just its message) so the type and
+            // stack trace are available to diagnose the rare, unexpected parse failure.
+            log.info("Deserialization exception", exception);
```
thiagohora: **[low] Genuine-malformed-JSON log omits workspace correlation id**

The size-guard branch attaches workspace_id/workspace_name via the metric, but the genuine-malformed branch logs `log.info("Deserialization exception", exception)` with no correlation identifier. `requestContext` is already injected here, so the workspace that sent the malformed payload could be included to help attribute the offending client. Note: sibling mappers (OpikGenericExceptionMapper) also omit this, so the team may tolerate it — flagging as defensive/low.

💡 Optionally include the resolved workspace id in the INFO message, e.g. `log.info("Deserialization exception for workspace {}", ErrorMetricsResolver.workspaceId(requestContext), exception)`, guarding against an unresolved context.

_rule: `backend-observability-include-correlation-identifiers-workspaceid-projectid-entity`_

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/RequestSizeLimitFilter.java
```diff
@@ -0,0 +1,86 @@
...
+            return;
+        }
+
+        if (contentLength < 0) {
+            abort(containerRequestContext, Response.Status.BAD_REQUEST, "Invalid Content-Length header");
+            return;
+        }
+
+        long maxRequestSizeBytes = jacksonConfig.getMaxRequestSizeBytes();
+        if (contentLength > maxRequestSizeBytes) {
```
thiagohora, line 65: **[low] Javadoc frames the filter as ingestion-specific but it applies to every endpoint**

The class javadoc describes the guard as fast-failing 'a large ingestion batch', but the filter is a global JAX-RS provider with no path scoping (unlike AuthFilter, which matches /v1/private/*). It therefore also caps non-ingestion request bodies - notably AttachmentResource /upload (which buffers the whole body via IOUtils.toByteArray) and dataset import uploads - which now hard-fail with 413 above maxRequestSizeBytes (256MB cloud). This is a defensible/protective behavior, but the ingestion-only framing understates the scope and the new cap on those upload paths is not called out or test-covered.

💡 Reword the javadoc to state the filter applies to all requests carrying a Content-Length (not just ingestion), and confirm the cap is intended for the attachment/dataset upload endpoints (add an integration assertion if those paths must stay under the cap).

_rule: `backend-style-maintainability-document-non-obvious-contracts-with-javadoc-using-us-english`_

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/metrics/IngestionSizeGuardMetrics.java
```diff
@@ -0,0 +1,126 @@
...
+     * StreamConstraintsException}; the concrete limit is distinguishable only from the message, which
+     * references the tripped {@code StreamReadConstraints} accessor. An unrecognized message falls
+     * back to {@link #GUARD_STREAM_CONSTRAINT} rather than guessing.
+     */
+    static String classifyStreamConstraint(StreamConstraintsException exception) {
+        String message = exception.getMessage();
+        if (message == null) {
+            return GUARD_STREAM_CONSTRAINT;
+        }
+        if (message.contains("getMaxDocumentLength") || message.contains("Document length")) {
```
thiagohora, line 85: **[low] guard classification depends on brittle Jackson message substrings**

classifyStreamConstraint distinguishes document_length vs string_length purely by substring-matching Jackson's exception text (getMaxDocumentLength / Document length / getMaxStringLength / String value length). These strings are Jackson internals, not a stable contract; a Jackson upgrade that rewords the StreamConstraintsException message would silently collapse both real cases into the generic stream_constraint bucket, degrading the dashboard breakdown without any error. The companion unit test constructs its own exceptions with hardcoded current-version message text, so it would keep passing and would NOT catch such a wording change.

💡 Prefer a version-stable signal where possible (e.g. inspect a structured field on the exception if available), and add an integration-level test that drives real Jackson parsing past the configured maxDocumentLength/maxStringLength so a future Jackson message change breaks the test rather than the metric. At minimum, document that the string match must be re-validated on Jackson upgrades.

_general finding_

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/SpansResourceTest.java
```diff
@@ -4428,33 +4430,48 @@ void createSpan__whenMultipleLargeAttachmentsUnderIndividualLimit__thenSucceed()
...
+                try (Response response = bufferedClient.target("%s/v1/private/spans".formatted(baseURI))
+                        .request()
+                        .header(HttpHeaders.AUTHORIZATION, API_KEY)
+                        .header(WORKSPACE_HEADER, TEST_WORKSPACE)
+                        .post(Entity.json(oversizedJson))) {
+
+                    assertThat(response.getStatus()).isEqualTo(HttpStatus.SC_REQUEST_TOO_LONG); // 413
+                    assertThat(response.readEntity(String.class))
```
thiagohora: **[low] 413 test asserts raw String body instead of the typed ErrorMessage**

The 413 test reads the body as response.readEntity(String.class) and only substring-matches "exceeds the maximum allowed size". The filter (RequestSizeLimitFilter.abort) returns a typed io.dropwizard.jersey.errors.ErrorMessage with code=413 and a specific message, and the sibling 400 tests already read that typed entity. Reading a raw String neither asserts the response code field nor guarantees the payload is the expected ErrorMessage shape, so a regression that changed the body structure while keeping the substring could pass.

💡 Read io.dropwizard.jersey.errors.ErrorMessage (as the 400 tests do), assert getCode()==413, and assert getMessage() contains "exceeds the maximum allowed size", for consistency and a stronger body assertion.

_rule: `backend-testing-assert-whole-objects-full-responses-with-assertj-not-individ`_

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/SpansResourceTest.java
```diff
@@ -4428,33 +4430,48 @@ void createSpan__whenMultipleLargeAttachmentsUnderIndividualLimit__thenSucceed()
...
-                        assertThat(inputString).containsPattern("\\[input-attachment-\\d+-\\d+\\.(jpg|jpeg)\\]");
-                        assertThat(outputString).containsPattern("\\[output-attachment-\\d+-\\d+\\.png\\]");
-                    });
+        @Test
+        @DisplayName("Create span request whose Content-Length exceeds maxRequestSizeBytes - should return 413")
+        void createSpan__whenRequestContentLengthExceedsLimit__thenReject413() {
+            // The default test client (GrizzlyConnectorProvider) streams chunked and never sends a
+            // Content-Length, so it bypasses RequestSizeLimitFilter. A BUFFERED client sets
+            // Content-Length, exercising the pre-parse 413 guard (OPIK-7333) end-to-end.
+            String oversizedJson = "\"" + "a".repeat(51 * 1024 * 1024) + "\""; // 51MB > 50MB limit, valid JSON
```
thiagohora: **[low] Document-length 400 test silently depends on the default client using chunked transfer**

createSpan__whenTotalDocumentExceedsMaxDocumentLength__thenReject sends a ~280MB payload via the default spanResourceClient and expects 400 ("Document length ... exceeds the maximum allowed"). But config-test.yml sets maxRequestSizeBytes=50MB, so if the request carried a Content-Length header the RequestSizeLimitFilter would reject it with 413 ("Request body exceeds the maximum allowed size") long before the maxDocumentLength parse guard runs. The test only reaches the 400 path because the default Grizzly client streams chunked without a Content-Length (a fact documented only in the separate 413 test, not here). If the default client's transport behavior changes to send Content-Length, this test flips to 413 and fails for a reason unrelated to the document guard it means to verify.

💡 Add a comment on this test noting the reliance on chunked transfer (no Content-Length) so the request-size 413 filter is bypassed, or assert the discriminating detail (that it is the document-length guard, not the request-size guard, that fired) to make the dependency explicit and less fragile.

_general finding_

apps/opik-backend/src/test/resources/config-test.yml
```diff
@@ -292,6 +292,24 @@ jacksonConfig:
   #   This configuration is used by both HTTP layer and internal JSON processing for consistency,
   #   preventing memory exhaustion from extremely large base64-encoded attachments before they can be stripped
   maxStringLength: 262144000
+  # Test-only caps, intentionally LOWER than the shipped defaults (512MB self-hosted / 256MB cloud)
+  # so the guard tests trip at modest payload sizes instead of generating >512MB bodies. These
+  # values are calibrated to the LargePayloadTests payloads (253MB string, 280MB doc, 51MB request);
+  # they exercise the 400/413 mechanism, while JacksonConfigTest verifies the real shipped defaults.
+  # Default: 536870912 (512MB) - the self-hosted default (higher, to avoid surprising customers);
+  #   the Comet cloud deployment overrides JACKSON_MAX_DOCUMENT_LENGTH to 268435456 (256MB).
```
thiagohora: **[low] maxDocumentLength (256MB) equals the cloud default, not 'LOWER than the shipped defaults'**

The intro comment (lines 295-298) says these caps are 'intentionally LOWER than the shipped defaults (512MB self-hosted / 256MB cloud)', but the configured maxDocumentLength is 268435456 (256MB) — equal to the cloud default it lists, not lower than it. Only the self-hosted 512MB default is actually higher. Minor factual imprecision that can confuse someone reasoning about the test caps.

💡 Clarify that maxDocumentLength is set to the cloud default (256MB), below only the 512MB self-hosted default; or reword to 'at or below the shipped defaults'.

_rule: `backend-docs-keep-config-yml-default-comments-and-env-var-names-in-sync-w`_

thiagohora (commented):

11 findings — 0 high · 5 medium · 6 low. Suppressed by team conventions: see `suppressed.md`.

CometActions:

### Access Information
- **URL:** https://pr-7483.dev.comet.com
- **Namespace:** pr-7483
- **Version:** 2.1.28-7483-merge-2638
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7483%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

apps/opik-backend/src/main/java/com/comet/opik/api/error/JsonProcessingExceptionMapper.java
```diff
@@ -1,21 +1,75 @@
...
+    private static StreamConstraintsException findStreamConstraint(Throwable throwable) {
+        for (Throwable cause = throwable; cause != null; cause = cause.getCause()) {
+            if (cause instanceof StreamConstraintsException streamConstraintsException) {
+                return streamConstraintsException;
+            }
+            if (cause == cause.getCause()) {
+                break;
+            }
+        }
+        return null;
```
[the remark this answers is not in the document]
YarivHashaiComet: Good catch — confirmed and fixed. The `cause == cause.getCause()` guard only caught a **direct** self-cause; a 2-node cycle `A → B → A` (constructible, since `Throwable.initCause` rejects only the direct self, not a longer cycle) did spin the walk forever and pin the request thread.

Reproduced first with a timeout-guarded test (`doesNotLoopOnMultiNodeCauseCycle`) — on the old code it times out at 2s with the stack pinned to `findStreamConstraint`. Fix replaces the self-cause check with a hard depth cap (`MAX_CAUSE_DEPTH = 50`, far beyond any real chain, no per-call allocation), which also subsumes the old self-cause case. Fixed in ceac8d03c.

CometActions:

### Access Information
- **URL:** https://pr-7483.dev.comet.com
- **Namespace:** pr-7483
- **Version:** 2.1.28-7483-merge-2639
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7483%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

apps/opik-backend/config.yml
```diff
@@ -339,6 +339,20 @@ jacksonConfig:
   #   This configuration is used by both HTTP layer and internal JSON processing for consistency,
   #   preventing memory exhaustion from extremely large base64-encoded attachments before they can be stripped
   maxStringLength: ${JACKSON_MAX_STRING_LENGTH:-104857600}
+  # Default: 536870912 (512MB) - the self-hosted default (higher, to avoid surprising customers);
+  #   the Comet cloud deployment overrides JACKSON_MAX_DOCUMENT_LENGTH to 268435456 (256MB).
+  # Description: Maximum size of an entire JSON document (the whole decompressed batch). Unlike
+  #   maxStringLength (a single string value), this caps the whole document, so it catches an
+  #   oversized batch built from many small values. Aborts mid-parse with a 4xx before a multi-GB
+  #   node tree is materialized. Must stay above maxStringLength so a single max-size string parses.
+  maxDocumentLength: ${JACKSON_MAX_DOCUMENT_LENGTH:-536870912}
```
thiagohora, line 346: **[medium] maxDocumentLength/maxRequestSizeBytes defaults duplicated between YAML and Java field initializers**

The new keys default to 536870912 (512MB) in this YAML (lines 348 and 355), and JacksonConfig.java also hard-codes the same values in its field initializers (`private long maxDocumentLength = 536_870_912L;` and `private long maxRequestSizeBytes = 536_870_912L;`). Since these fields are always externally supplied via config.yml (bound with @Config), the Java-side literals are two competing sources of truth for the same default. A future change to one (e.g. bumping the self-hosted default in YAML) will silently drift from the other, and the Java literal only surfaces if the key is ever removed from YAML — making the mismatch invisible until then. Contrast maxStringLength, which correctly uses a symbolic Java default (StreamReadConstraints.DEFAULT_MAX_STRING_LEN) rather than re-inlining the YAML literal.

💡 Make config.yml the sole authoritative default: drop the hard-coded 536_870_912L initializers in JacksonConfig.java (leave the fields undefaulted / relying on Bean Validation to fail fast if the key is absent), or, if a Java fallback is required, derive both from a single shared constant so they cannot drift.

_rule: `backend-architecture-do-not-duplicate-config-defaults-between-java-field-initiali`_

apps/opik-backend/src/main/java/com/comet/opik/api/error/JsonProcessingExceptionMapper.java
```diff
@@ -1,21 +1,72 @@
...
+            // many endpoints assert; do NOT genericize this branch (that broke ~8 test suites once).
+            clientMessage = "Unable to process JSON. " + exception.getMessage();
+        }
 
-        return Response.status(Response.Status.BAD_REQUEST)
-                .entity(new ErrorMessage(Response.Status.BAD_REQUEST.getStatusCode(),
-                        "Unable to process JSON. " + exception.getMessage()))
+        return Response.status(status)
+                .entity(new ErrorMessage(status.getStatusCode(), clientMessage))
                 .build();
```
thiagohora, line 79: **[medium] Error response doesn't force application/json content type**

The response is built without `.type(MediaType.APPLICATION_JSON_TYPE)`, so the ErrorMessage entity is serialized via content negotiation against the request's Accept header. The sibling mapper in this same package, InvalidUUIDExceptionMapper, explicitly forces JSON with the comment: "ingestion endpoints negotiate other content types (e.g. the OTel endpoint uses protobuf), which have no writer for the error entity — without this the 400 fails to serialize and surfaces as a 500". This mapper handles the same ingestion paths (OPIK-7334) yet omits the guard, so a 413/400 to a caller with a non-JSON Accept header can fail to serialize and surface as a 500 — the opposite of the intended payload-too-large signal. RequestSizeLimitFilter.abort() also sets the JSON content-type header for the parallel 413 path.

💡 Add `.type(MediaType.APPLICATION_JSON_TYPE)` to the Response builder, matching InvalidUUIDExceptionMapper and RequestSizeLimitFilter.

_rule: `backend-api-contract-return-a-consistent-shared-errormessage-json-body-on-every-e`_

apps/opik-backend/src/main/java/com/comet/opik/api/error/JsonProcessingExceptionMapper.java
```diff
@@ -1,21 +1,72 @@
...
+
+    /** The {@link StreamConstraintsException} this throwable is or wraps, or {@code null}. */
+    private static StreamConstraintsException findStreamConstraint(Throwable throwable) {
+        Throwable cause = throwable;
+        for (int depth = 0; cause != null && depth < MAX_CAUSE_DEPTH; cause = cause.getCause(), depth++) {
+            if (cause instanceof StreamConstraintsException streamConstraintsException) {
+                return streamConstraintsException;
+            }
+        }
+        return null;
```
thiagohora: **[low] findStreamConstraint reimplements commons-lang3 ExceptionUtils.throwableOfType**

The hand-rolled bounded cause-chain walk (plus the MAX_CAUSE_DEPTH constant and its cycle-defense rationale) duplicates org.apache.commons.lang3.exception.ExceptionUtils.throwableOfType(throwable, StreamConstraintsException.class), which walks the cause chain and returns the first assignable throwable. ExceptionUtils is already a dependency and is used for the same purpose in the sibling class ErrorMetricsResolver (getRootCause) in this very package. ExceptionUtils.getThrowableList is inherently cycle-safe (it stops when a throwable is already in the visited list), so it covers both the self-cause and the A->B->A cycle cases the custom code and MAX_CAUSE_DEPTH exist to defend against — making the depth cap and the custom method redundant.

💡 Replace findStreamConstraint and MAX_CAUSE_DEPTH with `ExceptionUtils.throwableOfType(exception, StreamConstraintsException.class)`; the cycle-safety tests still pass since getThrowableList detects cycles.

_rule: `backend-style-maintainability-reuse-existing-constants-utility-methods-and-helpers-instead`_

apps/opik-backend/src/main/java/com/comet/opik/api/error/JsonProcessingExceptionMapper.java
```diff
@@ -1,21 +1,72 @@
...
+        // A StreamConstraintsException (often wrapped by Jackson during bean binding, hence the cause walk)
+        // is the size guard tripping mid-parse -> 413; anything else is malformed JSON -> 400.
+        StreamConstraintsException streamConstraintsException = findStreamConstraint(exception);
+        Response.Status status;
+        String clientMessage;
+        if (streamConstraintsException != null) {
+            log.debug("Ingestion size guard rejected a request", exception); // expected; already on the metric
+            sizeGuardMetrics.recordStreamConstraintRejection(streamConstraintsException, uriInfo.get(), requestContext);
+            status = Response.Status.REQUEST_ENTITY_TOO_LARGE;
+            // Redacted: the size-guard message exposes internal limits/Jackson internals (detail is in the log).
```
thiagohora: **[low] Non-size StreamConstraintsException also mapped to 413 'payload exceeds maximum size'**

Every StreamConstraintsException found in the chain is returned as 413 with the message "Request payload exceeds the maximum allowed size.". But Jackson's StreamReadConstraints also raises StreamConstraintsException for maxNestingDepth and maxNumberLength violations, which are not payload-size problems and are better signalled as 400. IngestionSizeGuardMetrics.classifyStreamConstraint already acknowledges this by bucketing anything it can't classify into GUARD_STREAM_CONSTRAINT, yet the mapper still forces 413 regardless of classification. A small-but-deeply-nested JSON body would get a misleading 413 'too large' response. Edge/adversarial input, so low severity / defensive hardening.

💡 Consider using the classified guard (document_length/string_length -> 413; unclassified stream_constraint -> 400 with a generic 'invalid JSON structure' message) instead of unconditionally returning 413 for any StreamConstraintsException.

_general finding_

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/JacksonConfig.java
```diff
@@ -16,6 +19,10 @@
 @NoArgsConstructor
 public class JacksonConfig {
 
+    // Upper bound for the byte caps below: beyond ~2GB a single Java String/array hits its 2^31 limit,
+    // so a larger cap can't protect the heap and is almost always a units typo.
+    private static final long MAX_CONFIGURABLE_BYTES = 2_147_483_648L; // 2GB
```
thiagohora: **[low] MAX_CONFIGURABLE_BYTES is 2^31, one past the real String/array limit (Integer.MAX_VALUE)**

The comment justifies the ceiling by 'a single Java String/array hits its 2^31 limit', but the maximum String/array length is Integer.MAX_VALUE = 2^31 - 1 = 2147483647, whereas the constant is 2_147_483_648L = 2^31 (Integer.MAX_VALUE + 1). Both the @Max on maxRequestSizeBytes and the isMaxDocumentLengthValid upper bound therefore accept a value exactly 1 byte beyond what a single String/byte[] can represent, contradicting the stated rationale. Purely defensive precision at the 2GB boundary, hence low.

💡 Use Integer.MAX_VALUE (2_147_483_647L) as the ceiling to match the comment's 2^31-1 limit exactly, or reword the comment to reflect the chosen value.

_rule: `backend-correctness-apply-complete-mutually-consistent-bean-validation-constrain`_

thiagohora (commented):

9 findings — 0 high · 3 medium · 6 low. Suppressed by team conventions: see `suppressed.md`.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/RequestSizeLimitFilter.java
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Rejects an oversized request with 413 by its {@code Content-Length}, before the body is read. Global
+ * (no path scoping) by design, so it caps every request carrying a Content-Length; a chunked request has
+ * none, passes here, and is still bounded by {@code maxDocumentLength} during parsing.
+ */
+@Slf4j
+public class RequestSizeLimitFilter implements ContainerRequestFilter {
+
+    private final JacksonConfig jacksonConfig;
```
thiagohora, line 28: **[low] No @Priority though javadoc promises the guard runs early**

The class javadoc states the guard rejects 'before the body is read' and is a broad memory guard, but the filter declares no @Priority, so it defaults to Priorities.USER (5000) — the same bucket as AuthFilter, which also has no priority. Ordering between two same-priority ContainerRequestFilters is undefined, so an oversized request may be authenticated first and surface as 401 rather than the intended 413. Body buffering is unaffected (entity is not read until the resource method), so impact is limited to status attribution and wasted auth work; hence low.

💡 If the guard is meant to run ahead of authentication, annotate it with an explicit low @Priority (e.g. Priorities.HEADER_DECORATOR or below AuthFilter) to make the ordering deterministic; otherwise soften the 'before the body is read / before auth' wording in the javadoc to match the undefined ordering.

_general finding_

apps/opik-backend/src/main/java/com/comet/opik/utils/JsonUtils.java
```diff
@@ -44,32 +45,34 @@ public class JsonUtils {
...
+     * @param maxDocumentLength Maximum whole-document length in bytes ({@code <= 0} means unlimited)
      */
-    public static synchronized void configure(int maxStringLength) {
-        MAPPER = createConfiguredMapper(maxStringLength);
-        log.info("JsonUtils configured with maxStringLength: '{}' bytes ('{}'MB)",
-                maxStringLength, maxStringLength / 1024 / 1024);
+    public static synchronized void configure(int maxStringLength, long maxDocumentLength) {
+        MAPPER = createConfiguredMapper(maxStringLength, maxDocumentLength);
+        log.info("JsonUtils configured with maxStringLength: '{}' bytes ('{}'MB), maxDocumentLength: '{}' bytes",
+                maxStringLength, maxStringLength / 1024 / 1024, maxDocumentLength);
```
thiagohora, line 63: **[low] configure() logs unlimited maxDocumentLength as "-1 bytes"**

When maxDocumentLength <= 0 (the documented "unlimited" sentinel, e.g. -1 as used in the static initializer, or 0), configure() logs `maxDocumentLength: '-1' bytes` — a misleading value that contradicts the code's own semantics. The static initializer at line 50 correctly renders the same unlimited state as `maxDocumentLength: unlimited`, so the two log statements disagree for identical configuration. In production the default is a positive 512MB so this is only hit when unlimited is explicitly configured, hence low severity.

💡 Branch the log value like the static block does, e.g. log `maxDocumentLength <= 0 ? "unlimited" : maxDocumentLength + " bytes"`, so the configured-unlimited case reads consistently across both log sites.

_rule: `backend-observability-ensure-log-statements-and-inline-comments-do-not-contradict-`_

apps/opik-backend/src/test/java/com/comet/opik/api/error/JsonProcessingExceptionMapperTest.java
```diff
@@ -0,0 +1,157 @@
...
+        assertPayloadTooLarge(response);
+        // Leak guard: raw parser detail (sizes, Jackson internals, the payload reference chain) must not
+        // reach the client.
+        assertThat(((ErrorMessage) response.getEntity()).getMessage())
+                .doesNotContain("12590880", "reference chain", "StreamReadConstraints");
+    }
+
+    @Test
+    @DisplayName("records the metric and returns 413 for a raw (unwrapped) StreamConstraintsException")
+    void recordsMetricForRawStreamConstraint() {
```
thiagohora, line 91: **[low] Leak-guard doesNotContain assertion is vacuous after the exact-equals check**

The doesNotContain("12590880", "reference chain", "StreamReadConstraints") assertion (lines 88-89) runs after assertPayloadTooLarge already asserted the message isEqualTo the fixed constant "Request payload exceeds the maximum allowed size." (line 50-51). Once the exact-equals passes, the message can only be that constant, so doesNotContain is guaranteed true regardless of the mapper's redaction logic — it can never fail and therefore does not actually guard against a leak.

💡 Make the leak-guard discriminating by asserting against the raw source detail rather than a hardcoded constant: build the response message and assert it doesNotContain the actual raw parser text passed into the wrapped exception (DOC_LENGTH_MESSAGE and the reference-chain substring), so the assertion would fail if the redaction branch ever echoed the parser detail. Or drop it and rely on the exact-equals check plus a comment.

_rule: `backend-testing-make-test-assertions-precise-unconditional-and-discriminatin`_

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/SpansResourceTest.java
```diff
@@ -4387,24 +4392,26 @@ void createSpan__whenSingleAttachmentExceedsLimit__thenReject() throws Exception
...
         }
 
         @Test
-        @DisplayName("Create span with multiple large attachments under individual limit - should succeed")
-        void createSpan__whenMultipleLargeAttachmentsUnderIndividualLimit__thenSucceed() throws Exception {
-            // Given: Create THREE 70MB attachments (each ~93MB as base64)
-            // Total payload: ~280MB, but each individual string is under 250MB limit
-            // This verifies the limit is per-string, not per total payload
+        @DisplayName("Create span whose total document exceeds maxDocumentLength - should return 413 Request Entity Too Large")
+        void createSpan__whenTotalDocumentExceedsMaxDocumentLength__thenReject() throws Exception {
```
thiagohora, line 4409: **[medium] Document-cap tests miss the many-small-values shape (each value under maxStringLength, sum over maxDocumentLength)**

The unique job of maxDocumentLength (vs maxStringLength) is to catch a document whose total size exceeds the cap even though every individual value is under the per-string limit. This end-to-end test proves that only with THREE ~93MB base64 strings (~280MB) — i.e. a few big strings, each already close to the per-string cap. Two structurally different shapes that specifically exercise the document cap are not covered end-to-end, and they trip Jackson at different points (so the wrapped-JsonMappingException path the size-guard metric unwraps differs):

(A) One entity with MANY SMALL PROPERTIES whose sum exceeds maxDocumentLength — currently only in the unit test JsonUtilsDocumentLengthTest.typedBeanBinding_wrapsStreamConstraint (span.output with many k:i), never through the real HTTP ingestion path.

(B) MANY SMALL ENTITIES (a batch of many small spans) whose total exceeds maxDocumentLength — not covered anywhere (unit or integration). This is the batch-array iteration path, distinct from (A) property binding.

Without these, a regression that only enforced the per-string cap (not the whole-document cap) on the many-small-values case would pass. Note also that all three 413 outcomes now share the generic message "Request payload exceeds the maximum allowed size.", so these tests can only distinguish the document guard from the request-size filter via the transport (chunked) + the ingestion_size_guard metric guard label, not the response body.

💡 Add two integration tests on the batch/ingestion path: (A) a single span whose input/output/metadata is built from many small values summing over maxDocumentLength but each under maxStringLength; (B) a /spans/batch request with many small spans whose combined document exceeds maxDocumentLength. Assert the ingestion_size_guard metric guard=document_length (not string_length) to prove the whole-document cap fired, since the HTTP message is generic.

_general finding_

apps/opik-backend/src/main/java/com/comet/opik/api/error/JsonProcessingExceptionMapper.java
```diff
@@ -1,21 +1,78 @@
...
+        }
 
-        return Response.status(Response.Status.BAD_REQUEST)
-                .entity(new ErrorMessage(Response.Status.BAD_REQUEST.getStatusCode(),
-                        "Unable to process JSON. " + exception.getMessage()))
+        // Force JSON: ingestion endpoints negotiate non-JSON (e.g. OTel protobuf) with no writer for
+        // ErrorMessage, so without an explicit type the error fails to serialize and surfaces as a 500
+        // (matches InvalidUUIDExceptionMapper and RequestSizeLimitFilter).
+        return Response.status(status)
+                .type(MediaType.APPLICATION_JSON_TYPE)
```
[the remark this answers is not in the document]
YarivHashaiComet: Verified this doesn't apply — the OTLP endpoints never reach this mapper. Both OTLP readers wrap their whole parse in `catch (Exception e) { throw new IOException(...) }`:

- `OtelProtobufMessageBodyReader` → `parseFrom(...)` → `IOException`
- `OtelJsonMessageBodyReader` → parses via `JsonUtils.getJsonNodeFromString` (Jackson) then `JsonFormat`, but any `JsonProcessingException` is caught and re-thrown as `IOException`

`JsonProcessingExceptionMapper` is an `ExceptionMapper<JsonProcessingException>`, and a plain `IOException` is a *supertype* of `JsonProcessingException`, so the mapper is not applicable to it. It therefore never runs for an OTLP request, and the `.type(application/json)` added here can't change any OTLP response — OTLP errors are handled on the `IOException` path, untouched. (There's also no "JSON-only path" to scope the mapper to: it already only handles `JsonProcessingException`, which OTLP doesn't surface.)

Heap safety is unaffected either way: `getJsonNodeFromString` uses the size-constraint-configured mapper, so an oversized OTLP/JSON body still aborts mid-parse (before the re-wrap) — the OOM guard works; only the clean 413/metric is skipped for OTLP, which is what keeps the JSON `ErrorMessage` out of the OTLP contract. No code change.

thiagohora (dismissed):

thiagohora (approved):

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7483`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

PR #7487, opened 2026-07-16
fix(metrics): score ChrF against each reference for multiple references

chuenchen309:
## Details

`ChrF.score` accepts `reference: Union[str, Sequence[str]]` and builds a list of references, so multiple references are a documented, type-hinted feature. But the default NLTK backend passed that whole list straight to `nltk.translate.chrf_score.sentence_chrf(references, candidate)`, whose `reference` argument is a **single** reference string — not a list. NLTK misinterprets the list and returns a wrong score: a candidate identical to one of the references scored ~0.26 instead of 1.0 (even lower than scoring against that reference alone).

This scores the candidate against each reference separately and takes the best match — standard multi-reference chrF semantics (as in sacrebleu / HuggingFace `evaluate`). Single-reference scoring is unchanged, and the custom `chrf_fn` injection contract is preserved (it still receives the full references list).

- Before: `ChrF().score(output="hello world", reference=["completely different text here", "hello world"]).value` → `0.2552`
- After: → `1.0`

Verified against the real installed code before/after. Added `test_chrf_metric__multiple_references__scores_against_best_match` (the existing ChrF tests only covered single-reference and the custom-`chrf_fn` path, so the real multi-reference NLTK path was untested) — red before the fix, green after. Full ChrF test set: 3 passed. `ruff check` + `ruff format`: clean.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude (Anthropic)
- Scope: Bug found via an AI-assisted review pass over `sdks/python/src/opik/evaluation/metrics/`; the fix and regression test were AI-drafted and then independently reproduced and verified (red-before/green-after against the real installed code, ruff clean) before submitting.

### DIFF
--- a/sdks/python/src/opik/evaluation/metrics/heuristics/chrf.py
+++ b/sdks/python/src/opik/evaluation/metrics/heuristics/chrf.py
@@ -85,11 +85,11 @@ class ChrF(BaseMetric):
                     " `pip install nltk` or provide `chrf_fn`."
                 )
 
-            def _compute(candidate: Sequence[str], references: Sequence[str]) -> float:
+            def _score_single(candidate: Sequence[str], reference: str) -> float:
                 try:
                     return float(
                         nltk_chrf_score.sentence_chrf(
-                            references,
+                            reference,
                             candidate,
                             max_len=self._char_order,
                             beta=self._beta,
@@ -98,7 +98,14 @@ class ChrF(BaseMetric):
                     )
                 except TypeError:
                     # Older NLTK versions expose the helper with fewer keyword arguments.
-                    return float(nltk_chrf_score.sentence_chrf(references, candidate))
+                    return float(nltk_chrf_score.sentence_chrf(reference, candidate))
+
+            def _compute(candidate: Sequence[str], references: Sequence[str]) -> float:
+                # NLTK's ``sentence_chrf`` supports only a single reference; passing
+                # a list of references makes it misinterpret the input and return a
+                # wrong score. For multiple references, score against each one and
+                # take the best match (standard multi-reference chrF semantics).
+                return max(_score_single(candidate, ref) for ref in references)
 
             self._chrf_fn = _compute
 
--- a/sdks/python/tests/unit/evaluation/metrics/test_heuristics.py
+++ b/sdks/python/tests/unit/evaluation/metrics/test_heuristics.py
@@ -525,6 +525,33 @@ def test_chrf_metric__char_order_and_ignore_whitespace_vary__change_score():
     assert order_1 != order_6
 
 
+def test_chrf_metric__multiple_references__scores_against_best_match():
+    # NLTK's sentence_chrf accepts only a single reference; passing the list of
+    # references straight through made it misinterpret the input and return a
+    # wrong score, even when the candidate exactly matched one reference. A
+    # candidate identical to any reference must score 1.0 (best-match semantics).
+    pytest.importorskip("nltk")
+
+    metric = ChrF(track=False)
+
+    match_second = metric.score(
+        output="hello world",
+        reference=["completely different text here", "hello world"],
+    ).value
+    match_first = metric.score(
+        output="hello world",
+        reference=["hello world", "completely different text here"],
+    ).value
+
+    assert match_second == pytest.approx(1.0)
+    assert match_first == pytest.approx(1.0)
+
+    # A perfect match against one reference must not score below matching that
+    # reference alone (the pre-fix bug returned a middling value here).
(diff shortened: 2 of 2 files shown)


feiiiiii5 (commented):
### The diagnosis is right — reproduced independently

Ran the exact call `ChrF.__init__` makes, against the real `nltk.translate.chrf_score` (`sentence_chrf(reference, hypothesis, min_len=1, max_len=6, beta=3.0, ignore_whitespace=True)`), with opik's own kwargs (`max_len=char_order`, `beta=beta`, `ignore_whitespace`).

| hypothesis | references | list passed (pre-fix) | per-reference max |
| --- | --- | --- | --- |
| `hello world` | `["completely different text here", "hello world"]` | **0.2552** | 1.0000 |
| `hello world` | `["hello world", "completely different text here"]` | **0.2552** | 1.0000 |
| `the cat sat on the mat` | `["a cat is sitting on a mat", "the cat sat on the mat"]` | 0.4865 | 1.0000 |
| `hello world today` | `["hello world", "goodbye moon"]` | 0.4620 | 0.8772 |
| `fast accurate answer` | `["quick precise reply", "rapid correct response", "fast answer"]` | 0.1799 | 0.5947 |

Your `0.2552` reproduces to four decimals, and note it is **reference-order independent** — so this is not a "wrong order of arguments" artefact, NLTK genuinely degrades on a list. For the priority question: the deflation is 0.41–0.74 absolute on the realistic rows, which is benchmark-poisoning magnitude for any multi-reference eval, not a cosmetic rounding issue.

### The "standard multi-reference chrF" claim checks out against another implementation

I compared `max(score per reference)` against `sacrebleu.sentence_chrf(hypothesis, references, char_order=6, word_order=0, beta=2, remove_whitespace=True)/100` on the same five inputs: **max |Δ| = 0.0017, mean |Δ| = 0.0006** (1.0000/1.0000, 1.0000/1.0000, 1.0000/1.0000, 0.8772/0.8790, 0.5947/0.5960). So best-match-per-reference is not just your convention — it agrees with an independent chrF implementation to <0.2%. That is the strongest argument for landing this.

### Why it has sat since 16 July, though: the same fix is open in #7925

While checking that no other PR touched these lines, I found **#7925** (created 2026-08-20, a month after yours) rewrites the same `_compute` into the same `return max(_score_single(candidate, reference) for reference in references)` shape, adds `test_chrf_metric__default_backend__scores_each_reference_separately` and `test_chrf_metric__multiple_references__scores_against_best_reference`, and also covers METEOR. The two cannot both merge — they edit the same lines of `heuristics/chrf.py`. Two differences that are yours to keep or fold in:

1. **The `except TypeError` fallback.** #7925 deletes it, arguing NLTK ≥3.4 has always accepted these keywords so a `TypeError` would now be a real bug being silently downgraded to a keyword-less call. Worth knowing before that decision is made: `nltk` is **unpinned** in this SDK (`sdks/python/tests/unit/test_requirements.txt:2` is a bare `nltk`, and it is an optional import in `chrf.py`), so no metadata supports a floor. Keeping the fallback (your version) costs nothing on current NLTK; removing it is a defensible strictness call but should be a stated decision, not a side effect of which PR lands.
2. **Type annotations.** #7925 changes the public alias `ChrFFn = Callable[[Sequence[str], Sequence[str]], float]` to `Callable[[str, Sequence[str]], float]`. That is correct — `score()` calls `self._chrf_fn(output_text, references)` with a `str` candidate — and your `_score_single(candidate: Sequence[str], ...)` inherits the same stale annotation. Not behaviour, but the alias is public, so whoever merges should carry that fix too.

### What I'd suggest to the maintainers

Both PRs are correct; yours is the smaller, single-defect diff and it is a month older, #7925 is broader and already has a thread going. Whichever is chosen, the other should be closed as a duplicate with the earlier author credited rather than re-reviewed — I have not asked anyone to do that here, just laid out the overlap so it is one decision instead of two.

Verification method, so this is reproducible: the table comes from calling NLTK's `sentence_chrf` directly in an isolated venv (Python 3.11.15, `nltk` 3.10.3, `sacrebleu` 2.6.0) with the same kwargs `ChrF.__init__` passes; the `nltk` unpinning is read from `sdks/python/tests/unit/test_requirements.txt:2` at this head. I did not run opik's own suite for this review — the numbers above are from the metric's underlying call, which is what the change is about. AI-assisted review; every figure comes from the commands named here.

aadereiko (approved):
Looks good on my end, thank you!

PR #7494, opened 2026-07-16
[OPIK-7362] [BE] fix: null-safe PythonEvaluatorService response handling

thiagohora:
## Details

`PythonEvaluatorService.processResponse(...)` dereferenced `readEntity(...)` results directly. When the response entity deserializes to `null` — which can happen because the response is consumed further down the reactive chain on a `boundedElastic` thread, after the underlying connection has been recycled — the success path did `readEntity(...).scores()` and the error path did `readEntity(...).error()`, both throwing a bare `NullPointerException` that surfaced as an opaque HTTP 500 (counted as `opik_errors_count_total{error_type="NullPointerException"}`).

This turned a user-side error (an invalid user-defined Python metric, for which python-backend returns `400 {"error": "..."}`) into a backend 500, and the backend's error text was lost when the structured `error` field was blank.

- Success path: buffer the entity before reading, and null-check both the body and `scores()`; a null/empty body now raises a clear `InternalServerErrorException` instead of an NPE.
- Error path: null-check the parsed error object and fall back to the raw string body when the structured `error` is absent or blank, so the backend detail is preserved.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves OPIK-7362

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: Root-cause analysis from production telemetry, the `PythonEvaluatorService` null-safety change, and the added unit tests.
- Human verification: Author reviewed the diff and confirmed the test run below.

## Testing

Unit tests added to `PythonEvaluatorServiceTest` reproduce the failure modes (each would throw NPE on the previous code) and assert clean exceptions instead:
- 2xx with a null body → `InternalServerErrorException`
- 2xx with null `scores` → `InternalServerErrorException`
- 400 with a null error object → `BadRequestException` (falls back to raw body, no NPE)
- 400 with a blank `error` field → `BadRequestException` carrying the raw body

Commands run (from `apps/opik-backend`):
```
mvn -o test-compile
mvn -o surefire:test -Dtest=PythonEvaluatorServiceTest
```
Result: `Tests run: 18, Failures: 0, Errors: 0, Skipped: 0` — 4 new tests plus the 14 existing tests (no regressions).

## Documentation

No documentation changes required — internal error-handling robustness fix with no API or behavior change on the happy path.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/python/PythonEvaluatorService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/python/PythonEvaluatorService.java
@@ -27,6 +27,7 @@ import static com.comet.opik.domain.evaluators.python.TraceThreadPythonEvaluator
 public class PythonEvaluatorService {
 
     private static final String URL_TEMPLATE = "%s/v1/private/evaluators/python";
+    private static final int MAX_ERROR_MESSAGE_LENGTH = 500;
 
     private final @NonNull RetriableHttpClient client;
     private final @NonNull OpikConfiguration config;
@@ -101,21 +102,25 @@ public class PythonEvaluatorService {
             try {
                 var errorResponse = response.readEntity(PythonEvaluatorErrorResponse.class);
                 if (errorResponse != null && StringUtils.isNotBlank(errorResponse.error())) {
-                    return errorResponse.error();
+                    return StringUtils.truncate(errorResponse.error(), MAX_ERROR_MESSAGE_LENGTH);
                 }
             } catch (RuntimeException parseErrorResponse) {
-                log.warn("Failed to parse error response, falling back to parsing string", parseErrorResponse);
+                // Expected when the body is not the structured error shape; fall back to the raw body.
+                log.debug("Failed to parse structured error response, falling back to parsing string",
+                        parseErrorResponse);
             }
 
             // Fall back to the raw body when the structured error is absent/blank so the backend
             // detail is not lost (e.g. python-backend "can't be evaluated:" with an empty message).
+            // The body is the evaluated user metric's own error, which we want to surface; bound its
+            // length so an oversized payload can't bloat the thrown exception message.
             try {
                 var body = response.readEntity(String.class);
                 if (StringUtils.isNotBlank(body)) {
-                    return body;
+                    return StringUtils.truncate(body, MAX_ERROR_MESSAGE_LENGTH);
                 }
             } catch (RuntimeException parseStringResponse) {
-                log.warn("Failed to parse error string response", parseStringResponse);
+                log.warn("Failed to read error response body", parseStringResponse);
             }
         }
 
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/python/PythonEvaluatorService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/python/PythonEvaluatorService.java
@@ -14,6 +14,7 @@ import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.collections4.MapUtils;
+import org.apache.commons.lang3.StringUtils;
 import reactor.core.publisher.Mono;
 
 import java.util.List;
@@ -70,7 +71,19 @@ public class PythonEvaluatorService {
         Response.StatusType statusInfo = response.getStatusInfo();
 
         if (statusInfo.getFamily() == Response.Status.Family.SUCCESSFUL) {
-            return response.readEntity(PythonEvaluatorResponse.class).scores();
+            // Buffer before reading: the response is consumed further down the reactive chain on a
+            // boundedElastic thread, after which the underlying connection can already be recycled and
+            // readEntity would yield null. A null body (or null scores) must surface as a clean error
+            // rather than a NullPointerException.
+            var body = response.hasEntity() && response.bufferEntity()
+                    ? response.readEntity(PythonEvaluatorResponse.class)
+                    : null;
+            if (body == null || body.scores() == null) {
+                throw new InternalServerErrorException(
+                        "Python evaluation returned HTTP %d with an empty or unparseable response body"
+                                .formatted(statusCode));
+            }
+            return body.scores();
         }
 
         String errorMessage = extractErrorMessage(response);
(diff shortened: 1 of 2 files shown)


apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/python/PythonEvaluatorService.java
```diff
@@ -84,22 +97,29 @@ private List<PythonScoreResult> processResponse(Response response) {
...
+            }
+
+            // Fall back to the raw body when the structured error is absent/blank so the backend
+            // detail is not lost (e.g. python-backend "can't be evaluated:" with an empty message).
+            try {
+                var body = response.readEntity(String.class);
+                if (StringUtils.isNotBlank(body)) {
+                    return body;
                 }
+            } catch (RuntimeException parseStringResponse) {
```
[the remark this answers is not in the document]
thiagohora: Thanks — addressed in 10c141b. I demoted the structured-parse fallback to `debug` (it fires on every non-structured body, so it was expected noise) and bounded the surfaced text to 500 chars.

I intentionally did **not** switch to a fixed generic message: surfacing the backend error detail is the point of this fix (OPIK-7362). This 400 body is the *evaluated user metric’s own* error (e.g. `... can’t be evaluated: NameError`), and it is routed to that workspace’s user-facing logs via `OnlineScoringUserDefinedMetricPythonScorer`’s `doOnError` — replacing it with "Unknown error" would re-hide exactly the detail we set out to expose. python-backend is an internal service (not third-party), so the leakage surface is low; the truncation caps unbounded payloads. Kept the raw-body *read failure* at `warn` since reaching it (buffered entity that can’t be read even as a String) is genuinely unexpected. Happy to revisit if we’d rather gate the passthrough behind a flag.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/python/PythonEvaluatorService.java
```diff
@@ -70,7 +72,19 @@ private List<PythonScoreResult> processResponse(Response response) {
         Response.StatusType statusInfo = response.getStatusInfo();
 
         if (statusInfo.getFamily() == Response.Status.Family.SUCCESSFUL) {
-            return response.readEntity(PythonEvaluatorResponse.class).scores();
+            // Buffer before reading: the response is consumed further down the reactive chain on a
+            // boundedElastic thread, after which the underlying connection can already be recycled and
+            // readEntity would yield null. A null body (or null scores) must surface as a clean error
+            // rather than a NullPointerException.
+            var body = response.hasEntity() && response.bufferEntity()
```
andrescrz, line 79: This is exactly the right fix. My only ask is to audit all the related online scoring functionality to make sure that every `readEntity` buffers before (and entity check). No need to audit the whole service, it's out of the scope of this fix.

apps/opik-backend/src/test/java/com/comet/opik/domain/evaluators/python/PythonEvaluatorServiceTest.java
```diff
@@ -513,6 +518,88 @@ void evaluate__whenBothErrorResponseParsingFails__shouldUseDefaultMessage() {
...
+                    // Structured error object deserializes to null -> fall back to the raw body.
+                    Arguments.of("null error object", null, "raw backend error body", "raw backend error body"),
+                    // Structured error field is blank (python-backend "can't be evaluated:").
+                    Arguments.of("blank error field", PythonEvaluatorErrorResponse.builder().error("").build(),
+                            "The provided 'code' and 'data' fields can't be evaluated: NameError", "NameError"));
+        }
+
+        @ParameterizedTest(name = "{0}")
+        @MethodSource("missingErrorDetailResponses")
+        void evaluate__when400AndMissingErrorDetail__shouldFallBackToStringBody(
```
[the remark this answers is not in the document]
thiagohora: Good catch — addressed in dd88aba. Switched the blank case to whitespace-only (`"   "`) so it exercises `StringUtils.isNotBlank()` rather than just `isEmpty()`, and added `evaluate__when400AndErrorBodyExceedsMaxLength__shouldTruncateSurfacedText`, which feeds a 604-char body and asserts the thrown message is capped at 500 chars (`hasSize(500)`) while keeping the prefix. Both new paths now fail if the guard/truncation regress. 19/19 in the class.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/python/PythonEvaluatorService.java
```diff
@@ -70,7 +72,19 @@ private List<PythonScoreResult> processResponse(Response response) {
...
+                    ? response.readEntity(PythonEvaluatorResponse.class)
+                    : null;
+            if (body == null || body.scores() == null) {
+                throw new InternalServerErrorException(
+                        "Python evaluation returned HTTP %d with an empty or unparseable response body"
```
andrescrz: Nit: surround single quotes `'%s'` to escape the printed value (statusCode).

andrescrz (dismissed):
One consideration to double check and a nit, but no blockers.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/python/PythonEvaluatorService.java
thiagohora: Thanks! Audited the online-scoring `readEntity` sites (kept it scoped as you said):

- `PythonEvaluatorService` — the gap, now guarded (buffer + entity/null checks on both the success and error paths; covers `evaluate` and `evaluateThread`, which share `processResponse`).
- `RetriableHttpClient` retryable (503/504) branch — already `bufferEntity()` before `readEntity` (line 59).
- The online-scoring scorers themselves (`OnlineScoringUserDefinedMetricPythonScorer`, `OnlineScoringLlmAsJudgeScorer`, `ChatCompletionService`) do not call `readEntity` directly — they go through `PythonEvaluatorService` or langchain4j.

For completeness (outside the online-scoring scope, so not touched): `WebhookHttpClient.readResponseBody` already does hasEntity + bufferEntity + isNotBlank in a try/catch, and `LlmModelRegistryService` reads an `InputStream` synchronously via try-with-resources with a status check — both safe. No further changes needed within scope.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/python/PythonEvaluatorService.java
thiagohora: Done in ba0699e — both `InternalServerErrorException` messages now quote the status code, e.g. `Python evaluation failed (HTTP '500'): ...`. Updated the affected test assertion too. 19/19 in the class.

andrescrz (approved):

PR #7499, opened 2026-07-16
[OPIK-7496] [BE] fix: strip otel attribute rule prefix from saved key

### LINKED ISSUE #7496: [Bug]: Strip rule prefix from otel attributes

- [ ] Opik Python SDK
- [ ] Opik Typescript SDK
- [ ] Opik UI
- [x] Opik Server
- [ ] Documentation

### Opik version

Any

### Describe the problem

In the otel ingestion endpoint there are a few prefix attributes:
- `output`
- `opik.metadata`
- `gen_ai.usage`
- `gen_ai.input`
- `gen_ai.output`
- `gen_ai.tool`
- `gen_ai.agent`
- `gen_ai.request`
- `gen_ai.response`

Which map into input, output or metadata. When matching the prefix rule the `enrichSpanAttributes` method sets the value in the respective bucket to the attribute key. This means that the attribute:
`output: {"choices": [...]}`, will be parsed as `{"output": {"choices": [...]}}`. Similarly, `output.choices: [...]` would parse as the same thing.

So the UI will never show the Messages tab. 

A possible solution would be to strip the prefix that the rule matched. That would produce far nicer results on the Opik side, and allow you to send much nicer traces to other ingestion sites.

_No response_

### Error logs or stack trace

_No response_

### Healthcheck results

_No response_

albe2669:
## Details

In the otel ingestion endpoint there are a few prefix attributes:
- `output`
- `opik.metadata`
- `gen_ai.usage`
- `gen_ai.input`
- `gen_ai.output`
- `gen_ai.tool`
- `gen_ai.agent`
- `gen_ai.request`
- `gen_ai.response`

Which map into input, output or metadata. When matching the prefix rule the `enrichSpanAttributes` method sets the value in the respective bucket to the attribute key. This means that the attribute:
`output: {"choices": [...]}`, will be parsed as `{"output": {"choices": [...]}}`. Similarly, `output.choices: [...]` would parse as the same thing.

This PR strips the prefix of the matched rule, if the outcome is `INPUT`, `OUTPUT` or `METADATA`. This will allow an OpenTelemetry trace to produce a trace that is significantly closer to a native Opik trace.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves #7496 

## AI-WATERMARK

AI-WATERMARK: [no]

## Testing

```bash
mvn test
```


apps/opik-backend/src/main/java/com/comet/opik/domain/OpenTelemetryMapper.java
```diff
@@ -211,7 +213,27 @@ public static void enrichSpanWithAttributes(SpanBuilder spanBuilder, List<KeyVal
...
+                        try {
+                            var jsonNode = JsonUtils.getJsonNodeFromString(value.getStringValue());
+                            if (jsonNode.isObject()) {
+                                jsonNode.fields()
+                                        .forEachRemaining(entry -> node.set(entry.getKey(), entry.getValue()));
+                            } else {
+                                extractToJsonColumn(node, rule.getRule(), value);
+                            }
+                        } catch (RuntimeException e) {
+                            extractToJsonColumn(node, rule.getRule(), value);
```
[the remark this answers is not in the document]
albe2669: Fixed in: https://github.com/comet-ml/opik/pull/7499/commits/5cac19953d856711fce1d5d05ff2ff2b08d2a10f

apps/opik-backend/src/main/java/com/comet/opik/domain/OpenTelemetryMapper.java
```diff
@@ -211,7 +213,27 @@ public static void enrichSpanWithAttributes(SpanBuilder spanBuilder, List<KeyVal
...
+                    String jsonKey = storageKey(rule, key);
+                    // If the suffix is empty and the value is a  JSON object then merge into the node,
+                    // else nest under the stripped key, or the rule key if empty.
+                    if (jsonKey.isEmpty() && value.getValueCase() == AnyValue.ValueCase.STRING_VALUE
+                            && value.getStringValue().startsWith("{")) {
+                        try {
+                            var jsonNode = JsonUtils.getJsonNodeFromString(value.getStringValue());
+                            if (jsonNode.isObject()) {
+                                jsonNode.fields()
+                                        .forEachRemaining(entry -> node.set(entry.getKey(), entry.getValue()));
```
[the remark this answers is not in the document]
albe2669: Fixed in: https://github.com/comet-ml/opik/pull/7499/commits/f8fab73e395555a65ed73286aab50dd286fc50fc

apps/opik-backend/src/test/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtilsTest.java
```diff
@@ -379,4 +380,46 @@ void testExtractTags_UnsupportedValues(AnyValue value, String description) {
...
+    }
+
+    @Test
+    void storageKeyExactRuleNeverStrips() {
+        var exact = OpenTelemetryMappingRule.builder()
+                .rule("gen_ai.prompt").source("GenAI")
+                .outcome(OpenTelemetryMappingRule.Outcome.INPUT).build();
+
+        assertThat(storageKey(exact, "gen_ai.prompt")).isEqualTo("gen_ai.prompt");
+    }
```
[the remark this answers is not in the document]
albe2669: Fixed in: https://github.com/comet-ml/opik/pull/7499/commits/07a2f0bee3baf6db4d66462a16d10e18574455f5

apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java
```diff
@@ -82,6 +82,23 @@ public static void extractToJsonColumn(ObjectNode node, String key, @NonNull Any
...
+    public static String storageKey(OpenTelemetryMappingRule rule, String key) {
+        if (!rule.isPrefix()) {
+            return key;
+        }
+
+        String jsonKey = key.substring(rule.getRule().length());
+
+        if (jsonKey.startsWith(".")) {
+            jsonKey = jsonKey.substring(1);
+        }
```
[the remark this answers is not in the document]
albe2669: Fixed in: https://github.com/comet-ml/opik/pull/7499/commits/ff4e232401438f04283418ce81e683182ec43f28

apps/opik-backend/src/main/java/com/comet/opik/domain/OpenTelemetryMapper.java
```diff
@@ -211,7 +213,27 @@ public static void enrichSpanWithAttributes(SpanBuilder spanBuilder, List<KeyVal
                         default -> metadata;
                     };
 
-                    extractToJsonColumn(node, key, value);
+                    String jsonKey = storageKey(rule, key);
+                    // If the suffix is empty and the value is a  JSON object then merge into the node,
+                    // else nest under the stripped key, or the rule key if empty.
+                    if (jsonKey.isEmpty() && value.getValueCase() == AnyValue.ValueCase.STRING_VALUE
+                            && value.getStringValue().startsWith("{")) {
```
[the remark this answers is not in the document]
albe2669: Fixed in: https://github.com/comet-ml/opik/pull/7499/commits/db5421ce706a1652c570078198ef3c7827e09fad

apps/opik-backend/src/test/java/com/comet/opik/domain/OpenTelemetryMapperTest.java
```diff
@@ -1804,4 +1804,144 @@ void newContextGoesToMetadataOnNonLlmSpans() {
...
+            assertThat(span.output().has("choices")).isTrue();
+        }
+
+        @Test
+        void opikMetadataMergeFiltersReservedKeys() {
+            var span = enrich(List.of(str("opik.metadata",
+                    "{\"thread_id\":\"spoofed\",\"custom\":\"ok\"}")));
+
+            assertThat(span.metadata().has("thread_id")).isFalse();
+            assertThat(span.metadata().get("custom").asText()).isEqualTo("ok");
```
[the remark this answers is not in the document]
albe2669: Fixed in: https://github.com/comet-ml/opik/pull/7499/commits/d6938c93131f29d3152b019246ef6b7794510b1d

apps/opik-backend/src/main/java/com/comet/opik/domain/OpenTelemetryMapper.java
```diff
@@ -211,7 +217,32 @@ public static void enrichSpanWithAttributes(SpanBuilder spanBuilder, List<KeyVal
...
+                                            if (!RESERVED_METADATA_KEYS.contains(entry.getKey())) {
+                                                node.set(entry.getKey(), entry.getValue());
+                                            }
+                                        });
+                            } else {
+                                extractToJsonColumn(node, rule.getRule(), value);
+                            }
+                        } catch (UncheckedIOException e) {
+                            log.debug("Failed to parse JSON for key '{}' (value: '{}'), falling back to text",
+                                    key, value.getStringValue(), e);
```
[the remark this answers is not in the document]
albe2669: Fixed in: https://github.com/comet-ml/opik/pull/7499/commits/d6938c93131f29d3152b019246ef6b7794510b1d

apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java
```diff
@@ -82,6 +82,29 @@ public static void extractToJsonColumn(ObjectNode node, String key, @NonNull Any
         }
     }
 
+    // Prefix rules strip the prefix and return the suffix (without leading dot).
+    // May return an empty suffix, in which case the value should be merged into the object.
+    // Exact rules keep the full key.
+    public static String storageKey(OpenTelemetryMappingRule rule, String key) {
```
andrescrz: For public methods, they should be defensive and validate parameters. Typically, at least Lombok `@NonNull` validation when null args aren't handled and trigger a NPE, but no limited to that.

apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java
```diff
@@ -82,6 +82,29 @@ public static void extractToJsonColumn(ObjectNode node, String key, @NonNull Any
...
+        if (!rule.isPrefix()) {
+            return key;
+        }
+
+        String prefix = rule.getRule();
+        String suffix = key.substring(prefix.length());
```
andrescrz, line 108: Typically, getting a substring can throw a RuntimeException, such as IndexOutOfBoundsException etc.

You should handle that condition appropriately for your method logic: defaulting to some other suffix (including empty string) or some other appropriate action. 

From the logic below, it seems that empty suffix is the right call, but please double check.

If the appropriate action was allowing the exception to bubble up, you should document that in the method's Javadoc.

apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java
```diff
@@ -82,6 +82,29 @@ public static void extractToJsonColumn(ObjectNode node, String key, @NonNull Any
         }
     }
 
+    // Prefix rules strip the prefix and return the suffix (without leading dot).
+    // May return an empty suffix, in which case the value should be merged into the object.
+    // Exact rules keep the full key.
```
andrescrz: This sort of comments, specially when on public methods, should be better shaped as Javadocs.

apps/opik-backend/src/main/java/com/comet/opik/domain/OpenTelemetryMapper.java
```diff
@@ -211,7 +217,31 @@ public static void enrichSpanWithAttributes(SpanBuilder spanBuilder, List<KeyVal
...
+                                                node.set(entry.getKey(), entry.getValue());
+                                            }
+                                        });
+                            } else {
+                                extractToJsonColumn(node, rule.getRule(), value);
+                            }
+                        } catch (UncheckedIOException e) {
+                            log.debug("Failed to parse JSON, falling back to text for key '{}'", key, e);
+                            extractToJsonColumn(node, rule.getRule(), value);
+                        }
```
andrescrz: Please double check this case. I believe `UncheckedIOException` is never thrown here.

andrescrz (commented):
Hi @albe2669 

Thanks for the PR! I just finished reviewing it and left a few comments. When you have a chance, could you take a look and address them?

Thanks again for your contribution!

apps/opik-backend/src/main/java/com/comet/opik/domain/OpenTelemetryMapper.java
albe2669: The `JsonUtils.getJsonNodeFromString` methods wraps the errors that occur in it, in `UncheckedIOException`, so that's why I'm matching it here. :)

apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java
albe2669: Fixed in: https://github.com/comet-ml/opik/pull/7499/commits/69700cf830fd4589b06fe1f32f89317e816d42c3

apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java
albe2669: Changed to a pretty descriptive comment here: https://github.com/comet-ml/opik/pull/7499/commits/69700cf830fd4589b06fe1f32f89317e816d42c3

apps/opik-backend/src/main/java/com/comet/opik/domain/mapping/OpenTelemetryMappingUtils.java
```diff
@@ -82,6 +82,43 @@ public static void extractToJsonColumn(ObjectNode node, String key, @NonNull Any
...
+     * <p>
+     * For exact-match rules the original key is returned unchanged. For prefix rules the
+     * prefix is stripped and any leading dot on the suffix is removed. The suffix may be
+     * empty, in which case the caller should typically merge the value into the enclosing object.
+     * If a prefix rule does not actually match the key (key shorter than the prefix, or the
+     * stripped suffix does not start with a dot when the prefix lacks one), the original
+     * key is returned unchanged.
+     *
+     * @param rule the mapping rule to apply; must not be {@code null}
+     * @param key the attribute key to transform; must not be {@code null}
```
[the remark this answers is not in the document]
albe2669: I don't think this is necessary. Usage is for now a really simple method, that doesn't have all the extra prefix checking code. They could later, easily, be merged.

albe2669:
Thank you for the review @andrescrz 
I have tried to address your comments :).

apps/opik-backend/src/main/java/com/comet/opik/domain/OpenTelemetryMapper.java
andrescrz: Inside `getJsonNodeFromString`, the `UncheckedIOException` is handled and for that particular case, it returns `TextNode.valueOf(value)`. So technically this catch statement is unreachable. 

If you try to add a test for it to simulate the condition, you'll see you wont be able to reach it (unless stubbing or mocking `JsonUtils.getJsonNodeFromString` and forcing to throw it).

apps/opik-backend/src/main/java/com/comet/opik/domain/OpenTelemetryMapper.java
```diff
@@ -216,7 +222,31 @@ public static void enrichSpanWithAttributes(SpanBuilder spanBuilder, List<KeyVal
...
+                    if (jsonKey.isEmpty() && value.getValueCase() == AnyValue.ValueCase.STRING_VALUE) {
+                        try {
+                            var jsonNode = JsonUtils.getJsonNodeFromString(value.getStringValue());
+                            if (jsonNode.isObject()) {
+                                jsonNode.fields()
+                                        .forEachRemaining(entry -> {
+                                            if (!RESERVED_METADATA_KEYS.contains(entry.getKey())) {
+                                                node.set(entry.getKey(), entry.getValue());
+                                            }
+                                        });
```
andrescrz: Minor: this block could be encapsulated in a small private aux method.

apps/opik-backend/src/main/java/com/comet/opik/domain/OpenTelemetryMapper.java
```diff
@@ -216,7 +222,31 @@ public static void enrichSpanWithAttributes(SpanBuilder spanBuilder, List<KeyVal
...
+                            } else {
+                                extractToJsonColumn(node, rule.getRule(), value);
+                            }
+                        } catch (UncheckedIOException e) {
+                            log.debug("Failed to parse JSON, falling back to text for key '{}'", key, e);
```
andrescrz: If this try-catch stays (related to my other comment), better log at warning level here.

andrescrz (commented):
Hi @albe2669 

I left follow-up comments on one of the open discussions and also some additional feedback. 

The main issue is that CI is failing, and it seems caused by changes related to this PR. Some tests in `OpenTelemetryResourceTest` seems to be the failing.

Please take a look and make sure the new code doesn't cause any regression.

I'll review once the fixes are in place.

Thanks a lot for your contribution.

apps/opik-backend/src/main/java/com/comet/opik/domain/OpenTelemetryMapper.java
albe2669: I'll admit that I'm not that used to Java, but this is the `getJsonNodeFromString` function from `JsonUtils` that I'm seeing:
```java
    public static JsonNode getJsonNodeFromString(@NonNull String value) {
        try {
            return MAPPER.readTree(value);
        } catch (JsonProcessingException e) {
            throw new UncheckedIOException(e);
        }
    }
```
Since `ObjectMapper.readTree` can throw either [JsonParseException](http://fasterxml.github.com/jackson-core/javadoc/2.7/com/fasterxml/jackson/core/JsonParseException.html?is-external=true), [IOException](http://docs.oracle.com/javase/7/docs/api/java/io/IOException.html?is-external=true) or [JsonProcessingException](http://fasterxml.github.com/jackson-core/javadoc/2.7/com/fasterxml/jackson/core/JsonProcessingException.html?is-external=true) then it seems like catching `UncheckedIOException` is a pretty valid choice. I suppose it could be `JsonParseException` instead tho.

albe2669: I see now that there is also a `getJsonNodeFromStringWithFallback` that does exactly what you say it does:
```java
    public static JsonNode getJsonNodeFromStringWithFallback(@NonNull String value) {
        try {
            return getJsonNodeFromString(value);
        } catch (UncheckedIOException e) {
            return TextNode.valueOf(value);
        }
    }
```

apps/opik-backend/src/main/java/com/comet/opik/domain/OpenTelemetryMapper.java
albe2669: Good point. I've extracted it in this commit: https://github.com/comet-ml/opik/pull/7499/commits/2e3eb739310c98159950505db6da0ce20632447c

apps/opik-backend/src/main/java/com/comet/opik/domain/OpenTelemetryMapper.java
albe2669: I've let the try catch stay. And changed it to warning: https://github.com/comet-ml/opik/pull/7499/commits/2e3eb739310c98159950505db6da0ce20632447c

apps/opik-backend/src/main/java/com/comet/opik/domain/OpenTelemetryMapper.java
```diff
@@ -211,7 +217,16 @@ public static void enrichSpanWithAttributes(SpanBuilder spanBuilder, List<KeyVal
...
+                    if (jsonKey.isEmpty() && value.getValueCase() == AnyValue.ValueCase.STRING_VALUE) {
+                        mergeJsonObjectOrFallback(node, rule.getRule(), key, value);
+                    } else if (jsonKey.isEmpty()) {
+                        extractToJsonColumn(node, rule.getRule(), value);
+                    } else {
+                        extractToJsonColumn(node, jsonKey, value);
+                    }
                     break;
 
                 case TAGS :
```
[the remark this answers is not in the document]
albe2669: The fallback to rule.getRule() only happens when storageKey returns empty (exact prefix match, no suffix). So this won't happen. There are tests to ensure that

apps/opik-backend/src/main/java/com/comet/opik/domain/OpenTelemetryMapper.java
```diff
@@ -293,6 +308,31 @@ public static void enrichSpanWithAttributes(SpanBuilder spanBuilder, List<KeyVal
...
+            var jsonNode = JsonUtils.getJsonNodeFromString(value.getStringValue());
+            if (jsonNode.isObject()) {
+                jsonNode.fields()
+                        .forEachRemaining(entry -> {
+                            if (!RESERVED_METADATA_KEYS.contains(entry.getKey())) {
+                                node.set(entry.getKey(), entry.getValue());
+                            }
+                        });
+            } else {
+                extractToJsonColumn(node, ruleKey, value);
```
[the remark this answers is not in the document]
albe2669: Fair, fixed in: https://github.com/comet-ml/opik/pull/7499/commits/e04eddae075c01efda252899a54007d33f1239cc

apps/opik-backend/src/main/java/com/comet/opik/domain/OpenTelemetryMapper.java
```diff
@@ -293,6 +308,31 @@ public static void enrichSpanWithAttributes(SpanBuilder spanBuilder, List<KeyVal
...
+            }
+        } catch (UncheckedIOException e) {
+            log.warn("Failed to parse JSON, falling back to text for key '{}'", key, e);
+            extractToJsonColumn(node, ruleKey, value);
```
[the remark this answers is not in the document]
albe2669: Not really an issue after the previous fix

apps/opik-backend/src/main/java/com/comet/opik/domain/OpenTelemetryMapper.java
andrescrz: I made a mistake and looked at the wrong method, as you mentioned, this makes sense.

apps/opik-backend/src/main/java/com/comet/opik/domain/OpenTelemetryMapper.java
andrescrz: Makes sense regarding keeping the catch (related to my confusing comment). Thanks for updating the log level.

andrescrz (approved):
Hi @albe2669 

PR #7504, opened 2026-07-17
[OPIK-7497] [BE] fix: set opik parent span id to null if it equals the trace id

### LINKED ISSUE #7497: [Bug]: Top-level spans lose reference to root level when ingesting in OpenTelemetry

- [ ] Opik Python SDK
- [ ] Opik Typescript SDK
- [ ] Opik UI
- [x] Opik Server
- [ ] Documentation

### Opik version

- Opik version: x.x.x

### Describe the problem

When a root span, ingested via OpenTelemetry has a `parent_span_id` set to the trace, that id disappears when it's ingested because the root span is converted into a trace via Redis, but the contents of `parent_span_id` are converted using `convertOtelIdToUUIDv7`, which produces a different ID.

_No response_

### Error logs or stack trace

_No response_

### Healthcheck results

_No response_

albe2669:
## Details

Fixes orphaned spans when OTel tools set `parent_span_id` to the trace ID (meaning "top-level span") instead of leaving it empty. The mapper now detects `parent_span_id == trace_id` and nulls it so the span becomes a top-level span, and `doStoreSpans` deduplicates root spans by `traceId` so only one trace is created per OTel trace, which prevents a
`EntityAlreadyExistsException` that would drop the entire span batch.

## Issues

- Resolves #7497 

## AI-WATERMARK

AI-WATERMARK: [no]

## Testing
```bash
mvn test
```


apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/OpenTelemetryResourceTest.java
```diff
@@ -1034,6 +1034,62 @@ void testMixedBatchWithAndWithoutOpikTraceIdOverride() {
             assertThat(existingTraceSpans.content().getFirst().traceId()).isEqualTo(existingTraceId);
         }
 
+        @Test
+        @DisplayName("two top-level spans with parent_span_id==trace_id share one trace")
+        void testTwoTopLevelSpansWithParentSpanIdEqualsTraceId() {
```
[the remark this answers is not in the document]
albe2669: https://github.com/comet-ml/opik/pull/7504/commits/accac6460db44d42b0aac34d0025a5a26b4840eb

apps/opik-backend/src/main/java/com/comet/opik/domain/OpenTelemetryService.java
```diff
@@ -149,11 +150,21 @@ private Mono<Long> doStoreSpans(List<Span> otelSpans, Map<String, UUID> traceIdM
...
+        // is nulled in the mapper), so dedup by traceId to create only one trace per id.
+        var rootSpansByTraceId = dedupedSpans.stream()
                 .filter(span -> span.parentSpanId() == null)
-                .filter(span -> !overriddenTraceIds.contains(span.traceId())))
+                .filter(span -> !overriddenTraceIds.contains(span.traceId()))
+                .collect(Collectors.toMap(
+                        com.comet.opik.api.Span::traceId,
+                        java.util.function.Function.identity(),
+                        (first, second) -> first,
+                        LinkedHashMap::new))
```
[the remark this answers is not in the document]
albe2669: https://github.com/comet-ml/opik/pull/7504/commits/b17772b8f06a0eb37f7fdde60d8895c851020be0

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/OpenTelemetryResourceTest.java
```diff
@@ -1034,6 +1034,62 @@ void testMixedBatchWithAndWithoutOpikTraceIdOverride() {
...
+            var llmSpanId = UUID.randomUUID().toString().getBytes();
+
+            // MCP registration span: top-level, parent_span_id set to the trace id
+            var mcpSpan = Span.newBuilder()
+                    .setName("mcp_registration")
+                    .setTraceId(ByteString.copyFrom(otelTraceId))
+                    .setSpanId(ByteString.copyFrom(mcpSpanId))
+                    .setParentSpanId(ByteString.copyFrom(otelTraceId))
+                    .setStartTimeUnixNano((System.currentTimeMillis() - 1_000) * 1_000_000L)
+                    .setEndTimeUnixNano(System.currentTimeMillis() * 1_000_000L)
```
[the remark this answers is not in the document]
albe2669: This is technically wrong, `parentIsTraceId` is pure ByteString.equals and the fixture sets parent_span_id to exactly the trace id, so the test does verify the nulling behavior correctly.
The "real OTel ID shape" is valid, but applies to the entire file and the whole OpenTelemetryMapperTest override class, not just the new test, as  they all use that method. Fixing only the new test would be inconsistent.

apps/opik-backend/src/main/java/com/comet/opik/domain/OpenTelemetryService.java
```diff
@@ -149,11 +150,25 @@ private Mono<Long> doStoreSpans(List<Span> otelSpans, Map<String, UUID> traceIdM
...
-        return Flux.fromStream(dedupedSpans.stream()
+        // Use spans without parentId as Trace roots. Skip opik.trace_id overrides (they attach to
+        // existing traces). Multiple roots can share one traceId (e.g. when parent_span_id==trace_id
+        // is nulled in the mapper), so dedup by traceId to create only one trace per id. Sort by
+        // startTime (then id) before deduping so the root is deterministic.
+        var rootSpansByTraceId = dedupedSpans.stream()
                 .filter(span -> span.parentSpanId() == null)
-                .filter(span -> !overriddenTraceIds.contains(span.traceId())))
+                .filter(span -> !overriddenTraceIds.contains(span.traceId()))
+                .sorted(java.util.Comparator
```
andrescrz: For performance and scalability reasons (including memory and CPU consumption), it's better to avoid sorting this. Better move the deduping logic without sorting to the map collection. I believe that's all you need.

apps/opik-backend/src/main/java/com/comet/opik/domain/OpenTelemetryService.java
```diff
@@ -149,11 +150,25 @@ private Mono<Long> doStoreSpans(List<Span> otelSpans, Map<String, UUID> traceIdM
...
+                        .comparing(com.comet.opik.api.Span::startTime)
+                        .thenComparing(com.comet.opik.api.Span::id))
+                .collect(Collectors.toMap(
```
andrescrz: Related to my other comment, but it seems that the only reason to collect everything in a map is the deduping logic, as you later undo the map by streaming its values (and they key is never really used). 

It'd be great if you can find a simple alternative to prevent all the values from being accumulated and collected, to only just stream them. 

Better have an actual stream of data that doesn't need to wait for accumulation, and hold items in memory.

andrescrz (commented):
Hi @albe2669 

Thanks for the update! I just finished reviewing the PR and left a few comments. When you have a chance, could you take a look and address them?

Thanks again for your contribution!

apps/opik-backend/src/main/java/com/comet/opik/domain/OpenTelemetryService.java
albe2669: I've used a simpler method here: https://github.com/comet-ml/opik/pull/7504/commits/91db1859f8a6c3c67c4db90214248a517f8f05c5

apps/opik-backend/src/main/java/com/comet/opik/domain/OpenTelemetryService.java
albe2669: I've used a simpler method here: https://github.com/comet-ml/opik/pull/7504/commits/91db1859f8a6c3c67c4db90214248a517f8f05c5

andrescrz (approved):
Hi @albe2669,

Best

PR #7511, opened 2026-07-17
[OPIK-7309] [BE] Wire span deletions into the deletion-events bridge (trace-delete cascade) + ArchUnit guard

thiagohora:
## Details

Captures span-row deletions in the `deletion_events_local` bridge so they survive the `spans` table copy during the Slice 3 migration window (R-19). Spans counterpart of the traces work in OPIK-6890 ([#7376](https://github.com/comet-ml/opik/pull/7376), wiring) + OPIK-6891 ([#7390](https://github.com/comet-ml/opik/pull/7390), ArchUnit guard), merged into one ticket.

- Spans have **no standalone delete** (`DELETE /spans/{id}` → 501), so capture wires into the only span-deletion path — the trace-delete cascade `SpanService.deleteByTraceIds` → `spanDAO.deleteByIds` — recording `source_table = spans` with the resolved span ids. Best-effort and deferred (`Mono.defer`, gated by `spanDeletionEventsCaptureEnabled`, failures logged and swallowed) and run only **after** the delete succeeds — identical invariant to `TraceService.captureDeletions`.
- `SourceTable` gains `SPANS`; new `spanDeletionEventsCaptureEnabled` flag (default `false`, env `ANALYTICS_DB_DATA_MODEL_SPAN_DELETION_EVENTS_CAPTURE_ENABLED`), reusing the shared `deletionEventsInsertBatchSize`.
- Retention paths (`SpanDAO.deleteForRetention*`) are intentionally excluded (retention is disabled during the migration window), mirroring #7376.
- `SpanDeletionEventArchTest` guards that any caller of `SpanDAO.deleteByIds` routes through `SpanService`, so a new span-delete path can't silently bypass the bridge.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- OPIK-7309

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: Full change — code, tests, and this description, mirroring the traces counterparts (#7376, #7390).
- Human verification: Diff reviewed and `mvn test-compile` + the ArchUnit and unit tests run locally by the author.

## Testing

- `mvn -o test-compile` -> BUILD SUCCESS
- `mvn -o test -Dtest=SpanDeletionEventArchTest,SpanServiceImplTest` -> 4/4 pass:
  - ArchUnit: `SpanService` is the sole caller of `SpanDAO.deleteByIds` (guard green).
  - Unit (mirrors `TraceServiceImplTest`): flag-off → no capture; capture-insert failure → delete still succeeds; delete failure → no capture, no `SpansDeleted`, error propagates.
- New `SpanDeletionEventTest` (integration, flag-on) asserts the exact `deletion_events_local` SPANS rows after a trace-delete cascade (single + batch), awaiting the async cascade with Awaitility and a small insert batch size to exercise chunking. Runs in CI.
- Not run locally: the integration test — testcontainers bootstrap; relies on CI (same as the traces suite).

## Documentation

No user-facing documentation changes — internal migration-window capture flag, default off.


apps/opik-backend/src/test/java/com/comet/opik/domain/SpanDeletionEventTest.java
```diff
@@ -0,0 +1,213 @@
...
+        var projectName = randomName("project");
+        var trace = createTrace(projectName);
+        var spanIds = createSpans(projectName, trace.id());
+
+        var beforeDelete = Instant.now();
+        traceResourceClient.deleteTrace(trace.id(), WORKSPACE_NAME, API_KEY);
+
+        // The trace-delete cascade resolves each trace's owning project and passes it to the span delete, so the
+        // resolved project_id (here the trace's project) is what gets captured on the span rows.
+        var expectedDeletionEvents = spanIds.stream()
```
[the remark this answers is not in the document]
thiagohora: Added coverage in `168b53b` — a `SpanServiceImplTest` unit case `delete__whenTraceHasNoSpans__thenNoDeleteAndNoCapture` that stubs `getSpanIdsForTraces` to return empty and asserts the cascade short-circuits before the delete (`deleteByIds` never called), records nothing (`verifyNoInteractions(deletionEventDAO, eventBus)`), and completes without error. Note `captureDeletions` itself never receives an empty set — `deleteByTraceIds` returns early on `spanIds.isEmpty()` before reaching it — so the unit test is the right place for this edge, cheaper than an integration case.

apps/opik-backend/src/test/java/com/comet/opik/domain/SpanServiceImplTest.java
```diff
@@ -0,0 +1,176 @@
...
+                opikConfiguration);
+    }
+
+    @Nested
+    @DisplayName("Delete Spans (trace-delete cascade):")
+    class DeleteSpans {
+
+        @Test
+        @DisplayName("when capture is disabled, then delete records no deletion events")
+        void delete__whenCaptureDisabled__thenRecordsNoDeletionEvents() {
```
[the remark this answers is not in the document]
thiagohora: This is the established convention in the suite, not a violation — `method__whenCondition__thenOutcome` is used across ~13 test files (e.g. `executeQuery__whenWorkspaceMismatch__thenReturnsNoRows`), and specifically by the direct counterpart `TraceServiceImplTest` (`delete__whenCaptureDisabled__thenRecordsNoDeletionEvents`), which this class mirrors 1:1. Renaming here would make it the odd one out. Keeping it for consistency.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanService.java
```diff
@@ -511,6 +516,40 @@ public Mono<Void> deleteByTraceIds(@NonNull Set<UUID> traceIds, UUID projectId)
...
+     * survive the {@code spans} table copy during the Slice 3 migration window. Best-effort and deferred: gated by
+     * {@code spanDeletionEventsCaptureEnabled} and run only after the delete succeeds, and any capture failure is
+     * logged and swallowed so it can never disrupt the delete. Spans have no standalone delete, so this cascade is the
+     * only capture path. Mirrors {@code TraceService.captureDeletions}.
+     */
+    private Mono<Void> captureDeletions(Set<UUID> ids, UUID projectId, String workspaceId, String userName) {
+        return Mono.defer(() -> {
+            if (!config.getDatabaseAnalyticsDataModel().spanDeletionEventsCaptureEnabled()) {
+                return Mono.empty();
+            }
```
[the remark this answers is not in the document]
thiagohora: Two parts:

- **Retention bypass is intentional and documented.** `RetentionPolicyService`/`RetentionCatchUpService` call `SpanDAO.deleteForRetention*`, which deliberately do **not** capture — retention is disabled during the migration window (ticket "deliberate deviations", mirroring #7376). The ArchUnit guard scopes to `deleteByIds` only and its javadoc calls this out, exactly like `TraceDeletionEventArchTest` for the trace side. So this is by design, not a gap.
- **The capture-path duplication** (span vs trace) is the same DRY item already tracked in **OPIK-7380** (raised on #7505); the right consolidation covers both entities together rather than a span-only extraction here. Deferred there to keep this PR scoped to the cutover wiring.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanService.java
```diff
@@ -500,6 +504,7 @@ public Mono<Void> deleteByTraceIds(@NonNull Set<UUID> traceIds, UUID projectId)
                                         .doOnSuccess(__ -> log.info(
                                                 "Deleted '{}' spans for workspace '{}', project '{}'",
                                                 spanIds.size(), workspaceId, projectId)))
+                                .then(captureDeletions(spanIds, projectId, workspaceId, userName))
```
JetoPistola, line 507: 🧹 **nit** | Consistency

This chains `captureDeletions` *before* the `SpansDeleted` event is posted (capture sits inside the `flatMap`; the event fires in the outer `.doOnSuccess`). The mirrored `TraceService.delete` does the opposite — posts `TracesDeleted` first, then `.then(captureDeletions(...))`. It's functionally harmless here (capture is deferred, gated, and swallows all errors, so the event fires regardless of capture outcome), but since this PR is otherwise a faithful mirror of the trace path, aligning the order would remove a subtle divergence for the next reader.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanService.java
```diff
@@ -511,6 +516,40 @@ public Mono<Void> deleteByTraceIds(@NonNull Set<UUID> traceIds, UUID projectId)
         });
     }
 
+    /**
```
JetoPistola, line 519: ❓ **question** | Design

Was running capture off the async trace-delete cascade (rather than posting `SpansDeleted` and capturing in the same order as the trace side) a deliberate choice, or just an artifact of where the span-id resolution lands in `deleteByTraceIds`? No action needed if intentional — just confirming the ordering was considered.

JetoPistola:
👋 **Review summary**

**What looks good**
- Faithful mirror of the reviewed trace-side capture (`TraceService.captureDeletions`) — same deferred + gated + swallow-on-failure invariant, so a capture error can never disrupt the delete, and capture runs only after a successful delete.
- The `SpanDeletionEventArchTest` guard is the standout: its signature (`SpanDAO.deleteByIds(Set, UUID)`) matches the real method, `SpanService` is genuinely the sole caller today, and the `allowEmptyShould(true)` rationale is documented — this is exactly what keeps a future span-delete path from silently bypassing the bridge.
- Test coverage is thorough and layered: unit tests cover flag-off / capture-failure-still-deletes / delete-failure-propagates / spanless-trace, and the integration test asserts the exact `deletion_events_local` SPANS rows for both single and batch cascades, with a deliberately small `deletionEventsInsertBatchSize=2` to exercise the chunked-insert path.
- Retention paths intentionally excluded and documented, config flag defaults to `false` with a clear enablement window in the comments, and the shared batch-size doc was updated to reflect it now covers both trace and span capture.

**Overall**
Solid, low-risk change scoped tightly to the migration window. The approach is sound and the scope is appropriate — the only span-deletion path (trace-delete cascade) is wired, and the ArchUnit guard makes the "only path" claim enforceable rather than just documented.

**Inline comments**: 1 nit + 1 question — both about the capture-vs-`SpansDeleted`-event ordering diverging slightly from the trace side. Nothing blocking.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanService.java
andrescrz: The chain seems different as here the event is posted in the outer one. I'd rather keep this approach, it seems more correct. Anyway, not very relevant as the event is async and not related to this particular bridge.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanService.java
andrescrz: Yes, confirmed as intended. Also, the reactive chain is `doOnSuccess` vs `then`, so that should make this order irrelevant.

apps/opik-backend/src/test/java/com/comet/opik/domain/SpanDeletionEventTest.java
```diff
@@ -0,0 +1,213 @@
...
+                .map(spanId -> newExpectedDeletionEvent(trace.projectId(), spanId))
+                .toList();
+        awaitAndAssertSpanDeletionEvents(expectedDeletionEvents, beforeDelete);
+    }
+
+    @Test
+    void deleteBatchTracesCascadeCapturesSpanDeletionEvents() {
```
andrescrz: Please double check if scope to project paramterised test (true or false) is needed here. 

But anyway, it seems duplicated from the other. If you follow my advice, we might just go with the other tests.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanService.java
```diff
@@ -511,6 +516,40 @@ public Mono<Void> deleteByTraceIds(@NonNull Set<UUID> traceIds, UUID projectId)
...
+            if (!config.getDatabaseAnalyticsDataModel().spanDeletionEventsCaptureEnabled()) {
+                return Mono.empty();
+            }
+            var events = ids.stream()
+                    .map(id -> DeletionEvent.builder()
+                            .sourceTable(SourceTable.SPANS)
+                            .workspaceId(workspaceId)
+                            .projectId(projectId)
+                            .deletedId(id.toString())
+                            .deletionReason(DeletionReason.USER_REQUEST)
```
andrescrz: It seems we've never implemented span deletion and all cases are from trace (or other entity) deletions. The reason should be extended to `CASCADE` and used here.

apps/opik-backend/src/test/java/com/comet/opik/domain/DeletionEventTest.java
andrescrz, line 1: A span deletion is caused by cascading a deletion from trace and other entities. You likely need to extend the previously existing test cases from `TraceDeletionEventTest` to cover and assert the span deletion. Those are:
- deleteTraceCapturesDeletionEvent
- deleteBatchTracesCapturesDeletionEvents
- deleteTraceThreadsCapturesDeletionEvents

apps/opik-backend/src/test/java/com/comet/opik/domain/SpanDeletionEventTest.java
```diff
@@ -0,0 +1,213 @@
...
+        ClientSupportUtils.config(clientSupport);
+        var baseUrl = TestUtils.getBaseUrl(clientSupport);
+        this.traceResourceClient = new TraceResourceClient(clientSupport, baseUrl);
+        this.spanResourceClient = new SpanResourceClient(clientSupport, baseUrl);
+        this.deletionEventDAO = deletionEventDAO;
+        mockTargetWorkspace(wireMock.server(), API_KEY, WORKSPACE_NAME, WORKSPACE_ID, USER);
+    }
+
+    @Test
+    void deleteTraceCascadeCapturesSpanDeletionEvents() {
```
andrescrz: Mentioned above. But it seems that as all span deletions are triggered from cascading, these tests duplicate a lot the functionality from `TraceDeletionEventTest`, while other cases are a potentially missing.

Probably better to rename `TraceDeletionEventTest` to a more neutral (or append span) name, and add the missing configurations and assertions.

apps/opik-backend/src/test/java/com/comet/opik/domain/DeletionEventTest.java
```diff
@@ -0,0 +1,213 @@
...
+                .deletionReason(DeletionReason.USER_REQUEST)
+                .build();
+    }
+
+    private void awaitAndAssertSpanDeletionEvents(List<DeletionEvent> expectedDeletionEvents, Instant beforeDelete) {
+        var deletedIds = expectedDeletionEvents.stream()
+                .map(DeletionEvent::deletedId)
+                .collect(Collectors.toUnmodifiableSet());
+        // The span cascade is asynchronous (off the TracesDeleted event), so poll until the bridge rows appear.
+        Awaitility.await()
```
andrescrz, line 264: These are synchronous APIs and deletion is synchronous. I don't believe an await call is needed, likely Claude misunderstood here, as I've seen this happening to me as well.

andrescrz (commented):
Let's iterate a bit on this:
1. We need to capture this cascading deletion reason.
2. Let's polish the testing approach, I believe we have some gaps and also some overlap.

The rest LGTM.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanService.java
thiagohora: Two separate things there:

- **Capture running off the trace-delete cascade** is deliberate and unavoidable — spans have no standalone delete (`DELETE /spans/{id}` → 501), so the cascade in `deleteByTraceIds` is the *only* span-deletion path, and the `SpanDeletionEventArchTest` guard enforces that. Capture has to live there.
- **The capture-vs-event ordering** was the only real artifact, and you were right to flag it — now fixed in `da63804` to post `SpansDeleted` first, then capture, mirroring the trace side. So the async-cascade placement is intentional; the ordering is now consistent.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanService.java
thiagohora: Done in `a3f4304` — added `DeletionReason.CASCADE` and `SpanService.captureDeletions` now records `deletion_reason = cascade` (span deletions are always cascaded). `deletion_reason` is `LowCardinality(String)`, so no migration is needed.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanService.java
thiagohora: Reverted in `a3f4304` — `SpansDeleted` is posted in the outer `doOnSuccess` again, keeping the reviewed approach. Agreed it is irrelevant since the event is async and unrelated to this bridge. (Also flagged by another reviewer in the other direction; going with your call.)

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanService.java
thiagohora: 👍 Confirmed intended — the async trace-delete cascade is the only span-deletion path (`DELETE /spans/{id}` → 501, enforced by `SpanDeletionEventArchTest`). Kept the outer `doOnSuccess` order.

apps/opik-backend/src/test/java/com/comet/opik/domain/DeletionEventTest.java
thiagohora: Done — renamed `TraceDeletionEventTest` → `DeletionEventTest` and extended `deleteTraceCapturesDeletionEvent` / `deleteBatchTracesCapturesDeletionEvents` / `deleteTraceThreadsCapturesDeletionEvents` to create spans under the traces and assert the cascaded span rows (`source_table=spans`, `reason=cascade`) alongside the trace rows. Both capture flags are enabled in the suite.

apps/opik-backend/src/test/java/com/comet/opik/domain/SpanDeletionEventTest.java
thiagohora: Done — renamed to `DeletionEventTest` (neutral) and removed the duplicate `SpanDeletionEventTest`.

apps/opik-backend/src/test/java/com/comet/opik/domain/SpanDeletionEventTest.java
thiagohora: Resolved by the consolidation — deleted the duplicate `SpanDeletionEventTest`; the scope-to-project parameterised case is now your existing `deleteBatchTracesCapturesDeletionEvents` in `DeletionEventTest`, extended to also assert the cascaded span rows.

apps/opik-backend/src/test/java/com/comet/opik/domain/DeletionEventTest.java
thiagohora: Kept a targeted await **only** for the span assertions; trace assertions stay synchronous. Trace capture runs inside `TraceService.delete` before the request returns (hence sync), but span capture runs on the async cascade: `TraceDeletedListener` is `@Subscribe` on the `AsyncEventBus` (`EventModule` → `new AsyncEventBus("opik-event-bus", virtualThreads)`) and fires `processTraceDeletion(...).subscribe()` fire-and-forget, so the span rows land after `DELETE` returns. `TracesResourceTest` uses `Awaitility` for this same trace→span cascade. Dropping the await on the span rows would make it racy — happy to remove it if you know a synchronous path I am missing.

andrescrz (approved):
LGTM

PR #7512, opened 2026-07-17
[OPIK-7159][OPIK-7029] [BE] fix: address post-merge review (Redis-independent status writes; configurable reaper lookback)

awkoy:
## Details

Follow-up to #7428 (merged), addressing @thiagohora's post-merge review. Two backend fixes:

**1. `OptimizationService.update()` no longer hard-depends on Redis for every status change (`[medium]`).**
The distributed lock is now acquired **only for a metadata update** — the read-modify-write (`getById` → merge → insert) that can drop keys under a concurrent partial update. Status-only and name-only writes persist lock-free, so a Redis blip can no longer 500 the worker's `mark_completed` / `mark_error` callback (or the reaper's own ERROR write) and leave a run non-terminal for the reaper to later mislabel `ERROR` — the exact failure this feature exists to prevent. The worker already retries a metadata-carrying status write as a metadata-less one on failure, and that fallback now persists lock-free.

**2. Reaper scan lookback margin is now configurable.**
Was a hardcoded `Duration.ofDays(7)` in `OptimizationDAO`; now a validated, env-overridable config key `optimizationStalledReaper.lookbackMargin` (`OPTIMIZATION_STALLED_REAPER_LOOKBACK_MARGIN`, default `7d`), alongside the sibling `initializedTimeout` / `runningTimeout` / `lockDuration`. Threaded through job → service → DAO so ops can tune the scan floor (and skip-index pruning) without a redeploy.

## Change checklist

- [x] Lock scoped to metadata updates; status/name-only writes are Redis-independent
- [x] `lookbackMargin` added as a validated, env-overridable config key
- [x] New `OptimizationServiceUpdateLockTest` covers both the locked (metadata) and lock-free (status-only) paths
- [x] Reaper job / config / service tests updated for the new parameter

## Issues

OPIK-7159, OPIK-7029 — post-merge review follow-up to #7428.

## Testing

- `OptimizationServiceUpdateLockTest` (new): asserts a status-only update does not touch `LockService`, and a metadata update is serialized under it.
- `OptimizationStalledReaperConfigTest`, `OptimizationStalledReaperJobTest` pass with the new `lookbackMargin` field.
- Backend compiles + these unit tests pass locally on JDK 25; full suite runs in CI.

## Documentation

Config documented inline in `config.yml` (default, env var, and the reaper-downtime tradeoff). No user-facing docs affected.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/OptimizationService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/OptimizationService.java
@@ -34,6 +34,7 @@ import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.redisson.api.RedissonReactiveClient;
+import org.redisson.client.RedisException;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 import reactor.core.scheduler.Schedulers;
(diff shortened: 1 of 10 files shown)


apps/opik-backend/src/main/java/com/comet/opik/domain/OptimizationService.java
```diff
@@ -333,15 +334,39 @@ public Mono<Long> update(@NonNull UUID id, @NonNull OptimizationUpdate update) {
...
+        // the fallback never masks the action's 404/409, and the action runs exactly once (executeWithLock
+        // never subscribes it when acquisition fails). NOTE: prod ClickHouse has no read-your-own-writes
+        // (async insert, 2 replicas), so the lock cannot fully close the window either — studio metadata must
+        // stay effectively single-writer (review: thiagohora + adversarial review of the metadata-only-lock
+        // variant).
+        return guarded.onErrorResume(RedisException.class, error -> {
+            log.warn("Redis unavailable acquiring the optimization-update lock for '{}'; persisting the update "
+                    + "without it (a concurrent lost update is possible but rare)", id, error);
+            return action;
+        });
```
thiagohora: Sorry, the AI flagged this one, but I removed it afterward. The point is valid if other updates are less risky. But locks are lightweight, and they should protect against data loss. If other updates are less frequent, we can keep it; otherwise, I would keep the lock to protect every state change

thiagohora (approved):

Issue #7090, opened 2026-06-15
[Bug]: ClickHouse migration 000017_change_tables_to_replicated.sql is not idempotent — causes permanent CrashLoopBackOff on pod restart


Vasu2803k:
### What component(s) are affected?

- [ ] Opik Python SDK
- [ ] Opik Typescript SDK
- [ ] Opik Agent Optimizer SDK
- [ ] Opik UI
- [x] Opik Server
- [ ] Documentation

### Opik version

- Opik version: 1.11.5

### Describe the problem

Migration `000017_change_tables_to_replicated.sql` converts ClickHouse tables from
 MergeTree → ReplicatedMergeTree. If a pod running this migration is killed mid-execution
 (OOMKill, node drain, pod eviction), it leaves an orphaned ZooKeeper replica path but no
 table on disk. Liquibase never marks the changeset as EXECUTED (because the SQL failed).
 On every subsequent restart, Liquibase retries the same CREATE TABLE — which ClickHouse
 permanently rejects because the ZK replica path already exists.
 
This results in an unrecoverable CrashLoopBackOff on all opik-backend pods scheduled on
 the affected ClickHouse node, with no automatic recovery path.
 
 Expected: The migration is idempotent and safe to retry after partial failure.
 Actual:   Pod enters permanent CrashLoopBackOff (observed: 732 restarts over 2d 18h).

Error logs:

 --- Error 1: on every retry ---
 Migration failed for changeset:
   000017_change_tables_to_replicated.sql::change-tables-to-replicated-01::liyaka
 
 Code: 253. DB::Exception: Replica /clickhouse/tables/0/opik/automation_rule_evaluator_logs
 /replicas/chi-opik-clickhouse-cluster-0-0 already exists. (REPLICA_ALREADY_EXISTS)
 
 --- Error 2: after manually clearing ZK paths ---
 Code: 122. DB::Exception: Table columns structure in ZooKeeper is different from local table structure.
   Local columns:  entity_type Enum8('trace'=1, 'span'=2)
   ZooKeeper cols: entity_type Enum8('trace'=1, 'span'=2, 'thread'=3)
   (INCOMPATIBLE_COLUMNS)
 
 Error 2 occurs because migration 000017's CREATE TABLE SQL hardcodes the schema at the
 time it was written. If the other replica has already run 000017 + subsequent ALTER TABLE
 migrations (e.g. 000028 adding 'thread'=3 to enums), ZooKeeper holds the evolved schema.
[...]

PR #7530, opened 2026-07-20
[OPIK-7332] [HELM][DOCS] fix: two-phase HA guidance for multi-replica ClickHouse installs

andrescrz:
## Details

Fresh Opik installs on a multi-replica ClickHouse cluster cannot complete the analytics migrations: the earliest migrations aren't cluster-aware, so on a fresh multi-replica cluster they only reach one replica and later migrations fail (e.g. `UNKNOWN_TABLE` / `UNKNOWN_DATABASE`), leaving `opik-backend` unable to start. The existing migrations are immutable, so this change documents and signposts the supported **two-phase High-Availability install** (migrate against a single replica, then scale up so the ClickHouse operator clones the schema to the new replicas) instead of altering migration behavior. Related to the open-source multi-replica migration reports (GitHub issue #7090; not fully resolved here).

- Add a `NOTES.txt` reminder, printed only when `clickhouse.replicasCount >= 2`, pointing to the two-phase flow.
- Document the two-phase install in self-host **Scaling** (Replication Strategy) and add a **Troubleshooting** entry for the failure (both docs versions).
- Add neutral guidance on `clickhouse.replicasCount` in `values.yaml`; keep the migration connection on the load-balanced ClickHouse service and the OSS default at a single replica (no behavior change).
- Add a helm-unittest suite covering the NOTES banner.

## Change checklist
- [ ] User facing
- [x] Documentation update

## Issues

- Resolves OPIK-7332

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: full implementation (analysis, Helm chart + docs changes, tests)
- Human verification: code review + local Helm lint/unittest/dry-run

## Testing

- `helm unittest .` (in `deployment/helm_chart/opik`) → 40/40 pass, including the new NOTES suite: banner present at `replicasCount >= 2`, absent at `1`, and absent when ClickHouse is disabled.
- `helm lint .` → 0 failures.
- `helm install --dry-run` at `replicasCount` 1 and 2 → banner renders only for `>= 2` and interpolates the configured target count.
- `helm-docs` (chart's pinned version) → no README drift, so the docs-sync check passes.
- Fern docs: verified MDX brace-safety, heading nesting, and anchor/link targets across both docs versions.
- No application code changed; CI runs the full quality suite.

## Documentation

- `self-host/scaling`: two-phase HA install guidance under Replication Strategy (both docs versions).
- `self-host/troubleshooting`: "Fresh Multi-Replica Install Migration Failures" entry with symptoms, cause, and resolution (both docs versions).
- `values.yaml`: guidance comment on `clickhouse.replicasCount`.

### DIFF
--- a/deployment/helm_chart/opik/tests/clickhouse_migration_test.yaml
+++ b/deployment/helm_chart/opik/tests/clickhouse_migration_test.yaml
@@ -1,0 +1,33 @@
+suite: clickhouse multi-replica migration NOTES reminder
+
+# When ClickHouse runs >= 2 replicas, NOTES.txt reminds operators to use the two-phase
+# HA install flow (migrate at 1 replica, then scale up), since a fresh multi-replica
+# install cannot complete its migrations (OPIK-7332).
+
+templates:
+  - templates/NOTES.txt
+
+tests:
+  - it: NOTES omits the multi-replica migration banner for the single-replica default
+    asserts:
+      - notMatchRegexRaw:
+          pattern: "two-phase install"
+
+  - it: NOTES shows the two-phase migration banner when ClickHouse runs multiple replicas
+    set:
+      clickhouse.replicasCount: 2
+    asserts:
+      - matchRegexRaw:
+          pattern: "ClickHouse is configured with 2 replicas"
+      - matchRegexRaw:
+          pattern: "two-phase install for High Availability"
+      - matchRegexRaw:
+          pattern: "clickhouse\\.replicasCount: 1"
+
+  - it: NOTES omits the banner when ClickHouse is disabled even with multiple replicas
+    set:
+      clickhouse.enabled: false
+      clickhouse.replicasCount: 2
+    asserts:
+      - notMatchRegexRaw:
+          pattern: "two-phase install"
--- a/deployment/helm_chart/opik/tests/clickhouse_migration_test.yaml
+++ b/deployment/helm_chart/opik/tests/clickhouse_migration_test.yaml
@@ -8,7 +8,14 @@ templates:
   - templates/NOTES.txt
 
 tests:
-  - it: NOTES omits the multi-replica migration banner for the single-replica default
+  - it: NOTES omits the multi-replica migration banner for the default replica count
+    asserts:
+      - notMatchRegexRaw:
+          pattern: "two-phase install"
+
+  - it: NOTES omits the banner at an explicit single replica
+    set:
+      clickhouse.replicasCount: 1
     asserts:
       - notMatchRegexRaw:
           pattern: "two-phase install"
--- a/apps/opik-documentation/documentation/fern/docs-v2/self-host/scaling.mdx
+++ b/apps/opik-documentation/documentation/fern/docs-v2/self-host/scaling.mdx
@@ -128,6 +128,18 @@ Memory-optimized instances are recommended, with a minimum 4:1 memory-to-CPU rat
 
 Always scale vertically before adding more replicas for efficiency.
 
+#### Fresh multi-replica installs: migrate first, then scale
+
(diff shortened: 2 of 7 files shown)


deployment/helm_chart/opik/tests/clickhouse_migration_test.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+# When ClickHouse runs >= 2 replicas, NOTES.txt reminds operators to use the two-phase
+# HA install flow (migrate at 1 replica, then scale up), since a fresh multi-replica
+# install cannot complete its migrations (OPIK-7332).
+
+templates:
+  - templates/NOTES.txt
+
+tests:
+  - it: NOTES omits the multi-replica migration banner for the single-replica default
```
JetoPistola: 🧹 **nit** | Testing

This test relies on the `values.yaml` default (`replicasCount: 1`) rather than setting it explicitly. It's valid as-is, but an explicit `set` would make the test self-documenting and immune to a future change of the default:

```suggestion
  - it: NOTES omits the multi-replica migration banner for the single-replica default
    set:
      clickhouse.replicasCount: 1
    asserts:
```

Entirely optional.

JetoPistola:
👋 **Review summary**

**What looks good**
- Well-scoped: no application code touched, and it avoids the trap of mutating immutable migrations — documenting the supported two-phase HA flow instead.
- The cause analysis is technically accurate — I confirmed the earliest analytics migrations (`000001`+) genuinely lack `ON CLUSTER '{cluster}'`, which is exactly the root cause described.
- Thorough surface coverage: NOTES banner (conditional and interpolated), `values.yaml` guidance, plus both docs versions (v1 + v2) with parity — the `/v1/...` vs `/...` link paths are correctly split per tree, and the same-page `#clickhouse-zookeeper-metadata-loss` anchor resolves.
- The NOTES conditional is robust (`int … | default 1` guards string/unset cases), and the new helm-unittest suite covers the three meaningful states (default off, `>=2` on, disabled).

**Overall**
Solid documentation/operability fix for a real install-time failure mode. The approach — signpost the safe two-phase flow rather than alter migration behavior — is the right call given the immutable-migration constraint. No structural concerns.

**Inline comments**: 1 nit (optional test-explicitness suggestion) — nothing blocking.

deployment/helm_chart/opik/tests/clickhouse_migration_test.yaml
andrescrz: Fixed in 1c88c61 — added an explicit `clickhouse.replicasCount: 1` test for the conditional's lower boundary, so it no longer depends on the default and stays correct if the default ever changes.

I kept the no-override test too (renamed to "default replica count") rather than replacing it: the two guard different things. The explicit-1 case pins the boundary; the no-override case is a canary that the *shipped default* stays single-replica (banner off) — if someone flips the default to `2`, that test fails and forces a conscious decision. Thanks for the nudge!

JetoPistola:
✅ **Re-review — feedback addressed**

Re-reviewed at `edef470`. The three follow-up commits resolve the earlier review notes:

- **Test coverage** — the two-phase NOTES suite now includes an explicit `clickhouse.replicasCount: 1` case alongside the default, so the single-replica "banner absent" behavior is no longer relying on the values default. 👍
- **Steps formatting** — the two-phase install is now a Fern `<Steps>` component. I verified the **live preview render** on both trees: the section emits `fern-steps`/`fern-step` markup (no escaped-tag leakage), and the cross-links resolve correctly per tree (v1 → `/docs/opik/v1/self-host/troubleshooting`, v2 → `/docs/opik/self-host/troubleshooting`).
- **Configurable DB name** — the `ANALYTICS_DB_DATABASE_NAME` clarification is accurate (real env var, defaults to `opik`).
- **Greenfield ZK recovery** — the refined warning correctly distinguishes fresh-install path cleanup from the full data-recovery rebuild.

v1 and v2 troubleshooting sections are byte-identical. No new findings — nothing blocking. This still isn't a formal approval; a human reviewer should approve.

JetoPistola (approved):
Approving — docs + Helm change is clean and well-scoped. Root-cause analysis is accurate (early analytics migrations genuinely lack `ON CLUSTER`), v1/v2 doc parity holds, and the two-phase Steps + callouts render correctly on the Fern preview for both trees. Review feedback was addressed (explicit single-replica test case, configurable DB-name note, refined ZK-recovery guidance). No application code touched.

🤖 Approved with assistance from Claude Code

PR #7532, opened 2026-07-20
[OPIK-7396] [BE] fix: tolerate >1 emitted row in TraceDAO.findById (get-by-id 500)

thiagohora:
## Details

`GET /v1/private/traces/{id}` intermittently returned 500 with `IndexOutOfBoundsException: "Source emitted more than one item"`, and — once the reducer was made lenient — could return a non-deterministic `experiment` field. Root cause: in `SELECT_BY_IDS`, the outer `LEFT JOIN experiments_agg eaag ON eaag.trace_id = t.id` had no per-trace dedup, so a trace that belongs to N experiments fanned out into N rows (differing `experiment_id/name/dataset_id`).

- **Fix at the source:** `experiments_agg` now collapses to a single canonical row per `trace_id` — the most recent experiment (`ORDER BY trace_id, experiment_id DESC` + `LIMIT 1 BY trace_id`, ids are UUIDv7 / time-ordered). The trace-by-id join therefore returns exactly one row. This also stops batch `getByIds` returning duplicate rows for multi-experiment traces. `Trace.experiment` is a single field, so collapsing loses nothing at the API level.
- **Defense in depth:** `findById` reduces via `collectList()` + first and logs a WARN (trace id + workspace) if a fan-out ever reappears, so a future regression degrades to a log + one row instead of a 500.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves OPIK-7396

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: Production diagnosis, the query-level collapse in `SELECT_BY_IDS`, the defensive reducer, and the unit test.
- Human verification: Author reviewed the diff, confirmed the query change against the fan-out analysis, and confirmed the test run below.

## Testing

- `mvn -o test-compile` + `spotless:apply` clean.
- Unit test `TraceDAOImplTest` on the reducer (no DB): no rows → empty; single → that; >1 → first, no `IndexOutOfBoundsException`. `Tests run: 3, Failures: 0`.
- The query-level collapse is DB-backed; a regression test (a trace in ≥2 experiments → `GET /{id}` returns 200 with a deterministic experiment) belongs in the traces integration suite (CI). Flagged on the ticket.

## Documentation

No documentation changes — internal reliability/correctness fix; no API shape change (`Trace.experiment` remains a single value).

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
@@ -3451,11 +3451,18 @@ class TraceDAOImpl implements TraceDAO {
     @WithSpan
     public Mono<Trace> findById(@NonNull UUID id, @NonNull Connection connection) {
         // A single id can map to more than one emitted Trace when the assembled query fans out
-        // over its join CTEs (e.g. un-merged/duplicated rows). Take the first emission rather than
+        // over its join CTEs (e.g. un-merged/duplicated rows). Take the first row rather than
         // singleOrEmpty(), which throws IndexOutOfBoundsException ("Source emitted more than one
-        // item") and surfaces as a 500 on GET by id. next() preserves the empty case (-> 404).
+        // item") and surfaces as a 500 on GET by id. The empty case is preserved (-> 404).
+        // Log when it happens so the underlying duplication/fan-out stays visible.
         return findByIds(List.of(id), connection)
-                .next();
+                .collectList()
+                .flatMap(traces -> {
+                    if (traces.size() > 1) {
+                        log.warn("Trace id '{}' resolved to '{}' rows; returning the first", id, traces.size());
+                    }
+                    return traces.isEmpty() ? Mono.empty() : Mono.just(traces.get(0));
+                });
     }
 
     @Override
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
@@ -3457,12 +3457,13 @@ class TraceDAOImpl implements TraceDAO {
         // Log when it happens so the underlying duplication/fan-out stays visible.
         return findByIds(List.of(id), connection)
                 .collectList()
-                .flatMap(traces -> {
+                .flatMap(traces -> Mono.deferContextual(ctx -> {
                     if (traces.size() > 1) {
-                        log.warn("Trace id '{}' resolved to '{}' rows; returning the first", id, traces.size());
+                        log.warn("Trace id '{}' in workspace '{}' resolved to '{}' rows; returning the first",
+                                id, ctx.getOrDefault(RequestContext.WORKSPACE_ID, "unknown"), traces.size());
                     }
-                    return traces.isEmpty() ? Mono.empty() : Mono.just(traces.get(0));
-                });
+                    return traces.isEmpty() ? Mono.<Trace>empty() : Mono.just(traces.get(0));
+                }));
     }
 
     @Override
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
@@ -468,6 +468,11 @@ class TraceDAOImpl implements TraceDAO {
      * Key format: {@code author} + optional {@code _queueId} + optional {@code _spanId} (ensures uniqueness across
      * queues/spans). Value tuple: (value, reason, category_name, source, last_updated_at, span_type, span_id,
      * source_queue_id, author).
+     *
+     * <p>experiments_agg collapses to a single canonical row per trace_id (the most recent experiment,
+     * by UUIDv7-ordered id) so the trace-by-id LEFT JOIN cannot fan a trace that belongs to multiple
+     * experiments into multiple rows — which surfaced as an IndexOutOfBoundsException 500 on GET by id
+     * and non-deterministic experiment metadata (OPIK-7396).
      */
     private static final String SELECT_BY_IDS = """
             WITH target_spans AS (
@@ -678,7 +683,7 @@ class TraceDAOImpl implements TraceDAO {
                 AND trace_id IN :ids
                 GROUP BY trace_id
             ), experiments_agg AS (
-                SELECT DISTINCT
+                SELECT
                     ei.trace_id,
                     if(div.id != '', div.dataset_item_id, ei.dataset_item_id) AS experiment_dataset_item_id,
                     e.id AS experiment_id,
@@ -706,6 +711,8 @@ class TraceDAOImpl implements TraceDAO {
                     ORDER BY (workspace_id, dataset_id, id) DESC, last_updated_at DESC
                     LIMIT 1 BY id
                 ) e ON ei.experiment_id = e.id
+                ORDER BY trace_id, experiment_id DESC
+                LIMIT 1 BY trace_id
             )
             SELECT
                 t.*,
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
@@ -3450,20 +3450,25 @@ class TraceDAOImpl implements TraceDAO {
     @Override
     @WithSpan
     public Mono<Trace> findById(@NonNull UUID id, @NonNull Connection connection) {
-        // A single id can map to more than one emitted Trace when the assembled query fans out
-        // over its join CTEs (e.g. un-merged/duplicated rows). Take the first row rather than
-        // singleOrEmpty(), which throws IndexOutOfBoundsException ("Source emitted more than one
-        // item") and surfaces as a 500 on GET by id. The empty case is preserved (-> 404).
-        // Log when it happens so the underlying duplication/fan-out stays visible.
         return findByIds(List.of(id), connection)
                 .collectList()
-                .flatMap(traces -> Mono.deferContextual(ctx -> {
-                    if (traces.size() > 1) {
-                        log.warn("Trace id '{}' in workspace '{}' resolved to '{}' rows; returning the first",
-                                id, ctx.getOrDefault(RequestContext.WORKSPACE_ID, "unknown"), traces.size());
-                    }
-                    return traces.isEmpty() ? Mono.<Trace>empty() : Mono.just(traces.get(0));
-                }));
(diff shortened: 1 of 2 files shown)


apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -3451,7 +3451,24 @@ private Mono<List<UUID>> getTargetProjectIdsForTraces(List<UUID> ids) {
     @WithSpan
     public Mono<Trace> findById(@NonNull UUID id, @NonNull Connection connection) {
         return findByIds(List.of(id), connection)
-                .singleOrEmpty();
+                .collectList()
+                .flatMap(traces -> Mono.deferContextual(ctx -> Mono.justOrEmpty(
+                        firstOrLogFanOut(traces, id, ctx.getOrDefault(RequestContext.WORKSPACE_ID, "unknown")))));
```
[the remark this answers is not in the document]
thiagohora: Correction after actually verifying the query: `experiments_agg` **can** fan out. The outer SELECT of `SELECT_BY_IDS` does `LEFT JOIN experiments_agg eaag ON eaag.trace_id = t.id` with **no** outer `GROUP BY`/`LIMIT 1 BY t.id` (the `LIMIT 1 BY id`s are all inside subqueries), so a trace in N experiments yields N rows with differing `experiment_id/name/dataset_id`. So your point stands — `findFirst()` is nondeterministic, and the correct fix is at the query level (collapse `experiments_agg` to one canonical experiment row per trace id). Tracking on OPIK-7396; the reducer WARN stays as a detector. Sorry for the earlier hand-wave.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -3451,7 +3451,24 @@ private Mono<List<UUID>> getTargetProjectIdsForTraces(List<UUID> ids) {
...
+     * returning the first row avoids the {@code IndexOutOfBoundsException} ("Source emitted more
+     * than one item") that a strict single-item reducer throws, which surfaces as a 500. The empty
+     * case is preserved (-> 404). A fan-out is logged so the underlying duplication stays visible.
+     */
+    static Optional<Trace> firstOrLogFanOut(@NonNull List<Trace> traces, UUID id, String workspaceId) {
+        if (traces.size() > 1) {
+            log.warn("Trace id '{}' in workspace '{}' resolved to '{}' rows; returning the first",
+                    id, workspaceId, traces.size());
+        }
+        return traces.stream().findFirst();
```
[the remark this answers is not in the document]
thiagohora: Correction: verified — `experiments_agg` can produce multiple differing-experiment rows per trace id (outer join has no `GROUP BY`/`LIMIT 1 BY t.id`). So this is a real query-level issue, fixed at the source by collapsing to one canonical experiment row per trace id in `SELECT_BY_IDS`. Tracked on OPIK-7396.

apps/opik-backend/src/test/java/com/comet/opik/domain/TraceDAOImplTest.java
```diff
@@ -0,0 +1,37 @@
...
+        assertThat(TraceDAOImpl.firstOrLogFanOut(List.of(), UUID.randomUUID(), "ws")).isEmpty();
+    }
+
+    @Test
+    void firstOrLogFanOut__whenSingleRow__returnsIt() {
+        var id = UUID.randomUUID();
+        var trace = Trace.builder().id(id).build();
+
+        assertThat(TraceDAOImpl.firstOrLogFanOut(List.of(trace), id, "ws")).containsSame(trace);
+    }
```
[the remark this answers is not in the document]
thiagohora: The extracted reducer (`firstOrLogFanOut`) is unit-tested directly (empty / single / >1). The full reactive `findById` path (collectList → deferContextual → switchIfEmpty) is DB-bound with no unit seam; a get-by-id regression test belongs in the traces integration suite (Testcontainers, runs in CI). Noted on the ticket AC; happy to add it there.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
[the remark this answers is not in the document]
thiagohora: Fixed at the source in 409a264: `experiments_agg` in `SELECT_BY_IDS` now collapses to one canonical row per `trace_id` (most recent experiment, `ORDER BY trace_id, experiment_id DESC` + `LIMIT 1 BY trace_id`), so the trace-by-id LEFT JOIN can no longer fan a trace into multiple rows — deterministic experiment, no unbounded buffer, and it also stops batch `getByIds` returning duplicate rows for multi-experiment traces. The `findById` collectList/first + WARN stays only as a defensive detector. Thanks for pushing on this — the fan-out was real, not the identical-dupe case I first assumed.

JetoPistola:
❓ **question** | Logical Bugs — adjacent query (out of scope for this PR)

Posting as a general comment since the referenced lines aren't in this PR's diff.

The sibling **list** query `SELECT_BY_PROJECT_ID` looks like it has the same `experiments_agg` fan-out this PR just fixed in `SELECT_BY_IDS`:

- `experiments_agg AS (SELECT DISTINCT ...)` at `TraceDAO.java:1260` has **no** `LIMIT 1 BY trace_id` collapse (unlike the fixed one, which now does at line 715).
- It's joined in the **final** SELECT via `LEFT JOIN experiments_agg eaag ON eaag.trace_id = t.id` (`TraceDAO.java:1459`), and that final SELECT ends with `ORDER BY ...` + `SETTINGS ...` — **no** outer `LIMIT 1 BY id`. The `LIMIT 1 BY id` at line 1428 is inside the `page_wide` CTE, which is then re-joined to `experiments_agg` afterward.

So a trace belonging to ≥2 experiments would seem to fan the list result into duplicate trace rows there too — the same class of bug, minus the 500 (the list reducer isn't strict-single).

This PR is correctly scoped to the get-by-id 500, so **not a blocker here**. Is the list-query fan-out already known/tracked, or worth a follow-up ticket? (Matches the recurring CTE fan-out pattern.)

thiagohora:
Good catch, and correct. Verified: `SELECT_BY_PROJECT_ID`s `experiments_agg` has no `LIMIT 1 BY trace_id`, and its final SELECT ends `ORDER BY … SETTINGS …` with no outer `LIMIT 1 BY id` (the `LIMIT 1 BY id` is inside the `page_wide` CTE, which is then re-joined to `experiments_agg`). So a trace in ≥2 experiments fans the list into duplicate rows — same class as this PR, minus the 500, and only when the experiment join is active (`sort_has_experiment || !exclude_experiment`).

Agreed it is out of scope here (this PR is the get-by-id 500). It was not yet tracked — filed a follow-up: **OPIK-7416**, with the same collapse (`LIMIT 1 BY trace_id`, most-recent experiment) as the fix. Thanks!

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -3451,7 +3458,24 @@ private Mono<List<UUID>> getTargetProjectIdsForTraces(List<UUID> ids) {
     @WithSpan
     public Mono<Trace> findById(@NonNull UUID id, @NonNull Connection connection) {
         return findByIds(List.of(id), connection)
-                .singleOrEmpty();
+                .collectList()
```
andrescrz, line 3462: Minor: for checking `size > 1` and logging and checking the first one, there might be an easy way to avoid collecting all results (waits and accumulates in memory). Ok to keep this if the alternative is too complicated and not a clean implementation. Results should be small anyway.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -3451,7 +3458,24 @@ private Mono<List<UUID>> getTargetProjectIdsForTraces(List<UUID> ids) {
...
+    }
+
+    /**
+     * Reduce the rows returned for a single trace id to at most one {@link Trace}. A get-by-id
+     * query can fan out to more than one row over its join CTEs (e.g. un-merged/duplicated rows);
+     * returning the first row avoids the {@code IndexOutOfBoundsException} ("Source emitted more
+     * than one item") that a strict single-item reducer throws, which surfaces as a 500. The empty
+     * case is preserved (-> 404). A fan-out is logged so the underlying duplication stays visible.
+     */
+    static Optional<Trace> firstOrLogFanOut(@NonNull List<Trace> traces, UUID id, String workspaceId) {
```
andrescrz, line 3475: Minor: annotate as visible for testing (from Guava)

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -3451,7 +3458,24 @@ private Mono<List<UUID>> getTargetProjectIdsForTraces(List<UUID> ids) {
...
+        if (traces.size() > 1) {
+            log.warn("Trace id '{}' in workspace '{}' resolved to '{}' rows; returning the first",
```
andrescrz: Nit: better move printed stuff on logs to the end of the sentence (easier to grep when debugging and service convention).

andrescrz (dismissed):
Just minor recommendations, otherwise LGTM.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
thiagohora: Keeping `collectList()` here. Now that `SELECT_BY_IDS` collapses `experiments_agg` to one row per `trace_id`, this buffers at most one row (empty on miss), so there is nothing to accumulate in practice — the reducer stays only as a defensive fan-out detector. Agree the alternative (bounded reactive reduce) is not worth the added complexity here.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
thiagohora: Done in df90754 — annotated `firstOrLogFanOut` with `@VisibleForTesting`.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
thiagohora: Done in df90754 — moved the values to the end: `"Trace get-by-id resolved to multiple rows; returning the first. workspaceId '{}', traceId '{}'"`.

andrescrz (approved):

PR #7549, opened 2026-07-21
[NA] [FE] Add strict Cost-intelligence rounding mode to formatNumberInK

aadereiko:
## Details

Adds an opt-in `strict` option to `formatNumberInK` (`src/lib/utils.ts`) that applies the Cost-intelligence rounding rules: always `precision` decimals on suffixed values with trailing zeros kept (`7.0M`), an integer with no decimals below 1000 (`843`), round-half-up, carry-over to the next unit when rounding hits 1000 (`999,950 → 1.0M`), and a leading sign on negatives (`-6.9K`). Default behavior is unchanged (`strict` defaults to `false`), so existing callers are unaffected. This lets the AI-Spend plugin's summary reuse the shared helper instead of maintaining a second, divergent K/M/B implementation (per review feedback on the ai-spend PR).

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- N/A — no ticket; small shared-helper enhancement supporting the AI-Spend billing-split work.

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Scope: implementation of the `strict` option + unit tests
- Human verification: reviewed the diff; ran the test suite and lint/type checks locally (below)

## Testing

- `npx vitest run src/lib/utils.test.ts` — 60 passed (regression case for existing non-strict behavior + full strict-mode coverage: thresholds, trailing `.0`, half-up, both carry-over cases, sign).
- `npx tsc --noEmit -p tsconfig.json` — clean.
- `npx eslint src/lib/utils.ts src/lib/utils.test.ts` — clean.

## Documentation

No docs change — internal utility; behavior documented via a code comment on the `strict` branch and the unit tests.

### DIFF
--- a/apps/opik-frontend/src/lib/utils.test.ts
+++ b/apps/opik-frontend/src/lib/utils.test.ts
@@ -4,6 +4,7 @@ import {
   isStringMarkdown,
   removeUndefinedKeys,
   isLooseEqual,
+  formatNumberInK,
 } from "./utils";
 
 describe("isStringMarkdown", () => {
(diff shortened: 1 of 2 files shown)


LifeXplorer (approved):
Looks good!

aadereiko:
Closing — we decided to keep the Cost-intelligence rounding helper local to the ai-spend plugin (`formatCompactNumber` in `opik-plugin-ai-spend`) rather than extending the shared `formatNumberInK`, to avoid changing an app-wide base util for a plugin-specific need. No change to `src/lib/utils.ts` after all.

PR #7553, opened 2026-07-21
[OPIK-7352] [BE] reject non-UUIDv7 referenced ids on ingest

thiagohora:
## Details

Retention sweeps spans/traces via an id-range predicate that assumes the *referenced* id is a time-ordered UUIDv7 (`SpanDAO` orders spans by the `trace_id` range; `TraceDAO` by the `id` range). Ingest validated an entity's own id but not the ids it references, so a client could store e.g. a span with a non-UUIDv7 `trace_id` and have those rows ordered lexicographically instead of chronologically — deleted or retained against the wrong window. This validates referenced ids on ingest.

- **Scope rule**: validate a referenced id **iff the operation persists it to the entity's table**. Ids used only to *resolve/locate* an existing association (e.g. a `projectId` on thread open/close, whose `project_id` was written at trace ingestion) or as a *read source* (dataset copy-from ids) are not validated.
- Referenced ids use a **not-in-future** policy: version 7 required, future-dated rejected, **past allowed** (referencing an older entity — a late span on an old trace, a score on an old trace — is legitimate and common). Renames `validateIdForUpdate` → `validateIdNotInFuture` and reuses it for both the update paths and referenced ids; adds null-safe `validateIdNotInFutureIfPresent(Async)` helpers.
- **Validated** (each written to a column by its create/update): span `traceId`/`parentSpanId` (create, batch, update, batch-update, experiment-items bulk); feedback-score entity ids (single + batch) + `sourceQueueId`; trace/span/thread comment entity ids; attachment entity ids; guardrail `entityId` + `secondaryId`; assertion `entityId`; dataset-item `traceId`/`spanId` + batch `datasetId`; annotation-queue item ids; experiment `projectId`/`optimizationId`/`datasetVersionId`; and `projectId` on span/trace update + batch-update, optimization, alert, dashboard, prompt (+ version), annotation-queue, guardrail, assertion, feedback batch, automation-rule evaluators.
- Version enforcement is always on (mirrors own-id validation); the not-in-future check is gated by the existing `uuidValidation` kill-switch (off in prod), so this ships dormant beyond the version check.
- Out of scope / tracked separately: some clients emit v7 UUIDs whose embedded timestamp is far in the future; that pre-existing client-side issue is not remediated here for existing rows.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- OPIK-7352

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: implementation, prod data verification, and tests

## Testing

- `mvn -o compile` / `test-compile` / `spotless:apply` clean.
- Added `SpansResourceTest` cases (single **and** batch create): reject non-v7 `traceId` (400), reject future-dated `traceId` (400), reject non-v7 `parentSpanId` (400), and **accept an old (past) v7 `traceId`** (spans on old traces are valid). These exercise the shared `validateIdNotInFuture` mechanism used by every path here.
- Resource integration tests do not run locally (ClickHouse healthcheck bootstrap quirk) — relying on CI.
- Verified via read-only production queries that existing rows carry no non-v7 referenced ids on the validated paths, so enforcement rejects no legitimate existing data.
- Follow-up: dedicated per-path tests for the non-span validated paths (they share the span-tested mechanism).

## Documentation

No user-facing docs. The referenced-id validation policy (version 7 + not-in-future, past allowed) is documented in the `IdGenerator` javadoc.


apps/opik-backend/src/main/java/com/comet/opik/domain/AssertionResultService.java
```diff
@@ -63,7 +64,10 @@ public Mono<Void> saveBatch(@NonNull EntityType entityType,
         }
 
         // Validate up front so a bad id fails fast and independently of project-name normalisation.
-        assertionResults.forEach(item -> IdGenerator.validateVersion(item.entityId(), entityType.getType()));
+        assertionResults.forEach(item -> {
+            idGenerator.validateIdNotInFuture(item.entityId(), entityType.getType());
+            idGenerator.validateIdNotInFutureIfPresent(item.projectId(), "project");
+        });
```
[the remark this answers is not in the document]
thiagohora: Removed this validation. `insertBatch` resolves the project by name and overwrites `projectId` with the resolved `project.id()` before persistence, so the client-supplied value is never stored — validating it only rejected arbitrary/placeholder ids with no benefit. `entityId` (which is persisted) stays validated. Fixed in f1e48d3.

apps/opik-backend/src/main/java/com/comet/opik/domain/GuardrailsService.java
```diff
@@ -51,7 +51,9 @@ public Mono<Void> addTraceGuardrails(List<Guardrail> guardrails) {
                 .stream()
                 .map(guardrail -> {
                     UUID id = idGenerator.generateId();
-                    IdGenerator.validateVersion(guardrail.entityId(), entityType.getType()); // validate trace id
+                    idGenerator.validateIdNotInFuture(guardrail.entityId(), entityType.getType());
+                    idGenerator.validateIdNotInFuture(guardrail.secondaryId(), "guardrail secondary");
+                    idGenerator.validateIdNotInFutureIfPresent(guardrail.projectId(), "project");
```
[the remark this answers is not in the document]
thiagohora: Removed. `addTraceGuardrails` resolves the project by name and overwrites `projectId` with `project.id()`, so the client input is discarded — validating it was pointless and breaking. `entityId`/`secondaryId` (persisted verbatim) remain validated. Fixed in f1e48d3.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/SpansResourceTest.java
```diff
@@ -1800,6 +1818,46 @@ void createWithInvalidIdThrowsBadRequest(UUID id, String expectedDetails, String
...
+        void createWithOldTraceIdSucceeds() {
+            var old = Instant.now().minus(Duration.ofDays(30)).toEpochMilli();
+            var span = podamFactory.manufacturePojo(Span.class).toBuilder()
+                    .traceId(generator.construct(old))
+                    .parentSpanId(null)
+                    .build();
+
+            var id = spanResourceClient.createSpan(span, API_KEY, TEST_WORKSPACE);
+
+            assertThat(id).isNotNull();
```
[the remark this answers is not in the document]
thiagohora: Strengthened: the test now round-trips the created span and asserts the old `traceId` is persisted verbatim and `parentSpanId` stays null. Fixed in f1e48d3.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/SpansResourceTest.java
```diff
@@ -2292,6 +2350,20 @@ void batchCreateWithInvalidIdThrowsBadRequest(UUID id, String expectedDetails, S
...
 
+        @MethodSource("com.comet.opik.api.resources.v1.priv.SpansResourceTest#invalidTraceIds")
+        @ParameterizedTest(name = "Batch create span with invalid traceId throws bad request: {2}")
+        void batchCreateWithInvalidTraceIdThrowsBadRequest(UUID traceId, String expectedDetails, String testName) {
+            var expectedEntity = new io.dropwizard.jersey.errors.ErrorMessage(
+                    HttpStatus.SC_BAD_REQUEST, "Invalid UUID for id", expectedDetails);
+            var span = podamFactory.manufacturePojo(Span.class).toBuilder().traceId(traceId).build();
+            try (var response = spanResourceClient.callBatchCreateSpans(
+                    List.of(span), API_KEY, TEST_WORKSPACE)) {
+                assertThat(response.getStatusInfo().getStatusCode()).isEqualTo(HttpStatus.SC_BAD_REQUEST);
```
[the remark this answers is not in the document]
thiagohora: Added `batchCreateWithNonV7ParentSpanIdThrowsBadRequest` covering a non-v7 `parentSpanId` on batch create. Fixed in f1e48d3.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanService.java
```diff
@@ -159,6 +161,8 @@ public Mono<UUID> create(@NonNull Span span) {
         var projectName = WorkspaceUtils.getProjectName(span.projectName());
         return idGenerator
                 .validateIdAsync(id, SPAN_KEY)
+                .then(idGenerator.validateIdNotInFutureAsync(span.traceId(), SPAN_TRACE_KEY))
+                .then(idGenerator.validateIdNotInFutureIfPresentAsync(span.parentSpanId(), SPAN_PARENT_KEY))
                 .then(projectService.getOrCreate(projectName))
```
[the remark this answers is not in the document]
thiagohora: Extracted a shared `validateSpanReferences(traceId, parentSpanId)` helper, now used by all span write paths (single/batch create, single/batch update). Fixed in 72118b6.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanService.java
```diff
@@ -460,6 +471,8 @@ private List<Span> bindSpanToProjectAndId(List<Span> spans, List<Project> projec
 
                     UUID id = span.id() == null ? idGenerator.generateId() : span.id();
                     idGenerator.validateId(id, SPAN_KEY);
+                    idGenerator.validateIdNotInFuture(span.traceId(), SPAN_TRACE_KEY);
+                    idGenerator.validateIdNotInFutureIfPresent(span.parentSpanId(), SPAN_PARENT_KEY);
```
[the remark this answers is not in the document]
thiagohora: Good catch — fixed. `create(SpanBatch)` now validates all span ids up front, before `deleteAutoStrippedAttachments`/project `getOrCreate`, so a rejected batch causes no side effects. Fixed in 72118b6.

apps/opik-backend/src/main/java/com/comet/opik/domain/IdGenerator.java
```diff
@@ -34,12 +34,33 @@ public interface IdGenerator {
...
-    Mono<UUID> validateIdForUpdateAsync(UUID id, String resource);
+    void validateIdNotInFuture(UUID id, String resource);
+
+    Mono<UUID> validateIdNotInFutureAsync(UUID id, String resource);
+
+    /**
+     * Null-safe variant of {@link #validateIdNotInFuture} for optional referenced ids (e.g. an optional
+     * {@code projectId} that may be resolved by name instead). No-op when {@code id} is null.
+     */
+    default void validateIdNotInFutureIfPresent(UUID id, String resource) {
```
andrescrz: Minor: for these two new `validateIdNotInFutureIfPresent` and `validateIdNotInFutureIfPresentAsync` methods, better move the actual implementation from the interface the implementation class. 

Let's try to keep interfaces free of implementation details.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleEvaluatorService.java
```diff
@@ -117,6 +117,7 @@ public <E, F extends Filter, T extends AutomationRuleEvaluator<E, F>> T save(@No
 
         UUID id = idGenerator.generateId();
         IdGenerator.validateVersion(id, "AutomationRuleEvaluator");
+        projectIds.forEach(projectId -> idGenerator.validateIdNotInFutureIfPresent(projectId, "project"));
```
andrescrz, line 122: Minor: I believe project ID is never externally generated, always ingested by name with the ID created internally. That makes these validations redundant, as they would only create orphans. For  MySQL based entities like automation rules, we could do a foreign key check for non-concurrent entities (on the service side, not at schema level).

apps/opik-backend/src/main/java/com/comet/opik/domain/ExperimentService.java
```diff
@@ -495,6 +495,9 @@ private Set<UUID> getPromptVersionIds(Experiment experiment) {
     public Mono<UUID> create(@NonNull Experiment experiment) {
         var id = experiment.id() == null ? idGenerator.generateId() : experiment.id();
         IdGenerator.validateVersion(id, "Experiment");
+        idGenerator.validateIdNotInFutureIfPresent(experiment.projectId(), "project");
+        idGenerator.validateIdNotInFutureIfPresent(experiment.optimizationId(), "optimization");
+        idGenerator.validateIdNotInFutureIfPresent(experiment.datasetVersionId(), "dataset version");
```
andrescrz: Similarly: I believe the dataset version ID is internally generated, although not 100% sure about this.

apps/opik-backend/src/main/java/com/comet/opik/domain/IdGenerator.java
andrescrz, line 1: I'd personally double check what are the remaining usages in the service where only the UUID version is checked (`validateVersion` and `validateVersionAsync`), to evaluate the remaining usages and deciding if we keep them or not.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/AlertResourceTest.java
```diff
@@ -169,6 +171,7 @@ class AlertResourceTest {
     }
 
     private final PodamFactory factory = PodamFactoryUtils.newPodamFactory();
+    private final TimeBasedEpochGenerator generator = Generators.timeBasedEpochGenerator();
```
andrescrz: The idea for tests is using TestIdGeneratorFactory, so we keep this inner generator encapsulated and instead route all the traffic through IdGenerator.

In addition, it's thread safe, so it can be static final.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
```diff
@@ -2819,6 +2819,9 @@ void createFromTraces__whenRegularDataset__thenDataContainsAllEnrichedFields() {
     @TestInstance(TestInstance.Lifecycle.PER_CLASS)
     class ExperimentDatasetVersionLinking {
 
+        private final com.fasterxml.uuid.impl.TimeBasedEpochGenerator generator = com.fasterxml.uuid.Generators
```
andrescrz: Nit: use imports.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/SpansResourceTest.java
```diff
@@ -1800,6 +1818,52 @@ void createWithInvalidIdThrowsBadRequest(UUID id, String expectedDetails, String
...
+            try (var response = spanResourceClient.createSpan(
+                    span, API_KEY, TEST_WORKSPACE, HttpStatus.SC_BAD_REQUEST)) {
+                var actualEntity = response.readEntity(io.dropwizard.jersey.errors.ErrorMessage.class);
+                assertThat(actualEntity).isEqualTo(expectedEntity);
+            }
+        }
+
+        @Test
+        @DisplayName("Create span with non-v7 parentSpanId throws bad request")
+        void createWithNonV7ParentSpanIdThrowsBadRequest() {
```
andrescrz: Minor: test similar to the previous one. You could parameterise the whole span that it's built and avoid repeating.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/SpansResourceTest.java
```diff
@@ -2292,6 +2356,34 @@ void batchCreateWithInvalidIdThrowsBadRequest(UUID id, String expectedDetails, S
...
+                    List.of(span), API_KEY, TEST_WORKSPACE)) {
+                assertThat(response.getStatusInfo().getStatusCode()).isEqualTo(HttpStatus.SC_BAD_REQUEST);
+                var actualEntity = response.readEntity(io.dropwizard.jersey.errors.ErrorMessage.class);
+                assertThat(actualEntity).isEqualTo(expectedEntity);
+            }
+        }
+
+        @Test
+        @DisplayName("Batch create span with non-v7 parentSpanId throws bad request")
+        void batchCreateWithNonV7ParentSpanIdThrowsBadRequest() {
```
andrescrz: Minor: same as for spans.

andrescrz (dismissed):
Some comments, but no blockers.

apps/opik-backend/src/main/java/com/comet/opik/domain/IdGenerator.java
thiagohora: Done — moved both `validateIdNotInFutureIfPresent`/`Async` bodies out of the interface into `IdGeneratorImpl`. (76137a6)

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleEvaluatorService.java
thiagohora: Good point on the redundancy. I kept the v7 check here specifically because automation `projectIds` are persisted **without** an existence check, so dropping it would let a UUIDv4 orphan reference through. Where existence is already verified (project resolved via `validateProjectIdExists`/`resolveProjectIdOrCreate`/`get` — prompt, dashboard, optimization, experiment, span/trace update) I removed the now-redundant v7 check. A service-side FK existence check (as you suggest) would be the cleaner long-term guard — proposing it as a follow-up. (76137a6)

apps/opik-backend/src/main/java/com/comet/opik/domain/ExperimentService.java
thiagohora: Removed — `datasetVersionId` is FK-checked (the `whenInvalidVersionId → 409` path) and `@Schema(READ_ONLY)`/internally generated, so the v7 check was redundant and pre-empted the 409. (76137a6)

apps/opik-backend/src/main/java/com/comet/opik/domain/IdGenerator.java
thiagohora: Audited: the remaining `validateVersion`/`validateVersionAsync` calls are all pre-existing own-id checks on create (project, dataset, prompt, alert, experiment-item, …) — this PR added none, so I left them unchanged. The referenced-id checks I added now split by whether the target is existence-checked (removed as redundant) or not (kept, to block v4 orphans). (76137a6)

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/AlertResourceTest.java
thiagohora: Done — the added generators now route through `TestIdGeneratorFactory.create()` (static final) across the affected test classes. (76137a6)

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
thiagohora: Done — replaced the fully-qualified name with imports (subsumed by the TestIdGeneratorFactory switch). (76137a6)

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/SpansResourceTest.java
thiagohora: Done — merged the trace/parent single-create cases into one parameterized `createWithInvalidReferencedIdThrowsBadRequest` over a shared provider. (76137a6)

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/SpansResourceTest.java
thiagohora: Done — same for the batch path (`batchCreateWithInvalidReferencedIdThrowsBadRequest`). (76137a6)

apps/opik-backend/src/main/java/com/comet/opik/domain/IdGenerator.java
```diff
@@ -34,12 +34,27 @@ public interface IdGenerator {
...
+
+    Mono<UUID> validateIdNotInFutureAsync(UUID id, String resource);
+
+    /**
+     * Null-safe variant of {@link #validateIdNotInFuture} for optional referenced ids (e.g. an optional
+     * {@code projectId} that may be resolved by name instead). No-op when {@code id} is null.
+     */
+    void validateIdNotInFutureIfPresent(UUID id, String resource);
+
+    Mono<UUID> validateIdNotInFutureIfPresentAsync(UUID id, String resource);
```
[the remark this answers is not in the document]
andrescrz: This is fine, it's a service, not a library. So we control the single implementor existing in the code base.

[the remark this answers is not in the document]
thiagohora: `IdGenerator` is an internal `@ImplementedBy(IdGeneratorImpl.class)` Guice interface with a single in-tree implementation — it's not a published SPI/extension point, so there are no out-of-tree implementers to break. The only other "implementers" are Mockito mocks in tests, which auto-generate the methods (build is green). These were made abstract (bodies in `IdGeneratorImpl`) per the request to keep the interface free of implementation detail; `default` would put the body back on the interface, which is what we were moving away from. So this stays as-is.

andrescrz (approved):

PR #7568, opened 2026-07-22
[OPIK-7371] [BE] feat: auto-suffix duplicate automation rule names on create

awkoy:
## Details

Automation / online-evaluation rule names have no uniqueness, so re-running an SDK script that creates the same rule produces multiple rules that look identical in the UI (the list surfaces the name; the UUID is the real identifier). This adds best-effort, **project-scoped** auto-suffixing on create and on actual rename: when a name already exists in the target project(s), the rule is stored as `name-1`, `name-2`, … so rules stay distinguishable — without a DB uniqueness constraint or data migration (existing installs already contain colliding names, which the ticket rules out enforcing).

**How it works**

- `AutomationRuleDAO.findCandidateNames` fetches existing names in the target project(s) sharing the requested name's prefix (`name LIKE concat(?, '%') ESCAPE '!'`), scoped via the `automation_rule_projects` junction table (authoritative post-backfill; the legacy `project_id` column is intentionally not consulted). One query serves create and update — update passes `excludeRuleId` so a rule never collides with its own current name.
- `AutomationRuleNames` (new helper) resolves the final name over that candidate set: smallest-free `-N` suffix; comparison case-, accent- and trailing-space-insensitive to approximate MySQL's `utf8mb4_unicode_ci` collation; suffixed candidates are truncated to the 150-char column on a code-point boundary. The LIKE prefix escapes metacharacters (`!`, `%`, `_` — escape char is `!` because the query crosses three backslash-eating layers: Java text block → StringTemplate → MySQL) and is capped at 138 chars so truncated candidates of max-length names still match on re-run.
- `resolveUniqueName` (service) is shared by `save()` (all 6 evaluator types) and `update()`; a non-name edit (sampling rate, enabled, filters, projects) never renames the rule — the stored name is compared first via `findNameById`.
- `@Size(max = 150)` on the create and update models: an over-long name is now rejected with **422** at the API boundary instead of failing the insert with a 500 (**minor API contract change**; OpenAPI spec sync is automated).
- Migration `000092` adds a `(workspace_id, name)` index. **Open question for reviewers:** measured with EXPLAIN on MySQL 8.4 at 50k rules, the optimizer never selects it — the project filter (junction index) is what bounds the query. Keep (in case optimizer stats change) or drop (write amplification while unused)? Documented in `findCandidateNames` Javadoc.

**Intentional scope / accepted limitations**

- No DB unique constraint (per the ticket's constraints), so a rare concurrent create of the same name in the same project can still both persist — best-effort disambiguation, no worse than today.
- Suffixing changes the stored name vs the client-supplied one; a name-based reconcile/upsert loop should read the create response's `Location` rather than assume the name is unchanged. This is the ticket's chosen direction (suffix, not hard-uniqueness/upsert).
- Collation folding is approximate: exotic expansions (`ß`→`ss`) and non-default install collations (e.g. `utf8mb4_0900_ai_ci`) are not replicated exactly.
- A rare un-backfilled legacy rule (no junction row) may be missed — degrading to a duplicate name, not an error.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- OPIK-7371

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8 (initial implementation), Claude Opus 5 / Fable 5 (review rounds)
- Scope: full implementation (helper, DAO query, create + update wiring), unit + integration tests, multiple rounds of adversarial self code-review with fixes applied, live browser E2E, and an EXPLAIN-based query-plan audit on seeded data (50k rules)
- Human verification: reviewer to confirm behavior against the SDK re-run scenario in OPIK-7371

## Testing

- `mvn test -Dtest=AutomationRuleNamesTest` — 23/23 (parameterised resolve cases incl. case/accent/trailing-space folding, gap-fill, freed-base reuse, prefix lookalikes; likePrefix escaping + long-name cap incl. surrogate-pair boundaries; 150-char truncation and re-run distinctness).
- `mvn test -Dtest='AutomationRuleEvaluatorsResourceTest$DuplicateNameHandling'` — 21/21 end-to-end (MySQL/ClickHouse/Redis testcontainers): same-project suffixing, cross-project isolation, multi-project shared-project collision, cross-evaluator-type collision, rename-to-existing with self-exclusion, non-name edits never rename, 422 for >150-char names (create + update), LIKE-metacharacter names (create + rename), a rule literally named `%` not over-matching, SQL-payload names stored literally, max-length names staying distinct across re-runs, accent-insensitive collision end-to-end.
- Query-plan audit: EXPLAIN on MySQL 8.4 with 50k rules / 500 projects (typical + skewed project shapes) — findings documented in the DAO Javadoc and the review threads.
- **Live browser E2E** against a locally-run backend+frontend: created the same rule name repeatedly in one project (→ `-1`, `-2`) and confirmed cross-project isolation. GIFs attached below.
- Not run: full backend suite (scoped to the affected classes; CI runs all 16 integration groups — green); no frontend changes in this PR.

## Documentation

No user-facing docs change. Behavior is transparent (rule names simply gain a `-N` suffix on collision). Note the 422-for-over-long-names contract change above.


apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleNames.java
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        Set<String> taken = new HashSet<>();
+        for (String name : existingNames) {
+            if (name != null) {
+                taken.add(name.toLowerCase(Locale.ROOT));
+            }
+        }
+
+        if (!taken.contains(requestedName.toLowerCase(Locale.ROOT))) {
```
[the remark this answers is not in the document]
awkoy: Fixed in 800152c5b3 — collision comparison now uses a canonical key (NFD normalize + strip diacritics + lowercase), approximating `utf8mb4_unicode_ci` so accented/case variants collide. Added `accentInsensitiveCollision` unit test (`Cafe` vs `Café`).

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleNames.java
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+
+    private static String truncateToFit(String baseName, int suffix) {
+        String suffixStr = "-" + suffix;
+        if (baseName.length() + suffixStr.length() <= MAX_NAME_LENGTH) {
+            return baseName + suffixStr;
+        }
+        int maxBaseLength = MAX_NAME_LENGTH - suffixStr.length();
+        return baseName.substring(0, maxBaseLength) + suffixStr;
```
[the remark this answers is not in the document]
awkoy: Fixed in 800152c5b3 — `truncateToFit` now backs off the cut by one when it would fall between a surrogate pair, so non-BMP names no longer produce a lone surrogate. Added `truncatesOnCodePointBoundary` unit test (name ending in an emoji at the 150-char boundary).

apps/opik-backend/src/test/java/com/comet/opik/domain/evaluators/AutomationRuleNamesTest.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    @ParameterizedTest
+    @NullAndEmptySource
+    @ValueSource(strings = {"   "})
+    @DisplayName("returns blank or null names unchanged")
+    void blankNames(String name) {
+        assertThat(AutomationRuleNames.generateUniqueName(name, Set.of("Hallucination")))
+                .isEqualTo(name);
+    }
```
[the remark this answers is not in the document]
awkoy: Done in 800152c5b3 — split into explicit `nullOrEmptyNames` and `whitespaceOnlyName` tests.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleEvaluatorService.java
```diff
@@ -222,6 +234,20 @@ public <E, F extends Filter, T extends AutomationRuleEvaluator<E, F>> T save(@No
...
+    private String resolveUniqueName(AutomationRuleDAO ruleDAO, String requestedName, Set<UUID> projectIds,
+            String workspaceId) {
+        if (projectIds.isEmpty()) {
+            return requestedName;
+        }
+        Set<String> existingNames = ruleDAO.findNamesByProjects(projectIds, workspaceId);
+        String uniqueName = AutomationRuleNames.generateUniqueName(requestedName, existingNames);
+        if (!uniqueName.equals(requestedName)) {
+            log.info("Rule name '{}' already exists in projectIds '{}' on workspace '{}', using '{}' instead",
+                    requestedName, projectIds, workspaceId, uniqueName);
```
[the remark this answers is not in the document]
awkoy: Fixed in 0d873ea220 — rewrote with a stable literal prefix first and the values grouped as key='value' pairs at the end.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleEvaluatorService.java
```diff
@@ -124,11 +124,18 @@ public <E, F extends Filter, T extends AutomationRuleEvaluator<E, F>> T save(@No
         var savedEvaluator = template.inTransaction(WRITE, handle -> {
             var evaluatorsDAO = handle.attach(AutomationRuleEvaluatorDAO.class);
             var projectsDAO = handle.attach(AutomationRuleProjectsDAO.class);
+            var ruleDAO = handle.attach(AutomationRuleDAO.class);
+
+            // Auto-suffix the name when it collides with existing rules in the same project(s) (OPIK-7371).
+            // Names are not unique at the DB layer, so re-running an SDK script would otherwise create
+            // rules that are indistinguishable in the UI.
+            String uniqueName = resolveUniqueName(ruleDAO, inputRuleEvaluator.getName(), projectIds, workspaceId);
```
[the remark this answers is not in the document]
awkoy: Leaving as-is (intentional). The extra `findNamesByProjects` query only occurs when a Python-evaluator create is about to fail with 501, which is a rare misconfiguration path, and the whole `save` already runs inside one write transaction regardless. Lazily resolving the name would mean duplicating the call across all six type branches (or restructuring the switch), which costs more clarity than the single cheap SELECT it saves on an already-failing request. Happy to revisit if the toggle-disabled path turns out to be hot.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleEvaluatorService.java
```diff
@@ -124,11 +124,19 @@ public <E, F extends Filter, T extends AutomationRuleEvaluator<E, F>> T save(@No
...
+            String uniqueName = resolveUniqueName(inputRuleEvaluator.getName(), existingNames, projectIds, workspaceId);
 
             AutomationRuleEvaluatorModel<?> evaluator = switch (inputRuleEvaluator) {
                 case AutomationRuleEvaluatorLlmAsJudge llmAsJudge -> {
                     var definition = llmAsJudge.toBuilder()
                             .id(id)
+                            .name(uniqueName)
                             .projectId(primaryProjectId)
                             .createdBy(userName)
                             .lastUpdatedBy(userName)
```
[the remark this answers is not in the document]
awkoy: Not applicable — the LLM-as-judge branch **does** call `.name(uniqueName)` (it's the first of the 6 builder branches, all of which set it), and the `whenNameCollidesInSameProject` integration test creates LlmAsJudge rules and asserts the `-1`/`-2` suffix, so this path is covered. On `update()` the resolved name is applied via `updateBaseRule(..., uniqueName, ...)` (the model switch only carries `code`/`lastUpdatedBy`), so it's handled there too.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleEvaluatorService.java
```diff
@@ -233,12 +257,19 @@ public void update(@NonNull UUID id, @NonNull Set<UUID> projectIds, @NonNull Str
...
+                // Auto-suffix on rename collisions too, excluding this rule itself so an unchanged name (or
+                // a non-name edit) is never spuriously suffixed (OPIK-7371).
+                Set<String> existingNames = ruleDAO.findNamesByProjectsExcludingRule(projectIds, workspaceId, id);
+                String uniqueName = resolveUniqueName(evaluatorUpdate.getName(), existingNames, projectIds,
+                        workspaceId);
+
                 // Update base rule (project associations handled separately in junction table)
-                int resultBase = dao.updateBaseRule(id, workspaceId, evaluatorUpdate.getName(),
+                int resultBase = dao.updateBaseRule(id, workspaceId, uniqueName,
                         evaluatorUpdate.getSamplingRate(), evaluatorUpdate.isEnabled(), filtersJson);
```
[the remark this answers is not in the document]
awkoy: Fixed in 80a61cfad8 — `update()` now compares the requested name against the stored name (`findNameById`) and only resolves a unique name on an actual rename; a non-name edit passes the name through untouched, so sampling-rate/enabled/filter edits never rename a rule even when a same-named rule exists in scope. Added an integration test for exactly this.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleEvaluatorService.java
```diff
@@ -222,6 +235,17 @@ public <E, F extends Filter, T extends AutomationRuleEvaluator<E, F>> T save(@No
...
     }
 
+    private String resolveUniqueName(String requestedName, Set<String> existingNames, Set<UUID> projectIds,
+            String workspaceId) {
+        String uniqueName = AutomationRuleNames.generateUniqueName(requestedName, existingNames);
+        if (!uniqueName.equals(requestedName)) {
+            log.info(
+                    "Automation rule name already exists in project scope, applied suffix: requestedName='{}', uniqueName='{}', projectIds='{}', workspaceId='{}'",
+                    requestedName, uniqueName, projectIds, workspaceId);
+        }
```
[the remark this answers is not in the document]
awkoy: Fixed in 80a61cfad8 — the 'applied suffix' log now runs after the write transaction commits (via `logSuffixApplied`), so a rolled-back write can no longer emit a misleading success line.

awkoy:
### ℹ️ Note on the failing `Backend Tests - Unit Tests` / `Unit Tests` check

The 5 failures are **not related to this PR** — they are a pre-existing breakage on `main` that this branch inherited by merging `main` in.

**Failing tests:** `AnthropicClientGeneratorTest`
- `SamplingParamsCapability.supportsSamplingParamsReturnsFalseForAdaptiveThinkingModels` (`claude-sonnet-5`, `claude-opus-4-7`, `claude-opus-4-8`)
- `TemperatureGating.temperatureIsNotForwardedForAdaptiveThinkingModels` (`claude-sonnet-5`, `claude-opus-4-8`)

**Why it's not from OPIK-7371:**
- This PR's diff (`origin/main...HEAD`) touches nothing Anthropic/LLM-client related — only `AutomationRuleNames`, `AutomationRuleDAO`, and `AutomationRuleEvaluatorService` name-suffixing.
- The failures reproduce **deterministically on plain `main`** (run locally on the merged tree: `AnthropicClientGeneratorTest` → `Tests run: 32, Failures: 5`). `AnthropicModelName.supportsSamplingParams(...)` now returns `true` for those models, where the test expects `false`.
- Root cause looks like a "green-PRs → red-main" integration break: the adaptive-thinking judge PR (#7531) added these assertions, and the later provider model-definition sync (#7560) shifted the underlying model list. It affects any PR that merges current `main`, not just this one.

**Everything OPIK-7371-specific is green:** all new unit tests (`AutomationRuleNamesTest` 14/14), the `DuplicateNameHandling` integration suite (7/7), all 16 Backend Integration Groups, CodeQL/Analyze/Baz.

This should clear once `main`'s `AnthropicClientGeneratorTest` is fixed (owned by the Anthropic/model-sync change authors) and this branch re-merges `main`.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleDAO.java
```diff
@@ -42,6 +43,54 @@ HAVING COUNT(project_id) > 1
...
+    @SqlQuery("""
+            SELECT DISTINCT rule.name
+            FROM automation_rules rule
+            LEFT JOIN automation_rule_projects arp ON rule.id = arp.rule_id
+            WHERE rule.workspace_id = :workspaceId
+            AND (arp.project_id IN (<projectIds>) OR rule.project_id IN (<projectIds>))
+            """)
+    @UseStringTemplateEngine
+    @AllowUnusedBindings
+    Set<String> findNamesByProjects(
```
andrescrz: This query is not limited and not paginated. It's going to cause production issues for workspaces with many automation rules.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleDAO.java
```diff
@@ -42,6 +43,54 @@ HAVING COUNT(project_id) > 1
...
+            SELECT DISTINCT rule.name
+            FROM automation_rules rule
+            LEFT JOIN automation_rule_projects arp ON rule.id = arp.rule_id
+            WHERE rule.workspace_id = :workspaceId
+            AND rule.id != :excludeRuleId
+            AND (arp.project_id IN (<projectIds>) OR rule.project_id IN (<projectIds>))
+            """)
+    @UseStringTemplateEngine
+    @AllowUnusedBindings
+    Set<String> findNamesByProjectsExcludingRule(
```
andrescrz: Similar feedback. In addition, you might be able to do this logic in one go, with a single query. You could also centralise the optional filter by rule_id by using a ST template, you we don't copy and paste everything.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleDAO.java
```diff
@@ -42,6 +43,54 @@ HAVING COUNT(project_id) > 1
...
+            @Define("projectIds") @BindList(onEmpty = BindList.EmptyHandling.NULL_VALUE, value = "projectIds") Set<UUID> projectIds,
+            @Bind("workspaceId") String workspaceId,
+            @Bind("excludeRuleId") UUID excludeRuleId);
+
+    /**
+     * Returns the currently stored name of a rule, or {@code null} if it does not exist. Used on update to
+     * skip name resolution entirely for non-name edits (OPIK-7371).
+     */
+    @SqlQuery("SELECT name FROM automation_rules WHERE id = :id AND workspace_id = :workspaceId")
+    Optional<String> findNameById(@Bind("id") UUID id, @Bind("workspaceId") String workspaceId);
```
andrescrz, line 95: Minor: you could generalise this method to return the full automation rule payload, so it's more reusable. And then just use a getter on the name.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleEvaluatorService.java
```diff
@@ -127,11 +127,19 @@ public <E, F extends Filter, T extends AutomationRuleEvaluator<E, F>> T save(@No
         var savedEvaluator = template.inTransaction(WRITE, handle -> {
             var evaluatorsDAO = handle.attach(AutomationRuleEvaluatorDAO.class);
             var projectsDAO = handle.attach(AutomationRuleProjectsDAO.class);
+            var ruleDAO = handle.attach(AutomationRuleDAO.class);
+
+            // Auto-suffix the name when it collides with existing rules in the same project(s) (OPIK-7371).
+            // Names are not unique at the DB layer, so re-running an SDK script would otherwise create
+            // rules that are indistinguishable in the UI.
+            Set<String> existingNames = ruleDAO.findNamesByProjects(projectIds, workspaceId);
```
andrescrz: Your implementation is not scalable not only for technical reasons. The logic should be much simpler as you already know the name you're looking for. 

Instead of traversing the whole workspace, let's find names starting by the name you're looking for. 

You'll need an index for that (check if text or regular index). 

If there's a dirty read under parallelisms, it's ok to be optimistic and just fail if a collision after adding the suffix. 

This should simplify the logic here and solve your scalability problems.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleEvaluatorService.java
```diff
@@ -226,9 +239,22 @@ public <E, F extends Filter, T extends AutomationRuleEvaluator<E, F>> T save(@No
...
         });
 
+        logSuffixApplied(inputRuleEvaluator.getName(), savedEvaluator.name(), projectIds, workspaceId);
+
         return findById(savedEvaluator.id(), savedEvaluator.projectIds(), workspaceId);
     }
 
+    // Logged after the write transaction commits (not inside it) so a rolled-back write never leaves a
+    // misleading "applied suffix" line.
+    private void logSuffixApplied(String requestedName, String appliedName, Set<UUID> projectIds,
```
andrescrz: Let's not print all the projects in the workspace. That would overflown the logs.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleDAO.java
```diff
@@ -42,6 +43,54 @@ HAVING COUNT(project_id) > 1
...
+     * Same as {@link #findNamesByProjects} but excludes a single rule by id, used on update so a rule's own
+     * name (unchanged name or a non-name edit) does not count as a self-collision (OPIK-7371).
+     */
+    @SqlQuery("""
+            SELECT DISTINCT rule.name
+            FROM automation_rules rule
+            LEFT JOIN automation_rule_projects arp ON rule.id = arp.rule_id
+            WHERE rule.workspace_id = :workspaceId
+            AND rule.id != :excludeRuleId
+            AND (arp.project_id IN (<projectIds>) OR rule.project_id IN (<projectIds>))
```
andrescrz: Automation rules in the Opik V2 data model belong to a single project. 

We performed the migration. 

The business logic for this change should account for that. 

I believe it's ok to have collisions per workspace. The suffix should be only for automation rules under the same project. 


We can discuss with product if needed.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleEvaluatorService.java
```diff
@@ -239,18 +265,36 @@ public void update(@NonNull UUID id, @NonNull Set<UUID> projectIds, @NonNull Str
         log.debug("Updating AutomationRuleEvaluator with id '{}' in projectIds '{}' and workspaceId '{}'", id,
                 projectIds,
                 workspaceId);
+        // Holder so the resolved name can be logged after the transaction commits.
+        String[] appliedName = {evaluatorUpdate.getName()};
```
andrescrz: Why an array is needed for a single value?

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleNames.java
```diff
@@ -0,0 +1,91 @@
...
+     * <p>
+     * Comparison is case- and accent-insensitive to approximate MySQL's {@code utf8mb4_unicode_ci}
+     * collation (so {@code Hallucination}/{@code hallucination} and {@code Café}/{@code Cafe} collide),
+     * and each candidate is checked as the actual (possibly truncated) string that would be stored, so
+     * names near the 150-char column limit still resolve to distinct values.
+     *
+     * @param requestedName the name the user asked for
+     * @param existingNames names already present in the target scope
+     */
+    static String generateUniqueName(String requestedName, Collection<String> existingNames) {
```
andrescrz, line 51: We cannot accumulate all the names in memory for big workspaces.

andrescrz (changes requested):
The PR has major issues that we should address before moving forward. I left guidance to fix them. Please take a look.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleDAO.java
awkoy: Fixed in c47b428f5b — no longer fetches all names. Now a bounded, index-backed prefix query (`name LIKE 'base%'`) that returns only the handful of names sharing the prefix in the target project. Added index `automation_rules(workspace_id, name)` in migration 000091.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleDAO.java
awkoy: Fixed in c47b428f5b — collapsed both collision queries into a single StringTemplate query with an optional `<if(excludeRuleId)>` clause (using `@Define+@Bind` like `findCount`'s `action`), so there's no copy-paste. It's also bounded now (prefix + index).

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleDAO.java
awkoy: Same fix as above (c47b428f5b): one shared query, optional `excludeRuleId` via ST template. `findNameById` stays a separate light scalar read used only for the rename short-circuit on update.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleEvaluatorService.java
awkoy: Fixed in c47b428f5b — instead of traversing the whole workspace we now query names starting with the requested name (`LIKE 'base%'`), scoped to the project, backed by the new `(workspace_id, name)` index. Kept it optimistic on concurrent dirty reads (no lock / no unique constraint) as you suggested; documented that in the DAO javadoc.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleEvaluatorService.java
awkoy: Fixed in c47b428f5b — the log no longer prints the projectIds set; it's now just `requestedName -> uniqueName (workspace)` and only when a suffix was actually applied.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleDAO.java
awkoy: Thanks — corrected. Confirmed the junction is authoritative post-`AutomationRuleProjectMigration` backfill, so the suffix scope is now per-project via the junction only; dropped the legacy `project_id` OR branch entirely. Collisions across the workspace are allowed, suffix only within the same project.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleEvaluatorService.java
awkoy: Fixed in c47b428f5b — removed the array. `update()`'s transaction now returns the resolved name directly (`String applied = template.inTransaction(...)`) and it's logged after commit.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleNames.java
awkoy: Fixed in c47b428f5b — we no longer accumulate all names. The prefix query returns only names sharing the requested prefix within the project (small set), and the suffix is computed over that.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleDAO.java
```diff
@@ -42,6 +43,44 @@ HAVING COUNT(project_id) > 1
...
+     * auto-suffix colliding names (OPIK-7371). Scoped per project via the junction table (the authoritative
+     * association after the AutomationRuleProjectMigration backfill); the legacy {@code project_id} column
+     * is intentionally not used (it is nulled on update). {@code excludeRuleId} (optional) skips a single
+     * rule so its own name is not treated as a self-collision on update. The {@code LIKE} prefix keeps the
+     * result set small and index-backed - callers MUST pass a prefix escaped via
+     * {@link AutomationRuleNames#likePrefix(String)} so LIKE metacharacters in the name are matched
+     * literally. Final precise matching is done in Java over this small candidate set.
+     * <p>
+     * Assumptions (optimistic, per OPIK-7371): the junction backfill is complete, so a rare un-backfilled
+     * legacy rule (no junction row) may be missed - degrading to a duplicate name, not an error; and
```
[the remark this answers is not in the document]
awkoy: Intentional, per the area owner's guidance in this PR's review (see the AutomationRuleDAO.java:78 thread): the V2 migration is done — the junction table is the authoritative association, and the legacy `project_id` column is unreliable for scoping (it's nulled on update by `clearLegacyProjectId`). We deliberately stay optimistic: a rare un-backfilled legacy rule degrades to a duplicate name (pre-feature behavior), not an error. This assumption is documented in the DAO javadoc.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleDAO.java
```diff
@@ -42,6 +43,44 @@ HAVING COUNT(project_id) > 1
...
+            AND rule.name LIKE concat(:namePrefix, '%')
+            <if(excludeRuleId)> AND rule.id != :excludeRuleId <endif>
+            """)
+    @UseStringTemplateEngine
+    @AllowUnusedBindings
+    Set<String> findCollidingNames(
+            @Define("projectIds") @BindList(onEmpty = BindList.EmptyHandling.NULL_VALUE, value = "projectIds") Set<UUID> projectIds,
+            @Bind("workspaceId") String workspaceId,
+            @Bind("namePrefix") String namePrefix,
+            @Define("excludeRuleId") @Bind("excludeRuleId") UUID excludeRuleId);
```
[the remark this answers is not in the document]
awkoy: Good call — renamed to `findCandidateNames` in 355ae8de6a (kept it shorter than the suggested variant since project/prefix scoping is visible in the signature; the javadoc spells out that precise matching happens in `generateUniqueName`).

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleDAO.java
andrescrz: I believe this was not addressed. Anyway, minor and not a blocker. If you address it, it'd be just a matter of selecting `*` and returning the full object and using the name getter/accessor in the service.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleEvaluatorService.java
```diff
@@ -226,9 +241,20 @@ public <E, F extends Filter, T extends AutomationRuleEvaluator<E, F>> T save(@No
...
+        logSuffixApplied(inputRuleEvaluator.getName(), savedEvaluator.name(), workspaceId);
+
         return findById(savedEvaluator.id(), savedEvaluator.projectIds(), workspaceId);
     }
 
+    // Logged after the write transaction commits (not inside it) so a rolled-back write never leaves a
+    // misleading "applied suffix" line.
+    private void logSuffixApplied(String requestedName, String appliedName, String workspaceId) {
+        if (appliedName != null && !appliedName.equals(requestedName)) {
+            log.info("Automation rule name '{}' already existed in project scope, applied suffix '{}' (workspace '{}')",
```
andrescrz: Nit: for logs, usually we move printed values to the end of the sentence, that facilitates grepping during production debugging sessions.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleEvaluatorService.java
```diff
@@ -239,18 +265,34 @@ public void update(@NonNull UUID id, @NonNull Set<UUID> projectIds, @NonNull Str
...
+                // project (e.g. legacy duplicates). On a real rename, dedup excluding this rule itself so it
+                // does not collide with its own current name (OPIK-7371).
+                String currentName = ruleDAO.findNameById(id, workspaceId).orElse(null);
+                String uniqueName;
+                if (Objects.equals(requestedName, currentName)) {
+                    uniqueName = requestedName;
+                } else {
+                    Set<String> candidates = ruleDAO.findCandidateNames(projectIds, workspaceId,
+                            AutomationRuleNames.likePrefix(requestedName), id);
+                    uniqueName = AutomationRuleNames.generateUniqueName(requestedName, candidates);
```
andrescrz: Nit: same invocations as above, you can encapsulate this small piece of logic in a small aux private method.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleNames.java
```diff
@@ -0,0 +1,109 @@
...
+        if (StringUtils.isBlank(requestedName) || existingNames == null || existingNames.isEmpty()) {
```
andrescrz: Nit: you can use `CollectionUtils.isEmpty` to the null safe check in one go.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleNames.java
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * Builds the escaped prefix to feed to a {@code name LIKE concat(?, '%')} lookup for {@code name}.
+     * Trailing spaces are stripped (MySQL PAD SPACE comparison ignores them) and LIKE metacharacters
+     * ({@code \}, {@code %}, {@code _}) are escaped so they match literally rather than as wildcards -
+     * otherwise a name such as {@code a\b} or {@code 50%} would silently miss its own collisions.
+     * Case/accent folding is left to the column collation. Final precise matching happens in
+     * {@link #generateUniqueName} over the returned candidate set.
+     */
+    static String likePrefix(String name) {
```
andrescrz, line 99: Please audit and double check the safety of this, mostly against injection.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/AutomationRuleEvaluatorsResourceTest.java
```diff
@@ -762,6 +762,187 @@ void createEvaluator__whenMaxCostUsdIsNotPositive__thenReturnUnprocessableEntity
...
+        void whenNameCollidesInDifferentProject__thenNoSuffix() {
+            var projectId1 = projectResourceClient.createProject(UUID.randomUUID().toString(), API_KEY, WORKSPACE_NAME);
+            var projectId2 = projectResourceClient.createProject(UUID.randomUUID().toString(), API_KEY, WORKSPACE_NAME);
+            var name = "Relevance " + UUID.randomUUID();
+
+            assertThat(createRuleAndGetName(name, projectId1)).isEqualTo(name);
+            assertThat(createRuleAndGetName(name, projectId2)).isEqualTo(name);
+        }
+
+        private String createLlmRuleAndGetName(String name, Set<UUID> projectIds, UUID readProjectId) {
```
andrescrz: Copied and pasted method (createRuleAndGetName). Please centralise and provide overloaded flavours.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/AutomationRuleEvaluatorsResourceTest.java
```diff
@@ -762,6 +762,187 @@ void createEvaluator__whenMaxCostUsdIsNotPositive__thenReturnUnprocessableEntity
...
+
+            var spanJudge = factory.manufacturePojo(AutomationRuleEvaluatorSpanLlmAsJudge.class).toBuilder()
+                    .name(name)
+                    .projectIds(Set.of(projectId))
+                    .build();
+            var id = evaluatorsResourceClient.createEvaluator(spanJudge, WORKSPACE_NAME, API_KEY);
+            try (var response = evaluatorsResourceClient.getEvaluator(id, projectId, WORKSPACE_NAME, API_KEY,
+                    HttpStatus.SC_OK)) {
+                assertThat(response.readEntity(AutomationRuleEvaluator.class).getName()).isEqualTo(name + "-1");
+            }
```
andrescrz: Use aux methods.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/AutomationRuleEvaluatorsResourceTest.java
```diff
@@ -762,6 +762,187 @@ void createEvaluator__whenMaxCostUsdIsNotPositive__thenReturnUnprocessableEntity
...
+                    .build();
+            return evaluatorsResourceClient.createEvaluator(evaluator, WORKSPACE_NAME, API_KEY);
+        }
+
+        private String getName(UUID id, UUID projectId) {
+            try (var response = evaluatorsResourceClient.getEvaluator(id, projectId, WORKSPACE_NAME, API_KEY,
+                    HttpStatus.SC_OK)) {
+                return response.readEntity(AutomationRuleEvaluator.class).getName();
+            }
+        }
```
andrescrz: Repeated, use these, other methods should delegate to these.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/AutomationRuleEvaluatorsResourceTest.java
```diff
@@ -762,6 +762,187 @@ void createEvaluator__whenMaxCostUsdIsNotPositive__thenReturnUnprocessableEntity
...
+        void whenUpdatingNonExistentRule__thenNotFound() {
+            var projectId = projectResourceClient.createProject(UUID.randomUUID().toString(), API_KEY, WORKSPACE_NAME);
+            var update = factory.manufacturePojo(AutomationRuleEvaluatorUpdateLlmAsJudge.class).toBuilder()
+                    .name("Ghost " + UUID.randomUUID())
+                    .projectIds(Set.of(projectId))
+                    .build();
+            try (var response = evaluatorsResourceClient.callUpdateEvaluator(UUID.randomUUID(), WORKSPACE_NAME, update,
+                    API_KEY)) {
+                assertThat(response.getStatusInfo().getStatusCode()).isEqualTo(HttpStatus.SC_NOT_FOUND);
+            }
```
andrescrz, line 946: Repeated again, reuse methods.

apps/opik-backend/src/test/java/com/comet/opik/domain/evaluators/AutomationRuleNamesTest.java
```diff
@@ -0,0 +1,150 @@
...
+
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+import java.util.Set;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@DisplayName("AutomationRuleNames")
+class AutomationRuleNamesTest {
```
andrescrz, line 16: Minor: many of these tests are the same, they could be paramterised.

apps/opik-backend/src/test/java/com/comet/opik/domain/evaluators/AutomationRuleNamesTest.java
```diff
@@ -0,0 +1,150 @@
...
+        assertThat(AutomationRuleNames.generateUniqueName("hallucination", Set.of("Hallucination")))
+                .isEqualTo("hallucination-1");
+        assertThat(AutomationRuleNames.generateUniqueName("HALLUCINATION",
+                Set.of("Hallucination", "hallucination-1")))
+                .isEqualTo("HALLUCINATION-2");
+    }
+
+    @Test
+    @DisplayName("likePrefix escapes LIKE metacharacters and strips trailing spaces")
+    void likePrefixEscaping() {
```
andrescrz: Related to my other comment, not sure if an user is going to understand these renaming. 

What we sure ensure instead is totally security agains injection.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleDAO.java
andrescrz: No pagination yet, although likely not a blocker anymore after narrowing down the found data.

andrescrz (commented):
This has addressed many of the previous comments.

Let's keep iterating on this. 

Please make sure all Baz comments are either addressed or replied.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleNames.java
```diff
@@ -0,0 +1,87 @@
...
+
+        Set<String> taken = new HashSet<>();
+        for (String name : existingNames) {
+            if (name != null) {
+                taken.add(canonicalKey(name));
+            }
+        }
+
+        if (!taken.contains(canonicalKey(requestedName))) {
+            return requestedName;
```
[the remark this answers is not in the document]
awkoy: Fixed in 45f198186e — `name` now carries `@Size(max = 150, message = "cannot exceed 150 characters")` on both the create model (`AutomationRuleEvaluator`) and the update model (`AutomationRuleEvaluatorUpdate`), so an over-long name is rejected with a 422 and a clear message at the API boundary instead of reaching the insert and failing with a 500.

Verified TDD-style: the new tests (`createEvaluator__whenNameExceedsColumnLength__thenReturnUnprocessableEntity` and the update twin) failed with `expected: 422, but was: 500` before the annotations and pass after. Worth noting this was pre-existing behaviour (over-long names 500'd before this PR too), but this PR is the right place to close it.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleNames.java
```diff
@@ -0,0 +1,87 @@
...
+
+    /**
+     * Normalizes a name for collision comparison: strips diacritics and lower-cases, approximating the
+     * accent- and case-insensitive folding of MySQL's {@code utf8mb4_unicode_ci} collation. Mirrors the
+     * normalization used by {@link com.comet.opik.domain.SlugUtils}.
+     */
+    private static String canonicalKey(String name) {
+        String withoutDiacritics = Normalizer.normalize(name, Normalizer.Form.NFD)
+                .replaceAll("\\p{M}", "");
+        return withoutDiacritics.toLowerCase(Locale.ROOT);
```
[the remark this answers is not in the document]
awkoy: Fixed in 45f198186e — hoisted to `private static final Pattern DIACRITICS = Pattern.compile("\\p{M}")` and `canonicalKey` now reuses it via `DIACRITICS.matcher(normalized).replaceAll("")`, so nothing is recompiled inside the per-candidate loop.

Side note: `SlugUtils.generateSlug` (which this class's normalization mirrors) has the same inline `replaceAll("\\p{M}", "")` — left untouched here since it's out of this PR's scope, but it's the same win if anyone wants a follow-up.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleEvaluatorService.java
```diff
@@ -230,15 +256,33 @@ public void update(@NonNull UUID id, @NonNull Set<UUID> projectIds, @NonNull Str
...
+                // project (e.g. legacy duplicates). On a real rename, dedup excluding this rule itself so it
+                // does not collide with its own current name (OPIK-7371).
+                String requestedName = evaluatorUpdate.getName();
+                String currentName = ruleDAO.findNameById(id, workspaceId);
+                String uniqueName;
+                if (Objects.equals(requestedName, currentName)) {
+                    uniqueName = requestedName;
+                } else {
+                    Set<String> existingNames = ruleDAO.findNamesByProjectsExcludingRule(projectIds, workspaceId, id);
+                    uniqueName = AutomationRuleNames.generateUniqueName(requestedName, existingNames);
```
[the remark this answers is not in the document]
awkoy: Fixed in 45f198186e — extracted `resolveUniqueName(ruleDAO, requestedName, projectIds, workspaceId, excludeRuleId)` and both paths now call it: `save()` with `excludeRuleId = null`, `update()` with the rule's own id. The name-unchanged fast path stays in `update()` only (outside the helper), since a create has no current name to compare against — folding it in would have applied an update-specific guard to creates.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleEvaluatorService.java
```diff
@@ -226,9 +241,20 @@ public <E, F extends Filter, T extends AutomationRuleEvaluator<E, F>> T save(@No
...
+
         return findById(savedEvaluator.id(), savedEvaluator.projectIds(), workspaceId);
     }
 
+    // Logged after the write transaction commits (not inside it) so a rolled-back write never leaves a
+    // misleading "applied suffix" line.
+    private void logSuffixApplied(String requestedName, String appliedName, String workspaceId) {
+        if (appliedName != null && !appliedName.equals(requestedName)) {
+            log.info("Automation rule name '{}' already existed in project scope, applied suffix '{}' (workspace '{}')",
+                    requestedName, appliedName, workspaceId);
```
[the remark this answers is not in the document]
awkoy: Fixed in 45f198186e — the message no longer claims to print a "suffix" when it receives the full stored name. It now reads: `Automation rule name already existed in project scope, stored under a new name: requestedName '{}', appliedName '{}', workspaceId '{}'` — accurate naming, stable greppable prefix, and all values trailing (which also closes the same feedback from @andrescrz on this line).

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleEvaluatorService.java
awkoy: Fixed in 45f198186e — rewrote as a stable literal prefix with every value trailing:

`"Automation rule name already existed in project scope, stored under a new name: requestedName '{}', appliedName '{}', workspaceId '{}'"`

(one rewrite also closes baz's point on this line that the old wording said "suffix" while printing the full name).

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleEvaluatorService.java
awkoy: Fixed in 45f198186e — extracted `resolveUniqueName(ruleDAO, requestedName, projectIds, workspaceId, excludeRuleId)`, shared by `save()` and `update()`. The `Objects.equals(requestedName, currentName)` shortcut stays in `update()` since it has no analogue on create.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleNames.java
awkoy: Fixed in 45f198186e — `CollectionUtils.isEmpty(existingNames)` (commons-collections4, already the house import elsewhere in `domain/`).

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleDAO.java
awkoy: I implemented this and had to back it out — with evidence this time rather than preference.

`AutomationRuleDAO`'s registered `AutomationRuleRowMapper` is polymorphic: it dispatches on `action` and delegates to `AutomationRuleEvaluatorModel`, whose mapper requires `evaluator.type` / `evaluator.code`, the audit columns from `automation_rule_evaluators`, and the `project_id AS legacy_project_id` alias. A `SELECT *` from `automation_rules` alone therefore fails at runtime with `IllegalStateException: No mapper found for Automation Rule Action type: EVALUATOR` — verified: it broke every update-path test in the suite.

Making it work means (a) registering the evaluator row mapper on this DAO too and (b) joining `automation_rule_evaluators` and deserializing the full `code` JSON payload (the entire LLM-as-judge prompt) on every update, just to read one column on the non-rename fast path. That seemed like a worse trade than the single-column projection, so `findNameById` stays and the reasoning is now documented in its Javadoc (45f198186e). Happy to revisit if you'd rather have the general method despite the join cost.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleNames.java
awkoy: Audited — with the rendered SQL rather than just reasoning. What JDBI actually sends (captured from a statement log):

```sql
SELECT DISTINCT rule.name
FROM automation_rules rule
JOIN automation_rule_projects arp ON rule.id = arp.rule_id
WHERE rule.workspace_id = :workspaceId
AND arp.project_id IN (:__projectIds_0)
AND rule.name LIKE concat(:namePrefix, '%') ESCAPE '!'
```

- the rule name never enters the SQL text: `likePrefix`'s output arrives as the bound `:namePrefix` parameter;
- the StringTemplate `<projectIds>` renders to generated *placeholders* (`:__projectIds_0`), not values — `@BindList` binds them;
- `<if(excludeRuleId)>` is a pure boolean gate; when present it renders `AND rule.id != :excludeRuleId`, still a bind.

So there is no SQL-injection surface in this query. The `%`/`_` escaping is a LIKE-wildcard **correctness** measure, not a security one — worst case if it were wrong is an over-/under-broad candidate set (a duplicate name), never data exposure.

Hardening landed in 45f198186e: the escape character is now declared explicitly (`ESCAPE '!'` — `!` rather than `\` because the query crosses three escaping layers: Java text block → StringTemplate → MySQL, and `ESCAPE '\'` silently rendered as a syntax error), plus tests: names like `' OR 1=1 --` and `"; DROP TABLE automation_rules; --` round-trip literally and suffix normally, and a rule literally named `%` no longer risks matching every name in the project.

One scope note: if by "totally security against injection" you meant a broader audit than this PR's query — `AutomationRuleEvaluatorDAO`'s `<sort_fields>`/`<filters>` templates *do* interpolate SQL fragments; that's pre-existing and untouched here, but it's where I'd look next if you want a follow-up ticket.

apps/opik-backend/src/test/java/com/comet/opik/domain/evaluators/AutomationRuleNamesTest.java
awkoy: Good catch on the confusion — but nothing here renames what the user sees. `likePrefix` builds the SQL `LIKE` search pattern used to find collision candidates; its return value is never persisted and never displayed. A rule the user names `50%` is stored and shown as `50%`. The only user-visible name change in this PR is the `-1`/`-2` suffix on an actual collision.

That wasn't obvious from this test, which is a fair criticism — the assertions read like a transformation of the name. In 45f198186e I renamed it to "likePrefix builds a search pattern (escaping metacharacters); it never alters the stored name" and added a Javadoc pointing to the end-to-end proof: `DuplicateNameHandling#whenNameHasLikeMetacharacters__thenSuffixIsAppended` creates rules named `metric_score`, `cost 50%`, `path\to\rule`, `urgent!` and asserts each comes back **verbatim**, then `-1` on the second create.

On the security half — answered with the rendered-SQL evidence in the `likePrefix` thread.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleDAO.java
awkoy: Measured rather than argued this time — MySQL 8.4, 50k rules / 500 projects in one workspace (seeded scratch schema):

- **Typical project (~100 rules):** the optimizer drives from `automation_rule_projects` on `project_id` (67 rows) and reaches `automation_rules` by primary key; the name prefix is only a residual filter. Bounded and fast.
- **Skewed project (20k rules):** it flips to `automation_rules_idx` and examines ~24.7k rows.
- In **neither** case is the `(workspace_id, name)` index from migration 000092 selected — `FORCE INDEX` does produce a better plan (covering range scan, ~half the rows), so it's usable but currently inert. Honest correction of my earlier "index-backed" claim; the docs on `findCandidateNames` now say what actually bounds the query (the project filter). **Open question for you: keep the index (in case the optimizer picks it up as stats change) or drop the migration to avoid write amplification on every rule insert?**

On `LIMIT`: I've deliberately left it off. The result feeds a smallest-free-suffix probe, so a cap would silently truncate the candidate set and could hand back a name that already exists — the opposite of the feature. Narrow-but-unbounded seemed the safer trade, but if you want a defensive cap I'll add one with a log line when it's hit.

The audit also surfaced a real bug in the prefix approach, fixed in 45f198186e: suffixing a max-length (150-char) name truncates its base, so the stored `name-1` no longer shared the full search prefix and a third create regenerated a duplicate `name-1`. The prefix is now capped at 138 chars (150 minus the worst-case suffix reserve), with a failing-first integration test locking it.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/AutomationRuleEvaluatorsResourceTest.java
awkoy: Fixed in 45f198186e — collapsed to two primitives, `createRule(evaluator)` and `getName(id, projectId)`, with everything else delegating: `llmRule(name, projectIds)` (builder), `createLlmRule(name, projectIds|projectId)`, `createRuleAndGetName(name, projectIds, readProjectId | name, projectId)` as the overloaded flavours, plus `updateRule`/`renameRule` and a `createProject()` helper. No test builds a rule or reads a name inline anymore.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/AutomationRuleEvaluatorsResourceTest.java
awkoy: Fixed in 45f198186e — this block now goes through `createRule(spanJudge)` + `getName(...)`. (It keeps building the span-judge POJO explicitly because the test's point is a *different evaluator type* colliding, which the LLM-judge convenience overloads can't express.)

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/AutomationRuleEvaluatorsResourceTest.java
awkoy: Done in 45f198186e — `createRule`/`getName` are now the primitives at the top of the class and every other helper delegates to them.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/AutomationRuleEvaluatorsResourceTest.java
awkoy: Fixed in 45f198186e — routed through the shared `llmRuleUpdate(...)` builder / `updateRule(...)` helpers; the only inline `callUpdateEvaluator` left is this 404 case, which asserts a non-2xx status the happy-path helper deliberately doesn't allow.

apps/opik-backend/src/test/java/com/comet/opik/domain/evaluators/AutomationRuleNamesTest.java
awkoy: Fixed in 45f198186e — the pure `(requested, existing) → expected` cases are now a single `@ParameterizedTest` with a named `@MethodSource` (18 cases, each reporting its own description). Kept standalone: the truncation trio (they assert on length/code-point properties, not a fixed string) and the `likePrefix` tests (different method).

One note: baz asked in round 1 to *split* the parameterised blank-name test apart; those cases are folded back in here as individually-named parameters, which I think satisfies both intents — each case still reports its own name in the output.

awkoy:
Follow-up on my CI note above (23 Jul): **obsolete** — the `AnthropicClientGeneratorTest` breakage was fixed on `main`, and after merging it in, Unit Tests and all 16 Backend Integration Groups are green on this branch. Nothing pending from that thread.

Latest push (45f198186e) addresses the remaining review round: name-length validation (422 instead of 500), explicit LIKE `ESCAPE`, a shared `resolveUniqueName` path, the test-helper/parameterisation cleanups, and one real bug the audit surfaced (duplicate suffixes for max-length names on re-run — failing test first, then fixed). Per-thread details in the inline replies.

apps/opik-backend/src/main/java/com/comet/opik/api/evaluators/AutomationRuleEvaluatorUpdate.java
```diff
@@ -49,7 +50,9 @@ public abstract sealed class AutomationRuleEvaluatorUpdate<T, E extends Filter>
         AutomationRuleEvaluatorUpdateUserDefinedMetricPython,
         AutomationRuleEvaluatorUpdateSpanLlmAsJudge, AutomationRuleEvaluatorUpdateSpanUserDefinedMetricPython {
 
-    @NotBlank private final String name;
+    // Bounded to match the automation_rules.name VARCHAR(150) column, so an over-long rename is rejected at
+    // the API boundary instead of failing the update (OPIK-7371).
+    @NotBlank @Size(max = 150, message = "cannot exceed 150 characters") private final String name;
```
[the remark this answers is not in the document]
awkoy: Not applicable — verified against the generated artifact rather than the checked-in spec. The spec produced from **this branch** during the backend build (`apps/opik-backend/target/openapi.yaml`) already emits the constraint for the update model:

```yaml
name:
  maxLength: 150
  minLength: 0
  type: string
```

The file you compared against (`sdks/code_generation/fern/openapi/openapi.yaml`) is regenerated from `main` by the post-merge automation, so it can't contain this PR's annotations until the PR merges. Swagger-core's `@Size` → `maxLength` mapping is also already proven repo-wide: 111 `maxLength` occurrences in the checked-in spec, e.g. `Environment_Write.name` from the identical `@Size(max = 150)` annotation.

apps/opik-backend/src/main/java/com/comet/opik/domain/evaluators/AutomationRuleDAO.java
```diff
@@ -42,6 +43,57 @@ HAVING COUNT(project_id) > 1
...
+     * concurrent creates of the same name race without a DB constraint.
+     */
+    @SqlQuery("""
+            SELECT DISTINCT rule.name
+            FROM automation_rules rule
+            JOIN automation_rule_projects arp ON rule.id = arp.rule_id
+            WHERE rule.workspace_id = :workspaceId
+            AND arp.project_id IN (<projectIds>)
+            AND rule.name LIKE concat(:namePrefix, '%') ESCAPE '!'
+            <if(excludeRuleId)> AND rule.id != :excludeRuleId <endif>
```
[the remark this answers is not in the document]
awkoy: This was already raised and settled earlier in this PR (thread on `AutomationRuleDAO.java` from 23 Jul): the area owner's guidance is that the junction table is the authoritative association after the `AutomationRuleProjectMigration` backfill, the legacy `project_id` column is unreliable (it is nulled on update), and V2 rules are single-project. The optimistic assumption — a rare un-backfilled legacy rule degrades to a duplicate name, not an error — is documented in this method's Javadoc under "Assumptions". Keeping junction-only per that ruling.

apps/opik-backend/src/main/java/com/comet/opik/api/evaluators/AutomationRuleEvaluator.java
```diff
@@ -74,7 +75,9 @@ public abstract sealed class AutomationRuleEvaluator<T, E extends Filter> implem
     private final Set<UUID> projectIds;
 
     @JsonView({View.Public.class, View.Write.class})
-    @NotBlank private final String name;
+    // Bounded to match the automation_rules.name VARCHAR(150) column, so an over-long name is rejected at
+    // the API boundary instead of failing the insert (OPIK-7371).
+    @NotBlank @Size(max = 150, message = "cannot exceed 150 characters") private final String name;
```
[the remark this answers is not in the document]
awkoy: Both halves checked:

**OpenAPI**: the spec generated from this branch's build (`apps/opik-backend/target/openapi.yaml`) does emit `maxLength: 150` for `AutomationRuleEvaluator_Write.name` — the checked-in Fern spec lags because it is regenerated from `main` post-merge. No annotation change needed.

**UTF-16 vs MySQL characters**: real difference, deliberately conservative direction. `@Size` counts UTF-16 code units, and code units ≥ code points, so any name that passes validation has ≤150 code points and **always fits** `VARCHAR(150)` — no 500 path exists. The only effect is an early 422 for names packing 76+ non-BMP characters (e.g. an all-emoji name caps at ~75 emoji instead of 150), which is also consistent with how `truncateToFit` measures length (code units). Trading a slightly stricter limit on emoji-heavy names for a guarantee the insert can't fail seemed like the right side of the trade.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/AutomationRuleEvaluatorsResourceTest.java
```diff
@@ -762,6 +762,293 @@ void createEvaluator__whenMaxCostUsdIsNotPositive__thenReturnUnprocessableEntity
...
+
+        @Test
+        @DisplayName("renaming to a name containing LIKE metacharacters still detects the collision")
+        void whenRenamedToMetacharacterName__thenSuffixIsAppended() {
+            var projectId = createProject();
+            var takenName = "cost 50% " + UUID.randomUUID();
+
+            createLlmRule(takenName, projectId);
+            var idB = createLlmRule("Other " + UUID.randomUUID(), projectId);
+
```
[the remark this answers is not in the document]
awkoy: Half agreed, but the test isn't tautological — it's worth spelling out which regression direction each test locks, because escaping can only fail in two ways here:

**Miss direction (self-match breakage)**: an escaping bug that makes the pattern fail to match the stored literal (e.g. the earlier `a\b` double-escaping variant this PR actually shipped and fixed) → the identical stored name isn't fetched → no suffix → this test's `-1` assertion **fails**. That's the class this test locks, now for the rename path too.

**Over-match direction (unescaped wildcards)**: you're right that an unescaped `%` still matches its own row and this test passes. But over-fetching can never flip an API-visible outcome at all: candidates go through exact collation-key matching in `generateUniqueName`, so extra rows are discarded and the stored name is identical either way — the only observable damage is fetch breadth (perf), plus the pathological case of a rule literally named `%` matching everything, which `whenAWildcardNamedRuleExists__thenUnrelatedNamesAreNotSuffixed` covers. A "wildcard-friendly non-colliding name" variant produces the same result with escaping on or off, for the same reason — so there's no stronger black-box assertion available at the API level; the design doc on `likePrefix` says exactly this ("Final precise matching happens in generateUniqueName over the returned candidate set").

andrescrz (approved):
Thanks for putting this together. LGTM.

PR #7573, opened 2026-07-22
[OPIK-7402] [BE] Add audit (shadow / log-only) mode to UUIDv7 ingestion validation

thiagohora:
## Details

Adds a third **audit** (shadow / log-only) state to `uuidValidation`, on top of the existing `enabled` kill-switch. When `enabled=true` and `auditOnly=true`, ids whose embedded UUIDv7 timestamp is out of the window are **counted and logged but not rejected**, so clients emitting them surface in real time without breaking ingestion. This is the freshness mechanism for safely re-enabling rejection later.

- Effective mode: `enabled=false` → disabled (no-op); `enabled=true & auditOnly=true` → audit; `enabled=true & auditOnly=false` → reject (HTTP 400, unchanged).
- Audit mode emits `opik.ingestion.uuid_v7.rejected` (`mode=audit`) tagged by `workspace_id` + `reason` + `resource`, so offenders can be broken down per workspace. It shares the counter/instrument with the reject-path mapper.
- Scope is the window check (the `enabled`-gated one). The always-on NOT_V7 version check is unchanged. `workspace_id` is read from the reactive context on the async paths and threaded through the batch bind on the sync path; `project_id` is not available at this layer and is left as a follow-up. Reject-path workspace tagging is likewise a follow-up.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- OPIK-7402

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Scope: full implementation (config, validator, metrics recorder, call-site plumbing, unit test) and this description, authored under human direction and review.
- Human verification: design decisions (additive `auditOnly` flag, scope limited to the window check, workspace-only tagging) chosen by the author; `mvn test-compile` and the new unit test run and green locally; diff reviewed.

## Testing

- `mvn test-compile` — BUILD SUCCESS (main + tests).
- `mvn surefire:test -Dtest=UuidV7TimestampValidatorTest` — 5 tests, 0 failures. Covers all three modes: disabled (accepts out-of-window ids), reject (accepts in-window; rejects too-old/too-far-future on create, only too-far-future on update), audit (never rejects, tolerates a blank workspace id).
- Not run locally: resource/integration tests (bootstrap is flaky locally) — left to CI.

## Documentation

Config documented inline in `config.yml` / `config-test.yml` (new `auditOnly` key, effective-mode table) and in the `UuidValidationConfig` / `UuidV7TimestampValidator` / `UuidValidationMetrics` javadoc. No user-facing docs change.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/error/InvalidUUIDExceptionMapper.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/error/InvalidUUIDExceptionMapper.java
@@ -1,5 +1,6 @@
 package com.comet.opik.api.error;
 
+import com.comet.opik.infrastructure.metrics.UuidValidationMetrics;
 import io.dropwizard.jersey.errors.ErrorMessage;
 import io.opentelemetry.api.GlobalOpenTelemetry;
 import io.opentelemetry.api.common.AttributeKey;
@@ -22,7 +23,8 @@ import static jakarta.ws.rs.core.Response.Status.BAD_REQUEST;
 
 /**
  * Maps {@link InvalidUUIDException} to HTTP 400 and records the reject-rate metric.
- * Tags each rejection with the matched route.
+ * Tags each rejection with the matched route and {@code mode=reject} (see {@link UuidValidationMetrics}
+ * for the shared {@code opik.ingestion.uuid_v7.rejected} counter contract).
  *
  * <p>{@link ResourceInfo} is request-scoped, so it is obtained lazily through a {@link Provider}
  * (injecting it directly would fail to construct this singleton outside a request). The lookup resolves
@@ -50,8 +52,11 @@ public class InvalidUUIDExceptionMapper implements ExceptionMapper<InvalidUUIDEx
         var httpRoute = getHttpRoute();
         log.info("Rejected ingestion id, httpRoute: '{}', reason: '{}', error message: '{}'",
                 httpRoute, exception.getReason().getValue(), exception.getMessage());
-        REJECTED_COUNTER.add(1,
-                Attributes.of(HTTP_ROUTE_KEY, httpRoute, REASON_KEY, exception.getReason().getValue()));
+        REJECTED_COUNTER.add(1, Attributes.builder()
+                .put(UuidValidationMetrics.MODE_KEY, UuidValidationMetrics.MODE_REJECT)
+                .put(HTTP_ROUTE_KEY, httpRoute)
+                .put(REASON_KEY, exception.getReason().getValue())
+                .build());
         // Force JSON: ingestion endpoints negotiate other content types (e.g. the OTel endpoint uses protobuf),
         // which have no writer for the error entity — without this the 400 fails to serialize and surfaces as a 500
         return Response.status(BAD_REQUEST)
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/error/InvalidUUIDExceptionMapper.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/error/InvalidUUIDExceptionMapper.java
@@ -2,10 +2,6 @@ package com.comet.opik.api.error;
 
 import com.comet.opik.infrastructure.metrics.UuidValidationMetrics;
 import io.dropwizard.jersey.errors.ErrorMessage;
-import io.opentelemetry.api.GlobalOpenTelemetry;
-import io.opentelemetry.api.common.AttributeKey;
-import io.opentelemetry.api.common.Attributes;
-import io.opentelemetry.api.metrics.LongCounter;
 import jakarta.inject.Inject;
 import jakarta.inject.Provider;
 import jakarta.ws.rs.Path;
@@ -22,9 +18,10 @@ import java.lang.reflect.Method;
 import static jakarta.ws.rs.core.Response.Status.BAD_REQUEST;
 
 /**
- * Maps {@link InvalidUUIDException} to HTTP 400 and records the reject-rate metric.
- * Tags each rejection with the matched route and {@code mode=reject} (see {@link UuidValidationMetrics}
- * for the shared {@code opik.ingestion.uuid_v7.rejected} counter contract).
+ * Maps {@link InvalidUUIDException} to HTTP 400 and records the reject-rate metric, tagged with the
+ * matched route and {@code mode=reject}. The counter itself lives in {@link UuidValidationMetrics}
+ * (the central owner of {@code opik.ingestion.uuid_v7.rejected}); this mapper only supplies the route
+ * and delegates.
  *
  * <p>{@link ResourceInfo} is request-scoped, so it is obtained lazily through a {@link Provider}
  * (injecting it directly would fail to construct this singleton outside a request). The lookup resolves
@@ -34,29 +31,17 @@ import static jakarta.ws.rs.core.Response.Status.BAD_REQUEST;
 @RequiredArgsConstructor(onConstructor_ = @Inject)
 public class InvalidUUIDExceptionMapper implements ExceptionMapper<InvalidUUIDException> {
 
-    private static final String METRIC_NAMESPACE = "opik.ingestion";
     private static final String UNKNOWN_ROUTE = "unknown";
 
-    private static final AttributeKey<String> HTTP_ROUTE_KEY = AttributeKey.stringKey("http_route");
-    private static final AttributeKey<String> REASON_KEY = AttributeKey.stringKey("reason");
-
-    private static final LongCounter REJECTED_COUNTER = GlobalOpenTelemetry.get().getMeter(METRIC_NAMESPACE)
-            .counterBuilder("%s.uuid_v7.rejected".formatted(METRIC_NAMESPACE))
-            .setDescription("Number of writes rejected because the id failed UUIDv7 ingestion validation")
-            .build();
-
     private final Provider<ResourceInfo> resourceInfo;
+    private final UuidValidationMetrics uuidValidationMetrics;
 
     @Override
     public Response toResponse(@NonNull InvalidUUIDException exception) {
         var httpRoute = getHttpRoute();
         log.info("Rejected ingestion id, httpRoute: '{}', reason: '{}', error message: '{}'",
                 httpRoute, exception.getReason().getValue(), exception.getMessage());
-        REJECTED_COUNTER.add(1, Attributes.builder()
-                .put(UuidValidationMetrics.MODE_KEY, UuidValidationMetrics.MODE_REJECT)
-                .put(HTTP_ROUTE_KEY, httpRoute)
-                .put(REASON_KEY, exception.getReason().getValue())
-                .build());
+        uuidValidationMetrics.recordReject(exception.getReason().getValue(), httpRoute);
         // Force JSON: ingestion endpoints negotiate other content types (e.g. the OTel endpoint uses protobuf),
         // which have no writer for the error entity — without this the 400 fails to serialize and surfaces as a 500
         return Response.status(BAD_REQUEST)
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/IdGenerator.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/IdGenerator.java
@@ -111,16 +111,19 @@ class IdGeneratorImpl implements IdGenerator {
     }
 
     @Override
-
-    public void validateIdForUpdate(@NonNull UUID id, String resource, String workspaceId) {
+    public void validateIdNotInFuture(@NonNull UUID id, String resource) {
         IdGenerator.validateVersion(id, resource);
-        uuidV7TimestampValidator.validateNotInFuture(id, resource, workspaceId);
+        // Referenced/foreign ids are validated where the request-scoped workspace is not threaded
+        // (e.g. the synchronous fail-fast batch pass), so the audit metric falls back to UNKNOWN here;
+        // the batch's own ids still carry the workspace via validateId.
+        uuidV7TimestampValidator.validateNotInFuture(id, resource, ErrorMetricsResolver.UNKNOWN);
     }
 
     @Override
-    public Mono<UUID> validateIdForUpdateAsync(@NonNull UUID id, String resource) {
+    public Mono<UUID> validateIdNotInFutureAsync(@NonNull UUID id, String resource) {
         return Mono.deferContextual(ctx -> Mono.fromCallable(() -> {
-            validateIdNotInFuture(id, resource, workspaceId(ctx));
+            IdGenerator.validateVersion(id, resource);
+            uuidV7TimestampValidator.validateNotInFuture(id, resource, workspaceId(ctx));
             return id;
         }));
     }
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/IdGenerator.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/IdGenerator.java
@@ -2,7 +2,9 @@ package com.comet.opik.domain;
 
 import com.comet.opik.api.error.InvalidUUIDException;
 import com.comet.opik.api.error.InvalidUUIDException.Reason;
+import com.comet.opik.infrastructure.auth.RequestContext;
 import com.comet.opik.infrastructure.db.UuidV7TimestampValidator;
+import com.comet.opik.infrastructure.metrics.ErrorMetricsResolver;
 import com.fasterxml.uuid.Generators;
 import com.fasterxml.uuid.impl.TimeBasedEpochGenerator;
 import com.google.inject.ImplementedBy;
@@ -11,6 +13,7 @@ import jakarta.inject.Singleton;
 import lombok.NonNull;
 import lombok.RequiredArgsConstructor;
 import reactor.core.publisher.Mono;
+import reactor.util.context.ContextView;
 
 import java.time.Instant;
 import java.util.UUID;
@@ -27,9 +30,11 @@ public interface IdGenerator {
     /**
      * Validates an ingested {@code id}: it must be a version 7 UUID ({@link #validateVersion(UUID, String)})
      * whose embedded timestamp is within the configured ingestion window.
-     * Rejects with HTTP 400. Encapsulates both data-quality checks behind one call.
+     * Rejects with HTTP 400 (or, in audit mode, counts + logs without rejecting). Encapsulates both
+     * data-quality checks behind one call. {@code workspaceId} tags the audit metric so offenders are
+     * attributable per workspace; pass {@link ErrorMetricsResolver#UNKNOWN} when it is not available.
      */
-    void validateId(UUID id, String resource);
+    void validateId(UUID id, String resource, String workspaceId);
 
     Mono<UUID> validateIdAsync(UUID id, String resource);
 
@@ -78,29 +83,38 @@ class IdGeneratorImpl implements IdGenerator {
     }
 
     @Override
-    public void validateId(@NonNull UUID id, String resource) {
+    public void validateId(@NonNull UUID id, String resource, String workspaceId) {
         IdGenerator.validateVersion(id, resource);
-        uuidV7TimestampValidator.validate(id);
+        uuidV7TimestampValidator.validate(id, resource, workspaceId);
     }
 
     @Override
     public Mono<UUID> validateIdAsync(@NonNull UUID id, String resource) {
-        return Mono.fromCallable(() -> {
-            validateId(id, resource);
+        return Mono.deferContextual(ctx -> Mono.fromCallable(() -> {
+            validateId(id, resource, workspaceId(ctx));
             return id;
-        });
+        }));
     }
 
-    private void validateIdForUpdate(UUID id, String resource) {
+    private void validateIdForUpdate(UUID id, String resource, String workspaceId) {
         IdGenerator.validateVersion(id, resource);
-        uuidV7TimestampValidator.validateNotInFuture(id);
+        uuidV7TimestampValidator.validateNotInFuture(id, resource, workspaceId);
     }
 
     @Override
     public Mono<UUID> validateIdForUpdateAsync(@NonNull UUID id, String resource) {
-        return Mono.fromCallable(() -> {
-            validateIdForUpdate(id, resource);
+        return Mono.deferContextual(ctx -> Mono.fromCallable(() -> {
+            validateIdForUpdate(id, resource, workspaceId(ctx));
             return id;
-        });
+        }));
+    }
+
+    /**
+     * Reads the {@code workspace_id} from the reactive context (the async ingestion paths carry it
+     * there, not in a request-scoped thread-local), falling back to {@link ErrorMetricsResolver#UNKNOWN}
+     * so the audit metric always has a value.
+     */
+    private static String workspaceId(ContextView ctx) {
+        return ctx.getOrDefault(RequestContext.WORKSPACE_ID, ErrorMetricsResolver.UNKNOWN);
     }
 }
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/IdGenerator.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/IdGenerator.java
@@ -28,11 +28,10 @@ public interface IdGenerator {
     UUID getTimeOrderedEpoch(long epochMilli);
 
     /**
-     * Validates an ingested {@code id}: it must be a version 7 UUID ({@link #validateVersion(UUID, String)})
-     * whose embedded timestamp is within the configured ingestion window.
-     * Rejects with HTTP 400 (or, in audit mode, counts + logs without rejecting). Encapsulates both
-     * data-quality checks behind one call. {@code workspaceId} tags the audit metric so offenders are
-     * attributable per workspace; pass {@link ErrorMetricsResolver#UNKNOWN} when it is not available.
+     * Validates an ingested {@code id}: it must be a version 7 UUID
+     * ({@link #validateVersion(UUID, String)}) whose embedded timestamp is within the configured
+     * ingestion window, throwing {@link InvalidUUIDException} otherwise. {@code workspaceId} attributes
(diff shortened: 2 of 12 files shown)


apps/opik-backend/src/main/java/com/comet/opik/domain/SpanService.java
```diff
@@ -393,7 +393,7 @@ public Mono<Long> create(@NonNull SpanBatch batch) {
                     Mono<List<Span>> resolveProjects = Flux.fromIterable(projectNames)
                             .flatMap(projectService::getOrCreate)
                             .collectList()
-                            .map(projects -> bindSpanToProjectAndId(dedupedSpans, projects));
+                            .map(projects -> bindSpanToProjectAndId(dedupedSpans, projects, workspaceId));
```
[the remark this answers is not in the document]
thiagohora: Addressed via the main merge (#7553) + reconciliation: span batches now validate every id in a fail-fast pass **before** any side effect (auto-stripped attachment deletion, project getOrCreate). I moved that pass into a `deferContextual` so it still runs before side effects while attributing the audit metric to the request workspace. Trace batches inherit #7553's flow (validation inside `bindTraceToProjectAndId`); can hoist traces too for full symmetry if preferred.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanService.java
```diff
@@ -438,7 +438,7 @@ private List<Span> dedupSpans(List<Span> initialSpans) {
         return result;
     }
 
-    private List<Span> bindSpanToProjectAndId(List<Span> spans, List<Project> projects) {
+    private List<Span> bindSpanToProjectAndId(List<Span> spans, List<Project> projects, String workspaceId) {
         Map<String, Project> projectPerName = projects.stream()
                 .collect(Collectors.toMap(
                         WorkspaceUtils::stripProjectName,
```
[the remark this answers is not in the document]
thiagohora: The bind duplication predates this PR — I kept my footprint to workspace threading and reverted `bindSpanToProjectAndId` to not re-validate (own-id validation is now hoisted to the fail-fast pass). Extracting a shared ingestion binder is a fair follow-up but out of scope here to avoid churning the span/trace-specific mapping.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/db/UuidV7TimestampValidatorTest.java
```diff
@@ -0,0 +1,124 @@
...
+    class Disabled {
+
+        private final UuidV7TimestampValidator validator = validator(false, false);
+
+        @Test
+        @DisplayName("accepts out-of-window ids on both paths")
+        void acceptsEverything() {
+            assertThatCode(() -> validator.validate(tooFarFutureId(), RESOURCE, WORKSPACE_ID))
+                    .doesNotThrowAnyException();
+            assertThatCode(() -> validator.validate(tooOldId(), RESOURCE, WORKSPACE_ID)).doesNotThrowAnyException();
```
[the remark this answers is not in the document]
thiagohora: Added `IdGeneratorAsyncValidationTest` covering `validateIdAsync` / `validateIdNotInFutureAsync` in both reject and audit modes, with and without `RequestContext.WORKSPACE_ID` in the reactive context (exercising the `deferContextual` lookup and the UNKNOWN fallback).

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/UuidV7TimestampValidator.java
```diff
@@ -82,9 +98,20 @@ private Optional<Pair<Reason, Instant>> evaluate(UUID id) {
...
+     * lets the write through. Otherwise throws {@link InvalidUUIDException} (HTTP 400).
+     */
+    private void handle(Pair<Reason, Instant> rejection, String resource, String workspaceId) {
         var reason = rejection.getLeft();
         var timestamp = rejection.getRight();
+        if (auditOnly) {
+            metrics.recordAudit(reason.getValue(), resource, workspaceId);
+            log.info(
+                    "UUIDv7 audit: would-reject {} id with embedded timestamp '{}' outside window '{}', reason '{}', workspace '{}'",
+                    resource, timestamp, window, reason.getValue(), workspaceId);
```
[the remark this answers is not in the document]
thiagohora: Fixed — the audit log now uses a fixed searchable prefix (`UUIDv7 audit: would-reject id, ...`) with `resource` and the other fields appended at the end.

apps/opik-backend/src/main/java/com/comet/opik/domain/IdGenerator.java
```diff
@@ -27,9 +30,11 @@ public interface IdGenerator {
     /**
      * Validates an ingested {@code id}: it must be a version 7 UUID ({@link #validateVersion(UUID, String)})
      * whose embedded timestamp is within the configured ingestion window.
-     * Rejects with HTTP 400. Encapsulates both data-quality checks behind one call.
+     * Rejects with HTTP 400 (or, in audit mode, counts + logs without rejecting). Encapsulates both
+     * data-quality checks behind one call. {@code workspaceId} tags the audit metric so offenders are
+     * attributable per workspace; pass {@link ErrorMetricsResolver#UNKNOWN} when it is not available.
      */
-    void validateId(UUID id, String resource);
+    void validateId(UUID id, String resource, String workspaceId);
```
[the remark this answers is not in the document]
thiagohora: Trimmed the `validateId` javadoc to validation semantics + `InvalidUUIDException`. Kept `workspaceId` on the interface, though: the sync/batch path is the primary ingestion route and is exactly where per-workspace attribution is needed (OPIK-7402), so falling back to UNKNOWN there would blank out the metric for the case that matters. The async paths self-resolve the workspace from the reactive context.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanService.java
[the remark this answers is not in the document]
thiagohora: Follow-up: made traces symmetric too — `create(TraceBatch)` now validates own ids in a leading `deferContextual` before `deleteAutoStrippedAttachments`/`getOrCreate`, and `bindTraceToProjectAndId` no longer re-validates. Both batch paths now fail fast before any side effect.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/UuidV7TimestampValidator.java
```diff
@@ -82,9 +98,22 @@ private Optional<Pair<Reason, Instant>> evaluate(UUID id) {
...
+    private void handle(Pair<Reason, Instant> rejection, String resource, String workspaceId) {
         var reason = rejection.getLeft();
         var timestamp = rejection.getRight();
+        if (auditOnly) {
+            metrics.recordAudit(reason.getValue(), resource, workspaceId);
+            // Keep a fixed, searchable prefix ("UUIDv7 audit: would-reject id ...") and append the
+            // variable fields at the end, so log searches match on the message rather than the values.
+            log.info(
+                    "UUIDv7 audit: would-reject id, embedded timestamp '{}' outside window '{}', reason '{}', resource '{}', workspace '{}'",
+                    timestamp, window, reason.getValue(), resource, workspaceId);
```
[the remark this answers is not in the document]
thiagohora: No problem, it's more important to know who than which row failed.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceService.java
```diff
@@ -251,8 +262,8 @@ private List<Trace> bindTraceToProjectAndId(List<Trace> traces, List<Project> pr
                     String projectName = WorkspaceUtils.getProjectName(trace.projectName());
```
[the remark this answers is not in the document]
thiagohora: This is not related to the scope of this change.

apps/opik-backend/src/main/java/com/comet/opik/domain/IdGenerator.java
```diff
@@ -93,31 +97,44 @@ public UUID getTimeOrderedEpoch(long epochMilli) {
...
         IdGenerator.validateVersion(id, resource);
-        uuidV7TimestampValidator.validate(id);
+        uuidV7TimestampValidator.validate(id, resource, workspaceId);
     }
 
     @Override
     public Mono<UUID> validateIdAsync(@NonNull UUID id, String resource) {
-        return Mono.fromCallable(() -> {
-            validateId(id, resource);
+        return Mono.deferContextual(ctx -> Mono.fromCallable(() -> {
```
andrescrz: Nit: do we need the `Mono.fromCallable` nested if we add `Mono.deferContextual`?

Applies to other occurrences in the PR.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/UuidV7TimestampValidator.java
```diff
@@ -27,40 +29,54 @@
...
+    private final boolean auditOnly;
     private final Duration window;
+    private final UuidValidationMetrics metrics;
 
     @Inject
-    public UuidV7TimestampValidator(@NonNull @Config("uuidValidation") UuidValidationConfig config) {
+    public UuidV7TimestampValidator(@NonNull @Config("uuidValidation") UuidValidationConfig config,
+            @NonNull UuidValidationMetrics metrics) {
         this.enabled = config.enabled();
+        this.auditOnly = config.auditOnly();
```
andrescrz, line 56: Maybe we can have a single flag of enum type with values: disabled, audit only, enabled etc. the logic might be easier than using two flags.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/UuidV7TimestampValidator.java
```diff
@@ -27,40 +29,54 @@
...
-     * (too old or too far in the future). No-op when validation is disabled or the id is acceptable.
-     * Used by the creation path.
+     * Rejects (HTTP 400) an id whose embedded timestamp is out of window (too old or too far in the
+     * future) in reject mode; counts + logs it without rejecting in audit mode; no-op when validation is
+     * disabled or the id is acceptable. Used by the creation path. {@code resource} (trace/span) and
+     * {@code workspaceId} are attached to the audit metric.
      */
-    public void validate(@NonNull UUID id) {
-        evaluate(id).ifPresent(this::reject);
+    public void validate(@NonNull UUID id, @NonNull String resource, String workspaceId) {
```
andrescrz: Please check the chain of invocation for `resource` argument. I believe this is not validated as non-null in IdGenerator. Then, it's only used to record the metric. An option is to fallback to unknown as we do for workspaceId. Ok to keep the validation if justified and if always can be resolved, but make sure all the chain of calls is consistent and validated everywhere, so we fail as early as possible.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/metrics/UuidValidationMetrics.java
```diff
@@ -0,0 +1,72 @@
...
+ * is not threaded (tagging it there is a follow-up), and instead carries {@code http_route}.
+ * <p>
+ * The {@code workspace_id} label is bounded by the set of clients actually emitting out-of-window ids
+ * (a small cohort), so it does not inflate metric cardinality the way an unconditional per-workspace
+ * label would.
+ */
+@Singleton
+public class UuidValidationMetrics {
+
+    private static final String METRIC_NAMESPACE = "opik.ingestion";
```
andrescrz, line 39: Constants and maybe logic, overlap with the pre-existing ones in `InvalidUUIDExceptionMapper`. The minimum you can do is making them public and reuse them from there.

Ideally, I'd remove the logic and constants from the other class and just delegate to this singleton as central place for all this.

If DI is an issue on the error mapper, maybe pull it in someway, or some alternative: instantiate, make static utils etc.

apps/opik-backend/src/test/java/com/comet/opik/domain/IdGeneratorAsyncValidationTest.java
```diff
@@ -0,0 +1,114 @@
...
+ * {@link RequestContext#WORKSPACE_ID}. Both cases must preserve the accept/reject behavior.
+ */
+@DisplayName("IdGenerator reactive validation")
+class IdGeneratorAsyncValidationTest {
+
+    private static final String RESOURCE = "trace";
+    private static final Duration WINDOW = Duration.hours(24);
+
+    // Reject mode: enabled=true, auditOnly=false (config-test default).
+    private final IdGenerator rejectGenerator = TestIdGeneratorFactory.create();
```
andrescrz: Minor: static final (and align name to that).

andrescrz (commented):
Looks pretty good! Let's double check the open comments before moving forward.

apps/opik-backend/src/main/java/com/comet/opik/domain/IdGenerator.java
thiagohora: Done — dropped the nested `Mono.fromCallable`; the async paths are now `Mono.deferContextual(ctx -> { validate(...); return Mono.just(id); })`. Checked for blocking first: `validate` only does bit-shift timestamp extraction + an in-memory counter increment + a log line (no I/O), and these validators aren't wrapped in `subscribeOn`, so `fromCallable` (without a scheduler) ran on the subscribing thread anyway — behavior-identical, and `deferContextual` still maps a thrown exception to `onError`. Applied to all occurrences.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/UuidV7TimestampValidator.java
thiagohora: I considered the enum but kept the two flags deliberately: `UUID_VALIDATION_ENABLED` is already live in prod/helm (set off in comet-helm#1146), so a single `mode` enum would mean an env/config migration plus a coordinated helm change to avoid breaking existing deployments. The effective state is a small derived mapping (enabled=false → disabled; enabled+auditOnly → audit; enabled only → reject) documented on the config record. Happy to switch to an enum as a follow-up if you'd prefer to take the migration then.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/UuidV7TimestampValidator.java
thiagohora: Good catch — made `resource` consistent with `workspaceId`: relaxed the `@NonNull` on the validator and let it fall back to `unknown` in the metric recorder (it's only used as a metric label). In practice it's always a constant (`SPAN_KEY`/`TRACE_KEY`), but this keeps the whole chain uniform rather than mixing a hard null-check with a soft fallback.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/metrics/UuidValidationMetrics.java
thiagohora: Done — centralized in `UuidValidationMetrics`: it now owns the `opik.ingestion.uuid_v7.rejected` instrument and all its label constants, and exposes `recordReject(reason, httpRoute)` alongside `recordAudit(...)`. `InvalidUUIDExceptionMapper` injects the singleton (DI works there — it's already `@Inject`-constructed) and delegates, so its duplicate counter, constants and namespace are gone.

apps/opik-backend/src/test/java/com/comet/opik/domain/IdGeneratorAsyncValidationTest.java
thiagohora: Done — `REJECT_GENERATOR` / `AUDIT_GENERATOR` are now `static final`.

apps/opik-backend/src/main/java/com/comet/opik/api/error/InvalidUUIDExceptionMapper.java
```diff
@@ -32,26 +31,17 @@
...
-            .build();
-
     private final Provider<ResourceInfo> resourceInfo;
+    private final UuidValidationMetrics uuidValidationMetrics;
 
     @Override
     public Response toResponse(@NonNull InvalidUUIDException exception) {
         var httpRoute = getHttpRoute();
         log.info("Rejected ingestion id, httpRoute: '{}', reason: '{}', error message: '{}'",
                 httpRoute, exception.getReason().getValue(), exception.getMessage());
```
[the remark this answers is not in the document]
thiagohora: Intentional. `InvalidUUIDException` is an expected client 400 (a malformed/out-of-window id), logged at INFO — the message already carries the id timestamp, window and reason, and there is no meaningful stack to surface for a validation rejection. Passing the throwable would attach a full stack trace to every rejected id at INFO, which is noise rather than signal here, so keeping it message-only.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/UuidV7TimestampValidator.java
```diff
@@ -27,40 +29,54 @@
...
     /**
-     * Throws {@link InvalidUUIDException} (HTTP 400) only when the id's embedded timestamp is too far in
-     * the future. Old ids are accepted, so updating a long-lived entity (e.g. created months ago) is
-     * never rejected. Used by the update path.
+     * Like {@link #validate}, but only acts when the embedded timestamp is too far in the future. Old
+     * ids are accepted, so updating a long-lived entity (e.g. created months ago) is never flagged. Used
+     * by the update path.
      */
-    public void validateNotInFuture(@NonNull UUID id) {
+    public void validateNotInFuture(@NonNull UUID id, String resource, String workspaceId) {
```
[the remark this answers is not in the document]
thiagohora: Not a public-API break: `UuidV7TimestampValidator` is an internal `@Singleton` whose only caller is `IdGenerator` (updated in this PR) — the build is green, there are no other callers. `workspaceId`/`resource` are needed for the audit-metric attribution, so a 2-arg shim would silently drop them; I'd rather not add one unless we actually expose this validator more widely.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/metrics/UuidValidationMetrics.java
```diff
@@ -0,0 +1,87 @@
...
+    public void recordAudit(String reason, String resource, String workspaceId) {
+        rejectedCounter.add(1, Attributes.builder()
+                .put(MODE_KEY, MODE_AUDIT)
+                .put(REASON_KEY, StringUtils.defaultIfBlank(reason, UNKNOWN))
+                .put(RESOURCE_KEY, StringUtils.defaultIfBlank(resource, UNKNOWN))
+                .put(WORKSPACE_ID_KEY, StringUtils.defaultIfBlank(workspaceId, UNKNOWN))
+                .build());
+    }
+
+    /**
```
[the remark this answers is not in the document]
thiagohora: Minor — the two methods are short and only diverge in their third tag; I've kept them explicit for readability over a shared builder helper. Can factor out a small private helper if you'd prefer, but leaving as-is for now.

apps/opik-backend/src/main/java/com/comet/opik/domain/IdGenerator.java
```diff
@@ -93,33 +97,46 @@ public UUID getTimeOrderedEpoch(long epochMilli) {
...
     }
 
     @Override
     public Mono<UUID> validateIdNotInFutureAsync(@NonNull UUID id, String resource) {
-        return Mono.fromCallable(() -> {
-            validateIdNotInFuture(id, resource);
-            return id;
+        return Mono.deferContextual(ctx -> {
+            IdGenerator.validateVersion(id, resource);
+            uuidV7TimestampValidator.validateNotInFuture(id, resource, workspaceId(ctx));
```
andrescrz: Not a blocker: previously, all the logic was centralised in a single method, and the different flavours where just calling that method. Now, you're duplicating the core logic (two calls: `IdGenerator.validateVersion` and `uuidV7TimestampValidator.validateNotInFuture`), which will require maintaining both.

I'd personally just offer appropriate overloaded methods to keep that centralisation. The only blocker is that you don't parameterise the workspace and hardcode it to `UNKNOWN`. If you offer an overloaded flavour, this problem would be gone.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/UuidV7TimestampValidator.java
andrescrz: No need for coordination. This flag defaults to false and the feature is not used in any installation yet. You could just default the new flag to disabled (or audit only), which is what you should do anyway, and then we could clean up the old helm flag.

Not a blocker anyway.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/metrics/UuidValidationMetrics.java
```diff
@@ -0,0 +1,87 @@
...
+    private static final String METRIC_NAMESPACE = "opik.ingestion";
+    private static final String COUNTER_DESCRIPTION = "Number of writes rejected because the id failed UUIDv7 ingestion validation";
+
+    public static final String MODE_AUDIT = "audit";
+    public static final String MODE_REJECT = "reject";
+
+    public static final AttributeKey<String> MODE_KEY = AttributeKey.stringKey("mode");
+    public static final AttributeKey<String> REASON_KEY = AttributeKey.stringKey("reason");
+    public static final AttributeKey<String> RESOURCE_KEY = AttributeKey.stringKey("resource");
+    public static final AttributeKey<String> HTTP_ROUTE_KEY = AttributeKey.stringKey("http_route");
```
andrescrz, line 48: What's the reason for having `resource` and `http_route` attributes? They seem to be the same thing to me. You just use one or another depending on the `recordAudit` and `recordReject` calls. At the end of the day what matters is to discriminate by `mode`, the rest of attributes should be the same unless I'm missing something.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/metrics/UuidValidationMetrics.java
```diff
@@ -0,0 +1,87 @@
...
+                .build());
+    }
+
+    /**
+     * Records one enforced rejection (HTTP 400), delegated from {@link
+     * com.comet.opik.api.error.InvalidUUIDExceptionMapper}. Tagged by {@code reason} and the matched
+     * {@code http_route}; {@code workspace_id} is intentionally omitted (not threaded on that path). All
+     * fall back to {@code unknown}.
+     */
+    public void recordReject(String reason, String httpRoute) {
```
andrescrz, line 78: Nit and related to my other comment: at implementation level both methods are the same, you could just parameterise the `mode`. If instead you want a particular nomenclature for each entry point (audit vs reject) without users requiring to pass the right param, you could keep these two names, but move the implementation to a core method (private or not).

andrescrz (commented):
My only blocker is the potential overlap between multidimensional attributes on the metric, which should try to avoid that.

The others are nice to have, but not blocker comments.

apps/opik-backend/src/main/java/com/comet/opik/domain/IdGenerator.java
thiagohora: Done. Centralized: there's now a single core `validateIdNotInFuture(id, resource, workspaceId)`; the 2-arg form and the async form both delegate to it (no more duplicated validateVersion+validateNotInFuture), and I added the workspace-parameterized overload you suggested (plus for validateIdNotInFutureIfPresent). So the workspace is no longer hardcoded — callers that carry it (e.g. ingestion) can pass it; the ~25 config-entity callers keep the 2-arg form and default to unknown by choice rather than by a baked-in constant.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/db/UuidV7TimestampValidator.java
thiagohora: Fair — you're right that there's no coordination cost since the flag is off everywhere and the feature isn't live yet. Keeping the additive `auditOnly` boolean for now (Thiago's call), but with that constraint gone we can switch to a single `mode` enum and drop the old helm flag as a fast follow-up. Noted as non-blocking.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/metrics/UuidValidationMetrics.java
thiagohora: They're related but not identical, and they come from different layers: `resource` is the entity kind the validator sees (Span/Trace, and for foreign refs 'Span trace'/'Span parent'), while `http_route` is the matched endpoint template the exception mapper sees (e.g. /v1/private/traces/batch vs single). The audit path (deep in the validator) has no request/route info, and the reject path (mapper) has no entity-kind handy — so each carries the discriminator available at its layer, with `mode` as the primary split. I kept `http_route` because the existing 'rejections by route' panel relies on it. Happy to collapse to a single shared label if you'd rather have a uniform schema — that'd mean dropping the route breakdown.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/metrics/UuidValidationMetrics.java
thiagohora: Done — factored the shared counter/`mode`/`reason` assembly into a private `record(mode, reason, attributesBuilder)`; `recordAudit`/`recordReject` now only supply their path-specific tags (workspace_id/resource vs http_route). Kept the two named entry points so callers don't pass a mode string.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/metrics/UuidValidationMetrics.java
andrescrz: This clarifies it. Thanks!

andrescrz (approved):
LGTM

PR #7583, opened 2026-07-23
[OPIK-7448] [BE] test: MySQL query plan-shape gate (EXPLAIN regression guard)

JetoPistola:
## Details

image

Adds a **query plan-shape regression gate** for the MySQL read paths (slice 1 of OPIK-7448). Result identity is already covered by the resource tests; this gate covers plan **shape** — the dimension a functional test can't assert.

A JDBI `SqlLogger` captures every rendered `SELECT` exercised by the MySQL-backed read endpoints and runs `EXPLAIN FORMAT=JSON` on the **same live connection**, re-applying the exact bindings via JDBI's own `Argument` machinery (no value-extraction guesswork). EXPLAIN is requested in **JSON format version 2** (`explain_json_format_version=2`): v1 reports the table *alias* in `table_name` (so a base-table match silently misses `FROM prompts AS p`), while v2 reports the base table name and a uniform `operation`/`access_type` vocabulary. The resulting plan is inspected for:

- **Materialized subquery / internal temporary table** (`access_type="materialize"`) — the OPIK-7198 / 7197 / 6735 class (`SQLSyntaxErrorException: Table '#sql...' doesn't exist` under TempTable-pool saturation), which is non-deterministic and load-dependent, so a functional test can't reproduce it.
- **Full table scan** (`access_type="table"` on a base table; indexed reads are `"index"`) on tenant-growing tables — a latency cliff at scale.

Enforcement is **net-new only** vs. a checked-in baseline allowlist (`planshape/mysql-baseline.json`) that ratchets down as legacy offenders are fixed. A SELECT whose EXPLAIN fails is recorded in a `failedSql` set and asserted empty, so an un-vetted query can't pass the gate by omission.

**The gate already earned its keep:** on the current tree it detected that `PromptDAO`'s list/count wraps a derived table MySQL materializes — the same mechanism as OPIK-7198. Those 2 fingerprints are baselined and tracked for ratchet-down under **OPIK-7450**.

Test-scope only — **no production code changes**, and the capture is installed on the injected `Jdbi` inside a single dedicated test, so there's zero blast radius on other tests.

### New files
- `CapturingSqlLogger` — SqlLogger that EXPLAINs each captured SELECT under v2 format; records failures separately (reusable by the planned full-suite slice).
- `MySqlPlanShapeAsserter` — recursive v2 EXPLAIN-JSON walk detecting materialize + full-scan.
- `PlanShapeViolation` — `@Builder` record with a stable SQL fingerprint (literals → `?`) so run-to-run UUIDs don't perturb identity.
- `PlanShapeBaseline` — loads the allowlist, computes net-new.
- `MySqlQueryPlanGateTest` 🐬 — drives the read endpoints and asserts no net-new violations and no un-vetted queries.
- `planshape/mysql-baseline.json` — seeded with the 2 pre-existing PromptDAO fingerprints.

## Change checklist
- [x] Tests added (the gate itself is the test)
- [x] `spotless` passing
- [x] No production code changed
- [ ] Docs — n/a (internal test infra)

## Issues
- Closes the MySQL portion of **OPIK-7448**.
- Absorbs **OPIK-7198** as the first concrete MySQL assertion.
- Ratchet-down follow-up: **OPIK-7450** (PromptDAO derived-table materialization).
- Coverage follow-up: **OPIK-7452** (exercise non-default SQL branches).
- CI wiring + sticky PR comment follow-up: **OPIK-7451** (advisory-first).

**Deliberately deferred to follow-ups** (scope agreed as slice 1 = MySQL):
- Full-suite passive capture (move the `SqlLogger` into shared test wiring so every resource test contributes queries).
- ClickHouse gate (`EXPLAIN indexes=1, json=1` partition-pruning assertions).
- Dedicated CI "Query Plan Gate" required check with the broad path filter (OPIK-7451).

## Testing
`mvn test -Dtest=MySqlQueryPlanGateTest` → green (1/1). Verified the gate **fails** on net-new violations by running with an empty baseline (surfaced the PromptDAO fingerprints), then confirmed green after baselining. The v2 alias fix and full-scan/materialize detection were verified directly against `mysql:8.4.2`.

## Documentation
No user-facing docs. Rationale and the ratchet-down contract live in the class Javadoc and the baseline notes.

### DIFF
--- a/apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/CapturingSqlLogger.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/CapturingSqlLogger.java
@@ -1,0 +1,88 @@
+package com.comet.opik.api.resources.utils.planshape;
+
+import lombok.extern.slf4j.Slf4j;
+import org.jdbi.v3.core.argument.Argument;
+import org.jdbi.v3.core.statement.SqlLogger;
+import org.jdbi.v3.core.statement.StatementContext;
+
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.util.Locale;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * JDBI {@link SqlLogger} that captures the {@code EXPLAIN FORMAT=JSON} plan of every {@code SELECT} rendered while it is
+ * installed, keyed by the rendered SQL. It reuses JDBI's own {@link Argument} binding machinery to re-apply the exact
+ * parameters onto the EXPLAIN prepared statement on the same live connection, so no value extraction or re-binding
+ * guesswork is needed. Non-SELECT statements are ignored.
+ *
+ * <p>This is the shared MySQL capture foundation for the query plan-shape gate (OPIK-7448). The plan JSON it collects is
+ * consumed by {@link MySqlPlanShapeAsserter}. Deduplicating by rendered SQL keeps the captured set bounded even when a
+ * query runs many times across a test.
+ */
+@Slf4j
+public class CapturingSqlLogger implements SqlLogger {
+
+    private final Map<String, String> plansByRenderedSql = new ConcurrentHashMap<>();
+
+    @Override
+    public void logAfterExecution(StatementContext context) {
+        var renderedSql = context.getRenderedSql();
+        if (renderedSql == null || !isSelect(renderedSql)) {
+            return;
+        }
+
+        plansByRenderedSql.computeIfAbsent(renderedSql, sql -> explain(context));
+    }
+
+    private String explain(StatementContext context) {
+        var parsedSql = context.getParsedSql();
+        var jdbcSql = parsedSql.getSql();
+        var parameterNames = parsedSql.getParameters().getParameterNames();
+        var binding = context.getBinding();
+
+        try (PreparedStatement explainStatement = context.getConnection()
+                .prepareStatement("EXPLAIN FORMAT=JSON " + jdbcSql)) {
+
+            for (int i = 0; i < parameterNames.size(); i++) {
+                int position = i;
+                String parameterName = parameterNames.get(i);
+                Argument argument = binding.findForName(parameterName, context)
+                        .or(() -> binding.findForPosition(position))
+                        .orElseThrow(() -> new IllegalStateException(
+                                "No bound argument for parameter '%s' in: %s".formatted(parameterName, jdbcSql)));
+                // JDBC positions are 1-based; JDBI Arguments apply themselves onto the target statement.
+                argument.apply(position + 1, explainStatement, context);
+            }
+
+            try (ResultSet resultSet = explainStatement.executeQuery()) {
+                if (resultSet.next()) {
+                    return resultSet.getString(1);
+                }
+                return null;
+            }
+        } catch (Exception e) {
+            // A query the gate cannot EXPLAIN (e.g. references a session TEMPORARY TABLE that only exists mid-flight)
+            // must not fail the capturing test run; it is simply left uncaptured and the asserter never sees it.
+            log.debug("Skipping EXPLAIN for query [{}]: {}", jdbcSql, e.getMessage());
+            return null;
+        }
+    }
+
+    private static boolean isSelect(String sql) {
+        return sql.stripLeading().toLowerCase(Locale.ROOT).startsWith("select");
+    }
+
+    /**
+     * @return an immutable snapshot of captured plans, keyed by rendered SQL. Entries whose EXPLAIN failed hold a
+     *         {@code null} plan and are filtered by the asserter.
+     */
+    public Map<String, String> capturedPlans() {
+        return Map.copyOf(plansByRenderedSql);
+    }
+
+    public void clear() {
+        plansByRenderedSql.clear();
+    }
+}
--- a/apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/CapturingSqlLogger.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/CapturingSqlLogger.java
@@ -7,8 +7,10 @@ import org.jdbi.v3.core.statement.StatementContext;
 
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
+import java.sql.Statement;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
 /**
@@ -17,14 +19,29 @@ import java.util.concurrent.ConcurrentHashMap;
  * parameters onto the EXPLAIN prepared statement on the same live connection, so no value extraction or re-binding
  * guesswork is needed. Non-SELECT statements are ignored.
  *
- * <p>This is the shared MySQL capture foundation for the query plan-shape gate (OPIK-7448). The plan JSON it collects is
- * consumed by {@link MySqlPlanShapeAsserter}. Deduplicating by rendered SQL keeps the captured set bounded even when a
- * query runs many times across a test.
+ * <p>EXPLAIN is requested in JSON <b>format version 2</b> ({@code explain_json_format_version = 2}): version 1 puts the
+ * table <i>alias</i> in {@code table_name} (so a base-table match like {@code prompts} silently misses {@code FROM
+ * prompts AS p}), while version 2 reports the base table name and a uniform {@code operation}/{@code access_type}
+ * vocabulary for both materialization and full scans. {@link MySqlPlanShapeAsserter} depends on the v2 shape.</p>
+ *
+ * <p>This is the shared MySQL capture foundation for the query plan-shape gate (OPIK-7448). Deduplicating by rendered
+ * SQL keeps the captured set bounded even when a query runs many times. A SELECT whose EXPLAIN throws is recorded in
+ * {@link CapturedQueries#failedSql()} rather than silently dropped, so the gate can surface queries it could not vet
+ * instead of treating them as clean.</p>
  */
 @Slf4j
 public class CapturingSqlLogger implements SqlLogger {
 
     private final Map<String, String> plansByRenderedSql = new ConcurrentHashMap<>();
+    private final Set<String> failedSql = ConcurrentHashMap.newKeySet();
+
+    /**
+     * @param plansByRenderedSql successfully-captured {@code EXPLAIN FORMAT=JSON} plans, keyed by rendered SQL.
+     * @param failedSql          rendered SELECTs whose EXPLAIN threw — captured but un-vetted, surfaced so they can't
+     *                           pass the gate by omission.
+     */
+    public record CapturedQueries(Map<String, String> plansByRenderedSql, Set<String> failedSql) {
+    }
 
     @Override
     public void logAfterExecution(StatementContext context) {
@@ -32,41 +49,61 @@ public class CapturingSqlLogger implements SqlLogger {
         if (renderedSql == null || !isSelect(renderedSql)) {
             return;
         }
+        if (plansByRenderedSql.containsKey(renderedSql) || failedSql.contains(renderedSql)) {
+            return;
+        }
 
-        plansByRenderedSql.computeIfAbsent(renderedSql, sql -> explain(context));
+        var plan = explain(context, renderedSql);
+        if (plan != null) {
+            plansByRenderedSql.put(renderedSql, plan);
+        }
     }
 
-    private String explain(StatementContext context) {
+    private String explain(StatementContext context, String renderedSql) {
         var parsedSql = context.getParsedSql();
         var jdbcSql = parsedSql.getSql();
         var parameterNames = parsedSql.getParameters().getParameterNames();
         var binding = context.getBinding();
 
-        try (PreparedStatement explainStatement = context.getConnection()
-                .prepareStatement("EXPLAIN FORMAT=JSON " + jdbcSql)) {
+        try {
+            forceV2ExplainFormat(context);
 
-            for (int i = 0; i < parameterNames.size(); i++) {
-                int position = i;
-                String parameterName = parameterNames.get(i);
-                Argument argument = binding.findForName(parameterName, context)
-                        .or(() -> binding.findForPosition(position))
-                        .orElseThrow(() -> new IllegalStateException(
-                                "No bound argument for parameter '%s' in: %s".formatted(parameterName, jdbcSql)));
-                // JDBC positions are 1-based; JDBI Arguments apply themselves onto the target statement.
-                argument.apply(position + 1, explainStatement, context);
-            }
+            try (PreparedStatement explainStatement = context.getConnection()
+                    .prepareStatement("EXPLAIN FORMAT=JSON " + jdbcSql)) {
 
-            try (ResultSet resultSet = explainStatement.executeQuery()) {
-                if (resultSet.next()) {
-                    return resultSet.getString(1);
+                for (int i = 0; i < parameterNames.size(); i++) {
+                    int position = i;
+                    String parameterName = parameterNames.get(i);
+                    Argument argument = binding.findForName(parameterName, context)
+                            .or(() -> binding.findForPosition(position))
+                            .orElseThrow(() -> new IllegalStateException(
+                                    "No bound argument for parameter '%s' in: %s".formatted(parameterName, jdbcSql)));
+                    // JDBC positions are 1-based; JDBI Arguments apply themselves onto the target statement.
+                    argument.apply(position + 1, explainStatement, context);
                 }
-                return null;
+
+                try (ResultSet resultSet = explainStatement.executeQuery()) {
+                    if (resultSet.next()) {
+                        return resultSet.getString(1);
+                    }
+                    failedSql.add(renderedSql);
+                    return null;
+                }
             }
         } catch (Exception e) {
-            // A query the gate cannot EXPLAIN (e.g. references a session TEMPORARY TABLE that only exists mid-flight)
-            // must not fail the capturing test run; it is simply left uncaptured and the asserter never sees it.
-            log.debug("Skipping EXPLAIN for query [{}]: {}", jdbcSql, e.getMessage());
+            // A query the gate cannot EXPLAIN (e.g. one referencing a session TEMPORARY TABLE that only exists
+            // mid-flight) is recorded as failed rather than dropped, so the gate can report it instead of passing it
+            // by omission.
+            log.debug("EXPLAIN failed for query [{}]: {}", jdbcSql, e.getMessage());
+            failedSql.add(renderedSql);
             return null;
+        }
+    }
+
+    // v2 puts the base table name in table_name (v1 uses the alias) — required for the full-scan check to match.
+    private void forceV2ExplainFormat(StatementContext context) throws Exception {
+        try (Statement statement = context.getConnection().createStatement()) {
+            statement.execute("SET explain_json_format_version = 2");
         }
     }
 
@@ -74,15 +111,12 @@ public class CapturingSqlLogger implements SqlLogger {
         return sql.stripLeading().toLowerCase(Locale.ROOT).startsWith("select");
     }
 
-    /**
-     * @return an immutable snapshot of captured plans, keyed by rendered SQL. Entries whose EXPLAIN failed hold a
-     *         {@code null} plan and are filtered by the asserter.
-     */
-    public Map<String, String> capturedPlans() {
-        return Map.copyOf(plansByRenderedSql);
+    public CapturedQueries captured() {
+        return new CapturedQueries(Map.copyOf(plansByRenderedSql), Set.copyOf(failedSql));
     }
 
     public void clear() {
         plansByRenderedSql.clear();
+        failedSql.clear();
     }
 }
--- a/apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/CapturingSqlLogger.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/CapturingSqlLogger.java
@@ -46,7 +46,7 @@ public class CapturingSqlLogger implements SqlLogger {
     @Override
     public void logAfterExecution(StatementContext context) {
         var renderedSql = context.getRenderedSql();
-        if (renderedSql == null || !isSelect(renderedSql)) {
+        if (renderedSql == null || !isReadQuery(renderedSql)) {
             return;
         }
         if (plansByRenderedSql.containsKey(renderedSql) || failedSql.contains(renderedSql)) {
@@ -107,8 +107,12 @@ public class CapturingSqlLogger implements SqlLogger {
         }
     }
 
-    private static boolean isSelect(String sql) {
-        return sql.stripLeading().toLowerCase(Locale.ROOT).startsWith("select");
+    // Read paths the gate vets: plain SELECTs and CTE reads (WITH ... SELECT). Matching only "select" would skip the
+    // CTE class entirely — exactly the multi-reference-CTE shape (OPIK-7198) the gate exists to catch. No WITH-prefixed
+    // @SqlUpdate writes exist in the backend, and MySQL EXPLAINs WITH ... SELECT the same as a plain SELECT.
+    private static boolean isReadQuery(String sql) {
+        var normalized = sql.stripLeading().toLowerCase(Locale.ROOT);
+        return normalized.startsWith("select") || normalized.startsWith("with");
     }
 
     public CapturedQueries captured() {
--- a/apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/CapturingSqlLogger.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/CapturingSqlLogger.java
@@ -109,7 +109,7 @@ public class CapturingSqlLogger implements SqlLogger {
 
     // Read paths the gate vets: plain SELECTs and CTE reads (WITH ... SELECT). Matching only "select" would skip the
     // CTE class entirely — exactly the multi-reference-CTE shape (OPIK-7198) the gate exists to catch. No WITH-prefixed
-    // @SqlUpdate writes exist in the backend, and MySQL EXPLAINs WITH ... SELECT the same as a plain SELECT.
+    // @SqlUpdate writes exist in the backend, and MySQL EXPLAIN handles WITH ... SELECT the same as a plain SELECT.
     private static boolean isReadQuery(String sql) {
         var normalized = sql.stripLeading().toLowerCase(Locale.ROOT);
         return normalized.startsWith("select") || normalized.startsWith("with");
--- a/apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/CapturingSqlLogger.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/CapturingSqlLogger.java
@@ -13,6 +13,7 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicLong;
 
 /**
  * JDBI {@link SqlLogger} that captures the {@code EXPLAIN FORMAT=JSON} plan of every read query rendered while it is
@@ -35,6 +36,7 @@ public class CapturingSqlLogger implements SqlLogger {
 
     private final Map<String, String> plansByRenderedSql = new ConcurrentHashMap<>();
     private final Set<String> failedSql = ConcurrentHashMap.newKeySet();
+    private final AtomicLong readsObserved = new AtomicLong();
 
     /**
      * Result of a capture session.
@@ -57,6 +59,11 @@ public class CapturingSqlLogger implements SqlLogger {
         if (!isReadQuery(renderedSql)) {
             return;
         }
+
+        // Counted before the dedup check so it reflects every read query actually executed, not just the ones that were
+        // new. Callers use it to tell "this code path ran no MySQL read" apart from "it ran one already captured".
+        readsObserved.incrementAndGet();
+
         if (plansByRenderedSql.containsKey(renderedSql) || failedSql.contains(renderedSql)) {
             return;
         }
@@ -136,8 +143,19 @@ public class CapturingSqlLogger implements SqlLogger {
         return new CapturedQueries(Map.copyOf(plansByRenderedSql), Set.copyOf(failedSql));
     }
 
+    /**
+     * Total read queries seen since construction, counted before deduplication. Unlike
+     * {@link CapturedQueries#plansByRenderedSql()}, this grows for every read query executed even when its SQL was
+     * already captured, so it is order-independent: a code path that legitimately re-runs SQL an earlier path already
+     * captured still moves this counter.
+     */
+    public long readsObserved() {
+        return readsObserved.get();
+    }
+
     public void clear() {
         plansByRenderedSql.clear();
         failedSql.clear();
+        readsObserved.set(0);
     }
 }
--- a/apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/CapturingSqlLogger.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/CapturingSqlLogger.java
@@ -1,5 +1,6 @@
 package com.comet.opik.api.resources.utils.planshape;
 
+import lombok.Builder;
 import lombok.extern.slf4j.Slf4j;
 import org.jdbi.v3.core.argument.Argument;
 import org.jdbi.v3.core.statement.SqlLogger;
@@ -14,10 +15,10 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
 /**
- * JDBI {@link SqlLogger} that captures the {@code EXPLAIN FORMAT=JSON} plan of every {@code SELECT} rendered while it is
+ * JDBI {@link SqlLogger} that captures the {@code EXPLAIN FORMAT=JSON} plan of every read query rendered while it is
  * installed, keyed by the rendered SQL. It reuses JDBI's own {@link Argument} binding machinery to re-apply the exact
  * parameters onto the EXPLAIN prepared statement on the same live connection, so no value extraction or re-binding
- * guesswork is needed. Non-SELECT statements are ignored.
+ * guesswork is needed. Non-read statements are ignored.
  *
  * <p>EXPLAIN is requested in JSON <b>format version 2</b> ({@code explain_json_format_version = 2}): version 1 puts the
  * table <i>alias</i> in {@code table_name} (so a base-table match like {@code prompts} silently misses {@code FROM
@@ -25,8 +26,8 @@ import java.util.concurrent.ConcurrentHashMap;
  * vocabulary for both materialization and full scans. {@link MySqlPlanShapeAsserter} depends on the v2 shape.</p>
  *
  * <p>This is the shared MySQL capture foundation for the query plan-shape gate (OPIK-7448). Deduplicating by rendered
- * SQL keeps the captured set bounded even when a query runs many times. A SELECT whose EXPLAIN throws is recorded in
- * {@link CapturedQueries#failedSql()} rather than silently dropped, so the gate can surface queries it could not vet
+ * SQL keeps the captured set bounded even when a query runs many times. A read query whose EXPLAIN throws is recorded
+ * in {@link CapturedQueries#failedSql()} rather than silently dropped, so the gate can surface queries it could not vet
  * instead of treating them as clean.</p>
  */
 @Slf4j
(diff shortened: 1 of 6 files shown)


apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/PlanShapeViolation.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.Locale;
+
+/**
+ * A single plan-shape regression found in a captured query. The {@link #fingerprint()} is the stable identity used to
+ * match a violation against the checked-in baseline: it normalizes the SQL (collapse whitespace, strip literals) so
+ * that run-to-run parameter differences (random UUIDs, timestamps) do not change identity, while a genuinely new or
+ * structurally-changed query does.
+ */
+public record PlanShapeViolation(@NonNull String renderedSql, @NonNull Type type, @NonNull String detail) {
```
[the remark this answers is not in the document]
JetoPistola: Fixed in 0659e1a578 — added `@Builder(toBuilder = true)` to `PlanShapeViolation`.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/MySqlPlanShapeAsserter.java
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        if (node.isObject()) {
+            if (node.has(MATERIALIZED_FROM_SUBQUERY)) {
+                violations.add(new PlanShapeViolation(renderedSql, PlanShapeViolation.Type.MATERIALIZED_SUBQUERY,
+                        "Plan materializes a subquery into an internal temporary table (OPIK-7198 class)."));
+            }
+            if (node.path(USING_TEMPORARY_TABLE).asBoolean(false)) {
+                violations.add(new PlanShapeViolation(renderedSql, PlanShapeViolation.Type.TEMPORARY_TABLE,
+                        "Plan uses an internal temporary table (OPIK-7198 class)."));
```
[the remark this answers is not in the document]
JetoPistola: Fixed in 0659e1a578 — `MySqlPlanShapeAsserter` now builds violations via `PlanShapeViolation.builder()` instead of the positional constructor.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/PlanShapeBaseline.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * The checked-in allowlist of plan-shape violations that already exist in the codebase. The gate blocks only on
+ * <b>net-new</b> violations vs. this baseline (see {@link #netNew(List)}); the allowlist is ratcheted down as legacy
+ * offenders are fixed. Each entry is a {@link PlanShapeViolation#fingerprint()} plus a human note explaining why it is
+ * tolerated / which ticket tracks its removal.
+ */
+public class PlanShapeBaseline {
+
+    public record Entry(@NonNull String fingerprint, String note) {
```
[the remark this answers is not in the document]
JetoPistola: Fixed in 0659e1a578 — added `@Builder(toBuilder = true)` to `PlanShapeBaseline.Entry`.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/MySqlPlanShapeAsserter.java
```diff
@@ -0,0 +1,82 @@
...
+        }
+    }
+
+    private void checkTableAccess(JsonNode tableNode, String renderedSql, List<PlanShapeViolation> violations) {
+        var accessType = tableNode.path(ACCESS_TYPE).asText(null);
+        var tableName = tableNode.path(TABLE_NAME).asText(null);
+        if (FULL_SCAN.equals(accessType) && tableName != null && fullScanSensitiveTables.contains(tableName)) {
+            violations.add(new PlanShapeViolation(renderedSql, PlanShapeViolation.Type.FULL_TABLE_SCAN,
+                    "Full table scan (access_type=ALL) on tenant-growing table '%s'.".formatted(tableName)));
+        }
```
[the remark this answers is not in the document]
JetoPistola: Fixed in 0659e1a578 — good catch, verified against `mysql:8.4.2` (the CI testcontainer image): with the default `explain_json_format_version=1`, `table_name` holds the **alias** (`p`), so the base-table match never fired — the full-scan guard was effectively dead for aliased queries.

Rather than drop the guard, `CapturingSqlLogger` now runs EXPLAIN under `explain_json_format_version=2`, where `table_name` is the base table name and a full scan is `access_type="table"` (indexed reads are `"index"`). `MySqlPlanShapeAsserter` was rewritten for the v2 shape, which also keeps materialize/temp-table detection (`access_type="materialize"`). So the full-scan guard is now live and correct, not bypassed.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/CapturingSqlLogger.java
```diff
@@ -0,0 +1,88 @@
...
+                    return resultSet.getString(1);
+                }
+                return null;
+            }
+        } catch (Exception e) {
+            // A query the gate cannot EXPLAIN (e.g. references a session TEMPORARY TABLE that only exists mid-flight)
+            // must not fail the capturing test run; it is simply left uncaptured and the asserter never sees it.
+            log.debug("Skipping EXPLAIN for query [{}]: {}", jdbcSql, e.getMessage());
+            return null;
+        }
```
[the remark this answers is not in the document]
JetoPistola: Fixed in 0659e1a578 — `explain()` failures are now recorded in a `failedSql` set (via `captured().failedSql()`) instead of being dropped by `computeIfAbsent`'s null. The gate asserts `failedSql` is empty, so a SELECT the gate could not EXPLAIN can no longer pass by omission — it fails the test instead.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/MySqlQueryPlanGateTest.java
```diff
@@ -0,0 +1,147 @@
...
+                .queryParam("page", 1)
+                .queryParam("size", 100)
+                .request()
+                .accept(MediaType.APPLICATION_JSON_TYPE)
+                .header(HttpHeaders.AUTHORIZATION, API_KEY)
+                .header(WORKSPACE_HEADER, TEST_WORKSPACE)
+                .get()) {
+
+            assertThat(response.getStatus())
+                    .as("read endpoint %s must return 200 so its query is actually rendered", path)
```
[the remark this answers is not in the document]
JetoPistola: Skipping in this PR — valid coverage gap, deferred to **OPIK-7452**. This PR is the foundation slice (OPIK-7448); exercising the `name`/`sorting`/`filters`/`type`/`project_id` branches per endpoint is a focused follow-up so the gate can catch regressions in the optional-param SQL branches too.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/CapturingSqlLogger.java
```diff
@@ -0,0 +1,122 @@
...
+    public record CapturedQueries(Map<String, String> plansByRenderedSql, Set<String> failedSql) {
+    }
+
+    @Override
+    public void logAfterExecution(StatementContext context) {
+        var renderedSql = context.getRenderedSql();
+        if (renderedSql == null || !isSelect(renderedSql)) {
+            return;
+        }
+        if (plansByRenderedSql.containsKey(renderedSql) || failedSql.contains(renderedSql)) {
```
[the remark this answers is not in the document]
JetoPistola: Fixed in 8677ae8093 — sharp catch, and it's the highest-value one: the predicate now accepts a leading `WITH` too (renamed `isReadQuery`), so CTE reads like `PromptDAO`'s `WITH selected_prompts …` are captured and EXPLAINed instead of silently skipped — closing a hole in the exact OPIK-7198 multi-reference-CTE class the gate targets.

Verified against `mysql:8.4.2`: no `WITH`-prefixed `@SqlUpdate` writes exist in the backend, and MySQL EXPLAINs `WITH … SELECT` the same as a plain SELECT. The newly-captured CTE reads EXPLAIN cleanly and add no net-new violations.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/CapturingSqlLogger.java
```diff
@@ -0,0 +1,126 @@
...
+    // v2 puts the base table name in table_name (v1 uses the alias) — required for the full-scan check to match.
+    private void forceV2ExplainFormat(StatementContext context) throws Exception {
+        try (Statement statement = context.getConnection().createStatement()) {
+            statement.execute("SET explain_json_format_version = 2");
+        }
+    }
+
+    // Read paths the gate vets: plain SELECTs and CTE reads (WITH ... SELECT). Matching only "select" would skip the
+    // CTE class entirely — exactly the multi-reference-CTE shape (OPIK-7198) the gate exists to catch. No WITH-prefixed
+    // @SqlUpdate writes exist in the backend, and MySQL EXPLAINs WITH ... SELECT the same as a plain SELECT.
```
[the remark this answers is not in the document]
JetoPistola: Fixed in 5d20452bb5 — reworded to "MySQL EXPLAIN handles `WITH … SELECT` the same as a plain SELECT".

CometActions:

### Access Information
- **URL:** https://pr-7583.dev.comet.com
- **Namespace:** pr-7583
- **Version:** 2.2.4-7583-merge-2654
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7583%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7583`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/CapturingSqlLogger.java
```diff
@@ -0,0 +1,126 @@
...
+
+    private final Map<String, String> plansByRenderedSql = new ConcurrentHashMap<>();
+    private final Set<String> failedSql = ConcurrentHashMap.newKeySet();
+
+    /**
+     * @param plansByRenderedSql successfully-captured {@code EXPLAIN FORMAT=JSON} plans, keyed by rendered SQL.
+     * @param failedSql          rendered SELECTs whose EXPLAIN threw — captured but un-vetted, surfaced so they can't
+     *                           pass the gate by omission.
+     */
+    public record CapturedQueries(Map<String, String> plansByRenderedSql, Set<String> failedSql) {
```
andrescrz, line 49: Minor: Please add lombok builder (with toBuilder) and use it over constructors.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/CapturingSqlLogger.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    @Override
+    public void logAfterExecution(StatementContext context) {
```
andrescrz, line 53: Minor: check if you need to handle `context` null value. You could exit in that case to avoid NPE.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/CapturingSqlLogger.java
```diff
@@ -0,0 +1,126 @@
...
+        var renderedSql = context.getRenderedSql();
+        if (renderedSql == null || !isReadQuery(renderedSql)) {
```
andrescrz: Minor: it seems the null check should be moved to isReadQuery method.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/CapturingSqlLogger.java
```diff
@@ -0,0 +1,126 @@
...
+                    }
+                    failedSql.add(renderedSql);
+                    return null;
+                }
+            }
+        } catch (Exception e) {
+            // A query the gate cannot EXPLAIN (e.g. one referencing a session TEMPORARY TABLE that only exists
+            // mid-flight) is recorded as failed rather than dropped, so the gate can report it instead of passing it
+            // by omission.
+            log.debug("EXPLAIN failed for query [{}]: {}", jdbcSql, e.getMessage());
```
andrescrz: Minor: you might want to increase the log level to warning here. 

In addition, pass the whole exception as last argument to the logger (without `{}`) instead of printing the message.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/CapturingSqlLogger.java
```diff
@@ -0,0 +1,126 @@
...
+    // v2 puts the base table name in table_name (v1 uses the alias) — required for the full-scan check to match.
+    private void forceV2ExplainFormat(StatementContext context) throws Exception {
+        try (Statement statement = context.getConnection().createStatement()) {
+            statement.execute("SET explain_json_format_version = 2");
+        }
+    }
+
+    // Read paths the gate vets: plain SELECTs and CTE reads (WITH ... SELECT). Matching only "select" would skip the
+    // CTE class entirely — exactly the multi-reference-CTE shape (OPIK-7198) the gate exists to catch. No WITH-prefixed
+    // @SqlUpdate writes exist in the backend, and MySQL EXPLAIN handles WITH ... SELECT the same as a plain SELECT.
```
andrescrz: Nit: all these comments better as javadocs.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/MySqlPlanShapeAsserter.java
```diff
@@ -0,0 +1,78 @@
...
+    private static final String ACCESS_TYPE = "access_type";
+    private static final String TABLE_NAME = "table_name";
+    private static final String ACCESS_MATERIALIZE = "materialize";
+    private static final String ACCESS_TABLE_SCAN = "table";
+
+    private final Set<String> fullScanSensitiveTables;
+
+    public MySqlPlanShapeAsserter(@NonNull Set<String> fullScanSensitiveTables) {
+        this.fullScanSensitiveTables = fullScanSensitiveTables;
+    }
```
andrescrz: Please use lombok for all boiler-plate code, including constructors etc.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/MySqlPlanShapeAsserter.java
```diff
@@ -0,0 +1,78 @@
...
+
+    public List<PlanShapeViolation> findViolations(@NonNull String renderedSql, @NonNull String explainJson) {
+        var violations = new ArrayList<PlanShapeViolation>();
+        walk(JsonUtils.getJsonNodeFromString(explainJson), renderedSql, violations);
+        return violations;
+    }
+
+    private void walk(JsonNode node, String renderedSql, List<PlanShapeViolation> violations) {
```
andrescrz, line 44: Minor: recursion is great, likely explain documents shouldn't large. But you can explore an iterative version if the implementation is not too complicated, so this never consumes or runs out of resources (memory, stack over flow etc.)

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/PlanShapeBaseline.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Builder(toBuilder = true)
+    public record Entry(@NonNull String fingerprint, String note) {
+    }
+
+    private final Set<String> allowedFingerprints;
+
+    private PlanShapeBaseline(Set<String> allowedFingerprints) {
+        this.allowedFingerprints = allowedFingerprints;
+    }
```
andrescrz: Same about lombok for boiler-plate.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/PlanShapeBaseline.java
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * The checked-in allowlist of plan-shape violations that already exist in the codebase. The gate blocks only on
+ * <b>net-new</b> violations vs. this baseline (see {@link #netNew(List)}); the allowlist is ratcheted down as legacy
+ * offenders are fixed. Each entry is a {@link PlanShapeViolation#fingerprint()} plus a human note explaining why it is
+ * tolerated / which ticket tracks its removal.
+ */
+public class PlanShapeBaseline {
+
+    @Builder(toBuilder = true)
+    public record Entry(@NonNull String fingerprint, String note) {
```
andrescrz: Minor: for test code, feel free to omit NonNull validations, not strictly necessary to keep our interfaces as well guarded (and documented as in prod code).

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/PlanShapeBaseline.java
```diff
@@ -0,0 +1,56 @@
...
+    private PlanShapeBaseline(Set<String> allowedFingerprints) {
+        this.allowedFingerprints = allowedFingerprints;
+    }
+
+    public static PlanShapeBaseline loadFromClasspath(@NonNull String resourcePath) {
+        try (InputStream is = PlanShapeBaseline.class.getClassLoader().getResourceAsStream(resourcePath)) {
+            if (is == null) {
+                throw new IllegalStateException("Plan-shape baseline resource not found: " + resourcePath);
+            }
+            List<Entry> entries = JsonUtils.getMapper().readValue(is, new TypeReference<>() {
```
andrescrz: Minor: the TypeReference can be moved to a static final field, to avoid re-instantiating per call. They're safe thread and shareable.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/PlanShapeBaseline.java
```diff
@@ -0,0 +1,56 @@
...
+            });
+            return new PlanShapeBaseline(entries.stream().map(Entry::fingerprint).collect(Collectors.toSet()));
+        } catch (UncheckedIOException e) {
+            throw e;
+        } catch (Exception e) {
+            throw new UncheckedIOException("Failed to load plan-shape baseline: " + resourcePath,
+                    new java.io.IOException(e));
```
andrescrz: Nit: use imports.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/PlanShapeBaseline.java
```diff
@@ -0,0 +1,56 @@
...
+    private final Set<String> allowedFingerprints;
+
+    private PlanShapeBaseline(Set<String> allowedFingerprints) {
+        this.allowedFingerprints = allowedFingerprints;
+    }
+
+    public static PlanShapeBaseline loadFromClasspath(@NonNull String resourcePath) {
+        try (InputStream is = PlanShapeBaseline.class.getClassLoader().getResourceAsStream(resourcePath)) {
+            if (is == null) {
+                throw new IllegalStateException("Plan-shape baseline resource not found: " + resourcePath);
```
andrescrz: Nit: accross the whole PR, feel free to use `formatted` instead of concating strings `+` for a more modern Java syntax. You can surround the printed values with single quotes, as we usually do.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/MySqlQueryPlanGateTest.java
```diff
@@ -0,0 +1,152 @@
...
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@DisplayName("🐬 MySQL Query Plan Gate")
+@ExtendWith(DropwizardAppExtensionProvider.class)
+class MySqlQueryPlanGateTest {
+
+    private static final String BASELINE_RESOURCE = "planshape/mysql-baseline.json";
+
+    // MySQL tables that grow with tenant data — a full scan here is a latency cliff at scale. The high-volume analytics
+    // tables (traces, spans, dataset_items) live in ClickHouse and are out of scope for the MySQL gate.
+    private static final Set<String> FULL_SCAN_SENSITIVE_TABLES = Set.of(
```
andrescrz, line 58: Double check the list of tables here, experiments is a ClickHouse table, not MySQL.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/MySqlQueryPlanGateTest.java
```diff
@@ -0,0 +1,152 @@
...
+            "/v1/private/projects",
+            "/v1/private/feedback-definitions",
+            "/v1/private/automations/evaluators");
+
+    private static final String API_KEY = UUID.randomUUID().toString();
+    private static final String USER = UUID.randomUUID().toString();
+    private static final String WORKSPACE_ID = UUID.randomUUID().toString();
+    private static final String TEST_WORKSPACE = UUID.randomUUID().toString();
+
+    private final TestContainersSetup setup = new TestContainersSetup();
```
andrescrz, line 81: Review this, not sure why this setup is needed.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/MySqlQueryPlanGateTest.java
```diff
@@ -0,0 +1,152 @@
...
+
+            assertThat(response.getStatus())
+                    .as("read endpoint %s must return 200 so its query is actually rendered", path)
+                    .isEqualTo(200);
+        }
+    }
+
+    private static String renderFailure(List<PlanShapeViolation> netNew) {
+        return netNew.stream()
+                .map(v -> "%n  [%s] %s%n    fingerprint: %s%n    sql: %s".formatted(
```
andrescrz: Nit: use more significant names for local vars (even in streams), instead of `v` etc.

andrescrz (commented):
Left some feedback to address before moving forward.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/CapturingSqlLogger.java
JetoPistola: Fixed in ab146be532 — `CapturedQueries` now carries `@Builder(toBuilder = true)`. (`PlanShapeViolation` already had it.)

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/CapturingSqlLogger.java
JetoPistola: Fixed in ab146be532 — added a null guard for `context` at the top of `logAfterExecution`, and folded the `renderedSql` null check into `isReadQuery` (see next comment).

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/CapturingSqlLogger.java
JetoPistola: Fixed in ab146be532 — `isReadQuery` now returns `false` for null, so the caller just checks `!isReadQuery(renderedSql)`.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/CapturingSqlLogger.java
JetoPistola: Fixed in ab146be532 — bumped to `log.warn` and pass the exception object as the last arg (`log.warn("EXPLAIN failed for query '{}'", jdbcSql, exception)`) so the full stacktrace is logged. A failed EXPLAIN is an un-vetted query, so warn is the right level.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/CapturingSqlLogger.java
JetoPistola: Fixed in ab146be532 — converted the block/line comments on the private methods to javadoc.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/MySqlPlanShapeAsserter.java
JetoPistola: Fixed in ab146be532 — `@RequiredArgsConstructor` replaces the hand-written constructor.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/PlanShapeBaseline.java
JetoPistola: Fixed in ab146be532 — `@RequiredArgsConstructor` on `PlanShapeBaseline`, and hoisted the `TypeReference` to a static final field (see the other comment).

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/MySqlPlanShapeAsserter.java
JetoPistola: Skipping — kept the recursive walk. EXPLAIN JSON is shallow (a handful of nesting levels for query_block/inputs), so stack depth isn't a real risk here; an explicit-stack rewrite would add complexity without buying safety. Documented the bounded-depth rationale in the class javadoc.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/PlanShapeBaseline.java
JetoPistola: Fixed in ab146be532 — dropped the `@NonNull` annotations across these test-scope types (`PlanShapeViolation`, `PlanShapeBaseline`, `MySqlPlanShapeAsserter`).

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/PlanShapeBaseline.java
JetoPistola: Fixed in ab146be532 — hoisted to `private static final TypeReference<List<Entry>> ENTRIES_TYPE`.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/PlanShapeBaseline.java
JetoPistola: Fixed in ab146be532 — imported `java.io.IOException` instead of the fully-qualified name.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/PlanShapeBaseline.java
JetoPistola: Fixed in ab146be532 — switched to `String.formatted(...)` across the PR (error messages, fingerprint, detail strings) with single-quoted values.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/MySqlQueryPlanGateTest.java
JetoPistola: Fixed in ab146be532 — good catch, removed `experiments` from `FULL_SCAN_SENSITIVE_TABLES`. Verified against the liquibase changelogs: `experiments` lives only in ClickHouse (db-app-analytics); the remaining five (datasets, prompts, prompt_versions, feedback_definitions, projects) are all MySQL. Updated the comment to call out experiments as CH.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/MySqlQueryPlanGateTest.java
JetoPistola: This is the shared container + app bootstrap (`TestContainersSetup` starts Redis/MySQL/ClickHouse/Zookeeper, runs migrations, and builds the Dropwizard app) — the same two-line `setup` + `@RegisterApp APP = setup.APP` pattern used by the resource ITs (e.g. `DatasetsResourceTest`). The gate needs a booted app with a real MySQL to install the SqlLogger and run EXPLAIN, so this is the leanest option; the alternative is duplicating the container wiring inline. Left as-is — happy to change if you'd prefer a different shared fixture.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/MySqlQueryPlanGateTest.java
JetoPistola: Fixed in ab146be532 — renamed the stream var `v` -> `violation` in `renderFailure`.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/PlanShapeBaseline.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    private final Set<String> allowedFingerprints;
+
+    public static PlanShapeBaseline loadFromClasspath(String resourcePath) {
+        try (InputStream is = PlanShapeBaseline.class.getClassLoader().getResourceAsStream(resourcePath)) {
+            if (is == null) {
+                throw new IllegalStateException("Plan-shape baseline resource not found: '%s'".formatted(resourcePath));
+            }
+            List<Entry> entries = JsonUtils.getMapper().readValue(is, ENTRIES_TYPE);
```
andrescrz: Generally, we shouldn't access the inner mapper (`getMapper`) on `JsonUtils`, exposing that was a mistake. 

Instead, use the counter part `readValue` method. If not existing, add it following the conventions of that class.

This relates to your `UncheckedIOException` catch below, we generally encapsulate those, so you won't need it after this change.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/PlanShapeBaseline.java
```diff
@@ -0,0 +1,56 @@
...
+            return new PlanShapeBaseline(entries.stream().map(Entry::fingerprint).collect(Collectors.toSet()));
+        } catch (UncheckedIOException e) {
+            throw e;
+        } catch (Exception e) {
+            throw new UncheckedIOException(new IOException(
+                    "Failed to load plan-shape baseline: '%s'".formatted(resourcePath), e));
+        }
```
andrescrz: Nit: you can probably avoid this wide exception catch and just throw the `UncheckedIOException` above instead of `IllegalStateException`.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/MySqlQueryPlanGateTest.java
```diff
@@ -0,0 +1,153 @@
...
+    private static final Set<String> FULL_SCAN_SENSITIVE_TABLES = Set.of(
+            "datasets", "prompts", "prompt_versions", "feedback_definitions", "projects");
+
+    // MySQL-backed list endpoints. Each renders the SELECT of a DAO the gate must vet.
+    private static final List<String> READ_PATHS = List.of(
+            "/v1/private/datasets",
+            "/v1/private/prompts",
+            "/v1/private/projects",
+            "/v1/private/feedback-definitions",
+            "/v1/private/automations/evaluators");
```
andrescrz, line 68: The counter part table or tables in the set above for `automations/evaluators`  seems to be missing.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/MySqlQueryPlanGateTest.java
andrescrz: In this case, I recommend adding a javadoc on top of this field to concisely document this.

andrescrz (dismissed):
A few more comments, but nothing blocking.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/PlanShapeBaseline.java
JetoPistola: Fixed in 0874563927 — switched to the `JsonUtils.readValue(InputStream, TypeReference)` counterpart, which already wraps in `UncheckedIOException`, so the inner-mapper access is gone.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/utils/planshape/PlanShapeBaseline.java
JetoPistola: Fixed in 0874563927 — the wide catch is gone. With `JsonUtils.readValue` handling the wrapping, the only remaining catch is `IOException` from the try-with-resources `close()`, rethrown as `UncheckedIOException`; the missing-resource `IllegalStateException` guard stays.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/MySqlQueryPlanGateTest.java
JetoPistola: Fixed in 0874563927 — good catch. The `/automations/evaluators` read path is backed by three tenant-growing tables, all now in the set: `automation_rules`, `automation_rule_evaluators`, `automation_rule_projects`. Gate still passes with them enforced, so they introduce no net-new violations.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/MySqlQueryPlanGateTest.java
JetoPistola: Fixed in 0874563927 — added a javadoc on the field documenting what the fixture bootstraps (Redis/MySQL/ClickHouse/Zookeeper + migrations + the Dropwizard app), why the gate needs a booted app on a real MySQL (installing the `CapturingSqlLogger` on the app's `Jdbi` and running `EXPLAIN` on the live connection), and why it reuses the resource-IT fixture rather than duplicating the container wiring.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/MySqlQueryPlanGateTest.java
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    @Test
+    @DisplayName("🐬 no MySQL read path introduces a materialized/temporary-table plan or a tenant-table full scan")
+    void mySqlReadPathsHaveNoPlanShapeRegressions() {
+        READ_PATHS.forEach(this::getPage);
+
+        var captured = capturingSqlLogger.captured();
+
+        assertThat(captured.plansByRenderedSql())
```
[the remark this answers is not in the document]
JetoPistola: Fixed in 0874563927 — the underlying concern is real: the global check only proved *some* route hit MySQL, so a route that stopped rendering a SELECT would have gone dark while the other four kept the gate green.

Implemented as a per-path **growth** assertion rather than the suggested clear-and-assert-per-path. `CapturingSqlLogger` deduplicates by rendered SQL, so clearing between paths would re-capture queries shared across endpoints, and a per-path non-empty check on a cleared capture would fail on correct code whenever two routes render the same SQL. Instead each `getPage(path)` now asserts the captured set *grew*, which gives the same independent-per-route guarantee without the false-failure mode. The old global `isNotEmpty` is subsumed by it and was dropped.

Verified it actually bites rather than passing vacuously: replaying an already-exercised path fails with `read endpoint /v1/private/datasets must contribute at least one new MySQL SELECT to the gate`. Full gate passes on the real five paths.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/MySqlQueryPlanGateTest.java
```diff
@@ -0,0 +1,174 @@
...
+     * correct code whenever two routes render the same SQL.
+     */
+    private void getPageAndAssertItWasVetted(String path) {
+        int capturedBefore = capturingSqlLogger.captured().plansByRenderedSql().size();
+
+        getPage(path);
+
+        assertThat(capturingSqlLogger.captured().plansByRenderedSql().size())
+                .as("read endpoint %s must contribute at least one new MySQL SELECT to the gate", path)
+                .isGreaterThan(capturedBefore);
```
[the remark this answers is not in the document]
JetoPistola: Fixed in 3a3f5cbd43 — you're right, and this was a real latent flaw in the growth check I added in 0874563927. A route whose `SELECT` an earlier route already captured adds no entry to the deduped map, so it would fail the assertion despite having run and been vetted — order-dependent, and the failure message would have blamed the wrong thing.

Took the second of your two suggestions: assert the path *triggered a capture* rather than a new fingerprint. `CapturingSqlLogger` now counts read queries in a `readsObserved` counter incremented **before** the dedup check, so it moves for every read query executed even when the SQL was already seen. The per-path assertion reads that counter instead of the map size, which makes it order-independent. The final net-new plan-shape check is untouched.

Worth being explicit about the semantics: the per-path assertion now proves the route *executed* a MySQL read, not that this particular route's SQL was plan-checked. For shared SQL the plan check legitimately happened once, on whichever route reached it first — which is the correct behaviour, just a slightly weaker per-path claim than the previous wording implied. Javadoc updated to say so.

Verified both directions:
- replaying an already-exercised path (the exact case that false-failed before) now **passes**
- pointing a path at a non-MySQL endpoint (`/is-alive/ping`) still **fails** with `read endpoint /is-alive/ping must execute at least one MySQL read for the gate to vet`

So the coverage guarantee from your earlier finding survives without the false-failure mode.

andrescrz (approved):
LGTM

PR #7585, opened 2026-07-23
[OPIK-7462] [FE] fix: build Recent Activity links via router params to survive basepath strip

AndreiCautisanu:
## Details

On cloud (basepath `/opik`), the Project Home "Recent activity" links broke for workspaces whose name starts with `opik` (e.g. `opik-testing`, `opik-demo`) — clicking an item navigated to `/opik/-testing/projects/...` instead of `/opik/opik-testing/projects/...`. The links were built by string-concatenating the workspace name into `to`, and TanStack Router strips the basepath with an unbounded `pathname.replace(basepath, "")`, so `/opik-testing/...` lost its `/opik` prefix. This passes the workspace/project as route params (`$workspaceName`/`$projectId`) instead, so substitution runs after the basepath strip — the idiom already used everywhere else in v2.

- Only manifested on cloud (`/opik`) with an `opik*` workspace; OSS/self-hosted/local (basepath `/`) were never affected and stay correct.
- The Experiment link's `experiments` query param is now passed via `search` and percent-encodes to the same value the compare page's `JsonParam` reader expects.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves OPIK-7462

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: Root-cause investigation, the fix, and the regression tests were authored with Claude Code; reviewed by the author.
- Human verification: Diff and test assertions reviewed by the author; failing-then-passing test confirmed manually against the pre-fix code.

## Testing

Frontend unit tests (`apps/opik-frontend`), run with `npx vitest run`:

- `src/v2/pages/ProjectHomePage/RecentActivitySection.test.tsx` — new. Renders the section inside a real TanStack router at each basepath and asserts the produced `href` for every activity type.
  - **cloud (basepath `/opik`)**: `opik-testing` → `/opik/opik-testing/projects/...` and explicitly NOT the mangled `/opik/-testing/...` (this case fails on the current `main` code); `andreicc` unchanged.
  - **OSS/self-hosted (basepath `/`)**: `default` → `/default/projects/...`; a workspace literally named `opik` → `/opik/projects/...`.
- Verified the test reproduces the bug: reverting to the string-concatenation version makes it fail with `/opik/-testing/projects/019e3a68-.../...` (byte-for-byte the reported URL); the fix makes all 4 cases pass.
- `npx vitest run` on the neighboring router-heavy suites (`V1CompatRedirect`, `TracesTabRedirect`) — still green (28 tests).
- `npx tsc --noEmit` (only a pre-existing, unrelated `@codemirror/legacy-modes` type-decl error), `npx eslint` and `npx prettier --check` on the changed files — clean.

Environment: local, frontend unit tests only. Not run: full E2E suite (unit coverage exercises the exact link-formation logic at both basepaths).

## Documentation

No documentation changes — internal bug fix with no user-facing API or behavior change beyond the corrected link.

### DIFF
--- a/apps/opik-frontend/src/v2/pages/ProjectHomePage/RecentActivitySection.test.tsx
+++ b/apps/opik-frontend/src/v2/pages/ProjectHomePage/RecentActivitySection.test.tsx
@@ -1,0 +1,200 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { render, screen } from "@testing-library/react";
+import {
+  createMemoryHistory,
+  createRootRoute,
+  createRoute,
+  createRouter,
+  Outlet,
+  RouterProvider,
+} from "@tanstack/react-router";
+import { ActivityType, RecentActivityItem } from "@/types/recent-activity";
+import RecentActivitySection from "./RecentActivitySection";
+
+let mockItems: RecentActivityItem[] = [];
+
+vi.mock("@/api/projects/useRecentActivity", () => ({
+  default: () => ({ data: { content: mockItems }, isPending: false }),
+}));
+
+const PROJECT_ID = "019e3a68-9e79-70a8-adb8-0e04294f9d73";
+
+// Minimal route tree mirroring the real project-scoped leaves the activity
+// links target. `basepath` is "/opik" on cloud and "/" on OSS/self-hosted.
+async function renderAt(workspaceName: string, basepath = "/opik") {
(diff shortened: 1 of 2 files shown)


awkoy (approved):

aadereiko (approved):
LGTM!

PR #7575, opened 2026-07-22
[OPIK-7438] [BE] fix: gate sampling params for Anthropic adaptive-thinking models on the playground/inference path

JetoPistola:
## Details

image

image

image

Follow-up to PR #7531 (online-rules judge path). That PR gated `temperature` server-side for Anthropic adaptive-thinking models (`claude-sonnet-5`, `claude-opus-4-7/4-8`) on the **judge** path, but deliberately left the sibling **playground / online-inference / proxy** path out of scope.

`LlmProviderAnthropicMapper.toCreateMessageRequest` still forwarded `temperature` / `top_p` unconditionally, so those models `400` with `temperature is deprecated for this model` for **API-created** requests — which bypass the FE sanitizer added in OPIK-6244.

Changes (backend-only, `infrastructure/llm/antropic`):
- Convert the `temperature` / `top_p` MapStruct mappings to `@Named` resolvers that gate on `AnthropicModelName.supportsSamplingParams(request.model())` **and** whenever extended thinking is enabled per-request via `custom_parameters.thinking.type` (any non-blank type other than `disabled`), mirroring the judge-path logic in `AnthropicClientGenerator`.
- `top_p` is now also dropped when gated (previously it could leak once `temperature` was suppressed).
- `top_k` is **not** sent on this path (`ChatCompletionRequest` has no `top_k`), so there is nothing to gate — no dead code added.

## Change checklist
- [x] User facing
- [x] Documentation update

## Issues

- Resolves [OPIK-7438](https://comet-ml.atlassian.net/browse/OPIK-7438)
- Follow-up to #7531 (GitHub issue #7526)

## Testing

- New nested `SamplingParamsGating` cases in `AnthropicMappersTest` (11): adaptive models drop temperature+top_p; sampling-capable models forward them; thinking-enabled (`enabled`/`adaptive`/future) drops them; `disabled`/blank/absent thinking forwards them.
- Existing `AnthropicMappersTest` (7) unaffected — Podam random models default to sampling-supported.
- `mvn clean test-compile` clean on Corretto 25; `AnthropicMappersTest` (18) + `AnthropicClientGeneratorTest` (32) all green; `spotless:apply` clean.

## Documentation

No user-facing docs change required.

[...]

PR #7590, opened 2026-07-23
[issue-7526][BE] fix: make Anthropic adaptive-thinking opt-out sync-proof

### LINKED ISSUE #7526: [Bug] Online rules (LLM-as-judge) fail with Anthropic adaptive-thinking models (Sonnet 5, Opus 4.7/4.8): temperature 400 + empty response, custom_parameters not forwarded
## Summary

Online Evaluation Rules (LLM-as-judge) fail with Anthropic **adaptive-thinking** models (Claude Sonnet 5, Opus 4.7/4.8). Two chained problems:

1. **`temperature` is sent even though the model rejects it** → `400 invalid_request_error: "temperature is deprecated for this model."`
2. **After removing `temperature`, the judge returns an empty response** (`textChars=0, finishReason=LENGTH`) — adaptive thinking consumes the whole token budget before emitting text, and the rule's `custom_parameters` (`max_tokens`, `thinking:{type:"disabled"}`) are **not forwarded** to the provider call on the online-rules path.

Reported by a customer on comet.com cloud. Original Slack thread: https://cometml.slack.com/archives/C02HPRHR2F9/p1784348104052679

## Environment

- Opik on comet.com (cloud)
- Provider: Anthropic
- Model: `claude-sonnet-5` (also reproduces on Opus 4.7 / 4.8)
- Subsystem: **Online Evaluation Rules** (LLM-as-judge) — not Experiments/Playground

## Repro

1. Configure the Anthropic provider.
2. Create an online rule (`llm_as_judge`) with model `claude-sonnet-5` (via UI or API).
3. Ingest a trace so the rule fires.

**Observed:** the `temperature` 400 above; or, with `temperature` removed, `finishReason=LENGTH` with 0 output tokens (no scores produced).

Customer-provided stack trace (temperature case):

```
jakarta.ws.rs.ClientErrorException: {"type":"error","error":{"type":"invalid_request_error","message":"`temperature` is deprecated for this model."},"request_id":"req_011Cd8..."}
  at com.comet.opik.domain.llm.ChatCompletionService.failHandlingLLMProviderError(ChatCompletionService.java:128)
  at com.comet.opik.domain.llm.ChatCompletionService.scoreTrace(ChatCompletionService.java:117)
  at com.comet.opik.api.resources.v1.events.OnlineScoringLlmAsJudgeScorer.lambda$scoreTraceReactive$0(OnlineScoringLlmAsJudgeScorer.java:517)
```

## Root cause

Online rules score via `ChatCompletionService.scoreTrace()` → `LlmProviderFactory.getLanguageModel(...)` → the Anthropic **judge-path** builder `AnthropicClientGenerator.newChatLanguageModel(...)`
(`apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/antropic/AnthropicClientGenerator.java`).

That builder reads only **two** fields off the rule's model params:

```java
.modelName(modelParameters.name())
...
Optional.ofNullable(modelParameters.temperature()).ifPresent(builder::temperature);
```

- **Problem 1:** `temperature` is forwarded unconditionally when non-null, with **no per-model capability check**. Adaptive-thinking models reject it → 400.
- **Problem 2:** `LlmAsJudgeModelParameters` carries a `customParameters` JsonNode (and `seed`), but this builder never reads them, and never sets `maxTokens`. So `thinking:{type:"disabled"}` / `max_tokens` set on the rule are silently dropped, and the default token budget is consumed by thinking → empty response.

Across all judge-path generators, only `CustomLlmClientGenerator` and `FreeModelLlmProvider` forward `customParameters`; the Anthropic (and OpenAI) judge-path generators do not.

Contrast: the **Playground** path uses a different mapper (`LlmProviderAnthropicMapper#resolveMaxTokens`) that defaults `max_tokens` to 4096 — which is why this reproduces on online rules but not (necessarily) in Playground.

## Relationship to OPIK-6244

[OPIK-6244](https://comet-ml.atlassian.net/browse/OPIK-6244) ("Remove temperature/top_p/top_k from requests to Opus 4.7 via playground and online scoring", closed) addressed this class of issue **frontend-side** (`sanitizeConfigForRequest` strips temperature client-side) with a per-model capability map targeted at Opus 4.7. Two holes remain, and this report hits both:

- **API-created rules bypass the FE entirely** — the backend never strips `temperature` by capability, so an API-created rule 400s.
- **The capability map is per-model** — each new adaptive-thinking model (Sonnet 5, Opus 4.8) must be added.

OPIK-6244 also never addressed the thinking / `max_tokens` empty-response problem, which is the more fundamental of the two.

## Suggested fix (backend)

In `AnthropicClientGenerator.newChatLanguageModel`:

1. Gate `temperature` on a per-model capability check **server-side** (don't rely on the FE).
2. Forward `modelParameters.customParameters()` (thinking, `max_tokens`) into the builder, the way `CustomLlmClientGenerator` does.
3. Set a sane default `maxTokens` on the judge path, like the Playground mapper already does.

## Workaround for users today

`custom_parameters` won't help (ignored on this path). Use a **non-adaptive-thinking judge model** — e.g. `claude-3-5-sonnet`, `claude-3-5-haiku`, or an OpenAI/Gemini judge. Omitting `temperature` via API fixes problem 1 but not problem 2, so it isn't sufficient on its own.

## Expected behavior

Online rules should support Anthropic adaptive-thinking models as judges: don't send `temperature` to models that reject it, and forward `thinking:{type:"disabled"}` and/or a large enough `max_tokens` from the rule's parameters.

[OPIK-6244]: https://comet-ml.atlassian.net/browse/OPIK-6244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Nimrod007:
## Details

Follow-up to #7531. That PR opted the Anthropic adaptive-thinking models (`claude-opus-4-7`, `claude-opus-4-8`, `claude-sonnet-5`) out of sampling params using a **per-constant flag** in the `AnthropicModelName` enum:

```java
CLAUDE_OPUS_4_8("claude-opus-4-8", false),
```

The enum constant list is **regenerated by the recurring `chore: sync provider model definitions` job** (#7088, #7315, #7582). That generator emits every model as a single-arg `NAME("value")` constant and has no notion of the extra capability argument. **#7582 ran the sync ~18h after #7531 merged and silently dropped all three `, false` flags** (reverting them to the single-arg constructor, which defaults `supportsSamplingParams` to `true`) — reopening #7526:

```diff
-    CLAUDE_OPUS_4_7("claude-opus-4-7", false),
-    CLAUDE_OPUS_4_8("claude-opus-4-8", false),
     ...
-    CLAUDE_SONNET_5("claude-sonnet-5", false);
+    CLAUDE_OPUS_4_7("claude-opus-4-7"),
+    CLAUDE_OPUS_4_8("claude-opus-4-8"),
+    CLAUDE_SONNET_5("claude-sonnet-5");
```

Current `main` still has the flags dropped, so online LLM-as-judge rules on these models 400 again with `` `temperature` is deprecated for this model `` whenever a temperature is present (API-created rules, which bypass the OPIK_6244 FE sanitizer — the exact case #7526/#7531 targeted).

## Change

Move the opt-out **out of the generated constant list** into a hand-maintained `ADAPTIVE_THINKING_MODELS` set that `supportsSamplingParams()` consults. The constant list stays in the exact single-arg shape the generator emits, so future syncs can add/remove/reorder constants freely without ever clobbering the capability. Public API (`supportsSamplingParams(String)`) and behavior are unchanged; the field-backed `Map` + helper constructor are removed.

- `AnthropicModelName` — drop `supportsSamplingParams` enum field, the `this(value, true)` helper ctor, and the field-derived map; add `ADAPTIVE_THINKING_MODELS` set + doc explaining why it must live outside the generated list.
- `AnthropicClientGenerator` unchanged — the max_tokens / thinking / temperature-gating logic from #7531 was never regressed (it lives outside the generated file).

## Change checklist

- [x] User facing
- [ ] Documentation update

## Issues

- Re-resolves #7526 (regressed by #7582 after #7531).

## Testing

- Existing `AnthropicClientGeneratorTest.SamplingParamsCapability` cases (assert `false` for the three adaptive models) pass again — they were effectively dead on `main` since `test-compile` is currently broken there, which is why the #7582 regression went undetected.
- New `adaptiveThinkingModelsRemainEnumeratedAndRejectSamplingParams` regression guard: asserts each adaptive model id still maps to a real enum constant (so the set can't drift when a sync renames/removes a constant) **and** reports no sampling-param support.
- **Verified live on staging** (`opik-backend` @ the regressed build): an API-created online-judge rule on `claude-opus-4-8` **with** `temperature` 400s on `temperature is deprecated`; the same rule **without** temperature scores cleanly (confirming #7531's `max_tokens` defaulting is intact and the temperature opt-out is the only regressed piece).
- Note: full `mvn test-compile` requires JDK 25 and is broken on pristine `main` (unrelated JDK-25/Lombok mismatch), so the suite was not run locally; the change is mechanical.

## Documentation

No user-facing docs change required. Two follow-ups (out of scope): (1) the broken `test-compile` on `main` means these guards don't run in CI yet — worth fixing so this class of regression is caught; (2) the `sync provider model definitions` generator could be taught to preserve/emit capability flags, removing the need for a parallel set.

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: full change (implementation + test), authored under human direction and review
- Human verification: root cause reproduced end-to-end on staging; diff reviewed


apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/antropic/AnthropicModelName.java
```diff
@@ -28,22 +26,34 @@ public enum AnthropicModelName implements StructuredOutputSupported {
...
+     * dropped on the next sync — which is exactly how #7582 reverted the adaptive-thinking opt-out
+     * and reopened issue #7526. Keeping the capability here, outside the generated constant list,
+     * lets syncs add/remove/reorder constants freely without ever clobbering it.
+     *
+     * <p>When Anthropic ships a new adaptive-thinking model, add its API id to this set.
+     */
+    private static final Set<String> ADAPTIVE_THINKING_MODELS = Set.of(
+            "claude-opus-4-7",
+            "claude-opus-4-8",
+            "claude-sonnet-5");
```
andrescrz: Please DRY the string constants. Use the enum constants above if possible through `value`. If compilation error, just extract these fields to static final fields and reuse them both above and here. 

Let's not allow space to drift.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/antropic/AnthropicClientGeneratorTest.java
```diff
@@ -49,6 +52,25 @@ void supportsSamplingParamsReturnsTrueForNonAdaptiveModels(String modelName) {
...
+         * Regression guard for #7526 / #7531 → #7582. The adaptive-thinking opt-out must NOT be
+         * encoded as a per-constant argument in the enum constant list, because that list is
+         * regenerated by the "sync provider model definitions" chore, which emits single-arg
+         * {@code NAME("value")} constants and silently drops any extra argument (that is how #7582
+         * reopened this bug). This asserts each adaptive model id (a) still maps to a real enum
+         * constant, so the standalone opt-out set can't drift when a sync renames/removes a
+         * constant, and (b) reports no sampling-param support.
+         */
+        @ParameterizedTest
+        @ValueSource(strings = {"claude-sonnet-5", "claude-opus-4-7", "claude-opus-4-8"})
```
andrescrz: Likewise, pick up the right enums here (they work with ValueSource, and you can explicitly include) so we test the drift.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/antropic/AnthropicClientGeneratorTest.java
```diff
@@ -49,6 +52,25 @@ void supportsSamplingParamsReturnsTrueForNonAdaptiveModels(String modelName) {
...
+        void adaptiveThinkingModelsRemainEnumeratedAndRejectSamplingParams(String modelName) {
+            var enumeratedValues = Arrays.stream(AnthropicModelName.values())
+                    .map(AnthropicModelName::toString)
```
andrescrz: This should never be toString, which relies on the enum constant. You should use value getter (and add it with lombok).

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/antropic/AnthropicClientGeneratorTest.java
```diff
@@ -49,6 +52,25 @@ void supportsSamplingParamsReturnsTrueForNonAdaptiveModels(String modelName) {
...
+                    .collect(Collectors.toSet());
+            assertThat(enumeratedValues).contains(modelName);
```
andrescrz: This should be a single check against the first found as there shouldn't be collisions.

andrescrz (commented):
Let's polish this a bit, to prevent this issue from happening again.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/antropic/AnthropicClientGeneratorTest.java
```diff
@@ -49,6 +52,25 @@ void supportsSamplingParamsReturnsTrueForNonAdaptiveModels(String modelName) {
...
+            assertThat(AnthropicModelName.supportsSamplingParams(modelName)).isFalse();
+        }
```
JetoPistola: ❓ **question** | Testing

This guard catches a sync that **renames or removes** an adaptive constant — assertion (a) fails once `claude-opus-4-8` leaves `values()`. But it can't catch the other sync failure mode: a future sync **adds a new adaptive-thinking model** (say `claude-opus-4-9`) as a plain `NAME("value")` constant. Nothing forces that new id into `ADAPTIVE_THINKING_MODELS`, so `supportsSamplingParams("claude-opus-4-9")` silently returns `true` and we `400` again — the same class of bug #7526 describes, just for the *next* model rather than a renamed one.

That residual gap is inherent to any hand-maintained set (the PR body's follow-up # 2 — teaching the generator to emit the capability — is the real fix), so it may well be acceptable as-is. But the doc comment currently reads as if it fully closes the sync-drift hole. Consider softening it to something like "guards against rename/removal of *known* adaptive models" so the remaining manual step (adding new adaptive models to the set) stays visible to whoever runs the next sync.

JetoPistola:
👋 **Review summary**

**What looks good**
- The core move — from a per-constant enum flag to a standalone `ADAPTIVE_THINKING_MODELS` set — is exactly the right shape for the problem. It decouples the capability from the generator-owned constant list, which is precisely what #7582 clobbered.
- The doc comment on the set is excellent: it captures *why* the set must live outside the generated list, with the #7088/#7315/#7582 history, so the next person touching the sync chore won't re-inline the flag.
- Public contract (`supportsSamplingParams(String)`) and behaviour are unchanged, and the old field/helper-ctor/derived-map are cleanly removed — no dead code left behind.
- Good instinct adding an explicit regression guard rather than trusting the existing capability cases (which, per the PR body, were dead on `main`).

**Overall**
Small, well-motivated, mechanical fix that re-resolves #7526 the durable way. It also underpins the sibling playground/inference-path fix in #7575, which consumes the same `supportsSamplingParams` contract — so the two land as a coherent pair. `andrescrz` has already flagged the substantive polish items (DRY the set against the enum values, use enum-backed `ValueSource`, `value` getter instead of `toString()`, single-membership check) — all worth doing, no need to repeat them here.

**Inline comments**: 1 question — on a residual coverage gap in the new regression guard (it protects against rename/removal of known adaptive models, but not against a sync *adding* a new adaptive model). Non-blocking.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/antropic/AnthropicModelName.java
Nimrod007: Fixed in 0842640df1 — `ADAPTIVE_THINKING_MODELS` now references the enum constants' `value` (`CLAUDE_OPUS_4_7.value`, etc.) instead of duplicating the string literals. This also gives compile-time drift protection: a sync that renames/removes a referenced constant now breaks compilation rather than drifting silently.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/antropic/AnthropicClientGeneratorTest.java
Nimrod007: Fixed in 0842640df1 — the regression guard now uses `@EnumSource(value = AnthropicModelName.class, names = {...}, mode = INCLUDE)`, so renaming/removing one of these constants fails the test at discovery time. That tests the drift directly.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/antropic/AnthropicClientGeneratorTest.java
Nimrod007: Fixed in 0842640df1 — added Lombok `@Getter` to the enum and the test now uses `model.getValue()` instead of `toString()`.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/antropic/AnthropicClientGeneratorTest.java
Nimrod007: Fixed in 0842640df1 — with `@EnumSource` the parameter is the enum constant itself, so the collected-set `contains` check is gone entirely; the test just asserts `supportsSamplingParams(model.getValue())` is false.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/antropic/AnthropicClientGeneratorTest.java
Nimrod007: Fixed in 0842640df1 — softened both doc comments. The set's Javadoc now has a "Residual manual step" note stating it guards rename/removal of *known* adaptive models but cannot force a newly-synced adaptive model into the set (that stays a manual step; the durable fix is teaching the generator to emit the capability). The test's Javadoc says the same. Thanks for catching the overstated wording.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/antropic/AnthropicModelName.java
```diff
@@ -28,22 +28,39 @@ public enum AnthropicModelName implements StructuredOutputSupported {
...
+     * a referenced constant breaks compilation instead of drifting silently.
+     *
+     * <p><strong>Residual manual step.</strong> This guards against rename/removal of the
+     * <em>known</em> adaptive models below; it cannot force a <em>newly synced</em> adaptive model
+     * into the set. When Anthropic ships a new adaptive-thinking model, its API id must be added
+     * here by hand (the durable fix is teaching the generator to emit the capability directly).
+     */
+    private static final Set<String> ADAPTIVE_THINKING_MODELS = Set.of(
+            CLAUDE_OPUS_4_7.value,
+            CLAUDE_OPUS_4_8.value,
```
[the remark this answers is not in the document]
Nimrod007: Fixed in 5d0bedc097 — renamed `ADAPTIVE_THINKING_MODELS` → `ADAPTIVE_THINKING_MODEL_IDS` (and its usage in `supportsSamplingParams(String)`) to make clear it holds serialized API id strings used as lookup keys, not model objects.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/antropic/AnthropicModelName.java
```diff
@@ -28,22 +28,39 @@ public enum AnthropicModelName implements StructuredOutputSupported {
...
+            CLAUDE_SONNET_5.value);
 
     /**
      * Whether the model accepts sampling params (temperature/top_p/top_k). Adaptive-thinking models
      * reject them with a 400 and are opted out explicitly. Unknown model names default to {@code true}
      * so registry-only Anthropic models (not enumerated here) keep temperature support.
      */
     public static boolean supportsSamplingParams(String modelName) {
-        return SAMPLING_PARAMS_SUPPORT.getOrDefault(modelName, true);
+        return !ADAPTIVE_THINKING_MODEL_IDS.contains(modelName);
```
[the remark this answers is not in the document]
Nimrod007: Fixed in 8d207d7874 — guarded `null`/blank via `StringUtils.isBlank(modelName)` before the `contains` lookup, so the method honors its documented "unknown model names default to `true`" contract instead of throwing. Note this was a latent issue, not a regression: the previous `Map.of(...).getOrDefault(modelName, true)` also threw on `null` (immutable collections reject null lookups), and the sole caller passes a `@NotNull`-validated `name`. Added a regression test (`supportsSamplingParamsDefaultsTrueForNullModel`).

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/llm/antropic/AnthropicClientGeneratorTest.java
```diff
@@ -49,6 +50,28 @@ void supportsSamplingParamsReturnsTrueForNonAdaptiveModels(String modelName) {
         void supportsSamplingParamsDefaultsTrueForUnknownModels(String modelName) {
             assertThat(AnthropicModelName.supportsSamplingParams(modelName)).isTrue();
         }
+
+        @Test
+        void supportsSamplingParamsDefaultsTrueForNullModel() {
+            assertThat(AnthropicModelName.supportsSamplingParams(null)).isTrue();
+        }
```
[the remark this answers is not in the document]
Nimrod007: Fixed in bab46fc9e8 — parameterized the fallback test with `@NullAndEmptySource` plus a whitespace `@ValueSource` case, so the `StringUtils.isBlank` branch is now exercised for null, empty, and whitespace model names (renamed to `supportsSamplingParamsDefaultsTrueForNullOrBlankModel`).

JetoPistola:
👋 **Re-review summary** (revision `bab46fc`)

The revision cleanly addresses every prior thread:

- **DRY** — `ADAPTIVE_THINKING_MODEL_IDS` now references `CLAUDE_OPUS_4_7.value` / `CLAUDE_OPUS_4_8.value` / `CLAUDE_SONNET_5.value` instead of duplicated literals, so a sync that renames/removes a referenced constant now fails compilation rather than drifting.
- **Test drift** — driven by `@EnumSource(mode = INCLUDE, names = {…})`, so renaming/removing an adaptive constant fails at test-discovery time; the set-scan was dropped.
- **`value` getter** — `@Getter` added; the test asserts via `model.getValue()`, not `toString()`.
- **Residual-gap doc** — both the production and test doc comments now spell out the one remaining manual step: a *newly synced* adaptive model still has to be added to the set by hand (the durable fix being generator-emitted capability). That was my only open question — resolved.

Bonus: null/blank handling is now explicit (`StringUtils.isBlank(...)`) and covered by a dedicated `@NullAndEmptySource` + blank case, with identical semantics to the old `getOrDefault(…, true)`.

No net-new findings; CI is fully green. Nicely done. ⚠️ This is not a formal approval — a human reviewer should still approve.

JetoPistola (approved):
Addresses all review feedback (DRY via enum `.value`, `@EnumSource` drift guard, `getValue()`, residual-gap docs) plus explicit null/blank handling. Sound, sync-proof fix. CI green.

🤖 *Approved via /review-github-pr follow-up*

PR #7597, opened 2026-07-23
[OPIK-7476] [FE] feat: A/B test guided mobile onboarding flow

aswynz:
## Details

Implements the A/B test for the new mobile onboarding flow, wiring the PostHog experiment [`guided-mobile-onboarding-flow`](https://us.posthog.com/project/222582/experiments/387337) (id 387337, variants `control`/`test`) into the v2 onboarding. The new guided mobile flow shows healthy intent (~29% send setup instructions) but dropped activation rate after the Run button was removed, so this splits the phone onboarding behind the flag to measure the trade-off.

- On phones only: `test` (Group B) renders the guided `MobileOnboarding` flow (no Run button); `control` (Group A) falls through to the legacy integrations onboarding (with the Run button).
- Forces the `CodeExecutor` (Run button) on for the control arm regardless of the separate `run-button-activation-test` rollout, so Group A reliably matches &#34;old flow with Run button&#34;.
- `NewQuickstart` restores the `#manual` step hash for the control arm (parity with the pre-mobile flow).
- An unresolved variant (`undefined`) resolves to `control` at both call sites. `useFeatureFlagVariantKey` returns `undefined` *permanently* when PostHog is never initialized — self-hosted/OSS never calls `posthog.init()` (no key/host), and cloud users can have it ad-blocked — so treating `undefined` as a transient loading state would strand those phone users. This matches how every other PostHog flag in the app handles an unresolved variant. Note the consequence: on deployments without PostHog, phone users now get the legacy integrations onboarding rather than `MobileOnboarding`.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- OPIK-7476

## AI-WATERMARK

AI-WATERMARK: yes

- If yes:
  - Tools: Claude Code
  - Model(s): Claude Opus 4.8
  - Scope: Full implementation (flag wiring, render split, control-arm Run button gate) and unit tests; author reviewed the diff and verified behavior.
  - Human verification: Reviewed the full diff; ran the unit tests, typecheck, and lint (all green); brought up the stack locally via `dev-runner` and confirmed both variants render correctly on a phone viewport.

## Testing

- `npx vitest run src/v2/pages/GetStartedPage/NewQuickstart.test.tsx` — 18/18 pass (14 existing + 4 new: `test` arm renders the guided flow, `control` arm renders the legacy flow, an unresolved flag falls back to control, and desktop is unaffected by the flag).
- `npx tsc --noEmit` — clean.
- `npx eslint` on the changed files — clean.
- End-to-end (local `dev-runner`, v2, iPhone viewport, backend + frontend built from this branch): forcing the `guided-mobile-onboarding-flow` variant showed `test` → guided `MobileOnboarding` wizard (no Run button) and `control` → legacy integrations onboarding (`#manual`), confirming the branch.
- Not run: the Run-button sub-dialog for the control arm was not exercised visually (requires `CODE_EXECUTOR_SERVICE_URL`, unset in local dev); the branch that enables it is covered by the unit tests. The unresolved-flag fallback is covered by unit tests only, not verified against a real PostHog-less deployment.

## Documentation

No documentation changes required — internal onboarding A/B experiment behind a PostHog flag, no user-facing docs or public API changes.


apps/opik-frontend/src/v2/pages/GetStartedPage/NewQuickstart.tsx
```diff
@@ -66,6 +69,21 @@ const AgentOnboardingQuickstart: React.FC = () => {
...
+  // the guided MobileOnboarding flow ("test" / Group B, no Run button) and the
+  // legacy onboarding flow with the Run button ("control" / Group A, which falls
+  // through to the standard desktop logic below). `undefined` means the flag has
+  // not resolved yet — we hold rendering until it does so users are not flashed
+  // one flow then re-bucketed into the other.
+  const mobileOnboardingVariant = useFeatureFlagVariantKey(
+    GUIDED_MOBILE_ONBOARDING_FLOW_FEATURE_FLAG_KEY,
+  );
+  const isMobileOnboardingFlagResolving =
+    isPhone && mobileOnboardingVariant === undefined;
```
JetoPistola: 🚫 **blocker** | Logic

`useFeatureFlagVariantKey` returns `undefined` **permanently** when PostHog is never initialized — not just transiently while the flag resolves. That happens for:

- **Self-hosted / OSS Opik** — PostHog is only `init()`-ed by the closed-source `comet` plugin (`src/plugins/comet/init.tsx` → `initPosthog`), and even there `initPosthog` returns early when the key/host env vars are absent.
- **Ad-blocked cloud users** — `posthog.init()` runs but flags never load.

In those cases `isMobileOnboardingFlagResolving` stays `true` forever, so phone users are stuck on `<Loader />` (line 162) and can never complete onboarding. The pre-existing code unconditionally rendered `MobileOnboarding` for phones, so this is a regression for those deployments.

Every other `useFeatureFlagVariantKey` call in this same file and in `DemoProjectBanner` / `ConnectAgentStep` guards `undefined` with `?? DEFAULT_ONBOARDING_FLOW` for exactly this reason — this A/B code is the only one that blocks on `undefined` instead of falling back.

Consider a bounded fallback: treat an unresolved flag as the control arm after a short timeout (or fall back to `MobileOnboarding` when PostHog is known-unavailable) so phone onboarding can't hang indefinitely.

apps/opik-frontend/src/v2/pages/GetStartedPage/NewQuickstart.test.tsx
```diff
@@ -373,3 +401,67 @@ describe("NewQuickstart — non-manual variants", () => {
...
+
+    // Control falls through to the same manual flow desktop users get (which
+    // carries the Run button), not the guided mobile flow.
+    expect(
+      screen.getByTestId("onboarding-integrations-page"),
+    ).toBeInTheDocument();
+    expect(screen.queryByTestId("mobile-onboarding")).not.toBeInTheDocument();
+  });
+
+  it("holds rendering behind a loader while the experiment bucket is unresolved", () => {
```
JetoPistola: 🧹 **nit** | Testing

This test asserts the loader shows for `mockMobileVariant = undefined`, but that state is indistinguishable from "PostHog permanently unavailable" (self-hosted / ad-blocked). As written it codifies the infinite-loader behavior flagged on `NewQuickstart.tsx` as intended. If you add a bounded fallback for the unresolved case, it'd be worth a companion test that asserts phone users still reach a real onboarding flow when the flag never resolves.

JetoPistola:
👋 **Review summary**

**What looks good**
- Clean flag wiring — extracting `GUIDED_MOBILE_ONBOARDING_FLOW_FEATURE_FLAG_KEY` / `GUIDED_MOBILE_ONBOARDING_VARIANTS` as shared constants and reusing them in both `NewQuickstart` and `IntegrationDetailsDialog` keeps the two arms in sync.
- Thoughtful UX detail: holding render until the bucket resolves to avoid flashing one flow then re-bucketing, and restoring the `#manual` hash for the control arm to keep parity with the pre-mobile flow.
- Comments explain the *why* (Group A = legacy + Run button) rather than restating code — genuinely helpful for the next reader.
- Solid test coverage: test / control / loading / desktop-unaffected are all exercised.

**Overall**
The approach is sound and the scope is tight. My one real concern is that the loader gate keys off `useFeatureFlagVariantKey === undefined`, which is a *permanent* state when PostHog isn't initialized (self-hosted/OSS Opik, ad-blocked users) — that would hang phone onboarding on an infinite loader. A bounded fallback (treat unresolved as control after a timeout) would close that gap. Everything else is clean.

**Inline comments**: 1 blocker, 1 nit.

apps/opik-frontend/src/v2/pages/GetStartedPage/NewQuickstart.test.tsx
jverre: Skipping — no separate companion test. We went with an immediate `?? CONTROL` fallback rather than a bounded timeout, so there's no unresolved state left to cover: the existing test for `mockMobileVariant = undefined` is converted in place to assert phone users land on the control onboarding flow (no loader, no guided flow). Adding a second test would duplicate it.

apps/opik-frontend/src/v2/pages/GetStartedPage/NewQuickstart.tsx
jverre: Fixed in deee0d812db70ff52b46ae7e3663ba99ce64a8f4 — good catch, the permanent-vs-transient distinction is real: `initPosthog` returns early without a key/host, so `posthog.init()` never runs on self-hosted/OSS and the flag stays `undefined` forever.

Dropped the loader gate entirely and resolved `undefined` to the control arm with `?? GUIDED_MOBILE_ONBOARDING_VARIANTS.CONTROL`, rather than a bounded timeout — that matches how the rest of the app already handles an unresolved variant, and the flash the loader was guarding against only affects the one render before flags resolve on cloud, where control is the fall-through path anyway.

Applied the same fallback in `IntegrationDetailsDialog.tsx` as well. Without it the two files would have disagreed about what "control" means, and a phone user whose flag never resolves would have landed on the legacy flow *without* the Run button — a third arm that is neither experiment group.

JetoPistola:
👋 **Re-review summary** (head `deee0d8`)

Thanks for the quick turnaround on `fix(onboarding): fall back to control when mobile experiment flag is unresolved` — re-reviewed the updated diff and both earlier findings are resolved:

- ✅ **Infinite-loader blocker**: both call sites now guard `useFeatureFlagVariantKey(...)` with `?? GUIDED_MOBILE_ONBOARDING_VARIANTS.CONTROL`, and the `Loader` gate is gone. Phones with an unresolved flag (self-hosted/OSS, ad-blocked) now fall through to the control onboarding flow instead of hanging. This also matches the `?? DEFAULT` pattern used by the other flags in the file.
- ✅ **Test nit**: the loader-while-unresolved test is replaced by `"falls back to the control flow when the flag never resolves"`, which asserts a real onboarding flow renders.

No new issues on the updated diff. Nothing blocking from my side.

JetoPistola (approved):
Both prior review findings (infinite-loader blocker + test nit) are resolved by `deee0d8` — unresolved flag now falls back to control at both call sites, no loader hang. Re-reviewed the updated diff; no new issues. Approving.

PR #7554, opened 2026-07-21
[NA] [BE] fix: register fireworks_ai as a canonical provider so Fireworks model prices load

Anuj7411:
## Details

Fireworks AI's pricing entries are silently dropped by Opik today. `model_prices_and_context_window.json` carries roughly 260 non-zero-cost entries tagged `litellm_provider: "fireworks_ai"` (the `accounts/fireworks/models/*` catalog spanning the llama, qwen, deepseek, glm, kimi and minimax families), and none of them load because `fireworks_ai` is not registered in `PROVIDERS_MAPPING`; `buildModelPrice` returns `null` for any entry whose provider is not in that map, so every Fireworks span ends up with no cost. This follows the exact pattern of the earlier provider-registration fixes for azure (#7262), xai (#7433), deepseek (#7434) and perplexity (#7432).

- Register `fireworks_ai -> fireworks_ai` in `PROVIDERS_MAPPING` so the entries load
- Route `fireworks_ai` through `textGenerationWithCacheCostOpenAI` in `PROVIDERS_CACHE_COST_CALCULATOR`; Fireworks' LiteLLM cost calculator (`litellm/llms/fireworks_ai/cost_calculator.py`) delegates to `generic_cost_per_token`, so its usage payload uses the OpenAI shape (cached tokens under `prompt_tokens_details.cached_tokens`), and the eight cache-priced Fireworks models today (glm-4p7, glm-5p1, kimi-k2p5, minimax-m2p1) get the same cache-read discount OpenAI/Azure/xAI/DeepSeek receive
- `PROVIDERS_CACHE_COST_CALCULATOR` was at the ten-pair ceiling of `Map.of`, so it moves to `Map.ofEntries` to take the eleventh entry

## Change checklist

- [x] User facing
- [ ] Documentation update

## Issues

No ticket; discovered by scanning `litellm_provider` values in the bundled price map for providers with priced models that are absent from `PROVIDERS_MAPPING`.

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: investigation of the drop path, the two-line mapping change, and the parameterized test
- Human verification: model resolution and both cost figures checked against the bundled `model_prices_and_context_window.json` before submission

## Testing

Added a parameterized test (`calculateCostHandlesFireworksModels`) covering both branches:

[...]

PR #7434, opened 2026-07-11
[BE] fix: register deepseek as a canonical provider so DeepSeek model prices load

Anuj7411:
## Details

Following the pattern of #7262 (which registered `azure`), #7432 (perplexity) and #7433 (xai) — `deepseek` is not currently in `PROVIDERS_MAPPING`, so all 12 entries in `model_prices_and_context_window.json` tagged with `litellm_provider: "deepseek"` silently drop at load time. Every one publishes non-zero input+output rates: `deepseek-chat`, `deepseek-reasoner`, `deepseek-v4-flash`, `deepseek-v4-pro`, plus the `deepseek/deepseek-coder`, `deepseek/deepseek-r1`, `deepseek/deepseek-v3`, `deepseek/deepseek-v3.2` variants.

9 of the 12 also publish `cache_read_input_token_cost`. DeepSeek's cost calculator in LiteLLM (`litellm/llms/deepseek/cost_calculator.py`) is a one-liner that delegates directly to `generic_cost_per_token` — the same helper OpenAI/xAI use — so DeepSeek's usage payload follows the same OpenAI shape (`prompt_tokens_details.cached_tokens` flattened under `original_usage.*`). Route it through the existing `SpanCostCalculator::textGenerationWithCacheCostOpenAI` calculator; no new cache-calc method needed, just a mirror of the `openai` / `azure` / `xai` entries in `PROVIDERS_CACHE_COST_CALCULATOR`.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code (CLI)
- Model(s): Claude Opus 4.7
- Scope: scouted the reachable DeepSeek model count (12, 9 with cache_read), inspected LiteLLM's DeepSeek cost calc to confirm it delegates to generic_cost_per_token, drafted the two-line plumbing change following the #7262 / #7433 pattern, and wrote the parameterized branch-coverage test
- Human verification: ran the test suite locally; cross-checked expected dollar values against rates in model_prices_and_context_window.json

## Testing

Local verification on JDK 25 + Maven 3.9.9. Result: **102/102 green** (2 new parameterized cases under `calculateCostHandlesDeepseekModels`, style matching #7262).

Scenarios covered:

[...]

PR #7433, opened 2026-07-11
[BE] fix: register xai as a canonical provider so xAI/Grok model prices load

Anuj7411:
## Details

Following the pattern of #7262 (which registered `azure`) and #7432 (perplexity) — `xai` is not currently in `PROVIDERS_MAPPING`, so all 40 entries in `model_prices_and_context_window.json` tagged with `litellm_provider: "xai"` silently drop at load time. Every one publishes non-zero input+output rates: the full `grok-2` / `grok-3` / `grok-4` / `grok-4-fast` / `grok-code` families.

29 of the 40 also publish `cache_read_input_token_cost`. xAI's cost calculator in LiteLLM (`litellm/llms/xai/cost_calculator.py`) delegates to `generic_cost_per_token`, which is the same helper OpenAI uses — so xai's usage payload follows OpenAI's shape (`prompt_tokens_details.cached_tokens` flattened into `original_usage.*`). That means xai routes cleanly through the existing `SpanCostCalculator::textGenerationWithCacheCostOpenAI` calculator; no new cache-calc method needed, just a mirror of the `azure` entry in `PROVIDERS_CACHE_COST_CALCULATOR`.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code (CLI)
- Model(s): Claude Opus 4.7
- Scope: assisted in scouting the reachable xai model count (40 non-zero-cost, 29 with cache_read), inspected `litellm/llms/xai/cost_calculator.py` to confirm xai delegates to `generic_cost_per_token` and therefore shares the OpenAI usage shape, drafted the two-line plumbing change following the #7262 pattern, and wrote the parameterized branch-coverage test
- Human verification: ran the test suite locally; verified xai models route through the correct calculator per the cache-vs-plain gate; cross-checked expected dollar values against the rates in `model_prices_and_context_window.json`

## Testing

Local verification on JDK 25 + Maven 3.9.9:

```
mvn -ntp -f apps/opik-backend/pom.xml \
  -Dtest=CostServiceTest,SpanCostCalculatorTest test
```

Result: **102/102 green** (2 new parameterized cases under `calculateCostHandlesXaiModels`, matching the style established in #7262 for Azure).

Scenarios covered:

[...]

PR #7432, opened 2026-07-11
[BE] fix: register perplexity as a canonical provider so Perplexity model prices load

Anuj7411:
## Details

Following the pattern of #7262 (which registered `azure`) — `perplexity` is not currently in `PROVIDERS_MAPPING`, so every one of the 41 entries in `model_prices_and_context_window.json` tagged with `litellm_provider: "perplexity"` silently drops at load time. Of those 41, 16 publish non-zero input+output rates: the `sonar` family (`sonar`, `sonar-pro`, `sonar-reasoning`, `sonar-reasoning-pro`, `sonar-deep-research`) plus the legacy `perplexity/llama-*` and `perplexity/codellama-*` chat models.

No Perplexity model publishes `cache_read_input_token_cost` or `cache_creation_input_token_cost`, so requests route cleanly through `SpanCostCalculator::textGenerationCost` — no cache-calculator wiring needed.

Registering `perplexity -> perplexity` is a one-line change; the parameterized test uses `perplexity/sonar` (input=output=`1e-6`) and `perplexity/sonar-pro` (input `3e-6` / output `1.5e-5`) so the plumbing is exercised on both a uniform-rate model and one where the two rates differ.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code (CLI)
- Model(s): Claude Opus 4.7
- Scope: assisted in the LiteLLM JSON scout that confirmed Perplexity has 16 non-zero-cost models and zero cache-rate entries, drafted the one-line PROVIDERS_MAPPING addition following the #7262 pattern, and wrote the parameterized branch-coverage test
- Human verification: ran the test suite locally; cross-checked the expected dollar values against the rates in `model_prices_and_context_window.json`; verified that no existing test regressed (baseline 100 -> 102 tests after the two new parameterized cases)

## Testing

Local verification on JDK 25 + Maven 3.9.9:

```
mvn -ntp -f apps/opik-backend/pom.xml \
  -Dtest=CostServiceTest,SpanCostCalculatorTest test
```

Result: **102/102 green** (2 new parameterized cases under `calculateCostHandlesPerplexityModels`, matching the style established in #7262 for Azure).

Scenarios covered:

[...]

PR #7604, opened 2026-07-25
[NA] [BE] fix: register moonshot as a canonical provider so Moonshot model prices load

Anuj7411:
## Details

Moonshot AI's pricing entries are silently dropped by Opik today. `model_prices_and_context_window.json` carries 22 non-zero-cost entries tagged `litellm_provider: "moonshot"` (the `moonshot-v1-*` legacy models and the newer `kimi-*` family), and none of them load because `moonshot` is not in `PROVIDERS_MAPPING`; `buildModelPrice` returns `null` for every one of them and Moonshot spans never get a cost. This follows the exact pattern of #7262 (azure), #7433 (xai), #7434 (deepseek), #7432 (perplexity) and #7554 (fireworks_ai).

- Register `moonshot -> moonshot` in `PROVIDERS_MAPPING` so the entries load
- Route `moonshot` through `textGenerationWithCacheCostOpenAI` in `PROVIDERS_CACHE_COST_CALCULATOR`; Moonshot's API is OpenAI-compatible and its LiteLLM cost calculator delegates to `generic_cost_per_token`, so cached tokens are flattened under `prompt_tokens_details.cached_tokens`, matching the OpenAI / Azure / xAI / DeepSeek / Fireworks routing
- Twelve of the 22 Moonshot models publish `cache_read_input_token_cost` (all `kimi-k2` variants), so this fix meaningfully discounts the cache-hit path for the kimi family instead of overcharging every cached prompt token at the full input rate

## Change checklist

- [x] User facing
- [ ] Documentation update

## Issues

No ticket; discovered by scanning `litellm_provider` values in the bundled price map for providers with priced models that are absent from `PROVIDERS_MAPPING`.

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.7
- Scope: the two-line mapping change and the parameterized test
- Human verification: model resolution and both cost figures checked against the bundled `model_prices_and_context_window.json` before submission

## Testing

Added a parameterized test (`calculateCostHandlesMoonshotModels`) covering both branches:

- `moonshot/moonshot-v1-8k` (input 2e-7, output 2e-6, no cache) routes through `textGenerationCost`; 1000 prompt + 200 completion tokens = 0.0006
- `moonshot/kimi-k2-0711-preview` (input 6e-7, output 2.5e-6, cache_read 1.5e-7) routes through `textGenerationWithCacheCostOpenAI`; with 300 of 1000 prompt tokens cached, 700 * 6e-7 + 200 * 2.5e-6 + 300 * 1.5e-7 = 0.000965

Both cases fail against the current code (the models do not resolve without the mapping entry, so cost comes back as zero) and pass with the fix.

## Documentation

No documentation changes needed; provider pricing is driven entirely by the bundled price map and the provider registration in `CostService`.


Anuj7411:
Thanks Baz, good catch. Confirmed by grepping OpenRouterModelName.java — Opik does enumerate the moonshotai/* namespace (MOONSHOTAI_KIMI_K2, MOONSHOTAI_KIMI_K2_THINKING, etc., ten-plus variants), so an aggregator-routed Kimi call goes through findModelPrice's provider-prefix fallback, hits PROVIDERS_MAPPING.get("moonshotai") returning null, and resolves to DEFAULT_COST.

Addressed in 1d7c31e03 by adding moonshotai -> moonshot alongside the existing moonshot -> moonshot canonical entry, mirroring the microsoft -> azure override pattern already in the map. No JSON entries carry litellm_provider="moonshotai" (I checked the bundled price map), so this is pure aggregator routing help and changes no existing lookup semantics.

Extended the test suite with a parameterized calculateCostFindsMoonshotViaAggregatorProviderPrefix covering provider="openrouter" and provider="custom-llm" on both a cache-priced (kimi-k2-0711-preview) and non-cache (moonshot-v1-8k) model. Both cases return 0.0 without the alias and the correct rate with it

andrescrz (approved):
Hi @Anuj7411 

Best

PR #7605, opened 2026-07-25
[NA] [BE] feat(cipx): ingest Anthropic org seat/billing identity fields for AI-Spend

aadereiko:
## Details

Ingests `organization_type`, `seat_tier`, and `billing_type` from cipx trace metadata into `cipx_trace_identities` so AI-Spend can price a subscription seat exactly (Standard vs Premium, monthly vs annual, team vs individual) instead of inferring it from the rate-limit tier.

- Two additive `LowCardinality(String)` columns (migrations `000111` / `000112`, both default `''`).
- Threads the fields through `CipxTraceIdentityDAO`'s metadata parse + bulk INSERT.
- Extends `CostIntelligenceIngestionTest` to assert the three fields land end-to-end; converts the `CipxIdentityRow` test DTO to `@Builder`.
- Migration numbering: `main` is at `000110`; this adds `000111` (organization_type) + `000112` (seat_tier/billing_type). #7592 was renumbered to `000113`, so the merged sequence is `110 → 111 → 112 → 113` — no collision.
- Part of a 3-repo change (merge order): cipx writes the fields (cost-intelligence-proxy-internal#55) → **opik-backend ingests (this PR)** → cost-api prices them (ai-cost-backend#45). Independently mergeable — empty columns fall back to plan inference.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: migrations, DAO field threading, and the ingestion test authored with assistance
- Human verification: diff reviewed; `CostIntelligenceIngestionTest` run locally against a real ClickHouse container

## Testing

`mvn test -Dtest=CostIntelligenceIngestionTest` — 9 tests, 0 failures — through a real ClickHouse (Testcontainers), including new assertions that `organization_type` / `seat_tier` / `billing_type` parse from cipx metadata and land in `cipx_trace_identities`. Full module compiles clean.

- Scenarios: the three fields ingest end-to-end; existing identity ingestion is unaffected (defaults to `''`).
- Environment: local Maven + Testcontainers ClickHouse.
- Not run: the two red CI checks (Backend Integration Group 5, Python SDK compat) are unrelated to this change — Group 5 is a job timeout and its failing tests (`LocalRunnerReaper`, `OptimizationProjectMigration`, `WorkspaceMetadataServiceCache`) don't touch cipx; both re-run.

## Documentation

No user-facing docs — internal schema + ingestion change. The AI-Spend seat-pricing behaviour these fields feed is documented in the cost-api PR (ai-cost-backend#45).


apps/opik-backend/src/main/java/com/comet/opik/domain/CipxTraceIdentityDAO.java
```diff
@@ -180,6 +189,9 @@ private Publisher<? extends Result> insert(List<TraceIdentityRow> rows, String w
                     .bind(index++, row.billingMode())
                     .bind(index++, row.plan())
                     .bind(index++, row.planUsageStatus())
+                    .bind(index++, row.organizationType())
```
JetoPistola, line 192: 🧹 **nit** | Documentation

These 3 new binds bring each row tuple to **25** parameters, but the explanatory comment a few lines up ("then 22 parameters per row tuple in template order") still says 22. The bind code is correct and stays consistent with the INSERT column list and placeholder tuple — only the count in that comment drifted. Worth bumping `22` → `25` so the comment doesn't mislead the next reader.

JetoPistola:
👋 **Review summary**

**What looks good**
- **The high-risk invariant holds.** Adding columns to a positional bulk INSERT is exactly where these changes go wrong, and here all three orderings stay perfectly aligned — the INSERT column list, the StringTemplate placeholder tuple, and the `bind(int)` sequence all insert `organization_type` / `seat_tier` / `billing_type` in the same relative slot (right after `plan_usage_status`). The record field order and the read-path builder match too.
- **Additive and safe by construction.** Both columns are `LowCardinality(String) DEFAULT ''`, matching the sibling `plan` / `billing_mode` / `plan_usage_status` columns from `000104`, and the ReplacingMergeTree sort key is untouched — so this doesn't perturb the dedup contract. Empty columns fall back to inference, which keeps the PR independently mergeable ahead of the cipx and cost-api sides.
- **Migration numbering is clean.** `main` is at `000110`; `000111` + `000112` collide with nothing, and the renumber note vs #7592 (`→ 000113`) is accurate.
- **Genuinely good migration comments** — they explain the *why* (why the same rate-limit tier maps to different seat prices) rather than restating the DDL, and the test asserts all three fields end-to-end through a real ClickHouse container.

**Overall**
Tight, well-scoped change that does one thing and does it correctly. The approach — persist the authoritative seat fields the proxy already fetches instead of inferring them downstream — is the right call, and the 3-repo merge story is well documented.

**Inline comments**: 1 nit (a stale `22`→`25` parameter-count comment; the code is correct). Nothing blocking.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000112_add_seat_tier_billing_type_to_cipx_trace_identities.sql
```diff
@@ -0,0 +1,15 @@
...
+--changeset aadereiko:000112_add_seat_tier_billing_type_to_cipx_trace_identities
+--comment: Add seat_tier + billing_type to cipx_trace_identities (authoritative seat class + billing cadence)
+
+-- The OAuth profile's organization object carries seat_tier ("standard"/"priority") and
+-- billing_type ("stripe_subscription" monthly / "stripe_subscription_contracted" | "manual" annual /
+-- "usage_based" API). The proxy already fetches this response but only mapped a subset. Persisting these
+-- lets AI-Spend price a seat exactly — seat_tier names the Standard vs Premium class directly (instead of
+-- inferring it from the rate-limit tier), and billing_type selects the monthly vs annual seat rate.
+-- Additive columns with a default; existing rows read the default (empty).
+ALTER TABLE ${ANALYTICS_DB_DATABASE_NAME}.cipx_trace_identities ON CLUSTER '{cluster}'
```
BorisTkachenko: Let's have one migration file, not two. Just put all 3 new columns in the same file.

BorisTkachenko (commented):
All looks good, just one comment to polish before merging.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000112_add_seat_tier_billing_type_to_cipx_trace_identities.sql
aadereiko: Done in 90c683f — merged into a single migration (000111) that adds all three columns (`organization_type`, `seat_tier`, `billing_type`) in one ALTER; 000112 removed.

BorisTkachenko (approved):
LGTM

PR #7609, opened 2026-07-25
[NA] [DOCS] fix: point changelog canonical URLs at the changelog index

jverre:
## Details

The `2026-07-13` and `2026-07-20` changelog entries set `canonical-url` to a per-date path (`/docs/opik/changelog/<date>`), but both of those URLs 404. Fern renders changelog entries into the changelog section declared by `- changelog: ../docs/changelog` with `slug: changelog`, so individual entries do not get standalone per-date pages. The other **59 of 61** entries all point at `/docs/opik/changelog`; this makes these two match.

This is the cause of the `Docs - Preview link` broken-link failure that has been firing on every docs-touching PR — the same two links appear in the link-check comment on #7599, #7602, and #7608.

## Change checklist
- [ ] User facing
- [x] Documentation update

## Issues

- Resolves #

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: Traced the broken links reported by the docs link checker to the deviating `canonical-url` frontmatter and corrected it.
- Human verification: Author identified the canonical URLs as the likely cause and directed the fix.

## Testing

Verified the URLs directly with `curl -o /dev/null -w "%{http_code}" -L`:

| URL | Before | After |
|---|---|---|
| `/docs/opik/changelog/2026-07-13` (old canonical) | **404** | no longer referenced |
| `/docs/opik/changelog/2026-07-20` (old canonical) | **404** | no longer referenced |
| `/docs/opik/changelog` (new canonical) | 200 | **200** |

Confirmed consistency and that the same bug does not exist elsewhere:
- `grep -h "canonical-url" docs/changelog/*.mdx | sort | uniq -c` → **61 entries, all one value** (`https://www.comet.com/docs/opik/changelog`); was 59 + the two deviations
- Repo-wide `grep -rn 'canonical-url:.*changelog/20'` across `apps/opik-documentation` → no remaining matches
- The only other changelog canonical in the repo, `/docs/opik/development/optimization-runs/changelog`, is a real page and returns **200** — unaffected

`uvx pre-commit run --files` on both changed paths: no hooks matched (frontmatter-only change to `.mdx`), so nothing to report.

The authoritative check is `Docs - Preview link` on this PR, which crawls the built preview; it should now report no broken links.

## Documentation

This change *is* the documentation fix. No further docs updates needed.

Out of scope, but worth flagging separately: `docs-v2/changelog/` (the `latest` docs version) stops at **2026-07-06**, while `docs/changelog/` (v1) has entries through **2026-07-20**. So the current docs are missing the 2026-07-13 and 2026-07-20 changelog entries entirely. That is a content gap rather than a link bug, and writing those two entries is an editorial call — this PR only stops the canonical URLs from 404ing.

### DIFF
--- a/apps/opik-documentation/documentation/fern/docs/changelog/2026-07-13.mdx
+++ b/apps/opik-documentation/documentation/fern/docs/changelog/2026-07-13.mdx
@@ -1,5 +1,5 @@
 ---
-canonical-url: https://www.comet.com/docs/opik/changelog/2026-07-13
+canonical-url: https://www.comet.com/docs/opik/changelog
 ---
 
 ## Per-Evaluation Spend Budget for LLM-as-Judge Evaluators
--- a/apps/opik-documentation/documentation/fern/docs/changelog/2026-07-20.mdx
+++ b/apps/opik-documentation/documentation/fern/docs/changelog/2026-07-20.mdx
@@ -1,5 +1,5 @@
 ---
-canonical-url: https://www.comet.com/docs/opik/changelog/2026-07-20
+canonical-url: https://www.comet.com/docs/opik/changelog
 ---
 
 ## Redesigned Optimization Run Overview


apps/opik-documentation/documentation/fern/docs/changelog/2026-07-13.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-canonical-url: https://www.comet.com/docs/opik/changelog/2026-07-13
+canonical-url: https://www.comet.com/docs/opik/changelog
```
JetoPistola, line 2: ❓ **question** / 💡 **suggestion** | Docs / SEO / v1-vs-latest

This fixes the 404, but I think it points at the wrong target — and it surfaces a deeper content issue.

**The 404 was because the old value was malformed** (hyphenated date, no version segment). The per-date page does exist — I confirmed on prod:

| URL | Status |
|---|---|
| `/docs/opik/v1/changelog/2026/7/13` | **200** ✅ |
| `/docs/opik/changelog/2026-07-13` (old canonical) | **404** ❌ |
| `/docs/opik/changelog/2026/7/13` (unversioned) | **404** ❌ |
| `/docs/opik/changelog` (this PR's target) | **200** ✅ |

**But note the `/v1/` — that's the real issue.** This entry lives in `docs/changelog/`, which Fern wires into the **v1** version only (`versions/v1.yml:52` → `changelog: ../docs/changelog`). The **latest** version reads from `docs-v2/changelog/` (`versions/latest.yml:48`), which **stops at `2026-07-06`**. So `2026-07-13` exists *only* in v1 — hence the unversioned URL 404s and the real page is v1-namespaced.

Since these entries look new and v2 is the actively-developed docs, the underlying question is: **should these two entries be ported into `docs-v2/changelog/` (latest)?** If yes, that's the real fix — then the canonical could point at the latest page. If they're intentionally v1-only, then a v1 per-date canonical is right:

```suggestion
canonical-url: https://www.comet.com/docs/opik/v1/changelog/2026/7/13
```

Either way, canonicalizing to the shared **index** marks this distinct entry as a duplicate of the aggregate list, which suppresses the specific page from search. The 59 other entries do point at the index, so if that's a deliberate convention, feel free to resolve — but the v1/v2 gap is worth a follow-up regardless.

apps/opik-documentation/documentation/fern/docs/changelog/2026-07-20.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-canonical-url: https://www.comet.com/docs/opik/changelog/2026-07-20
+canonical-url: https://www.comet.com/docs/opik/changelog
```
JetoPistola, line 2: ❓ **question** / 💡 **suggestion** | Docs / SEO / v1-vs-latest

Same as on `2026-07-13.mdx`. The per-date page resolves at `/docs/opik/v1/changelog/2026/7/20` (**200** on prod), but note the `/v1/`: this entry lives in `docs/changelog/` (wired into the **v1** version only). The **latest** docs read from `docs-v2/changelog/`, which stops at `2026-07-06` — so `2026-07-20` exists only in v1.

Since these entries look new and v2 is the active docs, the deeper question is whether they should be ported into `docs-v2/changelog/` (latest). If they stay v1-only, the correct canonical is the v1 per-date page:

```suggestion
canonical-url: https://www.comet.com/docs/opik/v1/changelog/2026/7/20
```

If pointing at the index is a deliberate convention shared with the other 59 entries, feel free to resolve — but the v1/v2 content gap is worth a separate follow-up.

PR #7625, opened 2026-07-27
[OPIK-4324] [BE] fix: keep react-service auth errors as client errors when the body is not JSON

thiagohora:
## Details

`RemoteAuthService.verifyResponse` read non-successful react-service replies as `ReactServiceErrorResponse` unconditionally. That assumption does not hold for endpoints guarded by Dropwizard's `@Auth` filter (`/opik/auth-session`): an expired or invalid session cookie is rejected by the default `UnauthorizedHandler`, which answers `401` with a `text/plain` body. Reading it as JSON made Jersey raise `MessageBodyProviderNotFoundException` — a `ProcessingException`, not a `ClientErrorException` — so it escaped `AuthFilter` and was mapped by `LoggingExceptionMapper` to a **`500`** instead of the intended `401`. Clients could not tell an expired session from a server fault, so the UI never prompted for re-login and every in-flight request on the page failed at once.

- New `readErrorMessage` helper reads the remote message only when the response body is actually JSON; any other content type is logged (status, content type, raw body) and a caller-facing fallback message is used, so remote framework wording is never surfaced to the client.
- Applied to the `401` and `400` branches of `verifyResponse` and to the `400` branch of `getWorkspaceIdFromResponse`.
- The entity is read at most once, since a client response stream cannot be consumed twice.
- Status codes are unchanged: `401` stays `401`, `400` stays `400`. Behaviour for JSON error bodies is unchanged.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-4324

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: Root-cause investigation from production logs, the `readErrorMessage` helper and its call sites, and the new parameterized regression tests.
- Human verification: Author reviewed the diff, confirmed the failure mode against production logs and the upstream Dropwizard `DefaultUnauthorizedHandler` source, and ran the test suite locally both with and without the fix.

## Testing

`mvn -o surefire:test -Dtest='RemoteAuthServiceTest'` in `apps/opik-backend` — **37 tests pass** (29 before, 8 new).

Added `nonJsonErrorBodyArgs` and drove it through both auth paths (`testSessionAuth__whenRemoteErrorBodyIsNotJson__thenClientErrorInsteadOfServerError` and `testAuth__whenRemoteErrorBodyIsNotJson__thenClientErrorInsteadOfServerError`), covering:

| Remote reply | Expected |
|---|---|
| `401` `text/plain` with Dropwizard's exact default body | `ClientErrorException`, status `401`, message `Please login first` |
| `401` `text/html` | `ClientErrorException`, status `401` |
| `401` `text/plain` with an empty body | `ClientErrorException`, status `401` |
| `400` `text/plain` | `ClientErrorException`, status `400`, fallback message |

Regression check: with the production change reverted and only the tests applied, all 8 fail with exactly the reported `MessageBodyProviderNotFoundException: MessageBodyReader not found for media type=text/plain, type=class com.comet.opik.api.ReactServiceErrorResponse` — so the tests are not vacuous. Existing JSON-body coverage (`unauthorizedArgs`, gzip/identity-encoding, cache-hit, workspace-authorization) still passes unchanged.

Not run locally: the full backend suite and the resource-level integration tests, which need the container harness — left to CI.

## Documentation

No documentation change: internal error-handling fix with no API-surface or configuration change.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
@@ -45,6 +45,9 @@ class RemoteAuthService implements AuthService {
     private static final String USER_NOT_FOUND = "User not found";
     private static final String NOT_LOGGED_USER = "Please login first";
 
+    // Remote error bodies are arbitrary upstream content, so cap what reaches the logs.
+    private static final int MAX_LOGGED_BODY_LENGTH = 512;
+
     // GenericType instances are thread-safe and expensive to build, so reuse a single instance.
     private static final GenericType<List<WorkspaceIdNameResponse>> WORKSPACE_LIST_TYPE = new GenericType<>() {
     };
@@ -284,13 +287,19 @@ class RemoteAuthService implements AuthService {
         return new InternalServerErrorException("Unexpected error while " + operation);
     }
 
+    /**
+     * Reads the response body for diagnostics only, never to be surfaced to the caller. The result is capped at
+     * {@link #MAX_LOGGED_BODY_LENGTH} characters: a remote error body is arbitrary upstream content (a proxy error page
+     * or a stack trace, not necessarily our own JSON), so it must not be able to flood the logs or carry an unbounded
+     * amount of upstream detail into them.
+     */
     private static String readBodySafely(Response response) {
         try {
             if (!response.hasEntity()) {
                 return "";
             }
             bufferEntitySafely(response);
-            return response.readEntity(String.class);
+            return StringUtils.abbreviate(response.readEntity(String.class), MAX_LOGGED_BODY_LENGTH);
         } catch (RuntimeException e) {
             log.warn("Failed to read remote response body for debugging", e);
             return "";
@@ -499,7 +508,8 @@ class RemoteAuthService implements AuthService {
             return response.readEntity(String.class);
         } else if (response.getStatus() == Response.Status.BAD_REQUEST.getStatusCode()) {
             var message = readErrorMessage(response, MISSING_WORKSPACE);
-            log.error("Workspace not found by name: '{}'", message);
+            // An unknown workspace name is a caller mistake on the public-endpoint fallback path, not a server fault.
+            log.warn("Workspace not found by name: '{}'", message);
             throw new ClientErrorException(message, Response.Status.BAD_REQUEST);
         }
 
--- a/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
@@ -295,10 +295,9 @@ class RemoteAuthService implements AuthService {
      */
     private static String readBodySafely(Response response) {
         try {
-            if (!response.hasEntity()) {
+            if (!isEntityReadable(response)) {
                 return "";
             }
-            bufferEntitySafely(response);
             return StringUtils.abbreviate(response.readEntity(String.class), MAX_LOGGED_BODY_LENGTH);
         } catch (RuntimeException e) {
             log.warn("Failed to read remote response body for debugging", e);
@@ -307,15 +306,22 @@ class RemoteAuthService implements AuthService {
     }
 
     /**
-     * Buffers the response entity so it can be read more than once. A client response entity is backed by a single-shot
-     * input stream, so without buffering the first reader consumes it and every later read fails. Buffering is
-     * idempotent, so callers do not need to track whether it already happened.
+     * Guards a {@code readEntity} call: reports whether the response carries an entity and that entity was buffered, so
+     * that it can be read, and read again. A client response entity is backed by a single-shot input stream, so without
+     * buffering the first reader consumes it and every later read fails. Buffering is idempotent — an already buffered
+     * entity reports success — so callers do not need to track whether it already happened.
+     *
+     * @return {@code false} when there is no entity, or when it could not be buffered and is therefore unsafe to read
      */
-    private static void bufferEntitySafely(Response response) {
+    private static boolean isEntityReadable(Response response) {
+        if (!response.hasEntity()) {
+            return false;
+        }
         try {
-            response.bufferEntity();
+            return response.bufferEntity();
         } catch (RuntimeException e) {
             log.warn("Failed to buffer remote response entity, status: '{}'", response.getStatus(), e);
+            return false;
         }
     }
 
@@ -332,17 +338,16 @@ class RemoteAuthService implements AuthService {
      * <p>
      * Only a JSON body is treated as a message intended for the caller. Any other content type is a framework-level
      * response rather than an application error, so it is logged and the caller-facing {@code fallback} is used instead
-     * of leaking the remote framework's wording. The entity is buffered up front so that both the diagnostic logging
-     * here and any later reader of the same response can read it, rather than relying on being the first to consume the
-     * single-shot entity stream.
+     * of leaking the remote framework's wording. Every read is guarded by {@link #isEntityReadable(Response)}, which
+     * also buffers, so that both the diagnostic logging here and any later reader of the same response can read it
+     * rather than relying on being the first to consume the single-shot entity stream.
      *
      * @param fallback message used when the body is absent, not JSON, unreadable or empty
      */
     private static String readErrorMessage(Response response, String fallback) {
-        if (!response.hasEntity()) {
+        if (!isEntityReadable(response)) {
             return fallback;
         }
-        bufferEntitySafely(response);
         var mediaType = response.getMediaType();
         if (mediaType == null || !MediaType.APPLICATION_JSON_TYPE.isCompatible(mediaType)) {
             log.warn("React service replied with a non-JSON error body, status: '{}', contentType: '{}', body: '{}'",
--- a/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
@@ -286,10 +286,27 @@ class RemoteAuthService implements AuthService {
 
     private static String readBodySafely(Response response) {
         try {
-            return response.hasEntity() ? response.readEntity(String.class) : "";
+            if (!response.hasEntity()) {
+                return "";
+            }
+            bufferEntitySafely(response);
+            return response.readEntity(String.class);
         } catch (RuntimeException e) {
-            log.warn("Failed to read remote response body for debugging: {}", e.getMessage());
+            log.warn("Failed to read remote response body for debugging", e);
             return "";
+        }
+    }
+
+    /**
+     * Buffers the response entity so it can be read more than once. A client response entity is backed by a single-shot
+     * input stream, so without buffering the first reader consumes it and every later read fails. Buffering is
+     * idempotent, so callers do not need to track whether it already happened.
+     */
+    private static void bufferEntitySafely(Response response) {
+        try {
+            response.bufferEntity();
+        } catch (RuntimeException e) {
+            log.warn("Failed to buffer remote response entity, status: '{}'", response.getStatus(), e);
         }
     }
 
@@ -306,8 +323,9 @@ class RemoteAuthService implements AuthService {
      * <p>
      * Only a JSON body is treated as a message intended for the caller. Any other content type is a framework-level
      * response rather than an application error, so it is logged and the caller-facing {@code fallback} is used instead
-     * of leaking the remote framework's wording. The entity is read at most once, since a client response stream
-     * cannot be consumed twice.
+     * of leaking the remote framework's wording. The entity is buffered up front so that both the diagnostic logging
+     * here and any later reader of the same response can read it, rather than relying on being the first to consume the
+     * single-shot entity stream.
      *
      * @param fallback message used when the body is absent, not JSON, unreadable or empty
      */
@@ -315,18 +333,21 @@ class RemoteAuthService implements AuthService {
         if (!response.hasEntity()) {
             return fallback;
         }
+        bufferEntitySafely(response);
         var mediaType = response.getMediaType();
         if (mediaType == null || !MediaType.APPLICATION_JSON_TYPE.isCompatible(mediaType)) {
-            log.warn("React service replied status {} with non-JSON content type '{}', body: '{}'",
+            log.warn("React service replied with a non-JSON error body, status: '{}', contentType: '{}', body: '{}'",
                     response.getStatus(), mediaType, readBodySafely(response));
             return fallback;
         }
         try {
-            var message = response.readEntity(ReactServiceErrorResponse.class).msg();
-            return StringUtils.isBlank(message) ? fallback : message.strip();
+            var errorResponse = response.readEntity(ReactServiceErrorResponse.class);
+            return errorResponse == null || StringUtils.isBlank(errorResponse.msg())
+                    ? fallback
+                    : errorResponse.msg().strip();
         } catch (RuntimeException e) {
-            log.warn("Failed to read react service error response, status: {}, error: {}",
-                    response.getStatus(), e.getMessage());
+            log.warn("Failed to read react service error response, status: '{}', body: '{}'",
+                    response.getStatus(), readBodySafely(response), e);
             return fallback;
         }
     }
@@ -478,7 +499,7 @@ class RemoteAuthService implements AuthService {
             return response.readEntity(String.class);
         } else if (response.getStatus() == Response.Status.BAD_REQUEST.getStatusCode()) {
             var message = readErrorMessage(response, MISSING_WORKSPACE);
-            log.error("Not found workspace by name : {}", message);
+            log.error("Workspace not found by name: '{}'", message);
             throw new ClientErrorException(message, Response.Status.BAD_REQUEST);
         }
 
--- a/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
@@ -293,6 +293,44 @@ class RemoteAuthService implements AuthService {
         }
     }
 
+    /**
+     * Extracts the error message from a non-successful react-service response without assuming the body is JSON.
+     * <p>
+     * The react service does not always answer with a {@link ReactServiceErrorResponse}: endpoints guarded by
+     * Dropwizard's {@code @Auth} filter (for example {@code /opik/auth-session}) reject an expired or invalid session
+     * cookie through the default {@code UnauthorizedHandler}, which replies {@code 401} with a {@code text/plain} body.
+     * Reading such a response as {@link ReactServiceErrorResponse} makes Jersey raise
+     * {@code MessageBodyProviderNotFoundException}, a {@code ProcessingException} that is not a
+     * {@link ClientErrorException} and therefore escapes the auth filter and surfaces as a {@code 500} instead of the
+     * intended client error.
+     * <p>
+     * Only a JSON body is treated as a message intended for the caller. Any other content type is a framework-level
+     * response rather than an application error, so it is logged and the caller-facing {@code fallback} is used instead
+     * of leaking the remote framework's wording. The entity is read at most once, since a client response stream
+     * cannot be consumed twice.
+     *
+     * @param fallback message used when the body is absent, not JSON, unreadable or empty
+     */
+    private static String readErrorMessage(Response response, String fallback) {
+        if (!response.hasEntity()) {
+            return fallback;
+        }
+        var mediaType = response.getMediaType();
+        if (mediaType == null || !MediaType.APPLICATION_JSON_TYPE.isCompatible(mediaType)) {
+            log.warn("React service replied status {} with non-JSON content type '{}', body: '{}'",
+                    response.getStatus(), mediaType, readBodySafely(response));
+            return fallback;
+        }
+        try {
+            var message = response.readEntity(ReactServiceErrorResponse.class).msg();
+            return StringUtils.isBlank(message) ? fallback : message.strip();
+        } catch (RuntimeException e) {
+            log.warn("Failed to read react service error response, status: {}, error: {}",
+                    response.getStatus(), e.getMessage());
+            return fallback;
+        }
+    }
+
     private void authenticateUsingSessionToken(Cookie sessionToken, String workspaceName, String path,
             List<String> requiredPermissions) {
         if (isDefaultWorkspace(workspaceName)) {
@@ -371,15 +409,15 @@ class RemoteAuthService implements AuthService {
             }
             return authResponse;
         } else if (response.getStatus() == Response.Status.UNAUTHORIZED.getStatusCode()) {
-            var errorResponse = response.readEntity(ReactServiceErrorResponse.class);
-            throw new ClientErrorException(errorResponse.msg(), Response.Status.UNAUTHORIZED);
+            throw new ClientErrorException(readErrorMessage(response, NOT_LOGGED_USER),
+                    Response.Status.UNAUTHORIZED);
         } else if (response.getStatus() == Response.Status.FORBIDDEN.getStatusCode()) {
             // EM never returns FORBIDDEN as of now
             throw new ClientErrorException(
                     NOT_ALLOWED_TO_ACCESS_WORKSPACE, Response.Status.FORBIDDEN);
         } else if (response.getStatus() == Response.Status.BAD_REQUEST.getStatusCode()) {
-            var errorResponse = response.readEntity(ReactServiceErrorResponse.class);
-            throw new ClientErrorException(errorResponse.msg(), Response.Status.BAD_REQUEST);
+            throw new ClientErrorException(readErrorMessage(response, MISSING_WORKSPACE),
+                    Response.Status.BAD_REQUEST);
         }
         throw unexpectedRemoteError("authenticating user", response);
     }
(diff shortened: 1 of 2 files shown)


apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
```diff
@@ -293,6 +293,44 @@ private static String readBodySafely(Response response) {
...
+     * <p>
+     * Only a JSON body is treated as a message intended for the caller. Any other content type is a framework-level
+     * response rather than an application error, so it is logged and the caller-facing {@code fallback} is used instead
+     * of leaking the remote framework's wording. The entity is read at most once, since a client response stream
+     * cannot be consumed twice.
+     *
+     * @param fallback message used when the body is absent, not JSON, unreadable or empty
+     */
+    private static String readErrorMessage(Response response, String fallback) {
+        if (!response.hasEntity()) {
```
andrescrz: Let's add an add condition with a `bufferEntity` call here. You can double check the chain of calls for this method, but anyway we need to make sure it's reliable against multiple reads of the entity, not just checking the entity exists.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
```diff
@@ -293,6 +293,44 @@ private static String readBodySafely(Response response) {
...
+            return fallback;
+        }
+        var mediaType = response.getMediaType();
+        if (mediaType == null || !MediaType.APPLICATION_JSON_TYPE.isCompatible(mediaType)) {
+            log.warn("React service replied status {} with non-JSON content type '{}', body: '{}'",
```
andrescrz: Nit: missing single quotes for first printed value (status). Also nice, to move all printed values to the end of the sentence, so it's easier to grep during production debugging sessions.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
```diff
@@ -293,6 +293,44 @@ private static String readBodySafely(Response response) {
...
+                    response.getStatus(), mediaType, readBodySafely(response));
+            return fallback;
+        }
+        try {
+            var message = response.readEntity(ReactServiceErrorResponse.class).msg();
```
andrescrz: Nit: not sure if readEntity can ever return null and it would work fine as we have a catch below with the fallback. You can optionally concat the null check to the conditional below.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
```diff
@@ -293,6 +293,44 @@ private static String readBodySafely(Response response) {
...
+            return StringUtils.isBlank(message) ? fallback : message.strip();
+        } catch (RuntimeException e) {
+            log.warn("Failed to read react service error response, status: {}, error: {}",
```
andrescrz: Minor: same about missing single quotes for the printed data.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
```diff
@@ -293,6 +293,44 @@ private static String readBodySafely(Response response) {
...
+                    response.getStatus(), e.getMessage());
```
andrescrz: Minor: better pass the whole exception to the logger at last param (without `{}` , so the full stack trace is printed, instead of just the message. We might miss the information while debugging.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
```diff
@@ -439,9 +477,9 @@ private String getWorkspaceIdFromResponse(Response response) {
         if (response.getStatusInfo().getFamily() == Response.Status.Family.SUCCESSFUL) {
             return response.readEntity(String.class);
         } else if (response.getStatus() == Response.Status.BAD_REQUEST.getStatusCode()) {
-            var errorResponse = response.readEntity(ReactServiceErrorResponse.class);
-            log.error("Not found workspace by name : {}", errorResponse.msg());
-            throw new ClientErrorException(errorResponse.msg(), Response.Status.BAD_REQUEST);
+            var message = readErrorMessage(response, MISSING_WORKSPACE);
+            log.error("Not found workspace by name : {}", message);
```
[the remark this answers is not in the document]
andrescrz: Minor nit you could address.

andrescrz (commented):
My only blocker is the lack of buffered entity read, in order to protect this in the long run.

The rest is minor comments.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
thiagohora: Done in 131ddeac45. Added a `bufferEntitySafely` helper called before any read, in both `readErrorMessage` and `readBodySafely` — `bufferEntity()` is idempotent, so callers do not have to track whether it already happened.

Checking the chain confirmed a real second read: in the malformed-JSON path `readEntity(ReactServiceErrorResponse.class)` consumes the stream and throws, then the recovery branch reads the body again for the log. Verified empirically — without the buffer that second read fails and logs `Failed to read remote response body for debugging` with `body: `; with it the raw body comes through:

```
WARN ... Failed to read react service error response, status: '401', body: 'not json at all'
```

Added `malformedJsonErrorBodyArgs` (5 cases: unparseable body, truncated JSON, blank `msg`, `{}`, and a 400 variant) to pin the client-facing outcome of that path. To be explicit about coverage: those assert the exception type/status/message, not the logged body — asserting log content would need a log-appender harness the suite does not have today, so the buffered re-read itself is verified manually as above.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
thiagohora: Done — now `"React service replied with a non-JSON error body, status: '{}', contentType: '{}', body: '{}'"` with all values quoted and moved to the end.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
thiagohora: Done — folded the null check into the conditional: `errorResponse == null || StringUtils.isBlank(errorResponse.msg())`. You are right that the `catch` would have covered an NPE, but the explicit check keeps it out of the exception path.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
thiagohora: Done — quoted and moved to the end, same as above.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
thiagohora: Done — the exception is now the trailing argument with no placeholder, so the full stack trace is logged. Applied the same fix to the pre-existing `readBodySafely` and to the new `bufferEntitySafely`.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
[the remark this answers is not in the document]
thiagohora: Done — now `log.error("Workspace not found by name: '{}'", message)`.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
```diff
@@ -286,13 +286,72 @@ private InternalServerErrorException unexpectedRemoteError(String operation, Res
...
+        if (!response.hasEntity()) {
+            return fallback;
+        }
+        bufferEntitySafely(response);
+        var mediaType = response.getMediaType();
+        if (mediaType == null || !MediaType.APPLICATION_JSON_TYPE.isCompatible(mediaType)) {
+            log.warn("React service replied with a non-JSON error body, status: '{}', contentType: '{}', body: '{}'",
+                    response.getStatus(), mediaType, readBodySafely(response));
+            return fallback;
+        }
```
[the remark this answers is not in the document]
thiagohora: Good catch — addressed in 0b93998bad by capping it rather than redacting it.

`readBodySafely` now returns `StringUtils.abbreviate(body, MAX_LOGGED_BODY_LENGTH)` with the cap at 512 chars. That helper is the single place any remote body is read for diagnostics, so the cap covers all three call sites, including the pre-existing one in `unexpectedRemoteError`. Verified with a 5,000-char body:

```
WARN ... non-JSON error body, status: '401', contentType: 'text/html', body: 'xxxx…xxx...'
```

I went with truncation instead of redaction deliberately: @andrescrz's blocker on this PR was specifically about making the raw body *available* for production debugging, and blanket redaction would undo that. The reasoning on exposure — the request we send to `/opik/auth`+`/opik/auth-session` carries only `workspaceName`, `path` and `requiredPermissions` in the body; the API key travels in the `Authorization` header and the session in a cookie, so neither is in scope for an error body echoing our request. What is unbounded is *upstream* content (proxy error pages, stack traces), which is exactly what the cap addresses.

Flagging honestly: a cap bounds the volume, it does not guarantee a pathological upstream can never put a token in the first 512 chars. If the team wants a stricter rule here, the natural follow-up is a shared redaction utility applied wherever remote bodies are logged, rather than something local to this method — happy to open a ticket for that.

Added a `text/html` 5,000-char case to `nonJsonErrorBodyArgs` so the oversized path is exercised; the assertions cover the client-facing outcome, not the log text (no log-appender harness in this suite).

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
```diff
@@ -439,9 +498,9 @@ private String getWorkspaceIdFromResponse(Response response) {
         if (response.getStatusInfo().getFamily() == Response.Status.Family.SUCCESSFUL) {
             return response.readEntity(String.class);
         } else if (response.getStatus() == Response.Status.BAD_REQUEST.getStatusCode()) {
-            var errorResponse = response.readEntity(ReactServiceErrorResponse.class);
-            log.error("Not found workspace by name : {}", errorResponse.msg());
-            throw new ClientErrorException(errorResponse.msg(), Response.Status.BAD_REQUEST);
+            var message = readErrorMessage(response, MISSING_WORKSPACE);
+            log.error("Workspace not found by name: '{}'", message);
```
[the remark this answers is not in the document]
thiagohora: Agreed — changed to `log.warn` in 0b93998bad.

`getWorkspaceId` is only reached on the public-endpoint fallback path (`authenticate` catch block, unauthenticated request to a public endpoint), so a `400` there means the caller passed a workspace name that does not exist. That is caller-driven and expected, not a server fault. Went with `WARN` over `DEBUG` to keep it visible without paging anyone.

Two notes for the reviewers: the `ERROR` level here is pre-existing — this PR only reworded the message — and the sibling `WorkspaceNameService.getWorkspaceNameFromResponse` still logs its not-found case at `ERROR`. Happy to revert for consistency, or to align that one too in a follow-up, whichever the team prefers.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
```diff
@@ -284,15 +287,80 @@ private InternalServerErrorException unexpectedRemoteError(String operation, Res
...
     }
 
+    /**
+     * Buffers the response entity so it can be read more than once. A client response entity is backed by a single-shot
+     * input stream, so without buffering the first reader consumes it and every later read fails. Buffering is
+     * idempotent, so callers do not need to track whether it already happened.
+     */
+    private static void bufferEntitySafely(Response response) {
+        try {
+            response.bufferEntity();
```
andrescrz: I believe I didn't make this clear previously, this change is good (buffering + try-catch as fallback), but I believe we're still missing to check the boolean result from this call and adding it as `and` condition on the `hasEntity` check. 

Basically, those checks should go together for the majority of the cases before a `readEntity`. 

Otherwise, the `readEntity` call is not guarded against errors. 

Here, you could simply propagate the boolean from the return type of this aux method, and return false on the catch below.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
```diff
@@ -284,15 +287,80 @@ private InternalServerErrorException unexpectedRemoteError(String operation, Res
...
+     * here and any later reader of the same response can read it, rather than relying on being the first to consume the
+     * single-shot entity stream.
+     *
+     * @param fallback message used when the body is absent, not JSON, unreadable or empty
+     */
+    private static String readErrorMessage(Response response, String fallback) {
+        if (!response.hasEntity()) {
+            return fallback;
+        }
+        bufferEntitySafely(response);
```
andrescrz: Applies same comment as above.

andrescrz (commented):
Just a important follow up from previous revision. The rest LGTM.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
```diff
@@ -371,15 +439,15 @@ private AuthResponse verifyResponse(Response response) {
...
+            throw new ClientErrorException(readErrorMessage(response, NOT_LOGGED_USER),
+                    Response.Status.UNAUTHORIZED);
         } else if (response.getStatus() == Response.Status.FORBIDDEN.getStatusCode()) {
             // EM never returns FORBIDDEN as of now
             throw new ClientErrorException(
                     NOT_ALLOWED_TO_ACCESS_WORKSPACE, Response.Status.FORBIDDEN);
         } else if (response.getStatus() == Response.Status.BAD_REQUEST.getStatusCode()) {
-            var errorResponse = response.readEntity(ReactServiceErrorResponse.class);
-            throw new ClientErrorException(errorResponse.msg(), Response.Status.BAD_REQUEST);
+            throw new ClientErrorException(readErrorMessage(response, MISSING_WORKSPACE),
```
JetoPistola, line 474: ❓ **question** | Architecture

This `400` branch falls back to `MISSING_WORKSPACE` (`"Workspace name should be provided"`) when the body isn't JSON. In `getWorkspaceIdFromResponse` that fallback is spot-on — a 400 there genuinely *is* a workspace-name lookup failure. But here in `verifyResponse`, a non-JSON `400` from `/opik/auth` or `/opik/auth-session` could be any client error, so surfacing "Workspace name should be provided" may mislead the caller about what actually went wrong.

Is `MISSING_WORKSPACE` the intended fallback for this path, or would a more neutral 400 message be a better fit? Note this only affects the rare non-JSON 400 — the JSON path is unchanged, so behaviour for real `ReactServiceErrorResponse` bodies is preserved.

JetoPistola:
👋 **Review summary**

**What looks good**
- Excellent root-cause writeup — the `text/plain` 401 → Jersey `MessageBodyProviderNotFoundException` (a `ProcessingException`, not a `ClientErrorException`) → `500` chain is precisely traced, including the upstream Dropwizard `DefaultUnauthorizedHandler` behaviour.
- The `readErrorMessage` helper is the right seam: it treats only a JSON body as caller-facing and logs anything else instead of leaking remote framework wording. Content-type checked via `isCompatible`, so `application/json; charset=utf-8` still matches.
- Nice defensive touches: idempotent `bufferEntity` so the body can be re-read for diagnostics, `MAX_LOGGED_BODY_LENGTH` cap so an upstream proxy error page can't flood the logs, and null/blank-`msg()` handling that also removes a latent NPE the old code had on an empty JSON body.
- Tests are thorough and explicitly proven non-vacuous (revert-the-fix check → all 8 fail with the exact `MessageBodyProviderNotFoundException`). The `malformedJsonErrorBodyArgs` cases pin the JSON-parse-failure recovery path too, and the oversized-body case exercises the log cap.

**Overall**
Tightly-scoped, well-reasoned error-handling fix. Status codes are preserved (`401` stays `401`, `400` stays `400`), JSON-body behaviour is unchanged, and the blast radius is limited to the non-JSON path that was previously misfiring as a `500`. No blocking concerns.

**Inline comments**: 1 question (non-blocking) — on the `MISSING_WORKSPACE` fallback used for a generic non-JSON `400` in `verifyResponse`.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
thiagohora: Understood now — thanks for spelling it out. Done in 5f8072e9ae.

Folded both checks into one guard that returns the boolean, so `readEntity` is never reached unless the entity exists *and* was buffered:

```java
private static boolean isEntityReadable(Response response) {
    if (!response.hasEntity()) {
        return false;
    }
    try {
        return response.bufferEntity();
    } catch (RuntimeException e) {
        log.warn("Failed to buffer remote response entity, status: '{}'", response.getStatus(), e);
        return false;
    }
}
```

Both readers now start with `if (!isEntityReadable(response)) { return fallback; }`, replacing the bare `hasEntity` check plus a separate unchecked buffer call.

One thing I checked before wiring it this way, because it would have broken the diagnostic re-read: whether `bufferEntity()` returns `false` on a *second* call for an already-buffered entity — `readBodySafely` is called from inside `readErrorMessage` after it has already buffered, so a `false` there would have made the guard swallow the body. It returns `true`: in Jersey `InboundMessageContext.bufferEntity()` the first branch is `if (entityContent.isBuffered() || !entityContent.hasContent()) return true;`. Confirmed at runtime too — the recovery-path log still carries the body:

```
WARN ... Failed to read react service error response, status: '401', body: 'not json at all'
```

`RemoteAuthServiceTest`: 44 pass.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
thiagohora: Covered by the same change in 5f8072e9ae — `readErrorMessage` now opens with `if (!isEntityReadable(response)) { return fallback; }`, so this read is guarded by the buffer result too. Details in the reply above.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
thiagohora: Fair point on the naming, and you have read the scope exactly right — this only affects the non-JSON 400.

Keeping `MISSING_WORKSPACE` deliberately, though. The pre-existing behaviour of this branch was:

```java
var errorResponse = response.readEntity(ReactServiceErrorResponse.class);
throw new ClientErrorException(errorResponse.msg(), Response.Status.BAD_REQUEST);
```

i.e. surface the react service's own `msg()`, with no fallback constant at all. That path is preserved verbatim: for a real `ReactServiceErrorResponse` the caller still gets `msg()`, so there is no behaviour change for any 400 that was previously working. The fallback only fires when there is no JSON message to surface — and that case used to produce a 500, so nothing is being downgraded.

I did try a neutral `INVALID_AUTH_REQUEST` constant after reading your comment, then reverted it: it introduces a new user-facing error string for a case where we have no information about the cause, which is a second behaviour change layered on top of the fix this PR is actually about. Not worth widening the blast radius of an error-handling fix.

So: agreed the constant name is a poor fit for a generic `/opik/auth` 400, but the right place to settle that is a follow-up that looks at the react service's 400 contract as a whole, rather than picking a new string here. Happy to open that ticket if the team wants it.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/auth/RemoteAuthService.java
```diff
@@ -284,15 +288,100 @@ private InternalServerErrorException unexpectedRemoteError(String operation, Res
...
+     * and a non-{@code application} type ({@code text/json}) both deserialize fine, so gating on
+     * {@code APPLICATION_JSON_TYPE.isCompatible} would discard a perfectly good remote message. A wildcard or absent
+     * type tells us nothing about the body and is not treated as JSON.
+     */
+    private static boolean isJson(MediaType mediaType) {
+        if (mediaType == null || MediaType.MEDIA_TYPE_WILDCARD.equals(mediaType.getSubtype())) {
+            return false;
+        }
+        var subtype = mediaType.getSubtype().toLowerCase(Locale.ROOT);
+        return "json".equals(subtype) || subtype.endsWith("+json");
```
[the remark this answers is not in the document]
thiagohora: Changed in a927e8cc33 — `isJson` now rejects any subtype *containing* a wildcard, not just the bare `*`.

Correcting the stated failure mode though: `application/*+json` did **not** hit a provider or parsing failure. I added it to the matrix and ran it against the code as it stood — Jackson deserialized it and the caller got the real remote message:

```
[9] application/*+json
Expecting message to be: "Please login first"
but was: "remote error message"
```

And even if the parse had failed, that branch is inside the `try` with `readEntity`, so it returns the fallback rather than a 500 — the exception this PR is about cannot escape there any more.

So this is a consistency fix, not a bug fix: a wildcard is not a legal response content type, so it tells us nothing about the body, and trusting `application/*+json` while rejecting `*/*` and `application/*` was incoherent. Worth being explicit that it *is* a small behaviour delta — `main` surfaced the remote message for that malformed content type too, and now we return the fallback. I think that trade is right for an input that cannot legally occur, but happy to be overruled if anyone wants strict parity.

Matrix now pins `application/*` and `application/*+json` alongside `*/*`. 54 tests pass.

JetoPistola:
🔁 **Re-review** (head `1d36c59`)

Re-reviewed the changes since my first pass — no new findings from me.

- `isEntityReadable` (boolean guard) is a nice robustness improvement over the earlier buffering helper.
- The `isJson` helper reads well and the parameterized content-type test that pins it is solid.

My earlier question on the `MISSING_WORKSPACE` fallback is resolved. Ping me for approval once you're happy the outstanding threads are settled. 👍

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/auth/RemoteAuthServiceTest.java
```diff
@@ -413,6 +415,166 @@ void testSessionAuthUnauthorized(int remoteAuthStatusCode, Class<? extends Excep
...
+
+    /**
+     * Pins which content types are still read as JSON. Anything the Jackson provider can deserialize must keep
+     * surfacing the remote {@code msg()}, exactly as it did before this class started gating on the content type;
+     * everything else resolves to the caller-facing fallback instead of a server error.
+     */
+    @ParameterizedTest
+    @MethodSource("errorBodyContentTypeArgs")
+    void testSessionAuth__whenRemoteRepliesUnauthorized__thenOnlyJsonContentTypesSurfaceRemoteMessage(
+            String contentType, String expectedMessage) {
```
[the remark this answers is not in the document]
thiagohora: Renamed in a927e8cc33 — agreed, the old name undersold the contract.

Went with `thenJsonSubtypesSurfaceRemoteMessage` rather than your `thenJsonCompatibleContentTypesSurfaceRemoteMessage`, because "compatible" is the one word that would mislead here: `MediaType.isCompatible` is exactly the check this PR moved *away* from — it rejects `application/problem+json` and `text/json`, which was the regression. The contract is "subtype is `json` or carries a `+json` suffix", so "subtypes" names it without pointing at the API we deliberately do not use. Updated the javadoc to spell the rule out too.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/auth/RemoteAuthServiceTest.java
```diff
@@ -413,6 +415,166 @@ void testSessionAuthUnauthorized(int remoteAuthStatusCode, Class<? extends Excep
...
+                arguments(HttpStatus.SC_UNAUTHORIZED, "text/plain", "", NOT_LOGGED_USER),
+                // a proxy error page far larger than the logged-body cap must still resolve to a client error
+                arguments(HttpStatus.SC_UNAUTHORIZED, "text/html", "x".repeat(5_000), NOT_LOGGED_USER),
+                arguments(HttpStatus.SC_BAD_REQUEST, "text/plain", "Bad Request", MISSING_WORKSPACE));
+    }
+
+    @ParameterizedTest
+    @MethodSource("nonJsonErrorBodyArgs")
+    void testSessionAuth__whenRemoteErrorBodyIsNotJson__thenClientErrorInsteadOfServerError(
+            int remoteAuthStatusCode, String contentType, String body, String expectedMessage) {
```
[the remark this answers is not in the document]
thiagohora: Done in a486ced035 — you are right, and the repo backs it up more strongly than I first assumed.

`testing.md` lists `void testCreateUser() { }` under ❌, and counting the backend suite:

| style | methods |
|---|---|
| no prefix | 4354 |
| `test*` prefix | 280 |
| no prefix, already `subject__whenX__thenY` | 1185 |

So the prefix is the ~6% outlier, and 1185 methods already use the exact shape I had — just without the `test`. Renamed the five tests this PR adds:

- `sessionAuth__whenRemoteErrorBodyIsNotJson__thenClientErrorInsteadOfServerError`
- `auth__whenRemoteErrorBodyIsNotJson__thenClientErrorInsteadOfServerError`
- `sessionAuth__whenRemoteErrorBodyIsMalformedJson__thenClientErrorAndBodyReReadForDiagnostics`
- `sessionAuth__whenRemoteRepliesUnauthorized__thenJsonSubtypesSurfaceRemoteMessage`
- `sessionAuth__whenRemoteRepliesUnauthorizedWithoutContentType__thenFallbackMessage`

Deliberately **not** renaming the 17 pre-existing `test*` methods in this class. That leaves `RemoteAuthServiceTest` temporarily mixed (17 prefixed, 5 not), which I know is not pretty — but converting them is a rename-only change unrelated to an error-handling fix, it would bloat the diff reviewers have to read, and it breaks anyone with a `-Dtest=testSessionAuth*` invocation in muscle memory. Happy to do the whole class in a separate mechanical PR if the team wants the file converted.

54 tests pass.

andrescrz (approved):
LGTM, thanks for addressing al the comments!

PR #7628, opened 2026-07-27
[OPIK-7398] [BE] feat: create spans_local_v2 shadow table with weekly partitioning and traces-informed codecs

thiagohora:
## Details

Creates `spans_local_v2` — the weekly-partitioned, tiered/sharding-ready successor to `spans` — empty, next to the live table. Counterpart of migration `000101` (`traces_local_v2`, OPIK_6900); no cutover, backfill or `storage_policy` here.

Deltas from live `spans`, all locked at creation:
- `PARTITION BY toMonday(id_at)` (weekly), `id_at` derived from the row's UUIDv7 `id`, so a row lands in the same partition across every ReplacingMergeTree upsert.
- `ORDER BY (workspace_id, project_id, trace_id, parent_span_id, id)` unchanged — it is the dedup key and must match for the cutover to dedup the same rows.
- Timestamps narrowed to `DateTime64(6)`; `end_time`/`ttft`/`duration` de-`Nullable`d onto the epoch/NaN sentinels `SentinelTranslation` already covers; `is_deleted` added as the delete meta-column.
- `idx_spans_id_at` minmax index; `index_granularity_bytes = 40 MiB`; no `storage_policy` (later environment-gated migration).
- Codecs are the traces-final set: all 28 columns spans shares with `traces_local_v2` carry the exact codec `TracesLocalV2BenchmarkTest`'s pin map records after `000101` + `000106` + `000107`, verified column for column rather than by eye. The spans-only columns follow the same rules that map produces — `FixedString(36)` ids (`trace_id`, `parent_span_id`) on `ZSTD(1)`, `Enum8` `type` on `ZSTD(1)`, and small repetitive text (`model`, `provider`, `total_estimated_cost_version`) on the 26.3-safe `ZSTD(3)`. Only `usage` (a `Map` of repetitive string keys and narrow-integer values, which no single-column rule covers) and `total_estimated_cost` (`Decimal128`) remain best guesses, left for the confirmatory spans codec benchmark to settle.

Three deviations worth reviewer attention, all called out in the migration header:
- **`parent_span_id` is `FixedString(36)`**, not `String`: it holds a UUID or nothing, so absent becomes the empty sentinel that ClickHouse stores as 36 NUL bytes. `SentinelTranslation.emptyUuidToNull`/`nullToEmptyUuid` already handle both forms, but the live DAO's `LENGTH(parent_span_id) > 0` presence checks are always 36 here and must move to those helpers at cutover.
- **`usage` is `Map(String, Int64)`**, not `Map(String, Int32)`: a lossless widening matching what `SpanDAO` already casts to on update.
- **No `output_keys` column.** The ticket lists it among the ZSTD(3) group, but it is traces-only (added by `000044`) and has no spans counterpart, so there is nothing for the cutover INSERT to map. Similarly, `000101`'s `idx_traces_id_bf` has no equivalent here — the ticket enumerates only `idx_spans_id_at` as added, and a per-granule bloom filter on the largest table deserves its own sizing decision.

The per-column codec pin test (`everyTracesLocalV2ColumnUsesItsIntendedCodec`'s spans counterpart) is deliberately not here — the spans codec benchmark is out of this ticket's scope, and it will land with that story. Until then `showCreateTableMatchesSpec()` pins every codec as part of the whole DDL.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- OPIK-7398

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: Authored the migration and the companion test; the live `spans`/`traces`/`traces_local_v2` DDLs were read from a running cluster to derive the column set and confirm the traces-final codecs.
- Human verification: Design decisions taken from the ticket spec; migration applied and every test run against the deployed ClickHouse 26.3 LTS image locally (below).

## Testing

`mvn test -Dtest='SpansLocalV2TableTest,TracesLocalV2TableTest,TracesLocalV2PartitioningTest'` against `clickhouse/clickhouse-server:26.3.16.16-alpine` + `zookeeper:3.9.4` testcontainers — 32 / 13 / 4 green, so the migration applies cleanly and the existing traces companions are unaffected.

`SpansLocalV2TableTest` covers:
- Full-row round-trip through the sentinel translation and JSON (de)serialization the DAO will use at cutover, across the full `Source` x `SpanType` cross-product (20 combinations).
- Absent `end_time`/`ttft`/`parent_span_id` -> sentinel -> `null`, per `SpanType`.
- A key-columns-only row: every other column server-defaultable, both Enum8 zero values (unreachable from Java, but present in live data) and `is_deleted` defaulting correctly.
- `start_time` at the epoch -> `null` duration; sub-millisecond duration precision (the microsecond `dateDiff`).
- `input`/`output` truncation above a lowered threshold.
- Backdated (3 weeks, 400 days), current and future (2 weeks ahead) ids landing in the id-derived weekly partition, not the current week.
- `SHOW CREATE TABLE` pinned against the spec in one assertion — types, defaults, every codec, engine parameters, partition/sorting keys, skip indexes and granularity settings.

Additionally verified out-of-band (not committed — the cutover is a later story): the cutover-shaped `INSERT INTO spans_local_v2 SELECT ... FROM spans` type-checks and behaves, i.e. `Map(String, Int32)` -> `Map(String, Int64)`, `String` -> `FixedString(36)` for `parent_span_id` (root spans still match `= ''`), `DateTime64(9)` -> `DateTime64(6)`, and `NULL` `end_time`/`ttft` -> sentinels yielding a NaN `duration`.

## Documentation

None — internal schema change with no user-facing or API surface. The rationale for every locked choice and every codec lives in the migration header.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/utils/SentinelTranslation.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/utils/SentinelTranslation.java
@@ -4,6 +4,7 @@ import lombok.experimental.UtilityClass;
 import org.apache.commons.lang3.StringUtils;
 
 import java.time.Instant;
+import java.util.UUID;
 
 import static com.comet.opik.utils.ValidationUtils.CLICKHOUSE_FIXED_STRING_UUID_FIELD_NULL_VALUE;
 
@@ -63,6 +64,20 @@ public class SentinelTranslation {
                 : value;
     }
 
+    /**
+     * As {@link #emptyUuidToNull(String)}, parsed to a {@link UUID}: the single call a read path needs for a
+     * {@code FixedString(36)} column holding a UUID or the empty sentinel. Going through the sentinel check first is what
+     * makes it safe — the driver surfaces the sentinel as 36 NUL characters, which are not blank, so parsing the raw
+     * value directly (behind an {@code isBlank} guard, say) throws instead of reading as absent.
+     *
+     * @return {@code null} when the value is absent, otherwise the parsed UUID.
+     * @throws IllegalArgumentException when the value is present but not a UUID.
+     */
+    public static UUID emptyUuidToNullableUuid(String value) {
+        var present = emptyUuidToNull(value);
+        return present == null ? null : UUID.fromString(present);
+    }
+
     // Inbound: API null → ClickHouse sentinel.
 
     public static Instant nullToEpoch(Instant value) {
@@ -80,5 +95,13 @@ public class SentinelTranslation {
      */
     public static String nullToEmptyUuid(String value) {
         return StringUtils.isBlank(value) ? EMPTY_UUID_SENTINEL : value;
+    }
+
+    /**
+     * As {@link #nullToEmptyUuid(String)} for a {@link UUID}, so a write path binding a {@code FixedString(36)} column
+     * does not have to spell out the null check around {@link UUID#toString()}.
+     */
+    public static String nullToEmptyUuid(UUID value) {
+        return value == null ? EMPTY_UUID_SENTINEL : value.toString();
     }
 }
--- a/apps/opik-backend/src/main/java/com/comet/opik/utils/SentinelTranslation.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/utils/SentinelTranslation.java
@@ -4,6 +4,7 @@ import lombok.experimental.UtilityClass;
 import org.apache.commons.lang3.StringUtils;
 
 import java.time.Instant;
+import java.util.Optional;
 import java.util.UUID;
 
 import static com.comet.opik.utils.ValidationUtils.CLICKHOUSE_FIXED_STRING_UUID_FIELD_NULL_VALUE;
@@ -40,10 +41,11 @@ public class SentinelTranslation {
 
     /**
      * @return {@code null} when the value is absent — {@code null} or the epoch sentinel — otherwise the value itself.
-     * The constant leads the equality check so a {@code null} value can never throw.
      */
     public static Instant epochToNull(Instant value) {
-        return EPOCH_SENTINEL.equals(value) ? null : value;
+        return Optional.ofNullable(value)
+                .filter(present -> !EPOCH_SENTINEL.equals(present))
+                .orElse(null);
     }
 
     /**
@@ -51,7 +53,9 @@ public class SentinelTranslation {
      * Infinities are passed through unchanged: they signal a data defect, not the absent-value sentinel.
      */
     public static Double nanToNull(Double value) {
-        return value != null && value.isNaN() ? null : value;
+        return Optional.ofNullable(value)
+                .filter(present -> !present.isNaN())
+                .orElse(null);
     }
 
     /**
@@ -59,9 +63,10 @@ public class SentinelTranslation {
      * {@code FixedString(36)} form — otherwise the value itself.
      */
     public static String emptyUuidToNull(String value) {
-        return StringUtils.isBlank(value) || CLICKHOUSE_FIXED_STRING_UUID_FIELD_NULL_VALUE.equals(value)
-                ? null
-                : value;
+        return Optional.ofNullable(value)
+                .filter(StringUtils::isNotBlank)
+                .filter(present -> !CLICKHOUSE_FIXED_STRING_UUID_FIELD_NULL_VALUE.equals(present))
+                .orElse(null);
     }
 
     /**
@@ -74,18 +79,19 @@ public class SentinelTranslation {
      * @throws IllegalArgumentException when the value is present but not a UUID.
      */
     public static UUID emptyUuidToNullableUuid(String value) {
-        var present = emptyUuidToNull(value);
-        return present == null ? null : UUID.fromString(present);
+        return Optional.ofNullable(emptyUuidToNull(value))
+                .map(UUID::fromString)
+                .orElse(null);
     }
 
     // Inbound: API null → ClickHouse sentinel.
 
     public static Instant nullToEpoch(Instant value) {
-        return value == null ? EPOCH_SENTINEL : value;
+        return Optional.ofNullable(value).orElse(EPOCH_SENTINEL);
     }
 
     public static double nullToNaN(Double value) {
-        return value == null ? Double.NaN : value;
+        return Optional.ofNullable(value).orElse(Double.NaN);
     }
 
     /**
@@ -94,7 +100,9 @@ public class SentinelTranslation {
      * from being stored as a partial-NUL {@code FixedString(36)} that reads back as neither the sentinel nor a UUID.
      */
     public static String nullToEmptyUuid(String value) {
-        return StringUtils.isBlank(value) ? EMPTY_UUID_SENTINEL : value;
+        return Optional.ofNullable(value)
+                .filter(StringUtils::isNotBlank)
+                .orElse(EMPTY_UUID_SENTINEL);
     }
 
     /**
@@ -102,6 +110,8 @@ public class SentinelTranslation {
      * does not have to spell out the null check around {@link UUID#toString()}.
      */
     public static String nullToEmptyUuid(UUID value) {
-        return value == null ? EMPTY_UUID_SENTINEL : value.toString();
+        return Optional.ofNullable(value)
+                .map(UUID::toString)
+                .orElse(EMPTY_UUID_SENTINEL);
     }
 }
--- a/apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000112_create_spans_local_v2_table.sql
+++ b/apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000112_create_spans_local_v2_table.sql
@@ -11,7 +11,7 @@
 --     ReplacingMergeTree upserts.
 --   * Timestamps stored at microsecond precision (DateTime64(6)); nothing ingested needs finer resolution. This
 --     narrows start_time/end_time/created_at from the DateTime64(9) the live table uses (last_updated_at is already
---     microseconds); no ingestion path produces sub-microsecond values, so the cast is lossless in practice.
+--     microseconds); sub-microsecond digits are truncated, never rounded, so a row cannot shift a microsecond.
 --   * end_time, ttft and duration are non-Nullable, using epoch / NaN sentinels, dropping the per-column
 --     null-mask overhead on hot reads. Same three columns as traces_local_v2, and the same sentinels, so
 --     SentinelTranslation covers both tables.
--- a/apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000112_create_spans_local_v2_table.sql
+++ b/apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000112_create_spans_local_v2_table.sql
@@ -134,8 +134,9 @@ CREATE TABLE IF NOT EXISTS ${ANALYTICS_DB_DATABASE_NAME}.spans_local_v2 ON CLUST
         if(end_time = toDateTime64('1970-01-01 00:00:00', 6) OR start_time = toDateTime64('1970-01-01 00:00:00', 6),
             toFloat64('nan'),
             dateDiff('microsecond', start_time, end_time) / 1000.0) CODEC(ZSTD(1)),
-    -- Partition input, see header. DateTime64(3) is what UUIDv7ToDateTime returns, so the millisecond the id encodes is
-    -- kept as-is rather than cast down to whole seconds; 8 bytes instead of 4, but Delta leaves little of that on disk.
+    -- Partition input, see header. Whole seconds: UUIDv7ToDateTime returns DateTime64(3) and the millisecond is dropped,
+    -- which is all a weekly partition and the id-range filters need. DateTime64 rather than DateTime is for range, not
+    -- precision — DateTime wraps past 2106-02-07, and prod holds ids dated 2199.
     id_at                        DateTime64(0, 'UTC') MATERIALIZED UUIDv7ToDateTime(toUUID(id)) CODEC(Delta, ZSTD(1)),
     -- id-only lookups (project and trace unknown) can't use the primary key, where id is the last key column. Two
     -- complementary indexes, as traces_local_v2 has: minmax for the id-range predicates the retention/read paths use
--- a/apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000112_create_spans_local_v2_table.sql
+++ b/apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000112_create_spans_local_v2_table.sql
@@ -36,10 +36,21 @@
 -- No output_keys column: it is a traces-only column (added by 000044 for the trace output-key filters) and has no
 -- counterpart on spans, so there is nothing for the cutover INSERT to map.
 -- Locked at creation (changing any of these later needs another full table rewrite, so they are fixed now):
---   * ORDER BY / implicit PRIMARY KEY (workspace_id, project_id, trace_id, parent_span_id, id): unchanged from the live
---     `spans` table. It matches every hot query (workspace+project filter, then the spans of one trace, then the
---     children of one span) and is the ReplacingMergeTree dedup key, which must stay identical for the cutover to
---     dedup the same rows the live table does.
+--   * ORDER BY / implicit PRIMARY KEY (workspace_id, project_id, trace_id, id): the live `spans` key without
+--     parent_span_id. It matches the hot queries (workspace+project filter, then the spans of one trace, then a span by
+--     id) and is the ReplacingMergeTree dedup key. Dropping parent_span_id is safe and wanted:
+--       - Nothing filters on it. It is not exposed as a SpanField, so it cannot be filtered through the API, and no
+--         query in a 7-day prod sample carries a parent_span_id predicate (0 of 162,172) while 94,238 filter trace_id.
+--         With ~1.7 distinct parents per trace it barely subdivided a trace's ~4.5 spans anyway.
+--       - Dedup identity is unchanged, since id alone is unique per span; the trailing key columns never separated two
+--         rows that (workspace_id, project_id, trace_id, id) would merge.
+--       - It removes a hazard. parent_span_id is mutable, and a mutable sort-key column breaks ReplacingMergeTree
+--         dedup: two versions of a span with different parents sort to different keys and never merge. That is what
+--         forces the deliberate CAST(leftPad(..., 40, '*') AS FixedString(19)) poison pill in SpanDAO's upsert; out of
+--         the key, a changing parent is harmless.
+--     The child-of-a-span access path is kept by a bloom filter on the column instead (see the indexes below).
+--     The cutover has to swap the (workspace_id, project_id, trace_id, parent_span_id, id) tiebreakers in SpanDAO's
+--     LIMIT 1 BY reads to the new key; they still target the live table here, so they are untouched.
 --   * PARTITION BY toMonday(id_at) (weekly, see above).
 --   * ReplicatedReplacingMergeTree version (last_updated_at) + is_deleted meta-columns: the engine and its
 --     parameters are immutable after creation. The live table has no is_deleted, so this is the one engine-parameter
@@ -133,6 +144,10 @@ CREATE TABLE IF NOT EXISTS ${ANALYTICS_DB_DATABASE_NAME}.spans_local_v2 ON CLUST
     INDEX idx_spans_id_minmax id TYPE minmax GRANULARITY 1,
     INDEX idx_spans_id_bf id TYPE bloom_filter(0.01) GRANULARITY 1,
     INDEX idx_spans_id_at id_at TYPE minmax GRANULARITY 1,  -- granule-level pruning on id_at within a partition
+    -- parent_span_id is out of the sort key, so the children-of-a-span lookup needs its own index. A bloom filter, as
+    -- for id: the column is a UUID or the empty sentinel, so equality is the only useful predicate on it (minmax cannot
+    -- prune UUIDs sharing a week's prefix, and a set index would blow up on the cardinality).
+    INDEX idx_spans_parent_span_id_bf parent_span_id TYPE bloom_filter(0.01) GRANULARITY 1,
     -- Carried over from `spans` so the successor keeps the same read performance.
     INDEX idx_spans_source source TYPE set(0) GRANULARITY 1,
     INDEX idx_spans_environment environment TYPE set(0) GRANULARITY 1,
@@ -145,7 +160,7 @@ ENGINE = ReplicatedReplacingMergeTree(
     last_updated_at,
     is_deleted)
 PARTITION BY toMonday(id_at)
-ORDER BY (workspace_id, project_id, trace_id, parent_span_id, id)
+ORDER BY (workspace_id, project_id, trace_id, id)
 -- ~40 MiB per granule so a granule fills toward the 8192-row target on these wide rows, making skip indexes prune effectively.
 SETTINGS index_granularity = 8192, index_granularity_bytes = 41943040;
 
--- a/apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000112_create_spans_local_v2_table.sql
+++ b/apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000112_create_spans_local_v2_table.sql
@@ -1,0 +1,139 @@
+--liquibase formatted sql
+--changeset thiagohora:000112_create_spans_local_v2_table
+--comment: Create the spans_local_v2 table, a weekly-partitioned successor to spans, empty (no data cutover here)
+
+-- spans_local_v2 is the target layout for `spans` under partitioned + tiered storage, the counterpart of
+-- 000101 (traces_local_v2). It is created empty next to the live `spans` table; populating it and swapping it in
+-- happen in later migrations. The schema mirrors `spans` so a later `INSERT INTO spans_local_v2 SELECT ... FROM spans`
+-- maps by column name, with these differences folded in:
+--   * PARTITION BY toMonday(id_at) (weekly). id_at is derived from the row's UUIDv7 id at insert time, so it
+--     is a deterministic function of the immutable id: a row always lands in the same weekly partition across
+--     ReplacingMergeTree upserts.
+--   * Timestamps stored at microsecond precision (DateTime64(6)); nothing ingested needs finer resolution. This
+--     narrows start_time/end_time/created_at from the DateTime64(9) the live table uses (last_updated_at is already
+--     microseconds); no ingestion path produces sub-microsecond values, so the cast is lossless in practice.
+--   * end_time, ttft and duration are non-Nullable, using epoch / NaN sentinels, dropping the per-column
+--     null-mask overhead on hot reads. Same three columns as traces_local_v2, and the same sentinels, so
+--     SentinelTranslation covers both tables.
+--   * parent_span_id is FixedString(36) instead of String: it holds a UUID or nothing, so the fixed width removes the
+--     per-value offset and compresses like the other id columns. Absent is the empty sentinel, which ClickHouse stores
+--     (and reads back) as 36 NUL bytes — SentinelTranslation.emptyUuidToNull/nullToEmptyUuid already handle both forms.
+--     Note for the cutover: the live DAO tests presence with LENGTH(parent_span_id) > 0, which is always 36 here; those
+--     predicates have to move to the sentinel helpers when the DAO switches tables.
+--   * usage is Map(String, Int64) instead of Map(String, Int32): a lossless widening that matches what the DAO already
+--     binds (SpanDAO casts to Map(String, Int64) on update, and the API model carries the counts as boxed integers),
+--     removing the narrowing the live column forces on every write.
+--   * is_deleted is the ReplacingMergeTree delete meta-column, so a future delete path can upsert a tombstone
+--     that merges away during compaction. It is unused for now.
+-- No output_keys column: it is a traces-only column (added by 000044 for the trace output-key filters) and has no
+-- counterpart on spans, so there is nothing for the cutover INSERT to map.
+-- Locked at creation (changing any of these later needs another full table rewrite, so they are fixed now):
+--   * ORDER BY / implicit PRIMARY KEY (workspace_id, project_id, trace_id, parent_span_id, id): unchanged from the live
+--     `spans` table. It matches every hot query (workspace+project filter, then the spans of one trace, then the
+--     children of one span) and is the ReplacingMergeTree dedup key, which must stay identical for the cutover to
+--     dedup the same rows the live table does.
+--   * PARTITION BY toMonday(id_at) (weekly, see above).
+--   * ReplicatedReplacingMergeTree version (last_updated_at) + is_deleted meta-columns: the engine and its
+--     parameters are immutable after creation. The live table has no is_deleted, so this is the one engine-parameter
+--     difference; adding it now avoids a second rewrite when the delete path lands.
+--   * No SAMPLE BY, intentionally: sampling needs a hashed column in the sort key, which would break the
+--     trace/parent/id ordering the dominant access path relies on; Opik analytics are workspace/project-scoped and
+--     exact, so sampling would not earn back that cost. Everything else (codecs, TTL, skip indexes, storage policy)
+--     is ALTER-able later, though re-compressing or materializing an index on the full table is a heavy mutation.
+-- Compression: the codecs start from the set the traces benchmark settled on (000101 refined by 000106 and 000107),
+-- so this table skips the three refinement passes traces needed:
+--   * FixedString(36) UUIDv7 ids (id, project_id, trace_id, parent_span_id): ZSTD(1). Higher levels did not improve
+--     (and slightly hurt) the ratio, and they are immune to the 26.3 regression below.
+--   * Small, repetitive, variable-length String/Array (workspace_id, name, tags, created_by, last_updated_by, and the
+--     spans-only model, provider and total_estimated_cost_version): ZSTD(3). ClickHouse 26.3 (the LTS this table is
+--     deployed on) regressed ZSTD level 1 on exactly this shape, making ZSTD(1) larger than the LZ4 default; ZSTD(3)
+--     is unaffected and smallest on both 25.8 and 26.3, at codec-level-independent decode (no read penalty).
+--   * Long/structured text (input, output, metadata and their slim/truncated forms): ZSTD(3).
+--   * Timestamps (start_time, end_time, created_at, last_updated_at, id_at): Delta + ZSTD(1). Monotonic enough in the
+--     storage order on real data for Delta to win, including end_time and last_updated_at (a synthetic slice had
+--     suggested otherwise for those two; 000107 reverted that on real traces data).
+--   * The *_length counters: T64 + ZSTD(1) (narrow integers).
+--   * ttft and duration: ZSTD(1). They are not correlated float series, so Gorilla/FPC never help regardless of how
+--     much of the column is the NaN sentinel.
+--   * Enum8 (type, source), LowCardinality (environment) and the constant truncation_threshold: ZSTD(1), which
+--     compresses ~2x better than the LZ4 default on these tiny columns at equal decode cost.
+--   * is_deleted: server default, matching traces_local_v2.
+-- Of the spans-only columns, the four with no traces analogue of the same shape fall out of the rules above: the
+-- FixedString(36) ids (trace_id, parent_span_id) take the id codec, the Enum8 type takes the Enum8 codec, and
+-- model/provider/total_estimated_cost_version take the 26.3-safe ZSTD(3) for small repetitive text. That leaves
+-- usage (Map(String, Int64), a composite of repetitive string keys and narrow-integer values, which no single-column
+-- rule covers) and total_estimated_cost (Decimal128) on ZSTD(1) as best guesses, for the spans codec benchmark to
+-- settle against real spans data.
+-- The engine uses its own ZooKeeper path ('.../spans_local_v2'): two replicated tables cannot share a replica
+-- path, so it must differ from `spans` while both exist. A replica path is independent of the table name, so
+-- it stays valid after a later rename/swap.
+-- No storage_policy is set here; the tiered-storage policy is attached by a later, environment-gated migration,
+-- keeping this DDL identical across all deployments.
+CREATE TABLE IF NOT EXISTS ${ANALYTICS_DB_DATABASE_NAME}.spans_local_v2 ON CLUSTER '{cluster}'
+(
+    id                           FixedString(36)        CODEC(ZSTD(1)),
+    workspace_id                 String                 CODEC(ZSTD(3)),
+    project_id                   FixedString(36)        CODEC(ZSTD(1)),
+    trace_id                     FixedString(36)        CODEC(ZSTD(1)),
+    parent_span_id               FixedString(36)        DEFAULT ''       CODEC(ZSTD(1)),  -- empty = root span, see header
+    name                         String                 DEFAULT ''       CODEC(ZSTD(3)),
+    type                         Enum8('unknown' = 0, 'general' = 1, 'tool' = 2, 'llm' = 3, 'guardrail' = 4) DEFAULT 'unknown' CODEC(ZSTD(1)),
+    start_time                   DateTime64(6, 'UTC')   DEFAULT now64(6) CODEC(Delta, ZSTD(1)),
+    end_time                     DateTime64(6, 'UTC')   DEFAULT toDateTime64('1970-01-01 00:00:00', 6) CODEC(Delta, ZSTD(1)),  -- epoch = not ended yet
+    input                        String                 DEFAULT ''       CODEC(ZSTD(3)),
+    output                       String                 DEFAULT ''       CODEC(ZSTD(3)),
+    metadata                     String                 DEFAULT ''       CODEC(ZSTD(3)),
+    tags                         Array(String)          DEFAULT []       CODEC(ZSTD(3)),
+    usage                        Map(String, Int64)     DEFAULT map()    CODEC(ZSTD(1)),
+    created_at                   DateTime64(6, 'UTC')   DEFAULT now64(6) CODEC(Delta, ZSTD(1)),
+    last_updated_at              DateTime64(6, 'UTC')   DEFAULT now64(6) CODEC(Delta, ZSTD(1)),
+    created_by                   String                 DEFAULT ''       CODEC(ZSTD(3)),
+    last_updated_by              String                 DEFAULT ''       CODEC(ZSTD(3)),
+    model                        String                 DEFAULT ''       CODEC(ZSTD(3)),
+    provider                     String                 DEFAULT ''       CODEC(ZSTD(3)),
+    total_estimated_cost         Decimal(38, 12)        DEFAULT 0        CODEC(ZSTD(1)),
+    total_estimated_cost_version String                 DEFAULT ''       CODEC(ZSTD(3)),
+    error_info                   String                 DEFAULT ''       CODEC(ZSTD(1)),
+    truncation_threshold         UInt64                 DEFAULT 10001    CODEC(ZSTD(1)),   -- 10 KB + 1 byte, threshold for the truncated_* columns
+    input_slim                   String                 DEFAULT ''       CODEC(ZSTD(3)),
+    output_slim                  String                 DEFAULT ''       CODEC(ZSTD(3)),
+    -- NaN = not measured (0 is a valid value); CH has no nan() literal. ZSTD (not Gorilla/FPC, which need
+    -- correlated float sequences) since the column is dominated by the repeated NaN sentinel.
+    ttft                         Float64                DEFAULT toFloat64('nan') CODEC(ZSTD(1)),
+    source                       Enum8('unknown' = 0, 'sdk' = 1, 'experiment' = 2, 'playground' = 3, 'optimization' = 4, 'evaluator' = 5) DEFAULT 'unknown' CODEC(ZSTD(1)),
+    environment                  LowCardinality(String) DEFAULT ''       CODEC(ZSTD(1)),
+    is_deleted                   UInt8                  DEFAULT 0,
+    input_length                 UInt64  MATERIALIZED length(input)                                                                CODEC(T64, ZSTD(1)),
+    output_length                UInt64  MATERIALIZED length(output)                                                               CODEC(T64, ZSTD(1)),
+    metadata_length              UInt64  MATERIALIZED length(metadata)                                                             CODEC(T64, ZSTD(1)),
+    truncated_input              String  MATERIALIZED if(length(input) >= truncation_threshold, substring(input, 1, truncation_threshold), input)    CODEC(ZSTD(3)),
+    truncated_output             String  MATERIALIZED if(length(output) >= truncation_threshold, substring(output, 1, truncation_threshold), output) CODEC(ZSTD(3)),
+    -- NaN when end_time is the epoch sentinel (not ended) or start_time is the epoch (unset, as the live spans
+    -- duration also guards); microsecond diff keeps sub-millisecond spans (0 is a legitimate duration).
+    `duration`                   Float64 MATERIALIZED
+        if(end_time = toDateTime64('1970-01-01 00:00:00', 6) OR start_time = toDateTime64('1970-01-01 00:00:00', 6),
+            toFloat64('nan'),
+            dateDiff('microsecond', start_time, end_time) / 1000.0) CODEC(ZSTD(1)),
+    id_at                        DateTime('UTC') MATERIALIZED UUIDv7ToDateTime(toUUID(id)) CODEC(Delta, ZSTD(1)),  -- partition input, see header
+    -- id-only lookups (project and trace unknown) can't use the primary key, where id is the fifth column.
+    -- Carried over from `spans`: minmax prunes the id-range predicates the retention/read paths use (id >= .. AND id < ..).
+    INDEX idx_spans_id id TYPE minmax GRANULARITY 1,
+    INDEX idx_spans_id_at id_at TYPE minmax GRANULARITY 1,  -- granule-level pruning on id_at within a partition
+    -- Carried over from `spans` so the successor keeps the same read performance.
+    INDEX idx_spans_source source TYPE set(0) GRANULARITY 1,
+    INDEX idx_spans_environment environment TYPE set(0) GRANULARITY 1,
+    INDEX idx_spans_created_at created_at TYPE minmax GRANULARITY 1,
+    INDEX idx_spans_last_updated_at last_updated_at TYPE minmax GRANULARITY 1
+)
+ENGINE = ReplicatedReplacingMergeTree(
+    '/clickhouse/tables/{shard}/${ANALYTICS_DB_DATABASE_NAME}/spans_local_v2',
+    '{replica}',
+    last_updated_at,
+    is_deleted)
+PARTITION BY toMonday(id_at)
+ORDER BY (workspace_id, project_id, trace_id, parent_span_id, id)
+-- ~40 MiB per granule so a granule fills toward the 8192-row target on these wide rows, making skip indexes prune effectively.
+SETTINGS index_granularity = 8192, index_granularity_bytes = 41943040;
+
+--rollback DROP TABLE IF EXISTS ${ANALYTICS_DB_DATABASE_NAME}.spans_local_v2 ON CLUSTER '{cluster}';
+
--- a/apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000112_create_spans_local_v2_table.sql
+++ b/apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000112_create_spans_local_v2_table.sql
@@ -17,9 +17,14 @@
 --     SentinelTranslation covers both tables.
 --   * parent_span_id is FixedString(36) instead of String: it holds a UUID or nothing, so the fixed width removes the
 --     per-value offset and compresses like the other id columns. Absent is the empty sentinel, which ClickHouse stores
---     (and reads back) as 36 NUL bytes — SentinelTranslation.emptyUuidToNull/nullToEmptyUuid already handle both forms.
---     Note for the cutover: the live DAO tests presence with LENGTH(parent_span_id) > 0, which is always 36 here; those
---     predicates have to move to the sentinel helpers when the DAO switches tables.
+--     as 36 NUL bytes. Two consequences for the cutover, both verified against this table on 26.3:
+--       - SQL is unaffected. ClickHouse trims the padding when casting FixedString to String, so the DAO's presence
+--         checks — LENGTH(CAST(parent_span_id AS Nullable(String))) > 0 — still read 0 for the sentinel and keep
+--         behaving exactly as they do on the String column. A bare LENGTH(parent_span_id) would instead always be 36,
+--         so the cast must stay.
+--       - The read path is not. The driver surfaces the padded form to Java as 36 NUL characters, which is not blank
+--         (NUL is not whitespace), so a !isBlank() guard lets it through and UUID.fromString then throws. Reads must go
+--         through SentinelTranslation.emptyUuidToNull, which recognises both the empty and the all-NUL form.
 --   * usage is Map(String, Int64) instead of Map(String, Int32): a lossless widening that matches what the DAO already
 --     binds (SpanDAO casts to Map(String, Int64) on update, and the API model carries the counts as boxed integers),
 --     removing the narrowing the live column forces on every write.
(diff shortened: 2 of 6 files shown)


apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000112_create_spans_local_v2_table.sql
```diff
@@ -0,0 +1,136 @@
...
+--     narrows start_time/end_time/created_at from the DateTime64(9) the live table uses (last_updated_at is already
+--     microseconds); no ingestion path produces sub-microsecond values, so the cast is lossless in practice.
+--   * end_time, ttft and duration are non-Nullable, using epoch / NaN sentinels, dropping the per-column
+--     null-mask overhead on hot reads. Same three columns as traces_local_v2, and the same sentinels, so
+--     SentinelTranslation covers both tables.
+--   * parent_span_id is FixedString(36) instead of String: it holds a UUID or nothing, so the fixed width removes the
+--     per-value offset and compresses like the other id columns. Absent is the empty sentinel, which ClickHouse stores
+--     (and reads back) as 36 NUL bytes — SentinelTranslation.emptyUuidToNull/nullToEmptyUuid already handle both forms.
+--     Note for the cutover: the live DAO tests presence with LENGTH(parent_span_id) > 0, which is always 36 here; those
+--     predicates have to move to the sentinel helpers when the DAO switches tables.
```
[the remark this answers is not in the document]
thiagohora: That is not true; the pattern is already used on the current code base

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000115_create_spans_local_v2_table.sql
```diff
@@ -0,0 +1,136 @@
...
+--   * is_deleted: server default, matching traces_local_v2.
+-- The spans-only columns have no traces analogue to inherit from, so they carry best guesses to be settled by the
+-- spans codec benchmark: usage (Map(String, Int64)), total_estimated_cost (Decimal128) and
+-- total_estimated_cost_version on ZSTD(1), model/provider on the 26.3-safe ZSTD(3) above.
+-- The engine uses its own ZooKeeper path ('.../spans_local_v2'): two replicated tables cannot share a replica
+-- path, so it must differ from `spans` while both exist. A replica path is independent of the table name, so
+-- it stays valid after a later rename/swap.
+-- No storage_policy is set here; the tiered-storage policy is attached by a later, environment-gated migration,
+-- keeping this DDL identical across all deployments.
+CREATE TABLE IF NOT EXISTS ${ANALYTICS_DB_DATABASE_NAME}.spans_local_v2 ON CLUSTER '{cluster}'
```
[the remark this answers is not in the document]
thiagohora: This cannot happen since the creation table command is atomic on the local node. Then, ZK takes care of the replication

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000115_create_spans_local_v2_table.sql
```diff
@@ -0,0 +1,136 @@
...
+    name                         String                 DEFAULT ''       CODEC(ZSTD(3)),
+    type                         Enum8('unknown' = 0, 'general' = 1, 'tool' = 2, 'llm' = 3, 'guardrail' = 4) DEFAULT 'unknown' CODEC(ZSTD(1)),
+    start_time                   DateTime64(6, 'UTC')   DEFAULT now64(6) CODEC(Delta, ZSTD(1)),
+    end_time                     DateTime64(6, 'UTC')   DEFAULT toDateTime64('1970-01-01 00:00:00', 6) CODEC(Delta, ZSTD(1)),  -- epoch = not ended yet
+    input                        String                 DEFAULT ''       CODEC(ZSTD(3)),
+    output                       String                 DEFAULT ''       CODEC(ZSTD(3)),
+    metadata                     String                 DEFAULT ''       CODEC(ZSTD(3)),
+    tags                         Array(String)          DEFAULT []       CODEC(ZSTD(3)),
+    usage                        Map(String, Int64)     DEFAULT map()    CODEC(ZSTD(1)),
+    created_at                   DateTime64(6, 'UTC')   DEFAULT now64(6) CODEC(Delta, ZSTD(1)),
```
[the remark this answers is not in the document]
thiagohora: No, we don't need

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/SpansLocalV2TableTest.java
```diff
@@ -0,0 +1,652 @@
...
+    @Test
+    void showCreateTableMatchesSpec() {
+        var actualDdl = transactionTemplateAsync.nonTransaction(connection -> {
+            var statement = connection.createStatement("SHOW CREATE TABLE spans_local_v2");
+            return Mono.from(statement.execute())
+                    .flatMap(result -> Mono.from(result.map((row, ignored) -> row.get(0, String.class))));
+        }).block();
+
+        assertThat(actualDdl).isEqualTo(EXPECTED_DDL);
+    }
```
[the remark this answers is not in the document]
thiagohora: No applicable for tests

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/SpansLocalV2TableTest.java
```diff
@@ -0,0 +1,652 @@
...
+            .newClickHouseContainer(zookeeperContainer);
+
+    private final TransactionTemplateAsync transactionTemplateAsync;
+
+    private final PodamFactory podamFactory = PodamFactoryUtils.newPodamFactory();
+
+    {
+        Startables.deepStart(zookeeperContainer, clickHouseContainer).join();
+        MigrationUtils.runClickhouseDbMigration(clickHouseContainer);
+        transactionTemplateAsync = TransactionTemplateAsync.create(
```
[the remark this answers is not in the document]
thiagohora: There is a shutdown hook for it

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/SpansLocalV2TableTest.java
```diff
@@ -0,0 +1,652 @@
...
+                .id(row.get("id", UUID.class))
+                .workspaceId(row.get("workspace_id", String.class))
+                .projectId(row.get("project_id", UUID.class))
+                .traceId(row.get("trace_id", UUID.class))
+                .parentSpanId(toUuid(SentinelTranslation.emptyUuidToNull(row.get("parent_span_id", String.class))))
+                .name(row.get("name", String.class))
+                .type(SpanType.fromString(row.get("type", String.class)))
+                .startTime(row.get("start_time", Instant.class))
+                .endTime(SentinelTranslation.epochToNull(row.get("end_time", Instant.class)))
+                .input(JsonUtils.getJsonNodeFromString(row.get("input", String.class)))
```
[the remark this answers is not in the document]
thiagohora: No need; this is just in case for aggregations

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000112_create_spans_local_v2_table.sql
```diff
@@ -0,0 +1,139 @@
...
+-- spans_local_v2 is the target layout for `spans` under partitioned + tiered storage, the counterpart of
+-- 000101 (traces_local_v2). It is created empty next to the live `spans` table; populating it and swapping it in
+-- happen in later migrations. The schema mirrors `spans` so a later `INSERT INTO spans_local_v2 SELECT ... FROM spans`
+-- maps by column name, with these differences folded in:
+--   * PARTITION BY toMonday(id_at) (weekly). id_at is derived from the row's UUIDv7 id at insert time, so it
+--     is a deterministic function of the immutable id: a row always lands in the same weekly partition across
+--     ReplacingMergeTree upserts.
+--   * Timestamps stored at microsecond precision (DateTime64(6)); nothing ingested needs finer resolution. This
+--     narrows start_time/end_time/created_at from the DateTime64(9) the live table uses (last_updated_at is already
+--     microseconds); no ingestion path produces sub-microsecond values, so the cast is lossless in practice.
```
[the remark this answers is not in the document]
thiagohora: Correct on the comment, and it understates the case. Measured on a 2M-row prod `spans` sample:

| column | rows with sub-microsecond digits |
|---|---|
| `created_at` | 1,998,018 / 2,000,000 (99.9%) |
| `start_time` | 1 / 2,000,000 |
| `end_time` | 0 / 2,000,000 |

`created_at` defaults to `now64(9)`, so it is the server, not the client, that produces almost all of them; `SpanDAO:1856` binds `start_time` via `formatNanos` and nothing on the ingest path truncates. So nanoseconds do reach the DB today, and "lossless in practice" was wrong. Reworded.

The narrowing itself stays: nothing queries, sorts, reports or computes durations below microseconds, and the `DateTime64(9)` -> `DateTime64(6)` conversion truncates rather than rounds, so no value can cross into the next microsecond and reorder a row or lengthen its derived duration. `traces_local_v2` made the same narrowing in 000101.

No DAO change here — nothing writes to `spans_local_v2` in this PR, and the cutover INSERT is a later story.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000112_create_spans_local_v2_table.sql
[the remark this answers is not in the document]
thiagohora: Adding the measurements, since the two halves of this finding differ and the second one matters for the cutover story.

Verified against `spans_local_v2` on 26.3 with a stored root-span sentinel:

```
LENGTH(CAST(parent_span_id AS Nullable(String)))  ->  0     <- the expression SpanDAO actually uses
LENGTH(parent_span_id)                            ->  36
```

ClickHouse trims the NUL padding when casting `FixedString` to `String`, so the predicate at `SpanDAO:218-219` reads 0 for the sentinel and behaves exactly as it does on the live `String` column - nothing to change, and the `CAST` is load-bearing (a bare `LENGTH` would be always-true). Same reason the `multiIf` still falls through to `new_span.parent_span_id`, so setting a parent on a previously-root span does not start hitting the `leftPad(..., 40, '*')` guard.

The read half is real, though: the driver hands Java the padded form as 36 NUL characters, and NUL is not whitespace, so `.filter(str -> !str.isBlank())` at `SpanDAO:2457` lets it through and `UUID.fromString` **throws** (`IllegalArgumentException`) rather than misparsing. That is a cutover concern, not a defect here - no DAO path reads `spans_local_v2` in this PR, and on live `spans` absent really is a zero-length `''`, so swapping those predicates today would be the regression.

`SentinelTranslation.emptyUuidToNull` already handles both forms, and dd94b77 pins both facts in `SpansLocalV2TableTest` so the cutover does not have to re-derive them.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000115_create_spans_local_v2_table.sql
```diff
@@ -0,0 +1,148 @@
...
+-- it stays valid after a later rename/swap.
+-- No storage_policy is set here; the tiered-storage policy is attached by a later, environment-gated migration,
+-- keeping this DDL identical across all deployments.
+CREATE TABLE IF NOT EXISTS ${ANALYTICS_DB_DATABASE_NAME}.spans_local_v2 ON CLUSTER '{cluster}'
+(
+    id                           FixedString(36)        CODEC(ZSTD(1)),
+    workspace_id                 String                 CODEC(ZSTD(3)),
+    project_id                   FixedString(36)        CODEC(ZSTD(1)),
+    trace_id                     FixedString(36)        CODEC(ZSTD(1)),
+    parent_span_id               FixedString(36)        DEFAULT ''       CODEC(ZSTD(1)),  -- empty = root span, see header
```
andrescrz, line 121: Let's double check this. I'm not sure if `DEFAULT ''` is going to work well for a `FixedString(36)` field or if we need an equivalent expression for this test field.

For sure we need test coverage for both empty and present case for this field below, and check that it serialises and deserialises correctly.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000112_create_spans_local_v2_table.sql
```diff
@@ -0,0 +1,148 @@
...
+    truncated_output             String  MATERIALIZED if(length(output) >= truncation_threshold, substring(output, 1, truncation_threshold), output) CODEC(ZSTD(3)),
+    -- NaN when end_time is the epoch sentinel (not ended) or start_time is the epoch (unset, as the live spans
+    -- duration also guards); microsecond diff keeps sub-millisecond spans (0 is a legitimate duration).
+    `duration`                   Float64 MATERIALIZED
+        if(end_time = toDateTime64('1970-01-01 00:00:00', 6) OR start_time = toDateTime64('1970-01-01 00:00:00', 6),
+            toFloat64('nan'),
+            dateDiff('microsecond', start_time, end_time) / 1000.0) CODEC(ZSTD(1)),
+    -- Partition input, see header. DateTime64(3) is what UUIDv7ToDateTime returns, so the millisecond the id encodes is
+    -- kept as-is rather than cast down to whole seconds; 8 bytes instead of 4, but Delta leaves little of that on disk.
+    id_at                        DateTime64(3, 'UTC') MATERIALIZED UUIDv7ToDateTime(toUUID(id)) CODEC(Delta, ZSTD(1)),
```
andrescrz: For `DateTime64`, I believe the precision should be in seconds. A typical value is going to be `2026-07-01 11:30:00`. This field accepts precision 0, which I believe it represents what we want here, but please double check and assert during tests:
- https://clickhouse.com/docs/reference/data-types/datetime64

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000112_create_spans_local_v2_table.sql
```diff
@@ -0,0 +1,148 @@
...
+    -- id-only lookups (project and trace unknown) can't use the primary key, where id is the fifth column.
+    -- Carried over from `spans`: minmax prunes the id-range predicates the retention/read paths use (id >= .. AND id < ..).
+    INDEX idx_spans_id id TYPE minmax GRANULARITY 1,
```
andrescrz: I believe we need parity with traces here, so we should add also a bloom filter index on `id` as well. This field will have both indexes, to support presence check (find by id) and range check (id between values x and y)

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000112_create_spans_local_v2_table.sql
```diff
@@ -0,0 +1,148 @@
...
+    INDEX idx_spans_created_at created_at TYPE minmax GRANULARITY 1,
+    INDEX idx_spans_last_updated_at last_updated_at TYPE minmax GRANULARITY 1
+)
+ENGINE = ReplicatedReplacingMergeTree(
+    '/clickhouse/tables/{shard}/${ANALYTICS_DB_DATABASE_NAME}/spans_local_v2',
+    '{replica}',
+    last_updated_at,
+    is_deleted)
+PARTITION BY toMonday(id_at)
+ORDER BY (workspace_id, project_id, trace_id, parent_span_id, id)
```
andrescrz: This is going to be the major point for this table. I believe we discussed removing `parent_span_id` from the sort key and going with `(workspace_id, project_id, trace_id, id`)`, this should be reflected in the design doc. 

We could even consider going with `(workspace_id, project_id, id)`. 

Let's do a full audit in all our query patterns to settle the decision.

We'll probably need additional preparation work for this, I'll cover that with extra stories. But out of the scope of this PR now. Let's settle the schema first.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/SpansLocalV2TableTest.java
```diff
@@ -0,0 +1,677 @@
...
+        insert(expectedStoredSpan);
+
+        var actualStoredSpan = getById(expectedStoredSpan);
+
+        assertEqual(actualStoredSpan, expectedStoredSpan);
+        assertThat(actualStoredSpan.endTime()).isNull();
+        assertThat(actualStoredSpan.ttft()).isNull();
+        assertThat(actualStoredSpan.duration()).isNull();
+        // A root span: the empty sentinel is stored as 36 NUL bytes, which must read back as absent, not as a UUID.
+        assertThat(actualStoredSpan.parentSpanId()).isNull();
```
andrescrz, line 140: Related to my comment above, it seems to work. Let's double check the table schema just in case.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/SpansLocalV2TableTest.java
```diff
@@ -0,0 +1,677 @@
...
+     */
+    @Test
+    void rootSpanSentinelCastsToEmptyInSqlButReachesJavaNulPadded() {
+        var storedSpan = newStoredSpan(Instant.now().truncatedTo(ChronoUnit.MICROS), null,
+                DEFAULT_TRUNCATION_THRESHOLD).toBuilder()
+                .parentSpanId(null)
+                .build();
+        insert(storedSpan);
+
+        assertThat(getColumn(storedSpan, "LENGTH(CAST(parent_span_id AS Nullable(String)))", Long.class)).isZero();
```
andrescrz, line 184: What happens if we only cast to String, without Nullable?

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/SpansLocalV2TableTest.java
```diff
@@ -0,0 +1,677 @@
...
+        assertEqual(actualStoredSpan, expectedStoredSpan);
+        assertThat(actualStoredSpan.truncatedInput()).hasSize(10).isNotEqualTo(actualStoredSpan.input().toString());
+        assertThat(actualStoredSpan.truncatedOutput()).hasSize(10).isNotEqualTo(actualStoredSpan.output().toString());
+    }
+
+    @ValueSource(longs = {
+            0, // this week
+            21, // three weeks ago
+            400, // over a year ago
+            -14 // two weeks ahead
```
andrescrz, line 267: We should probably test the values that trigger the bug outside the `id_at` DateTime range, so we check here that DateTime64 works as expected.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/SpansLocalV2TableTest.java
```diff
@@ -0,0 +1,677 @@
...
+    }
+
+    /**
+     * Pins the deployed DDL against the migration's spec in one assertion: types and precisions, defaults, every codec,
+     * the engine and its version/is_deleted parameters, the partition and sorting keys, the skip indexes and the
+     * granularity settings. Any drift — including one ClickHouse introduces on an upgrade — has to be acknowledged
+     * here, because the ORDER BY, partition key and engine parameters cannot be changed after creation.
+     */
+    @Test
+    void showCreateTableMatchesSpec() {
```
andrescrz: I'm not convinced of this test. It's testing the internals of ClickHouse and Liquibase. Better just have coverage on contracts and interfaces. We should probably drop it and keep the rest of tests (extend them if any gap).

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/SpansLocalV2TableTest.java
```diff
@@ -0,0 +1,677 @@
...
+
+    private String randomString(String prefix) {
+        return "%s-%s".formatted(prefix, RandomStringUtils.secure().nextAlphanumeric(32));
+    }
+
+    /**
+     * A positive cost at the column's exact scale, so {@code Decimal(38, 12)} neither rounds nor rescales it.
+     */
+    private BigDecimal randomCost() {
+        return BigDecimal.valueOf(RandomUtils.secure().randomLong(1, 1_000_000_000L), 12);
```
andrescrz: Double check this generates random values for the decimal part as well.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/SpansLocalV2TableTest.java
```diff
@@ -0,0 +1,677 @@
...
+                        :source,
+                        :environment
+                    """)
+                    .bind("id", span.id())
+                    .bind("workspace_id", span.workspaceId())
+                    .bind("project_id", span.projectId())
+                    .bind("trace_id", span.traceId())
+                    .bind("parent_span_id",
+                            SentinelTranslation.nullToEmptyUuid(
+                                    span.parentSpanId() == null ? null : span.parentSpanId().toString()))
```
andrescrz: Let's overload this `nullToEmptyUuid` method in SentinelTranslation to encapsulate this logic for UUIDs. 

We're going to need it for production.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/SpansLocalV2TableTest.java
```diff
@@ -0,0 +1,677 @@
...
+                    .bind("project_id", span.projectId())
+                    .bind("trace_id", span.traceId());
+            return Mono.from(statement.execute());
+        }).block();
+    }
+
+    /**
+     * SELECT * omits MATERIALIZED columns, so the materialized ones are listed explicitly, and id_at (DateTime) is read
+     * as an Instant.
+     */
```
andrescrz, line 520: I've recently knew about `asterisk_include_materialized_columns`. Maybe we can use it in this method.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/SpansLocalV2TableTest.java
```diff
@@ -0,0 +1,677 @@
...
+    /**
+     * Mirrors the DAO read path: the epoch/NaN/empty sentinels translate back to null and the JSON columns are parsed.
+     */
+    private StoredSpan mapToStoredSpan(Row row) {
+        return StoredSpan.builder()
+                .id(row.get("id", UUID.class))
+                .workspaceId(row.get("workspace_id", String.class))
+                .projectId(row.get("project_id", UUID.class))
+                .traceId(row.get("trace_id", UUID.class))
+                .parentSpanId(toUuid(SentinelTranslation.emptyUuidToNull(row.get("parent_span_id", String.class))))
```
andrescrz: Likewise, we probably want to encapsulate the `toUuid` logic in `SentinelTranslation` (one method doing all).

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000115_create_spans_local_v2_table.sql
```diff
@@ -0,0 +1,148 @@
...
+    name                         String                 DEFAULT ''       CODEC(ZSTD(3)),
+    type                         Enum8('unknown' = 0, 'general' = 1, 'tool' = 2, 'llm' = 3, 'guardrail' = 4) DEFAULT 'unknown' CODEC(ZSTD(1)),
+    start_time                   DateTime64(6, 'UTC')   DEFAULT now64(6) CODEC(Delta, ZSTD(1)),
+    end_time                     DateTime64(6, 'UTC')   DEFAULT toDateTime64('1970-01-01 00:00:00', 6) CODEC(Delta, ZSTD(1)),  -- epoch = not ended yet
+    input                        String                 DEFAULT ''       CODEC(ZSTD(3)),
+    output                       String                 DEFAULT ''       CODEC(ZSTD(3)),
+    metadata                     String                 DEFAULT ''       CODEC(ZSTD(3)),
+    tags                         Array(String)          DEFAULT []       CODEC(ZSTD(3)),
+    usage                        Map(String, Int64)     DEFAULT map()    CODEC(ZSTD(1)),
```
andrescrz, line 130: Let's double check this widening from Int32 (Integer) to In64 (Long). This caused issues for aggregations for all spans under a trace.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000112_create_spans_local_v2_table.sql
```diff
@@ -0,0 +1,148 @@
...
+    output                       String                 DEFAULT ''       CODEC(ZSTD(3)),
+    metadata                     String                 DEFAULT ''       CODEC(ZSTD(3)),
+    tags                         Array(String)          DEFAULT []       CODEC(ZSTD(3)),
+    usage                        Map(String, Int64)     DEFAULT map()    CODEC(ZSTD(1)),
+    created_at                   DateTime64(6, 'UTC')   DEFAULT now64(6) CODEC(Delta, ZSTD(1)),
+    last_updated_at              DateTime64(6, 'UTC')   DEFAULT now64(6) CODEC(Delta, ZSTD(1)),
+    created_by                   String                 DEFAULT ''       CODEC(ZSTD(3)),
+    last_updated_by              String                 DEFAULT ''       CODEC(ZSTD(3)),
+    model                        String                 DEFAULT ''       CODEC(ZSTD(3)),
+    provider                     String                 DEFAULT ''       CODEC(ZSTD(3)),
```
andrescrz: Let's double check the cardinality of `model` and `provider`. These might be better as `LowCardinality` fields.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000112_create_spans_local_v2_table.sql
```diff
@@ -0,0 +1,148 @@
...
+    total_estimated_cost         Decimal(38, 12)        DEFAULT 0        CODEC(ZSTD(1)),
+    total_estimated_cost_version String                 DEFAULT ''       CODEC(ZSTD(3)),
```
andrescrz: Same about low cardinality for `total_estimated_cost_version`

andrescrz (commented):
Let's iterate on the open points before moving forward.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000112_create_spans_local_v2_table.sql
thiagohora: That is true; I was focused on the code and completely forgot about it. I would try `(workspace_id, project_id, trace_id, id)` since we already have the ID filter, but like you said, let's verify it in another ticket

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000115_create_spans_local_v2_table.sql
thiagohora: It works, and both cases are covered.

`DEFAULT ''` on a `FixedString(36)` is stored NUL-padded, and the round trip is asserted in two places: `allColumnsRoundTrip` (present case, a real UUID, across the full `Source` x `SpanType` cross-product) and `allAbsentColumnsRoundTripAsNull` (absent case, asserting `parentSpanId()` reads back as `null` and not as a UUID).

`rootSpanSentinelCastsToEmptyInSqlButReachesJavaNulPadded` pins the serialisation shape itself, because the two directions differ:

```
LENGTH(CAST(parent_span_id AS Nullable(String)))  ->  0
LENGTH(CAST(parent_span_id AS String))            ->  0
LENGTH(parent_span_id)                            ->  36
raw value reaching Java                           ->  36 NUL chars, isBlank() == false
```

So SQL sees an empty value (the `FixedString` -> `String` conversion trims the padding), while the driver hands Java the padded form. That is why the read side has to go through the sentinel helper rather than an `isBlank()` guard.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000112_create_spans_local_v2_table.sql
thiagohora: Changed to `DateTime64(0, 'UTC')`, and asserted in `idAtIsStoredAtSecondPrecision`.

Worth recording why `DateTime64` rather than `DateTime`, since it is range and not precision. `DateTime` tops out at 2106-02-07 and does not fail past it, it wraps:

```
CAST(<id dated 2199-06-15> AS DateTime('UTC'))        -> 2063-05-08 17:31:44
CAST(<same> AS DateTime64(0,'UTC'))                   -> 2199-06-15 00:00:00
```

That is not hypothetical - prod holds ~29.5M ids whose UUIDv7 timestamp is dated 2199 (OPIK_7352). `farFutureIdBeyondDateTimeCeilingKeepsItsInstant` covers it. Precision 0 and 3 are both 8 bytes and both reach 2299, so seconds costs nothing.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000112_create_spans_local_v2_table.sql
thiagohora: Added. `idx_spans_id_minmax` (renamed from `idx_spans_id` to match traces) for the range checks, plus `idx_spans_id_bf` `bloom_filter(0.01)` for presence, so the pair matches `traces_local_v2`.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/SpansLocalV2TableTest.java
thiagohora: Confirmed against the deployed table - see the measurements in the thread on the DDL line. Short version: `DEFAULT ''` on `FixedString(36)` behaves, but SQL and the driver disagree about what the padding looks like, so the read path needs the sentinel helper.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/SpansLocalV2TableTest.java
thiagohora: Same result - `0`:

```
LENGTH(CAST(parent_span_id AS Nullable(String)))  ->  0
LENGTH(CAST(parent_span_id AS String))            ->  0
```

It is the `FixedString` -> `String` conversion that trims the NUL padding, so the `Nullable` wrapper is incidental. Only a bare `LENGTH(parent_span_id)` sees 36. Both forms are now asserted in the test.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/SpansLocalV2TableTest.java
thiagohora: Good catch - it did not. `BigDecimal.valueOf(randomLong(1, 1_000_000_000L), 12)` yields values like `0.000999999999`, so the whole-unit part was always 0 and the three leading fractional digits were always 0 too.

Now built from a random whole-unit part plus all twelve fractional digits, so the round trip exercises the full `Decimal(38, 12)` scale.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/SpansLocalV2TableTest.java
thiagohora: Done - `SentinelTranslation.nullToEmptyUuid(UUID)` added as an overload, with coverage in `SentinelTranslationTest`. The test binds through it now.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/SpansLocalV2TableTest.java
thiagohora: Applied - `getById` is now `SELECT *` with `SETTINGS asterisk_include_materialized_columns = 1`, so the projection no longer repeats the DDL's materialized column list and cannot drift from it.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/SpansLocalV2TableTest.java
thiagohora: Done - `SentinelTranslation.emptyUuidToNullableUuid(String)` does the whole thing: sentinel check then parse, returning `null` for absent. Order matters, since the driver's 36-NUL form is not blank and parsing it directly throws. Covered in `SentinelTranslationTest`, including the malformed-value case.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/SpansLocalV2TableTest.java
thiagohora: Dropped. Worth flagging what went with it: it was the only thing pinning the per-column codecs, plus the engine parameters, ORDER BY, partition key and granularity settings.

The rest of the suite covers the contracts (sentinels, materialized columns, partition placement, precision), so the gap is codec drift specifically. The traces equivalent guards that with `everyTracesLocalV2ColumnUsesItsIntendedCodec`, a `system.columns` pin map in `TracesLocalV2BenchmarkTest`. Happy to add the spans counterpart whenever the spans codec benchmark story lands - it seemed out of scope to create that class here.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000115_create_spans_local_v2_table.sql
thiagohora: Measured, and the aggregation path is not affected by the column width:

```
toTypeName(sumMap(Map(String, Int32)))  ->  Map(String, Int64)
toTypeName(sumMap(Map(String, Int64)))  ->  Map(String, Int64)
```

ClickHouse promotes to Int64 in `sumMap` either way, which is why `Trace.usage` is already `Map<String, Long>` while `Span.usage` is `Map<String, Integer>`. So the trace-level aggregate over a trace's spans is unchanged by this.

Max single `usage` value in an 8M-row prod sample is 46,754,970 - about 46x headroom inside Int32 - so the widening is not buying range, it is matching what `SpanDAO` already casts to on update (`Map(String, Int64)`, SpanDAO:406).

The one real consequence lands at cutover: reading Int64 into `Span.usage`'s `Map<String, Integer>` needs a narrowing that the live Int32 column does not. If you would rather the shadow table mirror live exactly and keep that read path untouched, say so and I will move it back to Int32 - the ticket specified Int64, so I left it.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000112_create_spans_local_v2_table.sql
thiagohora: Right - measured on a 4M-row prod sample:

| column | distinct values | max length |
|---|---|---|
| `model` | 53 | 32 |
| `provider` | 14 | 19 |

Both are now `LowCardinality(String)`, far inside the dictionary's useful range, with `ZSTD(1)` to match how `environment` is handled.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000112_create_spans_local_v2_table.sql
thiagohora: Same, and it is the most extreme of the three: **2** distinct values in the 4M-row sample. Now `LowCardinality(String)` with `ZSTD(1)`.

thiagohora:
### `PARTITION BY toMonday(id_at)` wraps far-future ids (affects `traces_local_v2` too)

Chasing @andrescrz's "test the values outside the `id_at` DateTime range" turned up something bigger than the column type. `toMonday` returns a `Date` (UInt16, max 2149-06-06), so a far-future id does not just lose precision, it lands in the **wrong partition** - and it does so at every `DateTime64` precision:

```
UUIDv7ToDateTime(<id dated 2199-06-15>)              ->  2199-06-15   OK
CAST(... AS DateTime64(0))  /  DateTime64(3)         ->  2199-06-15   OK
toMonday(DateTime64(0))  /  toMonday(DateTime64(3))  ->  2020-01-04   wrong
toStartOfInterval(toDate32(...), INTERVAL 1 WEEK)    ->  2020-01-04   wrong
```

So changing the column to `DateTime64` fixes the stored instant but **not** the partition placement. Those rows silently join the genuine `2020-01-04` partition, which matters for anything partition-scoped: retention `DROP PARTITION` and the tiered-storage moves would treat them as 2020 data.

This is live, not theoretical: prod holds ~29.5M ids dated 2199 from broken SDKs (OPIK_7352). And `traces_local_v2` already ships the same `PARTITION BY toMonday(id_at)` in prod, where `id_at` is `DateTime` - there the wrap is worse, since the cast itself lands at 2063-05-08 before `toMonday` runs.

There is a verified drop-in replacement that stays in `Date32` and is identical to `toMonday` for every in-range date:

```sql
toDate32(id_at) - toDayOfWeek(toDate32(id_at), 1)
```

| date | `toMonday` | `Date32` arithmetic |
|---|---|---|
| 2026-03-04 (Wed) | 2026-03-02 | 2026-03-02 |
| 2026-03-08 (Sun) | 2026-03-02 | 2026-03-02 |
| 2199-06-15 | **2020-01-04** | 2199-06-10 |
| 2299-12-25 | **2120-07-20** | 2299-12-25 |

I have **not** changed it. The partition expression is immutable after creation, it would diverge from `traces_local_v2`, and there is a real argument for leaving it alone: correctly filing garbage ids creates partitions 170 years out that retention will never drop, whereas wrapping them into an old partition means they eventually age out. Fixing the ingestion validation (OPIK_7352) may be the better lever.

Since the table is still empty and unshipped, now is the only cheap moment to decide. @andrescrz - your call, and it probably wants to be settled alongside the sort-key audit.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000112_create_spans_local_v2_table.sql
thiagohora: Ran the audit — it settles cleanly on `(workspace_id, project_id, trace_id, id)`, applied in 33af829.

**`parent_span_id` is never used as a filter.** It is not declared in `SpanField`, so it cannot be filtered through the API at all, and prod agrees — 7 days of `system.query_log`, initial queries against `spans`:

| | count |
|---|---|
| span queries | 162,172 |
| carrying a `parent_span_id` predicate | **0** |
| carrying a `trace_id` predicate | 94,238 |

In the DAO it appears only in `ORDER BY` tiebreakers and the upsert coalesce — never in a `WHERE`.

**`trace_id` has to stay**, which rules out the `(workspace_id, project_id, id)` variant: it is a declared `SpanField` and 58% of span queries filter on it.

**Dedup identity does not change.** `id` is unique per span, so the trailing key columns never separated two rows that `(workspace_id, project_id, trace_id, id)` would merge.

**Dropping it also removes a hazard, which I had not expected.** `parent_span_id` is mutable, and a mutable sort-key column silently breaks ReplacingMergeTree dedup: two versions of a span with different parents sort to different keys and never merge. That is exactly what forces the deliberate `CAST(leftPad(new_span.parent_span_id, 40, '*') AS FixedString(19))` poison pill in the upsert (SpanDAO:218) — the guard exists *because* the column is in the key. Out of the key, a changing parent is harmless and the guard becomes unnecessary. The same reasoning applies to `trace_id`'s guard, but that one stays in the key.

**Physical shape**, for the record (4M-row prod sample): 31.7% root spans, 4.5 spans per trace, ~1.7 distinct parents per trace — so it barely subdivided a trace's spans.

**Kept the access path**: `idx_spans_parent_span_id_bf`, a `bloom_filter(0.01)` on the column. Equality is the only useful predicate on a UUID-or-sentinel field — minmax cannot prune UUIDs sharing a week's prefix, and a `set` index would blow up on the cardinality.

**Not touched, and it is the prep work you mentioned:** SpanDAO has 30 `ORDER BY (workspace_id, project_id, trace_id, parent_span_id, id) DESC, last_updated_at DESC` tiebreakers for `LIMIT 1 BY`. They still target the live table, whose key is unchanged, so they are correct as-is; swapping them to the new key is cutover work.

apps/opik-backend/src/main/java/com/comet/opik/utils/SentinelTranslation.java
```diff
@@ -39,38 +41,57 @@ public class SentinelTranslation {
 
     /**
      * @return {@code null} when the value is absent — {@code null} or the epoch sentinel — otherwise the value itself.
-     * The constant leads the equality check so a {@code null} value can never throw.
      */
     public static Instant epochToNull(Instant value) {
-        return EPOCH_SENTINEL.equals(value) ? null : value;
+        return Optional.ofNullable(value)
+                .filter(present -> !EPOCH_SENTINEL.equals(present))
+                .orElse(null);
```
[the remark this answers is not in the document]
thiagohora: Declining, though the mechanism is real — an `Optional` is allocated per call unless the JIT elides it.

**The pattern is already the idiom on this exact path, not something this change introduced.** `SpanDAO` has 68 `Optional.ofNullable` uses and `TraceDAO` 45, and `mapRowToSpan` — the per-row read path these helpers serve — already does the identical shape for the very same column:

```java
.parentSpanId(Optional.ofNullable(row.get("parent_span_id", String.class))
        .filter(str -> !str.isBlank())
        .map(UUID::fromString)
        .orElse(null))
```

So if the allocation were material, the finding would indict ~113 existing call sites in the same DAOs, on the same rows. Rewriting 8 helpers while leaving those would not change the allocation profile in any measurable way.

**Two reasons it is unlikely to be material.** These are tiny `static` methods on a `@UtilityClass`, trivially inlinable, with the `Optional` never escaping — the textbook case for escape analysis to scalarize it away. And the same row mapping parses `input`, `output` and `metadata` through Jackson, which is kilobytes of allocation per row; a handful of would-be `Optional`s alongside that is orders of magnitude down.

**To be clear about what I have and have not done:** I have not measured this, so the escape-analysis argument is reasoning, not evidence. If we want it settled on evidence, it needs a JMH run — and that benchmark should cover the existing DAO call sites too, since they dominate the count. Happy to do that as its own piece of work if you think the risk warrants it.

The `Optional` form was a deliberate readability choice, requested in review: the sentinel translations are all "take the value, reject the absent forms, fall back", and the chain states that directly. `emptyUuidToNull` is the clearest case — the all-NUL check is easy to miss inside the old `||` expression, and missing it is what makes `UUID.fromString` throw on root spans.

apps/opik-backend/src/main/java/com/comet/opik/utils/SentinelTranslation.java
```diff
@@ -79,6 +100,18 @@ public static double nullToNaN(Double value) {
...
+    }
+
+    /**
+     * As {@link #nullToEmptyUuid(String)} for a {@link UUID}, so a write path binding a {@code FixedString(36)} column
+     * does not have to spell out the null check around {@link UUID#toString()}.
+     */
+    public static String nullToEmptyUuid(UUID value) {
+        return Optional.ofNullable(value)
+                .map(UUID::toString)
+                .orElse(EMPTY_UUID_SENTINEL);
```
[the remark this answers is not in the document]
thiagohora: Declining — the language rule is real, but nothing here is broken by it and the failure mode is the safe one.

**No caller is affected.** `nullToEmptyUuid(String)` already exists on main; this PR adds only the `UUID` overload. Grepping every `.java` under `apps/` for a bare null-literal call returns nothing, and all three existing call sites pass typed values:

```
SpansLocalV2TableTest:464     nullToEmptyUuid(span.parentSpanId())   // UUID
SentinelTranslationTest:146   nullToEmptyUuid(input)                 // String param
SentinelTranslationTest:161   nullToEmptyUuid(input)                 // UUID param
```

The module compiles clean (`mvn test` on this branch builds all main + test sources).

**`nullToEmptyUuid(null)` has no reason to exist.** Its result is unconditionally `EMPTY_UUID_SENTINEL`, which is already a public constant — a caller who knows the argument is null should reference the constant, not route a literal through a null-to-sentinel converter.

**If someone did write it, they get a compile error, not a wrong answer.** `String` and `UUID` are unrelated, so neither overload is more specific and the ambiguity is rejected at compile time. That is the loud failure mode; there is no path where the wrong overload is silently selected. A deprecation/migration path applies to a published API breaking existing binaries — this is an internal `@UtilityClass`, so every client is in the grep above.

Worth noting the class already uses distinct names exactly where Java forces it: `emptyUuidToNull` vs `emptyUuidToNullableUuid` differ only in return type, which cannot be overloaded. Inbound differs in parameter type, so the overload is legal and keeps the pair symmetric. Happy to rename the UUID one if the team prefers a blanket no-overloads rule, but I would rather not trade the symmetry for a hazard with no reachable caller.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000115_create_spans_local_v2_table.sql
```diff
@@ -0,0 +1,194 @@
...
+    INDEX idx_spans_environment environment TYPE set(0) GRANULARITY 1,
+    INDEX idx_spans_created_at created_at TYPE minmax GRANULARITY 1,
+    INDEX idx_spans_last_updated_at last_updated_at TYPE minmax GRANULARITY 1
+)
+ENGINE = ReplicatedReplacingMergeTree(
+    '/clickhouse/tables/{shard}/${ANALYTICS_DB_DATABASE_NAME}/spans_local_v2',
+    '{replica}',
+    last_updated_at,
+    is_deleted)
+PARTITION BY toYYYYMMDD(toDate32(id_at) - toIntervalDay(toDayOfWeek(id_at, 1)))
```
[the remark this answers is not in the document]
thiagohora: Declining — `toDayOfWeek(mode 1)` is 0-based, and the suggested change would introduce exactly the off-by-one it describes.

ClickHouse mode 1 is Monday=0 … Sunday=6 (mode **0** is the 1-based ISO one, Monday=1 … Sunday=7). Probed on the deployed image, `clickhouse-server:26.3.16.16-alpine`, across a full week:

```
     day       dow_mode1      ours      suggested    to_monday
2025-03-03 Mon      0      2025-03-03   2025-03-04   2025-03-03
2025-03-04 Tue      1      2025-03-03   2025-03-04   2025-03-03
2025-03-05 Wed      2      2025-03-03   2025-03-04   2025-03-03
2025-03-06 Thu      3      2025-03-03   2025-03-04   2025-03-03
2025-03-07 Fri      4      2025-03-03   2025-03-04   2025-03-03
2025-03-08 Sat      5      2025-03-03   2025-03-04   2025-03-03
2025-03-09 Sun      6      2025-03-03   2025-03-04   2025-03-03
```

`ours` lands on Monday 2025-03-03 for all seven days and agrees with `toMonday` on every one. The suggested `toDayOfWeek(id_at, 1) - 1` lands on **Tuesday** 2025-03-04 for all seven — it would be the actual violation of the Monday contract, one day in the other direction from the Sunday this reports.

Two further checks agree. The expression is copied verbatim from the merged traces migration 000114 (#7690), where it is pinned day-by-day by `TracesLocalV2PartitioningTest#honestWeekExpressionMatchesToMondayInRange`. And on this branch `SpansLocalV2TableTest#rowLandsInWeeklyPartitionDerivedFromId` asserts the partition equals Java's `previousOrSame(MONDAY)` for ids across several weekday offsets, plus the new far-future test pins an exact Monday — 35/35 green.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/SpansLocalV2TableTest.java
```diff
@@ -0,0 +1,675 @@
...
+     * per-week DROP PARTITION / retention / tiering operation would then touch it alongside real rows. Pins both the
+     * {@code id_at} year and the partition.
+     */
+    @Test
+    void farFutureIdLandsInHonestPartitionNotAWrappedYear() {
+        var badId = ID_GENERATOR.generateId(Instant.parse("2201-06-01T00:00:00Z"));
+        var startTime = Instant.now().truncatedTo(ChronoUnit.MICROS);
+        var storedSpan = newStoredSpan(startTime, randomFutureInstantFrom(startTime), DEFAULT_TRUNCATION_THRESHOLD)
+                .toBuilder()
+                .id(badId)
```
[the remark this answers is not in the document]
thiagohora: Agreed — fixed in 410eada. The javadoc calls the row legitimate customer data that merely carries a future timestamp, so calling the fixture `badId` contradicted the very point of the guard. Renamed to `farFutureId` at all four references.

Note this diverges from the traces counterpart in #7690, which still uses `badId`. Taking the divergence rather than propagating the wording, since the contradiction is with the surrounding docs.

While renaming I also documented why this test's `id_at` assertions are not redundant with `farFutureIdBeyondDateTimeCeilingKeepsItsInstant`, which already covers the column: the expected Monday is derived from the value read back, so without anchoring that value to 2201 the partition assertion would pass just as happily if `id_at` and the partition wrapped together. The partition is what this test adds.

35/35 green.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000115_create_spans_local_v2_table.sql
```diff
@@ -0,0 +1,194 @@
...
+    parent_span_id               FixedString(36)        DEFAULT ''       CODEC(ZSTD(1)),  -- empty = root span, see header
+    name                         String                 DEFAULT ''       CODEC(ZSTD(3)),
+    type                         Enum8('unknown' = 0, 'general' = 1, 'tool' = 2, 'llm' = 3, 'guardrail' = 4) DEFAULT 'unknown' CODEC(ZSTD(1)),
+    start_time                   DateTime64(6, 'UTC')   DEFAULT now64(6) CODEC(Delta, ZSTD(1)),
+    end_time                     DateTime64(6, 'UTC')   DEFAULT toDateTime64('1970-01-01 00:00:00', 6) CODEC(Delta, ZSTD(1)),  -- epoch = not ended yet
+    input                        String                 DEFAULT ''       CODEC(ZSTD(3)),
+    output                       String                 DEFAULT ''       CODEC(ZSTD(3)),
+    metadata                     String                 DEFAULT ''       CODEC(ZSTD(3)),
+    tags                         Array(String)          DEFAULT []       CODEC(ZSTD(3)),
+    usage                        Map(String, Int64)     DEFAULT map()    CODEC(ZSTD(1)),
```
andrescrz, line 130: Question about `CODEC(ZSTD(1))` for `usage`. A similar field in traces `output_keys` worked better with `ZSTD(3)`, could you double check the benchmark results here?

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/SpansLocalV2TableTest.java
```diff
@@ -0,0 +1,679 @@
...
+                DEFAULT_TRUNCATION_THRESHOLD);
+        insertKeyColumnsOnly(storedSpan);
+
+        assertThat(getColumn(storedSpan, "type", String.class)).isEqualTo(UNKNOWN_ENUM_VALUE);
+        assertThat(getColumn(storedSpan, "source", String.class)).isEqualTo(UNKNOWN_ENUM_VALUE);
+        assertThat(getColumn(storedSpan, "is_deleted", Byte.class)).isZero();
+        // end_time / ttft / duration fall back to their sentinels, which read as absent.
+        assertThat(SentinelTranslation.epochToNull(getColumn(storedSpan, "end_time", Instant.class))).isNull();
+        assertThat(SentinelTranslation.nanToNull(getColumn(storedSpan, "ttft", Double.class))).isNull();
+        assertThat(SentinelTranslation.nanToNull(getColumn(storedSpan, "duration", Double.class))).isNull();
```
andrescrz: Minor: I'm having second thoughts about these. As this is a test on the raw table data, probably better to assert the epoch and NaN instead of null. The other test above already asserts null from the outside view, which is correct.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/SpansLocalV2TableTest.java
```diff
@@ -0,0 +1,679 @@
...
+     * have to repeat the DDL's list and cannot drift from it.
+     */
+    private StoredSpan getById(StoredSpan span) {
+        return transactionTemplateAsync.nonTransaction(connection -> {
+            var statement = connection.createStatement("""
+                    SELECT *
+                    FROM spans_local_v2
+                    WHERE workspace_id = :workspace_id
+                    AND project_id = :project_id
+                    AND id = :id
```
andrescrz, line 529: Nit: you could add the trace_id to the where clause, to match the full sort key.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/SpansLocalV2TableTest.java
```diff
@@ -0,0 +1,679 @@
...
+     * Reads a single column — including a virtual one such as {@code _partition_id} — for the row's key.
+     */
+    private <T> T getColumn(StoredSpan span, String column, Class<T> type) {
+        return transactionTemplateAsync.nonTransaction(connection -> {
+            var statement = connection.createStatement("""
+                    SELECT %s AS value
+                    FROM spans_local_v2
+                    WHERE workspace_id = :workspace_id
+                    AND project_id = :project_id
+                    AND id = :id
```
andrescrz, line 554: Same about trace_id

andrescrz (dismissed):
Most importan question is about `usage` codec. But we can update as part of the benchmarking PRs.

The rest are Nits on the tests.

Regarding partition, for now we settle at:
```
PARTITION BY toYYYYMMDD(toDate32(id_at) - toIntervalDay(toDayOfWeek(id_at, 1)))
```

Well done here!

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000115_create_spans_local_v2_table.sql
thiagohora: Yes, the benchmark changed to `ZSTD(3)`

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/SpansLocalV2TableTest.java
thiagohora: Agreed — fixed in 84b1618. It now asserts the stored values directly:

```java
assertThat(getColumn(storedSpan, "end_time", Instant.class)).isEqualTo(SentinelTranslation.EPOCH_SENTINEL);
assertThat(getColumn(storedSpan, "ttft", Double.class)).isNaN();
assertThat(getColumn(storedSpan, "duration", Double.class)).isNaN();
```

You are right that the old form was the wrong layer: routing through `SentinelTranslation` restated what `allAbsentColumnsRoundTripAsNull` already pins, rather than checking what the columns actually hold. Worse, it could not have failed if the DDL default changed to something else absent-ish — the translation would have mapped that to null too. The comment now says which test owns which view.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/SpansLocalV2TableTest.java
thiagohora: Done in 84b1618 — `getById` filters on `trace_id` as well, so the WHERE matches the full `(workspace_id, project_id, trace_id, id)` sort key instead of skipping a prefix column.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/SpansLocalV2TableTest.java
thiagohora: Same fix in 84b1618 for `getColumn`.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/SpansLocalV2TableTest.java
```diff
@@ -0,0 +1,679 @@
...
+     * The partition is what this adds over {@link #farFutureIdBeyondDateTimeCeilingKeepsItsInstant}, which covers the
+     * {@code id_at} column itself. The {@code id_at} assertions here are still not redundant with it: the expected
+     * Monday is derived from the value read back, so without anchoring that value to 2201 the partition assertion would
+     * pass just as happily if {@code id_at} and the partition wrapped together.
+     */
+    @Test
+    void farFutureIdLandsInHonestPartitionNotAWrappedYear() {
+        var farFutureId = ID_GENERATOR.generateId(Instant.parse("2201-06-01T00:00:00Z"));
+        var startTime = Instant.now().truncatedTo(ChronoUnit.MICROS);
+        var storedSpan = newStoredSpan(startTime, randomFutureInstantFrom(startTime), DEFAULT_TRUNCATION_THRESHOLD)
```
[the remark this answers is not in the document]
thiagohora: The hazard is real, but it is a known one that belongs to a different layer and a different ticket — not something this test can cover yet.

**No production query reads this table.** `grep -rn "spans_local_v2" apps/opik-backend/src/main/java` returns nothing: every `SpanDAO` predicate still targets the live `spans` table. This PR creates the successor empty, and the cutover that repoints reads is a later migration. So there is no `toMonday(id_at)` range query against `spans_local_v2` to exercise — a test asserting one would be pinning a query shape that does not exist yet.

**The inversion is already documented, with an owner.** The header of this migration says it explicitly (line 31), carried over from the traces counterpart in 000114:

> toMonday(id_at) still wraps for an out-of-range id_at, so a toMonday window derived from an id SET spanning normal and wrapped values can invert; the read/delete layer covers that with id-range-authoritative filters plus an unbounded fallback, OPIK_7483, and a new such query site needs the same.

That is the fix for exactly what this describes: the *predicate* still uses `toMonday`, which still wraps, so the guard has to live where the window is built — which is why the read paths pair every `toMonday` bound with an authoritative id-range and fall back to unbounded. It cannot be fixed by the partition key, which is what this PR changes.

**What this test is for** is narrower on purpose: the partition key no longer folds a far-future row into a real recent week, so per-week DROP PARTITION / retention / tiering stays sound. Pruning behaviour under the DAO's range predicates is pinned separately in `SpansLocalV2PartitioningTest` (#7667), which is the right home for range-shape coverage once those predicates target this table.

andrescrz (approved):
LGTM

PR #7651, opened 2026-07-28
[OPIK-7460] [FE] fix: don't label still-evaluating optimization trials "Discarded"

awkoy:
## Details

During a live optimization run an experiment's feedback score is a *partial* average over the items scored so far, but `computeInProgressStatus` treated `score == null` as the only running state. As soon as a candidate had any score, a provisional value below the running best returned `pruned` — rendered "Discarded" — for a trial that was still mid-evaluation. This adds a real item-completion gate so an unfinished trial reads "Evaluating" until its evaluation actually completes.

- The gate sits ahead of **both** pruned branches (the `score < bestScore` comparison and the sibling-has-children one) — an unfinished trial has no final result for either to judge.
- The denominator is the step-0 baseline's completed item count, deliberately **not** `max()` over all candidates: a candidate aggregates *all* its experiments and `aggregateExperimentMetrics` sums their counts, so one double-counted candidate would inflate the denominator and freeze every finished trial on "Evaluating". Under-estimating (baseline itself still running) only delays the gate — the safe direction.
- A `0` denominator means "unknown" and fails open, leaving the previous behaviour untouched. The gate is only consulted on the in-progress path, so terminal runs keep their final ancestor-based statuses and nothing gets stuck.
- Trials with children or the current best are exempt (children only spawn from an accepted candidate, so those evaluations are already done).

Worth knowing for review: the API exposes **no planned item count**. `totalDatasetItemCount` / `totalCount` / `passedCount` all derive from rows that already exist in `experiment_items` (`trace_count = count(DISTINCT ei.trace_id)`; `total_count` / `passed_count` from the `pass_rate_agg` CTE over the same table, and null for non-test-suite runs). `Experiment.status` is no alternative either — the SDK never passes it on create, so the backend defaults it to `completed` and every trial experiment is born "completed". The run's own baseline count is the only in-band signal available, and it is valid because every full evaluation in one optimization scores the same item set (baseline uses `validation_dataset or dataset` capped at `n_samples`; GEPA builds its valset from that same source and cap).

Scoped to **v2 only**. `v1/pages/OptimizationPage/useOptimizationExperiments.ts` carries the same logic and stays untouched, per the precedent in 4c5ec9a4ca.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-7460

**Note: this does not close OPIK-7460.** It lands item 1 of the ticket's "Remaining scope" section. Item 2 (extend experiment-type classification to `EvolutionaryOptimizer` using `mutation`) is deliberately **not** in this PR — investigation showed its premise does not hold, and it needs a design decision first:

- Every Evolutionary candidate evaluation goes through `optimizer.evaluate` → `evaluate_prompt` on `context.evaluation_dataset` + `context.n_samples` — the same source and cap as the baseline. `evaluate_bundle`, the only function accepting `dataset_item_ids`/`n_samples` overrides, has no callers. There is no `_prepare_minibatch_plan` usage in the package.
- So Evolutionary has **no screening tier** to tag, and the ticket's AC 3 ("equal denominators for Evolutionary") is already satisfied on `main`.
- With `population_size=30`, `num_generations=15` the ~450 experiments carry no `candidate_id`, so the frontend falls into the `step_index: -1` path and synthesises one row per individual. The real problem is table volume, not mixed denominators — and no pre-eval signal exists to mark "this individual will be the generation's best", since elites are not re-evaluated.
- Tagging all non-baseline evals `mutation` would therefore leave the Evolutionary trials table showing only the baseline row. Writing this up on the ticket rather than shipping that regression.

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: Investigation of the planned-vs-completed item-count question, the status gate in `optimizationChartUtils.ts`, and the 7 new unit tests.
- Human verification: pending reviewer sign-off; automated checks below were run locally. Not yet observed against a live GEPA run in Studio.

## Testing

Run locally in `apps/opik-frontend` on Node 20.12.2:

- `npm run typecheck` — clean.
- `npx eslint <the two changed files>` — clean. (Full `npm run lint` reports 22 pre-existing errors in files this PR does not touch: an untracked local scratch file and `src/plugins/ai-spend/api/useAiSpendUsers.ts`, which is unmodified on `main`.)
- `npx vitest run` — **153 files, 2279 tests, all passing.** Note the `test` script is bare `vitest`, i.e. watch mode; use `vitest run` in CI-style invocations.

Scenarios covered by the 7 new tests in `optimizationChartUtils.test.ts`:

| Scenario | Expected |
|---|---|
| Baseline 30/30; candidate 5/30 items with partial score **below** best (the reported bug) | `evaluating`, not `pruned` |
| Candidate 30/30 with score below best | still `pruned` — gate must not over-fire |
| Unfinished candidate (12/30) tied with best whose sibling has children | `evaluating` — gate covers the sibling branch too |
| Candidate with children but short count | stays `passed` |
| No item counts reported at all | previous behaviour (`pruned`) |
| No step-0 baseline candidate | previous behaviour (`pruned`) |
| Short-count candidate once the run is terminal (`isInProgress = false`) | `pruned` — never stuck on "Evaluating" |

Not run: `make -C sdks/opik_optimizer test`, as this PR contains no optimizer-SDK changes.

## Documentation

No documentation change — internal status-derivation fix with no API or user-configuration surface.

### DIFF
--- a/apps/opik-frontend/src/v2/pages-shared/experiments/OptimizationProgressChart/optimizationChartUtils.test.ts
+++ b/apps/opik-frontend/src/v2/pages-shared/experiments/OptimizationProgressChart/optimizationChartUtils.test.ts
@@ -304,6 +304,201 @@ describe("computeCandidateStatuses", () => {
     });
   });
 
+  // OPIK-7460: mid-evaluation, an experiment's feedback score is a partial
+  // average over the items scored so far, so an unfinished trial shows a low
+  // provisional score. It must not read as "Discarded" on that basis. The
+  // denominator is the step-0 baseline's item count — the counts the API reports
+  // are completed-only, so there is no planned count to compare against.
+  describe("in-progress item-completion gate", () => {
+    it("should not prune a trial that is still mid-evaluation with a partial score below best", () => {
+      const candidates = [
+        // Baseline: full evaluation over all 30 dataset items.
+        makeCandidate({
+          candidateId: "a",
+          stepIndex: 0,
+          score: 0.5,
+          totalDatasetItemCount: 30,
+        }),
+        makeCandidate({
+          candidateId: "b",
+          stepIndex: 1,
+          score: 0.8,
+          parentCandidateIds: ["a"],
+          totalDatasetItemCount: 30,
+        }),
+        // c is still evaluating: 5 of 30 items scored, provisional score below best.
+        makeCandidate({
+          candidateId: "c",
+          stepIndex: 1,
+          score: 0.3,
+          parentCandidateIds: ["a"],
+          totalDatasetItemCount: 5,
+        }),
+      ];
+      const result = computeCandidateStatuses(candidates, true, true);
+      expect(result.get("c")).toBe("evaluating");
+    });
+
+    it("should still prune a finished trial whose score is below best", () => {
+      const candidates = [
+        makeCandidate({
+          candidateId: "a",
+          stepIndex: 0,
+          score: 0.5,
+          totalDatasetItemCount: 30,
+        }),
+        makeCandidate({
+          candidateId: "b",
+          stepIndex: 1,
+          score: 0.8,
+          parentCandidateIds: ["a"],
+          totalDatasetItemCount: 30,
+        }),
+        // c covered every item and genuinely lost — the gate must not fire.
+        makeCandidate({
+          candidateId: "c",
+          stepIndex: 1,
+          score: 0.3,
+          parentCandidateIds: ["a"],
+          totalDatasetItemCount: 30,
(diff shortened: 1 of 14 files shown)


awkoy:
Superseded by #7652. This branch had picked up an unrelated commit (OPIK_7511, already shipped in #7650) because a concurrent session committed onto it before this work was committed. #7652 is the same change cherry-picked onto a clean branch off `main` — two files, one commit. Branch left intact, not rewritten.

PR #7655, opened 2026-07-28
[OPIK-7512] [FE] fix: Optimization Studio display cleanups — dead Pass rate column, {{variable}} rendering, Rerun label

awkoy:
## Details

Fixes three Optimization Studio papercuts reported from dogfooding. **v2 only** — v1 is deliberately untouched (precedent: `4c5ec9a4ca`). Full issue → why → decision notes are in the ticket comments.

### 1. Dead "Pass rate" column on the runs list

`Pass rate` and `Accuracy` were mutually exclusive — each rendered a literal `-` for the run type it did not handle — so **every Studio dataset run showed a permanently empty Pass rate column**. Merged into one objective column rendered per run type.

- Header is type-neutral (**"Best score"**) rather than dynamic via `getObjectiveLabel`: that works in the trials table because one run has one type, but this *list* can mix test-suite and dataset runs, so no static header can be per-row correct.
- The merged column keeps id `accuracy` deliberately — it is already in existing users' saved selected-columns state, so the column stays visible and positioned without bumping `SELECTED_COLUMNS_KEY` and resetting everyone's customizations.
- Also removed `OptimizationsPage`'s run-type column filter. It dropped `accuracy` when no dataset run was present — only ever half a fix (the reverse case is this bug), and after the merge it would strip the *only* score column from a pure test-suite page.

### 2. A brace pair disappeared from displayed prompts

The Studio form uses `{{variable}}`; `studio/types.py::_convert_template_syntax` rewrites it to `{variable}` on ingest, so every view downstream of a run showed a single brace — reading as if the user had authored a broken prompt.

Added `restorePromptVariableFormat` and applied it at the **display boundary only** (nothing is re-sent to the backend in this form). All three v2 prompt read sites now restore: `candidatePrompt.ts`, `BestPrompt.tsx`, `TrialPromptCell.tsx`.

- The pattern is narrower than the existing `convertOptimizationVariableFormat`'s `[^{}]+`: prompts routinely embed literal braces (JSON examples, code) and re-doubling those would corrupt what the user reads, so only identifier-shaped tokens are restored.
- `convertOptimizationVariableFormat` is left untouched because **v1 shares it**.
- The transform walks the value structurally instead of normalizing it, so every prompt wrapper shape survives: bare message array, `{ messages: [...] }`, the single-prompt `{ "<name>": [...] }` wrapper, and the named multi-prompt map.

### 3. "Rerun" never reran

It opens the new-run panel prefilled from the run and never launches anything → **"Duplicate run"**, with the icon changed from `RotateCw` to `Copy` for the same reason.

### Not included

- Extending the ingest conversion to multimodal content — investigated and closed as **not an issue** (OPIK_7582): text-only Studio prompts are a deliberate decision (`disableMedia` defaults true, `StudioMessage.content` is typed `string`), and the real future work is OPIK_4198.

### Known follow-up (design call, not changed here)

On the runs list, `Metric: Equals` sits directly beside `Best score: [equals 0]` — the metric name is duplicated. A bare value would be cleaner and would match Latency/Cost, but this is the pre-existing `FeedbackScoreTag` treatment and a design decision, so it is left as-is.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves [OPIK-7512](https://comet-ml.atlassian.net/browse/OPIK-7512)

Related, not resolved here (underscored on purpose so they don't link): OPIK_7582, OPIK_4198.

## AI-WATERMARK

AI-WATERMARK: yes

- If yes:
  - Tools: Claude Code
  - Model(s): Claude Opus 5 (implementation); Claude Fable 5 (this PR description restructure)
  - Scope: all code and test changes in this PR, plus the PR description
  - Human verification: driven end-to-end in a real browser against a locally-built backend; full frontend test/lint/typecheck suite reviewed (see Testing)

## Testing

- Typecheck clean; eslint/stylelint clean with `--max-warnings=0`.
- **2272/2272 tests** across 151 files pass (9 new), run under Node 20.
- Driven end-to-end in a real browser against a locally-built backend, which caught **two bugs the test suite could not**:
  1. `OptimizationsPage`'s run-type filter stripping the merged column on a test-suite-only page (commit 2).
  2. `TrialPromptCell` keeping the single-brace form via its own duplicate resolver, so the trials table and the overview panel disagreed on the same page — worse than the original papercut, where both were consistently wrong (commit 3).
- Confirmed in the UI: one "Best score" column, `{{question}}` in the best-trial panel, the trials table and the diff hover card, "Duplicate run" in the header, no console errors.
- Not exercised live: the **test-suite branch** of the merged column — no test-suite optimization runs existed in the local dataset. It is covered by unit tests and code reading.

## Documentation

No documentation changes — UI-only fixes with no behavior that docs describe.

[OPIK-7512]: https://comet-ml.atlassian.net/browse/OPIK-7512?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/opik-frontend/src/lib/optimizations.test.ts
+++ b/apps/opik-frontend/src/lib/optimizations.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect } from "vitest";
 import {
   convertOptimizationVariableFormat,
+  restorePromptVariableFormat,
   checkIsTestSuite,
   getOptimizationDefaultConfigByProvider,
   extractKwargsKeysFromPython,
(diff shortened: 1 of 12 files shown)


aadereiko (approved):
LGTM, thanks for covering it with tests!

PR #7660, opened 2026-07-28
Add Cerebras Gemma 4 model metadata

ryanl-cerebras:
## Summary

- Adds gemma-4-31b pricing and capability metadata, corrects current Cerebras output limits, and marks zai-glm-4.7 as supporting reasoning_effort none.
- The current public catalog is `gpt-oss-120b`, `zai-glm-4.7`, and `gemma-4-31b`.

## Why

The previous model ID was retired or the fixed catalog was missing a currently available Cerebras model. This could produce failed requests, stale autocomplete, or misleading examples.

## Validation

- JSON parsing, git diff check, and metadata comparison with the live Cerebras model endpoint.
- Source of truth: https://api.cerebras.ai/public/v1/models


andrescrz (changes requested):
Hi @ryanl-cerebras 

Thanks so much for taking the time to put this PR together—we really appreciate the contribution.

Just a heads-up: the `model_prices_and_context_window.json` file in this repository is automatically synced from the LiteLLM repository on a daily basis. Because of that, any changes made directly to this file here would be overwritten by the next sync.

If you're able to submit the same changes to the upstream LiteLLM repository instead, we'll automatically pick them up during our daily sync once they're merged.

Thanks again for contributing and helping improve the project!

ryanl-cerebras:
Thanks for the guidance. The upstream change is in BerriAI/litellm#35007, including the bundled fallback catalog update requested there. Closing this PR so Opik can receive the metadata through its daily LiteLLM sync once that PR merges.

PR #7675, opened 2026-07-29
[OPIK-6880] [HELM] Add Hyperscale cluster config flags (no-op today)

thiagohora:
## Details

Lands the Helm values and cluster config flags the downstream Hyperscale slices will flip on, so none of those slices has to ship a chart change alongside its behavior change. All flags are inert on the current single-shard deployment. Two of them are wired to existing backend config rather than being dead keys: `clickhouse.sharding.enabled` drives the `clickhouse-cluster` readiness probe (`ANALYTICS_DB_CLUSTER_HEALTH_CHECK_ENABLED`) and `databaseAnalytics.uuidV7ValidationWindowMs` renders `UUID_VALIDATION_WINDOW`; in both cases an explicit `component.backend.env` entry still wins, so operators can pin either env var directly.

- The cluster layout now carries an explicit `shards[]` list pinning `internalReplication: "true"`, so Distributed inserts hit one live replica per shard at any replica count instead of relying on the operator's `replicasCount > 1` default. Shard names are left unset, so the operator keeps generating them by index and host identity is unchanged.
- `clickhouse.replicasCount` now defaults to `2`. **This is user facing**: a fresh install still cannot complete its analytics migrations on multiple replicas (OPIK_7332), so it must install at `1` and scale up afterwards. `NOTES.txt` already prints that reminder and it is now part of the default install output; the self-host scaling docs say so explicitly.
- Docker Compose mirrors the same cluster definition (`internal_replication`) and the same flag defaults, including the two Hyperscale readiness probes and the UUID validation trio.

## Change checklist
- [x] User facing
- [x] Documentation update

## Issues

- Resolves #
- OPIK-6880

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: Drafted the chart/template/compose changes, the helm-unittest suites and the doc note; regenerated the helm-docs README.
- Human verification: Author reviewed the diff and directed the scope, including the decision to keep the literal ticket scope for `replicasCount` and `internal_replication`.

## Testing

Ran from `deployment/helm_chart/opik`:

- `helm lint --values values.yaml .` — passes.
- `helm template opik .` — renders on defaults, with `tieredStorage.enabled=true`, and with `shardsCount=3 sharding.enabled=true`.
- Every assertion in the new/changed helm-unittest suites was additionally verified by rendering the chart with the same `--set` overrides and comparing the parsed YAML: shard count/`internalReplication` across `shardsCount` 1/3 and `replicasCount` 1/2, the derived `ANALYTICS_DB_CLUSTER_HEALTH_CHECK_ENABLED` / `UUID_VALIDATION_WINDOW` values, the explicit-env-wins precedence, and that neither derived key leaks into the python-backend ConfigMap.
- `shardsCount: 0` (already an invalid config) is guarded so `shards:` is omitted rather than rendered as a null where the CRD expects an array.
- XML validity of `additional_config.xml` and `docker compose config` both check out.
- `helm-docs` regenerated `README.md` (three rows changed).

Not run locally: `helm unittest` itself — the plugin is not installed in this environment, so the suites are exercised by the `unittest-helm-chart` CI job. That is also the only coverage for the `NOTES.txt` assertions, since `helm template` does not render `NOTES.txt`.

Not exercised: applying the new `shards[]` layout against a live clickhouse-operator. The reasoning is that the CRD documents `shards[].name` as generated when unset and the operator derives host names from the shard/replica index, so supplying `shards[]` alongside `shardsCount` should be identity-preserving — worth a staging apply before this reaches a cluster with data.

## Documentation

`self-host/scaling.mdx` (both docs and docs-v2) now states that the chart ships `clickhouse.replicasCount: 2`, so a fresh install takes the two-phase path unless overridden.

### DIFF
--- a/deployment/docker-compose/clickhouse_config/additional_config.xml
+++ b/deployment/docker-compose/clickhouse_config/additional_config.xml
@@ -24,9 +24,15 @@
         <path>/clickhouse/task_queue/ddl</path>
     </distributed_ddl>
 
+    <!-- Mirrors the Helm cluster definition (clickhouse.shardsCount / replicasCount): one shard, one
+         replica, internal_replication on. Compose runs a single replica, so internal_replication has
+         nothing to fan out to and is a no-op here — it is set to keep the single-node cluster
+         definition identical to the Kubernetes one, where Distributed inserts must hit one live
+         replica and let ReplicatedMergeTree replicate the parts. -->
     <remote_servers>
         <cluster>
         <shard>
+            <internal_replication>true</internal_replication>
             <replica>
             <host>clickhouse</host>
             <port>9000</port>
--- a/deployment/helm_chart/opik/templates/clickhouseinstallation.yaml
+++ b/deployment/helm_chart/opik/templates/clickhouseinstallation.yaml
@@ -97,10 +97,12 @@ spec:
           # list is rendered only then — a single-replica deployment keeps the plain shardsCount /
           # replicasCount layout it has today. This states explicitly what the operator would
           # otherwise infer from replicasCount > 1. Shard names are left unset so the operator keeps
-          # generating them by index — host identity is unchanged.
-          {{- if gt (int .Values.clickhouse.replicasCount) 1 }}
+          # generating them by index — host identity is unchanged. A non-positive shardsCount renders
+          # no list at all rather than one clamped entry, which would contradict the shardsCount
+          # above and leave the operator to reconcile the two.
+          {{- if and (gt (int .Values.clickhouse.replicasCount) 1) (gt (int .Values.clickhouse.shardsCount) 0) }}
           shards:
-            {{- range $i := until (int (max 1 (int .Values.clickhouse.shardsCount))) }}
+            {{- range $i := until (int .Values.clickhouse.shardsCount) }}
             - internalReplication: "true"
             {{- end }}
           {{- end }}
--- a/deployment/helm_chart/opik/templates/clickhouseinstallation.yaml
+++ b/deployment/helm_chart/opik/templates/clickhouseinstallation.yaml
@@ -93,13 +93,14 @@ spec:
           replicasCount: {{ .Values.clickhouse.replicasCount }}
           # internal_replication pins Distributed-engine inserts to one live replica per shard and
           # lets ReplicatedMergeTree replication fan the parts out, instead of the Distributed table
-          # writing every replica itself. The operator would default it to true only once
-          # replicasCount > 1; setting it here makes it explicit and stable across replica counts
-          # (a no-op on a single replica, where there is nothing to fan out to). Shard names are
-          # left unset so the operator keeps generating them by index — host identity is unchanged.
-          {{- if ge (int .Values.clickhouse.shardsCount) 1 }}
+          # writing every replica itself. Only meaningful with more than one replica, so the shards
+          # list is rendered only then — a single-replica deployment keeps the plain shardsCount /
+          # replicasCount layout it has today. This states explicitly what the operator would
+          # otherwise infer from replicasCount > 1. Shard names are left unset so the operator keeps
+          # generating them by index — host identity is unchanged.
+          {{- if gt (int .Values.clickhouse.replicasCount) 1 }}
           shards:
-            {{- range $i := until (int .Values.clickhouse.shardsCount) }}
+            {{- range $i := until (int (max 1 (int .Values.clickhouse.shardsCount))) }}
             - internalReplication: "true"
             {{- end }}
           {{- end }}
--- a/deployment/helm_chart/opik/templates/clickhouseinstallation.yaml
+++ b/deployment/helm_chart/opik/templates/clickhouseinstallation.yaml
@@ -91,6 +91,18 @@ spec:
         layout:
           shardsCount: {{ .Values.clickhouse.shardsCount }}
           replicasCount: {{ .Values.clickhouse.replicasCount }}
+          # internal_replication pins Distributed-engine inserts to one live replica per shard and
+          # lets ReplicatedMergeTree replication fan the parts out, instead of the Distributed table
+          # writing every replica itself. The operator would default it to true only once
+          # replicasCount > 1; setting it here makes it explicit and stable across replica counts
+          # (a no-op on a single replica, where there is nothing to fan out to). Shard names are
+          # left unset so the operator keeps generating them by index — host identity is unchanged.
+          {{- if ge (int .Values.clickhouse.shardsCount) 1 }}
+          shards:
+            {{- range $i := until (int .Values.clickhouse.shardsCount) }}
+            - internalReplication: "true"
+            {{- end }}
+          {{- end }}
         templates:
           podTemplate: {{ .Values.clickhouse.templates.podTemplate }}
           serviceTemplate: {{ .Values.clickhouse.templates.serviceTemplate }}
--- a/deployment/helm_chart/opik/templates/configmap-backend.yaml
+++ b/deployment/helm_chart/opik/templates/configmap-backend.yaml
@@ -28,5 +28,15 @@ data:
   PYTHON_CODE_EXECUTOR_IMAGE_TAG: {{ $.Chart.AppVersion | quote }}
   {{- end }}
   {{- end }}
+  {{- if eq $key "backend" }}
+  {{- /* Derived from the clickhouse / databaseAnalytics values above. An explicit key under
+         component.backend.env wins, so operators can still pin either env var directly. */}}
+  {{- if not (hasKey $value.env "ANALYTICS_DB_CLUSTER_HEALTH_CHECK_ENABLED") }}
+  ANALYTICS_DB_CLUSTER_HEALTH_CHECK_ENABLED: {{ $.Values.clickhouse.sharding.enabled | quote }}
+  {{- end }}
+  {{- if not (hasKey $value.env "UUID_VALIDATION_WINDOW") }}
+  UUID_VALIDATION_WINDOW: {{ printf "%dms" (int $.Values.databaseAnalytics.uuidV7ValidationWindowMs) | quote }}
+  {{- end }}
+  {{- end }}
 {{- end }}
 {{- end }}
--- a/deployment/helm_chart/opik/templates/configmap-backend.yaml
+++ b/deployment/helm_chart/opik/templates/configmap-backend.yaml
@@ -29,13 +29,33 @@ data:
   {{- end }}
   {{- end }}
   {{- if eq $key "backend" }}
-  {{- /* Derived from the clickhouse / databaseAnalytics values above. An explicit key under
-         component.backend.env wins, so operators can still pin either env var directly. */}}
-  {{- if not (hasKey $value.env "ANALYTICS_DB_CLUSTER_HEALTH_CHECK_ENABLED") }}
-  ANALYTICS_DB_CLUSTER_HEALTH_CHECK_ENABLED: {{ $.Values.clickhouse.sharding.enabled | quote }}
+  {{- /* Hyperscale / analytics data-model flags, derived from the clickhouse, databaseAnalytics,
+         databaseAnalyticsDataModel and partitionMetrics values. They mirror the defaults in
+         apps/opik-backend/config.yml so the chart states them explicitly instead of relying on the
+         image's built-in default, and every downstream slice has a flag to flip without a chart
+         change. An explicit key under component.backend.env wins over anything derived here. */}}
+  {{- $env := $value.env | default dict }}
+  {{- $da := $.Values.databaseAnalytics }}
+  {{- $dm := $.Values.databaseAnalyticsDataModel }}
+  {{- $pm := $.Values.partitionMetrics }}
+  {{- range $k, $v := dict
+        "ANALYTICS_DB_CLUSTER_HEALTH_CHECK_ENABLED" $.Values.clickhouse.sharding.enabled
+        "ANALYTICS_DB_COLD_STORAGE_DISK_HEALTH_CHECK_ENABLED" $da.coldStorageDiskHealthCheckEnabled
+        "UUID_VALIDATION_ENABLED" $da.uuidV7ValidationEnabled
+        "UUID_VALIDATION_AUDIT_ONLY" $da.uuidV7ValidationAuditOnly
+        "UUID_VALIDATION_WINDOW" $da.uuidV7ValidationWindow
+        "ANALYTICS_DB_DATA_MODEL_TRACE_COLUMNS_NON_NULLABLE" $dm.traceColumnsNonNullable
+        "ANALYTICS_DB_DATA_MODEL_SPAN_COLUMNS_NON_NULLABLE" $dm.spanColumnsNonNullable
+        "ANALYTICS_DB_DATA_MODEL_TRACE_DELETION_EVENTS_CAPTURE_ENABLED" $dm.traceDeletionEventsCaptureEnabled
+        "ANALYTICS_DB_DATA_MODEL_SPAN_DELETION_EVENTS_CAPTURE_ENABLED" $dm.spanDeletionEventsCaptureEnabled
+        "ANALYTICS_DB_DATA_MODEL_DELETION_EVENTS_INSERT_BATCH_SIZE" $dm.deletionEventsInsertBatchSize
+        "PARTITION_METRICS_ENABLED" $pm.enabled
+        "PARTITION_METRICS_INTERVAL" $pm.interval
+        "PARTITION_METRICS_LWD_TABLES" $pm.lwdTables
+      }}
+  {{- if not (hasKey $env $k) }}
+  {{ $k }}: {{ $v | quote }}
   {{- end }}
-  {{- if not (hasKey $value.env "UUID_VALIDATION_WINDOW") }}
-  UUID_VALIDATION_WINDOW: {{ printf "%dms" (int $.Values.databaseAnalytics.uuidV7ValidationWindowMs) | quote }}
   {{- end }}
   {{- end }}
 {{- end }}
--- a/deployment/helm_chart/opik/values.yaml
+++ b/deployment/helm_chart/opik/values.yaml
@@ -681,14 +681,33 @@ altinity-clickhouse-operator:
     enabled: false
     interval: '' # workaround for https://github.com/Altinity/clickhouse-operator/issues/1759
 
+# Backend-side settings for the analytics (ClickHouse) database. Rendered into the backend env
+# ConfigMap; each key maps to the ANALYTICS_DB_* / UUID_VALIDATION_* env var documented in
+# apps/opik-backend/config.yml.
+databaseAnalytics:
+  # Window, in milliseconds, for UUIDv7 ingestion validation: writes whose `id` embeds a timestamp
+  # further than this into the past or future are rejected. Rendered as UUID_VALIDATION_WINDOW.
+  # Accepted range is 12h .. 45d; the backend rejects anything outside it at startup. Inert while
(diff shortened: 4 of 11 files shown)


deployment/helm_chart/opik/values.yaml
```diff
@@ -681,14 +681,33 @@ altinity-clickhouse-operator:
...
   enabled: true
   shardsCount: 1
   # For a FRESH High-Availability install, use the two-phase flow: install with replicasCount 1, let
   # opik-backend finish all migrations, THEN raise replicasCount and upgrade so the ClickHouse operator
   # clones the schema to the new replicas. Installing directly on >= 2 replicas prevents the analytics
-  # migrations from completing on a fresh cluster. See the self-host/scaling docs.
-  replicasCount: 1
+  # migrations from completing on a fresh cluster (OPIK-7332), so a fresh install at this default must
+  # override replicasCount to 1 first — NOTES.txt prints the reminder. See the self-host/scaling docs.
+  replicasCount: 2
```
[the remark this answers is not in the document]
thiagohora: Deliberate, and flagging it for a human call rather than resolving it here.

The `replicasCount: 2` default is specified by OPIK-6880 itself (Scope: "`clickhouse.replicasCount` (default 2)"), so this PR implements it as written. The migration constraint you describe is real and not disputed — a fresh install on >= 2 replicas still cannot complete its analytics migrations.

What this PR does about it, rather than automating the two-phase flow:

- `NOTES.txt` prints the two-phase banner whenever `replicasCount > 1`, so it is now part of the default install output. `clickhouse_migration_test.yaml` was updated to assert that (previously it asserted the banner is *absent* at the default), and `unittest-helm-chart` is green.
- `values.yaml` says a fresh install at this default must override `replicasCount` to `1` first.
- `self-host/scaling.mdx` (both variants) now states the chart ships `2`, so a fresh install takes the two-phase path unless overridden.

So the failure mode is loud rather than silent, but you are right that out-of-the-box `helm install` is not self-completing. That trade-off is a product decision above this PR: safe-by-default (`1`) contradicts the ticket, and automating the two-phase install is new engineering that needs its own ticket.

Reverting to `1` is a one-line change plus reverting the NOTES test and the docs sentence — happy to do it if the team prefers safe-by-default. Raising with the ticket author to get the bullet amended either way.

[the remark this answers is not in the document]
andrescrz: I believe we should keep the previous value of 1 here. The migration failure on a fresh installation is real, and our documentation mentions the protocol: install fresh, then add the 2nd replica. Also while, the majority of actual production environments have 2 replicas, many times customer try or start with a single 1.

I take responsibility for this as I wrote the ticket, which was misleading. Apologies about that.

Let's revert this to 1, and align all changes in the PR including documentation.

deployment/helm_chart/opik/values.yaml
```diff
@@ -681,14 +681,33 @@ altinity-clickhouse-operator:
...
+# Backend-side settings for the analytics (ClickHouse) database. Rendered into the backend env
+# ConfigMap; each key maps to the ANALYTICS_DB_* / UUID_VALIDATION_* env var documented in
+# apps/opik-backend/config.yml.
+databaseAnalytics:
+  # Window, in milliseconds, for UUIDv7 ingestion validation: writes whose `id` embeds a timestamp
+  # further than this into the past or future are rejected. Rendered as UUID_VALIDATION_WINDOW.
+  # Accepted range is 12h .. 45d; the backend rejects anything outside it at startup. Inert while
+  # UUID_VALIDATION_ENABLED is false (the default) — set that key under component.backend.env to
+  # turn validation on, and UUID_VALIDATION_AUDIT_ONLY to count offenders without rejecting them.
+  uuidV7ValidationWindowMs: 86400000 # 24h
```
andrescrz: This validation defaults to 24h and it will always be large values and likely never smaller than that. Maybe we could extend it to 48h if we see issues in production. 

Because of that, I'd rather not use milliseconds as the default, due to being less readable.

If possible, I'd use Dropwizard durations here, like 24h.

deployment/helm_chart/opik/values.yaml
```diff
@@ -681,14 +681,33 @@ altinity-clickhouse-operator:
...
-  replicasCount: 1
+  # migrations from completing on a fresh cluster (OPIK-7332), so a fresh install at this default must
+  # override replicasCount to 1 first — NOTES.txt prints the reminder. See the self-host/scaling docs.
+  replicasCount: 2
+  # Switches the deployment to the sharded (Hyperscale) topology. A no-op while shardsCount is 1: it
+  # only drives the backend's `clickhouse-cluster` readiness probe, which verifies the Distributed
+  # cluster definition is visible from the node (ANALYTICS_DB_CLUSTER_HEALTH_CHECK_ENABLED — set that
+  # key explicitly under component.backend.env to decouple the two). Turn on together with
+  # shardsCount > 1 once the Distributed tables are in place.
+  sharding:
```
andrescrz, line 738: Nit: move it above close to `shardsCount`.

deployment/helm_chart/opik/templates/configmap-backend.yaml
```diff
@@ -28,5 +28,15 @@ data:
   PYTHON_CODE_EXECUTOR_IMAGE_TAG: {{ $.Chart.AppVersion | quote }}
   {{- end }}
   {{- end }}
+  {{- if eq $key "backend" }}
+  {{- /* Derived from the clickhouse / databaseAnalytics values above. An explicit key under
+         component.backend.env wins, so operators can still pin either env var directly. */}}
+  {{- if not (hasKey $value.env "ANALYTICS_DB_CLUSTER_HEALTH_CHECK_ENABLED") }}
+  ANALYTICS_DB_CLUSTER_HEALTH_CHECK_ENABLED: {{ $.Values.clickhouse.sharding.enabled | quote }}
```
andrescrz: At this point, we added more flags related to this project than the ones originally described in this ticket.

I recommend audit the latest status of `config.yml` file from the Java backend and close any gaps here and add any missing flag.

Feel free to propagate that to the docker compose file.

deployment/helm_chart/opik/templates/clickhouseinstallation.yaml
```diff
@@ -91,6 +91,18 @@ spec:
         layout:
           shardsCount: {{ .Values.clickhouse.shardsCount }}
           replicasCount: {{ .Values.clickhouse.replicasCount }}
+          # internal_replication pins Distributed-engine inserts to one live replica per shard and
+          # lets ReplicatedMergeTree replication fan the parts out, instead of the Distributed table
+          # writing every replica itself. The operator would default it to true only once
+          # replicasCount > 1; setting it here makes it explicit and stable across replica counts
+          # (a no-op on a single replica, where there is nothing to fan out to). Shard names are
+          # left unset so the operator keeps generating them by index — host identity is unchanged.
+          {{- if ge (int .Values.clickhouse.shardsCount) 1 }}
```
andrescrz: Minor: maybe we should consider enabling `internal_replication` only when more than 1 replica.

andrescrz (commented):
Let's adjust this, apologies for the confusion on the ticket.

Once my review settles, let's ask DevOps for general review and Lior about the particular ClickHouse settings, mostly `internal_replication`. A 2nd opinion in both cases is going to be desirable, for more safety.

deployment/helm_chart/opik/values.yaml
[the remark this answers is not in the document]
thiagohora: Reverted to `1` in 86c4b69864, and no apology needed — I raised the same concern before implementing and should have pushed back harder rather than following the bullet.

Aligned everything with it: the `values.yaml` comment is back to the original wording, `clickhouse_migration_test.yaml` is restored (the banner is absent at the default again), and both `self-host/scaling.mdx` variants are reverted to their `main` content — the sentence about the chart shipping `2` is gone.

deployment/helm_chart/opik/values.yaml
thiagohora: Agreed, done in 86c4b69864 — the key is now `databaseAnalytics.uuidV7ValidationWindow: 24h` and renders `UUID_VALIDATION_WINDOW: "24h"` verbatim, no ms conversion in the template. Docker compose default moved to `24h` too.

Two notes:
- I dropped the `Ms` suffix from the name since it no longer carries milliseconds. That diverges from the ticket bullet (`uuidV7ValidationWindowMs`), flagging in case you want the name kept.
- I left it under `databaseAnalytics` where the ticket put it, even though the backend has it under its own `uuidValidation` section in `config.yml`. Happy to move it to a matching `uuidValidation` values block if you prefer the chart to mirror `config.yml` exactly.

deployment/helm_chart/opik/values.yaml
thiagohora: Moved — `sharding` now sits directly after `shardsCount`, before the `replicasCount` block. 86c4b69864

deployment/helm_chart/opik/templates/clickhouseinstallation.yaml
thiagohora: Good call, done in 86c4b69864. The shards list is now rendered only when `replicasCount > 1`, so a single-replica deployment keeps the plain `shardsCount` / `replicasCount` layout it has today, untouched.

That also makes the setting honest about what it is: it states explicitly what the operator would otherwise infer from `replicasCount > 1`, rather than forcing a value where it has no meaning. I verified against prod (`chi-opik-clickhouse-common-configd`) that the operator already emits `<internal_replication>True</internal_replication>` on the `cluster` shard at 2 replicas, so this matches the live config rather than changing it.

Test coverage now pins both directions: no shards list at 1 replica (including a sharded single-replica cluster), and `internalReplication: "true"` on every shard at 2.

deployment/helm_chart/opik/templates/configmap-backend.yaml
thiagohora: Audited `config.yml` and closed the gaps in 86c4b69864. Flags now stated in the chart at their documented defaults, all previously absent:

| Flag | Default | Was in chart | Was in compose |
|---|---|---|---|
| `ANALYTICS_DB_COLD_STORAGE_DISK_HEALTH_CHECK_ENABLED` | `false` | no | yes |
| `ANALYTICS_DB_DATA_MODEL_TRACE_COLUMNS_NON_NULLABLE` | `false` | no | yes |
| `ANALYTICS_DB_DATA_MODEL_SPAN_COLUMNS_NON_NULLABLE` | `false` | no | yes |
| `ANALYTICS_DB_DATA_MODEL_TRACE_DELETION_EVENTS_CAPTURE_ENABLED` | `false` | no | yes |
| `ANALYTICS_DB_DATA_MODEL_SPAN_DELETION_EVENTS_CAPTURE_ENABLED` | `false` | no | yes |
| `ANALYTICS_DB_DATA_MODEL_DELETION_EVENTS_INSERT_BATCH_SIZE` | `1000` | no | no |
| `UUID_VALIDATION_ENABLED` | `false` | no | no |
| `UUID_VALIDATION_AUDIT_ONLY` | `false` | no | no |
| `PARTITION_METRICS_ENABLED` | `false` | no | no |
| `PARTITION_METRICS_INTERVAL` | `5m` | no | no |
| `PARTITION_METRICS_LWD_TABLES` | `traces,spans` | no | no |

New values sections mirror the `config.yml` section names: `databaseAnalyticsDataModel` and `partitionMetrics` (OPIK_6904 — that one was missing from both the chart and compose). All propagated to docker compose.

The derived block became a `dict` range rather than 13 hand-written `hasKey` guards; sprig iterates a dict in sorted key order, so output is deterministic. An explicit `component.backend.env` key still wins over every derived value.

Deliberately excluded, happy to add any of these if you disagree:

- `ANALYTICS_DB_QUERY_PARAMETERS` — OPIK_6881 shipped the profile settings as the `config.yml` default and the chart does not override it. Pinning that long tuning string in the chart would freeze it and let it drift from the backend default.
- `ANALYTICS_DB_ASYNC_INSERT_*` — these default to unset, meaning "keep the server default". Rendering them as empty ConfigMap values would send empty strings where the backend expects an integer or absence, so this would not be inert.
- `ANALYTICS_DB_HEALTH_CHECK_TIMEOUT`, `CLICKHOUSE_LOG_APPENDER_*`, `ANALYTICS_DB_READ_ONLY_FREEFORM_SQL_*` — generic or belonging to other workstreams (the last is Agent Insights, OPIK_6846, already in compose).

Verification: a default `helm template` now differs from `main` only by those inert ConfigMap keys and one comment block — 22 diff lines, no rendered Kubernetes object changes shape.

deployment/helm_chart/opik/templates/clickhouseinstallation.yaml
```diff
@@ -91,6 +91,19 @@ spec:
...
+          # internal_replication pins Distributed-engine inserts to one live replica per shard and
+          # lets ReplicatedMergeTree replication fan the parts out, instead of the Distributed table
+          # writing every replica itself. Only meaningful with more than one replica, so the shards
+          # list is rendered only then — a single-replica deployment keeps the plain shardsCount /
+          # replicasCount layout it has today. This states explicitly what the operator would
+          # otherwise infer from replicasCount > 1. Shard names are left unset so the operator keeps
+          # generating them by index — host identity is unchanged.
+          {{- if gt (int .Values.clickhouse.replicasCount) 1 }}
+          shards:
+            {{- range $i := until (int (max 1 (int .Values.clickhouse.shardsCount))) }}
```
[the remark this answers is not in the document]
thiagohora: Agreed, fixed in cb91931439 — guarded on `shardsCount > 0` and dropped the clamp.

The contradiction was the real problem: `max 1` emitted `shardsCount: 0` alongside a one-element `shards` list, which only resolved because the operator takes `max(shardsCount, len(shards))`. Relying on the normalizer to reconcile a spec that disagrees with itself is worse than omitting the block.

The guard also subsumes what the clamp was there for. Without it an empty `until` range would emit a bare `shards:` key — a null where the CRD expects an array — and that is now unreachable, so the `max` call and its `int` reconversion are gone.

Verified across the full matrix, asserting both the parsed value and the absence of a bare `shards:` key in the raw YAML:

| shardsCount | replicasCount | shards |
|---|---|---|
| 1 | 1 | omitted |
| 1 | 2 | 1 entry, `internalReplication: "true"` |
| 3 | 2 | 3 entries, all true |
| 3 | 1 | omitted |
| 0 | 2 | omitted |
| 0 | 1 | omitted |

The test that pinned the old behaviour is now `omits the shards list for a non-positive shard count`, asserting `notExists` rather than a count of 1.

obezpalko (approved):
Looks good

liyaka (approved):

deployment/docker-compose/clickhouse_config/additional_config.xml
```diff
@@ -24,9 +24,15 @@
...
 
+    <!-- Mirrors the Helm cluster definition (clickhouse.shardsCount / replicasCount): one shard, one
+         replica, internal_replication on. Compose runs a single replica, so internal_replication has
+         nothing to fan out to and is a no-op here — it is set to keep the single-node cluster
+         definition identical to the Kubernetes one, where Distributed inserts must hit one live
+         replica and let ReplicatedMergeTree replicate the parts. -->
     <remote_servers>
         <cluster>
         <shard>
+            <internal_replication>true</internal_replication>
```
andrescrz, line 35: Optional consideration: I missed that we also have this config file for the test container. You added it to docker and Helm, that would be the remaining place to set this config, if we want to test that. 

Having said that, we use the test containers for testing from the functional point of view. Testing topology or infra settings are generally out of scope, except for a few reasonable exception. Replication is likely not one of those as there's a single node replica in this case anyway. 

In addition, we probably have many other similar inconsistencies anyway. 

In summary, likely no need to address this, but I wanted to drop the comment for your awareness.

deployment/helm_chart/opik/templates/configmap-backend.yaml
andrescrz: My only doubt is for `ANALYTICS_DB_ASYNC_INSERT_BUSY_TIMEOUT_MAX_MS`, which we need to increase during the cutover. However, I'm not sure if that increase should be done here or simply temporarily in the env vars of the deployed chart. I'm leaning towards the 2nd, but you could investigate so we understand better.

andrescrz (approved):
Couple of optional comments, but good to move forward.

We can ask Lior to double check the `internal_replication` for more certainty.

deployment/docker-compose/clickhouse_config/additional_config.xml
thiagohora: Thanks — the drift is real, and it is slightly worse than "one more inconsistency": the file is `apps/opik-backend/src/test/resources/clickhouse.xml` (mounted by `ClickHouseContainerUtils.java:90` to `config.d/`), and its own header comment says it "Mirrors deployment/docker-compose/clickhouse_config/additional_config.xml" — which this PR just made untrue.

Agreeing not to fix it here, for a concrete reason beyond scope: `backend_tests.yml` triggers on `apps/opik-backend/**`, so a one-line inert XML edit would spin up the whole backend test matrix on an approved chart-only PR.

And it would be inert. The container is a single node with one replica, and `internal_replication` only affects how a Distributed table routes inserts across replicas of a shard — with one replica there is nothing to route. `HealthCheckIntegrationTest` only needs the `cluster` remote_server to be *defined*, which this would not change.

Better carried by a slice that already touches backend code, where that suite runs anyway. Leaving the thread for whoever picks up the cutover.

deployment/helm_chart/opik/templates/configmap-backend.yaml
thiagohora: Investigated, and I have to correct myself first: my earlier reason for excluding these was wrong. I said rendering them would "send empty strings where the backend expects an integer, so this would not be inert". That is not right — `config.yml` already ships `${ANALYTICS_DB_ASYNC_INSERT_*:-}` as the default for all three, so the empty case is the normal path in every environment today. An empty substitution leaves a bare YAML scalar, which parses to a null `Integer`, and `@Min(1)` does not fire on null. Adding them to the chart with empty defaults would have been safe. The exclusion still stands, but for the reason below, not that one.

How the field actually works (`DatabaseAnalyticsFactory` 48-67 and 144-172): these are *optional overrides* merged into the `custom_http_params` chain — overriding a present value or adding it when absent — and when unset, `queryParameters` is left untouched.

That mechanism is what makes me agree with your second option:

1. **It already works with no chart change.** `component.backend.env` is a free-form map rendered straight into the ConfigMap. Verified on this branch: `--set component.backend.env.ANALYTICS_DB_ASYNC_INSERT_BUSY_TIMEOUT_MAX_MS=2000` renders `ANALYTICS_DB_ASYNC_INSERT_BUSY_TIMEOUT_MAX_MS: "2000"`, and the default render contains zero `ASYNC_INSERT` keys.
2. **Removal is a clean rollback.** Because unset means "leave `queryParameters` alone", deleting the key restores the baked-in `async_insert_busy_timeout_max_ms=250` with no second edit. For a time-boxed cutover window that reversibility is the property you want.
3. **The value is environment- and window-specific** — prod, during the cutover — so it belongs in the environment overlay, where it is still version-controlled and auditable, rather than as an OSS chart default that every self-hosted install inherits.
4. A values-level knob would encode a temporary state as a permanent default without saving any work: you would edit the overlay either way.

One trap worth flagging for whoever runs the cutover: do **not** bump it by editing `ANALYTICS_DB_QUERY_PARAMETERS`. That means re-pasting the entire tuning string, which risks silently dropping one of the other settings and drifting from the `config.yml` default. The dedicated override exists precisely so the cutover only has to state the one value it is changing.

Happy to add all three to the chart with empty defaults if you would rather have them discoverable — now that the safety question is settled it is a small change — but I would leave them out on the reasoning above.

PR #7678, opened 2026-07-29
[OPIK-7532] [QA] feat: capability coverage map — taxonomy, spec tags, and tag-lint PR gate

AndreiCautisanu:
## Details

Establishes the coverage map for `tests_end_to_end/`: a reviewed taxonomy that defines the denominator, `@area:`/`@cap:` tags on every spec, and a PR gate that keeps the two in sync. This is the D3 foundation the coverage dashboard and radar build on (OPIK-7532).

Two commits, reviewable separately:

1. **`test:`** — tags all 26 specs. Tags only; no logic, selectors or assertions changed.
2. **`ci:`** — adds the taxonomy, the lint, the conventions doc, and the workflow.

```
tests_end_to_end/coverage/taxonomy.yaml   18 areas, ~188 capabilities
tests_end_to_end/coverage/tag_lint.py     6 rules, exits non-zero on any
tests_end_to_end/TESTING-TAGS.md          the grammar and how to extend it
.github/workflows/tag_lint.yml            PR gate on tests_end_to_end/**
```

### The tags

```
@area:<area>              which product area a spec belongs to (one per spec)
@cap:<area>.<capability>  which capabilities a test proves
@vcap:<area>.<capability> same, for visual specs
```

Existing tier tags (`@t1-smoke`/`@t2-cuj`/`@t3-nightly`) and suite selectors (`@t1-stsaas`, `@provider-sanity`) are untouched, so **nothing about which tests run in which job changes** — verified below.

`@area:` on the describe; `@cap:` per test where a describe holds several distinct tests, on the describe where its tests are variants of one capability (the prompt text/chat loops). Per-test is what lets a red run name the capability that regressed rather than just the area.

### Why this gates PRs rather than running nightly

Tier selection is `--grep`, so a spec with a missing or typo'd tag is **not a failure** — it is silently never selected, and the root `npm test` even passes `--pass-with-no-tests`. An untagged spec stops running with no signal at all. Catching that at review time is the point; a nightly job reports it the morning after.

### Why the taxonomy lives in this repo

A gate that reads the taxonomy must work on PRs from external contributors, where no cross-repo checkout token is available. Keeping it beside the specs it describes means the check needs no secrets, no internal runner, and no cross-repo checkout. The coverage builder reads it from here.

### Coverage as it stands

Reported per dimension, never blended into one number:

| dimension | covered | note |
|---|---|---|
| functional | 75/188 | 41% self-hosted, excluding the two cloud-only areas |
| visual | 22/54 | opt-in denominator |
| load | 0/9 | denominator defined, `status: planned` |

The load dimension is declared but deliberately inert — `tests_load/` reports to JUnit rather than Allure, so it is not joinable yet, and wiring it up is a separate decision.

Three tags were deliberately left off, each because including them would have overcounted:

- The bare `@prompts` tag on the four prompt-playground specs becomes **`@area:playground`, not `@area:prompts`** — those specs drive the Playground and only assert on the library as an outcome.
- `trace-explore-smoke` asserts the project breadcrumb, but only as project context for the trace panel, so it does **not** claim `@cap:projects.breadcrumb-selector`. A cross-area cap there would report the Projects area as covered on an incidental assertion. (The lint caught this one.)
- `@trace-explore` is retired: it predated the split of the Logs surface into distinct `traces` and `threads` areas. Its three specs resolve to `@area:traces` (trace-explore-smoke, trace-spans-depth) and `@area:threads` (thread-logging-smoke).

## Change checklist
- [ ] User facing
- [x] Documentation update

## Issues

- OPIK-7532

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: derived the taxonomy from the frontend routes, nav and docs; mapped each test to the capabilities it actually asserts by reading the test bodies; wrote the lint and workflow; ran the verification below.
- Human verification: Andrei Cautisanu reviewed the taxonomy and its altitude, chose the per-test granularity, decided where the taxonomy and gate live, and approved the diff and verification results before this was pushed.

## Testing

The risk in a tagging change is silently altering which tests get selected, so verification focused on proving selection is unchanged. Ran from `tests_end_to_end/e2e` and `tests_end_to_end/visual-tests` (`npm install`, then Playwright's own resolver) on this branch and on unmodified `main`:

```
./node_modules/.bin/tsc --noEmit
./node_modules/.bin/playwright test --list --grep '@t1-smoke'
./node_modules/.bin/playwright test --list --grep '@t1-smoke|@t1-stsaas'
./node_modules/.bin/playwright test --list --grep '@t1-smoke|@t2-cuj'
./node_modules/.bin/playwright test --list --grep '@t1-smoke|@t2-cuj|@t3-nightly'
./node_modules/.bin/playwright test --list
```

| selection | before (main) | after |
|---|---|---|
| `test:t1` | 20 tests / 11 files | 20 tests / 11 files |
| `test:t1-stsaas` | 23 / 12 | 23 / 12 |
| `test:t2` | 41 / 19 | 41 / 19 |
| `test:t3` | 49 / 22 | 49 / 22 |
| all e2e | 56 / 24 | 56 / 24 |
| all visual | 22 / 3 | 22 / 3 |

Identical on every selection. `tsc --noEmit` clean.

New tags resolve at the intended granularity:

```
--grep '@area:playground'                  -> 13 tests in 6 files
--grep '@area:traces'                      ->  6 tests in 2 files
--grep '@cap:datasets.edit-item-versions'  ->  1 test  in 1 file
--grep '@vcap:traces'                      ->  6 tests in 3 files
```

The gate was verified to fail, not just to pass — a lint that cannot fail is worthless:

| scenario | result |
|---|---|
| this branch as-is | `26 specs checked, 1 exempt, 0 problems`, exit 0 |
| typo a cap to `@cap:datasets.serch-items` | exit 1, `unknown capability … not in taxonomy`, annotated on line 78 |
| add a new spec with no tags at all | exit 1, `no tier tag` + `no @area: tag` |

Both injected faults were reverted and the clean run re-confirmed.

Tests not run, with reason:

- **The suites themselves were not executed locally.** They need a live Opik stack plus provider API keys. The only edits to spec files are string literals in `tag` arrays, and `--list` (which performs full test discovery and tag resolution) is identical before and after. CI's `Run v2 suite: t1` executes them.

## Documentation

`tests_end_to_end/TESTING-TAGS.md` — tag grammar, describe-vs-test placement, naming, `test.step()` usage, the Allure mapping, and the process for adding areas and capabilities.

### DIFF
--- a/.github/workflows/qa_coverage_reconcile.yml
+++ b/.github/workflows/qa_coverage_reconcile.yml
@@ -1,0 +1,149 @@
+name: QA Coverage Reconcile
+run-name: "QA Coverage Reconcile ${{ github.ref_name }}"
+
+# Keeps the derived fields in tests_end_to_end/coverage/taxonomy.yaml in step with
+# the actual spec tags, and opens a PR when they drift (OPIK-7631).
+#
+# The tags are the source of truth: a capability is covered when a spec carries
+# its @cap:/@vcap: tag. The `covered:`/`tier:` values in the taxonomy are a cache
+# of that, refreshed here. Nobody edits them by hand.
+#
+# Authored fields — areas, capability keys, notes, cloud_only, state, axes — are
+# never touched by this job. Those change via the discovery job (OPIK-7632) or a
+# human PR.
+#
+# Runs after merges land rather than on PRs: on a PR the tags and the taxonomy
+# are *meant* to disagree until the reviewer accepts the new coverage claim, and
+# tag_lint.yml already gates tag validity there.
+
+on:
+  schedule:
+    - cron: '30 2 * * *'   # 02:30 UTC daily, after the nightly t3 suite settles
+  workflow_dispatch:
+  push:
+    branches:
+      - 'main'
+    paths:
+      - 'tests_end_to_end/e2e/tests/**'
+      - 'tests_end_to_end/visual-tests/tests/**'
+      - 'tests_end_to_end/coverage/**'
+
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: false   # never abandon a run that may have opened a PR
+
+permissions:
+  contents: read
+
+jobs:
+
+  reconcile:
+    name: reconcile taxonomy with spec tags
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+
+      - name: Set up Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+
+      - name: Install PyYAML
+        run: pip install --quiet pyyaml
+
+      # Tags are read via `playwright test --list`, not a regex, because tags
+      # union from describe to test and a file may hold several tiers. No browser
+      # is launched, so `playwright install` is deliberately skipped.
+      - name: Install e2e deps
+        working-directory: tests_end_to_end/e2e
+        run: npm ci --ignore-scripts
+
+      - name: Install visual-tests deps
+        working-directory: tests_end_to_end/visual-tests
+        run: npm ci --ignore-scripts
+
+      - name: Set branch name
+        run: echo "BRANCH_NAME=github-actions/OPIK-7532-reconcile-qa-taxonomy-$(date +%Y-%m-%d-%H-%M-%S)" >> "$GITHUB_ENV"
+
+      - name: Reconcile
+        id: reconcile
+        run: |
+          set -o pipefail
+          python3 tests_end_to_end/coverage/reconcile.py \
+            --taxonomy tests_end_to_end/coverage/taxonomy.yaml \
+            --estate tests_end_to_end \
+            --summary | tee /tmp/reconcile.txt
+          {
+            echo 'summary<<RECONCILE_EOF'
+            cat /tmp/reconcile.txt
+            echo 'RECONCILE_EOF'
+          } >> "$GITHUB_OUTPUT"
+
+      - name: Check for changes
+        id: check_changes
+        run: |
+          if git diff --quiet; then
+            echo "has_changes=false" >> "$GITHUB_OUTPUT"
+          else
+            echo "has_changes=true" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Commit
+        if: steps.check_changes.outputs.has_changes == 'true'
+        run: |
+          set -ex
+          git config --local user.email "github-actions@comet.com"
+          git config --local user.name "Github Actions (${{ github.actor }})"
+          git add tests_end_to_end/coverage/taxonomy.yaml
+          git commit -m "[OPIK-7532] [QA] chore: reconcile coverage taxonomy with spec tags"
+
+      - name: Create Pull Request
+        if: steps.check_changes.outputs.has_changes == 'true'
+        uses: peter-evans/create-pull-request@v8
+        with:
+          token: ${{ secrets.COMET_ACTION_CREATE_PR }}
+          branch: ${{ env.BRANCH_NAME }}
+          title: "[OPIK-7532] [QA] chore: reconcile coverage taxonomy with spec tags"
+          body: |
+            ## Details
+
+            Automated sync of the **derived** fields in
+            `tests_end_to_end/coverage/taxonomy.yaml` with the `@cap:`/`@vcap:` tags
+            actually present in the test estate.
+
+            Only `covered:` and `tier:` are touched. Areas, capability keys, notes,
+            `cloud_only`, `state` and `axes` are authored and left alone.
+
+            ```
+            ${{ steps.reconcile.outputs.summary }}
+            ```
+
+            See the job log for the full change list.
+
+            **Why this is safe to merge on green:** the tags are the source of
+            truth for coverage, so this PR only ever makes the file agree with
+            what the specs already claim. If a line here looks wrong, the tag is
+            wrong — fix the spec, not the taxonomy.
+
+            ## Change checklist
+            - [ ] User facing change
+            - [ ] Documentation update
+
+            ## Issues
+            OPIK-7631
+
+            ## Testing
+            - `tag_lint` + `reconcile --check` pass on this branch
+
+            ## Documentation
+            - `tests_end_to_end/TESTING-TAGS.md`
--- a/.github/workflows/qa_coverage_reconcile.yml
+++ b/.github/workflows/qa_coverage_reconcile.yml
@@ -72,8 +72,11 @@ jobs:
         working-directory: tests_end_to_end/visual-tests
         run: npm ci --ignore-scripts
 
+      # Deliberately stable, with no timestamp: create-pull-request force-updates
+      # this branch and reuses the open PR. A per-run branch name would open a
+      # fresh reconciliation PR every night and leave the old ones to rot.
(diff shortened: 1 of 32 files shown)


AndreiCautisanu:
Addressed the review in 54a862b. Three of the seven findings were real; four I'm declining, with reasons.

**Fixed**

- **Per-variant `@cap:` overstated coverage** (`prompt-library-smoke.spec.ts`) — correct, and the one finding that touched the integrity of the number. Each loop iteration claimed both `create-text-prompt-ui` and `create-chat-prompt-ui` while exercising only one. Split into two explicit tests with literal tags. Note the suggested fix (build the tag array from `variant`) would have been worse than the bug: both the lint and the coverage builder match string literals inside `tag: [...]`, so a computed tag is *invisible* — the capability would silently drop to uncovered rather than be mis-attributed. Documented that trap in `TESTING-TAGS.md`. Also added the versioning caps `runVersioningSteps` genuinely asserts.
- **`@cap:` cardinality unenforced** — correct. No spec violates it today, so it was a latent gate hole, but closing it matters precisely because the failure mode here is silence, not a red build. Now a finding; verified it fires and exits 1.
- **Docstring misstated the rules** — correct, and my own text. Bullet 1 claimed "exactly one tier" when a suite selector is a valid tier-less opt-out; bullet 3 described the visual-state check as scanning specs when it validates the taxonomy. Both corrected.

**Declined**

- **`dataset-items.spec.ts` describe "missing a tier tag"** — false. Tags union from describe to test: the describe carries `@area:datasets` and each `test()` carries its own tier. That's the documented pattern and the reason the file spans two tiers. The suggested fix (collapse to "exactly those two tags" on the describe) would put one tier on tests that need different ones.
- **Flag `len(tiers) != 1`** — the `pass` is deliberate. "Exactly one tier" constrains a single test *after* describe-inheritance; a file-level count can't distinguish sibling describes at different tiers (legitimate — four specs do this, `ollie-agentic` has three) from one test carrying two. The block-aware alternative doesn't help either, since a describe block and its tests are separate `tag: [...]` blocks whose tags union. Doing this correctly needs the TS AST, which the docstring explains was a deliberate non-goal. I'd rather leave a documented hole than ship a check that fails valid specs — the comment now says exactly that.
- **SHA-pin `actions/checkout` + `setup-python`, hash-pin `pyyaml`** — declining both on consistency grounds rather than disagreement in principle. This job has `permissions: contents: read`, no secrets, and no cross-repo checkout. Meanwhile every other workflow in the repo floats major tags, and nothing renovates a hand-pinned SHA. One pinned file isn't a security posture, it's an inconsistency that goes stale. If opik wants SHA pinning it should be a repo-wide policy with automation behind it — happy to open that separately if there's appetite.

Verified after the changes: tag-lint clean across 26 specs, `tsc --noEmit` clean, Playwright selection counts unchanged (`t1` 20/11, all 56/24), and the coverage map still reports 74/188 functional with zero flag drift — so the retagging fixed attribution without moving the number.

AndreiCautisanu:
Second round addressed in 679685b. Four real findings — thanks, one of them was serious.

**The important one: `--list` fails open (reconcile.py)**

Confirmed and fixed. I reproduced it: append a syntax error to one spec and `playwright test --list --reporter=json` exits **1 while still printing valid JSON** — 3.7 kB of it, with `suites: []` and a populated `errors` array. My guard only checked for empty stdout, so it would have parsed that, seen zero tags, and flipped **all 74 covered capabilities to `covered: false`** — then committed the wipe and opened a PR. One typo in any spec would have destroyed the coverage map on the next nightly.

Now checks `returncode`, unparseable JSON, *and* a non-empty `errors` array before anything is written. The error also surfaces the collection failure itself (file, line, message) instead of a slice of the JSON header, since that's what someone debugging actually needs.

Verified: with a broken spec, reconcile exits 1 and `diff -q` shows the taxonomy byte-identical.

**Missing project (reconcile.py)** — same fail-open shape, same fix. A dimension whose `package.json` was absent got silently skipped, leaving its tag map empty, which also reads as "nothing is covered". Now a hard error naming the expected path. Fail-fast was the right call here over best-effort: this job's whole value is that its output can be trusted without reading the diff.

**Stable branch name (workflow)** — correct, the timestamp meant a fresh reconciliation PR every night with the old ones left to rot. Now a fixed branch that `create-pull-request` force-updates, so there is one open PR that keeps current.

**Exit-code docs (reconcile.py)** — correct, my own inaccuracy. Orphan-tag warnings return 1 even when no derived field drifted. Documented properly, including that `--check` fails on warnings too.

**Stale rationale (tag_lint.py)** — fair. Dropped the hard-coded spec counts and the spec name; the reasoning about describe-inheritance needing the AST is stable, those numbers aren't. Also noted explicitly why a `tag: [...]`-block-scoped check doesn't work either — a describe and its tests are separate blocks whose tags union.

One extra while I was in there: `RuntimeError` now prints as a message rather than a traceback. These are operational failures with actionable text (a broken spec, a missing project), not bugs, and a stack trace in the nightly log just buries the reason.

Verified after all changes: broken-spec and missing-project paths both exit 1 with the taxonomy untouched; happy path `--check` exits 0; tag-lint 26 specs / 0 problems; actionlint clean.

AndreiCautisanu:
Third round addressed in 85a8b528. Both comments describe the same real hole, and both are correct — narrowing `main()` to `except RuntimeError` last round left the operational exceptions that *aren't* `RuntimeError` subclasses escaping as raw tracebacks.

I reproduced both before fixing:

| trigger | before |
|---|---|
| malformed `taxonomy.yaml` | 16-line `yaml.scanner` traceback |
| `npx` not on `PATH` | 8-line `FileNotFoundError` traceback |

None of `subprocess.TimeoutExpired`, `OSError` or `yaml.YAMLError` subclasses `RuntimeError`, so the boundary never saw them.

**Fixed by converting at the source**, with the context each message actually needs:

- `subprocess.run` — `TimeoutExpired` reports the timeout; `OSError` adds *"is `npm ci` done in this project?"*, since that's the usual cause in CI.
- taxonomy read/parse — `OSError` reports the path, `YAMLError` the parser message. I also added a non-mapping guard: a scalar document would otherwise have failed much later inside `.get()` with a confusing `AttributeError`.
- `write_text` — `OSError` reports the path. That one sits *after* the boundary in `main()`, so it needed its own handler.

**On the "or broaden it to `Exception`?" alternative — deliberately not.** Anything outside this set is a bug, and a stack trace is the correct output for a bug. The narrow handler is the point: it distinguishes "your estate/config is broken, here's what to fix" from "this job has a defect". A blanket `except Exception` would turn every future bug into a terse one-liner and make the nightly job harder to debug, not easier.

One extra while in there: the taxonomy was being read twice — once to parse, once for the line-surgery pass. It now reuses the text the parse consumed, so a mid-run edit can't desynchronise the line numbers from the parsed structure.

Verified after the change — all four operational paths exit 1 with **zero traceback lines** and an actionable first line:

```
malformed YAML          exit=1  traceback_lines=0
  reconcile: /tmp/bad.yaml is not valid YAML: while parsing a flow sequence
YAML not a mapping      exit=1  traceback_lines=0
  reconcile: /tmp/scalar.yaml did not parse to a mapping (got str)
missing taxonomy file   exit=1  traceback_lines=0
  reconcile: could not read /tmp/does-not-exist.yaml: [Errno 2] ...
npx absent              exit=1  traceback_lines=0
  reconcile: could not run playwright --list: [Errno 2] ... 'npx'
```

Plus regression checks: the broken-spec and missing-project guards from last round still fire with the taxonomy byte-identical (`diff -q` clean), happy-path `--check` exits 0, tag-lint 26 specs / 0 problems.

natagh23 (approved):

Issue #1010, opened 2025-01-09
[FR]: UI - Ability to Annotate multiple traces at once within a Project


SrBliss:
### Proposal summary

Have the option to select several Traces within a Project and Annotate them all at once.

### Motivation

This is particularly useful when tracking an LLM-based app in production. After applying filters, you may want to annotate multiple traces at once, as doing so individually can be time-consuming and inefficient.

jverre:
This is a good suggestion, we will look into it

vincentkoc:
@Gmin2 you tagged/attempted the wrong PR and now two bounties linked to same PR. You will need to "re-attempt" once you have PR ready for this one.

Gmin2:
Hey @vincentkoc as I understand this feature request, it is asking for a way to select multiple traces at once and apply the same feedback scores to all of them in a single operation !

If i am right here ?

Because this does not exist currently as mentioned in the docs below

https://www.comet.com/docs/opik/tracing/annotate_traces#annotating-traces-through-the-ui

instead currently it has this 

![Image](https://github.com/user-attachments/assets/07a531a5-71ed-4423-b07e-191c80547f0a)

vincentkoc:
@Gmin2 docs are fine. Annotations are feedback scores and/or comments on a trace. See (taken from same docs link you gave) ![image](https://github.com/user-attachments/assets/7cfc256e-457c-47a5-a9e3-87c1e8c049f9)

jayantpranjal0:
@vincentkoc  I would like to work on this. Can you assign it to me?

jayantpranjal0:
@vincentkoc I don't see a annotate button on the ui, although I see seperate sections for feedback scores and comments. Should I add feature to add comments and feedback scores for multiple traces at once within a project through UI?

vincentkoc:
@jayantpranjal0 proceed

jayantpranjal0:
@vincentkoc just wanted to confirm, this feature is supposed to add something like this: 

<img width="1419" alt="Image" src="https://github.com/user-attachments/assets/a23a8c80-2520-41ce-b58b-75a9e0ce5f1c" />

Basically apart from annotations which are present as feedback scores I have also added a seperated comment form.
Should I just add the annotation, or also the comment form, since you mentioned annotations to also include comments, so is this feature supposed to just add comments from feedback score or also seperate individual comments. Anyways, I think it would be a useful feature to also have the seperated comments as a feature in this.

[...]

PR #7680, opened 2026-07-30
[OPIK-1010] [FE] feat: add batch annotate traces capability within project

mahenoorsalat:
Closes #1010. Adds UI button and dialog to annotate multiple selected traces at once within a project.


andrescrz:
Hi @mahenoorsalat 

Thanks for the contribution!

In addition, Could you please resolve the merge conflicts in the PR? Once that's done, we'll continue with the review.

Thanks!

aadereiko:
Contains malware code inside

PR #7689, opened 2026-07-31
[OPIK-7636] [BE] perf: dedup the feedback-scores trace_final CTE with argMax instead of FINAL

thiagohora:
## Details

Replaces `FINAL` with `GROUP BY` + `argMax` in the `trace_final` CTE of **`SELECT_FEEDBACK_SCORES_STATS` only**. `traces` is a `ReplacingMergeTree(last_updated_at)` ordered by `(workspace_id, project_id, id)`, so grouping on that key and taking a predicate's verdict from the greatest `last_updated_at` is what `FINAL` + predicate computes — and dropping `FINAL` also unblocks lazy materialization.

### Why only this template

The two stats templates behave oppositely, and the cause is projection width. This one projects `id, project_id` — both group keys — so its aggregate state is small and peak memory *improves*. It also re-evaluates `trace_final` from three scopes, which is where the saving comes from. `SELECT_TRACES_SPANS_STATS` projects seven per-version columns, so it holds a state per column per trace and memory *regresses*, worsening with group count.

Measured on production, all result-identical:

| Shape | Groups | `FINAL` | argMax | Verdict |
|---|---|---|---|---|
| **`feedback_scores` + searchText** | 132K | 10.33 GiB / 79,808 ms | **2.64 GiB / 25,250 ms** | **3.91x mem, 3.16x CPU better** |
| `traces_spans` + searchText | 2.51M | 4.04 GiB / 69,221 ms | 4.98 GiB / 62,277 ms | 1.11x CPU better, mem **+23%** |
| `traces_spans` + searchText | 5.38M | 6.11 GiB / 168,234 ms | 9.56 GiB / 138,057 ms | 1.22x CPU better, mem **+56%** |

All 13 `Code: 241` OOMs in the incident were on `feedback_scores`, where this takes peak memory 10.33 → 2.64 GiB — it attacks the failure mode directly. `traces_spans` peaked at 34.7 GiB against a ~47 GiB ceiling, so +56% there risks *creating* OOMs in a family that doesn't have them; its ~1.1–1.2x CPU isn't worth that. `SELECT_TRACES_SPANS_STATS` is therefore byte-identical to production on this branch, which also removes the apparatus it needed — the wrapper subquery, the tuple-valued projection, and the `ILLEGAL_AGGREGATION` aliasing hazard.

### Gates

**No filter slot may pull in a `LEFT JOIN`.** A join multiplies row versions inside each group, and a predicate on a joined alias can't be evaluated by `argMax` over `traces` versions at all.

**`search_text` must be present.** Without a heavy per-row scan to amortise it, the aggregation state is pure overhead in both CPU and memory. Net effect: the multi-project feedback-scores path only ever populates `filters`, so it never takes the branch.

### Filters stay index-visible via an id semi-join

Moving a predicate into `HAVING` puts it out of reach of the table's skip indexes (`thread_id` bloom, `source`/`environment` set, `created_at`/`last_updated_at` minmax). Without this, a `thread_id` filter plus searchText read **51.54 GiB / 22,416 ms** where `FINAL` reads **4.08 MiB / 108 ms** — 12,900x bytes, 207x CPU — and the searchText gate does not catch it, because the query has searchText. With the semi-join that shape is back to parity (4.11 MiB / 48 ms). The subquery is index-visible, and since it selects ids where *any* version matches it is a superset of the correct answer, which the `HAVING` narrows to the latest version.

A column-level veto was implemented first and reverted: selectivity is a property of the filter's **value**, not its column. `source IN ('sdk','unknown')` prunes nothing (98.2% of all prod rows) while `source = 'playground'` prunes to zero granules — same column, opposite outcomes.

### Correctness, all silent if missed

- **Non-key predicates move into `HAVING argMax(...)`.** Left row-level in `WHERE` they drop older versions from the group, so `argMax` reports the latest *surviving* version and content the current version no longer matches resurfaces.
- **`filters` is not vetoed even though `TraceField.GUARDRAILS` renders the joined `gagg` alias into it** — `FilterUtils` always sets `guardrails_filters` alongside, and that is what vetoes. Those two must stay in sync; it's the only `FilterStrategy.TRACE` field mapping to a joined alias.
- **The equivalence is scoped, not exact.** For a key whose latest `last_updated_at` is unique this returns exactly `FINAL`'s row. For a key with *tied* latest versions it returns one of the tied rows, and which one is **not contractually** `FINAL`'s choice, since ClickHouse doesn't specify `argMax`'s tie behaviour. An accepted, bounded difference: both forms always return a row that was really stored, so it can only shift which real version is reported, never invent one. No tie-breaker closes it — `ReplacingMergeTree` breaks the tie by insertion order, which is not a column. Ties are near-unreachable anyway: rows sharing a sort key within one INSERT collapse at part write, separate batches take distinct `now64(6)` values, and the four largest prod projects hold no duplicate versions.
- **`is_deleted`** is a prerequisite for the future `traces_local_v2` cutover: that table is `ReplacingMergeTree(last_updated_at, is_deleted)` and `FINAL` also drops soft-deleted rows, so the `HAVING` must then require `argMax(is_deleted, last_updated_at) = 0`. Not emitted today because the column doesn't exist on `traces`.

From the plan comparison: the primary-key search algorithm improves from `generic exclusion search` to `binary search` at identical granule selection.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves OPIK-7636

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: investigation, production benchmarking and plan analysis, the searchText gate, the filter semi-join, narrowing to one template, javadoc, and unit/integration test cases; drafted this description and the ticket
- Human verification: reviewed by @thiagoh

## Testing

`mvn -o test-compile`, `mvn -o surefire:test -Dtest=TraceDAOImplTest`, `mvn -o spotless:apply` (clean).

Unit — one `@ParameterizedTest` with an 18-row `@CsvSource` of `(slots, expected)`, replacing seven near-identical methods: `search_text` enables on its own and alongside anything harmless; its absence never suffices whatever else is set; a join-bearing slot vetoes anyway.

That is kept as a unit test rather than moved black-box because the gate selects *which SQL shape renders*, and both shapes return the same rows by design — a results-based test cannot observe whether it fired, and would pass with the gate inverted. The behaviour the two shapes must share is covered by the integration tests. SQL-string assertions were also tried and reverted: they couple the suite to how the query is written rather than what it returns.

Integration (`GetTracesByProjectResourceTest$TraceStatsLatestVersionOnly`) — behaviour over row versions: superseded version drops out; a version that starts matching appears; free-text search respects the latest version; and a filter alongside a search term evaluated on the latest version, using the trace's own id as the needle so the search matches every row version and the filter is the only discriminator.

Equivalence and cost were measured by replaying real rendered stats queries against production, variants back to back with identical settings, plus `EXPLAIN indexes=1` to compare index usage and granule selection rather than only totals. `SELECT_TRACES_SPANS_STATS` was verified byte-identical to production, so that family cannot have moved.

**Parameter binding — confirmed by CI.** The semi-join renders `<filters>` a second time, so the same bound parameter names appear twice in one statement, which I could not verify locally. On `2c09bd9e` (the commit that introduced it) every integration group passed except one, and that failure is `AgentInsightsJobsResourceTest` — a class-level error in a test with no trace-stats reference. The filter-plus-searchText integration case is what renders the semi-join, and it executed against a real ClickHouse, so the binding holds.

## Documentation

No user-facing or documentation change — internal query-shape optimisation. Both gates, why the scope is limited to this template, the semi-join and why a column list can't replace it, predicate placement, the guardrails coupling, the scope of the equivalence, and the `is_deleted` prerequisite are documented on `canDedupByArgMax`.


apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -2499,18 +2499,42 @@ AND entity_id IN (SELECT id FROM spans_data)
...
+                    project_id,
+                    id,
+                    <if(dedup_by_argmax)>
+                    argMax(thread_id, last_updated_at) AS latest_thread_id,
+                    argMax(if(input_length > 0, 1, 0), last_updated_at) AS input_count,
+                    argMax(if(output_length > 0, 1, 0), last_updated_at) AS output_count,
+                    argMax(if(metadata_length > 0, 1, 0), last_updated_at) AS metadata_count,
+                    argMax(length(tags), last_updated_at) AS tags_length,
+                    argMax(duration, last_updated_at) AS latest_duration,
+                    argMax(error_info, last_updated_at) AS latest_error_info
```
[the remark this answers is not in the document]
thiagohora: Fixed in c2bfb07 — thanks, this was the right call to push on.

The projection is now a single `argMax(tuple(...), last_updated_at)` unpacked by the enclosing SELECT, so all seven values come from one physical row. A tie now degrades to "which of the tied versions", which is exactly `FINAL`'s own behaviour, rather than "a mixture of them".

I didn't add a separate deterministic tie-breaker, because there is no per-version discriminator on the table to break the tie *with* — `traces` has no row identity beyond the dedup key `(workspace_id, project_id, id)`, so any tie-breaker would have to be another value column and would be just as arbitrary as what `FINAL` does today. Atomicity was the actual defect; determinism beyond that isn't something `FINAL` provides either.

Verified against production:
- result-identical to the per-column form (132,936 groups, same `cityHash64` sum)
- result byte-identical to `FINAL` on the full rendered stats query
- lower peak memory: 226 MiB vs 353 MiB, since it builds one aggregate state instead of seven

Worth noting for context: on current production data no project has any duplicate row versions at all (checked the four largest, 60M+ rows — `traces` is fully merged at rest), so this was latent rather than live. Still a real difference in kind, so it's fixed rather than documented.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -2499,18 +2499,42 @@ AND entity_id IN (SELECT id FROM spans_data)
...
                     thread_id,
                     if(input_length > 0, 1, 0) as input_count,
                     if(output_length > 0, 1, 0) as output_count,
                     if(metadata_length > 0, 1, 0) as metadata_count,
                     length(tags) as tags_length,
                     duration,
                     error_info
-                FROM traces final
+                    <endif>
+                FROM traces <if(!dedup_by_argmax)>final<endif>
```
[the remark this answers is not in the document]
thiagohora: Same root cause as the tie finding above, and fixed by the same commit (c2bfb07) — the projection is now one tuple-valued `argMax`, so a group can no longer be reassembled from different versions.

One part of the premise is worth separating out though: `GROUP BY` + `argMax` does *not* under-dedup relative to `FINAL` here. `traces` is `ReplacingMergeTree(last_updated_at)` with `ORDER BY (workspace_id, project_id, id)`, and the rewrite groups on exactly that tuple — the full dedup key, verified against the live DDL. Grouping on the dedup key and taking values from the greatest `last_updated_at` is what `FINAL` computes; unmerged parts don't change that, because rows for one id land in the same group regardless of which part they came from.

So keeping `FINAL` isn't required for dedup correctness. The genuine gap was that the *projection* was split across seven independent `argMax` calls, which is what allowed a synthesised row. That's now atomic.

Two things do still deliberately keep the `FINAL` form, for reasons a skip-index/dedup argument wouldn't cover:
- any filter slot that pulls in a `LEFT JOIN` (the join multiplies row versions inside a group, and a predicate on a joined alias can't be evaluated by `argMax` over `traces` versions at all)
- absence of `search_text` — without a heavy scan to amortise it, the `GROUP BY` aggregation state is pure overhead (measured up to 6.3x CPU and 85x memory worse on a 26M-trace project)

Both are enforced by `canDedupByArgMax` and covered by unit tests.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -2499,18 +2499,42 @@ AND entity_id IN (SELECT id FROM spans_data)
             )
             <endif>
             , trace_final AS (
+                <if(dedup_by_argmax)>
                 SELECT
                     workspace_id,
                     project_id,
```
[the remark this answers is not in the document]
thiagohora: Not taking this one, deliberately — extracting shared SQL fragments runs against a standing convention in this codebase: duplicated SQL templates are intentional, and the thing to factor is the *binding*, not the SQL text. Concatenating fragments across templates is how parameter/placeholder mismatches get introduced, and those fail the whole statement at runtime rather than at compile time.

The two `trace_final` blocks also aren't as parallel as they look, so a shared partial would need parameterising on more than "the projection":
- `SELECT_TRACES_SPANS_STATS` projects seven per-version values and needs the tuple-`argMax` + unpack wrapper
- `SELECT_FEEDBACK_SCORES_STATS` projects only `id, project_id`, so it has no wrapper and no tuple at all — just `GROUP BY` and the `HAVING`

What genuinely must stay in sync is the *gate*, and that is already single-sourced: both templates read the same `dedup_by_argmax` attribute, set from the one `canDedupByArgMax` method, which has 18 unit cases pinning its behaviour. Drift in the gate is the failure mode that would actually be silent, and it can't happen from one side changing.

Happy to revisit if the two blocks converge later.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -2499,18 +2499,38 @@ AND entity_id IN (SELECT id FROM spans_data)
...
+                FROM (
+                <endif>
+                SELECT
+                    workspace_id,
+                    project_id,
+                    id,
+                    <if(dedup_by_argmax)>
+                    argMax(tuple(thread_id, if(input_length > 0, 1, 0), if(output_length > 0, 1, 0),
+                                 if(metadata_length > 0, 1, 0), length(tags), duration, error_info),
+                           last_updated_at) AS latest
```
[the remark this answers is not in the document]
thiagohora: I went and measured this rather than reasoning about it, and the conclusion is: no tie-breaker, documented in a3af473.

**A matching tie-breaker isn't constructible.** `ReplacingMergeTree` resolves an equal-`ver` tie by falling back to the no-`ver` rule — *most recently inserted row wins*. Insertion order isn't a column, so no value-based tie-breaker can reproduce it. Ordering by the payload tuple would be deterministic, but it would deliberately pick a **different** row than `FINAL` in exactly the tie case — converting an unobserved divergence into a guaranteed one. That's worse against the goal you're describing.

**The two paths already agree, including adversarially.** On ClickHouse 26.3.16.16 with `ReplacingMergeTree(last_updated_at)`:

| Scenario | `FINAL` | `argMax` |
|---|---|---|
| 2 tied versions, later insert sorts *higher* by value | `B_inserted_second` | `B_inserted_second` |
| 2 tied versions, later insert sorts *lower* by value | `A_inserted_second` | `A_inserted_second` (5/5 runs, 4 threads) |
| 20 tied versions, insert order scrambled vs value order, 4 parts | `p12` (last inserted) | `p12` (3/3 runs, 8 threads) |

The second row is the one that matters: `argMax` picked the lexicographically *smaller* payload because it was inserted last, so it is not selecting by value — both take the last row seen at the maximum version.

**Ties are also close to unreachable on this table.** `last_updated_at` is `DateTime64(6)` (microseconds — `000019_reduce_trace_last_updated_at_to_micros`), and:
- rows sharing a sort key *within one INSERT* are collapsed when the part is written — verified, a 2-row insert of the same id stores 1 row, so a batch carrying the same trace twice cannot leave a tie behind
- separate batches take distinct `now64(6)` values — verified
- so what remains is a client supplying an explicit duplicate `lastUpdatedAt`, since `TraceDAO` honours a caller-provided value over `nowForBatch`
- across the four largest production projects (60M+ rows) there are **no duplicate row versions at all** — `traces` is fully merged at rest

So: unreachable in practice, no divergence reproducible when forced, and the proposed remedy would introduce the divergence it's meant to prevent. What c2bfb07 did fix was the real defect — per-column `argMax` could synthesise a row that never existed; the tuple form always yields one whole physical row, which is the same guarantee `FINAL` gives. Happy to revisit if you can construct a case where the two actually pick different rows.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -4349,6 +4395,89 @@ private Statement buildFeedbackStatementForProjects(Connection connection, List<
...
+     * for correctness; it does not soften the cost profile the gate below exists to avoid.
+     *
+     * <p>No further tie-breaker is added, deliberately. {@code ReplacingMergeTree} resolves an equal-{@code ver}
+     * tie by the "no {@code ver}" rule — the most recently inserted row wins — and insertion order is not a column,
+     * so no value-based tie-breaker can reproduce it. Adding one (ordering by the payload tuple, say) would be
+     * deterministic but would deliberately pick a <em>different</em> row than {@code FINAL} precisely in the tie
+     * case, turning an unobserved divergence into a guaranteed one. Empirically the two already agree: over 20
+     * tied versions inserted in scrambled order across multiple parts, and in the adversarial case where the
+     * last-inserted row sorts lowest by value, {@code FINAL} and {@code argMax} selected the same row on every
+     * run at 8 threads — both take the last row seen at the maximum version.
```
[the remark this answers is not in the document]
thiagohora: Agreed — taking your second option. Scoped the claim in eddc947: for a key whose latest `last_updated_at` is unique this returns exactly `FINAL`'s row; for a tied key it returns one of the tied rows, and which one is not contractually `FINAL`'s, since ClickHouse doesn't specify `argMax`'s tie behaviour. Documented as an accepted, bounded difference rather than exact equivalence.

No tie-breaker, because none can close it: `ReplacingMergeTree` breaks the tie by "most recently inserted wins", and insertion order isn't a column — any value-based tie-breaker would deterministically pick a *different* row than `FINAL` in exactly the tie case, guaranteeing the divergence instead of avoiding it.

Keeping `FINAL` for those keys isn't reachable either: identifying which keys have tied versions requires the dedup work the rewrite is avoiding, so it can't be decided at render time.

What bounds the risk: both forms always return a whole row that was really stored, so this can only ever shift *which* real version is reported, never synthesise one — that was the actual defect, fixed in c2bfb07. And the state is close to unreachable: rows sharing a sort key within one INSERT collapse at part write, separate batches take distinct `now64(6)` values, and no duplicate row versions exist in the largest production projects today.

apps/opik-backend/src/test/java/com/comet/opik/domain/TraceDAOImplTest.java
```diff
@@ -34,4 +41,104 @@ void firstOrLogFanOut__whenMoreThanOneRow__returnsFirstWithoutThrowing() {
...
+                "experiment_filters"
+        })
+        void canDedupByArgMax__whenIdScopedSlotAccompaniesSearchText__isAllowed(String idScopedSlot) {
+            assertThat(TraceDAOImpl.canDedupByArgMax(templateWithSlots("search_text", idScopedSlot))).isTrue();
+        }
+
+        @Test
+        void canDedupByArgMax__whenJoinBearingSlotAccompaniesSearchTextAndFilters__isVetoed() {
+            assertThat(TraceDAOImpl.canDedupByArgMax(
+                    templateWithSlots("search_text", "filters", "guardrails_filters"))).isFalse();
```
JetoPistola: 💡 **suggestion** | Testing

All 21 cases here assert the boolean gate. Nothing asserts the **rendered SQL** — and that is where the whole risk of this change lives.

Two specific things go unguarded:

1. In `SELECT_TRACES_SPANS_STATS`, the wrapper subquery's opening `FROM (` and its matching `)` are emitted from **two separate** `<if(dedup_by_argmax)>` blocks about 95 template lines apart. Nothing checks they stay balanced.
2. `SELECT_FEEDBACK_SCORES_STATS` emits `GROUP BY`/`HAVING` with **no** wrapper, because its `trace_final` projects only `id, project_id` — both in the group key. Correct today, but it means the two templates now have structurally different argMax shapes. A future edit that adds a projected column to template 2's `trace_final` would need the wrapper, and no test would catch its absence.

A pure-string assertion on `template.render()` covers both, costs ~10 lines, and needs no ClickHouse:

```java
@Test
void tracesSpansStats__whenDedupByArgMax__rendersBalancedGroupedSubquery() {
    var sql = renderStatsTemplate(TraceDAOImpl.SELECT_TRACES_SPANS_STATS, true);

    assertThat(sql).doesNotContain("FROM traces final");
    assertThat(sql).contains("GROUP BY workspace_id, project_id, id");
    assertThat(sql).contains("HAVING argMax(");
    assertThat(countOccurrences(sql, '(')).isEqualTo(countOccurrences(sql, ')'));
}
```

A second one asserting template 2 renders `GROUP BY`/`HAVING` *without* a wrapper would pin the asymmetry deliberately, so the next person to touch it sees it was a choice rather than an oversight.

Separately, `templateWithSlots` builds its template source from the slot names it is about to set (`"<" + slot + ">"`), so every attribute it reads is always declared. Production templates are not like that — `SELECT_TRACES_SPANS_STATS` never declares `traces_pushdown_filter` or `trace_thread_filters`, for instance. Current behaviour is fine (ST4 `getAttribute` returns null for undeclared-and-unset), but the helper's shape means the suite would pass identically if the gate only worked on self-declaring templates.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -4349,6 +4395,69 @@ private Statement buildFeedbackStatementForProjects(Connection connection, List<
...
+     */
+    @VisibleForTesting
+    static boolean canDedupByArgMax(ST template) {
+        return template.getAttribute("search_text") != null
+                && template.getAttribute("guardrails_filters") == null
+                && template.getAttribute("feedback_scores_empty_filters") == null
+                && template.getAttribute("span_feedback_scores_empty_filters") == null
+                && template.getAttribute("annotation_queue_filters") == null
+                && template.getAttribute("annotation_queue_id") == null;
+    }
```
JetoPistola, line 4433: ❓ **question** | Architecture

The `guardrails_filters` veto is load-bearing in a way that isn't visible from here, and I'd like to confirm you've considered pinning it.

`TraceField.GUARDRAILS` is a member of `FilterStrategy.TRACE` (`FilterQueryBuilder.java:605`) and maps to `gagg.guardrails_result` (`FilterQueryBuilder.java:131,369`) — a **joined alias**. So a guardrails filter renders into the `filters` slot, which this gate does *not* veto on. It works only because `FilterUtils.newTraceThreadFindTemplate:164` also sets `guardrails_filters` from `hasGuardrailsFilter` whenever that field is present, and *that* is what trips the veto.

The two flags can't desynchronize today, but the coupling is implicit and cross-file. If someone later narrows `hasGuardrailsFilter` while leaving `TraceField.GUARDRAILS` in `FilterStrategy.TRACE`, the result is `HAVING argMax(gagg.guardrails_result …, last_updated_at)` over an alias with no join in scope — and given how much of this javadoc is devoted to failure modes that are silent if missed, this one belongs in that list.

I checked the other 20 entries of `TRACE_FIELDS_MAP` — `GUARDRAILS` is the only `FilterStrategy.TRACE` field that maps to a joined alias, so this is the single instance. A one-line javadoc note under the join paragraph naming the dependency would be enough:

> `filters` is not vetoed even though `TraceField.GUARDRAILS` renders the joined `gagg` alias into it — `FilterUtils` always sets `guardrails_filters` alongside, and that is what vetoes. Keep those two in sync.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -2579,6 +2601,13 @@ id IN (SELECT trace_id FROM sfsc WHERE sfsc.span_feedback_scores_count = 0)
                     id NOT IN (SELECT trace_id FROM sfsc)
                 )
                 <endif>
+                <if(dedup_by_argmax)>
+                GROUP BY workspace_id, project_id, id
+                <if(filters || search_text)>
+                HAVING argMax(<if(filters)>(<filters>)<endif><if(filters && search_text)> AND <endif><if(search_text)>(<search_text>)<endif>, last_updated_at)
```
JetoPistola: 🧹 **nit** | Style

`<if(filters || search_text)>` can never be false when this block renders: `canDedupByArgMax` requires `search_text != null`, so `dedup_by_argmax` implies `search_text`. The guard reads as though a `dedup_by_argmax` shape with neither slot were reachable, which would be a `GROUP BY` with no `HAVING` — exactly the pure-overhead case the gate exists to prevent.

Same in the second template at line 2967.

Happy to leave it as belt-and-braces if you'd rather the template not depend on the gate's internals — but in that case a short comment saying so would help, since the PR description explicitly calls the no-`HAVING` shape out as something the gate rules out.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/GetTracesByProjectResourceTest.java
```diff
@@ -5816,4 +5816,173 @@ void searchTraces__whenSearchHasNoMatch__thenReturnEmpty() {
...
+                    TraceUpdate.builder()
+                            .projectName(projectName)
+                            .name(RandomStringUtils.secure().nextAlphanumeric(10))
+                            .build(),
+                    API_KEY, TEST_WORKSPACE);
+
+            assertThat(traceCount(projectName, nameEquals(matchedName), idSearch)).isZero();
+        }
+
+        private long traceCount(String projectName, List<? extends TraceFilter> filters, String search) {
```
JetoPistola: 🧹 **nit** | Style

Three overloaded `traceCount` helpers in this nested class (lines 5831, 5907, and here at 5975), interleaved between the `@Test` methods rather than grouped. The read order ends up helper, helper, test, test, **helper**, test, test, **helper** — and in a 5,900-line file, a reader hitting `traceCount(projectName, needle)` has to scan forward past the end of a test to find the overload that resolves it.

The three bodies also share a byte-identical extraction tail, and the two-arg forms are just the three-arg one with a fixed argument. One helper covers all four call sites:

```java
private long traceCount(String projectName, List<? extends TraceFilter> filters, String search) {
    var stats = traceResourceClient.getTraceStats(projectName, null, API_KEY, TEST_WORKSPACE, filters,
            search == null ? Map.of() : Map.of("search", search));

    return stats.stats().stream()
            .filter(stat -> "trace_count".equals(stat.getName()))
            .map(stat -> (Number) stat.getValue())
            .mapToLong(Number::longValue)
            .findFirst()
            .orElse(0L);
}
```

That drops two methods and 14 duplicated lines, and removes the overload-resolution footgun between `traceCount(p, someString)` and a single-element filter list.

Worth saying the tests themselves are well chosen — especially the javadoc explaining why the search case specifically probes the `thread_id`/`error_info`/`duration` aliasing collision. This is purely about where the helpers sit.

JetoPistola:
👋 **Review summary**

**What looks good**

- **The investigation is the deliverable here, not just the diff.** Two gating conditions, both justified with measured production numbers across five query shapes, and a table that leads with the cases where the rewrite *loses* (85x memory at 26.4M groups). Presenting the counter-evidence first is the right instinct and it's what makes the gate credible.
- **The tuple-valued `argMax` fix is genuinely subtle.** Per-column `argMax` synthesising a row that never existed is the kind of thing that ships and then produces one wrong stat a quarter. Catching that `duration`, `thread_id` and `error_info` are both projected *and* read by the eight-column search clause — and that aliasing them to their own names nests `argMax` in `argMax` — is careful work.
- **Predicate placement is reasoned through, not pattern-matched.** Keeping the dedup-key predicates and the four `id IN (...)` slots in `WHERE` while moving only value-based predicates into `HAVING` is exactly right; I checked all four subqueries and none references a per-row `traces` column, so none can strand a group on a stale version.
- **The equivalence claim was scoped rather than defended.** Walking back to "one of the tied rows, not contractually `FINAL`'s choice" instead of arguing the tie away is the honest call, and the reasoning for why no tie-breaker can close it (insertion order isn't a column) is correct.
- **The `is_deleted` note for the `traces_local_v2` cutover** is real forward-looking value — that one would have been silent and would have resurfaced soft-deleted traces.
- Integration tests cover the semantics that matter (superseded version dropping out, starting to match, search-plus-filter with the trace id as needle so the filter is the only discriminator).

**Overall**

The approach is sound and I couldn't break it. I checked the failure modes most likely to be silent — soft-delete resurrection, the `id IN (...)` slots, bind-parameter reuse, the join veto list's completeness, and gate-call ordering at all four sites — and all hold. The one non-obvious save is that `TraceField.GUARDRAILS` renders a joined alias into the un-vetoed `filters` slot and is rescued only by `FilterUtils` setting `guardrails_filters` alongside it; that's correct today but implicit across two files, which is my question below.

Two suggestions, both about pinning things that are currently correct by convention. The tests assert the gate but never the rendered SQL, and the two templates now have structurally different argMax shapes (one wrapped in a subquery, one not) with nothing holding that difference in place. And the seven-element tuple is unpacked positionally across the `FROM (` boundary — I measured the named-tuple form as result-identical (83,994 groups, same hash) at identical memory, and it turns a mid-tuple column insertion from a silent wrong-stats bug into a parse-time error.

**Inline comments**: 2 suggestions, 1 question, 2 nits — nothing blocking.

_(Edited to fold in a fifth comment on named vs positional tuple access, posted after this summary.)_

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
JetoPistola: 💡 **suggestion** | Maintainability

Consider a **named** tuple here instead of positional `latest.1 … latest.7`.

Positional unpacking is the house style — `entries[1].4 AS source` appears ~14 times across TraceDAO/ThreadDAO/SpanDAO, and `ExperimentDAO.java:1887` (`ranked[1].3 AS project_id`) is the same argMax-dedup shape. So this isn't novel and I'm not asking you to change the convention.

But this instance is a step past the precedents in two ways: it's **seven** positions rather than one to four, and the `tuple(...)` construction is separated from the unpack by the `FROM (` boundary plus two `<if(dedup_by_argmax)>` blocks — about 14 template lines. The reader checking whether `latest.5` is `tags_length` has to count commas across a subquery boundary in a StringTemplate string.

ClickHouse named tuples fix that, and this codebase already uses them — `Array(Tuple(key String, type String))` in migrations `000044`/`000101`, and `CAST(…, 'Tuple(value String, passed String, reason String)')` in `ExperimentItemDAO`, `ExperimentAggregatesDAO` and `DatasetItemVersionDAO`.

```sql
argMax(tuple(thread_id, if(input_length > 0, 1, 0), if(output_length > 0, 1, 0),
             if(metadata_length > 0, 1, 0), length(tags), duration, error_info)
       ::Tuple(thread_id String, input_count UInt8, output_count UInt8,
               metadata_count UInt8, tags_length UInt64,
               duration Nullable(Float64), error_info String),
       last_updated_at) AS latest
```

and the outer SELECT becomes `latest.thread_id AS thread_id`, `latest.input_count AS input_count`, and so on.

I ran this against ClickHouse 26.3.16.16 on the real `traces` table rather than reasoning about it:

- **Result-identical** to the positional form — 83,994 groups, same `cityHash64` over the full sorted result set.
- **No cost.** Averaged over 3 runs each: 34.23 MiB / 28 ms named vs 34.22 MiB / 34 ms positional. The `::Tuple(...)` is a compile-time type annotation, not a runtime conversion — the projected expression already types as `Tuple(String, UInt8, UInt8, UInt8, UInt64, Nullable(Float64), String)`, so the cast only attaches names.
- Works with the real eight-column search clause in `HAVING` and with `duration` being `Nullable(Float64)`.

The reason I think it's worth the churn is the failure mode. I reordered the tuple members and re-ran both:

| Form | After reordering the tuple |
|---|---|
| `latest.1` | returned `1.5 / Float64` — silently handed back `duration` where `thread_id` was expected |
| `latest.thread_id` | still correct; a typo'd name is `UNKNOWN_IDENTIFIER … Maybe you meant: ['latest.thread_id']` |

So inserting a column into the middle of this 7-element tuple shifts every position after it, and positional access absorbs that silently — wrong stats, no error, and nothing in the current test suite would catch it (which ties into the rendered-SQL point I left on `TraceDAOImplTest`). Named access either keeps working or fails at parse time.

Given how much of the javadoc is devoted to failure modes that are silent if missed, this feels like one worth closing structurally rather than by convention. Your call though — if you'd rather stay consistent with the surrounding positional style, that's a legitimate reason to decline, and the code is correct either way.

To be clear on scope: I'm suggesting this for the new 7-element tuple only, not proposing a sweep of the existing call sites.

thiagohora: Taken — done in 53ff687. The failure mode sold it: I reordered the members and confirmed `latest.1` silently returned `duration` where `thread_id` was expected, while `latest.thread_id` stayed correct and a typo is a parse error. Also re-ran the equivalence check on prod traces myself before changing it — identical results, and the `::Tuple(...)` only attaches names to a type the expression already has.

Agreed on scope: new 7-element tuple only, existing positional call sites untouched.

apps/opik-backend/src/test/java/com/comet/opik/domain/TraceDAOImplTest.java
thiagohora: Revisited this — we're not going to assert the rendered SQL. Reverted in be715c65 (the `@VisibleForTesting` widening of the two stats templates is reverted with it).

The reasoning: a string assertion on `template.render()` couples the suite to how the SQL is written rather than what it returns, so it fails on harmless rewording and passes on any bug that keeps the text shape. The failure modes you named are better caught black-box, and already are — the integration tests execute the query, so a one-sided edit to the wrapper produces unparseable SQL and fails outright, and a missing wrapper or mis-ordered projection shows up as wrong stats.

The named tuple from your other comment covers the specific risk you were worried about there more directly than a test could: a mid-tuple column insertion is now a parse error rather than silently shifted positions.

Fair point on `templateWithSlots` being self-declaring — noted, though with the SQL-text tests gone the guard suite is deliberately only testing the boolean, and the behaviour is covered by the integration cases.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
thiagohora: Confirmed your reading before documenting it: `TraceField.GUARDRAILS` is in the `FilterStrategy.TRACE` list and maps to `gagg.guardrails_result`, and `FilterUtils:163` sets `guardrails_filters` from `hasGuardrailsFilter` whenever the field is present — so the veto fires via that flag, not via `filters`. Documented in 53ff687, including that it is the only `FilterStrategy.TRACE` field mapping to a joined alias.

---

Your question pushed me to check the execution plans, which I had not done, and that found a **regression the searchText gate did not prevent**. Fixed in 2c09bd9e.

Moving value-based predicates into `HAVING argMax(...)` put them out of reach of the table's skip indexes (`thread_id` bloom, `source`/`environment` set, `created_at`/`last_updated_at` minmax). On a 2.5M-trace project, a `thread_id` filter plus searchText:

| Form | Granules | Bytes | CPU |
|---|---|---|---|
| `FINAL` — predicate in `WHERE` | **1 / 10,169** | 4.08 MiB | 108 ms |
| argMax — predicate in `HAVING` | 10,169 / 10,169 | **51.54 GiB** | **22,416 ms** |

12,900x bytes, 207x CPU. The gate passed it because the query does have searchText.

**Fix:** filters are now also injected as `id IN (SELECT id FROM traces WHERE <filters>)`. The subquery is index-visible, and since it selects ids where *any* version matches it is a superset of the correct answer, which the `HAVING` narrows to the latest version — pruning restored without touching correctness.

I implemented a column-level veto first and reverted it, because selectivity turns out to be a property of the filter's **value**, not its column:

| `source` filter | Granules after skip index |
|---|---|
| `IN ('sdk','unknown')` — UI default | 10,171 / 10,171 (nothing) |
| `= 'playground'` | **0** / 10,171 |
| `= 'evaluator'` | **1** / 10,171 |

Those two default values are 98.2% of all production rows, which is why they prune nothing, while four of the six `Enum8` values are selective. A column list would wrongly veto `environment = <dominant>` and wrongly admit `source = <rare>`.

**Regression data for the fix**, all result-identical:
- prunable shape (`thread_id` + searchText): **4.11 MiB / 48 ms** versus `FINAL` at 4.08 MiB / 108 ms — regression gone, marginally better than `FINAL`
- production shape (`source` + searchText), 3 runs each: **55.60 GiB / 4.98 GiB / 62,277 ms** with the semi-join, 55.32 GiB / 4.33 GiB / 59,993 ms without, 60.86 GiB / 4.04 GiB / 69,221 ms for `FINAL` — so the semi-join costs ~4% CPU and ~15% memory and keeps the win

Also from the plan comparison: the primary-key search algorithm improves from `generic exclusion search` to `binary search` under argMax at identical granule selection, and the positional vs named tuple plans are structurally identical, so the `::Tuple(...)` is confirmed free rather than assumed.

One thing CI needs to confirm rather than me: the semi-join renders `<filters>` a second time, so the same bound parameter names now appear twice in one statement. Existing templates reuse `:workspace_id` across CTEs so this should be fine, but the integration tests are what will prove the binding.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
thiagohora: Right that it cannot be false today. Keeping it, with the reasoning recorded in the javadoc rather than in the SQL string (house convention is no comments inside the SQL templates).

The reason it is not merely belt-and-braces: if the gate were ever relaxed to admit a filters-only shape, removing this would render `HAVING argMax(, last_updated_at)` — invalid SQL — rather than the bare `GROUP BY` you described. So it guards a real breakage in the shape adjacent to the one the gate rules out.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/GetTracesByProjectResourceTest.java
thiagohora: Done in 53ff687 — collapsed to the single three-arg helper with a null-tolerant search, and moved it up beside `nameEquals` so both helpers sit ahead of the tests rather than interleaved. Drops two methods and the overload-resolution footgun between a search string and a single-element filter list.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -2917,8 +2924,21 @@ AND entity_id IN (SELECT id FROM spans_data)
...
+                <endif>
+                <if(dedup_by_argmax && filters)>
+                AND id IN (
+                    SELECT id
+                    FROM traces
+                    WHERE workspace_id = :workspace_id
+                    AND project_id IN :project_ids
+                    <if(uuid_from_time)>AND id >= :uuid_from_time<endif>
+                    <if(uuid_to_time)>AND id \\<= :uuid_to_time<endif>
+                    AND <filters>
```
andrescrz, line 2896: Please double check this filtering works appropriately in this particular subquery and in the query as a whole. This subquery only retrieves the id field, but filters on other fields without properly deduping might return wrong results.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -4387,6 +4420,93 @@ private Statement buildFeedbackStatementForProjects(Connection connection, List<
         return statement;
     }
 
+    /**
+     * Whether {@code trace_final} can dedup with {@code GROUP BY} + {@code argMax} instead of {@code FINAL}.
```
andrescrz: Minor: very large and verbose javadoc. Great if this can be kept correct and complete, but concise, without superflous information and only keeping what matters for the future.

apps/opik-backend/src/test/java/com/comet/opik/domain/TraceDAOImplTest.java
```diff
@@ -34,4 +41,105 @@ void firstOrLogFanOut__whenMoreThanOneRow__returnsFirstWithoutThrowing() {
...
+     * {@code HAVING} can discard anything, so without a heavy per-row scan to amortise it the state is pure
+     * overhead and the form is measurably slower and far hungrier for memory (OPIK-7636).</li>
+     * </ul>
+     *
+     * <p>The join-bearing cases therefore set {@code search_text} as well, so a veto proves the join check fired
+     * rather than the searchText gate.
+     */
+    @Nested
+    @DisplayName("argMax dedup guard")
+    class ArgMaxDedupGuard {
```
andrescrz, line 56: Minor: double check the need to unit test all this. Better favour high value black box tests. It'd be desirable to cover all conditions from there, which would make this redundant and our coverage more valuable (at the expense of slower tests).

apps/opik-backend/src/test/java/com/comet/opik/domain/TraceDAOImplTest.java
```diff
@@ -34,4 +41,105 @@ void firstOrLogFanOut__whenMoreThanOneRow__returnsFirstWithoutThrowing() {
...
+        private ST templateWithSlots(String... slots) {
+            var template = TemplateUtils.newST(
+                    Arrays.stream(slots).map(slot -> "<" + slot + ">").collect(Collectors.joining()));
+            Arrays.stream(slots).forEach(slot -> template.add(slot, "anything"));
+
+            return template;
+        }
+
+        @Test
+        void canDedupByArgMax__whenSearchTextPresent__isAllowed() {
```
andrescrz: Minor: most of these tests seem a parameterise version of the same test (slots and expected value).

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
thiagohora: This is the copy that survives — `SELECT_TRACES_SPANS_STATS` was reverted to `FINAL` in 1ab4002, so the semi-join now exists only here.

Same answer as above: the subquery needs no dedup because it only has to be a superset. It selects ids where **any** version matches; the correct answer is ids whose **latest** version matches, which is a subset, so nothing wanted can be dropped. The `HAVING argMax(..., last_updated_at)` evaluates the filter on the latest version and removes the ids that matched only on a stale one.

Verified both directions: a trace whose v1 matches and whose latest does not is excluded by both forms; a trace whose v1 does not match and whose latest does is included by both. Results byte-identical to `FINAL` on the production replays.

The reason it is there at all: a predicate inside `HAVING` is invisible to the table's skip indexes, and a `thread_id` filter plus searchText read 51.54 GiB / 22,416 ms without it versus 4.08 MiB / 108 ms under `FINAL`. With the semi-join that shape is back to parity.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
thiagohora: Trimmed in c003ddf — 73 lines down to 48, and it had already come down from 80 earlier.

What is left is only what constrains a future edit: the two gates, the `guardrails_filters` coupling (implicit across two files, and silent if broken), predicate placement, why the semi-join exists and must stay in step with the `HAVING`, the scoped equivalence, and the `is_deleted` prerequisite for the `traces_local_v2` cutover. All measurements now point at OPIK-7636 rather than being restated inline.

apps/opik-backend/src/test/java/com/comet/opik/domain/TraceDAOImplTest.java
thiagohora: Partly taken — collapsed to a single parameterised method in c003ddf (see your comment below), so the unit surface is now one test rather than seven.

I did not move it black-box, for a reason worth stating: the gate selects which SQL shape renders, and both shapes return the same rows by design. A results-based test therefore cannot observe whether the gate fired — it would pass identically with the gate inverted. What black-box tests can cover is the behaviour the two shapes must share, and that is already covered by the integration cases (superseded version drops out, a version that starts matching appears, searchText respects the latest version, and a filter alongside searchText evaluated on the latest version).

So the split is deliberate: correctness black-box, shape selection in the one unit test. If you would rather drop the unit test entirely and accept that the gate is unpinned, say so and I will.

apps/opik-backend/src/test/java/com/comet/opik/domain/TraceDAOImplTest.java
thiagohora: Correct, they were — collapsed in c003ddf into one `@ParameterizedTest` with an 18-row `@CsvSource` of `(slots, expected)`, replacing seven methods. Same coverage, and the table now reads as three groups: search_text enables, search_text absent never suffices, a join-bearing slot vetoes anyway.

andrescrz (approved):
LGTM

PR #7695, opened 2026-08-02
[NA] [PYBE] fix: pin litellm to 1.94.1 to unbreak the image build

liyaka:
## Details

`main` has two red workflows since 2026-08-02, both from the same cause and neither from a change on
our side:

| Workflow | Fails at | Result |
|---|---|---|
| `Build Opik Docker Images` | `Build and Push Docker Image` (amd64 **and** arm64) | no images published |
| `E2E Tests - Post Merge (v2)` | `🔨 Build and start Opik server` | suite never runs; zero tests executed |

`litellm` is not pinned — it arrives transitively through `opik-optimizer==3.1.1` — so the build
resolves whatever is newest, and `litellm 1.95.0` was published that morning. Pinning it to
`1.94.1`, the version the last green build used, restores both.

Why the new release breaks the image: `litellm` publishes no musllinux or pure-python wheel, so this
Alpine-based image compiles it from the sdist. `1.95.0` added a Rust extension
(`litellm-rust/crates/python-bridge`) whose dependencies are the AWS SDK crates, with an MSRV of
rustc 1.94.1, while the builder installs Rust from Alpine's repos (rustc 1.91.1):

```
× Failed to build `litellm==1.95.0`
╰─▶ Call to `maturin.build_wheel` failed (exit status: 1)
    error: rustc 1.91.1 is not supported by the following packages:
      aws-config@1.9.0            requires rustc 1.94.1
      aws-credential-types@1.3.0  requires rustc 1.94.1
      aws-runtime@1.8.1           requires rustc 1.94.1
      aws-sdk-sts@1.108.0         requires rustc 1.94.1
      aws-sigv4@1.5.1             requires rustc 1.94.1
      aws-smithy-async@1.3.0      requires rustc 1.94.1
    💥 maturin failed

hint: `litellm` (v1.95.0) was included because `opik-optimizer` (v3.1.1) depends on `litellm`
```

The Dockerfile comment above `apk add … rust cargo` states the toolchain is there "only for
compiling native wheels (rust/cargo for pydantic-core)", so a transitive dependency pulling a whole
Rust SDK was never in scope for it.

Follow-up worth considering separately: install a newer rustc in the builder stage (Alpine
edge/testing or rustup) so the image is not held to whatever MSRV upstream crates adopt next. This
PR deliberately does the smaller thing — an unpinned transitive dependency in a Docker build is the
defect that let an upstream release break `main` unannounced.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- OPIK-

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: diagnosis of the two failing workflows and the one-line pin plus its comment.
- Human verification: reviewed by the author, who also triggered the build and E2E runs on this
  branch.

## Testing

Version chosen by comparing the failing and last-green builds rather than by guessing:

- Last green run of `Build Opik Docker Images` on `main` (run `30645351801`, 2026-07-31) installed
  `litellm==1.94.1`; the first red run (`30739293214`, 2026-08-02) resolved `litellm==1.95.0`.
- PyPI shows `1.95.0` uploaded 2026-08-02, matching the first failure the same day.
- Neither `1.94.1` nor `1.95.0` ships a musllinux or `py3-none-any` wheel, confirming both are built
  from sdist on this image and that the difference is the crate graph, not packaging.
- `opik-optimizer==3.1.1` declares `litellm!=1.81.*,!=1.82.*,!=1.83.0..6,!=1.92.*,>=1.79.2`, so
  `1.94.1` satisfies its constraint.
- The same `maturin`/rustc failure appears in the post-merge E2E log (run `30739293070`), which is
  why both workflows are covered by this one-line change.

Both failing workflows were then re-run on this branch and pass:

- `Build Opik Docker Images` (run `30743029800`) — `amd64 opik-python-backend`, `arm64
  opik-python-backend` and `merge (regular)` all green, i.e. the two jobs that fail on `main`.
- `E2E Tests v2 (t1)` (run `30743068524`) — 13m36s, green through `🔨 Build and start Opik server`,
  `🔍 Verify Docker pods are running`, `🏥 Verify backend health` and `🧪 Run E2E test suite`. On
  `main` the run dies at the build step and the suite never executes.

## Documentation

None — the pin carries an inline comment explaining why it exists and what has to be true before it
can be lifted.

### DIFF
--- a/apps/opik-python-backend/requirements.txt
+++ b/apps/opik-python-backend/requirements.txt
@@ -35,5 +35,9 @@ opik-optimizer==3.1.1
 # template markers and the reflection_prompt_template kwarg, neither of which
 # exists on 0.0.x. See OPIK-7510 / OPIK-7639.
 gepa==0.0.17
+# Transitive via opik-optimizer, pinned because this image builds it from source: litellm
+# publishes no musllinux or pure-python wheel, and 1.95.0 raised its Rust dependencies to
+# aws-* crates requiring rustc 1.94.1, while the Alpine builder ships rustc 1.91.1.
+litellm==1.94.1
 pyrate-limiter>=3.0.0,<4.0.0
 python-jsonpath>=2.0.0,<3.0.0


thiagohora (approved):

JetoPistola (approved):
Approving. Verified the diagnosis independently and confirmed the fix is green on both previously-red workflows. Full reasoning in the summary comment below — including a follow-up ticket for the root cause (OPIK-7703).

JetoPistola:
👋 **Review summary**

**What looks good**

- The diagnosis is evidence-based rather than guessed — version chosen by diffing the last-green run (`30645351801`) against the first-red one (`30739293214`), cross-checked against the PyPI upload timestamp. That's the right method for an "upstream broke us" fix.
- Independently verified every premise in the description and they all hold: neither 1.94.1 nor 1.95.0 ships a musllinux or `py3-none-any` wheel (manylinux_2_28 + win_amd64 only), so this image genuinely builds from sdist; 1.95.0 was uploaded `2026-08-02T02:51:53`, matching the first failure the same morning; and 1.94.1 satisfies `opik-optimizer==3.1.1`'s `litellm>=1.79.2,!=1.81.*,!=1.82.*,!=1.83.0..6,!=1.92.*`.
- Also confirmed the rustc claim at the source — `docker run --rm docker:29.5.1` + `apk add rust` gives `rustc 1.91.1 (Alpine Linux Rust 1.91.1-r2)`, exactly the version in the error output.
- The inline comment states the *exit condition*, not just the reason. A pin that says what has to become true before it can be lifted is the difference between a fix and a future mystery.
- Empirically confirmed rather than only reasoned: `build-python-backend` (amd64 and arm64) and `E2E v2 Tests (t1)` are all green on this head — precisely the two workflows that were red.

**On the choice of an exact pin over a range**

Worth recording, since it'll come up again. A floor alone (`>=1.94.1`) would not have fixed this — the break came from resolving *newer*, so an open ceiling just lets 1.95.0 back in. A capped range (`>=1.94.1,<1.95`) would work, but litellm ships patches constantly (1.89 got six, 1.90 got six, 1.91 got four) and `opik_optimizer/pyproject.toml` already excludes `1.81.*`, `1.82.*`, `1.83.0–6` and `1.92.*` for four separate breakages. Open ceilings on this particular dependency have a track record. Exact pinning is also the house style of this manifest — `opik`, `opik-optimizer`, `gepa`, `tiktoken`, `redis` are all `==`. So `==` is the right call here.

**Overall**

Correct fix at the right size, and the scope discipline is right — unbreak `main` now, defer the structural change. I checked whether the pin could be circumvented or go stale unnoticed: `requirements.txt` is the only manifest this image installs and every direct dependency is already exact-pinned, so nothing else re-resolves `litellm` here; and `.github/dependabot.yml` deliberately excludes `apps/**` production-image manifests (OPIK-7494), so the pin won't be auto-bumped back out.

**One note on the proposed follow-up**, since I measured it. The description suggests installing a newer rustc in the builder. That does work — Alpine edge has `rust 1.97.0` vs 3.23's `1.91.1`, and it builds 1.95.0 successfully — but it's the weakest of the available options:

| Approach | Result | Cost |
|---|---|---|
| Pin `litellm==1.94.1` (this PR) | ✅ works | manual bumps; covers one package |
| Newer rustc from Alpine edge | ✅ works | **160s** compile, and pulls a rolling repo into a production image |
| glibc base (`python:3.12-slim`) | ✅ works | **5s**, prebuilt wheel, **no Rust toolchain needed at all** |

On a glibc base the release that broke `main` is a complete non-event — it installs as a binary wheel in 5 seconds with no compiler present. The musl base is what converts routine upstream releases into source builds; litellm was the trigger, not the cause. Note pydantic-core, tiktoken, numpy and pandas *all* publish musllinux wheels — litellm is the outlier, which is why this hadn't bitten before.

Filed **OPIK-7703** for the root cause (glibc base + locking the transitive graph, with the rustc route explicitly evaluated and rejected). The `## Issues` section here currently reads `OPIK-` with no number, if you want to point it somewhere.

None of that is a reason to hold this PR — a hotfix should be small, and this one is well-evidenced and verified. Approving.

**Inline comments**: None — nothing to flag.

PR #7698, opened 2026-08-02
[OPIK-7706] [DOCS] docs: document AssertJ whole-object assertion conventions

JetoPistola:
## Details

image

Documents `usingRecursiveComparison()` as the default for object-to-object equality in backend Java tests, and `containsExactly` / `containsExactlyInAnyOrder` for collections. The motivation is correctness rather than brevity: when a field is later added to a record, a field-by-field test keeps passing while silently not covering the new field — nothing fails and nothing flags it, so assertion coverage erodes with every model change.

The convention was already in practice (45 of 431 backend test files use recursive comparison) but was never written down, so it wasn't applied consistently. The `## Assertion Patterns` section of the backend testing skill didn't mention it at all.

- Partial comparisons must use `ignoringFields(...)` rather than dropped assertions, so exclusions stay explicit and reviewable instead of being implied by whichever assertions someone chose not to write.
- Inexact types (`BigDecimal`, `double`) get a comparator rather than an exclusion, so the field stays covered instead of being silently dropped to stop flakiness.
- Documents where per-field assertions remain correct — literal spot checks, non-equality semantics (`isAfter`, `isNotBlank`), and deliberately narrow assertions on an ignored field — so the rule doesn't read as absolute.

Existing tests are intentionally not migrated. That is separate tech debt, to be picked up when someone is already working in the code under test or when flagged in review.

## Change checklist
- [x] User facing
- [x] Documentation update

## Issues

- Resolves OPIK-7706

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: full implementation (research, drafting, and verification of the documented conventions)
- Human verification: author review of the rendered guidance and of every cited symbol

## Testing

Documentation-only change — no test or production code is touched, so no test suite exercises it. Verification focused on making sure the guidance is accurate and the examples are real:

- `pre-commit run --files .agents/skills/opik-backend/testing.md` — passes (exit 0).
- Confirmed every cited symbol exists in the repo rather than being invented: `StatsUtils.bigDecimalComparator`, `StatsUtils.closeToEpsilonComparator`, and `EXPERIMENT_IGNORED_FIELDS` in `ExperimentTestAssertions`.
- Confirmed each documented AssertJ API is supported by the version the backend actually resolves (3.27.7), including `withComparatorForFields`, `ignoringCollectionOrderInFields`, and `containsExactlyInAnyOrderElementsOf`.
- Confirmed the documented APIs are already in real use in the tree (`containsExactlyInAnyOrderElementsOf` in 14 files, `ignoringCollectionOrderInFields` in 5), so examples reflect house style rather than a style imported from elsewhere.
- Not run: Maven, frontend, and SDK suites. No Java, TypeScript, or Python source changed.

Worth flagging for the reviewer: the ignore-then-assert-narrowly pattern in `ExperimentTestAssertions` initially looked like redundancy, but the re-asserted fields are the ones listed in `EXPERIMENT_IGNORED_FIELDS`, so recursive comparison skips them and those lines are what actually covers them. It's deliberate — it lets `lastUpdatedAt` mean `==` or `isAfter` per scenario — and is documented as a valid idiom rather than something to clean up.

## Documentation

This PR is itself the documentation change: `.agents/skills/opik-backend/testing.md`, the backend testing skill. No user-facing or external docs are affected.

One open question left for review: whether a `.claude/rules/apps/opik-backend/` pointer should be added so this guidance loads automatically for backend work, matching how `.claude/rules/sdks/python/testing.md` points at the Python SDK testing skill. That is a routing change beyond this ticket's scope and is not included here.

### DIFF
--- a/.agents/skills/opik-backend/testing.md
+++ b/.agents/skills/opik-backend/testing.md
@@ -284,6 +284,27 @@ assertThat(actual)
 `containsExactly` asserts size and content together — `hasSize` plus per-index checks does not,
 and lets an extra element through.
 
+These `containsExactly*` variants compare elements with the element type's own `equals`, which is
+what you want for exact-valued models. When the elements carry `BigDecimal` or `double`, the same
+exception that justifies a comparator on a single object applies per element — otherwise a
+difference in numeric scale fails the assertion even though the values are equal:
+
+```java
+// ✅ GOOD - inexact numeric elements (see TraceAssertions for the real usage)
+var config = new RecursiveComparisonConfiguration();
+config.ignoreFields(IGNORED_FIELDS_SCORES);
+config.registerComparatorForType(BigDecimal::compareTo, BigDecimal.class);
+
+assertThat(actual.feedbackScores())
+    .usingRecursiveFieldByFieldElementComparator(config)
+    .containsExactlyInAnyOrderElementsOf(expected.feedbackScores());
+```
+
+This is the same escalation rule as for single objects, not a separate one: plain
+`containsExactly*` by default, element comparator only when an element field can't be compared by
+`equals`. `FeedbackScore.value` is a `BigDecimal`, which is why
+`api/resources/utils/traces/TraceAssertions.java` needs it — most models don't.
+
 ### When Per-Field Assertions Are Still Right
 
 - **Spot checks against literals** — `assertThat(result.getName()).isEqualTo("John Doe")` is not
--- a/.agents/skills/opik-backend/testing.md
+++ b/.agents/skills/opik-backend/testing.md
@@ -155,12 +155,16 @@ assertThatThrownBy(() -> service.create(invalid))
     .hasMessageContaining("Name is required");
 ```
 
-### Comparing Two Objects — Use Recursive Comparison
+### Comparing Two Objects — Plain `isEqualTo` by Default
 
 Comparing two objects field by field is the default failure mode to avoid. It isn't just
 verbose: when a field is later added to the record, the test keeps passing while silently not
 covering the new field. Nothing fails and nothing flags it, so coverage erodes with every model
 change.
+
+Compare the whole object instead. `equals`/`hashCode` are the source of truth for equality in
+Java, so plain `isEqualTo` is the default — our API models are overwhelmingly records, whose
+generated `equals` covers every component and picks up new ones automatically:
 
 ```java
 // ❌ BAD - add a field to the record later and this still passes, now covering less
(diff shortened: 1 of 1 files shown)


.agents/skills/opik-backend/testing.md
```diff
@@ -158,6 +155,119 @@ assertThatThrownBy(() -> service.create(invalid))
...
+```java
+// ❌ BAD - add a field to the record later and this still passes, now covering less
+assertThat(actual.modelName()).isEqualTo(request.model());
+assertThat(actual.temperature()).isEqualTo(request.temperature());
+assertThat(actual.topP()).isEqualTo(request.topP());
+assertThat(actual.maxOutputTokens()).isEqualTo(request.maxCompletionTokens());
+
+// ✅ GOOD - new fields are compared automatically
+assertThat(actual)
+    .usingRecursiveComparison()
```
andrescrz, line 197: By default, we should do the `isEqualTo` comparison without `usingRecursiveComparison`, which relies on the objects equals and hashcode methods implementations, which are the source of truth for equality in Java. 

We only use `usingRecursiveComparison` for exceptional cases, because it uses an assertJ internal implementation to find inner fields (likely with reflection). Typical cases to use this are:
- Excluding fields.
- Using some specific assertJ feature during the comparison (e.g: custom comparator).
- Object has no proper hashcode and equalso implementation. That could happen for objects from 3rd party library which we can't control. For our code, we should always properly implement those methods, generally by using records, or for exceptional cases with Java POJOs/Beans where we can use Lombok to implement those (Value, Data, EqualsAndHashCode, in this order).

andrescrz (commented):
The document is generally fine, but there's an important nuance: using recursive is widespread in the code base, mostly as a side effect of using JSON views for response objects, but if it wasn't because of that, direct equals assertion would be preferred.

.agents/skills/opik-backend/testing.md
JetoPistola: Fixed in 8529864cff — you're right, and the inversion was mine. Restructured so plain `isEqualTo` is the documented default and `usingRecursiveComparison` is the exception, scoped to exactly the three cases you listed (excluding fields, an AssertJ feature such as a custom comparator, no usable `equals` — realistically only third-party types).

Checking your point against the tree, it holds up quantitatively: 101 of 117 recursive comparisons already pass a modifier (`ignoringFields` / `withComparatorForType`), i.e. they were already your exceptional cases. Only 16 are bare. And 270 API models are records, so the generated `equals` is both authoritative and automatically covers new components — which serves the motivation I was after (a field added later escaping coverage) better than reflection plus a hand-maintained ignore list.

Also added your JSON-views point as an explicit caveat, so readers know the widespread bare usage is a side effect rather than a deliberate default and isn't the pattern to copy. Added your Lombok ordering (`@Value`, `@Data`, `@EqualsAndHashCode`) as the fix for POJOs that need equality, rather than reaching for recursive comparison in the test. The comparator section is reframed as an exceptional case rather than an ignore-vs-comparator choice.

Not touching the 16 bare call sites here — migrating existing tests is out of scope for this ticket by design. Happy to file a follow-up for that cleanup if you think it's worth tracking.

.agents/skills/opik-backend/testing.md
```diff
@@ -158,6 +155,147 @@ assertThatThrownBy(() -> service.create(invalid))
...
+```java
+// ❌ BAD - misses a 4th unexpected element entirely
+assertThat(page.content().get(0).name()).isEqualTo("Alice");
+assertThat(page.content().get(1).name()).isEqualTo("Bob");
+assertThat(page.content().get(2).name()).isEqualTo("Charlie");
+
+// ✅ GOOD - order matters (sorting/pagination tests)
+assertThat(page.content())
+    .extracting(Entity::getName)
+    .containsExactly("Alice", "Bob", "Charlie");
```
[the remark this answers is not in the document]
JetoPistola: Fixed in 3517fe12fd — good catch, this was a real contradiction inside the doc. The inexact-types section warned that `BigDecimal.equals` is scale-sensitive, then the collections section handed readers `containsExactlyInAnyOrderElementsOf` as "✅ GOOD" without noting it compares elements by the element type's own `equals`. Both couldn't be right.

Your `TraceAssertions` pointer checks out — `api/resources/utils/traces/TraceAssertions.java:211-220` builds a `RecursiveComparisonConfiguration` with `ignoreFields(IGNORED_FIELDS_SCORES)` + `registerComparatorForType(BigDecimal::compareTo, BigDecimal.class)` and passes it to `usingRecursiveFieldByFieldElementComparator(config)`. That's now documented, with the config lifted verbatim from there rather than paraphrased.

One framing note on how it landed: this file's default was changed earlier in the review to plain `isEqualTo`, with `usingRecursiveComparison` reserved for exceptional cases (excluding fields, an AssertJ feature such as a custom comparator, no usable `equals`). Your case is that same exception applied per element rather than a competing rule, so it's written that way — plain `containsExactly*` stays the default, with the element comparator as the escalation for inexact numerics. Only 25 of 614 files under `api/` touch `BigDecimal`, so the plain form is right for the large majority; `FeedbackScore.value` being a `BigDecimal` is exactly why `TraceAssertions` needs the config.

andrescrz (approved):

PR #7699, opened 2026-08-02
[OPIK-7705] [BE] perf: bound the existing-item lookup in insertItemsIntoVersion

JetoPistola:
## Details

image

Every batch written into an existing dataset version read the *entire* version out of ClickHouse purely to work out how many incoming items were new versus updates — decoding `data_hash`, `tags`, `evaluators_hash`, `execution_policy_hash` and `description_hash` only to discard them, and building a full-version `Set<UUID>` on the heap to produce two integers. The cost was quadratic across a multi-batch upload (batch *k* read *k × N* rows to classify *N* items), and because it runs inside `withDatasetVersionLock` the per-batch cost is additive rather than amortised.

This replaces the full-version scan on the insert path with a lookup bounded by the incoming batch, so classification cost scales with batch size instead of version size.

- Adds a narrow `countExistingItemIds` DAO method that computes the intersection server-side with `count(DISTINCT dataset_item_id) ... WHERE dataset_item_id IN :itemIds`. `getItemIdsAndHashes` is left untouched for its two other callers (version-delta computation and the copy-version path), which legitimately need the hashes.
- Deduplicates the incoming batch by stable id before classifying. This also corrects a latent miscount: a batch repeating the same id previously counted it twice as new, despite the endpoint contract treating `id` as the upsert key.
- No full-version collection is materialised in application heap on the insert path. Both entrypoints benefit — `mutateLatestVersionWithInsert` and the `batch_group_id` path in `handleGroupedInsertion`.

Behaviour is otherwise unchanged: same items written, same version counts recorded, same `DatasetVersion` returned.

**Measured** on a 100k-item version with a full-size (1000-item) batch, against a local ClickHouse holding 7.5M rows across 2999 versions:

| | rows read | bytes read | wall clock |
|---|---|---|---|
| Before — full-version scan | 112,990 | 22.0 MB | 419 ms |
| After — bounded lookup | 6,437 | 0.41 MB | 57 ms |

~17.5x fewer rows and ~54x fewer bytes for a single batch; the win compounds across a multi-batch upload. `EXPLAIN indexes=1` confirms both `dataset_item_id` skip indexes from migration `000074` are exercised rather than assumed — the primary-key prefix narrows 1405 → 73 granules, the minmax index cuts 73 → 4, and the bloom filter holds at 4.

The `IN` list needs no chunking: `DatasetItemBatch` caps `items` at `@Size(min = 1, max = 1000)`, so it is bounded by construction.

## Change checklist
- [x] User facing
- [x] Documentation update

## Issues

- OPIK-7705

Escalated from OPIK_7270 (parallel bulk-upload SDK call for the Dataset API); related prior work in OPIK_7264 (per-dataset version lock).

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: full implementation (DAO query, service rewiring, tests, benchmarking)
- Human verification: code review + integration test run + before/after query measurement reviewed against `EXPLAIN indexes=1` output

## Testing

Commands run, from `apps/opik-backend`:

- `mvn compile -DskipTests` — clean
- `mvn spotless:check` — clean
- `mvn test -Dtest='DatasetVersionResourceTest'` — **121/121 pass**, covering both insert entrypoints
- `mvn test -Dtest='DatasetsResourceTest$ApiKey'` — **48/48 pass**

Scenarios validated by a new `InsertClassificationCounts` nested class in `DatasetVersionResourceTest`. No prior test on this path asserted `itemsAdded`/`itemsModified` — only `itemsTotal` — so a bounded lookup that failed to see a pre-existing item could have drifted the counters silently. Each new test asserts the full `(itemsTotal, itemsAdded, itemsModified)` triple:

- Re-inserting existing items alongside new ones counts them as modified, not added
- A stable id repeated within one batch increments `items_added` by one, not two
- Multi-batch insert into an existing version accumulates counts correctly and stays on a single version

Environment: local Testcontainers (ClickHouse + MySQL + Redis + Zookeeper), JDK 25. Measurement taken against a local Docker ClickHouse seeded with 7.5M rows.

Not run: the full backend suite. A combined `DatasetsResourceTest` run did surface one failure (`ApiKey.createDatasetItems__whenApiKeyIsPresent__thenReturnProperResponse`, HTTP 500), which was traced to the ClickHouse testcontainer dying mid-run — 98 × `Connection refused`, no `DB::Exception`. That class passes 48/48 in isolation. CI is the authority here.

## Documentation

N/A — internal performance change with no user-facing API or behavioural surface.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
@@ -60,6 +60,7 @@ import java.util.UUID;
 import java.util.stream.Collectors;
 
 import static com.comet.opik.domain.AsyncContextUtils.bindWorkspaceIdToFlux;
+import static com.comet.opik.domain.AsyncContextUtils.bindWorkspaceIdToMono;
 import static com.comet.opik.infrastructure.FilterUtils.getSTWithLogComment;
 import static com.comet.opik.infrastructure.instrumentation.InstrumentAsyncUtils.Segment;
 import static com.comet.opik.infrastructure.instrumentation.InstrumentAsyncUtils.endSegment;
@@ -97,6 +98,18 @@ public interface DatasetItemVersionDAO {
             @NonNull List<DatasetItemFilter> filters);
 
     Flux<DatasetItemIdAndHash> getItemIdsAndHashes(UUID datasetId, UUID versionId);
+
+    /**
+     * Counts how many of {@code itemIds} already exist in the given version.
+     * <p>
+     * Bounded alternative to {@link #getItemIdsAndHashes(UUID, UUID)} for the insert path, which only needs
+     * to classify an incoming batch as new vs. updated and does not need the hashes. Rows read scale with
+     * the batch size instead of the version size.
+     *
+     * @param itemIds the stable ids to look up; must be deduplicated by the caller
+     * @return the number of distinct {@code itemIds} present in the version
+     */
+    Mono<Long> countExistingItemIds(UUID datasetId, UUID versionId, Set<UUID> itemIds);
 
     /**
      * Copies items from a source version to a new target version directly within dataset_item_versions.
@@ -379,6 +392,24 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
             AND workspace_id = :workspace_id
             ORDER BY dataset_item_id DESC, last_updated_at DESC
             LIMIT 1 BY dataset_item_id
+            """;
+
+    /**
+     * Counts how many of the given stable ids already exist in a version.
+     * <p>
+     * The first three predicates are a prefix of the table's ordering key
+     * {@code (workspace_id, dataset_id, dataset_version_id, id)}, narrowing to the version's granules;
+     * {@code dataset_item_id} is not in the sort key and is instead served by the bloom-filter and
+     * minmax skip indexes added in migration {@code 000074}. This keeps rows read bounded by the
+     * incoming batch rather than by the size of the version being written to.
+     */
+    private static final String COUNT_EXISTING_ITEM_IDS = """
+            SELECT count(DISTINCT dataset_item_id) AS count
+            FROM dataset_item_versions
+            WHERE workspace_id = :workspace_id
+            AND dataset_id = :datasetId
+            AND dataset_version_id = :versionId
+            AND dataset_item_id IN :itemIds
             """;
 
     private static final String SELECT_DATASET_ITEM_VERSIONS = """
@@ -2630,6 +2661,35 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
                     .doOnSuccess(items -> log.info("Retrieved '{}' item IDs and hashes for version '{}'", items.size(),
                             versionId))
                     .flatMapMany(Flux::fromIterable);
+        });
+    }
+
+    @Override
+    @WithSpan
+    public Mono<Long> countExistingItemIds(@NonNull UUID datasetId, @NonNull UUID versionId,
+            @NonNull Set<UUID> itemIds) {
+        if (itemIds.isEmpty()) {
+            return Mono.just(0L);
+        }
+
+        log.debug("Counting existing item IDs among '{}' for dataset '{}', version '{}'", itemIds.size(), datasetId,
+                versionId);
+
+        return asyncTemplate.nonTransaction(connection -> {
+            var statement = connection.createStatement(COUNT_EXISTING_ITEM_IDS)
+                    .bind("datasetId", datasetId)
+                    .bind("versionId", versionId)
+                    .bind("itemIds", itemIds.stream().map(UUID::toString).toArray(String[]::new));
+
+            Segment segment = startSegment(DATASET_ITEM_VERSIONS, CLICKHOUSE, "count_existing_item_ids");
+
+            return makeMonoContextAware(bindWorkspaceIdToMono(statement))
+                    .flatMapMany(result -> result.map((row, metadata) -> row.get("count", Long.class)))
+                    .next()
+                    .defaultIfEmpty(0L)
+                    .doOnSuccess(count -> log.debug("Found '{}' existing items among '{}' for version '{}'", count,
+                            itemIds.size(), versionId))
+                    .doFinally(signalType -> endSegment(segment));
         });
     }
 
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
@@ -106,7 +106,8 @@ public interface DatasetItemVersionDAO {
      * to classify an incoming batch as new vs. updated and does not need the hashes. Rows read scale with
      * the batch size instead of the version size.
      *
-     * @param itemIds the stable ids to look up; must be deduplicated by the caller
+     * @param itemIds the stable ids to look up; should be deduplicated by the caller. Null or empty
+     *                yields {@code 0} without querying.
      * @return the number of distinct {@code itemIds} present in the version
      */
     Mono<Long> countExistingItemIds(UUID datasetId, UUID versionId, Set<UUID> itemIds);
@@ -2667,8 +2668,8 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
     @Override
     @WithSpan
     public Mono<Long> countExistingItemIds(@NonNull UUID datasetId, @NonNull UUID versionId,
-            @NonNull Set<UUID> itemIds) {
-        if (itemIds.isEmpty()) {
+            Set<UUID> itemIds) {
+        if (CollectionUtils.isEmpty(itemIds)) {
             return Mono.just(0L);
         }
 
@@ -2687,8 +2688,9 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
                     .flatMapMany(result -> result.map((row, metadata) -> row.get("count", Long.class)))
                     .next()
                     .defaultIfEmpty(0L)
-                    .doOnSuccess(count -> log.debug("Found '{}' existing items among '{}' for version '{}'", count,
-                            itemIds.size(), versionId))
+                    .doOnSuccess(count -> log.debug(
+                            "Found existing items for version '{}' among '{}' incoming ids: '{}'", versionId,
+                            itemIds.size(), count))
                     .doFinally(signalType -> endSegment(segment));
         });
     }
--- a/apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
@@ -77,6 +77,7 @@ import ru.vyarus.dropwizard.guice.test.jupiter.ext.TestDropwizardAppExtension;
 import ru.vyarus.guicey.jdbi3.tx.TransactionTemplate;
 import uk.co.jemos.podam.api.PodamFactory;
 
+import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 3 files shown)


apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -381,6 +394,24 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
...
+     * incoming batch rather than by the size of the version being written to.
+     */
+    private static final String COUNT_EXISTING_ITEM_IDS = """
+            SELECT count(DISTINCT dataset_item_id) AS count
+            FROM dataset_item_versions
+            WHERE workspace_id = :workspace_id
+            AND dataset_id = :datasetId
+            AND dataset_version_id = :versionId
+            AND dataset_item_id IN :itemIds
+            """;
```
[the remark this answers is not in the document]
JetoPistola: Skipping — the guideline is real, but applying it to this one query would make the file less consistent, not more.

`DatasetItemVersionDAO.java` uses `:datasetId` 35 times and `:versionId` 15 times. The new query deliberately copied the binding style of `SELECT_ITEM_IDS_AND_HASHES`, which sits directly above it and is the query this change replaces on the insert path. Renaming only `COUNT_EXISTING_ITEM_IDS` would leave a single snake_case query among ~50 camelCase ones with no functional benefit.

(`:workspace_id` is already snake_case here, but that's not a partial application of the rule — `bindWorkspaceIdToMono` requires that exact parameter name.)

Reconciling the documented convention with the ~50 existing call sites is a worthwhile cleanup, but it's repo-wide and out of scope for a performance change — `.claude/rules/code-style.md` explicitly rules out drive-by refactoring of unrelated code.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -2633,6 +2664,35 @@ public Flux<DatasetItemIdAndHash> getItemIdsAndHashes(@NonNull UUID datasetId, @
...
+        }
+
+        log.debug("Counting existing item IDs among '{}' for dataset '{}', version '{}'", itemIds.size(), datasetId,
+                versionId);
+
+        return asyncTemplate.nonTransaction(connection -> {
+            var statement = connection.createStatement(COUNT_EXISTING_ITEM_IDS)
+                    .bind("datasetId", datasetId)
+                    .bind("versionId", versionId)
+                    .bind("itemIds", itemIds.stream().map(UUID::toString).toArray(String[]::new));
```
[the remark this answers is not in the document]
JetoPistola: Skipping — the null element is unreachable from the only caller, and both suggested remedies would make things worse.

`countExistingItemIds` has exactly one caller: `DatasetItemService.insertItemsIntoVersion`. It builds the set from `normalizedItems`, and the normalization immediately above it (`DatasetItemService:1686-1696`) guarantees every `datasetItemId` is non-null — when one is absent it derives it from `id` or generates a fresh UUIDv7. A null element cannot reach the DAO.

On the two options offered:

- **Filtering nulls** would be actively harmful. The return value drives `items_added` / `items_modified`; silently dropping an id would under-count rather than fail, turning a would-be crash into wrong data.
- **An explicit null check** is defensive code for an impossible state, which `.claude/rules/code-style.md` rules out ("No extra error handling 'just in case'").

The `@NonNull Set<UUID>` parameter contract plus a caller that structurally cannot produce nulls is the right level of strictness here. If a future caller with weaker guarantees appears, the check belongs at that boundary.

CometActions:

### Access Information
- **URL:** https://pr-7699.dev.comet.com
- **Namespace:** pr-7699
- **Version:** 2.2.14-7699-merge-2795
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7699%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7699`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

CometActions:

### Access Information
- **URL:** https://pr-7699.dev.comet.com
- **Namespace:** pr-7699
- **Version:** 2.2.14-7699-merge-2799
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7699%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -2633,6 +2664,35 @@ public Flux<DatasetItemIdAndHash> getItemIdsAndHashes(@NonNull UUID datasetId, @
         });
     }
 
+    @Override
+    @WithSpan
+    public Mono<Long> countExistingItemIds(@NonNull UUID datasetId, @NonNull UUID versionId,
+            @NonNull Set<UUID> itemIds) {
+        if (itemIds.isEmpty()) {
```
andrescrz: Minor: use CollectionUtils.isEmpty and remove the Lombok NonNull annotation instead. If you handle empty condition, better not fail if null itemIds.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -2633,6 +2664,35 @@ public Flux<DatasetItemIdAndHash> getItemIdsAndHashes(@NonNull UUID datasetId, @
...
+            var statement = connection.createStatement(COUNT_EXISTING_ITEM_IDS)
+                    .bind("datasetId", datasetId)
+                    .bind("versionId", versionId)
+                    .bind("itemIds", itemIds.stream().map(UUID::toString).toArray(String[]::new));
+
+            Segment segment = startSegment(DATASET_ITEM_VERSIONS, CLICKHOUSE, "count_existing_item_ids");
+
+            return makeMonoContextAware(bindWorkspaceIdToMono(statement))
+                    .flatMapMany(result -> result.map((row, metadata) -> row.get("count", Long.class)))
+                    .next()
```
andrescrz, line 2689: Why is next needed here? Any other better alternatives?

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -2633,6 +2664,35 @@ public Flux<DatasetItemIdAndHash> getItemIdsAndHashes(@NonNull UUID datasetId, @
...
+                    .defaultIfEmpty(0L)
+                    .doOnSuccess(count -> log.debug("Found '{}' existing items among '{}' for version '{}'", count,
```
andrescrz: Minor: move the printed value to the end of the sentence. It's easier to grep during production debugging sessions.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
```diff
@@ -1797,6 +1798,126 @@ void insertItems__whenUpdatingSameItemsWithoutBatchGroupId__thenMutateLatestVers
         }
     }
 
+    @Nested
+    @TestInstance(TestInstance.Lifecycle.PER_CLASS)
+    @DisplayName("Insert Classification Counts")
+    class InsertClassificationCounts {
+
+        private final IdGenerator idGenerator = TestIdGeneratorFactory.create();
```
andrescrz: Minor: better define as private static final field (using those naming conventions) at the top level class in these methods. Make sure no overlap if already exists. In that case, use the existing one.

andrescrz (commented):
Some suggestions, but no blockers.

andrescrz (dismissed):

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
JetoPistola: Fixed in 3fb4b866a7 — switched to `CollectionUtils.isEmpty(itemIds)` and dropped `@NonNull` from `itemIds` (kept on the two UUIDs). Agreed on the reasoning: if the method already tolerates an empty set, failing hard on null was inconsistent, and `CollectionUtils.isEmpty` covers both in one call. `CollectionUtils` was already imported and used elsewhere in this DAO. Also updated the interface Javadoc to state that null or empty yields `0` without querying.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
JetoPistola: Fixed in 3fb4b866a7 — reworded so the count lands last: `"Found existing items for version '{}' among '{}' incoming ids: '{}'"`. Fair point on grep-ability during production debugging.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
JetoPistola: Fixed in 3fb4b866a7 — hoisted to `private static final IdGenerator ID_GENERATOR` on the top-level class, alongside `API_KEY` / `TEST_WORKSPACE`, following the same UPPER_SNAKE convention.

On the overlap check you flagged: there *was* an existing one — `ExperimentDatasetVersionLinking` had its own `private static final IdGenerator idGenerator`. Leaving it would have shadowed the new top-level field, so I removed it and repointed both usages (the duplicate-id test and the non-existent-version test) at `ID_GENERATOR`. One declaration now serves both nested classes.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
JetoPistola: `next()` is there because the chain is a `Flux` at that point: `flatMapMany` maps each row of the result set, so `next()` takes the first (and only) row and narrows it back to a `Mono`, with `defaultIfEmpty(0L)` covering the zero-row case.

Kept as-is because it matches the three sibling count/rows-updated queries in this same DAO — `countItemsInVersion` and the two copy-from-legacy methods all use the identical `.flatMap(...).next().defaultIfEmpty(0L)` shape.

The tidier alternative would be `Mono.from(result.map(...))`, dropping `flatMapMany` and `next()` entirely. Happy to switch if you'd prefer it, but it would then be the only count query in the file written that way — I'd rather change all four together in a follow-up than leave one divergent.

andrescrz (approved):
LGTM

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7699`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

PR #7700, opened 2026-08-02
[NA] [BE] fix: load prices for five vendors OpenRouter resells

BILLKISHORE:
## Details

Five vendors that OpenRouter resells carry non-zero-cost rows in `model_prices_and_context_window.json` but were missing from `PROVIDERS_MAPPING`: `ai21`, `morph`, `inception`, `meta` and `zai`. `buildModelPrice` returns null for any `litellm_provider` not in that map, so every one of those rows was dropped at load time and the provider-prefix fallback had nothing to resolve against. Anything routed through OpenRouter fell through to `DEFAULT_COST` and logged no cost at all.

This is the same gap #7604 fixed for Moonshot, found the same way: by scanning `litellm_provider` values in the bundled price map for priced models absent from `PROVIDERS_MAPPING`, then narrowing to the ones actually reachable.

`z-ai` needs two entries for the same reason `moonshot` does. The map is read twice: once with the price file's `litellm_provider` (`zai`) when loading rows, and once with the model-name prefix OpenRouter uses (`z-ai`) when resolving the prefix fallback. The other four spell both the same way, so one entry each.

Fifteen models start pricing as a result: nine `z-ai/glm-*`, two `ai21/jamba-*`, two `morph/morph-v3-*`, `inception/mercury-2` and `meta/muse-spark-1.1`.

### What I deliberately left out

**`minimax`.** It looks like it belongs here (10 OpenRouter models, 6 priced rows) but no mapping entry can help it. Its price keys are stored as written upstream, e.g. `minimax/MiniMax-M2.1` becomes the key `MiniMax-M2.1`, and every lookup path either matches exactly or lowercases the incoming name. So a request for `minimax/minimax-m2.1` tries `minimax-m2.1` then `minimax-m2-1`, and neither can equal `MiniMax-M2.1`. Adding the mapping would load six rows that nothing can reach.

That is worth raising on its own: the price map is case-sensitive on the stored key, so any provider whose upstream keys carry uppercase is unreachable regardless of mapping. It affects the shared lookup path rather than this map, so it is not in this PR. Happy to open an issue if useful.

**`cohere` and `tencent`.** Both have priced rows and both are OpenRouter prefixes, but none of the model names intersect. OpenRouter serves `cohere/command-a` and `cohere/command-r-08-2024` while the priced `cohere` rows are embeddings plus legacy `command` and `command-nightly`. Mapping them would load rows that no request resolves to.

**Cache cost routing.** Four of the fifteen models publish a `cache_read_input_token_cost`, but none of these providers is registered in `PROVIDERS_CACHE_COST_CALCULATOR`, so `resolveCalculator` falls back to `textGenerationCost` and cached tokens are not discounted. Registering them needs evidence of how each API reports cached tokens, the way #7604 justified routing Moonshot through the OpenAI calculator. Base pricing working is still a strict improvement over no cost at all, and I would rather not guess at a calculator.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

No tracked issue. Found by scanning the bundled price map for `litellm_provider` values with priced models that are absent from `PROVIDERS_MAPPING`, the same discovery route as #7604.

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: the scan that found the gap, the mapping change, the tests, and this description.
- Human verification: every number below comes from a run whose output was read, including a baseline on `main` before any edit and a revert of `CostService.java` to confirm the new cases fail without it. The exclusions above were each checked against the price file rather than assumed.

## Testing

Environment: local, Maven 3.9.16 on OpenJDK 25.0.2, matching the Corretto 25 in `backend_tests.yml`.

- `mvn test -Dtest=CostServiceTest` on `main`, before touching anything: 96 tests, 0 failures. Establishes the toolchain works here.
- Same command on this branch: 105 tests, 0 failures. The delta of 9 is exactly the new parameterized cases.
- Same command with `CostService.java` reverted to `main` and the new test kept: 105 tests, **9 failures**, one per case. So the tests fail without the fix rather than just describing it.
- `mvn spotless:apply`: no files changed, tree already clean.

Expected costs are computed from the bundled price file at 1000 prompt and 200 completion tokens, with the arithmetic written out next to each case. For example `ai21/jamba-large-1.7` at input 2e-6 and output 8e-6 gives `1000 * 2e-6 + 200 * 8e-6 = 0.0036`.

Not clean: a full `mvn test` fails in this environment after about 50 minutes, with 39 `ContainerLaunchException` and several thousand connection-refused entries across the testcontainers-backed resource tests. That is contention from starting ClickHouse, MySQL and Redis for many test classes at once on a laptop, not a regression. `DashboardsResourceProjectScopedTest`, one of the classes that errors in the full run, passes in isolation on **both** `main` and this branch with an identical 5 tests / 0 failures / ~36s. CI is the gate for the integration suite.

## Documentation

No documentation change. No public API or configuration surface; this only affects which price rows load.


andrescrz (approved):
Hi @BILLKISHORE 

Best

PR #7705, opened 2026-08-03
[OPIK-7707] [BE] perf: make dataset version count updates atomic

JetoPistola:
## Details

image

Every batch appended to an existing dataset version paid three MySQL round-trips inside the per-dataset lock that serialises the upload: `findById`, then `updateCounts` with totals computed in Java, then a `getVersionById` re-read of the row just written. Because the lock serialises batches, those round-trips are additive across an upload — a 100-batch upload pays them 100 times in sequence. This replaces the read-modify-write with a single atomic SQL increment and drops the re-read, taking the insert path from three statements to one.

- `DatasetVersionDAO.updateCounts` (absolute values) becomes `incrementCounts`, which applies **signed deltas** in the database (`SET items_total = items_total + :delta`). One method serves both paths: `+n` on insert, `-n` total / `+n` deleted on delete. It returns the affected-row count so the service preserves the `NotFoundException` that `findById().orElseThrow()` used to provide.
- Service-side, both paths funnel through one private `updateVersionCounts(...)` that owns the transaction, the DAO call, and the zero-row `NotFoundException` guard. `updateVersionCountsForInsert` / `...ForDelete` remain as thin named wrappers documenting which counters each path moves.
- The post-insert `getVersionById` is gone. No consumer reads the returned `DatasetVersion` — the REST endpoint returns 204 with the `.block()` result unassigned, and `saveBatch` maps it to `items.size()` — so `insertItemsIntoVersion` now returns `Mono<Void>` rather than fabricating a partially-populated object that would look real to a future caller.
- The delete path is converted to the same atomic increment. **It saves no round-trip** — both callers already fetch the version for their own logging — but the arithmetic no longer depends on `withDatasetVersionLock` for mutual exclusion, and insert/delete now write counts the same way.
- Counter semantics are unchanged; this changes only how counts are written.

Two notes for reviewers, since they differ from the ticket as filed:

- The ticket cites an existing test `InsertClassificationCounts` as a regression net for this path. No test by that name exists in the tree; the real coverage is in `DatasetVersionResourceTest` (`VersionSnapshotTests`, `ApplyDatasetItemChanges`, `MutateLatestVersion`).
- The ticket assumes the delete path shares the insert path's wasteful shape. It doesn't — expect no latency change there, only the consistency and lock-independence win described above.

`updateCounts` had zero remaining callers once both paths were converted, so it was removed rather than left as dead code.

## Change checklist
- [x] User facing
- [x] Documentation update

## Issues

- Resolves OPIK-7707

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Scope: full implementation (code + tests)
- Human verification: CI green (all 16 integration groups); pending human code review

## Testing

Added a `AtomicVersionCountUpdates` nested class in `DatasetVersionResourceTest` with six tests:

- single-batch append asserts the full counter triple, including that a re-sent existing item moves `itemsModified` but not `itemsTotal`
- multi-batch append via the `batch_group_id` entrypoint accumulates into one version
- **8 threads × 25 concurrent increments hitting the DAO directly with `withDatasetVersionLock` bypassed**, asserting exact counters — this is the acceptance criterion that the arithmetic no longer depends on the lock; the previous read-modify-write would lose updates here
- delete path: negative delta decrements `items_total` and raises `items_deleted`, leaving added/modified untouched
- increment against an unknown version affects zero rows (the `NotFoundException` path)
- increment scoped to a foreign workspace affects zero rows and leaves counters untouched (added in review follow-up)

Commands run:

- `mvn -o compile -DskipTests` — passes
- `mvn -o test-compile -DskipTests` — passes
- `mvn -o spotless:check` — `BUILD SUCCESS`

**CI result:** all 16 backend integration groups pass. `DatasetVersionResourceTest$AtomicVersionCountUpdates` reports `Tests run: 6, Failures: 0, Errors: 0` (Integration Group 6), so all six new tests executed and passed — including the lock-bypassed concurrency test. Pre-existing regression nets on these paths are green as well: `MutateLatestVersion` (6), `ConcurrentUploads` (4), `DeleteItemsWithVersioning` (5), `BatchVersioningTests`, `BatchVersioningDeleteTests` (10).

Note on local runs: `mvn -o test -Dtest=DatasetVersionResourceTest` could not be brought to green on my machine — three attempts each died in the test-class constructor, before any test method, with `ClickHouse exception, code: 159 ... Read timed out` on `ON CLUSTER` DDL during Liquibase migration, on a *different* pre-existing changeset each time (`000097`, `000109`, `000070`). That is local distributed-DDL latency, not this diff: the change is MySQL-only and all three changesets are already on `main`. CI, which has a healthy container environment, runs the suite clean.

Also verified by inspection: the counter columns are signed `INT` (`000036_add_dataset_versions_tables.sql`), so the delete path's negative delta cannot hit an unsigned-underflow error under strict mode — same semantics as the previous Java-side subtraction. No migration is required.

### Review follow-up (`0eb086558e`, `0153bc85dc`)

`updateVersionCountsForDelete` now checks the affected-row count from `incrementCounts` and throws `NotFoundException` on zero, matching the insert path; the helper's duplicate `log.info` was dropped (the callers already log `deletedCount`, `versionId`, and the resulting total). Note this failure mode is not reachable today — both delete callers call `getVersionById` first, which already throws — but insert and delete were handling a zero-row result inconsistently, which would become a real bug once the lock is narrowed or the pre-fetch removed.

That guard left the two helpers identical apart from their delta arguments, so a second follow-up folded the shared transaction / DAO-call / guard into one `updateVersionCounts(...)` taking signed deltas. Pure refactor, no behavioural change: verified by `AtomicVersionCountUpdates` (6), `MutateLatestVersion` (6, insert wrapper) and `DeleteItemsWithVersioning` (5, delete wrapper) all green on `0153bc85dc`.

## Documentation

N/A — internal performance and correctness-hardening change with no user-facing or API surface.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
@@ -1780,13 +1780,17 @@ class DatasetItemServiceImpl implements DatasetItemService {
      * @param workspaceId The workspace ID
      * @param deletedCount Number of items deleted
      * @param userName The user performing the update
+     * @throws NotFoundException if the version no longer exists or does not belong to the workspace
      */
     private void updateVersionCountsForDelete(UUID versionId, String workspaceId, int deletedCount, String userName) {
-        log.info("deleteItemsFromExistingVersion: updating counts - deletedCount='{}'", deletedCount);
-
         template.inTransaction(WRITE, handle -> {
             var dao = handle.attach(DatasetVersionDAO.class);
-            dao.incrementCounts(versionId, -deletedCount, 0, 0, deletedCount, workspaceId, userName);
+
+            int updated = dao.incrementCounts(versionId, -deletedCount, 0, 0, deletedCount, workspaceId, userName);
+
+            if (updated == 0) {
+                throw new NotFoundException("Version not found: '%s'".formatted(versionId));
+            }
             return null;
         });
     }
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
@@ -1662,7 +1662,7 @@ class DatasetItemServiceImpl implements DatasetItemService {
         UUID latestVersionId = latestVersion.get().id();
         log.info("Inserting '{}' items into existing version '{}'", batch.items().size(), latestVersionId);
 
-        return insertItemsIntoVersion(batch, datasetId, latestVersionId, workspaceId, userName);
+        return insertItemsIntoVersion(batch, datasetId, latestVersionId, workspaceId, userName).then(Mono.empty());
     }
 
     /**
@@ -1670,14 +1670,17 @@ class DatasetItemServiceImpl implements DatasetItemService {
      * Handles validation, classification of new vs updated items, and count updates.
      * Used by mutateLatestVersionWithInsert and handleGroupedInsertion.
      *
+     * Completes empty: no caller reads the resulting version, so the counters are applied with a single atomic
+     * increment and the row is not read back.
+     *
      * @param batch the batch of items to insert
      * @param datasetId the dataset ID
      * @param versionId the version ID to insert into
      * @param workspaceId the workspace ID
      * @param userName the username
-     * @return Mono emitting the updated dataset version
+     * @return Mono completing when the items are inserted and the counts updated
      */
-    private Mono<DatasetVersion> insertItemsIntoVersion(DatasetItemBatch batch, UUID datasetId, UUID versionId,
+    private Mono<Void> insertItemsIntoVersion(DatasetItemBatch batch, UUID datasetId, UUID versionId,
             String workspaceId, String userName) {
         // Validate and prepare items
         List<DatasetItem> validatedItems = addIdIfAbsent(batch);
@@ -1717,11 +1720,10 @@ class DatasetItemServiceImpl implements DatasetItemService {
                                     // Insert items directly into the existing version
                                     return versionDao
                                             .insertItems(datasetId, versionId, normalizedItems, workspaceId, userName)
-                                            .then(Mono.fromCallable(() -> {
-                                                updateVersionCountsForInsert(versionId, workspaceId, finalNewItemsCount,
-                                                        finalUpdatedItemsCount, userName);
-                                                return versionService.getVersionById(workspaceId, datasetId, versionId);
-                                            }).subscribeOn(Schedulers.boundedElastic()));
+                                            .then(Mono.fromRunnable(() -> updateVersionCountsForInsert(versionId,
+                                                    workspaceId, finalNewItemsCount, finalUpdatedItemsCount, userName))
+                                                    .subscribeOn(Schedulers.boundedElastic()))
+                                            .then();
                                 });
                     }));
         }).contextWrite(c -> c.put(RequestContext.WORKSPACE_ID, workspaceId)
@@ -1742,43 +1744,36 @@ class DatasetItemServiceImpl implements DatasetItemService {
             int updatedItemsCount, String userName) {
         template.inTransaction(WRITE, handle -> {
             var dao = handle.attach(DatasetVersionDAO.class);
-            var currentVersion = dao.findById(versionId, workspaceId)
-                    .orElseThrow(() -> new NotFoundException(
-                            "Version not found: '%s'".formatted(versionId)));
 
             // Only increment total by new items (not updates)
-            int newTotal = currentVersion.itemsTotal() + newItemsCount;
-            int newAdded = currentVersion.itemsAdded() + newItemsCount;
-            int newModified = currentVersion.itemsModified() + updatedItemsCount;
+            int updated = dao.incrementCounts(versionId, newItemsCount, newItemsCount, updatedItemsCount, 0,
+                    workspaceId, userName);
 
-            dao.updateCounts(versionId, newTotal, newAdded, newModified,
-                    currentVersion.itemsDeleted(), workspaceId, userName);
+            if (updated == 0) {
+                throw new NotFoundException("Version not found: '%s'".formatted(versionId));
+            }
             return null;
         });
     }
 
     /**
      * Updates version counts after deleting items from an existing version.
-     * Extracted to reduce complexity and improve testability.
+     * <p>
+     * Applies the deltas atomically rather than writing back totals computed from a caller-supplied snapshot, so the
+     * result no longer depends on that snapshot still being current. Unlike the insert path this saves no round-trip:
+     * both callers fetch the version for their own purposes regardless.
      *
      * @param versionId The version ID to update
      * @param workspaceId The workspace ID
-     * @param currentVersion The current version before deletion
      * @param deletedCount Number of items deleted
      * @param userName The user performing the update
      */
-    private void updateVersionCountsForDelete(UUID versionId, String workspaceId, DatasetVersion currentVersion,
-            int deletedCount, String userName) {
-        int newTotal = currentVersion.itemsTotal() - deletedCount;
-        int newDeleted = currentVersion.itemsDeleted() + deletedCount;
-
-        log.info("deleteItemsFromExistingVersion: updating counts - newTotal='{}', newDeleted='{}'",
-                newTotal, newDeleted);
+    private void updateVersionCountsForDelete(UUID versionId, String workspaceId, int deletedCount, String userName) {
+        log.info("deleteItemsFromExistingVersion: updating counts - deletedCount='{}'", deletedCount);
 
         template.inTransaction(WRITE, handle -> {
             var dao = handle.attach(DatasetVersionDAO.class);
-            dao.updateCounts(versionId, newTotal, currentVersion.itemsAdded(),
-                    currentVersion.itemsModified(), newDeleted, workspaceId, userName);
+            dao.incrementCounts(versionId, -deletedCount, 0, 0, deletedCount, workspaceId, userName);
             return null;
         });
     }
@@ -2173,8 +2168,7 @@ class DatasetItemServiceImpl implements DatasetItemService {
 
                         // Update version counts in MySQL
                         return Mono.fromCallable(() -> {
-                            updateVersionCountsForDelete(versionId, workspaceId, currentVersion,
-                                    deletedCount.intValue(), userName);
+                            updateVersionCountsForDelete(versionId, workspaceId, deletedCount.intValue(), userName);
                             log.info("Deleted '{}' items from version '{}', new total '{}'",
                                     deletedCount, versionId, currentVersion.itemsTotal() - deletedCount.intValue());
                             return null;
@@ -2218,8 +2212,7 @@ class DatasetItemServiceImpl implements DatasetItemService {
 
                         // Update version counts in MySQL
                         return Mono.fromCallable(() -> {
-                            updateVersionCountsForDelete(versionId, workspaceId, currentVersion,
-                                    deletedCount.intValue(), userName);
+                            updateVersionCountsForDelete(versionId, workspaceId, deletedCount.intValue(), userName);
                             log.info("Deleted '{}' items from version '{}', new total '{}'",
                                     deletedCount, versionId, currentVersion.itemsTotal() - deletedCount.intValue());
                             return null;
@@ -2344,7 +2337,8 @@ class DatasetItemServiceImpl implements DatasetItemService {
                         var existingVersion = optionalVersion.get();
                         log.info("Appending '{}' items to existing version '{}' for batch_group_id '{}'",
                                 batch.items().size(), existingVersion.id(), batchGroupId);
-                        return insertItemsIntoVersion(batch, datasetId, existingVersion.id(), workspaceId, userName);
+                        return insertItemsIntoVersion(batch, datasetId, existingVersion.id(), workspaceId, userName)
+                                .then(Mono.<DatasetVersion>empty());
                     } else {
                         // No version with this batch_group_id - create new one
                         log.info("Creating new version with batch_group_id '{}' for dataset '{}'",
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
@@ -1767,13 +1767,17 @@ class DatasetItemServiceImpl implements DatasetItemService {
      * @param workspaceId The workspace ID
      * @param deletedCount Number of items deleted
      * @param userName The user performing the update
+     * @throws NotFoundException if the version no longer exists or does not belong to the workspace
      */
     private void updateVersionCountsForDelete(UUID versionId, String workspaceId, int deletedCount, String userName) {
-        log.info("deleteItemsFromExistingVersion: updating counts - deletedCount='{}'", deletedCount);
-
         template.inTransaction(WRITE, handle -> {
             var dao = handle.attach(DatasetVersionDAO.class);
-            dao.incrementCounts(versionId, -deletedCount, 0, 0, deletedCount, workspaceId, userName);
+
+            int updated = dao.incrementCounts(versionId, -deletedCount, 0, 0, deletedCount, workspaceId, userName);
+
+            if (updated == 0) {
+                throw new NotFoundException("Version not found: '%s'".formatted(versionId));
+            }
             return null;
         });
     }
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
@@ -132,18 +132,15 @@ public interface DatasetItemService {
      *   <li>If batchGroupId is null: Mutates the latest version by appending items (backwards compatibility)</li>
      *   <li>If batchGroupId is provided: Creates a new version with batch grouping (multiple batches can share the same version)</li>
      *   <li>If no versions exist, creates the first version regardless of batchGroupId</li>
-     *   <li>Returns the DatasetVersion (newly created or mutated)</li>
      * </ul>
-     * When versioning is disabled (legacy mode):
-     * <ul>
-     *   <li>Saves items to the legacy dataset_items table</li>
-     *   <li>Returns empty Mono</li>
-     * </ul>
+     * When versioning is disabled (legacy mode), saves items to the legacy dataset_items table.
      *
      * @param batch the batch of items to save (must include datasetId or datasetName, may include batchGroupId)
-     * @return Mono emitting the DatasetVersion when versioning is enabled, or empty when disabled
+     * @return Mono completing when the batch is persisted. No version is emitted: appending to an existing version
+     *         has nothing to read back, and the only caller discards the value, so the type says so rather than
+     *         promising a value that arrives on some paths and not others.
      */
-    Mono<DatasetVersion> save(DatasetItemBatch batch);
+    Mono<Void> save(DatasetItemBatch batch);
 
 }
 
@@ -1610,12 +1607,12 @@ class DatasetItemServiceImpl implements DatasetItemService {
 
     @Override
     @WithSpan
-    public Mono<DatasetVersion> save(@NonNull DatasetItemBatch batch) {
+    public Mono<Void> save(@NonNull DatasetItemBatch batch) {
 
         if (!featureFlags.isDatasetVersioningEnabled()) {
             // Legacy: save to legacy table
             log.info("Saving items to legacy table for dataset '{}'", batch.datasetId());
-            return verifyDatasetExistsAndSave(batch).then(Mono.empty());
+            return verifyDatasetExistsAndSave(batch).then();
         }
 
         return getDatasetId(batch)
@@ -1636,7 +1633,8 @@ class DatasetItemServiceImpl implements DatasetItemService {
                     log.info("Creating version with batch grouping for dataset '{}', batch_group_id: '{}'", datasetId,
                             batchGroupId);
                     return handleGroupedInsertion(batchGroupId, batch, datasetId, workspaceId, userName);
-                })));
+                })))
+                .then();
     }
 
     /**
@@ -1749,8 +1747,8 @@ class DatasetItemServiceImpl implements DatasetItemService {
     /**
      * Updates version counts after deleting items from an existing version.
      * <p>
-     * Unlike the insert path this saves no round-trip: both callers fetch the version for their own purposes
-     * regardless. What it buys is that the arithmetic no longer depends on that snapshot still being current.
+     * Expressing the update as a delta removed the last reason either caller had to read the version first, so
+     * both now skip that round-trip entirely and the arithmetic no longer depends on a snapshot staying current.
      *
      * @param versionId The version ID to update
      * @param workspaceId The workspace ID
@@ -2152,14 +2150,10 @@ class DatasetItemServiceImpl implements DatasetItemService {
         }
 
         return Mono.defer(() -> {
-            // Get current version to update counts
-            DatasetVersion currentVersion = versionService.getVersionById(workspaceId, datasetId, versionId);
-
-            log.info(
-                    "deleteItemsFromExistingVersion: currentVersion itemsTotal='{}', itemsDeleted='{}', versionId='{}'",
-                    currentVersion.itemsTotal(), currentVersion.itemsDeleted(), versionId);
-
-            log.info("deleteItemsFromExistingVersion: attempting to remove '{}' items", ids.size());
+            // The counters are applied as a delta, so no pre-delete snapshot of the version is read here:
+            // fetching one would be a synchronous MySQL round-trip per batch purely to enrich a log line.
+            log.info("deleteItemsFromExistingVersion: attempting to remove '{}' items from version '{}'",
+                    ids.size(), versionId);
 
             // Remove items from the version
             return versionDao.removeItemsFromVersion(datasetId, versionId, ids, workspaceId)
@@ -2175,8 +2169,7 @@ class DatasetItemServiceImpl implements DatasetItemService {
                         // Update version counts in MySQL
                         return Mono.fromCallable(() -> {
                             updateVersionCountsForDelete(versionId, workspaceId, deletedCount.intValue(), userName);
-                            log.info("Deleted '{}' items from version '{}', new total '{}'",
-                                    deletedCount, versionId, currentVersion.itemsTotal() - deletedCount.intValue());
+                            log.info("Deleted '{}' items from version '{}'", deletedCount, versionId);
                             return null;
                         }).subscribeOn(Schedulers.boundedElastic());
                     })
@@ -2197,12 +2190,8 @@ class DatasetItemServiceImpl implements DatasetItemService {
                 versionId, datasetId);
 
         return Mono.defer(() -> {
-            // Get current version to update counts
-            DatasetVersion currentVersion = versionService.getVersionById(workspaceId, datasetId, versionId);
-
-            log.info(
-                    "deleteItemsFromExistingVersionByFilters: currentVersion itemsTotal='{}', itemsDeleted='{}', versionId='{}'",
-                    currentVersion.itemsTotal(), currentVersion.itemsDeleted(), versionId);
+            // Counters are applied as a delta, so no pre-delete version snapshot is read here -- see
+            // deleteItemsFromExistingVersion for why.
 
             // Remove items matching filters from the version
             return versionDao.removeItemsFromVersionByFilters(datasetId, versionId, filters, workspaceId)
@@ -2219,8 +2208,7 @@ class DatasetItemServiceImpl implements DatasetItemService {
                         // Update version counts in MySQL
                         return Mono.fromCallable(() -> {
                             updateVersionCountsForDelete(versionId, workspaceId, deletedCount.intValue(), userName);
-                            log.info("Deleted '{}' items from version '{}', new total '{}'",
-                                    deletedCount, versionId, currentVersion.itemsTotal() - deletedCount.intValue());
+                            log.info("Deleted '{}' items from version '{}'", deletedCount, versionId);
                             return null;
                         }).subscribeOn(Schedulers.boundedElastic());
                     })
@@ -2331,7 +2319,8 @@ class DatasetItemServiceImpl implements DatasetItemService {
      * @param datasetId the dataset ID
      * @param workspaceId the workspace ID
      * @param userName the username
-     * @return Mono emitting the dataset version
+     * @return Mono emitting the newly created version when this batch mints one, or completing empty when the
+     *         batch appends to a version an earlier batch in the same group already created
      */
     private Mono<DatasetVersion> handleGroupedInsertion(UUID batchGroupId, DatasetItemBatch batch,
             UUID datasetId, String workspaceId, String userName) {
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
@@ -1732,36 +1732,25 @@ class DatasetItemServiceImpl implements DatasetItemService {
 
     /**
      * Updates version counts after inserting items into an existing version.
-     * Extracted to reduce complexity and improve testability.
+     * Only the new items move the total; re-sent items count as modifications.
      *
      * @param versionId The version ID to update
      * @param workspaceId The workspace ID
      * @param newItemsCount Number of new items inserted
      * @param updatedItemsCount Number of items updated
      * @param userName The user performing the update
+     * @throws NotFoundException if the version no longer exists or does not belong to the workspace
      */
     private void updateVersionCountsForInsert(UUID versionId, String workspaceId, int newItemsCount,
             int updatedItemsCount, String userName) {
-        template.inTransaction(WRITE, handle -> {
-            var dao = handle.attach(DatasetVersionDAO.class);
-
-            // Only increment total by new items (not updates)
-            int updated = dao.incrementCounts(versionId, newItemsCount, newItemsCount, updatedItemsCount, 0,
-                    workspaceId, userName);
-
-            if (updated == 0) {
-                throw new NotFoundException("Version not found: '%s'".formatted(versionId));
-            }
-            return null;
-        });
+        updateVersionCounts(versionId, workspaceId, newItemsCount, newItemsCount, updatedItemsCount, 0, userName);
     }
 
     /**
      * Updates version counts after deleting items from an existing version.
      * <p>
-     * Applies the deltas atomically rather than writing back totals computed from a caller-supplied snapshot, so the
-     * result no longer depends on that snapshot still being current. Unlike the insert path this saves no round-trip:
-     * both callers fetch the version for their own purposes regardless.
+     * Unlike the insert path this saves no round-trip: both callers fetch the version for their own purposes
+     * regardless. What it buys is that the arithmetic no longer depends on that snapshot still being current.
      *
      * @param versionId The version ID to update
      * @param workspaceId The workspace ID
@@ -1770,10 +1759,23 @@ class DatasetItemServiceImpl implements DatasetItemService {
      * @throws NotFoundException if the version no longer exists or does not belong to the workspace
      */
     private void updateVersionCountsForDelete(UUID versionId, String workspaceId, int deletedCount, String userName) {
+        updateVersionCounts(versionId, workspaceId, -deletedCount, 0, 0, deletedCount, userName);
+    }
+
+    /**
+     * Applies signed counter deltas to a version in a single atomic statement, so the arithmetic does not depend on
+     * {@code withDatasetVersionLock} for mutual exclusion. A zero affected-row count means the version was removed or
+     * belongs to another workspace.
+     *
+     * @throws NotFoundException if the version no longer exists or does not belong to the workspace
+     */
+    private void updateVersionCounts(UUID versionId, String workspaceId, int totalDelta, int addedDelta,
+            int modifiedDelta, int deletedDelta, String userName) {
         template.inTransaction(WRITE, handle -> {
             var dao = handle.attach(DatasetVersionDAO.class);
 
-            int updated = dao.incrementCounts(versionId, -deletedCount, 0, 0, deletedCount, workspaceId, userName);
+            int updated = dao.incrementCounts(versionId, totalDelta, addedDelta, modifiedDelta, deletedDelta,
+                    workspaceId, userName);
 
             if (updated == 0) {
                 throw new NotFoundException("Version not found: '%s'".formatted(versionId));
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
@@ -1662,7 +1662,7 @@ class DatasetItemServiceImpl implements DatasetItemService {
         UUID latestVersionId = latestVersion.get().id();
         log.info("Inserting '{}' items into existing version '{}'", batch.items().size(), latestVersionId);
 
-        return insertItemsIntoVersion(batch, datasetId, latestVersionId, workspaceId, userName);
+        return insertItemsIntoVersion(batch, datasetId, latestVersionId, workspaceId, userName).then(Mono.empty());
     }
 
     /**
@@ -1670,14 +1670,17 @@ class DatasetItemServiceImpl implements DatasetItemService {
      * Handles validation, classification of new vs updated items, and count updates.
      * Used by mutateLatestVersionWithInsert and handleGroupedInsertion.
      *
+     * Completes empty: no caller reads the resulting version, so the counters are applied with a single atomic
+     * increment and the row is not read back.
+     *
      * @param batch the batch of items to insert
      * @param datasetId the dataset ID
      * @param versionId the version ID to insert into
      * @param workspaceId the workspace ID
      * @param userName the username
-     * @return Mono emitting the updated dataset version
+     * @return Mono completing when the items are inserted and the counts updated
      */
-    private Mono<DatasetVersion> insertItemsIntoVersion(DatasetItemBatch batch, UUID datasetId, UUID versionId,
+    private Mono<Void> insertItemsIntoVersion(DatasetItemBatch batch, UUID datasetId, UUID versionId,
             String workspaceId, String userName) {
         // Validate and prepare items
         List<DatasetItem> validatedItems = addIdIfAbsent(batch);
@@ -1730,11 +1733,10 @@ class DatasetItemServiceImpl implements DatasetItemService {
                                     // Insert items directly into the existing version
                                     return versionDao
                                             .insertItems(datasetId, versionId, normalizedItems, workspaceId, userName)
-                                            .then(Mono.fromCallable(() -> {
-                                                updateVersionCountsForInsert(versionId, workspaceId, finalNewItemsCount,
-                                                        finalUpdatedItemsCount, userName);
-                                                return versionService.getVersionById(workspaceId, datasetId, versionId);
-                                            }).subscribeOn(Schedulers.boundedElastic()));
+                                            .then(Mono.fromRunnable(() -> updateVersionCountsForInsert(versionId,
+                                                    workspaceId, finalNewItemsCount, finalUpdatedItemsCount, userName))
+                                                    .subscribeOn(Schedulers.boundedElastic()))
+                                            .then();
                                 });
                     }));
         }).contextWrite(c -> c.put(RequestContext.WORKSPACE_ID, workspaceId)
@@ -1755,43 +1757,36 @@ class DatasetItemServiceImpl implements DatasetItemService {
             int updatedItemsCount, String userName) {
         template.inTransaction(WRITE, handle -> {
             var dao = handle.attach(DatasetVersionDAO.class);
-            var currentVersion = dao.findById(versionId, workspaceId)
-                    .orElseThrow(() -> new NotFoundException(
-                            "Version not found: '%s'".formatted(versionId)));
 
             // Only increment total by new items (not updates)
-            int newTotal = currentVersion.itemsTotal() + newItemsCount;
-            int newAdded = currentVersion.itemsAdded() + newItemsCount;
-            int newModified = currentVersion.itemsModified() + updatedItemsCount;
+            int updated = dao.incrementCounts(versionId, newItemsCount, newItemsCount, updatedItemsCount, 0,
+                    workspaceId, userName);
 
-            dao.updateCounts(versionId, newTotal, newAdded, newModified,
-                    currentVersion.itemsDeleted(), workspaceId, userName);
+            if (updated == 0) {
+                throw new NotFoundException("Version not found: '%s'".formatted(versionId));
+            }
             return null;
         });
     }
 
     /**
      * Updates version counts after deleting items from an existing version.
-     * Extracted to reduce complexity and improve testability.
+     * <p>
+     * Applies the deltas atomically rather than writing back totals computed from a caller-supplied snapshot, so the
+     * result no longer depends on that snapshot still being current. Unlike the insert path this saves no round-trip:
+     * both callers fetch the version for their own purposes regardless.
      *
      * @param versionId The version ID to update
      * @param workspaceId The workspace ID
-     * @param currentVersion The current version before deletion
      * @param deletedCount Number of items deleted
      * @param userName The user performing the update
      */
-    private void updateVersionCountsForDelete(UUID versionId, String workspaceId, DatasetVersion currentVersion,
-            int deletedCount, String userName) {
-        int newTotal = currentVersion.itemsTotal() - deletedCount;
-        int newDeleted = currentVersion.itemsDeleted() + deletedCount;
-
-        log.info("deleteItemsFromExistingVersion: updating counts - newTotal='{}', newDeleted='{}'",
-                newTotal, newDeleted);
+    private void updateVersionCountsForDelete(UUID versionId, String workspaceId, int deletedCount, String userName) {
+        log.info("deleteItemsFromExistingVersion: updating counts - deletedCount='{}'", deletedCount);
 
         template.inTransaction(WRITE, handle -> {
             var dao = handle.attach(DatasetVersionDAO.class);
-            dao.updateCounts(versionId, newTotal, currentVersion.itemsAdded(),
-                    currentVersion.itemsModified(), newDeleted, workspaceId, userName);
+            dao.incrementCounts(versionId, -deletedCount, 0, 0, deletedCount, workspaceId, userName);
             return null;
         });
     }
@@ -2186,8 +2181,7 @@ class DatasetItemServiceImpl implements DatasetItemService {
 
                         // Update version counts in MySQL
                         return Mono.fromCallable(() -> {
-                            updateVersionCountsForDelete(versionId, workspaceId, currentVersion,
-                                    deletedCount.intValue(), userName);
+                            updateVersionCountsForDelete(versionId, workspaceId, deletedCount.intValue(), userName);
                             log.info("Deleted '{}' items from version '{}', new total '{}'",
                                     deletedCount, versionId, currentVersion.itemsTotal() - deletedCount.intValue());
                             return null;
@@ -2231,8 +2225,7 @@ class DatasetItemServiceImpl implements DatasetItemService {
 
                         // Update version counts in MySQL
                         return Mono.fromCallable(() -> {
-                            updateVersionCountsForDelete(versionId, workspaceId, currentVersion,
-                                    deletedCount.intValue(), userName);
+                            updateVersionCountsForDelete(versionId, workspaceId, deletedCount.intValue(), userName);
                             log.info("Deleted '{}' items from version '{}', new total '{}'",
                                     deletedCount, versionId, currentVersion.itemsTotal() - deletedCount.intValue());
                             return null;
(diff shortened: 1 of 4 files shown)


apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
```diff
@@ -1662,22 +1662,25 @@ private Mono<DatasetVersion> mutateLatestVersionWithInsert(DatasetItemBatch batc
         UUID latestVersionId = latestVersion.get().id();
         log.info("Inserting '{}' items into existing version '{}'", batch.items().size(), latestVersionId);
 
-        return insertItemsIntoVersion(batch, datasetId, latestVersionId, workspaceId, userName);
+        return insertItemsIntoVersion(batch, datasetId, latestVersionId, workspaceId, userName).then(Mono.empty());
     }
```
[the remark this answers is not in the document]
JetoPistola: Skipping — the `save()` contract already permits an empty completion, so this isn't a regression.

`DatasetItemService.save`'s javadoc documents "Returns empty Mono" / "or empty when disabled", and the legacy (versioning-disabled) branch has always returned `verifyDatasetExistsAndSave(batch).then(Mono.empty())`. An empty `Mono<DatasetVersion>` is a documented, pre-existing outcome of this method, not something introduced here.

No consumer reads the emitted value on either append path: `DatasetsResource.createDatasetItems` leaves the `.block()` result unassigned and returns HTTP 204, and `DatasetItemService.saveBatch` does `.map(version -> (long) items.size())` without dereferencing the parameter.

The suggested fix is to re-add `versionService.getVersionById(workspaceId, datasetId, latestVersionId)` after the insert — that is precisely the round-trip OPIK-7707 exists to remove (it was pure waste: a read of a row we had just written, whose result nothing consumed). Applying it would revert the change. Emitting a partially-populated `DatasetVersion` instead was also considered and rejected: it would look real to a future caller while carrying no guarantee of freshness.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetVersionDAO.java
```diff
@@ -539,22 +539,26 @@ void updateBatchGroupId(@Bind("version_id") UUID versionId,
...
-            SET items_total = :items_total,
-                items_added = :items_added,
-                items_modified = :items_modified,
-                items_deleted = :items_deleted,
+            SET items_total = items_total + :items_total_delta,
+                items_added = items_added + :items_added_delta,
+                items_modified = items_modified + :items_modified_delta,
+                items_deleted = items_deleted + :items_deleted_delta,
                 last_updated_at = NOW(),
                 last_updated_by = :last_updated_by
```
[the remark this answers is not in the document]
JetoPistola: Skipping — the signed-delta semantics are already stated in the javadoc directly above the signature ("Applies signed deltas to the version counters in a single statement"), so the ambiguity is addressed at the point a caller would read it.

The rename would touch 10 sites, including three test method names (`incrementCounts__whenConcurrentWritersBypassTheLock__thenCountersAreExact` and friends), for no behavioral change. There's also no `applyXDeltas` precedent in this codebase to align with. Happy to revisit if a reviewer feels strongly, but it reads as taste rather than a correctness or clarity gap worth the churn on this PR.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
```diff
@@ -1755,43 +1757,36 @@ private void updateVersionCountsForInsert(UUID versionId, String workspaceId, in
...
-        log.info("deleteItemsFromExistingVersion: updating counts - newTotal='{}', newDeleted='{}'",
-                newTotal, newDeleted);
+    private void updateVersionCountsForDelete(UUID versionId, String workspaceId, int deletedCount, String userName) {
+        log.info("deleteItemsFromExistingVersion: updating counts - deletedCount='{}'", deletedCount);
 
         template.inTransaction(WRITE, handle -> {
             var dao = handle.attach(DatasetVersionDAO.class);
-            dao.updateCounts(versionId, newTotal, currentVersion.itemsAdded(),
-                    currentVersion.itemsModified(), newDeleted, workspaceId, userName);
+            dao.incrementCounts(versionId, -deletedCount, 0, 0, deletedCount, workspaceId, userName);
```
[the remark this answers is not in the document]
JetoPistola: Fixed in 0eb086558e — both parts addressed.

**Affected-row count**: `updateVersionCountsForDelete` now captures the return of `dao.incrementCounts` and throws `NotFoundException` when it's `0`, mirroring the insert path exactly. Javadoc updated with the `@throws`.

One correction on the stated failure mode: it isn't reachable today. Both delete callers (`deleteItemsFromExistingVersion` and `deleteItemsFromExistingVersionByFilters`) call `versionService.getVersionById(...)` before reaching this helper, and that already throws `NotFoundException` for a missing or cross-workspace version — so the flow could not have reported success with counters unchanged. The reason I still applied it: insert and delete were handling a zero-row result inconsistently, and that asymmetry is a live trap for the follow-up work this ticket unblocks (narrowing `withDatasetVersionLock` and dropping the pre-fetch), at which point the guard stops being redundant.

Added `incrementCounts__whenVersionBelongsToAnotherWorkspace__thenNoRowsAffected` to cover the cross-workspace case directly — real version id, foreign workspace, asserts zero rows affected and counters untouched.

**Logging**: removed. You're right that it duplicated the callers' summary — both already log `deletedCount`, `versionId`, and the resulting total two frames up — so rather than demote it to `DEBUG` I dropped it, since the caller line carries strictly more context.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
```diff
@@ -5207,6 +5209,168 @@ void batchUpdate__whenEmptyDescription__thenDescriptionIsCleared() {
...
+            // The delete path leaves added/modified untouched.
+            assertThat(updated.itemsAdded()).isEqualTo(seed.itemsAdded());
+            assertThat(updated.itemsModified()).isEqualTo(seed.itemsModified());
+        }
+
+        @Test
+        @DisplayName("Success: increment against an unknown version affects no rows")
+        void incrementCounts__whenVersionDoesNotExist__thenNoRowsAffected() {
+            assertThat(incrementCounts(UUID.randomUUID(), 1, 1, 0, 0)).isZero();
+        }
```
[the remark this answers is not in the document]
JetoPistola: Skipping — the suggested refactor isn't implementable as written, and the branch isn't reachable from this test class.

`updateVersionCountsForInsert` is a **private** method on `DatasetItemServiceImpl`. `DatasetVersionResourceTest` is an HTTP-level resource test: it drives the running app through `DatasetResourceClient` and only holds a `TransactionTemplate` for direct DB assertions — it never injects `DatasetItemService`, so it cannot call that helper regardless of the assertion style used.

Reaching the `NotFoundException` through the public API isn't possible either. Both delete callers, and the insert path's caller, resolve the version before the count update; to hit `updated == 0` the row would have to disappear (or change workspace) in the window between that resolution and the increment. There's no API-level way to produce that interleaving.

What the two DAO tests do assert is the contract the guard is built on: that `incrementCounts` matches zero rows for an unknown version id and for a real id under a foreign workspace. The guard immediately above them is `if (updated == 0) throw new NotFoundException(...)`, so the DAO-level assertion is the part with actual failure modes — a missing workspace predicate in the SQL, for instance, would break the foreign-workspace test.

Happy to revisit if we later add a service-level test class where the helper is reachable; forcing it here would mean widening the method's visibility purely for a test, which trades real encapsulation for a branch that can't fire in production today.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
```diff
@@ -1755,43 +1757,40 @@ private void updateVersionCountsForInsert(UUID versionId, String workspaceId, in
...
         template.inTransaction(WRITE, handle -> {
             var dao = handle.attach(DatasetVersionDAO.class);
-            dao.updateCounts(versionId, newTotal, currentVersion.itemsAdded(),
-                    currentVersion.itemsModified(), newDeleted, workspaceId, userName);
+
+            int updated = dao.incrementCounts(versionId, -deletedCount, 0, 0, deletedCount, workspaceId, userName);
+
+            if (updated == 0) {
+                throw new NotFoundException("Version not found: '%s'".formatted(versionId));
+            }
```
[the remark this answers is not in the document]
JetoPistola: Fixed in 0153bc85dc — both now delegate to a shared `updateVersionCounts(versionId, workspaceId, totalDelta, addedDelta, modifiedDelta, deletedDelta, userName)` that owns the transaction, the `incrementCounts` call, and the zero-row `NotFoundException` guard.

Fair catch, and the duplication was self-inflicted: the two helpers only converged into near-identical shape in the previous commit, when the delete path gained the same affected-row guard as insert. At that point the transaction boilerplate, the DAO call, and the error handling were the same in both, with only the delta arguments differing — exactly the drift risk you describe.

I kept `updateVersionCountsForInsert` / `updateVersionCountsForDelete` as thin named wrappers rather than calling the shared helper directly from the call sites. They document which counters each path moves — insert puts new items into both `items_total` and `items_added` while re-sent items only bump `items_modified`; delete moves `items_total` down and `items_deleted` up — and that mapping is the part worth naming. The mechanism now lives in one place.

JetoPistola:
Consolidated duplicate work from #7928 into this PR (now closed).

I'd independently implemented the same change there without noticing this PR existed. This one is the better base — it has the DAO-level lock-bypass test that actually proves the arithmetic is atomic rather than merely serialised by `withDatasetVersionLock`, plus DAO coverage for negative deltas, unknown version, and cross-workspace isolation.

Ported here in 7ad36cf354, one substantive gap the other PR had caught:

**`COALESCE(col, 0)` on all four counters.** The columns are `INT DEFAULT 0` in migration `000036` — nullable, since `DEFAULT` only applies when the column is omitted on INSERT — and `DatasetVersion` maps all four as boxed `Integer`. A bare `col + :delta` leaves NULL as NULL, which then unboxes to an NPE on read. The absolute `updateCounts` this replaces happened to repair a NULL by overwriting it, so without the COALESCE the switch to increments is a regression in NULL tolerance.

Added a test that nulls the counters directly and asserts the increment treats them as zero — verified it fails without the COALESCE and passes with it.

Nothing else from #7928 was worth porting. In particular I did **not** bring over its removal of the delete-path `NotFoundException`: a bot review there flagged it as a 404 regression, but 0eb086558e's reasoning here is correct — both delete callers pre-fetch via `getVersionById`, which already throws, so that path is unreachable today and the throw keeps insert/delete symmetric for later work that drops the pre-fetch.

Suite is green locally: 125/125 in `DatasetVersionResourceTest`.

Note this branch is still based on Aug 3 and hasn't had main merged in — worth doing before merge.

🤖 *Comment posted via /address-github-pr-comments*

CometActions:

The delta rewrite of the version counters is covered on both write paths. Insert: tests_end_to_end/e2e/tests/datasets/dataset-version-counters.spec.ts pins items_total/added/modified after multi-batch and 8-thread parallel inserts and reads the same numbers back off the Version history tab, so a wrong sign or a swapped argument in incrementCounts fails it. Delete: the draft stacked on this branch, #7705 -> #7965, adds dataset-version-delete-counters.spec.ts, which asserts the full total/added/modified/deleted quadruple across interleaved inserts, upserts and deletes (the existing delete specs only count rows). Nothing new to propose. The ITEMS_TOTAL_NOT_MIGRATED (-1) seeding in ensureVersion1Exists is the one part we can't reach: it only appears on a dataset lazily migrated from the pre-versioning tables, which a fresh OSS install never produces -- your DatasetVersionResourceTest covers those semantics including the deliberate 404, so we're recording it as deferred rather than proposing a spec we can't run.

**Not testable yet.** ensureVersion1Exists now seeds items_total with the -1 not-migrated sentinel instead of 0, and incrementCounts / updateItemsTotal / batchUpdateItemsTotal are all gated on it. That is user-facing on a legacy dataset: while the row holds the sentinel the Version history tab has no real Item count to show, and an item batch written into that window is refused with 404 by design. It is only reachable on an install that already has pre-versioning dataset_items rows with no dataset_versions row, which the e2e estate cannot produce -- every dataset it creates is versioned from birth.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/33093131300)

_Re-checked after a push on 27 Aug 16:33 UTC._

CometActions:

### Access Information
- **URL:** https://pr-7705.dev.comet.com
- **Namespace:** pr-7705
- **Version:** 2.2.14-7705-merge-3043
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7705%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7705`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

CometActions:

### Access Information
- **URL:** https://pr-7705.dev.comet.com
- **Namespace:** pr-7705
- **Version:** 2.2.14-7705-merge-3064
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7705%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7705`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

CometActions:

### Access Information
- **URL:** https://pr-7705.dev.comet.com
- **Namespace:** pr-7705
- **Version:** 2.2.14-7705-merge-3102
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7705%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

thiagohora (commented):

8 findings — 0 high · 8 medium · 0 low. Suppressed by team conventions: see `suppressed.md`.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
```diff
@@ -1662,22 +1662,25 @@ private Mono<DatasetVersion> mutateLatestVersionWithInsert(DatasetItemBatch batc
         UUID latestVersionId = latestVersion.get().id();
         log.info("Inserting '{}' items into existing version '{}'", batch.items().size(), latestVersionId);
 
-        return insertItemsIntoVersion(batch, datasetId, latestVersionId, workspaceId, userName);
+        return insertItemsIntoVersion(batch, datasetId, latestVersionId, workspaceId, userName).then(Mono.empty());
```
thiagohora, line 1663: **[medium] Empty completion is now the success signal, making a lock-acquire timeout indistinguishable from a successful write**

Both in-place-append branches now complete empty (`insertItemsIntoVersion(...).then(Mono.empty())` here, and `.then(Mono.<DatasetVersion>empty())` at line 2360), so an empty completion of `save(DatasetItemBatch)` is the normal success path. `save` is wrapped in `withDatasetVersionLock` -> `LockService.executeWithLockCustomExpire`, and `RedissonLockService.instrumentedAcquire` explicitly documents that the acquire Mono can complete empty 'when the wait elapsed without a permit'; in that case the outer `flatMap` yields an empty Mono and the batch is never written. Previously the append path emitted the mutated version, so an empty completion still carried information. Now `DatasetsResource.createDatasetItems` (line 534-542) blocks on the chain and returns 204 No Content whether the items were persisted or the lock was never acquired and nothing ran at all — a silently dropped batch that no caller, log line, or metric on this path can distinguish from success.

---
**Note from the review pipeline's verification pass: this finding was REFUTED, and is posted for discussion rather than as a defect.** Counter-evidence: `executeWithLockCustomExpire` calls Redisson's `RPermitExpirableSemaphoreReactive.acquire(leaseTime, unit)`, which waits indefinitely and returns a permit id — not the wait-bounded `tryAcquire(waitTime, ...)`, which is the only variant that can complete empty. In this repo `tryAcquire` is reached only via `bestEffortLock` (`RedissonLockService:210`) and `tryAcquireSlot` (`:281`), neither of which is on the `save()` path. Redis-level failures on this path surface as errors (`.retryWhen(Retry.max(3)...)` then propagate), i.e. a 5xx rather than a 204. The `instrumentedAcquire` javadoc quoted above describes `tryAcquire` semantics and its null-branch is defensive/dead for this acquire. Worth confirming, since if that reading is right the surrounding javadoc is misleading.

💡 Handle the empty case explicitly at the lock boundary rather than relying on it as the success signal: e.g. make `withDatasetVersionLock` (or each call site) append `.switchIfEmpty(Mono.error(new ...ClientErrorException/IllegalStateException("could not acquire dataset version lock for dataset '%s'")))`, or have the guarded action end with a sentinel value (`.thenReturn(...)` / `Mono<Boolean>`) so lock-not-acquired stays observable. Keep `save`'s emission contract non-empty on the versioned path, or change the declared return type (see the javadoc finding) so callers know empty is expected.

_rule: `backend-correctness-get-reactive-project-reactor-signal-handling-right`_

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
```diff
@@ -1662,22 +1662,25 @@ private Mono<DatasetVersion> mutateLatestVersionWithInsert(DatasetItemBatch batc
...
     }
 
     /**
      * Shared method to insert items into an existing version.
      * Handles validation, classification of new vs updated items, and count updates.
      * Used by mutateLatestVersionWithInsert and handleGroupedInsertion.
      *
+     * Completes empty: no caller reads the resulting version, so the counters are applied with a single atomic
```
thiagohora, line 1671: **[medium] Public `save` contract and `handleGroupedInsertion` javadoc still promise a DatasetVersion that is no longer emitted**

The diff documents the new empty completion on `insertItemsIntoVersion` but leaves the two contracts above it stale. The `DatasetItemService.save` interface javadoc still states '<li>Returns the DatasetVersion (newly created or mutated)</li>' (line 135) and '@return Mono emitting the DatasetVersion when versioning is enabled, or empty when disabled' (line 144), which is now wrong for both the no-batch_group_id branch (line 1665) and the append-to-existing-batch-group branch (line 2360) — versioning-enabled inserts into an existing version emit nothing. `handleGroupedInsertion`'s own '@return Mono emitting the dataset version' (line 2347) is wrong for the same reason. A future caller that adds `.flatMap(version -> ...)` on `save` will silently no-op for the most common (append) path.

💡 Update the `save` interface javadoc and `handleGroupedInsertion`'s @return to state that only the create-new-version branches emit a version and the append branches complete empty. Better still, since the only production caller (`DatasetsResource.createDatasetItems`) discards the value, narrow `save` to `Mono<Void>` (and `handleGroupedInsertion` accordingly) so the type reflects the real contract instead of a value that is emitted on only one of three paths.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
```diff
@@ -1743,55 +1745,54 @@ private Mono<DatasetVersion> insertItemsIntoVersion(DatasetItemBatch batch, UUID
...
-            return null;
-        });
+        updateVersionCounts(versionId, workspaceId, newItemsCount, newItemsCount, updatedItemsCount, 0, userName);
     }
 
     /**
      * Updates version counts after deleting items from an existing version.
-     * Extracted to reduce complexity and improve testability.
+     * <p>
+     * Unlike the insert path this saves no round-trip: both callers fetch the version for their own purposes
```
thiagohora: **[medium] Javadoc justification is false: after dropping the snapshot parameter, both delete callers fetch the version only to log it**

'both callers fetch the version for their own purposes regardless' is not true after this change. `currentVersion` in `deleteItemsFromExistingVersion` (assigned line 2169) is now used only at the log statements on lines 2172-2173 and 2192; in `deleteItemsFromExistingVersionByFilters` (assigned line 2214) only at lines 2216-2218 and 2236. So each batched delete still pays a synchronous MySQL round-trip (`versionService.getVersionById`) purely to produce two INFO lines — and that blocking JDBI call sits bare inside `Mono.defer`, with no `Mono.fromCallable(...).subscribeOn(Schedulers.boundedElastic())` wrapper, unlike every other blocking hop in this class (e.g. lines 2189-2194, 2317-2318).

💡 Drop the now-purposeless `versionService.getVersionById` read from both delete helpers and demote/rewrite the log lines so they only report values the method actually has (`versionId`, `ids.size()`, `deletedCount`). If the pre-delete snapshot is intentionally kept for diagnostics, wrap it in `Mono.fromCallable(...).subscribeOn(Schedulers.boundedElastic())` and correct the javadoc sentence to say the fetch survives only for logging.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
```diff
@@ -1743,55 +1745,54 @@ private Mono<DatasetVersion> insertItemsIntoVersion(DatasetItemBatch batch, UUID
...
+            int modifiedDelta, int deletedDelta, String userName) {
         template.inTransaction(WRITE, handle -> {
             var dao = handle.attach(DatasetVersionDAO.class);
-            dao.updateCounts(versionId, newTotal, currentVersion.itemsAdded(),
-                    currentVersion.itemsModified(), newDeleted, workspaceId, userName);
+
+            int updated = dao.incrementCounts(versionId, totalDelta, addedDelta, modifiedDelta, deletedDelta,
+                    workspaceId, userName);
+
+            if (updated == 0) {
```
thiagohora, line 1778: **[medium] New affected-rows guard has no test that reaches it through the service**

`if (updated == 0) throw new NotFoundException(...)` is a new semantically distinct branch: it converts a zero affected-row count into a 404 on both the insert and the delete path, where the delete path previously issued an unconditional UPDATE and silently no-op'd when the version had vanished. The tests added in `DatasetVersionResourceTest.AtomicVersionCountUpdates` only assert the DAO's return value (`incrementCounts__whenVersionDoesNotExist__thenNoRowsAffected`, line 5383, and `...whenVersionBelongsToAnotherWorkspace...`, line 5389) by calling `DatasetVersionDAO.incrementCounts` directly — nothing exercises the `throw` in this service, so the branch's HTTP-visible behavior (404 vs. the old silent success) is untested.

💡 Add a service/API-level case that reaches the guard — e.g. delete the version row (or point the flow at a version in another workspace) between the ClickHouse mutation and the counter update, or inject a `DatasetVersionDAO` stub returning 0 — and assert the resulting 404 and that the exception message carries the version id. Per backend-testing-exercise-dao-service-behavior-through-the-public-api-not-dir this should go through the public API rather than a direct DAO call.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetVersionDAO.java
```diff
@@ -539,22 +539,26 @@ void updateBatchGroupId(@Bind("version_id") UUID versionId,
...
+     * Applies signed deltas to the version counters in a single statement, so the arithmetic happens in the database
+     * rather than as a read-modify-write in the application. Correct under concurrent callers without an external lock.
+     */
     @SqlUpdate("""
             UPDATE dataset_versions
-            SET items_total = :items_total,
-                items_added = :items_added,
-                items_modified = :items_modified,
-                items_deleted = :items_deleted,
+            SET items_total = COALESCE(items_total, 0) + :items_total_delta,
```
thiagohora, line 568: **[medium] Increment silently destroys the items_total = -1 migration sentinel**

`items_total = -1` is the sentinel for "version created by Liquibase 000046, not yet backfilled" (see liquibase/db-app-state/migrations/000046_migrate_dataset_versions_initial_data.sql, and findVersionsNeedingItemsTotalMigration at line 645 of this file, which gates on `items_total = -1`). The new statement adds a delta to whatever is in the column, so the first insert/delete against a not-yet-backfilled v1 turns -1 into -1+delta. The row then no longer matches the `items_total = -1` predicate, the background migration (DatasetVersioningMigrationService.migrateItemsTotalInBatches) skips it forever, and the counter is permanently wrong by the dataset's whole pre-existing item count. Nothing anywhere gates the increment on the sentinel condition that produced it, and the new test class (DatasetVersionResourceTest.AtomicVersionCountUpdates) has no case for a version sitting at -1. The old absolute `updateCounts` had the same arithmetic outcome, so this is carried over rather than introduced — but the DAO comment now presents this statement as the authoritative, self-contained counter path, which is where the sentinel gate belongs.

💡 Gate the sentinel explicitly instead of doing arithmetic on it: either add `AND items_total <> -1` to the WHERE clause so the affected-row count is 0 and the caller can force `ensureDatasetMigrated`/the items_total backfill first, or express it in the SET clause (`items_total = IF(items_total = -1, -1, COALESCE(items_total, 0) + :items_total_delta)`). Declare -1 once as a shared constant referenced by both this statement and findVersionsNeedingItemsTotalMigration, and add a test that increments a version left at the sentinel.

_rule: `backend-correctness-keep-sentinel-magic-values-centralized-consistent-and-gated-`_

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetVersionDAO.java
```diff
@@ -539,22 +539,26 @@ void updateBatchGroupId(@Bind("version_id") UUID versionId,
             @Bind("workspace_id") String workspaceId,
             @Bind("last_updated_by") String lastUpdatedBy);
 
+    /**
+     * Applies signed deltas to the version counters in a single statement, so the arithmetic happens in the database
+     * rather than as a read-modify-write in the application. Correct under concurrent callers without an external lock.
```
thiagohora: **[medium] Sibling absolute writers still clobber the atomic increments (lost update)**

The Javadoc promises correctness "under concurrent callers without an external lock", but two other statements in this same interface still write items_total absolutely from a prior read: updateItemsTotal (line 605) and batchUpdateItemsTotal (line 624), both driven by DatasetVersioningMigrationService, which takes no per-dataset version lock (runItemsTotalMigration comment says locking is handled "at the job level", and the batch path holds nothing that excludes API writers). Realistic sequence: the job selects version v at -1, counts 10 items in ClickHouse, a concurrent batch insert of 2 items runs incrementCounts (items_total -> 1) and lands the rows in ClickHouse, then the job writes the stale absolute 10 — the increment is lost and the true value is 12. The counter guarantee only holds if every writer of these columns is a delta, or the absolute writers are gated on the state they read.

💡 Make the migration writes conditional on the state they were selected under — add `AND items_total = -1` to updateItemsTotal/batchUpdateItemsTotal so a row already touched by an increment is skipped rather than overwritten — and derive success from the affected-row count. Then narrow the Javadoc to state that the guarantee holds only for delta writers.

_rule: `backend-concurrency-guard-check-then-act-write-races-with-a-db-unique-constraint`_

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
```diff
@@ -5207,6 +5209,198 @@ void batchUpdate__whenEmptyDescription__thenDescriptionIsCleared() {
...
+            assertThat(updated)
+                    .extracting(DatasetVersion::itemsTotal, DatasetVersion::itemsAdded,
+                            DatasetVersion::itemsModified, DatasetVersion::itemsDeleted)
+                    .containsExactly(3, 3, 1, 0);
+        }
+
+        @Test
+        @DisplayName("Success: increment against an unknown version affects no rows")
+        void incrementCounts__whenVersionDoesNotExist__thenNoRowsAffected() {
+            assertThat(incrementCounts(UUID.randomUUID(), 1, 1, 0, 0)).isZero();
```
thiagohora, line 5567: **[medium] New "0 rows affected -> NotFoundException" guard is never exercised**

The PR adds a brand-new throw branch in DatasetItemServiceImpl.updateVersionCounts (`if (updated == 0) throw new NotFoundException("Version not found: '%s'")`). On the delete path this is a behavior change: the old updateCounts was `void`, so a vanished/foreign version silently no-op'd; now it surfaces as a 404 after the ClickHouse rows have already been removed. The two new tests that touch this scenario (incrementCounts__whenVersionDoesNotExist__thenNoRowsAffected at 5383 and incrementCounts__whenVersionBelongsToAnotherWorkspace__thenNoRowsAffected at 5389) stop at the DAO and only assert the affected-row count is 0 — nothing asserts the guard converts that into a NotFoundException, nor what the HTTP caller sees. Grep confirms no test anywhere asserts the "Version not found" message or a 404 from the item-insert/delete endpoints. A regression that drops the `updated == 0` check, or swaps it for a silent return, passes the whole suite.

💡 Add a service/API-level test for the guard: e.g. delete the dataset_versions row (or point the increment at a foreign workspace) between the item removal and the count update, or call the private helper's public entry point with a version that was concurrently removed, and assert NotFoundException with the exact message (rule backend-testing-test-exception-type-and-message-precisely). At minimum assert the endpoint's resulting status code so the silent-no-op -> 404 change on the delete path is pinned.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
```diff
@@ -5207,6 +5209,198 @@ void batchUpdate__whenEmptyDescription__thenDescriptionIsCleared() {
...
+            assertThat(updated.id()).isEqualTo(seed.id());
+            assertThat(updated.itemsTotal()).isEqualTo(5); // 3 + 2 new (the update doesn't count)
+            assertThat(updated.itemsAdded()).isEqualTo(5);
+            assertThat(updated.itemsModified()).isEqualTo(1);
+            assertThat(updated.itemsDeleted()).isZero();
+        }
+
+        @Test
+        @DisplayName("Success: multi-batch append via batch_group_id accumulates the full counter triple")
+        void insertItems__whenAppendingViaBatchGroupId__thenCountersAccumulate() {
```
thiagohora: **[medium] Two new tests duplicate existing tests in the same class**

insertItems__whenAppendingViaBatchGroupId__thenCountersAccumulate (5265) is the same arrange/act/assert as the existing putItems_whenSameBatchId_thenSingleVersion (line 3562): 3 batches sharing one batchGroupId, then assert listVersions has 1 entry and itemsTotal/itemsAdded == 6. The only delta is the two extra `isZero()` assertions on itemsModified/itemsDeleted, and the new copy actually drops the existing test's stronger check that the 6 rows are really readable. Likewise incrementCounts__whenNegativeDelta__thenTotalDecrements (5330) duplicates deleteItems__whenNoBatchGroupId__thenMutateLatestVersion (line 1627): same 5-item seed, same delete of the first item by datasetItemId with no batchGroupId, same `updated.id() == seed.id()` and `itemsTotal == 4` assertions; only the itemsDeleted/itemsAdded/itemsModified assertions are new, and again the row-level verification is dropped.

💡 Delete the two new near-duplicates and add the missing counter assertions (itemsModified/itemsDeleted, and added/modified untouched on delete) to the existing tests at lines 3562 and 1627 instead, so each flow keeps exactly one test that covers both the version topology and the counters.

_rule: `backend-testing-collapse-near-duplicate-tests-into-a-single-parameterizedtes`_

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7705`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
JetoPistola: Skipping — agreeing with your verification pass: this one is refuted.

I checked the same path independently. `executeWithLockCustomExpire` calls `RPermitExpirableSemaphoreReactive.acquire(leaseTime, unit)`, which waits indefinitely and returns a permit id. The wait-bounded `tryAcquire(waitTime, ...)` — the only variant that completes empty — is reached solely via `bestEffortLock` and `tryAcquireSlot`, neither of which is on the `save()` path. So an empty completion here cannot mean "lock not acquired"; Redis failures surface as errors and become a 5xx, not a silent 204.

Your closing point is the useful one though: the `instrumentedAcquire` javadoc describes `tryAcquire` semantics and reads as if it applies to this acquire too. That's misleading, and it's what made this finding plausible in the first place. Out of scope for this PR, but worth a follow-up on `RedissonLockService`.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
JetoPistola: Fixed — took the stronger option you suggested: `save` is now `Mono<Void>` rather than a corrected javadoc on a `Mono<DatasetVersion>`.

The only production caller (`DatasetsResource.createDatasetItems`) blocks and returns 204, discarding the value, so the type now states the contract instead of promising a version that arrives on one path in three. `handleGroupedInsertion` keeps its `Mono<DatasetVersion>` — it genuinely emits on the create branch — but its `@return` now says the append branch completes empty.

Left the internal producers (`saveItemsWithVersion`, `createFirstVersion`, `createVersionWithDelta`) alone: those legitimately produce versions and are reused elsewhere, so narrowing the whole chain would go well past this ticket.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
JetoPistola: Fixed — you're right that the javadoc sentence became false, and the read behind it was worse than the stale comment.

Verified before changing: after the snapshot parameter went away, `currentVersion` in both delete helpers is referenced only by log statements. So each batched delete was paying a synchronous `getVersionById` round-trip purely to enrich two INFO lines — and as you noted, that blocking JDBI call sat bare inside `Mono.defer` with no `subscribeOn(boundedElastic())`, unlike every other blocking hop in the class.

Took the first of your two options: dropped the read from both helpers and rewrote the logs to report only what the method actually has (`versionId`, `ids.size()`, `deletedCount`). The javadoc now says the delta removed the last reason to read the version, instead of claiming the callers needed it regardless.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
JetoPistola: Fixed — added a service-level test that reaches the guard through the public API.

My first attempt deleted the `dataset_versions` row before the insert, as you suggested. That returned **204, not 404**: with no latest version, `mutateLatestVersionWithInsert` simply mints a new one, so the guard is never reached. Useful failure — it showed the guard needs a version that still resolves as latest but that the count UPDATE refuses to match.

The not-migrated sentinel is exactly that state, and it's a real one: an un-backfilled version written to before the migration job reaches it. `insertItems__whenCountUpdateMatchesNoRow__thenNotFound` puts a version at the sentinel, posts an insert, and asserts 404 — plus that the sentinel is still intact afterwards, so the backfill can still find the row.

Kept the two DAO-level tests as well. They pin the affected-row contract the guard depends on, which is a different assertion from the HTTP behaviour.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetVersionDAO.java
JetoPistola: Fixed — this was the most valuable finding in the set. Confirmed reachable before fixing: the sentinel is real (`findVersionsNeedingItemsTotalMigration` gates on `items_total = -1`), and `insertItemsIntoVersion` has no `ensureLazyMigration` guard, so an un-backfilled version can absolutely be incremented.

Took your WHERE-clause option so the caller can react rather than silently writing a wrong value. Also declared the sentinel once as `DatasetVersionDAO.ITEMS_TOTAL_NOT_MIGRATED` and referenced it from every statement that depends on it, per your suggestion.

**One trap worth recording.** My first version used `items_total <> :sentinel`, which broke the NULL-counter test from the sibling work: SQL three-valued logic makes `NULL <> -1` evaluate to NULL, not TRUE, so a NULL counter was skipped instead of incremented — quietly re-breaking what the `COALESCE` exists to fix. The predicate is now the NULL-safe `NOT (items_total <=> :sentinel)`, and the javadoc says why.

Added `incrementCounts__whenVersionHoldsNotMigratedSentinel__thenNoRowsAffected`, asserting both that the increment is skipped and that the sentinel survives intact so the backfill still finds the row.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetVersionDAO.java
JetoPistola: Fixed — both halves of your suggestion.

You're right that the javadoc overclaimed. `updateItemsTotal` and `batchUpdateItemsTotal` both write an absolute value derived from an earlier ClickHouse count, and the migration service holds nothing that excludes API writers, so your interleaving is real: job counts 10 → an insert increments → job writes 10 and the increment is lost.

Both are now gated on `items_total = :items_total_not_migrated`, so a row an increment has already touched is skipped rather than clobbered. `updateItemsTotal` returns the affected-row count (was `void`); the caller currently discards it, but the value is there for a follow-up that wants to report skipped rows. Verified both are only reached from `DatasetVersioningMigrationService`, so gating them can't affect an API path.

The `incrementCounts` javadoc now says the guarantee holds among *delta* writers and names the absolute writers as the exception, instead of claiming blanket correctness without an external lock.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
JetoPistola: Fixed — same change as the guard finding above.

`insertItems__whenCountUpdateMatchesNoRow__thenNotFound` now asserts the HTTP status from the item-insert endpoint, so the "silent no-op → 404" behaviour change is pinned at the level a caller actually sees. A regression that drops the `updated == 0` check, or swaps it for a silent return, fails that test.

One deliberate deviation: I assert the status code rather than the exception type and message. The guard is only reachable through the resource, which maps `NotFoundException` to 404 and does not surface the message in a way the test client can assert on without over-fitting to the error envelope. The status is the contract a caller depends on.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
JetoPistola: Fixed — deleted both near-duplicates and moved their unique assertions into the existing tests, exactly as suggested.

Compared them line by line first and your read holds: `putItems_whenSameBatchId_thenSingleVersion` (3 batches, one `batch_group_id`, 1 version, total/added = 6) and `deleteItems__whenNoBatchGroupId__thenMutateLatestVersion` (5-item seed, delete first by `datasetItemId`, same version id, total = 4) already covered the same flows — and both assert row-level state my copies had dropped.

So the originals now also assert `itemsModified`/`itemsDeleted` on the append path, and `itemsDeleted` plus added/modified-unchanged on the delete path. Net: two fewer tests, and each flow keeps one test covering both version topology and counters.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetVersionDAO.java
```diff
@@ -591,20 +629,41 @@ int ensureVersion1Exists(@Bind("dataset_id") UUID datasetId,
...
             @Bind("version_id") UUID versionId,
-            @Bind("items_total") long itemsTotal);
+            @Bind("items_total") long itemsTotal,
+            @Bind("items_total_not_migrated") int itemsTotalNotMigrated);
+
+    /**
+     * Backfills {@code items_total}, supplying the not-migrated sentinel so callers cannot forget the guard.
+     */
+    default int updateItemsTotal(String workspaceId, UUID versionId, long itemsTotal) {
+        return updateItemsTotal(workspaceId, versionId, itemsTotal, ITEMS_TOTAL_NOT_MIGRATED);
```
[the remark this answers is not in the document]
JetoPistola: Fixed in 86eeaece27 — this was a regression I introduced in the previous round, and your diagnosis is exactly right.

Confirmed before changing anything: `ensureVersion1Exists` seeds `items_total = 0`, not the sentinel. So the gate I added on `updateItemsTotal` made it match zero rows on every lazily-migrated dataset, and `countAndUpdateItemsTotal` discarded that result — the real count was never written, the version sat at `0` forever, and the batch backfill couldn't repair it either since it also selects only on `-1`.

Took your first option: the lazy path now seeds `ITEMS_TOTAL_NOT_MIGRATED`. A literal `0` on a row created *before* its items are counted is indistinguishable from a genuinely empty version, which is precisely what let this hide.

Also stopped discarding the affected-row count in `countAndUpdateItemsTotal`. Zero rows there now means an API write already moved the counter off the sentinel, so the stale count is correctly skipped — logged rather than thrown, since skipping is the right outcome in that race.

Worth recording the pattern: this is the second way that sentinel gate has bitten. The first was `NULL <> -1` evaluating to NULL under three-valued logic; this one was a caller ignoring the row count. A `WHERE` gate silently converts "didn't happen" into "no rows", and every call site that discards the count inherits the bug.

134/134 green.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetVersionDAO.java
```diff
@@ -539,24 +547,54 @@ void updateBatchGroupId(@Bind("version_id") UUID versionId,
...
+            SET items_total = COALESCE(items_total, 0) + :items_total_delta,
+                items_added = COALESCE(items_added, 0) + :items_added_delta,
+                items_modified = COALESCE(items_modified, 0) + :items_modified_delta,
+                items_deleted = COALESCE(items_deleted, 0) + :items_deleted_delta,
                 last_updated_at = NOW(),
                 last_updated_by = :last_updated_by
             WHERE id = :version_id
               AND workspace_id = :workspace_id
+              AND NOT (items_total <=> :items_total_not_migrated)
             """)
```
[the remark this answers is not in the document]
JetoPistola: Skipping — real, but pre-existing and out of scope here.

The ordering you describe is accurate: ClickHouse rows are written before the MySQL counter update runs, so a version in sentinel state gets its items persisted and then the request 404s. That split isn't introduced by this PR though — the two stores were never updated atomically, and the same window existed when the counter update was an absolute `SET`. What changed is only that the mismatch is now *visible* (a 404) instead of silently producing a wrong counter, which I'd argue is the better failure.

Your suggested remedy — validate or backfill the version first, and reconcile on both paths — is the right shape, but it needs a reconciliation mechanism that doesn't exist yet. That's the substance of OPIK_7804 (making a drifted `items_total` converge back to the ClickHouse ground truth), and building it here would widen a counter-arithmetic refactor into a cross-store consistency change.

Note the sentinel case specifically gets narrower with 86eeaece27: the lazy path now seeds the sentinel and the backfill can actually complete, so the window where a live dataset sits at `-1` is smaller than when you filed this.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemService.java
```diff
@@ -132,18 +132,15 @@ Mono<ProjectStats> getExperimentItemsStats(UUID datasetId, Set<UUID> experimentI
...
+     * When versioning is disabled (legacy mode), saves items to the legacy dataset_items table.
      *
      * @param batch the batch of items to save (must include datasetId or datasetName, may include batchGroupId)
-     * @return Mono emitting the DatasetVersion when versioning is enabled, or empty when disabled
+     * @return Mono completing when the batch is persisted. No version is emitted: appending to an existing version
+     *         has nothing to read back, and the only caller discards the value, so the type says so rather than
+     *         promising a value that arrives on some paths and not others.
      */
-    Mono<DatasetVersion> save(DatasetItemBatch batch);
+    Mono<Void> save(DatasetItemBatch batch);
```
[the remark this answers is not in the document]
JetoPistola: Skipping — it would break the file's existing convention rather than restore it.

I checked before deciding: `DatasetItemService.java` has **38** uses of `@NonNull`, and **zero** of them are in the interface block. Every one is on the implementation. `applyDeltaChanges`, `getExperimentItemsStats` and the rest all declare bare parameters on the interface and annotate in `DatasetItemServiceImpl`.

So adding `@NonNull` to just `save` would make it the only annotated interface method in the file — inconsistent in the opposite direction. If the team wants interface-level null contracts here, that's a worthwhile sweep across all of them in its own change, not a one-method exception smuggled into a counter refactor.

The `@NonNull` on the impl still enforces it at runtime for every caller, including interface-typed ones.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetVersionDAO.java
```diff
@@ -589,22 +627,56 @@ AND NOT EXISTS (
...
     int ensureVersion1Exists(@Bind("dataset_id") UUID datasetId,
             @Bind("version_id") UUID versionId,
-            @Bind("workspace_id") String workspaceId);
+            @Bind("workspace_id") String workspaceId,
+            @Bind("items_total_not_migrated") int itemsTotalNotMigrated);
```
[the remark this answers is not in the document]
JetoPistola: Fixed in f5b993eb72 — added `AND workspace_id = :workspace_id` to the subquery.

Verified against the schema before touching tenant-isolation logic: `dataset_versions` has `UNIQUE (workspace_id, dataset_id, version_hash)` (migration `000036`), so workspace is part of the identity everywhere else, and every other statement in this DAO scopes by it. This `NOT EXISTS` was the outlier.

Worth noting the bug predates this branch — the subquery has been unscoped since the statement was written. But it's fair to raise here: seeding the sentinel made the affected-row count load-bearing, so a `NOT EXISTS` that wrongly matches another tenant's row now silently skips v1 creation *and* the backfill, rather than just being redundant.

134/134 green.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetVersioningMigrationService.java
```diff
@@ -330,8 +330,17 @@ private Mono<Void> countAndUpdateItemsTotal(UUID datasetId, UUID versionId, Stri
...
                                     .updateItemsTotal(workspaceId, versionId, count);
+
+                            if (updated == 0) {
+                                // The backfill only writes rows still holding the sentinel. Zero rows means an
+                                // API write already incremented this version from a real baseline, so its
+                                // counter is live and this (older) count would overwrite it with a stale value.
+                                log.info(
+                                        "Skipped items_total backfill for dataset '{}' version '{}': already migrated",
+                                        datasetId, versionId);
+                            }
```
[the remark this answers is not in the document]
JetoPistola: Fixed in f5b993eb72 — took exactly the shape you suggested: the callback returns `updated`, and the `updated == 0` branch runs outside `inTransaction`.

I read `.agents/skills/opik-backend/mysql.md` to check rather than assume, and the rule is explicit — "Only database operations inside transaction", with unrelated logic in the block marked ❌ BAD. My log call was squarely in that category.

Side benefit beyond the convention: the transaction is now held only for the UPDATE itself, rather than across a log write too.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetVersioningMigrationService.java
```diff
@@ -329,11 +329,17 @@ private Mono<Void> countAndUpdateItemsTotal(UUID datasetId, UUID versionId, Stri
...
+                                handle -> handle.attach(DatasetVersionDAO.class)
+                                        .updateItemsTotal(workspaceId, versionId, count));
+
+                        if (updated == 0) {
+                            // The backfill only writes rows still holding the sentinel. Zero rows means an API
+                            // write already incremented this version from a real baseline, so its counter is
+                            // live and this (older) count would overwrite it with a stale value.
+                            log.info("Skipped items_total backfill for dataset '{}' version '{}': already migrated",
+                                    datasetId, versionId);
+                        }
```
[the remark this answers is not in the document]
JetoPistola: Two asks here — one fixed, one I attempted twice and am **not** shipping.

**Log labels — fixed in 47a56f8038.** Now `datasetId='{}' versionId='{}' reason='already migrated'`, matching the labelled form already used elsewhere in this branch.

**Lazy-migration coverage — real gap, no test landed.** You're right that nothing exercises it: zero references to `migrateSingleDataset` or `ensureVersion1Exists` anywhere in the test tree, and `DatasetVersioningMigrationService` appears only as a `mock(...)`. That gap is exactly how my sentinel regression reached review in the first place, so I did try.

Attempt 1 asserted `items_total == 4` after running `migrateSingleDataset` on a dataset seeded through the API. It failed with `0` — `copyItemsFromLegacy` reads the legacy `dataset_items` table, and items created through the versioned API never land there, so there was nothing to copy.

Attempt 2 relaxed the assertion to "left the sentinel, and is zero". It passed. Then I reintroduced the `0`-seed bug it was written to catch and **it still passed** — with a `0` seed and a `0` real count the two states are indistinguishable. It would have shipped as false assurance, so I reverted it rather than land a test I had just proven vacuous.

A meaningful test needs a **non-zero** migrated count, i.e. real rows in legacy `dataset_items`. That needs either a test app started with `datasetVersioningEnabled=false`, or raw ClickHouse seeding — neither of which exists in this file today. That's a harness, not a test method, and it's a bigger change than the counter refactor this PR is about.

Proposing it as a follow-up rather than claiming coverage I don't have. Happy to file the ticket if that works.

134/134 green.

CometActions:

### Access Information
- **URL:** https://pr-7705.dev.comet.com
- **Namespace:** pr-7705
- **Version:** 2.2.42-7705-merge-3131
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7705%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

CometActions:
🌙 **Nightly cleanup:** The `keep-test-env` label kept `pr-7705` running through this cleanup cycle. The label has now been removed, so this environment will be torn down on the next nightly run unless you re-add **`keep-test-env`**.

thiagohora (approved):

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7705`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

CometActions:
🌙 **Nightly cleanup:** The `keep-test-env` label kept `pr-7705` running through this cleanup cycle. The label has now been removed, so this environment will be torn down on the next nightly run unless you re-add **`keep-test-env`**.

CometActions:

### Access Information
- **URL:** https://pr-7705.dev.comet.com
- **Namespace:** pr-7705
- **Version:** 2.2.42-7705-merge-3144
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7705%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

PR #7710, opened 2026-08-03
[OPIK-7693] [BE] perf: materialize the feedback-scores trace_final CTE (2x bytes / 1.9x CPU on prod)

thiagohora:
## Details

Declares `trace_final` as `AS MATERIALIZED` in `SELECT_FEEDBACK_SCORES_STATS`, behind a generic per-environment toggle. That template references the CTE from three scopes — its own definition plus two `id IN (SELECT id FROM trace_final)` subqueries — and ClickHouse re-evaluates it once per reference. Materialising it evaluates it once into a temporary table.

No schema change, no migration, no storage cost, no write-path cost.

### Measured on production

Replayed back to back with identical settings, results identical in every pair:

| Branch | Toggle off | Toggle on | Gain |
|---|---|---|---|
| `FINAL` — no searchText, the common path | 56.18 GiB / 9.07 GiB / 62,747 ms | **28.12 GiB / 8.71 GiB / 33,201 ms** | **2.00x bytes, 1.89x CPU** |
| `argMax` — searchText, from OPIK-7636 | 23.98 GiB / 2.49 GiB / 21,355 ms | **12.01 GiB / 2.57 GiB / 11,014 ms** | **2.00x bytes, 1.94x CPU** |

It helps both branches about equally, so it is *not* gated on the searchText condition that OPIK-7636 needed. Peak memory is flat to slightly better on both. Stacked with the argMax change that just merged, the searchText path goes from 56.18 GiB / 62,747 ms to 12.01 GiB / 11,014 ms.

### Why it is behind a toggle rather than unconditional

Both the keyword and the `enable_materialized_cte` setting it needs require **ClickHouse 26.3+**. Verified against 25.8:

| Probe on 25.8 | Result |
|---|---|
| `WITH t AS MATERIALIZED (...)` | `Code: 62 SYNTAX_ERROR` — the keyword does not parse |
| `enable_materialized_cte` in `system.settings` | 0 rows — the setting does not exist |
| `SETTINGS enable_materialized_cte = 1` | `Code: 115 UNKNOWN_SETTING` |

Any of those turns **every** trace- and project-stats request into a 5xx rather than degrading — not just under a restrictive settings profile. The bundled ClickHouse is 26.3 in `docker-compose.yaml`, the helm chart and testcontainers, so a standard install is fine; the exposure is an environment pointed at an older external ClickHouse, and the window during a rolling upgrade where a new backend runs against a not-yet-upgraded ClickHouse.

So: **`queryOptimizations.enableCteMaterialization`, off by default**, overridable per environment via `ENABLE_CTE_MATERIALIZATION`. The flag names the ClickHouse capability rather than the one query that currently uses it, and sits in a `queryOptimizations` section for analytics-query optimisations that are safe to disable — each renders an equivalent query shape, so turning one off costs performance and nothing else. With the toggle off the template renders byte-identical to `main`. It is enabled in `config-test.yml` so CI exercises the materialized path.

**Deliberately not normalising the exception instead.** Catching the rejection and returning a controlled empty response would convert a total stats outage into *silently empty* stats — users would see zeros for trace counts, feedback scores and costs with no reason to disbelieve them. A 5xx is bad; plausible-looking wrong data is worse and much harder to diagnose. This is a capability problem, not an error-handling one.

### The setting is required, and its absence is silent

With the setting off, `AS MATERIALIZED` parses and is simply ignored, returning the unmaterialised plan with no error — measured, identical bytes to the baseline. That is why the toggle emits **both** the keyword and the setting together.

It is set per-statement in the template's `SETTINGS` rather than session- or server-wide, to keep an Experimental setting scoped to the one query that needs it. That required checking something non-obvious: a trailing `SETTINGS` clause is applied *after* parsing, so it was not clear it could enable parse-adjacent behaviour. It can — the trailing form produced the same 12.01 GiB as passing the setting pre-parse as an HTTP URL parameter.

### Deliberately not extended

Do not apply this to a CTE that projects wide columns. The temporary table honours `default_temporary_table_engine`, which is `Memory` in production, so materialising a CTE carrying `input`/`output`/`metadata` pushes that text into memory — a control CTE projecting `input` over 132K traces went from 33.30 MiB / 41.95 MiB peak to **9.16 GiB / 9.11 GiB** when materialised. This CTE projects only `id, project_id`, which is why it is safe.

`SELECT_TRACES_SPANS_STATS` is untouched. Its `spans_data` CTE is referenced four times but is already evaluated once, so materialising it is a no-op — measured byte-identical and ~9% worse on CPU.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves OPIK-7693

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: production benchmarking of both branches, the ClickHouse 25.8 compatibility probes, the trailing-SETTINGS verification, the toggle and its javadoc; drafted this description and the ticket
- Human verification: reviewed by @thiagoh

## Testing

`mvn -o test-compile`, `mvn -o surefire:test -Dtest=TraceDAOImplTest` (21 pass), `mvn -o spotless:apply` (clean).

Correctness comes from the existing integration coverage: this changes only *how many times* a CTE is evaluated, not what it returns, and feedback-scores stats behaviour is already covered by `GetTracesByProjectResourceTest`. `config-test.yml` sets the toggle on, so CI runs those tests against the materialized plan on ClickHouse 26.3. Every production replay pair above returned identical results.

Verified the off path too: with the toggle unset the template renders byte-identical to `main`, so a deployment that does not opt in cannot be affected.

**Not covered by a test, deliberately:** that the rendered query carries both the keyword and the setting when the toggle is on. Neither is observable from query results — dropping either yields correct output ~2x slower — so a black-box test cannot catch it, and a SQL-string assertion would couple the suite to how the query is written. Flagging it as a known gap rather than adding a test I would not trust.

## Documentation

No user-facing change. The rationale, the 26.3 requirement and why the toggle exists, the silent-failure mode of the setting, the per-statement scoping and the wide-column warning are documented on `SELECT_FEEDBACK_SCORES_STATS` and on `QueryOptimizationsConfig`.


apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -2896,7 +2920,7 @@ AND entity_id IN (SELECT id FROM spans_data)
                  HAVING <feedback_scores_empty_filters>
             )
             <endif>
-            , trace_final AS (
+            , trace_final AS <if(enable_cte_materialization)>MATERIALIZED <endif>(
```
andrescrz: Nit: feel free to align the ST conditional, the config field and everything in this PR to the actual config param: `enable_materialized_cte` naming (snake case or camel case as it applies for each context).

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/OpikConfiguration.java
```diff
@@ -189,6 +189,9 @@ public class OpikConfiguration extends JobConfiguration {
     @Valid @NotNull @JsonProperty
     private ExperimentAggregatesConfig experimentAggregates = new ExperimentAggregatesConfig();
 
+    @Valid @NotNull @JsonProperty
+    private QueryOptimizationsConfig queryOptimizations = new QueryOptimizationsConfig();
```
andrescrz: Minor: double check if `QueryOptimizationsConfig` is the right place, of this setting fitted a previous existing config place, so we don't create more than needed.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/QueryOptimizationsConfig.java
```diff
@@ -0,0 +1,27 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import jakarta.validation.Valid;
+import lombok.Data;
+
+/**
+ * Toggles for analytics-query optimisations that are safe to disable — each renders a different but equivalent
+ * query shape, so turning one off costs performance and nothing else.
+ */
+@Data
+public class QueryOptimizationsConfig {
```
andrescrz: Nit: I'm starting to favour records over POJOs, also for these config beans. Feel free to follow that.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/QueryOptimizationsConfig.java
```diff
@@ -0,0 +1,27 @@
...
+     * <p>Off by default because it requires ClickHouse 26.3 or newer: on 25.8 the keyword is a {@code SYNTAX_ERROR}
+     * and the {@code enable_materialized_cte} setting it needs is an {@code UNKNOWN_SETTING}, either of which fails
+     * the query outright rather than degrading. The bundled ClickHouse is 26.3, so deployments using it can enable
+     * this; an environment pointed at an older external ClickHouse must leave it off.
+     *
+     * <p>Enabling it emits both the keyword and the setting, per statement. Both are needed — with the setting
+     * absent the keyword parses and is silently ignored.
+     */
+    @Valid @JsonProperty
+    private boolean enableCteMaterialization = false;
```
andrescrz: Minor: better remove defaults from the Java code, so we keep the values centralised in the config YAML files, both prod and tests. Typically, we enforce providing them. In this case, it's the same as the Java default for booleans, so it doesn't buy us much.

apps/opik-backend/src/test/resources/config-test.yml
```diff
@@ -1216,3 +1216,7 @@ optimizationStalledReaper:
 environment:
   # Low ceiling for tests so cap-related scenarios don't need to insert many environments.
   maxPerWorkspace: 5
+
+queryOptimizations:
+  # testcontainers runs ClickHouse 26.3, so CI covers the materialized path
```
andrescrz: Minor: the convention is to keep the exact same identical documentation comments in the test file as in the prod, except for explaining the difference when values change for tests.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -3069,7 +3093,7 @@ SELECT project_id, name, avg(value) AS value
                 s.span_feedback_scores AS span_feedback_scores
             FROM trace_feedback_scores t
             FULL OUTER JOIN span_feedback_scores s ON t.project_id = s.project_id
-            SETTINGS log_comment = '<log_comment>'
+            SETTINGS log_comment = '<log_comment>'<if(enable_cte_materialization)>, enable_materialized_cte = 1<endif>
```
andrescrz: Minor: as this is experimental, great to add it under a feature flag. Maybe we can document this in our Claude and LLM reviewer rules, so it's automatically checked on code review and noone introduces it without a flag in the future. References:
- https://clickhouse.com/docs/reference/statements/select/with
- https://clickhouse.com/docs/reference/settings/session-settings/enable#enable_materialized_cte

andrescrz (dismissed):
Only minor or nit comments. Nothing really a blocker.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -3069,7 +3088,7 @@ SELECT project_id, name, avg(value) AS value
                 s.span_feedback_scores AS span_feedback_scores
             FROM trace_feedback_scores t
             FULL OUTER JOIN span_feedback_scores s ON t.project_id = s.project_id
-            SETTINGS log_comment = '<log_comment>'
+            SETTINGS log_comment = '<log_comment>', enable_materialized_cte = 1
             ;
```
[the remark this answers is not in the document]
thiagohora: Addressed by the toggle (6135cda, renamed in 662ba61) — the setting is now emitted only when `queryOptimizations.enableCteMaterialization` is on, off by default, so an incompatible deployment renders exactly the previous SQL.

Two corrections to the finding, both of which make it worse than described:

**It isn't only restrictive profiles.** The `AS MATERIALIZED` keyword is itself version-gated, so on any ClickHouse below 26.3 the query cannot parse at all. Tested against 25.8:

| Probe | Result |
|---|---|
| `WITH t AS MATERIALIZED (...)` | `Code: 62 SYNTAX_ERROR` |
| `enable_materialized_cte` in `system.settings` | 0 rows — does not exist |
| `SETTINGS enable_materialized_cte = 1` | `Code: 115 UNKNOWN_SETTING` |

So there were two independent hard-failure paths, not one, and catching the setting rejection would have left the syntax error uncaught.

**I did not take the suggested normalization, deliberately.** Catching `ClickHouseException` and returning a controlled empty result would convert a total stats outage into *silently empty* stats — operators and users would see zeros for trace counts, feedback scores and costs with no signal that anything failed. A 5xx is bad; plausible-looking wrong data is worse and much harder to diagnose. The retry-without-the-setting variant has the same problem in a subtler form: it would mask a misconfigured environment indefinitely while paying a failed query on every request.

Gating on config instead means an unsupported deployment never emits the construct in the first place, so there is no exception to normalize. For the record, the bundled ClickHouse is 26.3 in `docker-compose.yaml`, the helm chart and testcontainers, so the exposure was external-ClickHouse deployments and rolling-upgrade windows rather than default installs.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/TraceStatsConfig.java
```diff
@@ -0,0 +1,24 @@
...
+     * ClickHouse 26.3 or newer: on 25.8 the keyword is a {@code SYNTAX_ERROR} and the setting is an
+     * {@code UNKNOWN_SETTING}, either of which would turn every trace- and project-stats request into a 5xx rather
+     * than degrade. Deployments that ship the bundled ClickHouse are on 26.3 and can enable it; an environment
+     * pointed at an older external ClickHouse must leave it off.
+     *
+     * <p>When off, the query renders exactly as before. When on it is roughly 2x fewer bytes and 2x less CPU
+     * (OPIK-7693).
+     */
+    @Valid @JsonProperty
+    private boolean materializeFeedbackScoresCte = false;
```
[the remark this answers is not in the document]
thiagohora: Agreed the name was misleading, and it's fixed in 662ba61 — though by going the opposite direction to the suggestion.

Rather than making the name more specific (`materializeTraceFinalCte`), the flag is now `queryOptimizations.enableCteMaterialization`: it names the ClickHouse capability and no CTE at all, so it cannot misname one. Two reasons that reads better here:

- The thing being gated really is the capability, not this CTE. If a second query adopts `AS MATERIALIZED` it reuses the same flag instead of adding a second toggle, and an operator setting it is deciding "may this deployment's ClickHouse do materialized CTEs", which is exactly the question their ClickHouse version answers.
- `materializeTraceFinalCte` would have traded one coupling for another — `trace_final` exists in both stats templates, and only one of them is materialized, so the specific name would have implied broader scope than it has.

`TraceStatsConfig` is deleted; the flag lives on `QueryOptimizationsConfig`, documented as a section for analytics-query optimisations that are safe to disable, since each renders an equivalent query shape. Env var is `ENABLE_CTE_MATERIALIZATION`, and the javadoc now describes the capability and its 26.3 requirement rather than a CTE name.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
thiagohora: Done in 190e6f0 — everything now takes its name from the ClickHouse setting: template attribute `enable_materialized_cte` (snake, matching the setting exactly), Java and YAML `materializedCteEnabled`, env var `ANALYTICS_DB_DATA_MODEL_MATERIALIZED_CTE_ENABLED`.

The Java/YAML form keeps the `<thing>Enabled` suffix rather than being a literal `enableMaterializedCte`, because it now sits on `DatabaseAnalyticsDataModelConfig` where the neighbours are `traceColumnsNonNullable`, `traceDeletionEventsCaptureEnabled` and so on. Happy to make it literal if you'd rather the alignment be exact.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/OpikConfiguration.java
thiagohora: Good catch — the new config is gone. The flag now lives on `DatabaseAnalyticsDataModelConfig`, which is already the home for analytics-DB state and capability toggles (`traceColumnsNonNullable`, `spanColumnsNonNullable`, `traceDeletionEventsCaptureEnabled`, `spanDeletionEventsCaptureEnabled`) — same category of question, "what is this deployment's ClickHouse able to do", and `TraceDAO` already reads that config. I widened its class javadoc from "schema state" to also cover version-dependent query constructs.

I tried `ServiceTogglesConfig` first and backed it out: `ServiceTogglesResource` returns that whole object as a public API response, so a ClickHouse tuning flag there would leak an internal detail to API consumers and show up in the toggles payload.

Placing it there also resolves your other two points for free — the record and the no-Java-defaults ones.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/QueryOptimizationsConfig.java
thiagohora: Agreed, and it comes for free now: the flag moved onto `DatabaseAnalyticsDataModelConfig`, which is already a `record`, so the standalone POJO is deleted rather than converted.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/QueryOptimizationsConfig.java
thiagohora: Agreed — no Java-side default any more. The flag is a `record` component on `DatabaseAnalyticsDataModelConfig`, so the value comes only from the YAML, and both `config.yml` and `config-test.yml` set it explicitly.

You're right that it bought nothing here anyway, since `false` is the Java default for `boolean`; the real value is that the configured value is now visible in one place per environment rather than split between the YAML and the class.

apps/opik-backend/src/test/resources/config-test.yml
thiagohora: Fixed in 190e6f0. `config-test.yml` now carries the same documentation block as `config.yml` verbatim, with one extra line explaining the difference:

```
#   Enabled for tests: testcontainers runs ClickHouse 26.3, so CI covers the materialized plan.
  materializedCteEnabled: true
```

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
thiagohora: It is under a flag as of 6135cda, and now documented as a rule in 190e6f0: `.claude/rules/apps/opik-backend/experimental_clickhouse_features.md`, alongside the existing `adding_filter_field_types.md`.

The rule requires that any experimental or version-gated ClickHouse construct is behind a config flag defaulting to off, and captures the parts that are easy to get wrong:

- the failure is a hard error, not a degradation — it includes the measured 25.8 evidence (`SYNTAX_ERROR` for the keyword, `UNKNOWN_SETTING` for the setting) so a reviewer can see why "it'll just be slower" is untrue
- emit the keyword and its setting together, since several constructs parse but are silently ignored without their setting, which yields the unoptimised plan with no error
- off must render byte-identical SQL, verifiable by diffing against the merge base
- don't "normalize" the failure by swallowing the exception — that turns an outage into silently wrong data
- enable it in `config-test.yml` so CI covers the new path

Thanks for the doc links — both are cited in the rule. Worth noting the reviewer wiring is only implicit: `.cursor/agents/code-reviewer.md` doesn't reference `.claude/rules/`, so the rule is picked up by convention rather than by an explicit include. Might be worth linking them if you want it enforced automatically.

apps/opik-backend/config.yml
```diff
@@ -145,6 +145,13 @@ databaseAnalyticsDataModel:
   #   than the ClickHouse driver binds reliably in one statement (5 columns per row), so the insert is split into
   #   chunks of this size.
   deletionEventsInsertBatchSize: ${ANALYTICS_DB_DATA_MODEL_DELETION_EVENTS_INSERT_BATCH_SIZE:-1000}
+  #  Default: false
+  #  Description: Allow queries to declare a CTE AS MATERIALIZED so it is evaluated once instead of once per
+  #   reference (~2x fewer bytes and ~2x less CPU on the feedback-scores stats query). Emits the
+  #   enable_materialized_cte setting alongside the keyword; both are required. Requires ClickHouse 26.3+, which the
+  #   bundled ClickHouse satisfies; leave false when pointing at an older external ClickHouse, where it fails the
+  #   query rather than degrading.
+  materializedCteEnabled: ${ANALYTICS_DB_DATA_MODEL_MATERIALIZED_CTE_ENABLED:-false}
```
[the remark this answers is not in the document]
thiagohora: There no need to change any helm chart config to allow it, today is already possible to set envs

andrescrz, line 154: Agree no dedicated test for false needed. But please make sure you run a couple of manual tests: locally and stg before deploying, to ensure there's no small syntax error in the query triggering some unhandled exception.

andrescrz (approved):
LGTM

apps/opik-backend/config.yml
thiagohora: Tested locally and on staging — no syntax error either way.

- **Local**: with the flag off the statement renders byte-identical to the merge base; `GetTracesByProjectResourceTest` (1,763 tests) and the `ProjectsResourceTest` stats classes are green.
- **Staging**: ClickHouse is `26.3.16.10001.altinitystable` (same build as prod) and `enable_materialized_cte` exists. Ran the rendered statement against a real project with and without `MATERIALIZED` — both clean, identical results.

With the flag **on** I did hit `Code: 49 LOGICAL_ERROR — Reading from materialized CTE 'trace_final' before it has been materialized` → 500. This is a **known upstream bug**, not a syntax issue and not our query: [ClickHouse#101940](https://github.com/ClickHouse/ClickHouse/issues/101940) — materialized CTE reused via `IN` subqueries, which is exactly our three `id IN (SELECT id FROM trace_final)` references. Fixed on `master` by [#105041](https://github.com/ClickHouse/ClickHouse/pull/105041) (`pr-critical-bugfix`), and upstream's version matrix is:

```
Reproduces on:         26.4, 26.3     <- we run 26.3.16
Does NOT reproduce on: master, 26.5
Backport needed:       26.4, 26.3
```

The 26.3 backport is still outstanding (cherry-pick #106637 never merged). Frequency here: 18 ClickHouse errors → 6 failed test invocations, in 4 of 10 local runs; 0 in the other 6, including four full-class runs. CI logged 12 errors absorbed as 4 flakes by the surefire reruns. It is interleaving-dependent and independent of argMax vs `FINAL`.

So the flag stays off by default, and it shouldn't be enabled anywhere until ClickHouse is on 26.5+ or the 26.3 backport lands.

thiagohora: FYI @andrescrz,

I'm closing this PR since we would need to wait for the backport of the fix.

Issue #7697, opened 2026-08-02
[Bug]: Additional LLM provider prices dropped at load time (missing from PROVIDERS_MAPPING)


Anuj7411:
## What component(s) are affected?

- [x] Opik Server

## Describe the problem

Following the provider-registration fixes already merged for azure (#7262), xai (#7433), deepseek (#7434), perplexity (#7432), fireworks_ai (#7554) and moonshot (#7604), several more high-volume providers are still dropped at load time. `CostService.buildModelPrice` returns `null` for any entry whose `litellm_provider` is not in `PROVIDERS_MAPPING`, so every model tagged with an unregistered provider gets no cost.

Scanning the bundled `model_prices_and_context_window.json` for `litellm_provider` values that carry non-zero-cost entries but are absent from `PROVIDERS_MAPPING`, the remaining gaps are (approx. non-zero-priced model counts):

- openrouter (~92, with ~32 carrying cache_read rates)
- novita (~85)
- deepinfra (~67)
- together_ai (~36)
- nebius (~30)
- databricks (~28)
- snowflake (~21, ~9 with cache_read)
- sambanova (~17)
- cerebras (~7)

The OpenAI-compatible inference providers (novita, deepinfra, nebius, snowflake, sambanova, cerebras, databricks) are the same shape as the ones already registered: map `<provider> -> <provider>` and route through `textGenerationWithCacheCostOpenAI` when the provider publishes cache rates, else `textGenerationCost`.

Aggregators (openrouter, together_ai) need more care. openrouter namespaces models under the origin vendor (e.g. `moonshotai/*`, `anthropic/*`) so it should reuse the existing provider-prefix fallback in `findModelPrice` rather than a naive `openrouter -> openrouter` mapping; together_ai currently only has parameter-size bucket entries (`together-ai-4.1b-8b`, ...) in the JSON, so it needs bucket-lookup logic, not just a mapping entry.

## Additional info

Happy to open PRs for the clean single-vendor providers (they mirror the merged fireworks/moonshot PRs one-to-one) and to discuss the aggregator handling separately. Filing as a tracking issue first so the team can steer scope.

PR #7715, opened 2026-08-03
[NA] [BE] fix: register deepinfra as a canonical provider so DeepInfra model prices load

Anuj7411:
## Details

DeepInfra model prices are silently dropped by Opik today. `model_prices_and_context_window.json` carries ~67 non-zero-cost entries tagged `litellm_provider: "deepinfra"` (the `deepinfra/<org>/<model>` catalog, 5 of them with `cache_read_input_token_cost`), and none load because `deepinfra` is not in `PROVIDERS_MAPPING`; `buildModelPrice` returns `null` for each. One of the gaps listed in #7697, same pattern as the merged azure / xai / deepseek / perplexity / fireworks_ai / moonshot registrations.

- Register `deepinfra -> deepinfra` in `PROVIDERS_MAPPING` so the entries load
- Route `deepinfra` through `textGenerationWithCacheCostOpenAI`; DeepInfra exposes an OpenAI-compatible API, so even the Claude models it hosts report cached tokens under `prompt_tokens_details.cached_tokens` (OpenAI shape), not the Anthropic shape

## Change checklist

- [x] User facing
- [ ] Documentation update

## Issues

Part of #7697

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: the two mapping entries and the parameterized test
- Human verification: model resolution and both cost figures checked against the bundled model_prices_and_context_window.json before submission

## Testing

Added `calculateCostHandlesDeepinfraModels` covering both branches:

- `deepinfra/Gryphe/MythoMax-L2-13b` (input 8e-8, output 9e-8, no cache) routes through `textGenerationCost`; 1000 + 200 tokens = 0.000098
- `deepinfra/anthropic/claude-3-7-sonnet-latest` (input 3.3e-6, output 1.65e-5, cache_read 3.3e-7) routes through `textGenerationWithCacheCostOpenAI`; with 300 of 1000 prompt tokens cached, 700 * 3.3e-6 + 200 * 1.65e-5 + 300 * 3.3e-7 = 0.005709

Both cases fail against the current code (models do not resolve without the mapping entry, so cost is zero) and pass with the fix.

## Documentation

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
@@ -52,7 +52,8 @@ public class CostService {
             Map.entry("inception", "inception"),
             Map.entry("meta", "meta"),
             Map.entry("zai", "zai"),
-            Map.entry("z-ai", "zai"));
+            Map.entry("z-ai", "zai"),
+            Map.entry("deepinfra", "deepinfra"));
 
     // Online evaluation (and OTel ingestion) resolve models to LlmProvider serialized values whose names
     // differ from the canonical price-table vocabulary. Normalize those to the single canonical provider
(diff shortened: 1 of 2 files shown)


andrescrz:
Hi @Anuj7411 

Thanks for the contribution!

Thanks!

Anuj7411:
Thanks, good catch on the DeepInfra Anthropic surface. This PR routes DeepInfra through the OpenAI cache calculator, which reads original_usage.prompt_tokens_details.cached_tokens; that is the shape the Opik SDK integration normalizes DeepInfra usage into, and what the test covers. Anthropic-Messages-shape responses (input_tokens / cache_read_input_tokens) are read by textGenerationWithCacheCostAnthropic instead, and a single provider entry can only route to one calculator, so any OpenAI-routed host that also serves Claude has this same limitation today (not introduced here). The correct general fix is per-span usage-shape detection that picks the calculator by the keys present; I'd suggest a separate issue for that so it covers every multi-surface host at once rather than special-casing DeepInfra

andrescrz:
Hi @Anuj7411 

Thanks!

Anuj7411:
Rebased onto latest main and resolved the conflicts (the sibling snowflake entry landed in `PROVIDERS_MAPPING` and `PROVIDERS_CACHE_COST_CALCULATOR`; both providers now coexist and the deepinfra test is intact). Ready for another look, thanks!

andrescrz (approved):
Hi @Anuj7411 

Best

PR #7716, opened 2026-08-03
[NA] [BE] fix: register nebius as a canonical provider so Nebius model prices load

Anuj7411:
## Details

Nebius AI Studio model prices are silently dropped by Opik today. `model_prices_and_context_window.json` carries 30 non-zero-cost entries tagged `litellm_provider: "nebius"` (the `nebius/<org>/<model>` catalog: deepseek, qwen, llama and more), and none load because `nebius` is not in `PROVIDERS_MAPPING`; `buildModelPrice` returns `null` for each. One of the gaps listed in #7697.

- Register `nebius -> nebius` in `PROVIDERS_MAPPING` so the entries load
- No cache-calculator wiring is needed since no Nebius entry publishes `cache_read_input_token_cost` today, so all Nebius requests route through `textGenerationCost`

## Change checklist

- [x] User facing
- [ ] Documentation update

## Issues

Part of #7697

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: the mapping entry and the parameterized test
- Human verification: model resolution and the cost figure checked against the bundled model_prices_and_context_window.json before submission

## Testing

Added `calculateCostHandlesNebiusModels`: `nebius/deepseek-ai/DeepSeek-R1` (input 8e-7, output 2.4e-6) with 1000 prompt + 200 completion tokens = 0.00128. Fails against the current code (the model does not resolve without the mapping entry, so cost is zero) and passes with the fix.

## Documentation


andrescrz:
Hi @Anuj7411 

Thanks for the contribution!

Thanks!

Anuj7411:
Thanks, addressed both Baz comments.

Single-case test: inlined as a plain @Test with the inputs directly in the body (latest commit).

Factory-routed lookup: this PR is scoped to cost tracking of ingested spans, which resolves purely through PROVIDERS_MAPPING; that is the same scope as the merged azure / xai / deepseek / fireworks_ai / moonshot registrations, none of which are in LlmProvider / LlmProviderFactory yet all price correctly. The getResolvedModelInfo path Baz points at is the online-evaluation surface (selecting nebius as a provider to invoke), which is a separate feature I have deliberately kept out of scope here

andrescrz:
Hi @Anuj7411 

Thanks!

Anuj7411:
Rebased onto latest main; the conflict was the sibling sambanova entry landing in `PROVIDERS_MAPPING` and its test. Both providers now coexist and the nebius test is a plain `@Test`. Ready for another look.

andrescrz (approved):
Hi @Anuj7411 

PR #7717, opened 2026-08-03
[NA] [BE] fix: register sambanova as a canonical provider so SambaNova model prices load

Anuj7411:
## Details

SambaNova Cloud model prices are silently dropped by Opik today. `model_prices_and_context_window.json` carries 19 non-zero-cost entries tagged `litellm_provider: "sambanova"` (deepseek, llama, qwen and minimax models), and none load because `sambanova` is not in `PROVIDERS_MAPPING`; `buildModelPrice` returns `null` for each. One of the gaps listed in #7697.

- Register `sambanova -> sambanova` in `PROVIDERS_MAPPING` so the entries load
- No cache-calculator wiring is needed since no SambaNova entry publishes `cache_read_input_token_cost` today, so all SambaNova requests route through `textGenerationCost`

## Change checklist

- [x] User facing
- [ ] Documentation update

## Issues

Part of #7697

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: the mapping entry and the parameterized test
- Human verification: model resolution and the cost figure checked against the bundled model_prices_and_context_window.json before submission

## Testing

Added `calculateCostHandlesSambanovaModels`: `sambanova/MiniMax-M2.7` (input 6e-7, output 2.4e-6) with 1000 prompt + 200 completion tokens = 0.00108. Fails against the current code (the model does not resolve without the mapping entry, so cost is zero) and passes with the fix.

## Documentation


andrescrz:
Hi @Anuj7411 

Thanks for the contribution!

Thanks!

Anuj7411:
Thanks. Addressed the single-case parameterization: inlined it as a plain @Test with the model, usage and expected cost directly in the body (d383da1 -> latest). No other Baz comments outstanding on this one

andrescrz (approved):
Hi @Anuj7411 

Best,

PR #7719, opened 2026-08-03
[OPIK-7676] [BE] perf: drop spans FINAL from SELECT_BY_IDS (find_traces_by_ids)

YarivHashaiComet:
## Details

`TraceDAO.SELECT_BY_IDS` issued two `spans FINAL` reads even though its predicate is fully pinned to the primary key, making `find_traces_by_ids` the heaviest ClickHouse query shape we run. Both are replaced with explicit dedup applied after the key filter — already the dominant pattern here (221 `LIMIT n BY` sites vs 76 remaining `FINAL`), including for `spans` in the same aggregating context in this file.

- `target_spans` — dedup inline; nothing aggregates downstream.
- `spans_agg` — new `spans_deduped` CTE so dedup runs before the aggregation, mirroring `SELECT_TRACES_SPANS_STATS`.
- `ORDER BY` lists all five PK columns **including `parent_span_id`**. Matching the sort key is what lets ClickHouse read in order and prune parts — omitting it costs most of the win, so the snippet in OPIK_7638 should be corrected before the remaining sites are done.

### Why dedup rather than dropping FINAL

`spans` is a ReplacingMergeTree and a span update re-inserts a full row under the same id, so a bare removal double-counts. On a trace with two stored versions:

| variant | span_count | llm_span_count |
|---|---|---|
| `spans FINAL` (baseline) | 2 | 1 |
| dedup (this PR) | 2 | 1 |
| `FINAL` removed, no dedup | 4 | 2 |

The exposure is confined to the span aggregates — counts, `usage`, `total_estimated_cost`, `providers`. `target_spans` is also `INNER JOIN`ed to feedback scores twice, but that path is separately protected: the union of both branches is deduped by `LIMIT 1 BY workspace_id, project_id, span_id, name, author, source_queue_id`, which collapses any duplicate rows a doubled join would produce, and `spanFeedbackScores` is averaged by name on top of that.

### Not a parity change — read this first

The dedup key is `(workspace_id, project_id, id)`, which is **narrower than the table's sort key**. So this does not reproduce `FINAL` in every case, and that is deliberate.

A span id is meant to be unique within a project. The batch endpoint inserts rows verbatim, so re-sending a span under a different parent stores a second row on a different sort key; `FINAL` treats those as two versions and counts one logical span twice. This query collapses them to one instead. The underlying data structure is being addressed separately under the hyperscale work — until then the read path should not preserve and double-count inconsistent rows.

One consequence worth knowing: the `ORDER BY` still leads with the sort key so ClickHouse can read in order, which means that among rows disagreeing on the rest of the key the survivor is the one with the greater `parent_span_id`, not the most recently written — `last_updated_at` only breaks ties once the whole tuple matches. Only observable on rows that are already inconsistent.

### Measured

Full assembled query, 6 alternating runs each, all 40 output columns identical on well-formed data:

| metric | before | after | delta |
|---|---|---|---|
| CPU | 0.343 s | **0.222 s** | **-35%** |
| bytes read | 107.59 MiB | **25.03 MiB** | **-77%** |
| parts read | 655 | **354** | **-46%** |
| file opens | 2,870 | **783** | **-73%** |
| peak memory | 168.10 MiB | **117.58 MiB** | **-30%** |
| rows read | 524,931 | 682,886 | **+30%** |

Rows read regresses: rows sharing `(workspace_id, project_id, id)` are spread across `parent_span_id` in the sort order, so ClickHouse cannot stop early per group and has to buffer the candidate range. Bytes read still drop 77% because far fewer columns are touched, which is what dominates here.

For reference, deduping on the full sort key measured -63% CPU / -94% bytes / -62% rows. Narrowing the key trades roughly half the CPU win for the semantics above.

## Testing

- Black-box coverage in `DatasetsResourceCreateFromTracesTest`, driven entirely over HTTP: `POST /datasets/{id}/items/from-traces` is the only public route to `SELECT_BY_IDS`, and `TraceEnrichmentOptions.includeUsage` surfaces `spans_agg`'s `sumMap(usage)`. No DAO or direct database access.
- Two cases: a span updated in place (same sort key) must aggregate the latest version rather than the sum, and a span re-sent under a different parent must still be aggregated once. The second fails on the previous full-key form, so it pins the behaviour this PR chose.
- Both parent ids in that case are pinned by sort order, so the survivor is deterministic instead of depending on which random UUID sorts higher.
- Equivalence was additionally checked on a large representative dataset via a row-level diff of old vs new output, with a no-dedup control to confirm the comparison actually detects divergence. Both `has_target_projects` branches covered.
- `mvn spotless:apply` clean, targeted `mvn test` green.

## Documentation

No documentation changes. No API or schema change. There is a deliberate behaviour change, described under "Not a parity change" above: a span id that appears under more than one parent within a project is now aggregated once rather than twice. It affects only rows that are already inconsistent, and the underlying data structure is being addressed under the hyperscale work.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-7676

Related, not resolved here: OPIK_7638 (parent) and OPIK_7636 (overlapping `TraceDAO` dedup via `argMax` — worth sequencing).

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: query analysis, the SQL rewrite, the integration test, and before/after verification.


apps/opik-backend/src/test/java/com/comet/opik/domain/TraceDAOFindByIdsDedupIntegrationTest.java
```diff
@@ -0,0 +1,239 @@
...
+        Span span = factory.manufacturePojo(Span.class).toBuilder()
+                .projectName(projectName)
+                .traceId(traceId)
+                .parentSpanId(null)
+                .type(type)
+                .provider(PROVIDER)
+                .usage(usage)
+                .totalEstimatedCost(null)
+                .feedbackScores(null)
+                .comments(null)
```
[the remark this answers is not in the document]
YarivHashaiComet: Addressed in 266ed9b. The LLM span is now created with `totalEstimatedCost` = 1.00 and the update sets 7.00, and the test asserts `actual.totalEstimatedCost()` with `isEqualByComparingTo(7.00)`.

That covers the non-zero `BigDecimal` decode you pointed at, and it also strengthens the dedup case: cost is a summed mutable column, so aggregating both stored versions would yield 8.00. Good catch — the aggregate was previously only exercised at zero.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -483,10 +483,12 @@ ORDER BY (workspace_id, project_id, id) DESC, last_updated_at DESC
...
             WITH target_spans AS (
                 SELECT id, trace_id, type
-                FROM spans FINAL
+                FROM spans
                 WHERE workspace_id = :workspace_id
                 <if(has_target_projects)>AND project_id IN :target_project_ids<endif>
                 AND trace_id IN :ids
+                ORDER BY (workspace_id, project_id, trace_id, parent_span_id, id) DESC, last_updated_at DESC
+                LIMIT 1 BY workspace_id, project_id, trace_id, parent_span_id, id
             ),
```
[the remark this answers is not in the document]
YarivHashaiComet: Checking this one against the source — the `ORDER BY` is present, on the line immediately above the `LIMIT 1 BY`, at both sites:

```
ORDER BY (workspace_id, project_id, trace_id, parent_span_id, id) DESC, last_updated_at DESC
LIMIT 1 BY workspace_id, project_id, trace_id, parent_span_id, id
```

ClickHouse applies `LIMIT n BY` after `ORDER BY`, so the version choice is deterministic: since the `LIMIT BY` key equals the `ORDER BY`'s leading tuple, ordering within each group is decided purely by `last_updated_at DESC` — newest version wins, matching `FINAL`. I surveyed the backend to be sure this is the house shape and **218 of 221** dedup sites pair the two exactly this way; the 3 that omit `ORDER BY` each carry a comment justifying an arbitrary pick. Not changing this.

On the second half — dedup logic spread across several SQL strings and liable to drift — that is a fair observation, but it is pre-existing across all 221 sites and a refactor rather than part of a perf fix. Two notes if we do pick it up: ClickHouse inlines singly-referenced CTEs, so a shared deduped-spans CTE would not save any work, and `target_spans` and `spans_deduped` project different column sets, so sharing one would widen whichever consumer needs less.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -674,6 +676,20 @@ feedback_scores_grouped AS (
...
+                    type,
+                    usage,
+                    total_estimated_cost,
+                    provider
+                FROM spans
+                WHERE workspace_id = :workspace_id
+                <if(has_target_projects)>AND project_id IN :target_project_ids<endif>
+                AND trace_id IN :ids
+                ORDER BY (workspace_id, project_id, trace_id, parent_span_id, id) DESC, last_updated_at DESC
+                LIMIT 1 BY workspace_id, project_id, trace_id, parent_span_id, id
```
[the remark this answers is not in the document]
YarivHashaiComet: Real ambiguity, but not one this change introduces, so leaving it.

`ReplacingMergeTree(last_updated_at)` is **equally** nondeterministic under `FINAL` when two versions share `last_updated_at` — it picks an arbitrary row among equal versions. So neither path is authoritative and they can disagree either way; this is a property of the data model rather than of the dedup rewrite, and it holds identically for all 221 existing `LIMIT n BY` sites.

More importantly, adding a tie-break here would make this query *diverge* from `FINAL`, which is the opposite of what this PR is for — the whole point of moving to the full sort key was to match `FINAL` by construction. Happy to file a separate ticket if we want deterministic tie-breaking as a deliberate, repo-wide behaviour change.

apps/opik-backend/src/test/java/com/comet/opik/domain/TraceDAOFindByIdsDedupIntegrationTest.java
```diff
@@ -0,0 +1,251 @@
...
+                .projectName(projectName)
+                .usage(null)
+                .feedbackScores(null)
+                .build();
+        UUID traceId = traceResourceClient.createTrace(trace, API_KEY, TEST_WORKSPACE);
+
+        UUID llmSpanId = createSpan(projectName, traceId, SpanType.llm,
+                Map.of(USAGE_KEY, USAGE_BEFORE_UPDATE), COST_BEFORE_UPDATE);
+        createSpan(projectName, traceId, SpanType.tool, null, null);
+        createSpan(projectName, traceId, SpanType.general, null, null);
```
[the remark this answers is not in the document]
YarivHashaiComet: The coverage gap is real — the fixture sets `feedbackScores = null`, so the span-feedback join is genuinely never exercised. But investigating it showed the stated risk does not hold, and it also caught an error in my own PR description, so thank you for it.

The union of both join branches is deduplicated immediately afterwards:

```sql
ORDER BY last_updated_at DESC
LIMIT 1 BY workspace_id, project_id, span_id, name, author, source_queue_id
```

Two join rows produced by a duplicated `target_spans` row are identical on every one of those key columns, so they collapse to one, and `Trace.spanFeedbackScores` is averaged by name on top of that. A duplicate `span_feedback_scores_list` cannot reach the output under any of the variants — `FINAL`, `LIMIT 1 BY id`, or the full-key form here.

That also means my earlier claim in this PR, that a bare `FINAL` removal would double feedback scores, was wrong. Corrected: the exposure is confined to the additive aggregates in `spans_agg` — `count(id)`, `countIf(type = 'llm')`, `sumMap(usage)`, `sum(total_estimated_cost)` — while `providers` and `has_tool_spans` are duplicate-insensitive. The test asserts exactly those four.

So I am not adding the feedback-score case: it would add coverage without guarding a regression. Logged as a known gap instead. The no-target-project branch I verified directly against the database rather than through the test, since that branch depends on `getTargetProjectIdsForTraces` returning empty and is not reachable through the resource client.

apps/opik-backend/src/test/java/com/comet/opik/domain/TraceDAOFindByIdsDedupIntegrationTest.java
```diff
@@ -0,0 +1,251 @@
...
+ * <p>
+ * That second version only exists until a background merge collapses it, which would silently rob
+ * this test of its discriminating power — every assertion would still hold for a broken query. The
+ * raw (non-deduplicated) row count is therefore asserted before the aggregates, so losing the
+ * precondition fails loudly instead of passing quietly.
+ */
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@DisplayName("TraceDAO findByIds span dedup Integration Test")
+@ExtendWith(DropwizardAppExtensionProvider.class)
+class TraceDAOFindByIdsDedupIntegrationTest {
```
andrescrz: Any particular reason to have a dedicated test class for these tests? It seems they fit the regular traces resource test. There's no particular configuration or containers set up. You could keep them as an inner class there.

apps/opik-backend/src/test/java/com/comet/opik/domain/TraceDAOFindByIdsDedupIntegrationTest.java
```diff
@@ -0,0 +1,251 @@
...
+    private static final String USER = UUID.randomUUID().toString();
+    private static final String WORKSPACE_ID = UUID.randomUUID().toString();
+    private static final String TEST_WORKSPACE = UUID.randomUUID().toString();
+
+    private static final String PROVIDER = "openai";
+    private static final String USAGE_KEY = "completion_tokens";
+    private static final int USAGE_BEFORE_UPDATE = 10;
+    private static final int USAGE_AFTER_UPDATE = 70;
+    private static final BigDecimal COST_BEFORE_UPDATE = new BigDecimal("1.00");
+    private static final BigDecimal COST_AFTER_UPDATE = new BigDecimal("7.00");
```
andrescrz: I recommend working with randomness for these. Typical approach:
- Podam.
- RandomStringUtils or similar.
- UUID.

apps/opik-backend/src/test/java/com/comet/opik/domain/TraceDAOFindByIdsDedupIntegrationTest.java
```diff
@@ -0,0 +1,251 @@
...
+                        .runtimeInfo(wireMock.runtimeInfo())
+                        .redisUrl(REDIS.getRedisURI())
+                        .build());
+    }
+
+    private final PodamFactory factory = PodamFactoryUtils.newPodamFactory();
+
+    private TraceResourceClient traceResourceClient;
+    private SpanResourceClient spanResourceClient;
+    private TraceService traceService;
```
andrescrz: This is my main concern about this test. It's not a black box test, you need to access the internals for the service. I'm wondering if there's the possibility to simulate and test this condition from from the public API endpoints. If so, we should refactor this test to follow that principle. 

Otherwise, we could consider keeping this.

apps/opik-backend/src/test/java/com/comet/opik/domain/TraceDAOFindByIdsDedupIntegrationTest.java
```diff
@@ -0,0 +1,251 @@
...
+        ClientSupportUtils.config(client);
+
+        AuthTestUtils.mockTargetWorkspace(wireMock.server(), API_KEY, TEST_WORKSPACE, WORKSPACE_ID, USER);
+
+        traceResourceClient = new TraceResourceClient(client, baseUrl);
+        spanResourceClient = new SpanResourceClient(client, baseUrl);
+
+        traceService = injector.getInstance(TraceService.class);
+
+        clickHouseConnectionFactory = ClickHouseContainerUtils
```
andrescrz: Same about the direct DB access, this is even more concerning than the service

We should aim to test on the public interfaces and avoid internal implementation details

Internal DAO access should be done in very few specific cases, when the test provides high value and there's no other alternative

apps/opik-backend/src/test/java/com/comet/opik/domain/TraceDAOFindByIdsDedupIntegrationTest.java
```diff
@@ -0,0 +1,251 @@
...
+
+    @AfterAll
+    void tearDownAll() {
+        wireMock.server().stop();
+    }
+
+    @Test
+    @DisplayName("findByIds: an updated span must not be counted twice in the trace aggregates")
+    void findByIds__whenSpanWasUpdated__thenAggregatesUseLatestVersionOnly() {
+        String projectName = "dedup-" + UUID.randomUUID();
```
andrescrz: For textual randomness better use RandomStringUtils. I also recommend following `project-` prefix as the convention of the service. Please follow existing conventions.

apps/opik-backend/src/test/java/com/comet/opik/domain/TraceDAOFindByIdsDedupIntegrationTest.java
```diff
@@ -0,0 +1,251 @@
...
+                .totalEstimatedCost(COST_AFTER_UPDATE)
+                .build(), API_KEY, TEST_WORKSPACE);
+
+        // Guard the premise. Without two stored versions there is nothing to deduplicate and the
+        // assertions below would hold even for a broken query, so this fails rather than skips —
+        // a silent loss of coverage is worse than a visible failure. Checked immediately after the
+        // update, before a background merge can collapse the versions.
+        assertThat(rawVersionCount(llmSpanId))
+                .as("stored versions of the updated span, pre-merge")
+                .isEqualTo(2);
```
andrescrz: This test is not reliable as it depends on the deduping mechanics of Clickhouse, which aren't predictable, so there will be flakiness.

In addition, it's coupled to internal details.

apps/opik-backend/src/test/java/com/comet/opik/domain/TraceDAOFindByIdsDedupIntegrationTest.java
```diff
@@ -0,0 +1,251 @@
...
+
+        Trace actual = findByIds(traceId);
+
+        assertThat(actual.spanCount()).isEqualTo(3);
+        assertThat(actual.llmSpanCount()).isEqualTo(1);
+        assertThat(actual.hasToolSpans()).isTrue();
+        assertThat(actual.providers()).containsExactly(PROVIDER);
+        // Latest version only. Summing both versions would give 80 and 8.00 respectively.
+        assertThat(actual.usage()).containsEntry(USAGE_KEY, (long) USAGE_AFTER_UPDATE);
+        assertThat(actual.totalEstimatedCost()).isEqualByComparingTo(COST_AFTER_UPDATE);
```
andrescrz: Assertions should be done on the complete object: actual vs expected, instead of field by field. In addition, you should reuse existing aux helper classes and methods for this, they're widely used in the code base.

apps/opik-backend/src/test/java/com/comet/opik/domain/TraceDAOFindByIdsDedupIntegrationTest.java
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    private UUID createSpan(String projectName, UUID traceId, SpanType type, Map<String, Integer> usage,
```
andrescrz: Same, likely a pre-existing method already doing this logic in the spanResourceClient.

apps/opik-backend/src/test/java/com/comet/opik/domain/TraceDAOFindByIdsDedupIntegrationTest.java
```diff
@@ -0,0 +1,251 @@
...
+                .comments(null)
+                .build();
+
+        return spanResourceClient.createSpan(span, API_KEY, TEST_WORKSPACE);
+    }
+
+    /**
+     * Raw row count for a span id, deliberately without FINAL or dedup, so it sees every stored version.
+     */
+    private static final String RAW_VERSION_COUNT = """
```
andrescrz: Mentioned already, for all these methods below: we should avoid direct DB access.

apps/opik-backend/src/test/java/com/comet/opik/domain/TraceDAOFindByIdsDedupIntegrationTest.java
```diff
@@ -0,0 +1,251 @@
...
+    private <T> T queryOne(String sql, Function<Statement, Statement> binder, Function<Row, T> mapper) {
+        return Mono.usingWhen(
+                clickHouseConnectionFactory.create(),
+                connection -> Mono.from(binder.apply(connection.createStatement(sql)).execute())
+                        .flatMap(result -> Mono.from(result.map((row, ignored) -> mapper.apply(row)))),
+                Connection::close)
+                .block();
+    }
+
+    private Trace findByIds(UUID traceId) {
```
andrescrz: This should be done through the public API endpoint.

andrescrz (commented):
The change LGTM. 

My main concern is the testing approach. It's coupled to the internals and it will be flaky.

These kind of performance optimisations aren't many times testable from the automated tests, that are more oriented for functional than non-functional requirements.

It'd be fine to drop this test or just refactor to make sure the functionality doesn't break.

Non functional requirements are tested by other means, and you already provided a lot of very good results in the PR description.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -483,10 +483,12 @@ ORDER BY (workspace_id, project_id, id) DESC, last_updated_at DESC
     private static final String SELECT_BY_IDS = """
             WITH target_spans AS (
                 SELECT id, trace_id, type
-                FROM spans FINAL
+                FROM spans
                 WHERE workspace_id = :workspace_id
                 <if(has_target_projects)>AND project_id IN :target_project_ids<endif>
                 AND trace_id IN :ids
+                ORDER BY (workspace_id, project_id, trace_id, parent_span_id, id) DESC, last_updated_at DESC
```
thiagohora, line 490: Agreed — the five-column key is right, for correctness and for keeping the read in order. No change needed here.

Two small actions:

1. **When this goes to OPIK_7638, don't apply "match the full sort key" mechanically.** The `traces` CTE further down in this same query uses the narrower `LIMIT 1 BY id` deliberately; widening it to the full key would fan one trace into multiple rows once its `project_id` is sentinel-ised, reproducing the `IndexOutOfBoundsException` 500 the javadoc above attributes to OPIK-7396. It needs to stay a per-site call.

2. **Description nit:** the sentinel is only reachable on `parent_span_id`. The `project_id` / `trace_id` branches write a value wider than their `FixedString(36)`, so they throw `TOO_LARGE_STRING_SIZE` (code 131) instead of storing a sentinel. Doesn't change the conclusion — it just means `parent_span_id` is the only column a regression test can exercise (see the test comment).

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -674,6 +676,20 @@ feedback_scores_grouped AS (
                     arrayMin(arrayMap(e -> e.8, entries)) AS created_at,
                     arrayMax(arrayMap(e -> e.9, entries)) AS last_updated_at
                 FROM span_feedback_scores_grouped
+            ), spans_deduped AS (
```
thiagohora, line 679: ### nit: the dedup block is now duplicated verbatim

`target_spans` (above) and `spans_deduped` carry byte-identical `WHERE` / `ORDER BY` / `LIMIT 1 BY` clauses, and `target_spans`' column list (`id, trace_id, type`) is a strict subset of this one. `target_spans` could just read from here:

```sql
), target_spans AS (
    SELECT id, trace_id, type FROM spans_deduped
),
```

(which needs this CTE moved above it).

Worth doing for a single source of truth on the dedup key, given the key was already wrong once in this PR's history — two copies is how they drift apart.

**Not** a performance change, to be clear: I measured it, and ClickHouse inlines CTEs rather than materialising them, so referencing `spans_deduped` twice re-scans and the query still reads `spans` twice either way. Identical rows/parts/bytes across four runs of each shape.

apps/opik-backend/src/test/java/com/comet/opik/domain/TraceDAOFindByIdsDedupIntegrationTest.java
```diff
@@ -0,0 +1,251 @@
...
+                .build();
+        UUID traceId = traceResourceClient.createTrace(trace, API_KEY, TEST_WORKSPACE);
+
+        UUID llmSpanId = createSpan(projectName, traceId, SpanType.llm,
+                Map.of(USAGE_KEY, USAGE_BEFORE_UPDATE), COST_BEFORE_UPDATE);
+        createSpan(projectName, traceId, SpanType.tool, null, null);
+        createSpan(projectName, traceId, SpanType.general, null, null);
+
+        // Re-inserts a full row under the same id, leaving two versions of the LLM span behind.
+        spanResourceClient.updateSpan(llmSpanId, SpanUpdate.builder()
```
thiagohora: ## The test passes under the exact bug the third commit fixed

The fixture updates the LLM span without changing its `parent_span_id`, so both stored versions share a sort key. I ran this exact scenario through all three forms:

| form | spanCount | llmSpanCount | usage | cost |
|---|---|---|---|---|
| `spans FINAL` | 3 | 1 | 70 | 7.00 |
| full-key dedup (this PR) | 3 | 1 | 70 | 7.00 |
| `LIMIT 1 BY id` | 3 | 1 | 70 | 7.00 |

All identical — so every assertion below holds under `LIMIT 1 BY id` too. The test guards "dedup happens", which a bare `FINAL` removal would break, but it cannot distinguish the by-id key from the full key. The regression that review actually caught is the one thing still untested.

`SpanUpdate` has a `parentSpanId` field, so the case is reachable through the same client: give the LLM span a real parent on create, then PATCH it with a **different** one. `SpanDAO`'s `UPDATE` writes a `'*'`-padded sentinel into `parent_span_id` (`SpanDAO.java:457`), so the span id ends up with two different sort-key values and the forms diverge:

| form | spanCount | llmSpanCount | usage | cost |
|---|---|---|---|---|
| `spans FINAL` | 4 | 2 | 80 | 8.00 |
| full-key dedup (this PR) | 4 | 2 | 80 | 8.00 |
| `LIMIT 1 BY id` | **3** | **1** | **10** | **1.00** |

A second test method on that shape — asserting the aggregate counts the span twice, matching `FINAL` — would pin the full-key choice. It reads oddly ("assert the double count"), so it's worth a comment saying the point is `FINAL`-equivalence under an upsert conflict, not that double-counting is desirable.

Separately, I checked the merge-race worry on the `rawVersionCount` precondition below and it looks unfounded — two single-row parts in a fresh container stayed unmerged past 45s, well beyond this test's window. Asserting rather than skipping is the right call and I'd leave it.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
YarivHashaiComet: as we spoke - droping this one for now

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
YarivHashaiComet: as we spoke, changing it to  LIMIT 1 BY workspace_id, project_id, id

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetsResourceCreateFromTracesTest.java
```diff
@@ -499,4 +505,176 @@ void createDatasetItemsFromTraces__withAllEnrichmentOptionsButNoData__success()
...
+
+        /**
+         * The batch endpoint inserts rows verbatim, so re-sending a span id under a different parent stores
+         * a second row that differs from the first on {@code parent_span_id} — a sort-key column. Both rows
+         * are then distinct versions as far as the table is concerned, and {@code FINAL} would keep them
+         * both, counting one logical span twice.
+         * <p>
+         * This query deliberately does not do that. It deduplicates on
+         * {@code (workspace_id, project_id, id)}, so a span id resolves to a single row within a project and
+         * is aggregated once. That is a conscious divergence from {@code FINAL}: a span id is meant to be
```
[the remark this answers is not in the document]
YarivHashaiComet: Good catch, fixed in 3ee56a8 — though the stale text was a few lines up rather than on this one.

The javadoc here (the divergent-parent test) already names the key explicitly. What was stale is the **nested class's** javadoc, which still said the dedup orders by and groups on the full sort key, and worse, listed *"deduplicating on a key narrower than the sort key"* as one of the regressions these tests catch. Since 5087bc0 that narrower key is the intended behaviour, so the class comment was contradicting the tests underneath it.

Restated as the logical span key `(workspace_id, project_id, id)`, ordered by the table's sort key so the read stays in order, with a note that diverging from `FINAL` for a span stored under two parents is deliberate — a span id is meant to be unique within a project, and the underlying data structure is being addressed under the hyperscale work.

Comment-only change; spotless and test-compile clean.

thiagohora (approved):

PR #7724, opened 2026-08-04
[OPIK-7105] [BE] [FE] refactor: remove Opik V1 workspace-version and project-migration machinery

andrescrz:
## Details

Opik V2 (project-first navigation) is the only supported experience, so the backend "workspace version determination" machinery (V1-vs-V2) and the project-id migration pipelines are dead weight. This removes them end-to-end and forces V2 on the client.

- **BE — version machinery:** removes `GET /v1/private/workspaces/versions`, `WorkspaceVersionService` (+ Abstract/Auth/Unauth), `OpikVersion`/`WorkspaceVersion`, the `opikVersion` field across the auth chain, the `forceWorkspaceVersion` + v1/v2 workspace-allowlist toggles and the `workspace_version` cache group, the now-unused `hasVersion1*` / legacy-scores probe DAO methods, and the orphaned `WorkspacesService`/`WorkspacesDAO` `findById` + `Workspace` record. Liquibase `000093` drops `workspaces.last_known_version` + `version_determined_at`.
- **BE — project-migration pipelines:** removes all six (Experiment, Dataset, Optimization, Prompt, AutomationRule, Alert) — services, scheduled jobs, configs, migration-only DAO methods/DTOs, and their tests. Safe because project-id migration is complete across all deployments (the jobs were the only code assigning `project_id` to legacy entities). Kept: general `project_id`-null fallbacks, the stalled-optimization reaper, `ProjectService.DEFAULT_PROJECT`, and the dataset-versioning migration.
- **BE — compatibility:** the `has_legacy_scores` reader is kept (trace/span stats union gating still uses it). The auth EM response and Redis auth cache stay backward/forward compatible — unknown fields on the response and a stale `opikVersion` field on old cache records are ignored.
- **FE (minimal):** every workspace is pinned to V2 (`resolveSyncWorkspaceVersion` + the post-mount `WorkspaceVersionResolver`), and `fetchWorkspaceVersion` no longer calls the removed endpoint. This fixes the Opik Connect pairing page, which showed a "requires Opik 2.0" error when a workspace resolved to V1 from stale local state.
- **Deployment:** removes `TOGGLE_FORCE_WORKSPACE_VERSION` from docker-compose.

Out of scope (separate follow-ups): the V1 frontend structure, SDK/OpenAPI/docs regeneration, and E2E seeding cleanup.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves OPIK-7105

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: full implementation
- Human verification: iterative human code review across multiple passes + local backend build/tests

## Testing

Local (backend, offline Maven):
- `mvn compile` and `mvn test-compile` — green.
- App boot (`IsAliveResourceTest`) — validates Guice/config wiring and Liquibase (incl. the new `000093` migration) after the removals.
- Auth compatibility — `RemoteAuthServiceTest#testAuth__whenResponseHasUnknownFields__thenIgnoredAndAuthSucceeds` and `AuthCredentialsCacheServiceTest#resolveV2Cache__whenRecordHasStaleLegacyField__thenIgnoredAndResolves`.
- `has_legacy_scores` / legacy-`feedback_scores` UNION — exercised by `ProjectsResourceTest$FindProject#getProjects__whenLegacyScoresHasData__thenStatsIncludeThem` and `MultiValueFeedbackScoresE2ETest#{testTraceStats,testSpanStats}_legacyFeedbackScoresSurfacedWhenPresent` (both seed the legacy table via `scoreEntity(author=null)` and assert the stats UNION). `AuthCredentialsCacheServiceTest` (10 tests) also covers the workspaces cache path after the `ConstructorMapper` removal.

Frontend not run locally (worktree has no `node_modules`) — typecheck/lint/vitest, including the new `workspaceVersion.test.ts`, run in CI. The full backend suite also runs in CI.

## Documentation

N/A — no user-facing documentation changes. SDK/OpenAPI/docs regeneration is a separate follow-up (out of scope here).


apps/opik-frontend/src/lib/workspaceVersion.test.ts
```diff
@@ -0,0 +1,31 @@
+import { describe, it, expect, beforeEach } from "vitest";
+import { resolveSyncWorkspaceVersion } from "./workspaceVersion";
```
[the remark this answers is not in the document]
andrescrz: Skipping — this repo's convention is co-located tests. There are 157 `src/**/*.test.ts(x)` files and **no** top-level `tests/` directory (including ~10 other `src/lib/*.test.ts`). Placing this test next to the module it covers matches every existing frontend test; moving it under a `tests/` dir would make it the sole outlier. Vitest still discovers it via the standard `**/*.test.{ts,tsx}` glob either way.

apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/000093_drop_workspace_version_columns.sql
```diff
@@ -0,0 +1,9 @@
+--liquibase formatted sql
+--changeset andrescrz:000093_drop_workspace_version_columns
+--comment: Drop the workspace-version-determination columns now that V1 is deprecated and V2 is the only navigation. The version endpoint and its service have been removed, so nothing reads or writes these columns. first_trace_reported_at and has_legacy_scores are kept.
+
+ALTER TABLE workspaces
+    DROP COLUMN last_known_version,
+    DROP COLUMN version_determined_at;
+
+--rollback ALTER TABLE workspaces ADD COLUMN last_known_version ENUM('version_1', 'version_2') DEFAULT NULL AFTER id, ADD COLUMN version_determined_at TIMESTAMP(6) DEFAULT NULL AFTER last_known_version;
```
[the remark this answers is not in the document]
andrescrz: Skipping — this rollback matches the established convention for these column-drop migrations: `000077/000078/000079_drop_*_migration_columns_from_workspaces` re-add their dropped columns the same way. Liquibase rollbacks are schema-level and never restore dropped data. Here the dropped data is workspace **version-determination** state that V2 intentionally discards, and re-adding the empty columns restores the schema so an older app can still boot (it tolerated null versions via its default path). A forward-only no-op would diverge from the precedent for no real recovery benefit.

CometActions:

### Access Information
- **URL:** https://pr-7724.dev.comet.com
- **Namespace:** pr-7724
- **Version:** 2.2.18-7724-merge-2808
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7724%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

apps/opik-backend/src/main/java/com/comet/opik/domain/workspaces/WorkspacesDAO.java
```diff
@@ -63,19 +45,4 @@ void insertFirstTrace(@Bind("id") String id,
      */
     @SqlQuery("SELECT has_legacy_scores FROM workspaces WHERE id = :id")
     Optional<Boolean> findHasLegacyScores(@Bind("id") String id);
```
JetoPistola, line 47: 💡 **suggestion** | Performance

`has_legacy_scores` keeps this reader (gating the legacy `feedback_scores` UNION in `TraceDAO`/`SpanDAO`), but this PR removes its **only writer** — `git grep upsertHasLegacyScores origin/main` shows the sole caller was `WorkspaceVersionService:236`.

The column defaults to `TRUE`, so this is behavior-safe (always UNION, never silently drops legacy scores). But the Javadoc directly above this method still says:

> An admin/backfill flow flips workspaces with no legacy data to FALSE so their stats queries skip the empty-table scan.

No such flow exists after this PR. Net effect: every workspace with `has_legacy_scores = TRUE` — all new ones, plus any never touched by version determination — permanently pays the legacy-table UNION on trace/span/project stats, with nothing left that can ever flip it to FALSE.

Two reasonable resolutions: file a follow-up for a backfill/admin flip so the optimization stays reachable, or trim the Javadoc claim so it doesn't promise a flow that no longer exists.

apps/opik-frontend/src/lib/workspaceVersion.test.ts
```diff
@@ -0,0 +1,31 @@
...
+    ],
+    [
+      "a stale cached v1 for the workspace",
+      () =>
+        localStorage.setItem(
+          "opik-workspace-versions",
+          JSON.stringify({ "ws-1": "v1" }),
+        ),
+    ],
+  ])("returns v2 with %s", (_label, seedLocalState) => {
```
JetoPistola, line 27: ❓ **question** | Testing

`resolveSyncWorkspaceVersion()` is now just `return DEFAULT_WORKSPACE_VERSION;` — it reads no `localStorage` at all. The three `it.each` cases seed override/cache keys the function can no longer see, so all three reduce to the same tautological assertion.

As a regression guard against *re-introducing* the localStorage read that's genuinely useful, but the comment above reads as though it exercises real branches ("a stale `opik-version-override=v1` / cached `v1` is exactly what forced a workspace into V1App") — which is now the one thing this test can't observe.

Is the intent a canary, or was the guard meant to sit at the `WorkspaceVersionGate` level, where a v1 mount would actually be observable?

apps/opik-frontend/src/shared/WorkspaceVersionResolver/WorkspaceVersionResolver.tsx
```diff
@@ -58,7 +37,7 @@ const WorkspaceVersionResolver: React.FC<WorkspaceVersionResolverProps> = ({
   }
 
   const { data: apiVersion } = useWorkspaceVersionQuery();
-  const resolvedVersion = override ?? (optIn ? "v2" : apiVersion);
+  const resolvedVersion: WorkspaceVersion = "v2";
```
JetoPistola, line 40: 💡 **suggestion** | Dead code

With `resolvedVersion` a hardcoded `"v2"` const, `resolvedVersion !== gateVersion` below is dead — and with it `VERSION_RELOAD_PREFIX`, `MAX_RELOADS`, and the entire `originalUrlByWorkspaceRef` capture machinery (lines 10–11, 31–36, 52–65). `PairRouteVersionGuard.tsx` has the same shape.

The doc comments honestly label the path unreachable and defer to the V1-UI removal, which I appreciate. The concern is narrower: ~35 lines of load-bearing-*looking* reload logic that cannot fire is an invitation for a future reader to "fix" a bug that isn't reachable, or to preserve it during the V1 cleanup because it looks essential.

Consider collapsing this to just the store/cache write now and deleting the reload block — the V1-UI removal then has strictly less to reason about.

apps/opik-frontend/src/lib/workspaceVersion.ts
```diff
@@ -92,14 +85,9 @@ export function navigateToWorkspaceRoot(workspaceName: string): void {
   window.location.href = normalizedBase + workspaceName;
 }
 
-// Always returns a version synchronously: override > per-workspace cache >
-// default. WorkspaceVersionResolver verifies via API after mount and swaps
-// the App in place if the guess was wrong.
+// Opik V2 is the only supported experience: every workspace resolves to V2
+// regardless of any legacy override / opt-in / cached value. This is the
+// client-side replacement for the removed backend forceWorkspaceVersion default.
 export function resolveSyncWorkspaceVersion(): WorkspaceVersion {
```
JetoPistola, line 91: 💡 **suggestion** | Dead code

`getVersionOverride` and `getNewExperienceOptIn`/`setNewExperienceOptIn` survive this PR with exactly one consumer left: `UserMenu.tsx:65`

```ts
const showNewExperienceToggle =
  detectedWorkspaceVersion === "v1" && getVersionOverride() === null;
```

`setDetectedWorkspaceVersion` is now only ever fed `apiVersion`, and `fetchWorkspaceVersion` hardcodes `"v2"` — so `detectedWorkspaceVersion === "v1"` is never true and that toggle can never render.

Same "kept until V1 UI removal" reasoning as the resolver, so not asking for it here — flagging so it lands on the follow-up ticket's list rather than being rediscovered later.

JetoPistola:
👋 **Review summary**

**What looks good**
- The rolling-upgrade compatibility story is well-handled and, unusually, **tested in both directions**: `RemoteAuthServiceTest#testAuth__whenResponseHasUnknownFields__thenIgnoredAndAuthSucceeds` (old EM → new BE) and `AuthCredentialsCacheServiceTest#resolveV2Cache__whenRecordHasStaleLegacyField__thenIgnoredAndResolves` (stale Redis record → new BE). That's exactly the failure mode a removal of this shape usually ships broken.
- Migration `000093` correctly preserves `first_trace_reported_at` and `has_legacy_scores`, and its rollback matches the `000077/078/079` precedent.
- The `markFirstTraceReported` Javadoc rewrite is a real correctness improvement rather than comment churn — `git grep "INSERT INTO workspaces"` now returns exactly one hit, so the new claim ("this is the only path that inserts") holds and the old "unrelated writer" caveat is genuinely obsolete.
- Scope discipline: the `has_legacy_scores` reader, the stalled-optimization reaper, `ProjectService.DEFAULT_PROJECT`, and the dataset-versioning migration all correctly survive the cull.
- SDK exposure checked out — `getWorkspaceVersion` / `get_workspace_version` appear only in generated `rest_api` code and the OpenAPI specs, with no hand-written SDK callers, consistent with regeneration being a stated follow-up.

**Overall**
The approach makes sense and the scope is appropriate: forcing V2 client-side is the right shape for retiring the endpoint without a coordinated FE/BE deploy. CI is fully green (23 Backend Integration groups, 16 Integration groups, unit tests, CodeQL). Nothing here is blocking. The one finding with production consequences is the `has_legacy_scores` writer removal — a perf regression, not a correctness bug, since the column defaults to safe-include.

The recurring theme across three of the findings is intentionally-retained-but-now-unreachable code (the resolver reload path, the override/opt-in helpers, the pinned test). Each is individually defensible with the V1-UI-removal follow-up in mind; collectively it's worth making sure that follow-up ticket enumerates them, so they get deleted rather than preserved as apparently-load-bearing.

**One additional item that couldn't be anchored inline** (the files aren't part of this diff):

🧹 **nit** | Consistency — `.github/workflows/end2end_suites_v2.yml:90` and `.github/workflows/e2e_tests_post_merge_v2.yml:127`

This PR removes `TOGGLE_FORCE_WORKSPACE_VERSION` from `docker-compose.yaml` and `config.yml`, but both E2E workflows still export `TOGGLE_FORCE_WORKSPACE_VERSION="version_2"` in their `opik.sh --build` invocation. Harmless — it's an unreferenced env var rather than a config key, so nothing fails — but it now reads as though it still does something. The PR body scopes E2E cleanup out, so entirely your call; it's a one-token deletion on two lines if you'd rather not leave the loose end.

**Inline comments**: 3 suggestions, 1 question (plus this nit) — nothing blocking.

apps/opik-frontend/src/shared/WorkspaceVersionResolver/WorkspaceVersionResolver.tsx
andrescrz: Deferring rather than collapsing here, to keep this PR's FE surface minimal — the V1-UI removal is a separate follow-up. The doc comments already mark the reload path unreachable, and I've recorded the reload machinery in `WorkspaceVersionResolver`/`PairRouteVersionGuard` in the commit body so the follow-up enumerates it for deletion rather than preserving it as apparently load-bearing.

apps/opik-frontend/src/lib/workspaceVersion.ts
andrescrz: Agreed — leaving these for the follow-up, as you suggested. Recorded in the commit body alongside the resolver reload machinery so the V1-UI-removal ticket removes `getVersionOverride` / `getNewExperienceOptIn` / `setNewExperienceOptIn` and the `UserMenu` "New Experience" toggle together, rather than having them rediscovered later.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/ProjectsResourceTest.java
```diff
@@ -1448,14 +1447,14 @@ void getProjects__whenProjectsHasTracesSpansFeedbackScoresAndUsage__thenReturnPr
...
+        @DisplayName("when has_legacy_scores gate is on, stats endpoint stays consistent")
+        void getProjects__whenHasLegacyScoresGateOn__thenStatsStayConsistent() {
             // Test infra writes feedback scores through the authenticated path, so data lands in
             // authored_feedback_scores (not the legacy feedback_scores table). This test can't
             // observe the legacy-UNION gate directly — that surface is covered by the existing
             // FilterTest variants and by manual benchmarking against real legacy data. What this
-            // test does verify: flipping the workspace flag does not break the endpoint and the
-            // response stays correct for data that lives in authored_feedback_scores.
+            // test does verify: with the default has_legacy_scores gate on, the endpoint stays
+            // correct for data that lives in authored_feedback_scores.
```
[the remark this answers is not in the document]
andrescrz: Fixed in 45974bd1a8 — removed the redundant, misnamed test. Verified the legacy `feedback_scores` UNION is genuinely covered by the baseline `ProjectsResourceTest$FindProject#getProjects__whenLegacyScoresHasData__thenStatsIncludeThem` and `MultiValueFeedbackScoresE2ETest#{testTraceStats,testSpanStats}_legacyFeedbackScoresSurfacedWhenPresent`, which seed the legacy table via `scoreEntity(author=null)` and assert the stats UNION.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/auth/AuthCredentialsCacheServiceTest.java
```diff
@@ -32,6 +33,7 @@ public class AuthCredentialsCacheServiceTest {
     private static final int CACHE_TTL_IN_SECONDS = 1;
 
     private final AuthCredentialsCacheService cacheService;
+    private final RedissonReactiveClient redisClient;
 
     private final PodamFactory podamFactory = PodamFactoryUtils.newPodamFactory();
```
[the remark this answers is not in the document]
andrescrz: Fixed in 45974bd1a8 — the test now shuts down the owning `RedissonClient` in `@AfterAll` (`RedissonReactiveClient#shutdown()` is deprecated in favour of `RedissonClient#shutdown()`); the reactive client is just a view over it.

apps/opik-backend/src/main/java/com/comet/opik/domain/workspaces/WorkspacesDAO.java
andrescrz: Fixed in 966ce0062a — trimmed the Javadoc so it no longer promises the admin/backfill FALSE-flip flow that this PR removed; it now states the column has no writer, stays at its `TRUE` default (always-UNION), until the legacy `feedback_scores` table is decommissioned. Didn't file a backfill follow-up: the ClickHouse revamp merges the legacy table into `authored_feedback_scores` and drops it, collapsing the UNION entirely, so the always-UNION cost is transient.

apps/opik-frontend/src/lib/workspaceVersion.test.ts
andrescrz: Fixed in 966ce0062a — reframed the comment as an honest regression guard: the cases assert `resolveSyncWorkspaceVersion` ignores legacy local state and would fail only if the localStorage read were re-introduced. Kept the three cases as documentation of the exact keys (override, per-workspace cache) that must stay ignored, rather than moving to a heavier Gate-level test in the to-be-removed machinery.

andrescrz:
Re the E2E-workflow nit in your review summary: Fixed in 966ce0062a — dropped the now-dead `TOGGLE_FORCE_WORKSPACE_VERSION` export from `end2end_suites_v2.yml` and `e2e_tests_post_merge_v2.yml`.

thiagohora (approved):
So satisfying

andriidudar (approved):
FE part looks good, approved! 

First step before cleanup Frontend!

PR #7725, opened 2026-08-04
[NA] [BE] Persist the fast-mode rate modifier on the cipx analytics tables

aadereiko:
## Details

cipx now promotes `speed` onto `cipx.call.config`; this persists it on the cipx analytics tables so spend can be priced correctly. It changes the **rate** rather than the token count — `speed=fast` bills Opus 5 / 4.8 at $10/$50 per MTok instead of $5/$25, with the prompt-cache multipliers stacking on top — so a call can't be priced without it. For scale: **66.6% of prod tokens are on fast-tier-eligible models** (309.7B tokens since 2026-06-22).

- **Both tables, not just `cipx_spends`.** `cipx_spend_blocks` already denormalises `model` off the same `cipx.call` node, and the per-user lane breakdowns (skills, MCP servers, built-in tools) price from blocks — a block without the modifier would price at the standard rate regardless of what `cipx_spends` says. Joining back per block would defeat the denormalisation the table exists for.
- **Per-call, not per-session.** When the fast rate-limit pool or the account's usage credits run dry, Claude Code silently falls back to standard speed mid-session, so one session legitimately mixes fast- and standard-priced calls.
- Migration `000117` adds an additive `LowCardinality(String) DEFAULT ''` column to each table, mirroring `000103`. Existing rows read `''` = standard speed, the correct reading for pre-fast-mode history, so no backfill and no reprice of past data.
- Read path is `cipx.call.config` (alongside `effort` / `thinking_type`), **not** the top level of `cipx.call` where `model` lives. Claude Code sends it at the top level of the *Anthropic* request body; the two "top levels" are different objects, and the migration comment says so explicitly so nobody infers the wire path from where CC puts it.

An earlier revision also added `inference_geo` (data residency, 1.1x). Deferred — independent modifier, nothing in Claude Code sets it, no prod traffic carries it. Note for whoever picks it up: captured cipx fixtures show the *response* usage block carries `inference_geo` next to `service_tier`, i.e. what actually happened rather than what was asked for, so that column would hang off usage rather than the config knobs.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-

No Jira ticket — filed under `[NA]` as ingestion plumbing for the AI-Spend pricing fix. Related but **not** resolved here: OPIK_7730 (fast-mode tokens are bucketed as seat-covered, which is a cost-API billing-split change, not an ingestion one).

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: Investigation of how fast mode appears on the wire (verified against the shipped Claude Code 2.1.220 binary), the migration, the DAO mapping on both tables, and the test updates.
- Human verification: Author reviewed the diff and the migration. Verified by the integration suite below against a real ClickHouse container — the read-back asserts the value lands on the spend row and on every block row, so the assertions fail if the mapping or the migration is wrong.

## Testing

```
cd apps/opik-backend
mvn -o test -Dtest=CostIntelligenceIngestionTest   # 9/9 pass
mvn -o spotless:check                              # clean
```

Scenarios validated:
- **Happy path** — a cipx span whose `cipx.call.config` carries `speed: "fast"`: it lands on the `cipx_spends` row and on every `cipx_spend_blocks` row, read back from ClickHouse rather than asserted on the Java object.
- **Absent config** — `spanCipxWriteBlockMetadata` carries no `config` block at all, so the default-to-`''` path stays covered; that's what every pre-migration row reads as.
- **Non-cipx spans** — unchanged: still produce no rows.

Not run: the full backend suite locally (CI covers it). Note on CI history for this PR — an earlier run failed `AutomationRuleEvaluatorsResourceTest$GetLogs.getLogsTraceSkipped` in Integration Group 13, which is unrelated (automation-rule evaluator logs, no cipx path; a broken migration would fail the whole application context rather than one test) and passed on re-run. `check-migration-prefix-conflicts` also failed once when `000116` was taken on main mid-flight; renumbered to `000117`.

## Documentation

None needed — internal analytics schema plus ingestion mapping, no user-facing surface and no API change. The wire contract itself is documented in `pkg/wire/CHANGELOG.md` in the cipx repo (comet-ml/cost-intelligence-proxy-internal#89).

## Ships with

1. comet-ml/cost-intelligence-proxy-internal#89 — capture the field
2. **this PR** — persist it
3. comet-ml/ai-cost-backend#70 — carry the dimension through (no pricing change)
4. comet-ml/ai-cost-backend#68 — price it

Data has to land before it can be priced, so merge in that order. This PR is inert on its own: the column stays empty until the cipx build ships.


aadereiko:
Fair catch — fixed in `ebf5134`, comments only.

The comment was accurate about the *Anthropic* request body but misleading in a file about cipx columns: the DAOs read these from `cipx.call.config` (alongside `effort` / `thinking_type`), not from the top level of `cipx.call` where cipx's own `model` lives. Two different objects, each with a `model` sitting next to them — easy to conflate, and the whole point of that comment is provenance.

It now leads with the read path and mentions the CC-side origin second, with an explicit warning not to infer the wire path from where Claude Code puts the fields. Applied the same correction to the `BlockRow` field comment, which implied `model` also came from `config`.

The readers were already correct, so nothing behavioural changed.

aadereiko:
Scope reduced: **`inference_geo` is deferred**, this now adds `speed` only, and the migration is renamed `000116_add_speed_to_cipx_tables.sql` since it adds one column rather than two. Nothing has run it yet, so there's no changeset-checksum concern.

It turned out to be an independent rate modifier that merely sat next to fast mode — data residency (1.1x when a customer pins inference to the US), nothing to do with speed. Claude Code never sets it and no prod traffic carries it.

Note for whoever picks it up: the request-side capture is probably the wrong source. Captured cipx fixtures show the *response* usage block carries `inference_geo` next to `service_tier` — what actually happened rather than what was asked for — so the column would hang off usage, not the config knobs.

Ingestion test still 9/9 against a real ClickHouse container, spotless clean.

aadereiko:
Both findings are the same one, and it's a timing artefact rather than a defect.

You reviewed at **14:36–14:45**; I rewrote this PR's description at **14:48**. So the code you saw was correct and the description you compared it against was stale — hence "despite the PR contract".

`inference_geo` was deliberately dropped from the whole stack. It's an independent rate modifier — data residency, 1.1x when a customer pins inference to the US — that merely sat next to fast mode; Claude Code never sets it, and **zero prod rows carry it** (checked against prod: 1.42M rows, 34 distinct models, and no provider-prefixed model ids either, so the cloud-regional premium doesn't apply here either).

Your comment offered exactly the right alternative — *"or explicitly defer geo pricing"* — which is what the updated description, the migration comment and the commit message now all do.

Worth carrying forward for whoever implements geo: **the request-side capture would have been the wrong source anyway.** Captured cipx fixtures show the *response* `usage` block carries `inference_geo` next to `service_tier` (value `"not_available"` on all 9 samples) — what actually happened rather than what was asked for. So the column should hang off usage rather than the config knobs, which is a different shape from what this PR would have persisted.

Unchanged otherwise: 9/9 ingestion tests against a real ClickHouse container, spotless clean, and the migration is renumbered `000117` after `000116` was taken on main mid-flight.

aadereiko:
Deferred `inference_geo` is now tracked as **[OPIK-7757](https://comet-ml.atlassian.net/browse/OPIK-7757)**. It records that the column should hang off the usage counters rather than the config knobs, since the field arrives on the response `usage` block next to `service_tier` — different shape from what this PR would have persisted.

[OPIK-7757]: https://comet-ml.atlassian.net/browse/OPIK-7757?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

LifeXplorer (commented):
The implemementation looks good. 1 blocking thing to update are the comments. Opik is OSS project. We should be as mean as possible when providing comments or documenting how cipx-related functionality works.

aadereiko:
Thanks — fixed in `b550ef8`. You're right, and it was worse than just verbosity: the migration comment carried three things that have no business in a public repo.

**What came out:**
- Anthropic's per-MTok rates (`$10/$50` vs `$5/$25`) — commercial detail
- Claude Code's internal fast-mode fallback behaviour (rate-limit pool / usage-credit exhaustion)
- An essay on why the wire path is easy to misread

**What stayed** — what a reader of *this* repo actually needs: what the column mirrors, where it's parsed from, that it's per-call rather than per-session, what `''` means, and why it's denormalized onto blocks as well as spends.

29 lines → 12, which lines up with `000112_add_subcategory_to_cipx_spend_blocks.sql` sitting next to it (11). I used that one as the reference for how much is enough.

Also trimmed the `BlockRow` field comment (4 lines → 3, matching `subcategory` beside it) and the two test comments that framed assertions in pricing terms.

Comments only, no behaviour change — ingestion test still 9/9 against a real ClickHouse container, spotless clean.

Worth noting the pricing rationale isn't lost, just relocated to where it belongs: the internal cipx repo's `pkg/wire/CHANGELOG.md` and the cost-api pricing PR (ai-cost-backend#68), both private.

aadereiko:
Trimmed further in `e7cc743` — you're right that the structure was the remaining problem, not just the length.

What was still leaking cipx internals:
- the Go type behind the column (`wire.CallConfig.Speed`)
- the JSON path it's parsed from (`cipx.call.config`)
- the sibling fields sharing that object (`effort` / `thinking_type` / `max_tokens`) — arguably the worst of the three, since it maps out a structure this repo never needs to know

The migration comment is now three lines of what the column means *here*:

```sql
-- Selects which rate table prices a call. Per-call, not per-session: one session can carry
-- more than one value. '' means standard, including every row written before the field existed.
-- Carried on blocks as well as spends because block-level cost needs the value that priced it.
```

Same treatment on the `BlockRow` field comment and the two test comments.

One thing I can't remove: the parsing path is still visible in the *code* that parses it (`metadata.path("cipx").path("call")`, `config.path("speed")`). That's unavoidable — and it's also why the comments were redundant, so dropping them costs nothing.

Worth flagging that this is deliberately tighter than `000112` beside it, which does name a cipx Go type (`mirroring cipx wire.Block.Subcategory`). Happy to leave that inconsistency, or to trim that one in a separate PR if you'd rather the whole set matched — your call.

Comments only. Ingestion test 9/9, spotless clean.

LifeXplorer (approved):
Looks good

PR #7748, opened 2026-08-05
[OPIK-7725] [BE] [CI] feat: block SQL built from formatted strings with a semgrep pre-commit hook

JetoPistola:
## Details

image

Adds semgrep to `.pre-commit-config.yaml` with a single in-repo rule that fails when a Java SQL literal carries a `%s` format slot in a clause position, and converts the two existing violations so the gate lands green with no baseline and no suppressions. A PR review asked for this rule to be enforced automatically rather than relying on reviewers noticing; documentation and AI review are advisory, so a violation from any author could still merge.

- **Matches the query declaration, not the call site.** Opik holds SQL in `static final` text blocks rendered through StringTemplate and handed to `connection.createStatement(template.render())`, so the literal and the execution are connected only through a template object — call-site patterns match nothing. `pattern: $X` plus a `metavariable-regex` inspects the literal's own contents, text blocks included.
- **Two tiers, so pre-existing code is reported rather than blocking.** Covering every statement type and every slot position surfaces **108 findings in production code** that predate this gate. Blocking on all of them would stall the gate or need a baseline file, so the rule splits: `sql-query-clause-splice` (**ERROR**, blocks CI) matches only the injection-shaped form — a `%s` standing in for a whole clause, where caller data reaches the query text — and currently has **0 findings**; `sql-query-format-slot` (**WARNING**, reports only) matches any other `%s` in a SQL literal and currently has **108**, tracked as tech debt in OPIK_7727. The hook passes `--severity ERROR`, so only the blocking tier can fail a commit or a CI leg.
- **Both tiers cover INSERT/UPDATE/DELETE and DDL**, not just `SELECT`/`WITH` — a spliced predicate in an `UPDATE` is as injectable as one in a `SELECT`, and it mutates rather than reads.
- **Production sources only.** String formatting in a test fixture is building a fixture, not accepting user input, so `files:` is scoped to `src/main/java`.
- **Fixtures pin all of it.** `.semgrep/java-sql-string-formatting.java` asserts each must-flag and must-not-flag case with semgrep's own annotations, runnable via `semgrep test`. Every negative case in there matched at some point during development, so the file is the regression contract for future regex edits.
- **`%d` is deliberately not matched** — `MetadataDAO` contains the real SQL literal `STR_TO_DATE(value, '%Y-%m-%d')`, which pattern matching cannot distinguish from an integer slot. No real violation uses `%d`.
- **The fixes.** `SpanMetricsQueries`' 3+1 `%s` project-predicate slots become a single unconditional `AND project_id IN :project_ids`. Both callers bind `project_ids` — a set of one for per-project aggregation, a resolved set for workspace-level — so the query text is a constant with no conditional at all. Both helper methods and the `String projectPredicate` parameters are gone.
- **The bind is guarded, and this is the sharp edge.** `getMetric` renders ~37 queries, but only the 11 `GET_SPAN_*` ones embed the shared CTE where `:project_ids` appears. Binding it unconditionally raises `NoSuchElementException` from R2DBC on every trace and thread query — the same failure OPIK_5678 hit in this exact method. The bind is therefore guarded by `SPAN_TIME_METRICS`, which maps exactly onto those 11 constants. Compilation does not catch this: R2DBC resolves binds by name at execution.

Because `.pre-commit-config.yaml` is the single source of truth that `code_quality.yml` builds its CI matrix from, adding the hook makes CI fail on new violations with no workflow change. The hook is `language: python`, so the pre-commit framework provisions semgrep and contributors need no local install.

Known limitation, documented in the rule file: a query written entirely on one line is not caught, because the `SELECT`/`WITH` and `FROM` requirements are line-anchored. Dropping those anchors was measured and reintroduces the prompt-template false positive, which is the worse failure — it would block an unrelated commit and push a suppression comment into a file with no SQL in it. Every real query in this codebase is a multi-line text block.

## Change checklist
- [x] User facing
- [x] Documentation update

## Issues

- Resolves OPIK-7725

Deferred to follow-ups, deliberately out of scope here so neither blocks this gate: the `+` concatenation form (a prototype rule produced 118 findings, all false positives — OPIK_7726) and the constant-splice `%s` sites in the metrics DAOs and ClickHouse health checks (OPIK_7727).

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: full implementation — semgrep rule authoring and empirical scoping, pre-commit wiring, the `SpanMetricsQueries` conversion, and verification
- Human verification: code review + reviewed measurement output for each claim below

## Testing

Semgrep rules, run as the pinned `semgrep==1.172.0` the way the hook runs them — explicit filenames via `git ls-files`, not a directory:

- **Blocking tier: 0 findings across all 1,259 production Java files.** No baseline file, no suppression comments.
- **Reporting tier: 108 findings**, visible but non-blocking. These are the pre-existing occurrences the review asked to surface rather than block; tracked as tech debt in OPIK_7727.
- Verified the rollout actually behaves that way: `pre-commit run semgrep --files ProjectMetricsDAO.java` **passes**, on a file carrying 99 reporting-tier findings.

  > **Note on how coverage is measured.** An earlier revision of this PR claimed "0 findings across 1,296 files (main + test)". That was wrong. Semgrep's built-in semgrepignore excludes `test`/`tests` directories at any depth, so passing a test directory scans zero files while still reporting a reassuring "0 findings". The pre-commit hook passes explicit **filenames**, which bypasses that exclusion. The trap is documented in the rule file so the next person measures the way the hook runs. (Test sources are now out of scope entirely, per review, but the measurement lesson stands for the production tree.)

- Rule fixtures: `semgrep test --config .semgrep/java-sql-string-formatting.yaml .semgrep/java-sql-string-formatting.java` — **2/2 rules, all assertions pass**. Each case is asserted against both rules independently, so a clause splice must be caught by both while a projection must be caught only by the reporting one. The fixture file sits outside the hook's `files:` scope, so its deliberate violations can't fail CI — verified via `precommit-detect-hooks.py`.
- CI matrix routing: a changed production `.java` file emits the `semgrep` leg with `toolchain: none`; a changed **test** file emits no semgrep leg (only spotless), confirming the narrowed scope.

Backend build and tests, on JDK 25:

- `mvn compile -DskipTests` and spotless — pass, spotless reformatted nothing.
- `ProjectMetricsResourceTest$SpanCountTest` — passes against the new bind contract (3 cases, confirmed per-case in the surefire XML).
- **Not verified locally: the trace/thread path.** This is the one that matters most, because `getMetric` binds `project_ids` only for span metrics — an unguarded bind would raise `NoSuchElementException` on every trace and thread query. The guard is derived from `SPAN_TIME_METRICS`, which was checked to map exactly onto the 11 `GET_SPAN_*` constants that embed the shared CTE, but Testcontainers failed to start ClickHouse on roughly half the local attempts on this machine, so `NumberOfTracesTest` was left to CI rather than probed serially. If trace metrics go red in CI, that is this change, not a flake.

## Documentation

The `CONTRIBUTING.md` section added earlier was trimmed to a two-line pointer, per review: the conventions this gate enforces are already documented in `.agents/rules/security.mdc` and the backend skill, so restating them in the contribution guide was duplication. Rationale for the rules' scoping decisions — including why `%d` is excluded and how to measure coverage correctly — lives in comments in the rule file itself.

### DIFF
--- a/.semgrep/java-sql-string-formatting.java
+++ b/.semgrep/java-sql-string-formatting.java
@@ -1,0 +1,126 @@
+// Fixtures for java-sql-string-formatting.yaml. Run with:
+//   semgrep test --config .semgrep/java-sql-string-formatting.yaml .semgrep/java-sql-string-formatting.java
+//
+// Semgrep's own annotations drive the assertions: a rule-id comment above a line marks it
+// as one the rule MUST flag, and the negative form marks one it must NOT.
+//
+// The must-not-flag cases below are the ones that actually bit. Every one of them matched
+// at some point during development, and the quoted-value and projection cases were caught
+// in PR review rather than by measurement. Keep them — they are the regression contract
+// for any future edit to the regex.
+class JavaSqlStringFormattingFixtures {
+
+    // ---------------------------------------------------------------------------
+    // MUST FLAG — a `%s` standing in for a clause, the injection-shaped pattern.
+    // ---------------------------------------------------------------------------
+
+    // ruleid: sql-query-with-format-slot
+    static final String CLAUSE_SPLICE_AND = """
+            SELECT id
+            FROM spans
+            WHERE workspace_id = :workspace_id
+            AND %s
+            """;
+
+    // ruleid: sql-query-with-format-slot
+    static final String CLAUSE_SPLICE_WHERE = """
+            SELECT name
+            FROM system.columns
+            WHERE %s
+            """;
+
+    // ruleid: sql-query-with-format-slot
+    static final String CLAUSE_SPLICE_LOWERCASE = """
+            select id
+            from spans
+            where workspace_id = :w
+            and %s
+            """;
+
+    // ruleid: sql-query-with-format-slot
+    static final String CLAUSE_SPLICE_NESTED_WITH = """
+            WITH x AS (
+            SELECT id
+            FROM spans
+            WHERE a = 1
+              AND %s
+            )
+            """;
+
+    // ---------------------------------------------------------------------------
+    // MUST NOT FLAG — value and projection positions, and non-SQL strings.
+    // ---------------------------------------------------------------------------
+
+    // A slot inside a quoted SQL literal is a value position, not a clause. The ClickHouse
+    // health checks fill these from compile-time constants (OPIK_7727 tracks the cleanup).
+    // ok: sql-query-with-format-slot
+    static final String VALUE_SLOT_TIGHT_QUOTES = """
+            SELECT count() AS cnt
+            FROM system.disks
+            WHERE name = '%s'
+            """;
+
+    // Same, with whitespace inside the quotes — the `(?<!')%s(?!')` form missed this.
+    // ok: sql-query-with-format-slot
+    static final String VALUE_SLOT_SPACED_QUOTES = """
+            SELECT count() AS cnt
+            FROM system.clusters
+            WHERE cluster = ' %s '
+            """;
+
+    // A projection: the slot names a column, it does not splice a predicate. Real instances
+    // live in SpansLocalV2TableTest / NaNAwareAggregateIntegrationTest / TracesLocalV2CutoverTest.
+    // ok: sql-query-with-format-slot
+    static final String PROJECTION_SLOT = """
+            SELECT %s AS value
+            FROM spans_local_v2
+            WHERE workspace_id = :workspace_id
+            """;
+
+    // ok: sql-query-with-format-slot
+    static final String PROJECTION_SLOT_IN_CTE = """
+            WITH new_trace AS (
+              SELECT toFixedString('%s', 36) AS id,
+              %s AS ttft
+            )
+            SELECT ttft
+            FROM new_trace
+            """;
+
+    // A whole-CTE prefix placeholder, filled at class init from a constant (OPIK_7727).
+    // ok: sql-query-with-format-slot
+    static final String PREFIX_SPLICE = """
+            %s
+            SELECT bucket
+            FROM spans_filtered
+            """;
+
+    // Correctly parameterized: nothing to flag.
+    // ok: sql-query-with-format-slot
+    static final String FULLY_BOUND = """
+            SELECT id
+            FROM spans
+            WHERE workspace_id = :workspace_id
+            AND project_id = :project_id
+            """;
+
+    // Not SQL: a log message that happens to contain the word "update".
+    // ok: sql-query-with-format-slot
+    static final String LOG_MESSAGE = "Completed bulk update for '%s' dataset items";
+
+    // Not SQL: an LLM prompt containing the jq expression `..|select(.error_info?)`.
+    // ok: sql-query-with-format-slot
+    static final String PROMPT_WITH_JQ = """
+            Use `%s` to inspect the trace.
+            Filter with `..|select(.error_info?)` to find errors.
+            Results come from the span tree, not from the top-level input.
+            """;
+
+    // Real SQL date format, not an integer slot — why the rule matches %s only, never %d.
+    // ok: sql-query-with-format-slot
+    static final String DATE_FORMAT_NOT_A_SLOT = """
+            SELECT value
+            FROM metadata
+            WHERE STR_TO_DATE(value, '%Y-%m-%d') < CURDATE()
+            """;
+}
--- a/.semgrep/java-sql-string-formatting.java
+++ b/.semgrep/java-sql-string-formatting.java
@@ -2,19 +2,22 @@
 //   semgrep test --config .semgrep/java-sql-string-formatting.yaml .semgrep/java-sql-string-formatting.java
 //
 // Semgrep's own annotations drive the assertions: a rule-id comment above a line marks it
-// as one the rule MUST flag, and the negative form marks one it must NOT.
+// as one that rule MUST flag, and the negative form marks one it must NOT.
 //
-// The must-not-flag cases below are the ones that actually bit. Every one of them matched
-// at some point during development, and the quoted-value and projection cases were caught
-// in PR review rather than by measurement. Keep them — they are the regression contract
-// for any future edit to the regex.
+// Two rules are asserted independently:
+//   sql-query-clause-splice — ERROR, blocks CI. Injection-shaped clause splices only.
+//   sql-query-format-slot   — WARNING, reports only. Any `%s` in a SQL literal.
+//
+// A clause splice is flagged by BOTH (it is also a format slot); everything else that
+// carries a slot is flagged by the reporting rule alone. Non-SQL strings are flagged by
+// neither — those cases are the ones that actually bit during development, so keep them.
 class JavaSqlStringFormattingFixtures {
 
     // ---------------------------------------------------------------------------
-    // MUST FLAG — a `%s` standing in for a clause, the injection-shaped pattern.
+    // BLOCKING — a `%s` standing in for a whole clause. Caller data reaches the query.
     // ---------------------------------------------------------------------------
 
-    // ruleid: sql-query-with-format-slot
+    // ruleid: sql-query-clause-splice,sql-query-format-slot
     static final String CLAUSE_SPLICE_AND = """
             SELECT id
             FROM spans
@@ -22,14 +25,14 @@ class JavaSqlStringFormattingFixtures {
             AND %s
             """;
 
-    // ruleid: sql-query-with-format-slot
+    // ruleid: sql-query-clause-splice,sql-query-format-slot
     static final String CLAUSE_SPLICE_WHERE = """
             SELECT name
             FROM system.columns
             WHERE %s
             """;
 
-    // ruleid: sql-query-with-format-slot
+    // ruleid: sql-query-clause-splice,sql-query-format-slot
     static final String CLAUSE_SPLICE_LOWERCASE = """
             select id
             from spans
@@ -37,66 +40,71 @@ class JavaSqlStringFormattingFixtures {
             and %s
             """;
 
-    // ruleid: sql-query-with-format-slot
-    static final String CLAUSE_SPLICE_NESTED_WITH = """
-            WITH x AS (
-            SELECT id
-            FROM spans
-            WHERE a = 1
-              AND %s
-            )
+    // Non-SELECT statements are in scope too: an UPDATE or DELETE predicate is exactly as
+    // injectable as a SELECT one, and a splice there mutates rather than reads.
+    // ruleid: sql-query-clause-splice,sql-query-format-slot
+    static final String CLAUSE_SPLICE_UPDATE = """
+            UPDATE spans
+            SET name = :name
+            WHERE workspace_id = :workspace_id
+            AND %s
+            """;
+
+    // ruleid: sql-query-clause-splice,sql-query-format-slot
+    static final String CLAUSE_SPLICE_DELETE = """
+            DELETE FROM spans
+            WHERE workspace_id = :workspace_id
+            AND %s
             """;
 
     // ---------------------------------------------------------------------------
-    // MUST NOT FLAG — value and projection positions, and non-SQL strings.
+    // REPORTED, NOT BLOCKING — the query text is still assembled by formatting, but the
+    // slot is not a clause. Prefer a bound parameter or a StringTemplate attribute anyway.
     // ---------------------------------------------------------------------------
 
-    // A slot inside a quoted SQL literal is a value position, not a clause. The ClickHouse
-    // health checks fill these from compile-time constants (OPIK_7727 tracks the cleanup).
-    // ok: sql-query-with-format-slot
+    // Value position. Filled from a compile-time constant in the ClickHouse health checks,
+    // so not injectable today — but the query text should still be a constant.
+    // ok: sql-query-clause-splice
+    // ruleid: sql-query-format-slot
     static final String VALUE_SLOT_TIGHT_QUOTES = """
             SELECT count() AS cnt
             FROM system.disks
             WHERE name = '%s'
             """;
 
-    // Same, with whitespace inside the quotes — the `(?<!')%s(?!')` form missed this.
-    // ok: sql-query-with-format-slot
+    // Same, with whitespace inside the quotes.
+    // ok: sql-query-clause-splice
+    // ruleid: sql-query-format-slot
     static final String VALUE_SLOT_SPACED_QUOTES = """
             SELECT count() AS cnt
             FROM system.clusters
             WHERE cluster = ' %s '
             """;
 
-    // A projection: the slot names a column, it does not splice a predicate. Real instances
-    // live in SpansLocalV2TableTest / NaNAwareAggregateIntegrationTest / TracesLocalV2CutoverTest.
-    // ok: sql-query-with-format-slot
+    // Projection: the slot names a column rather than splicing a predicate.
+    // ok: sql-query-clause-splice
+    // ruleid: sql-query-format-slot
     static final String PROJECTION_SLOT = """
             SELECT %s AS value
             FROM spans_local_v2
             WHERE workspace_id = :workspace_id
             """;
 
-    // ok: sql-query-with-format-slot
-    static final String PROJECTION_SLOT_IN_CTE = """
-            WITH new_trace AS (
-              SELECT toFixedString('%s', 36) AS id,
-              %s AS ttft
-            )
-            SELECT ttft
-            FROM new_trace
-            """;
-
-    // A whole-CTE prefix placeholder, filled at class init from a constant (OPIK_7727).
-    // ok: sql-query-with-format-slot
+    // A whole-CTE prefix placeholder, filled at class init from a constant. A StringTemplate
+    // attribute expresses this without formatting the query text.
+    // ok: sql-query-clause-splice
+    // ruleid: sql-query-format-slot
     static final String PREFIX_SPLICE = """
             %s
             SELECT bucket
             FROM spans_filtered
             """;
 
-    // Correctly parameterized: nothing to flag.
-    // ok: sql-query-with-format-slot
+    // ---------------------------------------------------------------------------
+    // FLAGGED BY NEITHER — correctly parameterized SQL, and strings that only look like it.
+    // ---------------------------------------------------------------------------
+
+    // ok: sql-query-clause-splice,sql-query-format-slot
     static final String FULLY_BOUND = """
             SELECT id
             FROM spans
@@ -105,19 +113,21 @@ class JavaSqlStringFormattingFixtures {
             """;
 
     // Not SQL: a log message that happens to contain the word "update".
-    // ok: sql-query-with-format-slot
+    // ok: sql-query-clause-splice,sql-query-format-slot
     static final String LOG_MESSAGE = "Completed bulk update for '%s' dataset items";
 
     // Not SQL: an LLM prompt containing the jq expression `..|select(.error_info?)`.
-    // ok: sql-query-with-format-slot
+    // Prompt templates carry their own injection risk, tracked separately in OPIK 7837 —
+    // it is not a SQL concern and this rule is not the right gate for it.
+    // ok: sql-query-clause-splice,sql-query-format-slot
     static final String PROMPT_WITH_JQ = """
             Use `%s` to inspect the trace.
             Filter with `..|select(.error_info?)` to find errors.
             Results come from the span tree, not from the top-level input.
             """;
 
-    // Real SQL date format, not an integer slot — why the rule matches %s only, never %d.
-    // ok: sql-query-with-format-slot
+    // Real SQL date format, not an integer slot — why both rules match %s only, never %d.
+    // ok: sql-query-clause-splice,sql-query-format-slot
     static final String DATE_FORMAT_NOT_A_SLOT = """
             SELECT value
             FROM metadata
--- a/.semgrep/java-sql-string-formatting.yaml
+++ b/.semgrep/java-sql-string-formatting.yaml
@@ -30,25 +30,52 @@ rules:
     #      `..|select(.error_info?)`, and log messages like "Completed bulk update for
     #      '%s' dataset items" match on the word "update".
     #   2. A FROM clause on its own line — a second, independent structural marker.
-    #   3. A `%s` that is NOT inside a single-quoted SQL string. A slot spliced into a
-    #      quoted literal (`WHERE cluster = '%s'`) is a value position, which the
-    #      health checks fill from a compile-time constant; those are a readability
-    #      concern, not injection, and are out of scope here (OPIK 7727). This gate
-    #      targets the clause-position splice the review actually flagged.
+    #   3. A `%s` in CLAUSE position: it must be the whole right-hand side of an
+    #      `AND` / `OR` / `WHERE` / `ON` / `HAVING` keyword, on a line containing no
+    #      single quote. This is what makes the rule injection-shaped rather than
+    #      merely "%s near SQL", and it excludes two large false-positive families:
+    #        - Value positions — `WHERE name = '%s'`, and the same with whitespace
+    #          inside the quotes (`= ' %s '`). The ClickHouse health checks fill these
+    #          from compile-time constants; a readability concern, not injection
+    #          (OPIK 7727).
+    #        - Projections — `SELECT %s AS value`, `%s AS ttft`. The slot names a
+    #          column, it does not splice a predicate. Real instances live in
+    #          SpansLocalV2TableTest, NaNAwareAggregateIntegrationTest and
+    #          TracesLocalV2CutoverTest; an earlier revision of this rule flagged all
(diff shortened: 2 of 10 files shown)


.semgrep/java-sql-string-formatting.yaml
```diff
@@ -0,0 +1,54 @@
...
+    # 1 and 2 are line-anchored. Dropping those anchors was measured and reintroduces
+    # the prompt-template false positive, which is the worse failure — it would block an
+    # unrelated commit and push a suppression comment into a file with no SQL in it.
+    # Every real query in this codebase is a multi-line text block, so the gap is
+    # theoretical; revisit if that stops being true.
+    patterns:
+      - pattern: $X
+      - metavariable-regex:
+          metavariable: $X
+          regex: (?sm)^"(?=.*^\s*(?i:SELECT|WITH)\b)(?=.*^\s*(?i:FROM)\b)(?=.*^[^\n]*\S[^\n]*(?<!')%s(?!')).*"$
```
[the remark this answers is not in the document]
JetoPistola: Fixed in dbca3ab810 — both false positives confirmed and fixed, and verifying them turned up something bigger.

Semgrep's built-in semgrepignore excludes `test`/`tests` directories at any depth, so passing `apps/opik-backend/src/test/java` as a **directory** scans zero files while still reporting "0 findings" — the test tree was never covered by any measurement in this PR. The pre-commit hook passes explicit **filenames**, which bypasses that exclusion. Measured the way the hook actually runs, the previous rule had **27 findings across three test files** and would have turned CI red the moment anyone touched them. So this was a live breakage, not a hypothetical — thanks for pulling the thread.

Both reported cases reproduce, and both are fixed by requiring the `%s` to sit in **clause position**: the whole right-hand side of `AND`/`OR`/`WHERE`/`ON`/`HAVING`, on a line with no single quote.

- `WHERE cluster = ' %s '` — no longer matches. Rather than making the exclusion quote-aware, the clause-position requirement removes the need for it: a value slot is never the entire RHS of a clause keyword, spaced quotes or not.
- `SELECT %s AS value` — no longer matches. Projections account for 26 of the 27 findings (`SpansLocalV2TableTest`, `NaNAwareAggregateIntegrationTest`, `TracesLocalV2CutoverTest`).

I went with clause-context rather than an enumerated keyword list plus a `SELECT ... AS ...` exclusion — same outcome for the cases you raised, but one positive condition instead of a growing set of carve-outs.

The 27th finding was a **true positive** your review indirectly surfaced: `TracesLocalV2CutoverTest#columnNames` spliced a caller-supplied predicate via `AND %s`. Converted to a bound `default_kind IN/NOT IN :default_kinds` with a StringTemplate conditional — same fix shape as `SpanMetricsQueries`.

Regression fixtures added as suggested, in `.semgrep/java-sql-string-formatting.java` — 13 assertions in semgrep's own annotation format (`semgrep test`), covering both of your cases plus every false-positive class that bit during development. The fixture sits outside the hook's `files:` scope, so its deliberate violations can't fail CI.

Result: **0 findings across all 1,687 backend Java files** (1,259 main + 428 test), verified the way the hook runs. The measurement trap is documented in the rule file so the next person doesn't repeat it.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanMetricsQueries.java
```diff
@@ -39,7 +40,8 @@ WITH feedback_scores_deduped AS (
                     FROM feedback_scores
                     WHERE entity_type = 'span'
                       AND workspace_id = :workspace_id
-                      AND %s
+                      <if(project_id)> AND project_id = :project_id<endif>
+                      <if(project_ids)> AND project_id IN :project_ids<endif>
```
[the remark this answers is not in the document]
JetoPistola: Fixed in dbca3ab810 — added, one test per converted template:

- `SpanCountTest#spanCountExcludesOtherProjectsInTheSameWorkspace`
- `SpanTokenUsageTest#spanTokenUsageExcludesOtherProjectsInTheSameWorkspace`

Both are `@ParameterizedTest` over the non-TOTAL intervals and seed a **second project in the same workspace**: different span counts per bucket for the count test, and distinct token-usage key names for the usage test. The distinct names matter — a dropped predicate would surface extra series rather than only inflating existing ones, and `getMetricsAndAssert` compares the full result set recursively, so either failure mode trips it.

Worth recording why this needed covering: an unset StringTemplate conditional renders as empty and only logs `attribute project_id isn't defined` — it does not throw. Verified directly at the template level: rendering `SPAN_FILTERED_PREFIX` with the flag set yields `project_id = :project_id` in all **3** CTE scopes; without it, **0** — every `WHERE` degrades to `workspace_id` alone and silently widens to the whole workspace.

Two caveats I'd rather state than paper over:

1. `spanCountExcludesOtherProjectsInTheSameWorkspace` passes locally across all 3 intervals (confirmed per-case in the surefire XML). `spanTokenUsageExcludesOtherProjectsInTheSameWorkspace` compiles and its surrounding class passes 17/17, but **that specific method has not executed locally** — Testcontainers failed to start ClickHouse on 4 of 7 attempts on this machine. Leaving it to CI.
2. The **negative** check — delete `template.add("project_id", true)` and confirm these tests fail — also couldn't be completed locally for the same reason. The failure mode is proven (above); what's unverified is that these two tests detect it. Both caveats are recorded in the PR description.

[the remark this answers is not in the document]
JetoPistola: Follow-up on the caveat above — CI has now run both tests, so the first of the two open items is closed.

`SpanCountTest` reports **18 tests** (was 15) and `SpanTokenUsageTest` reports **20** (was 17). The +3 each is the new `@ParameterizedTest` across the three non-TOTAL intervals, so `spanTokenUsageExcludesOtherProjectsInTheSameWorkspace` — the one that never got past container startup locally — did execute, and passed.

Still open, and I'd rather leave it visible than quietly drop it: the **negative** check (delete a `template.add("project_id", true)` line, confirm these tests go red) is unverified. The failure mode is proven at the template level — flag set yields the predicate in all 3 CTE scopes, unset yields 0 and nothing throws — but that these two tests *detect* it has not been demonstrated on a completed run.

Unrelated to this change, for the record: `ThreadDurationTest#happyPathWithFilter[7]` failed once in Integration Group 9 on this commit and passed on rerun with no code change. It reads `THREAD_FILTERED_PREFIX`, which no commit in this PR touches, and the delta between the passing and failing commits contains no production code at all.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/ProjectMetricsResourceTest.java
```diff
@@ -2744,6 +2744,51 @@ void emptyData(TimeInterval interval) {
...
+         * which is selected by a StringTemplate {@code <if(project_id)>} flag rather than spliced into the query text.
+         * An unset flag renders the conditional as empty and only logs a warning, so a render path that forgets to set
+         * it would silently drop the predicate and aggregate every project in the workspace.
+         * <p>
+         * Seeding a second project in the SAME workspace with a different span count is what makes that visible: with
+         * the predicate present the counts below are returned as-is, and without it every bucket would also include the
+         * other project's spans.
+         */
+        @ParameterizedTest
+        @EnumSource(value = TimeInterval.class, names = "TOTAL", mode = EnumSource.Mode.EXCLUDE)
```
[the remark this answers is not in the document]
JetoPistola: Skipping — I checked this against the code and the gap it describes isn't reachable.

**The span path.** `ProjectMetricsDAO#getMetric` sets the project flag *unconditionally, above* the `isTotal` branch:

```java
template.add("project_id", true);   // ← always runs
if (isTotal) { template.add("bucket", ...); }
else         { template.add("step", ...).add("bucket", ...).add("fill_from", ...); }
```

TOTAL and non-TOTAL share one predicate code path — TOTAL only changes `bucket` / `step` / `fill_from`, none of which touch `project_id`. So "a StringTemplate omission could remove the project predicate [under TOTAL]" describes a state the code can't reach: deleting that line breaks *every* interval, and the three the tests already parameterize over would go red.

**The token-usage path.** `getProjectTokenUsageNames` takes no interval parameter at all, and `TOKEN_USAGE_NAMES` contains no interval or bucket conditional. A "two-project token-name assertion for TOTAL" isn't an uncovered case — there's no TOTAL variant of that query to assert against.

**On the cited guideline.** `.agents/skills/opik-backend/testing.md` does have a "Parameterized Tests" section, but it's about consolidating duplicated test methods into one parameterized test, not about enumerating every enum constant. It doesn't recommend what's attributed to it here.

The `TimeInterval.TOTAL` exclusion in the `@EnumSource` is inherited from every sibling test in these classes (`happyPath`, `emptyData`, and the rest all use `names = "TOTAL", mode = EXCLUDE`), because TOTAL returns a single bucket that the shared `getMetricsAndAssert` three-bucket helper can't express. Including it would mean a differently-shaped assertion for no additional coverage of the predicate.

One thing this did surface that's worth recording: `TotalIntervalTest` covers only `TRACE_COUNT`, so TOTAL coverage across metric types is genuinely thin. That's pre-existing and unrelated to this PR's change — and notably it wouldn't be closed by the suggestion here either. Leaving it as-is rather than expanding this PR's scope.

.semgrep/java-sql-string-formatting.java
```diff
@@ -0,0 +1,126 @@
...
+            FROM new_trace
+            """;
+
+    // A whole-CTE prefix placeholder, filled at class init from a constant (OPIK_7727).
+    // ok: sql-query-with-format-slot
+    static final String PREFIX_SPLICE = """
+            %s
+            SELECT bucket
+            FROM spans_filtered
+            """;
```
andrescrz, line 101: These should be flagged as well, ST templates should be used instead.

.semgrep/java-sql-string-formatting.java
```diff
@@ -0,0 +1,126 @@
...
+    // ok: sql-query-with-format-slot
+    static final String LOG_MESSAGE = "Completed bulk update for '%s' dataset items";
+
+    // Not SQL: an LLM prompt containing the jq expression `..|select(.error_info?)`.
+    // ok: sql-query-with-format-slot
+    static final String PROMPT_WITH_JQ = """
+            Use `%s` to inspect the trace.
+            Filter with `..|select(.error_info?)` to find errors.
+            Results come from the span tree, not from the top-level input.
+            """;
```
andrescrz, line 127: This can be a source of prompt injection. We might need to think about it, but not as part of this PR. Requires user defined input to really happen. Maybe ST is safe alternative here as well.

.semgrep/java-sql-string-formatting.java
```diff
@@ -0,0 +1,126 @@
...
+            WHERE name = '%s'
+            """;
+
+    // Same, with whitespace inside the quotes — the `(?<!')%s(?!')` form missed this.
+    // ok: sql-query-with-format-slot
+    static final String VALUE_SLOT_SPACED_QUOTES = """
+            SELECT count() AS cnt
+            FROM system.clusters
+            WHERE cluster = ' %s '
+            """;
```
andrescrz, line 82: This should be flagged, bind should be used instead.

.semgrep/java-sql-string-formatting.yaml
```diff
@@ -0,0 +1,81 @@
...
+    #
+    # Also not caught: a spliced predicate that is not the entire right-hand side of a
+    # clause keyword (`AND %s AND deleted = 0`, or `AND (%s)`). Requirement 3 anchors the
+    # slot to end-of-line to keep projections out. Broadening it is possible but was not
+    # needed for any real or hypothetical case reviewed here.
+    patterns:
+      - pattern: $X
+      - metavariable-regex:
+          metavariable: $X
+          regex: (?sm)^"(?=.*^\s*(?i:SELECT|WITH)\b)(?=.*^\s*(?i:FROM)\b)(?=.*^[^\n']*\b(?i:AND|OR|WHERE|ON|HAVING)\s+%s\s*$).*"$
```
andrescrz: We need to cover any statement, not only SELECT. That means INSERT, UPDATE, DELETE etc. maybe even DDL, which should generally only exists in Liquibase SQL files which is safe, but in case added to the Java code. Probably DDL should be completely banned from the Java code.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanMetricsQueries.java
andrescrz: Minor: generally, better combine this condition only in the `IN` flavour. It works in both cases, just by passing a Set/List of the project_id.

Applies everywhere in the PR.

apps/opik-backend/src/main/java/com/comet/opik/domain/WorkspaceMetricsDAO.java
```diff
@@ -243,8 +243,7 @@ STEP toIntervalDay(1)
     // projects: WorkspaceMetricsService resolves the "all projects" request into every project id up front, so the
     // predicate is always a bounded `project_id IN :project_ids` list that prunes on the spans primary key
     // (workspace_id, project_id, ...) — never an unconstrained workspace-wide scan.
-    private static final String SPAN_FILTERED_PREFIX = SpanMetricsQueries
-            .spanFilteredPrefix("project_id IN :project_ids");
+    private static final String SPAN_FILTERED_PREFIX = SpanMetricsQueries.SPAN_FILTERED_PREFIX;
```
andrescrz: Nit: feel free to directly use the original constant directly (with a static import). The 2nd constant is simply redundant and doesn't add much value.

Applies everywhere in the PR.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/TracesLocalV2CutoverTest.java
andrescrz, line 1: Tests should be excluded from this rule. It's fine to use string formatting there. This should only apply to prod.

CONTRIBUTING.md
```diff
@@ -61,6 +61,16 @@ Workflows are also scanned for **security** issues with [zizmor](https://github.
 ## Dockerfiles
 Dockerfiles are linted with [hadolint](https://github.com/hadolint/hadolint), which runs as a hook in the unified `🐙 Code Quality` workflow (and locally via pre-commit) on changed Dockerfiles. It uses hadolint's default rule set; the handful of intentionally-suppressed rules are annotated inline in each Dockerfile with a `# hadolint ignore=` comment and a reason. The hook runs hadolint via its Docker image, so it needs only Docker — no manual install. To run it directly on a single file: `docker run --rm -i ghcr.io/hadolint/hadolint < path/to/Dockerfile`.
 
+## SQL query construction (Java backend)
```
andrescrz, line 64: This is a too level detail for the contribution guide. I wouldn't reflect it here. If CI fails, it's explicit already. I would target the claude/agents coding rules instead.

andrescrz (commented):
This is a nice first iteration.

The main thing to consider is that adding this will flag a lot of pre-existing code. We likely want to have nice way to roll it out, that flags new code but doesn't block existing one, only reports it. 

We'll fix the pre-existing issues as tech debt, only urgently if they suppose a security problem (like injection).

JetoPistola:
> @andrescrz wrote in their review (COMMENTED):
>
> This is a nice first iteration.
> 
> The main thing to consider is that adding this will flag a lot of pre-existing code. We likely want to have nice way to roll it out, that flags new code but doesn't block existing one, only reports it. 
> 
> We'll fix the pre-existing issues as tech d…

Fixed in b669c54d08 — you were right about the scale, and I measured it: widening as the inline comments asked (all statement types, all slot positions) surfaces **108 findings in production code**. The narrow rule only landed green by dodging them.

Split into two tiers:

| Rule | Severity | Matches | Prod findings |
|---|---|---|---|
| `sql-query-clause-splice` | **ERROR** — blocks CI | `%s` standing in for a whole clause, where caller data reaches the query text | **0** |
| `sql-query-format-slot` | **WARNING** — reports only | Any other `%s` in a SQL literal: value positions, projections, CTE prefixes | **108** |

The hook passes `--severity ERROR`, so only the blocking tier can fail a commit or a CI leg. Verified the rollout behaves as intended rather than assuming it: `pre-commit run semgrep --files ProjectMetricsDAO.java` **passes**, on a file carrying 99 reporting-tier findings.

That matches your split of the work — new injection-shaped code is blocked, and the pre-existing surface is visible as tech debt under OPIK_7727 rather than holding anyone up.

Both tiers now also cover INSERT/UPDATE/DELETE and DDL, not just SELECT/WITH, per the inline comment.

.semgrep/java-sql-string-formatting.java
JetoPistola: Fixed in b669c54d08 — now flagged by the reporting tier (`sql-query-format-slot`).

Rather than making it blocking, this and the value-position case sit in the WARNING tier: the query text is still assembled by formatting and should be a StringTemplate attribute, but no caller data reaches it, so it's the tech-debt category from your review body rather than an injection one. It shows up in the 108 reported findings under OPIK_7727.

Fixture updated to assert exactly that split — `ok:` for the blocking rule, `ruleid:` for the reporting one.

.semgrep/java-sql-string-formatting.java
JetoPistola: Fixed in b669c54d08 — same treatment: now flagged by the reporting tier, with `bind` being the right answer.

Kept it non-blocking for the same reason as the prefix splice — the ClickHouse health checks fill this slot from a compile-time constant, so it's a "the query text should be a constant" concern rather than an injection one. Both quote forms are asserted in the fixture, including `= ' %s '` with whitespace inside the quotes, which an earlier revision of the regex missed.

.semgrep/java-sql-string-formatting.java
JetoPistola: Agreed, and deferred as you suggested — filed as OPIK_7837 and linked to this ticket.

Scoped it to establish the exposure before proposing a fix, since your own caveat is the crux: it "requires user defined input to really happen", and whether that holds anywhere today is unverified. So the ticket asks for an inventory of prompt templates assembled by formatting, each splice classified by where its value comes from, and only then a decision on mitigation (StringTemplate, delimiting the untrusted span, or a separate message role).

Deliberately kept out of these rules: they match SQL literals, and a prompt template isn't SQL — gating this would be a differently-shaped rule. The fixture keeps the prompt case as must-not-flag, with a comment pointing at the ticket so it isn't mistaken for an oversight.

.semgrep/java-sql-string-formatting.yaml
JetoPistola: Fixed in b669c54d08 — both tiers now match `INSERT`, `UPDATE`, `DELETE`, `CREATE`, `ALTER`, `DROP` and `TRUNCATE` alongside `SELECT`/`WITH`.

Agreed on the reasoning: a spliced predicate in an `UPDATE` or `DELETE` is exactly as injectable as one in a `SELECT`, and it mutates rather than reads. Added fixtures for the `UPDATE ... WHERE ... AND %s` and `DELETE FROM ... AND %s` forms so that stays covered.

On banning DDL from Java entirely — I didn't take that on here, since it's a different rule (presence of a statement type, not the shape of a slot) and would want its own measurement against the current tree first. Happy to file it if you want it tracked; it seemed like a separate decision from this gate.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanMetricsQueries.java
JetoPistola: Fixed in b669c54d08 — collapsed to the `IN` flavour everywhere. Both callers now bind `project_ids`: a set of one for per-project aggregation, a resolved set for workspace-level. The conditional is gone entirely, so the query text is a plain constant rather than a template with two branches.

Worth flagging what this surfaced, because it would have failed CI. `getMetric` renders ~37 queries, but only the 11 `GET_SPAN_*` ones embed the shared CTE where `:project_ids` appears — the trace and thread prefixes use the scalar `:project_id`. Binding `project_ids` unconditionally raises `NoSuchElementException` from R2DBC on every one of those, which is the same failure OPIK_5678 hit in this exact method (its comment is still there a few lines up). So the bind is guarded by `SPAN_TIME_METRICS`, which I verified maps exactly onto those 11 constants.

Compilation doesn't catch it — R2DBC resolves binds by name at execution — and neither would a span-metrics test, since those are precisely the queries that do carry the placeholder.

apps/opik-backend/src/main/java/com/comet/opik/domain/WorkspaceMetricsDAO.java
JetoPistola: Fixed in b669c54d08 — dropped all four redundant constants across both DAOs in favour of static imports of `SPAN_FILTERED_PREFIX` and `TOKEN_USAGE_NAMES`.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/TracesLocalV2CutoverTest.java
JetoPistola: Fixed in b669c54d08 — the hook's `files:` regex is now scoped to `^apps/opik-backend/src/main/java/`, so test sources are out of the rule entirely.

Your call as code owner, and I've reverted this file with it: the conversion here only existed *because* tests were in scope, so it goes away with that decision. `TracesLocalV2CutoverTest` is back to its original form on `main`.

Verified the narrowed scope routes correctly — a changed test file now emits only the spotless leg, while a changed production file still emits semgrep.

For the record on the earlier reasoning: including tests is what surfaced this splice in the first place, and I'd argued fixtures teach patterns. Your framing is the better one though — string formatting in a fixture is building a fixture, not accepting user input, so the injection premise doesn't hold there.

CONTRIBUTING.md
JetoPistola: Fixed in b669c54d08 — cut down to a two-line pointer.

You're right that it was duplication: checking `.agents/rules/security.mdc` after your comment, the rule is already there ("Build SQL with string operations" under Never, "Express the variable parts of a query as template conditionals" under Always), pointing at the backend skill for detail. So the contribution guide was restating conventions that already had a home.

What's left in `CONTRIBUTING.md` is just where the rules live and where the conventions are documented. The scoping rationale — why `%d` is excluded, how to measure coverage correctly — stays in comments in the rule file, next to the regex it explains.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanMetricsQueries.java
```diff
@@ -15,9 +15,10 @@ final class SpanMetricsQueries {
     private SpanMetricsQueries() {
     }
 
-    // %s slots are the project predicate: "project_id = :project_id" (single project) or
-    // "project_id IN :project_ids" (a resolved set of projects). workspace_id is always bound separately.
-    private static final String SPAN_FILTERED_PREFIX_TEMPLATE = """
+    // The project predicate is bound, not spliced: both callers bind `project_ids`, a per-project set of one
+    // ({@link ProjectMetricsDAO}) or a resolved set of many ({@link WorkspaceMetricsDAO}). The `IN` form covers
+    // both, so the query text stays a constant with no conditional. workspace_id is always bound separately.
```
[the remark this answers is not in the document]
JetoPistola: Skipping here — deferring to OPIK_7727, which already tracks this category. Reasoning, with the numbers:

**Scale.** It isn't the one declaration in `SpanMetricsQueries` — the seam is 35 call sites: 14 `.formatted(TRACE_FILTERED_PREFIX)`, 13 `.formatted(SPAN_FILTERED_PREFIX)`, 8 `.formatted(THREAD_FILTERED_PREFIX)` across the two metrics DAOs.

**No injection exposure.** All 35 are `static final String` constants filled from other `static final String` constants, assembled once at class initialization. No caller-supplied value reaches any of them, which is why they sit in the reporting tier rather than the blocking one — they're part of the 108 WARNING findings, so they're surfaced rather than invisible.

That matches how the human review framed this category: fix pre-existing issues as tech debt, urgently only where they're an actual security problem. This isn't one.

**Why not in this PR.** The conversion means those 35 constants stop being constants — each becomes a per-request `TemplateUtils.newST(...)` render with the prefix as an attribute. That touches every render path in both DAOs, including `getMetric`, whose bind contract turned CI red twice today: collapsing the project predicate to `IN :project_ids` left 11 span queries no longer declaring `:project_id`, and R2DBC raises `NoSuchElementException` for any bind the rendered SQL doesn't declare. Moving 35 more queries onto a new rendering path, with the ST-attribute/bind seam in play, is a meaningful regression risk for a readability fix — and it also shifts work from class-init to request time on a hot metrics path.

The suggestion itself is right, and I've written it into OPIK_7727 with the sizing and the per-request-cost question so it isn't lost.

andrescrz (approved):
LGTM

PR #7774, opened 2026-08-06
[BE]: register hyperbolic as a canonical provider so Hyperbolic model prices load

Anuj7411:
## Details

`CostService.PROVIDERS_MAPPING` only registers a subset of the `litellm_provider` values present in `model_prices_and_context_window.json`. Any model whose `litellm_provider` is not in that map has its price dropped at load time (`buildModelPrice` returns null), so its spans are billed at `DEFAULT_COST` (zero) and cost tracking / the per-evaluation spend budget silently under-report.

`hyperbolic` is one such provider: the bundled price file carries 16 non-zero-cost Hyperbolic entries (Qwen, DeepSeek, Llama and more served on Hyperbolic) that currently never load. This registers `hyperbolic` as a canonical provider so those prices resolve. No Hyperbolic model publishes cache rates today, so requests route through the standard `textGenerationCost` path; no cache calculator entry is needed.

Part of the ongoing provider-coverage cleanup (#7697).

## Testing

Added `calculateCostHandlesHyperbolicModels` asserting a representative Hyperbolic model prices to a non-zero, exact expected cost (it returns zero before this change).

## Documentation

No documentation changes needed

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
@@ -53,7 +53,8 @@ public class CostService {
             Map.entry("meta", "meta"),
             Map.entry("zai", "zai"),
             Map.entry("z-ai", "zai"),
-            Map.entry("sambanova", "sambanova"));
+            Map.entry("sambanova", "sambanova"),
+            Map.entry("hyperbolic", "hyperbolic"));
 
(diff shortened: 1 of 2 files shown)


Anuj7411:
Consolidated into #7832, which registers all five OpenAI-compatible providers in one PR to avoid the mutual PROVIDERS_MAPPING merge conflicts. Closing in favor of that.

PR #7775, opened 2026-08-06
[BE]: register baseten as a canonical provider so Baseten model prices load

Anuj7411:
## Details

`CostService.PROVIDERS_MAPPING` only registers a subset of the `litellm_provider` values in `model_prices_and_context_window.json`. Models whose provider is missing from that map have their price dropped at load time (`buildModelPrice` returns null), so their spans bill at `DEFAULT_COST` (zero) and cost tracking / the per-evaluation spend budget silently under-report.

`baseten` is one such provider: the bundled price file carries 11 non-zero-cost Baseten entries (GLM, DeepSeek, Llama and more served on Baseten Model APIs) that never load today. This registers `baseten` as a canonical provider so those prices resolve. No Baseten model publishes cache rates today, so requests route through the standard `textGenerationCost` path; no cache calculator entry is needed.

Part of the ongoing provider-coverage cleanup (#7697).

## Testing

Added `calculateCostHandlesBasetenModels` asserting a representative Baseten model prices to a non-zero, exact expected cost (it returns zero before this change).

## Documentation

No documentation changes needed

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
@@ -53,7 +53,8 @@ public class CostService {
             Map.entry("meta", "meta"),
             Map.entry("zai", "zai"),
             Map.entry("z-ai", "zai"),
-            Map.entry("sambanova", "sambanova"));
+            Map.entry("sambanova", "sambanova"),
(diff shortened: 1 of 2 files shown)


Anuj7411:
Consolidated into #7832, which registers all five OpenAI-compatible providers in one PR to avoid the mutual PROVIDERS_MAPPING merge conflicts. Closing in favor of that.

PR #5837, opened 2026-03-25
[issue-5835] [BE] fix: add IF NOT EXISTS to non-idempotent ClickHouse migrations

rusherman:
## Details

Add `IF NOT EXISTS` guards to `ADD COLUMN` statements in ClickHouse migrations
000001, 000003, and 000014 to ensure idempotent execution. These three early
migrations are the only ones missing this guard — all subsequent migrations
(000004+) already follow the `IF NOT EXISTS` convention.

## Change checklist

- [ ] User facing
- [ ] Documentation update

## Issues

Resolves #5835

AI-WATERMARK: yes
- **Tools**: Claude Code (CLI)
- **Model(s)**: Claude Opus 4.6
- **Scope**: Added `IF NOT EXISTS` to 14 `ADD COLUMN` statements across 3 migration files
- **Human verification**: All changes reviewed by human operator

## Testing

**Commands run:**
```bash
# Verify all ADD COLUMN now have IF NOT EXISTS
grep -n 'ADD COLUMN' apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000001_init_script.sql
grep -n 'ADD COLUMN' apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000003_add_metadata_to_experiments.sql
grep -n 'ADD COLUMN' apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000014_add_thread_id_to_traces.sql
```

**Scenarios validated:**
- Fresh migration run: all tables/columns created successfully
- Replayed migration on pre-existing schema: no errors (idempotent)

**Environment:** macOS, JDK 21, ClickHouse 24.8 (Alibaba Cloud managed)

## Documentation

No documentation update needed.

PR #7795, opened 2026-08-09
fix(migrations): make early ClickHouse ADD COLUMN statements idempotent

### LINKED ISSUE #5835: [Bug]: Non-idempotent ADD COLUMN in early ClickHouse migrations (000001, 000003, 000014)
### Component

- [x] Server

### Opik Version

1.10.45

### Describe the problem

**Expected behavior:** All ClickHouse migration DDL statements should be idempotent — safe to replay without errors.

**Actual behavior:** Three early migration files use `ADD COLUMN` without `IF NOT EXISTS`, causing failures when migrations are replayed on a pre-existing schema:

| Migration | Statement | Impact |
|-----------|-----------|--------|
| `000001_init_script.sql` | 12x `ADD COLUMN` (created_by, last_updated_by on 6 tables) | `DB::Exception: column already exists` |
| `000003_add_metadata_to_experiments.sql` | `ADD COLUMN metadata` | `DB::Exception: column already exists` |
| `000014_add_thread_id_to_traces.sql` | `ADD COLUMN thread_id` | `DB::Exception: column already exists` |

All subsequent migrations (000004+) correctly use `ADD COLUMN IF NOT EXISTS`.

### Steps to reproduce

1. Run Liquibase `dbAnalytics migrate` once (creates all tables/columns)
2. Clear the `DATABASECHANGELOG` table (or lose it during recovery)
3. Run `dbAnalytics migrate` again
4. Migrations 000001, 000003, 000014 fail with `column already exists`

This scenario occurs in practice when:
- Deploying on managed ClickHouse where schema was pre-created manually
- Disaster recovery scenarios where changelog tracking is lost
- Development environments where the changelog is reset

### Proposed solution

Add `IF NOT EXISTS` to the `ADD COLUMN` statements in these three files, consistent with the convention used in all other migrations.

Xsidz:
## Summary

Adds `IF NOT EXISTS` to 15 `ADD COLUMN` statements across 3 early migration files that were written before the pattern was adopted. Replaying these migrations on a pre-existing schema previously threw `DB::Exception: column already exists`.

| File | Statements fixed |
|------|-----------------|
| `000001_init_script.sql` | 12 (`created_by` + `last_updated_by` on 6 tables) |
| `000003_add_metadata_to_experiments.sql` | 1 (`metadata` on experiments) |
| `000014_add_thread_id_to_traces.sql` | 1 (`thread_id` on traces) |

All migrations from `000004` onwards already use `IF NOT EXISTS`. This backfills the three exceptions.

Rollback lines are unchanged. On a fresh install the behavior is identical; on replay the statements become safe no-ops instead of errors.

Also adds `scripts/check_clickhouse_migrations_idempotent.sh` wired into the existing migration prefix check CI workflow to prevent regression.

## Validation

- Diff verified: only `IF NOT EXISTS` inserted — column names, types, defaults, rollback lines all unchanged
- Semantic simulation: fresh install → identical result; replay → no-op instead of error
- CI script tested: passes on patched tree, exits 1 with violation list on unpatched tree
- 000004+ migrations confirmed untouched

Fixes #5835

### DIFF
--- a/.github/workflows/migration_prefix_check.yml
+++ b/.github/workflows/migration_prefix_check.yml
@@ -6,9 +6,10 @@ permissions:
 
 on:
   pull_request:
-    paths: 
+    paths:
       - "apps/opik-backend/src/main/resources/liquibase/db-app-state/migrations/**"
       - "apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/**"
+      - "scripts/**"
   workflow_dispatch:
 
 
--- a/.github/workflows/migration_prefix_check.yml
+++ b/.github/workflows/migration_prefix_check.yml
@@ -29,3 +29,7 @@ jobs:
       - name: Check for migration prefix conflicts
         shell: bash
         run: ./scripts/check_backend_migration_conflicts.sh
+
+      - name: Check ADD COLUMN idempotency in analytics migrations
+        shell: bash
+        run: ./scripts/check_clickhouse_migrations_idempotent.sh
--- a/apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000001_init_script.sql
+++ b/apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000001_init_script.sql
@@ -98,28 +98,28 @@ CREATE TABLE IF NOT EXISTS ${ANALYTICS_DB_DATABASE_NAME}.experiment_items
     ORDER BY (workspace_id, experiment_id, dataset_item_id, trace_id, id);
 
 ALTER TABLE ${ANALYTICS_DB_DATABASE_NAME}.experiment_items
-    ADD COLUMN created_by String DEFAULT '',
-    ADD COLUMN last_updated_by String DEFAULT '';
+    ADD COLUMN IF NOT EXISTS created_by String DEFAULT '',
+    ADD COLUMN IF NOT EXISTS last_updated_by String DEFAULT '';
 
 ALTER TABLE ${ANALYTICS_DB_DATABASE_NAME}.spans
-    ADD COLUMN created_by String DEFAULT '',
-    ADD COLUMN last_updated_by String DEFAULT '';
+    ADD COLUMN IF NOT EXISTS created_by String DEFAULT '',
+    ADD COLUMN IF NOT EXISTS last_updated_by String DEFAULT '';
 
 ALTER TABLE ${ANALYTICS_DB_DATABASE_NAME}.traces
-    ADD COLUMN created_by String DEFAULT '',
-    ADD COLUMN last_updated_by String DEFAULT '';
+    ADD COLUMN IF NOT EXISTS created_by String DEFAULT '',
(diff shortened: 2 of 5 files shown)


Xsidz:
Fixed in 5b2416566e:

1. **grep filter bug**: rewrote the check to parse grep -n output line-by-line, extracting the content portion (after the second colon) before applying the comment exclusion. The old pattern-based filter didn't account for the path:linenum: prefix.

2. **Multi-clause ALTER gap**: the per-line parsing now checks each ADD COLUMN occurrence independently, so a mixed-clause ALTER can't hide an unsafe clause behind a safe one.

3. **scripts/** in CI trigger paths**: added `scripts/**` to the workflow's pull_request paths filter so checker changes always run through CI.

4. **Rollback ownership note**: added a comment above the rollback lines explaining that they're intentionally unconditional — production deployments run migrations once and don't roll back, so column ownership at rollback time is unknowable.

Xsidz:
Fixed in 58a5ab9e04: switched to grep -ni with [[:space:]]+ so ADD COLUMN variants are matched regardless of case or extra whitespace. IF NOT EXISTS is now checked via case-insensitive string comparison (tr '[:upper:]' '[:lower:]') rather than exact-match grep -v, so lowercase and mixed-case forms are also guarded. Verified with a lowercase add column fixture that the script exits 1 correctly.

Xsidz:
Fixed in 14335a8662:

**T1 (block comments):** Added explicit skip for lines matching block comment patterns (`/\*`, `\* `, `\*/`) so documented examples in comments don't trigger false positives.

**T2 (type compatibility):** This is intentionally out of scope for this script. The check guards against `DB::Exception: column already exists` on replay — a structural crash. Schema compatibility (divergent column types) is a separate concern that would require querying `system.columns` against a running ClickHouse instance, which a static CI lint cannot do. Documented in the issue.

**T3 (|| true masks errors):** Replaced with explicit exit code handling: `grep` exit 1 (no matches) = trivially OK; exit 2+ = loud failure with error output. Also added an upfront `if [ ! -d "$MIGRATIONS_DIR" ]` guard that exits 2 so a missing directory is caught immediately rather than silently succeeding.

Xsidz:
Fixed in 75f496e8a4: wrapped grep in an if-context (`if grep_output=$(grep ...); then grep_rc=0; else grep_rc=$?; fi`) so `set -e` doesn't abort on exit code 1 (no matches). The no-match branch now runs as intended.

Nimrod007:
Hi @Xsidz,

First of all, thank you very much for your contribution — we really appreciate the time and effort you put into it, and the writeup was thorough.

We're going to close this PR, as Liquibase migrations cannot be modified once they've been publicly released. Changing them would alter the checksum (hash), which would break existing installations that have already applied the migration — they'd fail validation on the next startup. This is the same reason we closed #5837, which proposed the same change to these three files.

In general, this situation shouldn't occur under normal conditions, since once a migration is executed, Liquibase tracks it and won't reapply it.

Thanks again for your contribution — much appreciated!

Best regards

PR #4383, opened 2025-12-08
[OPIK-3392] add logging capabilities to LLM providers

idoberko2:
## Details
To better debug playground interactions, this PR adds debug logging capabilities to the LLM proxy.

Since the playground uses streaming, we accumulate the chunks until the response is finished and only then write to the log, to avoid overwhelming the logs with many partial chunks.

At this point, Vertex-AI is not enhanced with this logic.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues
- OPIK-3392

## Testing
Tested the following manually:
1. OpenAI, Anthropic, Gemini and custom provider when no flag is on (`LLM_PROXY_REQUEST_LOG_LEVEL=INFO`) - logs are not changed
2. OpenAI, Anthropic, Gemini and custom provider when the flag is ON (`LLM_PROXY_REQUEST_LOG_LEVEL=DEBUG`) - LLM request and response are shown in the logs

## Documentation
No need

PR #7796, opened 2026-08-09
[OPIK-7847] [BE] Own and close the Vertex AI client per request to fix the GAX thread leak

YarivHashaiComet:
## Details

opik-backend built a fresh `VertexAI` SDK client on every LLM-as-judge call and never closed it. Each client owns a Google GAX `ScheduledThreadPoolExecutor` whose core threads are non-daemon and never expire, so the pool stays GC-rooted — every online-scoring / evaluator call on a Vertex AI model leaked a thread pool, and scoring pods climbed to ~15k `Gax-*` threads and thread-pinned until a rolling restart.

The root cause is ownership: nothing closed the per-call client. The langchain4j model can't — the two-arg `VertexAiGeminiChatModel(GenerativeModel, GenerationConfig)` constructor nulls its `VertexAI` handle, so its `close()` is a no-op.

Fix: build a fresh `VertexAI` per request and give it an owner that actually closes it.

- `CloseableVertexAiChatModel` / `CloseableVertexAiStreamingChatModel` wrap the langchain4j model, hold the `VertexAI` we built, and close it in `close()` — releasing the GAX threads.
- Close at the three request-scoped sites: `LlmProviderVertexAI.generate` (try-with-resources), `generateStream` (once, on the stream's terminal; build failures notify the caller), `ChatCompletionService.scoreTrace` (in a `finally`; other providers' models aren't `AutoCloseable`, so they're left to self-reclaim).
- No cache: no shared key, no process-lifetime credential residency, no eviction.

Only `VERTEX_AI` needed this. The other providers (OpenAI, Anthropic, Gemini, custom, free) also build a fresh model per call but over self-reclaiming clients (JDK `HttpClient` / OkHttp — daemon threads that idle-time-out), so none leak.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- OPIK-7847

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.8
- Scope: root-cause investigation (JVM thread dumps + local reproduction), the fix, a cross-provider audit, and the unit tests. Human-directed; the design (per-request ownership over caching) was decided by the author with reviewer input.
- Human verification: author reviewed the diff and tests, reproduced the leak locally, and validated the before/after on a running backend.

## Testing

Exact command (JDK 25):
- `mvn test -Dtest=VertexAIClientGeneratorTest,ChatCompletionServiceTest,LlmProviderVertexAITest` → **25/25**.

Scenarios validated:
- Ownership — closing the returned chat and streaming wrappers shuts down the underlying `VertexAI` (`isShutdown()` flips false→true).
- Streaming build failure notifies the caller (`handleError` + `handleClose`) instead of hanging.
- Multi-region SDK-host derivation unchanged.
- Standalone gRPC probe — 25 build→use cycles: **with `close()` +0** live `Gax-*` threads, **without `close()` +25**.
- Dev-runner (running backend) — 20 concurrent Vertex calls peaked at 20 Gax threads and settled back to 0; 30 sequential non-streaming + 30 streaming stayed flat; zero close-failure warnings.

Not run: the full `opik-backend` suite — the change is isolated to the Vertex path.

## Documentation

No config or docs changes — the fix is internal to the Vertex client generator and its call sites.


apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/vertexai/VertexAIClientGenerator.java
```diff
@@ -113,31 +126,41 @@ private Optional<String> apiEndpointFor(String canonicalLocation) {
...
                     .build();
-
-        } catch (Exception e) {
+        } catch (IOException e) {
             throw failWithError(e);
         }
     }
 
+    private static String credentialsDigest(String apiKey) {
+        return Hashing.sha256().hashString(apiKey, StandardCharsets.UTF_8).toString();
```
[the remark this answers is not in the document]
YarivHashaiComet: **Won't fix — intentional trade-off.**

The SHA-256 is a few microseconds over a ~2–3 KB payload, on the path to a Vertex call that takes hundreds of ms to seconds — and it's smaller than the per-call `decrypt()` of the same credential already done upstream in `LlmProviderFactoryImpl.buildConfig`. So it isn't a meaningful hot path.

The cache already retains only the digest, never the raw key. Memoizing the digest *per credential* (to skip the recompute) would require a `Map<rawApiKey, digest>` holding the plaintext service-account JSON in memory — which is exactly what keying by the digest is designed to avoid. The digest is also content-addressed, so it self-invalidates on key rotation (new key → new digest → new client).

Net: it would trade a negligible CPU saving for retaining the plaintext credential, so leaving as-is.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/vertexai/VertexAIClientGenerator.java
```diff
@@ -9,28 +9,41 @@
...
 import java.util.Objects;
 import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
 
-@RequiredArgsConstructor
 @Slf4j
 public class VertexAIClientGenerator implements LlmProviderClientGenerator<ChatModel> {
 
+    // One VertexAI per (credentials, location), reused for the process lifetime: building one per call leaked its
+    // non-daemon gRPC threads. No eviction, so a cached client is never closed while a request is still using it.
```
thiagohora: The symptom is right, but I think the cause is one level down: nothing *owns* the client, so nothing ever closes it.

Both constructors used here — `:56` and `:77` — delegate to a form that does `this.vertexAI = null`, and `close()` is `if (vertexAI != null) vertexAI.close()`. I read the field back on models built exactly as this generator builds them: `null` on both. So `close()` is a guaranteed no-op on this path — measured indistinguishable from not closing at all. That's langchain4j's ownership convention: the builder constructors create the `VertexAI` and close it, the `GenerativeModel` constructor receives one and disclaims it.

Fixing ownership instead of lifetime also fixes the leak: a small Opik-side type holding the `VertexAI` this generator built and overriding `close()`, generator back to its per-request form, try-with-resources at the three request-scoped call sites (`LlmProviderVertexAI.generate`, `generateStream`, `ChatCompletionService.scoreTrace`). Measured 0 channels / 0 threads for both chat and streaming, against a control at 25/25. All 11 pre-existing tests here pass, `ChatCompletionServiceTest` 11/11; only the two `reusesOneClientForEquivalentLocations` cases fail, and those exist to assert the cache.

That keeps the full `GenerationConfig` plus configured transport and scope, which the langchain4j builders can't express, and stays on the same deprecated constructor as today.

Is there a latency requirement behind caching that I'm not seeing? It avoids a token exchange and a TLS handshake per completion — real, but I haven't measured them.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/vertexai/VertexAIClientGenerator.java
```diff
@@ -9,28 +9,41 @@
...
     private final @NonNull LlmProviderClientConfig clientConfig;
+    private final Map<ClientKey, VertexAI> clients = new ConcurrentHashMap<>();
```
thiagohora: Worth checking this against the 8 production instances.

Without session affinity a workspace's calls scatter, so within a token lifetime a workspace making R calls cold-builds on roughly `8 × (1 − (7/8)^R)` instances — ~34% hit rate at 8 calls/hour, ~69% at 25, ~92% at 100. Below roughly 25 calls per workspace per hour, most requests still pay the token exchange and handshake anyway.

Retention doesn't dilute the same way. Without affinity every instance eventually serves every active workspace, so each JVM converges on the *full* set of Vertex-configured workspaces rather than an eighth — at ~1 channel and 1–2 threads per entry, no eviction, and key rotations adding permanently.

So the hit rate is divided by the fleet while retention is multiplied by it. How many workspaces have Vertex AI configured? That decides whether unbounded is fine or needs a bound before this ships.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/vertexai/VertexAIClientGenerator.java
```diff
@@ -9,28 +9,41 @@
...
+
+    public VertexAIClientGenerator(@NonNull LlmProviderClientConfig clientConfig) {
+        this.clientConfig = clientConfig;
+    }
+
+    // credentialsDigest, not the raw key, so the map never retains the service-account secret.
```
thiagohora: Only the *key* is hashed — the cached `VertexAI` value holds `ServiceAccountCredentials` built from the raw JSON, so the RSA private key and any cached token stay resident for the process lifetime. After an admin rotates or revokes a key the old one survives in heap (and in any heap dump) on every instance that served that workspace, with no path to evict it short of a restart. Suggest re-wording, since as written it reads as a stronger guarantee than the code provides.

thiagohora (commented):
Measured this before commenting, since the caching approach and the obvious alternative both look plausible on paper and only one survives contact with the runtime.

Probe on the **GRPC** transport — the one that leaks; the WireMock suite is REST, which is why it can't see this. 25 sequential clients, live channels via `InternalChannelz`, no server needed since gRPC connects lazily.

| variant | live channels | GAX threads (t+0 → t+15s) |
|---|---|---|
| never close (main today) | 25 | 25 → 50 |
| `ChatModel.close()` | 25 | 25 → 50 |
| close the underlying `VertexAI` | 0 | 0 |
| fix ownership + close (see inline) | 0 | 0 |
| control: same, not closed | 25 | 25 |

Thread count climbs *after* traffic stops, as each per-client executor ramps its pool — so it compounds on its own. Heap stays flat, so the exhaustion mode is threads and channels, not memory.

Holding the error-handling findings for a separate pass; this one is just the leak.

thiagohora:
**Context: this module is being superseded, and the replacement doesn't have this class of problem.**

`langchain4j-vertex-ai-gemini` sits on `com.google.cloud.vertexai`, whose `generativeai` package is deprecated by Google — langchain4j's own issue [#4383](https://github.com/langchain4j/langchain4j/issues/4383) describes the SDK as stopping in June 2026, and it was closed as completed on 2026-05-15 by shipping `langchain4j-google-genai` on the Google Gen AI SDK.

I ran the same probe against the new SDK, issuing real requests through a local HTTP stub:

| | Vertex AI SDK | Gen AI SDK |
|---|---|---|
| 25 clients, never closed | 25 channels + 25 → 50 threads | +1 thread total, ~2 MB |
| 25 clients, each closed | unchanged — `close()` is a no-op here | +0 |
| 1 shared client, 25 requests | n/a | +0 |

The Gen AI SDK is HTTP (OkHttp) rather than gRPC, and OkHttp's `TaskRunner` pool is shared and bounded — so per-request clients don't accumulate. Growth is flat instead of linear. `com.google.genai.Client` is also properly `AutoCloseable` and injectable via `client(Client)`, so there's no ambiguity about who owns it.

The new builder also closes the config gap that pins us to the deprecated constructor here: `stopSequences`, `presencePenalty`, `frequencyPenalty` and `seed` are all first-class, plus a `generateContentConfigCustomizer` escape hatch for anything else.

The one thing caching still buys on the new SDK is the OAuth token — per-request clients made 50 HTTP calls for 25 completions (25 token fetches + 25 generations) against 26 for a shared client. That's addressable by caching credentials rather than clients.

I'll open a ticket to migrate to `langchain4j-google-genai`. Not a blocker for this PR — flagging it so we don't over-invest in the caching design here.

YarivHashaiComet:
> **Context: this module is being superseded, and the replacement doesn't have this class of problem.**
> 
> `langchain4j-vertex-ai-gemini` sits on `com.google.cloud.vertexai`, whose `generativeai` package is deprecated by Google — langchain4j's own issue [#4383](https://github.com/langchain4j/langchain4j/issues/4383) describes the SDK as stopping in June 2026, and it was closed as completed on 2026-05-15 by shipping `langchain4j-google-genai` on the Google Gen AI SDK.
> 
> I ran the same probe against the new SDK, issuing real requests through a local HTTP stub:
> 
> Vertex AI SDK	Gen AI SDK
> 25 clients, never closed	25 channels + 25 → 50 threads	+1 thread total, ~2 MB
> 25 clients, each closed	unchanged — `close()` is a no-op here	+0
> 1 shared client, 25 requests	n/a	+0
> The Gen AI SDK is HTTP (OkHttp) rather than gRPC, and OkHttp's `TaskRunner` pool is shared and bounded — so per-request clients don't accumulate. Growth is flat instead of linear. `com.google.genai.Client` is also properly `AutoCloseable` and injectable via `client(Client)`, so there's no ambiguity about who owns it.
> 
> The new builder also closes the config gap that pins us to the deprecated constructor here: `stopSequences`, `presencePenalty`, `frequencyPenalty` and `seed` are all first-class, plus a `generateContentConfigCustomizer` escape hatch for anything else.
> 
> The one thing caching still buys on the new SDK is the OAuth token — per-request clients made 50 HTTP calls for 25 completions (25 token fetches + 25 generations) against 26 for a shared client. That's addressable by caching credentials rather than clients.
> 
> I'll open a ticket to migrate to `langchain4j-google-genai`. Not a blocker for this PR — flagging it so we don't over-invest in the caching design here.

cool - lets take this one as separated task

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/vertexai/LlmProviderVertexAI.java
```diff
@@ -41,18 +44,68 @@ public void generateStream(@NonNull ChatCompletionRequest request, @NonNull Stri
...
+                    }
+
+                    // Release the client's GAX threads exactly once, once the stream terminates
+                    // (onComplete/onError) — never when this task returns, which is before the first token.
+                    var closed = new AtomicBoolean(false);
+                    Runnable closeOnce = () -> {
+                        if (closed.compareAndSet(false, true)) {
+                            client.close();
+                        }
+                    };
```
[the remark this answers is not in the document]
YarivHashaiComet: Good catch — confirmed. On disconnect Jersey marks the `ChunkedOutput` closed, so `ChunkedOutputHandlers.handleMessage` hits its `isClosed()` guard and returns without cancelling the upstream stream — a Vertex stream that then never fires a terminal keeps its client (and its GAX pool).

Scoping it out of this PR, though: it's the pre-existing no-timeout / no-cancellation gap, not something the ownership-close change introduced. Before this PR *every* Vertex streaming call leaked its client; after it, only a stream that never terminates does — strictly better. The proper fix (a configurable Vertex call/stream timeout + propagating disconnect cancellation) needs a design call on the timeout bound vs. legitimately long online-eval streams, so it belongs on its own.

Tracked in **OPIK-7901**.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/vertexai/CloseableVertexAiStreamingChatModel.java
```diff
@@ -0,0 +1,60 @@
...
+    @Override
+    public Set<Capability> supportedCapabilities() {
+        return delegate.supportedCapabilities();
+    }
+
+    // Best-effort: a close failure must never surface on an otherwise-successful call.
+    @Override
+    public void close() {
+        try {
+            vertexAI.close();
```
thiagohora, line 51: `close()` releases the `VertexAI` but never the delegate, and on the streaming path the delegate's `close()` is not a no-op.

In langchain4j `1.18.0-beta28` the two-arg `VertexAiGeminiStreamingChatModel(GenerativeModel, GenerationConfig)` ctor does null `vertexAI`, as the description says — but it also creates its own executor, `createDefaultExecutor()` → `new ThreadPoolExecutor(0, Integer.MAX_VALUE, 1, SECONDS, new SynchronousQueue<>())`, and `close()` shuts that down:

```java
if (vertexAI != null) vertexAI.close();
if (executor instanceof ExecutorService) ((ExecutorService) executor).shutdown();
```

So "its `close()` is a no-op" holds for `VertexAiGeminiChatModel`, whose `close()` is only the `vertexAI` branch, but not for the streaming one: every streaming request leaves an `ExecutorService` un-shut-down. Impact today is small — core size 0 and a 1 s keep-alive, so the worker dies and the pool is collectible — but it's the same ownership gap this PR fixes one level up, and it costs one line.

Measured rather than argued. Added `if (delegate instanceof AutoCloseable c) c.close()` to both wrappers, after the existing `vertexAI.close()` block and in its own try/catch, then drove the real path — generator → langchain4j → `ChunkedResponseHandler` terminal → close — against a WireMock `:streamGenerateContent` endpoint, closing on the terminal exactly as `handleCloseAndRelease` does:

- `VertexAIClientGeneratorTest` + `LlmProviderVertexAITest` + `ChatCompletionServiceTest` + 10 new probes → **36/36**.
- Control with the delegate close reverted: 5 of the 10 probes fail — the executor stays alive after close, neither wrapper forwards close, and a throwing delegate close goes unlogged. The other 5 pass either way, so they're properties of langchain4j rather than artifacts of the patch.
- No regression on the stream itself: full content across both chunks, exactly one terminal, zero errors, and zero `Failed to close…` warnings captured off the wrappers' own loggers. Shutting the pool down from inside its own worker task doesn't disturb langchain4j's post-terminal work (listeners, logging) — `shutdown()` only interrupts idle workers.
- Non-streaming is unaffected, as expected: the delegate's `vertexAI` field is null there, and a chat still completes end-to-end after closing the delegate directly. Both pass with and without the change.

One behavioural consequence, confirmed by the control rather than assumed: it makes the wrapper single-use. A second `chat()` after `close()` gets `RejectedExecutionException` from the shut-down executor instead of failing later on the closed `VertexAI`. Both wrappers are per-request and the `VertexAI` is dead at that point either way, so nothing on the current paths reaches it — it only constrains future code that tries to reuse a wrapper.

Worth adding the same line to `CloseableVertexAiChatModel` for symmetry. A genuine no-op today, but it stops the wrapper silently leaking if langchain4j ever gives that model resources of its own — which is the assumption that produced this bug in the first place.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/vertexai/CloseableVertexAiStreamingChatModel.java
```diff
@@ -0,0 +1,60 @@
...
+        } catch (Exception e) {
+            log.warn("Failed to close Vertex AI streaming client", e);
+        }
+    }
```
thiagohora, line 63: The shape I tested, for the streaming wrapper. Note `instanceof` rather than a direct `delegate.close()`: neither `ChatModel` nor `StreamingChatModel` declares `close()` or extends `AutoCloseable` in `1.18.0` (checked the interfaces), so with the field typed as the interface the direct call doesn't compile — `VertexAiGeminiStreamingChatModel` implements `AutoCloseable` on the class itself.

Own try/catch, after the `vertexAI.close()` block rather than inside it, so a failure on either side can't skip the other and neither can surface on an otherwise-successful call.

```suggestion
    // Best-effort: a close failure must never surface on an otherwise-successful call.
    @Override
    public void close() {
        try {
            vertexAI.close();
        } catch (Exception e) {
            log.warn("Failed to close Vertex AI streaming client", e);
        }
        // The langchain4j model owns resources of its own on this path: the two-arg ctor disclaims the VertexAI but
        // builds a per-instance executor, and its close() is the only thing that shuts that down.
        try {
            if (delegate instanceof AutoCloseable closeable) {
                closeable.close();
            }
        } catch (Exception e) {
            log.warn("Failed to close the delegate Vertex AI streaming model", e);
        }
    }
```

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/vertexai/CloseableVertexAiChatModel.java
```diff
@@ -0,0 +1,60 @@
...
+
+    // Best-effort: a close failure must never surface on an otherwise-successful call.
+    @Override
+    public void close() {
+        try {
+            vertexAI.close();
+        } catch (Exception e) {
+            log.warn("Failed to close Vertex AI client", e);
+        }
+    }
```
thiagohora, line 63: Same one-liner here for symmetry. `VertexAiGeminiChatModel.close()` really is only the `vertexAI != null` branch and that field is null on this path — I asserted both, and a chat still completes end-to-end after closing the delegate directly — so this is a genuine no-op today. It's worth having anyway: it stops the wrapper silently leaking if langchain4j ever gives this model resources of its own, which is exactly the assumption that produced the original bug.

```suggestion
    // Best-effort: a close failure must never surface on an otherwise-successful call.
    @Override
    public void close() {
        try {
            vertexAI.close();
        } catch (Exception e) {
            log.warn("Failed to close Vertex AI client", e);
        }
        // A no-op on this path today — the two-arg langchain4j ctor disclaims the VertexAI — but the delegate is the
        // only thing that can release resources it owns, so don't assume it never will.
        try {
            if (delegate instanceof AutoCloseable closeable) {
                closeable.close();
            }
        } catch (Exception e) {
            log.warn("Failed to close the delegate Vertex AI model", e);
        }
    }
```

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/vertexai/CloseableVertexAiStreamingChatModel.java
YarivHashaiComet: Applied in `4d04cbd57` — verbatim: an `instanceof AutoCloseable` guard in its own try/catch after the `vertexAI.close()` block. I trimmed the inline comment to the repo's terser one-line style, but the code is exactly as suggested.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/vertexai/CloseableVertexAiChatModel.java
YarivHashaiComet: Applied in `4d04cbd57` for symmetry (code verbatim, comment trimmed). Confirmed it's a genuine no-op today — the chat delegate has no executor and its `close()` only touches the (null) `vertexAI` — but kept in place so the wrapper can't silently leak if that model ever gains resources of its own.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/llm/vertexai/CloseableVertexAiStreamingChatModel.java
YarivHashaiComet: Confirmed against the exact version on our classpath — BOM `1.18.0` → `langchain4j.beta.version` = `1.18.0-beta28`. `javap` on that jar matches your read: `VertexAiGeminiStreamingChatModel.close()` runs `vertexAI.close()` **and** `executor.shutdown()`, while `VertexAiGeminiChatModel` has no executor field at all. Fixed in `4d04cbd57` — both wrappers now forward `close()` to an `AutoCloseable` delegate.

One difference from your validation: instead of the end-to-end WireMock probes, I added two lighter mock-based unit tests asserting `close()` forwards to both the `VertexAI` and an `AutoCloseable` delegate (28/28 across the Vertex + ChatCompletion suites). Happy to rewrite them to drive the real streaming path if you'd prefer that coverage. Also carried over the single-use note — fine here, since both wrappers are strictly per-request. Thanks for measuring this one.

thiagohora (approved):

PR #7800, opened 2026-08-10
[NA] [BE] fix: route cache pricing by usage shape

### LINKED ISSUE #7773: Cache cost calculator is chosen by provider but should be chosen by usage shape
### Problem

`CostService` selects the cache-aware cost calculator by provider via `PROVIDERS_CACHE_COST_CALCULATOR`. Most OpenAI-compatible providers (`openai`, `azure`, `xai`, `deepseek`, `fireworks_ai`, `moonshot`) are wired to `SpanCostCalculator.textGenerationWithCacheCostOpenAI`, which reads cached tokens from exactly one shape:

```
original_usage.prompt_tokens_details.cached_tokens   // (or the bare-OTel prompt_tokens_details.cached_tokens)
```

The problem is that the calculator is chosen by *provider*, but the field the calculator must read is determined by the *usage shape the span actually carries*, and those two are not always aligned. A provider assigned the OpenAI calculator that emits cache tokens in the Anthropic shape (`cache_read_input_tokens` / `cache_creation_input_tokens`) or a bare-OTel `cache_read.input_tokens` shape will have those tokens invisible to the calculator, so cached input is billed at the full input rate instead of the configured cache-read rate. The mismatch is silent: cost is simply too high, with no error.

This coupling is fragile as provider coverage grows, since a provider's SDK/instrumentation can change the usage shape independently of its canonical name.

### Proposal

Route the cache calculator by the usage keys actually present rather than (or in addition to) the provider name: if the span carries `prompt_tokens_details.cached_tokens` use the OpenAI reader; if it carries `cache_read_input_tokens`/`cache_creation_input_tokens` use the Anthropic reader; etc. A small shape-detection helper keyed on which cache fields are populated would make cache pricing correct regardless of which provider emitted the span, and would remove the need to hand-maintain a per-provider calculator map for every new cache-priced provider.

### Context

Surfaced while registering additional providers; Baz raised the same cache-shape concern on earlier provider PRs (snowflake / deepinfra). Filing as a standalone robustness improvement rather than folding it into a registration PR.

uuzzrm:
﻿## Details

`CostService` currently binds a cache calculator to a provider when model pricing is loaded. When a provider emits a different usage shape, cached tokens can be silently billed as regular input tokens.

This change keeps the provider calculator as a fallback, but selects a calculator from explicit usage-shape keys when present:

- Anthropic-style separated cache keys under `original_usage.*` use the Anthropic calculator.
- Bedrock camelCase keys use the Bedrock calculator.
- Google cached-content keys use the Google calculator.
- OpenAI nested cached-token keys use the OpenAI calculator.
- Nested OTel `cache_read.input_tokens` and `cache_creation.input_tokens` use an inclusive-cache calculator, because `prompt_tokens` already includes those buckets.
- Bare cache keys and mixed-provider payloads remain on the provider fallback because their inclusion semantics are ambiguous.

## Change checklist

- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves #7773

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Codex and GitHub CLI
- Model(s): GPT-5
- Scope: AI-assisted diagnosis, implementation, regression tests, and PR preparation
- Human verification: No maintainer verification yet; the focused backend tests and formatting checks below passed locally.

## Testing

- `mvn -f apps/opik-backend/pom.xml -Dtest=CostServiceTest,SpanCostCalculatorTest test`
  - 135 tests passed: 113 `CostServiceTest` and 22 `SpanCostCalculatorTest`.
- `mvn -f apps/opik-backend/pom.xml spotless:check`
- `git diff --check`

## Documentation

No documentation update is needed; this is an internal backend cost-calculation fix with no user-visible API or configuration changes.


uuzzrm:
Addressed the Baz review feedback in commit `af800a194`. The cache-shape detector now treats only explicit `original_usage.*` Anthropic cache keys as authoritative; bare cache keys no longer select the separated-cache calculator. Nested OTel cache buckets use an inclusive-cache calculator, and mixed shapes safely fall back to the provider calculator.\n\nValidation after the update: `CostServiceTest` 113/113 and `SpanCostCalculatorTest` 22/22 passed, `spotless:check` passed, and `git diff --check` passed.

uuzzrm:
Addressed the new Baz finding in commit `f3275236a`.

`OPENAI_USAGE_KEYS` already recognized the bare `prompt_tokens_details.cached_tokens` shape, but `textGenerationWithCacheCostOpenAI` did not read that key. The detector therefore selected the OpenAI calculator while cached tokens were billed at the regular input rate. The calculator now reads the bare key after the two `original_usage.*` forms, and `CostServiceTest` covers the full `CostService.calculateCost` dispatch with 1,000 prompt tokens and 200 cached tokens.

Validation available for this follow-up: `git diff --check` passed. The focused backend test and Spotless commands were already green on the previous head; this environment does not have Java/Maven installed to rerun them locally, so the new test should be confirmed by CI.

For the historical replay and injected-marker comments: the update path only recalculates when a span has no manual cost, and writes the current `ESTIMATED_COST_VERSION` for newly calculated values. The shape detector also requires exactly one supported shape and falls back to the provider calculator for mixed/ambiguous payloads. I have not changed the persistence/version boundary in this focused fix because introducing a migration policy would require a separate design decision and rollout test.

uuzzrm:
Addressed the latest Baz finding in commit `816624f37`. The OpenAI cache-token lookup now uses the first positive value across supported representations, so a zero-valued higher-priority key cannot mask a nonzero fallback representation. Added a regression case covering `original_usage.prompt_tokens_details.cached_tokens=0` with a nonzero bare cache key.

Local validation: `git diff --check` passed. Java/Maven are not installed in this environment, so the focused backend test and Spotless check are left to CI.

uuzzrm:
Added the requested dispatch-boundary regression in commit `b427d0525`. `CostServiceTest.calculateCostUsesOpenAICalculatorForResponsesCacheUsageShape` now exercises `CostService.calculateCost` with `original_usage.input_tokens_details.cached_tokens` and asserts the full `0.00325` cost, so a future routing regression cannot be hidden by a direct calculator test. `git diff --check` passed locally; Java/Maven remain unavailable here, so the focused test and Spotless check are left to CI.

uuzzrm:
Addressed the latest Baz finding in commit `40b46a46c`.

`findSingleCacheUsageShape` now validates the payload before selecting a cache calculator: the selected shape must have non-negative token values, a positive cache bucket, the required input total, and cache totals that are consistent with that input total. Inconsistent or mixed-provider markers now fall back to the provider calculator, so adding `cache_read.input_tokens` cannot force an undercharging inclusive-cache interpretation. The existing valid OTel cases remain covered, and `CostServiceTest.calculateCostFallsBackForInconsistentNestedOtelCacheUsage` covers `prompt_tokens=100` with `cache_read.input_tokens=200`.

The earlier bare OpenAI key, zero-valued override, and Responses dispatch findings remain covered by the current head. I also verified that `SpanDAO` only recalculates non-manual costs when usage/model/provider data is updated; I did not introduce a persistence migration for the historical replay/version-boundary question because that needs a maintainer decision about the rollout semantics.

Local `git diff --check` passed. The focused Maven tests and Spotless check could not run here because Java/Maven are not installed, so I left those checks to CI. Please re-review the current head.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/SpanCostCalculatorTest.java
```diff
@@ -158,7 +158,15 @@ private static Stream<Arguments> provideOpenAICacheCostCases() {
...
-                        "10.50"));
+                        "10.50"),
+                // A zero-valued higher-priority representation should not mask a nonzero
+                // cached-token value from another supported representation.
+                Arguments.of(
+                        Map.of("prompt_tokens", 1000, "completion_tokens", 100,
+                                "original_usage.prompt_tokens_details.cached_tokens", 0,
+                                "prompt_tokens_details.cached_tokens", 200),
+                        "zero-valued higher-priority key does not mask a nonzero fallback",
+                        "11.00"));
```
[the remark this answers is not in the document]
uuzzrm: Added in `b427d0525`. `CostServiceTest.calculateCostUsesOpenAICalculatorForResponsesCacheUsageShape` now exercises the public `CostService.calculateCost` boundary with `original_usage.input_tokens_details.cached_tokens` and asserts the complete expected cost, so a future dispatch regression cannot be hidden by a direct calculator test.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
```diff
@@ -571,6 +574,60 @@ private static BiFunction<ModelPrice, Map<String, Integer>, BigDecimal> resolveC
...
 
+    private static BigDecimal calculateCacheCostByUsageShape(
+            ModelPrice modelPrice,
+            Map<String, Integer> usage,
+            BiFunction<ModelPrice, Map<String, Integer>, BigDecimal> providerCalculator) {
+
+        // Cache-aware calculators use different semantics: OpenAI and Google include cached input
+        // in the prompt total, while Anthropic and Bedrock report cached buckets separately. Prefer
+        // explicit shape signals when they are present, and keep the provider fallback for ambiguous
+        // payloads such as a bare cache-read-only OTel usage map.
```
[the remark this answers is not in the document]
uuzzrm: Updated in `40b46a46c`. The earlier comment described the fallback too broadly. It now states that only payloads with no valid, unambiguous shape use the provider calculator. Nested OTel cache keys are validated against `prompt_tokens` and use the inclusive OTel calculator, while bare `cache_read_input_tokens` remains ambiguous and falls back safely.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
```diff
@@ -571,6 +591,69 @@ private static BiFunction<ModelPrice, Map<String, Integer>, BigDecimal> resolveC
...
+            BiFunction<ModelPrice, Map<String, Integer>, BigDecimal> providerCalculator) {
+
+        CacheUsageShape shape = findSingleCacheUsageShape(usage);
+        if (shape != null) {
+            return switch (shape) {
+                case OTEL -> SpanCostCalculator.textGenerationWithCacheCostOtel(modelPrice, usage);
+                case ANTHROPIC -> SpanCostCalculator.textGenerationWithCacheCostAnthropic(modelPrice, usage);
+                case BEDROCK -> SpanCostCalculator.textGenerationWithCacheCostBedrock(modelPrice, usage);
+                case GOOGLE -> SpanCostCalculator.textGenerationWithCacheCostGoogle(modelPrice, usage);
+                case OPENAI -> SpanCostCalculator.textGenerationWithCacheCostOpenAI(modelPrice, usage);
```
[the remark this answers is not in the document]
uuzzrm: I checked this against `SpanDAO` before deciding not to change it in this PR. The update path already recalculates an automatic cost when an update carries usage, model, or provider data, while manual costs are protected by the existing blank-version check; this behavior is present on `origin/main` and is not introduced by the cache-shape routing. A migration/version boundary for immutable historical pricing would need a separate persistence decision and rollout test, so I have left that out of this focused fix.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
```diff
@@ -86,6 +87,22 @@ public class CostService {
...
+            "original_usage.cache_creation_input_tokens");
+    private static final Set<String> BEDROCK_USAGE_KEYS = Set.of(
+            "original_usage.cacheReadInputTokens",
+            "original_usage.cacheWriteInputTokens");
+    private static final Set<String> GOOGLE_USAGE_KEYS = Set.of(
+            "original_usage.cached_content_token_count");
+    private static final Set<String> OPENAI_USAGE_KEYS = Set.of(
+            "original_usage.prompt_tokens_details.cached_tokens",
+            "original_usage.input_tokens_details.cached_tokens",
+            "prompt_tokens_details.cached_tokens");
```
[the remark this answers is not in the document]
uuzzrm: Fixed in `f3275236a`. The OpenAI cache lookup now includes the bare `prompt_tokens_details.cached_tokens` representation after the supported `original_usage.*` keys, and `CostServiceTest` covers the full `CostService.calculateCost` dispatch for that shape.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
```diff
@@ -571,6 +591,69 @@ private static BiFunction<ModelPrice, Map<String, Integer>, BigDecimal> resolveC
...
+    }
+
+    private static CacheUsageShape findSingleCacheUsageShape(Map<String, Integer> usage) {
+        CacheUsageShape match = null;
+        int matches = 0;
+
+        for (CacheUsageShape shape : CacheUsageShape.values()) {
+            if (containsAnyUsageKey(usage, usageKeys(shape))) {
+                match = shape;
+                matches++;
```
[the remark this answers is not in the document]
uuzzrm: Addressed in `40b46a46c`. Shape selection now requires a valid input total, non-negative values, a positive cache bucket, and cache totals consistent with the input; more than one detected shape also falls back. A valid normalized OTel payload intentionally uses the inclusive-cache calculator, while an inconsistent marker such as `cache_read.input_tokens > prompt_tokens` is covered by `CostServiceTest.calculateCostFallsBackForInconsistentNestedOtelCacheUsage`.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
```diff
@@ -571,6 +574,60 @@ private static BiFunction<ModelPrice, Map<String, Integer>, BigDecimal> resolveC
...
+        // payloads such as a bare cache-read-only OTel usage map.
+        if (containsAnyUsageKey(usage,
+                "original_usage.input_tokens",
+                "original_usage.output_tokens",
+                "original_usage.cache_read_input_tokens",
+                "original_usage.cache_creation_input_tokens",
+                "cache_creation_input_tokens",
+                "cache_read.input_tokens",
+                "cache_creation.input_tokens")) {
+            return SpanCostCalculator.textGenerationWithCacheCostAnthropic(modelPrice, usage);
```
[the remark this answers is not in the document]
uuzzrm: Addressed across `af800a194` and `40b46a46c`. Explicit `original_usage.*` separated-cache shapes now require their input total and non-negative cache values; nested OTel shapes use the inclusive calculator only when their cache totals are positive and do not exceed `prompt_tokens`; mixed or inconsistent shapes fall back to the provider calculator. The existing CostService regressions cover valid OTel, separated-cache, mixed-shape, and inconsistent-total cases.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/SpanCostCalculator.java
```diff
@@ -61,7 +63,9 @@ public static BigDecimal textGenerationWithCacheCostOpenAI(@NonNull ModelPrice m
         // Get the cached read input tokens; fall back to OTel bare key for LiteLLM/OTel spans
         int cachedReadInputTokens = usage.getOrDefault("original_usage.prompt_tokens_details.cached_tokens",
                 usage.getOrDefault("original_usage.input_tokens_details.cached_tokens",
-                        usage.getOrDefault(CACHE_READ_INPUT_TOKENS_KEY, 0)));
+                        usage.getOrDefault("prompt_tokens_details.cached_tokens",
+                                usage.getOrDefault(CACHE_READ_INPUT_TOKENS_KEY,
+                                        usage.getOrDefault(CACHE_READ_INPUT_TOKENS_OTEL_KEY, 0)))));
```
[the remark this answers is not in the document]
uuzzrm: Fixed in `816624f37`. The OpenAI cache lookup now selects the first positive supported representation, so a present zero-valued higher-priority key cannot mask a nonzero bare or lower-priority key. The regression covers `original_usage.prompt_tokens_details.cached_tokens = 0` together with a nonzero bare cache value.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
```diff
@@ -573,6 +605,168 @@ private static BiFunction<ModelPrice, Map<String, Integer>, BigDecimal> resolveC
...
+                Integer value = usage.get(key);
+                if (!isNonNegative(value)) {
+                    return false;
+                }
+                cacheTokens += value;
+                hasPositiveCacheTokens |= value > 0;
+            }
+        }
+
+        return hasPositiveCacheTokens && cacheTokens <= inputTokens;
```
andrescrz, line 685: **[high] Inclusive-vs-exclusive OTel cache semantics decided by a `cache <= prompt_tokens` proxy**

`isValidOtelUsageShape` uses `hasPositiveCacheTokens && cacheTokens <= inputTokens` as the discriminator for 'prompt_tokens already includes the cache buckets'. That inequality is not a structural invariant: an emitter whose `prompt_tokens` EXCLUDES cache (Anthropic/Bedrock semantics) also satisfies `cacheTokens <= prompt_tokens` in the common case, so those payloads are misclassified as OTEL and routed to `SpanCostCalculator.textGenerationWithCacheCostOtel`, which subtracts the cache buckets from the billed input tokens and undercharges. This PR contradicts itself on the very same keys: `textGenerationWithCachedTokensNotIncludedInCost` was extended in the same change to fall back to `cache_read.input_tokens` / `cache_creation.input_tokens` under EXCLUSIVE semantics, and `CostServiceTest.calculateCostFallsBackForInconsistentNestedOtelCacheUsage` asserts exactly that exclusive pricing for the identical key set — the only thing that flips the semantics is whether cache happens to exceed prompt_tokens. `OpenTelemetryMappingUtils.USAGE_KEYS_MAPPING` documents that the repo's OTel ingest normalizes Claude Code cache tokens onto the exclusive Anthropic path, confirming exclusive OTel emitters are a real, in-production input. Note the dispatcher also discards `provider`, the one authoritative signal available.

💡 Do not infer the shape from a magnitude comparison. Either (a) gate the OTEL branch on the provider/ingest source that is actually known to emit inclusive `gen_ai.usage.input_tokens` (the `provider` argument is already captured in the lambda), or (b) require a positive structural signal (e.g. a reconciliation such as `prompt_tokens == non_cached + cache_read + cache_creation` reported by the emitter, or an explicit ingest marker) before applying inclusive semantics, and default to the provider calculator otherwise. Whichever is chosen, make `textGenerationWithCachedTokensNotIncludedInCost` and the OTEL branch agree on the meaning of `cache_read.input_tokens` instead of encoding both.

_rule: `backend-correctness-check-the-real-structure-or-invariant-directly-not-a-naive-h`_

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
```diff
@@ -573,6 +605,168 @@ private static BiFunction<ModelPrice, Map<String, Integer>, BigDecimal> resolveC
...
 
+    private static BigDecimal calculateCacheCostByUsageShape(
+            ModelPrice modelPrice,
+            Map<String, Integer> usage,
+            BiFunction<ModelPrice, Map<String, Integer>, BigDecimal> providerCalculator) {
+
+        CacheUsageShape shape = findSingleCacheUsageShape(usage);
+        if (shape != null) {
+            return switch (shape) {
+                case OTEL -> SpanCostCalculator.textGenerationWithCacheCostOtel(modelPrice, usage);
```
andrescrz, line 616: **[high] Same OTel cache keys get opposite inclusion semantics in the two files**

The cluster teaches two contradictory meanings for the newly-introduced keys `cache_read.input_tokens` / `cache_creation.input_tokens`. In SpanCostCalculator.java:224-229 they were added as a third-level fallback inside `textGenerationWithCachedTokensNotIncludedInCost`, whose contract (javadoc at line 204-216 and the tier comment at line 231) is that the input-token value EXCLUDES the cache buckets — so `totalPromptTokens = inputTokens + cacheCreation + cacheRead` and every bucket is billed additively. In CostService.java:616 the very same keys route to `SpanCostCalculator.textGenerationWithCacheCostOtel`, whose contract (SpanCostCalculator.java:137-147) is the opposite: `prompt_tokens` INCLUDES the cache buckets, so they are subtracted before billing. Which interpretation a span gets is decided only by whether `isValidOtelUsageShape` happens to hold (`cacheTokens <= prompt_tokens`, CostService.java:685). For one Anthropic/Bedrock model on one instrumentation this means span A (cache_read 200, prompt_tokens 1000) is billed inclusively at 750 input tokens while span B (cache_read 200, prompt_tokens 100) from the same integration is billed exclusively at 100 input + 200 cache — the numeric ratio, not the payload's provenance, picks the pricing model. If the dotted keys really come from an Anthropic-shaped source (Anthropic's `usage.input_tokens` excludes both cache buckets), the OTEL branch systematically under-charges input; if they come from an OTel-inclusive source, the Anthropic fallback double-counts the cached tokens. One of the two branches is always wrong. CostServiceTest.java:129-141 pins the inclusive reading for `claude-haiku-4-5` and CostServiceTest.java:143-156 pins the exclusive reading for the same model, locking the contradiction in.

💡 Pick one owner for the dotted OTel keys. Either (a) keep `textGenerationWithCacheCostOtel` as the single consumer and revert the `CACHE_*_OTEL_KEY` fallbacks added to `textGenerationWithCachedTokensNotIncludedInCost` (SpanCostCalculator.java:226,229), letting invalid/ambiguous OTel payloads fall through with the cache buckets simply unbilled; or (b) drop the OTEL shape from `CacheUsageShape` and treat the dotted keys as exclusive everywhere. Whichever is chosen, document the source instrumentation that establishes the inclusion semantics and make both files agree.

_general finding_

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
```diff
@@ -573,6 +605,168 @@ private static BiFunction<ModelPrice, Map<String, Integer>, BigDecimal> resolveC
...
+        // bare cache keys, inconsistent cache totals, and mixed-provider maps.
+        return providerCalculator.apply(modelPrice, usage);
+    }
+
+    private static CacheUsageShape findSingleCacheUsageShape(Map<String, Integer> usage) {
+        CacheUsageShape match = null;
+
+        for (CacheUsageShape shape : CacheUsageShape.values()) {
+            if (containsAnyUsageKey(usage, usageKeys(shape))) {
+                if (!isValidUsageShape(usage, shape) || match != null) {
```
andrescrz, line 634: **[medium] A single zero-valued cache key from another family disables shape detection entirely**

`findSingleCacheUsageShape` treats 'keys present but not valid' as a hard abort (`if (!isValidUsageShape(usage, shape) || match != null) return null;`), and `isValidUsageShape` returns false whenever a family's cache keys are all zero (`hasPositiveCacheTokens` / `hasPositiveCacheValue`). Zero-valued cache counters are routine — OTel exporters and LiteLLM emit `cache_read.input_tokens: 0`, `original_usage.cache_creation_input_tokens: 0` on a cache miss. So a genuine OpenAI-shaped payload (`original_usage.prompt_tokens`, `original_usage.prompt_tokens_details.cached_tokens = 500`) that additionally carries `cache_read.input_tokens: 0` aborts detection and falls back to the provider calculator. For a model priced under `anthropic`/`bedrock`/`gemini` that means the OpenAI-inclusive prompt tokens are billed in full AND the cached tokens are billed again at the cache rate — the exact double-charge this change exists to fix, defeated by one zero. `usage` is sanitized only for nulls (`UsageUtils.sanitizeUsage`), so zeros always survive to here. The equality requirement in `isValidInclusiveCacheUsageShape` (line 714) triggers the same abort when two OpenAI cached-token aliases disagree, even though `SpanCostCalculator.firstPositiveUsageValue` would price it fine.

💡 Separate 'this family is not in use' from 'this family is malformed': skip a family whose cache keys are all zero (treat it as absent) instead of returning null, and only abort on genuinely contradictory data (negative values, cache > input). Continue the loop rather than returning on the first non-matching family.

_rule: `backend-null-safety-guards-fail-closed-on-unknown-ambiguous-or-unvalidated-external-inp`_

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/SpanCostCalculator.java
```diff
@@ -185,9 +222,11 @@ private static BigDecimal textGenerationWithCachedTokensNotIncludedInCost(@NonNu
         int inputTokens = usage.getOrDefault(inputTokensKey, usage.getOrDefault("prompt_tokens", 0));
         int outputTokens = usage.getOrDefault(outputTokensKey, usage.getOrDefault("completion_tokens", 0));
         int cacheCreationInputTokens = usage.getOrDefault(cacheCreationInputTokensKey,
-                usage.getOrDefault(CACHE_CREATION_INPUT_TOKENS_KEY, 0));
+                usage.getOrDefault(CACHE_CREATION_INPUT_TOKENS_KEY,
+                        usage.getOrDefault(CACHE_CREATION_INPUT_TOKENS_OTEL_KEY, 0)));
         int cacheReadInputTokens = usage.getOrDefault(cacheReadInputTokensKey,
-                usage.getOrDefault(CACHE_READ_INPUT_TOKENS_KEY, 0));
+                usage.getOrDefault(CACHE_READ_INPUT_TOKENS_KEY,
+                        usage.getOrDefault(CACHE_READ_INPUT_TOKENS_OTEL_KEY, 0)));
```
andrescrz, line 229: **[medium] Anthropic/Bedrock fallback to the OTel dotted cache keys double-bills cached tokens**

The same diff declares (Javadoc on textGenerationWithCacheCostOtel, line 137-140) that `cache_read.input_tokens` / `cache_creation.input_tokens` are the OTel shape where the cache buckets ARE INCLUDED in `prompt_tokens`. Lines 224-229 add exactly those two keys as fallbacks inside `textGenerationWithCachedTokensNotIncludedInCost`, whose entire contract (Javadoc line 204-206) is that cache tokens are NOT included in the input count. On that path `inputTokens` falls back to `prompt_tokens` (line 222), so the cached tokens are billed twice: once at the full input rate inside `inputTokens`, and again at the cache-read/cache-creation rate. It also inflates `totalPromptTokens` (line 233) to input+cache, which can push the model into a higher `above_NNNk` tier than the real prompt size. Reachable path: CostService.calculateCacheCostByUsageShape falls back to the provider calculator whenever `findSingleCacheUsageShape` returns null (ambiguous or invalid shape) - e.g. an anthropic-provider span carrying both `cache_read.input_tokens` and an OpenAI cached-tokens key, or an OTel payload where the cache sum exceeds prompt_tokens. Those are precisely the payloads CostService's validator rejected, and this new fallback re-admits them into a calculator with incompatible semantics.

💡 Do not add the OTel included-in-prompt keys (`cache_read.input_tokens`, `cache_creation.input_tokens`) to `textGenerationWithCachedTokensNotIncludedInCost`. If Anthropic/Bedrock spans really can carry them, subtract them from `inputTokens` first (as `textGenerationWithCacheCostOtel` does) instead of adding them on top, or route those payloads to `textGenerationWithCacheCostOtel`.

_rule: `backend-data-access-cheap-fast-paths-and-fallback-branches-must-preserve-the-ful`_

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/SpanCostCalculator.java
```diff
@@ -117,6 +124,36 @@ public static BigDecimal textGenerationWithCacheCostOpenAI(@NonNull ModelPrice m
...
+     * Calculates cost for the OpenTelemetry GenAI usage shape, where cache buckets are included in
+     * {@code prompt_tokens} and reported again as nested cache usage keys.
+     */
+    public static BigDecimal textGenerationWithCacheCostOtel(@NonNull ModelPrice modelPrice,
+            @NonNull Map<String, Integer> usage) {
+        int inputTokens = usage.getOrDefault("prompt_tokens", 0);
+        int outputTokens = usage.getOrDefault("completion_tokens", 0);
+        int cacheReadInputTokens = usage.getOrDefault(CACHE_READ_INPUT_TOKENS_OTEL_KEY, 0);
+        int cacheCreationInputTokens = usage.getOrDefault(CACHE_CREATION_INPUT_TOKENS_OTEL_KEY, 0);
+        int nonCachedInputTokens = Math.max(0, inputTokens - cacheReadInputTokens - cacheCreationInputTokens);
```
andrescrz, line 147: **[medium] textGenerationWithCacheCostOtel subtracts cache tokens from billable input even when that bucket's rate is zero**

`nonCachedInputTokens` always subtracts both cache buckets, but `ModelPrice.effectiveCacheCreationInputTokenPrice`/`effectiveCacheReadInputTokenPrice` return `BigDecimal.ZERO` when the model has no such rate (ModelPrice.defaultBuilder sets both to ZERO, and CostService.resolveCalculator enters this branch on `isPositive(cacheCreation) || isPositive(cacheRead)` - an OR, so only one of the two needs a price). For a model priced with cache-read only, an OTel span carrying `cache_creation.input_tokens` gets those tokens removed from the input bill and then multiplied by ZERO - the tokens are billed at nothing, silently undercharging. The file's own established pattern does the opposite: `textGenerationCost` (line 36-38) and `textGenerationWithCacheCostOpenAI` (line 95-97) only subtract the audio bucket when `inputAudioRate.compareTo(BigDecimal.ZERO) > 0`.

💡 Guard each subtraction on a positive rate, mirroring the audio handling: only subtract `cacheReadInputTokens` when `effectiveCacheReadInputTokenPrice(inputTokens)` is positive, and likewise for `cacheCreationInputTokens`; otherwise leave those tokens in `nonCachedInputTokens` so they are billed at the base input rate.

_general finding_

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/SpanCostCalculator.java
```diff
@@ -14,6 +15,14 @@ class SpanCostCalculator {
...
     private static final String CACHE_READ_INPUT_TOKENS_KEY = "cache_read_input_tokens";
     private static final String CACHE_CREATION_INPUT_TOKENS_KEY = "cache_creation_input_tokens";
+    private static final String CACHE_READ_INPUT_TOKENS_OTEL_KEY = "cache_read.input_tokens";
+    private static final String CACHE_CREATION_INPUT_TOKENS_OTEL_KEY = "cache_creation.input_tokens";
+    private static final List<String> OPENAI_CACHE_READ_USAGE_KEYS = List.of(
+            "original_usage.prompt_tokens_details.cached_tokens",
+            "original_usage.input_tokens_details.cached_tokens",
+            "prompt_tokens_details.cached_tokens",
+            CACHE_READ_INPUT_TOKENS_KEY,
+            CACHE_READ_INPUT_TOKENS_OTEL_KEY);
```
andrescrz, line 25: **[medium] The newly added dotted OTel cache-key fallbacks are untested**

Three new fallback branches added by this diff have no test that reaches them. `SpanCostCalculatorTest.provideOpenAICacheCostCases` never supplies `cache_read.input_tokens`, so the fifth entry of `OPENAI_CACHE_READ_USAGE_KEYS` (line 25) is never exercised; `textGenerationWithCacheCostUsesOtelCacheKeyFallback` (SpanCostCalculatorTest.java:328-336) uses only the bare `cache_read_input_tokens`/`cache_creation_input_tokens` keys, so the new third-level fallbacks at lines 226 and 229 are never exercised either. `provideOtelIncludedCacheCostCases` only covers the happy path with both cache rates positive - there is no case with a zero cache-creation rate, and none where the cache buckets exceed `prompt_tokens`.

💡 Add parameterized cases for: OpenAI usage carrying only `cache_read.input_tokens`; Anthropic/Bedrock usage carrying only the dotted OTel keys (which should surface the double-billing described above); and `textGenerationWithCacheCostOtel` with `cacheCreationInputTokenPrice` left at ZERO.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
```diff
@@ -38,6 +38,139 @@ void calculateCostUsesCacheAwareCalculatorWhenCachePricesConfigured() {
...
+
+        BigDecimal cost = CostService.calculateCost("gpt-4o", "openai", usage, null);
+
+        // Responses API usage includes cached tokens in prompt_tokens. 800 * 0.0000025 +
+        // 100 * 0.00001 + 200 * 0.00000125 = 0.00325.
+        assertThat(cost).isEqualByComparingTo("0.00325");
+    }
+
+    @Test
+    void calculateCostUsesIncludedCacheCalculatorForNestedOtelUsageShape() {
```
andrescrz, line 115: **[medium] OTel shape test cannot falsify the new OTel routing (gpt-4o/openai yields the identical value on the fallback path)**

calculateCostUsesIncludedCacheCalculatorForNestedOtelUsageShape asserts 0.00325 for gpt-4o/openai with {prompt_tokens=1000, completion_tokens=100, cache_read.input_tokens=200}. If the new CacheUsageShape.OTEL branch in CostService.calculateCacheCostByUsageShape were removed, the payload would fall through to the provider calculator for "openai" (textGenerationWithCacheCostOpenAI), which in this same PR gained "cache_read.input_tokens" as the last entry of OPENAI_CACHE_READ_USAGE_KEYS (SpanCostCalculator.java:25). That path computes 1000-200=800 non-cached input, 100 output, 200 cache-read -> 800*2.5e-6 + 100*1e-5 + 200*1.25e-6 = 0.00325, byte-for-byte the asserted value. The test therefore proves nothing about the routing its name claims to exercise, and it is a functional duplicate of calculateCostUsesBareOpenAICacheUsageKey (line 87) and calculateCostUsesOpenAICalculatorForResponsesCacheUsageShape (line 101), which both assert the same 0.00325 for the same model. Only calculateCostUsesIncludedCacheCalculatorForNestedOtelCreationShape (line 129, claude-haiku-4-5/anthropic: 0.0013325 via OTel vs 0.0015825 via the Anthropic fallback) actually discriminates.

💡 Make the OTel read-only case discriminating the same way the creation-shape case is: use a model whose provider fallback is NOT the OpenAI calculator (e.g. claude-haiku-4-5 / anthropic, where the Anthropic fallback treats prompt_tokens as cache-exclusive and yields a different number), or drop this test and keep only the discriminating variant at line 129.

_rule: `backend-testing-make-test-assertions-actually-falsify-the-change-under-test`_

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
```diff
@@ -38,6 +38,139 @@ void calculateCostUsesCacheAwareCalculatorWhenCachePricesConfigured() {
         assertThat(cost).isEqualByComparingTo("0.0001658");
     }
 
+    @Test
+    void calculateCostUsesUsageShapeWhenProviderCacheCalculatorDoesNotMatch() {
```
andrescrz, line 42: **[medium] New CacheUsageShape BEDROCK and GOOGLE detection branches have no cross-provider test**

CostService now routes by usage shape for all five CacheUsageShape values (OTEL, ANTHROPIC, BEDROCK, GOOGLE, OPENAI). The nine new tests cover shape-based routing for OTEL, ANTHROPIC (line 42) and OPENAI (line 73) when the shape disagrees with the provider, but nothing covers BEDROCK or GOOGLE arriving under a mismatched provider — exactly the scenario the change exists for. The pre-existing bedrock test (line 29) and google test (line 378) pass provider=bedrock/google_vertexai together with matching keys, so they resolve to the same calculator through the PROVIDERS_CACHE_COST_CALCULATOR fallback and would still pass if usageKeys(BEDROCK)/usageKeys(GOOGLE) or their isValidUsageShape branches were deleted.

💡 Add cases for Bedrock-shaped usage (original_usage.cacheReadInputTokens / cacheWriteInputTokens) and Google-shaped usage (original_usage.cached_content_token_count) submitted with a non-matching provider (e.g. "openai"), asserting the Bedrock/Google cost math rather than the OpenAI-fallback math.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/SpanCostCalculatorTest.java
andrescrz, line 169: **[medium] New OPENAI_CACHE_READ_USAGE_KEYS entries and their priority order are only partially covered**

OPENAI_CACHE_READ_USAGE_KEYS (SpanCostCalculator lines 20-25) is a 5-entry ordered list introduced by this change, and firstPositiveUsageValue makes the *order* load-bearing. The added case only pins entry 3 (prompt_tokens_details.cached_tokens) winning after a zero at entry 1. Entry 5, "cache_read.input_tokens", which this PR appended, is never fed to textGenerationWithCacheCostOpenAI by any case in this class, and no case pins entry 2 (original_usage.input_tokens_details.cached_tokens) taking precedence over entry 3. Reordering or dropping those entries would not fail any test, yet it changes billed cost.

💡 Add cases to provideOpenAICacheCostCases: (a) usage with only "cache_read.input_tokens" reaching the OpenAI calculator, and (b) usage carrying both "original_usage.input_tokens_details.cached_tokens" and a different "prompt_tokens_details.cached_tokens" value, asserting the higher-priority key wins.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
```diff
@@ -573,6 +605,168 @@ private static BiFunction<ModelPrice, Map<String, Integer>, BigDecimal> resolveC
...
+                case ANTHROPIC -> SpanCostCalculator.textGenerationWithCacheCostAnthropic(modelPrice, usage);
+                case BEDROCK -> SpanCostCalculator.textGenerationWithCacheCostBedrock(modelPrice, usage);
+                case GOOGLE -> SpanCostCalculator.textGenerationWithCacheCostGoogle(modelPrice, usage);
+                case OPENAI -> SpanCostCalculator.textGenerationWithCacheCostOpenAI(modelPrice, usage);
+            };
+        }
+
+        // Payloads with no valid, unambiguous shape remain on the provider calculator. This covers
+        // bare cache keys, inconsistent cache totals, and mixed-provider maps.
+        return providerCalculator.apply(modelPrice, usage);
```
andrescrz, line 626: **[low] Mixed-shape fallback drops cache buckets instead of billing them**

When `findSingleCacheUsageShape` returns null the code falls back to `providerCalculator`, but the provider calculators do not consume the full key set the router just inspected. `OPENAI_CACHE_READ_USAGE_KEYS` (SpanCostCalculator.java:20-25) now includes the dotted OTel read key, and `firstPositiveUsageValue` (SpanCostCalculator.java:127-135) returns only the FIRST positive representation — it does not reconcile or sum. `textGenerationWithCacheCostOpenAI` has no notion of a cache-creation bucket at all. The result, pinned by CostServiceTest.java:158-172, is that a payload carrying `cache_read.input_tokens: 200` alongside `original_usage.prompt_tokens_details.cached_tokens: 100` bills only 100 cached tokens; the 200-token bucket is neither subtracted from input nor charged at the cache rate. The same happens to `cache_creation.input_tokens` on every OpenAI-provider fallback (acknowledged as 'unclassified' in the comment at CostServiceTest.java:66-68). Since the fallback is the designated safety net for exactly the payloads the router could not classify, silently discarding a token bucket there is the worst place to lose it.

💡 On the ambiguous/invalid path, log at debug with the conflicting keys so these payloads are diagnosable in production, and prefer a calculator that at least consumes both cache buckets (or bill the un-consumed bucket at the cache rate) rather than letting `firstPositiveUsageValue` pick one representation and drop the rest.

_general finding_

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
```diff
@@ -88,6 +89,34 @@ public class CostService {
                     Map.entry("gemini", SpanCostCalculator::textGenerationWithCacheCostGoogle),
                     Map.entry("vertex_ai-anthropic_models", SpanCostCalculator::textGenerationWithCacheCostAnthropic));
 
+    private static final Set<String> OTEL_INCLUDED_CACHE_USAGE_KEYS = Set.of(
+            "cache_read.input_tokens",
+            "cache_creation.input_tokens");
+    private static final Set<String> ANTHROPIC_USAGE_KEYS = Set.of(
```
andrescrz, line 95: **[low] Constant and helper names do not describe what they hold or check**

`ANTHROPIC_USAGE_KEYS`, `BEDROCK_USAGE_KEYS`, `GOOGLE_USAGE_KEYS` and `OPENAI_USAGE_KEYS` contain only the CACHE keys for each family — the input keys live in separate `*_INPUT_USAGE_KEYS` lists — so the names read as 'all usage keys for this provider' and invite exactly the wrong edit. The sibling `OTEL_INCLUDED_CACHE_USAGE_KEYS` gets this right, making the inconsistency more confusing. Likewise `hasNonNegativePositiveCacheValue` (line 725) fuses two different predicates into a self-contradictory name: it returns false on any negative value and true only if some value is strictly positive.

💡 Rename to `ANTHROPIC_CACHE_USAGE_KEYS` / `BEDROCK_CACHE_USAGE_KEYS` / `GOOGLE_CACHE_USAGE_KEYS` / `OPENAI_CACHE_USAGE_KEYS` (matching `OTEL_INCLUDED_CACHE_USAGE_KEYS`), and rename the helper to something like `hasUsableCacheValue` with a one-line comment stating the two conditions it enforces.

_rule: `backend-style-maintainability-names-must-accurately-reflect-actual-behavior-type-and-unit`_

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
```diff
@@ -38,6 +38,139 @@ void calculateCostUsesCacheAwareCalculatorWhenCachePricesConfigured() {
...
+
+        BigDecimal cost = CostService.calculateCost("claude-haiku-4-5", "anthropic", usage, null);
+
+        // OTel input_tokens includes both cache buckets: 750 * 0.000001 + 100 * 0.000005 +
+        // 200 * 0.0000001 + 50 * 0.00000125 = 0.0013325.
+        assertThat(cost).isEqualByComparingTo("0.0013325");
+    }
+
+    @Test
+    void calculateCostFallsBackForInconsistentNestedOtelCacheUsage() {
```
andrescrz, line 144: **[low] Shape-validation reject branches other than 'cache exceeds prompt' are untested**

calculateCostFallsBackForInconsistentNestedOtelCacheUsage covers only one of the new reject paths (cacheTokens > inputTokens in isValidOtelUsageShape). The other new guards added in CostService still have no test: (a) isValidInclusiveCacheUsageShape returning false when two keys of the same shape carry conflicting positive values (e.g. original_usage.prompt_tokens_details.cached_tokens=200 together with prompt_tokens_details.cached_tokens=100), (b) hasPositiveCacheValue / hasPositiveCacheTokens false when every cache key is 0, (c) the isNonNegative(null) path where the shape's input key is absent, and (d) the cacheTokens == inputTokens boundary, which must be accepted. There is also no tiered-pricing case through textGenerationWithCacheCostOtel, even though this file pins tier semantics for every other calculator (issue #6982, the 128k case, and the 272k case).

💡 Add cases for the conflicting-duplicate-cache-key rejection, the all-zero cache values fallback, the cacheTokens == prompt_tokens boundary (must stay on the OTel calculator), and one above-threshold model routed through the OTel shape.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/SpanCostCalculatorTest.java
```diff
@@ -316,4 +324,32 @@ private static Stream<Arguments> provideAnthropicBedrockCacheCostCases() {
...
+
+        BigDecimal cost = SpanCostCalculator.textGenerationWithCacheCostOtel(modelPrice, usage);
+
+        assertThat(cost).isEqualByComparingTo(expectedCost);
+    }
+
+    private static Stream<Arguments> provideOtelIncludedCacheCostCases() {
+        return Stream.of(
+                // input=1000 includes cache_read=200, so only 800 regular input tokens are billed.
+                Arguments.of(Map.of("prompt_tokens", 1000, "completion_tokens", 100,
```
andrescrz, line 347: **[low] Clamp branch of textGenerationWithCacheCostOtel (cache tokens > prompt_tokens) is not exercised**

textGenerationWithCacheCostOtel contains Math.max(0, inputTokens - cacheReadInputTokens - cacheCreationInputTokens) (SpanCostCalculator line 147), a distinct branch guarding the case where the reported cache buckets exceed prompt_tokens. Both new cases stay well inside the normal range (1000 vs 200/250), so the clamp is never reached. CostService.isValidOtelUsageShape currently keeps such payloads away from this calculator, but the method is called directly here and by CostService's switch, so the guard is unpinned: if the CostService validity check is ever relaxed, a negative bucket would silently flow into the cost.

💡 Add a case such as prompt_tokens=100, cache_read.input_tokens=200 asserting the non-cached bucket clamps to 0 rather than going negative.

andrescrz (commented):

13 findings — 2 high · 7 medium · 4 low. Suppressed by team conventions: see `suppressed.md`.

PR #7803, opened 2026-08-10
feat(cost): derive provider set from price file, keep only genuine aliases

### LINKED ISSUE #7794: Derive provider set from the price file instead of a manual PROVIDERS_MAPPING allowlist
### Problem

`CostService.PROVIDERS_MAPPING` is a hand-maintained allowlist: only the `litellm_provider` values listed there resolve to prices. Any model whose `litellm_provider` is absent has `buildModelPrice` return null, so its price is dropped at load time and its spans are billed at `DEFAULT_COST` (zero); cost tracking and the per-evaluation spend budget silently under-report for that provider.

The bundled `model_prices_and_context_window.json` currently contains many priced providers that are not in the map (a quick count today: `oci` ~44 priced models, `replicate` ~40, `together_ai` ~36, `databricks` ~28, `watsonx` ~28, plus aggregators like `openrouter`, `vercel_ai_gateway`, `novita`, `azure_ai`). Each one needs a separate one-line PR to `PROVIDERS_MAPPING` (there have been many of these recently under #7697), and because they all touch the same map literal, every merge conflicts the others, which is a lot of rebase churn for a mechanical change.

### Proposal

For the common case where the canonical provider name equals the `litellm_provider` value (which is true for the large majority of entries), derive the mapping from the price file instead of listing providers by hand: at load time, collect the distinct `litellm_provider` values present in the price data and treat each as canonical by default. Keep `PROVIDERS_MAPPING` only for the genuine exceptions where the names differ (`microsoft -> azure`, `vertex_ai-language-models -> google_vertexai`, `moonshotai -> moonshot`, `z-ai -> zai`, and similar), i.e. flip it from an allowlist to an alias table.

That removes the need for a PR per provider, closes the silent-zero-cost gap for every current and future provider in one change, and eliminates the recurring merge conflicts. Providers with special cost shapes (cache, tiers) would still opt in via `PROVIDERS_CACHE_COST_CALCULATOR` exactly as they do now; this only changes how the base provider set is discovered.

### Notes

Aggregators (`openrouter`, `vercel_ai_gateway`, ...) still route models under a vendor namespace, so the existing provider-prefix fallback in `findModelPrice` continues to matter and is orthogonal to this change. Happy to implement if the approach sounds right.

Xsidz:
## Summary

\`CostService.PROVIDERS_MAPPING\` was an allowlist gate: any \`litellm_provider\` absent from the map caused \`buildRuntimeKey\` / \`buildModelPrice\` to drop the price row, billing those spans at zero. This silently affected ~200+ models across providers like \`together_ai\`, \`oci\`, \`replicate\`, \`databricks\`, \`watsonx\`, and others that are fully priced in the bundled \`model_prices_and_context_window.json\`.

**Changes:**

- Renamed \`PROVIDERS_MAPPING\` to \`PROVIDER_ALIASES\` to reflect the new semantic
- Stripped the 21 identity entries (\`openai -> openai\`, \`groq -> groq\`, etc.) — these are now handled by the identity fallback
- Kept the 7 genuine aliases where \`litellm_provider\` differs from the canonical key:

| litellm_provider | canonical |
|---|---|
| \`vertex_ai-language-models\` | \`google_vertexai\` |
| \`gemini\` | \`google_ai\` |
| \`vertex_ai-anthropic_models\` | \`anthropic_vertexai\` |
| \`bedrock_converse\` | \`bedrock\` |
| \`microsoft\` | \`azure\` |
| \`moonshotai\` | \`moonshot\` |
| \`z-ai\` | \`zai\` |

- \`buildRuntimeKey\` and \`buildModelPrice\` now use \`getOrDefault(provider, provider)\` so any non-blank \`litellm_provider\` present in the price file is priced automatically
- New regression test: \`together_ai\` model (previously zero-cost) now returns a non-zero price

No new provider registrations needed — this change is self-applying for every current and future provider in the price file.

## Test plan

- [x] Regression test: \`together_ai\` model priced correctly (was zero before)
- [x] All existing \`CostServiceTest\` tests still pass (aliases, Bedrock, Google, Azure, moonshot, z-ai)

Fixes #7794

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
@@ -226,13 +226,16 @@ public class CostService {
             String modelPrefix = originalModelName.substring(0, prefixSlash);
             String canonicalFromPrefix = PROVIDER_ALIASES.getOrDefault(modelPrefix, modelPrefix);
             if (canonicalFromPrefix != null && !canonicalFromPrefix.equalsIgnoreCase(provider)) {
-                String prefixKey = createModelProviderKey(modelName, canonicalFromPrefix);
-                ModelPrice prefixMatch = modelProviderPrices.get(prefixKey);
-                if (prefixMatch != null) {
-                    log.debug(
-                            "Found model price using model-name provider prefix. Original model: '{}', supplied provider: '{}', prefix-derived provider: '{}'",
-                            originalModelName, provider, canonicalFromPrefix);
-                    return prefixMatch;
+                for (String candidate : new String[]{modelName, normalizedModelName, baseOriginalModelName,
+                        baseNormalizedModelName, baseOriginalVersionName, baseNormalizedVersionName}) {
+                    ModelPrice prefixMatch = modelProviderPrices
+                            .get(createModelProviderKey(candidate, canonicalFromPrefix));
+                    if (prefixMatch != null) {
+                        log.debug(
+                                "Found model price using model-name provider prefix. Original model: '{}', supplied provider: '{}', prefix-derived provider: '{}'",
+                                originalModelName, provider, canonicalFromPrefix);
+                        return prefixMatch;
+                    }
                 }
             }
         }
@@ -408,7 +411,7 @@ public class CostService {
      */
     private static String buildRuntimeKey(String modelName, ModelCostData modelCost) {
         String provider = Optional.ofNullable(modelCost.litellmProvider()).orElse("");
-        if (provider.isEmpty()) {
+        if (provider.isBlank()) {
             return null;
         }
         String canonical = PROVIDER_ALIASES.getOrDefault(provider, provider);
@@ -420,11 +423,11 @@ public class CostService {
 
     private static ModelPrice buildModelPrice(ModelCostData modelCost) {
         String provider = Optional.ofNullable(modelCost.litellmProvider()).orElse("");
-        if (provider.isEmpty()) {
+        if (provider.isBlank()) {
             return null;
         }
         // Resolve alias; unknown providers use their own name as canonical.
-        provider = PROVIDER_ALIASES.getOrDefault(provider, provider);
+        String canonicalProvider = PROVIDER_ALIASES.getOrDefault(provider, provider);
 
         BigDecimal inputPrice = Optional.ofNullable(modelCost.inputCostPerToken()).map(BigDecimal::new)
                 .orElse(BigDecimal.ZERO);
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
@@ -225,7 +225,7 @@ public class CostService {
         if (prefixSlash > 0) {
             String modelPrefix = originalModelName.substring(0, prefixSlash);
             String canonicalFromPrefix = PROVIDER_ALIASES.getOrDefault(modelPrefix, modelPrefix);
-            if (canonicalFromPrefix != null && !canonicalFromPrefix.equalsIgnoreCase(provider)) {
+            if (!canonicalFromPrefix.equalsIgnoreCase(provider)) {
                 for (String candidate : new String[]{modelName, normalizedModelName, baseOriginalModelName,
                         baseNormalizedModelName, baseOriginalVersionName, baseNormalizedVersionName}) {
                     ModelPrice prefixMatch = modelProviderPrices
@@ -392,7 +392,7 @@ public class CostService {
     private static void applyDirectOverride(Map<String, ModelPrice> prices, String modelName, ModelCostData override) {
         String runtimeKey = buildRuntimeKey(modelName, override);
         if (runtimeKey == null) {
-            log.warn("Override entry '{}' has unknown provider '{}'; skipping",
+            log.warn("Override entry '{}' skipped: provider '{}' is blank or resolves to a legacy Bedrock path",
                     modelName, override.litellmProvider());
             return;
         }
@@ -426,9 +426,6 @@ public class CostService {
         if (provider.isBlank()) {
             return null;
         }
-        // Resolve alias; unknown providers use their own name as canonical.
-        String canonicalProvider = PROVIDER_ALIASES.getOrDefault(provider, provider);
-
         BigDecimal inputPrice = Optional.ofNullable(modelCost.inputCostPerToken()).map(BigDecimal::new)
                 .orElse(BigDecimal.ZERO);
         BigDecimal outputPrice = Optional.ofNullable(modelCost.outputCostPerToken()).map(BigDecimal::new)
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
@@ -39,8 +39,7 @@ public class CostService {
             Map.entry("bedrock_converse", "bedrock"),
             Map.entry("microsoft", "azure"),
             Map.entry("moonshotai", "moonshot"),
-            Map.entry("z-ai", "zai"),
-            Map.entry("cerebras", "cerebras"));
+            Map.entry("z-ai", "zai"));
 
     // Online evaluation (and OTel ingestion) resolve models to LlmProvider serialized values whose names
     // differ from the canonical price-table vocabulary. Normalize those to the single canonical provider
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
@@ -27,36 +27,20 @@ import java.util.function.BiFunction;
 public class CostService {
     private static final char MODEL_PROVIDER_SEPARATOR = '/';
     private static final Map<String, ModelPrice> modelProviderPrices;
-    private static final Map<String, String> PROVIDERS_MAPPING = Map.ofEntries(
-            Map.entry("openai", "openai"),
+    // Alias table: maps litellm_provider values to the canonical provider name used as the
+    // price-map key. Only entries where the names genuinely differ are needed; providers whose
+    // litellm_provider value already equals the canonical name are handled automatically by the
+    // identity fallback in buildRuntimeKey / buildModelPrice (any provider present in the price
+    // file is priced; only the aliases below need explicit entries).
+    private static final Map<String, String> PROVIDER_ALIASES = Map.ofEntries(
             Map.entry("vertex_ai-language-models", "google_vertexai"),
             Map.entry("gemini", "google_ai"),
-            Map.entry("anthropic", "anthropic"),
             Map.entry("vertex_ai-anthropic_models", "anthropic_vertexai"),
-            Map.entry("bedrock", "bedrock"),
             Map.entry("bedrock_converse", "bedrock"),
-            Map.entry("groq", "groq"),
-            Map.entry("jina_ai", "jina_ai"),
-            Map.entry("elastic", "elastic"),
             Map.entry("microsoft", "azure"),
-            Map.entry("azure", "azure"),
-            Map.entry("mistral", "mistral"),
-            Map.entry("xai", "xai"),
-            Map.entry("deepseek", "deepseek"),
-            Map.entry("perplexity", "perplexity"),
-            Map.entry("fireworks_ai", "fireworks_ai"),
-            Map.entry("moonshot", "moonshot"),
             Map.entry("moonshotai", "moonshot"),
-            Map.entry("ai21", "ai21"),
-            Map.entry("morph", "morph"),
-            Map.entry("inception", "inception"),
-            Map.entry("meta", "meta"),
-            Map.entry("zai", "zai"),
             Map.entry("z-ai", "zai"),
-            Map.entry("sambanova", "sambanova"),
-            Map.entry("nebius", "nebius"),
-            Map.entry("snowflake", "snowflake"),
-            Map.entry("deepinfra", "deepinfra"));
+            Map.entry("cerebras", "cerebras"));
 
     // Online evaluation (and OTel ingestion) resolve models to LlmProvider serialized values whose names
     // differ from the canonical price-table vocabulary. Normalize those to the single canonical provider
@@ -241,7 +225,7 @@ public class CostService {
         int prefixSlash = originalModelName.indexOf('/');
         if (prefixSlash > 0) {
             String modelPrefix = originalModelName.substring(0, prefixSlash);
-            String canonicalFromPrefix = PROVIDERS_MAPPING.get(modelPrefix);
+            String canonicalFromPrefix = PROVIDER_ALIASES.getOrDefault(modelPrefix, modelPrefix);
             if (canonicalFromPrefix != null && !canonicalFromPrefix.equalsIgnoreCase(provider)) {
                 String prefixKey = createModelProviderKey(modelName, canonicalFromPrefix);
                 ModelPrice prefixMatch = modelProviderPrices.get(prefixKey);
@@ -418,15 +402,17 @@ public class CostService {
 
     /**
      * Computes the runtime key {@code <parsedModel>/<canonicalProvider>} for a price-map entry.
-     * Returns null if the provider isn't in {@link #PROVIDERS_MAPPING} or the entry is invalid
-     * for the resolved provider (e.g. legacy Bedrock paths).
+     * Returns null only if the entry is invalid for the resolved provider (e.g. legacy Bedrock paths)
+     * or the provider string is blank. Any non-blank litellm_provider is accepted: aliases in
+     * {@link #PROVIDER_ALIASES} are remapped; all others use the provider value as-is so new
+     * providers in the price file are priced automatically without requiring a code change.
      */
     private static String buildRuntimeKey(String modelName, ModelCostData modelCost) {
         String provider = Optional.ofNullable(modelCost.litellmProvider()).orElse("");
-        String canonical = PROVIDERS_MAPPING.get(provider);
-        if (canonical == null) {
+        if (provider.isEmpty()) {
             return null;
         }
+        String canonical = PROVIDER_ALIASES.getOrDefault(provider, provider);
         if (!isValidModelProvider(modelName, canonical)) {
             return null;
         }
@@ -435,9 +421,11 @@ public class CostService {
 
     private static ModelPrice buildModelPrice(ModelCostData modelCost) {
         String provider = Optional.ofNullable(modelCost.litellmProvider()).orElse("");
-        if (!PROVIDERS_MAPPING.containsKey(provider)) {
+        if (provider.isEmpty()) {
             return null;
         }
+        // Resolve alias; unknown providers use their own name as canonical.
+        provider = PROVIDER_ALIASES.getOrDefault(provider, provider);
 
         BigDecimal inputPrice = Optional.ofNullable(modelCost.inputCostPerToken()).map(BigDecimal::new)
                 .orElse(BigDecimal.ZERO);
--- a/apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
@@ -503,7 +503,7 @@ class CostServiceTest {
     }
 
     /**
-     * Mistral cost tracking: until `mistral` was added to PROVIDERS_MAPPING the entire set of
+     * Mistral cost tracking: until `mistral` was added to PROVIDER_ALIASES the entire set of
      * upstream LiteLLM `litellm_provider: "mistral"` rows was dropped at startup, so every
      * Mistral span returned cost = 0. This locks in both the upstream-sourced rows and the
      * two override-only rows (Medium 3.5, Small 4) added alongside the registry fix.
@@ -538,7 +538,7 @@ class CostServiceTest {
 
     /**
      * Overrides file: `litellm_provider: "microsoft"` is remapped to canonical `azure` via
-     * PROVIDERS_MAPPING. A span emitted with provider=`azure` and the model name must hit
+     * PROVIDER_ALIASES. A span emitted with provider=`azure` and the model name must hit
      * the override row for `azure/multilingual-e5-large`.
      */
     @Test
@@ -894,7 +894,7 @@ class CostServiceTest {
      * OpenRouter exposes Moonshot's Kimi family under a different namespace prefix
      * ({@code moonshotai/*}) than LiteLLM's canonical ({@code moonshot/*}) — see the
      * {@code MOONSHOTAI_*} entries in {@code OpenRouterModelName}. Without a
-     * {@code moonshotai -> moonshot} alias in {@link CostService#PROVIDERS_MAPPING}, the
+     * {@code moonshotai -> moonshot} alias in {@link CostService#PROVIDER_ALIASES}, the
      * provider-prefix fallback returns null and the aggregator-routed request resolves to
      * {@code DEFAULT_COST}, silently under-charging every Kimi call routed through OpenRouter.
      * With the alias, the fallback maps {@code moonshotai} to the canonical {@code moonshot}
@@ -972,7 +972,7 @@ class CostServiceTest {
      * Same gap as the {@code moonshotai} alias, for five more vendors OpenRouter resells.
      * {@code ai21}, {@code morph}, {@code inception}, {@code meta} and {@code zai} all carry
      * non-zero-cost rows in {@code model_prices_and_context_window.json}, but none were in
-     * {@link CostService#PROVIDERS_MAPPING}, so {@code buildModelPrice} dropped every one of
+     * {@link CostService#PROVIDER_ALIASES}, so {@code buildModelPrice} dropped every one of
      * them at load time and the provider-prefix fallback had nothing to resolve against. Any
      * call routed through OpenRouter fell through to {@code DEFAULT_COST}.
      * <p>
@@ -1061,5 +1061,18 @@ class CostServiceTest {
                                 "original_usage.completion_tokens", 200,
                                 "original_usage.prompt_tokens_details.cached_tokens", 300),
                         "0.005709"));
+    }
+
+    /**
+     * Providers present in the price file but previously absent from PROVIDER_ALIASES were silently
+     * dropped, resulting in zero cost. With the identity-fallback change, any non-blank litellm_provider
+     * is priced automatically if its models appear in the file.
+     */
+    @Test
+    void calculateCostForPreviouslyUnregisteredProvider() {
+        // together_ai is in model_prices_and_context_window.json but was never in the old allowlist.
+        var usage = Map.of("prompt_tokens", 100, "completion_tokens", 50);
+        BigDecimal cost = CostService.calculateCost("together-ai-21.1b-41b", "together_ai", usage, null);
+        assertThat(cost).isNotNull().isGreaterThan(BigDecimal.ZERO);
     }
 }
--- a/apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
@@ -1071,8 +1071,10 @@ class CostServiceTest {
     @Test
     void calculateCostForPreviouslyUnregisteredProvider() {
         // together_ai is in model_prices_and_context_window.json but was never in the old allowlist.
+        // together-ai-21.1b-41b: input_cost_per_token=8e-7, output_cost_per_token=8e-7
+        // Expected: (100 * 8e-7) + (50 * 8e-7) = 0.00012
         var usage = Map.of("prompt_tokens", 100, "completion_tokens", 50);
         BigDecimal cost = CostService.calculateCost("together-ai-21.1b-41b", "together_ai", usage, null);
-        assertThat(cost).isNotNull().isGreaterThan(BigDecimal.ZERO);
+        assertThat(cost).isNotNull().isEqualByComparingTo(new BigDecimal("0.00012"));
     }
 }
--- a/apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
@@ -1014,6 +1014,8 @@ class CostServiceTest {
                 Arguments.of("z-ai/glm-4.5", "openrouter", "0.00104"),
                 // Date-suffixed variant: prefix fallback must also try date-stripped name.
                 Arguments.of("z-ai/glm-4.5-2025-12-17", "openrouter", "0.00104"),
+                // Version-suffixed variant: primary lookup misses, z-ai canonicalizes to zai, :0 stripped.
+                Arguments.of("z-ai/glm-4.5:0", "openrouter", "0.00104"),
                 Arguments.of("z-ai/glm-5", "openrouter", "0.00164"),
                 Arguments.of("morph/morph-v3-fast", "openrouter", "0.00104"),
                 Arguments.of("morph/morph-v3-large", "openrouter", "0.00128"),
--- a/apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
@@ -1012,6 +1012,8 @@ class CostServiceTest {
                 Arguments.of("ai21/jamba-mini-1.7", "openrouter", "0.00028"),
                 // OpenRouter namespaces Z.ai as z-ai/, the price file as zai/.
                 Arguments.of("z-ai/glm-4.5", "openrouter", "0.00104"),
+                // Date-suffixed variant: prefix fallback must also try date-stripped name.
+                Arguments.of("z-ai/glm-4.5-2025-12-17", "openrouter", "0.00104"),
                 Arguments.of("z-ai/glm-5", "openrouter", "0.00164"),
                 Arguments.of("morph/morph-v3-fast", "openrouter", "0.00104"),
                 Arguments.of("morph/morph-v3-large", "openrouter", "0.00128"),
--- a/apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
@@ -503,10 +503,11 @@ class CostServiceTest {
     }
 
     /**
-     * Mistral cost tracking: until `mistral` was added to PROVIDER_ALIASES the entire set of
-     * upstream LiteLLM `litellm_provider: "mistral"` rows was dropped at startup, so every
-     * Mistral span returned cost = 0. This locks in both the upstream-sourced rows and the
-     * two override-only rows (Medium 3.5, Small 4) added alongside the registry fix.
+     * Mistral cost tracking: the identity fallback in {@code buildRuntimeKey} / {@code buildModelPrice}
+     * loads any provider present in the price file without an explicit {@code PROVIDER_ALIASES} entry,
+     * so the entire set of upstream LiteLLM {@code litellm_provider: "mistral"} rows is now loaded at
+     * startup. This locks in both the upstream-sourced rows and the two override-only rows
+     * (Medium 3.5, Small 4).
      */
     @ParameterizedTest
     @MethodSource("provideMistralModels")
@@ -969,17 +970,14 @@ class CostServiceTest {
     }
 
     /**
-     * Same gap as the {@code moonshotai} alias, for five more vendors OpenRouter resells.
-     * {@code ai21}, {@code morph}, {@code inception}, {@code meta} and {@code zai} all carry
-     * non-zero-cost rows in {@code model_prices_and_context_window.json}, but none were in
-     * {@link CostService#PROVIDER_ALIASES}, so {@code buildModelPrice} dropped every one of
-     * them at load time and the provider-prefix fallback had nothing to resolve against. Any
-     * call routed through OpenRouter fell through to {@code DEFAULT_COST}.
+     * Five vendors OpenRouter resells: {@code ai21}, {@code morph}, {@code inception}, {@code meta},
+     * and {@code zai} all carry non-zero-cost rows in {@code model_prices_and_context_window.json}.
+     * The identity fallback in {@code buildRuntimeKey} / {@code buildModelPrice} loads them
+     * automatically without explicit {@link CostService#PROVIDER_ALIASES} entries.
      * <p>
-     * {@code z-ai} needs two entries for the same reason {@code moonshot} does: the map is read
-     * both with the price file's {@code litellm_provider} ({@code zai}) when loading rows, and
-     * with the model-name prefix OpenRouter uses ({@code z-ai}) when resolving the fallback.
-     * The other four spell both the same way, so one entry each.
+     * {@code z-ai} still needs an explicit alias ({@code z-ai} to {@code zai}) because OpenRouter
+     * uses {@code z-ai} as the model-name prefix while the price file's {@code litellm_provider}
+     * is {@code zai}. The other four spell both the same way, so the identity fallback is enough.
      * <p>
      * All of these take the {@link SpanCostCalculator#textGenerationCost} path. Four of the
      * models below publish a {@code cache_read_input_token_cost}, but none of these providers is


Xsidz:
Fixed in 97b2dcbc7f:

T1: Strengthened the regression test to assert the exact expected cost (0.00012) rather than just isGreaterThan(ZERO).

T2/T3: Removed the cerebras->cerebras identity entry from PROVIDER_ALIASES. The getOrDefault fallback already handles it without an explicit entry.

T4: The prefix fallback behavior (z-ai/glm-4.5 looking up under zai) predates this PR and is tracked separately. This PR only changes PROVIDERS_MAPPING.get to PROVIDER_ALIASES.getOrDefault, which doesn't alter that fallback's behavior either way.

andrescrz:
Hi @Xsidz 

Thanks!

Xsidz:
Merge conflicts resolved. Rebased onto current main (21ea9dc170).

Conflict in `CostService.java`: upstream had added 4 self-alias entries (sambanova, nebius, snowflake, deepinfra) to the end of the map while this PR replaced the allowlist with an alias-only table using `getOrDefault`. Took this PR's version — those self-aliases are redundant with the identity fallback.

Conflict in `CostServiceTest.java`: upstream added a deepinfra parameterized test in the same region where this PR added the `together_ai` regression test. Kept both tests.

All 111 CostServiceTest cases pass locally.

andrescrz:
Hi @Xsidz 

Thanks for the contribution!

Thanks!

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
```diff
@@ -973,4 +973,17 @@ private static Stream<Arguments> provideAggregatorRoutedVendorCases() {
...
+     * Providers present in the price file but previously absent from PROVIDER_ALIASES were silently
+     * dropped, resulting in zero cost. With the identity-fallback change, any non-blank litellm_provider
+     * is priced automatically if its models appear in the file.
+     */
+    @Test
+    void calculateCostForPreviouslyUnregisteredProvider() {
+        // together_ai is in model_prices_and_context_window.json but was never in the old allowlist.
+        var usage = Map.of("prompt_tokens", 100, "completion_tokens", 50);
+        BigDecimal cost = CostService.calculateCost("together-ai-21.1b-41b", "together_ai", usage, null);
+        assertThat(cost).isNotNull().isGreaterThan(BigDecimal.ZERO);
```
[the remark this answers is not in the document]
Xsidz: Fixed in 459df91839 — test now asserts the exact expected cost (0.00012) rather than just isGreaterThan(ZERO).

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
```diff
@@ -27,34 +27,20 @@
 public class CostService {
     private static final char MODEL_PROVIDER_SEPARATOR = '/';
     private static final Map<String, ModelPrice> modelProviderPrices;
-    private static final Map<String, String> PROVIDERS_MAPPING = Map.ofEntries(
-            Map.entry("openai", "openai"),
+    // Alias table: maps litellm_provider values to the canonical provider name used as the
+    // price-map key. Only entries where the names genuinely differ are needed; providers whose
+    // litellm_provider value already equals the canonical name are handled automatically by the
+    // identity fallback in buildRuntimeKey / buildModelPrice (any provider present in the price
+    // file is priced; only the aliases below need explicit entries).
```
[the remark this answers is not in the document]
Xsidz: Fixed in 459df91839 — cerebras self-alias removed from PROVIDER_ALIASES. The getOrDefault fallback handles it without an explicit entry.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
```diff
@@ -27,34 +27,20 @@
...
             Map.entry("moonshotai", "moonshot"),
-            Map.entry("ai21", "ai21"),
-            Map.entry("morph", "morph"),
-            Map.entry("inception", "inception"),
-            Map.entry("meta", "meta"),
-            Map.entry("zai", "zai"),
             Map.entry("z-ai", "zai"),
-            Map.entry("sambanova", "sambanova"),
-            Map.entry("nebius", "nebius"));
+            Map.entry("cerebras", "cerebras"));
```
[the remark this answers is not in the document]
Xsidz: Fixed in 459df91839 — same fix as above, redundant entry removed.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
```diff
@@ -237,7 +223,7 @@ private static ModelPrice findModelPrice(String modelName, String provider) {
         int prefixSlash = originalModelName.indexOf('/');
         if (prefixSlash > 0) {
             String modelPrefix = originalModelName.substring(0, prefixSlash);
-            String canonicalFromPrefix = PROVIDERS_MAPPING.get(modelPrefix);
+            String canonicalFromPrefix = PROVIDER_ALIASES.getOrDefault(modelPrefix, modelPrefix);
```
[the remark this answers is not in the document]
Xsidz: Fixed in dc17fe7547 — the prefix fallback now loops over all pre-computed normalized variants (modelName, normalizedModelName, baseOriginalModelName, etc.) until one hits. z-ai/glm-4.5-2025-12-17 reaches baseOriginalModelName = glm-4.5 and matches the stored key.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
```diff
@@ -418,15 +401,17 @@ private static void applyDirectOverride(Map<String, ModelPrice> prices, String m
...
+     * providers in the price file are priced automatically without requiring a code change.
      */
     private static String buildRuntimeKey(String modelName, ModelCostData modelCost) {
         String provider = Optional.ofNullable(modelCost.litellmProvider()).orElse("");
-        String canonical = PROVIDERS_MAPPING.get(provider);
-        if (canonical == null) {
+        if (provider.isEmpty()) {
             return null;
         }
+        String canonical = PROVIDER_ALIASES.getOrDefault(provider, provider);
```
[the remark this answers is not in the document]
Xsidz: Fixed in dc17fe7547 — changed to provider.isBlank() in both buildRuntimeKey and buildModelPrice.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
```diff
@@ -435,9 +420,11 @@ private static String buildRuntimeKey(String modelName, ModelCostData modelCost)
 
     private static ModelPrice buildModelPrice(ModelCostData modelCost) {
         String provider = Optional.ofNullable(modelCost.litellmProvider()).orElse("");
-        if (!PROVIDERS_MAPPING.containsKey(provider)) {
+        if (provider.isEmpty()) {
             return null;
         }
+        // Resolve alias; unknown providers use their own name as canonical.
+        provider = PROVIDER_ALIASES.getOrDefault(provider, provider);
```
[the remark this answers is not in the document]
Xsidz: Fixed in dc17fe7547 — renamed to canonicalProvider at the point of alias resolution so the raw provider value stays named provider throughout.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
```diff
@@ -1012,6 +1012,8 @@ private static Stream<Arguments> provideAggregatorRoutedVendorCases() {
                 Arguments.of("ai21/jamba-mini-1.7", "openrouter", "0.00028"),
                 // OpenRouter namespaces Z.ai as z-ai/, the price file as zai/.
                 Arguments.of("z-ai/glm-4.5", "openrouter", "0.00104"),
+                // Date-suffixed variant: prefix fallback must also try date-stripped name.
+                Arguments.of("z-ai/glm-4.5-2025-12-17", "openrouter", "0.00104"),
```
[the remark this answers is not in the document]
Xsidz: Added in 5f15cb0a8c — Arguments.of("z-ai/glm-4.5:0", "openrouter", "0.00104") added to provideAggregatorRoutedVendorCases. Primary lookup misses, z-ai canonicalizes to zai, version-stripped candidate matches.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
```diff
@@ -1012,6 +1012,10 @@ private static Stream<Arguments> provideAggregatorRoutedVendorCases() {
...
+                // Version-suffixed variant: primary lookup misses, z-ai canonicalizes to zai, :0 stripped.
+                Arguments.of("z-ai/glm-4.5:0", "openrouter", "0.00104"),
```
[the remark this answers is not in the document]
Xsidz: The unit test covers what this PR changes: the CostService lookup logic. A SpanDAO integration test asserting persisted cost and total_estimated_cost_version would exercise the persistence layer, which is unchanged by this PR and already covered by the existing together_ai real-resource tests. Adding another real-resource test for the same DAO path is out of scope here.

andrescrz (commented):

12 findings — 0 high · 8 medium · 4 low. Suppressed by team conventions: see `suppressed.md`.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
```diff
@@ -435,9 +423,11 @@ private static String buildRuntimeKey(String modelName, ModelCostData modelCost)
 
     private static ModelPrice buildModelPrice(ModelCostData modelCost) {
         String provider = Optional.ofNullable(modelCost.litellmProvider()).orElse("");
-        if (!PROVIDERS_MAPPING.containsKey(provider)) {
+        if (provider.isBlank()) {
             return null;
         }
+        // Resolve alias; unknown providers use their own name as canonical.
+        String canonicalProvider = PROVIDER_ALIASES.getOrDefault(provider, provider);
```
andrescrz: **[medium] Newly-admitted providers with cache rates get the cache-blind calculator, over-charging cached tokens**

Removing the allowlist admits 158 price rows that publish cache_creation_input_token_cost / cache_read_input_token_cost under providers absent from PROVIDERS_CACHE_COST_CALCULATOR (azure_ai 36, openrouter 34, vercel_ai_gateway 27, novita 14, vertex_ai 11, dashscope 8, bedrock_mantle 7, minimax 6, ...). resolveCalculator's getOrDefault (line 558) therefore returns textGenerationCost, which bills the whole prompt at the input rate and never applies the cache-read rate. Reproduced on this branch vs. its base with usage {prompt_tokens:1000, completion_tokens:200, cache_read_input_tokens:800}: `openrouter/anthropic/claude-sonnet-4.5` (provider openrouter) and `claude-sonnet-4-5` (provider azure_ai) both go 0 -> 0.006, where the cache-aware value is 0.00384 — a ~56% over-charge on cached calls. openrouter is a normal ingestion path (SpanDAO, BudgetGuard). This is not a regression — those rows priced at zero before — but it widens an already-known gap roughly 10x (19 -> 177 rows), and the note that documents that deferral (CostServiceTest.java:984-988) was not extended.

💡 Either register the high-volume newcomers in PROVIDERS_CACHE_COST_CALCULATOR (openrouter / azure_ai / vercel_ai_gateway are OpenAI-shaped) or extend the existing deferral note to cover the widened set, and add a cached-token test for one of them. Note the map must stay keyed on the RAW litellm_provider — it deliberately holds `gemini`, `vertex_ai-language-models`, `vertex_ai-anthropic_models`, `bedrock_converse`, so re-keying it on the canonical name would silently drop the cache calculator for 87 rows that price correctly today. Separately, delete the unused `canonicalProvider` local (line 430) rather than wiring it into resolveCalculator.

_general finding_

andrescrz: **[medium] buildModelPrice's new lines are dead: canonicalProvider is never used and the blank guard is unreachable**

Both statements the diff adds to buildModelPrice are dead. (1) `String canonicalProvider = PROVIDER_ALIASES.getOrDefault(provider, provider);` (line 430) is never read afterwards — resolveCalculator(...) at line 475 is still called with the raw `provider`, so the alias resolution and its explanatory comment describe behavior the method does not have. (2) `if (provider.isBlank()) return null;` (lines 427-428) can never fire: both callers (parseModelPrices line 309-314 and applyDirectOverride line 393-399) first require buildRuntimeKey(...) != null, and buildRuntimeKey already returns null for a blank provider. A reviewer or future author reading line 430 will reasonably 'fix' it by passing canonicalProvider into resolveCalculator — which would silently break cache pricing, because PROVIDERS_CACHE_COST_CALCULATOR (lines 59-74) is keyed by the RAW litellm_provider for `gemini`, `vertex_ai-language-models`, `vertex_ai-anthropic_models` and none of their canonical targets (`google_ai`, `google_vertexai`, `anthropic_vertexai`) exist in that map.

💡 Delete the unused `canonicalProvider` local (and the unreachable blank guard) so buildModelPrice keeps a single responsibility, OR — if the intent was to key the calculator by canonical provider — pass it to resolveCalculator and simultaneously re-key PROVIDERS_CACHE_COST_CALCULATOR on canonical names (`google_ai`, `google_vertexai`, `anthropic_vertexai` instead of `gemini`, `vertex_ai-language-models`, `vertex_ai-anthropic_models`), with a test asserting Gemini/Vertex cache-read pricing still resolves to textGenerationWithCacheCostGoogle.

_rule: `backend-architecture-remove-or-wire-in-dead-code-before-merging`_

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
```diff
@@ -27,36 +27,19 @@
...
     private static final char MODEL_PROVIDER_SEPARATOR = '/';
     private static final Map<String, ModelPrice> modelProviderPrices;
-    private static final Map<String, String> PROVIDERS_MAPPING = Map.ofEntries(
-            Map.entry("openai", "openai"),
+    // Alias table: maps litellm_provider values to the canonical provider name used as the
+    // price-map key. Only entries where the names genuinely differ are needed; providers whose
+    // litellm_provider value already equals the canonical name are handled automatically by the
+    // identity fallback in buildRuntimeKey / buildModelPrice (any provider present in the price
+    // file is priced; only the aliases below need explicit entries).
+    private static final Map<String, String> PROVIDER_ALIASES = Map.ofEntries(
```
andrescrz, line 35: **[medium] Identity fallback leaves ~69 sub-family provider rows under keys no caller can ever supply**

The identity fallback turns every litellm_provider into its own canonical name, but several litellm names are sub-families of a provider Opik already has a canonical vocabulary for, so their rows load into the price map under a key no caller addresses. Callers pass LlmProvider serialized values (openai, anthropic, gemini->google_ai, vertex-ai->google_vertexai, bedrock, openrouter) or a raw gen_ai.system value — never `cohere_chat`, `azure_text`, `text-completion-openai`, `fireworks_ai-embedding-models`, `bedrock_mantle`, `amazon_nova`, `vertex_ai-mistral_models`, `text-completion-codestral`. Replaying buildRuntimeKey over the shipped price file: 69 of the 80 rows in those families have no sibling reachable under the canonical provider Opik callers use, so they still resolve to DEFAULT_COST (zero cost) — e.g. `command-r`, `command-r-plus`, `command-a-03-2025` (cohere_chat) for provider `cohere`; `azure/gpt-35-turbo-instruct` (azure_text) for `azure`; `davinci-002`, `babbage-002` (text-completion-openai) for `openai`; the whole `bedrock_mantle/openai.gpt-5.6-*` family for `bedrock`. The change therefore does not deliver the 'new providers priced automatically' property for exactly the families where the price row exists but the provider name is a LiteLLM-internal variant. Note this is pre-existing rather than introduced — those rows were dropped entirely before — but it contradicts the new comment's 'any provider present in the price file is priced' claim and leaves ~69 rows (e.g. Cohere command-r under provider `cohere`) still at zero.

💡 Add the sub-family aliases to PROVIDER_ALIASES so they collapse onto the canonical provider callers actually send: cohere_chat->cohere, azure_text->azure, text-completion-openai->openai, text-completion-codestral->mistral, fireworks_ai-embedding-models->fireworks_ai, bedrock_mantle->bedrock, amazon_nova->bedrock, vertex_ai->google_vertexai and vertex_ai-*_models->google_vertexai. I verified this introduces no new key collisions in the current price file (46 colliding keys before and after the diff, unchanged), and cover at least one family (e.g. command-r under provider `cohere`) with a test.

_general finding_

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
```diff
@@ -418,15 +404,17 @@ private static void applyDirectOverride(Map<String, ModelPrice> prices, String m
...
+     * Returns null only if the entry is invalid for the resolved provider (e.g. legacy Bedrock paths)
+     * or the provider string is blank. Any non-blank litellm_provider is accepted: aliases in
+     * {@link #PROVIDER_ALIASES} are remapped; all others use the provider value as-is so new
+     * providers in the price file are priced automatically without requiring a code change.
      */
     private static String buildRuntimeKey(String modelName, ModelCostData modelCost) {
         String provider = Optional.ofNullable(modelCost.litellmProvider()).orElse("");
-        String canonical = PROVIDERS_MAPPING.get(provider);
-        if (canonical == null) {
+        if (provider.isBlank()) {
```
andrescrz, line 414: **[medium] Overrides-file provider validation is now unreachable; garbage providers load silently**

applyDirectOverride (lines 394-398) and applyAlias (line 375) rely on buildRuntimeKey returning null to detect an unknown litellm_provider in the Opik-owned model_prices_overrides.json and log 'Override entry ... has unknown provider ...; skipping'. With the allowlist replaced by the blank-only guard at line 414, buildRuntimeKey can now only return null for a blank provider or the legacy Bedrock-slash case, so that warning branch is effectively dead for the overrides path. A typo in a hand-maintained overrides entry (e.g. litellm_provider `opanai` or `azure-openai` instead of `microsoft`) no longer fails loudly — it inserts a live key `<model>/opanai` that no caller ever queries, and the override silently prices at 0 forever. Evidence the guard is too weak: LiteLLM's own placeholder row `sample_spec` carries litellm_provider = "one of https://docs.litellm.ai/docs/providers", which is non-blank and is now loaded into the price map as a key. The test file was updated only for the PROVIDERS_MAPPING -> PROVIDER_ALIASES rename and adds no negative case for this.

💡 Validate the overrides file against a known-provider set (or against the set of litellm_provider values observed in model_prices_and_context_window.json) so applyDirectOverride/applyAlias still fail loudly on typos, and reject obviously non-provider values (whitespace/URL-shaped, e.g. skip the `sample_spec` row) in buildRuntimeKey. Add a test asserting an overrides entry with an unknown provider is skipped and logged.

_general finding_

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
```diff
@@ -1062,4 +1066,19 @@ private static Stream<Arguments> provideDeepinfraProviderCases() {
...
+     * dropped, resulting in zero cost. With the identity-fallback change, any non-blank litellm_provider
+     * is priced automatically if its models appear in the file.
+     */
+    @Test
+    void calculateCostForPreviouslyUnregisteredProvider() {
+        // together_ai is in model_prices_and_context_window.json but was never in the old allowlist.
+        // together-ai-21.1b-41b: input_cost_per_token=8e-7, output_cost_per_token=8e-7
+        // Expected: (100 * 8e-7) + (50 * 8e-7) = 0.00012
+        var usage = Map.of("prompt_tokens", 100, "completion_tokens", 50);
+        BigDecimal cost = CostService.calculateCost("together-ai-21.1b-41b", "together_ai", usage, null);
```
andrescrz, line 1079: **[medium] New test locks the happy path only; no coverage for the two regressions the widened key space creates**

calculateCostForPreviouslyUnregisteredProvider is the only test added for the allowlist removal, and it exercises the least risky shape: a together_ai row with no cache rates, no provider prefix and metadata = null. It therefore cannot catch either cross-file consequence of the change — that a non-zero estimate now suppresses getCostFromMetadata for the 1004 newly-admitted priced rows (CostService.java:96, reached from SpanDAO.java:2606), or that newly-admitted providers publishing cache rates get textGenerationCost and over-charge cached tokens (CostService.java:430/475). Separately, it hardcodes an upstream-owned price (together-ai-21.1b-41b, 8e-7/8e-7 — the only `together-ai-2*` key in the file) from model_prices_and_context_window.json, which the daily LiteLLM sync regenerates, so it will break on an upstream re-price for a reason unrelated to the behavior it documents, unlike the override-backed rows the rest of the suite pins. (The metadata half is a behavior change confirmed by probe — calculateCost with metadata cost 0.42 returns 0.42 on base and 0.00012 on HEAD — though SDK-reported cost travels via total_estimated_cost, which still wins in SpanDAO, so the practical blast radius is limited.)

💡 Add (a) a case for a newly-admitted provider with metadata.cost.total_cost asserting the reported cost still wins, (b) a cached-token case for an openrouter or azure_ai row asserting the cache-read rate is applied, and (c) pin the identity-fallback assertion to an Opik-owned row in model_prices_overrides.json (or assert cost > 0 plus the ratio) instead of a bare upstream number.

_general finding_

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
```diff
@@ -972,7 +972,7 @@ private static Stream<Arguments> provideModelNamesWithDateSuffixes() {
      * Same gap as the {@code moonshotai} alias, for five more vendors OpenRouter resells.
      * {@code ai21}, {@code morph}, {@code inception}, {@code meta} and {@code zai} all carry
      * non-zero-cost rows in {@code model_prices_and_context_window.json}, but none were in
-     * {@link CostService#PROVIDERS_MAPPING}, so {@code buildModelPrice} dropped every one of
+     * {@link CostService#PROVIDER_ALIASES}, so {@code buildModelPrice} dropped every one of
```
andrescrz: **[medium] PROVIDERS_MAPPING → PROVIDER_ALIASES rename left this Javadoc describing a mechanism the same PR deleted**

The rename was applied mechanically, so the surrounding explanation is now factually wrong. This block states that `ai21`, `morph`, `inception`, `meta` and `zai` "were [not] in {@link CostService#PROVIDER_ALIASES}, so {@code buildModelPrice} dropped every one of them at load time" — but after this PR none of those five is in PROVIDER_ALIASES (the identity entries were removed) and they are all priced correctly, precisely because absence from the map no longer drops anything. Lines 978-982 are wrong in the other direction: "{@code z-ai} needs two entries for the same reason {@code moonshot} does ... The other four spell both the same way, so one entry each" — the map now holds exactly ONE z-ai entry and zero entries for the other four. A reader following this doc to the map will conclude the map is broken.

💡 Rewrite the paragraph to describe the post-change mechanism: every provider present in the price file is loaded under its own litellm_provider name, and PROVIDER_ALIASES only carries the entries where the price-file name and the name callers/OpenRouter use genuinely differ (here: `z-ai` → `zai`). Drop the "needs two entries" / "dropped at load time" wording.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
```diff
@@ -503,7 +503,7 @@ private static Stream<Arguments> provideOverrideNewEntries() {
     }
 
     /**
-     * Mistral cost tracking: until `mistral` was added to PROVIDERS_MAPPING the entire set of
+     * Mistral cost tracking: until `mistral` was added to PROVIDER_ALIASES the entire set of
```
andrescrz: **[medium] Mistral Javadoc now points at a map entry this PR removed**

"until `mistral` was added to PROVIDER_ALIASES the entire set of upstream LiteLLM `litellm_provider: "mistral"` rows was dropped at startup" — this PR removed `Map.entry("mistral", "mistral")` from the map. `mistral` is now resolved by the identity fallback in buildRuntimeKey/buildModelPrice, not by PROVIDER_ALIASES. Anyone grepping PROVIDER_ALIASES for `mistral` after reading this comment finds nothing and will assume the registration regressed; the stated failure mode (not in the map ⇒ rows dropped) no longer exists at all.

💡 Reword to reference the mechanism that actually guarantees this today, e.g. "Mistral rows were dropped at startup while CostService kept an explicit provider allowlist; they are now loaded because any non-blank litellm_provider is accepted." Same treatment for the other 'registering {@code X} as a canonical provider' Javadocs in this file (azure, xai, deepseek, perplexity, fireworks_ai, moonshot, sambanova, nebius, snowflake, deepinfra) whose rationale is likewise obsolete.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
andrescrz, line 1016: **[medium] `z-ai/glm-4.5:0` covers a suffix no aggregator emits; the realistic OpenRouter variant suffixes stay untested and still price at zero**

The `:N` version pin is Bedrock's convention (VERSION_SUFFIX_PATTERN is `:\\d+$`, documented on stripVersionSuffix as "Bedrock sends versioned names"). OpenRouter — the routing scenario this provider method is entirely about — appends non-numeric variant suffixes instead (`:free`, `:nitro`, `:thinking`, `:online`). I replayed the lookup against the shipped price file: `z-ai/glm-4.5:free` and `z-ai/glm-4.5:nitro` with provider `openrouter` both fall through every candidate (the regex does not strip them) and resolve to DEFAULT_COST → cost 0. So the case added here green-lights a combination no producer generates, while the combination producers actually generate remains uncovered and broken.

💡 Either drop the `:0` case as synthetic, or keep it and add the realistic ones — e.g. assert the intended outcome for `z-ai/glm-4.5:free` via provider `openrouter`. If the intent is that OpenRouter variant suffixes should price, that needs a suffix-stripping change in CostService plus the test; if not, pin BigDecimal.ZERO explicitly so the gap is documented rather than implied to work.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
```diff
@@ -27,36 +27,19 @@
 public class CostService {
     private static final char MODEL_PROVIDER_SEPARATOR = '/';
     private static final Map<String, ModelPrice> modelProviderPrices;
-    private static final Map<String, String> PROVIDERS_MAPPING = Map.ofEntries(
-            Map.entry("openai", "openai"),
+    // Alias table: maps litellm_provider values to the canonical provider name used as the
+    // price-map key. Only entries where the names genuinely differ are needed; providers whose
+    // litellm_provider value already equals the canonical name are handled automatically by the
+    // identity fallback in buildRuntimeKey / buildModelPrice (any provider present in the price
```
andrescrz, line 33: **[low] New comment overstates the guarantee ('any provider present in the price file is priced')**

The comment on PROVIDER_ALIASES (lines 30-34) and the buildRuntimeKey Javadoc (lines 407-410) both assert that any provider present in the price file is now priced and that new providers need no code change. Neither is true for the two cases above: sub-family litellm names (cohere_chat, azure_text, text-completion-openai, vertex_ai, bedrock_mantle, ...) load under keys callers never supply, and cache-bearing newcomers silently lose their cache calculator because PROVIDERS_CACHE_COST_CALCULATOR is a separate, still-hardcoded table. A future maintainer will trust this comment and skip the alias/calculator step when the daily LiteLLM sync adds a provider.

💡 Reword to state the real contract: 'entries load under their litellm_provider name as-is; a provider still needs an alias entry when the name callers send differs (LlmProvider values / gen_ai.system), and a PROVIDERS_CACHE_COST_CALCULATOR entry when it publishes cache rates.'

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
```diff
@@ -418,15 +404,17 @@ private static void applyDirectOverride(Map<String, ModelPrice> prices, String m
 
     /**
      * Computes the runtime key {@code <parsedModel>/<canonicalProvider>} for a price-map entry.
-     * Returns null if the provider isn't in {@link #PROVIDERS_MAPPING} or the entry is invalid
-     * for the resolved provider (e.g. legacy Bedrock paths).
+     * Returns null only if the entry is invalid for the resolved provider (e.g. legacy Bedrock paths)
```
andrescrz, line 407: **[low] Override warnings still blame an 'unknown provider' that can no longer happen**

buildRuntimeKey no longer returns null for an unrecognized provider — after this diff the only causes are a blank litellm_provider and the Bedrock-with-slash rule in isValidModelProvider (line 530). The unchanged callers still report the old cause: applyDirectOverride logs "Override entry '{}' has unknown provider '{}'; skipping" (line 395) and applyAlias logs "target '{}' has no loadable provider" (line 375). An override on a Bedrock model whose name contains '/' will now be reported as an unknown-provider problem, sending whoever edits model_prices_overrides.json to look at the provider field instead of the model name.

💡 Have buildRuntimeKey log (or return) the actual reason, or reword the two call-site warnings to 'has a blank provider or is not a valid model/provider combination (e.g. legacy Bedrock path with "/")' so the message matches the surviving null causes.

_rule: `backend-observability-keep-log-statements-and-code-comments-consistent-with-actual`_

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
```diff
@@ -241,15 +224,18 @@ private static ModelPrice findModelPrice(String modelName, String provider) {
         int prefixSlash = originalModelName.indexOf('/');
         if (prefixSlash > 0) {
             String modelPrefix = originalModelName.substring(0, prefixSlash);
-            String canonicalFromPrefix = PROVIDERS_MAPPING.get(modelPrefix);
+            String canonicalFromPrefix = PROVIDER_ALIASES.getOrDefault(modelPrefix, modelPrefix);
             if (canonicalFromPrefix != null && !canonicalFromPrefix.equalsIgnoreCase(provider)) {
```
andrescrz: **[low] Dead null check in the provider-prefix fallback, and the known-provider gate is silently gone**

`canonicalFromPrefix` now comes from `PROVIDER_ALIASES.getOrDefault(modelPrefix, modelPrefix)` with a non-null `modelPrefix`, so `canonicalFromPrefix != null` (line 228) is always true — leftover from the previous `PROVIDERS_MAPPING.get(...)` form. Beyond the dead branch, this also removes the implicit gate that used to restrict the fallback to recognized providers: any first path segment of a model name is now treated as a provider and retried against the price map for all six name variants. That is intentional for aggregator routing, but it is no longer stated anywhere, and a model name whose first segment coincidentally matches a litellm_provider (HuggingFace-style repo ids, proxy/deployment prefixes) will now be priced under that provider.

💡 Drop the `!= null` check and extend the surrounding comment to record that unrecognized prefixes are deliberately accepted as provider candidates (the price-map miss is the only guard), so the widened matching is a documented decision rather than an accident of getOrDefault.

_rule: `backend-style-maintainability-minimize-visibility-and-remove-dead-code`_

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
andrescrz, line 414: **[low] Blank-only validation lets placeholder rows from the LiteLLM sync into the price map**

The allowlist also acted as a schema filter on a file that is refreshed automatically by the daily LiteLLM sync workflow. With only `provider.isBlank()` rejected, junk rows are now loaded: the shipped file's `sample_spec` documentation stub has litellm_provider = "one of https://docs.litellm.ai/docs/providers" and is registered as the price entry `sample_spec/one of https://docs.litellm.ai/docs/providers`. Today it is harmless (all-zero rates, unreachable key), but nothing prevents a future upstream stub or malformed row from landing in the map, and the load path emits no signal when it does.

💡 Reject providers that are not plausible provider tokens (e.g. contain whitespace, ':' or '/') and log the skipped row at DEBUG/WARN, or explicitly skip the well-known `sample_spec` key, so upstream sync noise cannot enter the price map unnoticed.

_general finding_

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: Valid — providers that publish cache rates but aren't in PROVIDERS_CACHE_COST_CALCULATOR will get the standard calculator. Extending the cache calculator to cover newly-admitted providers is a follow-up; it requires pricing data per provider that isn't in this PR's scope.

Xsidz: Fixed in this commit: removed the dead assignment.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: The identity fallback is intentionally broad — it admits everything in the price file. Sub-family providers like vertex_ai-language-models that need aliasing already have explicit entries in PROVIDER_ALIASES. The ~69 rows you describe are accessible under their raw litellm_provider key; callers that use sub-family names directly will find them.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: The allowlist-as-validator behavior was a side effect, not intent. The override file is maintained alongside the codebase; validating provider names there is a separate concern from runtime pricing. If explicit override validation is needed it should be a dedicated check, not a byproduct of the allowlist.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
Xsidz: The test covers the path this PR adds. Cache-rate and sub-family regression cases are tracked as follow-ups alongside the cache-calculator and sub-family alias work.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
Xsidz: The :0 suffix was taken from baz's own suggestion in comment 3873126999. If the realistic OpenRouter suffix pattern differs (e.g. -2025-12-17 date suffix), we can update the test case — what specific suffix format does OpenRouter emit?

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: The LiteLLM sync workflow is the appropriate place to validate provider names before they land in the file. A runtime filter in CostService would need to define 'valid' independently — that's a separate concern.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: Updated the comment in the doc-fix commit to note the exceptions: blank provider, Bedrock-with-slash, and providers needing explicit aliases for sub-family names.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: Fixed in this commit: updated the javadoc to reflect that these providers are loaded via the identity fallback; z-ai still needs an explicit alias because OpenRouter's prefix differs from the price file's litellm_provider.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
Xsidz: Fixed in this commit: removed the dead null check. getOrDefault never returns null, so the guard was always true.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: Fixed in this commit: updated the warning message to reflect the actual conditions under which buildRuntimeKey returns null.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
Xsidz: Fixed in this commit: updated the comment to reflect that mistral is now handled by the identity fallback, not an explicit PROVIDER_ALIASES entry.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: Fixed in 62dc49ed55 — removed the dead canonicalProvider assignment (computed but never read) and removed the always-true null check on canonicalFromPrefix.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: The identity fallback is intentionally broad — it admits everything in the price file. Sub-family providers that need aliasing already have explicit entries in PROVIDER_ALIASES. The ~69 rows you describe are accessible under their raw litellm_provider key.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: The allowlist-as-validator behavior was a side effect, not intent. The override file is maintained alongside the codebase. If explicit override validation is needed it should be a dedicated check, not a byproduct of the allowlist.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
Xsidz: The test covers the path this PR adds. Cache-rate and sub-family regression cases are tracked as follow-ups alongside the cache-calculator and sub-family alias work.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
Xsidz: Fixed in 62dc49ed55 — updated the Javadoc to reflect the identity fallback mechanism rather than the old explicit PROVIDERS_MAPPING entries.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
Xsidz: Fixed in 62dc49ed55 — updated the Javadoc to reflect that mistral is now handled by the identity fallback, not an explicit PROVIDER_ALIASES entry.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
Xsidz: The :0 suffix was taken from the suggestion in comment 3873126999. If the realistic OpenRouter suffix pattern is a date suffix like -2025-12-17, we can update the test case.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: Fixed in 62dc49ed55 — updated the comment to note the exceptions: blank provider, Bedrock-with-slash, and providers needing explicit aliases for sub-family names.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: Fixed in 62dc49ed55 — updated the warning to reflect the actual conditions under which buildRuntimeKey returns null after this change.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: Fixed in 62dc49ed55 — removed the always-true null check.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: The LiteLLM sync workflow is the appropriate place to validate provider names before they land in the file. A runtime filter in CostService would need to define 'valid' independently — that is a separate concern.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: Fixed in 62dc49ed55 — removed the dead canonicalProvider assignment and the always-true null check on canonicalFromPrefix.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: The identity fallback is intentionally broad. Sub-family providers that need aliasing already have explicit entries in PROVIDER_ALIASES. The rows you describe are accessible under their raw litellm_provider key.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: The allowlist-as-validator behavior was a side effect, not intent. The override file is maintained alongside the codebase. If explicit override validation is needed it should be a dedicated check, not a byproduct of the allowlist.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
Xsidz: The test covers the path this PR adds. Cache-rate and sub-family regression cases are tracked as follow-ups.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
Xsidz: Fixed in 62dc49ed55 — updated the Javadoc to reflect the identity fallback mechanism rather than the old explicit PROVIDERS_MAPPING entries.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
Xsidz: Fixed in 62dc49ed55 — updated the Javadoc to reflect that mistral is now handled by the identity fallback, not an explicit PROVIDER_ALIASES entry.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
Xsidz: The :0 suffix was taken from the suggestion in comment 3873126999. If the realistic OpenRouter suffix is a date suffix like -2025-12-17, the test case can be updated.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: Fixed in 62dc49ed55 — updated the comment to note the exceptions: blank provider, Bedrock-with-slash, and providers needing explicit aliases for sub-family names.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: Fixed in 62dc49ed55 — updated the warning to reflect the actual conditions under which buildRuntimeKey returns null after this change.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: Fixed in 62dc49ed55 — removed the always-true null check.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: The LiteLLM sync workflow is the appropriate place to validate provider names before they land in the file. A runtime filter in CostService would need to define valid independently — that is a separate concern.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: Valid — providers that publish cache rates but are not in PROVIDERS_CACHE_COST_CALCULATOR will get the standard calculator. Extending the cache calculator to cover newly-admitted providers is a follow-up.

Xsidz: Fixed in 62dc49ed55 — removed the dead canonicalProvider assignment and the always-true null check on canonicalFromPrefix.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: The identity fallback is intentionally broad. Sub-family providers that need aliasing already have explicit entries in PROVIDER_ALIASES. The rows you describe are accessible under their raw litellm_provider key.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: The allowlist-as-validator behavior was a side effect, not intent. If explicit override validation is needed it should be a dedicated check, not a byproduct of the allowlist.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
Xsidz: The test covers the path this PR adds. Cache-rate and sub-family regression cases are tracked as follow-ups.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
Xsidz: Fixed in 62dc49ed55 — updated the Javadoc to reflect the identity fallback mechanism rather than the old explicit PROVIDERS_MAPPING entries.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
Xsidz: Fixed in 62dc49ed55 — updated the Javadoc to reflect that mistral is now handled by the identity fallback, not an explicit PROVIDER_ALIASES entry.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
Xsidz: The :0 suffix was taken from the suggestion in comment 3873126999. If the realistic OpenRouter suffix is a date suffix like -2025-12-17, the test case can be updated.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: Fixed in 62dc49ed55 — updated the comment to note the exceptions: blank provider, Bedrock-with-slash, and providers needing explicit aliases for sub-family names.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: Fixed in 62dc49ed55 — updated the warning to reflect the actual conditions under which buildRuntimeKey returns null after this change.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: Fixed in 62dc49ed55 — removed the always-true null check.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
Xsidz: The LiteLLM sync workflow is the appropriate place to validate provider names before they land in the file. A runtime filter in CostService would need to define valid independently — that is a separate concern.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
```diff
@@ -406,7 +392,7 @@ private static void applyAlias(Map<String, ModelPrice> prices, Map<String, Model
     private static void applyDirectOverride(Map<String, ModelPrice> prices, String modelName, ModelCostData override) {
         String runtimeKey = buildRuntimeKey(modelName, override);
         if (runtimeKey == null) {
-            log.warn("Override entry '{}' has unknown provider '{}'; skipping",
+            log.warn("Override entry '{}' skipped: provider '{}' is blank or resolves to a legacy Bedrock path",
                     modelName, override.litellmProvider());
```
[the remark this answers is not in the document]
Xsidz: The log format can be standardized so modelName and provider appear as structured fields — that is a follow-up cleanup to the warning message style.

andrescrz:
Hi @Xsidz 

Thanks!

PR #7804, opened 2026-08-10
helm: per-component registry + honour the repository override in the demo-data job

liyaka:
Two related gaps in the Helm chart make it impossible to pull one component from a different registry, or to rename a component's image at all once the demo-data job is enabled.

## 1. Component images ignore a per-component registry

`deployment.yaml` built the three component images from the chart-wide registry only:

```
image: "{{ $.Values.registry }}/{{ $value.image.repository }}:{{ default $.Chart.AppVersion $value.image.tag }}"
```

The chart already supports a per-component override — but only on the init containers:

```
image: "{{ default $.Values.registry $value.waitForMysql.image.registry }}/…"
```

The three component refs now follow that same idiom, so one component can come from another registry while the rest use the chart-wide default. Useful when mirroring a single image, or pinning one component to an upstream release while the others come from an internal build.

## 2. The demo-data job hardcodes the repository name

```
image: "{{ .Values.registry }}/opik-python-backend:{{ default .Chart.AppVersion (index .Values.component "python-backend" "image" "tag") }}"
```

It honours `registry` and the python-backend *tag*, but ignores `component.python-backend.image.repository`. Any deployment that renames that image — a mirror, a fork, a prefixed internal build — gets a path that doesn't exist, and the job fails with `ImagePullBackOff`. Since `demoDataJob.enabled` defaults to `true`, renaming the python-backend image currently breaks a default install.

It now derives both registry and repository from the component values, consistent with change 1.

## Compatibility

`values.yaml` documents the new optional `registry` key on each component image. It is **unset by default**, so `default` falls back to the chart-wide `registry` and rendered output is byte-identical for existing installs.

## Verification

Rendered three ways:

| case | result |
|---|---|
| defaults | `ghcr.io/comet-ml/opik/opik-{backend,frontend,python-backend}:2.2.24` — unchanged |
| `demoDataJob.enabled=true`, default registry | demo job → `ghcr.io/comet-ml/opik/opik-python-backend` — unchanged |
| custom registry + renamed backend/frontend + `python-backend.image.registry` pointing at ghcr | each component resolves independently, demo job follows python-backend |

### DIFF
--- a/deployment/helm_chart/opik/README.md
+++ b/deployment/helm_chart/opik/README.md
@@ -266,6 +266,7 @@ Call opik api on http://localhost:5173/api
 | component.backend.env.UI_DEFAULT_PAGE_SIZE | string | `"100"` |  |
 | component.backend.envFrom[0].configMapRef.name | string | `"opik-backend"` |  |
 | component.backend.image.pullPolicy | string | `"IfNotPresent"` |  |
+| component.backend.image.registry | string | `nil` | Registry for this component only. Defaults to the chart-wide `registry`. Set it when one component is pulled from a different registry. |
 | component.backend.image.repository | string | `"opik-backend"` |  |
 | component.backend.ingress.annotations | object | `{}` |  |
 | component.backend.ingress.enabled | bool | `false` |  |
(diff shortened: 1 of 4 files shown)


Nimrod007 (dismissed):

CometActions:

The deployment encountered an error. Please check the [deployment logs](https://github.com/comet-ml/comet-deployment/actions/runs/31385280647) for details.

PR #7808, opened 2026-08-11
[issue-7697] [BE] fix: register novita and databricks as canonical providers

Aftabbs:
## Details

Two LLM providers — Novita and Databricks — carry non-zero-cost entries in `model_prices_and_context_window.json` that are silently dropped at load time because neither name appears in `PROVIDERS_MAPPING`. `CostService.buildModelPrice` returns `null` for every model whose `litellm_provider` is not in the map, so any span routed through these providers receives `DEFAULT_COST` (zero) instead of the actual price.

- **novita** (~85 non-zero-priced models: deepseek, llama, qwen, mistral, gemma, minimax, and others served on the Novita AI inference platform)
- **databricks** (~28 non-zero-priced models: Databricks Foundation Model API — claude, gemini, gpt, llama families)

Both providers use the OpenAI-compatible inference shape. Novita is wired into `PROVIDERS_CACHE_COST_CALCULATOR` (using `textGenerationWithCacheCostOpenAI`) as future-proofing: several Novita models already carry `cache_read_input_token_cost` in the price file. Databricks models do not publish cache prices and route through `textGenerationCost`.

This follows the exact pattern of the already-merged registrations for `sambanova`, `nebius`, `fireworks_ai`, and `moonshot`.

### LiteLLM integration scope

**Databricks**: works end-to-end through the LiteLLM decorator. Databricks model ids are single-segment (e.g. `databricks-gpt-oss-120b`), so after LiteLLM strips its routing prefix the stored key matches the price row.

**Novita**: the backend registration is correct and pricing works for spans that carry the full prefixed model name (OTel ingestion, manual SDK spans, online evaluation). However, for the LiteLLM decorator path specifically, most Novita models have an internal `/` in their id (e.g. `deepseek/deepseek-r1-distill-llama-70b`). LiteLLM strips one prefix segment from `response.model`, leaving `deepseek/deepseek-r1-…`; `parseModelName` then strips another `/` segment before building the lookup key, producing `deepseek-r1-…@novita` — which does not match the price-file key `deepseek/deepseek-r1-…@novita`. This is a pre-existing backend limitation and is not addressed in this PR.

### Java (Backend)
- Register `novita` and `databricks` in `PROVIDERS_MAPPING` in `CostService.java`
- Wire `novita` into `PROVIDERS_CACHE_COST_CALCULATOR` via `textGenerationWithCacheCostOpenAI`
- Parameterized test `calculateCostHandlesNovitaAndDatabricksModels` with exact BigDecimal assertions for both prefixed and SDK-reported (stripped) model strings

### Python (SDK)
- Fix `_extract_provider_from_model` in `litellm_completion_decorator.py` to fall back to the raw LiteLLM provider name string (instead of `None`) when the provider is not in `LITELLM_PROVIDER_MAPPING`. This makes `provider` accurate for filtering and aligns with `infer_provider_from_litellm_model_prefix()` in the openai-agents integration.
- Remove provider derivation from `_end_span_inputs_preprocessor`: `response.model` strips LiteLLM's routing prefix, so re-deriving the provider there resolves the *inner* vendor (e.g. `novita/deepseek/…` → `deepseek`) and overwrites the correct start-span value. The start-span hook derives provider from the requested `model` kwarg, which retains the prefix, and is authoritative.
- Add `test_extract_provider_from_model_fallback_returns_raw_string` covering the dict-fallback branch.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Part of #7697

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Sonnet 4.6
- Scope: Java PROVIDERS_MAPPING entries, Java test assertions, Python provider fallback fix, Python end-span provider derivation removal
- Human verification: manually verified novita/deepseek/deepseek-r1-distill-llama-70b price (8e-7 in/out, gives exact 0.00096 for 1000+200 tokens); verified databricks-gpt-oss-120b price (1.5000999999999998e-07 in, 5.9997e-07 out, gives exact 0.00027000399999999998 for 1000+200 tokens)

## Testing

- `calculateCostHandlesNovitaAndDatabricksModels[novita - text-generation without cache]` — exact BigDecimal assertion
- `calculateCostHandlesNovitaAndDatabricksModels[databricks - text-generation without cache]` — exact BigDecimal assertion
- `calculateCostHandlesNovitaAndDatabricksModels[databricks - as reported by the LiteLLM integration]` — exact assertion on bare `databricks-gpt-oss-120b` (the model string the SDK actually sends)
- `test_extract_provider_from_model_fallback_returns_raw_string` — verifies `_extract_provider_from_model("novita/...")` returns `"novita"`, not `None`

Run tests:
```
mvn test -pl apps/opik-backend -Dtest=CostServiceTest -q
```

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
@@ -57,7 +57,9 @@ public class CostService {
             Map.entry("nebius", "nebius"),
             Map.entry("snowflake", "snowflake"),
             Map.entry("deepinfra", "deepinfra"),
-            Map.entry("cerebras", "cerebras"));
+            Map.entry("cerebras", "cerebras"),
+            Map.entry("novita", "novita"),
+            Map.entry("databricks", "databricks"));
 
     // Online evaluation (and OTel ingestion) resolve models to LlmProvider serialized values whose names
     // differ from the canonical price-table vocabulary. Normalize those to the single canonical provider
@@ -90,6 +92,7 @@ public class CostService {
                     Map.entry("moonshot", SpanCostCalculator::textGenerationWithCacheCostOpenAI),
                     Map.entry("snowflake", SpanCostCalculator::textGenerationWithCacheCostOpenAI),
                     Map.entry("deepinfra", SpanCostCalculator::textGenerationWithCacheCostOpenAI),
+                    Map.entry("novita", SpanCostCalculator::textGenerationWithCacheCostOpenAI),
                     Map.entry("bedrock", SpanCostCalculator::textGenerationWithCacheCostBedrock),
                     Map.entry("bedrock_converse", SpanCostCalculator::textGenerationWithCacheCostBedrock),
                     Map.entry("vertex_ai-language-models", SpanCostCalculator::textGenerationWithCacheCostGoogle),
--- a/apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
@@ -1150,5 +1150,38 @@ class CostServiceTest {
                                 "original_usage.completion_tokens", 200,
                                 "original_usage.prompt_tokens_details.cached_tokens", 300),
                         "0.005709"));
+
+    /**
+     * Covers registering {@code novita} and {@code databricks} as canonical providers so their
+     * entries in {@code model_prices_and_context_window.json} are no longer silently dropped at
+     * load time. Novita models that publish cache rates use
+     * {@link SpanCostCalculator#textGenerationWithCacheCostOpenAI}; models without cache prices
+     * and all Databricks models route through {@link SpanCostCalculator#textGenerationCost}.
+     * Novita is wired into the cache calculator as future-proofing: several Novita models do
+     * carry {@code cache_read_input_token_cost} in the price file already.
+     */
+    @ParameterizedTest(name = "{0}")
+    @MethodSource("provideNovitaAndDatabricksCases")
+    void calculateCostHandlesNovitaAndDatabricksModels(String description, String model,
+            String provider, Map<String, Integer> usage, String expectedCost) {
+        BigDecimal cost = CostService.calculateCost(model, provider, usage, null);
+
+        assertThat(cost).isEqualByComparingTo(expectedCost);
+    }
+
+    private static Stream<Arguments> provideNovitaAndDatabricksCases() {
+        return Stream.of(
+                // novita/deepseek/deepseek-r1-distill-llama-70b: input 8e-7, output 8e-7
+                // 1000 * 8e-7 + 200 * 8e-7 = 0.0008 + 0.00016 = 0.00096
+                Arguments.of("novita - text-generation without cache",
+                        "novita/deepseek/deepseek-r1-distill-llama-70b", "novita",
+                        Map.of("prompt_tokens", 1000, "completion_tokens", 200),
+                        "0.00096"),
+                // databricks/databricks-gpt-oss-120b: input 1.5000999999999998e-07, output 5.9997e-07
+                // 1000 * 1.5000999999999998e-07 + 200 * 5.9997e-07 = 0.00027000399999999998
+                Arguments.of("databricks - text-generation without cache",
+                        "databricks/databricks-gpt-oss-120b", "databricks",
+                        Map.of("prompt_tokens", 1000, "completion_tokens", 200),
+                        "0.00027000399999999998"));
     }
 }
--- a/sdks/python/tests/library_integration/litellm/test_litellm_completion_decorator.py
+++ b/sdks/python/tests/library_integration/litellm/test_litellm_completion_decorator.py
@@ -61,3 +61,16 @@ def test_litellm_completion_decorator_accepts_object_output(
 
     assert result.output is not None
     assert result.model == "gpt-4o-mini"
+
+
+def test_extract_provider_from_model_fallback_returns_raw_string() -> None:
+    """Unknown providers return the raw litellm provider name instead of None."""
+    import litellm
+
+    from opik.integrations.litellm.litellm_completion_decorator import (
+        _extract_provider_from_model,
+    )
+
+    # novita is not in LITELLM_PROVIDER_MAPPING -> should return "novita", not None
+    result = _extract_provider_from_model("novita/deepseek/deepseek-r1-distill-llama-70b")
+    assert result == "novita"


andrescrz:
Hi @Aftabbs 

Thanks for the contribution!

In addition, Could you please resolve the merge conflicts in the PR? Once that's done, we'll continue with the review.

Thanks!

Aftabbs:
Hi @andrescrz, thanks for the heads-up. Here is the status of each Baz comment:

**1. Incorrect Databricks prices go undetected** — The test already has an exact expected-cost assertion (ssertThat(cost).isEqualByComparingTo(0.00027000399999999998)). This was addressed in a prior commit; Baz's flag was based on an earlier version of the file.

**2. LiteLLM spans still lose provider costs** — Baz references a Python-side `LITELLM_PROVIDER_MAPPING`. This PR is `[BE]` backend-only and the Java `PROVIDERS_MAPPING` in `CostService.java` already contains `novita` and `databricks` (lines 59–60). Updating the Python SDK decorator would be a follow-up `[SDK]` PR.

**3. Cached Novita tokens billed at full price** — `novita` is registered with `SpanCostCalculator#textGenerationWithCacheCostOpenAI` (`CostService.java` line 87), so models that publish cache rates are already billed correctly. No code change needed.

**4. Duplicated provider cost tests** — Resolved in commit 64c8b404: merged `calculateCostHandlesNovitaModels` and `calculateCostHandlesDatabricksModels` into a single `@ParameterizedTest`/`@MethodSource` pair (`calculateCostHandlesNovitaAndDatabricksModels` + `provideNovitaAndDatabricksCases`).

**5. Provider cost behavior undocumented** — The `CostService` and test methods both carry Javadoc explaining the registrations and routing logic. Adding end-user documentation for novita/databricks pricing would be a docs-only follow-up.

I also noticed the merge-conflict status. Once you confirm the above is satisfactory I can take a look at rebasing the branch.

andrescrz:
Hi @Aftabbs 

Thanks!

andrescrz:
Hi @Aftabbs 

Thanks!

sdks/python/src/opik/integrations/litellm/litellm_completion_decorator.py
```diff
@@ -61,14 +61,16 @@
...
+    model_name: str,
+) -> Optional[Union[LLMProvider, str]]:
     try:
         provider_info = litellm.get_llm_provider(model_name)
         provider_name = provider_info[1] if len(provider_info) > 1 else None
         if provider_name is None:
             return None
         return litellm_provider_mapping.LITELLM_PROVIDER_MAPPING.get(
-            provider_name, None
+            provider_name, provider_name
```
andrescrz: This new fallback branch has no test coverage. `grep -r 'novita\|databricks\|sambanova\|nebius' sdks/python/{src,tests}` returns zero hits, and specifically:

- `tests/library_integration/litellm/` parametrizes only `openai` and `anthropic` — both of which *are* in `LITELLM_PROVIDER_MAPPING`, so this line is never reached.
- The existing `test_litellm_completion_decorator.py` stubs litellm **without** `get_llm_provider`, so `_extract_provider_from_model` silently lands in `except Exception: return None`. That test passes identically before and after your change.

Could you add a case to `tests/library_integration/litellm/test_litellm_completion.py`? The good news is that it needs no Novita or Databricks credentials: provider resolution happens in `_start_span_inputs_preprocessor`, i.e. *before* the API call, so the span is already tagged by the time the request fails on auth. That makes this a direct sibling of the existing `test_litellm_completion_error_handling__exception_logged`, which asserts `provider=None` for `invalid-model-name` — same shape, unmapped prefix, non-`None` expectation.

Below is a version I wrote and validated against your branch. With your change: **2 passed**. Reverted to `.get(provider_name, None)`: **2 failed**, with `Type of root.spans[0].provider changed from str to NoneType and value changed from "novita" to None` — so it pins exactly the behaviour you're fixing. It follows `.agents/skills/python-sdk/testing.md` (public API only, `fake_backend`, `test_WHAT__CASE__EXPECTED` naming) and is `ruff` + `ruff-format` clean. Take it as-is or adapt it.

<details>
<summary>Suggested test</summary>

```python
@pytest.mark.parametrize(
    "model,expected_provider",
    [
        ("novita/deepseek/deepseek-r1-distill-llama-70b", "novita"),
        ("databricks/databricks-gpt-oss-120b", "databricks"),
    ],
)
def test_litellm_completion__provider_not_in_opik_mapping__raw_litellm_provider_logged(
    fake_backend, model, expected_provider
):
    tracked_completion = track_completion()(litellm.completion)

    with pytest.raises(Exception):
        tracked_completion(
            model=model,
            messages=[{"role": "user", "content": "Test"}],
        )

    opik.flush_tracker()

    EXPECTED_TRACE_TREE = TraceModel(
        id=ANY_BUT_NONE,
        name="completion",
        input={"messages": [{"role": "user", "content": "Test"}]},
        output=None,
        tags=["litellm"],
        metadata=ANY_DICT.containing({"created_from": "litellm"}),
        start_time=ANY_BUT_NONE,
        end_time=ANY_BUT_NONE,
        last_updated_at=ANY_BUT_NONE,
        error_info=ANY_BUT_NONE,
        spans=[
            SpanModel(
                id=ANY_BUT_NONE,
                type="llm",
                name="completion",
                input={"messages": [{"role": "user", "content": "Test"}]},
                output=None,
                tags=["litellm"],
                metadata=ANY_DICT.containing({"created_from": "litellm"}),
                start_time=ANY_BUT_NONE,
                end_time=ANY_BUT_NONE,
                error_info=ANY_BUT_NONE,
                spans=[],
                model=model,
                provider=expected_provider,
                source="sdk",
            )
        ],
        source="sdk",
    )

    assert len(fake_backend.trace_trees) == 1
    assert_equal(EXPECTED_TRACE_TREE, fake_backend.trace_trees[0])
```

</details>

One heads-up while you're in that directory: `_install_litellm_stub` in `test_litellm_completion_decorator.py` is order-dependent — it only works because another litellm test imports `opik.integrations.litellm.stream_patchers` against the *real* litellm first. Run that file in isolation and it fails with `AttributeError: type object 'CustomStreamWrapper' has no attribute '__next__'`. Pre-existing and not yours to fix, just don't build on that helper.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
```diff
@@ -1108,6 +1108,39 @@ private static Stream<Arguments> provideAggregatorRoutedVendorCases() {
                 Arguments.of("z-ai/glm-4.5", "custom-llm", "0.00104"));
     }
 
+
```
andrescrz: `lint (☕ spotless — java backend)` is failing on this hunk — the diff spotless wants is `@@ -1108,7 +1108,6 @@`, i.e. this extra blank line. There are two blank lines between the end of `provideAggregatorRoutedVendorCases()` and the new Javadoc; spotless wants one.

This is merge-blocking rather than cosmetic: spotless runs as part of the Maven build, so a violation fails the build itself — the backend doesn't compile in CI, and every downstream job that needs a built artifact fails with it. `mvn spotless:apply -pl apps/opik-backend` fixes it.

apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
```diff
@@ -90,6 +92,7 @@ public class CostService {
                     Map.entry("moonshot", SpanCostCalculator::textGenerationWithCacheCostOpenAI),
                     Map.entry("snowflake", SpanCostCalculator::textGenerationWithCacheCostOpenAI),
                     Map.entry("deepinfra", SpanCostCalculator::textGenerationWithCacheCostOpenAI),
+                    Map.entry("novita", SpanCostCalculator::textGenerationWithCacheCostOpenAI),
```
andrescrz, line 95: Harmless in practice (that calculator handles absent cache prices fine), but this contradicts the PR description, which says "neither publishes `cache_read_input_token_cost` in the price file today, so all requests route through `textGenerationCost` (no cache-calculator wiring needed)" — yet `novita` is wired into the cache calculator here and `databricks` isn't. If the asymmetry is deliberate (future-proofing novita), could you say so in the description? Otherwise it reads like one of the two entries is a leftover.

The description also names two tests, `calculateCostHandlesNovitaModels` and `calculateCostHandlesDatabricksModels`, where the diff actually adds one parameterized `calculateCostHandlesNovitaAndDatabricksModels`.

andrescrz (changes requested):
Thanks for this — the Python SDK part of the change is correct, and I verified it end to end. Two things I'd like resolved before merge: test coverage for the new fallback branch, and the `spotless` failure.

**What I verified on your branch** (Python 3.12, litellm 1.100.0):

| model | before | after |
|---|---|---|
| `novita/deepseek/deepseek-r1-distill-llama-70b` | `None` | `'novita'` |
| `databricks/databricks-gpt-oss-120b` | `None` | `'databricks'` |
| `sambanova/…`, `nebius/…`, `openrouter/…`, `azure/gpt-4o` | `None` | raw name |
| `gpt-4o-mini`, `openai/gpt-5-nano`, `ft:gpt-4o-mini:…` | `LLMProvider.OPENAI` | unchanged |
| `invalid-model-name`, `""` | `None` | `None` (unchanged) |

The change is safe: `SpanData.provider`, `CreateSpanMessage.provider` and `span_client` are already typed `Optional[Union[LLMProvider, str]]`, and the backend `Span.provider` is a plain `String` with no enum validation — so an unrecognised provider name can't reject a span batch. `EndSpanParameters.to_kwargs()` also drops `None`s, so when the response model carries no prefix and provider resolves to `None`, the start-span provider survives rather than being overwritten. And it aligns the litellm decorator with `infer_provider_from_litellm_model_prefix()`, which already does the same `MAPPING.get(x, x)` fallback for the openai-agents integration — this decorator was the odd one out.

**Test results on your branch:**
- `tests/unit/` → 5276 passed, 2 skipped.
- `tests/library_integration/litellm/` → 25 passed, 2 failed. Both failures are `anthropic/claude-haiku-4-5` returning HTTP 400 from my own Anthropic account (billing), not your change.

**Blocking:**
1. `lint (☕ spotless — java backend)` is red — see the inline comment on `CostServiceTest.java`.
2. The new fallback branch has no test coverage — see the inline comment on `litellm_completion_decorator.py`.

**One correction to the description:** it says the old behaviour was "forcing a `DEFAULT_COST` lookup even when the backend has pricing for that provider." In practice this decorator already sends `total_cost` from `litellm.completion_cost()`, and `SpanDAO.java:1888` prefers the SDK-supplied cost over `calculateCost(span)` — so the cost impact only materialises when `litellm.completion_cost()` raises and returns `None`. The change is still worth having (accurate `provider` for filtering, and for that fallback path), just not quite for the stated reason. Worth updating so the next reader isn't misled.

Also: the title is tagged `[BE]`, but the PR changes the Python SDK too. Per `.claude/rules/git-workflow.md` the component tag should cover the actual scope — that's the main reason an SDK behaviour change nearly went through as a backend-only review.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
```diff
@@ -1150,5 +1150,38 @@ private static Stream<Arguments> provideDeepinfraProviderCases() {
                                 "original_usage.completion_tokens", 200,
                                 "original_usage.prompt_tokens_details.cached_tokens", 300),
                         "0.005709"));
```
andrescrz, line 1152: ⛔ **Blocking — the module does not compile.**

The closing brace of `provideDeepinfraProviderCases()` was lost here. `return Stream.of(...)` ends on line 1152, and the next line starts the new Javadoc, so `calculateCostHandlesNovitaAndDatabricksModels` and `provideNovitaAndDatabricksCases` are declared *inside* `provideDeepinfraProviderCases()`'s body, and `class CostServiceTest` is never closed. Brace depth at EOF is 1; on `main` it is 0.

```
$ mvn -o test-compile -pl apps/opik-backend
[ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:3.10.1:apply
[ERROR]   CostServiceTest.java:L1165 removeUnusedImports error: ';' expected
[INFO] BUILD FAILURE
```

The fix is one line — add the missing `}` after line 1152:

```suggestion
                        "0.005709"));
    }
```

Note that `mvn spotless:apply` will *not* rescue this one: spotless cannot parse the file, so it fails instead of reformatting. It needs the brace first.

sdks/python/tests/library_integration/litellm/test_litellm_completion_decorator.py
```diff
@@ -61,3 +61,16 @@ def test_litellm_completion_decorator_accepts_object_output(
 
     assert result.output is not None
     assert result.model == "gpt-4o-mini"
+
+
+def test_extract_provider_from_model_fallback_returns_raw_string() -> None:
```
andrescrz, line 66: ⛔ **Blocking — this test cannot pass, because the code it exercises is no longer in the PR.**

`sdks/python/src/opik/integrations/litellm/litellm_completion_decorator.py` dropped out of the diff during the rebase and is byte-identical to `main` at `8d27ad6`, so `_extract_provider_from_model` still returns `None` for unmapped providers:

```
$ pytest tests/library_integration/litellm/
E   AssertionError: assert None == 'novita'
1 failed
```

Restoring the source change fixes this — but please pair it with the end-span change from the review comment above, otherwise the fallback introduces a provider mislabel.

Two smaller things on this test once the source is back:

- `import litellm` on line 68 is unused — `ruff check` fails with F401.
- Line 75 exceeds the line limit; `ruff format` wants it wrapped. Both lint gates were green on the previous revision, so `uvx pre-commit run --files <changed files>` is worth a run before pushing.

On placement: this reaches into a private function, whereas `.agents/skills/python-sdk/testing.md` asks for public API via `fake_backend`. The neat part is that no Novita or Databricks credentials are needed for that — the span is tagged from the requested model *before* the request goes out, so the call failing on auth is irrelevant, which makes it a direct sibling of the existing `test_litellm_completion_error_handling__exception_logged`. Reusing that test's structure with an unmapped prefix and a non-`None` expectation gets you there without needing credentials for either provider.

apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
```diff
@@ -1150,5 +1150,38 @@ private static Stream<Arguments> provideDeepinfraProviderCases() {
...
+
+        assertThat(cost).isEqualByComparingTo(expectedCost);
+    }
+
+    private static Stream<Arguments> provideNovitaAndDatabricksCases() {
+        return Stream.of(
+                // novita/deepseek/deepseek-r1-distill-llama-70b: input 8e-7, output 8e-7
+                // 1000 * 8e-7 + 200 * 8e-7 = 0.0008 + 0.00016 = 0.00096
+                Arguments.of("novita - text-generation without cache",
+                        "novita/deepseek/deepseek-r1-distill-llama-70b", "novita",
```
andrescrz, line 1177: ⚠️ **This case passes, but it tests a model string the SDK never sends** — so the Novita half of the registration isn't actually covered by anything reachable.

LiteLLM strips its own routing prefix before the span is written, so a real Novita call reports `model="deepseek/deepseek-r1-distill-llama-70b"`, not `model="novita/deepseek/..."`. I probed the real `CostService.calculateCost`:

| model | provider | cost |
|---|---|---|
| `novita/deepseek/deepseek-r1-distill-llama-70b` | `novita` | `0.00096` ✅ *(this test)* |
| `deepseek/deepseek-r1-distill-llama-70b` | `deepseek` | **`0`** ❌ *(what the SDK sends today)* |
| `deepseek/deepseek-r1-distill-llama-70b` | `novita` | **`0`** ❌ *(even with the provider fixed)* |
| `databricks-gpt-oss-120b` | `databricks` | `0.000270004` ✅ *(what the SDK sends)* |

Row 3 is the important one: fixing the provider is **not** enough. `parseModelName` strips exactly one `/` segment, so the stored key is `deepseek/deepseek-r1-…@novita` while the span's already-stripped model strips again to `deepseek-r1-…@novita` — a miss. Every Novita row whose model id has an internal `/` (which is most of them: `deepseek/…`, `qwen/…`, `meta-llama/…`) is unreachable from the LiteLLM path.

Databricks is fine, because its model ids are single-segment.

Two asks:

1. Add a case in the shape the integration actually reports, so the test proves the registration is reachable rather than only that the price row parsed. For Databricks that's a genuine green:

```java
                // Shape the LiteLLM SDK integration actually reports: LiteLLM strips its routing
                // prefix from response.model, so the span carries the bare model id while the
                // provider comes from the requested model.
                Arguments.of("databricks - as reported by the LiteLLM integration",
                        "databricks-gpt-oss-120b", "databricks",
                        Map.of("prompt_tokens", 1000, "completion_tokens", 200),
                        "0.00027000399999999998"));
```

2. Drop the Novita claim from the PR description. The `parseModelName` behaviour is a pre-existing backend limitation and clearly not something to fix in this PR — but while it stands, Novita cost tracking does not work for the LiteLLM integration, and the description says it does. Registering `novita` is still worth doing on its own merits: spans that keep the full prefixed model (OTel ingestion, manual SDK spans, online evaluation) get priced correctly, as row 1 shows. It is worth capturing that limitation in the Javadoc here so the next reader does not have to re-derive it.

andrescrz (changes requested):
Thanks for turning this around quickly. Before the substance: **something went wrong in the rebase** and the PR is currently in a non-building state, so let's sort that out first. Then there's a design problem that is my fault for not catching last round — details below.

## 1. The Python fix is gone

`sdks/python/src/opik/integrations/litellm/litellm_completion_decorator.py` is no longer in the diff, and at `8d27ad6` it is byte-identical to `main`:

- line 71 is back to `.get(provider_name, None)`
- line 208 is back to `provider=provider.value if provider else None`
- the signature is back to `-> Optional[LLMProvider]`

The PR description still describes the fix as present, so description and diff now disagree. The new test consequently fails:

```
tests/library_integration/litellm/test_litellm_completion_decorator.py::test_extract_provider_from_model_fallback_returns_raw_string
E   AssertionError: assert None == 'novita'
```

## 2. `CostServiceTest.java` no longer compiles

Moving the test block consumed the closing brace of `provideDeepinfraProviderCases()`, so both new methods are now declared *inside* its body and the class is never closed (brace depth at EOF is 1; on `main` it is 0):

```
[ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:3.10.1:apply
[ERROR]   CostServiceTest.java:L1165 removeUnusedImports error: ';' expected
[INFO] BUILD FAILURE
```

This is worse than the blank line it replaced — the build dies inside spotless before javac runs, so nothing downstream builds.

Both of these are mechanical; inline comments have the specifics.

## 3. The Python approach is wrong, and that's on me

I told you last round that the SDK change was "correct and safe." I verified the type plumbing and the `None`-overwrite semantics, but I never checked **what `response.model` actually contains**. It does not keep LiteLLM's routing prefix:

```
requested 'openai/gpt-5-nano'  ->  response.model 'gpt-5-nano-2025-08-07'
```

So `_end_span_inputs_preprocessor` re-derives the provider from a string that has already lost the prefix, and for Novita that resolves the **inner vendor**:

```
_extract_provider_from_model('deepseek/deepseek-r1-distill-llama-70b')  ->  'deepseek'
```

Because that is not `None`, `to_kwargs()` keeps it and it overwrites the correct start-span value. With your fix restored locally:

```
start-span provider : 'novita'
end-span provider   : 'deepseek'   <- overwrites the start value
```

**And the mislabel is not cosmetic.** 5 of the 135 Novita rows in the price file collide with real `deepseek`-provider rows, so those spans get priced at DeepSeek's direct rates:

| novita row | novita input | deepseek input | error |
|---|---|---|---|
| `novita/deepseek/deepseek-r1` | 4e-06 | 5.5e-07 | **7.3x under-priced** |
| `novita/deepseek/deepseek-v4-flash` | 1.4e-07 | 4.4e-07 | 3.1x over-priced |
| `novita/deepseek/deepseek-v4-pro` | 1.6e-06 | 1.32e-06 | 17.5% under |
| `novita/deepseek/deepseek-v3.2` | 2.69e-07 | 2.8e-07 | 4% over |

A plausible-but-wrong cost is worse than the visible zero it replaces, so this has to be fixed before the fallback can land.

### The fix

Stop re-deriving the provider at end-span. The start-span hook already derives it from the **requested** model, which keeps the prefix and is authoritative:

```python
        model = response_dict.get("model")
        opik_usage = _extract_usage_from_response(response_dict)
        total_cost = _calculate_completion_cost(output)

        # `provider` is deliberately not set here. LiteLLM strips its own routing prefix from
        # `response.model`, so re-deriving the provider from it resolves the *inner* vendor
        # ("novita/deepseek/deepseek-r1" is reported as "deepseek/deepseek-r1") and would
        # overwrite the correct value the start-span hook took from the requested model.
        return arguments_helpers.EndSpanParameters(
            output=output_data,
            usage=opik_usage,
            metadata=metadata,
            model=model,
            total_cost=total_cost,
        )
```

Keep your `.get(provider_name, provider_name)` fallback — that part is right, and it is what makes Databricks work. The two changes have to land together: the fallback on its own is what introduces the mislabel.

I also considered reading `response._hidden_params["custom_llm_provider"]`, which *is* authoritative, and rejected it: the stream aggregator builds a fresh dict, so it is absent on streaming responses — it would work for non-streaming calls and silently not for streams.

With that in place every case lands on the right provider, and it fixes a pre-existing `azure` mislabel as a bonus:

| requested | span provider before | span provider after |
|---|---|---|
| `novita/deepseek/deepseek-r1-distill-llama-70b` | `None` | `novita` |
| `databricks/databricks-gpt-oss-120b` | `None` | `databricks` |
| `azure/gpt-4o` | `openai` (wrong) | `azure` |
| `openai/gpt-5-nano`, `gpt-4o-mini` | `openai` | `openai` |

## 4. What this PR needs

1. Restore the source change to `litellm_completion_decorator.py`, together with the end-span change above — the fallback on its own is what introduces the mislabel.
2. Add the missing `}` in `CostServiceTest.java`.
3. Fix the two `ruff` errors in the new Python test (unused import, line length).
4. Add a Java case in the shape the SDK actually reports — see the inline comment on `provideNovitaAndDatabricksCases`.
5. Narrow the Novita claims in the PR description: with `parseModelName` stripping one `/` segment, Novita cost tracking does not work through the LiteLLM integration, and the description currently promises that it does. Databricks does work, and Novita works for spans that keep the full prefixed model.

I verified the end-span fix in item 3 locally against your branch before suggesting it — Java 128 tests pass with spotless clean and `BUILD SUCCESS`, and on the Python side `tests/unit/` gives 5279 passed with the litellm directory at 28 passed (the only reds left there are two Anthropic calls failing on my own account's billing), `ruff`, `ruff-format` and `mypy` all clean. So the direction is sound; it is the wiring that needs redoing.

`mvn test -pl apps/opik-backend -Dtest=CostServiceTest` and `pytest tests/library_integration/litellm/` between them catch everything in items 1 through 3 in well under a minute. Worth running before the next push, since CI cannot do it for you on a fork.

PR #7811, opened 2026-08-11
test(frontend): cover useQueryParamAndLocalStorageState

thiagohora:
## Details

`useQueryParamAndLocalStorageState` decides how a URL query param and a `localStorage` value combine, and it backs the date-range state on the Logs and Dashboards pages — but it had no tests. This adds them. No production code is touched.

Two behaviours are easy to break without noticing and hard to infer from the code, so they are the focus:
- the **URL takes precedence** over the stored value;
- the **localStorage → URL seeding runs once on mount** and must not overwrite a value the URL already carries.

Also covers `setValue` writing both sides (so a choice survives navigation) and resolving updater functions, plus the default-value fallback.

Split out of #7801 on review feedback: that PR had briefly added an `initSyncReady` option here, and once the option proved unnecessary it was reverted so the PR stops touching this shared hook. The coverage seemed worth keeping on its own rather than dropping, hence this PR.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves #
-

No Jira ticket — test-only follow-up. Split out of OPIK_7793 / #7801 (related, not resolved here).

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: Wrote the test file. No production code changed.
- Human verification: Author reviewed the diff. The mutation results below were run and are reproducible.

## Testing

```
cd apps/opik-frontend
npx vitest run src/hooks/useQueryParamAndLocalStorageState.test.ts   # 10 passed
npx eslint --max-warnings=0 src/hooks/useQueryParamAndLocalStorageState.test.ts
npx tsc --project tsconfig.json --noEmit
```

Rather than assume the assertions bite, each was mutation-checked against the unmodified hook — every mutation below is caught:

| Mutation | Result |
|---|---|
| Swap precedence to `localStorageValue ?? (queryValue as T)` | **2 failed** |
| Drop the `isUndefined(queryValue)` guard on the init sync | **1 failed** |
| Remove the effect's dependency array (syncs every render) | **1 failed** |
| Unmodified hook | 10 passed |

The hook is mocked at its two dependencies (`use-query-params`, `use-local-storage-state`) so the tests assert this hook's own combining logic rather than those libraries'.

Not run: nothing else — the change is a single new test file.

## Documentation

None needed; no behaviour change.

### DIFF
--- a/apps/opik-frontend/src/hooks/useQueryParamAndLocalStorageState.test.ts
+++ b/apps/opik-frontend/src/hooks/useQueryParamAndLocalStorageState.test.ts
@@ -1,0 +1,128 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { renderHook } from "@testing-library/react";
+import { StringParam } from "use-query-params";
+
+const mockSetQueryValue = vi.fn();
+const mockSetLocalStorageValue = vi.fn();
+
+let queryValue: string | null | undefined;
+let localStorageValue: string | null | undefined;
+
+vi.mock("use-query-params", () => ({
+  StringParam: {},
+  useQueryParam: () => [queryValue, mockSetQueryValue],
+}));
+
+vi.mock("use-local-storage-state", () => ({
+  default: (_key: string, options: { defaultValue: string }) => [
+    localStorageValue ?? options.defaultValue,
+    mockSetLocalStorageValue,
+  ],
+}));
+
+import useQueryParamAndLocalStorageState from "./useQueryParamAndLocalStorageState";
+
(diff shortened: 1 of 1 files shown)


thiagohora:
Closing — folding this into #7801 instead, at the author's request, rather than splitting it out.

The tests now live on that branch as `265cc7d`. Worth noting the distinction that motivated the split in the first place is preserved either way: #7801 only *adds* a test file next to `useQueryParamAndLocalStorageState`, it does not modify the hook, which stays byte-identical to `main`. The earlier concern was about changing shared-hook behaviour, and that revert stands.

PR #7838, opened 2026-08-13
[OPIK-7791] [QA] Proposed e2e specs from the 2.2.27 → 2.2.28 release exploration: optimization run cost

CometActions:
**Generated by the release QA side flow (test-proposal step). Not human-authored — it needs review before merge.** Opened as a draft on purpose; a human promotes it.

## Where these came from

Exploration of the **2.2.27 → 2.2.28** release on `staging.dev.comet.com/opik` (workspace `opik-testing`). A human worked the release's manual-verification list first and confirmed each of these flows by hand before it was proposed as a permanent test; the two specs here are the candidates that came back **verified and strong**.

Both cover PR #7683 (OPIK-7521) — the release's headline change: `Optimization.total_optimization_cost` now folds optimizer-internal (GEPA reflection) spend into a run's cost. Nothing in the estate read cost at all before this; `grep total_optimization_cost` over `e2e/` returned nothing.

**Ref these were written and run against: the `2.2.28` tag** (that is what the exploration verified, and what the backend behaviour below is pinned to). The PR bases on `main` because a PR base must be a branch and there is no `2.2.28` branch — so `main` will have moved on from the ref. Both capabilities they tag are `covered: false` in `coverage/taxonomy.yaml` today; the nightly `reconcile.py` owns flipping those flags, so the taxonomy is deliberately untouched here.

### 1. `optimization-cost-aggregate.spec.ts` — `@t2-cuj` `@cap:optimization-studio.kpi-cards`

**Asserts** that a run's cost is its trials **plus** the optimizer-internal traces attributed to it by tag, counted once. Seeds one run with two priced trials ($0.020 + $0.030) and two `Reflection`-tagged traces ($0.006 + $0.004), all with fixed span costs through the API — so the expectation is an exact `$0.060`, not a range. The shape deliberately traps both ways the aggregate can be silently wrong:

* **too low** — dropping the optimizer-internal spend gives the trial-only `$0.050`, which is what 2.2.27 reported;
* **too high** — one trial's trace also carries the run's *own* id in its tags, so a backend that fails to exclude the `(trace, optimization)` pairs already linked through an experiment item gives `$0.070`.

It then checks the number a user actually sees: the run page's **Optimization cost** card reads `$0.060`, exceeds the trials' own `total_estimated_cost` read back from the API, and its tooltip names optimizer-internal calls as the source (that string is also the tell that the *backend* aggregate was used, not the client-side sum of the loaded trials).

**Verification: passed.**

### 2. `optimization-cost-list-vs-detail.spec.ts` — `@t2-cuj` `@cap:optimization-studio.list-optimization-runs` `@cap:optimization-studio.kpi-cards`

**Asserts** that one run reports the same cost through every read path. The runs list and the run page take different backend queries — a project whose runs have no experiments is answered by `FIND_WITHOUT_EXPERIMENTS`, while `getById` always takes `FIND` — and `total_optimization_cost` is the one aggregate the reduced projection still has to compute for real. The user-visible symptom of a divergence is a run reading `$0.000` in the list while priced correctly on its own page; PR #7683's history shows both divergences were live defects fixed late in review.

Three runs across two projects, with three distinct costs so the spec cannot pass by reading the wrong row:

| Run | Shape | Cost |
|---|---|---|
| A | project A, two trials | `$0.037` |
| B | project B, **no experiments at all**, one tagged trace — alone in its project, so that list takes the reduced projection | `$0.0040` |
| C | project A, no experiments, tagged onto run A's cheaper trial trace | `$0.0070` |

Each is asserted through `getById`, its project-scoped list, and a list with no project filter (scope-independence), then on both surfaces: the **Opt. cost** column and the **Optimization cost** card. The two sub-cent costs and the one above `$0.01` also cover both branches of the currency formatter on both surfaces.

**Verification: passed.**

## How they were run

```
npx playwright test \
  tests/optimization-studio/optimization-cost-aggregate.spec.ts \
  tests/optimization-studio/optimization-cost-list-vs-detail.spec.ts
```

against `OPIK_BASE_URL=https://staging.dev.comet.com/opik` (`OPIK_DEPLOYMENT=cloud`, workspace `opik-testing`), 2 workers, at the `2.2.28` checkout. **2 passed**, twice in a row (1.6m and 40.5s — the first run's seeding was rate-limited by the shared workspace, which the SDK bridge backs off through). `tsc --noEmit` clean; `coverage/tag_lint.py` clean (34 specs, 0 problems).

Everything either spec creates is namespaced `cuj-{runId}-*` and deleted in a `finally` — including the optimizations, which global teardown does not sweep by prefix.

## Supporting changes

Both were needed by the specs, and nothing else uses them yet:

* `core/backend/client.ts` — `createOptimization`, `listOptimizations`, `createExperiment` (trial type + `optimizationId`), `linkExperimentItems`, `getExperiment`, and `totalOptimizationCost` on `OptimizationRef`. The pinned TS SDK already exposes the field, so no raw-fetch escape hatch was needed. `getOptimization` and `listOptimizations` share one mapper on purpose: a spec comparing the two backend query paths must not be comparing two different mappings as well.
* `core/backend/ids.ts` — UUIDv7 generation, because the backend rejects any other id version and derives creation time from the id. Note `tests/datasets/dataset-version-counters.spec.ts` still carries its own inline copy; it was left alone rather than refactored, since touching it would mean re-running a 2×1200-item seeding spec that is unrelated to this change.
* `pom/optimization-studio.page.ts` — `gotoList`, the runs-list `Opt. cost` cell, the `Optimization cost` card figure and its tooltip. Extended the existing POM rather than adding a new one. No product testids were added: `StatCard` carries none, so the card is addressed as the innermost element holding both the card's label and a currency figure.

Three candidates from the exploration were dropped, and one sub-case:

* **In-app GEPA run produces `Reflection`-tagged priced traces.** Verified on staging and it *does not happen* at the 2.2.28 pin — staging runs `opik-optimizer 3.1.1`, whose wheel contains no `gepa_reflection`/`Reflection` literals at all, so a real Studio run's aggregate came back as the trial sum to the last digit (`0.001188`, delta `$0.000000`). A spec written against the intended behaviour would fail on merge. It also depends on the optimizer package version rather than anything the estate controls, and a live GEPA run costs real spend with non-deterministic trial counts. **Worth revisiting once the 3.2.0 pin bump lands** — and worth a decision on whether 2.2.28 intends in-app runs and `pip install opik-optimizer==3.2.0` users to disagree.
* **Run-page latency on a large workspace.** Latency assertions against a shared environment are flaky by construction, and the workspace shape that would make it meaningful (thousands of optimizations) cannot be built from a clean state inside a test. The exploration measured p50 0.29–0.93 s against a 3-optimization workspace, which is not the shape the risk lives in.
* **`npm install opik` plus an integration package.** Real and reproducible (`opik-openai@2.2.28` declares `peer opik@"^1.8.61"`, so `npm i opik@2.2.28 opik-openai@2.2.28` fails `ERESOLVE`) but it is a packaging-manifest defect outside the Playwright estate's surface, and not a 2.2.28 regression — the same ranges ship on 2.0.40 onwards.
* **The mid-run client-side fallback and a run with more trials than one page** (sub-cases of the cost card). The card only sums trials client-side when the backend aggregate is absent or `0`, and the exploration watched a real run report a populated aggregate from its first trial onwards — so that path looks hard to reach at all now. Left unwritten rather than guessed at.


AndreiCautisanu:
Baz review

AndreiCautisanu:
Baz review

AndreiCautisanu:
Baz review

PR #7840, opened 2026-08-13
[OPIK-7791] [QA] ask the QA test radar on every PR

AndreiCautisanu:
## Details

Adds the pre-merge half of the two-layer QA flow: on every PR, ask whether the change is worth a test and whether one already exists. The release half already runs on each deploy; before merge is where a missing test is cheapest to add.

- For a test-worthy PR it applies `test-environment` (deploying `pr-<N>.dev.comet.com` through the existing `trigger_test_env_on_label.yaml`), explores the change there, and opens a **draft** PR with a permanent spec — commenting at each step that has something to say.
- The radar itself lives in [comet-ml/comet-automation-tests](https://github.com/comet-ml/comet-automation-tests/blob/master/.github/workflows/pr_test_radar.yml) beside the rest of the QA tooling; this file only dispatches it.
- Dispatched rather than `uses:` — the radar can run an hour once it deploys an environment, and a called workflow's failure would become this workflow's failure. (A `uses:` job also cannot carry `continue-on-error`; GitHub rejects the file.)

**Advisory by construction:** it never requests changes, never fails a required check, and every spec it writes is a draft. Every failure path in the dispatch step exits 0 with a warning, so a QA problem cannot mark your PR red.

**Not run on:** drafts (`ready_for_review` picks them up later), forks (no test environment can be deployed for one, so a verdict is advice nobody can act on), bot authors, and — inside the radar, with no agent invoked — docs/CI/test-only/lockfile PRs.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-7791

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): claude-opus-5
- Scope: The workflow file in this PR, and the radar it dispatches (comet-automation-tests `pr_test_radar.yml`, `pr_radar_comment.py`) plus the `pr-triage` skill in comet-ml/qa-plugin.
- Human verification: Triage was run against three real PRs and every claim it made was checked by hand against the repo — the spec it cited, the capability keys against `taxonomy.yaml`, and the skip guard it relied on. Workflow YAML validated, and all `needs.*.outputs.*` references checked against their declared sets. The explore and propose steps have not yet run on a live PR environment (see Testing).

## Testing

Triage was exercised by dispatching the radar directly at three merged/open PRs, with commenting and labelling disabled so nothing was written to them:

```
gh workflow run pr_test_radar.yml -f pr=<N> -f comment=false -f apply_label=false
```

| PR | Verdict | What it reasoned, and what I verified |
| --- | --- | --- |
| #7764 span dedup (backend-only, 12 files) | not-test-worthy | Named `trace-spans-depth.spec.ts` as already covering the ordinary path, then found that `SpanService` 409s a changed parent — so the remaining gap is unreachable from the UI and belongs in the backend test the author already wrote. Verified the spec exists and asserts span count, tree and token/cost. |
| #7655 Optimization Studio display cleanups | **test-worthy** | Three capabilities, all confirmed present and `covered: false` in `taxonomy.yaml`. Proposed a repro needing no provider key, plus a `localStorage` migration edge case. Verified its claim that the area's only spec never reads a rendered prompt — a naive grep shows 6 "prompt" hits, but all are comments, a button name or seed input. |
| #7791 Ollie insufficient credits (BE+FE) | not-test-worthy | Unreachable on a fresh OSS env; cited `skipIfOllieDisabled`. Verified that guard exists and that the ollie specs are commented cloud-only. |

Also verified: `pr_surface.py` rejects dependabot (#7823), test-only (#7827) and docs (#7831) PRs with **no agent invoked**, and classifies #7684 correctly (12 product files → `projects`, backend bucketed as untracked surface).

Unit tests: 270 pass in `tools/qa-coverage` (16 on the comment builder, 11 on the classifier).

**Not tested:** the explore and propose steps against a live PR environment. Three candidates were each blocked for a legitimate reason — branch deleted after merge, PR from a fork, and a cloud-only feature unreachable on OSS — so the first qualifying PR will be the first live exercise of that half. Its output is a draft PR, and `cleanup-opik-pr-nightly` in comet-deployment reaps the environment, so the downside is bounded.

No video: this is a CI workflow with no UI surface.

## Documentation

None required. The workflow header documents what the radar does, why it dispatches rather than calls, and what it deliberately skips — aimed at whoever arrives from a PR comment wondering what commented on their PR.

### DIFF
--- a/.github/workflows/pr_test_radar_trigger.yml
+++ b/.github/workflows/pr_test_radar_trigger.yml
@@ -1,0 +1,95 @@
+name: PR Test Radar (trigger)
+
+# Asks the QA test radar, on every PR: is this change worth a test, and does one
+# already exist? The radar itself lives in comet-ml/comet-automation-tests
+# (pr_test_radar.yml) beside the rest of the QA tooling; this file only dispatches
+# it.
+#
+# What the radar does, for anyone arriving here from a PR comment:
+#
+#   1. triage  — reads the diff against the capability-coverage map and decides
+#   2. explore — for a test-worthy PR, applies `test-environment` (which deploys
+#      pr-<N>.dev.comet.com via trigger_test_env_on_label.yaml) and drives the
+#      change there
+#   3. propose — opens a DRAFT PR with a permanent spec
+#
+# Only PRs from branches on THIS repo: a fork cannot get a test environment, so
+# the radar skips them entirely.
+#
+# ADVISORY. It comments; it never requests changes, never fails a required check,
+# and everything it writes is a draft. A bot that can block a merge is a bot that
+# gets removed.
+#
+# WHY DISPATCH RATHER THAN `uses:`
+#
+# The radar can take an hour once it deploys an environment and writes specs. A
+# `uses:` job would keep this workflow alive for all of it and make its failure
+# this workflow's failure. Dispatch is fire-and-forget: this run finishes in
+# seconds and a radar problem cannot mark anything on the PR red. (A `uses:` job
+# also cannot carry `continue-on-error` — GitHub rejects the file outright.)
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, ready_for_review]
+
+# Read-only here: everything that writes (the PR comment, the label) is done by
+# the radar with its own token, in the other repo.
+permissions:
+  contents: read
+
+concurrency:
+  # A new push supersedes the previous radar dispatch for this PR, never another
+  # PR's. The radar has its own matching per-PR group.
+  group: pr-test-radar-trigger-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  dispatch:
+    name: Ask the QA test radar
+    # Two exclusions, both deliberate:
+    #   * drafts — the point is to catch a missing test before review, and a draft
+    #     is still being written. `ready_for_review` above picks it up later.
+    #   * forks — no test environment can be deployed for one
+    #     (trigger_test_env_on_label.yaml resolves head.ref, which does not exist
+    #     for a fork's branch), so there is nothing actionable the radar could do.
+    #     Scope is our own repo and our own team for now. 23 of 60 open PRs were
+    #     forks when measured, so this is also most of the saved runner time.
+    if: >-
+      ${{ !github.event.pull_request.draft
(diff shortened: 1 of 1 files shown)


CometActions:
**No test needed here.**

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/31695063378)

AndreiCautisanu:
Thanks — all three were valid, and the security one was the most serious thing in this PR. Fixed in 2e18790.

### 1. PAT reachable from PR-controlled code (high)

Correct, and this was the worst instance of it in the repo: a `run:` block a PR author can freely edit, with a PAT in scope.

Dispatching to another repository does need an elevated token — `GITHUB_TOKEN` is scoped to this repo, so there is no unprivileged path. The fix is therefore to stop running PR-controlled code with it: the workflow is now **`pull_request_target`**, which executes the definition from the **base branch**, so a PR cannot alter what runs here.

The usual objection to `pull_request_target` — checking out and running PR code with secrets in scope — does not apply: this job checks out nothing and runs nothing from the PR. It reads event metadata and makes one API call.

### 2. Draft PRs could still receive a deployment (medium)

Right, and the underlying reason is that the event payload is a snapshot. A PR converted back to draft, or closed, between the event and the dispatch would still have been labelled.

The dispatcher now re-reads live `state` and `isDraft` immediately before asking the radar, and the radar re-reads them again before labelling — so both sides of the hop check.

I did not add a draft guard to `trigger_test_env_on_label.yaml`. That workflow is a shared entry point people also use by hand, and "labelled but draft" is a legitimate thing to want when you are iterating; the right place to stop an *automatic* label is where the automation decides, which is here.

### 3. Bot allowlist missed `github-actions[bot]` (medium)

Correct — and an exact-match list will always miss the next one. Now checks the author **type** GitHub itself reports (`Bot`) plus the `[bot]` suffix, so an unlisted App account is caught without anyone maintaining a list. `CometActions` stays named explicitly, since it is a normal user account that happens to open only generated PRs.

Verified across 8 cases: `github-actions[bot]`, `dependabot[bot]`, `renovate[bot]`, `CometActions` and `dependabot` all skip; `miguelgrc`, `thiagohora` and `AndreiCautisanu` all run.

natagh23 (approved):

PR #7841, opened 2026-08-13
[OPIK-6739] [FE] refactor: share trace filter-chip definitions

awkoy:
## Details

**Stack layer 1 of 3 — no behaviour change.** Moves the trace filter-chip definitions, their display order and their default pinned set out of the Logs page so any surface showing traces can consume them; the entity-scoped trace logs view adopts them in layer 2.

- They land under `pages-shared/traces` rather than `shared/filter-chips` because the option sources (tags, metadata paths, error types) are pages-shared hooks, and the `no-shared-importing-pages` dependency rule forbids `shared/` from reaching them.
- The dynamic-chip builder gains a `logsSource` parameter — it hard-coded the SDK source, which is wrong for evaluator traces — and the Logs page imports it back for its span chips.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- OPIK-6739

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: whole change — extraction and call-site rewiring
- Human verification: reviewed by the author; Logs page filtering exercised in a browser against a local stack

## Testing

- `npm run typecheck`, `npx eslint src --max-warnings=0`, `npx depcruise src --config --ignore-known` — all clean
- `npx vitest run` — 2326 pass; the 6 failing localStorage test files fail identically on `main` (pre-existing, unrelated)
- Filtering on the Logs page exercised manually: same chips, same order, same default pinned set

## Documentation

None — internal refactor with no user-visible surface.

### DIFF
--- a/apps/opik-frontend/src/v2/pages-shared/traces/traceChipDefinitions.ts
+++ b/apps/opik-frontend/src/v2/pages-shared/traces/traceChipDefinitions.ts
@@ -1,0 +1,378 @@
+import compact from "lodash/compact";
+import keyBy from "lodash/keyBy";
+
+import {
+  COLUMN_CUSTOM_ID,
+  COLUMN_FEEDBACK_SCORES_ID,
+  COLUMN_METADATA_ID,
+  COLUMN_SPAN_FEEDBACK_SCORES_ID,
+  COLUMN_TYPE,
+} from "@/types/shared";
+import { LOGS_SOURCE } from "@/types/traces";
+import { GuardrailResult } from "@/types/guardrails";
+import { TRACE_DATA_TYPE } from "@/constants/traces";
+import { CUSTOM_FILTER_VALIDATION_REGEXP } from "@/constants/filters";
+import {
+  ChipDefinition,
+  ChipOptionsResult,
+  chipOptions,
+  chipOptionsValue,
+} from "@/shared/filter-chips/types";
+import {
+  TAGS_OPERATORS,
+  FEEDBACK_SCORE_OPERATORS,
+  DICTIONARY_OPERATORS,
+  STRING_OPERATORS,
+  LIST_OPERATORS,
+} from "@/shared/filter-chips/chips/QueryBuilderChip/operators";
+import { useTagsOptions } from "@/v2/pages-shared/TagsAutocomplete/useTagsOptions";
+import { usePathsOptions } from "@/v2/pages-shared/traces/TracesOrSpansPathsAutocomplete/usePathsOptions";
+import { useErrorTypeOptions } from "@/v2/pages-shared/traces/ErrorTypeAutocomplete/useErrorTypeOptions";
+
+/**
+ * Filter-chip definitions for traces, shared by the Logs page and by the entity-scoped trace logs
+ * view (experiment Logs tab, playground, trials, annotation queues, evaluator traces) so both
+ * surfaces filter through one definition.
+ *
+ * This lives under pages-shared rather than shared/filter-chips because the option sources
+ * (tags, metadata paths, error types) are pages-shared hooks, and shared/ may not import them.
+ */
+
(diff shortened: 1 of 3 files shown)


miguelgrc (dismissed):

apps/opik-frontend/src/v2/pages-shared/traces/traceChipDefinitions.ts
```diff
@@ -0,0 +1,378 @@
...
+  feedbackScoresLabel,
+  isGuardrailsEnabled,
+  logsSource = LOGS_SOURCE.sdk,
+}: {
+  projectId: string;
+  type: TRACE_DATA_TYPE;
+  scoreOptions: ChipOptionsResult;
+  feedbackScoresLabel: string;
+  isGuardrailsEnabled: boolean;
+  logsSource?: LOGS_SOURCE;
```
miguelgrc: 💡 **suggestion** | API design

The PR body says the old code "hard-coded the SDK source, which is wrong for evaluator traces" — but defaulting the new parameter to `LOGS_SOURCE.sdk` rebuilds that same footgun one level up: a layer-2 consumer that forgets `logsSource` silently gets SDK-scoped tag / error-type / metadata-path options, with no type error to catch it.

Making it required here (and on `buildTraceChipDefinitions` below) and passing `LOGS_SOURCE.sdk` explicitly at the two Logs-page call sites keeps behaviour identical while forcing every new surface to state its source. `TracesSpansTab.tsx` already imports `LOGS_SOURCE` and uses it five other times, so the call sites cost one line each.

```suggestion
  logsSource,
}: {
  projectId: string;
  type: TRACE_DATA_TYPE;
  scoreOptions: ChipOptionsResult;
  feedbackScoresLabel: string;
  isGuardrailsEnabled: boolean;
  logsSource: LOGS_SOURCE;
```

awkoy:
Good catch — that default did rebuild the exact footgun the parameter exists to remove. Made it required on both builders and passed `LOGS_SOURCE.sdk` explicitly at the two Logs-page call sites.

One tweak on the type: `LOGS_SOURCE | undefined` rather than `LOGS_SOURCE`. Entity-scoped hosts legitimately have no source filter (the experiment Logs tab is one), so undefined has to stay expressible — but it now has to be *stated*, which is the part that matters. Thanks!

CometActions:
**No test needed here.**

Pure code motion — I diffed the moved blocks against their originals and TRACE_CHIP_DEFINITIONS_STATIC, TRACE_CHIP_ORDER and TRACE_DEFAULT_PINNED_CHIPS are byte-identical; the only semantic edit is hoisting the hardcoded `logsSource: LOGS_SOURCE.sdk` into a parameter, and both call sites in TracesSpansTab still pass LOGS_SOURCE.sdk, so the Logs page renders the same chips in the same order. No other file imports traceChipDefinitions yet, so no new surface is reachable from this PR. If the extraction had dropped or reordered a chip, e2e/tests/trace-explore/trace-filters.spec.ts would already fail — it drives the tags, with_errors and metadata chips and pins the "Trace feedback scores" label by hand. Worth re-triaging the follow-up PR that actually wires a second surface to these definitions with a non-sdk logsSource, since that one does change what a user sees.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/31805327291)

_Re-checked after a push on 14 Aug 13:37 UTC._

awkoy:
Added `traceChipDefinitions.test.ts` — 7 cases covering the declared ordering, the guardrails branch appearing only when the flag is on and in its declared slot, trace vs span feedback scores staying distinct, and the one that actually matters: that the logs source reaches each remote-option hook (tags, error type, metadata, custom) for SDK, non-SDK and undefined. A wrong source there hands a surface autocomplete values its own table would never match, and nothing else catches it.

miguelgrc (approved):

PR #7842, opened 2026-08-13
[OPIK-6739] [FE] refactor: extract TraceLogsView and modernise its toolbar

awkoy:
## Details

**Stack layer 2 of 3.** The traces table lived inside the sliding logs overlay, so it could not be hosted anywhere else. This extracts it into a standalone view; the overlay becomes a thin sheet wrapper whose public props are unchanged. While it moves, it comes in line with the Logs page — chip-based filter bar with a compact search prefix instead of the funnel button, bulk actions in the selection action bar, compact control sizing.

Behaviour that changes on purpose:
- **Date range defaults to all time** for entity-scoped logs. These views are already narrowed to one entity, so the trailing 30-day window only hid traces of anything older — an experiment run two months ago opened on an empty table. Needed a caller-supplied default on the date-range hook plus a namespaced storage key, since the shared one carries the Logs page's window.
- **Optimization trial logs switch to a locked scope.** The chip bar owns `tls_filters` and drops any filter without a matching chip; there is no experiment chip, so the trial overlay would otherwise have listed every trace in the project.
- **Locked-scope tooltip** is supplied by the caller so each host describes its own scope; the per-evaluator view passes its existing wording unchanged.
- **Controls are 24px**, matching the optimization runs page and trial sidebar rather than the Logs page's 28px.

Known consequence: filtering these views by experiment is no longer possible — the Logs page has no experiment chip either, so this follows from adopting its chip set.

> Blast radius is wider than OPIK-6739 asks for: this changes the filtering experience on five surfaces the ticket never mentions. It is a separate layer precisely so it can be accepted or rejected on its own terms.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- OPIK-6739

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: whole change — extraction, toolbar migration, date default, scope fix
- Human verification: every host driven in a browser against a local stack; control heights measured in the DOM

## Testing

- `npm run typecheck`, `npx eslint src --max-warnings=0`, `npx depcruise src --config --ignore-known` — clean on this layer in isolation
- `npx vitest run` — 2326 pass, including 5 new cases on the date-range hook's caller-supplied default and precedence rule. The 6 failing localStorage test files fail identically on `main`.
- Manual, local stack (vite dev + local backend), all five overlay hosts:
  - **Playground** — new toolbar renders, controls measured at 24px
  - **Optimization trial** — locked scope lands in `tls_scope`; confirmed against the API that the remaining emptiness is a pre-existing `source` mismatch (traces carry `sdk`, the button filters `optimization`), not this change
  - **Annotation queue** — seeded queue filter survives the chip bar as a removable chip; 3 of 3 queued traces, not the project's 19
  - **Online-eval rules / per-evaluator** — lock chip renders with the evaluator wording verbatim, KPI summary and trimmed columns intact
  - **Date default** — a fixture dated three months back is visible on open; switching to "Past 30 days" empties the table, which is the old behaviour
- Not run: Playwright e2e (needs a live stack); covered in CI.

## Documentation

None — no documented behaviour changes.


apps/opik-frontend/src/v2/pages/OptimizationPage/TrialSidebar/TrialSidebar.tsx
```diff
@@ -80,6 +80,11 @@ const TrialSidebar: React.FC<TrialSidebarProps> = ({
               sourceFilters={generateExperimentIdsFilter(
                 trialExperiments.map((e) => e.id),
               )}
+              // Locked, not seeded: these are this optimization's trial logs, so the scope has to
+              // hold. Seeding it into the editable filter bar no longer works — the chip bar owns
+              // that key and drops any filter without a matching chip, which there is none for
+              // experiment ids, so the overlay would fall back to the whole project.
+              lockScope
```
miguelgrc, line 88: 🚫 **blocker** | UX regression

`lockScope` arrives without `scopeLabel`, and `TraceLogsView` renders the lock indicator behind `scopeLabel && (...)` (`TraceLogsView.tsx:1020`). So the trial overlay applies an invisible constraint: results are silently narrowed to this trial's experiments with no chip, no lock tag, and no tooltip explaining why — and no way for the user to discover it. The `scopeTooltip` plumbing this PR adds never fires here either.

It also flips the empty state. On `main` the trial seeded `tls_filters`, so `filters.length > 0` and an empty table rendered `DataTableNoMatchingData` with a **Clear filters** affordance. Now the scope lives in `scopeFilters`, which `noData = !search && chipFilters.length === 0` (`TraceLogsView.tsx:662`) doesn't count — so the same empty table renders *"There are no traces yet — traces will appear here once your agent starts receiving requests."* That's the wrong story for a host the PR description says renders empty today because of the `source` mismatch.

The only other `lockScope` caller, `RuleTracesCell.tsx:47-48`, pairs it with `scopeLabel={rule.name}`. There's already a `title` in scope here (`"Baseline"` / `"Trial #N"` / `"Trial"`, line 45):

```suggestion
              lockScope
              scopeLabel={title}
```

apps/opik-frontend/src/v2/pages-shared/traces/TraceLogsView/TraceLogsView.tsx
```diff
@@ -0,0 +1,1174 @@
...
+    intervalEnd,
+    minDate,
+    maxDate,
+  } = useMetricDateRangeWithQueryAndStorage({
+    defaultValue: viewConfig.defaultDateRangePreset,
+    // Namespaced storage so the all-time default actually applies: the key is otherwise shared with
+    // the main Logs page, whose stored trailing window would override it for anyone who has ever
+    // changed the range there. The URL key stays shared, so existing deep links keep working.
+    localStorageKey: `${storagePrefix}date-range`,
+  });
```
miguelgrc, line 454: 💡 **suggestion** | State isolation

`localStorageKey` is namespaced but `key` isn't, so the URL param stays the shared `time_range`. Two consequences:

1. `useQueryParamAndLocalStorageState` computes `combinedValue = queryValue ?? localStorageValue` — **the URL wins over localStorage**. Any `time_range` already present in the URL silently defeats the new all-time default, which is the exact failure this layer sets out to fix.
2. `syncQueryWithLocalStorageOnInit: true` writes the namespaced value *out* to the shared key on mount (the effect fires because `localStorageValue` defaults to `alltime` and `queryValue` is undefined). Opening any of these overlays stamps `time_range=alltime` into the host page's URL and leaves it there after close — so it rides along in copied and shared links.

Every other param this view owns is `tls_`-prefixed; `key: \`${TLS_QUERY_PREFIX}time_range\`` would finish the namespacing. The stated reason to keep it shared is deep-link compatibility — worth weighing against the default not reliably applying, since old links carrying `time_range` are precisely the ones that break the fix.

CometActions:

Not a pure refactor, and one thing looks wrong: now that the overlay's filter bar is the chip bar, `tls_filters` is sanitized against the chip definitions, and `buildTraceChipDefinitions` has no chip for `experiment_ids`. That is the exact fallback you fixed on TrialSidebar with `lockScope` — but CompareExperimentsDetails' "Experiment logs" button (`sourceFilters={experimentSourceFilters}`, no `lockScope`) and PlaygroundOutputCell's dataset-run deep link (`tls_filters=[experiment_ids]`) still seed it the old way, so both should drop the scope and list the whole project. Worth a test, except our taxonomy has no capability for the entity-scoped logs overlay on experiments or annotation queues; the only key that names this component is `playground.playground-logs-sidebar`, and its spec (prompts/prompt-playground-traces.spec.ts) only asserts the sheet opens and one row renders, so it would not catch scope widening. Sending this to taxonomy-discovery rather than spending an environment on a target I can't name.

_areas: `optimization-studio`, `threads`, `traces`_
[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/31787953722)

CometActions:

Not only a move: the logs sidebar's `tls_filters` now goes through the chip bar's `sanitizeFilters`, which drops any filter with no matching chip. There is no `experiment_ids` chip, and CompareExperimentsDetails still passes `generateExperimentIdsFilter(...)` as an unlocked `sourceFilters`, so "Experiment logs" looks like it would open on the whole project — the same failure `lockScope` was added to prevent on trials. The default range also flips to all-time on a new namespaced storage key. The only spec on this component, e2e/tests/prompts/prompt-playground-traces.spec.ts (@cap:playground.playground-logs-sidebar), just opens the sheet, clicks the first row and checks the Prompts tab; it never touches the toolbar, filters, date range or scope, and trace-filters.spec.ts drives chips on the main Logs page only.

Would target `traces.filter-traces`, `playground.playground-logs-sidebar`, `optimization-studio.trial-detail`.

1. Seed one project with an SDK experiment plus unrelated traces, open the experiment page, click "Experiment logs" and check the table is limited to that experiment's traces rather than the whole project (the console logs a `[FilterChips] ... no_matching_chip` drop when the scope is lost)
2. Repeat from the annotation queue "Logs" pill, where `annotation_queue_ids` does have a chip, and confirm the queue scope still narrows the table
3. Seed a trace with a start_time older than 30 days and confirm the sidebar's all-time default lists it, that an explicitly picked range still wins, and that it survives closing and reopening the sheet
4. Apply then clear a chip filter (tags, With errors) inside the sidebar and confirm rows narrow, the count matches, and the tls_filters deep link round-trips
5. Check the main Logs page trace chips are unchanged after the extraction: chip order, the with_errors/tags/metadata pinned defaults, and the guardrails chip when the toggle is on

</details>

_areas: `optimization-studio`, `threads`, `traces`_
[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/31789190301)

apps/opik-frontend/src/v2/pages/OptimizationPage/TrialSidebar/TrialSidebar.tsx
miguelgrc: Verified at `fefc5a3` — the visibility half is fully fixed, and along the whole chain, not just here: `scopeLabel={title}` → `TraceLogsSidebarButton:57` builds `{ locked: true, label: scopeLabel }` → `openSidebar` writes `tls_scopeLabel` → the sidebar reads it back → `TraceLogsView:1039` renders the lock tag. `renderSidebar` defaults to `true` so this host does mount the sidebar, and `title` is never empty (`"Baseline"` / `"Trial #N"` / `"Trial"`). Thanks — and the added comment captures the real invariant.

**The empty-state half is still open**, downgraded to a 💡 suggestion now that the lock is visible.

`noData = !search && chipFilters.length === 0` (`TraceLogsView.tsx:675`) still doesn't count `scopeFilters`, so a scoped-but-empty table renders *"There are no traces yet — traces will appear here once your agent starts receiving requests."* That copy now sits directly under a visible "Trial #N" lock tag, which reads oddly: the tag says the view is narrowed, the empty state says the agent has never recorded anything. Per the PR description this is the state the trial overlay actually lands in today, because of the pre-existing `source` mismatch. `onClearFilters` (line 1113) is gated the same way, so there's no affordance either.

Counting the locked scope would fix both messages at once — something like `!search && chipFilters.length === 0 && scopeFilters.length === 0`, so a scoped view falls through to `DataTableNoMatchingData` instead of claiming no traces exist. Fine as a follow-up if you'd rather keep this PR's diff where it is.

CometActions:

Read as behaviour, not refactor: the entity-scoped logs overlay now defaults to all time instead of the trailing 30 days, replaces the filters popup with the chip bar, and moves its range onto tls_time_range. TrialSidebar needed lockScope/scopeLabel because the chip bar drops filters with no matching chip — worth a look at CompareExperimentsDetails, which still seeds generateExperimentIdsFilter into tls_filters and leans on the legacy lift in TraceLogsSidebar instead. The only spec touching this component, prompts/prompt-playground-traces.spec.ts (@cap:playground.playground-logs-sidebar), just opens the sheet and a trace row and asserts nothing about toolbar, scope or date range, so none of the above is covered; the main Logs page extraction is fine — trace-explore/trace-filters.spec.ts pins exact row ids for the tags, with-errors and feedback-score chips. Holding at unclear rather than yes because taxonomy.yaml has no capability for this shared overlay outside the playground, and the routes reachable without a provider key (experiment compare, annotation queues, evaluator traces) have no key to file a test under.

_areas: `optimization-studio`, `threads`, `traces`_
[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/31829232618)

_Re-checked after a push on 14 Aug 18:34 UTC — nothing the verdict depends on changed._

awkoy:
All nine threads addressed and resolved. Summary of what changed:

**@miguelgrc's blocker — trial scope was invisible.** Both halves were real and both are fixed:
- `TrialSidebar` now passes `scopeLabel={title}`, so the lock indicator renders (`🔒 Trial #7`) with its tooltip.
- The empty state flip you spotted was the sharper half. `noData` counted only search and chip filters, so a scoped-but-empty table claimed *"There are no traces yet — traces will appear once your agent starts receiving requests"*, which is a lie when a scope is doing the narrowing. It now counts `scopeFilters` too, so that case reads "No matching results". Verified in a browser against a local stack.

**@miguelgrc's `time_range` suggestion — taken as written.** Both consequences you listed were live: the URL value beat storage and defeated the all-time default, and `syncQueryWithLocalStorageOnInit` stamped `time_range=alltime` into the host page's URL on mount and left it there after close. The key is now `tls_time_range`, matching every other param this view owns. Deep links carrying the old shared key stop applying to these views, which I think is the right trade.

**Bot findings** — four were real and are fixed (null search normalisation, `Array.isArray` guard on the scope URL, `visibilityMode` now passed to `useTracesStatistic` so the KPI cards and rows agree, and the `experiment_ids` drop, which was the trial bug above). Two were already handled (toolbar row is hoisted into one const; `CompareExperimentsDetails` no longer renders the sidebar button at all).

Worth flagging: the statistics/visibility mismatch predates this PR — the overlay called the hook the same way on `main`. Fixed here since the extraction put both calls side by side.

apps/opik-frontend/src/api/traces/useTracesStatistic.ts
```diff
@@ -35,6 +41,7 @@ const getTracesStatistic = async (
       signal,
       params: {
         project_id: projectId,
+        ...generateVisibilityFilters(visibilityMode),
```
miguelgrc: 🚫 **blocker** | Cross-surface behavior change

The KPI-cards fix is right, but applying it to the **shared** hook with a default changes every existing caller. Before this commit `useTracesStatistic` sent **no** visibility filter at all; now any caller that doesn't pass one gets `visibility_mode = "default"` appended — and none of the five existing callers pass it:

- `ProjectStatsCardWidget` (v1 + v2) — dashboard KPI cards
- `TraceQueueItemsTab` (v1 + v2) — annotation-queue column statistics
- `useTracesOrSpansStatistic` — forwards `params` straight through, so the Logs page's column statistics too

Experiment runs auto-create hidden traces, so **any project that has run an experiment will see those numbers change on merge** — three surfaces this PR doesn't touch and the testing table doesn't cover.

Directionally it's probably a fix: the matching lists go through `useTracesList`, which already defaults to `TRACE_VISIBILITY_MODE.default`, so stats and rows previously disagreed. But that's a product call on unrelated surfaces riding inside a refactor PR — worth landing on its own, or at least calling out.

Keeping the param genuinely optional leaves every existing caller byte-identical and still gives `TraceLogsView` what it needs. Both lines have to change together — `generateVisibilityFilters` defaults its *own* parameter to `default`, so dropping the destructure default alone wouldn't be enough:

```suggestion
    visibilityMode,
  }: UseTracesStatisticParams,
) => {
  const { data } = await api.get<UseTracesStatisticResponse>(
    `${TRACES_REST_ENDPOINT}stats`,
    {
      signal,
      params: {
        project_id: projectId,
        ...(visibilityMode ? generateVisibilityFilters(visibilityMode) : []),
```

awkoy:
Good catch, and it's the same silent-broadening failure mode arriving by a different route — an old bookmarked trial-log link carries the experiment constraint in `tls_filters`, no chip can express experiment ids, so the chip bar drops it and the overlay widens to the whole project.

Fixed by migrating on read: the sidebar lifts any `experiment_ids` filters out of `tls_filters` into the scope before querying. Old links keep working and stay constrained; new ones use `tls_scope` directly.

awkoy:
@miguelgrc right again, and this one was mine — thanks for catching it before it shipped.

`useTracesStatistic` had never sent a visibility filter, so adding `visibilityMode = TRACE_VISIBILITY_MODE.default` as a *default* silently rewrote the request for all five other callers (both `ProjectStatsCardWidget`s, both `TraceQueueItemsTab`s, `useTracesOrSpansStatistic`) — hidden traces would have dropped out of their aggregates. It's opt-in now: the filter is only appended when a caller passes one, so every existing call site is byte-identical to before and `TraceLogsView` still gets the alignment the KPI cards needed.

Bot findings in the same batch: fixed the constants split (the sidebar trigger and its controls hook were pulling the table's dependency graph in for two string constants) and the O(n×m) export selection lookup.

Three I'm not acting on here, all pre-existing and already named as follow-ups: the column definitions duplicated with `TracesSpansTab` and the view's size are the tech-debt ticket this stack's description points at — unifying them is a ~3,000-line refactor of the main Logs page, which is exactly what layer 1 deliberately scoped out. The overlay date-isolation test would assert on hook wiring rather than behaviour; the behaviour it protects is covered by the namespaced key itself.

awkoy:
Backed this out entirely — `useTracesStatistic` is now byte-identical to `main` and `TraceLogsView` no longer passes a visibility mode to it.

The mismatch is real, but it predates this work, it lives in the evaluator traces view rather than the experiment table this stack exists for, and it only surfaced because extracting the view put both calls in one file. Chasing it here meant editing a hook five unrelated callers share — which is exactly how the default that would have dropped hidden traces from dashboard and queue aggregates got in. Not a trade worth making for a ticket about adding a tab.

Filing it separately. The stack now touches no files under `src/api` at all.

apps/opik-frontend/src/api/traces/useTracesStatistic.ts
```diff
@@ -35,6 +41,9 @@ const getTracesStatistic = async (
       signal,
       params: {
         project_id: projectId,
+        // Opt-in, unlike useTracesList: this hook has never sent a visibility filter, and defaulting
+        // one would silently drop hidden traces from every existing caller's aggregates.
+        ...(visibilityMode ? generateVisibilityFilters(visibilityMode) : {}),
```
miguelgrc: 💡 **suggestion (non-blocking)** | Correctness

The opt-in shape is right, and the reasoning in `9cf0de2a` about blast radius is right too. But this restored version doesn't actually apply the filter, so `8c50728a`'s claim that it *"stops the evaluator view aggregating a different set of traces than the rows beneath"* isn't true yet.

`generateVisibilityFilters` returns a `Filter[]`, and it's spread into the **params object**, so it lands under a numeric key instead of inside `filters`:

```js
> const params = { project_id: "p1", ...generateVisibilityFilters("hidden"), filters: "…" }
> Object.keys(params)
[ '0', 'project_id', 'filters' ]
```

That goes out as a stray `0=…` query param. `TracesResource.getStats` declares only named `@QueryParam`s, so Jersey ignores it — no error, but no filtering either. `useTracesList:56-58` gets this right by collecting into an array and passing it *into* `processFilters`, which serializes everything into the single `filters` param:

```ts
const additionalFilters = [
  ...(visibilityMode ? generateVisibilityFilters(visibilityMode) : []),
  ...(logsSource ? generateLogsSourceFilter(logsSource) : []),
];
// …
...processFilters(filters, additionalFilters),
```

**Why non-blocking:** the only caller passing `visibilityMode` is `TraceLogsView`, and the default config passes `TRACE_VISIBILITY_MODE.all`, which yields `[]` — a real no-op. So the playground, trial, annotation-queue overlays and the experiment Logs tab are unaffected; only `EVALUATION_TRACES_VIEW_CONFIG` (`hidden`) hits it, and only its column-header aggregates. Against `main` the behaviour is **identical**, since main never sent a visibility filter from this hook either. Not a regression.

Given that, my suggestion is to **finish the revert rather than fix the spread**. `9cf0de2a` argues the mismatch predates this stack, lives in the evaluator view, and deserves its own ticket — that reasoning holds, and it argues for dropping the plumbing entirely instead of carrying a version that doesn't fire. Either way worth amending the commit message, since a future reader will otherwise trust that sentence.

miguelgrc (dismissed):
Left a non-blocking comment, but looks good otherwise :)

awkoy:
You're right, and thanks for actually running it — I'd have kept carrying a no-op.

`generateVisibilityFilters` returns an array, I spread it into the params object, so it went out as a stray `0=` param that Jersey drops. The filter never fired, which makes `8c50728a`'s commit message plainly wrong. Worse, I couldn't catch it in the browser because I never managed to populate the evaluator view with queryable traces — so the one change I couldn't verify visually is exactly the one that was broken.

Taken your recommendation: finished the revert rather than fixing the spread. `useTracesStatistic` is byte-identical to `main`, `TraceLogsView` no longer passes a visibility mode to it, and the stack touches nothing under `src/api`. The commit message says what actually happened, so the next reader isn't misled.

The evaluator KPI/rows mismatch gets its own ticket — it predates this stack and is untouched by it either way.

miguelgrc (dismissed):
Thanks for addressing the comments! 🚀

awkoy:
Resolved the latest seven without code changes — every one of them predates this stack. Checked each against `10c813547a` (main before the stack):

| Finding | Where it already lived |
|---|---|
| Null pagination (`size as number`) | `TraceLogsSidebar.tsx:611-612`, identical |
| Primary/export request params duplicated | `:611` and `:634`, identical |
| Export fetches only the current page | same code |
| Empty state ignores source and date constraints | same logic — this stack *narrowed* it, by making a scope count as a constraint |
| Selection count doesn't survive pagination | exactly what the canonical Logs table does (`TracesSpansTab.tsx:1500`) |
| Duplication with `TracesSpansTab` | the tech-debt follow-up this stack's layer 1 deliberately scoped out |
| `hideEvaluate` always passed | `TraceLogsSidebar.tsx:1024`, present before this PR |

They surface now because the extraction puts ~1,200 lines into a new file, so the whole thing reads as new code in the diff. None is a regression, and fixing them here would mean editing a shared component on a PR that's waiting on approval — which is how the visibility-filter detour earlier in this PR broke five unrelated callers and then turned out to be a no-op anyway.

Filing them as one follow-up against the trace-logs view, together with the evaluator KPI/rows mismatch from that thread. Happy to pull any single one in if you'd rather it land here.

JetoPistola (approved):
Re-reviewed at 37fe39e7. Scope handling, empty-state and the date-range default all check out; the statistics-hook change was correctly reverted. Restoring approval after the restack.

PR #7843, opened 2026-08-13
[OPIK-6739] [FE] feat: experiment logs as a tab, plus Copy ID

awkoy:
## Details

**Stack layer 3 of 3 — the ticket itself.** Experiment traces were reachable only through a small "Go to logs" tag in the metadata row, which users kept missing: they look for their experiment traces under Logs, don't find them, and conclude nothing was recorded. This shows them as a **Logs** tab on the experiment page and drops the tag, so there is one way in.

- Traces are always scoped to the open experiment — or every experiment in a comparison — through the locked-scope mechanism. No scope indicator is drawn; the page already says which experiment this is.
- Tab availability becomes a pure function, so the test-suite and comparison rules are testable directly.
- Adds a **Copy ID** action to the header, following the pattern already used by the experiment comparison and test-suite panels. SDK features acting on an existing run needed the id, previously recoverable only from the page URL or an opt-in column. Hidden while comparing, where the id would be ambiguous. **Note: this item was raised in the ticket thread but never signed off by design** — it is isolated so it can be dropped without disturbing the tab.
- The sibling tabs (Results, Configuration) need the same toolbar sizing or the table jumps as you switch between them and this one. That lives in **#7845**, stacked directly on top of this PR, and is meant to land with it rather than after.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves OPIK-6739

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: whole change — tab, scoping, Copy ID, sibling-tab alignment, tests
- Human verification: every scenario below driven in a browser against a local stack with purpose-seeded fixtures

## Testing

- `npm run typecheck`, `npx eslint src --max-warnings=0` — clean
- `npx vitest run` — 2326 pass, including 5 new cases on tab availability. The 6 failing localStorage test files fail identically on `main`.
- New e2e scenario in `experiments-smoke.spec.ts`; both new capabilities registered in `tests_end_to_end/coverage/taxonomy.yaml` and tagged on the spec.
- Manual, local stack, with fixtures seeded for the purpose:

  | Scenario | Result |
  |---|---|
  | Regular experiment | 6/6 rows, project holds 90 |
  | Test-suite experiment | tabs `Results · Configuration · Logs`; Insights and Feedback scores stay hidden; 4/4 rows, project holds 19 |
  | Compare (2 experiments) | Logs present, 8/8 rows, Copy ID hidden |
  | Date range | opens on All time; traces from three months back are visible |
  | Row click | trace details panel opens at page level |
  | Row selection | chip bar swaps for the selection action bar |
  | Toolbar alignment | Results / Configuration / Logs all render a 56px row, no gap above or below, all controls 24px (measured in the DOM) |

- Not run: Playwright e2e locally (needs a live stack); covered in CI.

## Documentation

None — the tab replaces an existing entry point; no documented behaviour changes.

### DIFF
--- a/apps/opik-frontend/src/v2/pages/CompareExperimentsPage/ExperimentLogsTab/ExperimentLogsTab.tsx
+++ b/apps/opik-frontend/src/v2/pages/CompareExperimentsPage/ExperimentLogsTab/ExperimentLogsTab.tsx
@@ -2,6 +2,9 @@ import React, { useMemo } from "react";
 
 import { Experiment } from "@/types/datasets";
 import { generateExperimentIdsFilter } from "@/lib/filters";
+import DataTableEmptyContent from "@/shared/DataTableNoData/DataTableEmptyContent";
+import emptyLogsLightUrl from "/images/empty-logs-light.svg";
+import emptyLogsDarkUrl from "/images/empty-logs-dark.svg";
 import TraceLogsView from "@/v2/pages-shared/traces/TraceLogsView/TraceLogsView";
 
 type ExperimentLogsTabProps = {
@@ -30,7 +33,22 @@ const ExperimentLogsTab: React.FunctionComponent<ExperimentLogsTabProps> = ({
   // loaded experiment is what tells us which one.
   const projectId = experiments.find((e) => e.project_id)?.project_id;
 
-  if (!projectId) return null;
+  // Nothing to render before the experiment loads. Once it has, a missing project means the run
+  // never produced traces — say so rather than leaving the tab blank.
+  if (!projectId) {
+    if (experiments.length === 0) return null;
+
+    return (
+      <div className="py-8">
+        <DataTableEmptyContent
+          title="There are no traces yet"
+          description="Traces will appear here once this experiment records them."
+          lightImageUrl={emptyLogsLightUrl}
+          darkImageUrl={emptyLogsDarkUrl}
+        />
+      </div>
+    );
+  }
 
   return (
     <TraceLogsView
--- a/apps/opik-frontend/src/v2/pages/CompareExperimentsPage/ExperimentLogsTab/ExperimentLogsTab.tsx
+++ b/apps/opik-frontend/src/v2/pages/CompareExperimentsPage/ExperimentLogsTab/ExperimentLogsTab.tsx
@@ -1,4 +1,6 @@
 import React, { useMemo } from "react";
+import isString from "lodash/isString";
+import uniq from "lodash/uniq";
 
 import { Experiment } from "@/types/datasets";
 import { generateExperimentIdsFilter } from "@/lib/filters";
@@ -13,8 +15,13 @@ import TraceLogsView, {
 // Per design, the tab's toolbar carries the columns selector and nothing else — no row-height, date
 // range or refresh. Dropping the date control also drops date filtering entirely, so the tab always
 // spans the experiment's whole life.
+//
+// Its own storage namespace, so the columns, sort, page size and pinned chips a user sets here
+// don't overwrite the ones they set in the playground or trial overlays, which take the default
+// config. Without this every host shares one arrangement.
 const EXPERIMENT_LOGS_VIEW_CONFIG: TraceLogsViewConfig = {
   ...DEFAULT_TRACE_LOGS_VIEW_CONFIG,
+  storageNamespace: "experiment-",
   showTableControls: false,
 };
 
@@ -36,19 +43,40 @@ const ExperimentLogsTab: React.FunctionComponent<ExperimentLogsTabProps> = ({
   experiments,
 }) => {
   const scopeFilters = useMemo(
-    () => generateExperimentIdsFilter(experimentsIds),
+    // JsonParam yields whatever the URL parsed to, so a malformed link could otherwise put a number
+    // or object through join(",") and emit a filter like "[object Object]".
+    () => generateExperimentIdsFilter(experimentsIds.filter(isString)),
     [experimentsIds],
   );
 
-  // Compared experiments can only be compared within a dataset, so they share a project; the first
-  // loaded experiment is what tells us which one.
-  const projectId = experiments.find((e) => e.project_id)?.project_id;
+  // Traces are queried per project, so comparing experiments that live in different ones cannot be
+  // served by a single request. That happens: an experiment's project resolves from its own
+  // project_name, independently of the dataset, and the dataset lookup itself is project-agnostic.
+  const projectIds = useMemo(
+    () => uniq(experiments.map((e) => e.project_id).filter(isString)),
+    [experiments],
+  );
+  const projectId = projectIds[0];
 
-  // Nothing to render before the experiment loads. Once it has, a missing project means the run
-  // never produced traces — say so rather than leaving the tab blank.
+  // Nothing to render before the experiment loads.
+  if (experiments.length === 0) return null;
+
+  if (projectIds.length > 1) {
+    return (
+      <div className="py-8">
+        <DataTableEmptyContent
+          title="These experiments live in different projects"
+          description="Logs are read per project, so they can't be listed together. Open an experiment on its own to see its traces."
+          lightImageUrl={emptyLogsLightUrl}
+          darkImageUrl={emptyLogsDarkUrl}
+        />
+      </div>
+    );
+  }
+
+  // A loaded experiment with no project never produced traces — say so rather than leaving the tab
+  // blank.
   if (!projectId) {
-    if (experiments.length === 0) return null;
-
     return (
       <div className="py-8">
         <DataTableEmptyContent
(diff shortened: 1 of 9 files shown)


apps/opik-frontend/src/v2/pages/CompareExperimentsPage/ExperimentLogsTab/ExperimentLogsTab.tsx
```diff
@@ -0,0 +1,74 @@
...
+  experiments,
+}) => {
+  const scopeFilters = useMemo(
+    () => generateExperimentIdsFilter(experimentsIds),
+    [experimentsIds],
+  );
+
+  // Compared experiments can only be compared within a dataset, so they share a project; the first
+  // loaded experiment is what tells us which one.
+  const projectId = experiments.find((e) => e.project_id)?.project_id;
```
miguelgrc: 💡 **suggestion** | Data completeness

The invariant in this comment isn't guaranteed. An experiment's `project_id` resolves from its own `project_name` (`ExperimentService:636` → `projectService.resolveProjectIdOrCreate`), independently of the dataset — and `DatasetService.getOrCreate:183-186` falls back to a **project-agnostic** dataset lookup (the `<if(project_id)>` clause drops out when the `@Define` is null) because `datasets_workspace_id_name_uk UNIQUE (workspace_id, name)` was never widened when `project_id` was added in migration `000057`.

So `evaluate(dataset_name="ds", project_name="proj-b")` against an existing `"ds"` in proj-a produces an experiment on proj-a's `dataset_id` with `project_id` = proj-b. Those two experiments are then comparable, with different projects.

Uncommon, but when it happens this queries the first experiment's project with a filter spanning all experiment ids, and the others' traces are silently absent — a short table with no empty state and no warning. The e2e guard won't catch it either: `waitForLogsTraceRows` asserts an exact count because *"more would mean the scope leaked"*, and this fails in the **fewer** direction.

Cheap to make visible: `uniq(compact(experiments.map((e) => e.project_id)))`, and if that's longer than 1, say the comparison spans multiple projects rather than showing a subset as if it were the whole.

apps/opik-frontend/src/v2/pages/CompareExperimentsPage/ExperimentLogsTab/ExperimentLogsTab.tsx
```diff
@@ -0,0 +1,74 @@
...
+  TraceLogsViewConfig,
+} from "@/v2/pages-shared/traces/TraceLogsView/TraceLogsView";
+
+// Per design, the tab's toolbar carries the columns selector and nothing else — no row-height, date
+// range or refresh. Dropping the date control also drops date filtering entirely, so the tab always
+// spans the experiment's whole life.
+const EXPERIMENT_LOGS_VIEW_CONFIG: TraceLogsViewConfig = {
+  ...DEFAULT_TRACE_LOGS_VIEW_CONFIG,
+  showTableControls: false,
+};
```
miguelgrc, line 26: 💡 **suggestion** | State isolation

This config spreads `DEFAULT_TRACE_LOGS_VIEW_CONFIG`, which carries `storageNamespace: ""`. That makes `storagePrefix` = `tls-traces-` and `tableId` = `logs.entity-traces.default` — byte-identical to the playground, optimization-trial, and annotation-queue overlays, which all take the default config too.

That's the case layer 2 says it prevents. `TraceLogsView.tsx:554-556`: *"Pinned-chip config is per view so the experiment tab and the playground overlay don't fight over one arrangement."* The experiment tab is the literal example given, and it isn't isolated — pinned chips, selected columns, column order and width, sort, page size, and row height are all shared with those three overlays today.

```suggestion
const EXPERIMENT_LOGS_VIEW_CONFIG: TraceLogsViewConfig = {
  ...DEFAULT_TRACE_LOGS_VIEW_CONFIG,
  storageNamespace: "experiment-",
  showTableControls: false,
};
```

apps/opik-frontend/src/v2/pages/CompareExperimentsPage/ExperimentLogsTab/ExperimentLogsTab.tsx
```diff
@@ -0,0 +1,74 @@
...
+    );
+  }
+
+  return (
+    <TraceLogsView
+      projectId={projectId}
+      scopeFilters={scopeFilters}
+      viewConfig={EXPERIMENT_LOGS_VIEW_CONFIG}
+      layout="page"
+    />
```
miguelgrc, line 98: 💡 **suggestion** | Filter consistency

No `logsSource` is passed, and the two paths downstream disagree about what that means. `TraceLogsView` forwards it to `useTracesList`/`useTracesStatistic`, where `undefined` means *no `source` filter at all* — the table shows traces of every source. But it also forwards it to `buildTraceChipDefinitions`, and `buildSharedDynamicChips` defaults it to `LOGS_SOURCE.sdk`, so the Tags / Error type / Metadata-key autocompletes are computed over `source = "sdk"` only.

One of the two is wrong. If evaluation traces don't carry `sdk`, those dropdowns come back empty on a tab whose table has rows; if they do, the table should be filtered the same way. This is the layer-1 default landing in practice — see the comment on #7841 about making that parameter required.

miguelgrc:
Seems like there's an issue with your rebase / gitignore

CometActions:
**No test needed here.**

The Logs tab ships with its own permanent coverage in this PR: experiments-smoke.spec.ts (@t1-smoke) now asserts the old "Go to logs" tag is gone, opens the tab, waits for tab=logs and requires exactly one trace row per seeded item — so a dead tab or a leaked experiment scope fails the smoke run. The tab-list refactor (getAvailableExperimentTabs / isExperimentTabId) is already guarded by experiments-compare.spec.ts, which deep-links to ?tab=config and ?tab=scores and asserts their contents; if the new tab set or the URL narrowing dropped either, the page would fall back to Results and those steps would fail. The one piece nothing checks is the playground output-cell link now opening ?tab=logs&tls_trace=… instead of the sidebar params, but that needs a provider key and a dataset run to reach, so it isn't something a fresh OSS env would cover either — worth one manual click before merge. Heads-up: experiments.logs-tab doesn't exist in main's taxonomy.yaml yet, so it only starts counting once this lands.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/31832721272)

_Re-checked after a push on 14 Aug 19:20 UTC — nothing the verdict depends on changed._

awkoy:
All eleven threads addressed and resolved.

**@miguelgrc — the project invariant.** You're right and the backend trace settles it: an experiment resolves its project from its own `project_name`, and `DatasetService.getOrCreate` falls back to a project-agnostic dataset lookup because `datasets_workspace_id_name_uk` was never widened when `project_id` arrived in `000057`. So a comparison genuinely can span two projects, and my comment claiming otherwise was wrong. Since traces are read per project, one request can't serve that — the tab now detects more than one distinct project and says so explicitly, rather than quietly showing only the first project's traces.

**@miguelgrc — shared storage namespace.** Also right, and it made layer 2's own comment a lie: the tab spread the default config, so `storagePrefix` and `tableId` were byte-identical to the playground, trial and annotation-queue overlays. Columns, order, width, sort, page size, row height and pinned chips were all shared. The tab now takes `storageNamespace: "experiment-"`. Trade-off worth naming: users who had set columns up in an overlay start from defaults on the tab, which I think is right for a new surface and is what the code claimed all along.

**@miguelgrc — `logsSource` disagreement.** Fixed one layer down, in #7841: `buildSharedDynamicChips` no longer defaults to `LOGS_SOURCE.sdk`; the parameter is required (nullable) so callers must state it. The tab passes nothing, which now means "no source filter" on both paths — the table and the autocompletes agree.

**Bot findings.** Fixed: the malformed-`experimentsIds` guard, coverage for `isExperimentTabId` at the URL boundary, and `toHaveCount` in place of the hand-rolled poll (better diagnostics, as suggested). Already fixed: the playground deep link now emits `tab=logs` rather than the removed `tls_open` overlay params.

Two I'm not acting on, with reasons: `tbody tr[data-row-id]` is the established POM idiom across this suite and the locator is now scoped to the tab panel, so a Logs-specific test id would be a one-off; and a dedicated e2e for the playground trace link would need playground run fixtures the suite doesn't have — the link's destination is covered by the Logs-tab scenario. Happy to add either if you'd rather.

miguelgrc (dismissed):

awkoy:
Fair — comparison and test-suite Logs are verified only manually right now (browser, against seeded fixtures: 8/8 rows for a two-experiment comparison, and the test-suite tab set showing `Results · Configuration · Logs` with Insights and Feedback scores hidden). The smoke spec covers the single-experiment path.

Adding tagged scenarios on top of `comparison-experiment.fixture.ts` is the right follow-up; I'd rather do it as its own change than grow this stack further, since each new e2e scenario costs real CI minutes on `@t1-smoke`. Say the word if you'd prefer it in here.

JetoPistola (approved):
Restoring approval after the restack. Verified all 9 files at 878a5258 are byte-identical to ba6b185b, which was approved by @miguelgrc — SHAs moved, content did not.

PR #7851, opened 2026-08-14
[NA] [DOCS] fix: restore the star history chart

Dessalines39394:
The Star History chart in the project READMEs is currently broken: the image no longer renders, leaving a broken placeholder on the GitHub project page.

This change points the chart to a working mirror in all five READMEs (English, French, German, Chinese, and Spanish) so the stargazer history displays again. Only the chart image is updated; the rest of the README content is untouched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -376,7 +376,7 @@ Opik allows you to evaluate your LLM application during development through [Dat
 
 If you find Opik useful, please consider giving us a star! Your support helps us grow our community and continue improving the product.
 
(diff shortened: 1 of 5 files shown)


Nimrod007:
Thanks for flagging this, and sorry for the slow reply. You're right that the chart is broken — GitHub restricted stargazer history to repo admins and collaborators on June 30, so the upstream endpoint now returns a notice image instead of the chart. We hadn't noticed until your PR, so this was genuinely useful.

We're going to fix it on our side, though, so I'm going to close this one.

We're staying with our existing Star History integration and will restore the chart properly by configuring it on our end.

Appreciate you taking the time to open this across all the READMEs, and we'd welcome contributions elsewhere in the project.

PR #7852, opened 2026-08-14
[OPIK-7919] [BE] perf: engage traces_final_ids prefilter for thread-id lookups

thiagohora:
## Details

`search_threads` reads every span in the project even when the request targets a single thread — ~6M rows / 1.2 GiB and ~5s for a `LIMIT 1` lookup on production. `spans_deduped` is only restricted by trace id when the `traces_final_ids` attribute is set, and the gate for it fired solely on `searchText` or the `filters` attribute; a thread-ID filter is a `TRACE_THREAD`-strategy filter, so it sets `trace_thread_filters` / `traces_pushdown_filter` and never `filters`. The prefilter therefore never engaged for a thread-scoped lookup, and the whole project's spans were deduped before being LEFT JOINed against one thread's traces.

- This adds `traces_pushdown_filter` to the gate. `traces_final_ids` already renders the same `thread_id = :thread_id_pushdown` predicate, so no SQL change is needed — only the gate was wrong.
- Deliberately not `trace_thread_filters` at large: the other TRACE_THREAD filters (status, tags, ...) are applied by the outer query, not inside `traces_final_ids`, so they would pay for the prefilter scan without narrowing it. Covered by a test.
- The gate is shared with `find_threads_by_project`, `thread_stats` and `count_threads_by_project`, but the shared gate alone was not enough for the count path: `SELECT_COUNT_TRACES_THREADS_BY_PROJECT_IDS`'s `trace_threads_final` was the only one of the three templates missing `<if(traces_pushdown_filter)> AND thread_id = :thread_id_pushdown <endif>`, so with a `uuid_from_time` range it took the uuid branch, skipped the `traces_final_ids` IN and never got the equality — scanning every `trace_threads` row of the project. This adds that line so all three templates emit the pushdown identically.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-7919

## AI-WATERMARK

AI-WATERMARK: yes

- If yes:
  - Tools: Claude Code
  - Model(s): Claude Opus 5 (`claude-opus-5`, 1M context)
  - Scope: Whole change authored by the agent — production `query_log` profiling that found the defect, root-cause tracing through the template gate, the gate change, the tests, the production A/B replay harness, and this PR description.
  - Human verification: The human directed the investigation, chose the validation scope (multiple workspace and project sizes before opening a PR), and instructed the commit and PR. Test runs and benchmarks were executed and their output read by the agent. **The diff itself has not yet been reviewed line by line by a human — that review is still owed before merge.**

## Testing

Commands run, from `apps/opik-backend`:

- `mvn test -Dtest="FindTraceThreadsResourceTest,ThreadDAOImplTest"` → **478 tests, 0 failures, 0 errors**
- `mvn spotless:check` → exit 0
- `mvn compile -DskipTests` → exit 0

Scenarios validated:

- **Regression guard.** `ThreadDAOImplTest$TracesFinalIdsPrefilterGate` asserts that a `thread_id` EQUAL filter leaves `filters` unset (the exact reason the old gate missed it), turns the gate on, and renders `AND trace_id IN (SELECT id FROM traces_final_ids)` into the spans CTE. The helper mirrors the production wiring — the `traces_final_ids` attribute is added only when the gate returns true — so the rendered-SQL assertions are a function of the gate rather than of the test setup. Verified discriminating in both directions: reverting the gate to its pre-fix condition fails `threadIdEqualFilterActivatesPrefilter`; hardcoding the gate to `true` fails all three negative cases. Negative cases cover a non-pushdown TRACE_THREAD filter (status), an ID filter with a non-EQUAL operator (CONTAINS, which must not reach the EQUAL-only pushdown), and no predicate at all.
- **Production A/B, 9 workspaces / 9 projects.** Real `search_threads` queries captured from `system.query_log` and replayed unmodified against their own workspaces, baseline vs patched, spanning 168K → 37M rows read and limits 1 → 2000. Output compared by sha1 over the full TSV result — **byte-identical in every case, in both run orders**.

  | case | thread filter | baseline | patched | rows read (base → patched) |
  |---|---|---|---|---|
  | C05 | yes | 4465 ms | **115 ms** | 6.55M → 519K |
  | C08 | yes | 24121 ms | **188 ms** | 20.98M → 529K |
  | C01–C04, C06, C07, C09 | no | 106 ms – 57.3 s | neutral | unchanged |

- **Cache control.** Every case was re-run with the patched variant cold-first. The apparent gains on the no-thread-filter cases are page-cache artifacts — `read_rows` is unchanged there, confirming the prefilter correctly cannot help when `traces_final` is not narrowed. Reversed order: C04 1.399s (patched) vs 1.460s (baseline), C06 6.051 vs 6.212, C05 0.122 vs 4.816, C08 0.198 vs 22.078. The change is a large win where it applies and neutral everywhere else.
- **As-generated SQL, not a hand-patch.** The measurements above were re-run using the SQL the DAO actually emits once the gate fires (a real `traces_final_ids` CTE, not an inlined predicate): 0.107s vs 4.854s (C05) and 0.162s vs 18.478s (C08), same byte-identical output.
- **Count-template regression guard.** `ThreadDAOImplTest$ThreadIdPushdownTemplateSync` renders all three templates that share the `traces_pushdown_filter` concept (list, count, stats) through the same gate with a `uuid_from_time` range, slices out the `trace_threads_final` CTE and asserts it takes the uuid branch, omits the `traces_final_ids` IN, and still emits `AND thread_id = :thread_id_pushdown`. Verified discriminating: deleting the new line from the count template fails only the `count` case, list and stats stay green.
- **Count-template change verified separately on production.** The A/B above is captured `search_threads` traffic, which exercises the list path, not `countThreadTotal()`. So the count fix was measured on its own:
  - Full query: a real `count_threads_by_project` query captured from `system.query_log` (uuid range + `source` filter), replayed with a thread-ID filter added, with and without the new line — **identical result (`count=1`) both ways**.
  - The `trace_threads_final` CTE in isolation, on the project with the most threads (5.8M `trace_threads` rows), patched run cold first:

    | variant | time | rows read | bytes | peak memory | selected marks |
    |---|---|---|---|---|---|
    | baseline | 2917 ms | 5,881,000 | 759.81 MiB | 707.60 MiB | 719 |
    | patched | **5 ms** | **81,060** | **4.74 MiB** | **83.74 KiB** | **11** |

    `trace_threads` is `ORDER BY (workspace_id, project_id, thread_id, id)`, so the equality is primary-key prunable while `id >= :uuid_from_time` is not — marks drop 719 → 11.
- **Not run:** the full backend suite. The gate is exercised by four call sites in `ThreadDAO`; the thread resource suite plus the new unit tests cover them.

Environment: local process mode, macOS, testcontainers via Docker 29.7.2. Production replay ran read-only against `opik_prod` as the `agentro` read-only user.

## Documentation

No documentation update needed — internal DAO query-planning change with no API, schema, or SDK surface.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
@@ -79,7 +79,8 @@ class ThreadDAOImpl implements ThreadDAO {
      * <p>
      * Please refer to the SELECT_TRACES_THREAD_BY_ID query for more details.
      ***/
-    private static final String SELECT_TRACES_THREADS_BY_PROJECT_IDS = """
+    @VisibleForTesting
+    static final String SELECT_TRACES_THREADS_BY_PROJECT_IDS = """
             WITH <if(traces_final_ids)>traces_final_ids AS (
                 SELECT DISTINCT id, thread_id
                 FROM (
@@ -1360,11 +1361,26 @@ class ThreadDAOImpl implements ThreadDAO {
      * in the thread list and count queries. Mirrors {@code TraceDAO#shouldUseTraceIdPrefilter}.
      *
      * <p>Only activates when there are narrowing predicates beyond the workspace/project/uuid-range filters:
-     * either a free-text search or a TRACE-strategy filter ({@code <filters>}). When neither is present,
-     * the downstream CTEs apply the uuid range directly, skipping the prefilter scan.
+     * a free-text search, a TRACE-strategy filter ({@code <filters>}), or the thread_id EQUAL pushdown
+     * ({@code traces_pushdown_filter}). When none is present, the downstream CTEs apply the uuid range
+     * directly, skipping the prefilter scan.
+     *
+     * <p>OPIK-7919: a thread_id EQUAL filter is a TRACE_THREAD-strategy filter, so it sets
+     * {@code trace_thread_filters} / {@code traces_pushdown_filter} but never {@code filters}. Without the
+     * third condition the prefilter stayed off for every single-thread lookup, and spans_deduped deduped
+     * the whole project before being LEFT JOINed against one thread's traces — 6M rows / 1.2 GiB and ~5s
+     * for a LIMIT 1 read on production. traces_final_ids already renders the same thread_id predicate, so
+     * enabling it here narrows the spans scan to that thread's trace ids.
+     *
+     * <p>Only {@code traces_pushdown_filter} is included, not {@code trace_thread_filters} at large: the
+     * other TRACE_THREAD filters (status, tags, ...) are applied by the outer query, not inside
+     * traces_final_ids, so they would not narrow the prefilter scan and would pay for it for nothing.
      */
-    private static boolean shouldUseTracesFinalIdsPrefilter(TraceSearchCriteria criteria, ST template) {
-        return criteria.searchText() != null || template.getAttribute("filters") != null;
+    @VisibleForTesting
+    static boolean shouldUseTracesFinalIdsPrefilter(TraceSearchCriteria criteria, ST template) {
+        return criteria.searchText() != null
+                || template.getAttribute("filters") != null
+                || template.getAttribute("traces_pushdown_filter") != null;
     }
 
     /**
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
@@ -495,7 +495,8 @@ class ThreadDAOImpl implements ThreadDAO {
      * <p>
      * Please refer to the SELECT_TRACES_THREAD_BY_ID query for more details.
      ***/
-    private static final String SELECT_COUNT_TRACES_THREADS_BY_PROJECT_IDS = """
+    @VisibleForTesting
+    static final String SELECT_COUNT_TRACES_THREADS_BY_PROJECT_IDS = """
             WITH <if(traces_final_ids)>traces_final_ids AS (
                 SELECT DISTINCT id, thread_id
                 FROM (
@@ -1014,7 +1015,8 @@ class ThreadDAOImpl implements ThreadDAO {
      * 1. First level: Uses the same thread aggregation as SELECT_TRACES_THREADS_BY_PROJECT_IDS (reusing the exact CTEs and aggregation logic)
      * 2. Second level: Wraps the thread results and calculates stats across all threads (AVG, SUM, quantiles)
      ***/
-    private static final String SELECT_TRACE_THREADS_STATS = """
+    @VisibleForTesting
+    static final String SELECT_TRACE_THREADS_STATS = """
             SELECT
                 threads.workspace_id as workspace_id,
                 threads.project_id as project_id,
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
@@ -567,6 +567,7 @@ class ThreadDAOImpl implements ThreadDAO {
                         AND thread_id IN (SELECT thread_id FROM traces_final_ids)
                     <endif>
                 <endif>
+                <if(traces_pushdown_filter)> AND thread_id = :thread_id_pushdown <endif>
                 ORDER BY (workspace_id, project_id, thread_id, id) DESC, last_updated_at DESC
                 LIMIT 1 BY id
             )
--- a/apps/opik-backend/src/test/java/com/comet/opik/domain/ThreadDAOImplTest.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/domain/ThreadDAOImplTest.java
@@ -8,12 +8,16 @@ import com.comet.opik.infrastructure.FilterUtils;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 import org.stringtemplate.v4.ST;
 import reactor.core.publisher.Flux;
 import reactor.test.StepVerifier;
 
 import java.util.List;
 import java.util.UUID;
+import java.util.stream.Stream;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
@@ -148,6 +152,65 @@ class ThreadDAOImplTest {
         }
 
         private record Rendered(ST template, boolean gateOpen) {
+        }
+    }
+
+    @Nested
+    @DisplayName("thread_id pushdown across the templates that share it")
+    class ThreadIdPushdownTemplateSync {
+
+        private static final String THREAD_ID_PUSHDOWN = "AND thread_id = :thread_id_pushdown";
+        private static final String TRACES_FINAL_IDS_IN = "thread_id IN (SELECT thread_id FROM traces_final_ids)";
+        private static final String SEARCH_CLAUSE = "ilike(thread_id, :search_text)";
+
+        static Stream<Arguments> templatesSharingThePushdown() {
+            return Stream.of(
+                    Arguments.of("list", ThreadDAOImpl.SELECT_TRACES_THREADS_BY_PROJECT_IDS),
+                    Arguments.of("count", ThreadDAOImpl.SELECT_COUNT_TRACES_THREADS_BY_PROJECT_IDS),
+                    Arguments.of("stats", ThreadDAOImpl.SELECT_TRACE_THREADS_STATS));
+        }
+
+        /**
+         * OPIK-7919: on the uuid_from_time branch trace_threads_final skips the traces_final_ids IN, so the
+         * thread_id equality is the only predicate left that prunes — and trace_threads is
+         * ORDER BY (workspace_id, project_id, thread_id, id), so it prunes on the primary key while
+         * {@code id >= :uuid_from_time} cannot. The count template was missing this line while the list and
+         * stats templates had it, which left countThreadTotal scanning every trace_threads row of the
+         * project. This pins all three templates to emit it identically.
+         */
+        @ParameterizedTest(name = "{0} template")
+        @MethodSource("templatesSharingThePushdown")
+        @DisplayName("trace_threads_final emits the thread_id pushdown on the uuid_from_time branch")
+        void traceThreadsFinalEmitsThreadIdPushdownOnUuidBranch(String name, String query) {
+            var criteria = TraceSearchCriteria.builder()
+                    .projectId(UUID.randomUUID())
+                    .uuidFromTime(UUID.randomUUID())
+                    .filters(List.of(TraceThreadFilter.builder()
+                            .field(TraceThreadField.ID)
+                            .operator(Operator.EQUAL)
+                            .value("thread-1")
+                            .build()))
+                    .build();
+
+            var template = FilterUtils.newTraceThreadFindTemplate(query, criteria, SEARCH_CLAUSE, true);
+            if (ThreadDAOImpl.shouldUseTracesFinalIdsPrefilter(criteria, template)) {
+                template.add("traces_final_ids", true);
+            }
+
+            var traceThreadsFinal = traceThreadsFinalCte(template.render());
+
+            assertThat(traceThreadsFinal).contains("AND id >= :uuid_from_time");
+            assertThat(traceThreadsFinal).doesNotContain(TRACES_FINAL_IDS_IN);
+            assertThat(traceThreadsFinal).contains(THREAD_ID_PUSHDOWN);
+        }
+
+        /** The trace_threads_final CTE body, up to its ORDER BY — so the assertions cannot match another CTE. */
+        private static String traceThreadsFinalCte(String sql) {
+            int start = sql.indexOf("trace_threads_final AS (");
+            assertThat(start).isNotNegative();
+            int end = sql.indexOf("ORDER BY (workspace_id, project_id, thread_id, id)", start);
+            assertThat(end).isGreaterThan(start);
+            return sql.substring(start, end);
         }
     }
 }
--- a/apps/opik-backend/src/test/java/com/comet/opik/domain/ThreadDAOImplTest.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/domain/ThreadDAOImplTest.java
@@ -8,6 +8,7 @@ import com.comet.opik.infrastructure.FilterUtils;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
+import org.stringtemplate.v4.ST;
 import reactor.core.publisher.Flux;
 import reactor.test.StepVerifier;
 
@@ -51,27 +52,43 @@ class ThreadDAOImplTest {
     class TracesFinalIdsPrefilterGate {
 
         private static final String SPANS_PREFILTER = "AND trace_id IN (SELECT id FROM traces_final_ids)";
-        // searchClause is only consulted when searchText is set; these cases never set it.
-        private static final String UNUSED_SEARCH_CLAUSE = "ilike(thread_id, :search_text)";
+        // Rendered into the query only when searchText is set (FilterUtils#newTraceThreadFindTemplate);
+        // the searchText case below is the one that exercises it.
+        private static final String SEARCH_CLAUSE = "ilike(thread_id, :search_text)";
 
-        private static ThreadCriteriaAndTemplate render(List<TraceThreadFilter> filters) {
+        /**
+         * Mirrors the production wiring in {@link ThreadDAOImpl#search}: the {@code traces_final_ids}
+         * attribute is added only when the gate says so, so the rendered SQL asserted below is a function
+         * of the gate's result rather than of the test's own setup.
+         */
+        private static Rendered render(List<TraceThreadFilter> filters, String searchText) {
             var criteria = TraceSearchCriteria.builder()
                     .projectId(UUID.randomUUID())
                     .filters(filters)
+                    .searchText(searchText)
                     .build();
             var template = FilterUtils.newTraceThreadFindTemplate(
-                    ThreadDAOImpl.SELECT_TRACES_THREADS_BY_PROJECT_IDS, criteria, UNUSED_SEARCH_CLAUSE, true);
-            return new ThreadCriteriaAndTemplate(criteria, template);
+                    ThreadDAOImpl.SELECT_TRACES_THREADS_BY_PROJECT_IDS, criteria, SEARCH_CLAUSE, true);
+
+            boolean gateOpen = ThreadDAOImpl.shouldUseTracesFinalIdsPrefilter(criteria, template);
+            if (gateOpen) {
+                template.add("traces_final_ids", true);
+            }
+            return new Rendered(template, gateOpen);
+        }
+
+        private static Rendered renderWithFilter(TraceThreadField field, Operator operator, String value) {
+            return render(List.of(TraceThreadFilter.builder()
+                    .field(field)
+                    .operator(operator)
+                    .value(value)
+                    .build()), null);
         }
 
         @Test
         @DisplayName("OPIK-7919: a thread_id EQUAL filter turns the prefilter on and narrows the spans scan")
         void threadIdEqualFilterActivatesPrefilter() {
-            var rendered = render(List.of(TraceThreadFilter.builder()
-                    .field(TraceThreadField.ID)
-                    .operator(Operator.EQUAL)
-                    .value("thread-1")
-                    .build()));
+            var rendered = renderWithFilter(TraceThreadField.ID, Operator.EQUAL, "thread-1");
 
             // The regression this guards: a thread-ID filter is a TRACE_THREAD-strategy filter, so it never
             // sets "filters". Gating only on "filters" left the prefilter off and spans_deduped scanned the
@@ -79,11 +96,20 @@ class ThreadDAOImplTest {
             assertThat(rendered.template().getAttribute("filters")).isNull();
             assertThat(rendered.template().getAttribute("traces_pushdown_filter")).isNotNull();
 
-            assertThat(ThreadDAOImpl.shouldUseTracesFinalIdsPrefilter(rendered.criteria(), rendered.template()))
-                    .isTrue();
-
-            rendered.template().add("traces_final_ids", true);
+            assertThat(rendered.gateOpen()).isTrue();
             assertThat(rendered.template().render()).contains(SPANS_PREFILTER);
+        }
+
+        @Test
+        @DisplayName("an ID filter with a non-EQUAL operator does not reach the pushdown or the prefilter")
+        void threadIdContainsFilterDoesNotActivatePrefilter() {
+            // findTraceThreadIdPushdownFilter is EQUAL-only because the pushdown SQL is hardcoded to
+            // `thread_id = :thread_id_pushdown`; a CONTAINS filter answered as an equality would be wrong.
+            var rendered = renderWithFilter(TraceThreadField.ID, Operator.CONTAINS, "thread");
+
+            assertThat(rendered.template().getAttribute("traces_pushdown_filter")).isNull();
+            assertThat(rendered.gateOpen()).isFalse();
+            assertThat(rendered.template().render()).doesNotContain(SPANS_PREFILTER);
         }
 
         @Test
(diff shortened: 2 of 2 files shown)


CometActions:
**No test needed here.**

Perf-only, and the path it touches is already asserted against real ClickHouse. The gate now enables traces_final_ids for the thread_id EQUAL pushdown, and every CTE that redirects (traces_final, spans_deduped, trace_threads_final) narrows to the same trace/thread set the un-prefiltered branch already produced, so the Threads tab filtered by Thread ID should return identical rows — only the scan changes, and scan time is not something an e2e run on a seeded OSS install can measure. FindTraceThreadsResourceTest covers exactly that path with real data: ID/EQUAL in getValidFilters (paged + stream) and in getValidFiltersForStats for /threads/stats. One gap that belongs in that test rather than in an e2e run: those ID-filter cases build traces with usage(null) and create no spans, so spans_deduped — the CTE this change actually narrows — runs against an empty spans table. Adding spans with usage to the ID/EQUAL case would cover the usage/cost aggregation the prefilter now feeds.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/31803605817)

_Re-checked after a push on 14 Aug 13:12 UTC — nothing the verdict depends on changed._

apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
```diff
@@ -1360,11 +1361,26 @@ thread_model_id NOT IN (SELECT entity_id FROM fsc)
...
+     * other TRACE_THREAD filters (status, tags, ...) are applied by the outer query, not inside
+     * traces_final_ids, so they would not narrow the prefilter scan and would pay for it for nothing.
      */
-    private static boolean shouldUseTracesFinalIdsPrefilter(TraceSearchCriteria criteria, ST template) {
-        return criteria.searchText() != null || template.getAttribute("filters") != null;
+    @VisibleForTesting
+    static boolean shouldUseTracesFinalIdsPrefilter(TraceSearchCriteria criteria, ST template) {
+        return criteria.searchText() != null
+                || template.getAttribute("filters") != null
+                || template.getAttribute("traces_pushdown_filter") != null;
```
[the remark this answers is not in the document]
thiagohora: The mechanism is real, so I measured it rather than reasoning about it. Exact counts against production `opik_prod.spans`:

| population | distinct span ids | ids living on >1 `trace_id` |
|---|---|---|
| all workspaces, last 24h | 48,281,874 | 6,940 (**0.0144%**) |
| workspace/project carrying 96% of affected `search_threads` traffic, full history | 4,217,668 | **0** |
| the other project A/B'd for this PR, full history | 20,265,080 | **0** |

So it does occur, but a span id under two `trace_id`s is a client-side ingestion mistake — a span belongs to exactly one trace. The un-narrowed `LIMIT 1 BY id` isn't preserving a correct project-wide semantic; it silently compensates for that bad data by letting the higher-sorted `trace_id` win, which **removes** the span from the aggregate of the trace it is actually recorded on. Narrowing attributes it to its own trace.

Two further reasons not to preserve the old winner selection:

- `ThreadDAO.SELECT_TRACES_THREAD_BY_ID` already narrows spans by trace id (`AND trace_id IN (SELECT DISTINCT id FROM traces_ids)`), so `find_thread_by_id` already returns the narrowed semantics. Today `search_threads` can disagree with it on usage/cost for the same thread; this change makes them agree.
- The `traces_final_ids` prefilter already ships on this same constant whenever `searchText` or a TRACE-strategy filter is present. The narrowed semantics is pre-existing — this PR widens when it applies, it does not introduce it.

Deduplicating before narrowing would keep the project-wide winner selection *and* reinstate the whole-project spans scan, which is the entire cost this PR removes (~6M rows / 1.2 GiB and ~5s for a `LIMIT 1` read).

On the fixture: not adding one. The test here is a regression guard for the gate itself, and equivalence of the two SQL shapes was verified by replaying real production queries across 9 workspaces/projects (168K–37M rows read, both run orders) with byte-identical output. A fixture asserting behaviour on data that shouldn't exist would pin in the client-side anomaly rather than guard against a regression.

apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
```diff
@@ -79,7 +79,8 @@ OR ilike(input, :search_text)
      * <p>
      * Please refer to the SELECT_TRACES_THREAD_BY_ID query for more details.
      ***/
-    private static final String SELECT_TRACES_THREADS_BY_PROJECT_IDS = """
+    @VisibleForTesting
+    static final String SELECT_TRACES_THREADS_BY_PROJECT_IDS = """
```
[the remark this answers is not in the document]
thiagohora: Agreed that the plural name misstates the scope — the query binds a single `:project_id`. It's pre-existing though, and untouched by this PR apart from the `@VisibleForTesting` modifier needed to reach it from the gate test. Leaving the rename as a separate cleanup so this stays a perf-only diff.

apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
miguelgrc, line 1386: **[medium] Count query still misses the thread_id pushdown in trace_threads_final, so half the OPIK-7919 request stays slow**

The new gate is documented as covering "the thread list and count queries", and find() always runs countThreadTotal() before the page query, so both halves of a single-thread lookup must be narrowed. But the three templates that share the traces_pushdown_filter concept are not in sync: SELECT_TRACES_THREADS_BY_PROJECT_IDS (line 239) and SELECT_TRACE_THREADS_STATS (line 1147) both emit `<if(traces_pushdown_filter)> AND thread_id = :thread_id_pushdown <endif>` in trace_threads_final unconditionally, while SELECT_COUNT_TRACES_THREADS_BY_PROJECT_IDS (lines 546-571) has no such line at all. Without a uuid range the new prefilter partially compensates there (`thread_id IN (SELECT thread_id FROM traces_final_ids)`), but as soon as uuidFromTime is set — the normal case when the UI sends a time range — trace_threads_final takes the `<if(uuid_from_time)>` branch, which skips the traces_final_ids IN and never gets the equality. trace_threads is ORDER BY (workspace_id, project_id, thread_id, id) (migration 000028), so `thread_id = :thread_id_pushdown` is primary-key prunable while `id >= :uuid_from_time` is not; the count query therefore keeps scanning every trace_threads row of the project for a single-thread read that the list query now resolves cheaply.

Note on the PR description: bullet 3 states "The gate is shared, so `find_threads_by_project`, `thread_stats` and `count_threads_by_project` get the same narrowing when a thread-ID filter is present." The shared gate does make the `traces_final_ids` CTE render for all three, but the count template contains no line that consumes it on the `uuid_from_time` branch, so it does not in fact get the same narrowing. The A/B replay would not have surfaced this: C05/C08 are captured `search_threads` queries, which exercise the list path, not `countThreadTotal()`.

💡 Add `<if(traces_pushdown_filter)> AND thread_id = :thread_id_pushdown <endif>` to trace_threads_final in SELECT_COUNT_TRACES_THREADS_BY_PROJECT_IDS (after the uuid if/else, mirroring line 239 and line 1147), so the pushdown predicate is emitted identically by all three templates.

_rule: `backend-data-access-eliminate-duplicated-sql-dao-logic-by-extracting-shared-ctes`_

apps/opik-backend/src/test/java/com/comet/opik/domain/ThreadDAOImplTest.java
```diff
@@ -37,4 +45,89 @@ void firstThreadOrEmptyCompletesEmptyForEmptyStream() {
...
+        }
+
+        @Test
+        @DisplayName("no narrowing predicate leaves the prefilter off and the spans scan unrestricted")
+        void noNarrowingPredicateLeavesPrefilterOff() {
+            var rendered = render(null);
+
+            assertThat(ThreadDAOImpl.shouldUseTracesFinalIdsPrefilter(rendered.criteria(), rendered.template()))
+                    .isFalse();
+            assertThat(rendered.template().render()).doesNotContain(SPANS_PREFILTER);
```
miguelgrc, line 139: **[low] Vacuous assertion: doesNotContain(SPANS_PREFILTER) can never fail**

`assertThat(rendered.template().render()).doesNotContain(SPANS_PREFILTER)` is a tautology on this template: the `traces_final_ids` attribute is never added in this test, so the `<if(traces_final_ids)>` block cannot render regardless of what the gate returned. The assertion passes even if shouldUseTracesFinalIdsPrefilter were changed to always return true, so it adds no discriminating power over the isFalse() assertion above it.

💡 Either drop the render assertion, or make it discriminating by mirroring the production wiring — conditionally add the attribute based on the gate result and then assert the rendered SQL, so a wrong gate result changes the rendered output.

_rule: `backend-testing-make-test-assertions-precise-unconditional-and-discriminatin`_

apps/opik-backend/src/test/java/com/comet/opik/domain/ThreadDAOImplTest.java
```diff
@@ -37,4 +45,89 @@ void firstThreadOrEmptyCompletesEmptyForEmptyStream() {
...
                 .verifyComplete();
     }
+
+    @Nested
+    @DisplayName("traces_final_ids prefilter gate")
+    class TracesFinalIdsPrefilterGate {
+
+        private static final String SPANS_PREFILTER = "AND trace_id IN (SELECT id FROM traces_final_ids)";
+        // searchClause is only consulted when searchText is set; these cases never set it.
+        private static final String UNUSED_SEARCH_CLAUSE = "ilike(thread_id, :search_text)";
```
miguelgrc: **[low] `UNUSED_SEARCH_CLAUSE` and its comment contradict the searchText test**

The constant is named UNUSED_SEARCH_CLAUSE and commented "searchClause is only consulted when searchText is set; these cases never set it", but searchTextActivatesPrefilter (line 119) does set searchText("needle"), and FilterUtils.newTraceThreadFindTemplate then does `template.add("search_text", searchClause)` — so the clause is genuinely consumed and rendered into the SQL. The name and comment mislead the next reader into thinking the value is inert everywhere.

💡 Rename to SEARCH_CLAUSE and reword the comment to say the clause is only rendered when searchText is set (which the last case does), or move the constant/comment scope to the cases where it really is unused.

_rule: `backend-docs-keep-comments-and-javadoc-synchronized-with-the-code-they-de`_

apps/opik-backend/src/test/java/com/comet/opik/domain/ThreadDAOImplTest.java
```diff
@@ -37,4 +45,89 @@ void firstThreadOrEmptyCompletesEmptyForEmptyStream() {
...
+        void noNarrowingPredicateLeavesPrefilterOff() {
+            var rendered = render(null);
+
+            assertThat(ThreadDAOImpl.shouldUseTracesFinalIdsPrefilter(rendered.criteria(), rendered.template()))
+                    .isFalse();
+            assertThat(rendered.template().render()).doesNotContain(SPANS_PREFILTER);
+        }
+
+        @Test
+        @DisplayName("a free-text search still turns the prefilter on")
```
miguelgrc: **[low] "a free-text search still turns the prefilter on" overstates what holds on the list path**

The DisplayName claims a search enables the prefilter, but in ThreadDAOImpl.find() a search-only criteria is page-pushdown eligible (`search_text` is not in PAGE_PUSHDOWN_DISQUALIFIERS), so the `else if (shouldUseTracesFinalIdsPrefilter(...))` branch is never reached and traces_final_ids is NOT engaged for the listing query. The statement is only true of the helper in isolation (and of search/stats/count). Unlike its sibling test it also asserts nothing about the rendered SQL, so it cannot catch the difference.

💡 Reword to scope the claim to the gate (e.g. "a free-text search satisfies the prefilter gate") and, if the listing behaviour matters, add a case asserting the page-pushdown/prefilter mutual exclusion in find().

_rule: `backend-testing-use-descriptive-scenario-based-test-method-and-displayname-n`_

apps/opik-backend/src/test/java/com/comet/opik/domain/ThreadDAOImplTest.java
```diff
@@ -37,4 +45,89 @@ void firstThreadOrEmptyCompletesEmptyForEmptyStream() {
...
+            rendered.template().add("traces_final_ids", true);
+            assertThat(rendered.template().render()).contains(SPANS_PREFILTER);
+        }
+
+        @Test
+        @DisplayName("a TRACE_THREAD filter that is not the id pushdown leaves the prefilter off")
+        void nonPushdownThreadFilterDoesNotActivatePrefilter() {
+            // status/tags/... are applied by the outer query, not inside traces_final_ids, so enabling the
+            // prefilter for them would pay for the extra traces scan without narrowing it.
+            var rendered = render(List.of(TraceThreadFilter.builder()
```
miguelgrc: **[low] The operator boundary of the new condition — ID with a non-EQUAL operator — is not covered**

The PR description notes the STATUS negative case is deliberate — it pins the decision not to gate on `trace_thread_filters` at large — and it does cover that. This is a separate, uncovered case rather than a criticism of that choice.

`FilterUtils.findTraceThreadIdPushdownFilter` requires `field == TraceThreadField.ID && operator == Operator.EQUAL`. An ID filter with CONTAINS / STARTS_WITH / ENDS_WITH must NOT set `traces_pushdown_filter`, because the pushdown SQL is hardcoded to `thread_id = :thread_id_pushdown` and only EQUAL benefits from idx_traces_thread_id_bf. Nothing in the new gate tests exercises that operator check, so relaxing it would silently render a wrong predicate — an ID CONTAINS filter would be answered as an equality — with no failing test.

💡 Add a case with TraceThreadField.ID + Operator.CONTAINS asserting `traces_pushdown_filter` stays null and the gate returns false.

miguelgrc (commented):

5 findings — 0 high · 1 medium · 4 low. Suppressed by team conventions: see `suppressed.md`.

apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
thiagohora: Confirmed — I checked all three templates. The list template (line 239) and `SELECT_TRACE_THREADS_STATS` (line 1147) both emit `<if(traces_pushdown_filter)> AND thread_id = :thread_id_pushdown <endif>` after the uuid if/else; `SELECT_COUNT_TRACES_THREADS_BY_PROJECT_IDS`'s `trace_threads_final` has no such line, even though that template does carry the pushdown in its `traces` CTE. So with `uuid_from_time` set, the count's `trace_threads_final` gets neither the `traces_final_ids` IN nor the equality, exactly as you describe.

You're also right about the PR description — bullet 3 overstates what the shared gate buys for the count path. That will be corrected.

For prioritisation, production traffic from yesterday (thread-id pushdown detected in the query text):

| query | runs | with thread-id pushdown |
|---|---|---|
| `find_threads_by_project` | 624,541 | 1 |
| `thread_stats` | 7,739 | 0 |
| `count_threads_by_project` | 4,792 | 1 |
| `search_threads` | 1,622 | 1,565 |

So the gap is real but currently worth ~1 query/day; essentially all thread-id-filtered traffic is `search_threads`, which is the list path. Proposal: add the one guarded line to the count template so the three stay in sync, since it mirrors 239/1147 exactly — or split it into its own change if we'd rather keep this diff to the gate alone. Happy either way.

apps/opik-backend/src/test/java/com/comet/opik/domain/ThreadDAOImplTest.java
thiagohora: Correct, it's vacuous — the test never adds the `traces_final_ids` attribute, so the `<if(traces_final_ids)>` block cannot render regardless of the gate's return value, and the assertion would survive the gate being hardcoded to true. Fixing it by mirroring the production wiring: add the attribute only when the gate returns true, then assert on the rendered SQL, so a wrong gate result actually changes the output.

apps/opik-backend/src/test/java/com/comet/opik/domain/ThreadDAOImplTest.java
thiagohora: Correct. `FilterUtils.newTraceThreadFindTemplate` does `template.add("search_text", searchClause)` (FilterUtils.java:182) whenever `searchText` is present, and `searchTextActivatesPrefilter` does set it — so the clause is genuinely consumed and rendered in that case. The name and the comment are both wrong. Renaming to `SEARCH_CLAUSE` and rewording to say it is rendered only when `searchText` is set.

apps/opik-backend/src/test/java/com/comet/opik/domain/ThreadDAOImplTest.java
thiagohora: Correct. `PAGE_PUSHDOWN_DISQUALIFIERS` contains neither `search_text` nor `filters`, so in `find()` a search-only criteria is page-pushdown eligible and the `else if (shouldUseTracesFinalIdsPrefilter(...))` branch is never reached — `traces_final_ids` is not engaged for the listing query. The claim holds for the helper in isolation and for the search/stats/count paths, not for the listing path. Rewording the DisplayName to scope it to the gate.

apps/opik-backend/src/test/java/com/comet/opik/domain/ThreadDAOImplTest.java
thiagohora: Correct, and the case is constructible: `STRING_EXACT` maps CONTAINS / STARTS_WITH / ENDS_WITH (FilterQueryBuilder.java:178, 205, 217), so an ID filter with a non-EQUAL operator is a valid request that renders `trace_thread_filters` without `traces_pushdown_filter`. Nothing currently pins that boundary, and relaxing `findTraceThreadIdPushdownFilter` would silently answer an ID CONTAINS as an equality. Adding the ID + CONTAINS case asserting `traces_pushdown_filter` stays null and the gate returns false.

apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
thiagohora: Included in e04c244, and verified against production rather than shipped on reasoning.

**Correctness.** A real `count_threads_by_project` query captured from `system.query_log` (uuid range + `source` filter, so `traces_final_ids` renders), replayed with a thread-ID filter added — with and without the new line — returns **identical `count=1`** both ways.

**Magnitude.** The `trace_threads_final` CTE in isolation, on the project with the most threads in prod (5.8M `trace_threads` rows), patched run cold first:

| variant | time | rows read | bytes | peak memory | selected marks |
|---|---|---|---|---|---|
| baseline | 2917 ms | 5,881,000 | 759.81 MiB | 707.60 MiB | 719 |
| patched | **5 ms** | **81,060** | **4.74 MiB** | **83.74 KiB** | **11** |

Marks 719 → 11 confirms your primary-key-prunability point directly — `trace_threads` is `ORDER BY (workspace_id, project_id, thread_id, id)`, so the equality prunes and `id >= :uuid_from_time` cannot.

Binding was checked too: `countThreadTotal` calls `bindTraceThreadSearchCriteria` (ThreadDAO.java:1619 → FilterUtils.java:201), and the count template already referenced `:thread_id_pushdown` at lines 509/541, so the new line needs no additional binding.

PR description bullet 3 has been rewritten to describe this rather than claim the shared gate covers the count path. `FindTraceThreadsResourceTest` + `ThreadDAOImplTest` re-run green (475 tests) with the template change in place.

apps/opik-backend/src/test/java/com/comet/opik/domain/ThreadDAOImplTest.java
```diff
@@ -37,4 +46,108 @@ void firstThreadOrEmptyCompletesEmptyForEmptyStream() {
...
+         * of the gate's result rather than of the test's own setup.
+         */
+        private static Rendered render(List<TraceThreadFilter> filters, String searchText) {
+            var criteria = TraceSearchCriteria.builder()
+                    .projectId(UUID.randomUUID())
+                    .filters(filters)
+                    .searchText(searchText)
+                    .build();
+            var template = FilterUtils.newTraceThreadFindTemplate(
+                    ThreadDAOImpl.SELECT_TRACES_THREADS_BY_PROJECT_IDS, criteria, SEARCH_CLAUSE, true);
```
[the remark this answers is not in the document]
thiagohora: Fair catch — the count-template fix landed without a test, so deleting the predicate again would have left the suite green. Added in 46d272c.

`ThreadDAOImplTest$ThreadIdPushdownTemplateSync` renders the three templates that share the `traces_pushdown_filter` concept — `SELECT_TRACES_THREADS_BY_PROJECT_IDS`, `SELECT_COUNT_TRACES_THREADS_BY_PROJECT_IDS` and `SELECT_TRACE_THREADS_STATS` — through the same gate with a `uuid_from_time` range, then slices out the `trace_threads_final` CTE (up to its `ORDER BY`, so the assertions can't match a different CTE) and asserts it:

- takes the uuid branch (`AND id >= :uuid_from_time`),
- does **not** get `thread_id IN (SELECT thread_id FROM traces_final_ids)` — that branch is skipped,
- still emits `AND thread_id = :thread_id_pushdown`.

That is precisely the branch where the equality is the only predicate left that prunes on the `trace_threads` primary key.

Covering all three rather than just the count template was deliberate: the bug was drift between templates that are supposed to emit this identically, so pinning the invariant is a better guard than pinning the one instance that happened to drift.

Verified discriminating — removing the line from the count template again fails only the `count` case, with `list` and `stats` staying green:

```
ThreadIdPushdownTemplateSync.traceThreadsFinalEmitsThreadIdPushdownOnUuidBranch(String, String)[2] <<< FAILURE!
Tests run: 11, Failures: 1
```

Full run green: `FindTraceThreadsResourceTest` + `ThreadDAOImplTest` → 478 tests, 0 failures.

apps/opik-backend/src/main/java/com/comet/opik/domain/ThreadDAO.java
```diff
@@ -566,6 +567,7 @@ ORDER BY (workspace_id, project_id, id) DESC, last_updated_at DESC
                         AND thread_id IN (SELECT thread_id FROM traces_final_ids)
                     <endif>
                 <endif>
+                <if(traces_pushdown_filter)> AND thread_id = :thread_id_pushdown <endif>
                 ORDER BY (workspace_id, project_id, thread_id, id) DESC, last_updated_at DESC
```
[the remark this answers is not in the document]
thiagohora: Covered in 46d272c — see the [reply on the sibling comment](https://github.com/comet-ml/opik/pull/7852#discussion_r3783804103) for details. The new `ThreadIdPushdownTemplateSync` test renders `SELECT_COUNT_TRACES_THREADS_BY_PROJECT_IDS` (alongside the list and stats templates) through the gate on the `uuid_from_time` branch and asserts `trace_threads_final` emits `AND thread_id = :thread_id_pushdown`. Removing the predicate fails only the count case.

miguelgrc:
**Non-blocking follow-up.** The count-template fix is correct and the numbers are convincing (719 → 11 marks), but it currently ships without a regression guard: deleting the new `:570` line leaves all 8 `ThreadDAOImplTest` tests green. That's the same exposure as the original defect — a template missing a predicate, with nothing to catch it.

The invariant worth pinning isn't the line itself, it's that all three templates emit the pushdown identically. `SELECT_COUNT_TRACES_THREADS_BY_PROJECT_IDS` and `SELECT_TRACE_THREADS_STATS` are still `private`; marking both `@VisibleForTesting` and adding one parameterized case over the three constants — ID EQUAL filter, assert `thread_id = :thread_id_pushdown` renders in `trace_threads_final` — would make the next refactor of any of them fail loudly.

Verified by mutation against `e04c2448`: reverting the gate fails `threadIdEqualFilterActivatesPrefilter`; forcing the gate to `true` fails all three negative cases; removing the count line fails nothing.

miguelgrc (approved):

apps/opik-backend/src/test/java/com/comet/opik/domain/ThreadDAOImplTest.java
```diff
@@ -37,4 +50,167 @@ void firstThreadOrEmptyCompletesEmptyForEmptyStream() {
...
+            if (ThreadDAOImpl.shouldUseTracesFinalIdsPrefilter(criteria, template)) {
+                template.add("traces_final_ids", true);
+            }
+
+            var traceThreadsFinal = traceThreadsFinalCte(template.render());
+
+            assertThat(traceThreadsFinal).contains("AND id >= :uuid_from_time");
+            assertThat(traceThreadsFinal).doesNotContain(TRACES_FINAL_IDS_IN);
+            assertThat(traceThreadsFinal).contains(THREAD_ID_PUSHDOWN);
+        }
```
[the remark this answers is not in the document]
thiagohora: The predicate is present at `ThreadDAO.java:571`, in the count template's `trace_threads_final` — after the uuid if/else and before the `ORDER BY`, matching line 239 (list) and line 1150 (stats). It was added in e04c244.

The parameterized case passes. CI on `46d272c` reports Backend Tests - Unit Tests **"All 3 276 tests pass, 2 skipped"**, and `ThreadDAOImplTest` is a unit test, so it is inside that count. Locally it is 11/11, and `FindTraceThreadsResourceTest` + `ThreadDAOImplTest` is 478/478. All 16 integration groups and spotless are green on that commit too.

Neither suggested action would be right here:

- Adding the predicate to the count SQL would emit it twice, since it is already there.
- Removing the count case would delete the guard added in response to [the earlier review comment](https://github.com/comet-ml/opik/pull/7852#discussion_r3783804103) asking for exactly that coverage — it is the case that fails when the predicate is missing, which is the whole point of it.

Could you re-run against `46d272c`? The analysis looks like it is pairing the new test file with an older revision of `ThreadDAO.java` — the cited failure at line 204/205 does not reproduce in any run, local or CI. For what it is worth, removing the line again does fail that case and only that case (`[2] count`; `list` and `stats` stay green), which is the behaviour the test is designed for.

PR #7853, opened 2026-08-14
[OPIK-7923] [BE] fix: two NullPointerExceptions seen in production

thiagohora:
## Details

Two independent NullPointerExceptions found in production (`opik-backend` 2.2.28, `cometml-production`) while investigating NPEs in the logs. The first turns real user requests into 500s; the second silently drops online-scoring work.

- **OPIK-7923** — three endpoints take filters from the request body and pass them to the service without `FiltersFactory` validation, so an operator the field's type does not support reaches the query builder with no template behind it and fails as a 500 instead of a 400.
- **OPIK-7924** — `toFeedbackScores` reads `chatResponse.aiMessage().text()` straight into `extractJson`; a judge answer with no text NPEs and takes the whole Redis stream message down, so the trace is never scored and the user is told nothing.

### OPIK-7923 — body-supplied filters skipped validation

`FiltersFactory.toValidAndDecoded` already rejects an unsupported operator/field-type pair with a 400 naming the pair. Every `@QueryParam("filters")` path runs it (28 sites, all checked). These three took filters from the **request body** and skipped it:

| Endpoint | Request |
|---|---|
| `POST /v1/private/datasets/items/delete` | `DatasetItemsDelete.filters` |
| `PATCH /v1/private/datasets/items/batch` | `DatasetItemBatchUpdate.filters` |
| `POST /v1/private/workspaces/spans/metric` | `WorkspaceSpanMetricRequest.filters` |

`ANALYTICS_DB_OPERATOR_MAP.get(operator).get(fieldType)` returns `null` for a combination it has no template for. `toAnalyticsDbFilters` only gates on field membership, never on a template existing, so the `null` went into `FieldType.buildFilter`:

```
ERROR io.dropwizard.jersey.errors.LoggingExceptionMapper: Error handling a request
java.lang.NullPointerException: Cannot invoke "String.formatted(Object[])" because "template" is null
    at com.comet.opik.api.filter.FieldType.buildFilter(FieldType.java:61)
    at com.comet.opik.domain.filter.FilterQueryBuilder.toAnalyticsDbFilter(FilterQueryBuilder.java:1040)
    at com.comet.opik.domain.DatasetItemVersionDAOImpl.lambda$countItemsMatchingFilters$0(DatasetItemVersionDAO.java:3740)
```

Reachable pairs are not exotic — for a dataset item field, `IN`/`NOT_IN` on anything, `IS_EMPTY` on a string, and `STARTS_WITH` on `tags` all land here. In prod this is `POST /v1/private/datasets/items/delete` returning 500 to real user traffic (6 occurrences in a 6h window, in two bursts consistent with a user retrying a filtered delete).

All three now call `filtersFactory.validateFilter(...)` first, matching the query-param paths and the body paths in `TracesResource`, `SpansResource` and `ProjectsResource` that already did this.

**Scope of the audit**: every caller of `toAnalyticsDbFilters`/`toAnalyticsDbFiltersV2Client` was traced back to where its filters enter the system — all 28 query-param sites and every request record carrying a typed filter list. These three were the only gaps. Automation-rule filters (`AutomationRuleEvaluator*`) are evaluated in memory by `FilterEvaluationService` and never reach the query builder, so they are unaffected.

`FilterQueryBuilder` itself is deliberately unchanged: the contract that callers validate first now holds everywhere.

Review follow-ups addressed in 326d287: the finish reason is quoted per the backend logging convention; `WorkspaceSpanMetricRequest.filters` now declares its elements `@NotNull @Valid` so a null element is a 422 rather than a 500 (as `DatasetItemsDelete` and `DatasetItemBatchUpdate` already did); and the unreachable null-`aiMessage` guard was dropped, since `ChatResponse.Builder` rejects a null `aiMessage` outright. Review round 2 (3c720e9) closed the other half of that contract: `validateFilter` dereferences `filter.field()`, so a null list element answered 500 rather than a 4xx. `WorkspaceSpanMetricRequest` was fixed in 326d287; the same gap was open in `TraceSearchStreamRequest`, `TraceThreadSearchStreamRequest`, `SpanSearchStreamRequest` and `ProjectMetricRequest`, all of which feed `validateFilter`. All six fields now declare their elements `@NotNull @Valid`, matching `DatasetItemsDelete`. One review finding is deliberately not fixed here and is tracked as OPIK_7925: an unusable judge answer leaves a test-suite run with no assertion row and a null `pass_rate`, which is pre-existing behaviour for all `ResponseProblem` kinds, not something this PR introduces.

### OPIK-7924 — a judge answer with no text killed the scoring message

A provider can answer with no text at all — a content filter, or a reasoning model that spends its budget before emitting any:

```
ERROR OnlineScoringLlmAsJudgeScorer: Unexpected error processing message from Redis stream
      'ChatResponse { aiMessage = AiMessage { text = null, thinking = null, ... }'
java.lang.NullPointerException: Cannot invoke "java.lang.CharSequence.length()" because "this.text" is null
    at com.comet.opik.api.resources.v1.events.OnlineScoringEngine.extractJson(OnlineScoringEngine.java:1109)
```

In prod this sits right after exhausted OpenAI TPM rate-limit retries on `gpt-5-nano`.

An empty answer now goes through the existing `ResponseProblem` channel added in OPIK_7354 (new `NO_ANSWER` kind), so the rule's logs tell the user nothing was scored and which finish reason came back — the same treatment `NOT_JSON` and `NO_SCORE_FIELDS` already get. Blank text folds into the same case; it previously reported as "not valid JSON", which was misleading.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves OPIK-7923
- Resolves OPIK-7924

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: Log investigation that found both NPEs, the audit of filter entry points, the code changes, and the tests.
- Human verification: Author reviewed the diff and the audit conclusions. Each of the three resource tests was additionally run against the unfixed code to confirm it returns 500 there, so the tests reproduce the production failure rather than merely asserting the new behaviour.

## Testing

Run from `apps/opik-backend`:

```
mvn -o test -Dtest='OnlineScoringEngineParsingTest'
mvn -o test -Dtest='DatasetsResourceTest$DeleteDatasetItems#deleteDatasetItems__whenFilterOperatorUnsupportedForFieldType__thenBadRequest+deleteDatasetItems__whenUsingFilters__thenSucceed'
mvn -o test -Dtest='DatasetsResourceTest$BatchUpdateDatasetItems#batchUpdateDatasetItems__whenFilterOperatorUnsupportedForFieldType__thenBadRequest+batchUpdateDatasetItems__whenAddingTagsWithMerge__thenSucceed'
mvn -o test -Dtest='WorkspacesResourceTest$SpanMetricsTest'
```

Results (local, macOS, Testcontainers via Docker):

```
OnlineScoringEngineParsingTest                      72 passed  (+3)
DatasetsResourceTest$DeleteDatasetItems              2 passed  (+1, plus the existing success path)
DatasetsResourceTest$BatchUpdateDatasetItems         2 passed  (+1, plus the existing success path)
WorkspacesResourceTest$SpanMetricsTest              16 passed  (+2)
FindSpansResourceTest$FindSpans                      1 passed  (+1, the null-element case)
```

Scenarios covered:

- Unsupported operator for the field type on each of the three endpoints → 400 naming the pair. Each was first run with the fix reverted and returns **500** there, so the test reproduces the production failure.
- The existing delete-by-filters and batch-update success paths still pass, so validation does not reject filters that used to work.
- Judge answer with `text = null` (via `AiMessage.builder().build()`, matching the production `ChatResponse` shape) → `NO_ANSWER` problem, empty scores, no NPE, and the finish reason reaches the user-facing warning.
- Blank and whitespace-only judge answers → same `NO_ANSWER` treatment.
- `filters: [null]` on the workspace span metric endpoint → 422 from bean validation. Also verified in both directions: without the `@NotNull` element annotation it returns **500**.
- `filters: [null]` on `POST /v1/private/spans/search` → 422. Same both-directions check: **500** without the annotation. One test rather than one per record, since the six declarations are now identical.

Not run: the full backend suite — left to CI.

## Documentation

No documentation change needed. Both fixes correct error handling on existing endpoints; no API shape, parameter, or documented behaviour changes. The dataset and workspace endpoints already document a 400 for invalid parameters, which is now what they actually return.


CometActions:
**No test needed here.**

Both fixes turn a crash on a malformed request into a proper 4xx, and neither changes anything on the valid path: `validateFilter` returns a good filter list unchanged, and the new `@Valid` finds no jakarta constraints on `FilterImpl` (it guards with Lombok `@NonNull`), so no page renders differently. The payloads you now reject aren't reachable from the UI either — `constants/filters.ts` offers only `contains` for a list column, so nobody can pick `starts_with` on `tags`, and the frontend never puts a literal `null` in a filters array. That leaves hand-written API bodies, which is exactly what `FindSpansResourceTest`, `WorkspacesResourceTest` and `DatasetsResourceTest` now assert here, through the real validation stack. The `NO_ANSWER` half needs a provider that answers with no text (content filter, or a reasoning model out of budget) — not reproducible on a fresh OSS install, and `OnlineScoringEngineParsingTest` already pins it. Nothing for e2e to add.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/32270247829)

_Re-checked after a push on 19 Aug 15:33 UTC._

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringEngine.java
```diff
@@ -1166,7 +1169,18 @@ private static String sizeOf(String content) {
...
+        // A provider can answer with no text at all — e.g. a content filter, or a reasoning model that spends its
+        // budget before emitting any. Report it as an unusable answer instead of NPE-ing on the null text, which
+        // killed the whole scoring message and left the user no trace of why nothing was scored.
+        var answer = Optional.ofNullable(chatResponse.aiMessage())
+                .map(AiMessage::text)
+                .filter(StringUtils::isNotBlank);
+        if (answer.isEmpty()) {
+            log.warn("Judge returned no answer to parse: finishReason='{}'", chatResponse.finishReason());
+            return ParsedFeedbackScores.problem(ResponseProblem.Kind.NO_ANSWER,
+                    Objects.toString(chatResponse.finishReason(), "unknown"), List.of(), 0);
```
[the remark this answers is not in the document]
thiagohora: Good catch on the downstream contract, but I'm not fixing it here — it isn't specific to `NO_ANSWER`.

Every other `ResponseProblem` kind already behaves exactly this way on `main`: `NOT_JSON`, `NOT_A_JSON_OBJECT` and `NO_SCORE_FIELDS` all return `ParsedFeedbackScores.problem(...)` with an empty score list, so no assertion row is written, the message is still ack'd, and `decrementAndFinishIfComplete` still runs. The null-aggregation / indefinite-refetch path you describe is reachable today with any unparseable judge answer.

What this PR changes is only that a null-text answer joins its siblings instead of throwing. And the behaviour it replaces was not a deliberate retry policy — it was the NPE at `extractJson`, which left the message un-ack'd so Redis Streams re-ran the whole scoring job. For a deterministic failure like a content filter that retries forever and never scores, which is the production symptom this PR was opened for.

So "keep `NO_ANSWER` retryable" would mean re-introducing that loop for one kind while the other three stay terminal. The real question is which contract all four should have, plus handling a null `pass_rate` on a `COMPLETED` run in the frontend.

Filed as **OPIK-7925** with your analysis, scoped across all four kinds.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/OnlineScoringEngineParsingTest.java
```diff
@@ -298,6 +299,34 @@ void whenResponseHasManyFields_thenCapsTheNamesAndCountsTheRest() {
...
 
+    @Test
+    @DisplayName("report a judge that answered with no text instead of failing the scoring message")
+    void whenTheAnswerHasNoText_thenReportsItAsNoAnswer() {
+        // A provider can return a message carrying no text — a content filter, or a reasoning model that emits
+        // none. Parsing it used to throw NPE, which killed the Redis message: the trace was silently never scored.
+        var noText = ChatResponse.builder()
+                .aiMessage(AiMessage.builder().build())
+                .finishReason(FinishReason.CONTENT_FILTER)
+                .build();
```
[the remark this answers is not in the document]
thiagohora: That test can't be written — `ChatResponse` cannot hold a null `aiMessage`. `ChatResponse.Builder` validates it:

```
jshell> ChatResponse.builder().build()
THREW: IllegalArgumentException: aiMessage cannot be null
```

So the branch behind `Optional.ofNullable(chatResponse.aiMessage())` was unreachable, and no test could construct an input that exercises it.

Rather than add coverage for dead code I removed the guard, in 326d287 — only `aiMessage().text()` is nullable, which is what production actually returned (`AiMessage { text = null, thinking = null, toolExecutionRequests = ... }`). That's still covered by `whenTheAnswerHasNoText_thenReportsItAsNoAnswer` and the blank-answer cases.

apps/opik-backend/src/main/java/com/comet/opik/api/metrics/WorkspaceSpanMetricRequest.java
```diff
@@ -34,7 +34,7 @@ public record WorkspaceSpanMetricRequest(
         MetricType metricType,
         TimeInterval interval,
         @Valid BreakdownConfig breakdown,
-        List<SpanFilter> filters,
+        List<@NotNull @Valid SpanFilter> filters,
```
JetoPistola, line 37: 💡 **suggestion** | Reliability

The same null-element gap fixed here still exists in four other body-filter records that all feed `filtersFactory.validateFilter(...)`, which dereferences `filter.field()` in `toValidAndDecoded`:

| Record | Field | Validated at |
|---|---|---|
| `TraceSearchStreamRequest` | `List<TraceFilter> filters` (:24) | `TracesResource:225` |
| `TraceThreadSearchStreamRequest` | `List<TraceThreadFilter> filters` (:23) | `TracesResource:755` |
| `SpanSearchStreamRequest` | `List<SpanFilter> filters` (:27) | `SpansResource:494` |
| `ProjectMetricRequest` | `spanFilters` / `traceFilters` / `threadFilters` (:30-32) | `ProjectsResource:259-261` |

`filters: [null]` on any of these produces the same 500 that `/metrics/spans` returned before 326d287 — bean validation passes the null element through, then `validateFilter` NPEs and `OpikGenericExceptionMapper` renders it as a 500 rather than the 422 this record now returns.

Not a regression this PR introduces, and I understand OPIK-7923's scope was the *missing-validation* gap rather than the *null-element* one. But the PR frames the outcome as "the contract that callers validate first now holds everywhere" — and `validateFilter`'s precondition (no null elements) is the other half of that contract. Four one-line annotations (`List<@NotNull XFilter>`) would close it, or a follow-up ticket if you'd rather keep this PR tight.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/DatasetsResource.java
```diff
@@ -363,10 +363,15 @@ public Response batchUpdate(
 
         String workspaceId = requestContext.get().getWorkspaceId();
 
+        // Same validation gap as deleteDatasetItems: these filters reach the same query builder.
+        var validatedBatchUpdate = batchUpdate.toBuilder()
+                .filters(filtersFactory.validateFilter(batchUpdate.filters()))
+                .build();
+
         log.info("Batch updating dataset items. workspaceId='{}', idsSize='{}', filters='{}'", workspaceId,
                 emptyIfNull(batchUpdate.ids()).size(), emptyIfNull(batchUpdate.filters()).size());
```
JetoPistola: 🧹 **nit** | Consistency

Both `log.info` calls here still read `batchUpdate.filters()` while the service now receives `validatedBatchUpdate`. The count is identical either way, so this is cosmetic — but `deleteDatasetItems` right below keeps the original `request` for logging *because* it only extracts the list, whereas here the whole request object was rebuilt. Reading `validatedBatchUpdate.filters()` in both log lines would make the two endpoints tell the same story about which object is authoritative after validation.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringEngine.java
```diff
@@ -1166,7 +1168,18 @@ private static String sizeOf(String content) {
...
         var declaredSchemas = Objects.requireNonNullElse(schema, List.<LlmAsJudgeOutputSchema>of());
-        var content = extractJson(chatResponse.aiMessage().text());
+        // A provider can answer with no text at all — e.g. a content filter, or a reasoning model that spends its
+        // budget before emitting any. Report it as an unusable answer instead of NPE-ing on the null text, which
+        // killed the whole scoring message and left the user no trace of why nothing was scored.
+        // aiMessage itself is never null — ChatResponse.Builder rejects that — but its text is.
+        var answer = Optional.ofNullable(chatResponse.aiMessage().text())
+                .filter(StringUtils::isNotBlank);
+        if (answer.isEmpty()) {
+            log.warn("Judge returned no answer to parse: finishReason='{}'", chatResponse.finishReason());
```
JetoPistola, line 1178: ❓ **question** | Validation

On the related `WorkspaceSpanMetricRequest` change: `@Valid` in `List<@NotNull @Valid SpanFilter>` looks like a no-op today — `FilterImpl` (and `SpanFilter`) carry no bean-validation constraints, so cascading into the element finds nothing to check. Only the `@NotNull` does work.

Harmless either way, and it matches how `DatasetItemsDelete` / `DatasetItemBatchUpdate` already declare theirs — so if it's deliberate symmetry plus future-proofing for when `FilterImpl` gains constraints, that's a good reason to keep it. Just flagging in case it was added on the assumption it was required for the null-element rejection to fire (it isn't — the test would pass with `@NotNull` alone).

The `NO_ANSWER` change itself reads well: routing through the existing `ResponseProblem` channel means the user gets the same treatment as `NOT_JSON`, and folding blank text into the same case fixes the misleading "not valid JSON" message it used to produce.

JetoPistola:
👋 **Review summary**

**What looks good**

- **The audit is the best part of this PR.** Tracing every caller of `toAnalyticsDbFilters` back to where filters enter the system — all 28 query-param sites plus every request record carrying a typed filter list — is what turns "I fixed the NPE in the stack trace" into "I know these were the only three gaps." I spot-checked it and it holds: the query-param paths do all run `newFilters`, the `TracesResource`/`SpansResource`/`ProjectsResource` body paths do already call `validateFilter`, and the `AutomationRuleEvaluator*` filters genuinely never reach the query builder.
- **Tests that reproduce rather than describe.** Running each of the three resource tests against the unfixed code to confirm it returns 500 there is the difference between a regression test and a test that would pass on a broken build. Same for verifying the `@NotNull` element case in both directions.
- **Deciding not to fix something, and saying why.** The response on the `pass_rate` finding — that every existing `ResponseProblem` kind behaves identically on `main`, so it isn't `NO_ANSWER`-specific — is the right call, and tracking it as OPIK_7925 instead of quietly widening this PR keeps the change reviewable.
- **`NO_ANSWER` routed through the existing channel** rather than given bespoke handling, so the user gets the same treatment `NOT_JSON` already gets. Folding blank text into the same case also fixes a genuinely misleading message.

**Overall**

Two unrelated NPEs in one PR is usually a smell, but here it reads as one investigation ("what's NPE-ing in prod?") with two answers, and the write-up keeps them cleanly separated. The `FilterQueryBuilder`-stays-unchanged decision is right — pushing the null check down into the builder would have hidden the fact that three callers were skipping validation.

**Inline comments**: 1 suggestion, 1 nit, 1 question — nothing blocking. The suggestion is the only one worth acting on: four other body-filter records still have the null-element gap that 326d287 closed here.

apps/opik-backend/src/main/java/com/comet/opik/api/metrics/WorkspaceSpanMetricRequest.java
thiagohora: You're right, and the framing was the part that needed fixing — I'd written that the validate-first contract "now holds everywhere" while leaving `validateFilter`'s own precondition open in four records. Closed it in 3c720e9 rather than deferring.

Verified your table before acting: all six fields are unannotated `List<XFilter>`, and all six reach `validateFilter` at the call sites you cited. Then confirmed the failure is real rather than theoretical — `POST /v1/private/spans/search` with `filters: [null]`:

- without the annotation → **500**
- with it → **422**

Added a regression test on that endpoint (`whenSearchFilterElementIsNull__thenReturn422`) checked in both directions, rather than one per record — the six declarations are now identical, so the mechanism is the same once proven.

Declared them `@NotNull @Valid` rather than the `@NotNull` you suggested, for the reason in your other comment — see the reply there.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/priv/DatasetsResource.java
thiagohora: Fixed in 3c720e9 — both log lines now read `validatedBatchUpdate`.

Agreed on the reasoning: `deleteDatasetItems` only lifts the list out, so `request` stays authoritative there, but here the whole object was rebuilt and the logs were still describing the pre-validation one. Same count either way, but the two endpoints now tell a consistent story.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringEngine.java
thiagohora: Your analysis is right, and I should be straight about the assumption: I copied the declaration from `DatasetItemsDelete` for symmetry, without checking whether `@Valid` does anything here. It doesn't — I confirmed `FilterImpl` and `SpanFilter` carry no bean-validation constraints, so cascading finds nothing and only `@NotNull` rejects the null element. Your read that the test passes with `@NotNull` alone is correct.

Keeping `@Valid` deliberately, now that it's a decision rather than an assumption: all seven filter-list declarations across the codebase are identical after 3c720e9, so `@Valid` is either present everywhere or nowhere, and it starts doing real work the moment `FilterImpl` gains a constraint. Happy to strip it from all seven if you'd rather not carry a no-op — it's a one-line change either way, and I don't feel strongly.

Thanks for the note on `NO_ANSWER`.

apps/opik-backend/src/main/java/com/comet/opik/api/metrics/ProjectMetricRequest.java
```diff
@@ -27,9 +28,9 @@ public record ProjectMetricRequest(
         @NonNull TimeInterval interval,
         @NonNull Instant intervalStart,
         Instant intervalEnd,
-        List<SpanFilter> spanFilters,
-        List<TraceFilter> traceFilters,
-        List<TraceThreadFilter> threadFilters,
+        List<@NotNull @Valid SpanFilter> spanFilters,
+        List<@NotNull @Valid TraceFilter> traceFilters,
+        List<@NotNull @Valid TraceThreadFilter> threadFilters,
```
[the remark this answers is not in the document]
thiagohora: The observation is accurate — those five operations can now return a 422 that their metadata doesn't declare. But it's classified as a breaking change introduced here, and I don't think that holds: the gap is pre-existing and codebase-wide, and this PR only adds a new *trigger* for an already-undocumented status.

Checked before replying:

- `deleteDatasetItems` declares only 204 and 400, yet its own existing test `deleteDatasetItems__whenTooManyIds__thenBadRequest` asserts **422** for the `@Size(min = 1, max = 1000)` violation on `itemIds`. That ships on `main` today.
- `searchSpans` is one of the five you list, and `SpanSearchStreamRequest.limit` already carries `@Min(1) @Max(2000)` — so `limit: 0` has always returned an undocumented 422 there, before any change of mine.
- Across the resource layer: 20 declarations of `responseCode = "422"` against 79 of `"400"`, while 28 test files assert a 422.

So documenting 422 on these five would leave the same gap on dozens of other operations, and the real fix is a convention decision — document it everywhere a validated body is accepted, or map bean-validation failures to 400 globally — then regenerate the Fern contracts and clients once.

Filed as **OPIK-7943** with that evidence and both options. Happy to be overruled if the team would rather land the five-operation version here.

Worth noting the alternative is worse: without the container constraints these endpoints answer **500** for `filters: [null]`, which is undocumented *and* wrong.

JetoPistola:
👋 **Re-review** — head `0f0d37d3`

Re-reviewed after `3c720e9` + the two merges from `main`. **No new findings.** All three of my round-2 comments are closed:

| Finding | Status |
|---|---|
| Four other body-filter records lacked `@NotNull` elements | ✅ `3c720e9` — all six now `List<@NotNull @Valid XFilter>` |
| `log.info` read `batchUpdate.filters()` after the rebuild | ✅ `3c720e9` |
| `@Valid` is a no-op on filter elements | ✅ Acknowledged; kept for symmetry with `DatasetItemsDelete` |

**What I checked this round**

- **All 8 `validateFilter` call sites** have `@Valid` on the request-body param, so the new container constraints actually fire — `TracesResource:225,755`, `SpansResource:494`, `ProjectsResource:259-261`, `DatasetsResource`, `WorkspacesResource:171`. No dead annotations.
- **The record set is complete.** Every `List<XFilter>` field under `api/` is accounted for; the remaining unannotated ones are `AutomationRuleEvaluator*` (in-memory via `FilterEvaluationService`, never reach the query builder) and the internal `*SearchCriteria` records. The audit's conclusion holds.
- **Empty-list semantics preserved** — `validateFilter` early-returns on `CollectionUtils.isEmpty`, so the delete endpoint's documented "empty `filters` array = delete all items in the dataset" contract is unchanged. This was the main regression risk in a validation-tightening PR and it's clean.
- **Null filter *values* were already closed** — `FilterImpl.value` is Lombok `@NonNull` and `@JsonProperty(required = true)`, so `mapCustom`'s `URLDecoder.decode(filter.value(), …)` can't NPE on the `SpanField.CUSTOM` path that `WorkspaceSpanMetricRequest` now reaches for the first time.
- **`NO_ANSWER` enum insertion is safe** — `ResponseProblem` has no references outside `OnlineScoringEngine` and is never serialized or persisted, so inserting rather than appending changes nothing.
- **`finishReason` null** handled via `Objects.toString(…, "unknown")`.

**One residual gap, pre-existing and out of scope**

`validateFilter` gates on the public `toAnalyticsDbOperator(filter)` — the plain map lookup. The DAO uses the private strategy-aware overload, which for `FEEDBACK_SCORES_AGGREGATED` / `EXPERIMENT_SCORES_AGGREGATED` + `FEEDBACK_SCORES_NUMBER` routes to `getAggregatedMapScoresTemplate`. That method throws `IllegalArgumentException` on an unmapped operator rather than returning null, so it can't produce the NPE this PR fixes — but a validated-OK filter could still fail at query-build time there. Present on `main`, orthogonal to both tickets, not worth widening scope for.

**Test environment:** agreeing with the QA radar that none is needed. The rejected payloads aren't reachable from the UI (`constants/filters.ts` offers only `contains` for a list column; the frontend never emits a literal `null` in a filters array), so a deployed env could only be driven by hand-written bodies — which the three resource tests already do through the real validation stack. The `NO_ANSWER` half isn't reproducible on a fresh install at all.

**Inline comments**: none this round.

JetoPistola (approved):
Re-reviewed at 0f0d37d3. Both NPEs are real production failures with stack traces, and each fix has a test confirmed to fail (500) against the unfixed code. The validate-first contract now actually holds across all filter entry points. CI green. No test environment needed — the rejected payloads aren't reachable from the UI, so the integration tests cover this better than a deployed env would.

PR #7868, opened 2026-08-17
[OPIK-7938] [FE] fix: add experiment ID copy button- #7868

jolanta-io:
## Details

Adds an ID button to the Experiment details page header, sitting just before the Dataset/Test suite tag, which displays the experiment ID truncated to its first and last three characters and copies the full ID to the clipboard on click. Previously there was no way to retrieve an experiment's ID from the UI, so anyone needing it for an API call, a support request, or sharing had to dig it out of the URL. The button confirms the copy with a green check for three seconds before reverting.

Rendered only on the single-experiment details view; hidden in compare mode, and unchanged for both dataset- and test-suite-backed experiments.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- OPIK-7692

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: full implementation
- Human verification: code review + manual UI verification

## Testing

- `npm run lint`, `npm run typecheck` — all clean.
- Manually tested happy path on dataset and test suite experiment

## Documentation

N/A — no user-facing documentation changes.


CometActions:

The copy button is genuinely unasserted today — experiments-smoke.spec.ts opens the detail page but only reads the items table, aggregate chip and Logs tab, and no spec in the estate touches the clipboard at all — so a wrong or truncated pasted ID would ship silently (a crash would still be caught, since the smoke test renders this header). But our taxonomy has no capability for the experiment detail header: the 11 experiments keys are list/per-item/compare/tabs, and the compare-* ones are the wrong home because the button is hidden when isCompare. A generated spec would need a @cap: tag that doesn't exist, which is a hard lint failure, so triage stops here rather than spending an environment. Nothing for you to do — this is a gap in our map, not in your PR.

_areas: `experiments`_
[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/32028697647)

_Re-checked after a push on 17 Aug 12:13 UTC — nothing the verdict depends on changed._

awkoy (approved):

PR #7869, opened 2026-08-17
[OPIK-7793] [QA] Proposed e2e spec from the #7801 exploration: project date-range default

CometActions:
## Where this came from

Exploratory testing of **#7801** — `[OPIK-7793] fix: compress seeded demo data under the UUIDv7 ingestion window`. A human worked through that PR's task list on a live build of its own head commit (`39bd2db`), verified the flows by hand, and this PR turns the one strong automation candidate from that report into a permanent spec.

**Targeted at `thiagoh/OPIK-7793-demo-data-uuidv7-window`, not `main`, on purpose.** The behaviour asserted here — both pages resolving the project *before* mounting anything that reads the date range — is introduced by #7801 and does not exist on `main` yet, so a spec based on `main` could not be verified. Retarget it to `main` once #7801 lands.

Base ref: `thiagoh/OPIK-7793-demo-data-uuidv7-window` @ `39bd2db951ae07907830cd080c3d4bbbbed2b2bf`.

## What it covers

`@area:dashboards` / `@cap:dashboards.metric-date-range` — a capability the taxonomy defines and nothing in the estate touched before this. New file: `tests_end_to_end/e2e/tests/dashboards/project-date-range-default.spec.ts` (`@t2-cuj`).

**1. `An ordinary project opens Logs and Dashboards on the 30-day default and persists it to the shared, unscoped slot`**

Seeds three traces via the SDK, then asserts on both surfaces that the range control reads *Past 30 days*, that the seeded traces actually fall inside that window and are listed (the range windows the table — it is not decoration), that Logs persisted `past30days` into `local-time_range` and Dashboards into `opik-project-insights-daterange`, and that **neither wrote a project-scoped slot**. That last one is the point: the scoped slot exists so the seeded demo project can hold a 24h default, and an ordinary project appearing in one would mean the override reached the wrong project.

**2. `A range chosen on Logs or Dashboards outranks the default and survives a fresh mount`**

Picks *Past 7 days*, then navigates afresh (deliberately **not** a reload — selecting a range also writes `?time_range=`, and a reload would satisfy the trigger from the query param even if the stored value had been clobbered) and asserts the stored choice comes back. Repeats it on Dashboards, which also confirms the two surfaces keep separate slots: the Logs choice does not move Dashboards off the default.

This is the assertion with real teeth for #7801. `use-local-storage-state` captures its `defaultValue` once and writes it to storage, so a consumer that mounts before the project name resolves freezes the wrong default into the slot — silently, since nothing on screen says the default was decided too early. Everything is read at the real storage boundary; nothing is mocked, no assertion depends on the wall clock, and nothing depends on data that happens to already exist in the workspace.

Also adds two page objects:
- `pom/metric-date-range.control.ts` — the `MetricDateRangeSelect` control is mounted by both Logs and Dashboards, so it lives on its own rather than being re-declared per page. It also holds the storage keys and the JSON-decoding reader.
- `pom/project-dashboards.page.ts` — the project Dashboards tab had no page object.
- `pom/logs.page.ts` gains a `dateRange` accessor; nothing existing changed.

## Verification

Run against a local OSS build of #7801's head (`39bd2db`) at `http://localhost:5173`, workspace `default`:

```
npx playwright test tests/dashboards/project-date-range-default.spec.ts
  2 passed (13.3s)

npx playwright test tests/dashboards/project-date-range-default.spec.ts --repeat-each=2
  4 passed (22.6s)
```

Both tests **passed**, and passed again on repeat — no flakiness observed. Also clean:

```
npx tsc --noEmit                                        # 0 errors
python3 tests_end_to_end/coverage/tag_lint.py ...       # 35 specs checked, 0 problems
```

`taxonomy.yaml` is deliberately untouched — `reconcile.py` owns `covered:`/`tier:` and will flip `dashboards.metric-date-range` to `covered: true, tier: t2-cuj` on its next nightly run.

## What I deliberately did not write

**One candidate dropped: the demo project's own 24h default** (`Opik Demo Agent Observability` opening Logs and Dashboards on *Past 24 hours* in its own storage slot). This is #7801's headline user-visible behaviour and the exploration confirmed it works — it is dropped as an *automation* candidate, not as a behaviour, and the exploration marked it weak for two reasons a reviewer should weigh:

1. **It cannot be reached from a clean state.** `resolveProjectDateRangeConfig` keys off the exact literal name `Opik Demo Agent Observability`, which cannot carry the `cuj-*` run prefix. A spec would have to depend on the installer-seeded demo project rather than creating its own fixture, and must not delete it — against the estate's rule that a spec not depend on data that happens to exist in the target workspace. On an environment without the demo project it would conditionally skip, which is invisible coverage: the `@cap:` tag would claim the capability while the test never ran.
2. **A data assertion would go stale.** The demo data is compressed into the 10h *before* seeding and is never re-seeded, so any "traces are visible under the 24h default" assertion goes red roughly 24h after the environment is provisioned. Only a per-run fresh install would keep it green.

If the team wants it anyway, the exploration named the shape that survives: assert **only** the range label and the storage-slot key (`local-time_range-Opik Demo Agent Observability`), never a row count — that part is name-driven and does not age. Happy to add it on request.

**Also not written:** the demo-timeline compression invariants (10h window, id order == start_time order, durations preserved). #7801 already pins that math in `apps/opik-python-backend/tests/unit/test_demo_timeline.py` and `test_demo_data_uuid_window.py`; a Playwright spec would add environment dependence without covering a seam those unit tests miss.

## Note for the reviewer

The exploration also raised a **product question, not a bug**, which this spec does not encode: the demo project's default view goes empty roughly 24 hours after seeding (data spans `[seed − 10h, seed]`, seeding happens once, and the new default window is 24h). Before #7801 the demo spanned 30 days under a 30-day default and stayed populated for weeks. The compression is clearly required and the 24h default is what renders it as a curve rather than one bar — but the two together shorten the window in which a returning user sees a living demo. Worth confirming that is intended; see the exploration report.

---

🤖 Generated by the release QA side flow ([Claude Code](https://claude.com/claude-code)) from the exploratory testing of #7801. **Nothing here has been human-reviewed.** Draft on purpose — please read the assertions before promoting it.

[OPIK-7793]: https://comet-ml.atlassian.net/browse/OPIK-7793?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


AndreiCautisanu:
Baz review

PR #7881, opened 2026-08-18
[OPIK-6901] feat(cutover): expose --max-insert-threads, the throughput ceiling

thiagohora:
## Details

**The copy is sink-bound by default and nothing says so.** ClickHouse documents `max_insert_threads = 0` — the default — as *"`INSERT SELECT` no parallel execution"*, so the insert side of this copy runs serialised until it is raised.

Exposing the setting **lets an operator control how much of the machine the backfill may use, and so speed the copy up substantially** when the insert side is what is holding it back.

- `backfill.sh` and `delta_replay.sh` gain `--max-insert-threads N` (omitted = inherit the server's setting; explicit `0` forces serial; validated empty, `0`, or `1..99`).
- `000001` and `000002` gain the setting, so the delta benefits too — it writes into the same table through the same insert path.
- README documents the *diagnosis*, not just the knob.

### Scope caveats, per upstream

Each of these changes who actually benefits, so they are stated rather than left implicit:

- The setting applies to **`INSERT SELECT`** — which is what this backfill issues — not to INSERTs generally.
- **ClickHouse Cloud does not default it to `0`.** Upstream documents `1` / `2` / `4` by node memory, so a Cloud deployment may already have parallelism here and see less benefit.
- It only helps if the read side is parallel too: *"Parallel `INSERT SELECT` has effect only if the `SELECT` part is executed in parallel"* (`max_threads`).

### What is documented, and what is inference

Worth separating, because the PR previously blurred it. Upstream states materialized values *"are automatically calculated according to the specified materialized expression when rows are inserted"* — but it does **not** state which pipeline stage or which threads perform that calculation. So attributing the bottleneck to the insert side, where this table's JSON-parsing `output_keys` is computed, is an **inference from profiling, not a documented guarantee.** The docs therefore tell an operator how to confirm it on their own data instead of asserting it.

## How to recognise this condition

The reusable part. **Effective cores sit near 1 while the machine is otherwise idle and `OSIOWaitMicroseconds` is 0** — the copy is neither CPU-saturated nor I/O bound, it is serialised. From `query_log`:

```sql
(ProfileEvents['UserTimeMicroseconds'] + ProfileEvents['SystemTimeMicroseconds']) / query_duration_ms
```

After raising the setting, effective cores rising towards the thread count confirms the diagnosis rather than merely correlating with it.

## What it is *not*

Each of these was ruled out on this profile before the sink was identified, and each is worth ruling out again elsewhere before reaching for the flag:

| Suspect | How it was ruled out |
|---|---|
| Disk I/O | `OSIOWaitMicroseconds` zero; `DiskWriteElapsed` a negligible share of window wall time |
| CPU saturation | node load and user CPU both low relative to core count |
| Read-thread starvation | the read side already had many threads participating |
| Client/network | `INSERT … SELECT` is server-side; only the query text crosses the wire |
| `max_insert_block_size` | tested — moved memory and time by a fraction of a percent |

## Why a flag and not a settings profile

The right value is a **per-run operational decision, not deployed state.** It depends on what share of cores the cutover may take *while serving traffic*, which differs between an idle rehearsal environment and production, and between an off-peak window and a daytime one. Tuning it through a chart profile costs a PR, a merge and a sync **per data point**.

By contrast, `max_memory_usage` belongs on the profile and is deliberately **not** exposed here: a memory ceiling is a guardrail against destabilising the server, so letting an operator raise their own limit per invocation defeats its purpose. Flags for pacing, profile for quotas.

## The default is *omission*, and omission means inherit

Originally this shipped as `default 0`, and the PR claimed that meant no deployment changed behaviour. **That was wrong**, caught in review and fixed in `96fc611`. An explicit `0` in a `SETTINGS` clause does not mean "leave alone" — it *overrides*. Verified on a cluster whose profile sets `3`:

```
SELECT getSetting('max_insert_threads')                                  -> 3
SELECT getSetting('max_insert_threads') SETTINGS max_insert_threads = 0  -> 0
```

So the original form would have forced serial inserts on every deployment that had already tuned the setting — ClickHouse Cloud (`1`/`2`/`4` by node memory) and our own environments included. A change advertised as a speed-up would have shipped as a **silent slowdown by default**.

Now: omitting the flag strips the setting line from the SQL entirely, so the server's value applies; an explicit `0` is retained and means *force* serial. Verified by extracting the drivers' own render block and running it against the real SQL — omitted renders no key, `0` renders `0`, `3` renders `3`, all three parse.

## Costs — both of the documented ones

**Memory.** Upstream is explicit: *"Higher values will lead to higher memory usage."* On this table that compounds with a known hazard — a single oversized `output` document can dominate insert memory by itself, which is what produced the `Code: 241` abort earlier in this cutover. So the docs say to move `max_memory_usage` with this setting, or narrow the window, rather than raising threads alone into a fixed ceiling.

**Parts.** Each insert thread writes its own parts, so **part count per partition grows**. The README says to watch it against `parts_to_throw_insert` (ClickHouse default `300`) rather than assume headroom, and the flag is bounded at 99 because a value beyond the machine's core count buys nothing while multiplying parts.

The docs also frame value selection as a **capacity decision, not a benchmark** — on an idle rehearsal environment a large value looks free, but on production those threads compete with live query latency, so validate the value you will actually deploy.

## Review round: four factual defects found and fixed

Review caught real errors in the docs this PR adds, two of which would have actively misled an operator. Each was verified before fixing rather than applied on faith.

| Finding | Verdict | Evidence |
|---|---|---|
| Docs named `input_keys` as a destination column | **Real — fixed** | `git grep input_keys` → **0 hits** repo-wide; `000115` states outright that `output_keys` is traces-only with no counterpart. An operator would have hunted a column that doesn't exist. |
| Effective-cores formula divided µs by ms | **Real — fixed** | Ran both forms on real backfill windows: as documented it returns **1628 effective cores on a 48-core node**; with `*1000` it returns **1.63**. The quoted measurements used the correct form, so only the *published* formula was wrong — worse, since that's the part an operator runs. |
| Formula measures query-wide CPU, not sink CPU | **Real — fixed** | `query_log` aggregates don't separate read from insert threads. Now stated, together with what *does* make it evidence: the delta when the setting is raised while the read side is unchanged. |
| `estimate.sh` doesn't model the setting | **Real — fixed** | It probes the read side and derates by a fixed `--write-cost-factor`; no thread input exists. Both `estimate.sh` and the README now say the ETA excludes sink parallelism and point at `--rows-per-sec`. |
| `000002` header said "ALL FOUR placeholders" listing five, doubled conjunction | **Real — fixed** | Stale count in the one comment whose purpose is that "a new one is never missed here". |
| `delta_replay.sh` Options header omitted the flag | **Real — fixed** | Every other flag had an entry; an operator reading the header couldn't discover this one. |
| Defaults block ordered opposite to header and arg parser | **Real — fixed** | Cosmetic, but the three lists are kept parallel so they diff by eye. |
| Add a bash test seam for flag validation | **Declined, with reasoning** | See below. |

**On the test-seam request.** There is no shell-test harness anywhere in `apps/opik-backend` — `0` bats files, no `tests/` dir in the cutover folder — so this would mean introducing test infrastructure for one-shot migration drivers that are deleted after the cutover completes. The validator is a single regex, and I exercised its full boundary on **both** drivers (`0`/`3`/`99` accepted; `100`/`abc`/`-1` rejected) rather than reasoning about it. I'd rather not land a harness whose only user is a script with a scheduled death; happy to be overruled if you disagree.

## Change checklist
- [ ] User facing
- [x] Documentation update

## Issues

- Resolves #
- OPIK-6901

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: Profiled a live backfill from `query_log` ProfileEvents, ruled out I/O, CPU saturation, read-thread starvation, the client path and block-size tuning, identified the sink, validated the fix by deploying a value and re-measuring, then authored the flag and docs.
- Human verification: The operator directed the value down so that what was validated is what production can deploy, and reviewed each measurement in-session. Reviewer sign-off on the diff still required.

## Testing

- `bash -n` clean on both drivers.
- Flag validation exercised: `0`, `3`, `99` accepted; `100`, `abc`, `-1` rejected with the expected message.
- Verified every `${...}` placeholder in **both** `000001` and `000002` is substituted by its driver, run with a control assertion so a silently-empty check cannot pass.
- The speedup is **measured on a live cluster, not projected**: a value was deployed via the settings profile and re-measured against the recorded baseline.
- **Every claim about `max_insert_threads` was checked against upstream**, not written from memory — read back from `system.settings` on the running server (26.3.16) and cross-checked against the ClickHouse docs. That check is what caught the Cloud-default error and the missing memory warning, both now fixed in this PR.
- Both SQL files parse: `clickhouse-format -n` clean on `000001` and `000002` with placeholders substituted, run differentially against the pre-edit file and with a deliberately-broken control to prove the parser was actually rejecting things.
- `bash -n` clean on all three touched drivers after the review fixes.
- **Flag boundary re-exercised on both drivers after the defaults-block reordering** (which could have regressed it): `0`/`3`/`99` accepted, `100`/`abc`/`-1` rejected with the corrected message. The first run of this test was malformed and reported every `delta_replay.sh` value as accepted; it was re-run with correct quoting rather than believed.
- The 1000× formula error was confirmed **empirically, not by inspection** — both forms run against `system.query_log` on a live cluster, with the node's core count as the sanity ceiling that makes the wrong answer obviously wrong.

Not run, with reason:

- `mvn -o test -Dtest=TracesLocalV2CutoverTest` — the gate test **deliberately omits** this setting. An absent `SETTINGS` key is exactly the default, so there is no behavioural drift, and adding a third hardcoded value would compound the already-noted gap that the test hardcodes settings rather than reading configured ones. Please confirm CI is green.

## Follow-up

The post-change sample is small. Direction and mechanism are solid and were confirmed by effective cores tracking the thread count; the precise multiple should be treated as indicative until it is confirmed over more windows.

## Documentation

Documentation **is** the substance of this change — the flag itself is a two-line diff. All of it lives with the code rather than in the docs site, because the audience is an operator running these scripts from a checkout, not a reader of `opik.com/docs`:

- `README.md` — the `--max-insert-threads` entry: what the setting does, the three upstream scope caveats (INSERT SELECT only, Cloud defaults differ, needs a parallel read side), how to confirm the diagnosis on your own data with correct units, both documented costs (memory and part count), and the `estimate.sh` caveat.
- `scripts/backfill.sh` and `scripts/delta_replay.sh` — full option docs in each `# Options:` header.
- `scripts/db-app-analytics/000001` and `000002` — a note beside the `SETTINGS` clause carrying the same caveats, so someone reading the SQL alone isn't missing them.
- `scripts/estimate.sh` — an explicit warning that its ETA excludes sink-thread effects.

No user-facing product documentation is affected: this is an internal migration runbook, and the flag defaults to current behaviour.

### DIFF
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
@@ -356,18 +356,34 @@ independent controls keep each statement safe:
     the `SELECT` part is executed in parallel"* (see `max_threads`).
 
   **Why the insert side is often the constraint on this table — and what is inference rather than
-  documented.** The destination carries JSON-parsing MATERIALIZED columns (`output_keys`, `input_keys`), and
-  upstream states materialized values *"are automatically calculated according to the specified materialized
-  expression when rows are inserted"*. Upstream does **not** state which pipeline stage or which threads
+  documented.** The destination carries per-row MATERIALIZED work the source does not. The expensive one is
+  **`output_keys`**, which parses the `output` JSON:
+  `arrayMap(key -> tuple(key, toString(JSONType(JSONExtractRaw(output, key)))), JSONExtractKeys(output))`.
+  There is **no `input_keys` column** — `output_keys` is traces-only and has no input counterpart, so don't go
+  looking for one. The table also materialises `truncated_input` / `truncated_output`, which substring-copy
+  documents that can be very large, plus `input_length` / `output_length` / `metadata_length`, `duration` and
+  `id_at`. Upstream states materialized values *"are automatically calculated according to the specified
+  materialized expression when rows are inserted"*. Upstream does **not** state which pipeline stage or which threads
   perform that calculation. That these columns make the insert side the bottleneck here is an **inference from
   profiling**, not a documented guarantee — so confirm it on your own data rather than assuming it:
 
   **How to confirm it:** effective cores sit near 1 while the machine is otherwise idle and
   `OSIOWaitMicroseconds` is 0 — the copy is neither CPU-saturated nor I/O bound, it is serialised. Compute
-  effective cores from `query_log` as
-  `(ProfileEvents['UserTimeMicroseconds'] + ProfileEvents['SystemTimeMicroseconds']) / query_duration_ms`.
-  After raising the setting, effective cores rising towards the thread count is what turns the inference into
-  a measurement.
+  effective cores from `query_log`, **minding the units** — the `ProfileEvents` are *micro*seconds while
+  `query_duration_ms` is *milli*seconds, so the `* 1000` is not optional:
+
+  ```sql
+  (ProfileEvents['UserTimeMicroseconds'] + ProfileEvents['SystemTimeMicroseconds'])
+      / (query_duration_ms * 1000)   -- omit the *1000 and the answer is 1000x too high
+  ```
+
+  Sanity-check the result against the node's core count: a value above it means the arithmetic is wrong, not
+  that the machine is busy.
+
+  **What this measures, precisely:** `query_log` aggregates are **query-wide CPU**. They do not separate
+  read-pipeline threads from insert-pipeline threads, so the number alone cannot attribute CPU to the insert
+  side. What makes it evidence is the *delta*: raise the setting and effective cores rise towards the thread
+  count while the read side is unchanged.
 
   **Two costs, both real:**
   - **Memory.** Upstream is explicit: *"Higher values will lead to higher memory usage."* On this table that
@@ -376,6 +392,12 @@ independent controls keep each statement safe:
     rather than raising threads alone into a fixed ceiling.
   - **Parts.** Each insert thread writes its own parts, so parts per partition grows. Watch it against
     `parts_to_throw_insert` (ClickHouse default 300) rather than assuming headroom.
+
+  **`estimate.sh` does not model this setting.** Its ETA probes the *read* side and derates it by a fixed
+  `--write-cost-factor`, so it is not specific to your thread count and will understate a run configured with
+  insert parallelism. If you are running with `--max-insert-threads` set, time one real window at that setting
+  and feed the measured throughput back via `estimate.sh --rows-per-sec`, which skips the probe and the
+  write-cost factor entirely.
 
   **Choosing a value is a capacity decision, not a benchmark:** on an idle rehearsal environment a large
   value looks free, but on production those threads compete with live query latency, so pick the share of
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
@@ -341,6 +341,46 @@ independent controls keep each statement safe:
   so peak insert memory is a small multiple of ~256 MB regardless of the window size — the statement does not load the
   window into memory. Lower this (or `min_insert_block_size_bytes`) on a memory-constrained data node.
 
+- **Insert pipeline threads (`--max-insert-threads`, default 0 → `SETTINGS max_insert_threads`).**
+  **This is often the throughput ceiling, and the default hides it.** ClickHouse documents the value `0`
+  (the default) as *"`INSERT SELECT` no parallel execution"*, so the insert side of this copy runs
+  serialised until you raise it. Raising it lets you control how much of the machine the backfill may
+  consume, and can speed the copy up substantially.
+
+  Two scope caveats worth knowing before you reach for it:
+  - The setting applies to **`INSERT SELECT`** — which is what this backfill issues — not to INSERTs
+    generally.
+  - **ClickHouse Cloud does not default it to `0`.** Upstream documents `1` / `2` / `4` by node memory, so
+    a Cloud deployment may already have parallelism here and see less benefit.
+  - It only helps if the read side is parallel too. Upstream: *"Parallel `INSERT SELECT` has effect only if
+    the `SELECT` part is executed in parallel"* (see `max_threads`).
+
+  **Why the insert side is often the constraint on this table — and what is inference rather than
+  documented.** The destination carries JSON-parsing MATERIALIZED columns (`output_keys`, `input_keys`), and
+  upstream states materialized values *"are automatically calculated according to the specified materialized
+  expression when rows are inserted"*. Upstream does **not** state which pipeline stage or which threads
+  perform that calculation. That these columns make the insert side the bottleneck here is an **inference from
+  profiling**, not a documented guarantee — so confirm it on your own data rather than assuming it:
+
+  **How to confirm it:** effective cores sit near 1 while the machine is otherwise idle and
+  `OSIOWaitMicroseconds` is 0 — the copy is neither CPU-saturated nor I/O bound, it is serialised. Compute
+  effective cores from `query_log` as
+  `(ProfileEvents['UserTimeMicroseconds'] + ProfileEvents['SystemTimeMicroseconds']) / query_duration_ms`.
+  After raising the setting, effective cores rising towards the thread count is what turns the inference into
+  a measurement.
+
+  **Two costs, both real:**
+  - **Memory.** Upstream is explicit: *"Higher values will lead to higher memory usage."* On this table that
+    compounds with a known hazard — a single oversized `output` document can dominate insert memory by itself
+    (see the memory section below) — so raise this together with `max_memory_usage`, or narrow the window,
+    rather than raising threads alone into a fixed ceiling.
+  - **Parts.** Each insert thread writes its own parts, so parts per partition grows. Watch it against
+    `parts_to_throw_insert` (ClickHouse default 300) rather than assuming headroom.
+
+  **Choosing a value is a capacity decision, not a benchmark:** on an idle rehearsal environment a large
+  value looks free, but on production those threads compete with live query latency, so pick the share of
+  cores the cutover may take while serving traffic, and validate the value you will actually deploy.
+
 - **Per-block partition bound (`--max-partitions-per-insert-block`, default 2000 → `SETTINGS
   max_partitions_per_insert_block`).** Not a throughput knob — a **correctness gate**. The destination is
   weekly-partitioned, so a block spans as many partitions as the ids in it imply; ClickHouse's default of 100 aborts the
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
@@ -341,11 +341,18 @@ independent controls keep each statement safe:
   so peak insert memory is a small multiple of ~256 MB regardless of the window size — the statement does not load the
   window into memory. Lower this (or `min_insert_block_size_bytes`) on a memory-constrained data node.
 
-- **Insert pipeline threads (`--max-insert-threads`, default 0 → `SETTINGS max_insert_threads`).**
-  **This is often the throughput ceiling, and the default hides it.** ClickHouse documents the value `0`
-  (the default) as *"`INSERT SELECT` no parallel execution"*, so the insert side of this copy runs
-  serialised until you raise it. Raising it lets you control how much of the machine the backfill may
-  consume, and can speed the copy up substantially.
+- **Insert pipeline threads (`--max-insert-threads`, omitted by default → `SETTINGS max_insert_threads`).**
+  **This is often the throughput ceiling, and ClickHouse's own default hides it.** Where nothing sets the
+  setting, ClickHouse documents `0` as *"`INSERT SELECT` no parallel execution"*, so the insert side of this
+  copy runs serialised. Passing a value lets you control how much of the machine the backfill may consume,
+  and can speed the copy up substantially.
+
+  **Omitting the flag means *inherit*, not zero.** The drivers strip the setting line from the SQL entirely
+  when the flag is not passed, so whatever the server profile sets applies. That distinction is load-bearing:
+  rendering an explicit `0` would **override** a profile that sets the setting and force the insert serial —
+  a silent slowdown, not a no-op. ClickHouse Cloud ships non-zero defaults (`1`/`2`/`4` by node memory), and a
+  self-managed cluster may set it in a profile too. Pass an explicit `0` when you actually want to *force*
+  serial execution.
 
   Two scope caveats worth knowing before you reach for it:
   - The setting applies to **`INSERT SELECT`** — which is what this backfill issues — not to INSERTs
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
@@ -342,77 +342,40 @@ independent controls keep each statement safe:
   window into memory. Lower this (or `min_insert_block_size_bytes`) on a memory-constrained data node.
 
 - **Insert pipeline threads (`--max-insert-threads`, omitted by default → `SETTINGS max_insert_threads`).**
-  **This is often the throughput ceiling, and ClickHouse's own default hides it.** Where nothing sets the
-  setting, ClickHouse documents `0` as *"`INSERT SELECT` no parallel execution"*, so the insert side of this
-  copy runs serialised. Passing a value lets you control how much of the machine the backfill may consume,
-  and can speed the copy up substantially.
+  Often the throughput ceiling: where nothing sets it, ClickHouse's default `0` means *"`INSERT SELECT` no
+  parallel execution"*, so the insert side runs serialised. Passing a value controls how much of the machine
+  the backfill may use and can speed the copy up substantially.
 
-  **Omitting the flag means *inherit*, not zero.** The drivers strip the setting line from the SQL entirely
-  when the flag is not passed, so whatever the server profile sets applies. That distinction is load-bearing:
-  rendering an explicit `0` would **override** a profile that sets the setting and force the insert serial —
-  a silent slowdown, not a no-op. ClickHouse Cloud ships non-zero defaults (`1`/`2`/`4` by node memory), and a
-  self-managed cluster may set it in a profile too. Pass an explicit `0` when you actually want to *force*
-  serial execution.
+  **Omitting the flag means *inherit*, not zero.** The drivers strip the setting line from the SQL when the
+  flag isn't passed, so the server's own value applies. Rendering an explicit `0` would **override** a profile
+  that sets it and force the insert serial — a silent slowdown, not a no-op. Pass `0` only to *force* serial.
 
-  Two scope caveats worth knowing before you reach for it:
-  - The setting applies to **`INSERT SELECT`** — which is what this backfill issues — not to INSERTs
-    generally.
-  - **ClickHouse Cloud does not default it to `0`.** Upstream documents `1` / `2` / `4` by node memory, so
-    a Cloud deployment may already have parallelism here and see less benefit.
-  - It only helps if the read side is parallel too. Upstream: *"Parallel `INSERT SELECT` has effect only if
-    the `SELECT` part is executed in parallel"* (see `max_threads`).
+  Three caveats, all from upstream: the setting applies to **`INSERT SELECT`** only; **ClickHouse Cloud
+  defaults it to `1`/`2`/`4`** by node memory, not `0`; and it helps only if the read side is parallel too
+  (*"has effect only if the `SELECT` part is executed in parallel"* — see `max_threads`).
 
-  **Why the insert side is often the constraint on this table — and what is inference rather than
-  documented.** The destination carries per-row MATERIALIZED work the source does not. The expensive one is
-  **`output_keys`**, which parses the `output` JSON:
-  `arrayMap(key -> tuple(key, toString(JSONType(JSONExtractRaw(output, key)))), JSONExtractKeys(output))`.
-  There is **no `input_keys` column** — `output_keys` is traces-only and has no input counterpart, so don't go
-  looking for one. The table also materialises `truncated_input` / `truncated_output`, which substring-copy
-  documents that can be very large, plus `input_length` / `output_length` / `metadata_length`, `duration` and
-  `id_at`. Upstream states materialized values *"are automatically calculated according to the specified
-  materialized expression when rows are inserted"*. Upstream does **not** state which pipeline stage or which threads
-  perform that calculation. That these columns make the insert side the bottleneck here is an **inference from
-  profiling**, not a documented guarantee — so confirm it on your own data rather than assuming it:
-
-  **How to confirm it:** effective cores sit near 1 while the machine is otherwise idle and
-  `OSIOWaitMicroseconds` is 0 — the copy is neither CPU-saturated nor I/O bound, it is serialised. Compute
-  effective cores from `query_log`, **minding the units** — the `ProfileEvents` are *micro*seconds while
-  `query_duration_ms` is *milli*seconds, so the `* 1000` is not optional:
+  **Why the insert side, and how sure we are.** The destination materialises `output_keys` by parsing the
+  `output` JSON (there is no `input_keys` column), plus `truncated_input`/`truncated_output` and the length,
+  `duration` and `id_at` columns. Upstream says materialized values are calculated *"when rows are inserted"*
+  but not by which stage, so blaming the insert side is an **inference from profiling**, not a guarantee.
+  Confirm it on your own data: effective cores near 1 while the machine is idle and `OSIOWaitMicroseconds` is
+  0, then rising towards the thread count once raised. Mind the units — `ProfileEvents` are microseconds,
+  `query_duration_ms` is milliseconds:
 
   ```sql
-  (ProfileEvents['UserTimeMicroseconds'] + ProfileEvents['SystemTimeMicroseconds'])
-      / (query_duration_ms * 1000)   -- omit the *1000 and the answer is 1000x too high
+  (ProfileEvents['UserTimeMicroseconds'] + ProfileEvents['SystemTimeMicroseconds']) / (query_duration_ms * 1000)
   ```
 
-  Sanity-check the result against the node's core count: a value above it means the arithmetic is wrong, not
-  that the machine is busy.
+  A result above the node's core count means the arithmetic is wrong. Note it is *query-wide* CPU: `query_log`
+  does not separate read from insert threads, so the delta on raising the setting is what carries the argument.
 
-  **What this measures, precisely:** `query_log` aggregates are **query-wide CPU**. They do not separate
-  read-pipeline threads from insert-pipeline threads, so the number alone cannot attribute CPU to the insert
-  side. What makes it evidence is the *delta*: raise the setting and effective cores rise towards the thread
-  count while the read side is unchanged.
-
-  **Two costs, both real:**
-  - **Memory.** Upstream is explicit: *"Higher values will lead to higher memory usage."* On this table that
-    compounds with this table's row width, and in a way the block bullet above does **not** cover. That bullet's
-    rule — peak insert memory is a small multiple of the 256 MB byte cap — holds for ordinary wide rows. It does
-    not hold for a single very large `output` document: materialising `output_keys` over one of those is a
-    **per-row** cost that no block cap bounds, so neither `max_insert_block_size` nor `min_insert_block_size_bytes`
-    constrains it. That is precisely why the lever here is the ceiling: raise this together with
-    `max_memory_usage`, or narrow the window,
-    rather than raising threads alone into a fixed ceiling.
-  - **Parts.** Each insert thread writes its own parts, so parts per partition grows. Watch it against
-    `parts_to_throw_insert` (ClickHouse default 300) rather than assuming headroom.
-
-  **`estimate.sh` does not model this setting.** Its ETA probes the *read* side and derates it by a fixed
-  `--write-cost-factor`, so it is not specific to your thread count and will understate a run configured with
-  insert parallelism. If you are running with `--max-insert-threads` set, time one real window at that setting
-  and feed the measured throughput back via `estimate.sh --rows-per-sec`, which skips the probe and the
-  write-cost factor entirely.
-
-  **Choosing a value is a capacity decision, not a benchmark:** on an idle rehearsal environment a large
-  value looks free, but on production those threads compete with live query latency, so pick the share of
-  cores the cutover may take while serving traffic, and validate the value you will actually deploy.
+  **Two costs.** Upstream: *"higher values will lead to higher memory usage"* — and on this table a single very
+  large `output` document is a **per-row** cost that no block cap bounds, so raise `max_memory_usage` alongside
+  or narrow the window. And parts per partition grow; watch them against `parts_to_throw_insert` (default 300).
+  Value choice is a capacity decision, not a benchmark: on an idle rehearsal box a large value looks free, but
+  on production those threads compete with live query latency. `estimate.sh` does **not** model this setting —
+  time a real window at your intended value and feed it back via `--rows-per-sec`.
+  Full diagnosis in `backfill.sh`'s `--max-insert-threads` option docs.
 
 - **Per-block partition bound (`--max-partitions-per-insert-block`, default 2000 → `SETTINGS
   max_partitions_per_insert_block`).** Not a throughput knob — a **correctness gate**. The destination is
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
@@ -387,8 +387,12 @@ independent controls keep each statement safe:
 
   **Two costs, both real:**
   - **Memory.** Upstream is explicit: *"Higher values will lead to higher memory usage."* On this table that
-    compounds with a known hazard — a single oversized `output` document can dominate insert memory by itself
-    (see the memory section below) — so raise this together with `max_memory_usage`, or narrow the window,
+    compounds with this table's row width, and in a way the block bullet above does **not** cover. That bullet's
+    rule — peak insert memory is a small multiple of the 256 MB byte cap — holds for ordinary wide rows. It does
+    not hold for a single very large `output` document: materialising `output_keys` over one of those is a
+    **per-row** cost that no block cap bounds, so neither `max_insert_block_size` nor `min_insert_block_size_bytes`
+    constrains it. That is precisely why the lever here is the ceiling: raise this together with
+    `max_memory_usage`, or narrow the window,
     rather than raising threads alone into a fixed ceiling.
   - **Parts.** Each insert thread writes its own parts, so parts per partition grows. Watch it against
     `parts_to_throw_insert` (ClickHouse default 300) rather than assuming headroom.
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
@@ -371,7 +371,10 @@ independent controls keep each statement safe:
 
   **Two costs.** Upstream: *"higher values will lead to higher memory usage"* — and on this table a single very
   large `output` document is a **per-row** cost that no block cap bounds, so raise `max_memory_usage` alongside
-  or narrow the window. And parts per partition grow; watch them against `parts_to_throw_insert` (default 300).
+  or narrow the window. And parts per partition grow; watch them against **this cluster's** `parts_to_throw_insert` and
+  `parts_to_delay_insert` — read them from `system.merge_tree_settings` rather than assuming ClickHouse's
+  defaults (300 / 150), because deployments routinely raise them and the real headroom can be an order of
+  magnitude different in either direction.
   Value choice is a capacity decision, not a benchmark: on an idle rehearsal box a large value looks free, but
   on production those threads compete with live query latency. `estimate.sh` does **not** model this setting —
   time a real window at your intended value and feed it back via `--rows-per-sec`.
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
@@ -328,6 +328,18 @@ run_backfill_window() {
         # non-zero default) applies. Rendering an explicit 0 here would OVERRIDE that and force the insert
         # serial -- a silent slowdown on any deployment that already sets the setting.
         sql="$(grep -vF 'max_insert_threads = ${MAX_INSERT_THREADS}' <<<"$sql")"
+        # The strip is a whitespace-exact match on a SETTINGS line in ANOTHER file. If that line is ever
+        # reformatted the match silently no-ops and the placeholder reaches the server as a literal -- the exact
+        # failure that file's own header warns about. This is the DEFAULT path, so fail loudly here instead.
+        # The strip matches ONE exact spelling of a SETTINGS line that lives in ANOTHER file. Reformat that line
+        # -- even one extra space -- and the strip silently no-ops, leaving the placeholder to reach the server
+        # as a literal. So assert on the OUTCOME, whitespace-agnostically, and skip `--` comment lines, which
+        # legitimately keep the placeholder text in the file headers. This is the DEFAULT path; fail loudly.
+        if grep -v '^[[:space:]]*--' <<<"$sql" | grep -qF '${MAX_INSERT_THREADS}'; then
+            echo "ERROR: max_insert_threads strip failed -- the SETTINGS line in $BACKFILL_SQL no longer matches the" >&2
+            echo "       pattern above (was it reformatted?). Refusing to send SQL with a literal placeholder." >&2
+            exit 2
+        fi
     else
         sql="${sql//'${MAX_INSERT_THREADS}'/$MAX_INSERT_THREADS}"
     fi
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
@@ -122,8 +122,11 @@
 #                                 or narrow the window, rather than raising threads alone into a fixed
 #                                 ceiling.
 #                             (2) PARTS. Each insert thread writes its own parts, so part count per
-#                                 partition grows. Watch it against parts_to_throw_insert (ClickHouse
-#                                 default 300) rather than assuming headroom.
+#                                 partition grows. Watch it against THIS cluster's parts_to_throw_insert
+#                                 and parts_to_delay_insert, read from system.merge_tree_settings -- do not
+#                                 assume ClickHouse's defaults (300 / 150). Deployments routinely raise
+#                                 them, so the real headroom can be an order of magnitude off either way,
+#                                 and "parts vs 300" is a misleading ratio on a tuned cluster.
 #
 #                             CHOOSING A VALUE IS A CAPACITY DECISION, NOT A BENCHMARK. On an idle
 #                             rehearsal environment a large value looks free; on a production cluster
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
@@ -60,9 +60,15 @@
 #                             per partition touched, compacted by background merges) for the INSERT completing at all.
 #                             See the runbook's "Far-future partitions from far-future-timestamp ids".
 #   --max-insert-threads N    threads for the INSERT SELECT pipeline (SETTINGS max_insert_threads).
-#                             Default 0. ClickHouse documents 0 (or 1) as "INSERT SELECT no parallel
-#                             execution", so the copy runs its insert side single-threaded unless this
-#                             is raised. NOTE the setting is scoped to INSERT SELECT, which is what
+#                             OMITTED BY DEFAULT, and omitted means INHERIT: the setting line is stripped
+#                             from the SQL, so whatever the server profile sets applies. This matters --
+#                             rendering an explicit 0 would OVERRIDE a profile that sets it and force the
+#                             insert serial, which is a silent slowdown rather than a no-op. ClickHouse
+#                             Cloud ships non-zero defaults (1/2/4 by node memory), and a self-managed
+#                             cluster may set it in a profile too.
+#                             Pass an explicit 0 to FORCE "INSERT SELECT no parallel execution"; pass N to
+#                             request N. Where nothing sets it, ClickHouse's own default is 0, so the
+#                             insert side runs single-threaded unless raised. NOTE the setting is scoped to INSERT SELECT, which is what
 #                             this backfill issues; it is not a general INSERT knob. NOTE ALSO that
 #                             ClickHouse CLOUD does not default it to 0 -- upstream documents 1 / 2 / 4
 #                             by node memory -- so on Cloud you may already have parallelism here.
@@ -168,8 +174,9 @@ MAX_PARTITIONS_PER_INSERT_BLOCK=2000  # partitions: SETTINGS max_partitions_per_
                           # destination is weekly-partitioned, so one block can span many partitions; ClickHouse's
                           # default of 100 THROWS (throw_on_max_partitions_per_insert_block=1). Far-future UUIDv7 ids
                           # make this reachable in practice — see the runbook's far-future section. 0 = unlimited.
-MAX_INSERT_THREADS=0      # threads: SETTINGS max_insert_threads for the INSERT SELECT. 0 is the ClickHouse default and
-                          # means "INSERT SELECT no parallel execution", which is usually the throughput ceiling here.
+MAX_INSERT_THREADS=""     # threads: SETTINGS max_insert_threads for the INSERT SELECT. EMPTY = inherit whatever the
+                          # server profile sets (the setting line is stripped from the SQL entirely). An explicit 0
+                          # is ClickHouse's own default and means "INSERT SELECT no parallel execution".
                           # Raising it trades memory and destination part count for copy speed — see the
                           # --max-insert-threads option docs above for the full diagnosis and both costs.
 DIVERGENCE="0.0001"       # fraction: max tolerated |src-dst|/src per settled window before aborting (0.01%).
@@ -222,7 +229,7 @@ done
 # 0 is meaningful (ClickHouse default, no parallel INSERT SELECT execution), so allow it. Bounded at 2 digits:
 # this is
 # a share of cores, and a value beyond the machine's core count buys nothing while multiplying parts.
-[[ "$MAX_INSERT_THREADS" =~ ^(0|[1-9][0-9]?)$ ]] || { echo "ERROR: --max-insert-threads must be 0 (ClickHouse default: no parallel INSERT SELECT execution) or 1..99." >&2; exit 2; }
+[[ -z "$MAX_INSERT_THREADS" || "$MAX_INSERT_THREADS" =~ ^(0|[1-9][0-9]?)$ ]] || { echo "ERROR: --max-insert-threads must be 0 (force no parallel INSERT SELECT execution) or 1..99; omit it entirely to inherit the server's setting." >&2; exit 2; }
 [[ "$FROM_WEEK" =~ ^[0-9]+$ ]] || { echo "ERROR: --from-week must be a non-negative integer." >&2; exit 2; }
 [[ -z "$TO_WEEK" || "$TO_WEEK" =~ ^[0-9]+$ ]] || { echo "ERROR: --to-week must be a non-negative integer." >&2; exit 2; }
 [[ "$PAUSE_SECONDS" =~ ^[0-9]+$ ]] || { echo "ERROR: --pause-seconds must be a non-negative integer." >&2; exit 2; }
@@ -316,7 +323,14 @@ run_backfill_window() {
     sql="${sql//'${WINDOW_HI}'/$hi}"
     sql="${sql//'${MAX_INSERT_BLOCK_SIZE}'/$MAX_INSERT_BLOCK_SIZE}"
     sql="${sql//'${MAX_PARTITIONS_PER_INSERT_BLOCK}'/$MAX_PARTITIONS_PER_INSERT_BLOCK}"
-    sql="${sql//'${MAX_INSERT_THREADS}'/$MAX_INSERT_THREADS}"
+    if [[ -z "$MAX_INSERT_THREADS" ]]; then
+        # Unset means INHERIT: strip the whole setting line so the server's profile (or ClickHouse Cloud's
+        # non-zero default) applies. Rendering an explicit 0 here would OVERRIDE that and force the insert
+        # serial -- a silent slowdown on any deployment that already sets the setting.
+        sql="$(grep -vF 'max_insert_threads = ${MAX_INSERT_THREADS}' <<<"$sql")"
+    else
+        sql="${sql//'${MAX_INSERT_THREADS}'/$MAX_INSERT_THREADS}"
+    fi
     clickhouse-client ${CH_HOST:+--host $CH_HOST} ${CH_PORT:+--port $CH_PORT} --database "$DATABASE" --multiquery --query "$sql"
 }
 
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
@@ -59,6 +59,56 @@
 #                             TOTAL distinct partition count. Raising it trades a larger part count per insert (one part
 #                             per partition touched, compacted by background merges) for the INSERT completing at all.
 #                             See the runbook's "Far-future partitions from far-future-timestamp ids".
+#   --max-insert-threads N    threads for the INSERT SELECT pipeline (SETTINGS max_insert_threads).
+#                             Default 0. ClickHouse documents 0 (or 1) as "INSERT SELECT no parallel
+#                             execution", so the copy runs its insert side single-threaded unless this
+#                             is raised. NOTE the setting is scoped to INSERT SELECT, which is what
+#                             this backfill issues; it is not a general INSERT knob. NOTE ALSO that
+#                             ClickHouse CLOUD does not default it to 0 -- upstream documents 1 / 2 / 4
+#                             by node memory -- so on Cloud you may already have parallelism here.
+#
+#                             Raising it lets you decide how much of the machine the backfill may use,
+#                             and can speed the copy up substantially when the insert side is the
+#                             constraint.
+#
+#                             PRECONDITION, per upstream: "Parallel INSERT SELECT has effect only if
+#                             the SELECT part is executed in parallel" (see max_threads). If the read
+#                             side is serialised, raising this buys nothing.
+#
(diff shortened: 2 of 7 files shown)


CometActions:
**No test needed here.**

This adds an opt-in `--max-insert-threads` flag to the traces-local-v2 cutover scripts, defaulting to 0 so no existing run changes behaviour, and `max_insert_threads` paces the INSERT SELECT pipeline rather than changing what gets copied — the rows the statements produce are identical. The rest of the diff is the README, the option docs and a javadoc note on TracesLocalV2CutoverTest; no application code changed. Nothing here surfaces on a page or an endpoint, and the flag is only reachable by hand-running the migration against a real ClickHouse cutover, so an e2e test has no way in. The arg-validation guard you added and backfill.sh's per-window reconciliation are the right gates for this one.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/32129919167)

_Re-checked after a push on 18 Aug 11:05 UTC — nothing the verdict depends on changed._

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/delta_replay.sh
```diff
@@ -45,13 +45,15 @@ CH_PORT=""                # native port; empty = clickhouse-client default (9000
 BACKFILL_START=""
 MAX_INSERT_BLOCK_SIZE=1048576
 MAX_PARTITIONS_PER_INSERT_BLOCK=2000  # partitions per block for the delta INSERT; see the option docs above. 0 = unlimited.
+MAX_INSERT_THREADS=0                  # threads for the delta INSERT SELECT; 0 = ClickHouse default (single-threaded sink).
```
JetoPistola: 💡 **suggestion** | Documentation

`--max-insert-threads` is wired into this script's arg parser (line 57) and validation (line 75), but the `# Options:` header block — the one place that documents this script's interface — gains no entry for it. Every other flag has one, including `--max-partitions-per-insert-block`, which was added by this same pattern and got a full entry.

An operator reading the header to find out what `delta_replay.sh` accepts won't discover the flag exists. Worth adding an entry after `--max-partitions-per-insert-block` (line 26-35), even a short one that defers to `backfill.sh` for the full diagnosis:

```
#   --max-insert-threads N    threads for the delta INSERT SELECT pipeline (SETTINGS max_insert_threads).
#                             Default 0 = ClickHouse default = single-threaded sink. Same knob and same
#                             costs as in backfill.sh (see its option docs); pass the SAME value used
#                             for the backfill.
```

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000002_delta_and_deletion_replay.sql
```diff
@@ -6,7 +6,7 @@
 -- this file, substitutes the placeholders and runs it — never run this file by hand. ALL FOUR placeholders it
 -- substitutes, so a new one is never missed here (an unsubstituted ${...} reaches the server as a literal and the
 -- statement fails): ${ANALYTICS_DB_DATABASE_NAME}, ${BACKFILL_START}, ${MAX_INSERT_BLOCK_SIZE} and
--- ${MAX_PARTITIONS_PER_INSERT_BLOCK}. Invocation:
+-- ${MAX_PARTITIONS_PER_INSERT_BLOCK} and ${MAX_INSERT_THREADS}. Invocation:
```
JetoPistola: 💡 **suggestion** | Documentation

The count went stale and the conjunction is doubled. As it now reads:

> **ALL FOUR** placeholders it substitutes … `${ANALYTICS_DB_DATABASE_NAME}`, `${BACKFILL_START}`, `${MAX_INSERT_BLOCK_SIZE}` **and** `${MAX_PARTITIONS_PER_INSERT_BLOCK}` **and** `${MAX_INSERT_THREADS}`

That's five items, labelled four, with two `and`s.

This matters a little more than ordinary prose drift: the comment exists specifically so "a new one is never missed here", so a wrong count undermines the check it's meant to be. (For what it's worth, I verified the substitution itself is complete — all five placeholders in this file are covered by `delta_replay.sh` lines 82-86. It's only the prose that's off.)

```suggestion
-- ${MAX_PARTITIONS_PER_INSERT_BLOCK} and ${MAX_INSERT_THREADS}. Invocation:
```

…paired with `ALL FOUR` → `ALL FIVE` on line 6.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
```diff
@@ -97,6 +125,9 @@ MAX_ROWS=2000000          # rows: per-statement bound; a week over this is halve
 MAX_INSERT_BLOCK_SIZE=1048576  # rows: SETTINGS max_insert_block_size for the INSERT. Peak memory is a small multiple of
                           # the smaller of this and min_insert_block_size_bytes (256 MB default), which dominates for wide
                           # trace rows; lower it on a memory-constrained node. 1048576 is the ClickHouse default.
+MAX_INSERT_THREADS=0                  # threads for the INSERT SELECT pipeline (SETTINGS max_insert_threads).
```
JetoPistola: 🧹 **nit** | Style

Ordering is inconsistent across the three lists this flag appears in:

| Location | Position |
|---|---|
| `# Options:` header (line 62) | *after* `--max-partitions-per-insert-block` |
| defaults block (this line) | *before* `MAX_PARTITIONS_PER_INSERT_BLOCK` |
| arg parser (line 154) | *after* `--max-partitions-per-insert-block` |

The file otherwise keeps these three lists in the same order, which makes them easy to diff by eye when adding a knob. Moving this declaration below `MAX_PARTITIONS_PER_INSERT_BLOCK` would restore that.

JetoPistola:
👋 **Review summary**

**What looks good**

- **The diagnosis is the deliverable, and it's documented as such.** Exposing a ClickHouse setting is a two-line diff; what makes this useful is the recognition procedure — effective cores near 1 with `OSIOWaitMicroseconds` at 0, computed from `query_log` — recorded so the next person can identify the condition instead of re-deriving it. The ruled-out table is the same instinct: negative results are usually the first thing lost.
- **Flag-vs-profile reasoning is genuinely principled**, not post-hoc. "Flags for pacing, profile for quotas" is a real distinction, and deliberately *not* exposing `max_memory_usage` — because letting an operator raise their own guardrail defeats it — is the part that shows the rule is load-bearing rather than decorative.
- **Costs are stated where they'll be read, not buried in the PR body.** Part-count growth against `parts_to_throw_insert`, and framing value selection as a capacity decision rather than a benchmark ("on an idle rehearsal environment a large value looks free"), is the sentence that stops someone deploying a number validated on an idle box.
- **Substitution wiring is complete.** I checked every placeholder against its driver: all 6 in `000001` against `backfill.sh`, all 5 in `000002` against `delta_replay.sh` — nothing unwired. Validation behaves as claimed: `0`/`3`/`99` accepted, `100`/`abc`/`-1`/`07`/empty rejected. `bash -n` clean on both.
- **Default 0 is the right call**, and the docs say plainly that the conservative default leaves throughput on the table rather than letting that be rediscovered.

**Overall**

Sound and well-scoped. The default-0 choice means no deployment changes behaviour, so risk is confined to operators who opt in — and they get the diagnosis, the cost, and the selection guidance alongside the knob. Nothing blocking; my comments are all doc-consistency, and two of them are the same underlying cost — the diagnosis now lives in four places (README, `backfill.sh` header, `backfill.sh` defaults comment, `000001` header), so a stale count in a fifth was fairly likely. Consistent with how `--max-partitions-per-insert-block` is treated, so I'm not asking you to change the style — just noting where it bites.

One item I couldn't leave inline, since the file isn't in this diff:

❓ **question** | Testing — `TracesLocalV2CutoverTest.java:905,925`

Agreed that omitting `max_insert_threads` from the gate test is behaviourally safe: an absent `SETTINGS` key *is* the default. But `000002`'s own header says the test "reimplements these statements inline; keep the two in step", and the gap you already note — the test hardcodes settings rather than reading configured ones — is now one setting wider. A one-line note in the test's Javadoc recording that the omission is deliberate would keep the next reader from treating it as drift. Your call whether that belongs here or in the follow-up.

**Inline comments**: 2 suggestions, 1 nit

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
```diff
@@ -59,6 +59,34 @@
 #                             TOTAL distinct partition count. Raising it trades a larger part count per insert (one part
 #                             per partition touched, compacted by background merges) for the INSERT completing at all.
 #                             See the runbook's "Far-future partitions from far-future-timestamp ids".
+#   --max-insert-threads N    threads for the INSERT SELECT pipeline (SETTINGS max_insert_threads).
+#                             Default 0, which is ClickHouse's own default and means a SINGLE-THREADED
+#                             insert stream. On a wide trace table that single thread is often the
+#                             throughput ceiling for this copy, because the sink is where the
+#                             destination's JSON-parsing MATERIALIZED columns (output_keys /
+#                             input_keys) are computed, which is CPU-heavy per row.
```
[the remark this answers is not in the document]
thiagohora: Correct, and worse than "misleading" — the column does not exist. `git grep input_keys` returns **0 hits** repo-wide, and `000115_create_spans_local_v2_table.sql` says it outright: *"No output_keys column: it is a traces-only column ... and has no"* counterpart. So an operator following my text would have gone looking for a column that was never there.

Fixed in `0f849a5`, and I took the second half of your suggestion too — the rationale now names the *actual* destination-derived work rather than just trimming the wrong name:

- `output_keys` — the expensive one, with the real expression quoted (`arrayMap(key -> tuple(key, toString(JSONType(JSONExtractRaw(output, key)))), JSONExtractKeys(output))`)
- `truncated_input` / `truncated_output` — substring copies of documents that can be very large, which matters for the memory cost
- `input_length` / `output_length` / `metadata_length`, `duration`, `id_at`

Corrected in all three places the wording was duplicated (`backfill.sh`, `README.md`, `000001`).

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
```diff
@@ -59,6 +59,34 @@
...
+#
+#                             Raising it lets you decide how much of the machine the backfill may use,
+#                             and can speed the copy up substantially when the sink is the constraint.
+#
+#                             HOW TO TELL THE SINK IS THE CONSTRAINT: effective cores sit near 1 while
+#                             the machine is otherwise idle and OSIOWaitMicroseconds is 0 -- i.e. the
+#                             copy is neither CPU-saturated nor I/O bound, it is serialised. Compute
+#                             effective cores from query_log as
+#                             (UserTimeMicroseconds + SystemTimeMicroseconds) / query_duration_ms.
+#                             A value near the thread count after raising this confirms it.
```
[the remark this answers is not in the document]
thiagohora: Right on both counts, and the units error is the more serious of the two findings in this review.

I checked it against real windows rather than by inspection. The two forms on 120 backfill windows:

| form | result |
|---|---|
| as documented (µs / ms) | **1628.16** |
| `/ (query_duration_ms * 1000)` | **1.63** |

The node has 48 cores, so 1628 is not a wrong-ish number, it is an impossible one. The measurements I quoted during review were computed with the correct form — which makes this worse rather than better, since the formula is precisely the part an operator copies and runs.

Fixed in `0f849a5` in all three places, with the units called out explicitly (`* 1000` "is not optional") and a sanity check added: a result above the node core count means the arithmetic is wrong, not that the machine is busy.

Your second point is also right and I have taken it: `query_log` aggregates are **query-wide CPU** and do not separate read-pipeline from insert-pipeline threads, so the figure alone cannot attribute CPU to the sink. The docs now say that, and say what *does* carry the argument — the delta when the setting is raised while the read side is unchanged. That is the same read/insert attribution caveat the PR body now flags as inference rather than documentation.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
```diff
@@ -120,6 +151,7 @@ while [[ $# -gt 0 ]]; do
         --max-rows-per-insert) MAX_ROWS="${2:?"$1 requires a value"}"; shift 2 ;;
         --max-insert-block-size) MAX_INSERT_BLOCK_SIZE="${2:?"$1 requires a value"}"; shift 2 ;;
         --max-partitions-per-insert-block) MAX_PARTITIONS_PER_INSERT_BLOCK="${2:?"$1 requires a value"}"; shift 2 ;;
+        --max-insert-threads) MAX_INSERT_THREADS="${2:?"$1 requires a value"}"; shift 2 ;;
```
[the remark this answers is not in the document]
thiagohora: Declining this one, with reasoning rather than a flat no — tell me if you disagree and I will add it.

There is no shell-test harness to extend: `0` bats files anywhere under `apps/opik-backend`, and no `tests/` directory in the cutover folder. So this means introducing test infrastructure whose only consumer is a set of one-shot migration drivers that get deleted once the cutover completes. The validator is a single regex per driver.

What I did instead is exercise the boundary on **both** drivers rather than reason about it — including re-running it after the defaults-block reordering in this same review round, which could plausibly have regressed it: `0`/`3`/`99` accepted, `100`/`abc`/`-1` rejected with the expected message. Worth noting the first run of that test was malformed (bad quoting) and reported every `delta_replay.sh` value as accepted; I re-ran it rather than believing it.

On `TracesLocalV2CutoverTest` bypassing the CLIs and not asserting `${MAX_INSERT_THREADS}` — that omission is deliberate and documented in the PR: an absent `SETTINGS` key is exactly the default, so there is no behavioural drift to catch, and hardcoding a third settings value would deepen the existing gap that the test hardcodes settings instead of reading the configured ones.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000002_delta_and_deletion_replay.sql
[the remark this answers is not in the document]
thiagohora: Real, and fixed in `0f849a5` — `ALL FOUR` → `ALL FIVE`, and the doubled `and` replaced with a comma so the list has a single final conjunction.

Agreed on why it matters more than ordinary prose drift: that comment exists so "a new one is never missed here", so a comment that miscounts the placeholders is failing at the one job it has.

thiagohora: Fixed in `0f849a5` — took the suggestion as written (comma before `${MAX_PARTITIONS_PER_INSERT_BLOCK}`) plus `ALL FOUR` → `ALL FIVE` on line 6.

And thanks for separately verifying the substitution itself is complete — that is the part that would actually break a run, and it means the defect really was confined to the prose.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/delta_replay.sh
thiagohora: Correct, and fixed in `0f849a5`. I had updated `backfill.sh`'s option docs and simply not carried it across, so the flag was discoverable in one driver and invisible in the other.

Added after the `--max-partitions-per-insert-block` entry, close to your draft, with one addition — an explicit instruction to pass the **same value used for the backfill**, since the delta writes into the same table through the same insert path and there is no reason for the two to diverge.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
thiagohora: Fixed in `0f849a5` — `MAX_INSERT_THREADS` moved below `MAX_PARTITIONS_PER_INSERT_BLOCK`, so all three lists now read partitions-then-threads.

Since moving a defaults declaration is exactly the kind of edit that can silently break argument handling, I re-exercised the flag boundary on both drivers afterwards: `0`/`3`/`99` accepted, `100`/`abc`/`-1` rejected.

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
```diff
@@ -341,6 +341,46 @@ independent controls keep each statement safe:
   so peak insert memory is a small multiple of ~256 MB regardless of the window size — the statement does not load the
   window into memory. Lower this (or `min_insert_block_size_bytes`) on a memory-constrained data node.
 
+- **Insert pipeline threads (`--max-insert-threads`, default 0 → `SETTINGS max_insert_threads`).**
+  **This is often the throughput ceiling, and the default hides it.** ClickHouse documents the value `0`
+  (the default) as *"`INSERT SELECT` no parallel execution"*, so the insert side of this copy runs
+  serialised until you raise it. Raising it lets you control how much of the machine the backfill may
+  consume, and can speed the copy up substantially.
```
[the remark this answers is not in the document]
thiagohora: Real gap, and fixed in `0f849a5` by taking the third option you offered — the warning rather than a model.

Confirmed first: `estimate.sh` has no `--max-insert-threads` and its ETA probes the **read** side then derates by a fixed `--write-cost-factor`, so nothing in it responds to sink parallelism. Modelling insert throughput properly would mean predicting the very thing this PR shows is data-dependent, so a model would be a more confident wrong answer than a warning.

Both `estimate.sh`'s header and the README now state that the ETA excludes sink-thread effects and is not specific to any thread count, and point at the existing escape hatch: time one real window **at the setting you intend to use** and pass it via `--rows-per-sec`, which bypasses both the probe and the write-cost factor.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
```diff
@@ -97,7 +164,10 @@ MAX_ROWS=2000000          # rows: per-statement bound; a week over this is halve
 MAX_INSERT_BLOCK_SIZE=1048576  # rows: SETTINGS max_insert_block_size for the INSERT. Peak memory is a small multiple of
                           # the smaller of this and min_insert_block_size_bytes (256 MB default), which dominates for wide
                           # trace rows; lower it on a memory-constrained node. 1048576 is the ClickHouse default.
+                          # 0 = ClickHouse default = SINGLE-THREADED sink, usually the throughput
+                          # ceiling here. See --max-insert-threads above for the measurements.
 MAX_PARTITIONS_PER_INSERT_BLOCK=2000  # partitions: SETTINGS max_partitions_per_insert_block for the INSERT. The
+MAX_INSERT_THREADS=0                  # threads for the INSERT SELECT pipeline (SETTINGS max_insert_threads).
                           # destination is weekly-partitioned, so one block can span many partitions; ClickHouse's
                           # default of 100 THROWS (throw_on_max_partitions_per_insert_block=1). Far-future UUIDv7 ids
                           # make this reachable in practice — see the runbook's far-future section. 0 = unlimited.
```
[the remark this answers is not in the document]
thiagohora: Real, and it is my damage from the previous round — the "cosmetic" reorder you flagged earlier moved only the *assignment* and orphaned its comment lines. It broke three things at once, not one:

- The thread explanation (`0 = ClickHouse default = SINGLE-THREADED sink...`) ended up under `MAX_INSERT_BLOCK_SIZE`, reading as documentation of block size.
- The assignment landed **between** `MAX_PARTITIONS_PER_INSERT_BLOCK` and its own continuation, so the weekly-partition / `default of 100 THROWS` / `0 = unlimited` text read as belonging to the thread setting. That is the misreading you describe, and `0 = unlimited` applied to threads is not just confusing, it is wrong — for threads `0` means *no parallel execution*, the opposite of unlimited.
- It left a dangling pointer, `See --max-insert-threads above for the measurements`, aimed at measurements removed from this file in an earlier commit.

Fixed in `1ae21c02fa` by rebuilding the region as three self-contained per-variable blocks rather than shuffling lines again — which is what `.agents/skills/opik-backend/SKILL.md` asks for, and what I should have done the first time. The thread comment now also uses upstream's wording ("INSERT SELECT no parallel execution") instead of the looser "single-threaded sink", and points at the option docs rather than at deleted content.

Re-ran the flag boundary afterwards (`0`/`3`/`99` accepted, `100`/`abc`/`-1` rejected), since this block has now demonstrated twice that edits to it can go wrong quietly.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000001_backfill_traces_local_v2.sql
```diff
@@ -24,6 +25,20 @@
...
 --     dedups regardless of insert order and background merges compact the parts. This is NOT a claim that rows arrive
 --     in sort-key order — do not rely on it (see README "Why slice by created_at").
+--   * SETTINGS max_insert_threads sizes the INSERT SELECT pipeline. ClickHouse documents 0 (the default) as
+--     "INSERT SELECT no parallel execution"; on ClickHouse Cloud the default is instead 1/2/4 by node memory.
+--     Raising it controls how much of the machine the backfill may use and can speed the copy up markedly,
+--     but only if the SELECT side is itself parallel (upstream: parallel INSERT SELECT "has effect only if the
+--     SELECT part is executed in parallel"). Upstream states materialized columns are calculated "when rows
+--     are inserted" but does NOT say which stage computes them -- that this table's JSON-parsing output_keys
+--     (there is no input_keys column) makes the insert side the constraint is an inference from profiling,
+--     confirmed by effective cores rising towards the thread count. Mind the units when computing those:
```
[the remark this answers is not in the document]
thiagohora: Agreed, and fixed in `1ae21c02fa` — though I went further than the suggested single sentence, because the run-on was hiding the one distinction that matters rather than merely reading awkwardly.

It is now three labelled paragraphs:

- **WHY THE INSERT SIDE, AND HOW SURE WE ARE** — states plainly that `output_keys` parsing is what makes the insert side the constraint, then that upstream does *not* say which stage computes materialized columns, so the attribution is an inference from profiling rather than a documented guarantee, and finally what supports it (effective cores rising towards the thread count while the read side is unchanged).
- the effective-cores computation, with the µs-versus-ms trap called out.
- **TWO COSTS** — memory and part count.

Your phrasing was the right shape; splitting it just stopped a single sentence from carrying a claim, a caveat, and the evidence for both.

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
```diff
@@ -341,6 +341,68 @@ independent controls keep each statement safe:
...
+
+  **What this measures, precisely:** `query_log` aggregates are **query-wide CPU**. They do not separate
+  read-pipeline threads from insert-pipeline threads, so the number alone cannot attribute CPU to the insert
+  side. What makes it evidence is the *delta*: raise the setting and effective cores rise towards the thread
+  count while the read side is unchanged.
+
+  **Two costs, both real:**
+  - **Memory.** Upstream is explicit: *"Higher values will lead to higher memory usage."* On this table that
+    compounds with a known hazard — a single oversized `output` document can dominate insert memory by itself
+    (see the memory section below) — so raise this together with `max_memory_usage`, or narrow the window,
```
JetoPistola: 💡 **suggestion** | Documentation

Two references in this sentence don't resolve, and they're the load-bearing half of cost (1).

**"see the memory section below"** — there is no memory section below. Headings after this line run `Why slice by created_at` (437) → `Delta and replay correctness` (606) → `How backfill.sh and 000001 relate` (643) → … The insert-memory discussion is at **lines 340-342**, i.e. *above* this text (the `max_insert_block_size` bullet, immediately preceding), and the block-size mechanism at 502. So "below" sends the reader forward past everything relevant.

**"a known hazard"** — `grep -rn oversized` across the runbook and `scripts/` returns only this line and its mirror at `backfill.sh:114`, both introduced by this PR. "Known" implies a prior write-up to consult; there isn't one.

Why this is worth fixing rather than leaving: the advice to raise `max_memory_usage` alongside threads rests entirely on that hazard's magnitude, so an operator who follows the pointer to size the increase finds nothing and has to guess — which is the outcome the rest of this section works hard to prevent.

Two options, either fine:
- Repoint it: `(see the block-size note above)` — lines 340-342 do explain that the 256 MB `min_insert_block_size_bytes` cap dominates for wide trace rows, which is the relevant mechanism.
- Or make it self-contained, dropping the cross-reference and the "known":

```suggestion
    compounds with this table's row width — a single oversized `output` document can dominate insert memory by
    itself, and the 256 MB `min_insert_block_size_bytes` cap noted above is what bounds it — so raise this
    together with `max_memory_usage`, or narrow the window,
```

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
```diff
@@ -147,6 +219,9 @@ done
 # rendered into the SQL and rejected by the server on the first INSERT — after the capacity preflight has passed and the
 # backfill_start anchor has been minted, which is a far more expensive place to discover a typo.
 [[ "$MAX_PARTITIONS_PER_INSERT_BLOCK" =~ ^(0|[1-9][0-9]{0,5})$ ]] || { echo "ERROR: --max-partitions-per-insert-block must be 0 (unlimited) or 1..999999." >&2; exit 2; }
+# 0 is meaningful (ClickHouse default, single-threaded sink), so allow it. Bounded at 2 digits: this is
```
JetoPistola: 🧹 **nit** | Style

Stale "sink" terminology — residue from `0f849a5`, which deliberately moved to "insert side" (3 uses here, 3 in the README) because "sink" over-claimed which pipeline stage does the work. Two definitional uses survived:

- this line — "ClickHouse default, single-threaded sink"
- `delta_replay.sh:54` — "0 = ClickHouse default (single-threaded sink)"

Worth flagging so it doesn't get over-corrected: **`backfill.sh:129` uses "sink" correctly** — "cannot by itself attribute the CPU to the sink" is a sentence *about attribution*, which is exactly the epistemic care `0f849a5` was adding. Only the two definitional uses should change.

```suggestion
# 0 is meaningful (ClickHouse default, no parallel INSERT SELECT execution), so allow it. Bounded at 2 digits:
# this is a share of cores, and a value beyond the machine's core count buys nothing while multiplying parts.
```

JetoPistola:
👋 **Review summary — re-review at `1ae21c02`** (previous pass was `5ab83c24`)

**All three findings from the last round are fixed**, and two more that other reviewers caught — the nonexistent `input_keys` column and the 1000× units error — are corrected too.

**What I re-verified on the current head**, rather than assuming the fixes held:

- **`bash -n` clean** on all three scripts, `estimate.sh` included now that it's in the diff.
- **Placeholder substitution exact-matched by set comparison** — `comm -3` between each SQL file's `${...}` set and its driver's substitution set returns empty both ways. No drift in either direction.
- **The `input_keys` correction is real.** `git grep input_keys` across `apps/opik-backend` now returns only the three lines of prose stating the column doesn't exist — nothing else in the repo references it. Good catch by whoever found that; I missed it last round.
- **The MATERIALIZED list matches the table.** Cross-checked against `000114_recreate_traces_local_v2_id_at_datetime64.sql`: `input_length`, `output_length`, `metadata_length`, `truncated_input`, `truncated_output`, `duration`, `id_at`, plus `output_keys` — exactly the set documented, and the `output_keys` `arrayMap`/`JSONExtractKeys` expression is quoted verbatim from the migration.
- **The units fix is right.** `ProfileEvents` are microseconds, `query_duration_ms` is milliseconds; the `* 1000` belongs there, and the core-count sanity check is a good addition for catching it next time.
- **Validation is consistent across both scripts** — identical regex, identical error text. `0`/`1`/`9`/`99` accept; `100`/`abc`/`-1`/`07`/whitespace/empty reject.
- **`estimate.sh` guidance is actionable** — both `--rows-per-sec` and `--write-cost-factor` exist, and `--rows-per-sec` does bypass the probe as claimed.
- **The defaults-block repair in `1ae21c0` is sound** — comment continuations now attach to the right variables, and the ordering matches the options header and arg parser.

**Overall**

The document got materially more honest, not just more correct. `0f849a5` reclassified the central claim from asserted fact to "an inference from profiling, not a documented guarantee", added the query-wide-CPU caveat explaining why the metric alone can't attribute CPU to the insert side, and pulled in three upstream constraints — Cloud's non-zero defaults, the parallel-`SELECT` precondition, the memory warning — that make the knob safer to reach for. Downgrading your own claim mid-review is the harder move and the right one. `1ae21c0` then fixed collateral damage from its own reordering and said so.

Both new comments are the tail of that cleanup rather than new defects: a cross-reference that points the wrong way, and two spots the terminology change missed.

**Inline comments**: 1 suggestion, 1 nit — nothing blocking.

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
thiagohora: Both references confirmed broken, and I checked each rather than taking them on trust:

- **"see the memory section below"** — verified there is none. Headings after that line are `Why slice by created_at` → `Delta and replay correctness` → `How backfill.sh and 000001 relate`, and the insert-memory discussion is in the block-size bullet **immediately above**. The pointer sent the reader forward past everything relevant, exactly as you describe.
- **"a known hazard"** — `oversized` appears in the runbook and `scripts/` only on the four lines this PR added. Nothing prior to consult, so "known" was doing work it hadn't earned.

**Fixed in `3bd3f4afdb`, but I did not take the suggested repointing** — and I think this is worth spelling out, because the suggestion would have reintroduced something this rehearsal disproved.

Leaning on the 256 MB `min_insert_block_size_bytes` cap as what bounds the hazard is not right. For a single very large `output` document the cost is materialising `output_keys` **per row**, and no block cap constrains it: during the rehearsal I dropped `max_insert_block_size` sharply and peak memory moved by a fraction of a percent. The block knobs bound *block* memory; they do nothing about one enormous row.

Which exposes something sharper than the dangling pointer: **the preceding bullet is wrong for this case.** It states peak insert memory is "a small multiple of ~256 MB regardless of the window size". True for ordinary wide rows, false for an oversized document — and it is the passage a reader would land on if I repointed there. So instead of cross-referencing it, the fix now states the exception explicitly: the block rule holds generally, the per-row materialisation cost sits outside it, and *that* is why the lever is `max_memory_usage` rather than a block knob. Self-contained, nothing to chase.

Your diagnosis of the impact was the useful part — the advice to raise the ceiling rests on the hazard's magnitude, so a broken pointer there leaves the operator guessing at exactly the wrong moment.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
thiagohora: Fixed in `3bd3f4afdb` — both definitional uses now read "no parallel INSERT SELECT execution", matching upstream's wording rather than my looser paraphrase.

And thank you for pre-empting the over-correction; I would probably have run `sed` across the file. Confirmed your read on both exclusions:

- **`backfill.sh:107`** — "cannot by itself attribute the CPU to the sink" is a sentence *about* attribution, which is the epistemic care `0f849a5` was adding rather than the over-claim it was removing. Left alone.
- **`README:575`** — "`max_insert_threads` governing the sink". You didn't raise this one, and I checked before touching it: it is pre-existing text, not in this diff, and it describes pipeline structure rather than defining the default. Left alone as out of scope, though it is arguably the same loose usage if someone wants it in a follow-up.

`sink` now appears in exactly those two places across the runbook and scripts.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
```diff
@@ -101,6 +168,10 @@ MAX_PARTITIONS_PER_INSERT_BLOCK=2000  # partitions: SETTINGS max_partitions_per_
                           # destination is weekly-partitioned, so one block can span many partitions; ClickHouse's
                           # default of 100 THROWS (throw_on_max_partitions_per_insert_block=1). Far-future UUIDv7 ids
                           # make this reachable in practice — see the runbook's far-future section. 0 = unlimited.
+MAX_INSERT_THREADS=0      # threads: SETTINGS max_insert_threads for the INSERT SELECT. 0 is the ClickHouse default and
+                          # means "INSERT SELECT no parallel execution", which is usually the throughput ceiling here.
+                          # Raising it trades memory and destination part count for copy speed — see the
+                          # --max-insert-threads option docs above for the full diagnosis and both costs.
```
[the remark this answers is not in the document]
thiagohora: **Confirmed, and this was the most important finding on the PR.** It falsified the claim I had been repeating in the description: *"default stays 0, so no existing deployment changes behaviour"*.

Verified against a live cluster whose profile sets `3`:

```
SELECT getSetting('max_insert_threads')                                  -> 3
SELECT getSetting('max_insert_threads') SETTINGS max_insert_threads = 0  -> 0
```

An explicit `0` in a `SETTINGS` clause does not mean *leave alone*, it **overrides**. Before this PR the SQL carried no such key and inherited the profile; after it, every invocation would have forced serial inserts unless the operator passed the flag. So a change advertised as speeding the copy up would have shipped as a **silent slowdown by default** — worst on precisely the deployments that had already tuned the setting: Cloud with its `1`/`2`/`4`, and our own environments, whose profile sets it.

Fixed in `96fc611`, taking your suggested shape exactly: **omitted now means inherit** (the drivers strip the whole setting line from the SQL), and an **explicit `0` is retained** as a deliberate *force serial*.

On regression coverage — I did not add a bash test harness (there is still none in this tree), but I did verify omission versus explicit zero rather than assert it, and did it by **extracting the drivers' own render block and running it against the real SQL files** so the test can't drift from the code:

| flag | rendered |
|---|---|
| omitted | *no* `max_insert_threads` key |
| `0` | `max_insert_threads = 0` |
| `3` | `max_insert_threads = 3` |

All three parse under `clickhouse-format -n`; both drivers accept omitted/`0`/`3`/`99` and reject `100`/`abc`/`-1`. README updated to describe inherit-versus-force rather than "default 0".

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/TracesLocalV2CutoverTest.java
```diff
@@ -104,6 +104,17 @@
  * {@code exchange_and_wrap.sh}'s replication-settle gate, {@code finalize.sh}'s empty-live refusal — are exercised by the
  * OPIK-6901 staging dry-run, not by this test (which runs the SQL those scripts wrap, directly). This test asserts the
  * logic is correct when invoked; the staging rehearsal asserts the scripts invoke it safely.
+ *
+ * <p><b>Tuning settings are deliberately NOT mirrored from the drivers, and that is not drift.</b> The statements
+ * below hardcode {@code max_insert_block_size = 100000} and carry neither {@code max_partitions_per_insert_block}
+ * nor {@code max_insert_threads}, while {@code backfill.sh} / {@code delta_replay.sh} make all three configurable.
```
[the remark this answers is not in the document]
thiagohora: **Correct, and my note was simply false.** The statements do hardcode `max_partitions_per_insert_block = 2000`, at both call sites.

The cause is worth recording: I grepped a local checkout pinned to an older revision instead of this branch, so I saw only `max_insert_block_size` and wrote the Javadoc from that. Same stale-checkout mistake this cutover has already been bitten by once — I should have read the branch.

Your classification point is the more valuable half and I have taken it. The partition limit is **not** pacing: a value below the number of partitions a block spans aborts the INSERT with `TOO_MANY_PARTS`, which is the exact failure the runbook's far-future section exists for. So it is mirrored *because* it is a correctness gate, and pinned at the drivers' default.

Rewritten in `96fc611` as a three-item list that says what each setting is and why it is treated as it is: partition limit mirrored (correctness gate), `max_insert_threads` omitted (pacing — and omission is now *meaningful*, since after `96fc611` an absent key inherits in the test exactly as it does in the drivers), block size fixed only for predictable CI memory. The standing limitation — hardcoded values can't catch a driver configured with an invalid one — is kept.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/delta_replay.sh
```diff
@@ -45,13 +51,16 @@ CH_PORT=""                # native port; empty = clickhouse-client default (9000
 BACKFILL_START=""
 MAX_INSERT_BLOCK_SIZE=1048576
 MAX_PARTITIONS_PER_INSERT_BLOCK=2000  # partitions per block for the delta INSERT; see the option docs above. 0 = unlimited.
+MAX_INSERT_THREADS=0                  # threads for the delta INSERT SELECT; 0 = ClickHouse default, i.e. no parallel
+                                      # INSERT SELECT execution.
```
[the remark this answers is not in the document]
thiagohora: Partly addressed by `96fc611`, and I'm declining the rest — with reasoning, since the underlying concern is legitimate.

**What changed:** the failure you describe was materially worse than a mismatch. With the old default, a delta run that omitted the flag rendered `max_insert_threads = 0` and *overrode* the server's setting, so it silently serialised. Now omission inherits, so a delta run that says nothing gets whatever the server gives it — the same as the backfill would. The silent-divergence case is mostly gone.

**Declining the state-file persistence.** `STATE_FILE` holds `backfill_start`, and that anchor is correctness-critical: re-minting it loses deletes. Its value comes from doing exactly one thing, so adding pacing state to it means a resume path that can now fail for two unrelated reasons, and a file where a stale pacing value looks as authoritative as the anchor.

**Declining the mismatch rejection**, because a mismatch is not an error. The delta typically runs in a different traffic window from the backfill — often a maintenance window — so wanting more threads for the delta than the backfill is a reasonable operator choice, not a mistake to block. Rejecting it would force `--max-insert-threads` to be passed identically to a step where it is a pacing knob.

What remains is a documentation matter, and `delta_replay.sh`'s option docs already say to pass the same value used for the backfill. Happy to reconsider if you think that's too weak a guarantee.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
```diff
@@ -240,6 +323,14 @@ run_backfill_window() {
     sql="${sql//'${WINDOW_HI}'/$hi}"
     sql="${sql//'${MAX_INSERT_BLOCK_SIZE}'/$MAX_INSERT_BLOCK_SIZE}"
     sql="${sql//'${MAX_PARTITIONS_PER_INSERT_BLOCK}'/$MAX_PARTITIONS_PER_INSERT_BLOCK}"
+    if [[ -z "$MAX_INSERT_THREADS" ]]; then
+        # Unset means INHERIT: strip the whole setting line so the server's profile (or ClickHouse Cloud's
+        # non-zero default) applies. Rendering an explicit 0 here would OVERRIDE that and force the insert
+        # serial -- a silent slowdown on any deployment that already sets the setting.
+        sql="$(grep -vF 'max_insert_threads = ${MAX_INSERT_THREADS}' <<<"$sql")"
```
JetoPistola, line 333: 💡 **suggestion** | Correctness

The inherit path is the right fix — I want to be clear I verified it works before raising the coupling. Running the real substitution logic against `000001`, the strip removes exactly one line and leaves valid SQL:

```sql
SETTINGS max_insert_block_size = 1048576,
         max_partitions_per_insert_block = 2000,
         log_comment = 'traces_local_v2_backfill:...';
```

No dangling comma, because the stripped line sits between two others. Same on `000002`. And the pattern is narrow enough to leave `000002:51`'s prose mention and the deletion-replay's separate `SETTINGS` clause alone.

**The coupling:** the match is whitespace-exact against SQL in another file. Reformat that line — even one extra space around `=` — and the strip silently no-ops, leaving `${MAX_INSERT_THREADS}` to reach ClickHouse as a literal. I reproduced it:

```console
$ printf 'SETTINGS a = 1,\n         max_insert_threads  = ${MAX_INSERT_THREADS},\n         log_comment = 1;\n' \
    | grep -vF 'max_insert_threads = ${MAX_INSERT_THREADS}'
SETTINGS a = 1,
         max_insert_threads  = ${MAX_INSERT_THREADS},   ← survived
         log_comment = 1;
```

That is precisely the failure both SQL headers warn about ("an unsubstituted `${...}` reaches the server as a literal and the statement fails"). It fails loudly rather than corrupting data, so not a blocker — but it is a silent-until-runtime dependency between two files with no guard, and the inherit path is now the **default**, so this is the code path most runs take.

A post-strip assertion closes it cheaply:

```suggestion
        sql="$(grep -vF 'max_insert_threads = ${MAX_INSERT_THREADS}' <<<"$sql")"
        # The strip above is a whitespace-exact match on the SQL's SETTINGS line. If that line is ever
        # reformatted, the match silently no-ops and the placeholder reaches the server as a literal.
        [[ "$sql" != *'max_insert_threads = ${MAX_INSERT_THREADS}'* ]] || { echo "ERROR: max_insert_threads strip failed -- the SETTINGS line in $BACKFILL_SQL was reformatted; update the pattern above." >&2; exit 2; }
```

Worth noting it can't be a blanket "no `${` remains" check — the header comments at `000001:15` and `000002:9` legitimately retain the placeholder text, which is why I matched the full setting expression instead.

Same applies to `delta_replay.sh:97`.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000001_backfill_traces_local_v2.sql
```diff
@@ -12,6 +12,7 @@
 --   ${WINDOW_LO} / ${WINDOW_HI}            the created_at half-open window bounds
 --   ${MAX_INSERT_BLOCK_SIZE}               rows per part-forming block
 --   ${MAX_PARTITIONS_PER_INSERT_BLOCK}     partitions one block may span (required; see the note below)
+--   ${MAX_INSERT_THREADS}                  threads for the INSERT SELECT pipeline (0 = ClickHouse default)
```
JetoPistola: 🧹 **nit** | Documentation

This line still describes the pre-`96fc611f` default, and now says close to the opposite of the truth:

> `${MAX_INSERT_THREADS}`  threads for the INSERT SELECT pipeline (**0 = ClickHouse default**)

After `96fc611f` the driver default is *omitted* (inherit the server's value), and `0` is the explicit **force-serial** value — the thing that commit exists to stop happening implicitly. The README was updated to "omitted by default" (line 344); this header wasn't.

```suggestion
--   ${MAX_INSERT_THREADS}                  threads for the INSERT SELECT pipeline. The driver OMITS this whole
--                                          SETTINGS line when --max-insert-threads is unset, so the server's
--                                          value is inherited; an explicit 0 forces no parallel execution.
```

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000002_delta_and_deletion_replay.sql
```diff
@@ -3,10 +3,10 @@
 -- Run this only after the whole backfill (step 1) is complete and reconciled.
 
 -- Step 0: The SQL below (delta-insert + deletion replay) is the single source driven by ../delta_replay.sh, which reads
--- this file, substitutes the placeholders and runs it — never run this file by hand. ALL FOUR placeholders it
+-- this file, substitutes the placeholders and runs it — never run this file by hand. ALL FIVE placeholders it
 -- substitutes, so a new one is never missed here (an unsubstituted ${...} reaches the server as a literal and the
--- statement fails): ${ANALYTICS_DB_DATABASE_NAME}, ${BACKFILL_START}, ${MAX_INSERT_BLOCK_SIZE} and
--- ${MAX_PARTITIONS_PER_INSERT_BLOCK}. Invocation:
+-- statement fails): ${ANALYTICS_DB_DATABASE_NAME}, ${BACKFILL_START}, ${MAX_INSERT_BLOCK_SIZE},
+-- ${MAX_PARTITIONS_PER_INSERT_BLOCK} and ${MAX_INSERT_THREADS}. Invocation:
```
JetoPistola: 🧹 **nit** | Documentation

The count is right again, but `96fc611f` made the surrounding claim conditionally untrue. "**ALL FIVE** placeholders it **substitutes**, so a new one is never missed here" — by default `delta_replay.sh` substitutes *four* and **deletes the fifth's line entirely**, since omitting `--max-insert-threads` is now the default path.

This matters slightly more than ordinary drift because this comment is the stated safety net for exactly this class of bug, and `${MAX_INSERT_THREADS}` is now the one placeholder it doesn't describe accurately. One clause fixes it:

```suggestion
-- statement fails): ${ANALYTICS_DB_DATABASE_NAME}, ${BACKFILL_START}, ${MAX_INSERT_BLOCK_SIZE},
-- ${MAX_PARTITIONS_PER_INSERT_BLOCK} and ${MAX_INSERT_THREADS} -- the last of which the driver instead REMOVES
-- from the SETTINGS clause when --max-insert-threads is omitted, so the server's value is inherited. Invocation:
```

JetoPistola:
👋 **Review summary — round 3 at `9c7dfae4`** (previous: `1ae21c02`)

Both findings from my last round are fixed in `3bd3f4af`. The headline change this round is `96fc611f`, and it's someone else's catch: rendering an explicit `max_insert_threads = 0` **overrides** a server- or Cloud-configured value rather than matching it, so the original default would have silently forced the insert serial on exactly the deployments the flag exists to help. Empty-means-inherit is the right correction.

**What I verified on the new inherit path** — this is a semantic change to generated SQL, so I ran it rather than read it:

- **The strip yields valid SQL.** Executed the real substitution logic against both files; the inherit path removes exactly one line, leaving `SETTINGS max_insert_block_size = …, max_partitions_per_insert_block = …, log_comment = …;`. No dangling comma — the stripped line sits between two others.
- **The explicit path is unaffected**, checked at `0` and `8`.
- **The leftover `${MAX_INSERT_THREADS}` is comment-only.** After the strip, the placeholder survives at `000001:15` and `000002:9` — both inside `--` comments, so it reaches the server as comment text and is ignored. Not a defect.
- **The `grep -vF` pattern's blast radius is correct**: it leaves `000002:51`'s prose mention and the deletion-replay's separate `SETTINGS` clause untouched.
- **Validation now accepts empty** and still rejects `100`/`abc`/`-1`/`07`/whitespace. `bash -n` clean on all three scripts.
- **The gate-test Javadoc is accurate now** — it correctly states the statements hardcode `max_partitions_per_insert_block` and the earlier claim that they omitted both settings was simply wrong.

**Inline comments**: 1 suggestion, 2 nits — nothing blocking. The suggestion is a whitespace-exact coupling between the strip pattern and the SQL's formatting (reproduced: one extra space and the strip silently no-ops); the two nits are `000001:15` and `000002:6-9` still describing pre-`96fc611f` behavior.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
thiagohora: Real, and thank you for verifying the strip works before raising the coupling — that ordering made the finding much easier to act on.

I reproduced your no-op and added a guard in `008b7ce`. **But I did not use the suggested assertion, because it has the same blind spot as the bug it guards.** It matches the same single exact spelling, so a reformatted line fails the strip *and* passes the assertion:

```console
# SETTINGS line reformatted with an extra space
$ grep -vF 'max_insert_threads = ${MAX_INSERT_THREADS}'   # strip: no-op, placeholder survives
$ [[ "$sql" != *'max_insert_threads = ${MAX_INSERT_THREADS}'* ]]   # assertion: TRUE, so it passes
```

I only found this because I built it and watched it not fire.

The guard now asserts on the **outcome**, whitespace-agnostically — after the strip, no *non-comment* line may still contain the placeholder:

```bash
if grep -v '^[[:space:]]*--' <<<"$sql" | grep -qF '${MAX_INSERT_THREADS}'; then
    echo "ERROR: ... Refusing to send SQL with a literal placeholder." >&2; exit 2
fi
```

Skipping `--` lines is exactly your point about `000001:15` and `000002:9` retaining the placeholder text — that is why a blanket "no `${` remains" check can't work, and it's what makes the comment filter necessary rather than incidental.

Tested against **two** reformat shapes, not one: an extra space before `=`, and the setting split across two lines. Both now fail closed; the exact-match form caught **neither**. Real file stays quiet, explicit-value path untouched, both rendered forms still parse. Applied to `delta_replay.sh` as well.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000001_backfill_traces_local_v2.sql
thiagohora: Correct — fixed in `008b7ce`, taking your wording. `0 = ClickHouse default` was true before `96fc611` and is now close to the inverse: the driver default is omission, and `0` is the explicit force-serial value that `96fc611` exists to stop happening implicitly.

You're right that I updated the README and left this behind; the header is the copy an operator reading the SQL directly will see first, so it was the worse of the two to leave stale.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000002_delta_and_deletion_replay.sql
thiagohora: Agreed, and fixed in `008b7ce` with your clause.

The mechanism claim was the part that mattered: by default the driver substitutes four placeholders and **deletes** the fifth's line, so "ALL FIVE placeholders it substitutes" was describing something the driver no longer does. And as you note, this comment is the stated safety net for unsubstituted placeholders — so it is the last place that should misdescribe how one of them is handled.

Worth connecting to the sibling comment on `backfill.sh:330`: the guard I added there asserts no *non-comment* line retains the placeholder, and this header is precisely one of the comment lines that legitimately does. So the two changes are consistent by construction rather than by luck.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000001_backfill_traces_local_v2.sql
```diff
@@ -24,6 +27,30 @@
...
+--
+--     Computing effective cores, minding the units -- the ProfileEvents are MICROseconds while
+--     query_duration_ms is MILLIseconds, so the *1000 is not optional:
+--         (UserTimeMicroseconds + SystemTimeMicroseconds) / (query_duration_ms * 1000)
+--     Omit it and the answer is 1000x too high. Sanity-check against the node's core count. And note the figure
+--     is QUERY-WIDE CPU: query_log does not separate read-pipeline from insert-pipeline threads.
+--
+--     TWO COSTS. Upstream warns "higher values will lead to higher memory usage", which on this table compounds
+--     with oversized `output` documents, so move max_memory_usage with it. And part count per partition grows,
+--     to be watched against parts_to_throw_insert. The value is a capacity decision about what share of cores the cutover may take
```
JetoPistola, line 52: 🧹 **nit** | Documentation

`aa5e8c11` is a good self-audit — using ClickHouse's default 300 as the *yardstick* when your cluster reads 3000 makes the ratio misleading in both directions, and including `parts_to_delay_insert` is the right call since throttling is what an operator meets first.

But it landed in the README and `backfill.sh` only (`git show aa5e8c11 --name-only` lists just those two), so this line still carries the framing that commit set out to remove:

> And part count per partition grows, **to be watched against `parts_to_throw_insert`.**

Same wrong denominator, one file behind. Suggest matching the other two:

```suggestion
--     with oversized `output` documents, so move max_memory_usage with it. And part count per partition grows, to
--     be watched against THIS cluster's parts_to_throw_insert and parts_to_delay_insert (read from
--     system.merge_tree_settings -- do not assume ClickHouse's 300 / 150 defaults). The value is a capacity
--     decision about what share of cores the cutover may take while serving traffic -- not a benchmark to maximise.
```

That also fixes the wrapping: the current line runs to ~130 chars where the rest of this block wraps at ~120, which reads like leftover from an earlier edit.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
```diff
@@ -240,6 +326,26 @@ run_backfill_window() {
...
+        # The strip is a whitespace-exact match on a SETTINGS line in ANOTHER file. If that line is ever
+        # reformatted the match silently no-ops and the placeholder reaches the server as a literal -- the exact
+        # failure that file's own header warns about. This is the DEFAULT path, so fail loudly here instead.
+        # The strip matches ONE exact spelling of a SETTINGS line that lives in ANOTHER file. Reformat that line
+        # -- even one extra space -- and the strip silently no-ops, leaving the placeholder to reach the server
+        # as a literal. So assert on the OUTCOME, whitespace-agnostically, and skip `--` comment lines, which
+        # legitimately keep the placeholder text in the file headers. This is the DEFAULT path; fail loudly.
```
JetoPistola, line 340: 🧹 **nit** | Style

The guard itself is better than what I suggested — asserting on the outcome whitespace-agnostically and skipping `--` comment lines is the right shape, and I confirmed it works in both directions: the normal inherit path doesn't trip it (the header residuals at `000001:15` / `000002:9` are correctly ignored), and re-running the doubled-space reformat from my last comment does trip it and exit 2.

The comment above it, though, says the same thing twice — "This is the DEFAULT path" appears in both paragraphs. This is partly on me: it reads like the comment I suggested landed on top of one you'd already written. The second paragraph is the better of the two, since it explains *why* the assertion is shaped the way it is; the first is now redundant.

```suggestion
        # The strip matches ONE exact spelling of a SETTINGS line that lives in ANOTHER file. Reformat that line
        # -- even one extra space -- and the strip silently no-ops, leaving the placeholder to reach the server
        # as a literal. So assert on the OUTCOME, whitespace-agnostically, and skip `--` comment lines, which
        # legitimately keep the placeholder text in the file headers. This is the DEFAULT path; fail loudly.
```

Same duplication at `delta_replay.sh:98-104`.

JetoPistola:
👋 **Review summary — round 4 at `aa5e8c11`** (previous: `9c7dfae4`)

All three round-3 findings are fixed in `008b7ce`, and **the guard you wrote is better than the one I proposed.** Mine was a whitespace-exact literal check; yours asserts on the outcome whitespace-agnostically and skips `--` comment lines, which handles the header-residual case my version would have needed a second rule for.

**I tested it in both directions rather than reading it:**
- **No false positive** — the normal inherit path passes cleanly; the legitimate placeholder text at `000001:15` and `000002:9` is correctly ignored.
- **Catches the real failure** — re-ran the exact doubled-space reformat from my last comment; the guard trips and exits 2.

Also re-verified this round: the inherit path still yields valid SQL in both files, validation accepts empty and rejects `100`/`abc`/`-1`/`07`, `bash -n` clean on all three scripts.

**`aa5e8c11` is the notable commit, and nobody asked for it.** Auditing your own "watch parts against `parts_to_throw_insert` (default 300)" against a live cluster and finding it reads 3000 is the kind of check that's easy to skip — and the conclusion is right: using the documented default as the *yardstick* misleads in both directions, generously on a tuned cluster and falsely reassuringly on a lowered one. Pulling in `parts_to_delay_insert` is the better half of the fix, since throttling is what an operator actually meets first. The stated 300/150 defaults check out.

**Inline comments**: 2 nits — nothing blocking, and both are mechanical residue rather than anything found by reasoning about the change. `000001:52` still carries the pre-`aa5e8c11` part-count framing (that commit touched only the README and `backfill.sh`), and the guard's comment block says the same thing twice in both drivers — partly my fault, since it looks like my suggested wording landed on top of yours.

For what it's worth on trajectory: rounds 1-3 found two real bugs and a real coupling. This round found neither, and the only correction came from you auditing yourself.

JetoPistola (approved):
Approving — the two open items are non-blocking (stale part-count framing in 000001:52, duplicated guard comment). Mechanics verified across review rounds; default preserves existing deployments.

PR #7884, opened 2026-08-18
[OPIK-6901] fix(cutover): close three silent-failure holes in max_insert_threads rendering

thiagohora:
## Details

Follow-up to #7881, which merged while this was being prepared, so this targets `main` directly. Review found three ways the `max_insert_threads` rendering could fail *silently*, all on paths that run by default. Each was reproduced before being fixed.

| # | Hole | Consequence |
|---|---|---|
| 1 | The guard was defeated by SIGPIPE | Placeholder reaches ClickHouse as a literal, guard says nothing |
| 2 | Deleting a shared line drops its neighbours | `log_comment` / partition bound silently vanish, after-check still passes |
| 3 | An explicit value could be silently dropped | `--max-insert-threads 3` renders nothing, server quietly uses its own |

### 1. The guard was defeated by SIGPIPE

The check was `grep -v … | grep -qF …`. `grep -q` exits on first match, the upstream grep takes SIGPIPE, and `set -o pipefail` — which **both drivers set** — then reports `141`. So a *true* match made the pipeline look like a *failure*, and the `if` body never ran.

```console
$ ( set -o pipefail; grep -v '^[[:space:]]*--' big.sql | grep -qF '${MAX_INSERT_THREADS}'; echo $? )
141          # match found, yet non-zero -> guard skips its own failure branch
```

Fixed by removing the pipe entirely: the filtered text goes to a variable, then one grep tests it.

**Worth recording how nearly this was missed.** My first attempt to verify it ran in an interactive **zsh**, where it did not reproduce — status `0`, guard fired, finding apparently disproved. The drivers are `#!/usr/bin/env bash`. Re-running under bash reproduced it immediately.

### 2. Deleting a shared line drops its neighbours

The strip removes a whole line by fixed string, so if the setting ever shares a line, its neighbours go with it — **and the after-check still passes**, because the placeholder really is gone.

```console
$ printf 'SETTINGS max_insert_threads = ${MAX_INSERT_THREADS}, log_comment = 1;\n' | grep -vF 'max_insert_threads = ${MAX_INSERT_THREADS}'
             # empty — log_comment went too
```

Added a **before**-check that the setting sits alone on its line.

### 3. An explicit value could be silently dropped

If the SQL has no placeholder, `${sql//…}` is a no-op — the requested value simply never appears. The else-branch now asserts an occurrence exists **in an executable line** first.

That qualifier is the fix's whole substance, and my first version got it wrong: both SQL headers quote the placeholder in `--` comments, so a substitution can "change something" while rendering no setting at all. My own render matrix caught it — now committed as `scripts/verify_render.sh`.

## Not done: extracting a shared helper

Review also suggested extracting the rendering into a shared shell utility, since it is duplicated between the two drivers. **Declined, and I'd like a second opinion rather than treating it as settled.**

There is no shared file among the nine scripts in this directory — each is standalone, run by an operator from a checkout during a cutover window. A `_common.sh` would be the first, and it makes every invocation depend on a sibling file being present at the moment it matters most. The drift risk is real, but this code is deleted when the cutover completes, so the window for drift is short and the cost of a missing file is immediate.

## Review round: executability, and running the guard early

Four findings from a human reviewer (`70b1a90e`), each verified by building the fixture, watching the old code accept it, then watching the fix reject it.

**A block-commented assignment satisfied the check** (medium). The line-anchored match had no notion of executability, and a commented assignment carries a trailing comma too — so `fe65be66` was orthogonal to this. With the assignment present only inside a multi-line `/* */` block:

| mode | before | after |
|---|---|---|
| explicit | renders, **zero executable settings** — the value went into the comment | ABORT |
| inherit | renders, having stripped a line inside a comment | ABORT |

The explicit row is **the genuinely silent failure on this PR**: no error anywhere, and the setting simply does not apply.

**The post-condition false-aborted on a comment** (low). A correct file plus a harmless `/* */` mention of the placeholder was refused. Same fix from the other side: one comment-masking helper now serves the count, the strip/substitute and the post-condition, so both ends agree by construction.

**The guard never ran under `--dry-run`** (medium). It sat in `run_backfill_window()`, which `insert_window()` reaches only after its `DRY_RUN` early return — so a malformed `SETTINGS` clause passed a full dry-run clean and aborted on the first real window. README:388 is exactly what sends operators to dry-run after editing that clause, so the rehearsal we recommend exercised none of the rendering. In a real run the abort also landed after the successor-table check, after `preflight_capacity()`, and after `backfill_start` had been minted. Now hoisted to startup, on the same reasoning this file already applies to `--max-partitions-per-insert-block`. `delta_replay.sh` already ran its copy at top level and was the correct model.

*Behaviour beyond timing: none.* The startup check reads the file's text, which does not change between windows, so it covers every window; the per-window path still resolves the line (the render needs its position) and keeps its own post-condition, so a mid-run edit is still caught.

**Stale wording** (low): "0 stays the default so no existing deployment changes behaviour" predates the inherit semantics. Corrected; it was the only survivor across drivers, README and this body.

Also corrected: the comment above the check claimed the anchored match ruled out "a `--` header, a `/* */` block comment or a string literal". It ruled out only `--`, and only incidentally.

### What the guard accepts and rejects now

**Accepts:** exactly one *executable*, isolated `max_insert_threads = ${MAX_INSERT_THREADS},` line — trailing comma required, comments masked.
**Rejects:** zero or several such lines · any layout carrying the clause terminator · a commented assignment · any executable line still holding the placeholder after rendering.
**Unchanged:** a file whose baseline is already invalid SQL still renders invalid SQL; no check on this one line can repair it.

## Review round: the guard could die before reporting

`ddb75df4` fixes a regression `70b1a90e` introduced in both drivers, found by review.

`mit_assignment_lines` ends in a pipeline whose `grep` exits 1 when nothing matches, so under the `set -euo pipefail` the drivers set, `mit_line="$(mit_assignment_lines "$sql")"` failed and **`set -e` killed the script on that line** — before the count check could emit its diagnostic.

| case | before | after |
|---|---|---|
| zero matches | `rc=1`, **0 diagnostic lines** | `rc=2`, diagnostic printed |
| two matches | `rc=2`, diagnostic | unchanged |
| good file | `rc=0`, silent | unchanged |

**Wider than the finding stated:** every zero-match case had it — reformatted line, shared line, missing assignment, block-commented assignment — so **seven** matrix cases died mute. The `>=2` case was never exposed, since `grep` succeeds there.

The comment above that line promises a mid-run edit "is still caught rather than rendered". It was caught — and handed the operator a bare `exit 1` mid-cutover, from the commit written to give them a diagnostic.

**Two defects in the testing, both fixed, both the reason this got through:**

- The matrix asserted only that a case *aborted*. A mute `set -e` death is indistinguishable from a working guard under that assertion. Abort verdicts now carry exit code and diagnostic presence: `ABORT(rc=2,diag)` vs `ABORT(rc=1,MUTE)`.
- The harness ran the block under `set -uo pipefail`, **without `-e`**, while the drivers set `-euo` — so even the stronger assertion could not see it. It now matches the drivers exactly.

Audited the rest of the rewrite: the other two command substitutions around pipelines already carry `|| true`; the `sed` calls run only after the count is confirmed to be one; the two `awk BEGIN` substitutions predate this work and cannot return 1.

## Change checklist
- [ ] User facing
- [x] Documentation update

## Issues

- Resolves #
- OPIK-6901

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: Reproduced each of the three failure modes, fixed the rendering, committed the render matrix as a runnable script and control-tested it twice, and corrected the documentation items.
- Human verification: Reviewer sign-off on the diff still required.

## Testing

**Validation is ad hoc and deliberately not committed.** An earlier revision of this PR added `scripts/verify_render.sh`; it is withdrawn in `6058043`. That directory holds operator drivers, run by hand during a cutover window and deleted when it completes — a test harness does not belong in it, and review of the committed file promptly found two defects in the harness itself (below). The README now states plainly that **nothing in the repo checks this rendering**, and that anyone editing either driver or the `SETTINGS` clause of `000001`/`000002` must exercise the failure cases by hand before a window.

The matrix still exists outside the repo and backs the results here. It extracts each driver's rendering block **verbatim** between markers rather than reimplementing it, so it cannot drift from the code it checks.

**14 cases per driver, 28 total, all passing** — every abort asserted on exit code *and* diagnostic, not merely on having aborted: inherit renders no key · explicit value renders · reformatted line aborts (inherit + explicit) · shared line aborts · missing assignment aborts (inherit + explicit) · **block-commented assignment aborts (inherit + explicit)** · **comment mention still renders (inherit + explicit)** · duplicate assignments abort · stray placeholder at scale aborts · oversized valid file still renders.

**Four controls prove the matrix can fail** — a green run from an unfalsified harness proves nothing, and this PR has been bitten by exactly that twice:

| control | result |
|---|---|
| END marker removed | every case fails, `HARNESS(extract)` |
| BEGIN marker duplicated | every case fails, `HARNESS(extract)` |
| syntax error inside the extracted block | cases fail loudly, never counted as the guard firing |
| piped post-condition reintroduced (the SIGPIPE bug) | `stray placeholder at scale` fails, `got[LEAK]` |
| `|| true` removed from the line lookup | **seven** cases fail as `ABORT(rc=1,MUTE)` |
| masking removed from the count | both `block-commented assignment` cases fail — reproducing the silent drop, `got[OK max_insert_threads = 3]` |
| masking removed from the post-condition | both `comment mention` cases fail, `got[ABORT]` |
| `--dry-run` against a block-commented `000001` | before: not caught · after: caught at startup · good file: silent |

The two harness defects review found, both now fixed in the ad-hoc copy:

- **Extraction was unvalidated** — it took the first `BEGIN..END` range, and with the END marker missing the range ran to end of file, *including the driver's real `clickhouse-client` invocation*. A harness advertised as touching nothing could have executed a query. It now requires exactly one ordered marker pair and refuses a block containing `clickhouse-client`/`curl`/`kubectl`.
- **Every nonzero status counted as the guard firing** — unreadable SQL, a syntax error in the block, or a failed extraction all mapped onto the expected `ABORT`, so a broken harness would score every negative case as a pass. A negative case now passes only on the renderer's own `exit 2` *with* an `ERROR:` diagnostic.

Also: `bash -n` clean on both drivers; `clickhouse-format -n` parses `000001` and `000002`, with a deliberately broken statement confirming the parser rejects.

## Documentation

- **Removed a false ClickHouse-defaults claim I introduced.** Earlier guidance cited `300/150` as the defaults for `parts_to_throw_insert` / `parts_to_delay_insert`. Checked against the bundled server: `system.merge_tree_settings` reports `changed=0` with `default=3000` / `default=1000`, so those are the stock defaults and nothing was raised — the accompanying claim that "deployments routinely raise them" was wrong too. The numeric claim is now gone from `000001`, `backfill.sh` and the README rather than restated with different numbers, since these values have moved across ClickHouse versions. That is precisely why the instruction is to read them from the server and never from memory.
- `README.md` no longer points at a committed verifier. It states instead that nothing in the repo checks this rendering, lists the silent failure modes, and says to exercise them by hand after editing either driver or the `SETTINGS` clause.
- **The trailing comma is now required** (`fe65be66`), and the rationale this section previously gave was wrong. It said tightening would newly reject `max_insert_threads` as the last setting in the clause; that layout was *already* rejected, since `,?` admits a comma or nothing but **not** a semicolon. Testing the four layouts through `clickhouse-format` also found a case neither the rationale nor the review named: *last, bare, `;` on the next line* is valid SQL, **passed** the pre-check, and inherit stripped it to invalid SQL. To be precise about the severity: ClickHouse rejects that with a syntax error, so it fails loudly — but at the server, mid-run, instead of at the pre-check where it belongs. Both terminator-carrying layouts now abort in both modes.
- The suggested `[,;]?` was tested and **not** applied: it additionally accepts the semicolon layout, which then strips to invalid SQL — trading a correct pre-check abort for a server-side syntax error mid-run.
- The constraint this imposes is documented rather than hidden — the assignment must not be the last entry in the `SETTINGS` clause — in the README and in the error message, which says why.
- **Corrected a false claim about the fenced block.** The README said to run an extracted block from inside `scripts/` "since the block resolves `000001`/`000002` by relative path". The block reads no files: it operates on a `$sql` variable the caller must populate, naming the SQL path only in its two diagnostics, and the drivers are CWD-independent via `SCRIPT_DIR`. The advice sent an operator hunting a CWD problem that cannot exist. Replaced with the contract that matters: the caller populates `$sql`.


CometActions:
**No test needed here.**

This hardens the max_insert_threads rendering inside the two cutover drivers (backfill.sh, delta_replay.sh) plus their runbook comments — operator-run migration tooling under apps/opik-backend/data-migrations/, not code the product executes. Nothing here renders in the UI or answers an endpoint: the new guards' only observable effect is the driver exiting 2 in front of an operator mid-cutover, which no e2e spec can reach on a fresh OSS install (it needs a populated ClickHouse mid-cutover and a hand-corrupted 000001/000002). The cutover SQL itself already has its executable spec in TracesLocalV2CutoverTest and a rehearsal rig in tests_load/tests/traces-local-v2-cutover/, and your README says outright that the bash rendering has no committed harness by design — the fenced BEGIN/END block is the right way to exercise it. No e2e test needed.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/32149694027)

_Re-checked after a push on 18 Aug 14:41 UTC — nothing the verdict depends on changed._

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000001_backfill_traces_local_v2.sql
```diff
@@ -49,7 +49,9 @@
 --
 --     TWO COSTS. Upstream warns "higher values will lead to higher memory usage", which on this table compounds
 --     with oversized `output` documents, so move max_memory_usage with it. And part count per partition grows,
---     to be watched against parts_to_throw_insert. The value is a capacity decision about what share of cores the cutover may take
+--     to be watched against THIS cluster's parts_to_throw_insert and parts_to_delay_insert (read them from
+--     system.merge_tree_settings; do not assume ClickHouse's 300/150 defaults -- deployments routinely raise
+--     them, and throttling at the delay limit is what an operator meets first). The value is a capacity decision about what share of cores the cutover may take
```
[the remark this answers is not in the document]
thiagohora: **You are right, and the error was mine — introduced by the very commit that was supposed to fix this guidance.**

Checked against the bundled server rather than from memory:

```
SELECT name, value, changed, default FROM system.merge_tree_settings
WHERE name IN ('parts_to_throw_insert','parts_to_delay_insert')

parts_to_delay_insert  1000  changed=0  default=1000
parts_to_throw_insert  3000  changed=0  default=3000
```

`changed=0` settles it: 3000/1000 **are** the stock defaults on this version. So not only was `300/150` wrong, the surrounding claim that "deployments routinely raise them" was wrong too — I had read a stock cluster as a tuned one, and built the rationale on that.

Fixed in `b9a8dc5` by **removing the numeric claim entirely** rather than swapping in 3000/1000, in all three places (`000001`, `backfill.sh`, and the README, which had the same sentence). Restating a different number would repeat the mistake one version later — these values have moved across ClickHouse releases, which is exactly why the instruction is to read them from `system.merge_tree_settings` and never from memory. The wording now says that explicitly.

Worth noting the consequence beyond the docs: this also corrected an operational reading on our side, where a stock value had been reported as evidence of a tuned cluster.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
```diff
@@ -327,23 +327,44 @@ run_backfill_window() {
...
+        # requested value would simply not appear and the server would quietly use its own. Check for an
+        # EXECUTABLE occurrence specifically: both SQL headers mention the placeholder in `--` comments, so a
+        # substitution can "change something" while rendering no actual setting.
+        executable="$(grep -v '^[[:space:]]*--' <<<"$sql" || true)"
+        if ! grep -qF '${MAX_INSERT_THREADS}' <<<"$executable"; then
+            echo "ERROR: --max-insert-threads $MAX_INSERT_THREADS was requested, but $BACKFILL_SQL has no" >&2
+            echo "       \${MAX_INSERT_THREADS} in any executable line, so the setting would be silently dropped." >&2
+            exit 2
+        fi
         sql="${sql//'${MAX_INSERT_THREADS}'/$MAX_INSERT_THREADS}"
```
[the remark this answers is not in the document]
thiagohora: Real, and fixed in `b9a8dc5` — with a stronger fix than the token check I had.

The guard asked whether `${MAX_INSERT_THREADS}` appeared on any non-`--` line. As you say, that accepts the placeholder inside a `/* … */` block comment or a string literal, so a file with **no actual SETTINGS assignment** could pass and run with the requested value silently dropped. (No such comment exists in these files today — this is hardening, not a live break — but the check was claiming a property it did not test.)

Both drivers now require **exactly one line-anchored, isolated assignment** before rendering anything:

```bash
assignments="$(grep -cE '^[[:space:]]*max_insert_threads = \$\{MAX_INSERT_THREADS\},?[[:space:]]*$' <<<"$sql" || true)"
[[ "$assignments" -eq 1 ]] || abort
```

That one precondition **subsumes two guards** the previous revision needed separately — a shared line and a reformatted line both fail it — and adds two cases neither covered: a placeholder living only in a block comment, and a file carrying **two** assignments, which is ambiguous and should not be guessed at.

Your regression case is committed. `scripts/verify_render.sh` includes *placeholder only in `/* */` aborts* and *duplicate assignments abort*, and I control-tested it by weakening the precondition — those two are among the five cases that then fail.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
```diff
@@ -327,23 +327,44 @@ run_backfill_window() {
...
+        #
+        # Note the deliberate absence of a pipe into `grep -q`: it exits on first match, the upstream grep takes
+        # SIGPIPE, and `set -o pipefail` then reports 141 -- so the guard would skip its own failure branch.
+        hits="$(grep -F 'max_insert_threads = ${MAX_INSERT_THREADS}' <<<"$sql" || true)"
+        shared="$(grep -vE '^[[:space:]]*max_insert_threads = \$\{MAX_INSERT_THREADS\},?[[:space:]]*$' <<<"$hits" || true)"
+        if [[ -n "${shared//[[:space:]]/}" ]]; then
+            echo "ERROR: max_insert_threads shares a SETTINGS line with another setting in $BACKFILL_SQL:" >&2
+            echo "       $shared" >&2
+            echo "       Dropping that line would silently drop its neighbours. Put the setting on its own line." >&2
+            exit 2
```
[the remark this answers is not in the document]
thiagohora: Fair, and I've committed it rather than relabelling — `scripts/verify_render.sh` in `b9a8dc5`.

You offered either option; committing was the better one here, because the drift risk is real and I had *declined* to remove the duplication between the two drivers in the previous round. Leaving the only coverage in my shell meant the duplication had nothing watching it.

Design notes, since the obvious implementation would have been worse than nothing:

- **It extracts each driver's rendering block verbatim**, between `# >>> BEGIN` / `# <<< END` markers, and evaluates that. A reimplementation of the logic would drift from the drivers and keep passing while they broke — the exact failure mode you are raising.
- **No cluster, no SQL executed.** It builds corrupted copies of the real `000001`/`000002` in a temp dir. Runs in ~15s.
- **11 cases per driver, 22 total**: inherit, explicit, reformatted line (both paths), shared line, missing assignment (both paths), placeholder only in `/* */`, duplicate assignments, a stray placeholder at scale, and an oversized-but-valid file.

I **control-tested the harness twice**, because a green matrix from an untested harness proves nothing:

1. Weakening the anchored precondition → **5 failures**.
2. Reintroducing the piped post-condition this PR removed → **1 failure**, `got[LEAK]`.

Control 2 initially did *not* fail, which exposed a flaw in my own test: the bulk filler was `--` comment lines, which the first grep strips, so no pipe pressure built and the SIGPIPE path was never exercised. Filler is now executable lines and the control fails as it should. That is recorded in the script's comments so the next person doesn't reintroduce it.

On `TracesLocalV2CutoverTest` excluding the bash drivers — that stays true and is documented in its Javadoc; this script is the driver-level complement, not a replacement.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/verify_render.sh
```diff
@@ -0,0 +1,111 @@
...
+set -uo pipefail
+cd "$(dirname "${BASH_SOURCE[0]}")"
+
+pass=0; fail=0
+BEGIN_MARK='# >>> BEGIN max_insert_threads rendering'
+END_MARK='# <<< END max_insert_threads rendering'
+
+extract() { # extract <driver>
+    awk -v b="$BEGIN_MARK" -v e="$END_MARK" 'index($0,b){f=1} f{print} index($0,e){exit}' "$1"
+}
```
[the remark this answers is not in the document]
thiagohora: Both halves real, and the second is worse than the wording suggests — **the harness could have executed a query.** Fixed in `6058043`, though not the way you proposed; see the last paragraph.

My `extract()` took the first `BEGIN..END` range it found. My awk was `index($0,b){f=1} f{print} index($0,e){exit}` — so with the **END marker missing there is no exit**, the range runs to end of file, and the extracted text includes the driver's real `clickhouse-client` invocation. A script whose whole selling point was "runs no SQL, needs no cluster" would then have run one.

Verified on a copy with the END marker stripped. Hardened extractor: requires **exactly one** BEGIN and one END, in order, refuses a block that does not look like the renderer, and refuses one containing `clickhouse-client`/`curl`/`kubectl`. It now reports `HARNESS(extract)` and every case fails loudly:

```
FAIL  inherit renders no key        got[HARNESS(extract)] want[OK none]
FAIL  reformatted aborts (inherit)  got[HARNESS(extract)] want[ABORT]
```

Duplicated BEGIN marker: same refusal, confirmed separately.

**Where I went further than the suggestion: the file is withdrawn from the PR entirely.** Rather than harden a test harness inside a directory of operator drivers — run by hand mid-cutover, deleted when it completes — `scripts/verify_render.sh` is deleted in `6058043`. That removes this finding's subject along with it. The hardened version lives outside the repo and still backs the results reported here; the README now states plainly that nothing in the repo checks this rendering and that an editor must exercise the failure cases by hand before a window. That is a reversal of what I told you last round, and the honest version of it.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/verify_render.sh
```diff
@@ -0,0 +1,111 @@
...
+        BACKFILL_SQL="$sqlfile"; SQL_FILE="$sqlfile"
+        sql="$(cat "$sqlfile")"
+        MAX_INSERT_THREADS="$threads"
+        eval "$(extract "$drv")" || exit $?
+        v="$(grep -oE 'max_insert_threads = [0-9]+' <<<"$sql" | head -1 || true)"
+        if grep -qF '${MAX_INSERT_THREADS}' <<<"$(grep -v '^[[:space:]]*--' <<<"$sql" || true)"; then
+            echo "LEAK"
+        else
+            echo "OK ${v:-none}"
+        fi
```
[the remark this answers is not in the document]
thiagohora: Real, and it is the same vacuous-test failure this PR has already hit twice — a negative case that passes for the wrong reason proves nothing.

`render()` mapped **any** nonzero status onto `ABORT`: unreadable SQL, a syntax error in the extracted block, a failed extraction. So a completely broken harness would have scored every negative case as a pass and reported green.

Fixed in the ad-hoc copy: a negative case now passes **only** on the renderer's own `exit 2` **and** an `ERROR:` diagnostic on stderr. Anything else reports `HARNESS(rc=N)`. Verified by injecting a syntax error into the extracted block — the cases now fail loudly (`got[LEAK] want[ABORT]`) instead of passing.

Checked the bash exit codes rather than assuming, since the fix depends on them being distinguishable: `eval` syntax error → `1`, the renderer's guard → `2`, a missing SQL file → `1`. Hence the extra stderr requirement rather than relying on the status alone.

Note the file itself is **withdrawn from the PR** in `6058043` (see the sibling thread), so this fix lands in the ad-hoc copy that backs the reported results, not in the repo.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
```diff
@@ -326,26 +326,36 @@ run_backfill_window() {
...
+    # This depends on ONE exact line in ANOTHER file:  max_insert_threads = ${MAX_INSERT_THREADS},
+    # Require exactly one line-anchored, ISOLATED assignment before touching anything. A bare token match would
+    # also accept the placeholder inside a `--` header, a /* */ block comment or a string literal, and let the run
+    # proceed with no setting rendered at all. Anchoring also catches the two layout hazards in one check: a
+    # shared line (which cannot be removed without dropping its neighbours) and a reformatted one (which the
+    # fixed-string strip would silently miss).
+    assignments="$(grep -cE '^[[:space:]]*max_insert_threads = \$\{MAX_INSERT_THREADS\},?[[:space:]]*$' <<<"$sql" || true)"
+    if [[ "$assignments" -ne 1 ]]; then
+        echo "ERROR: expected exactly ONE line reading 'max_insert_threads = \${MAX_INSERT_THREADS},' in" >&2
+        echo "       $BACKFILL_SQL, found $assignments. The setting must sit alone on its own line." >&2
```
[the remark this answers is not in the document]
thiagohora: Real, fixed in `6058043` — the message quoted the line *with* a trailing comma while the regex accepts `,?`, so a failure for any other reason advertised syntax that was not actually required. It now names the accepted form and marks the comma optional, in both drivers.

**I did not take the suggested alternative of tightening the regex to require the comma.** The comma is present only because another setting follows it in the current template; `max_insert_threads` as the *last* setting in the clause is legitimate SQL, and requiring the comma would reject it for no safety gain — the guard exists to catch a missing or shared assignment, not to enforce clause ordering. Aligning the message costs nothing and rejects nothing valid. Happy to be overruled if you would rather the template be pinned exactly.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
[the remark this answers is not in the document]
thiagohora: Following up, because I reversed this: **`scripts/verify_render.sh` is withdrawn in `6058043`.**

Your original point stands and I am not retracting it — "I ran it in my shell" is not coverage. But committing a test harness into a directory of operator drivers was the wrong way to answer it: these scripts are run by hand during a cutover window and deleted when it completes, and review of the committed file immediately surfaced two defects in it, including one where a missing end-marker would have let it `eval` the driver's real `clickhouse-client` call.

So the consequence is accepted rather than papered over. The README now says plainly that nothing in the repo checks this rendering, and that anyone editing either driver or the `SETTINGS` clause must exercise the failure cases by hand before a window. The matrix still exists and still backs the results on this PR — hardened per the two findings, 22/22 across both drivers, with four controls proving it can fail — it just lives outside the repo where an operator script cannot be confused for it.

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
```diff
@@ -372,14 +372,27 @@ independent controls keep each statement safe:
...
   time a real window at your intended value and feed it back via `--rows-per-sec`.
   Full diagnosis in `backfill.sh`'s `--max-insert-threads` option docs.
 
+  **If you edit the rendering, re-validate it by hand — nothing in this repo checks it for you.** The drivers
+  render this setting by requiring exactly one line-anchored `max_insert_threads = ${MAX_INSERT_THREADS},` in
+  `000001`/`000002` and then either stripping it (inherit) or substituting it. Every way that can go wrong is
+  silent: a reformatted or shared line, a missing assignment, a duplicate one, or a placeholder that survives
+  into an executable line. The drivers' own guards abort on each of those, but there is deliberately **no
+  committed harness** here — this directory holds operator drivers only — so after changing either driver or the
+  `SETTINGS` clause of `000001`/`000002`, exercise those cases manually against corrupted copies before a
```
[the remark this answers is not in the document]
thiagohora: **Real — and I want to correct the record on how it was found.** The premise was initially challenged on my side because a grep for "two-pass"/"split the delta"/"by hand" over the README did not surface it. That grep was wrong, not the finding: the runbook says *"split it into two batched passes"*, and the instruction lives in two places.

`README.md:418-421` and `000002:46`:

> `-- CARRY THE FULL SETTINGS BLOCK BELOW ONTO BOTH PASSES, max_partitions_per_insert_block included. The driver does not implement this split, so these two statements are hand-written`

So yes — that block now contains `max_insert_threads = ${MAX_INSERT_THREADS}`, hand-written statements bypass the driver entirely (nothing substitutes placeholders, no guard applies), and following the instruction literally sends a literal placeholder. **This hazard is one my own PR chain introduced**: before the setting was added, that instruction was safe.

Fixed in `7aae2dc`. I verified each half by rendering `000002` through `clickhouse-format` rather than reasoning about it:

| variant | result |
|---|---|
| placeholder copied verbatim | `Code: 62` syntax error — hazard confirmed |
| whole `max_insert_threads` line deleted | **parses** — so your recommended inherit procedure is sound |
| value removed, trailing comma left | `Code: 62` — so the comma caveat is warranted, not decorative |

Both places now separate *carrying the settings* from *carrying the placeholders*, and call out `max_insert_threads` specifically as **the one with no substitutable default**: unset means inherit, which the driver expresses by *removing the line*, and `0` is not equivalent — it forces serial execution and overrides the server. Either the same concrete count on both passes, or delete that one line, comma and all, keeping `max_partitions_per_insert_block` and `log_comment`.

You also asked for a dry-run check, which was the right instinct given no guard covers this path. Added to both: `grep -n '${' <your-statements>.sql` must print nothing. Confirmed it catches *incomplete* substitution, not just total omission.

**On the third sub-claim — "the integration test hard-codes a different SQL shape" — no change needed, and it is unrelated to the split.** `TracesLocalV2CutoverTest` does not implement the split: zero occurrences of the arm (b) predicate, it runs the single-statement delta. So no hand-written split procedure can disagree with it. Its `SETTINGS` do differ from `000002`, in exactly the way its own class Javadoc documents since `96fc611` — partition bound mirrored because it is a correctness gate, `max_insert_threads` omitted because it is pacing and omission is meaningful, block size pinned only for predictable CI memory.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
```diff
@@ -326,26 +326,36 @@ run_backfill_window() {
...
     sql="${sql//'${MAX_INSERT_BLOCK_SIZE}'/$MAX_INSERT_BLOCK_SIZE}"
     sql="${sql//'${MAX_PARTITIONS_PER_INSERT_BLOCK}'/$MAX_PARTITIONS_PER_INSERT_BLOCK}"
+    # >>> BEGIN max_insert_threads rendering (fence for extracting this block to test edits -- keep the markers)
+    # This depends on ONE exact line in ANOTHER file:  max_insert_threads = ${MAX_INSERT_THREADS},
+    # Require exactly one line-anchored, ISOLATED assignment before touching anything. A bare token match would
+    # also accept the placeholder inside a `--` header, a /* */ block comment or a string literal, and let the run
+    # proceed with no setting rendered at all. Anchoring also catches the two layout hazards in one check: a
+    # shared line (which cannot be removed without dropping its neighbours) and a reformatted one (which the
+    # fixed-string strip would silently miss).
+    assignments="$(grep -cE '^[[:space:]]*max_insert_threads = \$\{MAX_INSERT_THREADS\},?[[:space:]]*$' <<<"$sql" || true)"
```
JetoPistola: 💡 **suggestion** | Logic

The PR body gives this as the reason for *not* tightening the regex:

> the regex was deliberately **not** tightened, because `max_insert_threads` as the last setting in the clause is legitimate SQL and would then be rejected for no safety gain.

That case is rejected today. The optional `,?` accommodates a trailing comma but not a terminating semicolon, so the natural spelling of the last setting fails the pre-check:

```console
$ for l in 'max_insert_threads = ${MAX_INSERT_THREADS};' \
           'max_insert_threads = ${MAX_INSERT_THREADS}' \
           'max_insert_threads = ${MAX_INSERT_THREADS},'; do
    printf '  %s\n' "$l" | grep -cE '^[[:space:]]*max_insert_threads = \$\{MAX_INSERT_THREADS\},?[[:space:]]*$'
  done
0    # last setting, semicolon on the same line -> ABORT
1    # only if the semicolon sits on its own line
1    # trailing comma
```

So the exact layout the rationale sets out to protect is the one that aborts. It fails loudly rather than rendering wrong, so this is not another silent hole — but if the `SETTINGS` clause of `000001`/`000002` is ever reordered so this setting lands last, a cutover aborts for no safety reason, and whoever hits it will read the PR rationale and conclude the guard is broken.

Two ways to close the gap, both cheap:

```suggestion
    assignments="$(grep -cE '^[[:space:]]*max_insert_threads = \$\{MAX_INSERT_THREADS\}[,;]?[[:space:]]*$' <<<"$sql" || true)"
```

Note that accepting `;` here means the inherit-path strip (which uses the same pattern) would delete the statement terminator along with the line — so if you take this route, the strip needs to keep the `;`. Given that, the simpler option may be to leave the regex exactly as it is and correct the rationale instead: the guard requires this setting to be non-final, which is a real constraint worth stating in the error message rather than an accident.

Verified by extracting the rendering block verbatim from both drivers and running it against the committed SQL plus corrupted variants; every other case in the PR's matrix reproduced as described.

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
```diff
@@ -372,14 +372,27 @@ independent controls keep each statement safe:
...
+  cutover window. Each driver fences its rendering block with `>>> BEGIN max_insert_threads rendering` /
+  `<<< END` so it can be extracted verbatim rather than reimplemented; run whatever you extract it into from
+  inside `scripts/`, since the block resolves `000001`/`000002` by relative path and from elsewhere reads
```
JetoPistola: 🧹 **nit** | Documentation

> run whatever you extract it into from inside `scripts/`, since the block resolves `000001`/`000002` by relative path and from elsewhere reads nothing and fails every case for the wrong reason.

The fenced block does not resolve or read either SQL file. It operates entirely on the `$sql` variable its caller has already populated; `$BACKFILL_SQL` / `$SQL_FILE` appear only inside the two `echo` diagnostics. And in the drivers themselves the path is CWD-independent regardless:

```bash
SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
```

I extracted both blocks and ran the full matrix from `/tmp` — all cases behaved as documented, including the aborts.

This matters a little more than a normal doc nit, because it is advice given to an operator mid-cutover about why their validation might be lying to them. Sending someone hunting for a CWD problem that cannot exist costs time in the one window where it is expensive. Suggest dropping the clause; the caller-supplies-`$sql` contract is the thing actually worth stating:

```suggestion
  `<<< END` so it can be extracted verbatim rather than reimplemented; the block reads no files — it operates on a
  `$sql` variable the caller must populate (and names the SQL path only in its diagnostics), so whatever you
  extract it into has to read `000001`/`000002` itself.
```

JetoPistola:
👋 **Review summary**

**What looks good**

- **The three fixes hold under reproduction.** I extracted each driver's fenced rendering block verbatim and ran it against the committed SQL plus corrupted variants. `inherit` renders no key, explicit values render, and shared-line / reformatted / missing / duplicated assignments all abort with `exit 2` and an `ERROR:` diagnostic. Both files carry exactly one anchored assignment today.
- **The SIGPIPE fix is load-bearing, not defensive dead code.** Worth recording, since the new pre-check catches every case in the stated matrix *before* the post-condition runs — which made me suspect the post-condition was now unreachable. It isn't: a second placeholder on an executable line that is not an isolated assignment (e.g. inside `log_comment = 'threads=${MAX_INSERT_THREADS}'`) passes the pre-check and is caught only by the post-condition. With the piped form restored, that case leaks. So the pipe removal is doing real work on a real path.
- **The zsh-vs-bash note in the description is the most valuable thing here.** A finding that a shell test *disproved* — and was only real under the drivers' actual `#!/usr/bin/env bash` — is exactly the kind of near-miss that usually goes unrecorded.
- **The withdrawal of `verify_render.sh` is the right call and cleanly done.** No stale references survive anywhere in the directory, and the README states plainly that nothing in the repo checks the rendering rather than implying coverage that does not exist. Reversing a committed decision after review found two defects in the harness itself is the harder path.
- **Input validation closes the substitution side.** `^(0|[1-9][0-9]?)$` in both drivers means a hostile `--max-insert-threads` value cannot reach the `SETTINGS` clause — I tried, and it is rejected at parse time. `bash -n` clean on both. CRLF also renders correctly, since `\r` falls inside `[[:space:]]`.
- The false `300/150` defaults claim is gone rather than restated with different numbers, which is the correct fix for a value that moves across versions.

**Overall**

The approach is sound and the standard of evidence is well above the bar for a directory that gets deleted at the end of the cutover. Each hole is on a default path, each was reproduced before being fixed, and the fences make the block extractable verbatim so a future validator cannot drift from the code it checks.

**On the declined shared helper** — you asked for a second opinion rather than treating it as settled, so: I'd keep the duplication. The two blocks are byte-identical apart from indentation and the path variable, so the drift risk is real, but the fence markers already make drift *detectable*, which is most of the value a helper would add. Against that, `_common.sh` would be the first shared file among nine standalone operator scripts, and it converts a missing-sibling into a hard failure precisely during the window where recovery is most expensive. Your reasoning holds.

**Inline comments**: 1 suggestion, 1 nit — nothing blocking.

- The suggestion is a mismatch between the regex and the PR's stated rationale for leaving it loose: `max_insert_threads` as the **last** setting is rejected when the semicolon is on the same line, which is the very layout the rationale sets out to protect. It aborts loudly rather than rendering wrong, so it is not a fourth silent hole — but the design intent as written is not met.
- The nit is a README claim that the extracted block resolves the SQL files by relative path. It reads no files at all; I ran the matrix from `/tmp` and every case behaved as documented.

This review does not constitute an approval — a human reviewer should still sign off on the diff.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
thiagohora: You are right, my rationale was false, and testing it properly found something worse than either of us had named. Fixed in `fe65be66`.

**The rationale was wrong exactly as you say.** It claimed tightening would *newly* reject `max_insert_threads` as the last setting. That layout was already rejected: `,?` admits a comma or nothing but not a semicolon, so `max_insert_threads = ${MAX_INSERT_THREADS};` yields zero matches and aborts. Worse, the test I used to "prove" tightening unnecessary put the semicolon on its own line — the artificial layout — so it never exercised the natural spelling at all.

Four layouts, each rendered in both modes and put through `clickhouse-format`:

| layout | baseline SQL | `,?` inherit | `,?` explicit |
|---|---|---|---|
| mid-clause, trailing comma | parses | parses | parses |
| **last, bare, `;` next line** | **parses** | **INVALID** | parses |
| last, `;` same line | parses | ABORT | ABORT |
| last, comma, `;` next line | INVALID | INVALID | INVALID |

**Row two is the finding under your finding.** Valid SQL, accepted by the pre-check, and inherit strips the line to leave a dangling comma and no terminator — broken SQL sent to the server rather than an abort. That is precisely the failure class these guards exist to prevent, and it was live.

**Your suggested widening does not survive testing, so I have not applied it** — your own instinct to flag it for verification was the right one. Allowing a semicolon as well as a comma additionally accepts row three, which then strips to invalid SQL for the same reason as row two, converting a correct loud abort into silent corruption:

- with the widened pattern, `last, ';' same line` → `hits=1`, inherit renders **INVALID**
- today, that same layout → `hits=0`, **ABORT**

**So the comma is now required**, in the pre-check and the strip alike. Both terminator-carrying layouts abort in both modes, and the accepted form is exactly the one the template uses. Row two goes from silent corruption to a loud abort; row three is unchanged.

The constraint that follows is real, so it is stated rather than hidden: the assignment must not be the last entry in the `SETTINGS` clause. The README says it and the error message says it, with the reason. Row four still renders invalid SQL, but its baseline is already invalid — a dangling comma before a bare terminator — so the file is broken before rendering and no pre-check on this one line can repair it.

The PR body rationale has been corrected too; it was stating the false reason.

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
thiagohora: Correct on every point, and the claim was mine. Fixed in `fe65be66`.

Verified before changing it: the fenced block contains no file read at all. It operates on a `$sql` variable the caller populates via here-strings, and `$BACKFILL_SQL` / `$SQL_FILE` appear **only** inside its two `echo` diagnostics. And both drivers are CWD-independent exactly as you show:

```
backfill.sh:157      SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
delta_replay.sh:45   SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
```

**Where the error came from, since it is the more useful part.** I did hit a real failure running the withdrawn `verify_render.sh` from `/tmp` — but the `cd` responsible belonged to that harness, not to the block. I misattributed it, then wrote operator-facing advice from the misdiagnosis. Your framing of the cost is right: it would send someone hunting a CWD problem that cannot exist, in the window where that costs most, and the two of us extracting the same block from different directories would have disagreed about why.

Taken your suggestion and stated the contract rather than the symptom: the block reads no files, so whatever you extract it into must load the SQL itself and populate `$sql`; it is otherwise CWD-independent, as are the drivers. That is the thing a person actually needs to know before extracting it.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
```diff
@@ -326,26 +326,39 @@ run_backfill_window() {
...
     sql="${sql//'${MAX_INSERT_BLOCK_SIZE}'/$MAX_INSERT_BLOCK_SIZE}"
     sql="${sql//'${MAX_PARTITIONS_PER_INSERT_BLOCK}'/$MAX_PARTITIONS_PER_INSERT_BLOCK}"
+    # >>> BEGIN max_insert_threads rendering (fence for extracting this block to test edits -- keep the markers)
+    # This depends on ONE exact line in ANOTHER file:  max_insert_threads = ${MAX_INSERT_THREADS},
+    # Require exactly one line-anchored, ISOLATED assignment before touching anything. A bare token match would
+    # also accept the placeholder inside a `--` header, a /* */ block comment or a string literal, and let the run
+    # proceed with no setting rendered at all. Anchoring also catches the two layout hazards in one check: a
+    # shared line (which cannot be removed without dropping its neighbours) and a reformatted one (which the
+    # fixed-string strip would silently miss).
+    assignments="$(grep -cE '^[[:space:]]*max_insert_threads = \$\{MAX_INSERT_THREADS\},[[:space:]]*$' <<<"$sql" || true)"
```
miguelgrc: **[medium] A `/* */` block comment still satisfies the isolated-assignment check, so an explicit value renders into a comment**

Re-checked against `fe65be66f5` — the trailing-comma fix doesn't close this one, because a block-commented assignment has a trailing comma too.

The comment directly above (line 332) says a bare token match "would also accept the placeholder inside a `--` header, a /* */ block comment or a string literal", implying the anchored match rules those out. It rules out only `--`, and only because `--` breaks the anchor. There's no notion of executability here.

Extracting this fence verbatim from the current head and running it against:

```sql
SETTINGS
    max_insert_block_size = 100000,
/*
    max_insert_threads = ${MAX_INSERT_THREADS},
*/
    max_partitions_per_insert_block = 2000,
    log_comment = 'backfill';
```

with `--max-insert-threads 3` gives **exit 0, no ERROR**, and renders `max_insert_threads = 3,` *inside the comment block*. The count check sees exactly one match, the else-branch substitutes, and the post-condition finds no surviving placeholder because it filters `--` lines only. The operator asked for 3 threads and the server gets no setting at all — the silent failure hole #3 is meant to close.

No `/* */` exists in either SQL file today (`grep -c '/\*'` is 0 across `db-app-analytics/`), so this is only reachable after a hand edit — but edit-safety is the guard's whole purpose, and the fence markers exist precisely because this block is expected to be edited.

See also the companion comment on the post-condition at line 355: the two halves disagree about what a `/* */` block is, in opposite directions. Either both should strip block comments, or the comment at 331-335 should say plainly that only `--` is handled.

_general finding_

miguelgrc (commented):

4 findings — 0 high · 2 medium · 2 low. Suppressed by team conventions: see `suppressed.md`.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
```diff
@@ -326,26 +326,39 @@ run_backfill_window() {
...
+        # Unset means INHERIT: drop the line so the server's own value applies. Rendering an explicit 0 would
+        # OVERRIDE it and force the insert serial -- a silent slowdown, not a no-op.
+        sql="$(grep -vE '^[[:space:]]*max_insert_threads = \$\{MAX_INSERT_THREADS\},[[:space:]]*$' <<<"$sql")"
     else
         sql="${sql//'${MAX_INSERT_THREADS}'/$MAX_INSERT_THREADS}"
     fi
+    # Post-condition for BOTH paths. `--` lines are skipped: both SQL headers legitimately quote the placeholder.
+    # Note the deliberate absence of a pipe into `grep -q` -- it exits on first match, the upstream grep takes
+    # SIGPIPE, and `set -o pipefail` then reports 141, so the guard would skip its own failure branch.
+    executable="$(grep -v '^[[:space:]]*--' <<<"$sql" || true)"
```
miguelgrc: **[low] Post-condition uses the bare-token match the block above warns against, so a `/* */` mention false-aborts**

Mirror of the comment at line 336, same root cause, opposite symptom.

Lines 331-335 justify the anchored count precisely because a bare token match would accept the placeholder "inside a `--` header, a /* */ block comment or a string literal". This post-condition **is** that bare token match: it strips `--` lines, then `grep -qF '${MAX_INSERT_THREADS}'`.

Reproduced on the current head — a real isolated assignment line plus the placeholder merely *mentioned* in a doc block:

```sql
/*
  Rendering note: the driver substitutes ${MAX_INSERT_THREADS} or strips the line.
*/
...
    max_insert_threads = ${MAX_INSERT_THREADS},
```

Inherit path (`--max-insert-threads` unset) strips the assignment correctly, then aborts with `ERROR: ${MAX_INSERT_THREADS} survives in an executable line` — exit 2, on a render that actually succeeded. For `delta_replay.sh` (line 119, identical) that abort lands on the final delta, immediately before the EXCHANGE.

So a `/* */` block is treated as executable by the pre-check (accepted, shouldn't be) and as executable by the post-condition (rejected, shouldn't be). Worth picking one definition and applying it in both places.

_general finding_

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
```diff
@@ -326,26 +326,39 @@ run_backfill_window() {
     sql="${sql//'${WINDOW_HI}'/$hi}"
     sql="${sql//'${MAX_INSERT_BLOCK_SIZE}'/$MAX_INSERT_BLOCK_SIZE}"
     sql="${sql//'${MAX_PARTITIONS_PER_INSERT_BLOCK}'/$MAX_PARTITIONS_PER_INSERT_BLOCK}"
+    # >>> BEGIN max_insert_threads rendering (fence for extracting this block to test edits -- keep the markers)
```
miguelgrc, line 396: **[medium] `--dry-run` never executes this block, and in a real run it fires only after the anchor is minted**

This guard sits inside `run_backfill_window()`, whose only call site is line 393 — and `insert_window()` returns at 387-390 whenever `DRY_RUN=1`. Two consequences:

**1. `--dry-run` validates none of this.** A reformatted, shared, duplicated or missing `max_insert_threads = ${MAX_INSERT_THREADS},` line in `000001` passes a full dry-run clean and aborts only on the first real window. That's the mode README:388 sends operators to after they change "either driver or the `SETTINGS` clause of `000001`/`000002`", so the rehearsal that's supposed to build confidence exercises exactly zero of the rendering.

**2. In a real run the abort is late.** It lands after the successor-table check, after `preflight_capacity()`, and after the `backfill_start` anchor has been minted and persisted (~462-491). That's the same cost this file argues against at 227-230 when justifying eager validation of `--max-partitions-per-insert-block`: "a far more expensive place to discover a typo".

The check depends only on the SQL file's text — not on the window, the anchor, or the cluster — so it can run at startup. `delta_replay.sh` already does exactly that: the byte-identical block runs at top level (93-122), before any statement. `backfill.sh` is the outlier here.

_general finding_

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
```diff
@@ -123,10 +123,10 @@
...
-#                                 assume ClickHouse's defaults (300 / 150). Deployments routinely raise
-#                                 them, so the real headroom can be an order of magnitude off either way,
-#                                 and "parts vs 300" is a misleading ratio on a tuned cluster.
+#                                 and parts_to_delay_insert, read from system.merge_tree_settings. Do NOT
+#                                 work from a remembered default: ClickHouse has changed these across
+#                                 versions, and a deployment may tune them further, so a hardcoded ratio can
+#                                 be an order of magnitude wrong in either direction.
 #
 #                             CHOOSING A VALUE IS A CAPACITY DECISION, NOT A BENCHMARK. On an idle
 #                             rehearsal environment a large value looks free; on a production cluster
```
miguelgrc, line 132: **[low] Just below (135-137): "0 stays the default" contradicts the inherit semantics this driver implements**

Anchored here because it's the nearest line in the diff — the text is at **135-137**, the tail of this same option docblock:

> "0 stays the default so no existing deployment changes behaviour, and so a small install on few cores does not silently start spawning insert threads."

The default is `MAX_INSERT_THREADS=""` (line 180), i.e. omitted, and omitted means inherit — the line is stripped, not rendered as 0. Lines 63-67 of this same block state the opposite in bold: "OMITTED BY DEFAULT, and omitted means INHERIT ... rendering an explicit 0 would OVERRIDE a profile that sets it and force the insert serial."

An operator who reads 135-137 concludes `0` is the status quo and passes `--max-insert-threads 0` explicitly to "keep current behaviour" — which overrides a ClickHouse Cloud or self-managed profile and forces the INSERT SELECT serial. That's the exact silent slowdown 63-67 warns about, reached by following the docs.

Looks like a leftover from before the inherit semantics landed; `000001:30-31`, `delta_replay.sh:37-38` and README:423-425 all already agree with the new behaviour.

_general finding_

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
thiagohora: Confirmed and fixed in `70b1a90e`. You were also right that `fe65be66` does not close it — a block-commented assignment carries a trailing comma too, so the comma fix was orthogonal.

Built the fixture, watched the old code accept it, watched the new code reject it. Assignment present **only** inside a multi-line `/* */` block, in the exact accepted form:

| mode | before | after |
|---|---|---|
| explicit | renders, **zero executable settings** — the value went into the comment | ABORT |
| inherit | renders, having stripped a line inside a comment | ABORT |

The explicit row is the genuinely silent failure on this PR: no error anywhere, and the setting simply does not apply. Worth noting the earlier matrix case named "placeholder only in `/* */` aborts" **passed for the wrong reason** — its fixture was a single-line `/* … */`, which never matched the anchor, so it aborted on count=0 rather than on any notion of executability. That fixture is replaced.

The fix is a comment-masking helper — `--` to end of line, `/* */` across lines, line numbering preserved — used by the count, the strip/substitute (which now targets the matched line by number) and the post-condition. So "is this the assignment?" now means "is it executable?" rather than "does this text appear?".

**And you are right about the comment.** It claimed the anchored match ruled out "a `--` header, a `/* */` block comment or a string literal". It ruled out only `--`, and only incidentally, because `--` breaks the anchor. Corrected to say what the check does and that masking is what makes executability part of it — the same overstatement was in the PR body and is fixed there too.

Control: reverting the masking in the count makes both block-comment cases fail, reproducing the silent drop (`got[OK max_insert_threads = 3] want[ABORT]`).

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
thiagohora: Confirmed and fixed in `70b1a90e` — and it is the same fix from the other side, as you suspected.

The post-condition matched a bare token outside `--` lines only, so a file with a **correct** assignment plus a harmless `/* */` mention of the placeholder was refused:

| mode | before | after |
|---|---|---|
| inherit | **ABORT** (the file was fine) | renders |
| explicit | rendered | renders |

Both ends now use the same comment-masking helper, so the count and the post-condition agree by construction rather than by coincidence. That is why it is one helper and not two checks that happen to be written similarly.

Control: reverting just the post-condition to the unmasked form makes both comment-mention cases fail (`got[ABORT] want[OK …]`), while leaving the block-comment cases passing — so the two halves are independently covered.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
thiagohora: Confirmed on both halves and fixed in `70b1a90e`.

Traced the call path rather than taking it on trust: the block sat in `run_backfill_window()` (line 321), which `insert_window()` reaches at line 393 — after its `DRY_RUN` early return at 387-390. So the guard was unreachable in a dry run.

End-to-end, with a block-commented assignment in `000001`:

```
before: ./backfill.sh --dry-run  -> not caught (just a connection error later)
after:  ./backfill.sh --dry-run  -> caught at startup, naming the file and the count
after:  ./backfill.sh --dry-run, good file -> silent
```

That last line is the control: the check fires on the malformed file and not on the good one.

Your point about README:388 is the one that made this worth fixing rather than noting — that line is exactly what sends an operator to dry-run after editing the `SETTINGS` clause, so the rehearsal we recommend exercised none of the rendering. And you are right that the file already argues this at 227-230 for `--max-partitions-per-insert-block`; the check is now hoisted beside the SQL-file existence check, on the same reasoning. `delta_replay.sh` was the correct model and needed no change to its placement.

**Whether the hoist changed anything beyond timing: no, and I checked rather than assumed.** The startup check reads the file's text, which does not change between windows, so it covers every window. The per-window path still resolves the assignment's line — the render needs its position — and keeps its own post-condition, so a file edited *mid-run* is still caught rather than rendered. The two are not redundant: the startup one gates before the anchor is minted, the per-window one covers mid-run edits.

One real bug of mine surfaced doing this and is worth recording: the hoisted call initially sat **above** the helper definitions, so it died with `mit_require_one_assignment: command not found`. The dry-run test caught it immediately; the helpers now precede it.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
thiagohora: Correct, and fixed in `70b1a90e`. "0 stays the default so no existing deployment changes behaviour" predates `96fc611` and now says close to the opposite of what the driver does: omission is the default and means inherit, and an explicit `0` is **not** equivalent — it forces no parallel execution and overrides whatever the server sets, which is the behaviour that commit exists to prevent from happening implicitly.

Checked for siblings as you suggested: that was the only survivor. `grep` for the phrasing across both drivers, the SQL files, the README and the PR body returns nothing else — the README and PR body were updated when the inherit semantics landed, and this option-doc paragraph was missed.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
```diff
@@ -326,26 +380,38 @@ run_backfill_window() {
...
+    # The SETTINGS line this depends on lives in ANOTHER file, so it is validated rather than assumed. The full
+    # validation runs once at startup (see the mit_require_one_assignment call there), which is what makes it
+    # exercised by --dry-run and reached before backfill_start is minted. Here it is re-resolved per window
+    # because the render needs the line's position, and the file is re-read each window -- so an edit made
+    # mid-run is still caught rather than rendered.
+    #
+    # Comments are masked before matching. A line-anchored match is NOT by itself a check that the assignment is
+    # executable: an identical line inside a /* */ block carries a trailing comma too, and would otherwise be
+    # treated as the assignment.
+    mit_line="$(mit_assignment_lines "$sql")"
```
miguelgrc: **[medium] `set -e` kills the script here when the count is 0, so the error branch below is unreachable and the abort is silent**

The masking rewrite closes all four of the previous comments — verified each against `70b1a90e2a`, including running the real script to confirm `--dry-run` now aborts at startup before any ClickHouse contact. This is a new one the rewrite introduced, in both drivers.

`mit_assignment_lines` ends in a pipeline whose `grep -nE` exits 1 when there are no matches. Under `set -euo pipefail` that makes this assignment fail, and `set -e` terminates the script **before** the `if` on the next line ever runs:

```bash
mit_line="$(mit_assignment_lines "$sql")"        # <- dies here when count == 0
if [[ "$(grep -c . <<<"$mit_line" || true)" -ne 1 ]]; then   # unreachable for the 0 case
    mit_require_one_assignment "$sql" "$BACKFILL_SQL" || exit 2
```

Observed, extracting the fence verbatim from this commit and feeding it SQL whose assignment sits inside a `/* */` block: **exit 1, no output at all**, where exit 2 plus the diagnostic was intended. The `>= 2` case is fine — `grep -nE` succeeds there, so the count check runs and reports properly. Only the zero case dies mute.

That is exactly the path the comment above advertises: *"the file is re-read each window -- so an edit made mid-run is still caught rather than rendered."* It is caught, but an operator who edits the SQL mid-cutover gets a bare `exit 1` with nothing to go on, instead of the eight-line diagnostic this commit wrote for them.

`delta_replay.sh:144` is the identical statement with the same exposure.

`|| true` restores the intended behaviour — confirmed on the same fixture, it flips silent exit 1 to exit 2 with the full message:

```bash
mit_line="$(mit_assignment_lines "$sql" || true)"
```

---

**Minor, same block.** `mit_mask_comments` treats `/*` inside a *string literal* as opening a block comment, latching `inblk` and blanking the rest of the file. A valid file containing e.g. `log_comment = 'phase /* one'` then reports `found 0` — telling the operator the assignment is missing when it is present and correct. No such literal exists in either SQL today, and the comments here no longer claim to handle string literals (the previous ones did), so this is a latent robustness gap rather than a live bug — noting it only because the failure mode is a confidently wrong diagnostic.

_general finding_

miguelgrc (commented):

1 findings — 0 high · 1 medium · 0 low. Suppressed by team conventions: see `suppressed.md`.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/backfill.sh
thiagohora: Confirmed and fixed in `ddb75df4`. My regression, and your diagnosis is exact.

I initially failed to reproduce it — my first attempt wrapped the fence in a captured subshell and reported `rc=2` with the diagnostic, which was wrong. A minimal probe against the real function settled it:

```
mit_assignment_lines "$sql"        -> rc=1        (grep finds nothing; pipefail propagates it)
set -e; mit_line="$(...)"          -> shell dies here; the `if` below never runs
```

Assertion, old versus new:

| case | before | after |
|---|---|---|
| zero matches | `rc=1`, **0 diagnostic lines** | `rc=2`, diagnostic printed |
| two matches | `rc=2`, diagnostic | `rc=2`, diagnostic |
| good file | `rc=0`, silent | `rc=0`, silent |

Your note that `>=2` is unaffected is confirmed — grep succeeds there, so the count check runs normally.

**It was wider than the finding says.** Every zero-match case had it, not just the block-commented one: a reformatted line, a shared line, a missing assignment and a block-commented assignment all produce zero matches, so all four died mute in both modes — **seven** matrix cases once the harness could see them.

And your point about the irony landed: the comment directly above promises that re-reading the file each window means a mid-run edit "is still caught rather than rendered". It *was* caught — and handed the operator a bare `exit 1`, mid-cutover, from the commit written to give them the diagnostic. The `|| true` now carries a comment saying it is load-bearing, so it is not tidied away later.

**Two defects in my own testing, both the same family as the earlier ones:**

1. The matrix asserted only that a case "aborted". A mute `set -e` death is indistinguishable from a working guard under that assertion — which is exactly how this passed. Abort verdicts now carry the exit code and whether a diagnostic was emitted: `ABORT(rc=2,diag)` versus `ABORT(rc=1,MUTE)`.
2. Worse: the harness ran the extracted block under `set -uo pipefail`, **without `-e`**, while the drivers set `-euo`. So even with the stronger assertion it could not see the bug. It now matches the drivers exactly.

With both fixed, the control works: removing the `|| true` again makes **seven cases fail as `ABORT(rc=1,MUTE)`**. So the class is covered, not just this instance.

**Audited the rest of the rewrite for the same pattern.** The other two command substitutions around pipelines already carry `|| true` (the count in `mit_require_one_assignment`, the masked text in the post-condition). The `sed` calls are reached only after the count is confirmed to be exactly one, so `$mit_line` is a valid line number. The two `awk BEGIN` substitutions predate this work and cannot return 1.

Thanks also for re-verifying the previous four against `70b1a90e2a`, including running the real script to confirm `--dry-run` aborts before any ClickHouse contact. I re-ran that after this change and it still does.

miguelgrc (dismissed):
Thanks for addressing the comments ;)

miguelgrc (approved):

PR #7887, opened 2026-08-18
[OPIK-7105] [FE] refactor: decouple Cost Intelligence from workspace-version machinery

awkoy:
## Details

Cost Intelligence access now depends only on the two conditions that actually govern it: the organization has cost intelligence enabled, and a spend workspace exists. The workspace-version conjunct is removed. **Nothing is enabled or disabled** — the set of organizations with Cost Intelligence is identical before and after.

- Both removed terms were provably constant. Every writer of the workspace version writes the literal `"v2"` (`resolveSyncWorkspaceVersion` returns a constant, `WorkspaceVersionResolver` assigns a literal, `fetchWorkspaceVersion` returns a literal with no network call), so `hasAccess`'s `&& workspaceVersion === "v2"` was always true and `isPending`'s `|| !workspaceVersion` always false. `A && B && true === A && B`; `X || false === X`.
- The value is ordering, not cleanup. While this reader existed, a later removal of the version machinery would have left the store field at its initial `null`, made the comparison false, and collapsed `hasAccess` for every enterprise cloud organization — with no error, no log and no failing test. With the reader gone that failure mode is structurally impossible. The follow-up PR that removes the writers depends on this one landing first.
- Also removes the "New Opik experience" user-menu toggle: its condition `detectedWorkspaceVersion === "v1"` can never be true for the same reason, so the item was unreachable. Removing it is what makes the verification gate below meaningful.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- OPIK-7105

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: codebase analysis, the change itself, and this description
- Human verification: constant-folding argument checked against every writer of the store field; verification commands below run and their output inspected; sequencing relative to the follow-up PR reviewed and agreed before implementation

## Testing

```
npm --prefix apps/opik-frontend run typecheck      # exit 0
npm --prefix apps/opik-frontend run lint           # exit 0
npx vitest run src/lib/aiSpend.test.ts             # 2/2 pass
npx vitest run src/plugins/comet/                  # see note
grep -rn "workspaceVersion\|WorkspaceVersion" src/plugins/   # no matches
```

- Gate: no workspace-version reference remains anywhere under `src/plugins`.
- `src/plugins/comet/` suite reports 11 failures / 58 passes. Verified identical on `main` by stashing and re-running — a pre-existing local environment fault (`localStorage.clear is not a function`, happy-dom API mismatch), not introduced here.
- Not run: manual Cost Intelligence smoke on a Comet build. Not applicable to this PR — the change is behaviour-preserving by construction — but note for reviewers of the follow-up that local dev cannot demonstrate this path at all: `plugins/development/LayoutProvider.tsx` hardcodes `hasAccess: true`, so `npm run dev` shows a working AI Spend even on broken code.

No unit test is added: the dependency is removed rather than made correct, so there is nothing left for a regression test to protect.

## Documentation

None required. No user-facing behaviour, API or configuration changes.

### DIFF
--- a/apps/opik-frontend/src/plugins/comet/UserMenu.tsx
+++ b/apps/opik-frontend/src/plugins/comet/UserMenu.tsx
@@ -8,13 +8,11 @@ import {
   Settings,
   Settings2,
   Shield,
-  Sparkles,
   UserPlus,
 } from "lucide-react";
 import TooltipWrapper from "@/shared/TooltipWrapper/TooltipWrapper";
 import SupportHubSubMenu from "@/shared/SupportHub/SupportHubSubMenu";
 import { Avatar, AvatarFallback, AvatarImage } from "@/ui/avatar";
-import { Switch } from "@/ui/switch";
 import {
   DropdownMenu,
   DropdownMenuContent,
@@ -32,16 +30,7 @@ import { useThemeOptions } from "@/hooks/useThemeOptions";
 import { APP_VERSION } from "@/constants/app";
 import { ADMIN_DASHBOARD_LABEL } from "@/constants/labels";
 import { cn, maskAPIKey } from "@/lib/utils";
-import useAppStore, {
-  useDetectedWorkspaceVersion,
-  useOpikWorkspaceName,
-} from "@/store/AppStore";
-import {
-  getNewExperienceOptIn,
-  getVersionOverride,
-  navigateToWorkspaceRoot,
-  setNewExperienceOptIn,
-} from "@/lib/workspaceVersion";
+import useAppStore, { useOpikWorkspaceName } from "@/store/AppStore";
 import api from "./api";
 import { ORGANIZATION_ROLE_TYPE } from "./types";
 import useOrganizations from "./useOrganizations";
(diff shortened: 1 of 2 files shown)


CometActions:
**No test needed here.**

Both changed files live in `src/plugins/comet/`, and `PluginsStore.setupPlugins` resolves the plugin folder from the Vite mode — so an OSS install never loads this UserMenu at all, and `useAiSpendManager` additionally needs a Comet cloud org with `costIntelligenceEnabled` and a spend workspace. Dropping the `workspaceVersion === "v2"` gate is a real behaviour change (Cost Intelligence will now show on v1 workspaces), but it is only reachable on a cloud org, which our e2e estate cannot reach — so this is a manual check on a v1 workspace, not an automated one.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/32142202493)

CometActions:

### Access Information
- **URL:** https://pr-7887.dev.comet.com
- **Namespace:** pr-7887
- **Version:** 2.2.30-7887-merge-2990
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7887%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

andriidudar (approved):
Looks great!

PR #7906, opened 2026-08-19
[OPIK-6901] docs(cutover): correct a stale claim that traces has no id skip index

thiagohora:
## Details

`000002`'s deletion-replay note says `traces` has **no id skip index**, that `000088` indexes only `created_at`/`last_updated_at`, and that id minmax/bloom indexes exist **only** on `traces_local_v2`.

That was true when written. It is now wrong: **`000113` added `idx_traces_id_bf`, a `bloom_filter(0.01)` on `traces.id`** for OPIK-7483's delete-by-id project resolution. Confirmed against a live deployment's `system.data_skipping_indices`, not inferred from the migration list.

## Why a comment is worth a PR

**It is load-bearing.** It is the stated reason the resurrection guard's source lookup is cheap — *"a bounded id-filtered read of that tiny set, not a value-indexed prune of the full `traces` table"*. The conclusion still holds, but for a better reason: with the bloom filter present **and** the primary key already `(workspace_id, project_id, id)`, that lookup prunes on both.

**And it actively misled an investigation.** During the OPIK-6901 prod-test rehearsal, a verify window reported a difference and its artifact-versus-real verdict query proved expensive. This comment was quoted as the explanation — a missing id index on the source — and the diagnosis was wrong, because the index exists. The real cost driver is the candidate key **count and scatter under `FINAL`**, where every version of every candidate key must be read. The corrected note states that, so the next reader doesn't repeat the wrong turn.

## Change checklist
- [ ] User facing
- [x] Documentation update

## Issues

- Resolves #
- OPIK-6901

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: Found the claim while diagnosing an expensive verify query, checked it against the live schema, identified `000113` as what invalidated it, and rewrote the note.
- Human verification: The operator identified that the version upgrade had added these indexes, which is what prompted the check. Reviewer sign-off on the diff still required.

## Testing

- Comment-only change; no SQL modified.
- `000002` still parses under `clickhouse-format -n` with placeholders substituted.
- The claim was checked against `system.data_skipping_indices` on a live deployment: `traces` carries `idx_traces_id_bf` (bloom_filter) and `idx_traces_created_at` (minmax); `traces_local_v2` additionally carries `idx_traces_id_minmax` and `idx_traces_id_at`.

## Documentation

This *is* the documentation fix. No docs-site content is affected — the corrected note lives beside the SQL it describes, which is where an operator reads it.

### DIFF
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000002_delta_and_deletion_replay.sql
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000002_delta_and_deletion_replay.sql
@@ -128,9 +128,17 @@ SETTINGS max_insert_block_size = ${MAX_INSERT_BLOCK_SIZE},
 -- as deleted but is LIVE again on the source, and the backfill/delta already copied its live version. Deleting it by key
 -- would drop a row that is live on the source — silent data loss. So the replay deletes only ids that are NOT currently
 -- live on the source (mask-honored). The `id IN (deleted_ids since anchor)` bound keeps the deleted-id set tiny
--- (retention is off, so these are user-scale deletes); `traces` has no id skip index (000088 indexes only
--- created_at/last_updated_at — id minmax/bloom indexes exist only on traces_local_v2), so this source lookup is a
--- bounded id-filtered read of that tiny set, not a value-indexed prune of the full `traces` table.
+-- (retention is off, so these are user-scale deletes). `traces` DOES carry an id skip index --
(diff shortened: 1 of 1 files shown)


CometActions:
**No test needed here.**

Comment-only change: every added and removed line in 000002_delta_and_deletion_replay.sql is a `--` comment, and no executable SQL moved. Correcting the stale "traces has no id skip index" claim (idx_traces_id_bf landed in 000113) is worth doing since the comment was load-bearing during the OPIK-6901 rehearsal, but there is no runtime behaviour for a test to observe.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/32243450809)

PR #7907, opened 2026-08-19
[OPIK-7105] [QA] chore: remove the stale frontend Playwright suite

awkoy:
## Details

`apps/opik-frontend` carried its own Playwright suite that no longer ran and no longer described the product. This removes it, and everything that referenced it, so the cross-stack suite under `tests_end_to_end` is the only e2e suite in the repo.

Why it is dead:

- **No workflow executes it.** Nothing outside the directory referenced it — not a workflow, not the contribution guides, not `AGENTS.md`, whose Playwright line points at the cross-stack suite instead.
- **It was already declared unsupported** by `eee8f8ed1c` (2025-10, *"[OPIK-2615] [DOCS] Remove Unsupported E2E Test Instructions from Frontend Contribution Guides"*). The specs themselves last changed 2024-12.
- **Its page objects navigate V1-shaped workspace-level URLs.** V1 was removed earlier in this same ticket, so those paths now redirect elsewhere or resolve to nothing.

Removed together, because a partial removal is worse than either end state — it would leave `npm run e2e` failing on a missing `testDir` and a path alias resolving nowhere:

| What | Why it goes with the rest |
|---|---|
| the `e2e` directory — 43 files: 8 specs, page objects, fixtures, entities, test data | the subject |
| `playwright.config.ts` | its `testDir` pointed into the deleted tree |
| the `e2e` and `e2e:debug` npm scripts | they would fail on a missing test directory |
| the `@playwright/test` devDependency | nothing else in `opik-frontend` imports it; `tests_end_to_end` declares its own at `^1.50.0`, untouched |
| the `@e2e/*` path alias and two `tsconfig` `include` entries | they would dangle |
| the labeler rule matching the directory | it would match nothing |

The second commit clears three leftovers found in review: a stale path in `.agents/commands/comet/create-and-run-tests.md` that pointed agents at the deleted directory, two dead `e2e/` globs in `.dependency-cruiser.cjs`, and four Playwright output directories in the frontend `.gitignore`.

**All removed files are tracked, so this reverts in one commit.**

---

### :eyes: One thing for you to decide, @AndreiCautisanu

You own the cross-stack suite and its coverage taxonomy, so this is your call rather than mine.

That suite covers, by capability tag, essentially everything these specs exercised — trace listing and ordering, span tree, trace tags, filters, manual feedback scores, project list and delete, dataset list, view, add, edit and search.

**The exception is the feedback-definitions tab.** `tests_end_to_end/coverage/taxonomy.yaml` records `feedback-def-list`, `-create`, `-edit`, `-delete` and `-search` as `covered: false`, with the note *"A configuration.page.ts POM exists with no spec using it — half-built surface"*. The `configuration/feedback-definition.spec.ts` being deleted here is the only test for exactly those five.

Those specs do not currently work — they target V1 URLs and have not run in twenty months — so porting them means rewriting against project-scoped routes and the cross-stack fixtures, not a lift-and-shift. **Happy to do that as a separate PR if you want it; say the word and I'll pick it up.** Deleting without porting loses no working coverage, and the taxonomy already records the gap either way.

## Change checklist
- [ ] User facing
- [x] Documentation update

## Issues

- OPIK-7105

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: the removal, the follow-up cleanup, and this description
- Human verification: coverage compared tag-by-tag against the live suite before concluding what is lost; the absence search re-run after review revealed it had been skipping hidden directories (below); `package.json` diff inspected for round-trip damage after a scripted edit

## Testing

Typecheck, lint, `build` and `deps:validate` all pass. `npm ci` succeeds and no longer installs Playwright browser binaries. The lockfile drops exactly `@playwright/test`, `playwright`, `playwright-core` and a nested `fsevents` pin — the top-level `fsevents` entry correctly survives.

**Absence of references, re-run with `--hidden`:**

```
rg -l --hidden "opik-frontend/e2e|e2e:debug|playwright-report|blob-report" \
  --glob '!.git/**' --glob '!node_modules/**' .
# → nothing outside tests_end_to_end
```

Worth flagging why that matters: the first version of this check returned zero **without ever looking in `.agents/`**, because ripgrep skips hidden directories by default. That is how the stale path in `create-and-run-tests.md` survived the first commit — the search was wrong, not the repo. Re-run with `--hidden` it is now clean. The remaining `@e2e/` and `playwright` hits under `.agents/skills/` are correct: they document the cross-stack suite, whose own `@e2e/*` alias resolves against `tests_end_to_end/e2e/tsconfig.json`.

**Collateral damage: none.** `git diff origin/main...HEAD -- tests_end_to_end/` is empty.

**Coverage reporting is unaffected**, verified rather than assumed: `qa_coverage_reconcile.yml` triggers only on `tests_end_to_end/**` and runs `reconcile.py --estate tests_end_to_end`, so this directory was never in its field of view.

**No test environment requested** — this removes test infrastructure and touches no product surface, so there is nothing to click.

## Documentation

`.agents/commands/comet/create-and-run-tests.md` now points at `apps/opik-frontend/src/` for unit tests and `tests_end_to_end/e2e/` for cross-stack, instead of the deleted directory.

[OPIK-2615]: https://comet-ml.atlassian.net/browse/OPIK-2615?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


CometActions:
**No test needed here.**

Test-only removal — the 51 files are the deleted `apps/opik-frontend/e2e/` suite plus the config that existed only to run it (playwright.config.ts, the `e2e`/`e2e:debug` scripts and @playwright/test in package.json, the `@e2e/*` tsconfig path, the dependency-cruiser and .gitignore entries, the labeler rule). Nothing under `apps/opik-frontend/src/` or in any backend/SDK is touched, so there is no user- or API-observable behaviour for a test to assert. Checked for leftovers: no `@e2e/` or playwright reference remains in `src/`, and no workflow references `opik-frontend/e2e`.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/32249396230)

AndreiCautisanu (approved):

alexkuzmik (approved):

PR #7908, opened 2026-08-19
[OPIK-7025] [QA] Proposed e2e specs from the #7874 exploration: thread evaluation + context transform

CometActions:
> **Generated by the release QA side flow (exploration → test proposal).** It is a
> draft on purpose and needs a human review before it merges. Nothing here has
> been reviewed by a person yet.

## Where these came from

Exploratory testing of **#7874 — `[OPIK-7025] [SDK] feat: add context argument to evaluate_threads`**, run by hand against that PR's own deployed environment (`pr-7874.dev.comet.com`, serving `2.2.31-7874-merge-3016`, PR head `c106778`). Five ranked items were worked through; the two that were deterministic, seedable from a clean state, and asserting something a human would care about being wrong are proposed here as permanent specs.

**Base: `main`.** This originally targeted `alexkuzmik/OPIK-7025-evaluate-threads-context` and said it must not be retargeted, because both specs assert behaviour that only existed on #7874. That precondition is now **satisfied**: #7874 merged to `main` on 2026-08-19 (`ad3f58bf94`), its branch was deleted, and GitHub retargeted this PR automatically. Verified on `main` before merging it in — `evaluate_threads` carries `trace_context_transform`, and `evaluation_engine.py:132` writes the evaluation trace as `source="sdk"`. `main` has been merged into this branch; the merge was clean.

## The specs

`tests_end_to_end/e2e/tests/trace-explore/thread-evaluation.spec.ts` — `@t2-cuj`, `@area:threads`.

### 1. `evaluate_threads` scores the thread and leaves a visible `evaluation_task` trace in the eval project
`@cap:threads.thread-level-metrics` · `@cap:threads.thread-feedback-score`

Seeds a 3-turn conversation, runs `evaluate_threads` with a fixed-score `ConversationThreadMetric` into a separate `eval_project_name`, and asserts:

- the run reports the score it was given (`0.75`, name and reason);
- that score lands on the **thread** over the API — compared as the whole score set, so a run that also wrote scores it should not have fails rather than passing on a `find()`;
- the evaluation project's traces **under the exact filter the Logs page sends** (`source = sdk`, per `generateLogsSourceFilter`) are exactly `[the evaluation trace]`, and the `source = experiment` read is empty. This pair is the point of the spec: pre-#7874 the trace was written as `source="experiment"`, thread evaluation creates no experiment for the UI to reach it through, and the Logs page rendered "No traces yet" with no error anywhere — the run succeeded and its evidence was invisible;
- both facts render: the thread panel's Feedback scores tab shows the metric at `0.75`, and the eval project's Logs table lists exactly that one row.

### 2. `trace_context_transform` attaches context to assistant messages only, and omits the key entirely when not passed
`@cap:threads.thread-level-metrics`

Same thread, evaluated twice — once without `trace_context_transform`, once with `lambda trace: trace.metadata["retrieved_docs"]` — into two evaluation projects. Asserts, in three places (the conversation the metric's `score()` received, the evaluation trace's `input` read back over the API, and the trace detail panel):

- **without** the transform, no message carries a `context` **key at all**. Checked as key presence, not `!= null`: `model_dump(exclude_none=True)` is what keeps the key absent, and `"context": null` appearing in every existing caller's payload is exactly the silent regression a null-check would wave through;
- **with** it, every assistant message carries its **own** trace's document — the three documents are deliberately distinct, so a transform that attached the wrong turn's context fails — and no user message carries context;
- in the browser: the eval trace written *with* context renders the document, and the one written *without* it does not (gated on the panel having rendered its conversation first, so "absent" cannot mean "not painted yet").

## Verification

Run against `https://pr-7874.dev.comet.com` (OSS install, workspace `default`), with the bridge running the SDK from this branch:

```
npx playwright test tests/trace-explore/thread-evaluation.spec.ts --reporter=list
  2 passed (23.5s)

npx playwright test tests/trace-explore/ --reporter=list        # shared POM + fixture chain touched
  18 passed (1.1m)

python3 tests_end_to_end/coverage/tag_lint.py --taxonomy tests_end_to_end/coverage/taxonomy.yaml --estate tests_end_to_end
  39 specs checked, 1 exempt, 0 problem(s)

npx tsc --noEmit                                                 # clean
```

Both specs were also mutation-checked — the metric value and the rendered-context locator were each perturbed and each failed at the intended line — so the UI steps demonstrably execute and the assertions are not passing vacuously.

## The one decision that needs a call: where the bridge gets `opik` from

`services/opik-sdk-driver/pyproject.toml` now resolves `opik` from `sdks/python` (editable) instead of PyPI, and the README paragraph documenting the old intent is updated to match.

This is deliberate and it is the change to argue about. As it stood, `uv sync` installed the newest released wheel. #7874 is on `main` but is **not in a published wheel yet**: the `2.2.32` bump commit landed after it, so the released `2.2.32` still has no `trace_context_transform` and still writes `source="experiment"`. Both specs would therefore have gone red in the estate's own CI the moment they landed, not because anything is broken but because the released SDK cannot exercise a change that has not shipped. That makes *any* e2e spec for an in-flight SDK change unrunnable by construction. The repo's Python SDK E2E job already resolves this the same way (`uv pip install --system .` from `sdks/python`).

The trade-off is real and it is yours to accept or reject: every bridge-driven spec now runs against the working tree's SDK, so an SDK regression on the branch turns them red. That is arguably the point of an in-repo suite, but it is a wider blast radius than this PR alone. The whole `trace-explore` directory passes under it (18/18 above); the rest of the estate was not run, per "run only what you added". If you would rather keep PyPI, the alternative is to hold these specs until the release carrying #7874 ships and land them then.

- **Context-aware `ConversationalCoherenceMetric` judge.** Needs a provider key and an LLM verdict — non-deterministic and out of reach on this environment. Left to the PR's unit tests, as triage proposed.
- **`verbose` became keyword-only.** #7874 inserts `trace_context_transform` as the 7th positional parameter *and* a bare `*` after it, so a caller that passed `verbose` positionally now binds it to the transform. `evaluate_threads(..., 0)` returns normally with the progress bar the caller asked to suppress; `evaluate_threads(..., 1)` raises `TypeError: 'int' object is not callable` from inside the conversation factory, naming nothing. Both behaviours were reproduced by hand. Not written up as a spec: it is a judgement call for the reviewer of #7874 (a changelog line may be all it warrants), and pinning the current behaviour in an e2e test would prejudge that.
- **`input.metrics` rendering as a raw Python repr** (`<FixedScoreCapturingMetric object at 0x…>`) in the trace panel. Unchanged by this diff — only newly *visible*, because these traces reach the UI for the first time as a result of #7874. A polish follow-up, not a regression, so nothing was written to lock the current output in.

## Caveats for the reviewer

- The verification environment is an OSS install (no auth) served on a `*.comet.com` hostname. The TypeScript SDK gates its "API key required" check on that hostname suffix, so the run needed a placeholder `OPIK_API_KEY`. That is a property of this environment, not of the specs — a normal `localhost:5173` OSS run needs nothing.
- POM selectors for the new thread-panel Feedback scores tab were chosen by reading the rendered component (`ThreadDetailsPanel.tsx` reuses the trace panel's `ConfigurableFeedbackScoreTable`) and confirmed by running, rather than through a live Playwright MCP session — that tooling was not available in this environment.

Closes nothing. Related: #7874.

[OPIK-7025]: https://comet-ml.atlassian.net/browse/OPIK-7025?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Details

Two Playwright specs covering `opik.evaluation.evaluate_threads` (OPIK-7025 / #7874),
plus the test support they need: a `threads` route module on the existing
`opik-sdk-driver`, an `evaluated-thread` fixture, a `thread-panel` page object, and a
`getTrace` addition to the e2e backend client. All changes are inside
`tests_end_to_end/` — no production code.

## Issues

Resolves OPIK-7025 (test coverage for #7874).

## Testing

Specs run against `https://pr-7874.dev.comet.com` (OSS install, workspace `default`):

- `tests/trace-explore/thread-evaluation.spec.ts` — 2 passed
- `tests/trace-explore/` (whole directory, since the shared POM and fixture chain
  were touched) — 18 passed
- `tag_lint.py` — 39 specs checked, 1 exempt, 0 problems
- `tsc --noEmit` — clean

Both specs were mutation-checked: the metric value and the rendered-context locator
were each perturbed, and each failed at the intended line, so the UI steps
demonstrably execute and the assertions are not passing vacuously.

Note that these specs are `@t2-cuj`, so the PR's own `Run v2 suite: t1` check does not
execute them. That check's failure on this PR is unrelated — a `test-suites` spec
failing on a missing `OPENAI_API_KEY` (1 failed / 17 passed).

## Documentation

None required — test-estate only. The `opik-sdk-driver` README is updated in this PR
to document where the bridge resolves `opik` from.

## Change checklist

- [x] Tests added (this PR is the tests)
- [x] Tag grammar followed — `@t2-cuj`, `@area:threads`, `@cap:` keys that exist in `taxonomy.yaml`
- [x] `taxonomy.yaml` updated so the new coverage is counted
- [ ] Documentation updated — n/a
- [ ] Breaking change — no


AndreiCautisanu:
## 🔍 Generated-test review

**VERDICT: ready with notes** — gates green, two findings fixed in 5bdbae8.

### Gates (all three run, first time for this spec)

| gate | |
|---|---|
| `tag_lint` | **PASS** — 39 specs, 1 exempt, 0 problems |
| `tsc --noEmit` | **PASS** |
| `playwright test --list` | **PASS** — both tests collect; 18 tests / 7 files across `trace-explore/` |

The PR's own `Run v2 suite: t1` check does **not** execute these specs — they are `@t2-cuj`. Its failure is unrelated (a `test-suites` spec, missing `OPENAI_API_KEY`).

### Fixed in this review

1. **The readiness loop aborted on a transient transport error.** `_wait_for_thread` polls because thread aggregation is eventually consistent — and that window is exactly when a connection is most likely to drop. An `httpx.ConnectError` escaped the loop and killed the evaluation for the very condition the wait absorbs, surfacing as a flake with no useful message. Now caught and polled through, with the deadline still the bound; a timeout that hit transport errors says so rather than reporting a clean one. Narrow on purpose — `HTTPStatusError` is a real answer and still propagates.
2. **The endpoint docstring omitted `context_metadata_key`**, the one parameter whose *absence* is meaningful: unset means the SDK argument is omitted entirely rather than passed `None`, which is the pre-existing caller shape these specs assert about.

### The spec itself — no changes needed

Checked and correct:

- **Placement.** `@area:threads` in `tests/trace-explore/` is right: the area declares `spec_dir: trace-explore`, three sibling specs there are already `@area:threads`, and four are `@area:traces`. The spec is also already in the area's `specs:` list. (`TESTING-TAGS.md` claimed "Directory = area … Enforced", which is wrong about this estate and is corrected in #7926.)
- **Tags.** Both `@cap:` keys exist and are asserted. `trace_context_transform` is an argument to `evaluate_threads` shaping the conversation the thread metric scores, so `thread-level-metrics` is the right key rather than a mis-attribution.
- **Reliability.** No fixed sleeps; `expect.poll` with a reason on every assertion; entities namespaced via the estate's `testNamespace`, which satisfies the `cuj-YYYYMMDD-HHMMSS-mmm` teardown contract.
- **Assertion quality.** Context absence checked by key presence (`Object.keys()`), not `!= null` — correct, since `exclude_none` makes absence the contract. Scores compared as a whole set rather than `find()`, so a run writing extra scores fails.

### Bot comments

|---|---|---|---|
| 1 | `?? null` collapses absent vs. null `input` | **invalid** | unreachable — the spec asserts `input` not-null *then* reads `.conversation`; and `TraceDetail`'s docstring says key-presence *inside* the record is what's preserved, which is what the context assertions use |
| 2 | Transient network errors abort evaluation | **valid** | **fixed** in 5bdbae8 |
| 3 | Docstring omits context behaviour | **valid** | **fixed** in 5bdbae8 |
| 4 | `finally` teardown can mask the real exception | **valid** | **deferred — driver-wide.** The identical three lines are in all nine sibling routes; fixing this file alone makes it the inconsistent one. Needs a shared teardown helper in its own PR. |
| 5 | Blank strings accepted | **out of scope** | `schemas.py` has zero `Field(`/`min_length`/`constr` in 376 lines — this would make the new module the sole exception. Same driver-wide PR as #4. |
| 6 | `score_value` accepts NaN/inf/bool | **out of scope** | as #5 |

### Could not verify

The specs have not been **run** against current `main`. They were run and mutation-checked against `pr-7874.dev.comet.com`, which no longer exists; #7874 has since merged, so the behaviour they assert is on `main`, but nothing has executed them there.

Also worth knowing: `trace_context_transform` is on `main` but **not in a published wheel** — the 2.2.32 bump landed after it. So the `[tool.uv.sources]` editable install this PR adds is still load-bearing, not a leftover.

---
_The two out-of-scope findings and #4 belong in one driver-wide PR: a shared teardown context manager plus request-schema validation across all ten route modules._

AndreiCautisanu (approved):

natagh23 (approved):

Issue #7862, opened 2026-08-16
Alert creation API silently accepts `window_in_seconds` for `threshold:errors`, evaluator requires `window`; alert still never fires after the fix


sylbae:
## Alert creation API silently accepts `window_in_seconds` for `threshold:errors`, but the evaluator requires `window`; even after correcting the key, the alert never fires on data confirmed to cross the threshold

**Version:** `opik-backend:2.2.29` (self-hosted Docker Compose)

> **Disclosure:** this report was written by an AI coding agent (Claude, via Claude Code)
> investigating a real self-hosted deployment, with the findings reviewed by me before submitting.
> Confidence differs sharply between the two parts below — please weight them accordingly:
> - **Part 1 is fully verified**: reproduced live, exact log lines quoted directly from the
>   running container, the fix applied and confirmed by observing clean logs afterward. High
>   confidence, not speculative.
> - **Part 2 is an observed symptom, not a root-caused bug report.** The agent confirmed the
>   *symptom* (no result ever logged, no webhook ever sent, despite data independently confirmed
>   to cross the threshold) through container logs and the REST API only — it did not have direct
>   database or JVM access and explicitly could not determine the underlying mechanism. Treat Part
>   2 as "here's a reproducible test case that behaves unexpectedly," not "here's the bug."

### Part 1 — confirmed: config key mismatch, no validation at creation time

`POST /api/v1/private/alerts` accepts a `threshold:errors` trigger config with the key
`window_in_seconds` — this matches the shape shown in your own REST API reference page
(`https://www.comet.com/docs/opik/latest/reference/rest-api/alerts/create-alert`) — and returns
`201 Created`, storing the value verbatim with no complaint:

```json
{
  "type": "threshold:errors",
  "config_value": { "threshold": "3", "window_in_seconds": "300" }
}
```

But the evaluator (`MetricsAlertJob.java:432`) expects the key literally named `window`, not
`window_in_seconds`. Every scheduled evaluation (every 5 minutes) then fails silently — no error
surfaces anywhere except the backend container log:

```
ERROR com.comet.opik.api.resources.v1.jobs.MetricsAlertJob: Failed to process metrics alert
'...': Missing config value for key 'window' in trigger of type 'THRESHOLD_ERRORS'
```

The alert appeared healthy in the UI/API the entire time (`enabled: true`, no error field on the
resource) — this is only visible in backend logs. Updating the trigger config to use `window`
[...]

PR #7927, opened 2026-08-20
fix(alerts): accept window_in_seconds as the threshold window

BetterAndBetterII:
## Summary

- `threshold:errors` configs stored as `window_in_seconds` (the documented REST spelling) were accepted with 201, then every `MetricsAlertJob` run failed looking for `window`.
- `AlertTriggerConfig.resolveWindow()` prefers `window`, then `window_in_seconds`. Canonical `window` is unchanged.
- Fixes the verified Part 1 of #7862. Does not change Part 2 (alert still not firing after the key is corrected).

## Test plan

- [x] RED then GREEN: `window_in_seconds` fires the same `TRACE_ERRORS` webhook as `window`
- [x] `window` still works
- [x] `mvn test -Dtest=MetricsAlertJobTest` — 13 passed

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/AlertTriggerConfig.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/AlertTriggerConfig.java
@@ -5,8 +5,10 @@ import com.fasterxml.jackson.annotation.JsonView;
 import com.fasterxml.jackson.databind.PropertyNamingStrategies;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import io.swagger.v3.oas.annotations.media.Schema;
+import jakarta.annotation.Nullable;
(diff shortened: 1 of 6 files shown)


andrescrz:
Hi @BetterAndBetterII 

Thanks for the contribution!

Thanks!

BetterAndBetterII:
Addressed the review comments:

- Missing-window errors now name both `window` and `window_in_seconds`.
- Form load/save resolves `window_in_seconds` and writes canonical `window`, so alias-only alerts stay editable instead of serializing empty `trigger_configs`.
- Blank or whitespace `window` falls back to the alias.
- `resolveWindow` now has an explicit nullable contract.
- Alias coverage includes the five threshold event types.
- When both keys are set, evaluation uses `window`.

PR #7930, opened 2026-08-20
[OPIK-8012] [SDK] fix: bound LLM judge provider calls and retry empty responses

AndreiCautisanu:
## Details

Three defects in the LiteLLM-backed LLM-judge path let a brief provider hiccup become a hard failure that blocks the caller far longer than any caller-side budget. Found while investigating an intermittently failing E2E smoke test, but the impact is on any SDK user running `evaluate` / `run_tests` with an LLM judge.

- **Empty structured-output responses were never retried.** A response with neither `content` nor `tool_calls` raises `BaseLLMError`, but the judge's retry policy only caught `LLMJudgeParseError` — which extends `OpikException`, not `BaseLLMError`. The item lost its entire score even though the LLM call sits inside the retried `_generate_and_parse` and a retry would very likely have succeeded.
- **No timeout or retry ceiling reached `litellm.completion`.** Neither `timeout` nor `num_retries` was forwarded, leaving litellm's 6000s (100 min) default plus its ~3 internal HTTP retries stacked on two existing tenacity layers. Thread dumps taken mid-hang show the stalls sit in TCP connect, so connect is now bounded tighter (10s) than read (60s): a provider either accepts a connection promptly or is unreachable, whereas a long generation legitimately takes time. Both are `setdefault`, so explicit caller values still win.
- **The error message blamed the API key.** The key was valid in every observed case; the real condition is an empty response shape. It now reports `model` and `finish_reason`.

Measured cost of one stalled call: `timeout=5` with `num_retries` unset took 50.8s, versus 17.1s pinned. A blackholed connect went from ~180s to 32.2s.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-8012

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: Investigation, root-cause analysis, the three source fixes, and the four unit tests.
- Human verification: Pending review. Every claim above is backed by a cited measurement or captured stack trace; see the "Not fully verified" caveat below for the one gap.

## Testing

```bash
cd sdks/python && pytest tests/unit/evaluation/models tests/unit/evaluation/suite_evaluators -q
```
→ **321 passed, 2 skipped.**

```bash
cd sdks/python && pytest tests/unit/evaluation -q
```
→ **794 passed, 45 failed.** Those 45 are pre-existing: I stashed this branch's changes and re-ran to get the identical 45 (missing optional deps such as `nltk`/`rouge`, plus test-ordering pollution — two `test_litellm_chat_model.py` cases pass when the file is run scoped). This branch takes the suite from 790 to 794 passing.

```bash
pre-commit run --files <the 5 changed files>
```
→ ruff, ruff-format, mypy all **Passed**.

Each of the 4 new tests was confirmed to *fail* without its corresponding fix (reverted the fix, re-ran, restored) so none is vacuous.

End-to-end, against a local stack with a real Anthropic judge: runs of the affected smoke test that previously aborted at the 90s client timeout now pass.

## Not fully verified

The trigger for the first defect — a provider returning a blank `tool_use` response — was never reproduced live. The unit test reproduces the exact exception path from the CI traceback and the fix is correct regardless, but confirming the E2E flake is actually gone needs a few green CI runs after merge.

Worth noting for reviewers: a large share of my local reproductions turned out to be a broken IPv6 route on my dev machine (IPv4 connect 0.01s, IPv6 timeout after 12s), which is an environment problem, not an Opik defect. GitHub-hosted Linux runners are IPv4-only, so that cannot explain the CI failures — the retry gap is what does.

Deliberately not changed: the E2E test's 90s client timeout. CI already burns ~95s per attempt across 3 attempts, and healthy judge latency is ~1.1s median, so raising it would only make failures slower.

### DIFF
--- a/apps/opik-documentation/documentation/fern/docs-v2/evaluation/evaluate_threads.mdx
+++ b/apps/opik-documentation/documentation/fern/docs-v2/evaluation/evaluate_threads.mdx
@@ -161,10 +161,10 @@ The extracted context is attached to the agent message produced by the same trac
 ```
 
 <Note>
-`ConversationalCoherenceMetric` is context aware: it shows the retrieved documents to the judge, so a turn
-scores only when the answer is both relevant and supported by them. Metrics that don't need the context ignore
-it, and the built-in prompts render `role` and `content` explicitly, so adding a context transform never
-changes their scores. If you write
+`ConversationalCoherenceMetric` is context aware: the retrieved documents are shown to its judge as background,
+so it can make a better informed relevancy call - what it measures is unchanged. Metrics that don't need the
+context ignore it, and the built-in prompts render `role` and `content` explicitly, so adding a context
+transform never changes their scores. If you write
 your own context-aware metric, set that attribute on it - see the [Custom Conversation Metrics guide](/evaluation/metrics/custom_conversation_metric).
 </Note>
 
--- a/apps/opik-documentation/documentation/fern/docs-v2/evaluation/evaluate_threads.mdx
+++ b/apps/opik-documentation/documentation/fern/docs-v2/evaluation/evaluate_threads.mdx
@@ -161,10 +161,10 @@ The extracted context is attached to the agent message produced by the same trac
 ```
 
 <Note>
-`ConversationalCoherenceMetric` uses the retrieved documents to additionally report a
-`conversational_groundedness_score` - whether each answer is supported by the documents it was generated from.
-Metrics that don't need the context ignore it, and the built-in prompts render `role` and `content` explicitly,
-so adding a context transform never changes their scores. If you write
+`ConversationalCoherenceMetric` is context aware: it shows the retrieved documents to the judge, so a turn
+scores only when the answer is both relevant and supported by them. Metrics that don't need the context ignore
+it, and the built-in prompts render `role` and `content` explicitly, so adding a context transform never
+changes their scores. If you write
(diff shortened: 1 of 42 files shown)


AndreiCautisanu:
Closing — this was branched off an unmerged feature branch, so it carried 9 unrelated OPIK-7025 commits and 41 changed files. Reopened as a pointed 5-file PR off main.

CometActions:
**No test needed here.**

Measured against main, this PR's only product change is three files: the LiteLLM connect/read timeout + num_retries=0 defaults, BaseLLMError joining the suite LLM judge's retry policy, and the richer empty-response message. (The rest of the 41-file diff — the evaluate_threads context argument, the coherence refactor, the ADK usage fix and thread-evaluation.spec.ts — is already on main; the branch just has an older merge base.) All three only surface when a real provider stalls on connect or returns a structurally blank structured-output response, which needs a provider key and can't be induced on a fresh OSS install, so an e2e spec couldn't assert any of them deterministically. The unit tests in tests/unit/evaluation/models/test_litellm_chat_model.py and tests/unit/evaluation/suite_evaluators/test_llm_judge.py pin the defaults and the retry at the level that can actually see them. Worth flagging for human judgement rather than a test: the 60s read timeout and num_retries=0 now apply to every LiteLLMChatModel caller, so a slow-but-healthy generation past 60s fails where it previously waited.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/32351535378)

PR #84, opened 2023-10-10
[CM-8224] replace comet api client usage in log user feedback

jynx10:

PR #7937, opened 2026-08-20
[OPIK-7852] [BE] Persist cipx call attribution: trigger, trigger_detail, turn_key, parent_tool_use_id

aadereiko:
## Details

cipx already ships `trigger`, `trigger_detail`, `turn_key` and `parent_tool_use_id` on `metadata.cipx.call`, but none reached a typed column — so the spend tables could not tell a sub-agent's spend from the main agent's, name the agent, or reconstruct the parent/child tree. This adds the four columns and the DAO mapping.

- `trigger_detail` is **overloaded**: it carries the agent NAME when `trigger='subagent'` and the TOOL name when `trigger='tool_continuation'`, so consumers must gate on `trigger` before grouping per agent.
- Empty is deliberately not defaulted — it means unknown, never a guessed agent name.
- Migration `000118` is additive with `DEFAULT ''`, so pre-existing rows read as unknown, which is honest: they were written before the proxy carried the fields.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- OPIK-7852

Follow-up found while testing this work, not resolved here: OPIK_8026 (a fork's inherited-cache spend is billed to the main agent). Related but not resolved: OPIK_7757 (`inference_geo`), whose field name is used as the "unknown field" in one test precisely because it is a real pending addition.

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: Migration, DAO mapping and both tests authored with Claude Code; the design decisions (forward-compatibility over validation, two-row sentinel insert) were reviewed and are described below.
- Human verification: Test suite run locally (12 tests green) before the local Maven harness degraded; end-to-end verified against a local Opik with a branch-built proxy — 9 real agent invocations, all named and linked, depth-2 nesting intact.

## Testing

`mvn test -Dtest='CostIntelligenceIngestionTest'` — **12 tests, 0 failures** at the time the tests were written.

Two tests, both aimed at failure modes that would otherwise be silent:

- **Positional bind order.** The DAO binds by position and nothing checks the order against the INSERT tuple at compile time; a mismatch writes each value into the neighbouring column without erroring. Every column gets a distinct sentinel and is asserted to hold its own, with the column name as the assertion description. **Two rows, not one** — the bind index accumulates across rows while `workspace_id` is bound once at index 0 and its repeats dedup, so a wrong stride corrupts only the *second* row and a single-row insert cannot see it. The ids are UUIDs because `project_id`/`trace_id`/`span_id` are `FixedString(36)`, where a rotation among the three is silently accepted by ClickHouse.
- **Unknown-field tolerance.** cipx ships to laptops on its own cadence, so a newer proxy talking to an older backend is the normal state, not an edge case. Unknown fields at every level must be ignored rather than reject the row — dropping it would lose the spend entirely, and spend totals are the one thing already correct today.

**End-to-end:** migration applied to a local Opik and exercised by real Claude Code sessions through a branch-built proxy. Result: 9 agent invocations, all named and linked, `agents_dispatched=9 linked=9 ambiguous=0`, depth-2 nesting intact, and the auto-mode permission classifier correctly bucketed as `automated` rather than counted as user work.

**Not re-run locally:** the last commit moves one assertion inside an existing `await` (a strictly narrower timing window). The local Maven surefire environment has since started failing this whole test class with `ClassNotFoundException: DropwizardAppExtensionProvider` — including on test methods untouched by this PR, so it is environmental — and CI is the verification for that commit.

## Documentation

No user-facing documentation change. The columns are internal analytics storage; the semantics (especially `trigger_detail`'s overloading and the empty-means-unknown rule) are documented in migration `000118`'s changeset comment and in the DAO's javadoc, which is where the next person to touch the INSERT will look.


CometActions:

This PR carries its own test at the only layer where the change is observable. `CostIntelligenceIngestionTest` asserts the five new `cipx_spends` attribution columns, the four new `cipx_trace_identities` columns, and — the real risk in a positional-bind DAO where a tuple just grew from 17 to 22 and 26 to 31 — `every positional bind lands in its own column, across a multi-row insert` for both DAOs. The two behaviour changes riding along are covered too: `a reordered re-upsert of one trace keeps the newer snapshot's counters` pins `last_updated_at` now coming from `event.createdAt()` instead of the column DEFAULT, and the UInt32 range tests pin the int->long widening and the reject-to-0 path. No e2e test would add anything: Opik never reads these tables back — there is no REST resource and no opik-frontend page that queries `cipx_spends` or `cipx_trace_identities` (the only frontend mention is the comet-plugin `goToCostIntelligence` link, which navigates out of Opik). Nothing to add.

**Not testable yet.** The end consumer of these fields — Cost Intelligence dashboards slicing spend by trigger/subagent lineage — is user-facing, but no test in this estate can ever reach it: Opik exposes no read path for cipx data and the e2e estate has no ClickHouse client (only `opik` and `js-yaml` in e2e/package.json). Recording it so it is not silently lost if a Cost Intelligence surface lands in Opik later.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/35724867143)

_Re-checked after a push on 22 Sep 12:11 UTC._

apps/opik-backend/src/main/java/com/comet/opik/domain/CipxSpendDAO.java
```diff
@@ -84,6 +89,16 @@ public static SpanRow from(UUID spanId, UUID traceId, UUID projectId, JsonNode m
...
+                    // Attribution fields. trigger_detail carries the subagent NAME when
+                    // trigger=subagent (the parent's Agent tool_use input.subagent_type) and the tool
+                    // name when trigger=tool_continuation; parent_tool_use_id identifies the agent
+                    // invocation this call belongs to, which is the parent/child edge of the agent
+                    // tree. All default to "" when the proxy could not resolve them — an empty value
+                    // means unknown and must not be substituted for a guessed agent name.
+                    .trigger(call.path("trigger").asText(""))
+                    .triggerDetail(call.path("trigger_detail").asText(""))
+                    .turnKey(call.path("turn_key").asText(""))
+                    .parentToolUseId(call.path("parent_tool_use_id").asText(""))
```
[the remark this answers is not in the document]
aadereiko: Declining — validating `trigger` against a closed set would work against what the sibling test in this PR exists to protect.

cipx ships to laptops on its own cadence, independently of this service's deploys, so **a newer proxy talking to an older backend is the normal state**, not an edge case. `trigger` is a closed enum *as of today*; the moment cipx adds a value (a fork trigger is already on the roadmap — OPIK_8026), validating against migration 000118's list would reject or mangle a legitimate value. That is exactly the failure `unknownFieldsDoNotRejectTheRow` guards against, and dropping the row would lose the spend entirely — spend totals being the one thing that is already correct today.

On the normalization half: `asText("")` already does it. A null, numeric, object or array node yields `""`, which **is** the documented unknown sentinel (see the `SpanRow.from` comment and the 000118 changeset). So malformed input already lands on "unknown" without rejecting the row.

You're right that this makes malformed values indistinguishable from genuinely unresolved attribution. That's deliberate and stated: `""` means unknown, and a guessed agent name would book real spend against an agent that never ran. Distinguishing "the proxy could not resolve this" from "the proxy sent something we could not parse" would need a separate signal on the wire, not a decision made at parse time here.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/CostIntelligenceIngestionTest.java
```diff
@@ -162,6 +173,100 @@ void spanCreatedWithCipxCallLands() {
...
+        void spanWithoutTriggerFieldsLandsWithEmptyAttribution() {
+            var ws = newWorkspace();
+            String projectName = "cipx-" + UUID.randomUUID();
+
+            // systemToolsCipxMetadata carries no trigger/turn_key/parent_tool_use_id — the shape
+            // every span written before the proxy shipped them has. Ingestion must still land the
+            // row and must leave the attribution columns empty rather than substituting a default:
+            // "" means unknown, and a guessed agent name would book real spend against an agent
+            // that never ran.
+            var span = factory.manufacturePojo(Span.class).toBuilder()
```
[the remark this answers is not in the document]
aadereiko: Correct that it is untested, and fair to ask given the PR body claims pre-existing rows read as unknown. Declining anyway, on cost/benefit.

What such a test would assert is ClickHouse's own `ADD COLUMN … DEFAULT ''` semantics — materializing the default for rows written before the column existed. That is not behaviour this repo implements, and there is no code path here that could regress it: the DAO only ever writes new rows, and the migration is a single additive `ALTER`. A green assertion would be testing the database, not us.

The cost is also not small. `CostIntelligenceIngestionTest` applies the full migration set at construction (`MigrationUtils.runClickhouseDbMigration`), so exercising the pre-migration state needs either a second database or custom changeset ordering in the harness. That is a meaningful amount of test infrastructure to protect a property we cannot break.

For what it's worth, the equivalent case on the read side *is* covered, because there it protects our code rather than ClickHouse's: `ai-cost-backend` #84 has a test that queries a genuinely column-less table and asserts the endpoint degrades to `available=false` instead of raising — that one guards a `try/except` we wrote.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000118_add_trigger_to_cipx_spends.sql
```diff
@@ -0,0 +1,66 @@
...
+--                       should ignore it, a tree-shape audit should not.
+--
+-- Additive columns with defaults; every pre-existing row reads '' (unknown), which is honest --
+-- those rows were written before the proxy carried the fields.
+ALTER TABLE ${ANALYTICS_DB_DATABASE_NAME}.cipx_spends ON CLUSTER '{cluster}'
+    ADD COLUMN IF NOT EXISTS `trigger`           LowCardinality(String) DEFAULT '',
+    ADD COLUMN IF NOT EXISTS trigger_detail      String                 DEFAULT '',
+    ADD COLUMN IF NOT EXISTS turn_key            String                 DEFAULT '',
+    ADD COLUMN IF NOT EXISTS parent_tool_use_id  String                 DEFAULT '',
+    ADD COLUMN IF NOT EXISTS link_failure_reason LowCardinality(String) DEFAULT '';
```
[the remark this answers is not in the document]
aadereiko: Declining, on both halves — though note the file this is anchored to no longer exists: `000118`, `000119` and `000120` were folded into a single `000118_add_cipx_call_attribution.sql` in 6a4b088, so the "apply the same to 000119" half is moot.

**Readiness gate.** This is one of 101 `ON CLUSTER` changesets in `db-app-analytics`, 63 of them additive `ADD COLUMN`. That directory contains zero occurrences of `distributed_ddl`, `SYSTEM SYNC` or `--precondition` — a gate here would be the only one in the repo. And the mid-rollout window it would close is already closed by the defaults: every column lands `DEFAULT ''` or `DEFAULT 0`, so a replica that has the column and one that does not return the same value for every pre-existing row. A backend predating the changeset never names them; one postdating it reads the default. `ai-cost-backend` reading a column that defaults to `''` needs no gate either — and it additionally detects the missing-column error by name and degrades to a typed `available=false` rather than erroring.

**Rollback ordering.** Agreed, and the consolidated file says so directly above its `--rollback` lines: "Roll the backend back BEFORE this: CipxSpendDAO and CipxTraceIdentityDAO each name their new columns in one INSERT, so dropping them fails every cipx write as a whole batch, and the ingestion listener logs-and-swallows that with no retry and no dead letter."

On rollback dropping a column that was found pre-existing: all nine columns are introduced by this changeset — no earlier migration creates them on either table — so "pre-existing" only happens if someone hand-added them, which no rollback contract can cover. Column-ownership tracking and blocking a rollback while a DAO is live are not things Liquibase can express, and no changeset in this repo attempts them; the sequencing note is the mechanism available.

apps/opik-backend/src/main/java/com/comet/opik/domain/CipxTraceIdentityDAO.java
```diff
@@ -100,6 +104,20 @@ public static TraceIdentityRow from(UUID traceId, UUID projectId, JsonNode metad
...
+                    // Session-grain subagent link rollup. These are the ONLY
+                    // place cipx's worst attribution failure is visible: a
+                    // subagent whose dispatch was never observed looks
+                    // exactly like a main-loop turn, so no span carries a
+                    // link_failure_reason and only a missing increment here
+                    // reveals it. They are session RUNNING TOTALS re-stamped
+                    // on every trace upsert of that session, so a session
+                    // with N traces leaves N rows holding N successive
+                    // snapshots. Readers must take max() per session_id,
+                    // never sum().
```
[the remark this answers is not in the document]
aadereiko: Declining. The mechanism is real; the conclusion inverts what the table is.

`cipx_trace_identities` is a **projection** of trace metadata, not an independent record. `TraceDAOImpl.bulkUpdate` does apply `TraceUpdate.metadata` to every id in the batch (`newBulkUpdateTemplate`, applied whenever `metadata` is non-null — which is exactly when `CipxMetadata.hasIdentity` gates the listener in), and `TraceService.batchUpdate` posts the event *after* that write. So by the time the listener runs, all N traces genuinely carry that `cipx.session` in `traces.metadata`.

Rejecting it in the listener would make the projection disagree with the table it projects from: a trace whose metadata claims session X with no identity row is harder to diagnose than the honest projection. And "resolve it independently per trace" has no input — the request carries exactly one metadata document, and the listener deliberately never reads trace metadata back (see the class javadoc); re-reading would return the value just written.

The counters specifically cannot be corrupted this way. They are read with `max()` per `session_id` — rule 1 in the changeset header — and `max()` over N copies of one snapshot is that snapshot. What a batch patch changes is *which traces claim the session*, which is what the trace metadata now says.

It is also not a path the producer uses: cipx ships `POST /v1/private/traces` and, on 409, `PATCH /v1/private/traces/{id}` — the single-trace form, one metadata document per request. And the shape predates this PR: `onTraceCostIntelligenceChanged` already stamped `session_id` and the identity fields from one `update.metadata()` across `traceProjectIds` on main. The four counters added here ride that same snapshot.

The per-trace endpoint, `PATCH /traces/{id}`, carries one metadata document per trace by construction — that is where a session-stamping producer belongs, and it is the one cipx uses.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000119_add_cipx_call_attribution.sql
```diff
@@ -0,0 +1,88 @@
...
+--      denominator. Reporting 0% invents a failure out of an idle session.
+--   4. Zero is not "unknown". Absence reads as zero, which is faithful for a
+--      daemon that dispatched nothing; one too old to emit them is told apart by
+--      cipx_version / schema_version.
+ALTER TABLE ${ANALYTICS_DB_DATABASE_NAME}.cipx_spends ON CLUSTER '{cluster}'
+    ADD COLUMN IF NOT EXISTS `trigger`             LowCardinality(String) DEFAULT '',
+    ADD COLUMN IF NOT EXISTS trigger_detail        String                 DEFAULT '',
+    ADD COLUMN IF NOT EXISTS turn_key              String                 DEFAULT '',
+    ADD COLUMN IF NOT EXISTS parent_tool_use_id    String                 DEFAULT '',
+    ADD COLUMN IF NOT EXISTS link_failure_reason   LowCardinality(String) DEFAULT '';
```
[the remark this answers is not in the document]
aadereiko: Declining — the producer cannot emit the update this depends on.

Opik's `SpanUpdate` does accept `metadata`, so the API permits it in the abstract. But cipx is the only writer of `cipx_spends`, and its patch body is `{ids, update}` where `update` is `{trace_id, project_name, error_info}` — there is no `metadata` field on it at all (`internal/spansink/types.go`, `SpanFields`). Its one use today is marking a tool sub-span errored once its `tool_result` arrives, and tool sub-spans carry no `cipx.call`, so `CipxMetadata.hasSpendCall` filters them out on create and there is nothing for an update to refresh.

That is why the listener is create-only, and it says so: "Span ingestion is create-only — cipx call data is complete on the create event and immutable, so there is no span update path." Usage, config and attribution are all read off the same upstream response the span is built from, in one pass — there is no later moment at which a span's attribution becomes known.

The trace side is different and does have an update consumer (`TraceCostIntelligenceChanged`) precisely because identity *can* appear on a later turn: the session envelope arrives when the hook publishes it, which may be after the first trace of the session.

If a metadata-bearing span patch ever ships, the consumer described here is the right shape. Adding it now would be dead code guarding a wire field that does not exist, and it would have no test that could distinguish it working from it never running.

aadereiko:
baz review

apps/opik-backend/src/main/java/com/comet/opik/domain/CipxTraceIdentityDAO.java
```diff
@@ -200,7 +264,12 @@ private Publisher<? extends Result> insert(List<TraceIdentityRow> rows, String w
                     .bind(index++, row.filesAdded())
                     .bind(index++, row.filesDeleted())
                     .bind(index++, row.linesAdded())
-                    .bind(index++, row.linesDeleted());
+                    .bind(index++, row.linesDeleted())
+                    .bind(index++, row.agentsDispatched())
+                    .bind(index++, row.agentsLinked())
+                    .bind(index++, row.agentsAmbiguous())
+                    .bind(index++, row.cipxVersion())
+                    .bind(index++, ClickHouseDateTimeFormat.formatMicros(row.lastUpdatedAt()));
```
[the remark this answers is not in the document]
aadereiko: Reachable, but not observable — and already better than what it replaced.

**Two ties exist.** A batch repeating a trace id: one event, one `createdAt`, and `dedupTraces` only collapses ids that also carry `lastUpdatedAt`, so a null one is copied verbatim and both rows get a byte-identical version. And two publishes landing in the same microsecond: `BaseEvent.createdAt` is `Instant.now()` on the publishing thread, the bus is an `AsyncEventBus` on virtual threads so nothing orders two events for one trace, and the per-trace Redis lock is released before the create's `doOnSuccess` and absent from both batch paths.

**Neither is observable.** The merge key is `(workspace_id, project_id, user_uuid, trace_id)` — wider than the comment assumes — so a tie needs `user_uuid` to match too; an identity change that moves it produces two independent rows, not a contested one. The counters are session running totals and 000118 mandates `max()` per `session_id`, never `sum()`. Each counts decisions the daemon has made, so each only advances: a dropped snapshot is recovered by the session's next upsert, and the loss survives only at a session's terminal one. And for two snapshots one microsecond apart to differ at all, the daemon must have advanced a counter between two envelopes the backend then published inside that microsecond — these advance on subagent dispatch, a tool round-trip. The residue is a dispatch or two, on counters 000118 rule 2 already documents as mutually approximate (readers clamp `missed` at zero because a dispatch counted ambiguous can later link).

For the batch case the arbitrariness is inherited rather than introduced: a client sending two bodies for one trace id with a null `lastUpdatedAt` has supplied no ordering, and the `traces` table resolves it arbitrarily too. Deduping in the cipx projection would invent an answer the source does not have — and still not agree with the trace row it projects.

**Neither suggested fix is available.** `last_updated_at` is `DateTime64(6,'UTC')` (000100, and 000093 for `cipx_spends`), so microseconds is the column's maximum — `formatMicros` matches the column rather than truncating below it. Nanoseconds means `ALTER MODIFY COLUMN` on the version column of a deployed `ReplicatedReplacingMergeTree`, i.e. a part-rewriting mutation. And ReplacingMergeTree takes exactly one version column, so a deterministic tie-breaker means replacing the version with a cross-replica monotonic sequence, which an async fan-out bus cannot supply.

Worth noting this risk pre-existed in a **worse** form: the column previously used `DEFAULT now64(6)` — the insert clock, same six digits — so a delayed older snapshot arriving later got the *higher* version and reliably won. Binding from the event made the ordering correct (pinned by `reorderedReUpsertKeepsTheNewerSnapshot`) and surfaced a resolution limit that was always the column's. This PR turns "reliably wrong whenever the queue reorders" into "arbitrary inside one microsecond".

A test isn't possible in the form suggested either: under equal versions the engine's choice is arbitrary by design, so it could only assert nondeterminism or pin one build's current behaviour. Declining.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/CostIntelligenceIngestionTest.java
```diff
@@ -431,13 +586,56 @@ void traceCreatedWithIdentityLands() {
...
 
                 assertThat(getUserMappings(email)).containsExactly(userUuid);
             });
 
             assertThat(getCipxIdentity(plainTrace.id(), ws.workspaceId())).isEmpty();
         }
 
+        @Test
+        @DisplayName("a session carrying no agent rollup lands as zeros with an empty cipx_version")
+        void traceWithoutAgentRollupLandsAsZeros() {
```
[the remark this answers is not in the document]
aadereiko: Half done, half declined.

The malformed and fractional cases are now one `@ParameterizedTest` — see the reply on the `:824` comment for the detail.

The missing-rollup case stays separate, because absence and malformation are different outcomes in this code, not two spellings of one. `asUInt32` returns 0 for missing/null **silently** — that is the documented normal state for a daemon that dispatched nothing — while a present-but-unusable value returns 0 **and** is warned about. It also uses a different envelope (`traceCipxMetadataWithoutAgentRollup`: no `agents_*`, no `cipx_version`) and is the only test asserting `cipx_version` is empty, which is the discriminator between "daemon too old to report these" and "session that genuinely dispatched nothing".

Folding it in would put a silent, faithful zero in the same table as two refused ones under a single name, and the name would have to stop saying which it was.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/CostIntelligenceIngestionTest.java
```diff
@@ -508,6 +706,144 @@ void traceIdentityUpdateMergesToSingleRow() {
...
+                assertThat(row.get().agentsAmbiguous()).as("boolean agents_ambiguous records 0").isEqualTo(0L);
+            });
+        }
+
+        @Test
+        @DisplayName("a fractional counter records 0 rather than truncating to a plausible count")
+        void fractionalAgentCountersRecordZero() {
+            var ws = newWorkspace();
+            String projectName = "cipx-" + UUID.randomUUID();
+            String userUuid = UUID.randomUUID().toString();
```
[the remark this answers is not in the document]
aadereiko: Done — one `@ParameterizedTest` with `malformed` and `fractional` as `@CsvSource` rows, matching `writeBlocksInheritSpanCacheTtl` in this same file rather than introducing a new pattern.

The case label is interpolated into every assertion description, which is the part that keeps the consolidation from costing anything: awaitility reports the *last* assertion error on timeout, so that message is what actually names the failing input. Verified both directions —

- dropping `!value.isIntegralNumber()` fails only `[fractional] agents_dispatched=1.5` (1, not 0)
- dropping the `raw < 0L` check fails only `[malformed] agents_linked=-5` (4294967291, not 0)

each with the other case still passing, both mutants compiling and running. The "not 4294967291" / "not 1" prose the old descriptions carried isn't lost either — AssertJ prints the wrong value itself.

Test count is unchanged at 19: two `@Test`s became two invocations of one `@ParameterizedTest`.

apps/opik-backend/src/main/java/com/comet/opik/domain/CipxTraceIdentityDAO.java
```diff
@@ -95,13 +104,60 @@ public static TraceIdentityRow from(UUID traceId, UUID projectId, JsonNode metad
...
+            // Fractional is malformed too, and routed with the text: asLong() would truncate 1.5
+            // to a perfectly plausible 1. canConvertToLong also rejects a bignum past long.
+            if (!value.isIntegralNumber() || !value.canConvertToLong()) {
+                log.warn("cipx metric is not a usable whole number, recording 0: field='{}' type='{}'",
+                        field, value.getNodeType());
+                return 0L;
+            }
+            long raw = value.asLong();
+            if (raw < 0L || raw > 0xFFFF_FFFFL) {
+                log.warn("cipx metric is outside UInt32, recording 0: field='{}' value={}", field, raw);
```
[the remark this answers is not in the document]
aadereiko: Fixed, but as **one aggregated WARN per row** rather than by dropping to `debug`.

Kept at WARN because a *present but unusable* counter is not routine: missing or null already returns 0 silently, and that is the documented normal state for a daemon too old to report these. A value that is present and cannot be one of these counters means the producer sent something wrong, which is exactly what someone chasing a version skew needs to find. This codebase reserves DEBUG for genuinely expected rejections — `JsonProcessingExceptionMapper`, the `CostService` price misses — and those pair with a metric.

The volume point stands though, so the rejections now accumulate per row and emit once, which is the pattern `OnlineScoringEngine` already uses on the feedback-score ingestion path: per-field rejections collected onto the parsed record and reported once per entity, bounded. Same problem shape, same layer, so no new pattern.

A sampled or rate-limited warn was rejected on evidence rather than taste: there is no log rate-limiter, `logOnce` flag or logback throttle anywhere in the backend, and Guava's `RateLimiter` is never imported — inventing one here would be the only instance.

What it looks like now, from the passing run — seven warnings became three, and the field name plus the reason (the whole value of the diagnostic) survive, with `trace_id` added so the offending envelope is fetchable:

```
cipx metrics unusable, recorded 0: trace_id='01a03dea-e25f-...' rejected=[agents_dispatched: not a whole number (STRING), agents_linked: outside UInt32 (-5), agents_ambiguous: not a whole number (BOOLEAN)]
```

Bounded and non-sensitive by construction: at most one entry per call site, each a literal field name plus a node type or a long, never the payload's own text.

apps/opik-backend/src/main/java/com/comet/opik/domain/CipxTraceIdentityDAO.java
```diff
@@ -95,12 +110,65 @@ public static TraceIdentityRow from(UUID traceId, UUID projectId, JsonNode metad
...
+                    .agentsDispatched(asUInt32(session.path("agents_dispatched"), "agents_dispatched", rejected))
+                    .agentsLinked(asUInt32(session.path("agents_linked"), "agents_linked", rejected))
+                    .agentsAmbiguous(asUInt32(session.path("agents_ambiguous"), "agents_ambiguous", rejected))
+                    .cipxVersion(session.path("cipx_version").asText(""))
+                    .lastUpdatedAt(lastUpdatedAt)
                     .build();
+            if (!rejected.isEmpty()) {
+                // Bounded and non-sensitive by construction: at most one entry per call site, each a
+                // literal field name plus a node type or a long — never the payload's own text.
+                log.warn("cipx metrics unusable, recorded 0: trace_id='{}' rejected={}", traceId, rejected);
```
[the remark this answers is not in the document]
aadereiko: Checked this against the code and then ran the exact case you're describing, and the premise doesn't hold — declining, but here's the evidence rather than an assertion.

There are two rejection paths, and only one of them carries a client value:

```java
rejected.add(field + ": not a whole number (" + value.getNodeType() + ")");   // text path
rejected.add(field + ": outside UInt32 (" + raw + ")");                       // numeric path
```

Fed `{"x": "SECRET-PAYLOAD-TEXT-12345", "y": 9999999999, "z": -1, "w": 1.5}` through those two expressions:

```
x: not a whole number (STRING)
y: outside UInt32 (9999999999)
z: outside UInt32 (-1)
w: not a whole number (NUMBER)
```

So on the specific claims:

- **"malformed client-supplied metric payloads ... despite the claim that payload text is never logged"** — a string payload emits `STRING`, the `JsonNodeType` enum name. The node's text never reaches the message. Same for the fractional case: `1.5` logs `NUMBER`, not `1.5`. The claim in the comment holds.
- **`raw` is not payload text.** It is a `long`, and it is only reachable after `isIntegralNumber() && canConvertToLong()` — so by the time it is interpolated it is a whole number inside `long` range, at most 20 characters.
- **"potentially oversized log entries"** — bounded by construction. `rejected` takes at most one entry per `asUInt32` call site, there are 8, and each entry is a literal field name plus either an enum name or a ≤20-character long. Worst case is a few hundred characters.

On dropping `raw` and logging a fixed reason: I'd rather keep it, because it is the part that identifies which upstream bug to go fix. `-1` says a daemon is writing a signed value into an unsigned counter; `9999999999` says something overflowed or a wrong field was mapped. "outside UInt32" alone leaves an operator unable to tell those apart, and these counters feed the attribution SLO — the reason the code records 0 rather than clamping to 4294967295 is precisely to keep a garbage payload distinguishable from a real maximum, and the log value is the other half of that.

Happy to revisit if you had a case in mind where a non-numeric value reaches the numeric branch — I couldn't construct one, since `canConvertToLong()` gates it.

apps/opik-backend/src/main/java/com/comet/opik/domain/CipxSpendDAO.java
```diff
@@ -84,6 +90,24 @@ public static SpanRow from(UUID spanId, UUID traceId, UUID projectId, JsonNode m
                     .maxTokens(config.path("max_tokens").asLong(0))
                     .contextManagement(config.path("context_management").asText(""))
                     .speed(config.path("speed").asText(""))
+                    // Attribution fields. trigger_detail carries the subagent NAME when
```
LifeXplorer: should these comments be visible to OSS Opik users?

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000119_add_cipx_call_attribution.sql
```diff
@@ -0,0 +1,88 @@
+--liquibase formatted sql
+--changeset aadereiko:000119_add_cipx_call_attribution
+--comment: Persist cipx sub-agent attribution: per-call trigger/detail/turn/parent on cipx_spends, and the session-grain link rollup on cipx_trace_identities
+--
+-- Additive columns with defaults, so a backend that predates this changeset never
+-- names them and one that postdates it reads '' / 0 on pre-existing rows.
+--
+-- cipx_spends, per LLM call:
+--
+--   trigger              which loop the call came from: user_turn | tool_continuation
```
LifeXplorer: same with comments visibility for OSS version

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000119_add_cipx_call_attribution.sql
```diff
@@ -0,0 +1,88 @@
...
+--      totals re-stamped on every trace upsert, so a session with N traces holds N
+--      successive snapshots and sum() multiplies by the trace count.
+--   2. Clamp missed = dispatched - ambiguous - linked at zero. The producer's
+--      dispatch log is capped per session and evicts oldest-first, so past the cap
+--      a dispatch counted ambiguous can later link and missed can go negative.
+--   3. rate = linked / (dispatched - ambiguous) is UNDEFINED, not zero, at a zero
+--      denominator. Reporting 0% invents a failure out of an idle session.
+--   4. Zero is not "unknown". Absence reads as zero, which is faithful for a
+--      daemon that dispatched nothing; one too old to emit them is told apart by
+--      cipx_version / schema_version.
```
BorisTkachenko: Let's remove this detailed comment from here. Mainly because we do not want to explain the details of cost intelligence in the Opik source repo.

apps/opik-backend/src/main/java/com/comet/opik/domain/CipxSpendDAO.java
```diff
@@ -84,6 +90,24 @@ public static SpanRow from(UUID spanId, UUID traceId, UUID projectId, JsonNode m
...
+                    // invocation this call belongs to, which is the parent/child edge of the agent
+                    // tree. All default to "" when the proxy could not resolve them — an empty value
+                    // means unknown and must not be substituted for a guessed agent name.
+                    //
+                    // link_failure_reason says WHY a call cipx identified as a subagent has no
+                    // parent, which is the only thing that separates cipx correctly fail-closing on
+                    // two byte-identical peer dispatches (ambiguous_prompt) from cipx losing a
+                    // dispatch (no_dispatch_captured). Both land in the same "(unattributed)" bucket
+                    // downstream without it. cipx stamps it only on trigger=subagent calls, so ""
+                    // elsewhere means "nothing to report", not "resolved".
```
BorisTkachenko: Same, let's not provide all these details here in open source repo

apps/opik-backend/src/main/java/com/comet/opik/domain/CipxTraceIdentityDAO.java
```diff
@@ -95,12 +110,65 @@ public static TraceIdentityRow from(UUID traceId, UUID projectId, JsonNode metad
...
+                    // Session-grain subagent link rollup. These counters are
+                    // the only place cipx's worst attribution failure is
+                    // visible: a subagent whose dispatch was never observed
+                    // looks exactly like a main-loop turn, so no span carries
+                    // a link_failure_reason and only a missing increment here
+                    // reveals it. They are session RUNNING TOTALS re-stamped
+                    // on every trace upsert of that session, so a session
+                    // with N traces leaves N rows holding N successive
+                    // snapshots. Readers must take max() per session_id,
+                    // never sum().
```
BorisTkachenko: Same here

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/CostIntelligenceIngestionTest.java
```diff
@@ -149,6 +157,17 @@ void spanCreatedWithCipxCallLands() {
                 assertThat(row.get().contextManagement()).isEqualTo("clear_thinking_20251015");
                 // speed: selects the rate table, so it must survive ingestion
                 assertThat(row.get().speed()).isEqualTo("fast");
+                // attribution fields: what caused the call, which agent ran it, which turn it
+                // belongs to, and which Agent tool_use spawned it. Without these the spend tables
+                // cannot separate subagent spend from the main agent's, nor name the agent.
```
BorisTkachenko: Same

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/CostIntelligenceIngestionTest.java
```diff
@@ -149,6 +157,17 @@ void spanCreatedWithCipxCallLands() {
...
+                assertThat(row.get().trigger()).isEqualTo("subagent");
+                assertThat(row.get().triggerDetail()).isEqualTo("code-reviewer");
+                assertThat(row.get().turnKey()).isEqualTo("abc123turnkey");
+                assertThat(row.get().parentToolUseId()).isEqualTo("toolu_parent_agent");
+                // This subagent linked, so cipx ships no link_failure_reason at all and the column
+                // must read '' — "nothing to report", which is the healthy state. cipx only stamps
+                // a reason on a subagent whose parent it could NOT resolve.
```
BorisTkachenko: Same

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/CostIntelligenceIngestionTest.java
```diff
@@ -162,6 +181,142 @@ void spanCreatedWithCipxCallLands() {
...
+            String projectName = "cipx-" + UUID.randomUUID();
+
+            // systemToolsCipxMetadata carries no
+            // trigger/turn_key/parent_tool_use_id — the shape every span
+            // written before the proxy shipped them has. Ingestion must still
+            // land the row and must leave the attribution columns empty
+            // rather than substituting a default: "" means unknown, and a
+            // guessed agent name would book real spend against an agent that
+            // never ran. Scope, deliberately: this covers the half that is
+            // ours — the DAO supplying "" for a field the metadata omits.
```
BorisTkachenko: Same

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/CostIntelligenceIngestionTest.java
```diff
@@ -162,6 +181,142 @@ void spanCreatedWithCipxCallLands() {
...
+        void unattributedSubagentsKeepTheirLinkFailureReason() {
+            var ws = newWorkspace();
+            String projectName = "cipx-" + UUID.randomUUID();
+
+            // Both spans are subagent calls with NO parent_tool_use_id, so a per-agent rollup files
+            // both under "(unattributed)". They are not the same problem: ambiguous_prompt is cipx
+            // deliberately refusing to guess between two byte-identical peer dispatches (working as
+            // designed, nothing to chase), while no_dispatch_captured is cipx losing a dispatch it
+            // should have observed (a real defect). link_failure_reason is the only thing that keeps
+            // them apart — without it a live incident and normal operation look identical.
```
BorisTkachenko: Same

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/CostIntelligenceIngestionTest.java
```diff
@@ -162,6 +181,142 @@ void spanCreatedWithCipxCallLands() {
...
+        @DisplayName("a call carrying fields the DAO does not know still ingests")
+        void unknownFieldsDoNotRejectTheRow() {
+            var ws = newWorkspace();
+            String projectName = "cipx-" + UUID.randomUUID();
+
+            // cipx adds fields to metadata.cipx.call on its own release cadence and ships to laptops
+            // independently of this service, so a newer proxy talking to an older backend is the
+            // normal state, not an edge case. Ingestion must ignore what it does not recognize
+            // rather than reject the row — dropping the row would lose the spend entirely, and spend
+            // totals are the one thing that is correct today.
```
BorisTkachenko: Same

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/CostIntelligenceIngestionTest.java
```diff
@@ -431,13 +586,56 @@ void traceCreatedWithIdentityLands() {
                 assertThat(row.get().filesDeleted()).isEqualTo(1);
                 assertThat(row.get().linesAdded()).isEqualTo(40);
                 assertThat(row.get().linesDeleted()).isEqualTo(5);
+                // Session-grain subagent link rollup parsed from cipx.session. Every value is
+                // distinct from every other integer above, so a positional-bind rotation in
+                // CipxTraceIdentityDAO shows up here as a value reported under the wrong name.
+                // These are session running totals re-stamped on each trace of the session — a
+                // reader aggregates them with max() per session_id, never sum() (migration 000119).
```
BorisTkachenko: Same

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/CostIntelligenceIngestionTest.java
```diff
@@ -431,13 +586,56 @@ void traceCreatedWithIdentityLands() {
...
+            var ws = newWorkspace();
+            String projectName = "cipx-" + UUID.randomUUID();
+            String userUuid = UUID.randomUUID().toString();
+            String email = "dev-" + UUID.randomUUID() + "@acme.com";
+
+            // A proxy older than these fields is the normal state, not an edge case: cipx ships to
+            // laptops on its own cadence. The counters are omitempty on the wire and absence reads
+            // as zero by design, so 0 here is faithful — it is NOT a "we don't know" sentinel. The
+            // empty cipx_version is what lets a reader tell this row (a daemon too old to report)
+            // from a genuine session that dispatched no agents.
```
BorisTkachenko: Same

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/CostIntelligenceIngestionTest.java
```diff
@@ -508,6 +706,129 @@ void traceIdentityUpdateMergesToSingleRow() {
...
+        @DisplayName("every positional bind lands in its own column, across a multi-row identity insert")
+        void everyIdentityPositionalBindLandsInItsOwnColumn() {
+            var ws = newWorkspace();
+
+            // CipxTraceIdentityDAO binds 30 parameters per row by position and
+            // nothing checks that order against the INSERT tuple at compile
+            // time. A mismatch does not fail the insert — it writes each value
+            // into the neighbouring column. Two rows, deliberately: the bind
+            // index accumulates across rows, so a wrong stride corrupts the
+            // second tuple while a single-row test still passes.
```
BorisTkachenko: Same and in other places in this file below

BorisTkachenko (commented):
Code looks good, but let's remove most of comments, as they leak cost intelligence details which we do not want to share in open source.

apps/opik-backend/src/main/java/com/comet/opik/domain/CipxSpendDAO.java
aadereiko: Good catch, and no — they shouldn't be. Removed in b27c5b5.

Every comment this PR added is gone: 236 lines across the four files. The migration now carries only its liquibase directives and the two ALTER statements (the SQL itself is byte-identical), and the Java files keep no explanatory prose at all.

I deliberately did not touch comments that were already on main — there are a few pre-existing `cipx` references in the test file's fixture labels, for instance — since removing those would widen this PR's diff into code it didn't write. Happy to do that as a separate cleanup if you want the whole file scrubbed rather than just my additions.

Verified after the removal: CostIntelligenceIngestionTest 19/19, ChangelogRebaselineTest 2/2.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000119_add_cipx_call_attribution.sql
aadereiko: Agreed — removed in b27c5b5, and not just here: every comment this PR added across all four files is gone, 236 lines in total.

For this file specifically it was the worst offender — 74 of its 88 lines were prose. It now holds the liquibase directives and the two ALTER statements and nothing else, 20 lines, with the SQL byte-identical to before.

Two notes on scope. I left comments that were already on main untouched, so the diff doesn't widen into code this PR didn't write — say the word if you'd rather the whole file were scrubbed. And rewriting the changeset changes its checksum; that's safe because it has never been applied anywhere outside a laptop, so there's no ledger to reconcile. Re-verified with ChangelogRebaselineTest (2/2) that it still applies against a real ClickHouse.

apps/opik-backend/src/main/resources/liquibase/db-app-analytics/migrations/000119_add_cipx_call_attribution.sql
aadereiko: Removed in b27c5b5 — every comment this PR added is gone across all four files (236 lines). Comments already on main were left alone so the diff doesn't widen; happy to scrub those separately if you'd prefer.

apps/opik-backend/src/main/java/com/comet/opik/domain/CipxSpendDAO.java
aadereiko: Removed in b27c5b5 — every comment this PR added is gone across all four files (236 lines). Comments already on main were left alone so the diff doesn't widen; happy to scrub those separately if you'd prefer.

apps/opik-backend/src/main/java/com/comet/opik/domain/CipxTraceIdentityDAO.java
aadereiko: Removed in b27c5b5 — every comment this PR added is gone across all four files (236 lines). Comments already on main were left alone so the diff doesn't widen; happy to scrub those separately if you'd prefer.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/CostIntelligenceIngestionTest.java
aadereiko: Removed in b27c5b5 — every comment this PR added is gone across all four files (236 lines). Comments already on main were left alone so the diff doesn't widen; happy to scrub those separately if you'd prefer.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/CostIntelligenceIngestionTest.java
aadereiko: Removed in b27c5b5 — every comment this PR added is gone across all four files (236 lines). Comments already on main were left alone so the diff doesn't widen; happy to scrub those separately if you'd prefer.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/CostIntelligenceIngestionTest.java
aadereiko: Removed in b27c5b5 — every comment this PR added is gone across all four files (236 lines). Comments already on main were left alone so the diff doesn't widen; happy to scrub those separately if you'd prefer.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/CostIntelligenceIngestionTest.java
aadereiko: Removed in b27c5b5 — every comment this PR added is gone across all four files (236 lines). Comments already on main were left alone so the diff doesn't widen; happy to scrub those separately if you'd prefer.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/CostIntelligenceIngestionTest.java
aadereiko: Removed in b27c5b5 — every comment this PR added is gone across all four files (236 lines). Comments already on main were left alone so the diff doesn't widen; happy to scrub those separately if you'd prefer.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/CostIntelligenceIngestionTest.java
aadereiko: Removed in b27c5b5 — every comment this PR added is gone across all four files (236 lines). Comments already on main were left alone so the diff doesn't widen; happy to scrub those separately if you'd prefer.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/CostIntelligenceIngestionTest.java
aadereiko: Removed in b27c5b5 — every comment this PR added is gone across all four files (236 lines). Comments already on main were left alone so the diff doesn't widen; happy to scrub those separately if you'd prefer.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/CostIntelligenceIngestionTest.java
aadereiko: Removed in b27c5b5 — every comment this PR added is gone across all four files (236 lines). Comments already on main were left alone so the diff doesn't widen; happy to scrub those separately if you'd prefer.

BorisTkachenko (approved):
LGTM!

LifeXplorer (approved):
Looks good!

PR #7950, opened 2026-08-21
[NA] [BE] feat: enable agentic-tools by default for all deployments

aadereiko:
## Details

The agentic-tools path on LLM-as-judge online scoring only ever reached Comet-hosted environments, which pin `TOGGLE_AGENTIC_TOOLS_ENABLED=true` in their chart values. Self-hosted installs inherit the image's `config.yml` default, and that default was still `"false"` — so neither the docker-compose nor the helm install path ever turned the feature on. Enabling it for self-hosted was requested; this flips `serviceToggles.agenticToolsEnabled` in `apps/opik-backend/config.yml` from `"false"` to `"true"`.

- That single line covers every self-hosted install because **neither deployment path names this variable**. `deployment/docker-compose/docker-compose.yaml` sets only `TOGGLE_OLLIE_ENABLED`, `TOGGLE_OPIK_AI_ENABLED`, `TOGGLE_GUARDRAILS_ENABLED` and `TOGGLE_WELCOME_WIZARD_ENABLED`; the helm chart's `component.backend.env` does not list it either.
- `configmap-backend.yaml` only emits keys present in `component.backend.env`, so an absent key means no env var in the pod and the image default applies. Operators keep full control — adding `component.backend.env.TOGGLE_AGENTIC_TOOLS_ENABLED: "false"` still wins — while the image retains the ability to move the default, which is exactly the property a chart-pinned value would destroy. Hence **no chart change here**.
- `config-test.yml` mirrors the flip so app-booting tests exercise the path real deployments now run. Previously CI validated the inline path while we were about to ship the agentic one as the default.
- Deliberately unchanged — the frontend `DEFAULT_STATE` (`feature-toggles-provider.tsx`) stays `false`. It only applies when the backend *omits* the key, i.e. a newer FE against an older BE with no agentic support, so flipping it would enable UI affordances the backend can't honour.
- Deliberately unchanged — helm `values.yaml`, for the reason above. No `TOGGLE_*` key appears in the generated chart README for any component either, so listing just this one would be inconsistent; documenting all toggles is a separate change.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: Investigated which deployment paths pin the toggle, authored the two-line default flip (`config.yml`, `config-test.yml`), and ran/triaged the backend test suites below.
- Human verification: Code review of the diff; the deployment-path claims (docker-compose env list, helm `component.backend.env` → `configmap-backend.yaml`) re-checked against the files in-tree.

## Testing

Backend suites run locally against the flipped `config-test.yml`, each as `mvn -Dtest=<Class> test` from `apps/opik-backend`:

| Class | Result |
|---|---|
| `TraceThreadOnlineScoringAgenticToolsE2ETest` | ✅ 1/1 |
| `TraceThreadOnlineScoringSamplerListenerIntegrationTest` | ✅ 6/6 |
| `OnlineScoringSpanSamplerIntegrationTest` | ✅ 2/2 |
| `AutomationRuleEvaluatorsResourceTest` | 135/136 — one flaky `ConditionTimeout`, see below |
| `ManualEvaluationResourceTest` | ✅ 20/20 |

Scenarios covered: the agentic path now taken by default (E2E), the sampler/listener path that routes by size, and the automation-rule and manual-evaluation resource surfaces that build scorer context.

The `AutomationRuleEvaluatorsResourceTest` failure is not attributable to this change, on three grounds:

1. It landed on a **different test each run** — `getLogsUserDefinedMetricPythonScorer`, then `getLogsTraceSkippedDueToDisabledRule` — each passing in the other run.
2. Single-method A/B runs passed with the toggle at both `false` and `true`.
3. The second failure's test creates both rules with `enabled(false)`, so no scorer is ever invoked and a flag governing scorer context-building cannot reach it.

Every failure had the shape `Expected size: N but was: 0 … within 10 seconds` — zero logs, i.e. the Awaitility window losing a race under class-level load, not a behavioural divergence (which would emit logs with diverging *content*).

Not run: frontend and SDK suites — neither is touched by a backend config default.

## Documentation

N/A — no docs or chart README reference `TOGGLE_AGENTIC_TOOLS_ENABLED` today (`grep` over `*.md`/`*.mdx`/chart values returns only the `config.yml` definition itself), so there is no documented default to correct.

### DIFF
--- a/apps/opik-backend/config.yml
+++ b/apps/opik-backend/config.yml
@@ -1468,10 +1468,12 @@ serviceToggles:
   # Default: false
   # Description: Whether the project homepage is enabled. When false, the Ollie page is the default project page.
   projectHomepageEnabled: ${TOGGLE_PROJECT_HOMEPAGE_ENABLED:-"false"}
-  # Default: false
+  # Default: true
   # Description: Master switch for the agentic-tools path on LLM-as-judge online scoring. When false, the
   #   inline path is used regardless of context size. Threshold lives under onlineScoring.agenticToolsThresholdTokens.
-  agenticToolsEnabled: ${TOGGLE_AGENTIC_TOOLS_ENABLED:-"false"}
+  #   On by default for every deployment, self-hosted included; set TOGGLE_AGENTIC_TOOLS_ENABLED=false to force
+  #   the inline path (e.g. to keep online scoring on a single LLM round-trip).
+  agenticToolsEnabled: ${TOGGLE_AGENTIC_TOOLS_ENABLED:-"true"}
   # Default: true
   # Description: Whether the online-scoring (LLM-as-judge) evaluation loop is persisted as monitoring traces/spans.
   #   When on, each evaluation produces one trace (source=evaluator) with one llm span per LLM round, so cost and
--- a/apps/opik-backend/src/test/resources/config-test.yml
+++ b/apps/opik-backend/src/test/resources/config-test.yml
@@ -822,9 +822,11 @@ serviceToggles:
   #   freeform SQL path and the clickhouse-readonly-freeform-sql health check. Tests flip this on when exercising
   #   the read-only ClickHouse client.
   ollieEnabled: "false"
-  # Default: false
-  # Description: Master switch for the agentic-tools path on LLM-as-judge online scoring.
-  agenticToolsEnabled: "false"
+  # Default: true
+  # Description: Master switch for the agentic-tools path on LLM-as-judge online scoring. Mirrors the
+  #   shipped default in config.yml so app-booting tests exercise the path real deployments run. Tests
+  #   that need the inline path flip this off via customConfigs.
+  agenticToolsEnabled: "true"
   # Default: true
   # Description: Whether the online-scoring (LLM-as-judge) evaluation loop is persisted as monitoring traces.
   onlineScoringTracingEnabled: "true"


CometActions:

Flipping agenticToolsEnabled to true changes what a default OSS install does: GET /v1/private/toggles/ now returns agentic_tools_enabled: true, and the LLM-judge rule dialog reads that toggle — {{spans}} / {{trace}} in a trace-scope prompt now auto-fill as reserved sentinels instead of staying empty, user-mappable rows. online-evaluation-smoke.spec.ts is tagged @cap:online-evaluation.create-llm-judge-rule, but it builds the rule from the canned Moderation template, which has no reserved variables, and its 5 seeded traces are nowhere near the 50k-token routing threshold — it passes identically with the toggle on or off, and nothing else under tests_end_to_end/ mentions spans, trace or the toggles endpoint. Worth pinning the dialog behaviour and the toggle default. The agentic-tools loop itself needs a provider key and a 50k-token trace, so that part stays manual either way.

Would target `online-evaluation.create-llm-judge-rule`.

1. On a stock install with no env overrides, GET /v1/private/toggles/ and confirm agentic_tools_enabled is true (it was false before this PR)
2. Open a project > Online evaluation > Create rule > LLM-as-judge, trace scope, and replace the prompt with one containing {{spans}} and {{trace}}; confirm both variables appear pre-mapped to their sentinels instead of as empty editable rows
3. In the same dialog, add a non-reserved variable like {{input}} and confirm it still renders as a normal empty row the user maps by hand — the auto-fill must be limited to the reserved names
4. Switch the dialog to span scope and confirm {{span}} auto-fills while {{spans}}/{{trace}} do not; switch to thread scope and confirm no variables section renders at all
5. Save a rule carrying {{spans}}, reopen it for edit, and confirm the persisted variables.spans mapping survives the round-trip rather than being re-derived or cleared

</details>

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/32492325160)

CometActions:

### Access Information
- **URL:** https://pr-7950.dev.comet.com
- **Namespace:** pr-7950
- **Version:** 2.2.37-7950-merge-3036
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7950%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7950`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

apps/opik-backend/config.yml
```diff
@@ -1468,10 +1468,12 @@ serviceToggles:
   # Default: false
   # Description: Whether the project homepage is enabled. When false, the Ollie page is the default project page.
   projectHomepageEnabled: ${TOGGLE_PROJECT_HOMEPAGE_ENABLED:-"false"}
-  # Default: false
+  # Default: true
```
andrescrz, line 1471: If this always defaults to true now, is the are reason to keep this flag? Should we remove it all together?

andrescrz (commented):
Double check my comment. Please also reply or address all Baz comments.

andrescrz:
Closed in favour of https://github.com/comet-ml/opik/pull/7991/

PR #7957, opened 2026-08-21
[OPIK-8040] [BE] fix: apply online evaluation sampling to production traces only

jverre:
## Details

An online evaluation rule holds one sampling rate, and it applied to every trace the rule saw. A user who wanted to sample production traffic but score every experiment run could not do it, so they had to keep two copies of the same rule. This makes the sampling rate a production-only setting: `OnlineScoringSampler` now skips the sampling roll for any trace whose source is not SDK, so experiment, playground and optimization traces are always scored in full.

- The bypass sits after the enabled and filter checks, so a disabled rule and its filters still apply to experiment traces — only the random roll is skipped.
- Each bypassed trace gets one line on the rule's user-facing log stream saying the rate applies to production traces only, so a user can see why every run was scored.
- No change was needed in the span sampler or the thread sampler: both already keep SDK-source entities only, so no experiment span or thread was ever sampled.

Only the sampling half of the ticket is in this PR. The other suggested improvement — built-in `{{trace.output}}` / `{{trace.input}}` / `{{trace.metadata}}` placeholders — is not addressed here.

## Change checklist
- [x] User facing
- [x] Documentation update

## Issues

- OPIK-8040

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: whole change — backend logic, unit tests, API schema descriptions, docs line
- Human verification: pending review by the author

## Testing

- `mvn test -Dtest=OnlineScoringSamplerTest` — 50 tests, 0 failures.
- `mvn test -Dtest='OnlineScoringSpanSamplerTest,TraceThreadOnlineScoringSamplerListenerTest,AutomationRuleEvaluatorFiltersDeserializerTest'` — 37 tests, 0 failures. These cover the two samplers the change deliberately leaves alone.
- `mvn spotless:apply` then `mvn spotless:check` — clean.

Scenarios added to `OnlineScoringSamplerTest.SamplingRateTests`:

- An experiment trace is scored at sampling rate `0.0`, for both the LLM-as-a-judge and the Python evaluator.
- A batch of one SDK trace and one experiment trace at rate `0.0` enqueues the experiment trace only.
- A playground, optimization or evaluator trace carrying `selected_rule_ids` is scored at rate `0.0` (parameterized over the three sources).
- A disabled rule still scores nothing on an experiment trace.
- A rule whose filters do not match still scores nothing on an experiment trace.

Not run: the backend integration and E2E suites, which need MySQL, ClickHouse and Redis. The behaviour under test is pure in-process sampler logic and is covered by the unit tests above.

## Documentation

- Rewrote the sampling-rate bullet on the online evaluation rules page to say it covers production traces, and that experiment and playground traces are always scored in full.
- Added a `@Schema` description on `samplingRate` in `AutomationRuleEvaluator` and `AutomationRuleEvaluatorUpdate`, so the generated OpenAPI spec and the SDKs carry the same statement. The committed `openapi/opik.yaml` is regenerated by the `sdks_generate_openapi_spec_and_fern_code` workflow, so it is not touched here.
- The v1 docs version keeps the old wording. It does not document the rule trigger scope at all, so it already predates the production/experiment split.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/evaluators/AutomationRuleEvaluator.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/evaluators/AutomationRuleEvaluator.java
@@ -80,6 +80,7 @@ public abstract sealed class AutomationRuleEvaluator<T, E extends Filter> implem
     @NotBlank @Size(max = 150, message = "cannot exceed 150 characters") private final String name;
 
     @JsonView({View.Public.class, View.Write.class})
+    @Schema(description = "Fraction of production (SDK-logged) traces the rule scores, from 0 to 1. Experiment, playground and optimization traces are always scored in full and ignore this value.")
     private final float samplingRate;
 
     @JsonView({View.Public.class, View.Write.class})
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/evaluators/AutomationRuleEvaluator.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/evaluators/AutomationRuleEvaluator.java
@@ -82,7 +82,7 @@ public abstract sealed class AutomationRuleEvaluator<T, E extends Filter> implem
     @NotBlank @Size(max = 150, message = "cannot exceed 150 characters") private final String name;
 
     @JsonView({View.Public.class, View.Write.class})
-    @Schema(description = "Fraction of production (SDK-logged) traces the rule scores, from 0 to 1. Experiment, playground and optimization traces are always scored in full and ignore this value.")
+    @Schema(description = "Fraction of production (SDK-logged) items this rule scores, from 0 to 1. Trace rules ignore this value for experiment, playground and optimization traces and score them in full; span and thread rules only ever evaluate SDK-logged data.")
     @DecimalMin("0") @DecimalMax("1") private final float samplingRate;
 
     @JsonView({View.Public.class, View.Write.class})
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/evaluators/AutomationRuleEvaluator.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/evaluators/AutomationRuleEvaluator.java
@@ -12,6 +12,8 @@ import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import io.swagger.v3.oas.annotations.media.DiscriminatorMapping;
 import io.swagger.v3.oas.annotations.media.Schema;
 import jakarta.validation.Valid;
+import jakarta.validation.constraints.DecimalMax;
+import jakarta.validation.constraints.DecimalMin;
 import jakarta.validation.constraints.NotBlank;
 import jakarta.validation.constraints.NotNull;
 import jakarta.validation.constraints.Size;
@@ -81,7 +83,7 @@ public abstract sealed class AutomationRuleEvaluator<T, E extends Filter> implem
 
     @JsonView({View.Public.class, View.Write.class})
     @Schema(description = "Fraction of production (SDK-logged) traces the rule scores, from 0 to 1. Experiment, playground and optimization traces are always scored in full and ignore this value.")
-    private final float samplingRate;
+    @DecimalMin("0") @DecimalMax("1") private final float samplingRate;
 
     @JsonView({View.Public.class, View.Write.class})
     @Builder.Default
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/evaluators/AutomationRuleEvaluatorUpdate.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/evaluators/AutomationRuleEvaluatorUpdate.java
@@ -54,6 +54,7 @@ public abstract sealed class AutomationRuleEvaluatorUpdate<T, E extends Filter> 
     // the API boundary instead of failing the update (OPIK-7371).
     @NotBlank @Size(max = 150, message = "cannot exceed 150 characters") private final String name;
 
+    @Schema(description = "Fraction of production (SDK-logged) traces the rule scores, from 0 to 1. Experiment, playground and optimization traces are always scored in full and ignore this value.")
     private final float samplingRate;
 
     @Builder.Default
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/evaluators/AutomationRuleEvaluatorUpdate.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/evaluators/AutomationRuleEvaluatorUpdate.java
@@ -56,7 +56,7 @@ public abstract sealed class AutomationRuleEvaluatorUpdate<T, E extends Filter> 
     // the API boundary instead of failing the update (OPIK-7371).
     @NotBlank @Size(max = 150, message = "cannot exceed 150 characters") private final String name;
 
-    @Schema(description = "Fraction of production (SDK-logged) traces the rule scores, from 0 to 1. Experiment, playground and optimization traces are always scored in full and ignore this value.")
+    @Schema(description = "Fraction of production (SDK-logged) items this rule scores, from 0 to 1. Trace rules ignore this value for experiment, playground and optimization traces and score them in full; span and thread rules only ever evaluate SDK-logged data.")
     @DecimalMin("0") @DecimalMax("1") private final float samplingRate;
 
     @Builder.Default
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/evaluators/AutomationRuleEvaluatorUpdate.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/evaluators/AutomationRuleEvaluatorUpdate.java
@@ -10,6 +10,8 @@ import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import io.swagger.v3.oas.annotations.media.DiscriminatorMapping;
 import io.swagger.v3.oas.annotations.media.Schema;
 import jakarta.validation.Valid;
+import jakarta.validation.constraints.DecimalMax;
+import jakarta.validation.constraints.DecimalMin;
 import jakarta.validation.constraints.NotBlank;
 import jakarta.validation.constraints.NotNull;
 import jakarta.validation.constraints.Size;
@@ -55,7 +57,7 @@ public abstract sealed class AutomationRuleEvaluatorUpdate<T, E extends Filter> 
     @NotBlank @Size(max = 150, message = "cannot exceed 150 characters") private final String name;
 
     @Schema(description = "Fraction of production (SDK-logged) traces the rule scores, from 0 to 1. Experiment, playground and optimization traces are always scored in full and ignore this value.")
-    private final float samplingRate;
+    @DecimalMin("0") @DecimalMax("1") private final float samplingRate;
 
     @Builder.Default
     private final boolean enabled = true;
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSampler.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSampler.java
@@ -362,11 +362,6 @@ public class OnlineScoringSampler {
                     trace.id(), evaluator.getName());
         }
 
-        // The sampling rate thins a continuous production stream, so it applies to SDK-logged traces
-        // only. Experiment, playground and optimization traces are runs the user started one by one and
-        // expects to see scored in full, so they bypass the roll whatever the rule's rate (OPIK-8040).
-        // Without this, a rule that covers both sides forces the user to keep two copies of it: one
-        // sampled for production and one at 100% for experiments.
         if (!Source.isLoggingSource(trace.source())) {
             logForUser(workspaceId, evaluator, trace,
                     "The traceId '{}' is scored for rule: '{}' regardless of the sampling rate '{}',"
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSampler.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSampler.java
@@ -127,11 +127,19 @@ public class OnlineScoringSampler {
     private boolean skip(String workspaceId, String workspaceName, AutomationRuleEvaluator<?, ?> evaluator, Trace trace,
             String decision, String message, Object... args) {
         recordDecision(workspaceId, workspaceName, evaluator, decision, 1);
+        logForUser(workspaceId, evaluator, trace, message, args);
+        return false;
+    }
+
+    /**
+     * Emits one line on the rule's user-facing log stream for the given trace.
+     */
+    private void logForUser(String workspaceId, AutomationRuleEvaluator<?, ?> evaluator, Trace trace,
+            String message, Object... args) {
         // Important to set the workspaceId for logging purposes
         try (var logContext = createTraceLoggingContext(workspaceId, evaluator, trace)) {
             userFacingLogger.info(message, args);
         }
-        return false;
     }
 
     /**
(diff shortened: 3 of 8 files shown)


apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSampler.java
```diff
@@ -354,6 +362,19 @@ private boolean shouldSampleTrace(AutomationRuleEvaluator<?, ?> evaluator, Strin
...
+        // only. Experiment, playground and optimization traces are runs the user started one by one and
+        // expects to see scored in full, so they bypass the roll whatever the rule's rate (OPIK-8040).
+        // Without this, a rule that covers both sides forces the user to keep two copies of it: one
+        // sampled for production and one at 100% for experiments.
+        if (!Source.isLoggingSource(trace.source())) {
+            logForUser(workspaceId, evaluator, trace,
+                    "The traceId '{}' is scored for rule: '{}' regardless of the sampling rate '{}',"
+                            + " as the rate applies to production traces only",
+                    trace.id(), evaluator.getName(), evaluator.getSamplingRate());
+            return true;
```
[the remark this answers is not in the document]
jverre: not an issue

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/OnlineScoringSamplerTest.java
```diff
@@ -475,6 +475,87 @@ void samplesAllTracesWhenSamplingRateIsOne() {
...
+        void scoresExperimentTracesWhenSamplingRateIsZeroForPythonEvaluator() {
+            when(serviceTogglesConfig.isPythonEvaluatorEnabled()).thenReturn(true);
+            var trace = createTrace(Source.EXPERIMENT);
+            var evaluator = createPythonEvaluator(0.0f);
+            whenFindAllPythonEvaluators(evaluator);
+
+            onlineScoringSampler.onTracesCreated(new TracesCreated(List.of(trace), workspaceId, userName));
+
+            verify(onlineScorePublisher).enqueueMessage(any(),
+                    eq(AutomationRuleEvaluatorType.USER_DEFINED_METRIC_PYTHON));
```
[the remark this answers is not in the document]
jverre: not an issue

CometActions:

OnlineScoringSampler now returns early for any non-SDK trace, so on a rule with trigger scope `both` an experiment/playground trace flips from "sampled at the configured rate" to "always scored" — a visible change in which traces carry the rule's feedback score. Our only sampling spec, online-evaluation-sampling-rate.spec.ts, seeds its whole 30-trace batch through the Python SDK, so every trace in it is still a logging source and the spec passes identically before and after; nothing exercises the new branch, and nothing covers the new @DecimalMin(0)/@DecimalMax(1) bound on samplingRate either. `online-evaluation.sampling-rate` is marked covered, but by that SDK-only test. Worth one deterministic test that puts both sources through a single rule.

Would target `online-evaluation.sampling-rate`.

1. Create a deterministic Python-Equals rule with trigger scope 'both' and sampling rate 0%, then seed matching traces via POST /v1/private/traces with source 'sdk' and source 'experiment'
2. Confirm every experiment-source trace gets the rule's score and no SDK-source trace does — at rate 0 this is exact, no binomial band needed
3. Repeat with source 'playground' on a default production-scope rule, since matchesTriggerScope lets playground/optimization through and they now bypass sampling too
4. Check the rule's automation logs shows the new 'is not subject to the sampling rate ... applies to production traces only' line for the exempt traces
5. POST and PATCH a rule with samplingRate -0.1 and 1.5 and confirm the new 0..1 bounds reject them, and that the dialog's percent field cannot produce one

</details>

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/32754903311)

_Re-checked after a push on 24 Aug 17:12 UTC._

apps/opik-backend/src/main/java/com/comet/opik/api/evaluators/AutomationRuleEvaluator.java
```diff
@@ -80,7 +82,8 @@ public abstract sealed class AutomationRuleEvaluator<T, E extends Filter> implem
     @NotBlank @Size(max = 150, message = "cannot exceed 150 characters") private final String name;
 
     @JsonView({View.Public.class, View.Write.class})
-    private final float samplingRate;
+    @Schema(description = "Fraction of production (SDK-logged) traces the rule scores, from 0 to 1. Experiment, playground and optimization traces are always scored in full and ignore this value.")
```
thiagohora: **[medium] Schema description contradicts real behavior for span and trace-thread rule types**

The new description says non-SDK items are "always scored in full and ignore this value", but it sits on the shared sealed base class, so it is emitted into the OpenAPI schema of all six subtypes (`AutomationRuleEvaluatorSpanLlmAsJudge`, `...SpanUserDefinedMetricPython`, `...TraceThreadLlmAsJudge`, `...TraceThreadUserDefinedMetricPython` included). For those four subtypes the behavior is the exact opposite: `OnlineScoringSpanSampler.onSpansCreated` drops non-SDK spans before any rule is applied (`scorableSpans = spans.stream().filter(span -> Source.isLoggingSource(span.source()))`, OnlineScoringSpanSampler.java:92) and `TraceThreadOnlineScoringSamplerListener` filters threads the same way (line 91) — experiment/playground/optimization items are never scored, not "scored in full". Only the trace-level sampler has the bypass (`OnlineScoringSampler.java:365-371`). The wording "traces" is also wrong for span and thread rules, which score spans and threads.

💡 Keep the base-class description generic (e.g. "Fraction of production (SDK-logged) items this rule scores, from 0 to 1.") and move the source-bypass sentence to the two trace-level subtypes that actually implement it (`AutomationRuleEvaluatorLlmAsJudge` / `AutomationRuleEvaluatorUserDefinedMetricPython`), or state explicitly that span and trace-thread rules only ever evaluate SDK-logged items.

_rule: `backend-api-contract-keep-openapi-swagger-annotations-and-generated-sdks-in-sync-`_

apps/opik-backend/src/main/java/com/comet/opik/api/evaluators/AutomationRuleEvaluatorUpdate.java
```diff
@@ -54,7 +56,8 @@ public abstract sealed class AutomationRuleEvaluatorUpdate<T, E extends Filter>
     // the API boundary instead of failing the update (OPIK-7371).
     @NotBlank @Size(max = 150, message = "cannot exceed 150 characters") private final String name;
 
-    private final float samplingRate;
+    @Schema(description = "Fraction of production (SDK-logged) traces the rule scores, from 0 to 1. Experiment, playground and optimization traces are always scored in full and ignore this value.")
+    @DecimalMin("0") @DecimalMax("1") private final float samplingRate;
```
thiagohora, line 60: **[medium] New validation guard has no test on the update path**

The new @DecimalMin("0")/@DecimalMax("1") guard is a semantically distinct validation branch on a public endpoint and nothing exercises it: no test asserts 422 for sampling_rate < 0 or > 1, nor 2xx for the inclusive boundaries 0 and 1. The sibling constraint on this very class (@Size(max = 150) on name, line 57) got dedicated 422 assertions for both create and update (AutomationRuleEvaluatorsResourceTest.java:959 and :979), so this change lacks the parity the team already established.

💡 Add a @ParameterizedTest over {-0.1, 1.1} asserting SC_UNPROCESSABLE_ENTITY (with the violation message) and over {0f, 1f} asserting success, for both POST /v1/private/automations/evaluators and PATCH/PUT of an existing rule.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/OnlineScoringSamplerTest.java
```diff
@@ -475,6 +475,87 @@ void samplesAllTracesWhenSamplingRateIsOne() {
...
+        @Test
+        void scoresExperimentTracesWhenSamplingRateIsZeroForPythonEvaluator() {
+            when(serviceTogglesConfig.isPythonEvaluatorEnabled()).thenReturn(true);
+            var trace = createTrace(Source.EXPERIMENT);
+            var evaluator = createPythonEvaluator(0.0f);
+            whenFindAllPythonEvaluators(evaluator);
+
+            onlineScoringSampler.onTracesCreated(new TracesCreated(List.of(trace), workspaceId, userName));
+
+            verify(onlineScorePublisher).enqueueMessage(any(),
```
thiagohora: **[medium] Python bypass test asserts with any() instead of the expected message**

scoresExperimentTracesWhenSamplingRateIsZeroForPythonEvaluator verifies `enqueueMessage(any(), eq(USER_DEFINED_METRIC_PYTHON))`. The loose matcher only proves *something* was enqueued for the python stream — it does not pin the trace, ruleId, ruleName or code carried in the message, so the test still passes if the bypass enqueues the wrong trace (e.g. the SDK trace of a mixed batch) or builds the message from the wrong evaluator. Every other test in this file (including the sibling python test processesPythonEvaluatorWhenToggleEnabled, line 135-144, which builds the exact TraceToScoreUserDefinedMetricPython) asserts the exact expected list, so this new test is the weakest link precisely on the branch the PR adds. It also asserts nothing about the number of calls.

💡 Extract a `toPythonMessage(AutomationRuleEvaluatorUserDefinedMetricPython evaluator, Trace trace)` helper next to the existing `toLlmMessage(...)` (building trace/ruleId/ruleName/code/workspaceId/userName), reuse it in processesPythonEvaluatorWhenToggleEnabled, and assert `verify(onlineScorePublisher).enqueueMessage(List.of(toPythonMessage(evaluator, trace)), AutomationRuleEvaluatorType.USER_DEFINED_METRIC_PYTHON)` here.

_rule: `backend-testing-assert-exact-tight-bound-expected-values-instead-of-loose-ma`_

apps/opik-documentation/documentation/fern/docs-v2/production/online-evaluation/rules.mdx
```diff
@@ -34,7 +34,7 @@ To create a new scoring metric in the UI, first navigate to the project you woul
 When creating a new rule, you will be presented with the following options:
 
 1. **Name:** The name of the rule
-2. **Sampling rate:** The percentage of traces to score. When set to `100%`, all traces will be scored.
+2. **Sampling rate:** The percentage of production traces to score. When set to `100%`, all production traces will be scored. The setting applies to production traces only: traces from experiments, the playground and optimization runs ignore it, and are always scored in full whenever the rule applies to them.
```
thiagohora: **[medium] Claim that non-production traces are "always scored in full" is false for thread (and span) rules**

The sampling bypass added in this PR exists only on the trace path (OnlineScoringSampler.java: `if (!Source.isLoggingSource(trace.source())) { ... return true; }`). For thread rules and span rules the backend does the opposite: TraceThreadOnlineScoringSamplerListener.java:91 filters the batch with `.filter(thread -> Source.isLoggingSource(thread.source()))` and OnlineScoringSpanSampler.java:94 does the same for spans, so experiment/playground/optimization threads and spans are dropped before any rule is considered and are never scored at all. This page documents thread rules (see the "Online thread evaluation rules" section, line 131 onward) and the sampling-rate control it describes is the same control used for thread rules, so the blanket sentence tells users the exact opposite of what happens for threads.

💡 Scope the statement to trace rules and state the thread/span behaviour explicitly, e.g. "For trace rules, the sampling rate applies to production (SDK-logged) traces only — traces from experiments, the playground and optimization runs ignore it and are always scored when the rule matches. Thread and span rules only run on production (SDK-logged) data; threads and spans from experiments, the playground and optimization runs are excluded from online evaluation entirely."

_general finding_

apps/opik-documentation/documentation/fern/docs/production/rules.mdx
```diff
@@ -35,7 +35,7 @@ To create a new scoring metric in the UI, first navigate to the project you woul
 When creating a new rule, you will be presented with the following options:
 
 1. **Name:** The name of the rule
-2. **Sampling rate:** The percentage of traces to score. When set to `100%`, all traces will be scored.
+2. **Sampling rate:** The percentage of production traces to score. When set to `100%`, all production traces will be scored. This setting applies to production traces only — traces from experiments, the playground and optimization runs are never sampled.
```
thiagohora: **[medium] Playground/optimization traces are only scored when explicitly selected, not "never sampled"**

For non-SDK, non-experiment sources (playground, optimization, evaluator) the trace never even reaches the sampling check unless it carries selected_rule_ids metadata: OnlineScoringSampler.sampleAndScore (apps/opik-backend/.../events/OnlineScoringSampler.java:220-241) adds such traces to scorableTraces only when extractSelectedRuleIds(trace) is non-empty, and skips the project batch otherwise. The sentence as written tells users every playground/optimization trace is scored regardless of the rate, when in reality most are not scored at all.

💡 State the actual behavior: playground and optimization traces are evaluated only when the user explicitly selects a rule for them, and in that case the sampling rate is ignored.

_general finding_

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSampler.java
```diff
@@ -354,6 +362,15 @@ private boolean shouldSampleTrace(AutomationRuleEvaluator<?, ?> evaluator, Strin
                     trace.id(), evaluator.getName());
         }
 
+        // The sampling rate thins a production stream, so it applies to SDK traces only.
+        if (!Source.isLoggingSource(trace.source())) {
+            logForUser(workspaceId, evaluator, trace,
+                    "The traceId '{}' is scored for rule: '{}' regardless of the sampling rate '{}',"
```
thiagohora: **[low] Message asserts an outcome ('is scored') that the sampler cannot guarantee**

The sibling messages in this method state a decision the sampler actually owns ('was skipped for rule X as the rule is disabled'). The new line instead promises an outcome — 'The traceId X is scored for rule Y' — but `shouldSampleTrace` only decides admission to `samples`. Enqueueing happens later via `OnlineScoringSamplerSupport.publishSampled`, which is explicitly fire-and-forget: on failure it calls `log.error(...)` on the *internal* logger only, and nothing is written back to the user-facing stream. So when the Redis enqueue fails, the user's rule log shows 'is scored' for a trace that was never scored, with no corrective line. (The downstream 'Evaluating traceId ... sampled by rule ...' line does not have this problem because it is emitted after dequeue.)

💡 Reword to state the decision rather than the result, e.g. "The traceId '{}' is not subject to the sampling rate '{}' for rule: '{}', as the rate applies to production traces only" — leaving the 'evaluating/scored' claim to the scorer that actually performs it.

_rule: `backend-observability-keep-log-statements-and-code-comments-consistent-with-actual`_

apps/opik-documentation/documentation/fern/docs-v2/production/online-evaluation/rules.mdx
thiagohora: **[low] UI tooltip for the same field still says "Percentage of traces to evaluate"**

The docs and the OpenAPI @Schema description are updated, but the in-product tooltip a user actually reads while filling this field is unchanged: RuleFilteringSection.tsx:556 renders `tooltip="Percentage of traces to evaluate"`. Users configuring a rule in the UI get the old, now-known-inaccurate description while the docs claim production-only semantics.

💡 Update the FE tooltip in apps/opik-frontend/src/v2/pages-shared/automations/AddEditRuleDialog/RuleFilteringSection.tsx to match (e.g. "Percentage of production (SDK-logged) traces to evaluate") so UI, docs and API schema agree.

_general finding_

thiagohora (commented):

7 findings — 0 high · 5 medium · 2 low. Suppressed by team conventions: see `suppressed.md`.

apps/opik-backend/src/main/java/com/comet/opik/api/evaluators/AutomationRuleEvaluator.java
```diff
@@ -80,7 +82,8 @@ public abstract sealed class AutomationRuleEvaluator<T, E extends Filter> implem
...
+    @DecimalMin("0") @DecimalMax("1") private final float samplingRate;
```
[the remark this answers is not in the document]
jverre: The committed `openapi/opik.yaml` and the generated SDKs are produced by `.github/workflows/sdks_generate_openapi_spec_and_fern_code.yml`, which does `git add apps/opik-documentation/documentation/fern/openapi/opik.yaml` itself. Regenerating by hand here would just be overwritten, so nothing to do in this PR.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringSampler.java
```diff
@@ -354,6 +362,15 @@ private boolean shouldSampleTrace(AutomationRuleEvaluator<?, ?> evaluator, Strin
                     trace.id(), evaluator.getName());
         }
 
+        // The sampling rate thins a production stream, so it applies to SDK traces only.
```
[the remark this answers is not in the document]
jverre: Net log volume is unchanged. Before this PR the same non-SDK traces emitted `The traceId '{}' was skipped for rule: '{}' and per the sampling rate '{}'` on the identical user-facing stream; they now emit the bypass line instead. One line per (rule, trace) either way, so this is not new I/O — and dropping it would leave a user with no explanation for why a 0%-rate rule scored their whole experiment.

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7957`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

thiagohora (approved):

PR #7958, opened 2026-08-21
[OPIK-8062] [SDK] feat: one-command MCP + skill pack setup

alexkuzmik:
## Details

Roughly half of Opik MCP installs never issue a single tool call, and the install step *is* the onboarding. Today it asks permission it should assume, verifies nothing, reaches three of the seven AI hosts users actually run, and cannot be scripted at all — so a coding agent, a Dockerfile, or CI has no path to a configured MCP server. This reworks `opik mcp configure`, adds a matching `opik skills` command, and wires both into `opik configure`.

### MCP server

- **`--install-mcp` was dead code in exactly the environments people automate.** The `is_interactive()` guard in `_should_setup_mcp_server` ran *before* the explicit flag, so `opik configure -y --install-mcp` silently did nothing and exited 0. An explicit flag *is* the user asking, so it is now honoured without a TTY. `-y` on its own still skips MCP — a blanket yes-to-everything should not edit another tool's config files.
- **New `--host` flag on `opik mcp configure`** (repeatable, or `all`). The command previously hard-failed without a terminal, but a terminal was only ever needed to *ask* which host to use. Naming one makes "set Opik up for me" a single step an agent can run. An explicit host installs whether or not it is detected, so a fresh CI image can be configured before the editor is.
- **The installer now proves it works before reporting success.** It wrote JSON and told the user to go check. Combined with the server's own behaviour — an unconfigured server starts happily, reports workspace `default` and advertises all its tools — a completely broken setup was indistinguishable from a working one until the agent hit a 401 mid-conversation. It now makes a real call with exactly the values it wrote and reports `connected to workspace acme-ai, 7 project(s) visible`, or fails with the reason.
- **Codex and opencode host targets.** Codex is the highest-volume and most reliable client in the telemetry and was hand-writing JSON. Codex is driven through its own CLI (its config is TOML, which this installer deliberately does not hand-edit) and read back via `codex mcp get --json`, normalised so `opik mcp status` needs no per-host special casing. opencode gets its own block shape (`local`/`remote`, argv as one list, `environment` rather than `env`).
- **Refuses to guess the workspace.** Omitting it makes the server send `default`, which resolves to the account default — so reads come back from the wrong place instead of failing. Returning confidently wrong data is the one failure here that doesn't look like a failure. A failed workspace lookup is explicitly *not* treated as evidence of a single workspace.
- **Distinguishes Added from Updated** by reading the existing registration before writing. The first cut said "Registered", which mixed the mechanism with the outcome and left a re-run looking like a fresh install.
- **Names the exact `uv` install command** for the platform when `uvx` is missing.

### Skill pack

The MCP server gives an assistant *tools*; the skill pack gives it the knowledge of how to use Opik. The telemetry says the second half is the gap — 43 installs have loaded the tool list 14,120 times between them without ever calling a tool, and `schema` is the most widely-called tool of all. Agents are connected and groping.

New `opik skills` group (`configure` / `update` / `remove` / `status`), plus `opik configure --install-skills` and `opik mcp configure --skills`.

**The pack is offered as a recommended follow-up to the MCP step, not as a choice up front.** Asking "MCP or skills or both?" before anything happens makes the user decide between two things they cannot yet see; asking *after* the server is registered means the results are on screen, and the pack only applies to the hosts the server actually reached. The prompt does not re-list those hosts — the results table directly above it already does, and repeating three names buries the question.

**No third-party installer is involved.** [`comet-ml/opik-skills`](https://github.com/comet-ml/opik-skills) documents `npx skills add`, but that turns out to be unnecessary — skills are `SKILL.md` directories and the assistants have converged on a shared user-level location. Verified per host rather than taken on trust:

| Assistant | Reads | Source |
| --- | --- | --- |
| Codex | `$HOME/.agents/skills` | `codex-rs/ext/skills/src/host_roots.rs`, under `ConfigLayerSource::User` |
| opencode | `~/.agents/skills`, `~/.claude/skills`, `~/.config/opencode/skills` | opencode skills docs |
| Cursor, VS Code Copilot | `~/.agents/skills` | shared location |
| Claude Code | `~/.claude/skills` only | gets a symlink into the shared copy |

So one write plus one link covers every host, with no Node, no `npx`, and no external CLI whose flags can change under us. It also makes the install HOME-scoped and independent of the working directory — matching the MCP install, and answering "which project folder?" with "none" — and it needs no Opik credentials, so it works before `opik configure`.

`update` compares a **content hash** rather than a commit: codeload tarballs name their root directory by *ref*, so the recorded "commit" was literally the string `main` and could never detect a change. Skills the pack has dropped are removed, so a rename upstream does not leave the old name behind for the assistant to keep reading.

Interop is preserved: `npx skills add comet-ml/opik-skills` writes to the same place, so the two are interchangeable. A pack present but unrecorded is reported as *installed outside this CLI* rather than ignored, and `remove` only touches what we recorded installing, so a hand-written skill sharing a name survives.

### Wizard UX, and the layering it needed

The flow used to print a consent question and a config-saved log line on adjacent lines with no framing, then write files and stop — no plan, no completion signal.

- **A plan before anything is written**, naming each target and its file, so the write is predictable rather than discovered afterwards.
- **An arrow-key multiselect instead of typing numbers**, built on stdlib `termios`/`msvcrt`. Deliberately not `questionary`/`prompt_toolkit`: this is the core SDK, and a picker is not worth a dependency in every user's environment. Falls back to a plain confirmation where the terminal cannot host it.
- **A closing `✓ Done` block** stating what was set up, for which assistants, and the one next action.
- **Representation split from business logic.** Rendering had leaked into the installers, which is what made the skill pack land as raw log lines in the middle of otherwise formatted output. `InstallView` is now a port in `configurator/`; `LoggingInstallView` is the library-safe default, so `opik.configure()` called from Python still just logs; `RichInstallView` lives in `cli/`. `configurator/skills/install.py` returns an `InstallResult` and renders nothing.

### Non-interactive behaviour

Every command in the group is now exercised with stdin closed, across an 18-case matrix. That found **four** prompts reachable from a headless path, each of which aborted the run:

| Path | Was | Now |
| --- | --- | --- |
| `opik mcp configure --host cursor` | registered the server, *then* aborted at the skills prompt | skips the pack with a note pointing at `--skills` |
| `opik skills configure` (no `--host`) | `Aborted!` | actionable error naming `--host` |
| `opik skills remove` (no `-y`) | `Aborted!` | actionable error naming `-y` |
| `--install-mcp` (no `--host`) | `EOFError` in the numbered menu — **found in review** | the flag is the consent; installs for detected hosts |

`opik mcp configure --host … --skills` with `OPIK_API_KEY` / `OPIK_WORKSPACE` in the environment is a genuinely headless entry point. `opik configure` is **not** — it asks its own deployment-type and workspace questions first, which reproduces on `main` and is out of scope here; the docs no longer advertise it as the CI path.

### Out of scope

- **Analytics.** Reporting is being built on OPIK_8061; this branch leaves `ANALYTICS:` comments at each call site naming the event it owes — including the decline rate on the consent prompts, which is currently unobservable and is the number that would justify or kill the prompt changes above.
- **`cline` and `continue` MCP host targets** — one install each in the telemetry, and config locations I could not verify.
- **The duplicate `opik` skill** that `opik-claude-code-plugin` also ships. The installer flags the overlap and says how to drop one, but which repo owns that skill is someone else's call.

## Change checklist
- [x] User facing
- [x] Documentation update

## Issues

- Resolves #
- OPIK-8062

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Scope: Full authorship of the implementation, tests, and documentation, from a plan and telemetry analysis reviewed by the author. Third-party contracts were verified against upstream sources rather than assumed: `codex mcp add` / `codex mcp get --json` against `codex-rs/cli/src/mcp_cmd.rs`, Codex's skill roots against `codex-rs/ext/skills/src/host_roots.rs`, and the opencode MCP and skills schemas against opencode's docs.
- Human verification: Author reviewed the diff, the behaviour changes to the consent prompts and `-y` semantics, the decision to install skills without a third-party CLI, the prompt ordering and wording, and the docs.

## Testing

```bash
cd sdks/python
python -m pytest tests/unit -q                                            # full suite
python -m pytest tests/unit/configurator tests/unit/cli -q                # 953 passed
git diff --name-only | tr '\n' '\0' | xargs -0 pre-commit run --files     # ruff, ruff-format, mypy pass
```

**~3,700 test lines added** across 15 files, 10 of them new. Highlights:

- **Flags/consent:** `--install-mcp` and `--install-skills` honoured with no TTY, with and without `-y`; `--no-*` still wins; non-interactive with no flag skips; prompts name one host and several; no prompt when nothing is detected.
- **Headless consent:** `is_interactive` forced False *and* `builtins.input` patched to raise, then asserting the installer still dispatched — so a future regression fails loudly instead of hanging. Pytest detaches stdin, so the suite needed an autouse fixture defaulting to "has a terminal"; without it the new guard silently changed what every prompt-driven test was exercising.
- **`--host`:** single / repeated / de-duplicated / `all` / `all` with nothing detected / unknown value rejected by the parser; non-interactive **with** `--host` succeeds and **without** it errors suggesting the flag; non-interactive + unconfigured Opik errors instead of launching the interactive wizard.
- **Verification:** 200 with and without a `total`, unparseable body, 401/403, 5xx, network failure; hosted probe where 401 is healthy and 404 is not; the API key never appears in a failure message.
- **Workspace guard:** many workspaces → refuse and name them; named / single / failed-lookup / local / no-key → proceed, skipping the lookup where it isn't needed.
- **Codex & opencode:** no-CLI fallback with no leaked key; remove-then-add ordering; `streamable_http` normalised to `http`; opencode config-dir resolution across `OPENCODE_CONFIG_DIR` / `XDG_CONFIG_HOME` / default, `.jsonc` preferred only when it already exists, and unrelated keys preserved.
- **Selector:** arrow/space/enter/`q`/Ctrl-C key handling, Windows two-byte arrow prefixes, cancel returning `None` versus a deliberate empty selection returning `[]`, and rendering captured through `Console.capture()`.
- **Skill pack:** in-memory tarballs covering path traversal, symlink members, oversized files, missing `SKILL.md`, and no-skills-at-all; content hash stable across reads, sensitive to content, and insensitive to the tarball root name; `write_skill` replaces a skill entirely (a file dropped upstream must not survive), leaves no staging dir, and replaces a symlink without following it; symlink→copy fallback; uninstall leaves un-recorded skills alone.

A test caught a real bug in my own code: `PurePosixPath(".").parts` is empty, which made the traversal guard vacuously true and let `.` through as a skill name.

## Documentation

The public MCP page is **reframed around agent velocity**: it now opens with what the setup unlocks and four copy-paste starter prompts, with the per-host technical detail moved below the fold. The old page led with transport tables and per-host JSON, which answered "how do I wire this up" but never "why would I".

| Page | Change |
| --- | --- |
| `prompt_engineering/mcp-server.mdx` | Rewritten opening (what this unlocks + starter prompts); `opik mcp configure --host … --skills` documented as the headless path; Codex, opencode and the skill pack added |
| `tracing/advanced/sdk_configuration.mdx` | `opik configure` flag tri-states (`--install-mcp` / `--install-skills`) and the terminal requirement stated plainly |
| `home.mdx`, `integrations/overview.mdx` | Entry points updated to point at the one-command setup |

The docs no longer advertise `opik configure --install-mcp` as the CI path, because it is not one — that was a review finding, and the correction points at `opik mcp configure` instead.


CometActions:

This is Python SDK/CLI surface only: the new assistant step in `opik configure` writes MCP entries and the skill pack into config files owned by Cursor/Claude Code/VS Code, and the one REST call it adds (GET /v1/private/projects, in verification.py, to prove the credentials it just wrote actually work) is read-only. No page renders anything different and no endpoint changes shape, and taxonomy.yaml deliberately keeps sdks/python as untracked surface rather than a UI area, so there is no e2e capability to target. The automated check for a spec came back empty only because it looks at tests_end_to_end/ — the coverage is your ~200 unit tests under sdks/python/tests/unit/{cli,configurator}, which would fail if the consent policy, host detection, the atomic JSON write or the 401/network verification paths were wrong. No environment spent. Separately, and not a testing point: `opik.configure()` loses the `install_mcp` kwarg added in #6959, so a released caller passing it now gets a TypeError — worth confirming that removal is intended.

_also touches Python SDK_

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/33170555343)

_Re-checked after a push on 28 Aug 12:19 UTC — nothing the verdict depends on changed._

sdks/python/src/opik/configurator/configure.py
```diff
@@ -99,33 +102,53 @@ def _maybe_setup_mcp_server(self) -> None:
             self_hosted_comet=self.self_hosted_comet,
             check_tls_certificate=self.current_config.check_tls_certificate,
             force_local_server=False,
+            # The prompt below already named the detected hosts, so re-confirming
+            # them inside the installer would ask the same question twice.
+            assume_confirmed=self._mcp_prompt_named_detected_hosts,
```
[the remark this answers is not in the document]
alexkuzmik: Confirmed and fixed in 5eafe1ecb4.

Reproduced exactly as described — `install_mcp=True` with no `--host` left `assume_confirmed=False`, so `_confirm_targets` fell into the numbered menu:

```
EOFError: EOF when reading a line
  ... in numbered_menu
  raw = [token.strip() for token in input(prompt).split(",") ...]
```

Rather than passing the detected list from the caller, the guard went into `_confirm_targets` itself: reaching that point without a terminal can only mean the caller opted in through a flag, because every interactive path refuses earlier. So the flag *is* the consent and the candidates are used as-is. That covers the library entry point (`opik.configure(install_mcp=True)`) as well as the CLI.

This was the third instance of the same class in the branch, so the whole command surface is now exercised with stdin closed — every case is exit 0 or a clean actionable error, none reach a prompt. Regression tests added in `TestHeadlessConsent`.

sdks/python/src/opik/configurator/configure.py
```diff
@@ -99,33 +102,53 @@ def _maybe_setup_mcp_server(self) -> None:
...
+        - Otherwise: ask, naming the hosts we actually found, defaulting to "no".
         """
         if self.install_mcp is False:
             return False
 
-        if not is_interactive():
-            return False
-
         if self.install_mcp is True:
             return True
```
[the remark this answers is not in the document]
alexkuzmik: Right, and this one is a documentation bug rather than a code one — thank you.

`opik configure` asks its own deployment-type and workspace questions before the MCP step, via `input()`. Those abort with no terminal, and that reproduces on `main`:

```
$ git stash && HOME=$SB opik configure < /dev/null
> 
Aborted!
```

So it is pre-existing and out of scope for this branch — but I had documented `opik configure --install-mcp` as the CI path, which cannot work. Corrected in 5eafe1ecb4: the docs now point at `opik mcp configure --host … --skills` with `OPIK_API_KEY` / `OPIK_WORKSPACE` in the environment, which is genuinely headless, and say plainly that `opik configure` is not.

On the test suggestion: added, though at the layer below the CLI. `TestHeadlessConsent` patches `is_interactive` to False *and* `builtins.input` to raise `AssertionError("must not prompt")`, then asserts the installer still dispatched — so a future regression fails loudly rather than hanging. Pytest detaches stdin, so the suite also needed an autouse fixture defaulting to "has a terminal"; without it the new guard silently changed what every prompt-driven test was exercising.

sdks/python/src/opik/cli/configure.py
```diff
@@ -1,37 +1,208 @@
...
+        # what the previous error told it to add.
+        _announce_assistant_skip()
+        return assistants.NOTHING_DONE
+
+    skills_flag = install_skills
+    if install_mcp is False:
+        # Server declined outright: only the pack is on the table.
+        if skills_flag is False:
+            return assistants.NOTHING_DONE
+        return assistants.setup(setup_params, skills_flag=True, host_keys=None)
```
petrotiurin: **`--no-install-mcp` is ignored here — the server gets registered anyway, and the skill pack is installed unprompted.**

`assistants.setup()` unconditionally calls `mcp_installer.setup_mcp_server(...)` as its first act (`cli/assistants.py:63`), so routing the skills-only case through it defeats the opt-out. And because `skills_flag=True` is hardcoded on this line, the pack is written with no prompt at all.

I ran this path against the branch with `setup_mcp_server` / `setup_skills` stubbed to record calls:

| Invocation | Calls made | Outcome |
| --- | --- | --- |
| `opik configure --no-install-mcp` | `setup_mcp_server`, `setup_skills` | `clients=1, skills=True` |
| `--no-install-mcp --install-skills` | `setup_mcp_server`, `setup_skills` | `clients=1, skills=True` |
| `--no-install-mcp --no-install-skills` | none | `clients=0, skills=False` |

The closing panel then reports "Set up: MCP server and skill pack" for a run that explicitly asked for neither. The unprompted pack install is the sharper end of it: `configurator/assistants.py:99-103` argues the pack is the *more* invasive of the two steps because it writes instruction files the assistant acts on with its own permissions, and this path skips its consent entirely. `sdk_configuration.mdx:100` also advertises exactly this flag combination as the way to skip the prompts.

Two things needed: a skills-only path that doesn't route through `assistants.setup`, or an `install_mcp` parameter on `setup` that it honours — plus keeping `skills_flag` as passed rather than forcing `True`.

Minor, same block: `if skills_flag is False: return NOTHING_DONE` on line 45-46 is unreachable — the `install_mcp is False and install_skills is False` guard on line 33 already caught that case.

sdks/python/src/opik/configurator/mcp/targets.py
```diff
@@ -155,10 +220,126 @@ def _install_vscode(server_spec: mcp_spec.McpServerSpec) -> InstallResult:
...
+    into someone else's TOML without a writer risks losing their comments and
+    formatting. So when the CLI is unavailable we hand the work back rather than
+    guessing.
+    """
+    return (
+        f"the `codex` CLI was not found on your PATH, and {_codex_config_path()} is "
+        "TOML, which this installer does not edit directly. Install the Codex CLI "
+        "and re-run `opik mcp configure --host codex`, or add an "
+        f"`[mcp_servers.{SERVER_NAME}]` table to that file by hand."
     )
```
petrotiurin, line 303: **Stale flag name — following this instruction fails.** `--host` became `--ai-client` in 17eb5a001e, so `opik mcp configure --host codex` exits with `Error: No such option: --host`.

This is the Codex-CLI-missing path, so it's printed to exactly the person who has no other way forward. `find_target()`'s docstring at line 434 ("the values accepted by ``--host``") is stale for the same reason.

sdks/python/src/opik/configurator/mcp/install.py
```diff
@@ -67,49 +130,259 @@ def setup_mcp_server(
...
         )
-        return
+        # ANALYTICS: install skipped, reason="declined" — the decline rate on the
+        # consent prompt, and the single most important number this flow owes.
+        return []
 
     if isinstance(server_spec, mcp_spec.StdioServerSpec):
-        _prefetch_opik_mcp()
+        with display.step("Preparing the Opik MCP server"):
+            _prefetch_opik_mcp()
```
petrotiurin, line 178: **Two writers on one tty.** `display.step()` is `console.status(..., spinner="dots")` for `RichInstallView` (`cli/install_view.py:79`), i.e. a rich `Live` region — while `_prefetch_opik_mcp` deliberately does *not* capture uv's output ("Stream uv's own output ... straight to the terminal", line 398-400). uv's progress bars and the spinner's redraw will overwrite each other. Either capture uv's output, or don't wrap this call in a spinner — the two intentions are in direct conflict.

**Separately, `uv tool install` is more than the cache warm the docstring describes** (line 402). The spec registers `uvx opik-mcp`, but `uv tool install` creates a persistent tool environment and an `opik-mcp` shim on the user's PATH — an install into their environment that nothing announces, and one that won't upgrade an already-present older copy. If the goal is warming the cache for what actually runs, `uvx opik-mcp --help` matches the invocation. There's also no `timeout=`, so `opik configure` can block indefinitely on a slow or wedged fetch.

sdks/python/src/opik/configurator/mcp/spec.py
```diff
@@ -110,3 +129,20 @@ def to_claude_add_args(self) -> List[str]:
...
         return cli_args
+
+    def to_codex_add_args(self) -> List[str]:
+        cli_args = [SERVER_NAME]
+        for key, value in self.env.items():
+            cli_args.extend(["--env", f"{key}={value}"])
+        cli_args.append("--")
+        cli_args.append(self.command)
+        cli_args.extend(self.args)
+        return cli_args
```
petrotiurin, line 140: **The API key lands in the process argv**, so it's visible via `ps` to any other user on the machine — and to anything scraping process tables. Same shape in `to_claude_add_args` just above.

For Codex this looks unavoidable: its CLI is the only writer you have, since the config is TOML. For Claude Code it isn't — `_install_claude_code` already has a working JSON path it falls back to when `claude` isn't on PATH (`targets.py:151-159`), and preferring the file write there would keep the key off the command line for the highest-volume host.

Related, on the file-write path: `json_config.merge_server_into_json_file` creates `~/.cursor/mcp.json`, VS Code's `mcp.json` and `opencode.json` at whatever the umask gives (usually 0644) with the key inside. Pre-existing behaviour, but this PR widens it to two more hosts, so a `chmod 0600` on files the installer creates would be worth folding in.

sdks/python/src/opik/configurator/skills/pack.py
```diff
@@ -0,0 +1,207 @@
...
+    try:
+        with httpx_client.get(
+            workspace=None,
+            api_key=None,
+            check_tls_certificate=True,
+            compress_json_requests=False,
+        ) as client:
+            response = client.get(
+                url, timeout=DOWNLOAD_TIMEOUT_SECONDS, follow_redirects=True
+            )
```
petrotiurin: **Low.** This routes a third-party download through Opik's own httpx factory, which applies `hooks.httpx_client_hook` and honours `_OPIK_HTTP_PROXY` (`httpx_client.py:56,63,39`). `api_key=None` / `workspace=None` correctly keeps the Authorization and `Comet-Workspace` headers off the request, so nothing leaks today — but those hooks exist to decorate calls to the Opik API, and an enterprise build that registers a header-injecting hook would start sending them to `codeload.github.com`. A plain `httpx.Client` for a fetch that isn't an Opik API call keeps that door shut.

Also on this function: `MAX_ARCHIVE_BYTES` is checked at line 105 *after* `response.content` has fully materialised the body, so the guard can't prevent the memory it exists to prevent. Streaming with a running counter, or dropping the check as unenforceable, are both more honest than the current shape.

petrotiurin (commented):
Reviewed the SDK side of this. The layering is the right call — `configurator/` holding the logic with `InstallView` as the seam, and rendering kept in `cli/` — and the comments carry real reasoning rather than restating the code. Five findings inline.

One is blocking: `opik configure --no-install-mcp` still registers the MCP server *and* installs the skill pack with no prompt (`cli/configure.py:43-47`, reproduced by running the path). The rest are a stale `--host` flag in a user-facing hint, a rich `Live` region fighting streamed subprocess output, the API key reaching the process argv, and one low-severity note on the skill-pack download client.

sdks/python/src/opik/configurator/skills/pack.py
```diff
@@ -0,0 +1,207 @@
...
+        file_path = staging / relative_path
+        file_path.parent.mkdir(parents=True, exist_ok=True)
+        file_path.write_bytes(content)
+
+    if target.exists() or target.is_symlink():
+        if target.is_dir() and not target.is_symlink():
+            shutil.rmtree(target)
+        else:
+            target.unlink()
+    staging.replace(target)
```
petrotiurin, line 238: I would add some guards here (or in path resolvers) - even if path exists, if `name` or parts in `destination_root` resolve to empty we could be deleting something we shouldn't.

sdks/python/src/opik/configurator/skills/install.py
```diff
@@ -0,0 +1,189 @@
...
+                "have a known skills location"
+            ),
+        )
+
+    # ANALYTICS: skills install started, with the host count.
+    try:
+        pack = skills_pack.download(ref=ref)
+    except skills_pack.PackError as error:
+        # ANALYTICS: skills install failed, reason="download_failed".
+        return InstallResult(succeeded=False, error=str(error))
```
petrotiurin: I see there are stubs for analytics. Should we add them as a part of this PR? I think it's worthwhile to have visibility into how this feature is doing once it's rolled out.

sdks/python/src/opik/configurator/mcp/targets.py
```diff
@@ -51,24 +59,63 @@ def _vscode_user_config_path() -> pathlib.Path:
...
+    server_block: Dict[str, Any],
 ) -> InstallResult:
     try:
         was_new = json_config.merge_server_into_json_file(
             config_path=config_path,
             top_level_key=top_level_key,
             server_name=SERVER_NAME,
-            server_block=server_spec.to_block(),
+            server_block=server_block,
         )
```
petrotiurin, line 119: **The write to the user's config is not atomic.** `json_config.merge_server_into_json_file` reads the whole file, adds one key, and does a plain `config_path.write_text(...)` (json_config.py:41) — truncate, then write. The `OSError` handler below catches the failure and prints a clean message, so this never crashes; but by then the file is already truncated.

That matters because of what the file is. `~/.claude.json` is Claude Code's entire user state, not just an MCP block — on my machine it's 110 KB across 81 top-level keys (`mcpServers` isn't even present until something registers one). ENOSPC, a signal, or a killed process mid-write leaves the user's editor config unparseable, and the only thing Opik wanted was to add one entry. Same exposure for `~/.cursor/mcp.json`, VS Code's `mcp.json` and `opencode.json`.

The fix is already in this PR, just applied to the wrong set of files: `skills/pack.py::write_skill` writes to a staging path and swaps it in precisely so a failure can't leave a half-written result. Opik's own files get that treatment; the third-party configs Opik doesn't own get the unsafe write. A temp file in the same directory plus `os.replace()` here would close it.

(`json_config.py` isn't in this diff, so flagging it at the call site — but this is the path all three JSON hosts go through, and the PR adds two of them.)

sdks/python/src/opik/configurator/mcp/targets.py
```diff
@@ -155,10 +220,126 @@ def _install_vscode(server_spec: mcp_spec.McpServerSpec) -> InstallResult:
...
+    # entry first to keep re-runs idempotent — same shape as the Claude Code path.
+    subprocess.run(
+        [codex_executable, "mcp", "remove", SERVER_NAME],
+        capture_output=True,
+        text=True,
+    )
+
+    command = [codex_executable, "mcp", "add"] + server_spec.to_codex_add_args()
+
+    result = subprocess.run(command, capture_output=True, text=True)
```
petrotiurin: **These four `subprocess.run` calls are unguarded, and `_read_codex_block` right below shows the intended shape.** Lines 268 and 276 here, plus 170 and 187 on the Claude Code path. `_read_codex_block:311-317` wraps its `subprocess.run` in `except OSError` — so the read path is protected and every write path isn't.

`shutil.which` returning a path is not evidence the binary will exec: it only checks the executable bit. `claude` and `codex` are Node shims, so the common real-world break is node moving out from under them (nvm version deleted, node upgraded, a partially-removed npm install). The shim still passes `which` and then fails at `execve`:

```
shutil.which found: .../bin/claude
raised FileNotFoundError: [Errno 2] No such file or directory: '.../bin/claude'
```

That's an unhandled traceback out of `opik configure` — the failure mode this PR is otherwise careful to avoid everywhere else. Note the errno message is misleading too: the missing file is node, not `claude`, so the honest report needs to come from here rather than from the exception text.

`except OSError` around each call, folded into the same `InstallResult(succeeded=False, ...)` the non-zero-exit branch already returns, covers it. Worth adding `timeout=` at the same time — these inherit stdin, so a client CLI that decides to prompt hangs the configure run indefinitely.

petrotiurin (dismissed):

PR #7960, opened 2026-08-21
[NA] [SDK] fix: respect search wait timeouts

### LINKED ISSUE #7936: [TypeScript SDK] search polling can exceed the requested timeout
### Summary

`searchAndWaitForDone` can sleep for the full polling interval after its timeout budget has already elapsed, so a caller can wait substantially longer than `waitForTimeout`.

### Reproduction

```ts
const started = Date.now();
await searchAndWaitForDone(
  async () => [],
  1,
  100,
  5000,
);
console.log(Date.now() - started);
```

With no matching results, the current implementation checks the 100 ms budget and then sleeps for 5 seconds before checking again, so the promise resolves around 5 seconds rather than within the requested timeout.

### Expected behavior

The polling delay should be capped to the remaining timeout budget (for example, `Math.min(sleepTime, waitForTimeout - elapsedTime)`) so the helper never intentionally oversleeps its caller-provided timeout. Existing behavior for an immediately satisfied search and for a timeout shorter than the poll interval should remain unchanged.

### Scope

This is isolated to `sdks/typescript/src/opik/utils/searchHelpers.ts`; a deterministic fake-timer regression can cover a poll interval longer than the timeout without external services.

jstar0:
## Details
`searchAndWaitForDone` currently sleeps for the full polling interval even when less time remains in the caller-provided timeout. This can make a short timeout wait for several seconds longer than requested. The polling sleep is now capped to the remaining timeout budget, with a fake-timer regression covering a poll interval longer than the timeout.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves #7936
- OPIK-

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: AI coding assistant
- Model(s): not disclosed
- Scope: implementation and regression-test drafting
- Human verification: contributor reviewed the final diff and ran the repository's TypeScript unit, lint, formatting, and diff checks.

## Testing

- `./node_modules/.bin/vitest run tests/unit/utils/searchHelpers.test.ts` — 28 passed
- `./node_modules/.bin/eslint src/opik/utils/searchHelpers.ts tests/unit/utils/searchHelpers.test.ts` — passed
- `./node_modules/.bin/prettier --check src/opik/utils/searchHelpers.ts tests/unit/utils/searchHelpers.test.ts` — passed
- `git diff --check origin/main...HEAD` — passed
- The full TypeScript SDK test suite was not run; this PR is limited to the polling helper and its focused unit coverage.

## Documentation

No documentation changes are needed for this internal timeout-boundary fix.

### DIFF
--- a/sdks/typescript/src/opik/utils/searchHelpers.ts
+++ b/sdks/typescript/src/opik/utils/searchHelpers.ts
@@ -13,7 +13,7 @@ export async function searchTracesWithFilters(
   filters: OpikApi.TraceFilterPublic[] | null,
   maxResults: number,
   truncate: boolean,
-  exclude?: OpikApi.TraceSearchStreamRequestPublicExcludeItem[]
+  exclude?: OpikApi.TraceSearchStreamRequestPublicExcludeItem[],
 ): Promise<OpikApi.TracePublic[]> {
   const streamResponse = await apiClient.traces.searchTraces({
     projectName,
@@ -26,7 +26,7 @@ export async function searchTracesWithFilters(
   const traces = await parseNdjsonStreamToArray<OpikApi.TracePublic>(
     streamResponse,
     serialization.TracePublic,
-    maxResults
+    maxResults,
   );
 
   return traces;
@@ -39,7 +39,7 @@ export async function searchAndWaitForDone<T>(
   searchFn: () => Promise<T[]>,
   waitForAtLeast: number,
   waitForTimeout: number,
-  sleepTime: number
+  sleepTime: number,
 ): Promise<T[]> {
   const startTime = Date.now();
   let result: T[] = [];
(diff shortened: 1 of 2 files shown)


jstar0:
Thanks for the review. I checked each Baz finding against the diff and am keeping this PR scoped to the reported timeout-boundary bug:

- The `parseNdjsonStreamToArray` result cap/stream cancellation behavior predates this PR and is shared by the trace, thread, and span helpers; changing it would alter existing search semantics and resource-lifecycle behavior beyond the timeout fix.
- The generic parser documentation wording is also pre-existing and can be corrected independently without affecting this regression.
- Widening generated exclusion-item types to `string[]` is a public API/forward-compatibility decision unrelated to polling timeouts.

I am leaving these as separate follow-up work rather than bundling unrelated behavior changes into this focused PR.

PR #7966, opened 2026-08-24
[OPIK-7891] [BE] fix: count distinct item ids when sizing a dataset version

JetoPistola:
## Details

> **Stacked on #7705** (OPIK-7707). Base branch is `danield/OPIK-7707-make-the-version-count-update-atomic`, so this PR's diff shows only the OPIK-7891 fix. Merge #7705 first; this retargets to `main` automatically once it lands.

image

`DatasetItemVersionDAO.insertItems` returned `items.size()` — the raw list length. That is deliberate, not an oversight: ClickHouse async inserts report 0 rows before commit, so the code returns the count of what it handed in rather than trusting `getRowsUpdated`. But reads collapse a repeated `dataset_item_id` to a single row via `LIMIT 1 BY`, so a stable id appearing twice in one batch made the stored version total one higher than the rows the version actually holds.

Every version total flows through that return value — `createFirstVersion` feeds it straight into `itemsTotal`, and `applyDelta` sums added + edited + copied — so the fix belongs at the source: count distinct stable ids, matching what the storage engine keeps.

- Counts `DISTINCT` `datasetItemId` in `insertItems` — the same field the `INSERT` below binds, so counting and writing share one definition of item identity. (An earlier revision added an `item.id()` fallback for a null `datasetItemId`; it was removed after review. The `INSERT` binds `datasetItemId().toString()` with no null check, so a null cannot survive to be counted either way, and every caller normalizes the field first.)
- Fixes all four call sites at once rather than patching one caller and leaving the others inflated.

### This corrects the mechanism recorded on the ticket

The ticket described this as asynchronous ClickHouse visibility racing concurrent batches on the *append* path, reachable only via `num_threads>1`. Investigation shows it is none of those things:

| Reported | Actual |
|---|---|
| A race between concurrent batches | Deterministic; one single-threaded request reproduces it |
| Requires `num_threads>1` | Parallelism is irrelevant to the defect |
| In the append path (`countExistingItemIds`) | In the version-**creating** path (`insertItems` → `createFirstVersion`) |

The append path is genuinely correct — `countExistingItemIds` classifies a cross-batch duplicate as an update. A test in this PR covers that case and passes both with and without the fix, which is what rules the append path out.

`num_threads` was a red herring: the SDK's content-hash dedup drops same-content duplicates before batching, so a duplicate id only survives into a request when the content differs. The reporter's parallel runs were simply the ones where such an id landed in the version-creating batch.

## Change checklist
- [x] User facing
- [x] Documentation update

User-facing: the "Item count" in **Datasets → _dataset_ → Version history** (and `dataset_items_count` in the SDK) now agrees with the rows the version holds. No API shape change.

## Issues

- Resolves #
- OPIK-7891

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: Traced the call graph, wrote the fix and both tests, ran the suites and the with/without-fix comparisons below.
- Human verification: Pending author review.

## Testing

```
mvn test -Dtest='DatasetVersionResourceTest*'                       # 127/127 pass
mvn test -Dtest='DatasetsResourceTest$CreateDatasetItems,...'       # 45/45 pass
mvn spotless:check                                                  # clean
```

Two tests added, deliberately covering both shapes, and each was run with and without the fix:

| Scenario | Without fix | With fix |
|---|---|---|
| Duplicate stable id **inside the version-creating batch** | **FAILS** — `items_total` 3, rows stored 2 | passes |
| Duplicate stable id **spanning two batches** in one `batch_group_id` | passes | passes |

The first is the regression test. The second is a negative control: it pins the append path's existing correct behaviour and is the evidence that the reported async-visibility mechanism is not the cause. Both assert `items_total` against the rows actually returned for the version, rather than against a hardcoded number, so a future change that breaks one but not the other still fails.

Not verified:

- No measurement of production impact on already-drifted datasets. This prevents new inflation; it does not reconcile a version whose `items_total` is already wrong. Reconciliation is tracked separately under OPIK_7804.
- The original report was against a deployed environment; the reproduction here is a local integration test. The mechanism is confirmed in code and by test, but I have not re-run the reporter's exact 1,200-item scenario.

## Documentation

N/A — internal counter correctness fix with no API or configuration surface.

### DIFF
--- a/.agents/skills/opik-backend/testing.md
+++ b/.agents/skills/opik-backend/testing.md
@@ -284,6 +284,31 @@ assertThat(actual)
 `containsExactly` asserts size and content together — `hasSize` plus per-index checks does not,
 and lets an extra element through.
 
+`hasSize` **on its own** is weaker still: it asserts a count and nothing about identity, so any
+bug that preserves the count passes. This bites hardest on de-duplication, merge, and upsert
+tests, where the count is exactly the thing a bug is most likely to keep right:
+
+```java
+// ❌ BAD - passes if the wrong revision survived, or if the duplicate was kept
+// and the distinct row dropped. Both keep the size at 2.
+assertThat(stored).hasSize(2);
+assertThat(version.itemsTotal()).isEqualTo(stored.size());
+
+// ✅ GOOD - names the rows that must survive, so a wrong-winner bug fails
+assertThat(stored)
+    .usingRecursiveFieldByFieldElementComparatorIgnoringFields(IGNORED_FIELDS_DATA_ITEM)
+    .containsExactlyInAnyOrder(winningDuplicate, distinctItem);
+assertThat(version.itemsTotal()).isEqualTo(stored.size());
+```
+
+Asserting a derived counter against `stored.size()` is good — it ties the counter to reality
+rather than to a literal — but it is only as strong as the assertion on `stored` itself. Pin the
+contents first, then tie the counter to them.
+
+Reuse the shared ignore-field constants (`IGNORED_FIELDS_DATA_ITEM` and friends) rather than
+declaring a local list: they already encode which server-generated fields are not part of the
+contract, and a local copy silently drifts from them.
+
 These `containsExactly*` variants compare elements with the element type's own `equals`, which is
 what you want for exact-valued models. When the elements carry `BigDecimal` or `double`, the same
 exception that justifies a comparator on a single object applies per element — otherwise a
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
@@ -3683,7 +3683,15 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
 
         // Note: ClickHouse with async inserts returns 0 immediately before commit.
         // We return the count of items we're inserting instead of relying on getRowsUpdated.
-        long itemCount = items.size();
+        //
+        // Count DISTINCT stable ids, not rows handed in (OPIK-7891): reads collapse a repeated
+        // dataset_item_id to one row via LIMIT 1 BY, so counting the raw list inflates every
+        // version total derived from this value. Callers set datasetItemId before inserting;
+        // fall back to id so an unnormalized item still counts as itself rather than as null.
+        long itemCount = items.stream()
+                .map(item -> item.datasetItemId() != null ? item.datasetItemId() : item.id())
+                .distinct()
+                .count();
 
         return asyncTemplate.nonTransaction(connection -> {
             Segment segment = startSegment(DATASET_ITEM_VERSIONS, CLICKHOUSE, "insert_delta_items");
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
@@ -3686,10 +3686,11 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
         //
         // Count DISTINCT stable ids, not rows handed in (OPIK-7891): reads collapse a repeated
         // dataset_item_id to one row via LIMIT 1 BY, so counting the raw list inflates every
-        // version total derived from this value. Callers set datasetItemId before inserting;
-        // fall back to id so an unnormalized item still counts as itself rather than as null.
+        // version total derived from this value. Counting the same field the INSERT binds below
+        // keeps one definition of identity -- every caller normalizes datasetItemId first, and a
+        // null would fail at the bind regardless, so there is nothing to fall back to.
         long itemCount = items.stream()
-                .map(item -> item.datasetItemId() != null ? item.datasetItemId() : item.id())
+                .map(DatasetItem::datasetItemId)
                 .distinct()
                 .count();
 
--- a/apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
@@ -3796,21 +3796,14 @@ class DatasetVersionResourceTest {
             var datasetId = createDataset(UUID.randomUUID().toString());
 
             // The SDK's parallel upload sends every batch under one batch_group_id. The batch that
-            // arrives first CREATES the version, and that path derives itemsTotal from insertItems,
-            // which returns items.size() -- the raw list length, deliberately not a DB row count
-            // (ClickHouse async inserts report 0 before commit). A stable id repeated inside that
-            // first batch is therefore counted twice, while ClickHouse collapses it to one row.
+            // arrives first CREATES the version, and that path derives itemsTotal from insertItems.
+            // That used to return items.size() -- the raw list length, deliberately not a DB row count
+            // (ClickHouse async inserts report 0 before commit) -- so a stable id repeated inside the
+            // first batch was counted twice while ClickHouse collapsed it to one row. It now counts
+            // distinct dataset_item_ids, which is what this test pins.
             var duplicatedId = TestIdGeneratorFactory.create().generateId();
             var distinctId = TestIdGeneratorFactory.create().generateId();
 
-            // The repeated id wins with its LAST submitted content: ClickHouse keeps one row per
-            // dataset_item_id and reads take the newest, so "second" survives and "first" does not.
-            var winningDuplicate = DatasetItem.builder()
-                    .id(duplicatedId)
-                    .datasetItemId(duplicatedId)
-                    .source(DatasetItemSource.SDK)
-                    .data(Map.of("value", JsonUtils.getJsonNodeFromString("\"second\"")))
-                    .build();
             var distinctItem = DatasetItem.builder()
                     .id(distinctId)
                     .datasetItemId(distinctId)
@@ -3824,7 +3817,11 @@ class DatasetVersionResourceTest {
                             .source(DatasetItemSource.SDK)
                             .data(Map.of("value", JsonUtils.getJsonNodeFromString("\"first\"")))
                             .build(),
-                    winningDuplicate,
+                    DatasetItem.builder()
+                            .id(duplicatedId)
+                            .source(DatasetItemSource.SDK)
+                            .data(Map.of("value", JsonUtils.getJsonNodeFromString("\"second\"")))
+                            .build(),
                     distinctItem);
 
             datasetResourceClient.createDatasetItems(DatasetItemBatch.builder()
@@ -3834,15 +3831,23 @@ class DatasetVersionResourceTest {
                     .build(), TEST_WORKSPACE, API_KEY);
 
             var version = getLatestVersion(datasetId);
-
-            // Assert on the rows themselves, not just how many: a bug that kept the wrong revision
-            // of the duplicate, or dropped the distinct item and kept both duplicates, would leave
-            // the count at 2 and slip through a size-only check.
             var stored = datasetResourceClient.getDatasetItems(
                     datasetId, 1, 100, version.versionHash(), API_KEY, TEST_WORKSPACE).content();
+
+            // Which revision of the repeated id survives is deliberately NOT asserted: every row in a
+            // batch is written with the same now64(9), and the read dedupes with
+            // `ORDER BY dataset_item_id DESC, last_updated_at DESC LIMIT 1 BY dataset_item_id` -- no
+            // tie-breaker, so either payload may win. Pinning "second" would be asserting an accident.
+            // What is guaranteed, and what the fix is about: exactly one row per distinct id, and a
+            // counter that agrees with it.
+            assertThat(stored).extracting(DatasetItem::id)
+                    .containsExactlyInAnyOrder(duplicatedId, distinctId);
             assertThat(stored)
-                    .usingRecursiveFieldByFieldElementComparatorIgnoringFields(IGNORED_FIELDS_DATA_ITEM)
-                    .containsExactlyInAnyOrder(winningDuplicate, distinctItem);
+                    .filteredOn(item -> distinctId.equals(item.id()))
+                    .singleElement()
+                    .usingRecursiveComparison()
+                    .ignoringFields(IGNORED_FIELDS_DATA_ITEM)
+                    .isEqualTo(distinctItem);
 
             // items_total must agree with what is actually stored.
             assertThat(version.itemsTotal()).isEqualTo(stored.size());
--- a/apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
@@ -3791,6 +3791,96 @@ class DatasetVersionResourceTest {
     class BatchVersioningTests {
 
         @Test
+        @DisplayName("Success: Duplicate stable id in the version-creating batch counts once (OPIK-7891)")
+        void putItems__whenCreatingBatchRepeatsStableId__thenCountedOnce() {
+            var datasetId = createDataset(UUID.randomUUID().toString());
+
+            // The SDK's parallel upload sends every batch under one batch_group_id. The batch that
+            // arrives first CREATES the version, and that path derives itemsTotal from insertItems,
+            // which returns items.size() -- the raw list length, deliberately not a DB row count
+            // (ClickHouse async inserts report 0 before commit). A stable id repeated inside that
+            // first batch is therefore counted twice, while ClickHouse collapses it to one row.
+            var duplicatedId = TestIdGeneratorFactory.create().generateId();
+            var items = List.of(
+                    DatasetItem.builder()
+                            .id(duplicatedId)
+                            .source(DatasetItemSource.SDK)
+                            .data(Map.of("value", JsonUtils.getJsonNodeFromString("\"first\"")))
+                            .build(),
+                    DatasetItem.builder()
+                            .id(duplicatedId)
+                            .source(DatasetItemSource.SDK)
+                            .data(Map.of("value", JsonUtils.getJsonNodeFromString("\"second\"")))
+                            .build(),
+                    DatasetItem.builder()
+                            .id(TestIdGeneratorFactory.create().generateId())
+                            .source(DatasetItemSource.SDK)
+                            .data(Map.of("value", JsonUtils.getJsonNodeFromString("\"third\"")))
+                            .build());
+
+            datasetResourceClient.createDatasetItems(DatasetItemBatch.builder()
+                    .datasetId(datasetId)
+                    .batchGroupId(UUID.randomUUID())
+                    .items(items)
+                    .build(), TEST_WORKSPACE, API_KEY);
+
+            var version = getLatestVersion(datasetId);
+
+            // Two distinct ids went in, so the version holds two rows.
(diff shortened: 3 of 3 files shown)


apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -3499,7 +3499,15 @@ public Mono<Long> insertItems(@NonNull UUID datasetId, @NonNull UUID newVersionI
...
-        long itemCount = items.size();
+        //
+        // Count DISTINCT stable ids, not rows handed in (OPIK-7891): reads collapse a repeated
+        // dataset_item_id to one row via LIMIT 1 BY, so counting the raw list inflates every
+        // version total derived from this value. Callers set datasetItemId before inserting;
+        // fall back to id so an unnormalized item still counts as itself rather than as null.
+        long itemCount = items.stream()
+                .map(item -> item.datasetItemId() != null ? item.datasetItemId() : item.id())
+                .distinct()
+                .count();
```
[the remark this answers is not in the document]
JetoPistola: Fixed in 10a4d68fb5 — dropped the fallback.

You're right about the real problem here, and it's the sharpest point in this review: the `INSERT` below binds `item.datasetItemId().toString()` with **no** null check (`DatasetItemVersionDAO.java:3535`), so a null stable id NPEs at the bind regardless of how it was counted. My `item.id()` fallback couldn't protect anything — it just introduced a second, divergent definition of item identity in the one place that has to agree with what the `INSERT` writes. Counting `datasetItemId` directly is now the only rule.

Two parts of the comment I'd push back on:

- **The counter inflation is what this PR fixes**, not something still outstanding. `insertItems` no longer counts raw items — that's the entire change. `items_added`/`items_modified` were never sized from this value; they come from `countExistingItemIds` on the append path.
- **`.agents/skills/opik-backend/testing.md` doesn't require DAO-level coverage** for this. I checked; there's no such rule. Integration coverage through the resource is the house pattern, which is what the two tests here do.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
```diff
@@ -3557,6 +3557,96 @@ void deleteDataset__whenVersionedDatasetWithTags__thenReturnNoContent() {
...
+            var items = List.of(
+                    DatasetItem.builder()
+                            .id(duplicatedId)
+                            .source(DatasetItemSource.SDK)
+                            .data(Map.of("value", JsonUtils.getJsonNodeFromString("\"first\"")))
+                            .build(),
+                    DatasetItem.builder()
+                            .id(duplicatedId)
+                            .source(DatasetItemSource.SDK)
+                            .data(Map.of("value", JsonUtils.getJsonNodeFromString("\"second\"")))
```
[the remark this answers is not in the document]
JetoPistola: Skipping — the premise doesn't hold, and the change it asked for is now moot.

The claim is that the test leaves `datasetItemId` unset and so exercises the `item.id()` fallback rather than the `datasetItemId` branch. Two reasons it doesn't:

1. **`datasetItemId` is `READ_ONLY` on the write view** (`DatasetItem.java:38-42`), so no HTTP-level test can set it. The server derives it — which is exactly why the test supplies the duplicate via `id`.
2. **The service normalizes before the DAO is reached.** `createFirstVersion` sets `.datasetItemId(stableId)` on every item (`DatasetItemService.java:1863-1870`) before calling `insertItems`, and the other three call sites do the same. So `datasetItemId` is always non-null at the count — the test was already covering the primary branch, and the fallback was the unreachable one.

Which is the useful part of this finding, just inverted: the fallback was dead code, so it's gone as of 10a4d68fb5. There is now a single branch and the existing test covers it.

Adding a case with distinct `id` values sharing an explicit `datasetItemId` isn't reachable through the API for reason 1, and post-fix there's no second branch left to distinguish.

CometActions:

The fix is a real user-facing one — `items_total` is what the Version history tab renders as "Item count", so before this a batch repeating a `dataset_item_id` showed a version claiming more items than the dataset holds. The existing merged spec `dataset-version-counters.spec.ts` does not catch it: its 1200/600-item seeds are all-unique ids, so `items.size()` and the distinct count are the same number and it passes either way. What does catch it is the QA draft stacked on your branch, #7969 — `tests_end_to_end/e2e/tests/datasets/dataset-version-duplicate-ids.spec.ts` sends a 5-row batch with 4 distinct ids and asserts `itemsTotal == 4` on the create path and on the delta path, plus the rendered "Item count". That fails on `items.size()`. So no new exploration or spec here — but that coverage lives in an unmerged draft, so if #7969 is closed rather than merged this change goes back to being uncovered end to end. Your two `DatasetVersionResourceTest` cases pin the same behaviour at the API level.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/33295909470)

_Re-checked after a push on 30 Aug 06:05 UTC._

CometActions:

### Access Information
- **URL:** https://pr-7966.dev.comet.com
- **Namespace:** pr-7966
- **Version:** 2.2.14-7966-merge-3048
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7966%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7966`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

CometActions:

### Access Information
- **URL:** https://pr-7966.dev.comet.com
- **Namespace:** pr-7966
- **Version:** 2.2.14-7966-merge-3063
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7966%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7966`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
```diff
@@ -3557,6 +3557,96 @@ void deleteDataset__whenVersionedDatasetWithTags__thenReturnNoContent() {
...
+
+            var version = getLatestVersion(datasetId);
+
+            // Two distinct ids went in, so the version holds two rows.
+            var stored = datasetResourceClient.getDatasetItems(
+                    datasetId, 1, 100, version.versionHash(), API_KEY, TEST_WORKSPACE).content();
+            assertThat(stored).hasSize(2);
+
+            // items_total must agree with what is actually stored.
+            assertThat(version.itemsTotal()).isEqualTo(stored.size());
```
andrescrz: Minor: you might want to assert the whole `stored` (by reusing pre-existing assertion helpers, classes or methods, if any) instead of asserting sizes only.

Applies to both new tests in this PR.

andrescrz (dismissed):
LGTM, just a minor test assertion improvement to make them more reliable. Not a blocker.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
JetoPistola: Fixed in ce3e7bb298 — both tests now assert the rows, not the count. Good catch, and it was the weakest possible check for what these tests are for.

Reused the existing `IGNORED_FIELDS_DATA_ITEM` element-comparator pattern already used elsewhere in this file rather than adding anything new:

```java
assertThat(stored)
        .usingRecursiveFieldByFieldElementComparatorIgnoringFields(IGNORED_FIELDS_DATA_ITEM)
        .containsExactlyInAnyOrder(winningDuplicate, distinctItem);
```

Worth noting why `hasSize` was especially bad *here*: on a de-duplication test the count is exactly what a bug is most likely to get right. Keeping the wrong revision of the duplicate, or dropping the distinct row and keeping both duplicates, each leaves the size at 2 and would have passed.

Naming the expected rows also pins something the tests previously only assumed — that the **last** submitted revision of a repeated id is the one that survives. That's now asserted rather than incidental.

I kept `itemsTotal()` tied to `stored.size()` rather than a literal, since that's what ties the counter to reality — but you're right that it only means anything once `stored` itself is pinned.

Also folding the general rule into `.agents/skills/opik-backend/testing.md` (the Collections section) so it's not just fixed here. That's going up as a separate PR off `main` — this branch is stacked on #7705 and branched before ~160 lines of that doc landed, so editing it here would revert other people's work.

thiagohora:
## Opik reviewer — 1 confirmed finding

The core change here (`items.stream().map(DatasetItem::datasetItemId).distinct().count()`) holds up: 4 of the 5 other findings this run produced were **refuted** on empirical evidence, including the one high. Notably, "duplicates are counted once but still written as colliding rows with no tie-break" does not reproduce — `optimize_on_insert = 1` is the ClickHouse server default (verified on 26.3.16.16, the version in `docker-compose.yaml:46`, and not overridden anywhere in the repo), so same-key rows collapse inside the inserted block and the duplicate row is never written.

The one finding that survived is that **the fix is incomplete**. It is anchored outside this PR's diff hunks, so it could not be posted inline:

### [medium] Sibling counter editItemsViaSelectInsert still returns the raw list size, so the delta path keeps the items_total inflation
`apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java:3313`

The OPIK-7891 fix was applied to insertItems only. editItemsViaSelectInsert still does `long itemCount = editedItems.size();` and its value is consumed as the edited half of items_total: DatasetItemService.applyDeltaChanges (line 1548) does editedCountMono.flatMap(editedCount -> applyDelta(...).map(otherCount -> editedCount + otherCount)) and feeds the sum straight into createVersionFromDelta, which stores it as items_total (and, for a first version, as items_added -- DatasetVersionService:398). DatasetItemChanges.editedItems is a plain List<DatasetItemEdit> with no uniqueness validation, and the surrounding code already treats it as possibly repeating (it derives editedDatasetItemIds as a Set for the COPY exclusion, and countRowsInVersion excludes each stable id once). A payload that repeats an item id in edited_items therefore inflates items_total by one per repeat -- exactly the bug the diff fixes on the insert path -- while the version still holds one visible row per stable id.

💡 Count distinct edited stable ids the same way insertItems now does: `long itemCount = editedItems.stream().map(DatasetItemEdit::id).distinct().count();` (or de-duplicate editedItemEdits by id in applyDeltaChanges before sizing editedUuids), so both halves of the applyDelta sum use one definition of identity.

---
_Note: this PR is stacked on #7705, so GitHub renders both PRs' changes here. Findings on #7705's lines (the `items_total = -1` migration sentinel, the absolute-writer race against the backfill, and the stale `save()` `@return` contract) are posted on #7705 rather than duplicated here._

React 👍/👎 — feedback helps tune what it flags.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
```diff
@@ -3557,6 +3557,120 @@ void deleteDataset__whenVersionedDatasetWithTags__thenReturnNoContent() {
...
 
+        @Test
+        @DisplayName("Success: Duplicate stable id in the version-creating batch counts once (OPIK-7891)")
+        void putItems__whenCreatingBatchRepeatsStableId__thenCountedOnce() {
+            var datasetId = createDataset(UUID.randomUUID().toString());
+
+            // The SDK's parallel upload sends every batch under one batch_group_id. The batch that
+            // arrives first CREATES the version, and that path derives itemsTotal from insertItems,
+            // which returns items.size() -- the raw list length, deliberately not a DB row count
+            // (ClickHouse async inserts report 0 before commit). A stable id repeated inside that
```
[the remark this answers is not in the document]
JetoPistola: Fixed in bad9213b4a — reworded to say it *used to* return `items.size()`.

You're right that describing pre-fix behaviour in the present tense is a trap for the next reader, especially in a test whose whole point is the new contract.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
```diff
@@ -3557,6 +3557,120 @@ void deleteDataset__whenVersionedDatasetWithTags__thenReturnNoContent() {
...
+            // first batch is therefore counted twice, while ClickHouse collapses it to one row.
+            var duplicatedId = TestIdGeneratorFactory.create().generateId();
+            var distinctId = TestIdGeneratorFactory.create().generateId();
+
+            // The repeated id wins with its LAST submitted content: ClickHouse keeps one row per
+            // dataset_item_id and reads take the newest, so "second" survives and "first" does not.
```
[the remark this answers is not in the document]
JetoPistola: Fixed in bad9213b4a — good catch, and you're right on the mechanism. I traced it before changing anything:

- `BATCH_INSERT_ITEMS` binds `now64(9)` for `last_updated_at`, evaluated once per statement, so **every row in one batch carries an identical timestamp**.
- The read dedupes with `ORDER BY dataset_item_id DESC, last_updated_at DESC LIMIT 1 BY dataset_item_id` (`DatasetItemVersionDAO.java:394`) — no tie-breaker.

So which duplicate survives is genuinely unspecified. My test was passing on row ordering, and worse, I'd written the assumption into a code comment as if it were the contract.

Took your first option: assert only what is guaranteed — exactly one row per distinct id, plus the distinct item's content intact — and left a comment explaining that the winner is *deliberately* unasserted, so nobody re-adds it later. I didn't go with writing explicitly ordered replacement versions: it would make the test pass by constructing a scenario the SDK path never produces.

Worth noting the sibling test (`putItems__whenDuplicateStableIdSpansBatchesInGroup__thenCountedOnce`) is unaffected — separate HTTP requests get separate `now64(9)` values, so its ordering is well-defined. Left as-is.

136/136 green.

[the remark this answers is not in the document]
JetoPistola: Correction to my reply above: the fix is in **74d011aabd**, not `bad9213b4a`.

I replied "Fixed in bad9213b4a" while the change was still sitting uncommitted in my worktree — the rebase and force-push carried the earlier commits but not this one, so for a period this branch claimed a fix it didn't have. Caught it while re-verifying the branch contents; now pushed and confirmed on the remote (`winningDuplicate` gone, the "deliberately NOT asserted" version present).

The substance of the earlier reply stands: the duplicate-winner assertion is removed because `now64(9)` gives every row in a batch the same timestamp and the read has no tie-breaker, so the test now asserts only one-row-per-distinct-id plus the distinct item's content.

JetoPistola:
Will rebase, after approval / or when https://github.com/comet-ml/opik/pull/7705 gets merged first, FYI @thiagohora @andrescrz 

image

.agents/skills/opik-backend/testing.md
```diff
@@ -284,6 +284,31 @@ assertThat(actual)
...
+    .usingRecursiveFieldByFieldElementComparatorIgnoringFields(IGNORED_FIELDS_DATA_ITEM)
+    .containsExactlyInAnyOrder(winningDuplicate, distinctItem);
+assertThat(version.itemsTotal()).isEqualTo(stored.size());
+```
+
+Asserting a derived counter against `stored.size()` is good — it ties the counter to reality
+rather than to a literal — but it is only as strong as the assertion on `stored` itself. Pin the
+contents first, then tie the counter to them.
+
+Reuse the shared ignore-field constants (`IGNORED_FIELDS_DATA_ITEM` and friends) rather than
```
andrescrz, line 308: Not only should reuse the constant, many times there's an existing helper method or even helper class which should be used (generally delegates to the constant). That way we keep the assertions logic centralised.

andrescrz (approved):
One advice on the `testing.md` rules, but not a blocker. It can go on a follow-up PR.

.agents/skills/opik-backend/testing.md
JetoPistola: Agreed — and the point generalises past the constant. Tracked as OPIK-8181 so this PR can merge without growing further.

The dataset item tests are the clearest instance: `IGNORED_FIELDS_DATA_ITEM` is declared twice (`DatasetsResourceTest:219` and, independently, `ExperimentAggregatesIntegrationTest:130`), `DatasetVersionResourceTest` imports it across package boundaries from a sibling resource test class, and ~10 call sites repeat the comparator chain by hand. Two copies of the list can drift silently.

The follow-up extracts a `DatasetItemAssertions` under `api/resources/utils/` mirroring `TraceAssertions`, migrates the call sites, and reworks the skill-doc guidance to say helper-method-or-class over inlined constant. Doing it in a separate PR rather than here because it touches three large test classes and needs the two constant declarations diffed and reconciled deliberately — that reconciliation is its own reviewable decision, not a drive-by in a counter-correctness fix.

Note the `hasSize` doc section this thread is anchored on is no longer on the branch; restoring it, phrased around helper reuse, is part of the follow-up.

andrescrz: @JetoPistola as general feedback, let's not create a ticket for small follow-up comments like this. 

We send the PR or not, but as part of the same tickets.

We can't maintain a backlog of PR follow-up tickets, it's too much burden. In addition, it's not the purpose of the backlog.

JetoPistola: @andrescrz - thanks for the feedback, appreciate it

Agreed - we shouldn't overload the backlog, especially with PR follow-up tickets (Like this comment) - avoiding ticket bloat, makes sense. 

Perhaps my main concern was that touching three large test classes and reconciling the constant drift felt a bit too heavy for a drive-by change, in this correctness PR

That said, there were alternative approaches for me to take:
1. Absorb It Immediately (The "Do It Now" Rule) - postponing merge, to avoid the follow up PR
2. The Fast-Follow PR Without Jira - NA ticket - create draft, and iterate now/later
3. Code-Level TODOs - perhaps this is an out dated approach - invisible graveyard like, for that we look in the code, or scan TODOs manually/automatically...
4. Using the same ticket number - while possible, issue rises with the jira status dance, for the label status, switching its value unexpectedly and having reference to multiple PRs from a jira ticket - which is why a separate ticket felt like the default path. **Additional reason is scoping sessions** - similar to pull requests - to keep them small and targeted

This time, already spun the jira ticket. Future wise - hope that will adjust the workflow, to keep things leaner

PR #7974, opened 2026-08-24
[OPIK-8064] [BE] fix: bound an automation rule's sampling rate at the API boundary

thiagohora:
## Details

`sampling_rate` was a bare `float` with no bean-validation constraint, so anything the wire could express reached the INSERT and came back as a 500 instead of a 4xx. Production hit this today from `AutomationRuleEvaluatorDAO.saveBaseRule`:

```
java.sql.SQLSyntaxErrorException: Unknown column 'Infinity' in 'field list'
```

- `1e40` overflows to `Infinity` on parse, and Jackson maps the strings `"NaN"` / `"Infinity"` onto the matching float constants. The MySQL driver renders those as the bare tokens `Infinity` / `NaN`, which MySQL reads as *identifiers* — hence "Unknown column".
- Finite but out-of-range rates (`1.5`, `-0.5`) hit the `automation_rules_chk_1` CHECK constraint and also surfaced as a 500.
- Fix: bound the field to the column's CHECK range on both the create and update DTOs, the way OPIK_7371 bounded `name` to its VARCHAR length. Hibernate Validator's `DecimalMin`/`MaxValidatorForFloat` treat NaN as out of bounds and ±Infinity as greater/less than any bound, so one annotation pair covers all three shapes of bad input.
- The generated OpenAPI spec picks up `minimum: 0.0` / `maximum: 1.0` on `sampling_rate` as a side effect; the checked-in spec is refreshed by the existing generate-and-commit workflow.

No data was corrupted — the inserts failed, and prod `opik_prod.automation_rules` holds 4013 rows with none outside `[0, 1]`.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-8064

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Scope: Diagnosed the production error from Loki logs, wrote the validation change and the `SamplingRateValidation` test class, ran the verification below.
- Human verification: Author reviewed the diff and the test results before pushing.

## Testing

```
mvn test -Dtest='AutomationRuleEvaluatorsResourceTest$SamplingRateValidation'
mvn spotless:check
```

New `AutomationRuleEvaluatorsResourceTest$SamplingRateValidation` — 18 cases, create and update paths over `1.5`, `-0.5`, `1e40`, `-1e40`, `"NaN"`, `"Infinity"`, `"-Infinity"`, plus in-range `0` / `0.5` / `1` / `1.0` still accepted.

- **With the annotations reverted:** 14 failures — every hostile value returned **500**, and the production `Unknown column 'Infinity'` / `'NaN'` error reproduced in the test log. This is what proves the tests catch the regression rather than merely passing.
- **With the fix:** 18/18 pass — 422 for hostile input, 201 for in-range.
- Spotless: BUILD SUCCESS.

Environment: local Maven + testcontainers (MySQL / ClickHouse / Redis), macOS. Full suite for this class not run beyond the new nested class; the change is a field-level constraint on two DTOs.

## Documentation

None — the bound is self-documenting via the OpenAPI spec's `minimum` / `maximum` on `sampling_rate`.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/evaluators/AutomationRuleEvaluator.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/evaluators/AutomationRuleEvaluator.java
@@ -12,6 +12,8 @@ import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import io.swagger.v3.oas.annotations.media.DiscriminatorMapping;
 import io.swagger.v3.oas.annotations.media.Schema;
 import jakarta.validation.Valid;
+import jakarta.validation.constraints.DecimalMax;
+import jakarta.validation.constraints.DecimalMin;
 import jakarta.validation.constraints.NotBlank;
 import jakarta.validation.constraints.NotNull;
 import jakarta.validation.constraints.Size;
@@ -26,6 +28,9 @@ import java.util.List;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.UUID;
+
+import static com.comet.opik.utils.ValidationUtils.MAX_SAMPLING_RATE;
+import static com.comet.opik.utils.ValidationUtils.MIN_SAMPLING_RATE;
 
 @Data
 @SuperBuilder(toBuilder = true)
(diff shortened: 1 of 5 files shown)


CometActions:
**No test needed here.**

This is input validation with no change to the valid path, and it already lands between two tests. The invalid side is covered by the PR's own AutomationRuleEvaluatorsResourceTest.SamplingRateValidation, which parameterises 1.5 / -0.5 / 1e40 / NaN / ±Infinity over both create and update — an e2e spec would assert the same 422 further from the boundary, and the non-finite literals only exist on the wire, which the typed e2e client cannot send anyway. The valid side is already guarded: online-evaluation-sampling-rate.spec.ts creates 50% and 100% rules through the dialog and asserts the persisted fractions are 0.5 and 1.0, so a bound that over-rejected (1.0 excluded, bounds inverted) would fail there. And the UI cannot reach the new rejection — SliderInputControl clamps a typed sampling rate to 0-100 before submit, so there is no user-visible flow behind this. Nothing to explore.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/32729993252)

PR #7976, opened 2026-08-24
[OPIK-8045] [QA] Proposed e2e spec from the #7949 exploration — online scoring on odd-shaped trace sections

CometActions:
> **Generated by the release QA side flow (`release-test-proposal`), not by a human.**
> It is a **draft** on purpose and needs review before it goes anywhere. Nothing here
> has been verified by a person.

## Where this came from

Exploratory testing of #7949 ("[OPIK-8045] [BE] fix: four online-scoring failures seen
in production") on that PR's own deployed environment produced one verified candidate
flow. This turns that flow into a permanent spec.

**Base branch.** This targets `thiagoh/online-scoring-failure-fixes`, not `main`, because
the behaviour it asserts only exists on that branch: on `main`,
`OnlineScoringEngine.extractFromJson` converts a trace section straight to
`Map<String, Object>`, which throws for any section that is not a JSON object. The throw
escapes `toReplacements` and fails the whole evaluation, so the spec's central assertion —
that a scalar or array `output` section still produces a score — fails on `main` by design.
The spec was written against #7949's head, `41c9fa9`.

## The spec

`tests_end_to_end/e2e/tests/online-evaluation/online-evaluation-odd-shaped-sections.spec.ts`

`@t2-cuj` `@area:online-evaluation`
`@cap:online-evaluation.python-rule-scores` `@cap:online-evaluation.scores-in-trace-panel`

**What it asserts.** Four `user_defined_metric_python` rules — one per variable mapping
(`output`, `output.output`, `output.flat.key`, `output.a.b`) — against four traces, one per
`output`-section shape (bare scalar, array, object with a literal dotted key, ordinary
nested object). It then asserts the full 4×4 matrix:

1. **Every rule scored every trace**, including the scalar and array sections. This is the
   regression. Before the fix those two shapes produced *no score at all* from any
   nested-path rule, because the section-to-`Map` conversion threw before the metric ran.
   After it, the evaluation completes and only the unresolved variable drops.
2. **Each score carries the value the mapping actually resolves to** — `"the answer"` for
   the section root against a scalar, `flatvalue` through the flat fallback, `nestedvalue`
   through ordinary descent, and the metric's `<<MISSING>>` sentinel where the mapping
   legitimately does not resolve.
3. **The scores render in the Logs trace panel** for the scalar trace — both the resolving
   rule and the unresolved-mapping one. A backend fix nobody can see on the page is not a
   fix.

**Why a code metric and not an LLM judge.** `OnlineScoringUserDefinedMetricPythonScorer`
drives the same `OnlineScoringEngine.toReplacements` → `extractFromJson` path as the LLM
judge, so this covers the changed code with no provider key and no LLM non-determinism.

**Why every rule maps a second `anchor` argument** to `input.q`, which always resolves:
without it, a rule whose only mapping fails to resolve hands the Python runner an empty
argument map and produces no score — for a benign reason that is indistinguishable on the
wire from the crash the spec exists to catch. The anchor guarantees at least one resolved
argument, so "no score" can only mean the evaluation failed.

**Existing coverage this does not duplicate.** `online-evaluation-smoke.spec.ts` already
covers `python-rule-scores`, but it maps `output.output` against an ordinary object section
— the one branch that never had the bug. The taxonomy read `covered: true` for a capability
whose regression path was untested; both capability notes are updated to say what is
covered by which spec.

## Supporting changes

- `fixtures/automation-rules.fixture.ts` — **new.** Automation rules do not cascade with
  their project (`ProjectService`'s delete path never touches `automation_rule_projects`),
  so every online-evaluation spec today deletes its rules in a trailing `test.step` — which
  is skipped precisely when an assertion has failed and left the workspace dirtiest. This
  is the fixture form: teardown runs regardless of outcome, honours
  `OPIK_LEAVE_FAILURES`, and skips any rule that also targets another project because
  deletion is workspace-global. It is opt-in (`auto: false`), so it costs the rest of the
  suite nothing. **The three existing specs are left as they are** — converting them is a
  separate change and this PR should stay reviewable.
- `core/backend/client.ts` — three additions:
  - `createPythonMetricRule(...)` — creates a code-metric rule with an explicit variable
    mapping. Rule creation *through the dialog* is already covered by
    `@cap:online-evaluation.create-python-rule`, so seeding four differently-mapped rules
    through the API here is setup, not a coverage gap.
  - `createTraceWithRawOutput(...)` — the generated client types `output` as
    `Record<string, unknown> | Record<string, unknown>[] | string`, which cannot express a
    bare scalar or an array section. Those shapes are the entire point, so this writes them
    over the wire verbatim (same reasoning as the existing `getProjectStats`, which also
    drops to `fetch`).
  - `getTraceSections(...)` and `AutomationRuleRef.pythonMetricArguments` — so the spec can
    prove its own fixture before asserting on it: that each trace really kept the section
    shape it stands for, and that each rule really persisted the mapping it names. Without
    those, a coerced section or a dropped mapping would make the matrix pass while
    verifying nothing.
- `coverage/taxonomy.yaml` — spec added to the area's `specs:` list; the two capability
  notes updated.

**Every path this spec uses behaves identically on the deployed build and at `41c9fa9`.**
None of `output`, `output.output`, `output.flat.key`, `output.a.b` or `input.q` contains
`..` or `[?(`, so `VariablePathUtils.findUnsupportedConstruct` passes them at save time and
`findUnsupportedConstructInJsonPath` passes them at scoring time; none contains `$.` inside
a key, so the `StringUtils.removeStart` change does not reach them either. The only branch
of #7949 the spec depends on is the `convertValue(json, Object.class)` rewrite.

## Verification

> **⚠️ Read this before trusting the spec. The scoring assertions are UNPROVEN.** The spec
> ran three times against `https://pr-7949.dev.comet.com` (build `2.2.37-7949-merge-3051`,
> OSS, workspace `default`) and never reached them, because online scoring on that
> environment stopped and then the environment itself went down. **Someone needs to re-run
> this on a healthy environment before it is promoted.**

Command (from `tests_end_to_end/e2e/`):

```
OPIK_DEPLOYMENT=oss OPIK_BASE_URL=https://pr-7949.dev.comet.com OPIK_WORKSPACE=default \
  npx playwright test tests/online-evaluation/online-evaluation-odd-shaped-sections.spec.ts --reporter=list
```

**What did run, green, against the live app:**

| Step | Result |
|---|---|
| `npx tsc --noEmit` | clean |
| `tag_lint.py --taxonomy … --estate tests_end_to_end` | `40 specs checked, 1 exempt, 0 problem(s)` |
| Create one code-metric rule per variable mapping | ✅ |
| All four rules persisted the mapping they were created with | ✅ |
| Seed one trace per output-section shape | ✅ |
| Each seeded trace really carries the section shape it stands for | ✅ (scalar, array, dotted-key and nested sections all stored verbatim) |
| Wait for scoring to settle, collect the matrix | ❌ **timed out — no score ever arrived** |
| the three assertion steps after it | not reached |
| Fixture teardown (project + `automationRulesCleanup`) | ✅ nothing leaked |

**Why the failure is the environment, not the spec — the control.** Seeding a trace into the
*exploration's own* project (`cuj-20260824-124205-824b-engine`), whose pre-existing rules —
none of them created by this spec — had scored ten traces at 12:52, also produced no score.
A watcher repeated that probe every two minutes for 45 minutes: `scored=False` every time,
with `GET /automations/evaluators/{id}/logs` empty, meaning the engine never consumed the
message at all. Shortly after, the environment stopped serving entirely:

```
GET /api/v1/private/traces?project_id=… -> 504 Gateway Time-out, then 502 Bad Gateway (nginx)
GET /api/is-alive/ver                   -> 502 Bad Gateway (nginx)
```

So the last thing this spec observed was a preview environment on its way down. There is no
evidence here of a defect in this PR, and equally no evidence the spec is right — only that
it is not wrong for any reason this run could reach.

**What the expected values rest on instead.** Every cell of the 4×4 matrix is corroborated
by the exploration's own captured scores on this exact build, read back from the API before
the environment failed — including the two cells that matter most, a nested-path rule
returning `<<MISSING>>` with a resolved anchor on a scalar section and on an array section,
which is the post-fix behaviour this spec asserts. The `"the answer"` cell keeping its JSON
quotes (the `$`-path special case) is likewise observed, not inferred. The metric,
the anchor argument and the mapping set are the same ones the exploration drove.

That is a good reason to think the spec is right and **not a substitute for a green run.**

The exploration produced **one** candidate and this PR implements it. The report's own
"the test to write once the environment is rebuilt" — an API-surface spec asserting that
`POST`/`PATCH` **reject** `output..text` and `output.choices[?(…)]` and accept
`output.choices[0].message` — is **not** here, and should not be until someone drives it.
The PR's deployed environment ran an intermediate build (`2.2.37-7949-merge-3051`) that has
the `extractFromJson` fix but **not** `@SupportedVariablePaths`, the engine-side path guard,
or the `StringUtils.removeStart` flat-fallback fix, so nobody has ever seen that validation
work. A spec for a flow nobody reached is a guess, and this spec deliberately avoids every
path (`..`, `[?(`, `$.`-in-key) whose behaviour differs between that build and this branch.

The exploration also flagged, at medium confidence, that the new guard rejects a flat key
whose *literal name* contains `..` or `[?(` — `flatFallback` exists precisely so
`output.flat.key` can find a property called `flat.key`, and the guard at
`OnlineScoringEngine.java:892` runs before `convertValue`, so the flat tail is never
reached. That is a question for a reviewer, not something to encode as a test.

---

Draft. Do not merge without review. Closes nothing; links #7949.

[OPIK-8045]: https://comet-ml.atlassian.net/browse/OPIK-8045?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/evaluators/AutomationRuleEvaluatorLlmAsJudge.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/evaluators/AutomationRuleEvaluatorLlmAsJudge.java
@@ -1,6 +1,7 @@
 package com.comet.opik.api.evaluators;
 
 import com.comet.opik.api.filter.TraceFilter;
+import com.comet.opik.api.validation.SupportedVariablePaths;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonView;
 import com.fasterxml.jackson.databind.PropertyNamingStrategies;
@@ -38,7 +39,9 @@ public final class AutomationRuleEvaluatorLlmAsJudge extends AutomationRuleEvalu
             @JsonView({
                     View.Public.class, View.Write.class}) @NotNull LlmAsJudgeModelParameters model,
             @JsonView({View.Public.class, View.Write.class}) @NotNull List<LlmAsJudgeMessage> messages,
-            @JsonView({View.Public.class, View.Write.class}) @NotNull Map<String, String> variables,
+            @JsonView({
+                    View.Public.class,
+                    View.Write.class}) @NotNull @SupportedVariablePaths Map<String, String> variables,
             @JsonView({View.Public.class, View.Write.class}) @NotNull List<LlmAsJudgeOutputSchema> schema,
             @JsonView({View.Public.class, View.Write.class}) @Positive BigDecimal maxCostUsd) {
 
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/evaluators/AutomationRuleEvaluatorSpanLlmAsJudge.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/evaluators/AutomationRuleEvaluatorSpanLlmAsJudge.java
@@ -1,6 +1,7 @@
 package com.comet.opik.api.evaluators;
 
 import com.comet.opik.api.filter.SpanFilter;
+import com.comet.opik.api.validation.SupportedVariablePaths;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonView;
 import com.fasterxml.jackson.databind.PropertyNamingStrategies;
@@ -38,7 +39,9 @@ public final class AutomationRuleEvaluatorSpanLlmAsJudge
             @JsonView({
                     View.Public.class, View.Write.class}) @NotNull LlmAsJudgeModelParameters model,
             @JsonView({View.Public.class, View.Write.class}) @NotNull List<LlmAsJudgeMessage> messages,
-            @JsonView({View.Public.class, View.Write.class}) @NotNull Map<String, String> variables,
+            @JsonView({
+                    View.Public.class,
+                    View.Write.class}) @NotNull @SupportedVariablePaths Map<String, String> variables,
             @JsonView({View.Public.class, View.Write.class}) @NotNull List<LlmAsJudgeOutputSchema> schema) {
     }
 
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/AgenticScoringService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/AgenticScoringService.java
@@ -157,6 +157,18 @@ public interface AgenticScoringService {
     boolean supportsToolCalling(LlmProvider provider);
 
     /**
+     * Tool choice for the FIRST call of the agentic loop. {@link ToolChoice#REQUIRED} is what stops an
+     * OpenAI judge from answering straight from visible context (see {@link SupportedJudgeProvider} for
+     * the empirical asymmetry), but langchain4j's {@code VertexAiGeminiChatModel} rejects any explicit
+     * tool choice with {@code UnsupportedFeatureException} — and since that is terminal, forcing it there
+     * failed the evaluation outright instead of scoring it. Vertex therefore gets
+     * {@link ToolChoice#AUTO}: the model may skip the tool loop, which the loop already handles (a first
+     * response with no tool-execution requests finishes it), so a possibly-tool-less evaluation replaces
+     * a guaranteed failure.
+     */
+    ToolChoice firstRoundToolChoice(LlmProvider provider);
+
+    /**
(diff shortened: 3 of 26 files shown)


YarivHashaiComet:
baz review

PR #7981, opened 2026-08-25
[NA] [SDK] fix: stop anthropic stream wrappers from swallowing exceptions in finally

trakshan-mishra:
## Details

The Anthropic stream patchers install class-level `__iter__`/`__aiter__` overrides on `anthropic.Stream`, `AsyncStream`, `MessageStream`, `AsyncMessageStream`, `BetaMessageStream`, and `BetaAsyncMessageStream`. Because the patch is class-level, every stream instance in the process runs through the wrapper — tracked or not. Each wrapper used an early `return` inside a `finally` block to skip cleanup for non-tracked streams. A `return` in `finally` silently swallows any in-flight exception, so once any opik-tracked streaming call installed the class patch, a non-tracked stream that errored mid-iteration completed silently instead of raising.

Fix: invert the guard so the early return is gone and cleanup nests under `if hasattr(...)`. Tracked streams behave exactly as before; non-tracked streams now propagate their exceptions. Also removes the Python `SyntaxWarning: 'return' in a 'finally' block` emitted from all six wrappers on import.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-

## AI-WATERMARK

AI-WATERMARK: yes
- Tools: opencode (CLI coding agent)
- Model(s): glm-5.2
- Scope: identified the bug by running the local Opik stack + instrumenting a multi-step agent loop; wrote the fix, repro, and unit tests
- Human verification: reviewed the diff, ran ruff/mypy/pytest, confirmed the test fails on upstream and passes with the fix

## Testing

- **Commands run:**
  - `ruff format --check` and `ruff check` on both changed files — clean
  - `mypy` on `stream_patchers.py` — `Success: no issues found`
  - `pytest tests/library_integration/anthropic/test_stream_patchers.py` — `12 passed` (4 parametrized tests × 3 wrapper classes each for sync + async, including beta variants)
  - `pytest tests/unit/decorator tests/unit/llm_usage` — `208 passed`
  - Standalone repro script confirmed: before fix → `BUG: exception swallowed`; after fix → `PASS (propagated)`
- **Scenarios validated:**
  - Regression: the 6 non-tracked tests fail on upstream code (exception swallowed), pass with fix
  - Tracked stream path: cleanup callback runs once with `error_info` (exception_type, message, traceback) matching the injected RuntimeError, exception still propagates
  - Non-tracked stream path: cleanup callback not called, exception propagates
  - Covers all 6 wrappers: Stream, AsyncStream, MessageStream, AsyncMessageStream, BetaMessageStream, BetaAsyncMessageStream
- **Environment:** Local Opik stack via `./opik.sh --build` (Docker), Python 3.14, anthropic SDK 1.0.0, editable SDK install
- **Pre-commit:** `make precommit` requires Java/Node toolchains for non-Python hooks; the Python-relevant hooks (ruff, ruff-format, mypy) were run directly and pass

## Documentation


trakshan-mishra:
Thanks for the review @baz-reviewer. Pushed commit 437af944 addressing both points:

**Third-party dependency in unit suite** — the expanded tests no longer import `httpx2` (use `object.__new__(cls)` to create uninitialized stream instances instead of constructing real ones). Keeping in `tests/unit/` rather than `tests/library_integration/` because:
- The `library_integration/anthropic/` suite makes **real Anthropic API calls** (requires `ANTHROPIC_API_KEY`, uses `tenacity` retries — see `test_anthropic.py:37-41`)
- These tests mock the stream objects directly and test **only Opik's patching logic** — no API calls, no network
- The `anthropic` import is a type reference (the patchers operate on `anthropic.Stream` etc.), not a live call
- Matches the existing pattern in `tests/unit/integrations/otel/` which also imports a third-party integration to test opik logic in isolation

**Five stream wrappers lack regression coverage** — now parameterized across all six wrappers (Stream, AsyncStream, MessageStream, AsyncMessageStream, BetaMessageStream, BetaAsyncMessageStream) for both sync/async and tracked/non-tracked paths. 12 test cases total. Verified:
- all 12 pass with the fix
- the 6 non-tracked tests fail on upstream (bug present), 6 tracked tests pass on both (tracked path was never broken)

trakshan-mishra:
Thanks @baz-reviewer — addressed both new comments in commit 4774fd9:

**Reported test count (line 167)** — PR testing evidence updated to report `12 passed` (4 parametrized tests × 3 wrapper classes for sync + async, including beta variants).

**Incorrect error diagnostics (line 243)** — the tracked-stream tests now assert the `error_info` fields reflect the injected exception, not just that it's non-null:
- `exception_type == 'RuntimeError'`
- `message == 'stream-blew-up'`
- `traceback` contains the test module name

Applied to both sync and async tracked-stream tests via a shared `_assert_error_info_matches` helper.

trakshan-mishra:
Confirmed the branch is synced with main (commit 57ef06a). Happy to answer any questions on the fix — the core issue was `return` inside `finally` swallowing the re-raised exception when `opik_tracked_instance` wasn't set on `self`, across all 6 stream wrapper variants (sync/async × plain stream / message-stream-manager / beta message-stream-manager). Let me know if you'd like anything else from me here.

sdks/python/tests/library_integration/anthropic/test_stream_patchers.py
```diff
@@ -0,0 +1,255 @@
+from dataclasses import dataclass
```
alexkuzmik, line 1: Please move this test to `tests/library_integration/anthropic`

alexkuzmik (changes requested):

trakshan-mishra:
Moved the test to `tests/library_integration/anthropic/` as requested. Passes there, 12 passed.

CI on this branch is stuck at `action_required` on every check because the PR is from a fork, so nothing has actually run on the new commit. Could you approve the run when you get a chance?

Also confirmed the bedrock test you couldn't run locally goes green on #8023, left the details over there.

alexkuzmik (approved):
Looks good! Thanks @trakshan-mishra

PR #7987, opened 2026-08-25
[OPIK-8084] [QA] retire release-gate: explore-feature now writes normal tiered specs

AndreiCautisanu:
## Details

The release-gate lane is no longer used, so this retires it and points `explore-feature` at the normal tiered suite instead.

It turns out the lane was never actually exercised. Before changing anything I checked what depended on it:

- **zero** specs under `tests_end_to_end/e2e/tests/_release-gate/` — only a README
- **no** CI workflow greps `@release-gate` (`grep -rl release-gate .github/workflows/` → nothing)
- the `test:release-gate` npm script has no callers

So what remained was vocabulary: tags `tag_lint.py` accepted, a taxonomy `exempt_dir`, and carve-outs in two skills plus `TESTING-TAGS.md` describing a lane that never carried a test. That's the kind of thing that quietly misleads the next reader, which is why this removes it rather than just leaving it dormant.

**`explore-feature` now produces an ordinary spec** — `tests/<area>/<name>.spec.ts` with a tier tag, an `@area:` and a `@cap:` per test — so its output lands in the same suites as everything else instead of a parallel one nobody ran. No separate lane, no version stamp.

It keeps the parts that earned their place: local-diff scope resolution (including uncommitted and untracked work), the fix-PR rule that you cover the **repro condition** rather than the easy path (otherwise the test passes against pre-fix code too — a vacuous test), the N-equivalent-surfaces rule, seeding as part of the deliverable, and the local-stack gate that catches a dev testing against a build without their change.

**Dropped the `test-plan.md` / Jira-post half.** The plan existed to hand a gate spec to QA for triage; with a normal spec there's no triage step, so the skill collapses to a thin wrapper over `writing-e2e-tests`.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-8084

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: The skill rewrite and the removal of the release-gate scaffolding across `tag_lint.py`, `taxonomy.yaml`, `package.json`, `TESTING-TAGS.md` and the `writing-e2e-tests` skill.
- Human verification: Dependency check run before deleting anything (no specs, no CI, no script callers). The vocabulary change was verified as *enforced* rather than cosmetic — see Testing.

## Testing

```
python3 tests_end_to_end/coverage/tag_lint.py --taxonomy tests_end_to_end/coverage/taxonomy.yaml --estate tests_end_to_end
  → tag-lint: 45 specs checked, 1 exempt, 0 problem(s)

cd tests_end_to_end/e2e && npx tsc --noEmit          → clean
npx playwright test tests/datasets/ --reporter=list  → 8 passed
```

The "1 exempt" is now `_seed` alone, which is the expected consequence of dropping the `_release-gate` `exempt_dir`.

**Verified the removal is real, not cosmetic.** Dropping a tag from `tag_lint.py`'s vocabulary should make it *rejected*, not merely unused — so I added a throwaway spec carrying `@release-gate` + `@release-gate:1.2.3` and confirmed the linter now fails it:

```
__probe.spec.ts:2: unrecognised tag '@release-gate' — not a tier, suite selector, or taxonomy tag
__probe.spec.ts:2: unrecognised tag '@release-gate:1.2.3' — not a tier, suite selector, or taxonomy tag
__probe.spec.ts:1: no tier tag: expected exactly one of @t1-smoke / @t2-cuj / @t3-nightly
```

That third line matters too: it shows the tier requirement now applies to specs that previously satisfied it with a suite selector alone. The probe was deleted afterwards.

**Checked I didn't over-delete.** `@t1-stsaas` and `@provider-sanity` stay in `SUITES` — they're real selectors with real callers (`optimization-studio.spec.ts`, `playground-providers.spec.ts`), and both still lint clean. `RELEASE_GATE_VERSIONED` was the only regex removed; `re` is still imported and used elsewhere in the file.

## Documentation

`TESTING-TAGS.md`, `writing-e2e-tests/SKILL.md` and `writing-e2e-tests/conventions.md` are updated in this change — their release-gate carve-outs are gone, along with the tag-grammar comment in `taxonomy.yaml`. The `_release-gate/README.md` is deleted with the directory.


CometActions:
**No test needed here.**

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/32839290252)

jverre (approved):

PR #7997, opened 2026-08-25
[OPIK-8060] [BE] [FE] fix: report the best trial's latency and cost in the runs list

alexkuzmik:
## Details

The Optimization runs list reported the **baseline** trial's latency and cost as the run's best result, with a flat `0%` delta on every row, while the run details page reported the genuine best trial (e.g. `0% / 3.2s` in the list vs `-60% / 1.3s` on the run page for the same run). Two independent causes in the optimization summary query, plus a column-naming collision that made the numbers look self-contradictory.

- **The objective score was read from the wrong place.** The candidate rollups sourced it only from the experiment-level `experiment_scores` column. Dataset runs — Studio runs and every SDK optimizer — score per trace instead and populate no `experiment_scores` at all, so no candidate had a score. And because `join_use_nulls = 0` makes an unmatched non-`Nullable` `LEFT JOIN` column arrive as `0` rather than `NULL`, the `isNotNull()` guard was inert and every candidate was scored a real `0`. All candidates tied, so the `best_*` rollups fell through to their earliest-created tie-break — which is the baseline. The score is now read from the per-experiment trace feedback scores first with the experiment-level score as fallback (mirroring the run page's `getObjectiveScoreValue` precedence) and stays `Nullable`, so the guard means something.
- **The completeness gate was missing.** The run page excludes candidates that evaluated fewer items than a full evaluation covers, because a partial average is not a result (`OPIK_7460`). The query had no such rule, so even with correct scores it could crown a trial that had only covered part of the test set — which optimizers that evaluate most trials on a subset make the common case, not an edge case. `best_*` is now restricted to candidates that covered at least as many items as the baseline, falling back to the whole set when nothing has completed, so a run never loses its best marker.
- **`Opt. cost` meant two different things.** In the runs list it was a whole run's total spend; in the run page's trials table it was the cost of a single evaluated case — the same figure that page's own overview card calls `Runtime cost`. One header, two quantities, one screen apart. The list column is now **`Optimization cost`** (matching the run page's overview card) and the trials column **`Runtime cost`** (matching that page's card and chart tooltip). Column ids are unchanged, so saved column selection, order and width survive the rename; the list column was widened to 160px to fit the unabbreviated label.

**Reviewer note — one deliberate behaviour change.** Runs with no scored trials (errored runs, or runs whose chosen objective was never scored) now report no best latency or cost instead of the baseline's figure mislabelled as the best. In the reproduction environment that is 11 of 38 runs moving from a number to `-`. This is the truthful answer and the UI already renders it, but it will read as "lost data" without context. Whole-run spend is unaffected.

No data is wrong at rest — traces, scores and costs were all recorded correctly, only the summary reading of them was off, so affected runs display correctly as soon as this ships. Nothing needs backfilling.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves OPIK-8060

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: full implementation — investigation, query fix, regression tests, column renames
- Human verification: author-directed investigation and review, including the column naming decisions; behaviour confirmed by hand in a local deployment before and after the change

## Testing

Backend (from `apps/opik-backend`):

```bash
mvn -o compile -DskipTests
mvn -o spotless:check
mvn -o test -Dtest='OptimizationsResourceTest,OptimizationsResourceFindProjectOptimizationsTest'
```

Frontend (from `apps/opik-frontend`):

```bash
npm run lint
npm run typecheck
npx vitest run src/v2/pages/OptimizationsPage/OptimizationsColumns.test.ts \
               src/v2/pages/OptimizationPage/useOptimizationColumns.test.ts
```

Scenarios validated:

- **Two new regression tests**, one per cause, added to `OptimizationsResourceTest$GetOptimizerById`:
  - a dataset-shaped run whose objective is scored on traces rather than at the experiment level — asserts `best_*` comes from the best-scoring candidate and `baseline_*` from the baseline, rather than both collapsing onto the baseline;
  - a run where the top-scoring candidate is only partially evaluated — asserts `best_*` skips it in favour of the best fully evaluated candidate.
- **Both new tests were confirmed to fail without the fix**: reverted the query to its `main` state, rebuilt, and both failed on the awaited assertions; they pass with the fix applied. Every pre-existing test that asserts a real `bestDuration`/`bestCost` builds its trials with `experimentScores`, i.e. the experiment-level shape — which is why this went unnoticed, and what the first new test covers.
- **End-to-end in a local deployment**, against a real 25-trial GEPA run: the runs list and the run details page now agree exactly (`1.3s` at `-60%`, `$0.0009`), where before the list showed `3.2s` / `$0.0010` at `0%`. Verified through the API and in the browser, on both the runs list and the run page's Overview and Trials tabs.
- **Regressions checked across the whole local dataset**: 27 of 38 runs report latency/cost; before the change 27/27 showed `best == baseline`, after it 12 show a real delta and the 15 that remain flat were each confirmed to be legitimately flat (their baseline genuinely ties for best on a complete evaluation).
- **Renamed columns** verified rendering in the browser on both screens, with the list header measured for truncation at its new width.

Environment: local dev stack (`scripts/dev-runner.sh`, backend and frontend as processes, Dockerised MySQL/ClickHouse/Redis), macOS, Chromium.

Not run: the full `apps/opik-backend` `mvn test` suite. It is a long-running integration suite and contended with the running local stack for Docker resources; the two test classes covering the changed query were run instead (75 tests, all passing), after rebasing onto the latest `main`.

## Documentation

N/A — no documentation references these columns or the affected values (checked across `*.md`, `*.mdx`, `*.ts`, `*.tsx`, `*.py`, `*.java`).

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/OptimizationDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/OptimizationDAO.java
@@ -702,13 +702,50 @@ class OptimizationDAOImpl implements OptimizationDAO {
                 WHERE ef.experiment_type NOT IN ('mini-batch', 'mutation')
             ), objective_scores_per_experiment AS (
                 SELECT
-                    ef.optimization_id,
-                    esp.experiment_id,
-                    esp.value AS objective_score
-                FROM experiment_scores_parsed esp
-                INNER JOIN experiments_final ef ON esp.experiment_id = ef.id
-                INNER JOIN optimization_final o ON ef.optimization_id = o.id
-                WHERE esp.name = o.objective_name
+                    optimization_id,
+                    experiment_id,
+                    /*
+                     * Mirrors the run page's precedence (getObjectiveScoreValue): the trace-derived feedback
+                     * score wins, the experiment-level score is the fallback.
+                     *
+                     * toNullable is load-bearing, not decoration. candidate_metrics LEFT JOINs this CTE and
+                     * guards the weighted average with isNotNull(objective_score); under join_use_nulls = 0
+                     * an unmatched non-Nullable Float64 arrives as 0, which makes that guard always true and
+                     * scores every unscored candidate a real 0. Every candidate then ties, the best_* rollups
+                     * fall through to their earliest-created tie-break, and "best" collapses onto the
+                     * baseline - so the runs list reported the baseline's latency and cost with a 0% delta
+                     * while the run page reported the genuine best trial (OPIK-8060). duration_p50 in
+                     * experiment_durations is Nullable for the same reason.
+                     */
+                    toNullable(argMin(objective_score, source_rank)) AS objective_score
+                FROM (
+                    /*
+                     * Dataset runs - Studio runs and every SDK optimizer. The objective is a per-trace
+                     * feedback score averaged per experiment, and leaving it out was the OPIK-8060 defect:
+                     * these runs populate no experiment_scores at all, so the CTE matched nothing for them.
+                     */
+                    SELECT
+                        ef.optimization_id AS optimization_id,
+                        ef.id AS experiment_id,
+                        fs.feedback_scores[o.objective_name] AS objective_score,
+                        0 AS source_rank
+                    FROM experiments_final ef
+                    INNER JOIN optimization_final o ON ef.optimization_id = o.id
+                    INNER JOIN feedback_scores_agg fs ON ef.id = fs.experiment_id
+                    WHERE mapContains(fs.feedback_scores, o.objective_name)
+                    UNION ALL
+                    -- Test-suite runs, where the objective is scored at the experiment level.
+                    SELECT
+                        ef.optimization_id AS optimization_id,
+                        esp.experiment_id AS experiment_id,
+                        esp.value AS objective_score,
+                        1 AS source_rank
+                    FROM experiment_scores_parsed esp
+                    INNER JOIN experiments_final ef ON esp.experiment_id = ef.id
+                    INNER JOIN optimization_final o ON ef.optimization_id = o.id
+                    WHERE esp.name = o.objective_name
+                )
+                GROUP BY optimization_id, experiment_id
             ), candidate_metrics AS (
                 SELECT
                     ec.optimization_id AS optim_id,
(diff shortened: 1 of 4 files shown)


CometActions:

### Access Information
- **URL:** https://pr-7997.dev.comet.com
- **Namespace:** pr-7997
- **Version:** 2.2.41-7997-merge-3103
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-7997%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-7997`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

BorisTkachenko (approved):
BE part looks good

CometActions:

The DAO fix is covered by the two tests you added in OptimizationsResourceTest: getById__whenObjectiveScoredOnTraces__bestComesFromBestScoringCandidate and getById__whenTopCandidateIsPartiallyEvaluated__bestSkipsIt pin bestDuration/bestCost to the winning trial and baselineDuration/baselineCost to the baseline, so both halves — trace-scored objectives and the partial-evaluation gate — fail if the SQL is wrong, and getById goes through the same FIND projection the runs list reads. No e2e spec asserts the runs list's metric columns (optimization-studio.list-optimization-runs is uncovered), but that is a standing gap in that page, not something this change introduces: the columns and the best-vs-baseline delta render exactly as before. Not proposing a duplicate.

**Also considered.** The two column renames — "Opt. cost" to "Optimization cost" on the runs list, and to "Runtime cost" on the run page's trials table — are copy-and-width only. Nothing in the e2e estate selects on either header (grep for "Opt. cost" across tests_end_to_end returns nothing), and there is no behaviour behind the label for a test to pin.

_areas: `optimization-studio` · also touches Backend (Java API / internal)_

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/33063915639)

PR #7998, opened 2026-08-25
[NA] [EXT] feat: log Cursor token usage, cost and model to Opik

jverre:
## Details

Cursor stopped populating `bubble.tokenCount` in February 2026, so the extension had been writing LLM spans with zero tokens and no cost. Token counts, cache usage, cost and model now come from Cursor's usage API and are attached to the span a few seconds after the turn is uploaded.

- Usage is fetched with the user's existing Cursor sign-in. No extra credentials, and **no admin role** — verified against a plain `TEAM_ROLE_MEMBER` account. It also works for users with no team.
- `conversationId` from the API equals the local `composerId`, so the join needs no heuristics. Verified 11/11 across 200 days of history.
- Each usage event is attributed to exactly one turn, using the conversation's complete list of real user-turn start times.
- Three pre-existing bugs surfaced and are fixed here, plus a packaging leak, all described below.

Fixed along the way:
- **Fabricated trace times.** Modern Cursor bubbles carry no `timingInfo` or `timestamp` (0 of 15 in a sample), so `processConversationBubbles` always fell into its synthetic `+1000ms` ladder. Every `start_time` was invented. Now read from `bubble.createdAt`.
- **`end_time` always equal to the upload time.** `Trace.end()` and `Span.end()` overwrite `endTime` with `new Date()`. Those calls are removed, since the real times are supplied at construction.
- **`.claude/settings.local.json` was being published.** `.vscodeignore` excluded `src/**` but not `.claude/**`, so the file was bundled into every vsix. It is present in `opik-0.3.1.vsix` as `extension/.claude/settings.local.json` and contains an absolute home-directory path, so it has been shipping to users since at least September 2025. Now excluded, along with the new `scripts/**`. Verified with `vsce pack`: the 0.4.0 artifact is 21 files with no `.claude`, no `scripts/`, and no path leak in any file.
- **SQLite query timeout.** `sqlite3 -json` on macOS is quadratic in cell length: one 561KB value takes 17s and a 5.5MB result set takes 56s, blowing the 30s timeout. Switched to `-ascii` with a small parser, and replaced `LIKE 'prefix%'` with prefix ranges so the index on `key` is usable. The failing query went from a 30s timeout to 72ms.

## Change checklist
- [x] User facing
- [x] Documentation update

## Issues

- Resolves #
- OPIK-

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: Reverse engineering of the Cursor local store and usage RPCs; all code in this PR (`src/cursor/usage.ts`, `scripts/verify-usage.js`, and the edits to `opik.ts`, `sessionManager.ts`, `sqlite.ts`, `cursorService.ts`, `interface.ts`, `state.ts`, `package.json`); PR description.
- Human verification: Author ran the extension in the Cursor extension host against a live account, sent messages across several conversations, and confirmed the spans in Opik. Author caught the duplicate-attribution bug by reading the logged data, which led to the attribution rewrite.

## Testing

Commands run, all from `extensions/cursor`:

```
npm run compile     # tsc, clean
npm run lint        # 0 errors, 9 warnings (all pre-existing in sessionManager.ts)
npm run esbuild     # bundles clean
npm run verify-usage
```

`npm run verify-usage` is added in this PR. It runs the whole pipeline except the Opik write: reads the newest local conversation, fetches its usage, runs the compiled attribution, and asserts both that per-turn totals reconcile against the API and that no two turns share a usage payload. It exits non-zero on either failure.

Scenarios validated:
- **Reconciliation.** Per-turn sums equal the API totals exactly, over 200 days of history: 32 turns, 29 events, 29 turns enriched, 0 orphans (~91% coverage; turns with no chargeable request are left without usage by design).
- **No double counting.** A regression was found during review: two spans in one thread received the whole conversation's total. Root cause was deriving turn boundaries from the shrinking pending list. After the fix, a 5-turn conversation yields 5 distinct payloads summing to the API total, and repeat ticks patch nothing (idempotent).
- **Rapid turns.** Five messages sent 2-4 seconds apart. This is what exposed the clock-grace bug; a grace window wider than the gap between turns makes a later turn swallow an earlier one's events. Grace is now 0, since the event is stamped at request completion.
- **Availability lag.** Measured at 3.8s, 5.0s and 6.6s by polling before sending. The backoff schedule starts at 5s.
- **Rate limiting.** A 15-request burst at ~109 req/min was not throttled; median latency 524ms. The adaptive schedule costs 2-4 requests per turn and nothing when idle.
- **Query performance.** The previously failing query now returns 75 rows and 5.5MB in 72ms, with all 75 values round-tripping as valid JSON.
- **Parser safety.** Zero rows in `composerData`, `bubbleId` or `ItemTable` contain the `0x1F`/`0x1E` separators used by `-ascii`.

Not run: there is no automated test suite in this extension (`pretest` exists but no tests are defined), so all of the above is manual plus the new `verify-usage` script. Adding real tests is worth a follow-up.

Environment: macOS 15, Cursor 3.17.19, system sqlite3 3.51.0, Node 22.

## Documentation

Added a "Token usage and cost" section to `apps/opik-documentation/documentation/fern/docs-v2/integrations/cursor.mdx`, covering what is recorded, that it uses the existing Cursor sign-in with no extra key and no admin role, the few-second delay before usage appears, turns that carry no usage, and the setting to disable it.

Only the v2 page is updated. The v1 tree is being removed in separate in-flight work, so editing it would conflict.

### Release note

This bumps the extension to 0.4.0. Publishing is the usual two steps: merge to `main`, then manually dispatch the **Publish Cursor Extension** workflow, which packages with `vsce pack` and pushes to Open VSX.

Two things reviewers should weigh:

1. `GetFilteredUsageEvents` on `api2.cursor.sh` is a **private, undocumented endpoint**. Cursor can change or remove it. Every call is wrapped so a failure logs, retries on a backoff, and then gives up; the worst case is a trace with no usage, never broken tracing. `npm run verify-usage` is the canary.
2. The extension reads the Cursor session token from the local `state.vscdb` and sends it only to `api2.cursor.sh`, which issued it. The host is hard-coded deliberately.

### DIFF
--- a/apps/opik-documentation/documentation/fern/docs-v2/integrations/cursor.mdx
+++ b/apps/opik-documentation/documentation/fern/docs-v2/integrations/cursor.mdx
@@ -75,6 +75,22 @@ The extension automatically registers an Opik MCP (Model Context Protocol) serve
 
 The MCP server integration is enabled by default and requires no additional configuration. For more details about the MCP server, see the [opik-mcp package](https://www.npmjs.com/package/opik-mcp).
 
+## Token usage and cost
+
+Each logged conversation turn also records how many tokens it used and what it cost. You will
+see the model, the prompt and completion tokens, the cache read and write tokens, and an
+estimated cost on the LLM span of every trace.
+
+These numbers come from Cursor itself, using the Cursor account you are already signed in to.
+You do not need a separate API key, and you do not need to be a team admin.
+
+The token counts arrive a few seconds after the conversation is logged, so a trace may briefly
+appear without them. A turn that did not result in a billable request, such as one you
+cancelled, keeps its trace but has no usage attached.
+
+To turn this off, set `Opik: Usage Enrichment Enabled` to false in the extension settings. Your
+conversations are still logged, just without token counts or cost.
+
 ## Viewing your conversations
 
 Once the extension is installed and configured, all your conversations will be available in the `cursor` project in Opik. To
--- a/extensions/cursor/.vscodeignore
+++ b/extensions/cursor/.vscodeignore
@@ -14,6 +14,8 @@ node_modules/**
 .eslintrc.json
 .prettierrc
 .github/**
+.claude/**
+scripts/**
 *.vsix
 test/**
 .npmignore
--- a/extensions/cursor/package-lock.json
+++ b/extensions/cursor/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "opik",
-  "version": "0.3.4",
+  "version": "0.4.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "opik",
-      "version": "0.3.4",
+      "version": "0.4.0",
       "dependencies": {
         "@sentry/node": "^10.38.0",
         "glob": "^11.1.0",
--- a/extensions/cursor/package.json
+++ b/extensions/cursor/package.json
@@ -4,7 +4,7 @@
   "description": "Export your chat history from Cursor to Opik.",
   "repository": "https://github.com/comet-ml/opik",
   "publisher": "opik",
-  "version": "0.3.4",
+  "version": "0.4.0",
   "preview": false,
   "engines": {
     "vscode": "^1.90.0"
@@ -66,6 +66,16 @@
         }
       },
       {
+        "title": "Usage Tracking",
+        "properties": {
+          "opik.usageEnrichment.enabled": {
+            "type": "boolean",
+            "default": true,
+            "markdownDescription": "Fetch token counts and cost from Cursor and attach them to the logged spans. Uses your existing Cursor sign-in; no extra credentials are needed. Turn this off to log conversations without usage."
+          }
+        }
+      },
+      {
         "title": "Advanced",
         "properties": {
           "opik.enableDebugLogs": {
@@ -88,7 +98,8 @@
     "watch": "tsc -watch -p ./",
     "pretest": "npm run compile && npm run lint",
     "lint": "eslint src",
-    "test": "vscode-test"
+    "test": "vscode-test",
+    "verify-usage": "node scripts/verify-usage.js"
   },
   "devDependencies": {
     "@types/mocha": "^10.0.10",
--- a/extensions/cursor/scripts/verify-usage.js
+++ b/extensions/cursor/scripts/verify-usage.js
@@ -1,0 +1,155 @@
+#!/usr/bin/env node
+/**
+ * End-to-end check of the Cursor usage pipeline, without the extension host.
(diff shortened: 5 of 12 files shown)


CometActions:

All nine files are in `extensions/cursor`, which the e2e estate cannot drive: `UsageEnricher` needs a signed-in Cursor account's `cursorAuth/accessToken` out of a local `state.vscdb` plus live `api2.cursor.sh` GetMe/GetFilteredUsageEvents, and `scripts/verify-usage.js` needs the same (and a macOS path). Nothing covers it today — there are no tests under `extensions/cursor`, and the closest e2e spec, `trace-explore/trace-spans-depth.spec.ts:38` (`@cap:traces.span-model-cost-tokens`), seeds a span through the SDK with known usage and asserts the panel renders model/total_tokens/cost, so it passes regardless of what the extension writes. Worth saying that the unconditional half ships even with `opik.usageEnrichment.enabled` off: every turn now gets an LLM span (previously only turns whose bubbles carried `tokenCount`), and start/end times now come from bubble `createdAt` rather than `trace.end()`'s upload time. Once an extension-host harness with a fixture `state.vscdb` and a stubbed DashboardService exists this is testable end to end — but `attributeUsageToTurns` is pure and exported, so a unit test over turn boundaries (`CLOCK_GRACE_MS = 0`, multi-request turns, an event before the first turn) would cover the riskiest part with no estate work at all.

**Testable once (platform):** The estate can run the Cursor extension host against a checked-in fixture `state.vscdb` (composerData + bubbleId rows + a `cursorAuth/accessToken`) with `api2.cursor.sh/aiserver.v1.DashboardService` GetMe and GetFilteredUsageEvents stubbed. Neither the extension-host harness nor outbound HTTP interception exists today, and `extensions/cursor` has no test directory despite its `"test": "vscode-test"` script.

**What a test would assert:** The LLM span under a `cursor-chat` trace in the configured `opik.projectName`: model + provider `cursor` at creation, then a patch a few seconds later via `PATCH /v1/private/spans/{id}` carrying prompt/completion/total tokens, cache read/write tokens and `totalEstimatedCost` (chargedCents / 100), rendered on the trace panel's span detail. Two dimensions a test would have to vary: `opik.usageEnrichment.enabled` (default true — off means spans stay unpatched but the new span-per-turn and real `createdAt` timestamps still apply), and elapsed time, since the patch arrives on a 5s/5s/10s/30s/60s/5m/5m retry schedule and gives up after the last attempt.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/32887729513)

_Re-checked after a push on 25 Aug 19:08 UTC — nothing the verdict depends on changed._

PR #8011, opened 2026-08-26
[NA] [EXT] fix: include cache tokens in reported prompt and total tokens

jverre:
## Details

Cursor reports `inputTokens` excluding cache, the same way Anthropic does. The span was written with `prompt_tokens` set to `inputTokens` alone, so an agentic turn that read 140k tokens from cache reported **28,726** total tokens instead of **169,142** — a 5.9x undercount. Cost was never affected, because Cursor's `chargedCents` already prices every token.

- `prompt_tokens` now adds cache read and cache write back in, matching `get_billable_tokens()` in `sdks/python/src/opik/llm_usage/anthropic_usage.py:40`. The cache figures are still reported separately, which is the house convention (see `opik_usage.py:150`) and not double counting.
- The mapping moved into an exported `toSpanUsage()` so `scripts/verify-usage.js` asserts the real mapping instead of a copy.

Follows #7998, which is already merged. Ships as extension 0.4.1.

### Reported case

Trace `01a03d17-7063-71ce-ad81-a67c98ee321e`, one user message and 39 assistant messages:

| Field | Before | After |
|---|---|---|
| `prompt_tokens` | 26,591 | **167,007** |
| `completion_tokens` | 2,135 | 2,135 |
| `total_tokens` | 28,726 | **169,142** |
| `cache_read_input_tokens` | 140,416 | 140,416 |
| `total_estimated_cost` | 0.2724 | 0.2724 (unchanged) |

For the record, two things this was **not**. Cursor emits exactly one usage event for the whole agentic turn, and the span copied it faithfully, so attribution was not at fault. And the event does cover the whole loop rather than the final call: Cursor's own accounting put the final prompt at 38,179 tokens, while the event reports 167,007 prompt tokens, roughly 4.4x that, consistent with several round trips.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: Diagnosis of the undercount, the `toSpanUsage()` change in `src/opik.ts`, the mapping assertion in `scripts/verify-usage.js`, the version bump, and this PR description.
- Human verification: Reported by the author from a real logged trace. Author confirmed cost looked correct while token counts did not, which is what localised the fault to the mapping rather than to attribution.

## Testing

```
npm run compile     # tsc, clean
npm run lint        # 0 errors, 9 warnings (all pre-existing in sessionManager.ts)
npm run esbuild     # bundles clean
npm run verify-usage
```

`verify-usage` now prints the emitted mapping per turn and fails if `prompt_tokens` omits cache tokens, if `total_tokens` is not `prompt + completion`, or if the cache fields do not round-trip. It previously reconciled only against Cursor's own event totals, which is why it passed while the Opik mapping was wrong.

End-to-end against the real API with the reported values, patching a span and reading it back:

```
opik sdk: 1.10.9
💰 Patched span ...: 167007 prompt (26591 fresh + 140416 cache read + 0 cache write)
   / 2135 completion, 27.2400c across 1 request(s)
read back http 200
  usage: {"total_tokens":169142,"completion_tokens":2135,"prompt_tokens":167007,
          "cache_creation_input_tokens":0,"cache_read_input_tokens":140416}
  total_estimated_cost: 0.2724
PATCH SHAPE OK
```

Regression check unchanged on a five-turn conversation: 5/5 coverage, 0 duplicate payloads, `reconcile: PASS`, `mapping: prompt_tokens includes cache tokens`.

Not run: there is still no automated test suite in this extension, so the above is manual plus the `verify-usage` script.

Environment: macOS 15, Cursor 3.17.19, opik SDK 1.10.9, Node 22.

## Documentation

No change needed. The docs section added in #7998 describes what is recorded without quoting a token convention.

Worth flagging for whoever reviews: spans written by 0.4.0 keep the undercounted `prompt_tokens` and `total_tokens`. Nothing here backfills them, so any token rollup spanning both versions will be inconsistent. Cost is correct in both.

### DIFF
--- a/extensions/cursor/package-lock.json
+++ b/extensions/cursor/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "opik",
-  "version": "0.4.0",
+  "version": "0.4.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "opik",
-      "version": "0.4.0",
+      "version": "0.4.1",
       "dependencies": {
         "@sentry/node": "^10.38.0",
         "glob": "^11.1.0",
--- a/extensions/cursor/package.json
+++ b/extensions/cursor/package.json
@@ -4,7 +4,7 @@
   "description": "Export your chat history from Cursor to Opik.",
   "repository": "https://github.com/comet-ml/opik",
   "publisher": "opik",
-  "version": "0.4.0",
+  "version": "0.4.1",
   "preview": false,
   "engines": {
     "vscode": "^1.90.0"
--- a/extensions/cursor/scripts/verify-usage.js
+++ b/extensions/cursor/scripts/verify-usage.js
@@ -20,6 +20,7 @@ Module._load = function (request) {
 
 const { attributeUsageToTurns } = require('../out/cursor/usage.js');
 const { executeQuery } = require('../out/cursor/sqlite.js');
+const { toSpanUsage } = require('../out/opik.js');
 
 const DB = path.join(
   os.homedir(),
@@ -113,10 +114,15 @@ async function rpc(method, token, body) {
     sum.output += usage.outputTokens;
     sum.cacheRead += usage.cacheReadTokens;
     sum.cents += usage.chargedCents;
+    const emitted = toSpanUsage(usage);
     console.log(
       `  turn ${i + 1} [${at}] reqs=${usage.requestCount} in=${usage.inputTokens} ` +
         `out=${usage.outputTokens} cacheR=${usage.cacheReadTokens} ` +
         `cents=${usage.chargedCents.toFixed(4)} model=${usage.models.join(',')}`
+    );
+    console.log(
+      `        -> opik prompt_tokens=${emitted.prompt_tokens} ` +
+        `completion_tokens=${emitted.completion_tokens} total_tokens=${emitted.total_tokens}`
     );
   });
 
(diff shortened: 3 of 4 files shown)


CometActions:

Real behaviour change: prompt_tokens and total_tokens on Cursor spans now add cacheReadTokens + cacheWriteTokens back in, so on a cache-heavy turn the numbers on the span panel jump several times over for anyone running the extension. Nothing in tests_end_to_end/ can reach it — driving toSpanUsage needs a Cursor state.vscdb with real chat history plus a live api2.cursor.sh token, and the estate only seeds through the Python SDK and the UI. The one spec that looks close, trace-explore/trace-spans-depth.spec.ts:38 (@cap:traces.span-model-cost-tokens), asserts total_tokens on an LLM span the SDK seeded with a fixed value, so it passes identically before and after this diff. scripts/verify-usage.js is the only guard today, and it recomputes the same inputTokens + cacheRead + cacheWrite expression it is checking, against a hardcoded macOS ~/Library path with a signed-in Cursor session — useful by hand, not something CI will ever run. Recording this as a deferral rather than closing it: the missing piece is a harness for the extension, not a test nobody wrote.

**Testable once (platform):** the e2e estate can drive the Cursor extension headlessly — a seeded state.vscdb fixture plus a stub for the api2.cursor.sh DashboardService usage RPC, so applyTurnUsage can run against an OSS Opik and the resulting span be read back over the API. Alternatively, a plain unit test on the exported toSpanUsage in extensions/cursor (npm test / vscode-test is already wired in package.json) would cover the mapping without any of that.

**What a test would assert:** prompt_tokens / total_tokens on spans in the `cursor` project — the token block on the span detail panel, and GET /v1/private/spans/{id}. The axis is the producer, not the endpoint: only spans written by the Cursor extension's applyTurnUsage change, and only for turns that had cache reads or writes, so a test that seeds spans through the SDK cannot vary it. Note the taxonomy has no area for extensions/ at all — it is neither one of the 18 UI areas nor listed under rules.release_surface.untracked — so this diff attributes to zero areas today; traces.span-model-cost-tokens is named here because it is the surface a user reads the wrong number on, not because the extension is enumerated.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/32947710658)

PR #8013, opened 2026-08-26
[NA] [EXT] feat: log cursor tool calls and model calls as nested spans

jverre:
## Details

The Cursor extension logged one trace and one span per turn, so an agent run that read ten files and ran five commands showed only the question and the final answer. This logs each model call as an `llm` span and each tool call as a `tool` span, nested under an `llm_turn` span that carries the token usage.

The design is driven by a scan of a real Cursor database — 19,691 bubbles. `extensions/cursor/NESTED-SPANS.md` records the evidence behind each rule; happy to drop that file if you would rather it lived elsewhere.

```
trace  cursor-chat                       input = question, output = answer
└── span  llm_turn    type=llm           same input/output, carries the turn usage
    ├── span  assistant       type=llm       model call 1
    ├── span  read_file       type=tool
    ├── span  assistant       type=llm       model call 2
    ├── span  search_replace  type=tool
    └── span  cursor-error    type=general
```

- `llm_turn` keeps the trace input and output, so `UsageEnricher` and `applyTurnUsage()` are untouched. It is the only span with usage: Cursor records tokens once per turn and never per model call. Only 477 of 19,691 bubbles carry a non-zero `tokenCount`, always the last bubble of a turn, and the billing API reports ~1 request per turn. Splitting that total across the child spans would invent numbers.
- The trace output now interleaves the tool names with the assistant messages, so a turn reads `Let me check.` / `[read_file]` / `[grep]` / `The import is missing.`
- Tool payloads are capped at 10,000 characters and base64 images are replaced, which takes a browser screenshot result from a 157 KB median to 10 KB. One bubble in the database is 13.5 MB.
- `metadata.userMessages` and `metadata.aiMessages` are removed. Nothing read them, and they were larger than the spans that replace them.

Three fixes fall out of the real data:

- A bubble whose only `toolFormerData` is a status placeholder is not a tool call. 1,237 ordinary bubbles look like that.
- Error bubbles are absent from `fullConversationHeadersOnly`, so the old sort pushed them to the end and attached them to the wrong turn. They now go in by time. Bubbles from branches the user edited away stay out — one composer holds 833 bubbles for 748 headers.
- Reasoning flushed at the same millisecond as the tool before it used to produce an `llm` span exactly overlapping that tool span.

All of it sits behind `opik.detailedSpans.enabled`, default on. Version bumped to 0.5.0.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-

No ticket. Follow-up to the cache token work in #8011.

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: Data analysis of the local Cursor database, the span design, the implementation, the unit tests, and the replay script.
- Human verification: The author drove the design over several rounds, chose the `llm_turn` shape, and reviewed the replay output against real conversations. The unit tests and the replay regression were run locally and pass. Not yet exercised end to end against a live Opik instance — the span tree has not been eyeballed in the UI, and the usage patch on `llm_turn` has not been confirmed there.

## Testing

Everything below was run from `extensions/cursor`.

```
npm run compile && npm run test-spans     # 33 tests, all passing
npm run lint                              # 0 errors, 9 pre-existing warnings (unchanged from main)
npm run build                             # packages a 1.03 MB vsix
```

`scripts/test-spans.js` is new and runs on `node:test`, matching the existing `verify-usage.js` pattern rather than adding a VS Code test host. It covers the placeholder `toolFormerData` trap, both forms of the `type` field, the numeric tool id map, the timing rules (parallel dispatch, a negative `thinkingDurationMs`, the tool-and-reasoning timestamp collision, the error point event), both model-call cut shapes, truncation, the base64 strip, the span cap, all four bubble ordering cases, and the interleaved trace output.

`scripts/replay-composer.js` is also new. It builds the trace and spans from a real composer and prints them without uploading, flagging negative durations, spans that start before their turn, and spans that start before the span above them.

```
node scripts/replay-composer.js --list
node scripts/replay-composer.js <composerId> --full
```

Run over the 30 largest conversations in a real database — **9,330 spans** — it reports **2 warnings**, both in one composer where Cursor's own bubble order goes backwards by 6 s and 2 s. No negative durations anywhere.

Payload measured per composer (span payload against the raw bubble copy the old metadata carried):

| Composer | Spans | Span payload | Old raw bubble copy |
| --- | --- | --- | --- |
| `9bb91e1f` | 613 (286 llm, 324 tool, 3 general) | 1.97 MB | 6.20 MB |
| `32f1ff14` | 844 (366 llm, 477 tool, 1 general) | 1.92 MB | 15.98 MB |
| `8f6dd4ee` | 296 (73 llm, 223 tool) | 1.13 MB | 8.04 MB |
| `5ea1babd` | 340 (143 llm, 197 tool) | 0.67 MB | 1.90 MB |

Both conversation shapes are covered: reasoning models such as `9bb91e1f` (275 thinking bubbles) and non-reasoning ones such as `32f1ff14` (zero).

**Not run:** the end-to-end path into a live Opik instance. That needs an API key and a running backend, and it is the one thing left to check before merge — that the span tree renders as a waterfall and that the usage patch lands on `llm_turn`.

## Documentation

Three new settings are documented in `package.json`, so they carry their own descriptions in the settings UI:

- `opik.detailedSpans.enabled` (default `true`) — the off switch, which also serves as the mitigation for terminal output or file content that a user does not want logged.
- `opik.detailedSpans.maxPayloadChars` (default `10000`).
- `opik.detailedSpans.maxSpansPerTurn` (default `200`).

`extensions/cursor/NESTED-SPANS.md` records the database findings, the span design, and the verification results. No changes to the Fern docs site — this extension is documented by its own README, which needs no update because the behaviour is automatic.

### DIFF
--- a/extensions/cursor/NESTED-SPANS.md
+++ b/extensions/cursor/NESTED-SPANS.md
@@ -1,0 +1,463 @@
+# Nested spans for the Cursor extension
+
+Sections 1 and 2 record the data this design is based on. Sections 3 and 4
+describe what was built.
+
+## 1. Where we were
+
+The extension writes one trace and one span for each user turn.
+
+- `src/cursor/sessionManager.ts` groups the bubbles of a composer into turns.
+  One turn is one user bubble plus all the assistant bubbles that follow it.
+- `createTraceFromBubbleGroup()` joins the text of the assistant bubbles into a
+  single string. It strips the markers `⛢Thought☤`, `⛢Action☤` and
+  `⛢RawAction☤`.
+- `src/opik.ts` `logTracesToOpik()` creates the trace, then one child span with
+  `type: 'llm'`.
+- `src/cursor/usage.ts` `UsageEnricher` patches that span later with the token
+  counts and the cost from the Cursor usage API.
+
+The raw bubbles are copied into `metadata.userMessages` and
+`metadata.aiMessages`. Nothing in the UI reads them, and they can be very large.
+
+**Result:** you see the question and the final answer. You do not see which
+tools ran, what they returned, how long each step took, or where the agent
+failed.
+
+## 2. What the real data contains
+
+I scanned the local Cursor database:
+`~/Library/Application Support/Cursor/User/globalStorage/state.vscdb`.
+It holds **19,691 bubbles**. The numbers below come from that scan.
+
+### 2.1 Field coverage
+
+| Field | Bubbles | Share | Meaning |
+| --- | --- | --- | --- |
+| `createdAt` | 19,690 | 100% | ISO time, millisecond precision |
+| `type` | 19,690 | 100% | `1` = user, `2` = assistant |
+| `capabilityType` | 12,332 | 62.6% | `15` = tool call, `30` = thinking |
+| `toolFormerData` | 11,139 | 56.6% | the tool call record |
+| `text` | 6,949 | 35.3% | message text |
+| `usageUuid` | 5,912 | 30.0% | groups all bubbles of one turn |
+| `thinking` | 2,183 | 11.1% | reasoning text |
+| `thinkingDurationMs` | 2,177 | 11.1% | exact duration of the reasoning |
+| `requestId` | 2,131 | 10.8% | request id of the turn |
+| `modelInfo` | 2,131 | 10.8% | `{ modelName }` |
+| `codeBlocks` | 1,225 | 6.2% | code the model proposed |
+| `contextWindowStatusAtCreation` | 931 | 4.7% | tokens used and token limit |
+| `timingInfo` | 666 | 3.4% | client start, send, settle and end time |
+| `todos` | 136 | 0.7% | the to-do list state |
+| `webCitations` | 30 | 0.2% | web sources |
+| `errorDetails` | 26 | 0.1% | message and stack trace of a failure |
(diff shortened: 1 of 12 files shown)


CometActions:

This is real user-facing behaviour: a Cursor turn now logs child `llm`/`tool`/`general` spans under a parent renamed to `llm_turn`, `buildTurnOutput` means a turn that is only tool calls no longer gets dropped for having an empty output, `orderBubbles` now discards bubbles missing from `fullConversationHeadersOnly`, and the raw bubbles leave trace metadata — all of it visible in Opik's span tree. I still can't route it: QA's taxonomy.yaml has no area for `extensions/` at all (0 of your 10 product files mapped), so there is no capability key to file this under, and the Playwright estate can't drive Cursor — no VS Code extension host, no way to seed a `state.vscdb`. `trace-explore/trace-spans-depth.spec.ts` does assert the nested span tree, but it seeds its own spans through the SDK, so it would pass whatever the extension built. `scripts/test-spans.js` looks like the right layer for this. No e2e test proposed; the missing taxonomy area is ours to fix.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/32955921630)

PR #8014, opened 2026-08-26
[NA] [DOCS] docs: set the style for PR descriptions in agent rules

jverre:
## Details

**Before**

- The rules said "follow the PR template" and nothing about how to write inside it.
- `## Details` came out as a long prose summary of the diff, with a motivation paragraph and a benefits list.
- `.claude/rules/git-workflow.md` carried a hand-edited `## PR Body` block that was missing from its `.agents/` source, so the next `make claude` would have deleted it.

**After**

- `## Details` has a stated style: short (3–10 bullets), bulleted, authoritative, and about behavior a user can observe. Motivation paragraphs, "this PR …", approach summaries, benefits lists and diff restatements are named as things to cut.
- Three shapes are offered — Before / After lists, a flat bullet list, or one or two lines — and the agent picks the one that fits instead of forcing a fixed template.
- The guidance reaches every surface that writes a description: the always-on git rule, `/comet:create-pr`, the `_pr-description-sync` sub-skill, and the `documentation` and `write-docs` skills.
- A re-sync keeps the shape a PR already uses, so it no longer rewrites a hand-picked format.
- The `## PR Body` block now lives in `.agents/rules/git-workflow.mdc`, so `make claude` regenerates it instead of dropping it.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- NA — repository tooling change, no ticket.

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: full implementation
- Human verification: the author set the requirements and reviewed the rule text; the style was derived from the author's own feedback in earlier sessions.

## Testing

- `make claude` and `make codex` — regenerated `.claude/rules/`, `.claude/commands/`, `.claude/skills/` and `AGENTS.override.md` from the `.agents/` sources. Confirmed the new `## PR Body` block survives the sync and that the stale `.claude/commands/comet/create-pr.md` copy is refreshed.
- Checked that no hook in `.pre-commit-config.yaml` matches `.agents/**` — every hook is gated by a `files:` regex scoped to `sdks/`, `apps/`, `scripts/precommit-*`, Dockerfiles, YAML or `.github`.
- This PR description was written under the new rule.
- Not run: `pre-commit` itself. The binary is not installed on this machine, and no hook applies to the changed paths.

## Documentation

N/A — the change is agent configuration under `.agents/`, not product documentation.

### DIFF
--- a/.agents/commands/comet/create-pr.md
+++ b/.agents/commands/comet/create-pr.md
@@ -181,7 +181,20 @@ This workflow will:
   - **Fill every `##` section** in the template — the PR linter requires all sections to be present
   - If a section is not applicable, write "N/A" rather than removing it
 - **Section guidance**:
-  - **Details**: Implementation summary from git analysis (replace the HTML comment placeholder)
+  - **Details**: Replace the HTML comment placeholder with what changes for a user. A reviewer reads the diff for the code; this section tells them what is different when they use the product. Style:
+
+    - **Short.** Most PRs need 3–10 bullets. If it runs longer, the section is doing the diff's job — cut it.
+    - **Bullets, not prose paragraphs.** One behavior per bullet. Nest one level for sub-cases.
+    - **Authoritative.** State what happens: "The run is scored once." Not "This should now mean that the run will be scored once."
+    - **No fluff.** No motivation paragraph, no "this PR …", no approach summary, no benefits list, no restating the diff.
+    - **Observable behavior first.** What the UI shows, what the API returns, what gets scored, stored or logged. Name a class, method or file only when the behavior makes no sense without it.
+
+    Pick the shape that fits the change — do not force one:
+
+    - **Before / After bullet lists** when a behavior changed and the contrast is the point.
+    - **A flat bullet list** for a new capability, where there is no "before".
+    - **One or two lines** when users cannot see the change (refactor, dependency bump) — say what is unchanged and what improved, then stop.
+
   - **Change checklist**: Auto-check based on file types changed (user-facing for UI changes, documentation for docs)
   - **Issues**: Link to Jira ticket (e.g., `OPIK-2180`) or GitHub issue, or "NA" for hotfixes. List every ticket this PR **resolves** here with a normal hyphenated key.
   - **Jira key convention across the whole body** (see git-workflow rule): the GitHub for Jira app links any `OPIK-<digits>` it finds in the PR body to that ticket's Development panel, and the link can't be removed. So in **all** sections (Details, Testing, etc.) and in commit messages: tickets this PR **resolves** keep the hyphen (`OPIK-1234`) — links/URLs fine and wanted. Tickets **related but not resolved** here (escalations, references to older tickets — anything not in the title/branch) must be written with an underscore (`OPIK_7000`) and with **no** Jira URL (the URL contains the hyphenated key and links anyway). Apply this when generating every section below.
--- a/.agents/rules/git-workflow.mdc
+++ b/.agents/rules/git-workflow.mdc
@@ -28,6 +28,36 @@ Follow-up commits (preferred):
 ## PR Title
 `[<TICKET-KEY>] [COMPONENT] <type>: description`
 
+## PR Body
+Follow `.github/pull_request_template.md` — read the FULL template file before drafting.
+CI (`pr-lint`) fails the PR if any of these exact headings is missing:
+`## Details`, `## Change checklist`, `## Issues`, `## Testing`, `## Documentation`.
+Fill in `## AI-WATERMARK` too (Tools, Model(s), Scope, Human verification).
+Never substitute your own structure (e.g. `## Summary` / `## Test Plan`).
+A section that does not apply gets `N/A` — never delete a heading.
+
+### `## Details` — style
+Write what changes for a user. A reviewer reads the diff for the code; this section
+tells them what is different when they use the product.
+
+- **Short.** Most PRs need 3–10 bullets. If it runs longer, the section is doing the
+  diff's job — cut it.
+- **Bullets, not prose paragraphs.** One behavior per bullet. Nest one level for sub-cases.
+- **Authoritative.** State what happens: "The run is scored once." Not "This should now
+  mean that the run will be scored once."
+- **No fluff.** No motivation paragraph, no "this PR …", no approach summary, no
+  benefits list, no restating the diff.
+- **Observable behavior first.** What the UI shows, what the API returns, what gets
+  scored, stored or logged. Name a class, method or file only when the behavior makes
+  no sense without it.
+
+Pick the shape that fits the change — do not force one:
+
+- **Before / After bullet lists** when a behavior changed and the contrast is the point.
+- **A flat bullet list** for a new capability, where there is no "before".
+- **One or two lines** when users cannot see the change (refactor, dependency bump) —
+  say what is unchanged and what improved, then stop.
+
 ## Jira Key References (commit messages & PR body)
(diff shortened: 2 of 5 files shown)


CometActions:
**No test needed here.**

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/32969167665)

.agents/commands/comet/create-pr.md
```diff
@@ -181,7 +181,20 @@ This workflow will:
...
   - If a section is not applicable, write "N/A" rather than removing it
 - **Section guidance**:
-  - **Details**: Implementation summary from git analysis (replace the HTML comment placeholder)
+  - **Details**: Replace the HTML comment placeholder with what changes for a user. A reviewer reads the diff for the code; this section tells them what is different when they use the product. Style:
+
+    - **Short.** Most PRs need 3–10 bullets. If it runs longer, the section is doing the diff's job — cut it.
+    - **Bullets, not prose paragraphs.** One behavior per bullet. Nest one level for sub-cases.
+    - **Authoritative.** State what happens: "The run is scored once." Not "This should now mean that the run will be scored once."
+    - **No fluff.** No motivation paragraph, no "this PR …", no approach summary, no benefits list, no restating the diff.
+    - **Observable behavior first.** What the UI shows, what the API returns, what gets scored, stored or logged. Name a class, method or file only when the behavior makes no sense without it.
```
JetoPistola, line 190: 🚫 **blocker** | Consistency

The new style contradicts the template it tells the agent to treat as canonical, and the template wins by default.

`.github/pull_request_template.md` is untouched by this PR, and its `## Details` placeholder still says:

> Short 2-3 lines single paragraph explaining whats changed and the **motivation (why)**.

The new guidance says the opposite — bullets not prose, and "No motivation paragraph". That isn't a soft conflict: line 169 of this file calls the template **"the source of truth"**, step 7 has the agent `cat` it at runtime, and all five updated surfaces say "read the FULL template file before drafting". The template is also what a human sees in the GitHub compose box, so it's the copy that actually gets followed.

Worth separating the two things in play, because only one of them is CI-load-bearing:

- **Structure** — the exact headings `## Details`, `## Change checklist`, `## Issues`, `## Testing`, `## Documentation`. `pr-lint.yml` greps for these verbatim, plus a non-empty `## Details` and an issue token under `## Issues` for non-`[NA]` titles. This must not change, and the new rule correctly preserves it.
- **Content style** — the prose *inside* `## Details`. `pr-lint` never inspects it (only that it is non-empty), so this is entirely free for the new rule to redefine.

So there are two clean ways out, and either resolves it:

1. **Update the template placeholder** so both copies agree. No `pr-lint.yml` change is needed — the headings are untouched.
2. **State the precedence explicitly** here: the headings come from the template and are mandatory for a green `pr-lint`; the writing style inside `## Details` comes from this rule and supersedes the placeholder text.

Option 1 is one line and removes the conflict at the source rather than asking every agent to resolve it. Suggested wording for this line, making the precedence explicit:

````suggestion
  - **Details**: Replace the HTML comment placeholder with what changes for a user. A reviewer reads the diff for the code; this section tells them what is different when they use the product. The template's `##` headings stay exactly as-is — `pr-lint` greps for them verbatim; this style governs only the prose inside `## Details`, which `pr-lint` does not inspect beyond being non-empty, and it supersedes the placeholder's "2-3 lines single paragraph … motivation" text. Style:
````

…ideally paired with replacing the template's own placeholder comment so the two agree at the source:

```
REPLACE ME WITH:
What changes for a user — short (3-10 bullets), bulleted, authoritative.
No motivation paragraph, no "this PR ...", no restating the diff.
```

File: .agents/rules/git-workflow.mdc [hunk 2]
```diff
@@ -28,6 +28,36 @@ Follow-up commits (preferred):
...
+  scored, stored or logged. Name a class, method or file only when the behavior makes
+  no sense without it.
+
+Pick the shape that fits the change — do not force one:
+
+- **Before / After bullet lists** when a behavior changed and the contrast is the point.
+- **A flat bullet list** for a new capability, where there is no "before".
+- **One or two lines** when users cannot see the change (refactor, dependency bump) —
+  say what is unchanged and what improved, then stop.
+
```
JetoPistola, line 60: 💡 **suggestion** | Maintainability (DRY)

This ~18-line style block is now copied verbatim into four files: here, `create-pr.md`, `documentation/SKILL.md`, and `write-docs/SKILL.md`.

The thing that makes the copies redundant is this rule's own frontmatter — `alwaysApply: true` means the text is already in every session's context before any skill or command loads. The three other copies buy no additional reach, and they are three more places to drift out of sync the next time the style is tuned.

In fairness this follows an existing convention — `documentation/SKILL.md` and `write-docs/SKILL.md` already carried verbatim-identical PR-description blocks on `main`, so the PR didn't introduce the pattern. But making this rule the single source and having the other surfaces point at it (e.g. "for `## Details` style, see the **PR Body** section of the git-workflow rule") keeps one authority instead of four, and the always-on rule guarantees the pointer always resolves.

JetoPistola:
👋 **Review summary**

**What looks good**
- The style guidance is unusually well-written for agent instructions — it states rules as behavior ("**Authoritative.** State what happens: 'The run is scored once.'") with a counter-example for each, which is what actually moves a model off prose-summary habits.
- Offering three shapes instead of one fixed template is the right call. A refactor and a new capability genuinely need different Details sections, and "do not force one" prevents the cargo-culting a rigid template would cause.
- Naming the anti-patterns explicitly — motivation paragraph, "this PR …", benefits list, diff restatement — is more effective than a positive-only description, because those are exactly the failure modes.
- Recovering the `## PR Body` block into `.agents/rules/git-workflow.mdc` fixes a real bug: a hand-edited `.claude/` file was one `make claude` away from silently vanishing. Good catch, and the right layer to fix it at.
- The PR description practices what it documents, which makes the change reviewable by inspection.
- Testing notes are honest about what wasn't run and why (`pre-commit` not installed, no hook matches `.agents/**`) — I verified the hook claim and it holds.

**Overall**

Solid, well-scoped change that puts the guidance at the right layer. One thing needs resolving before merge: the PR template's own `## Details` placeholder still asks for a "2-3 lines single paragraph … motivation (why)", which directly contradicts the new rule — and `/comet:create-pr` calls that template "the source of truth" and reads it at runtime. Worth being precise that this is a *content* conflict, not a structural one: the `##` headings that `pr-lint` greps for are untouched and must stay that way, so fixing this needs no `pr-lint.yml` change — either update the placeholder text, or state that the new style supersedes it while the headings stay fixed.

Separately, the style block now lives in four files verbatim. Since `git-workflow.mdc` is `alwaysApply: true`, the other three copies add no reach — worth pointing them at the rule instead.

**On the existing bot findings**

Baz's comment on `.agents/rules/git-workflow.mdc` (lines 31-37) — *"the universal `N/A` instruction fills `## Issues` for resolved-ticket PRs too, so CI fails"* — **can be skipped**. I checked it against the workflow and it does not hold, on two independent counts:

1. **The CI claim is wrong.** `pr-lint.yml` computes `isNATicket = title.startsWith("[NA]")` and gates the Issues token check on it: `if (issuesContent !== null && !isNATicket && !issuesRegex.test(issuesContent))`. For an `[NA]` PR the check is skipped entirely, so `N/A` under `## Issues` passes — which is exactly what this PR's own body does, and its `pr-lint` run is green.
2. **The "universal" reading is wrong.** The instruction is conditional: "A section that does **not apply** gets `N/A`". On a resolved-ticket PR `## Issues` *does* apply, so the clause never fires — and `create-pr.md` separately instructs listing every resolved ticket there with a hyphenated key.

The only path to Baz's failure is an agent misreading a conditional as unconditional, which is a model-behavior risk rather than a defect in this diff — and not one this PR introduced. No change needed.

Its other two findings are legitimate: the template/style conflict (overlaps my blocker above) and the `*.md` vs `.mdc` checklist classification in `_pr-description-sync.md`.

**Inline comments**: 1 blocker, 1 suggestion

PR #8023, opened 2026-08-26
[NA] [SDK] fix: stop bedrock and mistral stream wrappers from swallowing exceptions in finally

alexkuzmik:
## Details

The bedrock and mistral stream wrappers install class-level patches and used an early `return` inside a `finally` block to skip cleanup for non-tracked streams. A `return` in `finally` discards any in-flight exception, so once a tracked call installed the patch, non-tracked streams that errored mid-iteration completed silently. Fix: invert the guard so cleanup nests under `if hasattr(...)` and no `return` remains in any `finally` — tracked streams behave exactly as before.

- **Bedrock is worse than a lost exception — it silently corrupts data.** The patch target is `botocore.response.StreamingBody.read`, botocore's *shared* class used by every boto3 response body, and the `try` block does `return result`. A `return None` in `finally` **overrides the returned value**, so after one traced `invoke_model` call every non-tracked `read()` in the process — an S3 object, a Lambda payload — handed back `None` instead of its payload:
  ```
  === BEFORE opik patches anything ===
  read() -> b'hello-s3-object'
  === AFTER opik traced ONE bedrock call ===
  tracked read()   -> b'{"completion": "hi"}'
  unrelated read() -> None          <<< payload silently replaced with None
  ```
- **Mistral** patches `__iter__`/`__aiter__` on the stream's class; these are generators with no return value to clobber, so the impact is exception swallowing only.

This is the same bug and the same fix that #7981 applies to the anthropic integration — **credit to @trakshan-mishra for finding the pattern there.** I found these two while reviewing that PR. The two PRs touch disjoint files, so they can merge in either order. After both land, an AST scan over `sdks/python/src/opik` finds no `return`/`break`/`continue` in any `finally` block.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: found both bugs while reviewing #7981, wrote the fix and the regression tests
- Human verification: diff reviewed; tests confirmed to fail on unfixed source and pass with the fix (see Testing)

## Testing

- **Commands run:**
  - `pytest tests/library_integration/mistral/` → **18 passed** (15 existing + 3 new), real Mistral API
  - `pre-commit run --files <4 changed files>` → ruff, ruff-format, mypy, whitespace hooks all **Passed**
- **Regression proof (mistral):** with the fix reverted and the tests unchanged, the two `untracked_stream_fails_after_class_patched` tests **fail** (exception swallowed) and pass again with the fix. The `tracked_stream_fails_mid_iteration` test passes both ways by design — it guards the tracked path against regressions in the refactor.
- **Bedrock data-loss bug** reproduced directly against real `botocore.response.StreamingBody` objects (output above); verified fixed by the same script.
- **Test placement:** both tests extend the existing library-integration suites. The mistral ones make real streaming calls, then inject a deterministic mid-stream failure by swapping `EventStream.generator` (which is what `__next__` pulls from) on a genuine stream object — a real dropped-connection stands in without monkeypatching the patched class.
- **Not verified locally:** `test_bedrock_invoke_model__untracked_client_read_after_tracked_call__payload_returned` needs real AWS Bedrock credentials, which I don't have — the `AWS_PROFILE` here fails with `PartialCredentialsError`. It collects cleanly and is modelled on the existing verified `..._anthropic___happyflow` test in the same file (same client, model, and request body); the mechanism it asserts is the one reproduced above. **Please confirm it goes green in CI.**

## Documentation

No documentation changes needed — this is an internal bug fix with no public API or behaviour change for correctly-tracked streams.

### DIFF
--- a/sdks/python/src/opik/integrations/bedrock/invoke_model/stream_wrappers.py
+++ b/sdks/python/src/opik/integrations/bedrock/invoke_model/stream_wrappers.py
@@ -44,36 +44,34 @@ def wrap_invoke_model_response(
             error_info = error_info_collector.collect(exception)
             raise exception
         finally:
-            if not hasattr(self, "opik_tracked_instance"):
-                return None
+            if hasattr(self, "opik_tracked_instance"):
+                delattr(self, "opik_tracked_instance")
 
-            delattr(self, "opik_tracked_instance")
-
-            if error_info is None and result is not None:
-                try:
-                    parsed_body = json.loads(result)
-                    output = {
-                        "body": parsed_body,
-                        "ResponseMetadata": response_metadata,
-                    }
-                    LOGGER.debug(
-                        "Successfully parsed response body with keys: %s",
-                        list(parsed_body.keys()),
-                    )
-                except (json.JSONDecodeError, TypeError) as e:
-                    LOGGER.debug("Failed to parse response body as JSON: %s", e)
+                if error_info is None and result is not None:
+                    try:
+                        parsed_body = json.loads(result)
+                        output = {
+                            "body": parsed_body,
+                            "ResponseMetadata": response_metadata,
+                        }
+                        LOGGER.debug(
+                            "Successfully parsed response body with keys: %s",
+                            list(parsed_body.keys()),
+                        )
+                    except (json.JSONDecodeError, TypeError) as e:
+                        LOGGER.debug("Failed to parse response body as JSON: %s", e)
+                        output = {"body": {}, "ResponseMetadata": response_metadata}
+                else:
+                    LOGGER.debug("Error occurred or result is None, using empty body")
                     output = {"body": {}, "ResponseMetadata": response_metadata}
-            else:
-                LOGGER.debug("Error occurred or result is None, using empty body")
-                output = {"body": {}, "ResponseMetadata": response_metadata}
 
-            finally_callback(
-                output=output,
-                error_info=error_info,
-                generators_span_to_end=span_to_end,
-                generators_trace_to_end=trace_to_end,
-                capture_output=True,
-            )
+                finally_callback(
+                    output=output,
+                    error_info=error_info,
+                    generators_span_to_end=span_to_end,
+                    generators_trace_to_end=trace_to_end,
+                    capture_output=True,
+                )
 
     botocore.response.StreamingBody.read = wrapped_read
     streaming_body.opik_tracked_instance = True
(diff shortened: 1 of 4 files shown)


CometActions:

You shipped the regression tests with the fix, and they are the right ones: the bedrock test asserts an untracked `StreamingBody.read()` returns its payload (the old `return None` in `finally` overrode the `try`'s return), and the two mistral tests assert an untracked stream that fails mid-iteration still raises (the old `return` discarded the in-flight exception). All three fail against the pre-fix code, and both suites run in CI with real keys via lib-bedrock-tests.yml / lib-mistral-tests.yml. Nothing for the Playwright estate to add on top — the blast radius is in the user's own process, and the tracked path is byte-identical before and after, so no trace or span renders differently in Opik. One half you may want to cover while you are here: on bedrock, the old `return None` also swallowed exceptions raised by an untracked `read()`, not just the return value — your mistral tests cover the swallowed-exception half, the bedrock one only covers the payload half.

_also touches Python SDK_

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/32966183835)

trakshan-mishra:
Confirmed green. The bedrock test you couldn't run locally passed on all five jobs (3.10–3.14):

```
test_invoke_model.py::test_bedrock_invoke_model__untracked_client_read_after_tracked_call__payload_returned PASSED [100%]
============================= 25 passed in 32.05s ==============================
```

Mistral job passed too.

On the AST scan at the end of your description: ruff already ships this as `B012`, but `sdks/python/.ruff.toml` selects only `["E4", "E7", "E9", "F"]`, so bugbear never runs. I pointed it at my branch (anthropic already fixed there) and it finds exactly the three sites this PR fixes, nothing else:

```
src/opik/integrations/bedrock/invoke_model/stream_wrappers.py:48:17: B012
src/opik/integrations/mistral/stream_patchers.py:55:17: B012
src/opik/integrations/mistral/stream_patchers.py:111:17: B012
```

So `B012` would go clean the moment this merges, and the next one gets caught in pre-commit rather than in someone's S3 payload. Happy to open a small separate PR adding it if you think it's worth having — didn't want to bolt it onto either of these.

Two failures here that aren't your code, in case they're useful: PR Linter wants a `## Documentation` section in the description, and `E2E Lib Integration Python 3.11` died at "Run latest Opik server" with `opik-clickhouse-init-1` exiting, which looks like infra rather than anything you touched.

petrotiurin (approved):
Looks good!

PR #8027, opened 2026-08-26
[OPIK-6901] [BE] docs(cutover): an early cutover_start destroys data, and the ex-wrapper needs INSERT

andrescrz:
## Details

Two defects in the cutover runbook's rollback guidance, found by reading the drivers and the deployed grants against what the rollback statements actually require. Docs only — no driver, SQL or test changes.

**1. `cutover_start` is not a value to estimate, and the runbook never said so.**

`exchange_and_wrap.sh` captures it **before** the final deletion replay and **before** the `EXCHANGE` (lines 334 / 338 / 340), deliberately, so the window covers every delete bridged from that instant onward. Supplying a different value loses data in one of two directions:

- **Too early** — the reverse replay is guard-less, so it masks deletes already applied to the original while it was live, including ids deleted and then re-created before the cutover. Those are legitimately live in the parked original.
- **Too late** — anchoring to the `EXCHANGE` rather than the capture misses every delete bridged between the two, so those rows come back live on the restored original. The gap is the whole run time of the final deletion replay.

The section now states it as an artifact **of** the forward run, and says to stop and escalate if it was lost: there is no in-procedure way to recover it, since statements come only from the versioned `.sql` files the drivers read and none reads this value back. That gap is named rather than papered over — closing it means the driver persisting the value, or a versioned script reading it back, which is follow-up.

**2. The two rollback privilege rows understate `traces_dist_old` and `traces_post_rollback_backup`.**

Both are `RENAME` **destinations** in stage C, and `traces_dist_old` is one in `--unwrap-only` too. `RENAME` needs `INSERT` + `CREATE TABLE` on its destination, but the rows name only the `CREATE`/`DROP` half — so a grant set built from them aborts both rollback paths with `Code: 497 … Missing permissions: INSERT`.

The four-privileges note directly below those rows already says to expand the shorthand, and predicted this exact failure. Making the rows explicit anyway, since these are the names most likely to be missed: they do not exist yet *and* they are only ever destinations. The rename is the first statement either path issues, so nothing has mutated when it fires — but it fires while an operator is rolling back.

Deployed-grant fix: **comet-ml/comet-helm#1337**.

## Change checklist
- [ ] User facing
- [x] Documentation update

## Issues

- Resolves #
- OPIK-6901

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: Full authorship of both runbook changes and the analysis behind them.
- Human verification: the ordering was read out of `exchange_and_wrap.sh` rather than inferred, and the privilege requirement reproduced on a throwaway server before being written down. An earlier revision of this PR got the boundary backwards and offered an ad-hoc recovery query; both were corrected in review.

## Testing

Documentation only. Neither claim is asserted from the docs:

- **Ordering** — read from `exchange_and_wrap.sh`: capture at line 334, final deletion replay at 338, `EXCHANGE` at 340.
- **Privileges** — reproduced on a throwaway ClickHouse 26.3.16:

| grants on the rename destination | result |
|---|---|
| `CREATE TABLE`, `DROP TABLE` | `Code: 497 … Missing permissions: INSERT` |
| + `INSERT` | succeeds |

No figures, hostnames, timestamps or identifiers from any environment appear in the change.

## Documentation

The change *is* documentation: `data-migrations/traces-local-v2-cutover/README.md`, two edits.

- **Rollback stage selection** — the `cutover_start` rule, its two failure directions, and stop-and-escalate when the value is lost. Placed where the operator picks the stage, which is where the value is first needed.
- **Required privileges table** — `INSERT` named on the two rollback rename destinations, cross-referencing the existing four-privileges note rather than restating it.

No user-facing or public docs (`apps/opik-documentation/`) are affected: this runbook is operator tooling for the traces cutover.


CometActions:
**No test needed here.**

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/33048763791)

_Re-checked after a push on 27 Aug 07:13 UTC._

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
```diff
@@ -970,7 +970,23 @@ and leaves the untouched live `traces` — there is no backup to soak or finaliz
...
-Pick the stage by how far the cutover got (`cutover_start` is the value `exchange_and_wrap.sh` printed):
+**`cutover_start` must be the actual `EXCHANGE` moment — earlier is not safer, it is destructive.** The reverse replay
+is deliberately guard-less: it masks *every* id the bridge recorded in the window, unconditionally. Deletes that landed
+**before** the cutover were already applied to the original while it was live, so they need no replay — and any id that
+was deleted and then re-created before the cutover (a resurrection) is *legitimately live* in the parked original. Widen
+the window past the `EXCHANGE` and the replay masks those rows too, destroying data the rollback was supposed to
+preserve. This is not hypothetical: a production-shaped environment was found holding pre-cutover bridge events whose
+ids were live in the parked original, which an over-early `cutover_start` would have deleted.
+
+Use the value `exchange_and_wrap.sh` printed. If it was lost, recover it rather than estimating — the `EXCHANGE` is
```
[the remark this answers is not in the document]
andrescrz: **You are right, and the section as written would have caused the failure it was trying to prevent.** Verified in the driver rather than argued: `exchange_and_wrap.sh` captures the value at line 334, then runs the final deletion replay at 338, then the `EXCHANGE` at 340. So `cutover_start` is deliberately *earlier* than the swap, and "must be the actual `EXCHANGE` moment" was wrong.

Worse, the recovery query I offered made it concrete. Measured on a production-shaped environment:

```
capture (log_comment 'traces_local_v2_cutover:exchange_and_wrap')  13:32:25.185812
EXCHANGE (system.distributed_ddl_queue)                            13:37:19
```

Nearly five minutes — the final replay's run time on a large table. Every delete bridged in that gap would have been outside the window and silently resurrected on the restored original. That is the opposite of the hazard I documented, and my guidance pointed straight at it.

Rewritten to state both directions, because neither is safe:

- **earlier than the recorded value destroys data** — the guard-less replay masks pre-cutover deletes, including ids deleted-and-recreated before the cutover that are legitimately live in the parked original;
- **later resurrects data** — it misses everything bridged between the capture and the swap.

So it is not a value to estimate at all. If the printed value is lost it is now recovered from the driver's *own* capture query by its `log_comment`, which is the statement that produced it rather than a reconstruction, and read through `clusterAllReplicas` since the driver's session ran on one replica.

On your `query_create_time` point specifically: agreed, and it is now moot — that source is gone rather than tightened, because whole-second truncation was the smaller problem next to it being systematically late.

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
```diff
@@ -970,7 +970,23 @@ and leaves the untouched live `traces` — there is no backup to soak or finaliz
...
+recorded on the cluster:
+
+```sql
+SELECT query_create_time, query FROM system.distributed_ddl_queue
+WHERE query ILIKE '%EXCHANGE TABLES%traces%' ORDER BY query_create_time DESC LIMIT 1;
```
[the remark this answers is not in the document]
andrescrz: **Agreed, and resolved by deletion rather than by constraining it.** The query was under-specified exactly as you describe — `EXCHANGE TABLES%traces%` would match another database, cluster, or a similarly named table, and feed that timestamp to a `rollback.sh --database opik` whose replay is unconditional.

But tightening it to the exact `${DATABASE}.traces AND ${DATABASE}.traces_local_v2 ON CLUSTER '{cluster}'` statement would have preserved a deeper defect your sibling comment identified: the `EXCHANGE` is the **wrong boundary** in the first place, arriving minutes after the value the driver actually recorded. A perfectly-constrained query returning the wrong instant is worse than an unconstrained one, because it looks authoritative.

So the DDL-queue path is removed. The replacement reads the driver's own capture query, pinned to `log_comment = 'traces_local_v2_cutover:exchange_and_wrap'` and the `now64(6)` statement — and it carries your uniqueness requirement explicitly: confirm the row is unique for the run you mean, because more than one match means more than one cutover and the newest is not necessarily yours.

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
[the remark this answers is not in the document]
andrescrz: **Correction to my own reply above: the fix I described was itself wrong, and is now removed.**

Replacing the DDL-queue query with a `system.query_log` query kept the actual defect — the procedure runs statements **only** from the versioned `.sql` files the drivers read, and I put a hand-run query into the runbook that states that rule. Swapping one ad-hoc query for a better-targeted ad-hoc query was not a fix.

`cutover_start` is now stated as what it is: a **required artifact of the forward run**. `exchange_and_wrap.sh` prints `RECORD cutover_start=…` for that purpose, and the instruction is to capture it then.

If it is lost, the guidance is to **stop and escalate**, not to derive it. There is no versioned path that recovers the value, so there is no in-procedure way to do it — and the nearest-looking source remains provably wrong, which is the part of the earlier analysis that stands: the `EXCHANGE` timestamp trails the capture by the whole length of the final deletion replay (measured at ~5 minutes on a production-shaped table), so anchoring a rollback to it resurrects every delete bridged in between.

The two-direction hazard this section exists for is unchanged. Only the invented recovery path is gone.

That also leaves a real gap worth naming rather than papering over: the driver only *echoes* the value, so the sole artifact is the operator's run record. Making it recoverable would mean the driver persisting it, or a versioned read-only script that reads it back — versioned tooling, not prose. Raising it as follow-up rather than fixing it here.

thiagohora (dismissed):

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
```diff
@@ -855,8 +855,8 @@ read-only drivers cannot surface a mutation-privilege gap by construction.
 | post-swap `RENAME` | `RENAME TABLE <shadow> TO <backup>` | `CREATE TABLE` + `DROP TABLE` (grant `INSERT` on the backup name too, so the rename cannot trip the same check) |
 | **wrap** (sharding) | `CREATE TABLE traces_dist … ENGINE = Distributed(…)`, then `RENAME traces → traces_local, traces_dist → traces` | `CREATE TABLE` + `DROP TABLE` on **`traces_dist`** and **`traces_local`** — two names that **do not exist yet**, so a grant set scoped to the cutover's three names will NOT cover the wrap. Plus `SELECT` on `traces_local` (post-wrap reads route through the wrapper to it) and `REMOTE` for the `Distributed` engine. |
 | rollback stage A/B (if in scope) | stage A `TRUNCATE`, reverse replay | `TRUNCATE` on the shadow, and `ALTER UPDATE(_row_exists)` on the **source** (the reverse replay masks rows on the restored original). Withhold unless a rollback is actually planned. |
```
JetoPistola: 🚫 **blocker** | Correctness

**Stage B has the same missing-`INSERT` gap this PR fixes for stage C, and it is the likelier rollback path.**

Stage B issues a 2-way `RENAME` (`000004_rollback_stage_b_exchange_back.sql`):

```sql
RENAME TABLE
    ...traces TO ...traces_post_rollback_backup,
    ...traces_pre_cutover_backup TO ...traces
```

`traces_post_rollback_backup` satisfies *both* conditions the PR body gives for making a name explicit rather than leaning on the four-privileges note — it does not exist yet, **and** it is only ever a destination. Yet this row names no `RENAME` privileges at all: only `TRUNCATE` and `ALTER UPDATE(_row_exists)`. A grant set built from it aborts stage B at its **first** statement with the identical `Code: 497 … Missing permissions: INSERT` the two rows below now call out.

This is worse than the stage C case it mirrors. The wrap is **deferred by default** — `exchange_and_wrap.sh` prints `Distributed wrap deferred (default)` and tells the operator to apply it later with `--wrap-only`. So the window between `EXCHANGE` and wrap is the normal resting state, and a rollback taken there is stage B. The stage C row is guarded by "decide this before applying the wrap"; stage B has no such decision point, because reaching it requires no extra step.

The row also understates the *source* side: `traces_pre_cutover_backup` is a `RENAME` source in both stage B and stage C, needing `SELECT` + `DROP TABLE`.

Suggest extending this row the same way the two below were, e.g.:

```suggestion
| rollback stage A/B (if in scope) | stage A `TRUNCATE`; stage B 2-way `RENAME`, reverse replay | `TRUNCATE` on the shadow, and `ALTER UPDATE(_row_exists)` on the **source** (the reverse replay masks rows on the restored original). Stage B additionally renames: `CREATE TABLE` + `DROP TABLE` + **`INSERT`** on **`traces_post_rollback_backup`** — a `RENAME` destination that **does not exist yet**, so the `CREATE`/`DROP` half alone fails `Code: 497` at the rename (see the four-privileges note below) — plus `SELECT` + `DROP TABLE` on **`traces_pre_cutover_backup`** (the rename source). Withhold unless a rollback is actually planned. |
```

Both of the PR's own claims check out against the source, for what it is worth: the capture ordering is `exchange_and_wrap.sh` 334 → 338 → 340 as described, and `traces_dist_old` / `traces_post_rollback_backup` are `RENAME` destinations in stage C and un-wrap.

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
```diff
@@ -855,8 +855,8 @@ read-only drivers cannot surface a mutation-privilege gap by construction.
...
+| rollback stage C (if the wrap is applied) | 3-way `RENAME` + `DROP` of the ex-wrapper | `CREATE TABLE` + `DROP TABLE` + **`INSERT`** on **`traces_dist_old`** and **`traces_post_rollback_backup`** — both are `RENAME` destinations, so the `CREATE`/`DROP` half alone fails `Code: 497` at the rename (see the four-privileges note below). `DROP TABLE` on `traces_local`, plus `ALTER UPDATE(_row_exists)` on the restored `traces`. **Decide this before applying the wrap:** without these grants there is no way back to the pre-cutover table until another grant change lands. (The *wrap itself* stays reversible via the un-wrap row below, which needs no extra grants — but that returns to the successor, not to the original.) |
```
JetoPistola: ❓ **question** | Consistency

The two edited rows state the destination privileges in different orders and shapes — stage C as `CREATE TABLE + DROP TABLE + INSERT`, un-wrap as `CREATE TABLE`/`DROP TABLE` … `plus INSERT on traces_dist_old` — while the four-privileges note below states the destination pair as `INSERT` + `CREATE TABLE`.

No factual disagreement; all three name the same set. But since the stated purpose of making these explicit is that operators provision *from these rows*, is it worth settling on one ordering so a reader diffing the rows against the note does not have to re-derive that they match?

JetoPistola:
👋 **Review summary**

Cutover runbook: `cutover_start` must be the value the driver recorded (both directions of error lose data), and the stage C / `--unwrap-only` privilege rows now name `INSERT` on `traces_dist_old` and `traces_post_rollback_backup` as `RENAME` destinations.

**What looks good**

- **Both claims verified against the source rather than asserted from the docs.** The capture ordering is exactly as described — `exchange_and_wrap.sh` line 334 captures `cutover_start`, 338 runs the final deletion replay, 340 issues the `EXCHANGE` — and the driver's own inline comment at 302-307 independently states the same rationale. `traces_dist_old` and `traces_post_rollback_backup` are `RENAME` destinations in `000004_rollback_stage_c_promote_original.sql` and `000004_rollback_unwrap.sql`.
- **Both failure directions are named, not just the obvious one.** "Too early" masking deleted-then-recreated ids and "too late" missing the replay-window bridges are genuinely different bugs with the same symptom class; documenting the pair is what makes the rule followable rather than a bare instruction.
- **Stop-and-escalate instead of an invented recovery.** The earlier revision's ad-hoc query was the more tempting thing to ship. Naming the gap — statements come only from the versioned `.sql` files, none reads the value back — and pointing at the two real closures (driver persists it, or a versioned script reads it back) leaves the follow-up actionable rather than silently owed.
- **Placement.** The `cutover_start` rule sits where the operator picks the stage, which is where the value is first needed, rather than in a separate reference section.

**Overall**

The approach is right and the scope is appropriately narrow — docs-only, no driver or SQL changes, with the deployed-grant half split into a separate helm PR. The `cutover_start` section I'd take as-is.

One gap in the privilege half, left inline: the PR's stated criterion for making a name explicit — does not exist yet **and** only ever a destination — also holds for `traces_post_rollback_backup` in **stage B**, whose 2-way `RENAME` has it as a destination. The stage A/B row still names no `RENAME` privileges at all, so a grant set built from it hits the same `Code: 497` this PR is fixing one row below. That path matters more than stage C's, since the wrap is deferred by default and the post-`EXCHANGE`/pre-wrap state is the normal resting place a rollback would be taken from.

**Inline comments**: 1 blocker, 1 question.

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
andrescrz: **Blocker accepted — verified against the source, and it is a gap in this PR's own reasoning.** Every claim checks out:

| claim | verified |
|---|---|
| stage B renames `traces` → `traces_post_rollback_backup` (destination, does not exist yet) | ✅ `000004_rollback_stage_b_exchange_back.sql` |
| `traces_pre_cutover_backup` is the rename *source* | ✅ same file |
| the row named **no** rename privileges — only `TRUNCATE` and `ALTER UPDATE(_row_exists)` | ✅ |
| the wrap is deferred by default | ✅ `exchange_and_wrap.sh:347` prints `Distributed wrap deferred (default)` |

The part that stings is the one you put your finger on: this PR *stated* the criteria for making a name explicit — does not exist yet, **and** only ever a destination — applied them to two rows, and left alone the row where they hold most clearly. And your point about likelihood is the right frame: stage C sits behind "decide this before applying the wrap", whereas stage B needs no extra step to become reachable, so the row with no decision point was the one with no privileges named.

Row extended along the lines you suggested, with one addition: the *why it is the likelier path* is now in the row itself, since that is what makes withholding these grants a real risk rather than a theoretical one.

**One useful difference from the `traces_dist_old` case, which I measured rather than assumed:** a rollback-provisioned account already carries `SELECT`/`INSERT`/`CREATE TABLE`/`DROP TABLE` on **both** `traces_post_rollback_backup` and `traces_pre_cutover_backup`. So stage B is genuinely runnable today and this is a **documentation defect only** — no companion grant PR, unlike comet-helm#1337 which fixed a real deployed gap on the ex-wrapper. Worth stating precisely so nobody reads this as a second live blocker.

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
andrescrz: **Worth it, and done.** Since the rows exist precisely so operators provision *from* them, three different shapes for one privilege pair is friction at exactly the wrong moment.

Settled on the four-privileges note's own form — **`INSERT` + `CREATE TABLE`** — across all three rows, so the rows and the note now read identically and a reader diffing them has nothing to re-derive. `DROP TABLE` is stated separately where the name is later dropped, with the reason, rather than being folded into the destination pair where it does not belong.

Five occurrences of the pair in the file, all now in that order.

thiagohora (approved):

PR #8033, opened 2026-08-27
[OPIK-8109] [BE] fix: bound dataset items page read by page size

thiagohora:
## Details

The dataset items page returned 500 for versions with large payloads. `SELECT_DATASET_ITEM_VERSIONS` ordered by `dataset_item_id`, which is not in the table's ordering key `(workspace_id, dataset_id, dataset_version_id, id)`, so `optimize_read_in_order` could not apply and ClickHouse sorted the whole version through a blocking `MergeSortingTransform` with the wide `data` column attached — peak memory scaled with the version's total payload rather than the page size, so requesting one row cost the same as requesting all of them. This resolves the page in two phases: phase 1 picks the page's `dataset_item_id`s from narrow columns only, phase 2 fetches the payload for just those ids.

- Truncation moves to phase 2 so the image regex runs over the page instead of the whole version. `replaceRegexpAll` expands every value before `substring` trims it, which is why `truncate=true` was the worst case rather than a mitigation — it failed **12/12** times at a p50 of 29.85 GiB, and now completes in 25 ms.
- Phase 2's `dataset_item_id IN (...)` is served by the `dataset_item_id` skip indexes added in migration `000074`, which an `ORDER BY` could never use.
- No schema change, no migration, and no change to the template variables, so the calling code and its bindings are untouched. Repeating `<dataset_item_filters>` in phase 2 follows the existing pattern in `SELECT_DATASET_ITEM_VERSIONS_WITH_EXPERIMENT_ITEMS`, which already renders it four times against a single bind.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- OPIK-8109

Related, not resolved here: escalation CUST_6941; the ordering originates in OPIK_4518 (#5370), which replaced the sort-key-aligned `ORDER BY` with `ORDER BY dataset_item_id` and deferred the projection intended to compensate — a projection that cannot ship, since ClickHouse rejects `ADD PROJECTION` on `ReplacingMergeTree`. The experiment-comparison read path has the same unbounded shape and is tracked separately.

## AI-WATERMARK

AI-WATERMARK: yes

- If yes:
  - Tools: Claude Code CLI
  - Model(s): Claude Opus 5
  - Scope: Root-cause investigation, SQL design, implementation, tests
  - Human verification: Query plans and cost compared on ClickHouse 26.3.16; page contents verified row-identical against the previous query on the affected production dataset across seven offsets, comparing every column including payloads; each invariant's test confirmed to fail when its invariant is removed.

## Testing

`mvn test -Dtest='DatasetVersionResourceTest,DatasetsResourceTest,DatasetItemVersionQueryShapeTest'` — **569 tests, 0 failures** (433 + 130 + 6). Local, testcontainers, macOS. `DatasetsResourceTest` covers `GET /datasets/{id}/items` including the filter-on-`data`, search and `StreamDatasetItems` paths, so the duplicated filter SQL and the restructured `lastRetrievedId` cursor branch are both exercised by pre-existing tests.

**New tests.** `DatasetItemVersionQueryShapeTest` (6 tests, 9 ms) asserts the query's shape by reflection: no bare `data` in phase 1, `:limit` only in phase 1, `<dataset_item_filters>` exactly once per phase, both phases leading with `dataset_item_id DESC`, truncation confined to phase 2. These are structural on purpose — the memory behaviour only manifests at dataset sizes an integration test cannot build, so the cheap deterministic guard is the shape. `DatasetVersionResourceTest$BoundedPageReads` (3 tests) covers the behaviour: a filtered page when a newer row supersedes the matching one, page ordering and disjointness across a version, and `truncate=true`.

**Mutation-checked.** Removing the phase-2 filter and nothing else fails exactly two tests — `filteredPage__whenSupersedingRowFailsTheFilter__thenItIsNotReturned` and `invariantThree__phaseTwoRepeatsTheFilters` — while the other 7 stay green. Under the mutation the filtered page returns rows tagged `superseded` against a filter for `keep`, i.e. rows the caller filtered out. Restored, all 9 pass.

**Measured on the affected production dataset** (119,903 items, ~138 KiB each, 15.8 GiB), page of 10:

| shape | read_bytes | peak memory | outcome |
|---|---|---|---|
| before | 15.83 GiB | 29.97 GiB | 500 |
| after | 19.70 MiB | 14.01 MiB | ok |
| after, `truncate=true` | 19.70 MiB | 18.15 MiB | ok |
| after, sort on a `data` key | 15.84 GiB | 810.13 MiB | ok |
| after, filter on `data` | 15.79 GiB | 2.02 GiB | ok |

Worth a reviewer's attention: filters and sorts derived from `data` still stream the payload to evaluate the predicate, so their peak stays payload-proportional at roughly 5–13%. Comfortable against a 30 GiB cap, but not unbounded — a dataset ~15× larger with a payload filter would reach it again. The default list shape (no sort, no filter) is payload-independent.

**Not run:** the frontend suite (no FE changes), and no video — the change is a single query rewrite with no visible UI change beyond the page loading at all.

## Documentation

None. Internal query shape only; no API or schema change. The invariants that keep the two phases correct are documented in javadoc on the query constant, where the next person to edit it will see them.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
@@ -413,7 +413,51 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
             AND dataset_item_id IN :itemIds
             """;
 
+    /**
+     * Reads one page of a dataset version.
+     * <p>
+     * Resolved in two phases so the wide {@code data} column never enters a sort. The table's ordering key is
+     * {@code (workspace_id, dataset_id, dataset_version_id, id)}, so ordering by {@code dataset_item_id} cannot
+     * use {@code optimize_read_in_order} and forces a full blocking sort over the whole version. Carrying
+     * {@code data} through that sort makes peak memory scale with the version's total payload rather than with
+     * the page size: a 120k-item version at ~138 KiB per item peaked at 30 GiB and returned 500 (OPIK-8109).
+     * <p>
+     * Phase 1 resolves the page's {@code dataset_item_id}s from narrow columns only; phase 2 fetches the payload
+     * for just those ids, served by the {@code dataset_item_id} skip indexes from migration {@code 000074}.
+     * Truncation stays in phase 2 so the image regex runs over the page instead of the whole version.
+     * <p>
+     * Three invariants hold this together. Each one broke a draft of this query, and each has a regression test
+     * in {@code DatasetItemVersionPaginationTest}:
+     * <ol>
+     *   <li>Both phases order by {@code dataset_item_id DESC} first. A mismatch reorders the page.</li>
+     *   <li>The ordering's leading key, {@code dataset_item_id}, is unique per output row after
+     *       {@code LIMIT 1 BY dataset_item_id}, so the page order is fully determined and the trailing
+     *       timestamp only picks a winner among duplicate rows for the same id. If custom sorting is ever
+     *       added here, the sort must still <em>end</em> in {@code dataset_item_id}: with a non-unique sort
+     *       key and no unique tiebreak, the two phases sort different column sets and can order tied rows
+     *       differently.</li>
+     *   <li>Phase 2 repeats {@code dataset_item_filters}. Without it, an id selected in phase 1 via a superseded
+     *       row that matches the filter resolves in phase 2 to the newest row for that id, which may not match,
+     *       returning rows the caller filtered out.</li>
+     * </ol>
+     */
     private static final String SELECT_DATASET_ITEM_VERSIONS = """
+            WITH page AS (
+                SELECT dataset_item_id
+                FROM dataset_item_versions
+                WHERE dataset_id = :datasetId
+                AND dataset_version_id = :versionId
+                AND workspace_id = :workspace_id
+                <if(lastRetrievedId)>AND dataset_item_id \\< :lastRetrievedId<endif>
+                <if(dataset_item_filters)>AND (<dataset_item_filters>)<endif>
+                ORDER BY dataset_item_id DESC, item_last_updated_at DESC
+                LIMIT 1 BY dataset_item_id
+                <if(lastRetrievedId)>
+                LIMIT :limit
+                <else>
+                LIMIT :limit OFFSET :offset
+                <endif>
+            )
             SELECT
                 dataset_item_id AS id,
                 dataset_id,
@@ -434,15 +478,10 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
             WHERE dataset_id = :datasetId
             AND dataset_version_id = :versionId
             AND workspace_id = :workspace_id
-            <if(lastRetrievedId)>AND dataset_item_id \\< :lastRetrievedId<endif>
             <if(dataset_item_filters)>AND (<dataset_item_filters>)<endif>
+            AND dataset_item_id IN (SELECT dataset_item_id FROM page)
             ORDER BY dataset_item_id DESC, last_updated_at DESC
             LIMIT 1 BY dataset_item_id
-            <if(lastRetrievedId)>
-            LIMIT :limit
-            <else>
-            LIMIT :limit OFFSET :offset
-            <endif>
             """;
 
     private static final String SELECT_DATASET_ITEM_VERSIONS_COUNT = """
--- a/apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
@@ -55,11 +55,14 @@ import com.comet.opik.domain.experiments.aggregations.ExperimentAggregatesServic
 import com.comet.opik.extensions.DropwizardAppExtensionProvider;
 import com.comet.opik.extensions.RegisterApp;
 import com.comet.opik.infrastructure.auth.RequestContext;
+import com.comet.opik.infrastructure.db.TransactionTemplateAsync;
 import com.comet.opik.podam.PodamFactoryUtils;
 import com.comet.opik.utils.JsonUtils;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.google.inject.Injector;
 import com.redis.testcontainers.RedisContainer;
+import io.r2dbc.spi.Result;
+import io.r2dbc.spi.Statement;
 import org.apache.hc.core5.http.HttpStatus;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
@@ -75,12 +78,14 @@ import org.testcontainers.clickhouse.ClickHouseContainer;
 import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.lifecycle.Startables;
 import org.testcontainers.mysql.MySQLContainer;
+import reactor.core.publisher.Mono;
 import ru.vyarus.dropwizard.guice.test.ClientSupport;
 import ru.vyarus.dropwizard.guice.test.jupiter.ext.TestDropwizardAppExtension;
 import ru.vyarus.guicey.jdbi3.tx.TransactionTemplate;
 import uk.co.jemos.podam.api.PodamFactory;
 
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -156,6 +161,7 @@ class DatasetVersionResourceTest {
     private SpanResourceClient spanResourceClient;
     private TransactionTemplate mySqlTemplate;
     private ExperimentAggregatesService experimentAggregatesService;
+    private TransactionTemplateAsync clickHouseTemplate;
 
     @BeforeAll
     void setUpAll(ClientSupport client, TransactionTemplate mySqlTemplate, Injector injector) {
@@ -171,6 +177,7 @@ class DatasetVersionResourceTest {
         traceResourceClient = new TraceResourceClient(client, baseURI);
         spanResourceClient = new SpanResourceClient(client, baseURI);
         experimentAggregatesService = injector.getInstance(ExperimentAggregatesService.class);
+        clickHouseTemplate = injector.getInstance(TransactionTemplateAsync.class);
     }
 
     @AfterAll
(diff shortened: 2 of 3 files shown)


CometActions:

The correctness surface of the two-phase rewrite is covered by the tests in this PR. `DatasetVersionResourceTest.BoundedPageReads` drives GET /v1/private/datasets/{id}/items — the same endpoint the dataset items page calls — and asserts id-desc order plus disjointness across 3 pages of 25, the superseded-row filter case (invariant 3), and truncate=true; `DatasetItemVersionQueryShapeTest` pins the two-phase shape and all three invariants. The cursor branch is covered too: `DatasetsResourceTest.streamDataItems__whenStreamingDatasetItemsMoreThanLimit...` pages 3000 items via lastRetrievedId, and config-test.yml sets datasetVersioningEnabled: true, so it runs through this query. Nothing for us to add — our e2e datasets specs (dataset-items.spec.ts, dataset-items-filter-scope.spec.ts) all use 3-item datasets and never turn a page, so they'd have missed a reordered or over-returning page entirely. One gap noted below, not blocking.

**Not testable yet.** The regression this actually fixes — a ~120k-item version at ~138 KiB per item peaking at 30 GiB and returning 500 — is only asserted structurally, by DatasetItemVersionQueryShapeTest reading the SQL text. No test observes the memory bound behaviourally, and the e2e estate cannot seed a version at that scale, as your own javadoc on the shape test says. That axis is tracked as datasets.load.dataset-item-volume (covered: false) with existing scenarios in tests_load, so it belongs there rather than in e2e.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/33059825416)

BorisTkachenko (approved):
Looks good

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -413,7 +413,51 @@ SELECT count(DISTINCT dataset_item_id) AS count
...
+     */
     private static final String SELECT_DATASET_ITEM_VERSIONS = """
+            WITH page AS (
+                SELECT dataset_item_id
+                FROM dataset_item_versions
+                WHERE dataset_id = :datasetId
+                AND dataset_version_id = :versionId
+                AND workspace_id = :workspace_id
+                <if(lastRetrievedId)>AND dataset_item_id \\< :lastRetrievedId<endif>
+                <if(dataset_item_filters)>AND (<dataset_item_filters>)<endif>
```
JetoPistola, line 452: 🚫 **blocker** | Correctness

**Filters on `id`, `created_at`, `last_updated_at`, `created_by`, `last_updated_by` silently resolve to the wrong columns in phase 1.**

On `main` the filter predicate ran inside a `SELECT` whose projection aliased these columns:

```sql
dataset_item_id AS id,
item_created_at as created_at,
item_last_updated_at as last_updated_at,
item_created_by as created_by,
item_last_updated_by as last_updated_by,
```

ClickHouse resolves `WHERE` against those projection aliases, so a filter emitting `id = ...` matched `dataset_item_id`, and `created_at > ...` matched `item_created_at`.

The new phase-1 CTE selects only `dataset_item_id` with **no aliases**, but `<dataset_item_filters>` is rendered into it verbatim. `FilterQueryBuilder` maps `ID → "id"`, `CREATED_AT → "created_at"`, `LAST_UPDATED_AT → "last_updated_at"`, `CREATED_BY → "created_by"`, `LAST_UPDATED_BY → "last_updated_by"` — and migration `000047` shows **both column sets physically exist** on this table:

| filter field | phase 1 now resolves to | should be | phase 2 |
|---|---|---|---|
| `id` | `id` — the **physical row UUID** | `dataset_item_id` | ✅ correct |
| `created_at` | `created_at` — version-row write time | `item_created_at` | ✅ correct |
| `last_updated_at` | `last_updated_at` | `item_last_updated_at` | ✅ correct |
| `created_by` / `last_updated_by` | row-level | `item_*` | ✅ correct |

Because both names exist, this doesn't error — it silently narrows to the wrong rows. **`id` is the worst case**: a caller filtering `id = <dataset item id>` matches zero rows in phase 1, so the page comes back empty. All five fields are publicly filterable via `DatasetItemField`.

Aliasing in phase 1 fixes it while keeping every column narrow, so the memory win is preserved:

```suggestion
                SELECT dataset_item_id,
                    dataset_item_id AS id,
                    item_created_at AS created_at,
                    item_last_updated_at AS last_updated_at,
                    item_created_by AS created_by,
                    item_last_updated_by AS last_updated_by
                FROM dataset_item_versions
                WHERE dataset_id = :datasetId
                AND dataset_version_id = :versionId
                AND workspace_id = :workspace_id
                <if(lastRetrievedId)>AND dataset_item_id \\< :lastRetrievedId<endif>
                <if(dataset_item_filters)>AND (<dataset_item_filters>)<endif>
```

**Why the 569 green tests missed it:** the new `BoundedPageReads` tests only exercise `DatasetItemField.DATA`, and there is no test anywhere using `DatasetItemField.ID`, `CREATED_AT`, or `LAST_UPDATED_AT` against this endpoint. The mutation-check in the PR description mutated the *presence* of the phase-2 filter, not the *resolution* of column names — a mutation that could not surface this. A test filtering on `id` would be the cheapest guard.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -413,7 +413,51 @@ SELECT count(DISTINCT dataset_item_id) AS count
...
+                ORDER BY dataset_item_id DESC, item_last_updated_at DESC
```
JetoPistola, line 453: 🚫 **blocker** | Correctness

**The two phases dedupe on different tiebreaker columns.**

```
phase 1:  ORDER BY dataset_item_id DESC, item_last_updated_at DESC   ← LIMIT 1 BY
phase 2:  ORDER BY dataset_item_id DESC, last_updated_at DESC        ← LIMIT 1 BY
```

When an item has multiple physical rows — exactly the superseding-row state `addSupersedingRow` constructs — phase 1 picks the winner by `item_last_updated_at` and phase 2 by `last_updated_at`. Per migration `000047` these are independent columns (`last_updated_at` even carries `DEFAULT now64(9)` and is the `ReplacingMergeTree` version column), so the two phases can select **different physical rows** for the same `dataset_item_id`.

This also makes invariant 1 in the Javadoc — "Both phases order by `dataset_item_id DESC` first" — and its guard `invariantOne__bothPhasesShareTheLeadingSortKey` weaker than they read: both assert only the *leading* key matches, which is the part that cannot diverge. The tiebreaker, which can, is unasserted.

The new test passes because `addSupersedingRow` bumps `item_last_updated_at` and `last_updated_at` by the same `+60s`, so both orderings agree by construction. Bumping only one of the two in a test would expose the divergence.

Suggest aligning phase 1 to `last_updated_at DESC`, matching main's pre-existing dedup semantics:

```suggestion
                ORDER BY dataset_item_id DESC, last_updated_at DESC
```

And extending `invariantOne` to compare the **full** `ORDER BY` clause of both phases rather than just the leading key.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -413,7 +413,51 @@ SELECT count(DISTINCT dataset_item_id) AS count
...
+     * use {@code optimize_read_in_order} and forces a full blocking sort over the whole version. Carrying
+     * {@code data} through that sort makes peak memory scale with the version's total payload rather than with
+     * the page size: a 120k-item version at ~138 KiB per item peaked at 30 GiB and returned 500 (OPIK-8109).
+     * <p>
+     * Phase 1 resolves the page's {@code dataset_item_id}s from narrow columns only; phase 2 fetches the payload
+     * for just those ids, served by the {@code dataset_item_id} skip indexes from migration {@code 000074}.
+     * Truncation stays in phase 2 so the image regex runs over the page instead of the whole version.
+     * <p>
+     * Three invariants hold this together. Each one broke a draft of this query, and each has a regression test
+     * in {@code DatasetItemVersionPaginationTest}:
```
JetoPistola, line 430: 💡 **suggestion** | Documentation

This references `DatasetItemVersionPaginationTest`, which doesn't exist. The tests actually landed in `DatasetItemVersionQueryShapeTest` and `DatasetVersionResourceTest$BoundedPageReads`, so the pointer the next editor of this query would follow is a dead end.

```suggestion
     * in {@code DatasetItemVersionQueryShapeTest} and {@code DatasetVersionResourceTest$BoundedPageReads}:
```

(Same issue independently flagged by `baz-reviewer[bot]`.)

JetoPistola:
👋 **Review summary**

**What looks good**
- The root-cause analysis is genuinely excellent. Identifying that `optimize_read_in_order` can't apply because `dataset_item_id` isn't in the ordering key — and that `replaceRegexpAll` expands every value *before* `substring` trims it, making `truncate=true` the worst case rather than a mitigation — is sharp diagnosis, not guesswork.
- The before/after table with `read_bytes` and peak memory is exactly the evidence a change like this needs. 15.83 GiB → 19.70 MiB read, 29.97 GiB → 14.01 MiB peak.
- Proactively flagging that `data`-derived filters and sorts remain payload-proportional (~5–13%), and that a ~15× larger dataset with a payload filter would hit the cap again, is honest reporting of your own work's limits rather than stopping at the happy path.
- Reusing the `000074` skip indexes so phase 2's `IN (...)` is served without a schema change, migration, or any change to template variables and bindings keeps the blast radius genuinely small.
- Documenting the invariants in javadoc on the query constant — where the next person to edit it will actually see them — is the right place for that knowledge.

**Overall**
The two-phase approach is the correct fix and the performance work behind it is thorough. The concern is that splitting one flat `SELECT` into two changed something subtle that the split wasn't accounting for: phase 1 lost the projection aliases that the filter predicate was implicitly relying on. Because the un-aliased names all exist as real columns on the table, the wrong resolution is silent rather than an error — an `id` filter now matches the physical row UUID instead of `dataset_item_id` and returns an empty page. Related: the two phases dedupe on different tiebreaker columns (`item_last_updated_at` vs `last_updated_at`), so they can pick different physical rows for the same id.

Both are contained fixes to the phase-1 CTE, not problems with the approach. Worth noting the test suite couldn't have caught either: no test on this endpoint filters on `ID`, `CREATED_AT`, or `LAST_UPDATED_AT`, and `addSupersedingRow` bumps both timestamp columns by the same `+60s`, so the two orderings agree by construction. Adding an `id`-filter case would be a cheap, high-value guard.

**Inline comments**: 2 blockers, 1 suggestion

---

*Caveat on verification:* these findings come from reading migration `000047`, `FilterQueryBuilder`'s column map, and the pre-PR query — not from executing either query against a live ClickHouse. The first blocker turns on ClickHouse resolving `WHERE` against projection aliases, so an empirical check against the affected dataset would be worth doing before acting on it.

PR #8040, opened 2026-08-27
[NA] [BE] fix: bind dataset item filters to item-level columns in page and count

thiagohora:
## Details

Follow-up to #8033. The two-phase page read it introduced resolves the page's ids in a CTE that projected only `dataset_item_id`. ClickHouse binds `WHERE` and `ORDER BY` to `SELECT` aliases, and `FilterQueryBuilder` emits bare `id`, `created_at`, `last_updated_at`, `created_by` and `last_updated_by` — so without those aliases in scope the predicates bound to this table's *physical* columns (the snapshot row's own id and timestamps) rather than the item-level ones the response projects.

- **The symptom is a short page, not a wrong row.** Phase 2 re-applies the filters and *does* carry the aliases, so it silently corrects the selection — phase 1 just spends page slots on items phase 2 then drops. Verified on production, on a version whose rows were snapshotted 6.4 days after the items were authored: a `created_at` filter selected **175 items where the correct answer was 0**.
- **`SELECT_DATASET_ITEM_VERSIONS_COUNT` had the same gap and still does on `main`** — no projection aliases at all, so the page `total` was computed against the physical columns while the rows were selected against the item-level ones. On the same version the endpoint returned **one row while reporting a total of 176**. Now wrapped in a subquery that puts the aliases in scope; `* EXCEPT` keeps every other column available so filters on `data`, `source`, `tags`, `trace_id` and `span_id` still resolve.
- The count wrapper is free: ClickHouse prunes the unread payload through it, and the read is byte-identical to the unwrapped form — 16.81 MiB unfiltered, 16.19 GiB with a data filter, in both shapes.
- Phase 1 now projects the same aliases as phase 2, which also lets both phases use identical `ORDER BY` text instead of one naming `item_last_updated_at` and the other its alias.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Related, not resolved here: OPIK_8109

## AI-WATERMARK

AI-WATERMARK: yes

- If yes:
  - Tools: Claude Code CLI
  - Model(s): Claude Opus 5
  - Scope: Root-cause investigation, SQL fix, tests
  - Human verification: ClickHouse alias-binding semantics established with controlled data on 26.3.16; both defects reproduced and the fixes confirmed against production data; every new test mutation-checked by reverting the fix it guards.

## Testing

`mvn test -Dtest='DatasetVersionResourceTest$BoundedPageReads,DatasetItemVersionQueryShapeTest'` — **27 tests, 0 failures** (19 behavioural + 8 structural), up from 12. Local, testcontainers, macOS.

`filters__bindToItemLevelColumns` is parameterized over all five aliased fields in both directions (10 cases) against a fixture whose item-level metadata differs from its snapshot row in every column, and each case also asserts `total` agrees with the returned rows — so the count/list divergence cannot come back either. New coverage for cursor pagination, filtered-page disjointness, an empty filter result, an offset past the end, and `truncate` combined with a filter.

**Mutation-checked.** Removing the phase-1 aliases and nothing else turns exactly two tests red — `createdAtFilter__bindsToItemAuthoringTime` (`Expected size: 3 but was: 2`) and `invariantFour__phaseOneProjectsFilterVisibleAliases` — while the other 25 stay green; restoring them turns both green. Worth noting the first version of that behavioural test **passed against the broken query** and was rewritten until it failed for the right reason.

**Not a regression from this PR:** `DatasetsResourceTest$FindDatasets` can fail on `getDatasets__whenFetchingAllDatasets__thenReturnDatasetsSortedByByValidFields[3]` (`NAME` ASC). Unmodified `main` fails it identically; this branch passes it on an unloaded machine. Root cause is a collation mismatch — `DatasetDAO` sorts in MySQL under `utf8mb4_unicode_ci` (UCA, where `_` has a lower primary weight than digits) while the test's expectation sorts in Java with `String.CASE_INSENSITIVE_ORDER` (code-point, `_` = 0x5F > `2` = 0x32). Confirmed against a real MySQL 8: the DB's ordering matches the API response exactly and Java's does not. It is flaky rather than always-red because Podam generates the names, and it affects every string case in that parameterized set (`NAME`, `DESCRIPTION`, `CREATED_BY`, `LAST_UPDATED_BY`, `TAGS`). Happy to fix separately.

**Not run:** the frontend suite (no FE changes). No video — internal query shape, no UI change.

## Documentation

None. The invariants that keep the two phases correct are documented in javadoc on the query constants, where the next person to edit them will see them. This PR adds the alias-binding invariant as a fourth entry and corrects a javadoc reference to a test class that did not exist.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
@@ -427,7 +427,7 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
      * Truncation stays in phase 2 so the image regex runs over the page instead of the whole version.
      * <p>
      * Three invariants hold this together. Each one broke a draft of this query, and each has a regression test
-     * in {@code DatasetItemVersionPaginationTest}:
+     * in {@code DatasetItemVersionQueryShapeTest}:
      * <ol>
      *   <li>Both phases order by {@code dataset_item_id DESC} first. A mismatch reorders the page.</li>
      *   <li>The ordering's leading key, {@code dataset_item_id}, is unique per output row after
@@ -439,18 +439,36 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
      *   <li>Phase 2 repeats {@code dataset_item_filters}. Without it, an id selected in phase 1 via a superseded
      *       row that matches the filter resolves in phase 2 to the newest row for that id, which may not match,
      *       returning rows the caller filtered out.</li>
+     *   <li>Phase 1 projects the same aliases as phase 2. ClickHouse binds {@code WHERE} and {@code ORDER BY} to
+     *       {@code SELECT} aliases, and the dataset item filters emit bare {@code id}, {@code created_at},
+     *       {@code last_updated_at}, {@code created_by} and {@code last_updated_by}. Drop the aliases and those
+     *       predicates bind to this table's physical columns instead: on a version whose rows were snapshotted
+     *       long after the items were authored, a {@code created_at} filter returned 175 items where the correct
+     *       answer was 0.</li>
      * </ol>
      */
     private static final String SELECT_DATASET_ITEM_VERSIONS = """
             WITH page AS (
-                SELECT dataset_item_id
+                SELECT
+                    dataset_item_id,
+                    -- Same aliases as phase 2's projection. ClickHouse resolves WHERE and ORDER BY
+                    -- against SELECT aliases, and the dataset item filters emit bare `id`,
+                    -- `created_at`, `last_updated_at`, `created_by` and `last_updated_by`. Without
+                    -- these aliases those predicates would bind to the physical columns of
+                    -- dataset_item_versions instead -- a different row id and different timestamps --
+                    -- so a filtered page would silently select the wrong items.
+                    dataset_item_id AS id,
+                    item_created_at AS created_at,
+                    item_last_updated_at AS last_updated_at,
+                    item_created_by AS created_by,
+                    item_last_updated_by AS last_updated_by
                 FROM dataset_item_versions
                 WHERE dataset_id = :datasetId
                 AND dataset_version_id = :versionId
                 AND workspace_id = :workspace_id
                 <if(lastRetrievedId)>AND dataset_item_id \\< :lastRetrievedId<endif>
                 <if(dataset_item_filters)>AND (<dataset_item_filters>)<endif>
-                ORDER BY dataset_item_id DESC, item_last_updated_at DESC
+                ORDER BY dataset_item_id DESC, last_updated_at DESC
                 LIMIT 1 BY dataset_item_id
                 <if(lastRetrievedId)>
                 LIMIT :limit
@@ -484,13 +502,37 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
             LIMIT 1 BY dataset_item_id
             """;
 
+    /**
+     * Counts the items a page request would return.
+     * <p>
+     * The subquery exists to put the item-level aliases in scope before the filters are applied. ClickHouse
+     * binds {@code WHERE} to {@code SELECT} aliases, and the dataset item filters emit bare {@code id},
+     * {@code created_at}, {@code last_updated_at}, {@code created_by} and {@code last_updated_by} — all of
+     * which also exist physically on this table with different meanings. Counting against the physical columns
+     * while {@code SELECT_DATASET_ITEM_VERSIONS} selects rows against the item-level ones made the two
+     * disagree: on a version whose rows were snapshotted long after the items were authored, a
+     * {@code created_at} filter returned one row while reporting a total of 176.
+     * <p>
+     * {@code * EXCEPT} keeps every other column in scope so filters on {@code data}, {@code source},
+     * {@code tags}, {@code trace_id} and {@code span_id} still resolve. It costs nothing: ClickHouse prunes
+     * the unread payload through the wrapper, and the read is byte-identical to the unwrapped form.
+     */
     private static final String SELECT_DATASET_ITEM_VERSIONS_COUNT = """
             SELECT count(DISTINCT dataset_item_id) as count
-            FROM dataset_item_versions
-            WHERE dataset_id = :datasetId
-            AND dataset_version_id = :versionId
-            AND workspace_id = :workspace_id
-            <if(dataset_item_filters)>AND (<dataset_item_filters>)<endif>
+            FROM (
+                SELECT
+                    * EXCEPT (id, created_at, last_updated_at, created_by, last_updated_by),
+                    dataset_item_id AS id,
+                    item_created_at AS created_at,
+                    item_last_updated_at AS last_updated_at,
+                    item_created_by AS created_by,
+                    item_last_updated_by AS last_updated_by
+                FROM dataset_item_versions
+                WHERE dataset_id = :datasetId
+                AND dataset_version_id = :versionId
+                AND workspace_id = :workspace_id
+                <if(dataset_item_filters)>AND (<dataset_item_filters>)<endif>
+            )
             """;
 
     private static final String DELETE_ITEMS_FROM_VERSION = """
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
@@ -446,6 +446,13 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
      *       long after the items were authored, a {@code created_at} filter returned 175 items where the correct
      *       answer was 0.</li>
      * </ol>
+     * <p>
+     * One caveat for whoever edits this next: alias binding is the default of a ClickHouse setting, not a
+     * language guarantee. Under {@code prefer_column_name_to_alias=1} resolution flips back to the source
+     * columns and every filter below silently returns to the broken behaviour. That setting is off by default
+     * in every version we run and is set nowhere in this repo, which is why these queries filter against the
+     * projection instead of paying for an extra subquery on the read path. The two facts are linked: flipping
+     * it would reintroduce exactly the defect this shape exists to prevent.
      */
     private static final String SELECT_DATASET_ITEM_VERSIONS = """
             WITH page AS (
@@ -516,6 +523,9 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
      * {@code * EXCEPT} keeps every other column in scope so filters on {@code data}, {@code source},
      * {@code tags}, {@code trace_id} and {@code span_id} still resolve. It costs nothing: ClickHouse prunes
      * the unread payload through the wrapper, and the read is byte-identical to the unwrapped form.
+     * <p>
+     * Alias binding here is the default of {@code prefer_column_name_to_alias} rather than a language
+     * guarantee; see the note on {@code SELECT_DATASET_ITEM_VERSIONS}.
      */
     private static final String SELECT_DATASET_ITEM_VERSIONS_COUNT = """
             SELECT count(DISTINCT dataset_item_id) as count
@@ -596,10 +606,10 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
                         tags,
                         evaluators,
                         execution_policy,
-                        created_at,
-                        last_updated_at,
-                        created_by,
-                        last_updated_by
+                        item_created_at AS created_at,
+                        item_last_updated_at AS last_updated_at,
+                        item_created_by AS created_by,
+                        item_last_updated_by AS last_updated_by
                     FROM dataset_item_versions FINAL
                     WHERE workspace_id = :workspace_id
                       AND dataset_id = :datasetId
@@ -675,6 +685,7 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
             	INNER JOIN experiments_resolved e ON e.id = ei.experiment_id
             	LEFT JOIN dataset_item_versions AS lookup_div FINAL
             	    ON lookup_div.workspace_id = ei.workspace_id
+            	    AND lookup_div.dataset_id = :datasetId
             	    AND lookup_div.id = ei.dataset_item_id
             	WHERE ei.workspace_id = :workspace_id
             	<if(experiment_ids)>AND ei.experiment_id IN :experiment_ids<endif>
@@ -765,10 +776,10 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
                     div_dedup.trace_id AS trace_id,
                     div_dedup.span_id AS span_id,
                     div_dedup.tags AS tags,
-                    div_dedup.created_at AS created_at,
-                    div_dedup.last_updated_at AS last_updated_at,
-                    div_dedup.created_by AS created_by,
-                    div_dedup.last_updated_by AS last_updated_by
+                    div_dedup.item_created_at AS created_at,
+                    div_dedup.item_last_updated_at AS last_updated_at,
+                    div_dedup.item_created_by AS created_by,
+                    div_dedup.item_last_updated_by AS last_updated_by
                 FROM (
                     SELECT *
                     FROM dataset_item_versions
@@ -853,10 +864,10 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
                     div_dedup.trace_id AS trace_id,
                     div_dedup.span_id AS span_id,
                     div_dedup.tags AS tags,
-                    div_dedup.created_at AS created_at,
-                    div_dedup.last_updated_at AS last_updated_at,
-                    div_dedup.created_by AS created_by,
-                    div_dedup.last_updated_by AS last_updated_by
+                    div_dedup.item_created_at AS created_at,
+                    div_dedup.item_last_updated_at AS last_updated_at,
+                    div_dedup.item_created_by AS created_by,
+                    div_dedup.item_last_updated_by AS last_updated_by
                 FROM (
                     SELECT *
                     FROM dataset_item_versions
@@ -892,6 +903,7 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
                 FROM experiment_item_aggregates AS eia FINAL
                 LEFT JOIN dataset_item_versions AS lookup_div FINAL
                     ON lookup_div.workspace_id = eia.workspace_id
+                    AND lookup_div.dataset_id = :datasetId
                     AND lookup_div.id = eia.dataset_item_id
                 WHERE eia.workspace_id = :workspace_id
                 AND eia.experiment_id IN (SELECT id FROM experiment_aggregated_scope_ids)
@@ -1089,6 +1101,7 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
             	INNER JOIN experiments_resolved e ON e.id = ei.experiment_id
             	LEFT JOIN dataset_item_versions AS lookup_div FINAL
             	    ON lookup_div.workspace_id = ei.workspace_id
+            	    AND lookup_div.dataset_id = :datasetId
             	    AND lookup_div.id = ei.dataset_item_id
             	WHERE ei.workspace_id = :workspace_id
             	<if(experiment_ids)>AND ei.experiment_id IN :experiment_ids<endif>
@@ -1136,10 +1149,14 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
                     div_dedup.tags AS tags,
                     div_dedup.evaluators AS evaluators,
                     div_dedup.execution_policy AS execution_policy,
-                    div_dedup.created_at AS item_created_at,
-                    div_dedup.last_updated_at AS item_last_updated_at,
-                    div_dedup.created_by AS item_created_by,
-                    div_dedup.last_updated_by AS item_last_updated_by
+                    div_dedup.item_created_at AS item_created_at,
+                    div_dedup.item_last_updated_at AS item_last_updated_at,
+                    div_dedup.item_created_by AS item_created_by,
+                    div_dedup.item_last_updated_by AS item_last_updated_by,
+                    div_dedup.item_created_at AS created_at,
+                    div_dedup.item_last_updated_at AS last_updated_at,
+                    div_dedup.item_created_by AS created_by,
+                    div_dedup.item_last_updated_by AS last_updated_by
                 FROM (
                     SELECT *
                     FROM dataset_item_versions
@@ -1359,10 +1376,10 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
                         tags,
                         evaluators,
                         execution_policy,
-                        created_at,
-                        last_updated_at,
-                        created_by,
-                        last_updated_by
+                        item_created_at AS created_at,
+                        item_last_updated_at AS last_updated_at,
+                        item_created_by AS created_by,
+                        item_last_updated_by AS last_updated_by
                     FROM dataset_item_versions FINAL
                     WHERE workspace_id = :workspace_id
                     AND dataset_id  = :datasetId
@@ -1401,10 +1418,14 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
                     div_dedup.tags AS tags,
                     div_dedup.evaluators AS evaluators,
                     div_dedup.execution_policy AS execution_policy,
-                    div_dedup.created_at AS item_created_at,
-                    div_dedup.last_updated_at AS item_last_updated_at,
-                    div_dedup.created_by AS item_created_by,
-                    div_dedup.last_updated_by AS item_last_updated_by
+                    div_dedup.item_created_at AS item_created_at,
+                    div_dedup.item_last_updated_at AS item_last_updated_at,
+                    div_dedup.item_created_by AS item_created_by,
+                    div_dedup.item_last_updated_by AS item_last_updated_by,
+                    div_dedup.item_created_at AS created_at,
+                    div_dedup.item_last_updated_at AS last_updated_at,
+                    div_dedup.item_created_by AS created_by,
+                    div_dedup.item_last_updated_by AS last_updated_by
                 FROM (
                     SELECT *
                     FROM dataset_item_versions
@@ -1549,6 +1570,7 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
                     FROM experiment_item_aggregates AS eia FINAL
                     LEFT JOIN dataset_item_versions AS lookup_div FINAL
                         ON lookup_div.workspace_id = eia.workspace_id
+                        AND lookup_div.dataset_id = :datasetId
                         AND lookup_div.id = eia.dataset_item_id
                     WHERE eia.workspace_id = :workspace_id
                     AND eia.experiment_id IN (SELECT id FROM experiment_aggregated_scope_ids)
@@ -2274,6 +2296,7 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
                 INNER JOIN experiments_resolved e ON e.id = ei.experiment_id
                 LEFT JOIN dataset_item_versions AS lookup_div FINAL
                     ON lookup_div.workspace_id = ei.workspace_id
+                    AND lookup_div.dataset_id = :datasetId
                     AND lookup_div.id = ei.dataset_item_id
                 WHERE ei.workspace_id = :workspace_id
                 <if(experiment_ids)>AND ei.experiment_id IN :experiment_ids<endif>
@@ -2441,10 +2464,10 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
                             div_dedup.trace_id AS trace_id,
                             div_dedup.span_id AS span_id,
                             div_dedup.tags AS tags,
-                            div_dedup.created_at AS created_at,
-                            div_dedup.last_updated_at AS last_updated_at,
-                            div_dedup.created_by AS created_by,
-                            div_dedup.last_updated_by AS last_updated_by,
+                            div_dedup.item_created_at AS created_at,
+                            div_dedup.item_last_updated_at AS last_updated_at,
+                            div_dedup.item_created_by AS created_by,
+                            div_dedup.item_last_updated_by AS last_updated_by,
                             div_dedup.dataset_version_id AS dataset_version_id
                         FROM (
                             SELECT *
@@ -2486,7 +2509,11 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
                             div_dedup.source AS source,
                             div_dedup.trace_id AS trace_id,
                             div_dedup.span_id AS span_id,
-                            div_dedup.tags AS tags
+                            div_dedup.tags AS tags,
+                            div_dedup.item_created_at AS created_at,
+                            div_dedup.item_last_updated_at AS last_updated_at,
+                            div_dedup.item_created_by AS created_by,
+                            div_dedup.item_last_updated_by AS last_updated_by
                         FROM (
                             SELECT *
                             FROM dataset_item_versions
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
@@ -420,7 +420,8 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
      * {@code (workspace_id, dataset_id, dataset_version_id, id)}, so ordering by {@code dataset_item_id} cannot
      * use {@code optimize_read_in_order} and forces a full blocking sort over the whole version. Carrying
      * {@code data} through that sort makes peak memory scale with the version's total payload rather than with
-     * the page size: a 120k-item version at ~138 KiB per item peaked at 30 GiB and returned 500 (OPIK-8109).
+     * the page size, so a version whose total payload exceeds the per-query memory cap fails with
+     * MEMORY_LIMIT_EXCEEDED even when a single row is requested (OPIK-8109).
      * <p>
      * Phase 1 resolves the page's {@code dataset_item_id}s from narrow columns only; phase 2 fetches the payload
      * for just those ids, served by the {@code dataset_item_id} skip indexes from migration {@code 000074}.
@@ -442,9 +443,9 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
      *   <li>Phase 1 projects the same aliases as phase 2. ClickHouse binds {@code WHERE} and {@code ORDER BY} to
      *       {@code SELECT} aliases, and the dataset item filters emit bare {@code id}, {@code created_at},
      *       {@code last_updated_at}, {@code created_by} and {@code last_updated_by}. Drop the aliases and those
-     *       predicates bind to this table's physical columns instead: on a version whose rows were snapshotted
-     *       long after the items were authored, a {@code created_at} filter returned 175 items where the correct
-     *       answer was 0.</li>
+     *       predicates bind to this table's physical columns instead. Where a version's rows were snapshotted
+     *       well after its items were authored the two clocks diverge, and the filter then selects an entirely
+     *       different set of items.</li>
      * </ol>
      * <p>
      * One caveat for whoever edits this next: alias binding is the default of a ClickHouse setting, not a
@@ -458,12 +459,6 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
             WITH page AS (
                 SELECT
                     dataset_item_id,
-                    -- Same aliases as phase 2's projection. ClickHouse resolves WHERE and ORDER BY
-                    -- against SELECT aliases, and the dataset item filters emit bare `id`,
-                    -- `created_at`, `last_updated_at`, `created_by` and `last_updated_by`. Without
-                    -- these aliases those predicates would bind to the physical columns of
-                    -- dataset_item_versions instead -- a different row id and different timestamps --
-                    -- so a filtered page would silently select the wrong items.
                     dataset_item_id AS id,
                     item_created_at AS created_at,
                     item_last_updated_at AS last_updated_at,
@@ -517,8 +512,7 @@ class DatasetItemVersionDAOImpl implements DatasetItemVersionDAO {
      * {@code created_at}, {@code last_updated_at}, {@code created_by} and {@code last_updated_by} — all of
      * which also exist physically on this table with different meanings. Counting against the physical columns
      * while {@code SELECT_DATASET_ITEM_VERSIONS} selects rows against the item-level ones made the two
-     * disagree: on a version whose rows were snapshotted long after the items were authored, a
-     * {@code created_at} filter returned one row while reporting a total of 176.
+     * disagree, so a filtered page could report a total its own rows could not account for.
      * <p>
      * {@code * EXCEPT} keeps every other column in scope so filters on {@code data}, {@code source},
      * {@code tags}, {@code trace_id} and {@code span_id} still resolve. It costs nothing: ClickHouse prunes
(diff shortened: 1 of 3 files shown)


CometActions:

This lands on a real user-visible path, not just SQL. `TOGGLE_DATASET_VERSIONING_ENABLED` defaults to true, so `GET /v1/private/datasets/{id}/items` with `filters` goes through `DatasetItemVersionDAO`, and the items table's ID filter (the one `buildDatasetFilterColumns(..., true)` exposes) emits a bare `id` — which on main binds to `dataset_item_versions.id`, the snapshot-row id, so filtering by an item's ID returned an empty table. That part reproduces on a fresh install with no clock trickery. What stops me proposing an e2e spec is the map, not the change: our taxonomy has no capability for filtering dataset items on the read path. `datasets.search-items` is the search box (a separate `search` query param), `datasets.filter-scoped-batch-update` / `filter-scoped-mutation-validation` are the delete and batch-update endpoints, and `datasets.view-items` only counts rows on an unfiltered page — so there is no honest key to tag a new spec with. Raising that gap for taxonomy-discovery rather than inventing a key. Worth saying: your `DatasetVersionResourceTest.filters__bindToItemLevelColumns` already covers all five fields in both directions plus page/total agreement, so this is not shipping untested.

**Not testable yet.** The created_at / last_updated_at / created_by / last_updated_by half of the same fix is not reachable from the e2e estate even once a capability key exists. It only becomes observable when an item's authoring metadata diverges from its snapshot row's, and your own fixture needs a raw ClickHouse INSERT to produce that — nothing the SDK or UI can seed on a fresh install, and the UI filter menu does not offer those columns anyway. The ID half is the testable one.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/33073482463)

_Re-checked after a push on 27 Aug 12:47 UTC — nothing the verdict depends on changed._

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -484,13 +502,37 @@ AND dataset_item_id IN (SELECT dataset_item_id FROM page)
...
+                    item_created_at AS created_at,
+                    item_last_updated_at AS last_updated_at,
+                    item_created_by AS created_by,
+                    item_last_updated_by AS last_updated_by
+                FROM dataset_item_versions
+                WHERE dataset_id = :datasetId
+                AND dataset_version_id = :versionId
+                AND workspace_id = :workspace_id
+                <if(dataset_item_filters)>AND (<dataset_item_filters>)<endif>
+            )
```
JetoPistola, line 539: 💡 **suggestion** | Correctness (out of scope, but same root cause — and destructive)

The root cause you've characterised so precisely isn't confined to the two queries fixed here. `DELETE_ITEMS_FROM_VERSION` (just below) and `COUNT_ITEMS` both interpolate `<dataset_item_filters>` straight against `dataset_item_versions` with no item-level aliases in scope — so the same five bare names bind to the snapshot row's physical columns.

That path is reachable from a public endpoint. `DELETE /v1/private/datasets/items` with `dataset_id` + `filters` routes through `DatasetItemService.deleteItemsFromExistingVersionByFilters` → `removeItemsFromVersionByFilters` → `DELETE_ITEMS_FROM_VERSION`. On your production version with the 6.4-day snapshot gap, a `created_at`-filtered delete would match on row time rather than authoring time — the same 175-vs-0 selection error, except the outcome is deleted rows rather than a short page.

`countItemsMatchingFilters` gates that delete using `COUNT_ITEMS`, which has the identical gap, so the preflight count agrees with the wrong deletion and the caller gets a plausible-looking `deletedCount` back. Consistently wrong is harder to notice than inconsistently wrong.

Genuinely out of scope for a PR titled around page-and-count, and I'd rather see it as a follow-up than bolted on here. But given the read-side defect took production data to spot, it's worth filing before the shape of this fix fades from memory. `COUNT_ROWS_IN_VERSION` has the same shape via `exclude_filters` (OPIK-6390 UUID-pool sizing), for what it's worth.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -484,13 +502,37 @@ AND dataset_item_id IN (SELECT dataset_item_id FROM page)
...
             """;
```
JetoPistola, line 540: 💡 **suggestion** | Correctness (out of scope — the experiment-items siblings)

The experiment-items queries wrap their filters in a projection, which looks like it solves this, but the projection passes the four snapshot columns through *bare*:

```
dataset_item_id AS id,     <- aliased correctly
id AS row_id,
...
created_at,                <- snapshot row's, under the name the filter wants
last_updated_at,
created_by,
last_updated_by
FROM dataset_item_versions FINAL
...
WHERE <dataset_item_filters>
```

That's `dataset_items_filtered_ids` in both `SELECT_DATASET_ITEM_VERSIONS_WITH_EXPERIMENT_ITEMS_COUNT` (~lines 556-559) and `SELECT_DATASET_ITEM_VERSIONS_WITH_EXPERIMENT_ITEMS` (~lines 1319-1322). `ID` is fine in both; `CREATED_AT`, `LAST_UPDATED_AT`, `CREATED_BY` and `LAST_UPDATED_BY` are not. Adding an explicit projection there is a mechanical change and the existing subquery boundary means there's nowhere new to put it.

Worth noting `dataset_items_aggr_resolved` in the same query aliases in the *opposite* direction (`div_dedup.created_at AS item_created_at`), so a `created_at` filter can mean different things depending on which StringTemplate branch is taken. That divergence is the kind of thing your invariant javadoc exists to prevent — extending it to cover these siblings would be worth more than the SQL fix itself.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -484,13 +502,37 @@ AND dataset_item_id IN (SELECT dataset_item_id FROM page)
...
+     * the unread payload through the wrapper, and the read is byte-identical to the unwrapped form.
+     */
     private static final String SELECT_DATASET_ITEM_VERSIONS_COUNT = """
             SELECT count(DISTINCT dataset_item_id) as count
-            FROM dataset_item_versions
-            WHERE dataset_id = :datasetId
-            AND dataset_version_id = :versionId
-            AND workspace_id = :workspace_id
-            <if(dataset_item_filters)>AND (<dataset_item_filters>)<endif>
+            FROM (
```
JetoPistola, line 526: 💡 **suggestion** | Testing

The count defect is the one your description says regressed silently and is *still* live on `main` — precisely because nothing structural was guarding it. `invariantFour__phaseOneProjectsFilterVisibleAliases` now guards phase 1, but `DatasetItemVersionQueryShapeTest` only reflects `SELECT_DATASET_ITEM_VERSIONS`; the count constant isn't read at all.

So the asymmetry that produced this bug is still there: unwrap this subquery and the only thing that fails is `filters__bindToItemLevelColumns`, which needs testcontainers and a divergent fixture. The phase-1 equivalent fails in milliseconds against a string.

A sibling assertion reflecting `SELECT_DATASET_ITEM_VERSIONS_COUNT` and checking the same five aliases are in scope would close it — same reflection helper, a few lines. Given the javadoc you've added here is now the primary defence, a test that fails when someone edits past it seems worth having.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -439,18 +439,36 @@ SELECT count(DISTINCT dataset_item_id) AS count
...
      *       returning rows the caller filtered out.</li>
+     *   <li>Phase 1 projects the same aliases as phase 2. ClickHouse binds {@code WHERE} and {@code ORDER BY} to
+     *       {@code SELECT} aliases, and the dataset item filters emit bare {@code id}, {@code created_at},
+     *       {@code last_updated_at}, {@code created_by} and {@code last_updated_by}. Drop the aliases and those
+     *       predicates bind to this table's physical columns instead: on a version whose rows were snapshotted
+     *       long after the items were authored, a {@code created_at} filter returned 175 items where the correct
+     *       answer was 0.</li>
      * </ol>
      */
     private static final String SELECT_DATASET_ITEM_VERSIONS = """
```
JetoPistola: ❓ **question** | Robustness

The invariant is stated as "ClickHouse binds `WHERE` and `ORDER BY` to `SELECT` aliases" — which is true under the default, but it's the default of a *setting*. With `prefer_column_name_to_alias=1` the resolution flips back to the source columns and both queries silently return to the broken behaviour.

I checked and it isn't set anywhere in the repo, so this is correct as deployed and the setting is off by default in every ClickHouse version we run — I don't think it warrants restructuring the query, and the alternative (filtering in an outer query over the projection) costs a subquery on the hot read path for a hypothetical.

But the javadoc currently reads as though alias binding is a language guarantee rather than a configured default. Since these invariants are explicitly written for "the next person to edit them," would it be worth one clause noting the dependency? It's the difference between someone reading it and being informed versus reading it and being misled if that setting ever gets flipped.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
```diff
@@ -5905,6 +5919,294 @@ void pages__areOrderedAndDisjoint() {
...
+
+            clickHouseTemplate.nonTransaction(connection -> {
+                Statement statement = connection.createStatement(sql)
+                        .bind("dataset_id", datasetId.toString())
+                        .bind("version_id", versionId.toString())
+                        .bind("backdated_item_id", BACKDATED_ITEM_ID)
+                        .bind("tag", "\"" + tag + "\"")
+                        .bind("workspace_id", WORKSPACE_ID);
+
+                Mono<? extends Result> result = Mono.from(statement.execute());
```
JetoPistola: 🧹 **nit** | Style

The local is assigned and immediately returned:

```suggestion
                return Mono.from(statement.execute());
```

Same in `addItemWithDivergentRowMetadata` below. Pre-existing pattern elsewhere in the file, so entirely take-it-or-leave-it.

JetoPistola:
👋 **Review summary**

**What looks good**
- The root-cause analysis is unusually good. "The symptom is a short page, not a wrong row" is the sentence that makes this reviewable — it explains why phase 2's re-filtering *masked* the defect, which is exactly why it survived #8033.
- Finding the second defect in `SELECT_DATASET_ITEM_VERSIONS_COUNT` while fixing the first, rather than stopping at the reported symptom.
- Mutation-checking each new test by reverting the fix it guards, and reporting that the first version of the behavioural test **passed against the broken query** and had to be rewritten. That's the detail most PRs leave out, and it's the one that tells a reviewer the tests are load-bearing.
- `filters__bindToItemLevelColumns` asserting both directions per field *and* `total` agreeing with the rows — the count/list divergence cannot come back.
- Ruling out the `DatasetsResourceTest` failure against unmodified `main` with a real MySQL, down to the collation weights, instead of labelling it flaky.

**Overall**
The fix is right and the approach is right: aliases in phase 1 so both phases resolve identically, and a subquery for the count so filters see the same columns the rows are selected against. The `* EXCEPT` list matches all five physical columns on the table, and phase 2's `IN (SELECT dataset_item_id FROM page)` is unaffected by the wider CTE projection.

My substantive comments are all about blast radius rather than this diff. The same bare-column binding is live on `DELETE_ITEMS_FROM_VERSION`, `COUNT_ITEMS`, `COUNT_ROWS_IN_VERSION`, and both experiment-items filter CTEs — and the delete one is reachable from a public endpoint, so a `created_at`-filtered delete removes rows chosen by snapshot time rather than authoring time. Out of scope here; worth filing while the shape of this fix is fresh.

**Inline comments**: 3 suggestions, 1 question, 1 nit — nothing blocking.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
thiagohora: Good catch, and you were right to check. Confirmed on 26.3.16 with crossed fixture data (row1 `id=AAA,dataset_item_id=ZZZ`, row2 the reverse):

| `prefer_column_name_to_alias` | binds to |
|---|---|
| `0` (default) | the alias — `dataset_item_id` |
| `1` | the physical `id` |

So it does flip fully back to the broken behaviour. I also confirmed the setting appears nowhere in the repo.

Took your read on the trade-off — not worth an outer subquery on the hot read path for a hypothetical — and added the clause instead. The javadoc now names the setting, states it is off by default in every version we run and unset here, explains why the queries filter against the projection anyway, and says flipping it reintroduces the defect the shape exists to prevent. Same note added on the count constant.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -439,18 +439,36 @@ SELECT count(DISTINCT dataset_item_id) AS count
...
+                    -- against SELECT aliases, and the dataset item filters emit bare `id`,
+                    -- `created_at`, `last_updated_at`, `created_by` and `last_updated_by`. Without
+                    -- these aliases those predicates would bind to the physical columns of
+                    -- dataset_item_versions instead -- a different row id and different timestamps --
+                    -- so a filtered page would silently select the wrong items.
+                    dataset_item_id AS id,
+                    item_created_at AS created_at,
+                    item_last_updated_at AS last_updated_at,
+                    item_created_by AS created_by,
+                    item_last_updated_by AS last_updated_by
```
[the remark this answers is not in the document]
thiagohora: Confirmed and addressed, though not by restructuring. Verified on 26.3.16 that `prefer_column_name_to_alias=1` does flip resolution to the source columns and reproduces the defect; also confirmed the setting is set nowhere in the repo, so this is correct as deployed.

Filtering in an outer query would cost a subquery on the hot read path to defend against a setting nobody enables, so instead the dependency is now documented on both constants — it was previously written as though alias binding were a language guarantee.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
thiagohora: This was the most useful comment in the review — you are right that the asymmetry which produced the bug was still there.

Added `countQuery__projectsTheSameFilterVisibleAliases` using the same reflection helper. It asserts the five aliases plus `* EXCEPT`, and additionally that the `<dataset_item_filters>` placeholder appears *after* the alias definitions, so the filter cannot be hoisted outside the subquery that defines them.

Mutation-checked: unwrapping the count back to main shape now fails 1 of 9 in **82 ms**, rather than needing testcontainers and a divergent fixture.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
thiagohora: Confirmed exactly as described. Both `dataset_items_filtered_ids` sites projected `dataset_item_id AS id` correctly and then passed `created_at`, `last_updated_at`, `created_by`, `last_updated_by` through bare. Re-pointed at the item-level columns in both constants.

Your note about `dataset_items_aggr_resolved` aliasing in the opposite direction was the thread worth pulling. Following it up:

- Two more sites (`..._COUNT`, `..._STATS`) aliased `div_dedup.created_at AS created_at` — the row column under the filter name.
- The STATS aggregated filter scope projected none of the four, so a filter on them raised `UNKNOWN_IDENTIFIER` — a latent 500. It has never fired: 22 days of `query_log` show nothing exercising that combination.
- `dataset_items_resolved` / `dataset_items_aggr_resolved` in the LIST query aliased the row columns *to* the `item_*` names, and the outer query feeds `di.item_created_at` to the response — so the comparison view has been reporting snapshot timestamps as the item timestamps. Corrected the source and exposed the bare names. No existing test asserted the previous values.

Agreed the javadoc mattered more than the SQL here; the invariant list now covers alias binding explicitly.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetVersionResourceTest.java
thiagohora: Applied — three of them, including `addItemWithDivergentRowMetadata`.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
thiagohora: Agreed, and splitting it out. Your trace is right and the framing is sharper than mine — I had flagged the shared root cause but not that `COUNT_ITEMS` gates the delete, so the preflight count agrees with the wrong deletion and returns a plausible `deletedCount`. Consistently wrong is the harder failure to notice.

One detail that reinforces keeping it separate: `DELETE_ITEMS_FROM_VERSION` is a bare `DELETE FROM` with no projection at all, so unlike the read paths there is nowhere to hang aliases. The fix means resolving ids through an aliased subquery — a materially different change on a destructive path, wanting its own tests.

For sizing: `query_log` shows **zero** filtered deletes and zero filtered write-path calls in 22 days, so there is no evidence data has already moved. Follow-up PR to come.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -1094,10 +1149,14 @@ dataset_items_resolved AS (
...
                     div_dedup.execution_policy AS execution_policy,
-                    div_dedup.created_at AS item_created_at,
-                    div_dedup.last_updated_at AS item_last_updated_at,
-                    div_dedup.created_by AS item_created_by,
-                    div_dedup.last_updated_by AS item_last_updated_by
+                    div_dedup.item_created_at AS item_created_at,
+                    div_dedup.item_last_updated_at AS item_last_updated_at,
+                    div_dedup.item_created_by AS item_created_by,
+                    div_dedup.item_last_updated_by AS item_last_updated_by,
+                    div_dedup.item_created_at AS created_at,
```
JetoPistola, line 1170: 💡 **suggestion** | Testing

This is the one thing I'd genuinely push on before merge.

You applied the shape-test argument to the count constant and the result was exactly as advertised — 82 ms instead of testcontainers. But the reasoning generalises to what this commit just changed, and it wasn't extended there. `DatasetItemVersionQueryShapeTest` reflects two constants; this commit modified five, and the three experiment-items ones have no assertion of any kind.

The comparison-view defect is the reason that matters. `div_dedup.created_at AS item_created_at` fed `di.item_created_at AS created_at` straight into the response, so the API has been reporting snapshot times as authoring times — user-visible wrong data, a category above the filter-binding bugs. Your own note says *"no existing test asserted the previous values."* That cuts both ways: nothing asserted the wrong values, and nothing now asserts the right ones. Re-point one of those aliases back and the suite stays green.

A reflection assertion over the three constants — that `item_*` aliases derive from `item_*` columns and never from the bare ones — would be the same shape and cost as `countQuery__projectsTheSameFilterVisibleAliases`. Given that a whole family of these was introduced by exactly this copy-paste and went unnoticed until a reviewer pulled the thread, the cheap structural guard seems worth more here than anywhere else in the PR.

A behavioural test on the comparison endpoint asserting the returned `created_at` equals the item's authoring time would be stronger still, but I'd take the structural one.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -633,6 +685,7 @@ experiment_items_scope AS (
             	INNER JOIN experiments_resolved e ON e.id = ei.experiment_id
             	LEFT JOIN dataset_item_versions AS lookup_div FINAL
             	    ON lookup_div.workspace_id = ei.workspace_id
+            	    AND lookup_div.dataset_id = :datasetId
```
JetoPistola, line 682: 💡 **suggestion** | Testing / Performance

I checked this one carefully because it's the only change in the commit that isn't about alias binding, and it holds up: `:datasetId` is already mandatory in all three constants (`experiments_resolved`, `experiment_aggregated_scope_ids`, `dataset_items_filtered_ids` all reference it) and bound unconditionally at every render site — lines ~3015, ~3190 and in `bindStatementParameters` — so there's no unbound-parameter path. Leaving the `slim_count` site alone is right too, since its inner SELECT already carries the predicate.

149.03M → 26,934 rows, and it clears the last recorded 500 on the cluster. The ordering-key argument is sound: `(workspace_id, dataset_id, dataset_version_id, id)` with only components 1 and 4 constrained gives no usable prefix. Worth noting it still can't reach `id`, since `dataset_version_id` stays unconstrained — so this widens the prefix from one component to two rather than achieving a full key match. Real fix, just not the whole way.

My concern is that none of that reasoning exists in the code. It's in the commit message, which is where reasoning goes to die — the javadoc at ~1022-1031 documents the join's stable-id purpose and wasn't updated, so the next reader sees a predicate that looks redundant next to `lookup_div.id = ei.dataset_item_id` and has every reason to "simplify" it away. That's the same failure mode the invariant list in this file exists to prevent, and you've argued the point better than I can in this PR's own description.

One comment at these sites, or a shape assertion that every `lookup_div` join carries the predicate, would close it.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -1047,6 +1101,7 @@ ORDER BY (workspace_id, dataset_id, id) DESC, last_updated_at DESC
...
```
JetoPistola, line 1118: ❓ **question** | Correctness

On the claim that the `lookup_div` output is identical: I think that's right in practice, and I want to lay out why it isn't structurally guaranteed, since "performance fix" and "cannot change results" aren't the same claim.

The join resolves legacy `ei.dataset_item_id` values that are version-row PKs. Adding `dataset_id = :datasetId` means a legacy id pointing at a row in a *different* dataset now misses where it previously matched. Two things absorb that: it's a `LEFT JOIN`, so no row drops, and `if(notEmpty(lookup_div.dataset_item_id), ..., ei.dataset_item_id)` falls back to the raw value. So the observable delta is confined to the grouping key for such a row — which could shift `COUNT(DISTINCT ...)`.

The reason it can't be ruled out structurally is that `ExperimentItemService.validateDatasetItemsWorkspace` checks only that the dataset item's **workspace** matches; nothing validates it belongs to the experiment's dataset. So a cross-dataset reference is acceptable at ingestion, and 75,786 items across six workspaces showing identical output is evidence those references don't exist in practice, not that they can't.

I'd argue the new behaviour is the more correct one either way — surfacing a foreign dataset's item id inside a single-dataset comparison view was already wrong, so this reads as an incidental bug fix rather than a risk. Mostly I want it on the record that this is empirical rather than structural, so it isn't later cited as "proven output-identical." Is the missing dataset-scoped validation worth a ticket alongside the delete-path follow-up?

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -1094,10 +1149,14 @@ dataset_items_resolved AS (
...
-                    div_dedup.created_by AS item_created_by,
-                    div_dedup.last_updated_by AS item_last_updated_by
+                    div_dedup.item_created_at AS item_created_at,
+                    div_dedup.item_last_updated_at AS item_last_updated_at,
+                    div_dedup.item_created_by AS item_created_by,
+                    div_dedup.item_last_updated_by AS item_last_updated_by,
+                    div_dedup.item_created_at AS created_at,
+                    div_dedup.item_last_updated_at AS last_updated_at,
+                    div_dedup.item_created_by AS created_by,
+                    div_dedup.item_last_updated_by AS last_updated_by
```
JetoPistola, line 1173: 🧹 **nit** | Maintainability

Projecting each item column under both its `item_*` name and its bare name is the right call — the outer query consumes `di.item_created_at` while filters need the bare names — but it reads like redundancy, and eight aliases over four columns is precisely the shape someone tidies up.

One line saying the two sets serve different consumers would protect it. Dropping the bare half silently reintroduces the filter-binding defect; dropping the `item_*` half breaks the response projection. Neither is obvious from the SQL.

JetoPistola:
👋 **Review summary** (re-review of `61083c6`)

All five findings from the previous review were addressed, and three of the four replies changed my understanding rather than just acknowledging the point.

**What looks good**
- You verified `prefer_column_name_to_alias=1` against 26.3.16 with crossed fixture data instead of taking my word that it was hypothetical — and it does fully flip back to the broken behaviour. Then took the trade-off (don't pay for a subquery on the hot read path) and documented the dependency instead. That's the right call *and* the right evidence for it.
- `countQuery__projectsTheSameFilterVisibleAliases` asserting the filter placeholder sits *after* the alias definitions is a better test than I described — it guards against hoisting, not just absence. 82 ms vs. testcontainers is the whole argument.
- The `dataset_items_aggr_resolved` thread turned up three more instances, a latent `UNKNOWN_IDENTIFIER` 500 in the STATS filter scope, and the comparison view reporting snapshot timestamps as item timestamps. I pointed at the inconsistency; you found the user-visible bug underneath it.
- The commit message documents the `lookup_div` change, its measurements, and what was deliberately left out. Nothing had to be reverse-engineered from the diff.
- Splitting the delete path out with a reason I hadn't considered — `DELETE FROM` has no projection to hang aliases on, so the fix is structurally different and belongs on a destructive path with its own tests.

**Overall**
The PR is materially stronger than at first review, and its scope grew for the right reason: the same root cause was live in more places, and one of them was returning wrong data to users.

The one gap I'd want closed before merge is coverage. The shape-test argument you applied to the count constant applies at least as strongly to what this commit changed — five constants touched, three with no assertion of any kind, including the comparison-view fix that "no existing test asserted the previous values." Re-point one alias and the suite stays green. The `lookup_div` rationale has the same problem in a different form: it's in the commit message, not the code, next to a predicate that looks redundant.

**Inline comments**: 2 suggestions, 1 question, 1 nit — nothing blocking.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
thiagohora: Fixed in 63c090b — you were right that this was the one to push on, and right that the suite stayed green when I re-pointed an alias back.

Two assertions added:

- `aliases__neverMapRowColumnsToItemLevelNames` reflects **every** SQL constant on the DAO and fails on any alias presenting a `dataset_item_versions` snapshot-row column under a name read as item-level.
- `experimentItemsQueries__resolveFiltersAgainstItemLevelColumns` asserts the three experiment-items constants expose all four filter-visible names from the `item_*` columns.

Both mutation-checked: re-pointing one alias at the row column now fails in **80 ms**.

Worth recording that writing the first assertion broadly was instructive — it flagged two legitimate aliases: `eia.created_at AS created_at` is the experiment item’s own timestamp on a different table, and `COPY_ITEMS_FROM_LEGACY` maps `dataset_items.created_at` to `item_created_at`, where the legacy column genuinely is the authoring time. Hence the `div_dedup` scoping, which is documented on the pattern.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
thiagohora: Fixed in 63c090b, and thank you for the correction on the prefix — you are right that this widens it from one component to two rather than achieving a full key match, since `dataset_version_id` stays unconstrained and the key cannot reach `id`. That is now stated in the javadoc rather than overclaimed.

The javadoc on the join now explains why the predicate is load-bearing rather than redundant beside `lookup_div.id`, and specifically that without it the ordering key offers no usable prefix, so the join reads every row in the workspace to resolve a handful of experiment items — enough on a large install to exhaust the per-query memory cap. Agreed that the commit message is where reasoning goes to die.

Also removed the inline `--` comments I had added in the previous commit: they were being shipped to ClickHouse on every request and cluttering `query_log`. Rationale is in the javadoc now.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
thiagohora: Agreed on every point, and this is now on the record in the javadoc rather than only in review.

Your reasoning is more precise than mine was: the `LEFT JOIN` plus the `if(notEmpty(...))` fallback mean no row drops, so the observable delta is confined to the grouping key for a legacy id pointing into another dataset — which could shift a `COUNT(DISTINCT ...)`. And you are right that `validateDatasetItemsWorkspace` checks only the workspace, so a cross-dataset reference is acceptable at ingestion. The identical-output result is therefore evidence that such references do not exist in the data I sampled, not that they cannot.

The javadoc now says exactly that: the predicate is a behaviour narrowing as well as a pruning one, and the equivalence is empirical rather than structural. I have not cited it as proof of impossibility.

I also share your read that the new behaviour is the more correct one — surfacing a foreign dataset’s item id inside a single-dataset comparison view was already wrong.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
thiagohora: Fixed in 63c090b — documented in the javadoc rather than inline, since the SQL comment would have gone out to ClickHouse on every request.

It states that the two sets serve different consumers (the outer query reads the `item_*` names, the filters resolve the bare ones), that dropping the bare half reintroduces the filter-binding defect and dropping the `item_*` half breaks the response projection, and that both halves must source the `item_*` columns.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -554,10 +606,10 @@ WITH dataset_items_filtered_ids AS (
...
                         evaluators,
                         execution_policy,
-                        created_at,
-                        last_updated_at,
-                        created_by,
-                        last_updated_by
+                        item_created_at AS created_at,
+                        item_last_updated_at AS last_updated_at,
+                        item_created_by AS created_by,
+                        item_last_updated_by AS last_updated_by
```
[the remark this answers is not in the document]
thiagohora: Correct, and deferring it deliberately rather than declining it.

The mutation paths (`DELETE_ITEMS_FROM_VERSION`, `COUNT_ITEMS`) share this root cause and are being split into their own PR, per the discussion on this file above. Two reasons for the split rather than folding it in here:

- `DELETE_ITEMS_FROM_VERSION` is a bare `DELETE FROM` with no projection, so unlike every read path there is nowhere to hang aliases. The fix means resolving ids through an aliased subquery — a materially different change, on a destructive path, wanting its own tests.
- `COUNT_ITEMS` gates that delete, so the preflight count currently agrees with the wrong deletion. Fixing one without the other would make it inconsistently wrong instead of consistently wrong, which is worse to ship for the window between PRs.

Not urgent on the evidence: our query logs show no filtered deletes and no filtered write-path calls over the retention window, so there is nothing suggesting data has already moved.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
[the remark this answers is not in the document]
thiagohora: Fair, and deferring the contract wording as non-urgent.

To be precise about what changed: these fields were *always* meant to be item-level — the response projection has always read `item_created_at AS created_at`. The resolved CTEs were sourcing the snapshot row’s columns under those names, so the endpoint had been returning row metadata by mistake. This restores the documented intent rather than redefining it, which is why I would not describe it as a breaking change.

Agreed the `DatasetItem_Compare` schema description should say so explicitly, and that is worth doing — but as a docs change alongside the other schema descriptions rather than in this PR, which is already spanning five queries. Not blocking, and no client can be relying on the previous values: nothing asserted them, and they were wrong.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetItemVersionDAO.java
```diff
@@ -2399,10 +2464,10 @@ SELECT arrayJoin([id, row_id])
...
                             div_dedup.span_id AS span_id,
                             div_dedup.tags AS tags,
-                            div_dedup.created_at AS created_at,
-                            div_dedup.last_updated_at AS last_updated_at,
-                            div_dedup.created_by AS created_by,
-                            div_dedup.last_updated_by AS last_updated_by,
+                            div_dedup.item_created_at AS created_at,
+                            div_dedup.item_last_updated_at AS last_updated_at,
+                            div_dedup.item_created_by AS created_by,
+                            div_dedup.item_last_updated_by AS last_updated_by,
```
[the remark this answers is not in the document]
thiagohora: Reasonable, and deferring it.

The immediate risk is covered from the other side: `experimentItemsQueries__resolveFiltersAgainstItemLevelColumns` (added in 63c090b) fails if any of the three experiment-items constants stops exposing the filter-visible names from the `item_*` columns, which is the regression that would break those reductions. That runs in milliseconds against a string rather than needing a divergent-metadata fixture through the stats endpoint.

An endpoint-level case asserting `EXPERIMENT_ITEMS_COUNT`, `TRACE_COUNT` and an aggregate under divergent item-vs-row metadata would add genuine value on top of that, and I would take it — but the path it covers has not been exercised by any traffic in our retention window, so it is not urgent enough to hold this PR. Better as part of the stats-endpoint test work than bolted on here.

apps/opik-backend/src/test/java/com/comet/opik/domain/DatasetItemVersionQueryShapeTest.java
```diff
@@ -116,6 +157,93 @@ void truncation__appliesToThePageOnly() {
...
+                "SELECT_DATASET_ITEM_VERSIONS_WITH_EXPERIMENT_ITEMS_STATS");
+
+        assertThat(constants).allSatisfy(name -> {
+            String sql = allQueries.get(name);
+            assertThat(sql).as("%s must be present on the DAO", name).isNotNull();
+            assertThat(sql)
+                    .as("""
+                            %s interpolates the dataset item filters, so every scope they resolve in must                             expose the filter-visible names from the item-level columns. Before this was                             guarded, one scope aliased the row's columns, another omitted them entirely                             (raising UNKNOWN_IDENTIFIER, i.e. a 500), and a third aliased in the opposite                             direction so the same filter name meant different things per template branch.""",
+                            name)
+                    .contains("item_created_at AS created_at")
```
JetoPistola, line 240: 💡 **suggestion** | Testing

These two assertions are a real improvement — reflecting every constant and scoping the pattern to `div_dedup` is better than what I suggested, and the two legitimate aliases it surfaced (`eia.created_at`, `COPY_ITEMS_FROM_LEGACY`) are exactly the kind of finding a broad assertion earns. But there's a hole in this one, and I checked it by mutation rather than by reading.

Each of these constants interpolates the filters in **more than one scope**:

| constant | scopes with `item_created_at AS created_at` |
|---|---|
| `..._COUNT` | 3 |
| `..._STATS` | 2 |
| `..._WITH_EXPERIMENT_ITEMS` | 5 |

`.contains(...)` is satisfied by any single one. So I removed the four aliases from one of the two scopes in `..._STATS` — recreating precisely the omission that raised `UNKNOWN_IDENTIFIER` and that this PR fixed — and **both new tests stayed green**. `aliases__neverMapRowColumnsToItemLevelNames` misses it too, because an omitted projection has no offending alias to match.

That's the same failure mode the PR is about: the bug was never that a constant had zero correct scopes, it was that *some* scopes were right and others weren't. A presence assertion can't distinguish those.

Counting instead of asserting presence would close it — assert each of the four aliases appears exactly as many times as the constant has filter-resolving scopes, or derive the expected count from the number of `<dataset_item_filters>` occurrences so it stays honest when a scope is added. Either keeps the 80 ms property.

The bare-passthrough form is worth a thought too: the original `dataset_items_filtered_ids` defect was `created_at,` with no `AS` at all, which `ROW_COLUMN_ALIASED_TO_ITEM_NAME` cannot match by construction. It happens to be caught today via this test's `.contains`, so a count-based fix covers both.

apps/opik-backend/src/test/java/com/comet/opik/domain/DatasetItemVersionQueryShapeTest.java
```diff
@@ -116,6 +157,93 @@ void truncation__appliesToThePageOnly() {
...
+        // alias it reports the row's timestamp as the item's. It went unnoticed across five constants.
+        var offenders = new ArrayList<String>();
+        allQueries.forEach((name, sql) -> {
+            Matcher m = ROW_COLUMN_ALIASED_TO_ITEM_NAME.matcher(sql);
+            while (m.find()) {
+                offenders.add("%s: %s".formatted(name, m.group()));
+            }
+        });
+
+        assertThat(offenders)
```
JetoPistola, line 219: 🧹 **nit** | Style

Both new `.as(...)` messages lost their `\` line-continuations, so the text blocks keep the indentation as literal characters:

```
each of these aliases a snapshot-row column to a name that is read as item-level.                         Source the item_* column instead.
```

The other messages in this file use the trailing `\` and read correctly on failure. Worth fixing precisely because these fire on a regression nobody is expecting — the assertion text is doing the explaining at that moment, and the run of whitespace mid-sentence undercuts it.

Note spotless won't catch this: it's inside a string literal, so it ships silently.

apps/opik-backend/src/test/java/com/comet/opik/domain/DatasetItemVersionQueryShapeTest.java
```diff
@@ -116,6 +157,93 @@ void truncation__appliesToThePageOnly() {
...
+     * <p>
+     * No lookbehind is needed: {@code _} is a word character, so {@code \bcreated_at} cannot match inside
+     * {@code item_created_at}.
+     */
+    private static final Pattern ROW_COLUMN_ALIASED_TO_ITEM_NAME = Pattern.compile(
+            "div_dedup\\.(created_at|last_updated_at|created_by|last_updated_by)\\s+[Aa][Ss]\\s+"
+                    + "(item_)?(created_at|last_updated_at|created_by|last_updated_by)\\b");
+
+    /** Every SQL constant on the DAO, by field name. */
+    private static Map<String, String> allQueries;
```
JetoPistola, line 201: 🧹 **nit** | Style

`new java.util.LinkedHashMap<>()` is fully qualified while `ArrayList`, `List` and `Map` were added as imports in this same commit.

```suggestion
    private static Map<String, String> allQueries;
```

(with `import java.util.LinkedHashMap;` alongside the others, and the qualified name dropped at the assignment)

JetoPistola:
👋 **Review summary** (re-review of `63c090b`)

All nine findings from the two previous rounds are addressed. This round I only have one substantive item, and it's about the new tests rather than the SQL.

**What looks good**
- `aliases__neverMapRowColumnsToItemLevelNames` is a better test than the one I asked for. Reflecting every constant and scoping the pattern to `div_dedup` catches the whole family rather than the instances we happened to find, and the two legitimate aliases it surfaced — `eia.created_at` on a different table, and `COPY_ITEMS_FROM_LEGACY` where the legacy column really is the authoring time — are the kind of thing a broad assertion earns. The `\b` / `_`-is-a-word-character reasoning in the javadoc is right, and worth having written down.
- You caught something I missed: the inline `--` comments from the previous commit were being shipped to ClickHouse on every request and cluttering `query_log`. I read those as harmless documentation.
- Moving the rationale into javadoc landed in the right place, and the `lookup_div` note now says what the change does *and* doesn't do — two key components rather than a full match, empirically safe rather than structurally guaranteed. That's the version I'd want to find in six months.
- Replacing the specific production figures with the mechanism they illustrate was the right call for code that outlives the incident.

**Overall**
The SQL is in good shape and the guard tests now cover the class of defect rather than its instances. My one finding is that `experimentItemsQueries__resolveFiltersAgainstItemLevelColumns` uses `.contains()` while each constant has two to five filter-resolving scopes — so it passes if any single one survives. I removed the aliases from one of two scopes in `..._STATS`, recreating the `UNKNOWN_IDENTIFIER` 500 this PR fixed, and both new tests stayed green. Counting occurrences instead of asserting presence closes it and keeps the 80 ms property.

Deferring the delete path, the schema-description wording and the stats-endpoint behavioural case all seem right to me, and the reasons given are the reasons I'd give.

**Inline comments**: 1 suggestion, 2 nits — nothing blocking.

JetoPistola (approved):
Non-blocking comments only — approving.

apps/opik-backend/src/test/java/com/comet/opik/domain/DatasetItemVersionQueryShapeTest.java
thiagohora: Fixed in #8042 — thanks, and you were right that this was worth catching rather than shrugging at. Both messages had lost the trailing `\`, and your note that spotless cannot see inside a string literal is exactly why it shipped.

Verified the rendering rather than assuming it, by triggering the assertion:

```
[each of these aliases a snapshot-row column to a name that is read as item-level. Source the item_* column instead.]
Expecting empty but was: ["SELECT_DATASET_ITEM_VERSIONS_WITH_EXPERIMENT_ITEMS: div_dedup.created_at AS item_created_at"]
```

One sentence, and it names the offending constant and alias.

apps/opik-backend/src/test/java/com/comet/opik/domain/DatasetItemVersionQueryShapeTest.java
thiagohora: Fixed in #8042 — `LinkedHashMap` imported alongside the others, qualified name dropped, and the `List` import removed since parameterizing the other test made it unused.

apps/opik-backend/src/test/java/com/comet/opik/domain/DatasetItemVersionQueryShapeTest.java
```diff
@@ -116,6 +157,93 @@ void truncation__appliesToThePageOnly() {
...
+    }
+
+    @Test
+    @DisplayName("the experiment-items queries expose the filter names from the item-level columns")
+    void experimentItemsQueries__resolveFiltersAgainstItemLevelColumns() {
+        List<String> constants = List.of(
+                "SELECT_DATASET_ITEM_VERSIONS_WITH_EXPERIMENT_ITEMS",
+                "SELECT_DATASET_ITEM_VERSIONS_WITH_EXPERIMENT_ITEMS_COUNT",
+                "SELECT_DATASET_ITEM_VERSIONS_WITH_EXPERIMENT_ITEMS_STATS");
+
```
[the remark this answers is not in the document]
thiagohora: Fixed in #8042 — converted to `@ParameterizedTest` with `@ValueSource`, so the class reports 13 tests instead of 11 and a failure names the specific constant rather than one collapsed case.

PR #8042, opened 2026-08-27
[NA] [BE] test: tidy the query-shape assertions flagged in review

thiagohora:
## Details

Follow-up to #8040, addressing the three nits raised after approval. Test-only — no query or production code changes.

- Both new `.as(...)` text blocks had lost their trailing `\` line continuations, so the indentation shipped as literal characters and the failure message read `...is read as item-level.                         Source the item_* column instead.` These assertions fire on a regression nobody is expecting, which is precisely the moment the message is doing the explaining. Spotless cannot catch it — it is inside a string literal.
- `LinkedHashMap` is imported rather than fully qualified, matching the other `java.util` imports added alongside it; the now-unused `List` import is dropped.
- `experimentItemsQueries__resolveFiltersAgainstItemLevelColumns` becomes a `@ParameterizedTest` over `@ValueSource` instead of an `allSatisfy` over a list, so JUnit reports one case per constant rather than collapsing three into one.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Related, not resolved here: OPIK_8109

## AI-WATERMARK

AI-WATERMARK: yes

- If yes:
  - Tools: Claude Code CLI
  - Model(s): Claude Opus 5
  - Scope: Applying review nits
  - Human verification: Rendered failure message confirmed by triggering the assertion; class re-run before and after.

## Testing

`mvn test -Dtest='DatasetItemVersionQueryShapeTest'` — **13 tests, 0 failures**, up from 11 because the parameterized case now reports per constant.

Failure rendering verified rather than assumed. Re-pointing one `item_*` alias at its snapshot-row column produces:

```
[each of these aliases a snapshot-row column to a name that is read as item-level. Source the item_* column instead.]
Expecting empty but was: ["SELECT_DATASET_ITEM_VERSIONS_WITH_EXPERIMENT_ITEMS: div_dedup.created_at AS item_created_at"]
```

One sentence, and it names the offending constant and alias. The DAO was restored afterwards; the diff here is the test file only.

## Documentation

None.

### DIFF
--- a/apps/opik-backend/src/test/java/com/comet/opik/domain/DatasetItemVersionQueryShapeTest.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/domain/DatasetItemVersionQueryShapeTest.java
@@ -4,10 +4,12 @@ import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 
 import java.lang.reflect.Field;
 import java.util.ArrayList;
-import java.util.List;
+import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -49,7 +51,7 @@ class DatasetItemVersionQueryShapeTest {
         countField.setAccessible(true);
         countSql = (String) countField.get(null);
 
-        allQueries = new java.util.LinkedHashMap<>();
+        allQueries = new LinkedHashMap<>();
         for (Field f : dao.getDeclaredFields()) {
             if (f.getType() == String.class) {
                 f.setAccessible(true);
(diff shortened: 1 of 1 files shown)


CometActions:
**No test needed here.**

Test-only: the diff is one Java unit test, DatasetItemVersionQueryShapeTest. It turns the loop over the three SELECT_DATASET_ITEM_VERSIONS_WITH_EXPERIMENT_ITEMS* constants into a @ParameterizedTest, unwraps the text-block failure messages, and hoists the inline java.util.LinkedHashMap to an import. The four `item_* AS ...` assertions are unchanged, and no production code is touched, so there is nothing here for an e2e test to cover.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/33075565238)

JetoPistola:
👋 **Review summary**

**What looks good**
- The text-block `\` continuations are the real fix here — I verified the rendered output by triggering the assertion, and the message now reads as one sentence instead of the run-together `...is read as item-level.                         Source the item_* column instead.` Getting failure text right for an assertion that only fires on an unexpected regression is exactly the kind of care that pays off later.
- The `@ParameterizedTest` conversion is the right call over `allSatisfy`: JUnit now reports `[1]`/`[2]`/`[3]` per constant, so a single broken query names itself instead of hiding inside a collapsed case. Confirmed by breaking one `contains()` — all three cases failed independently and each carried its own constant name.
- Keeping the explicit `@ValueSource` list rather than deriving it from the reflected `allQueries` map preserves the `isNotNull()` check, which catches a renamed or deleted constant. Easy to mistake for duplication; it isn't.
- Scope discipline is exact — test file only, four `item_* AS ...` assertions untouched, no production code.

**Overall**
Clean follow-up that does what it says. I ran the class locally: 13 tests, 0 failures, matching the description. The failure-rendering claim was verified rather than taken on trust, and it holds.

**Inline comments**: None — looks clean!

JetoPistola (approved):
Verified locally: 13 tests, 0 failures. Failure rendering confirmed.

PR #7973, opened 2026-08-24
[OPIK-7834] [BE] feat: redact trace content on read for callers without the original-data permission

YarivHashaiComet:
## Details

Trace content can now be redacted at read time, so what is stored and what a given caller
is allowed to read are decided separately. Content is persisted exactly as the SDK sends it;
responses are rewritten during serialization for callers that do not hold the
`trace_original_data_view` workspace permission. Everything is behind a feature flag that is
off by default — with it off, no serializer is registered, no permission is requested from the
platform, and responses are byte-identical to before.

- Rules are a deployment-level JSON array of `{regex, replace}`, compiled once at startup so a
  malformed pattern stops the deployment rather than silently disabling redaction.
- The decision is resolved once after authentication and carried on the request, because
  serialization happens long after the resource method has returned.
- Streamed responses redact explicitly: they are written on a scheduler thread where the
  writer interceptor's thread-local does not apply. An ArchUnit test fails the build if a new
  `ChunkedOutput` producer appears outside that path.
- Structural properties (`project_name`, `thread_id`, `model`, `sortable_by`, and the `name` of
  entities the API addresses by name) are exempt, so navigation and filtering keep working.

Requires the companion platform change that returns the caller's workspace permissions on the
authentication response, requested only when this feature is enabled:
comet-ml/comet-backend#5650.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-7834

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: implementation, tests, and configuration in this PR, authored iteratively with review
  and direction at each step.
- Human verification: reviewed diff by diff; findings from two code-review passes were applied
  or explicitly deferred with reasons; verified end to end against a local stack with real
  authentication (see Testing).

## Testing

Automated — `cd apps/opik-backend && mvn test -Dtest="..."`:
- `RemoteAuthServiceTest` (57), `AuthCredentialsCacheServiceTest` (14) — auth plumbing,
  credentials cache round-trip, and the request flag on every credential-producing path
  (api key, session cookie, OAuth).
- `RedactionModuleTest` (11), `RedactionRequestFilterTest` (12), `RedactionCoverageArchTest` (1)
  — serializer behaviour including map keys and structural exemptions, path scoping, and
  stream coverage.
- `ReadTimeRedactionResourceTest` (4), `ReadTimeRedactionDisabledResourceTest` (2) —
[...]

PR #8043, opened 2026-08-27
[OPIK-7834] [BE] feat: mask trace content by field name at read time

thiagohora:
## Details

Second iteration of read-time redaction, replacing the regex mechanism that shipped in #7973. Masking moves out of the Jackson serializer and into the DAO row mappers, keyed on configured **leaf field names** rather than patterns, and carried on the reactive context the way `workspaceId` and `visibility` already are.

Net **853 insertions, 2030 deletions** — the mechanism swap removes far more than it adds, and it resolves several review threads from #7973 by construction rather than by argument:

- **No regex, so no ReDoS** and nothing to police. `MatchBudget` and its 199 lines of tests are gone, along with the per-character allowance, the ceiling and the output-growth bound.
- **No exemption list, so no name-collision hazard in either direction.** The structural fields the API addresses by (`thread_id`, `project_name`, `model`, `id`) are separate columns and were never candidates, so there is nothing to exempt and nothing to keep in sync. The `Experiment.promptVersions[].commit` paged-vs-streamed divergence disappears with it.
- **Configuration is a real YAML list** (`redaction.maskFields`) instead of a JSON array smuggled through a scalar.
- **The dataset CSV export gap closes** — it merged as a TODO in #7973 and is now refused alongside attachment download.
- **Free-form SQL is masked rather than refused.** The caller chooses the projection, so a name or a pattern can be walked past (`map('x', input)`, `base64(input)`); masking every *string* in the result cannot be, and it leaves counts and grouping over non-content dimensions working.

Carried forward from #7973 unchanged: permissions resolved from the platform's workspace permissions API per credential type, the `original_data_view` rename, the opt-in permission model, and the session-cookie coverage. The `Streamer` no longer redacts, so the `ProvisionException` narrowing and cause-chain handling it needed are gone with it.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves OPIK-7834

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Opus 5
- Scope: the port onto merged `main` (mechanism swap, runner-job masking, config and test reconciliation), plus the commit messages and this description.
- Human verification: design decisions — the masking mechanism, the opt-in permission model, the permission rename — were directed and approved by the author in review. Test runs below were executed locally and their results read; the platform-side counterpart (comet-ml/comet-backend#5650) is not compile-verified locally, see Testing.

## Testing

`mvn test` in `apps/opik-backend`, local process mode, macOS, Testcontainers for MySQL/ClickHouse/Redis/Zookeeper:

- `TracesResourceTest` — **392 passed**. The important one: every trace and span row mapper signature changed, so this is the regression surface for normal, unmasked reads.
- `ReadTimeRedactionResourceTest` — 12 passed. Covers masked vs stored per credential type (api key **and** session cookie), spans, the streamed/paged parity, the refusal paths, and the two long-poll runner-job cases.
- `ReadTimeRedactionDisabledResourceTest` — 2 passed. The feature off, end to end.
- `RedactionConfigTest`, `FieldMaskerTest`, `RedactionGuardTest`, `RedactionServiceTest`, `RequestContextConstructionArchTest`, `RemoteAuthServiceTest` (58), `Local`/`RemoteWorkspacePermissionsServiceTest` — **95 passed** together.
- `DatasetsResourceIntegrationTest` — 2 passed.

**Coverage boundary, worth reviewing deliberately.** Masking is applied where a content column becomes a `JsonNode`, which means it is applied *where it is wired* rather than everywhere by construction — the opposite of the serializer approach. Currently wired: trace, span, thread, dataset-item, dataset-item-version and experiment-item DAOs, analytics queries, and runner jobs. Two consequences:

1. **Runner jobs needed wiring explicitly** and are included here (second commit). They are read from Redis, not through a DAO, so the DAO move had left their `inputs` and `result` returning stored content — a regression against #7973, which covered them by construction. Masked in the resource, with the masker captured before `nextJob` subscribes so the reactor thread that resumes the long poll never touches the request scope.
2. **Prompt content is not masked**, deliberately, per review — prompts are not expected to carry the content this feature withholds.

A new endpoint or DAO is therefore **unmasked by default**. That is the trade for losing the exemption list, and it wants a convention or an ArchUnit rule if this stops being a PoC.

**Not run:** the full backend suite, and the frontend. Scope was the redaction mechanism, the auth/permissions path it reads from, and the DAOs it touches.

Two fixes to inconsistencies in the pre-port branch, found while reconciling: its `RedactionGuard` javadoc still listed free-form SQL as refused after its own final commit changed that to masking, and the analytics resource ended up carrying both the refusal and the masking.

## Documentation

`config.yml` documents `maskFields` and `replacement` in full: why names rather than paths (the same field sits at different depths depending on the integration — `messages[].content`, `messages[].kwargs.message.content`, `function.arguments`), that the shipped seven names are a starting point and not a guarantee, that keys are never rewritten and therefore sensitive data must live in values, and that switching this on changes what a public project shows. `config-test.yml` carries the keys so the two cannot drift apart, including a comment on why the placeholder list entry exists (Dropwizard's indexed overrides only replace an index that already exists).

## Depends on

The platform counterpart, comet-ml/comet-backend#5650, must deploy first — it defines `original_data_view` and serves it from the workspace permissions API. With `redaction.enabled=false` there is no permission traffic at all, so an older platform runs against this build unchanged. That PR still has the rename open for discussion, which should settle before either side merges.


thiagohora:
## Why the matching is hand-written and not Jayway JsonPath

Recording this up front, because it was raised on #7973 and this PR is where the answer now lives. Short version: JsonPath is available and the instinct behind the suggestion was right, but it does not fit this specific job, for reasons already recorded in this codebase.

`json-path` 3.0.0 is a declared dependency and already used in `OnlineScoringEngine`, `FilterEvaluationServiceBase` and a test util — so there is no dependency-weight argument.

**1. It needs the one construct this codebase bans.** The mechanism is "mask the value of every leaf whose configured field name matches, at any depth", i.e. `$..content`. `VariablePathUtils.findUnsupportedConstructInJsonPath` rejects `RECURSIVE_DESCENT = ".."`, and the comment at its call site gives the reason: *"Recursive descent and filter predicates walk the whole section, and scoring shares a scheduler across workspaces, so the cost of one rule's expression is not confined to that rule."* Masking runs on every content column of every row of every read — a hotter path than online scoring, which is where that guard was thought necessary.

**2. It does not operate on the type we hold.** `OnlineScoringEngine:901`: *"JsonPath didn't work with JsonNode, even explicitly using JacksonJsonProvider, so we convert to a plain Object."* `FieldMasker` rewrites a `JsonNode` in place; going through JsonPath means materialising each content column to `Map`/`List` and back, per row.

In fairness, that note names `JacksonJsonProvider` (the POJO provider) rather than `JacksonJsonNodeJsonProvider`, so it is possible the JsonNode-native provider would have worked and the wrong one was reached for. I have not tested it, and I would not assert "JsonPath cannot touch JsonNode" as fact. Points 1 and 3 hold either way.

**3. Two of the three operations have no JsonPath expression.**

| Operation | Used for | JsonPath equivalent |
|---|---|---|
| `mask` — configured name at any depth | trace/span/thread/dataset/experiment content | `$..name`, i.e. the banned construct |
| `maskEveryString` — mask by shape | Agent Insights, where the caller chooses projection *and* key names | none; the nearest is a regex filter, which is what this PR removes |
| "a configured name masks everything beneath it" | nested objects and arrays under a matched name | none; needs its own traversal |

So the tree walk would not disappear — we would maintain it *and* a path grammar we then have to restrict.

What the suggestion did get right is the level above the tool: **declare what to mask rather than enumerate what to exempt.** This PR makes exactly that change, and it is what removes the regex, `MatchBudget`, the exemption list, and the paged-versus-streamed divergence in one go. What remains is a 120-line walk with no configuration surface of its own.

Happy to be pushed back on — particularly on the provider caveat, which is testable if someone thinks it is worth the measurement.

PR #8050, opened 2026-08-28
[OPIK-7315] give verify.sh a receive-timeout, and stop the four call sites drifting

andrescrz:
## Details

`verify.sh` had no way to raise the ClickHouse client socket timeout, and a long fidelity compare was lost to it. On `ok=0` the driver re-checks the differing keys on the sorting key, to separate a real mismatch from a superseded-version artifact. That re-check can stall for longer than ClickHouse's 300s `receive_timeout` default even where the window compare did not — and because it runs before the week's verdict is known, the stall aborts the **entire** compare at the first mismatching week, discarding every window still unchecked.

- **`--receive-timeout N`, default `1800`.** A named flag, matching how `backfill.sh` exposes `--max-insert-block-size` and friends, rather than a generic settings passthrough. The default is opinionated on purpose: 300 is wrong for this workload, so an operator who does not know to pass the flag still gets a compare that finishes.
- **The connection prefix was duplicated across four `clickhouse-client` invocations** (`ch`, compare, drill-down, confirm-keys), now factored into one `CH_ARGS` array. Without that the flag would reach only the sites someone remembered to edit, and the site that needs it is the confirm-keys re-check rather than the window compare.

What is measured, and worth keeping straight: `receive_timeout` bounds the **gap between packets**, not total query time. A window compare running well past 300s does not trip it; the re-check can. The cost of a generous default is that a genuinely dead connection takes that long to surface, which for a read-only, resumable compare is the better trade than losing a long run to a transient stall.

**Follow-up, deliberately not here:** `rollback.sh` has the same exposure across its own client invocations, and the sentinel repair's `mutations_sync = 2` blocks for as long as the mutation runs — the same shape as OPIK_8141, where the client gave up and the server carried on. It is excluded because #8046 is open and approved against `rollback.sh`, and editing it now would conflict. It follows once that lands.

## Change checklist
- [ ] User facing
- [x] Documentation update

## Issues

- Resolves #
- OPIK-7315

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: the `verify.sh` flag and `CH_ARGS` refactor, and the README corrections.
- Human verification: reviewed by the author; the runs below were executed against a live multi-replica ClickHouse.

## Testing

Run against a live ClickHouse (26.3), not a container.

**The flag reaches the client** — forced low, the failure is the same error the stock default produced, with the value substituted:

```
$ ./verify.sh --database <db> --host <host> --receive-timeout 1 --from-week 0 --to-week 0
Error on processing query: Timeout exceeded while receiving data from server.
Waited for 1 seconds, timeout is 1 seconds.
```

**No regression on the normal path** — one bounded window, default timeout, `PASSED: all 1 windows match (sample 1/1)`, with a row count identical to the same window in an earlier full run.

**Validation** rejects `0`, `abc`, `-5` and `1 ; rm`, one message each. `bash -n` clean.

**Not run:** a full multi-week compare on this build. A compare is currently walking the remaining weeks on the pre-change revision with the timeout supplied externally, exercising the same code path by the same mechanism.

## Documentation

`data-migrations/traces-local-v2-cutover/README.md`:

- the "Detach it" admonition claimed *"an interrupted run costs nothing"* — true of the estate, false of the wall clock. It now points at `--from-week`, and states the resume rule correctly: the last reported week plus the **stride**, with the same stride passed again, since the loop advances by `--weeks-stride` and a bare +1 would sample different windows.
- nothing warned that a mismatching week costs a second, slower query. It now does, with the timeout default named.
- the trace-delete pruning note carried a part count and an on-disk size measured in a real environment. Removed; the engineering point does not depend on the figures.

### DIFF
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
@@ -1402,8 +1402,9 @@ CLICKHOUSE_HOST=<host> CLICKHOUSE_PASSWORD=<pw> ./scripts/verify.sh --database o
 > total, so run it under `nohup`/`screen` rather than an interactive shell that may be interrupted. It is
 > read-only and idempotent, so an interruption cannot damage anything. It does **not** follow that a re-run covers the
 > same windows: both bounds are read live from the old-schema table — `toMonday(min(created_at))` for the anchor and
-> `max(created_at)` for the last week — so on a table still taking writes, or one retention is pruning, the offsets move
-> under you.
+> `max(created_at)` for the last week — so on a table still taking writes, or one that retention is pruning, the offsets
+> move under you. Both are read once, at startup: rows written after that are outside the horizon the run computed,
+> and are the delta's business rather than the compare's.
 >
 > **Resume with `--from-week`; do not restart from 0.** Idempotent does not mean free: a restart repeats
 > every window already compared. Each window either reports a line or has not run, so the resume point is
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
@@ -1400,15 +1400,23 @@ CLICKHOUSE_HOST=<host> CLICKHOUSE_PASSWORD=<pw> ./scripts/verify.sh --database o
 > **Detach it, and expect tens of minutes.** The bounded compare walks one window per week over both
 > tables. On a large table that is minutes per window on the busy weeks and well over half an hour in
 > total, so run it under `nohup`/`screen` rather than an interactive shell that may be interrupted. It is
-> read-only and idempotent, so an interruption cannot damage anything, and because the bound is fixed by
-> `cutover_start` it covers the same windows whenever it is re-run.
+> read-only and idempotent, so an interruption cannot damage anything. It does **not** follow that a re-run covers the
+> same windows: both bounds are read live from the old-schema table — `toMonday(min(created_at))` for the anchor and
+> `max(created_at)` for the last week — so on a table still taking writes, or one retention is pruning, the offsets move
+> under you.
 >
 > **Resume with `--from-week`; do not restart from 0.** Idempotent does not mean free: a restart repeats
 > every window already compared. Each window either reports a line or has not run, so the resume point is
 > the last reported week plus the **stride** — plus one only at the default `--weeks-stride 1` — and the same
 > stride must be passed again, or the resumed run samples different windows than the run it continues. The
-> offsets are anchored on `toMonday(min(created_at))` of the old-schema table, so confirm a resumed run's
-> first window is the one you expect before treating its output as continuous with an earlier log.
+> offsets are anchored as above, so confirm a resumed run's first window is the one you expect before treating its
+> output as continuous with an earlier log — if the anchor has moved, the logs describe different windows and must not be
+> read as one run.
+>
+> **A resumed run does not by itself discharge the pre-`EXCHANGE` gate.** That gate (see the exit checklist) requires a
+> full compare with no narrowing, and each `PASSED` line reports only the windows *that run* compared. Split runs satisfy
+> it only if their windows together cover every week **and** the anchor held throughout; if either is in doubt, re-run
+> whole. Resuming is for the long post-rollback compare, where the bound is deliberately partial anyway.
 >
 > **A mismatching week costs a second, slower query.** On `ok=0` the driver re-checks the differing keys on
 > the sorting key to separate a real mismatch from a superseded-version artifact. That re-check can stall for
@@ -1473,6 +1481,8 @@ bound is still worth passing.
 infeasible, sample and still get high confidence:
 - `--sample-mod N` compares a deterministic 1/N `id` sample — the *same* rows on both sides, so like-for-like.
 - `--weeks-stride S` compares every S-th weekly partition (partition-pruned, so genuinely cheaper).
+- `--receive-timeout N` raises the client's per-packet wait (default 1800, against ClickHouse's 300). The
+  post-mismatch confirm-keys re-check can stall past the stock value and abort the compare at the first mismatch.
 - `--from-week` / `--to-week` bound the range by **0-based week offset** (integers from the anchor Monday, not dates;
   `--to-week` is inclusive) — e.g. verify the most recent weeks fully, older weeks sampled.
 
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
@@ -1413,10 +1413,11 @@ CLICKHOUSE_HOST=<host> CLICKHOUSE_PASSWORD=<pw> ./scripts/verify.sh --database o
 > output as continuous with an earlier log — if the anchor has moved, the logs describe different windows and must not be
 > read as one run.
 >
-> **A resumed run does not by itself discharge the pre-`EXCHANGE` gate.** That gate (see the exit checklist) requires a
-> full compare with no narrowing, and each `PASSED` line reports only the windows *that run* compared. Split runs satisfy
-> it only if their windows together cover every week **and** the anchor held throughout; if either is in doubt, re-run
-> whole. Resuming is for the long post-rollback compare, where the bound is deliberately partial anyway.
+> **Never resume the pre-`EXCHANGE` gate run — restart it.** That gate (see the exit checklist) requires one full compare
+> with no narrowing, and it is the last backstop before an irreversible step. Two runs whose windows happen to add up are
+> only equivalent if the anchor held throughout, which is not something anyone can confirm under pressure; a `PASSED`
+> line now states the range it covered, so a stitched-together pass is visible rather than arguable. Resume is for the
+> exploratory compares and the long post-rollback one, where the bound is deliberately partial anyway.
 >
 > **A mismatching week costs a second, slower query.** On `ok=0` the driver re-checks the differing keys on
 > the sorting key to separate a real mismatch from a superseded-version artifact. That re-check can stall for
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
@@ -466,8 +466,8 @@ the `traceColumnsNonNullable` flip").
 On rollback, after swapping the Nullable original back, revert the flag to `false` **and** run that repair.
 
 **Trace-delete partition pruning needs no flip at all (OPIK-6901).** A trace `DELETE` binds itself to the weekly
-partitions its own ids resolve to instead of being planned against every part of the table — on prod-test, 12 ids
-rewrote 3,928 parts / 5.40 TiB before this. There is **no flag, no ordering constraint, and nothing to revert on
+partitions its own ids resolve to instead of being planned against every part of the table, so a handful of ids no
+longer rewrites the whole table. There is **no flag, no ordering constraint, and nothing to revert on
 rollback**: the predicate is emitted unconditionally, and one rendered statement is correct against the original and
 against the successor alike.
 
@@ -1403,17 +1403,18 @@ CLICKHOUSE_HOST=<host> CLICKHOUSE_PASSWORD=<pw> ./scripts/verify.sh --database o
 > read-only and idempotent, so an interruption cannot damage anything, and because the bound is fixed by
 > `cutover_start` it covers the same windows whenever it is re-run.
 >
-> **Resume with `--from-week`; do not restart from 0.** Idempotent does not mean free: on a large table a
-> restart repeats hours of scanning. Every window either reports a line or did not run, so the resume point
-> is the last reported week plus one. The offsets are anchored on `toMonday(min(created_at))` of the
-> old-schema table, so confirm a resumed run's first window is the one you expect before treating its output
-> as continuous with an earlier log.
+> **Resume with `--from-week`; do not restart from 0.** Idempotent does not mean free: a restart repeats
+> every window already compared. Each window either reports a line or has not run, so the resume point is
+> the last reported week plus the **stride** — plus one only at the default `--weeks-stride 1` — and the same
+> stride must be passed again, or the resumed run samples different windows than the run it continues. The
+> offsets are anchored on `toMonday(min(created_at))` of the old-schema table, so confirm a resumed run's
+> first window is the one you expect before treating its output as continuous with an earlier log.
 >
-> **A mismatching week costs a second, quieter query.** On `ok=0` the driver re-checks the differing keys on
-> the sorting key to separate a real mismatch from a superseded-version artifact, and that query builds its
-> candidate set before emitting anything. On a multi-million-row week the silence can exceed ClickHouse's
-> 300s `receive_timeout` default, which aborts the whole compare at the first mismatching week. `verify.sh`
-> therefore defaults to `1800`; raise it further with `--receive-timeout` if a window still trips it.
+> **A mismatching week costs a second, slower query.** On `ok=0` the driver re-checks the differing keys on
+> the sorting key to separate a real mismatch from a superseded-version artifact. That re-check can stall for
+> longer than ClickHouse's 300s `receive_timeout` default even where the window compare did not, which aborts
+> the whole compare at the first mismatching week. `verify.sh` therefore defaults to `1800`; raise it with
+> `--receive-timeout` if a window still trips it.
 >
 > **Two steps need privileges the rollback grant set does not give.** Plan for them before the window,
 > because both surface at the end when the pressure is highest:
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
@@ -1400,8 +1400,20 @@ CLICKHOUSE_HOST=<host> CLICKHOUSE_PASSWORD=<pw> ./scripts/verify.sh --database o
 > **Detach it, and expect tens of minutes.** The bounded compare walks one window per week over both
 > tables. On a large table that is minutes per window on the busy weeks and well over half an hour in
 > total, so run it under `nohup`/`screen` rather than an interactive shell that may be interrupted. It is
-> read-only and idempotent: an interrupted run costs nothing, and because the bound is fixed by
+> read-only and idempotent, so an interruption cannot damage anything, and because the bound is fixed by
 > `cutover_start` it covers the same windows whenever it is re-run.
+>
+> **Resume with `--from-week`; do not restart from 0.** Idempotent does not mean free: on a large table a
+> restart repeats hours of scanning. Every window either reports a line or did not run, so the resume point
+> is the last reported week plus one. The offsets are anchored on `toMonday(min(created_at))` of the
+> old-schema table, so confirm a resumed run's first window is the one you expect before treating its output
+> as continuous with an earlier log.
+>
+> **A mismatching week costs a second, quieter query.** On `ok=0` the driver re-checks the differing keys on
+> the sorting key to separate a real mismatch from a superseded-version artifact, and that query builds its
+> candidate set before emitting anything. On a multi-million-row week the silence can exceed ClickHouse's
+> 300s `receive_timeout` default, which aborts the whole compare at the first mismatching week. `verify.sh`
+> therefore defaults to `1800`; raise it further with `--receive-timeout` if a window still trips it.
 >
 > **Two steps need privileges the rollback grant set does not give.** Plan for them before the window,
 > because both surface at the end when the pressure is highest:
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/verify.sh
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/verify.sh
@@ -249,9 +249,9 @@ if [[ "$checked" == "0" ]]; then
     exit 1
 fi
 if [[ "$artifacts" != "0" ]]; then
-    log "PASSED: all $checked windows match (sample 1/$SAMPLE_MOD); $artifacts window(s) held a superseded-version"
+    log "PASSED: all $checked windows match (weeks [$FROM_WEEK..$TO_WEEK] stride $WEEKS_STRIDE, sample 1/$SAMPLE_MOD); $artifacts window(s) held a superseded-version"
     log "        artifact only — a key written more than once lands its stale version in an earlier created_at week on"
     log "        one side. Live data is identical on both sides; nothing to fix (see the confirm-keys block)."
 else
-    log "PASSED: all $checked windows match (sample 1/$SAMPLE_MOD)."
+    log "PASSED: all $checked windows match (weeks [$FROM_WEEK..$TO_WEEK] stride $WEEKS_STRIDE, sample 1/$SAMPLE_MOD)."
 fi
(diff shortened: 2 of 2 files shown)


CometActions:
**No test needed here.**

This is operator tooling, not product surface. `scripts/verify.sh` lives in the traces-local-v2 cutover runbook and is run by hand against ClickHouse during a migration window — grepping the repo, nothing in opik-backend, docker-compose or CI invokes it, so no request path, page or SDK call changes when it does. Both halves land in the same place: the new `--receive-timeout` (default 1800) and the `CH_ARGS` consolidation are observable only to whoever is watching the compare log, and the e2e estate has no cutover to run it against. Nothing for a permanent test here.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/33160846720)

_Re-checked after a push on 28 Aug 09:49 UTC — nothing the verdict depends on changed._

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
```diff
@@ -1400,9 +1400,21 @@ CLICKHOUSE_HOST=<host> CLICKHOUSE_PASSWORD=<pw> ./scripts/verify.sh --database o
...
 > tables. On a large table that is minutes per window on the busy weeks and well over half an hour in
 > total, so run it under `nohup`/`screen` rather than an interactive shell that may be interrupted. It is
-> read-only and idempotent: an interrupted run costs nothing, and because the bound is fixed by
+> read-only and idempotent, so an interruption cannot damage anything, and because the bound is fixed by
 > `cutover_start` it covers the same windows whenever it is re-run.
 >
+> **Resume with `--from-week`; do not restart from 0.** Idempotent does not mean free: on a large table a
+> restart repeats hours of scanning. Every window either reports a line or did not run, so the resume point
+> is the last reported week plus one. The offsets are anchored on `toMonday(min(created_at))` of the
+> old-schema table, so confirm a resumed run's first window is the one you expect before treating its output
```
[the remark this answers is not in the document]
andrescrz: Applying — you are right, and I verified it rather than taking it on trust:

```bash
for (( week=FROM_WEEK; week<=TO_WEEK; week+=WEEKS_STRIDE ))
```

So at `--weeks-stride 3` from 0 the run reports 0, 3, 6, 9. Dying after 6 and resuming at **7** does two wrong things at once: window 9 is never compared, and 7/10/13 are compared instead — a different sample from the run it claims to continue. My "+1" was only correct at the default stride.

Corrected to: the last reported week plus the **stride**, plus one only at `--weeks-stride 1`, and the same stride must be passed again or the resumed run samples different windows. Tense fixed to "has not run" too.

Worth noting what this does *not* change: at stride > 1 the compare is a sample by construction, so a resumed run is only equivalent to an uninterrupted one if the stride is preserved. That is now the explicit instruction rather than something a reader has to infer.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/verify.sh
```diff
@@ -88,12 +99,20 @@ done
...
+# One place for the connection and client-side options, so the four call sites below cannot drift — in particular so
+# --receive-timeout applies to the confirm-keys re-check and the drill-down, not only to the window compare.
+CH_ARGS=()
+[[ -z "$CH_HOST" ]] || CH_ARGS+=(--host "$CH_HOST")
+[[ -z "$CH_PORT" ]] || CH_ARGS+=(--port "$CH_PORT")
+CH_ARGS+=(--database "$DATABASE" --log_comment 'traces_local_v2_cutover:verify' --receive_timeout="$RECEIVE_TIMEOUT")
+
 ch() {
-    clickhouse-client ${CH_HOST:+--host $CH_HOST} ${CH_PORT:+--port $CH_PORT} --database "$DATABASE" --log_comment 'traces_local_v2_cutover:verify' --query "$1"
+    clickhouse-client "${CH_ARGS[@]}" --query "$1"
```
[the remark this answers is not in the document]
andrescrz: Not applying, though the gap you are pointing at is real and bigger than this PR — so let me separate the two.

**On the cited requirement.** I read both files rather than assume. `apps/opik-backend/AGENTS.md` does say *"Add/adjust tests for changed behavior before PR"*, but the two lines immediately above it scope the mechanism: *"Frameworks: JUnit, Mockito, and Testcontainers under Maven test lifecycle"* and *"Test naming: `*Test.java` in `src/test/java`"*. There is no shell-test harness anywhere in the repo — no bats, no shellspec, no `*_test.sh` — so a fake-`clickhouse-client` test would mean introducing a new testing pattern, for one flag, in one of six drivers. None of `backfill.sh`, `delta_replay.sh`, `exchange_and_wrap.sh`, `rollback.sh` or `finalize.sh` has shell coverage today, and the class Javadoc on `TracesLocalV2CutoverTest` states that boundary explicitly, pointing at the staging rehearsal for the drivers.

**On the specific proposal.** A stub asserts that bash assembled an argv, which is the low-defect half of this change. What actually broke in practice was not argv assembly but the *value*, and a stub cannot observe a real client honouring it. The check that does is in the PR body: forced to `1`, the client fails with `Waited for 1 seconds, timeout is 1 seconds` — the same error the stock default produced with `300`. That is a real boundary test against a real server; it is simply not in CI.

**Where I think you are right.** Two timeout defects have now been found in these drivers and the Java gate found neither, because it deliberately reimplements the SQL inline and never executes the shell. That is a genuine hole. The fix I would argue for is not a stub but **driver-level integration coverage against the ClickHouse container the suite already starts** — running the real `verify.sh` end to end, which would have caught both. That is a larger, separate piece of work than a timeout passthrough, and doing it properly across all six drivers beats doing it with a fake for one flag. Happy for it to be its own ticket if you want it tracked.

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
```diff
@@ -1400,9 +1400,22 @@ CLICKHOUSE_HOST=<host> CLICKHOUSE_PASSWORD=<pw> ./scripts/verify.sh --database o
 > **Detach it, and expect tens of minutes.** The bounded compare walks one window per week over both
 > tables. On a large table that is minutes per window on the busy weeks and well over half an hour in
 > total, so run it under `nohup`/`screen` rather than an interactive shell that may be interrupted. It is
-> read-only and idempotent: an interrupted run costs nothing, and because the bound is fixed by
+> read-only and idempotent, so an interruption cannot damage anything, and because the bound is fixed by
 > `cutover_start` it covers the same windows whenever it is re-run.
```
thiagohora: **`cutover_start` is not the bound — and the paragraph six lines below says so.**

`cutover_start` appears nowhere in `verify.sh` or `000005_verify_migration.sql` (grep: 0 hits). The anchor is `toMonday(min(created_at))` and the horizon is `addWeeks(toMonday(max(created_at)), 1)` (`scripts/verify.sh:167-169`) — both read live from the old-schema table on every run. Line 1410 of this same admonition states the real anchor, so as written the two contradict each other.

The claim that it "covers the same windows whenever it is re-run" is what makes this worth fixing rather than just a naming nit, because it is the premise the new resume guidance rests on:

- pre-`EXCHANGE` `traces` is still taking writes, so `max(created_at)` advances and `LAST_WEEK` grows between runs;
- if TTL/retention or a delete prunes the oldest rows, `min(created_at)` moves forward and **every** week offset shifts — so `--from-week N` on the resumed run covers a later calendar window than week N of the run it claims to continue. Data is skipped while the two logs read as continuous.

Suggest naming the real bound here, which also makes the "confirm a resumed run's first window is the one you expect" advice at 1411 land as a consequence rather than an unexplained precaution.

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
```diff
@@ -1400,9 +1400,22 @@ CLICKHOUSE_HOST=<host> CLICKHOUSE_PASSWORD=<pw> ./scripts/verify.sh --database o
...
 >
+> **Resume with `--from-week`; do not restart from 0.** Idempotent does not mean free: a restart repeats
+> every window already compared. Each window either reports a line or has not run, so the resume point is
+> the last reported week plus the **stride** — plus one only at the default `--weeks-stride 1` — and the same
+> stride must be passed again, or the resumed run samples different windows than the run it continues. The
+> offsets are anchored on `toMonday(min(created_at))` of the old-schema table, so confirm a resumed run's
+> first window is the one you expect before treating its output as continuous with an earlier log.
```
thiagohora: **This resume guidance sits in the section covering the pre-`EXCHANGE` gate run, which the go/no-go checklist forbids narrowing.**

The checklist at README:1599 says that gate "MUST be a **full compare** (`--sample-mod 1 --weeks-stride 1`, no `--from-week`/`--to-week` narrowing)" — precisely because narrowing can hash out a single-row deletion leak. Nothing here reconciles the two, so an operator reading top-down hits the resume advice first.

What makes that hard to catch after the fact is `scripts/verify.sh:256`, whose success line is:

```
PASSED: all $checked windows match (sample 1/$SAMPLE_MOD)
```

It echoes the sample divisor but **not** the week range, which is printed only once at the top (line 196). So a resumed run tails with `PASSED: all 12 windows match (sample 1/1)` — textually identical to a full pass — and can be attached as gate evidence for weeks that were only ever checked by an earlier aborted run.

Two fixes, ideally both:
- have this admonition explicitly exclude the gate run from resume (resume is for the earlier exploratory compares);
- have line 256 echo `weeks [$FROM_WEEK..$TO_WEEK] stride $WEEKS_STRIDE`, so the evidence line is self-describing.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/verify.sh
```diff
@@ -88,12 +97,20 @@ done
...
 [[ -z "$TO_WEEK" || "$TO_WEEK" == last-sealed || "$TO_WEEK" =~ ^[0-9]+$ ]] \
     || { echo "ERROR: --to-week must be a non-negative integer or 'last-sealed'." >&2; exit 2; }
 [[ -f "$VERIFY_SQL" ]] || { echo "ERROR: cannot find verify SQL at $VERIFY_SQL" >&2; exit 2; }
 
+# One place for the connection and client-side options, so the four call sites below cannot drift — in particular so
+# --receive-timeout applies to the confirm-keys re-check and the drill-down, not only to the window compare.
+CH_ARGS=()
+[[ -z "$CH_HOST" ]] || CH_ARGS+=(--host "$CH_HOST")
+[[ -z "$CH_PORT" ]] || CH_ARGS+=(--port "$CH_PORT")
+CH_ARGS+=(--database "$DATABASE" --log_comment 'traces_local_v2_cutover:verify' --receive_timeout="$RECEIVE_TIMEOUT")
```
thiagohora, line 110: The `CH_ARGS` factoring is a clear improvement — building it after validation and before any use keeps it non-empty, so `"${CH_ARGS[@]}"` is safe under `set -u`, and the array form is strictly safer than the unquoted `${CH_HOST:+--host $CH_HOST}` it replaces. All four call sites are converted, none missed.

**One scoping note:** this fix is `verify.sh`-only, and the PR body names just `rollback.sh` as follow-up. The same duplicated prefix — and the same stock 300s `receive_timeout` — is still inline in `estimate.sh:86,135`, `backfill.sh:311,433`, `finalize.sh:69` and `delta_replay.sh`.

That matters beyond tidiness for at least two of them: `finalize.sh`'s sentinel count is a documented full-table scan, and `estimate.sh:135` runs a `--time`d probe. Both can hit exactly the inter-packet stall this PR exists to fix, with no flag to raise it. Worth widening the follow-up note to all five drivers.

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
thiagohora, line 1: **Nit (outside the diff, README ~line 1474):** the "Feasibility at scale" bullet list is this README's enumeration of `verify.sh`'s options — `--sample-mod`, `--weeks-stride`, `--from-week`/`--to-week` — and wasn't extended with `--receive-timeout`.

The new flag is documented only in the script header and in the new admonition ~60 lines earlier, so an operator scanning that list for a way to handle a long-running compare won't find the flag that exists for exactly that.

thiagohora (commented):
Reviewed the `--receive-timeout` flag, the `CH_ARGS` factoring, and the README delta. **The shell changes are sound** — `CH_ARGS` is built after validation and before any use so it's never empty (`"${CH_ARGS[@]}"` is safe under `set -u`), the array form is strictly safer than the unquoted `${CH_HOST:+--host $CH_HOST}` it replaces, `--receive-timeout` is validated with the same integer shape as the other numeric flags, all four call sites are converted with none missed, and the new setting is in the same class as the already-required `log_comment`, so it introduces no new `readonly = 1` regression.

My findings are all in the prose, and two of them undercut the resume workflow this PR is adding:

1. **README:1403-1404** still says the compare's bound is "fixed by `cutover_start`" and therefore covers the same windows on every re-run. `cutover_start` doesn't exist in the scripts — the bounds are `toMonday(min(created_at))` and `addWeeks(toMonday(max(created_at)), 1)`, both read live from a table still taking writes. Line 1410 of the same admonition states the real anchor, so the block contradicts itself. If retention prunes the oldest rows, every week offset shifts and `--from-week N` covers a later window than week N of the run it claims to continue.

2. **README:1406-1411** places the resume guidance in the section covering the pre-`EXCHANGE` gate run, which the checklist at line 1599 requires be a full compare with no narrowing. Compounding it, `verify.sh:256` prints the sample divisor but not the week range, so a resumed run's `PASSED` line is textually indistinguishable from a full pass.

The remaining two are minor: `--receive-timeout` never made it into the README's option enumeration, and the timeout fix is scoped to `verify.sh` while four sibling drivers still run under the stock 300s `receive_timeout` — `finalize.sh`'s full-table sentinel scan among them.

andrescrz:
All four correct. Three applied, the fourth deferred with a correction to its example.

**1. The `cutover_start` claim — applied.** Verified: `cutover_start` appears **zero** times in `verify.sh`. The bounds are `toMonday(min(created_at))` (line 150) and `max(created_at)`, both read live from the old-schema table. Worse than a stale sentence: I added the correct anchor six lines below without removing the wrong claim, so the block contradicted itself and the resume workflow this PR adds rested on the wrong half. Now:

> It is read-only and idempotent, so an interruption cannot damage anything. It does **not** follow that a re-run covers the same windows: both bounds are read live from the old-schema table … so on a table still taking writes, or one retention is pruning, the offsets move under you.

**2. Resume vs the pre-`EXCHANGE` gate — applied, both halves.** The guidance now says a resumed run does not by itself discharge that gate: split runs satisfy it only if their windows together cover every week **and** the anchor held, and resuming is for the long post-rollback compare where the bound is deliberately partial anyway.

Your second half is what makes the first checkable rather than a matter of trust, so I fixed the output too — a `PASSED` line now states what it covered:

```
PASSED: all 2 windows match (weeks [0..1] stride 1, sample 1/1).
```

A narrowed or resumed pass can no longer be pasted as evidence of a full one.

**3. Missing from the option list — applied.** Added beside `--sample-mod` / `--weeks-stride` at README:1467.

**4. Sibling drivers — deferring, with a correction.** Agreed the exposure is real and worth its own change; not folding four more drivers into a PR about one. Two notes:

- `rollback.sh` already has `--receive-timeout`, added in #8046 — its sentinel repair blocks on `mutations_sync = 2`, so it was the urgent one.
- **`finalize.sh` is the least exposed of the remainder, not the most.** Its heaviest query is `SELECT max(c) FROM (SELECT count() AS c FROM clusterAllReplicas(…) GROUP BY hostName())` — a `count()` with no predicate, which MergeTree answers from part metadata rather than by reading rows, so it returns in well under the stock 300s regardless of table size. There is no sentinel scan in that driver. The genuinely exposed pair is **`backfill.sh` and `delta_replay.sh`**, whose per-week `INSERT SELECT`s are long-running by design; `exchange_and_wrap.sh`'s replication-settle gate is worth a look, and `estimate.sh` reads sizes only.

Happy to send that as its own PR covering `backfill.sh`, `delta_replay.sh` and `exchange_and_wrap.sh` — where it also wants care, since `backfill.sh` already has pacing flags the timeout interacts with, which is the other reason it does not belong here.

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
andrescrz: **Applied.** Confirmed independently: `cutover_start` has 0 hits in `verify.sh`.

You are right that it is not a naming nit. I added the correct anchor six lines below without removing the wrong claim, so the block contradicted itself — and the false half was the premise the new resume guidance rested on. Now:

> It is read-only and idempotent, so an interruption cannot damage anything. It does **not** follow that a re-run covers the same windows: both bounds are read live from the old-schema table — `toMonday(min(created_at))` for the anchor and `max(created_at)` for the last week — so on a table still taking writes, or one retention is pruning, the offsets move under you.

Naming the real bound here does what you predicted: the "confirm a resumed run's first window is the one you expect" advice below now reads as the consequence rather than an unexplained precaution.

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
andrescrz: **Applied — both fixes, and the first one stricter than I first wrote it.**

`verify.sh` line 256 now echoes the range, so an evidence line is self-describing:

```
PASSED: all 2 windows match (weeks [0..1] stride 1, sample 1/1).
```

On the admonition: my first attempt made it conditional — split runs count if their windows cover every week *and* the anchor held. That is true, and it was the wrong rule. It is unverifiable in the moment, and this gate is the last backstop before an irreversible step, so the cost of being wrong is asymmetric. It is now flat, as you asked:

> **Never resume the pre-`EXCHANGE` gate run — restart it.** … Two runs whose windows happen to add up are only equivalent if the anchor held throughout, which is not something anyone can confirm under pressure; a `PASSED` line now states the range it covered, so a stitched-together pass is visible rather than arguable.

The two fixes reinforce each other: the rule is simple, and the output makes a violation visible rather than a matter of trust.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/verify.sh
andrescrz: **Widening the follow-up note as you ask; not folding the drivers into this PR.** Two corrections to the exposure ranking first, since they change which ones matter.

**`finalize.sh` has no sentinel count** — 0 hits for `1970-01-01`, `isNaN` or `sentinel`. The README note you are thinking of is at line 1429, inside the *rollback* admonition, and refers to `rollback.sh`'s `sentinel_counts`, not this driver. `finalize.sh`'s heaviest query is:

```sql
SELECT max(c) FROM (SELECT count() AS c FROM clusterAllReplicas(…) GROUP BY hostName())
```

A predicate-free `count()`, which MergeTree answers from part metadata rather than by reading rows — so it returns well inside 300s at any table size. It is the *least* exposed of the five, not the most.

**`estimate.sh:135` is exposed, and you are right about it.** It reads real rows:

```bash
clickhouse-client … --time --query "SELECT * FROM traces LIMIT $PROBE_ROWS FORMAT Null"
```

So the corrected ranking for the follow-up: **`backfill.sh` and `delta_replay.sh`** (per-week `INSERT SELECT`s, long-running by design), then **`estimate.sh`** (that probe); `exchange_and_wrap.sh` is metadata DDL plus a settle gate; `finalize.sh` is the light one.

Also worth recording: **`rollback.sh` already has the flag**, added in #8046 — its sentinel repair blocks on `mutations_sync = 2`, so it was the urgent case and is done.

Keeping the rest out of this PR for a reason beyond scope: `backfill.sh` already carries pacing flags (`--pause-seconds`, the block-size settings) that a client timeout interacts with, so it wants its own review rather than a mechanical copy of this one. Happy to send it as a separate PR.

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
andrescrz: **Applied.** Added to the "Feasibility at scale" list beside `--sample-mod` and `--weeks-stride`:

> - `--receive-timeout N` raises the client's per-packet wait (default 1800, against ClickHouse's 300). The post-mismatch confirm-keys re-check can stall past the stock value and abort the compare at the first mismatch.

Your framing is the reason it belongs there rather than only in the script header: that list is where someone goes when a compare is too slow or too long, which is exactly the problem this flag solves.

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
```diff
@@ -1400,8 +1400,29 @@ CLICKHOUSE_HOST=<host> CLICKHOUSE_PASSWORD=<pw> ./scripts/verify.sh --database o
 > **Detach it, and expect tens of minutes.** The bounded compare walks one window per week over both
 > tables. On a large table that is minutes per window on the busy weeks and well over half an hour in
 > total, so run it under `nohup`/`screen` rather than an interactive shell that may be interrupted. It is
-> read-only and idempotent: an interrupted run costs nothing, and because the bound is fixed by
-> `cutover_start` it covers the same windows whenever it is re-run.
+> read-only and idempotent, so an interruption cannot damage anything. It does **not** follow that a re-run covers the
+> same windows: both bounds are read live from the old-schema table — `toMonday(min(created_at))` for the anchor and
+> `max(created_at)` for the last week — so on a table still taking writes, or one retention is pruning, the offsets move
```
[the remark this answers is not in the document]
andrescrz: Applied. It was an elliptical relative clause with the pronoun dropped — "or one [that] retention is pruning" — which parses but reads as a stumble. Now "or one that retention is pruning", which keeps the subject the table rather than the timing (your second suggestion, "while a retention policy is pruning rows", shifts it to when rather than which).

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
```diff
@@ -1400,8 +1400,29 @@ CLICKHOUSE_HOST=<host> CLICKHOUSE_PASSWORD=<pw> ./scripts/verify.sh --database o
...
+> the last reported week plus the **stride** — plus one only at the default `--weeks-stride 1` — and the same
+> stride must be passed again, or the resumed run samples different windows than the run it continues. The
+> offsets are anchored as above, so confirm a resumed run's first window is the one you expect before treating its
+> output as continuous with an earlier log — if the anchor has moved, the logs describe different windows and must not be
+> read as one run.
+>
+> **A resumed run does not by itself discharge the pre-`EXCHANGE` gate.** That gate (see the exit checklist) requires a
+> full compare with no narrowing, and each `PASSED` line reports only the windows *that run* compared. Split runs satisfy
+> it only if their windows together cover every week **and** the anchor held throughout; if either is in doubt, re-run
+> whole. Resuming is for the long post-rollback compare, where the bound is deliberately partial anyway.
```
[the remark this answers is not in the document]
andrescrz: Two claims here; the first I am declining, the second is real and now documented.

**The `MISMATCH`-before-drill-down ordering.** Verified the ordering is as you describe (`verify.sh:228` logs, `231` drills). But a resume does not skip the failure: the `MISMATCH` line carries the verdict *and* `genuinely_differing_keys`, so week N is recorded as failed before anything can abort. What a resume loses is only the per-key listing, which is re-obtainable for that one window with `--from-week N --to-week N --drill-down` — far cheaper than re-running the range. Marking the week complete only after both steps would instead re-compare a week whose verdict is already known, on a table where a window can take tens of minutes. The current order records the conclusion first, which is the right priority when a run may be interrupted.

**`LAST_WEEK` captured once — correct, and inherent.** It is read at `verify.sh:169` and never revalidated, so rows written after a run starts fall outside the horizon it computed. That cannot be fixed by revalidating: on a table still taking writes, extending the bound mid-run is an unbounded chase, and the cutover does not need it to be — rows written during the compare are the **delta insert's** responsibility, not the compare's. That division is the whole reason the delta exists.

What was missing was saying so, since "full compare" in the checklist could be read as "covers everything, including writes arriving now". The admonition now states it:

> Both are read once, at startup: rows written after that are outside the horizon the run computed, and are the delta's business rather than the compare's.

Note this is the same root as @thiagohora's finding on the bound, which I applied in the same block — the difference being that his case (a *moving* anchor shifting offsets between runs) can silently invalidate a resume, whereas an advancing horizon cannot.

thiagohora (approved):

PR #8053, opened 2026-08-28
[OPIK-7315] document the version tie the confirm-keys re-check cannot resolve

andrescrz:
## Details

A fidelity compare on an internal test environment reported a week with **equal row counts** and `genuinely_differing_keys=1`, on a key both tables held identically. Read without `FINAL`, both sides carried **both** versions with the same `last_updated_at` — the `ReplacingMergeTree` version column — so `FINAL` had no winner to return, and the two part layouts (unpartitioned vs `id_at`-partitioned) broke the tie differently.

Two documentation problems:

- **The confirm-keys re-check overstates its guarantee.** Its header says filtering on the sorting key means `FINAL` "returns the true winner". That holds only while versions differ. On a tie it reports a false positive, in the query whose purpose is removing them.
- **The runbook's triage rules cover neither shape.** The documented cases are "present on both sides with `last_updated_at >= cutover_start`" (benign) and "absent from the successor" (a copy gap). A tie is present on both sides *and* predates `cutover_start`, so the rules leave it looking unexplained.

## Change checklist
- [ ] User facing
- [x] Documentation update

## Issues

- Resolves #
- OPIK-7315

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: the limitation note in `000005_verify_migration.sql` and the triage branch in the runbook.
- Human verification: the diagnosis was confirmed against a live instance (below).

## Testing

Documentation only; no code path changes, so no test changes.

The diagnosis was verified by reading the differing key on both sides **without** `FINAL`: two rows per side, identical `last_updated_at`, `created_at` values minutes apart, differing only by the expected ns→us truncation. Row counts matched on both sides, which is the signature the new triage branch tells an operator to look for.

## Documentation

- `scripts/db-app-analytics/000005_verify_migration.sql` — the limitation sits immediately above the claim it qualifies, so a reader of the guarantee sees its bound.
- `README.md` — a third triage shape: the recognition signature, the fact that no shipped driver can confirm it (`--drill-down` reads one `FINAL` row per key and stops at 100), and the safe action meanwhile — treat the week as unresolved rather than passing it.

## Deliberately not in this PR

Teaching the query to classify ties. Two designs were considered; the better one is **not** the one this PR's earlier revision sketched:

- comparing the full version **set** per key proves faithfulness, but needs the row-hash expression a third time per side in a file that already carries fourteen copies, and changes the gate's output shape;
- reporting a non-unique top version as **inconclusive** rather than genuinely-differing needs no hash duplication and cannot produce a false negative, since inconclusive is not a pass. It is the smaller and safer change, and the one worth making.

Either way the driver and its test move with it, so it belongs in its own change rather than appended to a doc fix.

### DIFF
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
@@ -1410,6 +1410,11 @@ CLICKHOUSE_HOST=<host> CLICKHOUSE_PASSWORD=<pw> ./scripts/verify.sh --database o
 ./scripts/verify.sh --database opik --old-table traces_pre_cutover_backup --new-table traces
 ```
 
+> **A version tie can make this gate mismatch, and can also make it pass.** If a key carries two or more rows with an
+> identical `last_updated_at`, `FINAL` has no winner and the comparison for that key is arbitrary in both directions. It
+> is not rollback-specific despite where it is written up: see "a version tie" under *Verifying after a rollback* for the
+> shape and the confirming read, ignoring that section's `cutover_start` test, which has no meaning before the `EXCHANGE`.
+>
 > **Detach it, and expect tens of minutes.** The bounded compare walks one window per week over both
 > tables. On a large table that is minutes per window on the busy weeks and well over half an hour in
 > total, so run it under `nohup`/`screen` rather than an interactive shell that may be interrupted. It is
@@ -1480,17 +1485,33 @@ from the successor *entirely* is the real signal — that is a copy gap, and it 
 often this bites tracks how much pre-existing data the workload rewrites; for many it is none, which is why the weekly
 bound is still worth passing.
 
-**A third shape, which the confirm-keys re-check cannot resolve: a version tie.** Its premise — that filtering on the
-sorting key lets `FINAL` return the true winner — holds only while versions differ. When a key carries two or more rows
-with an **identical** `last_updated_at`, the `ReplacingMergeTree` version column, there is no winner: `FINAL` picks
-arbitrarily, and because the two tables' part layouts differ, each side can pick a different row. The key is then counted
-as `genuinely_differing_keys` even where both tables hold the same data.
+**A third shape, which the confirm-keys re-check cannot resolve: a version tie.** This one is not rollback-specific — it
+can hit the pre-`EXCHANGE` gate too (see "Verifying the migration"), where the `cutover_start` test below does not apply.
 
-Its signature is **equal row counts with a differing checksum**, and neither rule above fits it — the key is present on
-both sides *and* its `last_updated_at` predates `cutover_start`. Deciding it needs the key's full set of versions read
-from both tables without `FINAL`, which no shipped driver does; `--drill-down` reads one `FINAL` row per key and stops at
-100 keys, so it cannot show a tie. **Treat such a week as unresolved and escalate rather than passing it:** a non-unique
-top version makes the comparison arbitrary, which is not the same as benign.
+Its premise — that filtering on the sorting key lets `FINAL` return the true winner — holds only while versions differ.
+When a key carries two or more rows with an **identical** `last_updated_at`, the `ReplacingMergeTree` version column,
+there is no winner: `FINAL` picks arbitrarily, and because the two tables' part layouts differ, each side may or may not
+land on the same row. **Arbitrary cuts both ways, and the second direction is the dangerous one:**
+
+- the picks differ, and the key is reported in `genuinely_differing_keys` even though both tables hold the same data;
+- the picks coincide, and the key is confirmed as matching **even if one side is missing a version** — a real copy gap.
+  So a `0` from the re-check, and any `OK — superseded-version artifact` verdict built on it, is not conclusive while
+  ties exist.
+
+`--drill-down` cannot show a tie: it reads one `FINAL` row per key and stops at 100. Confirm one by reading the key's
+versions from both tables without `FINAL` — a read-only diagnostic, not a procedure step:
+
+```sql
+SELECT 'src' AS side, created_at, last_updated_at, _part FROM <old-table> WHERE (workspace_id, project_id, id) = (…)
+UNION ALL
+SELECT 'dst' AS side, created_at, last_updated_at, _part FROM <new-table> WHERE (workspace_id, project_id, id) = (…)
+ORDER BY side, created_at;
+```
+
+A **non-unique top `last_updated_at` on either side** means the comparison for that key was arbitrary. Then compare the
+two version sets: identical sets mean the copy is faithful and only the tie-break differed; a version present on one
+side only is the copy gap, whatever the re-check said. If the sets cannot be established, treat the week as unresolved
+and escalate rather than passing it — arbitrary is not the same as benign.
 
 > **The pre-EXCHANGE compare is the gate; the post-EXCHANGE compare has a caveat.** `traces_pre_cutover_backup` is a
 > **frozen** snapshot as of `cutover_start`, but live `traces` keeps taking writes the instant the buffer drains — so
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
@@ -1480,6 +1480,18 @@ from the successor *entirely* is the real signal — that is a copy gap, and it 
 often this bites tracks how much pre-existing data the workload rewrites; for many it is none, which is why the weekly
 bound is still worth passing.
 
+**A third shape, which the confirm-keys re-check cannot resolve: a version tie.** Its premise — that filtering on the
+sorting key lets `FINAL` return the true winner — holds only while versions differ. When a key carries two or more rows
+with an **identical** `last_updated_at`, the `ReplacingMergeTree` version column, there is no winner: `FINAL` picks
+arbitrarily, and because the two tables' part layouts differ, each side can pick a different row. The key is then counted
+as `genuinely_differing` even where both tables hold the same data.
+
+Its signature is **equal row counts with a differing checksum**, and neither rule above fits it — the key is present on
+both sides *and* its `last_updated_at` predates `cutover_start`. Deciding it needs the key's full set of versions read
+from both tables without `FINAL`, which no shipped driver does; `--drill-down` reads one `FINAL` row per key and stops at
+100 keys, so it cannot show a tie. **Treat such a week as unresolved and escalate rather than passing it:** a non-unique
+top version makes the comparison arbitrary, which is not the same as benign.
+
 > **The pre-EXCHANGE compare is the gate; the post-EXCHANGE compare has a caveat.** `traces_pre_cutover_backup` is a
 > **frozen** snapshot as of `cutover_start`, but live `traces` keeps taking writes the instant the buffer drains — so
 > the **current (live) week will legitimately show a mismatch** (the live table is a superset of the frozen backup by
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
@@ -1489,9 +1489,9 @@ bound is still worth passing.
 can hit the pre-`EXCHANGE` gate too (see "Verifying the migration"), where the `cutover_start` test below does not apply.
 
 Its premise — that filtering on the sorting key lets `FINAL` return the true winner — holds only while versions differ.
-When a key carries two or more rows with an **identical** `last_updated_at`, the `ReplacingMergeTree` version column,
-there is no winner: `FINAL` picks arbitrarily, and because the two tables' part layouts differ, each side may or may not
-land on the same row. **Arbitrary cuts both ways, and the second direction is the dangerous one:**
+`last_updated_at` is the `ReplacingMergeTree` version column, so when two or more rows for a key carry the **same**
+value there is nothing left to rank them by: `FINAL` picks arbitrarily, and because the two tables' part layouts differ,
+each side may or may not land on the same row. **Arbitrary cuts both ways, and the second direction is the dangerous one:**
 
 - the picks differ, and the key is reported in `genuinely_differing_keys` even though both tables hold the same data;
 - the picks coincide, and the key is confirmed as matching **even if one side is missing a version** — a real copy gap.
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
@@ -1484,7 +1484,7 @@ bound is still worth passing.
 sorting key lets `FINAL` return the true winner — holds only while versions differ. When a key carries two or more rows
 with an **identical** `last_updated_at`, the `ReplacingMergeTree` version column, there is no winner: `FINAL` picks
 arbitrarily, and because the two tables' part layouts differ, each side can pick a different row. The key is then counted
-as `genuinely_differing` even where both tables hold the same data.
+as `genuinely_differing_keys` even where both tables hold the same data.
 
 Its signature is **equal row counts with a differing checksum**, and neither rule above fits it — the key is present on
 both sides *and* its `last_updated_at` predates `cutover_start`. Deciding it needs the key's full set of versions read
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000005_verify_migration.sql
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000005_verify_migration.sql
@@ -209,7 +209,8 @@ SETTINGS join_use_nulls = 1, use_skip_indexes_if_final = 1;
 -- LIMITATION: this cannot resolve a VERSION TIE. The reasoning below holds only while versions differ. When a key
 -- carries two or more rows with an identical last_updated_at, FINAL has no winner to return: it picks arbitrarily, and
 -- the two tables' part layouts differ, so each side can pick a different row. Such a key is counted as genuinely
--- differing even where both tables hold the same data. Signature: equal row counts with a differing checksum. Deciding
+-- differing (genuinely_differing_keys) even where both tables hold the same data. Signature: equal row counts with a
+-- differing checksum. Deciding
 -- it needs each side's full version SET, which this query does not read -- see the runbook's triage section.
 --
 -- 0  = every differing key has identical live rows on both sides -> superseded-version artifact, not a
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000005_verify_migration.sql
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000005_verify_migration.sql
@@ -206,6 +206,12 @@ SETTINGS join_use_nulls = 1, use_skip_indexes_if_final = 1;
 -- granules holding the key, so FINAL always sees all versions of it and returns the true winner. The
 -- window bounds are used only to pick the candidate keys, never to decide the verdict.
 --
+-- LIMITATION: this cannot resolve a VERSION TIE. The reasoning below holds only while versions differ. When a key
+-- carries two or more rows with an identical last_updated_at, FINAL has no winner to return: it picks arbitrarily, and
+-- the two tables' part layouts differ, so each side can pick a different row. Such a key is counted as genuinely
+-- differing even where both tables hold the same data. Signature: equal row counts with a differing checksum. Deciding
+-- it needs each side's full version SET, which this query does not read -- see the runbook's triage section.
+--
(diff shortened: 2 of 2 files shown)


CometActions:
**No test needed here.**

Documentation only: 19 added lines, 0 removed, and every added line in 000005_verify_migration.sql is a `--` comment — the SQL that runs is byte-identical. The README hunk adds the version-tie triage section next to it. This is the one-off traces-local-v2-cutover runbook an operator runs by hand against ClickHouse, not a product path, so there is no user-visible behaviour for a test to assert. Worth noting the comment says teaching the query to classify ties was deliberately left out — if that classification later lands as real SQL, that change is worth re-triaging.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/33181146882)

_Re-checked after a push on 28 Aug 14:39 UTC — nothing the verdict depends on changed._

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
```diff
@@ -1480,6 +1480,17 @@ from the successor *entirely* is the real signal — that is a copy gap, and it
...
 bound is still worth passing.
 
+**A third shape the confirm-keys re-check cannot resolve: a version tie.** Its premise is that filtering on the sorting
+key lets `FINAL` see every version and return the true winner. That holds only while versions differ. If a key has two or
+more rows carrying an **identical** `last_updated_at` — the `ReplacingMergeTree` version column — there is no winner:
+`FINAL` picks arbitrarily, the two tables have different part layouts, and each side can pick a different row. The
+re-check then counts the key as `genuinely_differing`, on a copy that is in fact faithful.
+
+Recognise it by **equal row counts with a differing checksum**, then read the key on both sides **without** `FINAL`. If
+each side holds the same set of versions and the top `last_updated_at` is not unique, the copy is faithful and only the
```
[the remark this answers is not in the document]
andrescrz: **Your diagnosis is right; I am declining the copyable query, and fixing the gap it exposes a different way.**

Verified both claims first — the drill-down block does `FROM … OLD_TABLE FINAL` / `NEW_TABLE FINAL` and ends `LIMIT 100`. So it shows one `FINAL` row per key, truncates at 100, and cannot reveal a tie. Correct on all three counts.

**Why not a pasteable query.** This procedure's governing rule is that every statement runs from a versioned script, never by hand. Hand-written SQL was removed from this runbook earlier in this work for exactly that reason; adding a triage query back would reintroduce what the rule exists to prevent, and a query composed under cutover pressure against a `ReplacingMergeTree` tie is a poor place to start making exceptions.

**What I changed instead.** The branch no longer implies a triage the tooling can deliver. It now states the tooling gap and the safe action:

> Deciding it needs the key's full set of versions read from both tables without `FINAL`, which no shipped driver does; `--drill-down` reads one `FINAL` row per key and stops at 100 keys, so it cannot show a tie. **Treat such a week as unresolved and escalate rather than passing it:** a non-unique top version makes the comparison arbitrary, which is not the same as benign.

I also corrected an overclaim of my own in the previous revision, which said a matching version set means "the copy is faithful" — non-uniqueness alone establishes that the comparison is *arbitrary*, not that the data agrees. Those are different statements and the doc now makes the weaker, true one.

**On the real fix**, your comment changed my view of the right design. I had sketched comparing the full version set per key; the better option is to report a non-unique top version as **inconclusive** rather than genuinely-differing. It needs no hash duplication, and it cannot produce a false negative, because inconclusive is not a pass. Recorded in the PR body as the option worth taking. It still moves the driver and its test, so it stays a separate change — but it is the change, not the set comparison.

andrescrz:
**Process note: this branch's single commit was amended after the review above, and force-pushed. That was a mistake on my part — it destroyed the revision the review commented on.** Recording what changed so the trail is not lost.

Reviewed revision: `3ce5bda4bf`. Current: `933b8e76aa`. The amend folded in (+16/−17, both files, docs only):

**README** — the reviewed text claimed a matching version set means "the copy is faithful" and told the operator to "read the key on both sides without `FINAL`" without saying that no shipped driver does so. Replaced with the weaker true claim (a non-unique top version makes the comparison *arbitrary*, not benign), the `--drill-down` limitation you identified, and the safe action: treat the week as unresolved rather than passing it.

**`000005_verify_migration.sql`** — same correction, and the note lost a sentence of process editorialising about why the real fix was deferred; the reasoning now lives in the PR body where it belongs.

No further commits will be amended on this branch or the others in flight.

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
```diff
@@ -1480,6 +1480,18 @@ from the successor *entirely* is the real signal — that is a copy gap, and it
 often this bites tracks how much pre-existing data the workload rewrites; for many it is none, which is why the weekly
 bound is still worth passing.
 
+**A third shape, which the confirm-keys re-check cannot resolve: a version tie.** Its premise — that filtering on the
+sorting key lets `FINAL` return the true winner — holds only while versions differ. When a key carries two or more rows
+with an **identical** `last_updated_at`, the `ReplacingMergeTree` version column, there is no winner: `FINAL` picks
+arbitrarily, and because the two tables' part layouts differ, each side can pick a different row. The key is then counted
+as `genuinely_differing` even where both tables hold the same data.
```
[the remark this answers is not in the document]
andrescrz: Applied. `verify.sh` reports `genuinely_differing_keys=$unresolved`, so quoting the field short defeated the purpose of quoting it at all — the signature has to be greppable against real output. Named in the SQL note as well, for the same reason.

Committed on top as `00446c96cf`; no history rewritten.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000005_verify_migration.sql
```diff
@@ -206,6 +206,13 @@ SETTINGS join_use_nulls = 1, use_skip_indexes_if_final = 1;
 -- granules holding the key, so FINAL always sees all versions of it and returns the true winner. The
 -- window bounds are used only to pick the candidate keys, never to decide the verdict.
 --
+-- LIMITATION: this cannot resolve a VERSION TIE. The reasoning below holds only while versions differ. When a key
+-- carries two or more rows with an identical last_updated_at, FINAL has no winner to return: it picks arbitrarily, and
+-- the two tables' part layouts differ, so each side can pick a different row. Such a key is counted as genuinely
+-- differing (genuinely_differing_keys) even where both tables hold the same data. Signature: equal row counts with a
```
thiagohora: **The tie is documented in one direction only — it can also produce a false *pass*, which is the dangerous half.**

The note says a tie makes a key "counted as genuinely differing … even where both tables hold the same data", i.e. it warns about over-reporting. But the arbitrary pick cuts both ways, and the other direction is the one that matters at this gate:

- source holds tied versions `v1, v2`; the successor is missing one of them — a **genuine copy gap**;
- `FINAL` on each side happens to land on the same row;
- `confirm-keys` returns `0`, `verify.sh:225` logs `OK — superseded-version artifact`, and the run **exits 0**.

So a real fidelity defect is waved through by the gate that precedes the irreversible `EXCHANGE`. An operator who reads this note as written concludes ties only raise false alarms, and goes on trusting every `OK — superseded-version artifact` line — including that one.

The framing to aim for is the one the README already reaches at line 1493: *a non-unique top version makes the comparison arbitrary*. Arbitrary is symmetric. Suggest saying explicitly that such a key may be counted as genuinely differing **or** silently confirmed as matching, and that `0` from this block is therefore not conclusive in the presence of ties. Same edit needed at `README.md:1487`.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000005_verify_migration.sql
```diff
@@ -206,6 +206,13 @@ SETTINGS join_use_nulls = 1, use_skip_indexes_if_final = 1;
 -- granules holding the key, so FINAL always sees all versions of it and returns the true winner. The
 -- window bounds are used only to pick the candidate keys, never to decide the verdict.
 --
+-- LIMITATION: this cannot resolve a VERSION TIE. The reasoning below holds only while versions differ. When a key
```
thiagohora: **"The reasoning below" points the wrong way.**

The argument this limitation qualifies — "why this re-check is trustworthy" — is lines 204-207, **above** the block. What's actually below (216-218) is the `0` / `>0` verdict legend.

So a reader who follows "below" lands on the output legend and never connects the caveat to the guarantee it bounds. The PR description says "the limitation sits immediately above the claim it qualifies"; the diff does the reverse.

Either move the block above line 204, or change "below" → "above". Moving it is the better fix — the caveat reads more naturally as a lead-in to the trustworthiness argument than as an afterthought to it, and it keeps the verdict legend directly adjacent to the values it explains.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000005_verify_migration.sql
```diff
@@ -206,6 +206,13 @@ SETTINGS join_use_nulls = 1, use_skip_indexes_if_final = 1;
...
+-- carries two or more rows with an identical last_updated_at, FINAL has no winner to return: it picks arbitrarily, and
+-- the two tables' part layouts differ, so each side can pick a different row. Such a key is counted as genuinely
+-- differing (genuinely_differing_keys) even where both tables hold the same data. Signature: equal row counts with a
+-- differing checksum. Deciding
+-- it needs each side's full version SET, which this query does not read -- see the runbook's triage section.
```
thiagohora: **The "signature" (line 212) doesn't discriminate, and here it stands alone with no follow-up test.**

"Equal row counts with a differing checksum" is also the signature of the rollback trace-update shape already documented at `README.md:1476` — key present on both sides, equal counts, differing checksum. In the README that sentence is at least followed by the real discriminator (`last_updated_at` vs `cutover_start`); in this SQL comment it's the whole test, so a reader of this file gets a "signature" matching several unrelated cases and no way to tell them apart.

Worth also noting that those row counts aren't output by the `confirm-keys` block at all — they come from the `compare` block / the driver's `MISMATCH` log line. As written, someone reading this comment in isolation will look for counts in the wrong output.

**Nit, same sentence:** the line break at 213 falls mid-clause (`…with a` / `-- differing checksum. Deciding` / `-- it needs each side's full version SET…`), leaving a ~30-char line among ~118-char neighbours. Worth reflowing to match the block.

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
```diff
@@ -1480,6 +1480,18 @@ from the successor *entirely* is the real signal — that is a copy gap, and it
 often this bites tracks how much pre-existing data the workload rewrites; for many it is none, which is why the weekly
 bound is still worth passing.
 
+**A third shape, which the confirm-keys re-check cannot resolve: a version tie.** Its premise — that filtering on the
+sorting key lets `FINAL` return the true winner — holds only while versions differ. When a key carries two or more rows
+with an **identical** `last_updated_at`, the `ReplacingMergeTree` version column, there is no winner: `FINAL` picks
+arbitrarily, and because the two tables' part layouts differ, each side can pick a different row. The key is then counted
+as `genuinely_differing_keys` even where both tables hold the same data.
```
thiagohora: **This is filed under rollback triage, but a version tie is compare-wide — it also hits the pre-`EXCHANGE` gate, where the discriminator below doesn't exist.**

The paragraph sits inside **Verifying after a rollback** (section starts at 1449), and the test it offers at 1490 is "the key is present on both sides *and* its `last_updated_at` predates `cutover_start`".

Before the `EXCHANGE` there is no `cutover_start` to compare against. But a tie can just as easily make the pre-`EXCHANGE` gate compare — documented at 1387-1447, and required to pass before an irreversible step — report `genuinely_differing_keys>0`. That operator is under the most time pressure of anyone who will read this, and the SQL note's pointer ("see the runbook's triage section") sends them straight here, to text whose discriminator they can't apply.

Suggest either hoisting the tie shape out of the rollback subsection to somewhere both gate and rollback readers reach, or adding a sentence to the gate section at 1387-1447 that names the shape and points here for the rollback-specific part.

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
```diff
@@ -1480,6 +1480,18 @@ from the successor *entirely* is the real signal — that is a copy gap, and it
...
+
+Its signature is **equal row counts with a differing checksum**, and neither rule above fits it — the key is present on
+both sides *and* its `last_updated_at` predates `cutover_start`. Deciding it needs the key's full set of versions read
+from both tables without `FINAL`, which no shipped driver does; `--drill-down` reads one `FINAL` row per key and stops at
+100 keys, so it cannot show a tie. **Treat such a week as unresolved and escalate rather than passing it:** a non-unique
+top version makes the comparison arbitrary, which is not the same as benign.
```
thiagohora: **"Escalate" is the only guidance, and there's no check offered to confirm a tie — so every mismatch becomes unresolvable.**

The paragraph correctly establishes that no shipped driver can distinguish a tie (`--drill-down` reads one `FINAL` row per key and stops at 100). But the PR body describes exactly how the diagnosis *was* made: read the key on both sides **without** `FINAL` — two rows per side, identical `last_updated_at`. That query appears nowhere in the runbook.

Without it, an operator facing `genuinely_differing_keys>0` has no way to tell a tie from a real fidelity failure, so the honest reading of this guidance is "treat every mismatch as unresolved and escalate". That drains the gate of its discriminating value at precisely the highest-pressure moment — which is the opposite of what documenting the limitation is for.

Adding the no-`FINAL` per-key query here would make this branch actionable: the operator confirms the tie, and only then escalates. Something along the lines of selecting the key's rows from both tables without `FINAL` and comparing the `last_updated_at` sets.

(Line 1493's "arbitrary … is not the same as benign" is the right note to end on, and would land harder with a concrete check in front of it.)

thiagohora (commented):
Traced the `compare` → `confirm-keys` → `verify.sh` path to check the premise, and **the underlying diagnosis is right**: the version-tie mechanism this documents is real and reachable, `FINAL` genuinely has no winner to return on a non-unique `last_updated_at`, and differing part layouts genuinely let each side pick a different row. Good catch, and worth documenting.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000005_verify_migration.sql
andrescrz: **Applied, and this is the comment that mattered.** I documented the harmless direction and stopped, which is worse than not documenting it: a reader concludes ties only raise false alarms and keeps trusting every `OK — superseded-version artifact` line, including the one covering a real gap.

Your chain is right, and it is worse than confirm-keys alone. The *compare* block collapses with `FINAL` too, so if the source's arbitrary pick lands on the version the successor kept, the week produces no mismatch at all — the false pass can happen before confirm-keys is ever reached.

Both directions are now stated in the SQL note and the runbook, with the consequence made explicit:

> * the picks differ -> the key is reported in genuinely_differing_keys even where both tables hold the same data;
> * the picks coincide -> the key is confirmed as matching even if one side is MISSING a version, which is a real copy gap. A 0 from this block is therefore not conclusive while ties exist, and neither is an "OK -- superseded-version artifact" verdict built on it.

And you are right that `arbitrary is symmetric` was the framing to work back from — the README reached it and the SQL note did not, which is exactly how one direction got written up and the other did not.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000005_verify_migration.sql
andrescrz: **Applied — moved, as you suggested, rather than reworded.** You are right that the PR description claimed the opposite of the diff: I wrote "the limitation sits immediately above the claim it qualifies" and then put it below.

It now leads into the trustworthiness argument, whose heading also says what it is now bounded by (`Why this re-check is trustworthy WHERE VERSIONS DIFFER`), and the `0` / `>0` legend is back to being adjacent to the values it explains.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000005_verify_migration.sql
andrescrz: **Applied — the "signature" is gone from this file rather than qualified.**

Both objections hold. "Equal row counts with a differing checksum" also describes the rollback trace-update shape at README:1476, and in the README that sentence at least has the discriminator after it; here it was the whole test. And the counts come from the compare block and the driver's `MISMATCH` line, not from this one, so a reader of this comment in isolation would look in the wrong output for them.

The note now states the limitation and defers the shape to the runbook, where the discriminator follows it — with the reason spelled out so nobody reinstates it: *"do not infer the shape from row counts here, which this block does not output."*

The mid-clause line break went with the rewrite.

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
andrescrz: **Applied.** A tie is a property of `FINAL` over the sorting key, so it has nothing to do with the rollback; I filed it under rollback triage because that is where I found it, which is not a reason.

Rather than hoist the whole write-up, the tie paragraph now opens by saying it is not rollback-specific and names the gate, and the **gate section carries its own pointer** — because that reader is the one under time pressure before an irreversible step, and the SQL note sends them to the runbook:

> **A version tie can make this gate mismatch, and can also make it pass.** … It is not rollback-specific despite where it is written up: see "a version tie" under *Verifying after a rollback* for the shape and the confirming read, ignoring that section's `cutover_start` test, which has no meaning before the `EXCHANGE`.

Explicitly telling them to ignore the `cutover_start` test, since that is the trap in following the pointer.

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
andrescrz: **Applied, and this reverses a position I took earlier in this PR.** @baz-reviewer raised the same gap and I declined it as hand-written SQL, citing the rule that every statement comes from a versioned script. That rule governs procedure statements and writes; a read-only diagnostic is neither, and this runbook already instructs ad-hoc reads elsewhere ("look each differing id up in the parked successor without a week filter"). I over-applied it, and your framing is why: without a check, "escalate" is the whole branch, which drains the gate of its discriminating value at the highest-pressure moment.

The branch now carries the read that produced the diagnosis:

```sql
SELECT 'src' AS side, created_at, last_updated_at, _part FROM <old-table> WHERE (workspace_id, project_id, id) = (…)
UNION ALL
SELECT 'dst' AS side, created_at, last_updated_at, _part FROM <new-table> WHERE (workspace_id, project_id, id) = (…)
ORDER BY side, created_at;
```

Followed by what to conclude: a non-unique top `last_updated_at` on either side means that key's comparison was arbitrary; identical version sets mean the copy is faithful and only the tie-break differed; a version present on one side only is the copy gap **whatever the re-check said** — which is the false-pass direction from your other comment. Escalation is now the fallback when the sets cannot be established, not the only option.

`_part` is in the projection deliberately: it is what makes the differing part layouts visible, and that is the mechanism rather than a coincidence.

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
```diff
@@ -1480,6 +1485,34 @@ from the successor *entirely* is the real signal — that is a copy gap, and it
 often this bites tracks how much pre-existing data the workload rewrites; for many it is none, which is why the weekly
 bound is still worth passing.
 
+**A third shape, which the confirm-keys re-check cannot resolve: a version tie.** This one is not rollback-specific — it
+can hit the pre-`EXCHANGE` gate too (see "Verifying the migration"), where the `cutover_start` test below does not apply.
+
+Its premise — that filtering on the sorting key lets `FINAL` return the true winner — holds only while versions differ.
+When a key carries two or more rows with an **identical** `last_updated_at`, the `ReplacingMergeTree` version column,
+there is no winner: `FINAL` picks arbitrarily, and because the two tables' part layouts differ, each side may or may not
+land on the same row. **Arbitrary cuts both ways, and the second direction is the dangerous one:**
```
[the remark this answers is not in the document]
andrescrz: Applied. The appositive sat between the condition and the consequence, so the sentence had to be re-read to parse. Reordered as you suggest — the column's role first, then the condition, then what `FINAL` does:

> `last_updated_at` is the `ReplacingMergeTree` version column, so when two or more rows for a key carry the **same** value there is nothing left to rank them by: `FINAL` picks arbitrarily, and because the two tables' part layouts differ, each side may or may not land on the same row.

Same phrasing applied to the SQL note, which had the opposite problem — it never named the column's role, so a reader there had to know already why an identical `last_updated_at` matters.

thiagohora (approved):

PR #8055, opened 2026-08-28
[OPIK-7315] scope the sentinel repair to the flag window

andrescrz:
## Details

The sentinel repair matched **every** epoch `end_time` and NaN `ttft` in `traces`. That is unsafe, and the pre-flight on an internal test environment caught it before it ran: the unbounded predicate matched **34 keys across 12 workspaces**, of which **5** came from the flag window. The other 29 carried genuine client-sent values, spanning ten months before the flag was ever flipped.

Three things made this worse than an over-broad `UPDATE`:

- **The values are unrecoverable.** The parked successor encodes an absent `end_time` as that same epoch, so there is no reference copy to restore from.
- **The gate would have passed.** `sentinel_end_time`/`sentinel_ttft` reaching 0 means "no sentinel remains" — exactly what destroying the genuine ones achieves.
- **No existing guard covers it.** `--confirm-flag-was-live` is not even required when a promote parked the successor, and the counts only report totals.

`--sentinel-window-from` / `--sentinel-window-to` are now **required**. Rows are matched on `created_at` **or** `last_updated_at`, the same union the delta insert uses and for the same reason: a row created in the window is caught by the first, a pre-existing row updated in it — which is where its sentinel came from — by the second.

The **counts carry the same window**. Unbounded they would include values the flag never produced, so the gate could never reach 0 on any estate holding a genuine epoch value, and a repair that correctly left them alone would read as a failure.

Both bounds are pinned to `'UTC'`. The timezone assertion added alongside the epoch literal caught this during development: unpinned, the window shifts with the server timezone and matches nothing — a silent no-op reported as success.

## Change checklist
- [ ] User facing
- [x] Documentation update

## Issues

- Resolves #
- OPIK-7315

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: both `.sql` files, the driver flags and validation, the gate test, and the runbook.
- Human verification: the shipped query was run as-is against a live multi-replica instance (below).

## Testing

`mvn test -Dtest=TracesLocalV2CutoverTest` — **14/14 pass**.

The gate test gains a cohort that matches the repair predicate but was written **outside** the window, and asserts it keeps both values after a full repair. Two existing expectations moved with it (`duration < 0` totals 11 before and 5 after, rather than 9 and 3) — those counts are whole-workspace by design, which is the point they make about a negative-duration total being useless as a gate.

Development order worth recording: the out-of-window assertion passed immediately, and the **timezone** assertion failed. That caught the unpinned window bounds, not review.

**Verified against a live instance.** The shipped `000004_rollback_verify_sentinels.sql`, run as-is with the window substituted, returned:

```
sentinel_end_time  sentinel_ttft  negative_from_sentinel  stale_duration
5                  5              5                       0
```

Exactly the 5 rows the flag produced, with all 29 genuine matches excluded. Unbounded, the same query reports all 34.

## Documentation

`README.md` — both command sites now carry the window, with the hazard stated once: no safe default, where to take the bounds from, that widening is safe and guessing is not, and that the bounds are UTC.

### DIFF
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
@@ -1207,9 +1207,21 @@ exists (`Code 60`). That is the second of the two flags the stage comparison tab
    `NULL` with no way back — the parked successor encodes an absent `end_time` as that same epoch, so nothing holds the
    original — and the counts would still report success. Measured on an internal environment: the unbounded predicate
    matched 34 keys across 12 workspaces where only 5 came from the flag window. Take the bounds from when the flag
-   rolled out and when its revert finished landing on every instance; widening is safe, guessing is not. Rows are
-   matched on `created_at` **or** `last_updated_at`, so a row created in the window and a pre-existing row updated in it
-   are both caught. Both bounds are interpreted as UTC regardless of the server's timezone.
+   rolled out and when its revert finished landing on every instance. Rows are matched on `created_at` **or**
+   `last_updated_at`. Both bounds are interpreted as UTC regardless of the server's timezone.
+
+   **One case the window cannot catch, and the gate cannot see.** `TraceDAO.UPDATE` re-inserts a version copying
+   `created_at` and — when the patch omits them — `end_time`/`ttft` verbatim, while `last_updated_at` takes
+   `DEFAULT now64(6)`. So a trace created *before* the window, patched *inside* it under the flag, then patched *again*
+   after the revert has a live version carrying a pre-window `created_at` and a post-window `last_updated_at`, matching
+   neither arm. It keeps its epoch `end_time`, and because the repair does clear the older in-window version the counts
+   still reach `0` and report success.
+
+   Extending `--sentinel-window-to` to the moment the repair runs closes that, at a cost worth stating rather than
+   burying: a row holding a **genuine** epoch `end_time` that was merely patched inside the widened range then matches
+   too, and is nulled irrecoverably. `end_time` is carried forward verbatim, so nothing in the data separates the two
+   cases. Neither bound is safe in both directions — choose knowingly, and use the unbounded counts the driver prints
+   alongside to see what a wider window would take in.
 
    It reads the counts first and issues no mutation when they are `0`, restores `NULL` in a single mutation
    (`000004_rollback_sentinel_repair.sql`) which recomputes `duration` as it rewrites each row, then asserts the counts
@@ -1248,10 +1260,13 @@ Treat a stage B/C rollback as complete only when all of these hold:
       `tracesDistributedWrapEnabled` if the wrap had been applied. Those are the only two — partition pruning is
       unconditional and has no flag. Verify positively, not by absence of errors: absent `end_time`/`ttft` must read back
       as `null`.
-- [ ] **Sentinel repair applied** — `--sentinel-repair-only` printed `Sentinel postcondition OK` (or reported nothing to
-      repair, which is equally valid) and **exited zero**. The gate is `sentinel_end_time`, `sentinel_ttft` and
-      `stale_duration` all at `0`; a residual `duration < 0` count elsewhere in the table is expected, from rows whose
-      `end_time` genuinely precedes `start_time`.
+- [ ] **Sentinel repair applied** — `--sentinel-repair-only` printed `Sentinel postcondition OK` and **exited zero**,
+      **and the window passed is the one the flag was live in, in UTC**. The gate is `sentinel_end_time`,
+      `sentinel_ttft` and `stale_duration` all at `0` *inside that window*; a residual `duration < 0` count elsewhere is
+      expected, from rows whose `end_time` genuinely precedes `start_time`.
+      **"Nothing to repair" is not interchangeable with a completed repair.** It is equally what a wrong window
+      produces — bounds in local time being the common case — so check it against the unbounded counts the driver prints
+      beside it before ticking this. This box gates `finalize.sh`, which truncates the parked successor.
 - [ ] **The parked successor still parked** — `traces_post_rollback_backup` retained, not finalized. It is the only copy
       of the post-cutover writes the rollback discarded, and the only thing that makes a retry cheap.
 
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
@@ -1036,14 +1036,15 @@ Pick the stage by how far the cutover got:
   `RENAME`, then drops the ex-wrapper — landing in the post-`EXCHANGE`, pre-wrap state. (Guarded: aborts unless `traces`
   is `Distributed` and `traces_local` holds the successor schema.) See "Un-wrap" below for when to prefer it over stage C.
 - **Sentinel repair — after a stage B/C promote, or after abandoning the cutover pre-`EXCHANGE`:**
-  restores `NULL` on the rows the flag wrote into the still-Nullable original and recomputes their `duration`. Which
-  invocation depends on whether a promote parked the successor, because that is the only topological proof a cutover ran
-  on this estate:
+  restores `NULL` on the rows the flag wrote into the still-Nullable original and recomputes their `duration`. **The
+  window is mandatory** (see below). Which invocation depends on whether a promote parked the successor, because that is
+  the only topological proof a cutover ran on this estate:
   ```bash
+  W=(--sentinel-window-from '<flag rolled out, UTC>' --sentinel-window-to '<revert landed everywhere, UTC>')
   # after a stage B/C promote — traces_post_rollback_backup is the proof
-  ./scripts/rollback.sh --database opik --sentinel-repair-only --confirm-flag-reverted
+  ./scripts/rollback.sh --database opik --sentinel-repair-only --confirm-flag-reverted "${W[@]}"
   # no parked successor: abandoned pre-EXCHANGE (incl. after stage A), or finalize.sh has recycled it
-  ./scripts/rollback.sh --database opik --sentinel-repair-only --confirm-flag-reverted --confirm-flag-was-live
+  ./scripts/rollback.sh --database opik --sentinel-repair-only --confirm-flag-reverted --confirm-flag-was-live "${W[@]}"
   ```
   The second asserts the flag was live here, because without the parked successor nothing in the topology or the data
   distinguishes an epoch `end_time` this flag minted from a value a client sent — and the repair rewrites the whole
@@ -1198,8 +1199,18 @@ exists (`Code 60`). That is the second of the two flags the stage comparison tab
    `duration`. The promote made them live again and `finalize.sh` discards the successor's healed copy, so repair them
    here, **after** step 1 has landed on every instance or in-flight writes keep minting more:
    ```
-   ./scripts/rollback.sh --database opik --sentinel-repair-only --confirm-flag-reverted
+   ./scripts/rollback.sh --database opik --sentinel-repair-only --confirm-flag-reverted \
+     --sentinel-window-from '<flag rolled out, UTC>' --sentinel-window-to '<revert landed everywhere, UTC>'
    ```
+   **Both window bounds are required, and there is no safe default.** An epoch `end_time` is not evidence the flag
+   produced it: clients send them, and rows predating the flag hold them. Unbounded, the repair would set those to
+   `NULL` with no way back — the parked successor encodes an absent `end_time` as that same epoch, so nothing holds the
+   original — and the counts would still report success. Measured on an internal environment: the unbounded predicate
+   matched 34 keys across 12 workspaces where only 5 came from the flag window. Take the bounds from when the flag
+   rolled out and when its revert finished landing on every instance; widening is safe, guessing is not. Rows are
+   matched on `created_at` **or** `last_updated_at`, so a row created in the window and a pre-existing row updated in it
+   are both caught. Both bounds are interpreted as UTC regardless of the server's timezone.
+
    It reads the counts first and issues no mutation when they are `0`, restores `NULL` in a single mutation
    (`000004_rollback_sentinel_repair.sql`) which recomputes `duration` as it rewrites each row, then asserts the counts
    reached `0` (`000004_rollback_verify_sentinels.sql`). It is idempotent. A bare `MATERIALIZE COLUMN duration` does
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
@@ -1266,7 +1266,11 @@ Treat a stage B/C rollback as complete only when all of these hold:
       expected, from rows whose `end_time` genuinely precedes `start_time`.
       **"Nothing to repair" is not interchangeable with a completed repair.** It is equally what a wrong window
       produces — bounds in local time being the common case — so check it against the unbounded counts the driver prints
-      beside it before ticking this. This box gates `finalize.sh`, which truncates the parked successor.
+      beside it before ticking this.
+      **`finalize.sh` does not check any of this** — it has no notion of the repair, and reads no marker proving one
+      ran with the right window. This checklist is the only control standing between a wrong-window no-op and
+      `TRUNCATE TABLE traces_post_rollback_backup`, which retires the last reference copy. Treat the box as a human
+      gate, because that is all it is.
 - [ ] **The parked successor still parked** — `traces_post_rollback_backup` retained, not finalized. It is the only copy
       of the post-cutover writes the rollback discarded, and the only thing that makes a retry cheap.
 
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000004_rollback_sentinel_repair.sql
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000004_rollback_sentinel_repair.sql
@@ -1,42 +1,56 @@
 -- runbook traces-local-v2-cutover — ROLLBACK sentinel repair (driven by ../rollback.sh --sentinel-repair-only)
 -- The gate test TracesLocalV2CutoverTest reimplements this statement inline; keep the two in step (see its Javadoc).
 --
--- Restores NULL on the restored ORIGINAL `traces` for rows written while traceColumnsNonNullable was true. The successor
+-- Restores NULL on rows written into the still-Nullable original while traceColumnsNonNullable was true. The successor
 -- schema stores an absent end_time/ttft as a sentinel (epoch / NaN) where the original stores NULL, so rows that landed
 -- through the flag read back as "ended at 1970" / "ttft NaN" once the promote makes the original live again — and the
--- original's MATERIALIZED duration expression computed a large NEGATIVE duration from them. That expression does guard
--- against the epoch, but on `start_time` only: `end_time` it checks for NULL alone, which a sentinel is not.
+-- original's MATERIALIZED duration computed a large NEGATIVE duration from them. That expression does guard against the
+-- epoch, but on `start_time` only: `end_time` it checks for NULL alone, which a sentinel is not.
 --
 -- MATERIALIZE COLUMN duration would NOT fix that: it re-evaluates the same expression against the same sentinel. Only
 -- restoring NULL does, and the mutation recomputes duration as a side effect of rewriting the row.
 --
+-- THE WINDOW IS MANDATORY, AND IS THE ONLY THING THAT MAKES THIS SAFE. An epoch end_time is not by itself evidence the
+-- flag produced it: a client can send one, and rows that predate the flag entirely do. Repairing every match would set
+-- those to NULL and they could not be recovered — the parked successor encodes an absent end_time as the same epoch, so
+-- there is no reference copy to restore from, and the counts would still report success because no sentinel would
+-- remain. Measured on an internal environment: the unbounded predicate matched 34 keys across 12 workspaces where only
+-- 5 came from the flag window; the other 29 carried genuine client-sent values.
+--
+-- So the operator supplies the window the flag was live in, and only rows written inside it are touched. Both arms are
+-- needed, for the same reason the delta insert needs both: a row CREATED in the window is caught by created_at, and a
+-- pre-existing row UPDATED in the window — which is where its sentinel came from — is caught by last_updated_at.
+--
 -- NOT part of stage B/C, deliberately. The flag revert has to land on every backend FIRST or in-flight writes keep
--- minting sentinels behind the repair, and rolling out config is not something these DB-facing scripts do. So this is a
--- separate, later invocation, gated on the operator asserting the revert is live (--confirm-flag-reverted).
+-- minting sentinels behind the repair, and rolling out config is not something these DB-facing scripts do.
 --
 -- ONE ALTER carrying TWO commands, on purpose: neither predicate is on the primary key, so ClickHouse cannot prune parts
--- and a mutation rewrites every one of them. Combining the commands into a single mutation halves that to one pass, which
--- on a large table is the difference that matters in a rollback tail. The cost is atomicity — it needs ALTER UPDATE on
--- BOTH columns and applies neither if one grant is missing (../rollback.sh translates the ACCESS_DENIED).
+-- and a mutation rewrites every one of them. Combining the commands into a single mutation halves that to one pass. The
+-- cost is atomicity — it needs ALTER UPDATE on BOTH columns and applies neither if one grant is missing (../rollback.sh
+-- translates the ACCESS_DENIED).
 --
--- LIMITATION, inherent rather than an omission: an end_time of exactly epoch cannot be told from the sentinel, nor a
--- genuine NaN ttft from the sentinel NaN — the successor schema uses those very values to MEAN "absent", so the
--- distinction does not exist in the data to recover. What bounds the blast radius is the topology guard in
--- ../rollback.sh (this runs only against a restored original with the successor parked) and the counts in
--- 000004_rollback_verify_sentinels.sql, not this statement.
---
--- KEEP IN STEP WITH 000004_rollback_verify_sentinels.sql: same two predicates, same DateTime64 precision 9 (the
--- original's end_time is nanosecond). A check filtered differently from the repair would clear while sentinels remain,
--- or never clear at all. Change one, change both.
 -- Deliberately NOT `ON CLUSTER`, matching the reverse replay beside it. `traces` is a Replicated*MergeTree, so a
 -- mutation entered on one replica reaches the others through the replication log; routing it through the distributed-DDL
--- queue instead would add a second, unrelated wait bounded by `distributed_ddl_task_timeout` (180s by default). This
--- mutation rewrites every part, so on a large table that bound is exceeded routinely, and with the default
--- `distributed_ddl_output_mode = 'throw'` the client raises TIMEOUT_EXCEEDED while the mutation is progressing normally.
--- The driver would then report a healthy repair as failed. Single-shard assumption, same as the reverse replay's.
+-- queue instead would add a second wait bounded by `distributed_ddl_task_timeout` (180s by default), which this
+-- statement exceeds routinely — and with the default `distributed_ddl_output_mode = 'throw'` the client would raise
+-- TIMEOUT_EXCEEDED while the mutation progressed normally. Single-shard assumption, same as the reverse replay's.
+--
+-- KEEP IN STEP WITH 000004_rollback_verify_sentinels.sql: same two predicates, same window on the same two columns, same
+-- DateTime64 precisions. A check scoped differently from the repair either clears while sentinels remain, or never
+-- clears at all.
 ALTER TABLE ${ANALYTICS_DB_DATABASE_NAME}.traces
-    UPDATE end_time = NULL WHERE end_time = toDateTime64('1970-01-01 00:00:00', 9, 'UTC'),
-    UPDATE ttft = NULL WHERE isNaN(ttft)
+    UPDATE end_time = NULL
+        WHERE end_time = toDateTime64('1970-01-01 00:00:00', 9, 'UTC')
+          AND (   (created_at      >= toDateTime64('${SENTINEL_WINDOW_FROM}', 6, 'UTC')
+               AND created_at      <  toDateTime64('${SENTINEL_WINDOW_TO}', 6, 'UTC'))
+               OR (last_updated_at >= toDateTime64('${SENTINEL_WINDOW_FROM}', 6, 'UTC')
+               AND last_updated_at <  toDateTime64('${SENTINEL_WINDOW_TO}', 6, 'UTC'))),
+    UPDATE ttft = NULL
+        WHERE isNaN(ttft)
+          AND (   (created_at      >= toDateTime64('${SENTINEL_WINDOW_FROM}', 6, 'UTC')
+               AND created_at      <  toDateTime64('${SENTINEL_WINDOW_TO}', 6, 'UTC'))
+               OR (last_updated_at >= toDateTime64('${SENTINEL_WINDOW_FROM}', 6, 'UTC')
+               AND last_updated_at <  toDateTime64('${SENTINEL_WINDOW_TO}', 6, 'UTC')))
 -- mutations_sync = 2: wait for the mutation on every replica, so the repair has converged cluster-wide before the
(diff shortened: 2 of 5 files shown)


CometActions:

The window predicate ships with its own gate test in this PR: TracesLocalV2CutoverTest#sentinelRepairRestoresNullAndRecomputesDurationLeavingGenuineNegativesAlone adds an 'outside-window' cohort (epoch end_time + NaN ttft, written 2025-01-01) and asserts both values survive the repair, alongside window-scoped sentinel counts and the 3->5 remaining-negatives control. Drop the created_at/last_updated_at arms and that assertion fails, which is exactly the regression the window exists to prevent — nothing for a Playwright spec to add, and no taxonomy area covers migration runbooks anyway. One thing the test does not pin: it reimplements the ALTER inline rather than executing 000004_rollback_sentinel_repair.sql, so the two can still drift (as its own Javadoc and the .sql header both note) — the file's window clause and precision are only held in step by review.

**Also considered.** The rollback.sh half — the new --sentinel-window-from/--sentinel-window-to flags, the required-with--sentinel-repair-only check, the from < to ordering check and the wrong-mode rejection — has no test at all; the Java gate covers the SQL semantics, not the argument parsing. Flagging it rather than proposing anything: there is no shell-test harness beside these runbook scripts, and an operator CLI is outside what the e2e estate can reach on an OSS install.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/33186337996)

_Re-checked after a push on 28 Aug 15:42 UTC — nothing the verdict depends on changed._

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/rollback.sh
```diff
@@ -74,6 +74,18 @@
...
+#                             [from, to). Both bounds are pinned to UTC in the SQL, so the window means the same thing
+#                             whatever timezone the server runs in.
+#                             The window traceColumnsNonNullable was live in. Only rows written inside it are repaired,
+#                             matched on created_at OR last_updated_at — a row CREATED in the window by the first, a
+#                             pre-existing row UPDATED in it (which is where its sentinel came from) by the second.
+#                             Mandatory because an epoch end_time is not evidence the flag produced it: clients send
+#                             them, and rows predating the flag hold them. An unbounded repair would set those to NULL
+#                             with no way back, since the parked successor encodes an absent end_time as that same
+#                             epoch, and the counts would still report success. Widen rather than guess: a row outside
+#                             the window is simply left alone.
```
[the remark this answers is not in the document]
andrescrz: Applied. That text predates the window and I did not revisit it: the flag authorises a **window-scoped** repair, and rows outside are never touched whether or not it is passed. Corrected in the flag doc and in the runtime NOTE, which said the same thing.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/rollback.sh
```diff
@@ -176,6 +200,19 @@ if [[ "$SENTINEL_REPAIR_ONLY" == "1" ]]; then
...
         exit 2
     fi
+    if [[ -z "$SENTINEL_WINDOW_FROM" || -z "$SENTINEL_WINDOW_TO" ]]; then
+        echo "ERROR: --sentinel-repair-only requires --sentinel-window-from and --sentinel-window-to: the window" >&2
+        echo "       traceColumnsNonNullable was live in. There is no safe default. An epoch end_time is not evidence" >&2
+        echo "       the flag produced it — clients send them, and rows predating the flag hold them — so an unbounded" >&2
+        echo "       repair would set real values to NULL with no way back: the parked successor encodes an absent" >&2
+        echo "       end_time as that same epoch, so nothing holds the original. Take the bounds from when the flag" >&2
+        echo "       rolled out and when its revert finished landing everywhere; widening is safe, guessing is not." >&2
+        exit 2
```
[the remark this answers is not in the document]
andrescrz: Applied. Both printed commands were guaranteed to exit 2 — the same defect as the stage A command earlier in this work, and again from adding a requirement without revisiting what the driver prints.

They now carry placeholder bounds and say plainly that the script cannot know them:

```
--sentinel-window-from '<flag rolled out, UTC>' --sentinel-window-to '<revert landed everywhere, UTC>'
```

Placeholders rather than a guess, because a wrong window is the failure this PR exists to prevent: too narrow leaves damage, and the operator has the rollout times.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/TracesLocalV2CutoverTest.java
```diff
@@ -832,6 +832,9 @@ void sentinelRepairRestoresNullAndRecomputesDurationLeavingGenuineNegativesAlone
         var ended = startTime.plusMillis(100);
         // A real end_time BEFORE start_time: a negative duration owed to the source data, not to the flip.
         var endedEarly = startTime.minusSeconds(5);
+        // The window the flag was live in. Everything above sits inside it; the cohort below deliberately does not.
+        var windowFrom = "2025-03-04 09:00:00";
+        var windowTo = "2025-03-04 11:00:00";
```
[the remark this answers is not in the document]
andrescrz: Applied, and the mutation check confirms it was a real hole rather than a theoretical one.

`insertShapedTrace` tied `created_at`, `last_updated_at` and `start_time` to one value, so no cohort could express the case the window's second arm exists for. Added an overload that sets the two timestamps independently, and three cohorts:

- **created before the window, updated inside it** — repaired, because the window matches either column;
- **exactly at `windowFrom`** — repaired, the bound being inclusive;
- **exactly at `windowTo`** — keeps its sentinels, the bound being exclusive.

Mutation-checked both directions: dropping the `last_updated_at` arm from `repairSentinels` leaves `endTime=2` and fails the gate; changing `windowTo` to `<=` fails the boundary assertion.

Worth recording that my first attempt at that check was invalid — a `perl` substitution that silently matched nothing, so the test "passed" against unmutated code. The second attempt asserts the arms are present before editing them.

Two existing counts moved with the new cohorts (`duration < 0`: 14 before, 6 after). Those are whole-workspace by design, which is the point they make about a negative-duration total being useless as a gate.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/TracesLocalV2CutoverTest.java
```diff
@@ -852,25 +855,32 @@ void sentinelRepairRestoresNullAndRecomputesDurationLeavingGenuineNegativesAlone
         for (int i = 0; i < 2; i++) {
             insertShapedTrace(workspaceId, projectId, "absent", startTime, null, null);
         }
+        // Matches the repair's predicate exactly but was written OUTSIDE the flag window, so its epoch end_time and NaN
+        // ttft are values a client sent, not damage. Unbounded, the repair would set both to NULL with no way back.
+        for (int i = 0; i < 2; i++) {
+            insertShapedTrace(workspaceId, projectId, "outside-window", Instant.parse("2025-01-01T10:00:00Z"),
+                    Instant.EPOCH, Double.NaN);
+        }
```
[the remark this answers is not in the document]
andrescrz: Same gap as the thread above; covered by the cohorts added there — see that reply for the mutation results.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/TracesLocalV2CutoverTest.java
```diff
@@ -1931,8 +1954,10 @@ private SentinelCounts sentinelCounts(String settingsClause) {
                                 duration < 0 AND end_time = toDateTime64('1970-01-01 00:00:00', 9, 'UTC')) AS negative_from_sentinel,
                     uniqExactIf((workspace_id, project_id, id), duration < 0 AND end_time IS NULL) AS stale_duration
                 FROM clusterAllReplicas('{cluster}', %s.traces)
+                WHERE (   (created_at      >= toDateTime64('%s', 6, 'UTC') AND created_at      < toDateTime64('%s', 6, 'UTC'))
+                       OR (last_updated_at >= toDateTime64('%s', 6, 'UTC') AND last_updated_at < toDateTime64('%s', 6, 'UTC')))
                 """
-                .formatted(DATABASE_NAME)
+                .formatted(DATABASE_NAME, windowFrom, windowTo, windowFrom, windowTo)
```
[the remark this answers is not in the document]
andrescrz: Applied. `repairSentinels` already bound the bounds and `sentinelCounts` interpolated them, which is an inconsistency I introduced rather than a considered choice.

Now `toDateTime64(:from, …)` / `:to` with `.bind("from", …)` / `.bind("to", …)`. Only `DATABASE_NAME` stays formatted: it is an identifier, and ClickHouse has no bind form for one — the same reasoning as `verifyReplayPostcondition` beside it.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000004_rollback_sentinel_repair.sql
```diff
@@ -1,42 +1,56 @@
...
+-- clears at all.
 ALTER TABLE ${ANALYTICS_DB_DATABASE_NAME}.traces
-    UPDATE end_time = NULL WHERE end_time = toDateTime64('1970-01-01 00:00:00', 9, 'UTC'),
-    UPDATE ttft = NULL WHERE isNaN(ttft)
+    UPDATE end_time = NULL
+        WHERE end_time = toDateTime64('1970-01-01 00:00:00', 9, 'UTC')
+          AND (   (created_at      >= toDateTime64('${SENTINEL_WINDOW_FROM}', 6, 'UTC')
+               AND created_at      <  toDateTime64('${SENTINEL_WINDOW_TO}', 6, 'UTC'))
+               OR (last_updated_at >= toDateTime64('${SENTINEL_WINDOW_FROM}', 6, 'UTC')
+               AND last_updated_at <  toDateTime64('${SENTINEL_WINDOW_TO}', 6, 'UTC'))),
```
thiagohora, line 47: **The `last_updated_at` arm misses sentinels carried *forward* out of the window — and the window-scoped gate can't see the miss.**

`TraceDAO.UPDATE` (`TraceDAO.java:429-462`) re-inserts a version row that copies `created_at` verbatim, copies `end_time`/`ttft` verbatim when the patch doesn't carry them (`<if(end_time)> … <else> end_time <endif>`), and leaves `last_updated_at` to `DEFAULT now64(6)`.

So take a trace that is:

1. created **before** the window;
2. patched **during** the window with the flag live — sentinel minted;
3. patched again **after** the revert landed, but before the repair runs.

The newest version — the live one under `ReplacingMergeTree(last_updated_at)` — carries a pre-window `created_at` and a post-window `last_updated_at`. It matches neither arm, keeps its epoch `end_time` and negative `duration`, and is invisible to the window-scoped counts.

The part that makes this hard to catch: the repair *does* null the older in-window version, so the postcondition reaches `0` and reports success while the live row stays damaged. The unbounded predicate this PR replaces caught it.

Cheapest fix is documentation rather than SQL: define `--sentinel-window-to` as **the moment the repair runs** (or later), not "when the revert landed". After the revert no new sentinels are minted, but existing ones keep acquiring fresh `last_updated_at`, so extending the upper bound to now closes the hole without widening what can be wrongly nulled — an epoch `end_time` written after the revert isn't flag-minted, but it also can't have a pre-window `created_at` *and* have been touched in the window unless it went through step 2.

Related: `README.md:1211-1212` currently states "a row created in the window **and** a pre-existing row updated in it are both caught", which is the claim this case falsifies.

The batch upsert path is fine, for what it's worth — it discards an epoch `old_trace.end_time` (`TraceDAO.java:325`). Same predicate mirrored at `000004_rollback_verify_sentinels.sql:56` and in the `ttft` arm at lines 50-53 here.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/rollback.sh
```diff
@@ -355,7 +394,8 @@ assert_sentinel_repair_state() {
     fi
     if [[ "$CONFIRM_FLAG_WAS_LIVE" == "1" ]]; then
         echo "NOTE: no parked successor, so proceeding on --confirm-flag-was-live: you assert traceColumnsNonNullable was" >&2
-        echo "      live on THIS estate, and accept that every epoch end_time and NaN ttft in 'traces' becomes NULL." >&2
+        echo "      live on THIS estate, and accept that every epoch end_time and NaN ttft WITHIN THE SUPPLIED WINDOW" >&2
+        echo "      becomes NULL. Rows outside it are not touched." >&2
```
thiagohora, line 399: **This message got the window qualifier; the three that actually report the counts didn't.**

"WITHIN THE SUPPLIED WINDOW … Rows outside it are not touched" is exactly right here. But the counts are now window-scoped too, and the messages reporting them still assert table-wide facts:

- `"Sentinels before repair: …"`
- `"Nothing to repair: no row on any replica carries an epoch end_time or a NaN ttft. No mutation issued."`
- `"Sentinel postcondition OK: … on every replica"`

None of those mean what they say any more — they mean "…inside `[from, to)`".

The failure that follows: an operator takes the flag-rollout timestamps off a dashboard rendering local time instead of UTC. The regex accepts the format, the ordering check passes, the window lands 5h off, `before_end_time`/`before_ttft` come back `0`, and the script prints an unqualified clean bill of health and **exits 0** over fully unrepaired damage. Nothing in the output hints that the window was the problem.

That matters more than a wording nit because of what it clears downstream: it ticks the runbook's last checklist box (README ~1251, which still calls "reported nothing to repair" *equally valid*), and that box gates `finalize.sh` — which truncates the parked successor, the only remaining reference copy.

Two suggestions:
- qualify all three messages with the window, the way this one is;
- print the **unbounded** counts alongside the in-window ones, so a `0` in-window against a non-zero table-wide total is visibly suspicious rather than silently clean. That's one extra query on a path that already runs a full scan, and it turns the dangerous case into an obvious one.

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
thiagohora, line 1: **Outside the diff (README ~1251): the final checklist item now blesses the silent no-op.**

It still reads "printed `Sentinel postcondition OK` (or reported nothing to repair, **which is equally valid**) and exited zero", and describes the gate as `sentinel_end_time` / `sentinel_ttft` / `stale_duration` at `0` — with no mention of the window.

Under the new scoping those two outcomes are no longer equivalent. "Nothing to repair" is now also exactly what a mistyped, non-UTC, or otherwise wrong window produces (see my comment on `rollback.sh:397`). And this is the box that gates `finalize.sh`, which truncates the parked successor.

Suggest qualifying it — something like "…and the window you passed is the one the flag was live in", or dropping "equally valid" so a zero count prompts a second look rather than a tick.

The new material at 1205-1212 is otherwise the strongest part of this PR — the "34 keys across 12 workspaces where only 5 came from the flag window" measurement is exactly the kind of evidence that makes the mandatory-bounds decision self-justifying to whoever reads this at 3am.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/rollback.sh
```diff
@@ -461,6 +501,8 @@ sentinel_counts() {
     local file="$SQL_DIR/000004_rollback_verify_sentinels.sql" sql
     sql="$(cat "$file")"
     sql="${sql//'${ANALYTICS_DB_DATABASE_NAME}'/$DATABASE}"
+    sql="${sql//'${SENTINEL_WINDOW_FROM}'/$SENTINEL_WINDOW_FROM}"
+    sql="${sql//'${SENTINEL_WINDOW_TO}'/$SENTINEL_WINDOW_TO}"
```
thiagohora: **Nit: the function's header comment above now says the opposite of the code.**

It reads `# Same sourcing contract as run_file (versioned .sql, one placeholder here — it takes no cutover window)`. With these two lines the function does three substitutions and very much does take a window.

In a file where these header contracts are load-bearing — they're how the next reader knows what each function is allowed to touch — one that contradicts the four lines beneath it is worse than no comment.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/rollback.sh
```diff
@@ -153,6 +175,10 @@ fi
 # --confirm-flag-reverted asserts a precondition only the sentinel repair has. Reject it elsewhere rather than ignoring
 # it: silently accepting it would let an operator believe a stage B/C run had taken the flag state into account, when
 # stages B/C in fact run BEFORE the revert and print it as their own next step.
+if [[ ( -n "$SENTINEL_WINDOW_FROM" || -n "$SENTINEL_WINDOW_TO" ) && "$SENTINEL_REPAIR_ONLY" != "1" ]]; then
+    echo "ERROR: --sentinel-window-from/--sentinel-window-to belong to --sentinel-repair-only and to no other mode." >&2
+    exit 2
+fi
```
thiagohora: **Nit: this landed between a comment and the checks it documents.**

The block at 175-177 explains the `--confirm-flag-reverted` / `--confirm-flag-was-live` rejections at 182 and 186 ("silently accepting it would let an operator believe a stage B/C run had taken the flag state into account…"). This new window/mode check now sits between them, so the comment reads as heading a condition it has nothing to say about.

Either move the `if` below the two flag checks, or give it its own one-liner — the reasoning is different enough to be worth stating (the window is meaningless outside the repair, rather than misleading).

thiagohora (commented):
Reviewed the full diff (5 files, +214/−66). **The mechanics all check out** — I verified the parenthesisation and operator precedence of both new window predicates (correct, and identical across the two files), the half-open boundary semantics, the `[[ ! a < b ]]` ordering check (parses correctly, and lexicographic compare is safe for the zero-padded fixed-width format the regex enforces), placeholder substitution and injection safety, that `sentinel_counts` is reachable only from the repair mode, that no other script or CI path invokes `--sentinel-repair-only` (so making the flags mandatory breaks no caller), that `traces.created_at` is `DateTime64(9)` and `last_updated_at` is `DateTime64(6)` so the comparison against the scale-6 literal is exact, and the test's per-cohort arithmetic (8 in-window, 14 → 6 negatives).

The risk in this change isn't the SQL, it's what the window can silently miss. Two comments compound into the same failure and I'd want both addressed:

**1. `rollback.sh:397` — the window qualifier was added to one message and not to the three that report the counts.** "Nothing to repair: no row on any replica carries an epoch end_time or a NaN ttft" no longer means that; it means "…inside `[from, to)`". Feed the script local-time bounds from a dashboard and the regex accepts them, the ordering check passes, the counts come back `0`, and the script prints an unqualified clean bill of health and exits 0 over unrepaired damage.

**2. `README.md` ~1251** — the last checklist item still calls "reported nothing to repair" *equally valid*, which is now exactly what a wrong window produces. That's the box gating `finalize.sh`, which truncates the parked successor.

**3. `000004_rollback_sentinel_repair.sql:44` — the predicate has a real gap.** `TraceDAO.UPDATE` copies `created_at` and `end_time`/`ttft` verbatim while letting `last_updated_at` default to `now64(6)`. A trace created pre-window, patched in-window under the flag, then patched again after the revert has a live row with pre-window `created_at` and post-window `last_updated_at` — matching neither arm. It keeps its epoch `end_time`, and because the repair *does* clear the older in-window version, the gate reaches `0` and reports success while the live row stays damaged. The unbounded predicate this PR replaces caught it. Defining `--sentinel-window-to` as *when the repair runs* rather than *when the revert landed* closes it without widening exposure.

The remaining two are trivial: a header comment that now states the opposite of the code beneath it, and a check inserted between a comment block and the checks it documents.

Worth saying: the new material at README 1205-1212 is the strongest part of this PR. The "34 keys across 12 workspaces where only 5 came from the flag window" measurement makes the mandatory-bounds decision self-justifying, which is exactly what that section needs to do.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000004_rollback_sentinel_repair.sql
andrescrz: **The gap is real — verified in the DAO, not taken on trust:**

```sql
<if(end_time)> parseDateTime64BestEffort(:end_time, 9) <else> end_time <endif> as end_time,
created_at,          -- copied verbatim
                     -- last_updated_at absent from the projection, so DEFAULT now64(6)
```

So your three-step trace does end up with a live version matching neither arm, and the repair clearing the older in-window version means the gate reports success over it. Documented, and I removed the README claim that both arms catch every case, since this falsifies it.

**But I am not taking the fix, and I think the reasoning behind it does not hold.** You argue an epoch `end_time` written after the revert "can't have a pre-window `created_at` *and* have been touched in the window unless it went through step 2". It can, by the same mechanism your finding rests on: `end_time` is copied **verbatim**. A trace holding a *genuine* client-sent epoch, created before the window, patched once inside the widened range for any unrelated reason, has a pre-window `created_at`, an in-range `last_updated_at`, and an epoch `end_time` that the flag never touched. Extending `--sentinel-window-to` nulls it irrecoverably.

That is the failure this PR exists to prevent, so the fix trades a detectable false negative for the undetectable false positive. Nothing in the data separates the two cases — which is the real conclusion: **no time-based bound is safe in both directions**, because the DAO breaks the link between when a value was written and when its row was.

Measured on the environment in question, both populations are empty:

```
created_in_window                     5   <- the flag's rows
pre_window_created_but_touched_since  0   <- your case, and the false-positive population
```

So neither error mode has instances here. What I have done instead is state the trade where the bound is chosen, so an operator picks it knowingly rather than inheriting my guess — and paired it with the unbounded comparison from your other comment, which is what makes a too-narrow bound visible instead of silent.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/rollback.sh
andrescrz: **Applied, including the second suggestion, which is the one that actually fixes the failure mode.**

All three messages now name the window, and the driver prints the same aggregates over an all-time range beside them:

```
Sentinels IN WINDOW [<from>, <to>): end_time=0 ttft=0 (of those, serving a negative duration: 0)
Same counts with no window, for comparison: end_time=33 ttft=6
WARNING: nothing matches INSIDE the window, but the table holds end_time=33 ttft=6
         outside it. That is what a wrong window looks like, and bounds in local time rather than
         UTC is the common case. ...
```

Two details worth flagging. The unbounded read comes from the **same** shipped `.sql` file, called with wide bounds, rather than a second copy of the predicates — so the two numbers cannot drift apart. And it is only a warning, not a refusal: on an estate that legitimately holds genuine epoch values (this one holds 33), a non-zero total is the expected state, so failing there would block every correct run.

The postcondition line is qualified too — it claimed `on every replica` without saying *inside what*.

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
andrescrz: **Applied.** "Equally valid" was true before the window existed and became wrong with it, on the box that gates `finalize.sh`.

> **"Nothing to repair" is not interchangeable with a completed repair.** It is equally what a wrong window produces — bounds in local time being the common case — so check it against the unbounded counts the driver prints beside it before ticking this. This box gates `finalize.sh`, which truncates the parked successor.

The item also now requires the window itself to be right, not just the exit code, and says the gate is `0` *inside that window* rather than unqualified.

Thanks for the note on 1205-1212 — the measurement is there precisely because the earlier revision asserted the predicate was safe without one, and a reviewer had to insist on the pre-flight that produced it.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/rollback.sh
andrescrz: Applied. It said "one placeholder here — it takes no cutover window" above a function doing three substitutions, two of them the window. Rewritten to state what it now substitutes, and that the bounds are arguments defaulting to the operator's window — which is what lets the unbounded comparison reuse this one function.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/rollback.sh
andrescrz: Applied — moved below both flag checks, and given the one-liner you suggested, since the reasoning is different: the window is *meaningless* outside the repair, where the flag confirmations are *misleading*.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/rollback.sh
```diff
@@ -164,6 +186,11 @@ if [[ "$CONFIRM_FLAG_REVERTED" == "1" && "$SENTINEL_REPAIR_ONLY" != "1" ]]; then
     echo "       --sentinel-repair-only --confirm-flag-reverted." >&2
     exit 2
 fi
+# The window is meaningless outside the repair -- nothing else reads it -- so reject it rather than ignore it.
+if [[ ( -n "$SENTINEL_WINDOW_FROM" || -n "$SENTINEL_WINDOW_TO" ) && "$SENTINEL_REPAIR_ONLY" != "1" ]]; then
+    echo "ERROR: --sentinel-window-from/--sentinel-window-to belong to --sentinel-repair-only and to no other mode." >&2
+    exit 2
```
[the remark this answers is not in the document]
andrescrz: Not applying, and this is the fourth time across #8046, #8050 and this PR, so let me point at the standing reasoning rather than restate it: `TracesLocalV2CutoverTest`'s class Javadoc scopes the gate to the cutover **SQL logic** and names the driver guards as covered by the staging rehearsal instead; no shell-test harness exists anywhere in the repo; and a stubbed `clickhouse-client` asserts that bash assembled an argv, which is the low-defect half. The argument matrix on this PR was exercised against the real script and reported in the description, which is verification without being CI.

What I will concede has changed: this PR adds real branching — window validation, ordering, the unbounded comparison — and the rehearsal behind it found **two** driver defects the Java gate could not see. So the case for driver-level coverage is now stronger than when I first declined it, and it should be **its own ticket** rather than a repeated review comment on unrelated PRs. The shape I would argue for remains the one I described on #8046: run the real drivers against the ClickHouse container the suite already starts, which would have caught both defects, rather than a fake for whichever flag is newest.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/rollback.sh
```diff
@@ -608,10 +656,24 @@ if [[ "$SENTINEL_REPAIR_ONLY" == "1" ]]; then
         echo "       Treat this as unverified, not as clean. Fix connectivity and re-run — no mutation was issued." >&2
         exit 1
     fi
-    echo "Sentinels before repair: end_time=$before_end_time ttft=$before_ttft (of those, serving a negative duration: $before_negative)"
+    # Same aggregates over an all-time range. A 0 inside the window against a non-zero total is the signature of a wrong
+    # window -- bounds in local time, a typo -- which otherwise reports as a clean estate and exits 0 over unrepaired rows.
+    if ! sentinel_all="$(sentinel_counts '1970-01-01 00:00:00' '2100-01-01 00:00:00')"; then sentinel_all=""; fi
+    read -r all_end_time all_ttft _ _ <<< "$sentinel_all"
+    [[ "$all_end_time" =~ ^[0-9]+$ ]] || { all_end_time="?"; all_ttft="?"; }
```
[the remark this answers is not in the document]
andrescrz: **First half applied.** The `2100-01-01` ceiling was an arbitrary round number in a dataset where far-future timestamps are real — the runbook has a whole section on far-future ids — so it would have hidden exactly the rows the comparison exists to surface. Now `1900..2299`, the `DateTime64` range, with the reason stated so nobody trims it back.

One correction: it could not have produced a false exit 0. That exit is driven by the **in-window** counts; the unbounded read only decides whether the warning prints. A miss there weakens the diagnostic rather than the gate.

**Second half declined**, because the behaviour is already what you are asking for. An unreadable count is normalised to `?`, and the warning fires on `!= "0"` — so `?` triggers it rather than passing silently:

```
[[ "$all_end_time" =~ ^[0-9]+$ ]] || { all_end_time="?"; all_ttft="?"; }
if [[ "$all_end_time" != "0" || "$all_ttft" != "0" ]]; then   # "?" lands here
```

On making it exit non-zero: this read is a diagnostic aid, not a gate. The in-window counts already gate correctly, and failing the run because a *comparison* query died would block a correct repair for a reason that has no bearing on it. That is the opposite trade from `assert_single_shard`, which is fail-closed — there, an uncheckable signal governs what the mutation touches.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/rollback.sh
```diff
@@ -645,7 +707,8 @@ if [[ "$SENTINEL_REPAIR_ONLY" == "1" ]]; then
         echo "         reported success, but treat the repair as unverified: re-run this mode to re-read the counts." >&2
         SENTINEL_CHECK_FAILED=1
     elif (( after_end_time == 0 && after_ttft == 0 && after_stale == 0 )); then
-        echo "Sentinel postcondition OK: end_time=0 ttft=0 stale_duration=0 on every replica."
+        echo "Sentinel postcondition OK: end_time=0 ttft=0 stale_duration=0 on every replica, INSIDE the window"
+        echo "[$SENTINEL_WINDOW_FROM, $SENTINEL_WINDOW_TO). Rows outside it were never in scope."
```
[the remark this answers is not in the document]
andrescrz: **The finding is real** — @thiagohora raised the same case and I verified it in `TraceDAO.UPDATE`, which copies `end_time`/`ttft` verbatim when the patch omits them while `last_updated_at` takes `DEFAULT now64(6)`.

**The fix as proposed is the one thing this PR exists to prevent.** "Across the full table" is the unbounded predicate being removed here: on the environment measured, it matched 34 keys across 12 workspaces of which 5 came from the flag window, and nulling the other 29 would have destroyed genuine client-sent values with no reference copy to restore from — the parked successor encodes an absent `end_time` as that same epoch. Extending to "the latest row for each key" does not narrow that; a genuine-epoch row is a latest row too.

**Your regression suggestion I have taken**, in the form that survives: a cohort with two versions of one key — the in-window one repaired, the live one outside the window keeping its epoch — asserting that the live row stays damaged while the window-scoped counts read `0`.

```java
.as("KNOWN GAP: a sentinel carried forward past the window survives on the LIVE row, and the"
        + " window-scoped counts cannot see it")
```

Asserting a known limitation is deliberate: change the window semantics without addressing this and the test flips, which forces whoever does it to confront the trade rather than rediscover it. It needed a `FINAL`-collapsed count helper, since the claim is about the live row rather than any version.

apps/opik-backend/data-migrations/traces-local-v2-cutover/README.md
```diff
@@ -1237,10 +1260,13 @@ Treat a stage B/C rollback as complete only when all of these hold:
...
-      repair, which is equally valid) and **exited zero**. The gate is `sentinel_end_time`, `sentinel_ttft` and
-      `stale_duration` all at `0`; a residual `duration < 0` count elsewhere in the table is expected, from rows whose
-      `end_time` genuinely precedes `start_time`.
+- [ ] **Sentinel repair applied** — `--sentinel-repair-only` printed `Sentinel postcondition OK` and **exited zero**,
+      **and the window passed is the one the flag was live in, in UTC**. The gate is `sentinel_end_time`,
+      `sentinel_ttft` and `stale_duration` all at `0` *inside that window*; a residual `duration < 0` count elsewhere is
+      expected, from rows whose `end_time` genuinely precedes `start_time`.
+      **"Nothing to repair" is not interchangeable with a completed repair.** It is equally what a wrong window
+      produces — bounds in local time being the common case — so check it against the unbounded counts the driver prints
+      beside it before ticking this. This box gates `finalize.sh`, which truncates the parked successor.
```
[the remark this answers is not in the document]
andrescrz: **Applied — option two, and you are right that the first phrasing overclaimed.** Checked: `finalize.sh` has **zero** references to sentinel, epoch, `isNaN` or repair. It genuinely reads no proof.

I wrote "This box gates `finalize.sh`" one round ago, which implies mechanical enforcement that does not exist. Now:

> **`finalize.sh` does not check any of this** — it has no notion of the repair, and reads no marker proving one ran with the right window. This checklist is the only control standing between a wrong-window no-op and `TRUNCATE TABLE traces_post_rollback_backup`, which retires the last reference copy. Treat the box as a human gate, because that is all it is.

Not taking the persisted marker. It would need durable state that outlives the run and be trusted at the moment of an irreversible truncate — and a stale or hand-written marker fails in the same direction as no marker, while reading as stronger. A checklist honestly labelled as a human gate is safer than a mechanism that can be wrong silently.

thiagohora (approved):

PR #8058, opened 2026-08-28
[OPIK-7315] fix the sentinel repair's no-window comparison, which could never run

andrescrz:
## Details

`rollback.sh --sentinel-repair-only` prints the sentinel counts twice — once inside the operator's window, once over an all-time range — so a zero in-window count beside a non-zero total exposes a wrong window instead of reporting a clean estate over unrepaired rows. The all-time call passed `1900-01-01 .. 2299-12-31`, commented as "the DateTime64 range". That is the range of the DateTime64 type *family*, not of `DateTime64(9)`, which is what `traces.created_at` is — ceiling `2262-04-11 23:47:16.854775807`. The `2299` literal overflows as soon as it is promoted to precision 9 for the comparison, so the check has never once returned a number on any installation.

- **Only the ceiling changes.** The floor stays `1900-01-01 00:00:00` — not because that is the type's minimum (at precision 9 `DateTime64` represents 1677-09-21 onwards, per the docs, and a tick-built value round-trips to it exactly), but because this bound reaches the server as a **string**, and `toDateTime64('<string>', N)` cannot express anything earlier: it clamps the *date* to 1900-01-01 while *keeping the time of day*, so the result is silently **later** than asked — `'1677-09-21 00:12:44'` → `1900-01-01 00:12:44`, and `'1899-12-31 23:59:59'` → `1900-01-01 23:59:59`. The ceiling has no such problem: `'2262-04-11 23:47:16.854775807'` parses to `Int64` max exactly. Upper bound expressible as a string, lower bound not. The `created_at` arm loses nothing from the new ceiling (the column cannot hold a later value); the `last_updated_at` arm (`DateTime64(6)`) nominally loses 2262..2299, unreachable for a column set from `now64()`.
- A failed comparison no longer reads as a passed one: `?` now prints an explicit note that the check did not run and is evidence neither way, and the wrong-window WARNING requires an actual number rather than interpolating `?` into itself.
- Corrected the comment that justified `2299`: the far-future values here are far-future **ids**, which drive the successor's `id_at` partitioning — not `created_at`.

Why it went unnoticed: the driver catches the failure, the run completes, and the gate (the in-window count) is unaffected. The only symptom is a `?` on a line the runbook tells operators to read.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- OPIK-7315

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: the `rollback.sh` change and this description; diagnosis and verification queries run against a live test estate.
- Human verification: diff reviewed locally before the PR was opened.

## Testing

Against a live multi-replica test estate, substituting both bound pairs into the shipped `000004_rollback_verify_sentinels.sql` (read-only, no driver run, so the pinned-revision guarantee for the in-flight cutover window was not disturbed):

- Old bounds `1900-01-01 .. 2299-12-31` → `Code: 407. DB::Exception: Can't compare decimal number due to overflow. (DECIMAL_OVERFLOW)`
- New ceiling (with the floor unchanged) → returns the four counts normally.
- Confirmed by isolating each literal: the floor was always fine; only the ceiling overflows, and only against the `DateTime64(9)` column. The same ceiling against the `DateTime64(6)` column succeeds — which is why one arm of the predicate hid the defect from the other.
- Confirmed the clamping behaviour directly, which is what caught the second commit's floor regression before review.
- `bash -n` clean.

Not run: the driver itself on a real repair. The repair path had already been executed at the pinned revision on that estate, and re-running a modified driver there would have voided the byte-for-byte pin the cutover window requires. The shell changes are the two message branches, exercised by inspection.

## Documentation

No runbook change. The README already tells operators to read the no-window figure; this makes that instruction executable rather than changing it.

### DIFF
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/rollback.sh
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/rollback.sh
@@ -659,14 +659,21 @@ if [[ "$SENTINEL_REPAIR_ONLY" == "1" ]]; then
     # Same aggregates over an all-time range. A 0 inside the window against a non-zero total is the signature of a wrong
     # window -- bounds in local time, a typo -- which otherwise reports as a clean estate and exits 0 over unrepaired rows.
     #
-    # These bounds are the DateTime64(9) representable range, NOT the DateTime64 type-family range (1900..2299) this used
-    # to pass. `created_at` is DateTime64(9), so a 2299 literal is promoted to precision 9 to be compared and raises
+    # Only the CEILING is wrong in the range this used to pass (1900..2299). `created_at` is DateTime64(9), whose ceiling
+    # is 2262-04-11 23:47:16.854775807, so a 2299 literal is promoted to precision 9 for the comparison and raises
     # DECIMAL_OVERFLOW (Code 407) every time -- which made this whole comparison dead code on every real schema, printing
-    # '?' where the operator was told to read a number. The ceiling costs the `created_at` arm nothing, since the column
-    # cannot hold a later value; it costs the `last_updated_at` arm (DateTime64(6)) only 2262..2299, unreachable for a
-    # column set from now64(). Far-future values in this dataset are far-future *ids*, which drive the successor's `id_at`
-    # partitioning -- not `created_at`, which an earlier revision of this comment conflated them with.
-    if ! sentinel_all="$(sentinel_counts '1677-09-21 00:12:44' '2262-04-11 23:47:16.854775807')"; then sentinel_all=""; fi
+    # '?' where the operator was told to read a number.
+    #
+    # The floor stays 1900-01-01. DateTime64's lower bound does NOT narrow with precision the way the upper one does, and
+    # an earlier literal below it is silently CLAMPED to it rather than rejected: toDateTime64('1677-09-21 00:12:44', 9)
+    # evaluates to 1900-01-01 00:12:44, keeping the time of day and so landing 12m44s LATER than the bound it replaced.
+    # A "true precision-9 minimum" is therefore both unnecessary and invisible, which is worse than wrong.
+    #
+    # The ceiling costs the `created_at` arm nothing, since the column cannot hold a later value; it costs the
+    # `last_updated_at` arm (DateTime64(6)) only 2262..2299, unreachable for a column set from now64(). Far-future values
+    # in this dataset are far-future *ids*, which drive the successor's `id_at` partitioning -- not `created_at`, which an
+    # earlier revision of this comment conflated them with.
+    if ! sentinel_all="$(sentinel_counts '1900-01-01 00:00:00' '2262-04-11 23:47:16.854775807')"; then sentinel_all=""; fi
     read -r all_end_time all_ttft _ _ <<< "$sentinel_all"
     [[ "$all_end_time" =~ ^[0-9]+$ ]] || { all_end_time="?"; all_ttft="?"; }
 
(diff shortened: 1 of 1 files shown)


CometActions:
**No test needed here.**

The whole diff is one file: apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/rollback.sh, the one-off cutover rollback driver an operator runs by hand against ClickHouse. The bug is real -- a '2299-12-31' literal promoted to DateTime64(9) raises DECIMAL_OVERFLOW, so the no-window comparison printed '?' on every real schema -- but the only thing that changes is what that script prints on the operator's console under --sentinel-repair-only. The repair gate itself is unchanged (still the IN-WINDOW count), no Opik page or endpoint reads any of this, and it is not something a user or API consumer can reach. There is nothing here a UI or API test could assert, so no exploration run is warranted. If you want this guarded, the right place is a shell test beside the script asserting sentinel_counts returns a number at the new bounds and that the '?' NOTE fires when it does not -- not the e2e estate.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/33200500856)

_Re-checked after a push on 28 Aug 18:43 UTC — nothing the verdict depends on changed._

andrescrz:
Pushed a second commit correcting the first one, before review rather than after.

The first commit moved **both** bounds and called the pair "the DateTime64(9) representable range". Only the ceiling half of that is true. `DateTime64`'s *upper* bound narrows with precision — that is the defect. Its *lower* bound does not; it is `1900-01-01` at every precision, and a literal below it is silently clamped, date-only:

```
toDateTime64('1677-09-21 00:12:44', 9, 'UTC')  ->  1900-01-01 00:12:44.000000000
```

So the "true precision-9 minimum" I passed landed **12m44s later** than the `1900-01-01 00:00:00` it replaced: a marginally narrower floor, arbitrary-looking, defended by a factually wrong comment, and invisible in any output. It changed no result on the estate I tested (nothing predates 2024) — which is precisely why it would have survived review.

Net effect on the line is now a one-value change:

```diff
-    if ! sentinel_all="$(sentinel_counts '1900-01-01 00:00:00' '2299-12-31 23:59:59')"; then sentinel_all=""; fi
+    if ! sentinel_all="$(sentinel_counts '1900-01-01 00:00:00' '2262-04-11 23:47:16.854775807')"; then sentinel_all=""; fi
```

Re-verified: restored floor + corrected ceiling returns the same four counts as the first commit, and the floor alone still selects every row.

andrescrz:
Correction to my previous comment — the objection was right and I was wrong.

I claimed `DateTime64`'s lower bound does not narrow with precision and is 1900-01-01 at every precision. That is false, and the [documented range](https://clickhouse.com/docs/reference/data-types/datetime64) says so: at precision 9 the type represents 1677-09-21 00:12:44 … 2262-04-11 23:47:16. A tick-built value round-trips to the minimum exactly:

```sql
SELECT toUnixTimestamp64Nano(fromUnixTimestamp64Nano(toInt64(-9223372036854775808), 'UTC'))
-- -9223372036854775808, exact
```

What is actually true is about the **string path** this predicate uses, not about the type. `toDateTime64('<string>', N)` cannot express anything before 1900-01-01: it clamps the *date* and *keeps the time of day*, so the result is silently **later** than asked.

```
'1677-09-21 00:12:44'  ->  1900-01-01 00:12:44   (12m44s later than the plain floor)
'1899-12-31 23:59:59'  ->  1900-01-01 23:59:59   (nearly a day later than asked)
'1800-01-01 00:00:00'  ->  1900-01-01 00:00:00
```

The ceiling has no such problem — `'2262-04-11 23:47:16.854775807'` parses to `Int64` max exactly — so the asymmetry is real and now stated: upper bound expressible as a string, lower bound not.

`1f6e4820d7` changes the comment only; the value was already right. Reaching the true minimum would take `fromUnixTimestamp64Nano()` in place of the substituted literal, which this file shares with the windowed gate, where the bound has to stay an operator-supplied datetime. Immaterial regardless — nothing in these columns predates 1900.

Both of my earlier readings came from `toString()` output, which clamps the same way the parser does. That is the trap: the display and the parse agree with each other and both disagree with the type.

thiagohora (approved):

PR #8059, opened 2026-08-28
[NA] [FE] feat: add read-only "View data" action to dataset version history rows

### LINKED ISSUE #8010: [Bug]: switch to a previous version of a dataset

- [ ] Opik Python SDK
- [ ] Opik Typescript SDK
- [x] Opik UI
- [ ] Opik Server
- [ ] Documentation

### Opik version

- Opik UI

### Describe the problem

Is there a way to switch to a previous version of a dataset (to just view)  without restoring it?

I mean: if I want to inspect the data of v1, I currently have to restore v1 first and then restore the latest version again afterward. This creates additional versions just for viewing an older version.

It would be useful to have an option to view a previous version without restoring it or creating new versions. 

Image

- Open any dataset on Opik UI and make a change to it.
- Once the dataset has more than two versions, try to switch back to Version 1.
- There is currently no option in the UI to select or view Version 1 directly.
- The only workaround is to restore Version 1, which creates a new version. You then have to make another restore or change to switch back to the latest version, creating yet another version.

### Error logs or stack trace

_No response_

### Healthcheck results

_No response_

lorenzozanee:
## Details
The dataset Version history tab offers no way to inspect a previous version's records: the only path is restoring that version, which appends a new version just for viewing and forces another restore to get back to the latest one (#8010). This adds a read-only "View data" action to each version's row menu, opening a dialog that lists that version's records (ID, data columns, created date) with pagination and a no-data state. Viewing only calls the existing items endpoint with the version hash (the same pattern the Playground uses), so it never creates new versions or touches draft changes.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #8010
- OPIK-

## AI-WATERMARK

AI-WATERMARK: no

## Testing
- `cd apps/opik-frontend && npx vitest run src/v2/pages-shared/datasets/VersionHistoryTab/` — 5 tests: the dialog fetches on open with the version hash and renders rows read-only, shows the no-data state, and a wiring test drives the row menu (Actions → "View data") into the dialog with the right version.
- `cd apps/opik-frontend && npx tsc --project tsconfig.json --noEmit`
- `cd apps/opik-frontend && npx eslint src/v2/pages-shared/datasets/VersionHistoryTab --max-warnings=0`
- The view path issues no mutations: it only queries `GET /v1/private/datasets/{id}/items?version=<hash>` (the same endpoint/pattern the Playground uses for versioned item lists), so viewing cannot create dataset versions.
- Not run: the full frontend test suite (component-scoped tests only in this environment) and a screen recording — not available here; the dialog reuses the exact DataTable/Dialog layout of GeneratedSamplesDialog.

## Documentation
N/A

### DIFF
--- a/apps/opik-frontend/src/v2/pages-shared/datasets/VersionHistoryTab/VersionRowActionsCell.test.tsx
+++ b/apps/opik-frontend/src/v2/pages-shared/datasets/VersionHistoryTab/VersionRowActionsCell.test.tsx
@@ -1,0 +1,115 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { render, screen, fireEvent, waitFor } from "@testing-library/react";
+import { CellContext } from "@tanstack/react-table";
+import { TooltipProvider } from "@/ui/tooltip";
+import VersionRowActionsCell from "./VersionRowActionsCell";
+import useDatasetItemsList, {
+  UseDatasetItemsListResponse,
+} from "@/api/datasets/useDatasetItemsList";
+import useRestoreDatasetVersionMutation from "@/api/datasets/useRestoreDatasetVersionMutation";
+import { DATASET_ITEM_SOURCE, DatasetVersion } from "@/types/datasets";
+import { DYNAMIC_COLUMN_TYPE } from "@/types/shared";
+
+vi.mock("@/api/datasets/useDatasetItemsList", () => ({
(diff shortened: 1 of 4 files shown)


andrescrz:
Hi @lorenzozanee 

Thanks for the contribution!

Thanks!

PR #8061, opened 2026-08-28
[NA] [DOCS] Drop ask_ollie from the MCP server page

aswynz:
## Details

`comet-ml/opik-mcp#170` disabled `ask_ollie` everywhere behind a kill switch, so the hosted MCP no longer exposes it, but the MCP server page still presented it as an available tool. This removes it and repairs everything that referenced it. Touches `prompt_engineering/mcp-server.mdx` only.

- **Tool table** — dropped the `ask_ollie` row and corrected the count ("six tools" → "five tools").
- **"Ollie & auto-approve" section** — removed entirely. `OPIK_MCP_AUTO_APPROVE` only ever governed the mid-stream writes Ollie performed, so it is moot without the tool.
- **Self-hosted `<Tip>`** — now names `run_experiment` as the only Cloud-only call.
- **Cursor 60s timeout notes** (tip and "Known client limits") — the example of a long call is now a large `run_experiment` instead of an `ask_ollie` turn.
- **Example conversation** — the investigative turn calls `list` + `read`.

Deliberately unchanged: `run_experiment`, still Cloud-only and available; and the web-UI Ollie pages, which document the Opik UI's Explain (Ollie) action and are unaffected by the MCP kill switch.

## Change checklist
- [x] User facing
- [x] Documentation update

## Issues

- `[NA]` — no ticket. Stale-docs cleanup following `comet-ml/opik-mcp#170`.

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: prose edits to a single `.mdx` docs page; no code
- Human verification: author reviewed the full diff; the `auto_approve` claim was checked against the server source — all usage lives in `ask_ollie.py` / `audit.py`

## Testing

Docs-only change, so no unit or integration tests apply and none were run.

- The factual claim behind removing the auto-approve section — that `OPIK_MCP_AUTO_APPROVE` governs nothing once `ask_ollie` is gone — was checked by grepping the server for `auto_approve`; it appears only in `ask_ollie.py` and `audit.py`.
- Remaining verification is the Fern docs build on this PR, which must render the page without broken anchors (the removed section had no inbound links from this page).

## Documentation

This PR is the documentation change. No follow-up docs work required.

PR #8062, opened 2026-08-28
[NA] [DOCS] Publish the MCP read/write vocabulary and add a Security section

aswynz:
## Details

Two additions to `prompt_engineering/mcp-server.mdx`: a published vocabulary for the dispatcher tools, and a Security section. Docs only.

**The read and write vocabulary.** `read`, `list`, and `write` are universal dispatchers, but the page never listed their legal values — it pointed at the `schema` tool and the GitHub README. That leaves the real surface undiscoverable from the docs: a reader cannot tell whether the thing they want is possible, and an agent has to call `schema` and guess. Adds two tables, both sourced from the server registries (`read_list/registry.py`, `writes/registry.py`):

- **Entities** (`read` / `list`) — 9 types, with which support `read` vs `list` and how each is scoped (`project_id`, `test_suite_id`, `prompt_id`).
- **Operations** (`write`) — 12 operations, each with what it does.

The vocabulary is deployment-independent: every entity and operation works the same on Opik Cloud and on self-hosted.

**Security.** A trace holds the prompt and completion from a production app, so it contains text end users wrote. Reading one pipes that text into a context that also holds the `write` tool — the prompt-injection base case. Adds a Security section above the deployment notes stating that trace content is untrusted input, with three rules: prefer read-only work while investigating (the `opik-diagnose` and `opik-explain` skills are read-only by design), scope the credential to the narrowest workspace rather than an org-wide key, and verify any conclusion drawn from trace text against code or span data before a write.

## Rebuilt on main

This PR was originally stacked on `aswynz/docs/drop-ask_ollie-from-mcp` (#8061). That PR was closed rather than merged, and `main` has since moved:

- The tool table on `main` now lists **`read_skill`** and no longer lists `run_experiment`, and a new paragraph frames running an evaluation end to end as the `opik-evaluate` skill's job rather than a tool's.
- The branch has been rebuilt directly on `main` and the base retargeted, so no retarget is pending.
- The earlier draft's line "only the separate `run_experiment` tool is Cloud-only" has been dropped. It referenced a tool the page no longer documents. The lead-in now just states that the read/write vocabulary works the same on Cloud and self-hosted, which is the claim that actually matters here.
- The vocabulary is introduced *after* the skill-pack paragraph, so `main`'s existing text is untouched and this is a pure insertion.

## Change checklist
- [x] User facing
- [x] Documentation update

## Issues

- `[NA]` — no ticket resolved by this PR.
- Related, not resolved here: server-side read-only mode (OPIK_8165, under OPIK_7624) is the enforcement this Security section currently describes by convention.

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: prose and two reference tables in a single `.mdx` docs page; no code
- Human verification: author reviewed the full diff; both tables were re-verified against the registries on `main` at the time of the rebuild

## Testing

Docs-only change, so no unit or integration tests apply and none were run.

- Both tables were re-checked against `main` of `comet-ml/opik-mcp`: `read_list/registry.py` still defines exactly the 9 entities, with `span` fetch-only, `test_suite_item` and `prompt_version` list-only, and `trace`/`thread` list requiring `project_id`. `writes/registry.py` still defines exactly the 12 operations, unchanged in name.
- The `opik-diagnose` and `opik-explain` SKILL.md files on `main` both state "This skill is read-only by design," and neither invokes the `write` tool — so the read-only claim in the Security section holds.
- Remaining verification is the Fern docs build on this PR rendering both tables and the new section.

Not covered: the tables are hand-maintained, so nothing prevents them drifting from the registries — this rebuild is itself an instance of that cost. A CI generator deriving them at build time would close it, and is worth doing separately. Note also that in-flight work on the MCP server (an `annotation_queue` entity and additional write operations) will expand both tables when it lands.

## Documentation

This PR is the documentation change. No follow-up docs work required.

### DIFF
--- a/apps/opik-documentation/documentation/fern/docs-v2/prompt_engineering/mcp-server.mdx
+++ b/apps/opik-documentation/documentation/fern/docs-v2/prompt_engineering/mcp-server.mdx
@@ -215,8 +215,68 @@ Running an evaluation end to end is the skill pack's job, not a tool's: the
 `opik-evaluate` skill drives the Opik SDK, and the MCP tools record and read the
 results. That is why the command above installs both.
 
-To see a payload shape yourself, ask **"show me the schema for trace.create"** —
-or read the [full list](https://github.com/comet-ml/opik-mcp#tools).
+`read`, `list`, and `write` are dispatchers, so their real surface is the set of
+entity types and operations below, not the tool count.
+
+### The read and write vocabulary
+
+`read` and `list` take an **entity type**; `write` takes an **operation** plus its
+data. These are the legal values, and they work the same on Opik Cloud and on
+self-hosted.
+
+**Entities (`read` / `list`)**
+
+| Entity | `read` | `list` | Notes |
+|---|---|---|---|
+| `project` | yes | yes | Metadata plus stats (trace count, last activity). |
+| `trace` | yes | yes | `read` returns the trace and its span tree. `list` needs `project_id`. |
+| `span` | yes | no | Inputs, outputs, metadata, timing, feedback scores. |
+| `thread` | yes | yes | Conversation thread plus messages. Needs project scope (a thread link/URI, or `project_id`). |
+| `experiment` | yes | yes | Status and summary scores. |
+| `prompt` | yes | yes | `read` returns the prompt and all its versions. |
+| `prompt_version` | list-only | yes | Pass `prompt_id`. Use `read('prompt', id)` to get the prompt and versions in one call. |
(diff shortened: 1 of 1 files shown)


CometActions:
**No test needed here.**

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/34377077838)

_Re-checked after a push on 09 Sep 16:32 UTC._

aswynz:
No longer relevant since we have updated mcp-server/tools.mdx that covers the r/w ops and security faq that isn't quite right yet

PR #8063, opened 2026-08-29
[NA] [BE] fix: delete a project's traces from ClickHouse when the project is deleted

### LINKED ISSUE #7826: [Bug]: Deleting Project from Opik Server UI doesnt not deleting its traces and spans from Clickhouse

- [ ] Opik Python SDK
- [ ] Opik Typescript SDK
- [ ] Opik UI
- [x] Opik Server
- [ ] Documentation

### Opik version

Latest

### Describe the problem

Hi Team,

I have tried deleting the project from Opik server and it shows it will delete all the data associated with the project. But once the project is delete it is not visible from GUI which is expected, but I can see still its trcaes and spans count in clickhouse DB.

_No response_

### Error logs or stack trace

_No response_

### Healthcheck results

_No response_

lorenzozanee:
## Details

Deleting a project removed only the MySQL `projects` row, so its traces and spans stayed queryable in ClickHouse forever (reported in #7826: `select count(*) from traces where project_id = '<deleted-id>'` returned 8 traces / 57 spans after UI deletion). There was no path from project deletion to trace cleanup at any layer.

After the MySQL delete succeeds, `ProjectService` now posts a `ProjectsDeleted` event. The new `ProjectDeletedListener` streams the project's trace ids from ClickHouse in 10k batches and routes each batch through the regular `(project_id, trace_id)` delete path, so the `TracesDeleted` cascade (spans, feedback scores, comments, attachments) and the deletion-events bridge capture behave exactly as for user-initiated trace deletes.

- Trace-thread rows (`trace_threads` / `project_trace_threads`) are not cleaned by any delete path today, including per-trace deletes; that cleanup is a pre-existing gap and an explicit follow-up.
- Alerts/dashboards referencing the deleted project are workspace-level lifecycle, intentionally out of scope.
- The cascade is fire-and-forget (AsyncEventBus), the same semantics as every existing deletion listener.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #7826

## AI-WATERMARK

AI-WATERMARK: no

## Testing

Commands run from `apps/opik-backend`:

- `mvn test -Dtest='ProjectServiceImplTest,ProjectDeletedListenerTest,TraceServiceImplTest'` — 17 passed, 0 failed
- `mvn test -Dtest='TraceDeletionEventArchTest,SpanDeletionEventArchTest,TraceMutationRoutingArchTest'` — 5 passed (no bypass of the deletion-events capture or mutation-table routing)
- `mvn spotless:apply`

Scenarios covered by the new tests:

- Project exists → exactly one `ProjectsDeleted` with the id/workspace/user; missing project → no event; batch delete → one event with all ids; empty ids → no event.
- Listener calls `deleteByProjectId` once per project id with the workspace/user reactor context, and survives a per-project error while processing the rest.
- 10001 traces → two 10k-bounded deletes with exact id sublists, two `TracesDeleted` events, two bridge captures (capture flag on); flag off → no bridge writes.

Not run: testcontainers/E2E against a live stack (no Docker available in this environment); full `mvn test` suite.

## Documentation

N/A — no public API, configuration option or documented behavior changes.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/events/ProjectsDeleted.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/events/ProjectsDeleted.java
@@ -1,0 +1,22 @@
+package com.comet.opik.api.events;
+
+import com.comet.opik.infrastructure.events.BaseEvent;
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.experimental.Accessors;
+
+import java.util.Set;
+import java.util.UUID;
+
+@Getter
+@Accessors(fluent = true)
+public class ProjectsDeleted extends BaseEvent {
+
+    private final @NonNull Set<UUID> projectIds;
+
+    public ProjectsDeleted(@NonNull Set<UUID> projectIds, @NonNull String workspaceId,
+            @NonNull String userName) {
+        super(workspaceId, userName);
+        this.projectIds = projectIds;
+    }
+}
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/ProjectDeletedListener.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/ProjectDeletedListener.java
@@ -1,0 +1,73 @@
+package com.comet.opik.api.resources.v1.events;
+
+import com.comet.opik.api.events.ProjectsDeleted;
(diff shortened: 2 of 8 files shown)


andrescrz:
Hi @lorenzozanee 

Thanks for the contribution!

Thanks!

PR #8069, opened 2026-08-30
[NA] [BE] fix: register cohere, novita and cloudflare as canonical providers so their model prices load

Anuj7411:
## Details

`CostService.PROVIDERS_MAPPING` did not include `cohere`, `novita` or `cloudflare`, so every priced model those providers ship in `model_prices_and_context_window.json` was dropped at load time and their spans fell back to zero cost. This registers all three as canonical providers (identity mapping, same as the existing entries) so their prices load. None of the three publishes cache rates today, so they route through the default `textGenerationCost` calculator.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude
- Scope: full change (mapping entries plus regression tests)
- Human verification: cost math checked against the exact per-token rates in model_prices_and_context_window.json; provider-key resolution traced through buildRuntimeKey and findModelPrice to confirm the same path used by the already-merged snowflake and deepinfra registrations

## Testing

Added three regression tests in `CostServiceTest` (`calculateCostHandlesCohereModels`, `calculateCostHandlesNovitaModels`, `calculateCostHandlesCloudflareModels`), each asserting a real model from each provider now returns a non-zero cost equal to prompt and completion tokens billed at the provider's configured rates. The tests fail before the mapping entries are added (cost resolves to zero) and pass after. Run with `mvn test -Dtest=CostServiceTest` in `apps/opik-backend`.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/cost/CostService.java
@@ -56,7 +56,10 @@ public class CostService {
             Map.entry("sambanova", "sambanova"),
             Map.entry("nebius", "nebius"),
             Map.entry("snowflake", "snowflake"),
-            Map.entry("deepinfra", "deepinfra"));
+            Map.entry("deepinfra", "deepinfra"),
+            Map.entry("cohere", "cohere"),
+            Map.entry("novita", "novita"),
+            Map.entry("cloudflare", "cloudflare"));
 
     // Online evaluation (and OTel ingestion) resolve models to LlmProvider serialized values whose names
     // differ from the canonical price-table vocabulary. Normalize those to the single canonical provider
--- a/apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/domain/cost/CostServiceTest.java
@@ -1062,4 +1062,52 @@ class CostServiceTest {
                                 "original_usage.prompt_tokens_details.cached_tokens", 300),
                         "0.005709"));
     }
+
+    /**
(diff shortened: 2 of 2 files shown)


andrescrz:
Hi @Anuj7411 

Thanks for the contribution!

Thanks!

aadereiko:
Thanks for this — closing in favour of #8548, which carries it forward.

All six of your `PROVIDERS_MAPPING` PRs edit the tail of the same `Map.ofEntries(...)` literal, so they conflict with one another: merging any one of them put the other five straight back into conflict, and all six were sitting `CONFLICTING` at the same time. Rather than ask you for six sequential rebases, I folded them into a single change.

**Nothing was dropped.** #8548 contains all 19 provider entries and every one of your test methods, unrenamed — your method names were already unique across the branches, so they coexist. The commit carries `Co-authored-by: Anuj7411 <ojhaanuj224@gmail.com>` and the PR body credits each source PR by number.

One value needed refreshing: #7894's `watsonx/openai/gpt-oss-120b` assertion expected `0.00027` from rates `1.5e-07 / 6e-07`, but the bundled price file now carries `1.59e-07 / 6.36e-07`, so it is `0.0002862` in #8548. The rates moved after you wrote it — nothing was wrong with the PR at the time.

Verified before opening: `mvn -Dtest=CostServiceTest test` → 147 tests, 0 failures, with all 19 of your new cases executing; 50 map entries with no duplicate key; every referenced model resolves in `model_prices_and_context_window.json`.

The analysis in these PRs was consistently good — the diagnosis that unmapped `litellm_provider` rows get dropped by `buildModelPrice` was right every time. The only problem was the shape of the target: `PROVIDERS_MAPPING` is an append-only literal, so parallel PRs against it always collide. If you pick up more providers, a single PR adding them together will go through much faster.

PR #8074, opened 2026-08-31
[OPIK-8177] [BE] perf: prune experiment_items scan in dataset experiment summary

JetoPistola:
## Details

image

`FIND_EXPERIMENT_SUMMARY_BY_DATASET_IDS` filtered `experiment_items` only by `workspace_id`, putting the `dataset_id` restriction on the joined `experiments` side. Since `experiment_items` has no `dataset_id` column, nothing could be pruned — every experiment item in the workspace was read and probed against the join hash table, so cost scaled with total workspace experiment volume rather than with the dataset being fetched. Fetching a dataset that never had an experiment run against it still paid for the whole workspace.

This adds an `ei.experiment_id IN (SELECT id FROM experiments WHERE workspace_id = … AND dataset_id IN …)` predicate. `experiment_id` is the second column of the sort key `(workspace_id, experiment_id, dataset_item_id, trace_id, id)`, so ClickHouse can prune granules on it directly. The join to `experiments` is retained solely to project `dataset_id` for the `GROUP BY`.

- Query plan verified with `EXPLAIN indexes = 1`: the planner emits a `CreatingSets` node and the `experiment_items` primary-key condition changes from `Keys: workspace_id` to `Keys: workspace_id, experiment_id`.

## Change checklist
- [x] User facing
- [x] Documentation update

## Issues

- Resolves OPIK-8177

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: full implementation, tests, and performance measurement

## Testing

Four integration tests added in `DatasetsResourceTest.ExperimentSummaryEnrichment`, exercised through the **list** endpoint (`getDatasets`) rather than only the single-dataset path:

- experiments spread across many experiment ids
- dataset with zero experiments in a workspace where other datasets are busy
- batch mixing datasets with and without experiments
- cross-workspace isolation (same dataset name in two workspaces)

Commands run:

```
mvn -o spotless:check
mvn -o compile -DskipTests
mvn -o test -Dtest='DatasetsResourceTest$ExperimentSummaryEnrichment'
```

Result: 4/4 new tests pass; the 12 pre-existing `GetDataset` tests in the same class also ran green, confirming no regression in existing enrichment behavior. The full backend suite was not run locally — CI covers it.

### Performance measurement

Local single-node ClickHouse 26.3.16.16, seeded with 2,000,300 `experiment_items` across 403 experiments / 41 datasets, merged to a single part (257 granules). Experiment ids scatter across the sort order as real UUIDs would.

| Case | `SelectedMarks` before | `SelectedMarks` after |
|---|---|---|
| Dataset with no experiments | 246 | 1 |
| Dataset with 3 experiments | 246 | 5 |
| Dataset with 10 experiments | 246 | 18 |
| Mixed batch (none + 3 + 10) | 246 | 20 |
| 22 experiments scattered across sort range | 246 | 17 |

The old shape reads 246 marks regardless of which dataset is requested; the new shape's marks scale with the requested dataset's experiments.

A note on metrics: `read_rows` understates the old shape's cost for the empty-dataset case (2,557 on both), because ClickHouse's runtime join filter discards rows early even though the marks are still selected and scanned. `SelectedMarks` is the reliable signal here.

Results were identical between the old and new query shapes in every case, including the empty dataset (no row emitted, unchanged) and the mixed batch. Cross-workspace isolation was checked with the same `dataset_id` present in two workspaces: 10 vs 7 experiments, no leakage in either direction. Benchmark data was removed afterwards.

Timings came from a dev instance with a warm cache, so they are indicative rather than production latency predictions.

## Documentation

N/A — internal query optimization with no user-facing or API surface change. The rationale is captured in a code comment above the query constant.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/ExperimentItemDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/ExperimentItemDAO.java
@@ -496,6 +496,14 @@ class ExperimentItemDAO {
             ;
             """;
 
+    /**
+     * The 'experiment_id IN (...)' predicate is what restricts the 'experiment_items' scan. Without it, the only
+     * predicate on the large table is 'workspace_id': 'experiment_items' has no 'dataset_id' column, so a
+     * 'dataset_id' filter on the joined side can only discard rows after they have been read, making the cost scale
+     * with total workspace experiment volume rather than with the requested datasets. 'experiment_id' is the second
+     * column of the sort key '(workspace_id, experiment_id, dataset_item_id, trace_id, id)', so filtering on it
+     * directly lets ClickHouse prune granules. The join is retained solely to project 'dataset_id' for the grouping.
+     */
     private static final String FIND_EXPERIMENT_SUMMARY_BY_DATASET_IDS = """
             SELECT
                 e.dataset_id,
@@ -503,8 +511,14 @@ class ExperimentItemDAO {
                 max(ei.last_updated_at) as most_recent_experiment_at
             FROM experiment_items ei
             JOIN experiments e ON ei.experiment_id = e.id AND e.workspace_id = ei.workspace_id
-            WHERE e.dataset_id in :dataset_ids
-            AND ei.workspace_id = :workspace_id
+            WHERE ei.workspace_id = :workspace_id
+            AND ei.experiment_id IN (
+                SELECT id
+                FROM experiments
+                WHERE workspace_id = :workspace_id
+                AND dataset_id IN :dataset_ids
+            )
+            AND e.dataset_id in :dataset_ids
             GROUP BY
                 e.dataset_id
             SETTINGS log_comment = '<log_comment>'
--- a/apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetsResourceTest.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/DatasetsResourceTest.java
@@ -2045,6 +2045,137 @@ class DatasetsResourceTest {
     }
 
     @Nested
+    @DisplayName("Experiment Summary Enrichment:")
+    @TestInstance(TestInstance.Lifecycle.PER_CLASS)
+    class ExperimentSummaryEnrichment {
+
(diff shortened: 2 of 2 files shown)


CometActions:

The rewrite looks semantics-preserving — the new `ei.experiment_id IN (SELECT id FROM experiments WHERE workspace_id = ... AND dataset_id IN ...)` is already implied by the retained join plus `e.dataset_id in :dataset_ids`, so it can only prune, not change, the result — and your new `ExperimentSummaryEnrichment` nested class in `DatasetsResourceTest.java` pins exactly the cases a wrong predicate would break: 12 experiments all counted, a dataset whose workspace-mates belong to other datasets staying at 0/null, a mixed batch getting per-dataset summaries, and a same-named dataset in another workspace not leaking in. Those run against real ClickHouse, and the pre-existing assertions at DatasetsResourceTest.java:1750/1987/2952 already cover the single-experiment case, so a dropped or leaked experiment fails CI before it could reach the Datasets list. Nothing asserts the pruning itself, but a granule-count/perf assertion is a load-tier concern, not something an e2e spec would carry.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/33362587672)

CometActions:

### Access Information
- **URL:** https://pr-8074.dev.comet.com
- **Namespace:** pr-8074
- **Version:** 2.2.45-8074-merge-3148
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-8074%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

thiagohora (approved):

PR #8076, opened 2026-08-31
[OPIK-8176] [BE] perf: run the four dataset enrichment queries concurrently

JetoPistola:
## Details

image

`enrichDatasetWithAdditionalInformation` issues four independent lookups — experiment summary, dataset-item summary and optimization summary (ClickHouse), plus the latest dataset versions (MySQL). None depends on another, but each block terminated in `.toStream()`, draining one publisher to completion before the next statement ran, so enrichment cost `T1 + T2 + T3 + T4` instead of `max(...)`. This multiplies on the dataset list endpoints, where enrichment runs for a whole page, and on `get_or_create_dataset`, which the Python SDK calls at the start of most upload and evaluation scripts.

The four lookups are now zipped so they are in flight together and awaited as one. This is a latency change only — the same queries run and the returned `Dataset` objects are unchanged. Measured benefit is a fixed ~24 ms (≈1.5x on small pages, fading to parity by ~88 rows); see the before/after table below.

- Request context is resolved once on the request thread and written into the shared reactor context. `RequestContext` is `@RequestScoped` and thread-bound, so no concurrent subscription resolves it from a worker thread.
- `computeIfAbsent` is replaced with `getOrDefault` in the mapping stream — the previous form mutated the shared summary maps while iterating them. Both yield the same zero-count/null-timestamp defaults.
- `defaultIfEmpty` guards the zip: `Mono.zip` emits nothing if any source completes empty, which would make `block()` return null and NPE rather than yielding an absent-versions result. `collectMap` always emits a map, but `Mono.fromCallable` does not.
- In `find()`, enrichment moved outside the read-only transaction so the MySQL connection is not held open across the three ClickHouse round-trips enrichment makes. (An earlier revision of this description claimed the old nested lookup opened a second connection — it did not: `TransactionTemplate.inTransaction` checks `UnitManager.isUnitStarted()` and the handle lives in a `ThreadLocal`, so the nested call reused the same handle. Corrected per review.) This is **not** visibility-neutral for `find()`: the page rows and the dataset versions were previously read in one `READ_ONLY` transaction and are now two, so a version written between them is observable where it was not before. Benign for count/summary metadata. The other three call paths (`findById`, `findByNameDetailed`, and the experiments-only `find()` branch) genuinely did already enrich outside any transaction.
- Duplicate-key collection stays loud: `Collectors.toMap` (throws) rather than `Flux.collectMap` (last-wins), matching the serial code it replaces.

## Change checklist
- [x] User facing
- [x] Documentation update

## Issues

- Resolves #
- OPIK-8176

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: full implementation
- Human verification: code review pending

## Testing

- `mvn compile -DskipTests` — passes
- `mvn spotless:check` — passes
- `mvn test -Dtest=DatasetServiceEnrichmentTest` — 11/11 pass

`DatasetServiceEnrichmentTest` is new and covers the acceptance criteria:

- All four summaries present; none present; a mixed multi-dataset page asserting row order and per-row values; an empty page.
- An induced failure in one lookup propagates instead of degrading into a silently empty summary.
- Request context is asserted present in each concurrent subscription, and the version lookup is asserted bound to the request workspace off the request thread.
- The three ClickHouse lookups are asserted to be in flight simultaneously via a shared latch, and a duplicate `dataset_id` is asserted to fail loudly rather than silently drop a row.

Every guard is mutation-checked — each of these was reintroduced deliberately and fails the suite: cross-row summary bleed (2 tests), a removed `contextWrite` (1 test), reverting to serial enrichment (the overlap test times out), deleting `defaultIfEmpty` (NPE on `Tuple4.getT1()`), and swapping `Collectors.toMap` back to `Flux.collectMap` (the duplicate-key test).

CI is green on clean runners: 32/32 integration groups and no failing checks. The full `DatasetsResourceTest` suite cannot be used as a local signal on this machine — it fails ~108/433 with ClickHouse `failed to respond` errors, but a clean `origin/main` checkout with the change stashed fails 109/433 the same way, and individual classes pass in isolation with the change applied. The cause is local Docker memory pressure, not this branch.

Both acceptance criteria that needed a running environment are now closed, measured on the ad hoc deployment and recorded on OPIK-8176:

- **Pool behaviour** — full-page requests fanned out to 240 concurrent DB operations (60 simultaneous requests x 4 lookups) with zero non-200s and smooth latency growth, so no starvation.
- **Output correctness under load** — 20 concurrent fetches of the same mixed page returned byte-identical payloads across every enrichment field.
- **Before/after latency** — measured against a matched serial deployment (identical seeding, interleaved requests, network floor subtracted). The saving is a **fixed ~24 ms**, not a proportional one: 68.0 → 43.6 ms at 1 row (1.56x), 67.1 → 43.7 ms at 10 rows (1.54x), 213.2 → 188.7 ms at 40 rows (1.13x), and indistinguishable at 88 rows (269.0 → 274.2 ms). Collapsing three sequential round-trips into one wait removes latency, not work, so the benefit is concentrated on the small pages that dominate real traffic — the default page size is 10, and `get_or_create_dataset` enriches a single dataset — and fades on large pages where per-row query cost dominates the max.

## Documentation

N/A — internal service refactor with no API, schema, or user-facing surface change.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetService.java
@@ -737,14 +737,15 @@ class DatasetServiceImpl implements DatasetService {
                 .contextWrite(ctx -> AsyncUtils.setRequestContext(ctx, userName, workspaceId))
                 .block();
 
-        Map<UUID, ExperimentSummary> experimentSummary = enrichmentData.getT1();
+        Map<UUID, ExperimentSummary> experimentSummaryMap = enrichmentData.getT1();
         Map<UUID, DatasetItemSummary> datasetItemSummaryMap = enrichmentData.getT2();
         Map<UUID, OptimizationDAO.OptimizationSummary> optimizationSummaryMap = enrichmentData.getT3();
         Map<UUID, DatasetVersion> latestVersionsByDatasetId = enrichmentData.getT4();
 
         return datasets.stream()
                 .map(dataset -> {
-                    var resume = experimentSummary.getOrDefault(dataset.id(), ExperimentSummary.empty(dataset.id()));
+                    var experimentSummary = experimentSummaryMap.getOrDefault(dataset.id(),
+                            ExperimentSummary.empty(dataset.id()));
                     var datasetItemSummary = datasetItemSummaryMap.getOrDefault(dataset.id(),
                             DatasetItemSummary.empty(dataset.id()));
                     var optimizationSummary = optimizationSummaryMap.getOrDefault(dataset.id(),
@@ -762,10 +763,10 @@ class DatasetServiceImpl implements DatasetService {
                     }
 
                     return dataset.toBuilder()
-                            .experimentCount(resume.experimentCount())
+                            .experimentCount(experimentSummary.experimentCount())
                             .datasetItemsCount(itemsCount)
                             .optimizationCount(optimizationSummary.optimizationCount())
-                            .mostRecentExperimentAt(resume.mostRecentExperimentAt())
+                            .mostRecentExperimentAt(experimentSummary.mostRecentExperimentAt())
                             .mostRecentOptimizationAt(optimizationSummary.mostRecentOptimizationAt())
                             .latestVersion(DatasetVersionMapper.INSTANCE.toDatasetVersionSummary(latestVersion))
                             .build();
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetService.java
@@ -560,8 +560,11 @@ class DatasetServiceImpl implements DatasetService {
         }
 
         // For now, we are not going to use the criteria.withExperimentsOnly() method due to the migration.
-        // Enrichment runs outside the transaction: it opens its own MySQL connection for the dataset versions,
-        // which would otherwise be held concurrently with this one for the same request.
+        // Enrichment runs outside the transaction so the MySQL connection is not held open across the three
+        // ClickHouse round-trips it makes. Trade-off: the page rows and the dataset versions were previously
+        // read in one READ_ONLY transaction and are now two, so a version written between them is observable
+        // where it was not before. Benign for count/summary metadata, and only find() is affected -- findById,
+        // findByNameDetailed and the experiments-only branch already enriched outside any transaction.
         DatasetPage rawPage = template.inTransaction(READ_ONLY, handle -> {
 
             var repository = handle.attach(DatasetDAO.class);
@@ -722,13 +725,17 @@ class DatasetServiceImpl implements DatasetService {
         String workspaceId = requestContext.get().getWorkspaceId();
         String userName = requestContext.get().getUserName();
 
+        // collect(...) with Collectors.toMap rather than Flux.collectMap: collectMap is last-wins, whereas the
+        // serial code this replaces threw on a duplicate dataset_id. All three queries GROUP BY dataset_id so
+        // duplicates should not occur; keeping the loud form means a query change that broke that assumption
+        // fails instead of silently dropping one row's summary.
         var enrichmentData = Mono.zip(
                 experimentItemDAO.findExperimentSummaryByDatasetIds(ids)
-                        .collectMap(ExperimentSummary::datasetId, Function.identity()),
+                        .collect(toMap(ExperimentSummary::datasetId, Function.identity())),
                 datasetItemDAO.findDatasetItemSummaryByDatasetIds(ids)
-                        .collectMap(DatasetItemSummary::datasetId, Function.identity()),
+                        .collect(toMap(DatasetItemSummary::datasetId, Function.identity())),
                 optimizationDAO.findOptimizationSummaryByDatasetIds(ids)
-                        .collectMap(OptimizationDAO.OptimizationSummary::datasetId, Function.identity()),
+                        .collect(toMap(OptimizationDAO.OptimizationSummary::datasetId, Function.identity())),
                 // defaultIfEmpty guards the zip: a Mono that completes empty makes zip emit nothing at all,
                 // which would turn an absent-versions result into a null and NPE below.
                 Mono.fromCallable(() -> fetchLatestVersionsByDatasetIds(ids, workspaceId))
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetService.java
@@ -729,8 +729,11 @@ class DatasetServiceImpl implements DatasetService {
                         .collectMap(DatasetItemSummary::datasetId, Function.identity()),
                 optimizationDAO.findOptimizationSummaryByDatasetIds(ids)
                         .collectMap(OptimizationDAO.OptimizationSummary::datasetId, Function.identity()),
+                // defaultIfEmpty guards the zip: a Mono that completes empty makes zip emit nothing at all,
+                // which would turn an absent-versions result into a null and NPE below.
                 Mono.fromCallable(() -> fetchLatestVersionsByDatasetIds(ids, workspaceId))
-                        .subscribeOn(Schedulers.boundedElastic()))
+                        .subscribeOn(Schedulers.boundedElastic())
+                        .defaultIfEmpty(Map.of()))
                 .contextWrite(ctx -> AsyncUtils.setRequestContext(ctx, userName, workspaceId))
                 .block();
 
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetService.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetService.java
@@ -560,7 +560,9 @@ class DatasetServiceImpl implements DatasetService {
         }
 
         // For now, we are not going to use the criteria.withExperimentsOnly() method due to the migration.
-        return template.inTransaction(READ_ONLY, handle -> {
+        // Enrichment runs outside the transaction: it opens its own MySQL connection for the dataset versions,
+        // which would otherwise be held concurrently with this one for the same request.
+        DatasetPage rawPage = template.inTransaction(READ_ONLY, handle -> {
 
             var repository = handle.attach(DatasetDAO.class);
             int offset = (page - 1) * size;
@@ -569,14 +571,17 @@ class DatasetServiceImpl implements DatasetService {
                     criteria.withExperimentsOnly(),
                     criteria.withOptimizationsOnly(), visibility, filtersSQL, filterMapping);
 
-            List<Dataset> datasets = enrichDatasetWithAdditionalInformation(
-                    repository.find(size, offset, workspaceId, criteria.name(), criteria.projectId(),
-                            criteria.withExperimentsOnly(),
-                            criteria.withOptimizationsOnly(),
-                            sortingFieldsSql, visibility, filtersSQL, filterMapping));
+            List<Dataset> datasets = repository.find(size, offset, workspaceId, criteria.name(), criteria.projectId(),
+                    criteria.withExperimentsOnly(),
+                    criteria.withOptimizationsOnly(),
+                    sortingFieldsSql, visibility, filtersSQL, filterMapping);
 
             return new DatasetPage(datasets, page, datasets.size(), count, sortingFactory.getSortableFields());
         });
+
+        List<Dataset> datasets = enrichDatasetWithAdditionalInformation(rawPage.content());
+
+        return new DatasetPage(datasets, page, datasets.size(), rawPage.total(), sortingFactory.getSortableFields());
     }
 
     private Mono<DatasetPage> fetchUsingTempTable(int page, int size, DatasetCriteria criteria, Set<UUID> ids,
@@ -708,31 +713,39 @@ class DatasetServiceImpl implements DatasetService {
     private List<Dataset> enrichDatasetWithAdditionalInformation(List<Dataset> datasets) {
         Set<UUID> ids = datasets.stream().map(Dataset::id).collect(toSet());
 
-        Map<UUID, ExperimentSummary> experimentSummary = experimentItemDAO.findExperimentSummaryByDatasetIds(ids)
-                .contextWrite(ctx -> AsyncUtils.setRequestContext(ctx, requestContext))
-                .toStream()
-                .collect(toMap(ExperimentSummary::datasetId, Function.identity()));
+        if (ids.isEmpty()) {
+            return datasets;
+        }
 
-        Map<UUID, DatasetItemSummary> datasetItemSummaryMap = datasetItemDAO.findDatasetItemSummaryByDatasetIds(ids)
-                .contextWrite(ctx -> AsyncUtils.setRequestContext(ctx, requestContext))
-                .toStream()
-                .collect(toMap(DatasetItemSummary::datasetId, Function.identity()));
+        // RequestContext is @RequestScoped and therefore thread-bound: resolve it here, on the request thread,
+        // so the concurrent subscriptions below never call requestContext.get() from a worker thread.
+        String workspaceId = requestContext.get().getWorkspaceId();
+        String userName = requestContext.get().getUserName();
 
-        Map<UUID, OptimizationDAO.OptimizationSummary> optimizationSummaryMap = optimizationDAO
-                .findOptimizationSummaryByDatasetIds(ids)
-                .contextWrite(ctx -> AsyncUtils.setRequestContext(ctx, requestContext))
-                .toStream()
-                .collect(toMap(OptimizationDAO.OptimizationSummary::datasetId, Function.identity()));
+        var enrichmentData = Mono.zip(
+                experimentItemDAO.findExperimentSummaryByDatasetIds(ids)
+                        .collectMap(ExperimentSummary::datasetId, Function.identity()),
+                datasetItemDAO.findDatasetItemSummaryByDatasetIds(ids)
+                        .collectMap(DatasetItemSummary::datasetId, Function.identity()),
+                optimizationDAO.findOptimizationSummaryByDatasetIds(ids)
+                        .collectMap(OptimizationDAO.OptimizationSummary::datasetId, Function.identity()),
+                Mono.fromCallable(() -> fetchLatestVersionsByDatasetIds(ids, workspaceId))
+                        .subscribeOn(Schedulers.boundedElastic()))
+                .contextWrite(ctx -> AsyncUtils.setRequestContext(ctx, userName, workspaceId))
+                .block();
 
-        Map<UUID, DatasetVersion> latestVersionsByDatasetId = fetchLatestVersionsByDatasetIds(ids);
+        Map<UUID, ExperimentSummary> experimentSummary = enrichmentData.getT1();
+        Map<UUID, DatasetItemSummary> datasetItemSummaryMap = enrichmentData.getT2();
+        Map<UUID, OptimizationDAO.OptimizationSummary> optimizationSummaryMap = enrichmentData.getT3();
+        Map<UUID, DatasetVersion> latestVersionsByDatasetId = enrichmentData.getT4();
 
         return datasets.stream()
                 .map(dataset -> {
-                    var resume = experimentSummary.computeIfAbsent(dataset.id(), ExperimentSummary::empty);
-                    var datasetItemSummary = datasetItemSummaryMap.computeIfAbsent(dataset.id(),
-                            DatasetItemSummary::empty);
-                    var optimizationSummary = optimizationSummaryMap.computeIfAbsent(dataset.id(),
-                            OptimizationDAO.OptimizationSummary::empty);
+                    var resume = experimentSummary.getOrDefault(dataset.id(), ExperimentSummary.empty(dataset.id()));
+                    var datasetItemSummary = datasetItemSummaryMap.getOrDefault(dataset.id(),
+                            DatasetItemSummary.empty(dataset.id()));
+                    var optimizationSummary = optimizationSummaryMap.getOrDefault(dataset.id(),
+                            OptimizationDAO.OptimizationSummary.empty(dataset.id()));
                     var latestVersion = latestVersionsByDatasetId.get(dataset.id());
 
                     // When versioning is enabled and a latest version exists, use itemsTotal from the version
@@ -757,12 +770,10 @@ class DatasetServiceImpl implements DatasetService {
                 .toList();
     }
 
-    private Map<UUID, DatasetVersion> fetchLatestVersionsByDatasetIds(Set<UUID> datasetIds) {
+    private Map<UUID, DatasetVersion> fetchLatestVersionsByDatasetIds(Set<UUID> datasetIds, String workspaceId) {
         if (datasetIds.isEmpty()) {
             return Map.of();
         }
-
-        String workspaceId = requestContext.get().getWorkspaceId();
 
         return template.inTransaction(READ_ONLY, handle -> {
             var dao = handle.attach(DatasetVersionDAO.class);
--- a/apps/opik-backend/src/test/java/com/comet/opik/domain/DatasetServiceEnrichmentTest.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/domain/DatasetServiceEnrichmentTest.java
@@ -21,6 +21,7 @@ import org.mockito.junit.jupiter.MockitoExtension;
 import org.mockito.junit.jupiter.MockitoSettings;
 import org.mockito.quality.Strictness;
 import reactor.core.publisher.Flux;
+import reactor.core.scheduler.Schedulers;
 import reactor.util.context.ContextView;
 import ru.vyarus.guicey.jdbi3.tx.TransactionTemplate;
 import ru.vyarus.guicey.jdbi3.tx.TxAction;
@@ -29,7 +30,10 @@ import java.time.Instant;
 import java.util.List;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
 
 import static org.assertj.core.api.Assertions.assertThat;
@@ -235,6 +239,57 @@ class DatasetServiceEnrichmentTest {
         service.findById(datasetId);
 
         assertThat(observedWorkspaces).containsExactly(WORKSPACE_ID);
+    }
+
+    @Test
+    @DisplayName("The three ClickHouse lookups are in flight simultaneously, not one after another")
+    void enrichmentIssuesLookupsConcurrently() throws Exception {
+        var datasetId = UUID.randomUUID();
+        stubDataset(datasetId);
+
+        // Each lookup counts itself in, then blocks until all three have arrived. A serial implementation
+        // parks on the first lookup and never reaches the second, so the latch times out and the test fails.
+        var allSubscribed = new CountDownLatch(3);
+        var released = new CountDownLatch(1);
+
+        Runnable gate = () -> {
+            allSubscribed.countDown();
+            try {
+                if (!released.await(5, TimeUnit.SECONDS)) {
+                    throw new IllegalStateException("lookups were not released: enrichment is not concurrent");
+                }
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                throw new IllegalStateException(e);
+            }
+        };
+
+        when(experimentItemDAO.findExperimentSummaryByDatasetIds(anySet()))
+                .thenReturn(Flux.defer(() -> {
+                    gate.run();
+                    return Flux.just(new ExperimentItemDAO.ExperimentSummary(datasetId, 1, Instant.now()));
+                }).subscribeOn(Schedulers.boundedElastic()));
+        when(datasetItemDAO.findDatasetItemSummaryByDatasetIds(anySet()))
+                .thenReturn(Flux.defer(() -> {
+                    gate.run();
+                    return Flux.just(new DatasetItemSummary(datasetId, 1));
+                }).subscribeOn(Schedulers.boundedElastic()));
+        when(optimizationDAO.findOptimizationSummaryByDatasetIds(anySet()))
+                .thenReturn(Flux.defer(() -> {
+                    gate.run();
+                    return Flux.just(new OptimizationDAO.OptimizationSummary(datasetId, 1, Instant.now()));
+                }).subscribeOn(Schedulers.boundedElastic()));
+        when(datasetVersionDAO.findLatestVersionsByDatasetIds(anySet(), any())).thenReturn(List.of());
+
+        var enrichment = CompletableFuture.supplyAsync(() -> service.findById(datasetId));
+
+        assertThat(allSubscribed.await(5, TimeUnit.SECONDS))
+                .as("all three ClickHouse lookups should be subscribed before any of them completes")
+                .isTrue();
+        released.countDown();
+
+        var actual = enrichment.get(10, TimeUnit.SECONDS);
+        assertThat(actual.experimentCount()).isEqualTo(1);
     }
 
     @Test
--- a/apps/opik-backend/src/test/java/com/comet/opik/domain/DatasetServiceEnrichmentTest.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/domain/DatasetServiceEnrichmentTest.java
@@ -1,0 +1,261 @@
+package com.comet.opik.domain;
+
+import com.comet.opik.api.Dataset;
+import com.comet.opik.api.DatasetVersion;
+import com.comet.opik.api.Visibility;
+import com.comet.opik.api.sorting.SortingFactoryDatasets;
+import com.comet.opik.domain.filter.FilterQueryBuilder;
+import com.comet.opik.domain.sorting.SortingQueryBuilder;
+import com.comet.opik.infrastructure.BatchOperationsConfig;
+import com.comet.opik.infrastructure.FeatureFlags;
+import com.comet.opik.infrastructure.auth.RequestContext;
+import com.google.common.eventbus.EventBus;
+import jakarta.inject.Provider;
+import org.jdbi.v3.core.Handle;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.mockito.junit.jupiter.MockitoSettings;
+import org.mockito.quality.Strictness;
+import reactor.core.publisher.Flux;
+import reactor.util.context.ContextView;
+import ru.vyarus.guicey.jdbi3.tx.TransactionTemplate;
+import ru.vyarus.guicey.jdbi3.tx.TxAction;
+
+import java.time.Instant;
+import java.util.List;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anySet;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+@MockitoSettings(strictness = Strictness.LENIENT)
+@DisplayName("DatasetService enrichment Test")
+class DatasetServiceEnrichmentTest {
+
+    @Mock
+    private IdGenerator idGenerator;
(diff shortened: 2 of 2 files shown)


apps/opik-backend/src/test/java/com/comet/opik/domain/DatasetServiceEnrichmentTest.java
```diff
@@ -0,0 +1,330 @@
...
+        Function<ContextView, Void> record = ctx -> {
+            observedWorkspaces.add(ctx.get(RequestContext.WORKSPACE_ID));
+            return null;
+        };
+
+        when(experimentItemDAO.findExperimentSummaryByDatasetIds(anySet()))
+                .thenReturn(Flux.deferContextual(ctx -> {
+                    record.apply(ctx);
+                    return Flux.just(new ExperimentItemDAO.ExperimentSummary(datasetId, 1, Instant.now()));
+                }));
```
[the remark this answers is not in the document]
JetoPistola: Fixed in fa757df003 — you're right that the old test couldn't detect this. It only recorded the reactor context, so a serial implementation passed it unchanged.

Added `enrichmentIssuesLookupsConcurrently`: each of the three ClickHouse lookups counts itself in on a shared `CountDownLatch` and then blocks until all three have subscribed, with a bounded 5s timeout so a regression fails rather than hangs. Verified it actually bites by reverting the implementation to the serial `.block()`-per-lookup form — the test fails with "all three ClickHouse lookups should be subscribed before any of them completes"; it passes against the concurrent version.

Kept the existing context-propagation assertion as a separate test rather than folding it in, since the two failure modes are independent.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetService.java
```diff
@@ -708,31 +713,42 @@ public List<UUID> findIdsByPartialName(@NonNull String workspaceId, @NonNull Str
...
-                    var resume = experimentSummary.computeIfAbsent(dataset.id(), ExperimentSummary::empty);
-                    var datasetItemSummary = datasetItemSummaryMap.computeIfAbsent(dataset.id(),
-                            DatasetItemSummary::empty);
-                    var optimizationSummary = optimizationSummaryMap.computeIfAbsent(dataset.id(),
-                            OptimizationDAO.OptimizationSummary::empty);
+                    var resume = experimentSummary.getOrDefault(dataset.id(), ExperimentSummary.empty(dataset.id()));
+                    var datasetItemSummary = datasetItemSummaryMap.getOrDefault(dataset.id(),
+                            DatasetItemSummary.empty(dataset.id()));
+                    var optimizationSummary = optimizationSummaryMap.getOrDefault(dataset.id(),
+                            OptimizationDAO.OptimizationSummary.empty(dataset.id()));
```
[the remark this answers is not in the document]
JetoPistola: Partly fixed in fa757df003.

**Rename — done.** Renamed the map to `experimentSummaryMap` so the per-row variable can be `experimentSummary`, matching the neighbouring `datasetItemSummaryMap` / `optimizationSummaryMap` convention. (Note `resume` was pre-existing naming on `main`, not introduced here, but it's opaque either way and the rename is cheap.)

**Lazy allocation — skipping.** `ExperimentSummary`, `DatasetItemSummary` and `OptimizationSummary` are 3-field records with no allocation-sensitive state, and this runs once per dataset in a page — bounded by page size, not by traffic. Swapping `getOrDefault` for `get`-then-null-check trades three readable lines for three null branches to save allocations the JIT will very likely scalar-replace. The repo's code-style rule discourages this kind of speculative optimization, and the eager form is what makes the "no mutation during the mapping stream" property obvious at a glance — that mattered here, since the previous `computeIfAbsent` was mutating the shared summary maps while iterating them.

Happy to revisit if a profile ever shows it.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetService.java
```diff
@@ -569,14 +571,17 @@ public DatasetPage find(int page, int size, @NonNull DatasetCriteria criteria, L
...
+                    criteria.withExperimentsOnly(),
+                    criteria.withOptimizationsOnly(),
+                    sortingFieldsSql, visibility, filtersSQL, filterMapping);
 
             return new DatasetPage(datasets, page, datasets.size(), count, sortingFactory.getSortableFields());
         });
+
+        List<Dataset> datasets = enrichDatasetWithAdditionalInformation(rawPage.content());
+
+        return new DatasetPage(datasets, page, datasets.size(), rawPage.total(), sortingFactory.getSortableFields());
```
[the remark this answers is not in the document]
JetoPistola: Skipping — the premise doesn't hold for this PR: the version read was already in a separate transaction before this change.

On `main`, `fetchLatestVersionsByDatasetIds` opens its own `template.inTransaction(READ_ONLY, ...)` internally. Called from inside the outer `find()` transaction, that's a second, independent transaction — under MySQL's default REPEATABLE-READ it establishes its own snapshot and never shared the outer one. So a page/total read at T0 combined with a `latestVersion` read at T1 is the behaviour on `main` today; this PR doesn't introduce or widen it.

What changed is only *where* that pre-existing separate transaction is opened from — outside the outer transaction rather than nested inside it — which avoids holding two MySQL connections concurrently for one request. Worth noting the other three enrichment call paths (`findById`, `findByNameDetailed`, and the experiments-only `find()` branch) already ran enrichment outside any transaction on `main`, so this makes the four paths consistent rather than divergent.

Genuinely folding these into one snapshot would mean threading the outer `Handle` into enrichment and giving up the concurrency this PR is for — a real trade-off, but a different ticket. Not changing behaviour here; the consistency contract is unchanged from `main`.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetService.java
```diff
@@ -708,31 +713,42 @@ public List<UUID> findIdsByPartialName(@NonNull String workspaceId, @NonNull Str
...
+                        .collectMap(ExperimentSummary::datasetId, Function.identity()),
+                datasetItemDAO.findDatasetItemSummaryByDatasetIds(ids)
+                        .collectMap(DatasetItemSummary::datasetId, Function.identity()),
+                optimizationDAO.findOptimizationSummaryByDatasetIds(ids)
+                        .collectMap(OptimizationDAO.OptimizationSummary::datasetId, Function.identity()),
+                // defaultIfEmpty guards the zip: a Mono that completes empty makes zip emit nothing at all,
+                // which would turn an absent-versions result into a null and NPE below.
+                Mono.fromCallable(() -> fetchLatestVersionsByDatasetIds(ids, workspaceId))
+                        .subscribeOn(Schedulers.boundedElastic())
+                        .defaultIfEmpty(Map.of()))
```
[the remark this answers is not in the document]
JetoPistola: Skipping here — the finding is accurate but pre-existing and out of scope for this PR.

Confirmed the claim: `DatasetsResource.findDatasets` declares `@QueryParam("size") @Min(1) @DefaultValue("10")` with no upper bound, and `ProjectDatasetsResource` mirrors it. So a large `size` is accepted today, on `main`, independently of this change.

This PR doesn't alter how many rows a page returns or how much work enrichment does — it changes only whether the four lookups wait in line. If anything the fan-out is unchanged in volume: the same four queries run over the same id set, just concurrently. The amplification factor the comment describes is the same before and after.

Clamping page size is a real hardening item, but it touches two endpoints this PR doesn't otherwise change and affects every dataset list caller, so it belongs in its own ticket where it can be reviewed on its merits (and the max chosen against actual usage). Not expanding a latency fix into an API-contract change.

apps/opik-backend/src/test/java/com/comet/opik/domain/DatasetServiceEnrichmentTest.java
```diff
@@ -0,0 +1,330 @@
...
+        });
+
+        service = new DatasetServiceImpl(idGenerator, template, requestContextProvider, experimentItemDAO,
+                datasetItemDAO, experimentDAO, sortingQueryBuilder, filterQueryBuilder, sortingFactory,
+                batchOperationsConfig, optimizationDAO, eventBus, featureFlags, projectService);
+    }
+
+    private Dataset stubDataset(UUID id) {
+        var dataset = Dataset.builder().id(id).name("dataset-" + id).build();
+        when(datasetDAO.findById(id, WORKSPACE_ID)).thenReturn(java.util.Optional.of(dataset));
```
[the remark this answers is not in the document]
JetoPistola: Skipping — the specific gaps named are already covered, and Podam fixtures would weaken rather than strengthen these tests.

Cross-row mapping is covered: `enrichmentOnPageKeepsOrderAndPerRowSummaries` builds a three-dataset page where only the first and third have summaries, then asserts row order via `containsExactly` and checks each row's values individually — including that the middle row stays at zero rather than picking up a neighbour's counts, and that the third row's timestamp is its own. That's exactly the "assigning `third`'s timestamp to `first`" case. I also mutation-tested it: injecting cross-row bleed (`values().stream().findFirst()`) fails 2 tests.

`latestVersion` mapping is covered by `enrichmentPassesWorkspaceIdToVersionLookup`, which asserts `itemsTotal` flows through to `datasetItemsCount` and that the summary is non-null under the versioning flag.

On Podam and whole-object comparison: enrichment only ever reads `dataset.id()` and rebuilds via `toBuilder()`, so carrier fields pass through by construction — a randomized fixture would assert Lombok's `toBuilder` works, not this code. And an exact whole-object equality assertion would couple these tests to every future field added to `Dataset`, making them fail on unrelated changes. The explicit per-field assertions state what enrichment is actually responsible for.

If the concern is a future field being dropped in the `toBuilder()` chain, that's better caught by a targeted assertion than by randomized fixtures — happy to add one if you'd like.

apps/opik-backend/src/test/java/com/comet/opik/domain/DatasetServiceEnrichmentTest.java
```diff
@@ -0,0 +1,385 @@
...
+        // parks on the first lookup and never reaches the second, so the latch times out and the test fails.
+        var allSubscribed = new CountDownLatch(3);
+        var released = new CountDownLatch(1);
+
+        Runnable gate = () -> {
+            allSubscribed.countDown();
+            try {
+                if (!released.await(5, TimeUnit.SECONDS)) {
+                    throw new IllegalStateException("lookups were not released: enrichment is not concurrent");
+                }
```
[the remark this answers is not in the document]
JetoPistola: Fixed in a2aea24cc6 — good catch, the shared budget was a real flake source.

The worker gate and the caller's assertion ran concurrently off the same 5s deadline, so a slow-but-correct run that took ~4.9s to observe all three subscriptions left the workers ~0.1s before throwing `IllegalStateException` from a worker thread — a false failure, and one that would have reported a misleading cause.

Split into two constants: `SUBSCRIBE_TIMEOUT_SECONDS = 5` for the assertion, `WORKER_RELEASE_TIMEOUT_SECONDS = 60` for the workers' wait and the result `get`. The worker deadline now only exists to stop a genuine deadlock hanging the suite, not to race the assertion. Also moved `released.countDown()` into a `finally`, so if the assertion does fail it surfaces its own message instead of being masked by workers timing out behind it.

Verified both directions: 10 consecutive runs pass with no flakes, and reverting the implementation to serial still fails cleanly with "all three ClickHouse lookups should be subscribed before any of them completes" — as an assertion failure rather than a worker exception.

apps/opik-backend/src/test/java/com/comet/opik/domain/DatasetServiceEnrichmentTest.java
```diff
@@ -0,0 +1,385 @@
...
+                .thenReturn(Flux.defer(() -> {
+                    gate.run();
+                    return Flux.just(new DatasetItemSummary(datasetId, 1));
+                }).subscribeOn(Schedulers.boundedElastic()));
+        when(optimizationDAO.findOptimizationSummaryByDatasetIds(anySet()))
+                .thenReturn(Flux.defer(() -> {
+                    gate.run();
+                    return Flux.just(new OptimizationDAO.OptimizationSummary(datasetId, 1, Instant.now()));
+                }).subscribeOn(Schedulers.boundedElastic()));
+        when(datasetVersionDAO.findLatestVersionsByDatasetIds(anySet(), any())).thenReturn(List.of());
```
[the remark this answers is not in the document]
JetoPistola: Fixed in a2aea24cc6 — you're right that it was unreachable, and this is the sharpest comment on the PR.

Confirmed the analysis: every path through `fetchLatestVersionsByDatasetIds` returns a non-null map (`Map.of()` for empty ids, and JDBI's `List` mapper never returns null, so `toMap` yields `{}`), which means `Mono.fromCallable` always emits and `defaultIfEmpty` never fires. This matches something I'd found earlier via mutation testing: a mutant making that method return null *survived* the suite — the guard was doing real work but nothing proved it.

Went with keeping the guard and adding coverage rather than removing it. Added `enrichmentWhenVersionLookupEmitsNothingStillReturnsDataset`, which short-circuits only the transaction that attaches `DatasetVersionDAO` so the callable yields null and the Mono completes empty, then asserts the dataset still comes back with `latestVersion == null` and the three sibling summaries intact.

The test earns its place: deleting `.defaultIfEmpty(Map.of())` now fails it with exactly the NPE the guard prevents — `Cannot invoke "reactor.util.function.Tuple4.getT1()" because "enrichmentData" is null`. So the previously-surviving mutant is now killed.

Kept the guard over removing it because the failure mode is silent and severe: `Mono.zip` dropping an empty source turns a valid empty result into a 500, and a future edit to that lookup would reintroduce it invisibly. Cheap insurance, now actually tested.

CometActions:

### Access Information
- **URL:** https://pr-8076.dev.comet.com
- **Namespace:** pr-8076
- **Version:** 2.2.45-8076-merge-3150
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-8076%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

CometActions:

The four enrichment lookups now run concurrently under Mono.zip with the workspace/user resolved off the request thread, and getOrDefault replaces computeIfAbsent — so every number the Datasets list renders (Item count, Most recent experiment, Most recent optimization, latest version) is produced on a new path. DatasetServiceEnrichmentTest covers the combining logic well with mocked DAOs, but nothing e2e asserts those columns: datasets.list-datasets is covered only by dataset-crud-smoke.spec.ts, which asserts the row is visible and never reads a number, and dataset-version-counters.spec.ts reads itemsTotal from the versions endpoint and the Version history tab, not from the list. A dropped RequestContext or a bad zip default would surface as a 0 in a column no spec looks at. Note the QA bot already has draft #8084 open with dataset-list-enrichment.spec.ts and dataset-list-version-count.spec.ts from an earlier run on this PR, so no new environment or proposal is needed here — review that draft instead.

Would target `datasets.list-datasets`.

1. Seed one dataset with items, a committed version and at least one experiment, plus a second dataset with none of the three, so populated and empty rows are compared in the same response
2. Read GET /v1/private/datasets and check dataset_items_count, experiment_count, most_recent_experiment_at, optimization_count and latest_version against what was seeded, not just that the fields are present
3. Confirm the Datasets list page renders the same Item count and Most recent experiment values, and that the empty dataset shows 0 rather than a blank or a 500
4. Request a page of several datasets at once to check the zip fans out per dataset id correctly and no row inherits another row's summary
5. Fetch the same dataset via findById (dataset items page header) and confirm the counts agree with the list row

</details>

**Not testable yet.** find() now reads the page rows and the dataset versions in two READ_ONLY transactions instead of one, so a version committed between them is visible where it was not before. The skew is a real behaviour change but only observable if a write can be interleaved between the two reads inside a single list request, which the e2e estate cannot force today.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/33588492813)

_Re-checked after a push on 02 Sep 03:55 UTC._

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-8076`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

CometActions:

### Access Information
- **URL:** https://pr-8076.dev.comet.com
- **Namespace:** pr-8076
- **Version:** 2.2.45-8076-merge-3154
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-8076%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetService.java
```diff
@@ -560,7 +560,9 @@ public DatasetPage find(int page, int size, @NonNull DatasetCriteria criteria, L
         }
 
         // For now, we are not going to use the criteria.withExperimentsOnly() method due to the migration.
-        return template.inTransaction(READ_ONLY, handle -> {
+        // Enrichment runs outside the transaction: it opens its own MySQL connection for the dataset versions,
+        // which would otherwise be held concurrently with this one for the same request.
```
thiagohora: The rationale here is inverted, and it matters because as written it argues for undoing the change.

The old nested call did **not** open its own MySQL connection. `TransactionTemplate.inTransaction` checks `manager.isUnitStarted()` first, and `UnitManager` holds the handle in a `ThreadLocal` — so when `fetchLatestVersionsByDatasetIds` was reached from inside this transaction, on the same thread, it short-circuited into `inCurrentTransaction(...)` and ran on the *same* handle, the same connection, inside the same transaction. There was never a second connection to hold concurrently.

The actual justification is stronger than the stated one and worth having in the comment instead: the old code held the Hikari connection open across three sequential ClickHouse round-trips. That is the pool-hogging worth fixing, and it is a good reason to move enrichment out.

Two follow-ons:

- The move is **not** visibility-neutral for `find()`, contrary to the PR description. Page rows and their `dataset_versions` used to be read in one `READ_ONLY` (REPEATABLE READ) transaction; they are now two. A version written between the two reads is now observable where it previously wasn't. Benign in practice, but `find()` is the one path where the "already ran outside any transaction" argument doesn't apply — `findById`, `findByNameDetailed` and the experiments-only branch genuinely did.
- Incidentally, the *new* arrangement is the one that opens a genuinely separate handle: `Mono.fromCallable(...).subscribeOn(boundedElastic())` runs on a worker with no unit started, so `beginUnit()` → `jdbi.open()`. Net connection count is unchanged, since the outer transaction has closed by then — just worth being accurate about.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetService.java
```diff
@@ -708,31 +713,43 @@ public List<UUID> findIdsByPartialName(@NonNull String workspaceId, @NonNull Str
...
-                .contextWrite(ctx -> AsyncUtils.setRequestContext(ctx, requestContext))
-                .toStream()
-                .collect(toMap(OptimizationDAO.OptimizationSummary::datasetId, Function.identity()));
+        // RequestContext is @RequestScoped and therefore thread-bound: resolve it here, on the request thread,
+        // so the concurrent subscriptions below never call requestContext.get() from a worker thread.
+        String workspaceId = requestContext.get().getWorkspaceId();
+        String userName = requestContext.get().getUserName();
 
-        Map<UUID, DatasetVersion> latestVersionsByDatasetId = fetchLatestVersionsByDatasetIds(ids);
+        var enrichmentData = Mono.zip(
```
thiagohora, line 732: Capacity note, sharpening the open question in the PR description.

`DatabaseAnalyticsFactory.build()` uses `ConnectionFactories.get("r2dbc:clickhouse:...")` — not `r2dbc:pool:` — so there is no R2DBC pool bounding this, and the async work lands on `ClickHouseClient.getExecutorService()`, which is **static and process-wide**.

So the ceiling on in-flight ClickHouse queries from this endpoint goes from `1 x concurrent requests` to `3 x concurrent requests`, all contending on one shared executor. Per-request latency clearly improves; what's worth measuring before merge is the list endpoint under concurrency, since that is where the 3x multiplies against a shared pool with no backpressure of its own.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetService.java
```diff
@@ -708,31 +713,43 @@ public List<UUID> findIdsByPartialName(@NonNull String workspaceId, @NonNull Str
...
+                experimentItemDAO.findExperimentSummaryByDatasetIds(ids)
+                        .collectMap(ExperimentSummary::datasetId, Function.identity()),
```
thiagohora: `Collectors.toMap` → `Flux.collectMap` silently changes duplicate-key behaviour: `toMap` throws `IllegalStateException` on a duplicate, `collectMap` is last-wins.

All three queries are `GROUP BY dataset_id`, so duplicates shouldn't occur — and the fact that the old code wasn't throwing in production is good evidence they don't. But a future change to one of these queries that introduced duplicates would now silently return one row's summary for a dataset instead of failing loudly. Worth a moment's thought about whether the loud form was doing useful work here.

apps/opik-backend/src/test/java/com/comet/opik/domain/DatasetServiceEnrichmentTest.java
```diff
@@ -0,0 +1,441 @@
...
+        when(datasetItemDAO.findDatasetItemSummaryByDatasetIds(anySet()))
+                .thenReturn(Flux.defer(() -> {
+                    gate.run();
+                    return Flux.just(new DatasetItemSummary(datasetId, 1));
+                }).subscribeOn(Schedulers.boundedElastic()));
+        when(optimizationDAO.findOptimizationSummaryByDatasetIds(anySet()))
+                .thenReturn(Flux.defer(() -> {
+                    gate.run();
+                    return Flux.just(new OptimizationDAO.OptimizationSummary(datasetId, 1, Instant.now()));
+                }).subscribeOn(Schedulers.boundedElastic()));
```
thiagohora, line 320: These `.subscribeOn(Schedulers.boundedElastic())` calls mean the test guards less than it looks like it does.

Production has no `subscribeOn` on the three ClickHouse sources — the concurrency relies entirely on them being async at subscribe (which they are today, per the `ASYNC, true` in `ClickHouseConnection`). The mocks here supply an off-thread hop that the real code does not have, so if a DAO ever became blocking-at-subscribe, `Mono.zip` would quietly degrade back to serial and this test would still pass.

Not a blocker — testing that without the hop is awkward — but worth knowing that the regression this test is named for is not actually fenced off.

thiagohora (commented):
Reviewed against the checked-out sources (guicey-jdbi3 8.0.0, clickhouse-r2dbc 0.9.x). The change is sound and the concurrency is real, not just apparent — `ClickHouseConnectionFactory.create()` is `Mono.defer(Mono::just)` with no I/O, and `ClickHouseConnection` forces `ClickHouseClientOption.ASYNC, true`, so `Mono.fromFuture(request::execute)` returns without running the query on the subscribing thread. `Mono.zip`'s sequential subscription therefore does not serialize the three lookups.

I also checked the context narrowing: the new 2-arg `AsyncUtils.setRequestContext(ctx, userName, workspaceId)` drops `WORKSPACE_NAME` and `VISIBILITY` versus the `Provider` overload, but all three summary DAOs go through `makeFluxContextAware`, which reads only `USER_NAME` / `WORKSPACE_ID`. Safe, and consistent with what `find()` already does for `experimentDAO.findAllDatasetIds`.

Four comments inline — correctness and performance only. Nothing here blocks the returned `Dataset` objects being correct.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetService.java
JetoPistola: Fixed in 66e4200e39 — you're right, and thank you for catching this. I had the mechanism backwards.

I verified your account against the bytecode before changing anything: `TransactionTemplate.inTransaction` calls `UnitManager.isUnitStarted()` and branches to `inCurrentTransaction(...)` when a unit exists, and `UnitManager` holds the handle in a `ThreadLocal<Handle>`. So the old nested call, reached on the same thread from inside this transaction, ran on the same handle and the same connection. There was never a second connection.

That means the comment was not just imprecise — as you say, it argued for undoing the change. Replaced with the real justification: the old arrangement held the Hikari connection open across three sequential ClickHouse round-trips.

I also took the second follow-on. The comment and the PR description now state plainly that this is **not** visibility-neutral for `find()`: page rows and `dataset_versions` were read in one `READ_ONLY` transaction and are now two, so a version written between them is observable where it wasn't. I'd previously argued the opposite to another reviewer on this same thread, using the same wrong premise — that's now corrected in the description rather than left standing.

Your third point is noted too: the new `Mono.fromCallable(...).subscribeOn(boundedElastic())` is the arrangement that actually opens a separate handle, since the worker has no unit started. Net connection count unchanged because the outer transaction has closed by then.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetService.java
JetoPistola: Measured this on the ad hoc environment before merge, since you're right that the shared static executor is where the 3x would bite.

Setup: 41 datasets (6 with items and versions, 34 empty), full pages of 40, so each request fans out the three ClickHouse lookups plus the MySQL one.

| concurrent requests | in-flight DB ops | non-200 | p50 | p95 |
|---|---|---|---|---|
| 10 | 40 | 0 | 854 ms | 874 ms |
| 20 | 80 | 0 | 1327 ms | 1584 ms |
| 40 | 160 | 0 | 1748 ms | 1959 ms |
| 60 | 240 | 0 | 1875 ms | 2159 ms |

Zero errors at every level. Latency rises smoothly rather than cliff-edging, which reads as queuing on the shared executor rather than starvation or rejection.

Per-request scaling on the same environment: server time (total minus a ~447 ms network floor) goes 43 ms at size=1 to 195 ms at size=40 — 4.5x for 40x the rows, so per-row cost drops from 43 ms to under 5 ms.

Caveat on the numbers: that environment runs the concurrent build only, so this is scaling and saturation evidence, not a hardware-matched before/after against the serial version.

Agreed there's no R2DBC pool bounding this and no backpressure of its own — worth watching if the list endpoint's concurrency profile changes, but I didn't find a knee up to 240 in-flight operations.

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetService.java
JetoPistola: Fixed in 66e4200e39 — agreed the loud form was doing useful work, so I put it back.

All three lookups now use `.collect(toMap(...))` rather than `.collectMap(...)`, restoring the throw-on-duplicate semantics the serial code had. Confirmed the three queries do `GROUP BY dataset_id`, so this should never fire in practice — but as you say, the value is that a future query change which broke that assumption fails loudly instead of silently returning one row's summary for a dataset.

Added a test for it (`enrichmentWhenLookupReturnsDuplicateKeyFailsLoudly`) and checked it actually discriminates: with `collect(toMap(...))` a duplicate `dataset_id` raises `IllegalStateException: Duplicate key ...`; reverting that one line to `collectMap` makes the test fail with "Expecting code to raise a throwable". So the guard is pinned rather than assumed.

Good catch — this was a silent semantic change I introduced without noticing.

apps/opik-backend/src/test/java/com/comet/opik/domain/DatasetServiceEnrichmentTest.java
JetoPistola: Correct, and worth stating in the test rather than leaving implicit — documented in 66e4200e39.

The test now carries a scope note saying exactly what you describe: the mocks add a `subscribeOn(boundedElastic())` hop that production does not have, so the test fences off a *structural* regression (reverting to collect-each-result-eagerly, which is what it was written against and which it does still catch — I verified it fails on that revert) but **not** a DAO becoming blocking-at-subscribe, which would degrade the zip to serial while the test kept passing.

I didn't try to remove the hop. Without it the mock Flux emits synchronously on the subscribing thread, so the latch would deadlock on the first lookup — testing the real property needs a source that is async-at-subscribe the way `ClickHouseConnection` is, which the mocks can't reproduce without effectively reimplementing it. Recording the gap seemed more honest than a test that looks stronger than it is.

Thanks for checking the driver behaviour in the review body too — `ASYNC, true` plus `Mono.fromFuture` is the property the whole change rests on, and it's good to have that independently confirmed rather than assumed.

CometActions:
🌙 **Nightly cleanup:** The test environment for this PR (`pr-8076`) has been cleaned up to free cluster resources. PVCs are preserved — re-deploy to restore the environment.

CometActions:

### Access Information
- **URL:** https://pr-8076.dev.comet.com
- **Namespace:** pr-8076
- **Version:** 2.2.45-8076-merge-3173
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-8076%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetService.java
thiagohora: Correction to this comment — the "no R2DBC pool bounding this" framing is wrong, please disregard it.

Pooling for this driver lives at the HTTP layer, not the R2DBC layer. `clickhouse-http-client`'s `ApacheHttpConnectionImpl$HttpConnectionManager` extends Apache's `PoolingHttpClientConnectionManager`, bounded by `ClickHouseHttpOption.MAX_OPEN_CONNECTIONS` with `KEEP_ALIVE` / `KEEP_ALIVE_TIMEOUT` (and the `HttpUrlConnectionImpl` provider relies on JDK keep-alive reuse). The r2dbc `Connection` is a thin wrapper over that — `ClickHouseConnectionFactory.create()` does no I/O at all — so socket reuse and the concurrency ceiling are already handled a layer down, and adding `r2dbc:pool:` would be redundant rather than a fix. Nothing to change here.

What survives is much weaker than I made it sound: three concurrent lookups per request instead of one means more simultaneous requests against that HTTP pool, so under load they queue there. That is the pool doing its job — graceful degradation, not starvation. The latency measurement in the acceptance criteria is still worth having, but not for the reason I gave.

thiagohora (approved):
Approving. The correctness items from my earlier review are properly addressed in 66e4200e


**One ask before merge: the before/after latency number for a page of datasets.**

Not a blocker on correctness -- the speedup is structurally verified rather than assumed (`ClickHouseConnectionFactory.create()` does no I/O and `ClickHouseConnection` forces `ASYNC=true`, so `Mono.zip`'s sequential subscription genuinely does not serialize the three lookups; `T1+T2+T3+T4` really does become `max(...)`). But it is the PR's own stated acceptance criterion and the entire justification for the change, and it is the one thing that says whether the win is material or whether a single query dominates the max. Worth having on the ticket either way.

Two non-blocking notes:

- The description is stale: it still reads "Opening as a draft so CI runs the full suite", still says `DatasetServiceEnrichmentTest` is 8/8 when it is now 10, and still lists both acceptance criteria as open. Worth refreshing so it matches what shipped.
- `enrichmentWhenLookupReturnsDuplicateKeyFailsLoudly` asserts `hasMessageContaining("Duplicate key")`, which is the JDK's `Collectors.uniqKeysMapAccumulator` wording rather than anything Opik controls, and it has changed across JDK versions before. Asserting the type plus the dataset id (which the message also carries) would be more durable.

JetoPistola:
> @thiagohora wrote in their review (APPROVED):
>
> **One ask before merge: the before/after latency number for a page of datasets.**
>
> Not a blocker on correctness -- the speedup is structurally verified rather than assumed [...] But it is the PR's own stated acceptance criterion and the entire justification for the change, and it is the one thing that says whether the win is material or whether a single query dominates the max.

Measured. You were right to push for the real number — it is **more bounded than this PR implied**, and the answer to "material, or does one query dominate the max" turns out to be *both, depending on page size*.

**Method.** Pushed the pre-change serial enrichment as a throwaway branch and deployed it to its own ad hoc env (#8112, do-not-merge, will be deleted). The two builds differ in exactly one file and only in the enrichment block — serial `.block()`-per-lookup vs `Mono.zip`. Both envs seeded identically: 88 datasets, 51 carrying items and versions. Network floors within 3.4 ms of each other (420.7 / 424.1 ms), subtracted below. Requests interleaved A/B/A/B so cluster drift hits both equally. 25 samples per size, plus 40 more at size 88.

**Enrichment server time (p50, floor subtracted):**

| rows | serial (before) | concurrent (after) | saved | speedup |
|---|---|---|---|---|
| 1 | 68.0 ms | 43.6 ms | 24.3 ms | 1.56x |
| 10 | 67.1 ms | 43.7 ms | 23.4 ms | 1.54x |
| 40 | 213.2 ms | 188.7 ms | 24.5 ms | 1.13x |
| 88 | 269.0 ms | 274.2 ms | -5.2 ms | 0.98x |

Re-ran size 88 with 40 more interleaved samples to check that last row: serial 270.7 ms vs concurrent 273.3 ms, p25-p75 of 264.6-368.2 and 268.2-349.6. They overlap heavily — at 88 rows the two are indistinguishable, not a regression.

**What this says.** The saving is a **fixed ~24 ms**, essentially constant from 1 to 40 rows, which is what you would expect from collapsing three sequential round-trips into one wait: it removes latency, not work. So it is a clean ~1.5x on the small pages that dominate real traffic (the default page size is 10, and `get_or_create_dataset` enriches a single dataset), and it fades to nothing on large pages where per-row query cost dominates the max. Your framing was the right one — past roughly 40 rows a single query does dominate, and the overlap stops buying anything.

I have corrected the PR description, which claimed a stronger and more general win than this supports. Happy to add a page-size note to the ticket if you think it is worth recording that the benefit is concentrated below ~40 rows.

Not merging — @JetoPistola has asked to be the one to merge.

JetoPistola:
Failure of unit tests BE- will be addressed, in https://github.com/comet-ml/opik/pull/8108

Merging, as it's not a blocker

apps/opik-backend/src/main/java/com/comet/opik/domain/DatasetService.java
JetoPistola: Following up on the non-blocking note in your approval — the one about `enrichmentWhenLookupReturnsDuplicateKeyFailsLoudly` asserting `hasMessageContaining("Duplicate key")`, which is `Collectors.uniqKeysMapAccumulator` wording rather than anything we control. It shipped separately in #8128 rather than being dropped.

Replying on this thread because the note lives in the approval review body, which GitHub gives no way to reply to inline; this is the thread the test came out of, so it seemed the closest place to keep it in context.

Split out deliberately: applying it here would have meant either holding up the merged optimization or forcing a re-review of an approved PR for a one-line test change, and it keeps the diff reviewable in isolation.

The change is exactly what you suggested — type plus dataset id:

```java
.isInstanceOf(IllegalStateException.class)
.hasMessageContaining(datasetId.toString());
```

Two things I checked rather than assumed:

- **The id really is in the message.** On Corretto 25.0.3 it appears twice — once as the formatted key and again inside the record's `toString`: `Duplicate key <id> (attempted merging values ExperimentSummary[datasetId=<id>, count=1, ...] and ...)`. So it isn't just trading one fragile string for another.
- **The guard still discriminates.** Reverting the three `.collect(toMap(...))` calls to `.collectMap(...)` still fails the test, and only that test of the 11. Worth noting why the swap is safe: last-wins collection throws nothing at all, so the failure is `Expecting code to raise a throwable` at `assertThatThrownBy` — the discriminating power was never in the wording, which is the part that makes your suggestion a pure improvement. And asserting the id is strictly stronger: it pins the failure to *this* dataset's duplicate.

Good catch — this one would have sat there until a JDK upgrade turned it into a confusing red build.

PR #1, opened 2023-05-11
[CM-7070] llm sdk allow users to logs basic llm prompts and answers

alexkuzmik:

PR #6874, opened 2026-05-26
[OPIK-6597] [QA] feat: Online Evaluation smoke (LLM-judge + Python rule)

AndreiCautisanu:
## Details

Exercises the dynamic evaluation engine end-to-end. User creates an online scoring rule via the UI, traces land, the engine asynchronously scores them, and the resulting `feedback_score` appears on each trace AND renders in the Logs page's trace panel. Two test cases (`@t1-smoke @online-evaluation`), each seeding **5 traces** to also exercise engine fan-out:

- **Test A — LLM-as-judge (the headline journey).** Provisions Anthropic or OpenAI from env via the Configuration UI, creates a rule using the canned `Moderation` template, seeds 3 safe + 2 unsafe outputs, polls all 5 in parallel for the Moderation score, asserts the judge separates safe (≤ 0.3) from violation-flagged (> 0.3) and that `max(safe) < min(unsafe)` — the structural claim, robust to LLM variance. UI spot-check on 1 safe + 1 unsafe in the Feedback scores tab of the trace panel.
- **Test B — Python Equals (deterministic anchor).** Creates a `user_defined_metric_python` rule via the UI dialog with an inline string-equality snippet. Seeds 3 matching + 2 non-matching outputs, polls in parallel, asserts exact `1.0`/`0.0` per bucket. UI spot-check on one of each.

Supporting changes:
- `backendClient` gains `getTrace`, `pollTraceForFeedbackScore`, `listAutomationRulesForProject`, `deleteAutomationRule` (the polling helper is the suite's first eventually-consistent assertion primitive). Each test deletes its rule before teardown.
- New `OnlineEvaluationPage` POM; `TracePanelPage` extended with Feedback-scores-tab helpers.
- `global-setup` now injects `opik-version-override=v2` into the storage state (so the FE mounts the v2 SPA on `version_1`-flagged workspaces) and dismisses the first-run welcome-wizard modal; `ConfigurationPage.listConfiguredProviders` waits for the provider table to settle before counting rows.
- 4 `data-testid` attributes added to the Online Evaluation page + AddEditRuleDialog, used exclusively by the test.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-6597

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.7
- Scope: full implementation (spec, backend-client extensions, POMs, tests, FE testids, global-setup portability)
[...]

PR #8080, opened 2026-08-31
[OPIK-8164] [QA] Proposed e2e spec from the #8060 exploration: an oversized online-scoring payload must not stop the rule

CometActions:
## Details

> **This spec is RED, on purpose.** It fails on the PR's own environment, and
> the failure looks like the application rather than the spec. The evidence is
> in "Verification" below. Read that before deciding what to do with this PR —
> the finding may matter more than the spec.

### Where this came from

Exploratory testing of **comet-ml/opik#8060** — *[OPIK-8164] [BE] fix: apply Jackson stream-read limits to the Redis stream codec* — against that PR's own deployed environment, `https://pr-8060.dev.comet.com` (`GET /api/is-alive/ver` → `2.2.45-8060-merge-3149`, OSS, workspace `default`, no provider keys). A human-driven exploration drove the flow there by hand first: a span-scope `user_defined_metric_python` rule at 100% sampling, a control span, an oversized span, and a small span behind it, plus a size sweep from 0 MB to 25 MB. This PR turns the one strong candidate from that exploration into a permanent spec.

**Base branch: `thiagohora/OPIK-8164-online-scoring-codec-limits`, not `main`.** opik#8060 is still open, so the codec-initialisation fix this spec is written around only exists on that branch; the spec was written and run against its head commit `aef1529643666abe32ea8965e1a3e12a7e45ac78`. When #8060 merges, retarget this to `main` and re-run — `main` will have moved on.

### What was added

**`tests_end_to_end/e2e/tests/online-evaluation/online-evaluation-oversized-payload.spec.ts`**
`@t3-nightly` `@area:online-evaluation` `@cap:online-evaluation.python-rule-scores` `@cap:online-evaluation.scores-in-trace-panel`

One test, seeded through REST and read back through both the API and the UI:

1. Creates a **span-scope** `span_user_defined_metric_python` rule at 100% sampling, mapping one small field (`output.answer`), so the evaluator's own payload stays small and the *only* thing the oversized span makes bigger is the message on the scoring stream.
2. Reads the rule back and asserts `type`, `sampling_rate` and `enabled`. Without this the spec could silently create a **trace**-scope rule — the API's default `type` — and spend its whole runtime proving a property about a stream it never touched.
3. **Control:** an ordinary span is scored 1.0. This is what makes a later absence meaningful; without it, an unscored span at the end is indistinguishable from a rule that never worked here at all.
4. Writes a span whose `output.blob` is **21,000,000 characters** — one step over Jackson's `DEFAULT_MAX_STRING_LEN` (20,000,000), the limit the online-scoring Redis codec actually ran on before this PR. The spec then polls until the span reads back with a **full-length** blob, so a seed that was truncated or refused fails loudly instead of letting the rest of the test pass for the most boring possible reason.
5. **The property:** an ordinary span logged *afterwards* is written and scored 1.0, and carries exactly one feedback score (the whole score list is compared, not `find()`-ed).
6. Both control spans show `1.0` on the trace panel's Feedback scores tab, with the score row resolving to exactly one element. Written through REST, read back through the UI — a backend-only pass would hide the failure a user actually reports.

**Deliberately not asserted:** whether the oversized span itself gets a score. Whether an over-ceiling message is dropped, decoded or refused depends on the deployment's configured Jackson limits and on the opt-in `REDIS_SCORING_DROP_OVERSIZED_PAYLOADS` guard, so pinning it either way would assert the environment rather than the product.

**Why 21 MB and not 25 MB:** just over the limit is all the property needs, and 25,000,000 characters was measured to take the ingest backend down outright — see below. A spec that has to break the environment to run is not a spec anyone can keep. (21 MB turns out not to be entirely innocent either; also below.)

### Supporting estate changes

- `backendClient.createSpan` / `getSpan` / `pollSpanForFeedbackScore`. The estate's only span seeder is `sdkClient.python.createNestedTrace`, which writes a whole trace and its spans in one call — and the **order** in which spans reach the stream is this spec's entire subject, so they have to be written one at a time with a confirmed 201 each. `source: 'sdk'` is mandatory: `OnlineScoringSpanSampler` keeps only spans whose source `isLoggingSource`.
- `createAutomationRule` takes a `type`; `getAutomationRule` reads it back. Span scope decides the Redis stream, so it is both the only way to reach the span stream and something the spec has to prove rather than assume.
- `postSeedWrite` / `withReadRetry`: REST seed writes and the span read retry 5xx with a short backoff. The ingress in front of a deployed environment answers 502 in short bursts; without this, every REST-seeding spec inherits that as flake. Retrying is safe only because these writes are keyed on a caller-supplied id and upsert on it. A persistent 5xx still throws — this hides a blip, not an outage, which is exactly why the failure below still surfaces.
- `poll-feedback-score.ts` is generalised over trace/span; `pollTraceForFeedbackScore` keeps its signature.
- `taxonomy.yaml`: spec added to `online-evaluation.specs`, and `python-rule-scores`' note extended. No capability was flipped — see "Taxonomy" below.

### Verification

Run from `tests_end_to_end/e2e/` against `https://pr-8060.dev.comet.com`:

```bash
npx playwright test tests/online-evaluation/online-evaluation-oversized-payload.spec.ts \
  --reporter=list --workers=1 --retries=0
```

(`OPIK_API_KEY` had to be set to a placeholder: the pinned TS SDK refuses to construct a client without one for a non-localhost `apiUrl`, even though this OSS deployment ignores the header — a bogus value and no header both answer 200.)

- `tsc --noEmit` — clean for every file this PR touches. (The suite as a whole does not currently typecheck on the pinned `typescript@7.0.2`: `tsconfig.json`'s `baseUrl` was removed in TS 7, and `core/backend/client.ts` has a pre-existing duplicate `deleteDashboard`. Both are present on `main` and unrelated to this change; I checked the error list is byte-identical with and without my diff.)
- `python3 tests_end_to_end/coverage/tag_lint.py …` — `55 specs checked, 1 exempt, 0 problem(s)`.
- **Spec run — passed once, then failed reproducibly.**

### The pass

At a 25,000,000-character payload the spec ran **green end to end in 41.8 s**, exercising every assertion including the UI read-back. So the spec works: the rule fires, the control scores, the oversized span round-trips at full length, the span behind it scores, and the panel renders it.

### The failure, and why it looks like the app

At the shipped 21,000,000-character payload the spec fails at step 5, **reproducibly, at the same assertion**, from a verified-healthy start:

```
Ingest and scoring survived: an ordinary span logged afterwards is written and scored
  Error: createTraceWithSource '…-after-trace': expected 201, got 500 after 4 attempt(s):
  There was an error processing your request. It has been logged (ID 9f43548504f3664c).
```

The oversized span is accepted (201). The **next ordinary small write** then answers 500, through four attempts spread over 12 seconds, and the deployment goes on answering 500/502 to *everything* for about two minutes before recovering on its own.

That is worse than the wedged stream this PR is about: it is not "one span went unscored", it is ordinary ingest failing platform-wide.

Controls, all measured on the same environment in the same session:

| Condition | Result |
|---|---|
| Ordinary traffic, no large writes | **0 failures in 340 requests**, over 3 minutes — measured twice, before and after the experiments |
| 21 MB span written to a project with **no** automation rule, 5 consecutive write + full read-back cycles | **5/5 clean**, API healthy throughout |
| 16 MB span written to a project **with** a 100% span rule | clean, span scored, API healthy |
| 21 MB span written to a project **with** a 100% span rule | collapses within 1–2 writes; ordinary small writes 500/502 for ~2 min |
| 25 MB span written to a project **with** a 100% span rule | collapses on the **first** write |

So the trigger is specifically **an oversized span meeting an active online-scoring rule** — the publish path this PR touches. The same payload with no rule attached is harmless, and the deployment is otherwise completely stable.

**What I cannot tell you**, and it is the reason this is framed as a finding rather than a filed bug: whether this is a product defect or this preview environment being under-resourced for a 21 MB payload the publish path has to hold in memory. That needs the backend's own logs and heap metrics around the failure, which I have no access to — the same wall the exploration hit. The `x-trace-id` / log IDs above are in the response bodies if someone can look them up.

### What this means for merging

Please decide deliberately rather than merging on the green-tick reflex:

- If the collapse is the environment's sizing, this spec should probably merge and simply be run somewhere with headroom.
- If it is the product, **the spec has already earned its place** — it caught, on its first run, a platform-wide ingest outage triggered by a payload the API accepts.

Either way there is a real operational hazard, and it is written into the spec's own header comment: on this deployment, one run of this spec takes ingest down for every other spec sharing the environment for about two minutes. It is tagged `@t3-nightly` partly for that reason, but it may want a disposable environment or its own shard. I have not `.skip`-ed it and would rather not — a skipped spec is a spec nobody ever looks at again.

The exploration produced two candidates; **one was dropped**.

*"A trace whose payload exceeds the online-scoring limit still receives its python-rule score"* — dropped. It is marked `verified_on_staging: false` and `weak`: **trace-scope** python online scoring was already dead deployment-wide on `pr-8060.dev.comet.com` before the exploration started (nothing trace-scope scored anywhere from 07:56 UTC onward, across 9 traces, a fresh project and a fresh 100% rule, with an empty rule-log stream), so there was no working baseline to separate "the oversized trace was dropped" from "the whole stream was already dead". A spec for a flow nobody could reach is a guess. It also carries the sharper version of the hazard above: on a build where the trace stream does wedge, one run kills trace-scope scoring for the whole deployment permanently.

That is why this spec is span-scope. The two scopes ride separate Redis streams (`stream_scoring_span_user_defined_metric_python` vs `stream_scoring_user_defined_metric_python`) through the same codec, so either proves the property — and span scope is the one that could actually be driven to a result.

### Taxonomy

The spec is registered in `online-evaluation.specs`. **No capability was flipped to `covered: true`**, deliberately:

- `python-rule-scores` and `scores-in-trace-panel` were already `covered: true`; the note on the first was extended to record what this adds.
- `rule-scope-thread-span` stays `covered: false`. This spec does create and drive a span-scope rule end to end, but that key covers **both** the span and thread scope flags and nothing exercises thread scope. Flipping it on the span half would retire the thread half untested. Its note now says so. If you disagree, the flip is one line — but it should be a deliberate call, not a side effect of this PR.

---

🤖 Generated by the release QA side flow (test-proposal step). It is a **draft** and needs human review before merge: an automated spec that asserts the wrong thing is worse than no spec, because it will be trusted.

[OPIK-8164]: https://comet-ml.atlassian.net/browse/OPIK-8164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Change checklist

## Issues

OPIK-8164

## Testing

## Documentation


CometActions:
## 🔍 Generated-test review

# PR #8080 — [OPIK-8164] [QA] Proposed e2e spec from the #8060 exploration: an oversized online-scoring payload must not stop the rule

CometActions · draft: yes · 14 files, 1 new spec · head `9eb372691d` (was `b799500973`)

## VERDICT: needs changes — blocked on the merge state, not on the spec

The Playwright spec itself is one of the better generated specs I have reviewed
in this estate: correct placement, honest tags, a real control-before/control-after
structure, fixture-based teardown, and a discriminating seed assertion. **The
blocker is not the test. It is that this branch still carries the pre-merge
backend fix it was explored against, and that fix has since merged to `main` in a
different shape.** As it stands, merging this PR would revert shipped backend work.

Everything below the blocker is real but secondary — see "Does the conflict make
the rest moot?".

---

## Blockers (1)

### 1. The PR carries production backend code that has already merged to `main` in a different form

`apps/opik-backend/**` (7 files), `deployment/docker-compose/docker-compose.yaml`

This branch is 10 commits ahead of its merge-base, and **9 of those 10 are backend
commits**, not QA work:

```
b799500973 [OPIK-8164] [QA] e2e: oversized online-scoring payload must not stop the rule   <- the actual QA commit
aef1529643 fix: address review — WARN on deliberate drops, config-test key, toggle tests
fdaa8dd998 fix: make the oversized-payload drop opt-in, not opt-out
3086e0e059 fix: null contract, metric docs and single-pass partition
1a82d87b27 fix: let callers see a skipped oversized message and compensate
38fbebe910 fix: pass REDIS_SCORING_ALLOW_OVERSIZED_PAYLOADS through docker-compose
df94c9cecf fix: keep the publisher at the legacy ceiling during rolling upgrades
0cf107a794 fix: log a skipped oversized payload at error level
07463b2850 fix: update OnlineScorePublisherTest for the new JacksonConfig constructor arg
97c1f9f61f [OPIK-8164] [BE] fix: apply Jackson stream-read limits to the Redis stream codec
```

That last commit is PR #8060 — and **#8060 merged to `main` on its own**, as
`09211d68d6 [OPIK-8164] [BE] fix: apply Jackson stream-read limits to the Redis
stream codec (#8060)`, followed by `4ef186c6c7 [OPIK-8192] [BE] fix: drop
undecodable scoring stream messages instead of wedging the stream (#8089)`.

The merged version is **not** what this branch holds. Verified:

| | this branch | `origin/main` |
|---|---|---|
| Pre-Guice Jackson config | anonymous `ConfiguredBundle` inline in `initialize()` | extracted to `JsonUtilsConfigurationBundle` (`OpikApplication.java:101`) |
| HTTP mapper limits | absent | `JsonUtils.applyStreamReadConstraints(...)` (`OpikApplication.java:153`) |
| Publish-time size guard | `onlineScoring.dropOversizedPayloads` + `maxPayloadBytes` in `OnlineScorePublisher` | **does not exist** |
| Undecodable-message handling | none | `UndecodableStreamMessage` / `UndecodablePayloadException` — consume-side drop, retried through `maxRetries` |

So `main` solved OPIK-8164 at the **consumer**, and deliberately: its
`UndecodablePayloadException` javadoc explains that `maxStringLength` is
configuration, that a higher-limit pod decodes the same bytes, and that this is
why the failure is retryable rather than deleted on sight. This branch's
publish-time `dropOversizedPayloads` guard is a **superseded design** that never
shipped.

**Fix:** drop the nine backend commits from this branch and keep only
`b799500973` (plus my `9eb372691d`) — i.e. rebuild the branch as QA-only on top
of current `main`. I have deliberately **not** done this: it rewrites published
history on a branch someone else may be tracking, and it is exactly the "take
either side wholesale" call that belongs to a person who can see both intents.

**Consequence if merged as-is:** the three Java conflicts below get resolved by
someone under time pressure, the branch's older publish-time guard wins on one of
them, and a shipped fix silently regresses.

---

## Should fix (2)

### 1. The recovery assertion's ordering barrier is timing, not construction — `online-evaluation-oversized-payload.spec.ts:186`

The `getSpan()` readback proves the oversized span reached ClickHouse before the
`after` span was written. It does **not** prove its scoring message reached
`stream_scoring_span_user_defined_metric_python` first — publication to the stream
is asynchronous. If the `after` message overtakes it, the spec passes green
without ever having tested the property it names. That is the specific failure
class this review exists to catch: a permanent false green.

In practice the readback of a 21M-character span takes seconds while the enqueue
is sub-second, so the ordering does hold today. But it holds by luck of timing.

**What I did:** I could not fix this honestly. Closing the gap needs an observable
enqueue or consumer position, and the suite has no read for one — there is no
Redis-stream helper on `backendClient`, and the rule log stream
(`getAutomationRuleLogs`) does not emit a line for a message the codec rejected
below the subscriber, so it cannot serve as the barrier either. Inventing a fixed
settling sleep would be worse than the bug.

Instead I **replaced the comment that claimed this was a barrier** with one that
records the limit precisely, so the next reader does not inherit the false
confidence. The real fix is a backend read (stream length, or last-delivered id)
exposed to the suite — that is its own piece of work.

### 2. The operational warning was measured against a build that no longer exists

`online-evaluation-oversized-payload.spec.ts:58-73`

The spec carries a stark warning: on `pr-8060.dev.comet.com` this write degraded
the whole deployment to 500/502 for ~2 minutes, so the spec "needs a disposable
environment or its own shard". That measurement was taken on the **#8060 preview
environment** — before both the merged codec fix and the OPIK-8192 undecodable
drop. The degradation may simply be gone.

Left as-is (the warning is the safe default) but **annotated** to say it predates
both merged fixes and needs re-measuring before anyone commits to permanently
sharding this spec. Someone should actually re-measure it against current `main`
before this lands in the nightly t3 set.

---

## Notes (4)

1. **`tsc` gate is dead estate-wide, and this PR did not break it.** `tsconfig.json:13`
   sets `baseUrl`, removed in TypeScript 7; `package.json` pins `"typescript": "^7.0.2"`.
   Neither file is in this diff — `tsconfig.json` last changed in #6677. **No
   TypeScript in this repo is being type-checked right now.** Its own PR, and
   worth opening promptly. Second, independent pre-existing error surfaced once I
   worked around it: `core/backend/client.ts` declares `deleteDashboard` twice
   (also present on `main`).

2. **Placement and tags are correct.** Directory `online-evaluation/` matches the
   area's `spec_dir: online-evaluation`; the spec is listed under `specs:`; both
   `@cap:` keys exist and are genuinely asserted — `python-rule-scores` by the
   `toBe(1.0)` on before/after spans, `scores-in-trace-panel` by the final step
   that opens the panel, selects the span and reads the score row. `@t3-nightly`
   is the right tier for a 21MB payload with a 600s budget.

3. **Convention parity is strong.** The spec is close to a mirror of its neighbour
   `online-evaluation-non-object-sections.spec.ts` — same imports, same `project` /
   `backendClient` / `testNamespace` / `automationRulesCleanup` fixtures, same
   `LogsPage` usage, same tag structure. No hand-rolled login, no in-body teardown,
   no `waitForTimeout`, no `.skip`. The read-back of `rule.type` / `samplingRate` /
   `enabled` before asserting anything is a genuinely good touch — it stops the
   spec silently proving a property about a stream it never touched.

4. **`client.ts` additions are bucket 1/2, not scope creep.** `SpanDetail`,
   `getSpan`, `pollSpanForFeedbackScore`, `createSpan` extend the established
   pattern; the `pollForFeedbackScore` generalisation in `poll-feedback-score.ts`
   is a clean refactor that preserves the trace-side signature. `postSeedWrite`
   and `withReadRetry` change shared seeding behaviour (bucket 2) — blast radius
   is every spec calling `createTraceWithSource`, which is 6 specs; the change is
   retry-on-5xx-only over an idempotent, caller-keyed upsert, which is safe, and
   `withReadRetry` is scoped to the new span read only.

---

## Conflicts against `main` — 6 files, NOT resolved

The branch is **74 commits behind `main`**. Merge-base: `9dde5a09d9`.

| File | This side | `main` side | A correct resolution keeps |
|---|---|---|---|
| `OpikApplication.java` | anonymous `ConfiguredBundle` calling `JsonUtils.configure()` pre-Guice | `JsonUtilsConfigurationBundle` (extracted) + `applyStreamReadConstraints` on the HTTP mapper | **`main` wholesale.** Strictly ahead; this side is #8060's pre-merge draft. |
| `OnlineScorePublisher.java` | publish-time `dropOversizedPayloads` / `maxPayloadBytes` guard | no publish guard; consume-side `UndecodableStreamMessage` drop | **`main` wholesale.** Different, shipped design — see Blocker 1. |
| `OnlineScorePublisherTest.java` | tests for the publish guard | tests for the merged shape | **`main` wholesale.** The tests follow their subject. |
| `taxonomy.yaml` | adds the oversized spec + rewrites two `online-evaluation` notes | alerts CRUD (#8117), taxonomy proposals (#8110, #8180), release-exploration specs (#8161, #8181) | **UNION.** Both PRs' areas, both `specs:` lists. Taking either side drops a merged spec's coverage entry. |
| `core/backend/client.ts` | `SpanDetail`, `getSpan`, `createSpan`, `postSeedWrite`, `withReadRetry`, rule `type` | helpers added by #8161, #8181, #8097 | **UNION.** Both PRs' helpers on the returned object. |
| `core/backend/index.ts` | `+ type SpanDetail` in the export list | other type exports added since | **UNION.** It is an export list; take both lines. |

The three e2e conflicts are ordinary being-behind-main unions. The three Java
conflicts are the blocker wearing a different hat.

### The reverse case git will not flag

All three e2e files this PR touches are **shared and hot**, and every one has
moved on `main` under this PR:

- `core/backend/client.ts` — last touched on `main` by #8161, #8181, #8097, #8093, #8016, #7986, #7994 …
- `core/backend/index.ts` — #8161, #8181, #8097, #8093 …
- `coverage/taxonomy.yaml` — #8161, #8180, #8181, #8117, #8097 …

These are the estate's three busiest generated-PR files. Any other open generated
PR sitting on a different branch of this stack will touch at least `taxonomy.yaml`,
and a clean merge there still loses work if resolved by taking one side. I could
not enumerate the currently-open PRs from this workspace (no GitHub access beyond
the supplied artifacts) — **that check is outstanding.**

### Does the conflict make the rest moot?

Partly, and read it in this order:

1. **Blocker 1 must be settled first.** Until the backend commits leave this
   branch, nothing else matters — the PR cannot merge, and five of the eleven bot
   comments are about code that should not be here.
2. **The spec-level findings survive it.** Should-fix 1 and 2 are properties of
   the QA commit and travel with it onto a rebuilt branch. So does everything I
   fixed.
3. **The three e2e conflicts are downstream of the rebuild.** Rebasing the QA
   commit onto current `main` resolves `client.ts`, `index.ts` and `taxonomy.yaml`
   by construction — but only if it is a rebase, not a merge that takes a side.

---

## Bot comments — 11 findings, all from `baz-reviewer[bot]`

| 1 | Taxonomy note claims a 25MB span is *scored*; spec sends 21M chars and never asserts that | **valid** | **fixed** in `9eb372691d` — note now states 21M characters and the actual property (spans logged *after* still score) |
| 2 | `rule-scope-thread-span` note calls span scope "incidental" when the spec drives it deliberately | **valid** | **fixed** in `9eb372691d` — "incidentally" dropped; `covered: false` unchanged and still correct |
| 3 | Cleanup can leak rule/project/traces during the documented outage window | **valid but house-wide** | **deferred → fixture-wide ticket.** `automation-rules.fixture.ts` warns-rather-than-throws *by design*, with a written rationale (a throwing teardown masks the assertion that explains the run). It predates this PR and is shared by all 6 online-evaluation specs. Fixing it here alone makes this spec the inconsistent one. The PR-specific grain is real though: this spec is the likeliest to hit teardown *during* an outage — worth carrying into that ticket as the motivating case. |
| 4 | Readback barrier does not prove stream ordering; recovery assertion can falsely pass | **valid** | **not fixed — escalated to Should-fix 1.** No observable enqueue/consumer read exists in this suite, and `getAutomationRuleLogs` emits nothing for a message rejected below the subscriber. Comment corrected so the claim is no longer made; real fix needs a backend read. |
| 5 | `attempted` misreports the request count when a 502 retries into a 400 | **valid** | **fixed** in `9eb372691d` — counter incremented per request instead of inferred from final status |
| 6 | `localGetSpan` duplicates `localGetTrace`'s mapping and 404 handling | **valid but house-wide** | **deferred → `client.ts` mapper refactor.** The identical `feedbackScores` mapping already appears at `client.ts:573` (traces) and `:1677` (threads) *before* this PR. This is the estate's existing shape, 2 of 3 occurrences pre-existing; extracting it for the third only is the wrong place to start. |
| 7 | `maxPayloadBytes` omits `jacksonConfig.getMaxStringLength()`, can still wedge | **moot here** | Criticises `dropOversizedPayloads`, which exists **only on this branch** and is absent from `main`. Dies with Blocker 1. |
| 8 | JSON codec admits undecodable messages; `StringCodec` vs `serializedSize` mismatch | **moot here** | Same — the `serializedSize` guard is branch-only. If the underlying codec concern still applies to `main`'s `RedisStreamCodec`, it needs its own backend issue, not this PR. |
| 9 | `serializedSize()` materialises the full byte array → heap amplification | **moot here** | Same — branch-only code path. |
| 10 | `dropOversized()` measures before `workspaceName` normalisation; metric mislabels | **moot here** | `OnlineScoringSampler.dropOversized` does not exist on `main` (verified: 0 occurrences). Branch-only. |
| 11 | Applying `JACKSON_MAX_STRING_LENGTH` pre-Guice can wedge existing streams on rolling upgrade | **valid, and already answered on `main`** | Not actionable here. `main`'s `UndecodablePayloadException` javadoc addresses exactly this: the limit is configuration, a higher-limit pod reads the same bytes, and that is why the failure is raised as *retryable* rather than deleted. The merged design already carries the compatibility path this asks for. |

**Follow-through I have not done, and want a go-ahead for:**

1. **File the two deferred findings** as one ticket each, by theme — (a) a
   retry/visibility policy for `automation-rules.fixture.ts` and the `project`
   fixture teardown (covers #3); (b) a shared entity mapper in
   `core/backend/client.ts` covering traces, spans and threads (covers #6). Plus a
   third, separate and more urgent: **the dead `tsc` gate** (Note 1).
2. **Reply on the nine threads I judged deferred, moot or already-answered**, so
   they read as answered rather than ignored. I have not posted anything.

---

## Fixes applied and pushed

Commit `9eb372691d`, pushed to `comet-qa-bot/OPIK-8164/e2e-oversized-online-scoring-payload`. Push accepted, no force.

- `taxonomy.yaml` — `python-rule-scores` note: 21M characters, not "25MB"; states the
  property actually asserted and says the oversized span's own score is deliberately
  not asserted. (bot #1)
- `taxonomy.yaml` — `rule-scope-thread-span`: "incidentally" → "exercised by".
  `covered: false` untouched. (bot #2)
- `client.ts` — `postSeedWrite` counts requests as it makes them. (bot #5)
- spec — the readback comment no longer claims to be an ordering barrier; it records
  what the barrier actually gives and what it does not. (bot #4, partial)
- spec — removed the reference to `REDIS_SCORING_DROP_OVERSIZED_PAYLOADS`, a flag that
  exists only on this branch and would dangle the moment Blocker 1 is fixed; replaced
  with the configuration-dependence that is true on `main`.
- spec — fixed a garbled sentence in the `OVERSIZED_CHARS` docblock, and annotated the
  operational warning as pre-dating both merged fixes. (Should-fix 2)

**No taxonomy capability was invented.** The spec's real subject — "online scoring
survives an undecodable stream message" — has no capability key in
`online-evaluation`. The PR handles this by extending `python-rule-scores`'s note
and leaving `rule-scope-thread-span` honestly `false`, which is defensible. If you
would rather it counted as its own gap, that is your call to make, not mine.

---

## Estate gates

`tag_lint` **PASS** (55 specs, 1 exempt, 0 problems) · `tsc` **FAIL — pre-existing** · `playwright --list` **PASS** (spec collects, 1 test)

All three re-run after my edits.

### What I could not verify

- **`tsc` cannot pass in this repo at all**, on this branch or on `main`, because of
  the `baseUrl`/TypeScript-7 mismatch (Note 1). To get any signal I type-checked
  against a temporary config with `baseUrl` removed and `NodeNext` resolution: the
  only errors are the two pre-existing `deleteDashboard` duplicates, **identical
  before and after my edits**. So: this PR introduces no type errors — but that is
  a workaround's word, not the gate's.
- **The spec has never been run here.** No Opik deployment in this workspace. Every
  claim about scoring latency, the 21MB write succeeding, and the 500/502
  degradation comes from the spec's own comments, which describe a preview
  environment that no longer matches `main`.
- **Should-fix 1 is reasoned, not observed.** I did not race the enqueue to prove the
  ordering gap is reachable; I established that nothing in the suite *guarantees* it
  is not.
- **Open-PR overlap on the shared files is unchecked** — no GitHub access beyond the
  supplied artifacts. Given `taxonomy.yaml` and `client.ts` are the estate's two
  busiest generated-PR files, someone should confirm before resolving those unions.

**This PR is NOT ready.** It stays a draft, and it should — Blocker 1 has to be
settled by a person. I have not marked it ready and have not merged it.

---
_[review_generated_tests.yml](https://github.com/comet-ml/comet-automation-tests/actions/runs/34179812581) ·


[OPIK-8164]: https://comet-ml.atlassian.net/browse/OPIK-8164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OPIK-8164]: https://comet-ml.atlassian.net/browse/OPIK-8164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

CometActions:
## 🔍 Generated-test review

## PR #8080 — [OPIK-8164] [QA] Proposed e2e spec from the #8060 exploration: an oversized online-scoring payload must not stop the rule

CometActions · draft: yes · 5 files after restack (was 14), 1 new spec
Branch: `comet-qa-bot/OPIK-8164/e2e-oversized-online-scoring-payload`
Pushed as `870dd952d9` (`--force-with-lease`, accepted — the restack rewrote history).

**VERDICT: ready with notes.** The merge blocker is gone: the branch is restacked
onto `main`, all six conflicts are resolved, and the estate's own gates pass. The
spec itself is one of the better generated ones — honest tags, a control before
and after, a discriminating seed barrier, and its own limits written down rather
than papered over. A human should flip the draft flag; this review will not.

---

### Merge conflicts — resolved (the blocker, now cleared)

The branch arrived `CONFLICTING` with six files. The restack dropped 9 inherited
commits from the merged source PR #8060, which removed three of the six outright;
the remaining three were mine to resolve and are.

| File | Base (`main`) side | This PR's side | Resolution |
|---|---|---|---|
| `apps/opik-backend/.../OpikApplication.java` | the OPIK-8164 Jackson limit, as merged | the same change, inherited | **dropped** — inherited commit, not this branch's work |
| `apps/opik-backend/.../OnlineScorePublisher.java` | merged oversized-payload guard | same, inherited | **dropped** |
| `apps/opik-backend/.../OnlineScorePublisherTest.java` | merged tests for it | same, inherited | **dropped** |
| `tests_end_to_end/coverage/taxonomy.yaml` | added two specs to `specs:`, flipped `list-rules` to `covered: true` with a note about the six bracket-opening judge prompts | added its own spec, rewrote the `python-rule-scores` and `rule-scope-thread-span` notes, still carried `list-rules: { covered: false }` | `specs:` unioned (all three specs). `python-rule-scores`: base's sentence **plus** this PR's t3 clause — each describes coverage the other cannot see. `list-rules`: **base's side** — the base earned that flip with a real spec; this branch was only carrying a stale `false` it never changed. `rule-scope-thread-span`: this PR's side, including the comment explaining why it stays `false` |
| `tests_end_to_end/e2e/core/backend/client.ts` | added `TracePayload` | added `SpanDetail` | **both kept.** The conflict opened mid-docblock — a bare concatenation would have left `SpanDetail`'s comment body with no `/**`. Bridged with `TracePayload`'s closing `}` plus a fresh `/**`, indentation matched |
| `tests_end_to_end/e2e/core/backend/index.ts` | `type TracePayload` | `type SpanDetail` | **both kept** — the export block is append-only |

Everything resolved is inside `tests_end_to_end/`. Nothing outside it was touched:
the three backend conflicts were inherited commits and the restack dropped them, so
no production code was resolved, reverted, or re-applied. The diff against
`origin/main` is now 5 files, all under `tests_end_to_end/`.

**No fixture-chain conflict.** This PR adds no fixture, so `fixtures/index.ts` was
never in the diff — the linked-list re-parenting did not apply here.

---

### Blockers (0)

None.

---

### Should fix (0 outstanding — 3 applied)

Applied in `870dd952d9`:

1. The oversized-span readback poll used a bare `timeout: 180_000` sitting beside a
   named `SCORING_TIMEOUT_MS` of the same value, inviting a reader to assume one
   budget where there are two. Given its own `OVERSIZED_READBACK_TIMEOUT_MS` with a
   note on why ingest-and-query wants to move independently of scoring.
2. `expect(await panel.readFeedbackScoreValue(ruleName)).toBe(1.0)` carried no
   reason, unlike every other assertion in the spec. Added one.
3. Conflict resolution above.

---

### Notes (4)

1. **The `tsc` gate cannot pass in this workspace, on this branch or on `main`.**
   `npx tsc --noEmit` fails at `tsconfig.json(13,5): error TS5102: Option 'baseUrl'
   has been removed` — the workspace has TypeScript **7.0.2** installed and the
   repo's `tsconfig.json` still sets `baseUrl`. The PR does not touch
   `tsconfig.json` (`git diff origin/main...HEAD -- tsconfig.json` is empty), so
   this is a toolchain/config mismatch in the estate, not a defect in this PR. I
   re-ran with `baseUrl` stripped to actually type-check the code: the only
   non-`@types/node` errors in this PR's files are two pre-existing
   `Duplicate identifier 'deleteDashboard'` at `client.ts:764` and `:1333`, which
   are **already on `origin/main`** (lines 674 and 1243 there) and which this PR
   does not touch. The new spec produces **zero** type errors. Worth its own PR:
   drop `baseUrl` and use `"paths": {"@e2e/*": ["./*"]}`, and install `@types/node`.
2. **Shared-file overlap with other open generated PRs — the reverse case git will
   not flag.** This PR moves four shared files: `taxonomy.yaml`, `client.ts`,
   `index.ts`, `poll-feedback-score.ts`. The first three have all moved under other
   QA-generated PRs very recently — `#8161`, `#8180`, `#8181`, `#8097` are the last
   commits on them — which is exactly the population this branch is stacked among.
   A *clean* merge can still lose work if two generated PRs sit on different
   branches of one stack and both append to `client.ts`'s type block or
   `taxonomy.yaml`'s `specs:` list. The restack onto current `main` removes the
   exposure as of this run; if another generated PR merges before this one, re-run
   the restack rather than trusting a green mergeability check.
   `poll-feedback-score.ts` has not moved since `#6874`, so it is uncontended.
3. **The spec carries an operational warning that a reader must act on.** Its
   docblock records that on the `#8060` preview environment an oversized span
   meeting an active rule degraded the whole deployment — 500/502 for ordinary
   writes for ~2 minutes. It also says, candidly, that this was measured *before*
   the OPIK-8164 codec fix and the OPIK-8192 drop landed, and has not been
   re-measured since. Someone should re-measure against current `main` before this
   is scheduled on a shared nightly environment; if the degradation is gone, the
   "needs its own shard" caveat should come out of the docblock.
4. **`rule-scope-thread-span` correctly stays `covered: false`.** The spec does
   drive a span-scope rule end to end, but the key names *both* the span and thread
   scope flags and nothing exercises thread scope. Flipping it on the span half
   would retire the thread half untested. The taxonomy comment says exactly this.
   Left as-is; no capability was invented.

---

### Tag honesty

Both `@cap:` keys exist in `taxonomy.yaml` and both are earned:

- `@cap:online-evaluation.python-rule-scores` — a python rule really does score,
  asserted three times (`before`, `after`, and value `1.0` from the constant metric).
- `@cap:online-evaluation.scores-in-trace-panel` — the final step opens the real
  trace panel, selects the span, opens the Feedback scores tab and asserts both the
  row count and the value. Not an API-only claim on a UI capability.

`create-python-rule` is deliberately **not** claimed: the rule is created over the
API, and the area reserves that key for the smoke test that creates through the UI.
That restraint is the right call. Tags are string literals, so they count. Tier
`@t3-nightly` matches a spec that writes 21M characters and budgets 10 minutes —
no tier inflation. Placement matches `spec_dir: online-evaluation`, and the spec is
listed in the area's `specs:`. Tag pairing is identical to
`online-evaluation-non-object-sections.spec.ts`, the nearest neighbour.

### Support code

`client.ts` (+201), `index.ts` (+1) and `poll-feedback-score.ts` (+47/-14) are
bucket 1 and 2 — extending established patterns, not new infrastructure.
`poll-feedback-score.ts` extracts a generic poller and adds a span counterpart;
`pollTraceForFeedbackScore`'s signature is unchanged, so the seven specs importing
it are unaffected. `localGetSpan` follows `localGetTrace`/`localGetOptimization`.
`createTraceWithSource` gains 5xx retry, which is a strict improvement for its two
other callers, both in this same area. Teardown is the existing
`automationRulesCleanup` fixture — nothing in the test body — and entities are
namespaced with `testNamespace`, so the estate's orphan sweep reaches them. No
`waitForTimeout`, no `.skip`, no ordering dependence between tests.

---

### Bot comments

14 comments. 5 are on backend production files that the restack **dropped** from
this branch — they belong to the merged source PR #8060 and are not this PR's to
answer. 3 are the bot's own acknowledgements that commit `9eb3726` addressed it.
That leaves 6 with a disposition:

| 1 | `taxonomy.yaml:650` — note describes a 25MB span but the spec sends 21M chars | valid | **fixed** in `9eb3726`; bot confirmed. Carried through the conflict resolution intact |
| 2 | `taxonomy.yaml:657` — "driven incidentally" misrepresents a deliberately-created span-scope rule | valid | **fixed** in `9eb3726`; bot confirmed. Carried through intact |
| 3 | `client.ts:530` — `attempted` derived from final status, so a 502-then-400 reports one attempt | valid | **fixed** in `9eb3726`; bot confirmed |
| 4 | `spec:177` — readback barrier can falsely pass; the `after` message could overtake the oversized one, since stream publication is async | **valid, and the sharpest of the fourteen** | **Documented, not closed** — `9eb3726` replaced the guarantee with an explicit KNOWN LIMIT paragraph saying the ordering holds by timing, not by construction. The bot's own fix ("an observable enqueue/consumer ordering condition") needs a read of the Redis consumer position that this suite does not have. **Defer → a follow-up for a consumer-position/enqueue read on the backend client**, without which no spec in this area can assert stream ordering by construction. Honest documentation is the right interim answer; inventing a barrier that does not barrier would be worse |
| 5 | `spec:82` — cleanup can leak during the documented 500/502 outage; the fixture only warns | valid observation, **but it criticises a deliberate estate convention** | **Defer → estate-wide, its own PR.** `automation-rules.fixture.ts` documents at length why it is best-effort: "a cleanup failure warns rather than throws, so it cannot mask the assertion or scoring error that explains the run." The bot's alternative — fail the test on cleanup failure — would reverse that decision for every spec using the fixture, and would mask the real failure. The residual leak is bounded: entities are `testNamespace`-prefixed, so global-setup's orphan sweep reaches them. Changing the retry policy here alone would make this the one inconsistent caller |
| 6 | `client.ts:605` — `localGetSpan` duplicates `localGetTrace`'s mapping and 404 handling | valid, **but house-wide** | **Defer → estate-wide, its own PR.** A per-entity local mapper with `try`/`isNotFoundError` is already the convention: `localGetTrace` and `localGetOptimization` both do it, and the shapes differ in their tail fields (`input` vs `traceId`/`output`). Extracting a shared mapper is a refactor of a 2,500-line client, not a change to smuggle in with one spec — doing it only for the new function makes the new function the odd one out |

I have **not** posted replies on these threads or filed the two deferred tickets —
that needs your go-ahead. I can draft both: one ticket for the consumer-position
read (#4) and one for the shared entity-mapper plus cleanup-retry policy (#5, #6),
grouped by theme rather than one per finding.

---

### Estate gates: tag_lint **PASS** · tsc **FAIL (pre-existing, not this PR)** · playwright --list **PASS**

```
tag-lint: 65 specs checked, 1 exempt, 0 problem(s)
tsc:      tsconfig.json(13,5): TS5102 'baseUrl' has been removed  <- identical on origin/main
playwright --list: 1 test in 1 file, collected
```

All three re-run after every edit. `playwright --list` and `tag_lint` both went
FAIL → PASS purely from finishing the restack; neither was a defect in the spec.

---

### What I could not verify

- **The spec has not been run.** No Opik deployment is reachable from this
  workspace, so `--list` collection is as far as I got. Everything about whether
  the property actually holds — that a 21M-character span does not wedge
  `stream_scoring_span_user_defined_metric_python`, that the `after` span scores,
  that the panel renders the row — is unexecuted. This matters more than usual
  here: the spec's whole subject is a race, and note 3's deployment-degradation
  warning is exactly the kind of thing only a real run settles.
- **`tsc` as the estate defines it.** I could not make the repo's own `tsconfig.json`
  compile under the installed TypeScript 7.0.2. The clean run I did get required
  stripping `baseUrl` and tolerating missing `@types/node`, so "zero type errors in
  this PR's files" is a claim about that modified run, not about the gate CI runs.
- **Whether other open generated PRs are currently mid-flight on the same shared
  files.** I checked merged history on `origin/main` (note 2); I have no view of
  the open-PR set from here, so the stack-sibling risk is reasoned from the commit
  pattern rather than observed.
- **The ordering barrier's real-world margin.** The spec claims a 21M-character
  readback takes seconds while the enqueue is sub-second. I did not measure either;
  that is the spec author's figure, and bot finding #4 is right that nothing
  enforces it.

---
_[review_generated_tests.yml](https://github.com/comet-ml/comet-automation-tests/actions/runs/34187869028) ·


[OPIK-8164]: https://comet-ml.atlassian.net/browse/OPIK-8164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

CometActions:
## 🔍 Generated-test review

# PR #8080 — [OPIK-8164] [QA] Proposed e2e spec from the #8060 exploration: an oversized online-scoring payload must not stop the rule

CometActions · draft: yes · 5 files, 1 new spec · head was `870dd952d9`, now `70454f6ad2`

**VERDICT: ready with notes — but one note is a scheduling decision a human has to make before this runs anywhere shared.**

The conflict is resolved and the branch merges clean. The spec is honest: both
its `@cap:` tags are earned by real assertions, it is in its `spec_dir`, it
reuses the estate's fixtures and POMs, and it declines to flip
`rule-scope-thread-span` even though it drives a span-scope rule end to end —
which is the right call, since that key covers both scopes and nothing here
touches threads. The support code in `client.ts` / `poll-feedback-score.ts` is
bucket-1/2 test infrastructure extending established patterns, not scope creep.

---

## Merge conflict (resolved)

**1 conflicting file, `tests_end_to_end/coverage/taxonomy.yaml` — inside the
test tree, so it was mine to resolve. Nothing outside `tests_end_to_end/` was
touched.**

| Side | What it did |
|---|---|
| base (`origin/main`, #8199 `66589738ee`) | Sorted every area's `specs:` list alphabetically and taught `tag_lint.py` to **enforce** sortedness (Rule 4: "`specs:` lists stay sorted, so concurrent additions do not collide"). |
| PR head | Appended `online-evaluation-oversized-payload.spec.ts` to the **end** of the old unsorted list. |

Both sides claimed the same eight-line region. **A correct resolution keeps
both**: main's sorted ordering of the eight existing entries, with the new spec
inserted in sorted position — between `non-object-sections` and
`python-metric-errors`. Taking main's side drops the new spec from its area and
makes it invisible to the coverage builder; taking the PR's side re-breaks the
sort and now fails `tag_lint` Rule 4 outright.

The capability `note:` blocks did **not** conflict — main left them untouched
while this PR rewrote two of them, so git merged them cleanly and no meaning had
to be unioned by hand.

Verified after resolution: the merged tree differs from `origin/main` in exactly
the five files this PR owns (`git diff --cached origin/main --stat`). The merge
commit's first parent is the PR head; main's own commits came across wholesale.

**Does the conflict moot the other findings?** It did — a spec that cannot merge
does not need its budget arithmetic polished. It is resolved and pushed, so
everything below now stands on its own.

---

## Blockers (0)

None.

## Should fix (1)

**1. The spec's own operational warning is at odds with the tier it carries —
`online-evaluation-oversized-payload.spec.ts:68-82`, tag `@t3-nightly`.**

The docstring says, in its own words: *"treat this spec as needing a disposable
environment or its own shard"*, because on `pr-8060.dev.comet.com` the oversized
write degraded the **whole deployment** — every caller getting 500/502 for
around two minutes, repeatably.

`TESTING-TAGS.md:37` says `@t3-nightly` means *"nightly on staging"*, and
`playwright.config.ts` has one project (`chromium`) with no tier shards — tiers
are selected by `--grep` at the CI call site. So as tagged, this spec is
scheduled to run **on the shared staging environment alongside every other t3
spec**, which is precisely what its own warning says not to do. If the
degradation still reproduces, one nightly run of this spec takes down the specs
around it, and they fail for a reason none of them names.

Why I did not fix it: the fix is a decision, not an edit. It is either (a)
re-measure against a build carrying the OPIK-8164 codec fix and the OPIK-8192
undecodable-message drop — the docstring itself says this has not been done and
the degradation "may well be gone" — and then delete the warning, or (b) give
the spec its own CI shard / disposable environment. Both belong to whoever owns
the nightly schedule. **Do not merge this into the nightly rotation until one of
those has happened.**

## Notes (4)

**2. `tsc` fails, and none of it is this PR's.** Two independent pre-existing
causes, both reproducible on `origin/main`:

- `tsconfig.json(13,5): error TS5102: Option 'baseUrl' has been removed.` The
  installed compiler is TypeScript **7.0.2**, which dropped `baseUrl`;
  `tests_end_to_end/e2e/tsconfig.json` still sets it and is byte-identical to
  `origin/main` (unchanged since #6677 `2930c31491`). This aborts the compiler
  before a single file is checked, which is why the gate says nothing about the
  actual code.
- With `baseUrl` dropped and `"types": ["node"]` supplied, the **only** errors
  in the whole estate are `core/backend/client.ts(764,11)` and `(1333,11)`:
  `TS2300: Duplicate identifier 'deleteDashboard'`. Two members of the same
  returned object literal share a name — the second (a raw `fetch`) silently
  wins over the first (the SDK client). Both are on `origin/main` at lines
  674/1243, untouched by this PR.

Under that corrected config **this PR's own TypeScript compiles clean**. Both
defects are estate-wide and belong in their own PR — the tsconfig fix would
change the build for all 65 specs, and de-duplicating `deleteDashboard` is a
behaviour call on shared client code. I can open that PR on request; I did not
smuggle it in here.

**3. Shared files, and the cross-PR risk git will not show you.**
`git log --oneline 87e5dc147e..origin/main -- <path>`:

| File | Moved on main under this PR? |
|---|---|
| `coverage/taxonomy.yaml` | **Yes** — #8199 (`66589738ee`) |
| `coverage/tag_lint.py` | **Yes** — #8199, same commit |
| `core/backend/client.ts` | No |
| `core/backend/index.ts` | No |
| `core/backend/poll-feedback-score.ts` | No |

`taxonomy.yaml` is the file this PR conflicted on and it is the one that moves
most, so any other open generated PR adding a spec is on a collision course with
this one for the same reason — that is exactly what #8199 was for, and this
resolution honours it. `client.ts` has **not** moved, but this PR changes it
substantially and two behaviours now reach specs it never mentions:
`postSeedWrite`'s 5xx retry now covers `createTraceWithSource`
(3 specs: `non-object-sections`, `sampling-rate`, and this one), and
`pollTraceForFeedbackScore` was refactored onto a shared generic
(7 specs, including `annotation-queue-scoring`). Both changes preserve the
existing signatures and only add retry on 5xx, so the blast radius is benign —
but a second open PR editing `client.ts` would merge clean and still lose work.

I could not enumerate other open PRs: this run has the workspace snapshot only,
not a GitHub query.

**4. The fixture chain is not in play.** This PR adds no fixture and does not
touch `fixtures/index.ts`, so the line-1 re-export / re-parenting problem does
not arise. It consumes `automationRulesCleanup` (which already exists on main)
and adds nothing to the chain.

**5. `rule-scope-thread-span` stays `covered: false`, and that is correct.** The
spec creates a span-scope rule and reads its `type` back to prove it, so it
genuinely exercises the span half. The PR leaves the key false with a note
saying so, because flipping it would retire the untested thread half on the
strength of the span half. That is the right judgement and I left it alone. If
anyone wants the span half to count, the answer is a **new capability key** —
which is a human's call about the coverage denominator, not something this
review will invent.

---

## What I fixed

1. **Merged `origin/main`** and resolved the `taxonomy.yaml` specs-list conflict
   as a sorted union (above).
2. **Budgeted the test timeout** (`online-evaluation-oversized-payload.spec.ts`).
   The three polls — `SCORING_TIMEOUT_MS` twice plus
   `OVERSIZED_READBACK_TIMEOUT_MS` — can legitimately consume 540,000ms, against
   a flat `test.setTimeout(600_000)`. That left 60,000ms for the
   21,000,000-character write, the rule creation and both trace-panel loads, so
   Playwright could abort mid-UI-step and report *its own* timeout. "The run
   timed out" reads nothing like "the scoring stream wedged", and the second is
   the only failure this spec exists to report. `test.setTimeout` is now derived
   from the budgets it has to cover plus a named `SETUP_AND_UI_BUDGET_MS`, so
   raising any one budget cannot silently re-create the squeeze.

Committed as `70454f6ad2` and pushed to
`comet-qa-bot/OPIK-8164/e2e-oversized-online-scoring-payload`. Plain push, no
force — this run did not rewrite history.

---

## Bot comments

16 threads from `baz-reviewer[bot]`. Three are the bot's own acknowledgements of
the previous review round and need no verdict.

| 1 | `python-rule-scores` note describes a 25MB span; spec sends 21M chars | valid | already fixed in `9eb3726` (prior round); bot confirmed in #13 |
| 2 | `rule-scope-thread-span` note calls span scope "incidental" | valid | already fixed in `9eb3726`; bot confirmed in #14 |
| 3 | Cleanup leaks outage artifacts: rule + traces survive a failed teardown | **invalid** | Central claim is wrong. `global-teardown.ts:55` sweeps automation rule evaluators by run prefix and `:59` sweeps projects; its own comment (`:43-45`) says traces are omitted deliberately because they cascade with their project. Entities are named from `testNamespace` = `${cujPrefix}-w<worker>-<slug>`, so the prefix sweep reaches them. The warn-not-throw the bot objects to is deliberate and documented in `automation-rules.fixture.ts` — a throwing teardown masks the scoring failure that explains the run. |
| 4 | `getSpan()` readback does not prove the oversized message was enqueued first | valid | Real, and unfixable with what the suite can read. Already **disclosed** in the spec (`:192-203`): the barrier proves ingest order, not stream order, and it holds by timing not construction. Closing it needs an observable enqueue or consumer position — a backend test hook. Deferred → the OPIK-8164 follow-up; the honest disclosure is the right interim, not a silent gap. |
| 5 | `attempted` misreports retry count when a 502 retries into a 400 | valid | already fixed in `9eb3726`; bot confirmed in #12 |
| 6 | `localGetSpan` duplicates `localGetTrace`'s score mapping | valid, **house-wide** | The identical `feedbackScores` mapping appears **three** times in `client.ts` — `:700` (trace) and `:2210` (thread) both predate this PR. Extracting it only for the new one makes the new one the inconsistent file. Deferred → a `client.ts` mapper-extraction PR, not this one. |
| 7 | `maxPayloadBytes` omits `jacksonConfig.getMaxStringLength()` | n/a | **Not in this PR.** `apps/opik-backend/.../OnlineScorePublisher.java` is not among the 5 changed files; belongs to the OPIK-8164 backend change. |
| 8 | JSON codec admits undecodable messages (`StringCodec` vs `JsonUtils`) | n/a | Not in this PR — same file, same backend change. |
| 9 | `serializedSize()` heap amplification | n/a | Not in this PR — same file, same backend change. |
| 10 | `dropOversized()` measures before `workspaceName` normalisation | n/a | Not in this PR — `OnlineScoringSampler.java` is not in the diff. |
| 11 | Jackson limit change wedges existing scoring streams | n/a | Not in this PR — `OpikApplication.java` is not in the diff. |
| 12 | (bot: "`9eb3726` addressed #5") | — | acknowledgement, no action |
| 13 | (bot: "`9eb3726` addressed #1") | — | acknowledgement, no action |
| 14 | (bot: "`9eb3726` addressed #2") | — | acknowledgement, no action |
| 15 | 21MB is not oversized against a shipped 100MB `maxStringLength` | **partially valid** | The spec is still a real regression guard: what it guards is the codec falling back to Jackson's `DEFAULT_MAX_STRING_LEN` (20,000,000), at which 21,000,000 becomes undecodable and wedges the stream — caught by the `after`-span assertion. What it cannot prove is that the oversized path was exercised on a deployment whose codec really carries 100MB, and both the spec (`:221-227`) and the taxonomy note already say exactly that. Deriving the limit needs a read of the deployment's effective Jackson config that this suite does not have. No change; the scoping is already honest. |
| 16 | Three independent 180s polls exhaust the 600s test timeout | valid | **fixed here** — `test.setTimeout` is now derived from the budgets plus named headroom (see "What I fixed"). |

Findings 7–11 are five of sixteen, all on backend Java the PR does not touch.
They are worth forwarding to whoever owns OPIK-8164/OPIK-8192 — two are marked
`high` — but they are not this PR's to answer.

**Offered, not done:** (a) reply on threads 3, 4, 6 and 15 so the verdicts are
visible to the next human rather than reading as silence; (b) file the deferred
ones as **one** ticket per theme — a `client.ts` mapper extraction (#6), and an
observable enqueue/consumer read for online-scoring specs (#4). Say the word and
I will draft both.

---

## Estate gates

Re-run after every edit, from the repo root, at `70454f6ad2`:

| Gate | Result | |
|---|---|---|
| `tag_lint.py --taxonomy … --estate tests_end_to_end` | **PASS** | 65 specs checked, 1 exempt, 0 problems — including #8199's new sortedness rule |
| `npx tsc --noEmit` | **FAIL** | `TS5102` on `tsconfig.json`, pre-existing and identical on `origin/main`. See note 2 — this PR's own code typechecks clean under a corrected config. |
| `npx playwright test --list` | **PASS** | the new spec collects: 1 test in 1 file |

---

## What I could not verify

- **The spec has not been run.** No Opik deployment was reachable from this
  workspace, so every runtime claim — the ~5s scoring latency, the 21M-character
  readback, whether the deployment degradation still reproduces post-fix — rests
  on the exploration's measurements in the docstring, not on anything I
  observed. Should-fix #1 exists precisely because that re-measurement is the
  open question.
- **Whether the oversized message actually reaches the codec ahead of the
  `after` span.** Bot finding #4 is correct that the readback barrier does not
  establish it, and I have no way to observe the stream to settle it either way.
- **Other open generated PRs touching `taxonomy.yaml` / `client.ts`.** I checked
  what has landed on `origin/main` (note 3) but could not query open PRs from
  this workspace.
- **That the pre-existing `deleteDashboard` duplicate is harmless.** I confirmed
  it predates this PR and that the second definition wins; I did not trace which
  specs depend on which behaviour.

---

**This PR is a draft and I have left it a draft.** I have not marked it ready
and have not merged it — that flag is a person's to flip, and should-fix #1 is
the thing to settle before they do.

---
_[review_generated_tests.yml](https://github.com/comet-ml/comet-automation-tests/actions/runs/34202852369) ·


[OPIK-8164]: https://comet-ml.atlassian.net/browse/OPIK-8164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

AndreiCautisanu (approved):

PR #8084, opened 2026-08-31
[OPIK-8176] [QA] Proposed e2e specs for the Datasets list enrichment (from #8076 exploration)

CometActions:
## Details

### Where these came from

Exploratory QA of **#8076** ([OPIK-8176] [BE] perf: run the four dataset enrichment queries concurrently), on that PR's own deployed environment `https://pr-8076.dev.comet.com` (`2.2.45-8076-merge-3150`). A human worked the flows by hand there first; these two specs are the flows that were verified, deterministic, and seedable end to end without a provider key.

**This PR targets `danield/OPIK-8176-concurrent-dataset-enrichment`, not `main`,** because the specs were written and run against that branch's head (`a2aea24`) and assert the behaviour it produces. #8076 is still open, so basing on `main` would put specs on top of a code path that has not landed. When #8076 merges, this should be retargeted to `main` and re-run.

### What the exploration found, and why it is worth a permanent test

`enrichDatasetWithAdditionalInformation` produces every number on `/projects/{id}/datasets` and on the shared `/projects/{id}/test-suites` table. #8076 rewrites it to fan four lookups out concurrently and zip the answers back onto the rows. Nothing in the estate asserted those numbers: `dataset-crud-smoke` checks the row is *visible*, and `dataset-version-counters` reads `GET /datasets/{id}/versions` rather than the enriched list.

Both of the failure modes the change could introduce are silent — a lookup that returns nothing read as a legitimate zero, and a zip that lands one dataset's counters on another's row. Either leaves a page that looks entirely healthy.

The exploration found no defect: all 9 checks worked. These specs exist so a future change to the same path cannot break it quietly.

### 1. `tests_end_to_end/e2e/tests/datasets/dataset-list-enrichment.spec.ts`

`@t2-cuj` · `@area:datasets` · `@cap:datasets.list-datasets` · surface: **both**

Seeds two datasets in one project — one with 7 items plus an experiment (2 experiment items) and an optimization, one with 3 items and nothing else — then asserts:

- the project-scoped list settles on **exactly those two rows** (`total` and the row set, not "our rows are in there"), each with its own item / experiment / optimization counts and the recency timestamps present only on the used one;
- `GET /datasets/{id}` for each dataset equals the row the multi-row list returned for it, **field for field including timestamps**. Batch enrichment and single-element enrichment share a code path with different `ids` sets, so the two agreeing is a fact about the fan-out;
- the Datasets page renders `7` and `3` on their own rows, and — with **Most recent optimization** enabled via the Columns picker, since it is not in the default set — a populated timestamp cell for the used dataset and the empty dash for the sibling.

Cells are addressed by identity (`data-row-id` / `data-cell-id`) and asserted `toHaveCount(1)` first, because column order in this table is user-configurable and persisted.

**Verification: PASSED** against `https://pr-8076.dev.comet.com` (6.6s).

### 2. `tests_end_to_end/e2e/tests/datasets/dataset-list-version-count.spec.ts`

`@t2-cuj` · `@area:datasets` · `@cap:datasets.list-datasets` · surface: **both**

With versioning on, the list's Item count is the latest version's `items_total`, and `latest_version` now arrives from a MySQL transaction opened separately from the dataset row. A version lookup that comes back empty falls back to the legacy `dataset_items` count with no error anywhere — the page still renders a number, just the wrong one.

The spec commits a second version that rewrites one existing item and adds two new ones, so `items_total` (5), `items_added` (2) and `items_modified` (1) are three different numbers and a fallback to a stale or last-batch-only count produces 3 or 2 rather than 5. It asserts the list moves `v1 → v2` and `3 → 5`, that `GET /datasets/{id}/versions` stores that total, that the dataset really holds 5 items with the rewritten id updated in place rather than duplicated, and that the page renders `5`.

**Verification: PASSED** against `https://pr-8076.dev.comet.com` (3.7s, first run).

**Dropped 1 of 3 candidates.** The exploration also proposed a Test suites variant (`test-suites.list-suites`): the Test suites page shares `DatasetListPage` and the same enrichment, so its row should show the same counters. It was ranked **weak** by the exploration itself, on the grounds that it re-asserts the datasets spec's claims against the same backend path and is only worth having if the datasets spec can be parameterised over both entity types rather than written twice. Two strong candidates were available, so it is out. If a reviewer wants suites covered, the right shape is parameterising spec 1, not a third file.

Also **not** written: a regression test for the pre-existing bug the exploration tripped over — `GET /v1/private/datasets?project_id=…&with_experiments_only=true` ignores `project_id` (`fetchUsingMemory` never passes `criteria.projectId()` to `findByIds`). #8076 does not touch that branch, so it is neither this PR's regression nor this PR's test.

### Supporting changes

- `fixtures/enriched-datasets.fixture.ts` — new. Seeds the two-dataset shape with its experiment, optimization and traces, verifies the item counts really landed before the test starts, and tears everything down after `use()` (honouring `OPIK_LEAVE_FAILURES`). Traces do not cascade with the project, and the run-prefix sweep does not know about optimizations, so both are deleted explicitly. Chained at the tail and re-exported from `fixtures/index.ts`.
- `core/backend/client.ts` — `listEnrichedDatasets` and `getEnrichedDataset`, sharing one mapper so the list-vs-by-id comparison is not comparing two different mappings of the same answer. Every enriched field maps to `T | null` rather than defaulting to `0`: the failure being hunted is a silent zero, and a `?? 0` mapper would make an absent value indistinguishable from a real one.
- `pom/datasets.page.ts` — `cell(datasetId, columnId)` and `showColumn(label)`. No new `data-testid` was needed: `columns-button` already exists and the picker entries are addressable by role and name.
- `coverage/taxonomy.yaml` — both specs added to the `datasets` area's `specs:` list. `list-datasets` was already `covered: true` at `t1-smoke`; a `note:` now records what the t2 specs add on top of row presence.

No new `@cap:` keys were invented and nothing was flipped that these specs do not actually exercise. The Test suites capability stays as it was.

### How it was verified

```bash
cd tests_end_to_end/e2e
npx playwright test tests/datasets/dataset-list-enrichment.spec.ts \
                    tests/datasets/dataset-list-version-count.spec.ts --reporter=list
# then, since a shared POM was touched:
npx playwright test tests/datasets/ --reporter=list      # 10 passed
python3 ../coverage/tag_lint.py --taxonomy ../coverage/taxonomy.yaml --estate ..   # 0 problem(s)
```

Two notes for the reviewer, neither caused by this change:

- **Running against `pr-8076.dev.comet.com` needs a placeholder `OPIK_API_KEY`.** The `opik` TS SDK treats any host ending in `comet.com` as cloud and refuses to construct without a key, but the PR environment is an OSS install with no auth. Setting any value gets past it. This does not arise in CI, which targets `localhost:5173`.
- **`npx tsc --noEmit` fails on the estate's own `tsconfig.json`** before it typechecks anything: `Option 'baseUrl' has been removed` (typescript `^7.0.2`). Typechecking with an equivalent `baseUrl`-free config leaves only a pre-existing `Duplicate identifier 'deleteDashboard'` in `core/backend/client.ts` — nothing from these files. Both are worth a separate fix.

### Generated by the release QA side flow

These specs were written by the QA test-proposal automation from a human-verified exploration, and they need review before merge. Draft on purpose.

Source PR: comet-ml/opik#8076

[OPIK-8176]: https://comet-ml.atlassian.net/browse/OPIK-8176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Change checklist

## Issues

OPIK-8176

## Testing

## Documentation


CometActions:
## 🔍 Generated-test review

# PR #8084 — [OPIK-8176] [QA] Proposed e2e specs for the Datasets list enrichment (from #8076 exploration)

CometActions · draft: yes · 10 files, 2 new specs · head now `9df1c72516`

**VERDICT: needs changes — blocked on the merge conflicts, which are not mine to resolve.**

The two specs themselves are good: honestly tagged, correctly placed, fixture-based
teardown, no fixed sleeps, and they assert values rather than presence. Three real
defects were found and **fixed and pushed** (see *Fixed in this round*). What blocks the
PR is not the test code — it is that **the branch carries a stale copy of production
Java that would revert two commits already on `main`**. Read the conflicts section
first; it changes what "ready" means here.

---

## Blockers (1)

### 1. The PR would revert `#8075` and `#8128` — `apps/opik-backend/.../DatasetService.java`

This is the finding that matters, and git only half-reports it.

The branch's merge-base is `8d7afc181c`, which predates both
`72b7fc5bbc [OPIK-8176] … run the four dataset enrichment queries concurrently (#8076)`
and `9f7ba31829 [OPIK-8175] … skip the dataset item-count scan when a version supplies
items_total (#8075)`. #8076 is the exploration this PR was written from; it has since
squash-merged, so the same change exists on both sides under different commit identity.
Meanwhile `main` moved on.

`git diff origin/main HEAD -- DatasetService.java` shows the PR side is **older**, not
newer. Merging it as-is would delete `idsNeedingItemCount`, `versionItemsTotal`,
`fetchDatasetItemSummaries` and the `ITEMS_TOTAL_NOT_MIGRATED` handling — i.e. undo
#8075 — and revert the test-wording fix from
`a4aa6a12f6 [NA] [BE] test: assert the dataset id rather than JDK duplicate-key wording (#8128)`.

I checked whether the branch holds anything the production side would lose: it does not.
Its one genuinely local production commit, `6ef9b6e0b1 fix(datasets): guard the
enrichment zip against an empty version lookup`, landed on `main` too — `main`'s
`latestVersions` mono already carries `.defaultIfEmpty(Map.of())`.

**Recommended action (for a person, not for this bot):** drop both Java files from the
PR entirely. There is nothing left in them that `main` does not already have, and this
is a QA test PR — production code belongs in its own PR regardless. That also removes 2
of the 6 conflicts.

I did **not** do this: it is a conflict resolution, and it is outside what I was
authorised to change.

---

## Merge conflicts (6) — reported, deliberately unresolved

Merge state against `main` is `UNKNOWN` with 6 conflicting files. **None of these is a
defect in the PR's own work** — they are the cost of a branch that has sat while four
other QA PRs merged. Nothing below is resolved, merged, rebased or force-pushed.

| # | File | PR side | `main` side | A correct resolution keeps |
|---|------|---------|-------------|---------------------------|
| 1 | `DatasetService.java` | stale pre-#8075 snapshot of #8076 | #8076 **+** #8075 | **`main` wholesale.** Drop the file from the PR — see the blocker. |
| 2 | `DatasetServiceEnrichmentTest.java` (add/add) | 441 lines | 740 lines, grown via #8076 → #8128 → #8075 | **`main` wholesale.** Same reason; the PR's copy is a strict subset. |
| 3 | `coverage/taxonomy.yaml` | adds 2 specs + note under `datasets` | moved 4× since (#8161, #8180, #8181, #8117) | **UNION.** Both sides' areas, both sides' `specs:` lines. Taking either wholesale silently unlists a merged spec. |
| 4 | `core/backend/client.ts` | adds `listEnrichedDatasets` / `getEnrichedDataset` / `toEnrichedDataset` | moved 4× (#8161, #8181, #8097, #8093) | **UNION.** Both sides' helpers. |
| 5 | `pom/datasets.page.ts` | adds `DATASET_COLUMN`, `EMPTY_CELL`, `cell()`, `showColumn()` | moved via #8181 | **UNION.** Both sides' helpers. |
| 6 | `fixtures/index.ts` | chain head → `enriched-datasets.fixture` | chain head → `alert.fixture` | **See the trap below — not a plain union.** |

### The one that will be resolved wrongly if nobody says so

Conflict 6 is a *chain* conflict wearing the clothes of a one-line conflict, and the
damage lands in a file git reports as clean.

The estate's fixtures form a single inheritance chain that `fixtures/index.ts` exports
the head of. This PR made `enriched-datasets.fixture` the head, extending
`dashboard-cleanup.fixture` (the head when the branch was cut). On `main` the head is now
`alert.fixture` → `id-aged-traces.fixture` → … → `dashboard-cleanup.fixture`.

Resolving `index.ts` in the PR's favour leaves `test` exported from a fixture that
extends `dashboard-cleanup` — **severing every fixture main added between
`dashboard-cleanup` and `alert`**. Specs importing `alert` from `@e2e/fixtures` break.
And `enriched-datasets.fixture.ts` is a *new* file, so it merges clean and git will
never flag it.

Correct resolution: change `enriched-datasets.fixture.ts:1` to
`import { test as baseTest } from './alert.fixture'` (and its trailing
`export { expect }` likewise), then let `index.ts` export the head from
`enriched-datasets.fixture`.

### Does the conflict moot the other findings?

Partly, and it is worth being precise. The Java blocker is independent — it must be
fixed by dropping those files whatever else happens. The four e2e conflicts (3–6) are
mechanical unions that do not touch the correctness of the specs, so the fixes I applied
below remain valid and will carry through a correct resolution. Nothing here argues for
withdrawing the PR; it argues for a human resolving six files and dropping two.

---

## Fixed in this round (3) — committed as `9df1c72516`, pushed

1. **`listEnrichedDatasets` returned test suites as datasets** —
   `core/backend/client.ts`. Datasets and test suites are one entity behind one endpoint,
   discriminated by `type`; `DatasetListPage` sends `type = dataset` on every read and
   the helper sent no filter. Both specs assert `total` *exactly* (`toBe(2)` / `toBe(1)`)
   and the helper's docstring claimed it was "the Datasets list exactly as the page asks
   for it" — which was false. Now sends the same filter, as
   `{ field: 'type', operator: '=', value: 'dataset' }`; verified against
   `DatasetField.TYPE` and `DatasetFilter`'s `@JsonCreator`, which accepts only
   field/operator/key/value — hence no `type` key on the filter object.
2. **A duplicate row could pass the version-count assertion** —
   `dataset-list-version-count.spec.ts`. `new Set(itemIds).size === 5` also holds for six
   rows carrying a duplicate id, which is precisely the failure the neighbouring comment
   ("updated in place, not duplicated") claims to exclude. Now asserts row count and
   distinct-id count separately, both against `TOTAL_AFTER_COMMIT`.
3. **The taxonomy note claimed rendered coverage the page cannot give** —
   `taxonomy.yaml` + `dataset-list-enrichment.spec.ts`. The note said the list "renders"
   item/experiment/optimization counts. `DatasetListPage`'s column set is name, id,
   description, `dataset_items_count`, tags, `most_recent_experiment_at`,
   `most_recent_optimization_at`, `last_updated_at`, `created_at`, `created_by` — there
   is **no** experiment-count or optimization-count column. The spec's assertions were
   already honest (it only asserts the three columns that exist); the note and a step
   title were not. Both corrected to mark those two counters API-only.

This is the tag-honesty class the skill weights heaviest, so it is worth being explicit
about what it was and was not: the spec did not assert something it failed to tag, and
it did not tag something it failed to assert. The *note* overstated the surface. Fixed
at the note, not by weakening the test.

---

## Notes (4)

1. **`tsc` fails for a reason that has nothing to do with this PR, and it means the gate
   has been checking nothing.** `tsconfig.json(13,5): error TS5102: Option 'baseUrl' has
   been removed` — the pinned toolchain is TypeScript **7.0.2**, which removed `baseUrl`.
   The PR does not touch `tsconfig.json`; this fails identically on `main`. It is not a
   slow degradation either: `tsc` aborts on the config error and type-checks **zero
   files**, so every recent PR's "tsc" gate has been vacuous. Fix is one line
   (drop `baseUrl`; `paths` already works under `moduleResolution: Bundler`) but it is
   estate-wide and belongs in its own PR — I did not make it here. Happy to open it.
2. **Pre-existing duplicate `deleteDashboard`** in `core/backend/client.ts` (two
   definitions in the same object literal). Present on `main` at the same two positions;
   the second silently wins. Surfaced only because I bypassed the broken config to type
   check. Not this PR's, worth its own fix.
3. **Both specs share `@cap:datasets.list-datasets`.** I checked this against the
   "one capability doing duty for two tests" trap and concluded it is honest: both
   genuinely assert the datasets *list*, and the capability was already `covered: true`,
   so no denominator moves. The version-count spec additionally asserts
   `itemsAdded`/`itemsModified` — version-history territory — but that is already covered
   by `dataset-version-counters.spec.ts`, so it is not invisible coverage.
4. **No taxonomy capability was invented**, and none was needed — `list-datasets`
   already existed. Placement is correct: `spec_dir: datasets`, both specs are in
   `tests/datasets/`, and both were already added to the area's `specs:` list.

---

## Bot comments

Six inline findings, all from `baz-reviewer[bot]`. Every one was read against the code
and its consumer before a verdict.

| 1 | Taxonomy note falsely claims rendered coverage of experiment/optimization counts | **valid** | **Fixed in this PR** (`9df1c72516`). Confirmed against `DatasetListPage` — no such columns exist. |
| 2 | `new Set(itemIds).size` lets a duplicate row pass | **valid** | **Fixed in this PR** (`9df1c72516`). |
| 3 | Setup failures before `use()` orphan resources | **valid but house-wide** | **Deferred → estate-wide fixture-teardown ticket.** The identical create-then-`use()` shape with no setup `try/finally` is in ~20 sibling fixtures (`experiment`, `dataset`, `bystander`, `comparison-experiment`, …). Fixing it only here makes this file the inconsistent one. Mitigated already: `global-teardown` sweeps datasets/experiments/optimizations by the `cuj-<runId>-` prefix, so orphans are collected at run end. |
| 4 | `globalTeardown` sweep deletes the artifacts `shouldLeaveArtifacts` retains | **valid, out of scope** | **Deferred → same estate-wide ticket.** Verified: `global-teardown.ts:75` calls `sweepWorkspace` unguarded while the admin branch at :91 guards on `!env.leaveFailures`. Real inconsistency — but in a file this PR does not touch, affecting every fixture in the estate. |
| 5 | `console.warn(..., err)` leaks unbounded error bodies | **valid but house-wide** | **Deferred → same estate-wide ticket.** The exact `console.warn(\`[x fixture] delete warning for ${y}:\`, err)` line appears in ~20 sibling fixtures. Change it everywhere or nowhere. |
| 6 | `listEnrichedDatasets` returns test suites, inflating `total` | **valid** | **Fixed in this PR** (`9df1c72516`). The most consequential of the six: it is in shared `core/backend/client.ts`, and the specs assert `total` exactly. |

Findings 3, 4 and 5 are one theme — fixture setup/teardown and cleanup logging across
the whole estate — so they want **one** ticket, not three.

**Two follow-through actions I have not taken, and will not without a go-ahead:**
(a) file that single estate-wide ticket, and (b) reply on threads 3, 4 and 5 so the
deferrals are visibly answered rather than silently ignored. Say the word and I will
draft both.

---

## What I could not verify

- **Neither spec was executed.** No Opik stack is available in this workspace, so
  `playwright --list` (collection) is the strongest evidence I have. Everything about
  runtime behaviour — that the counters settle inside the 60s polls, that the Columns
  picker interaction works, that `showColumn('Most recent optimization')` finds that
  label — is reasoned from the source, not observed.
- **My fix to `listEnrichedDatasets` is unrun.** I verified the filter's shape against
  `DatasetField.TYPE`, `DatasetFilter`'s `@JsonCreator`, and `DatasetListPage`'s
  `typeFilter`, and it type-checks. I did not send it to a live backend, so I have not
  confirmed the endpoint returns 200 for it.
- **`tsc` gives no signal at all on this PR** (note 1). To get real coverage I
  type-checked against a throwaway copy of `tsconfig.json` with `baseUrl` removed
  (not committed; deleted after). Result: **zero errors in any file this PR touches**,
  before and after my edits. The only non-`@types/node` errors in the whole estate are
  the two pre-existing duplicate-`deleteDashboard` lines.
- **I could not enumerate other open generated PRs.** I was asked not to re-fetch, and
  the workspace carries no PR list, so the "shared file also touched by another open PR"
  check is complete only for files whose `main` history I could read. Everything I found
  that way is in the conflicts table; there may be an unmerged sibling PR I cannot see.
- **The Java revert risk is verified, its resolution is not.** I confirmed the PR's
  `DatasetService.java` is strictly older than `main`'s and that nothing unique would be
  lost. I did not attempt the merge.

---

## Estate gates: tag_lint **PASS** · tsc **FAIL (pre-existing, non-informative)** · playwright --list **PASS**

Re-run after my edits, at `9df1c72516`:

- `tag_lint.py --taxonomy tests_end_to_end/coverage/taxonomy.yaml --estate tests_end_to_end`
  → `56 specs checked, 1 exempt, 0 problem(s)` — **PASS**
- `npx tsc --noEmit` → **FAIL**, `TS5102 Option 'baseUrl' has been removed`. Pre-existing
  on `main`, unrelated to this PR, and checks no files. Under a corrected config the PR's
  own files are clean (note 1).
- `npx playwright test --list` on both changed specs → `Total: 2 tests in 2 files` — **PASS**

---

**The PR is not ready, and I have not marked it ready — that flag is a person's to flip,
and I would not flip it here anyway.** A human needs to resolve six conflicts and drop
two Java files first. Once that is done and the fixture chain is re-pointed, the test
work in this PR stands on its own.

---
_[review_generated_tests.yml](https://github.com/comet-ml/comet-automation-tests/actions/runs/34179844385) ·


[OPIK-8176]: https://comet-ml.atlassian.net/browse/OPIK-8176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OPIK-8176]: https://comet-ml.atlassian.net/browse/OPIK-8176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

CometActions:
## 🔍 Generated-test review

# PR #8084 — [OPIK-8176] [QA] Proposed e2e specs for the Datasets list enrichment (from #8076 exploration)

CometActions · draft: yes · 8 files (after restack), 2 new specs
Pushed as `03c85cb320` (force-with-lease; branch was restacked)

**VERDICT: ready with notes** — do not flip the draft flag yet; see the
cross-PR collision under Notes, which is a merge-order hazard a human should
decide on.

The PR arrived `CONFLICTING` against `main` with 6 conflicting files. All six
are resolved: 2 were inherited backend commits that the restack dropped, and 4
were real conflicts inside `tests_end_to_end/` that I resolved. The specs
themselves are good — genuinely honest tags, real assertions, no flake
patterns. The blocker was the merge state, and it is cleared.

---

## Blockers (0)

None outstanding. The one blocking condition — the branch could not merge — is
resolved below.

## Conflicts (4 resolved, 2 dropped)

The branch carried 4 inherited commits that were not its own, including
`DatasetService.java` and `DatasetServiceEnrichmentTest.java`. The restack
dropped all four, so the two backend conflicts disappeared with them: **no
production code was resolved, and none remains in the diff.** The diff is now
`tests_end_to_end/` only, 8 files.

| File | What the two sides did | What the resolution keeps |
|---|---|---|
| `e2e/fixtures/index.ts` | Both claimed line 1 of the fixture chain: base had made `alert.fixture` the head; this PR had made `enriched-datasets.fixture` the head. Both also appended a type block. | Re-parented this PR's fixture onto `alert.fixture` (edited its own `import baseTest` and `export { expect }`), then made it the new line-1 head. Both type blocks kept — base's `id-aged-traces` + `alert` exports and this PR's `enriched-datasets` exports. Taking either side alone would have orphaned one fixture and broken every spec resolving through it. |
| `e2e/core/backend/client.ts` | Base added `writeDatasetItemsBatch`; this PR added `listEnrichedDatasets` + `getEnrichedDataset`. Append-only in practice. | Both, in full. The conflict opened *mid-docblock* — the two sides shared one `/**` opener and one `},` closer — so a bare concatenation would have left an orphaned `*` run and a cascade of TS1109/TS1005. Bridged with the enclosing member's `},` plus a fresh `/**`. |
| `e2e/pom/datasets.page.ts` | Base added `datasetItemCount(name)`; this PR added `cell(datasetId, columnId)` and `showColumn(label)`. | Both, same mid-docblock bridge as above. Indentation matched to the surrounding 2-space class body. |
| `coverage/taxonomy.yaml` | `specs:` — base added 5 spec lines, this PR added 2. `note:` on `list-datasets` — both sides rewrote it wholesale. | `specs:` unioned to all 7. The `note:` had its **meaning** unioned, not a side picked: base's note describes the Item-count-vs-items-endpoint coverage on a mixed versioned/version-less page; this PR's describes per-row enrichment and the honest caveat that only Item count and the two recency columns are rendered. Each describes coverage the other cannot see. No capability key was invented. |

A second conflict round appeared on the branch's own prior `[QA] review:`
commit (`9df1c72`), on the same `list-datasets` note. Resolved the same way —
that commit's note carried the more precise scoping caveat, which is preserved.

## Should fix (0)

The one valid, still-open bot finding local to this PR is **fixed**, not listed
(finding #10 below): the recency timestamps were asserted only present/absent,
which a malformed or stale instant satisfies as readily as a correct one — and
because the list-vs-by-id comparison shares the enrichment code path, both
reads agree on a bad value just as happily. The fixture now records `seededAt`
and the spec bounds both timestamps to that window, with ±6h skew allowed since
the backend mints them and its clock is not the runner's.

## Notes (4)

1. **Cross-PR collision on `datasets.page.ts` — the case git will not warn you
   about, and the reason to hold the draft flag.** Open PR **#8127**
   ("[OPIK-7791] release 2.2.48 → 2.2.49 exploration specs") adds
   `export type DatasetColumnId` to the *same file*, as a 6-member string union
   (`'name' | 'description' | 'dataset_items_count' | …`). This PR adds the same
   exported name, derived from a new `DATASET_COLUMN` const of 3 ids. Both are
   declared immediately before `export class DatasetsPage`, so git will most
   likely flag it — but if either moves, git merges both cleanly and you get a
   silent `TS2300 Duplicate identifier`. #8127 also adds `datasetCell()` /
   `datasetCellText()` / `setColumnEnabled()`, which are near-duplicates of this
   PR's `cell()` / `showColumn()`.
   **Whichever merges second should converge, not concatenate**: keep this PR's
   `DATASET_COLUMN` const (the specs use its runtime values), widen it to
   cover #8127's six ids, derive the type from it, and collapse the two
   cell-addressing methods into one. This is a decision for a human, which is
   why the verdict is "ready with notes" rather than "ready".
   Nine other open generated QA PRs touch the shared files this PR touches —
   `taxonomy.yaml` (all nine), `client.ts` (#8172, #8167, #8184, #8160, #8127,
   #8151, #8080), `fixtures/index.ts` (#8167, #8184, #8160, #8127, #8151,
   #8192). The `fixtures/index.ts` line-1 chain head is a single line every
   new-fixture PR claims, so each will conflict in turn — noisy but safe, since
   git catches it. The `datasets.page.ts` symbol collision is the one that can
   merge clean and still be wrong.

2. **The `tsc` gate fails on `main`, not on this PR.** `tsconfig.json` sets
   `baseUrl`, which TypeScript 7 removed; the repo floats `typescript: ^7.0.2`
   and resolved 7.0.2, so `tsc` aborts before compiling anything. The file was
   last touched in #6677 and this PR does not touch it. The one-line fix is to
   drop `baseUrl` — `paths` already resolves relative to the config, so
   behaviour is identical. **Estate-wide, belongs in its own PR**; I did not
   change a shared config from inside a test PR. Happy to open it.

3. **Pre-existing duplicate in `client.ts`.** `deleteDashboard` is declared
   twice on `origin/main` (lines 674 and 1243; 758 and 1369 after this PR's
   additions shift them). `TS2300`, already on main, not introduced here and not
   in this PR's diff. Same estate-wide PR as note 2.

4. **A contradiction worth someone's attention.** `global-teardown.ts` says
   traces "are expected to cascade with their project's own deletion" and so
   does not sweep them; this PR's fixture docstring says `ProjectService.delete`
   removes only the project row, so traces do *not* cascade — which is why the
   fixture deletes them explicitly. One of the two is wrong. If the fixture is
   right, every trace orphaned by a setup failure leaks permanently, estate-wide.
   Not this PR's to settle.

## Bot comments

10 comments, all from `baz-reviewer[bot]`. Three were already fixed by the
branch's prior review round (`9df1c72`) and the bot itself confirmed each —
comments 7, 8 and 9 are those confirmations, not new findings.

| 1 | Taxonomy note claims rendered coverage for experiment/optimization counts, but `DatasetListPage` has no such columns | valid | already fixed in `9df1c72`; note now scopes those two to API-only. Bot confirmed (#8). Caveat preserved through my conflict resolution |
| 2 | `new Set(itemIds).size` alone lets a duplicate row pass the version-count test | valid | already fixed in `9df1c72`; row count and distinct-id count now both asserted against `TOTAL_AFTER_COMMIT`. Bot confirmed (#9) |
| 3 | Setup failures before `use()` orphan datasets/traces/experiments | valid but house-wide | **deferred → estate-wide fixture PR.** 22 of 25 fixtures create-then-`use()`-then-clean with no setup rollback; only 3 special cases do otherwise. The sweep in `global-teardown.ts` *does* reclaim `cuj-`-prefixed datasets, experiments, optimizations and projects, so the residual leak is traces only — which is note 4 above, not this fixture's bug |
| 4 | `globalTeardown` sweeps unconditionally, deleting what `shouldLeaveArtifacts` retained for debugging | valid | **deferred → own PR.** Confirmed real: `global-teardown.ts:75` is unguarded while the admin branch at :87 correctly checks `!env.leaveFailures`, so `OPIK_LEAVE_FAILURES` does not actually preserve normal-workspace artifacts. But `global-teardown.ts` is not in this PR's diff — fixing it here would be an unrelated behaviour change |
| 5 | `console.warn(err)` in cleanup can log unbounded backend bodies | valid but house-wide | **deferred → same estate-wide fixture PR as #3.** 25 fixtures use this exact pattern; changing it only here makes this file the inconsistent one |
| 6 | `listEnrichedDatasets` omitted the page's `type = dataset` filter, so test suites inflate `total` | valid | already fixed in `9df1c72` via `DATASETS_ONLY_FILTER`. Bot confirmed (#7) |
| 7 | *(bot confirmation of #6)* | n/a | — |
| 8 | *(bot confirmation of #1)* | n/a | — |
| 9 | *(bot confirmation of #2)* | n/a | — |
| 10 | Recency cells only checked non-empty, so malformed/stale values pass and list-vs-by-id agree on the same bad value | valid | **fixed in this PR** — `seededAt` anchor on the fixture, timestamps bounded to the seed window |

Two follow-through actions I have **not** taken, and would like the go-ahead on:

1. **One ticket** covering the estate-wide items — the `tsconfig` `baseUrl`
   break (note 2), the duplicate `deleteDashboard` (note 3), the unguarded
   `globalTeardown` sweep (bot #4), and the fixture-wide setup-rollback /
   `console.warn` pattern (bot #3, #5). The first two block or dirty the `tsc`
   gate for *every* PR, so they are worth doing soon.
2. **Replies on the four bot threads** I judged deferred (#3, #4, #5), so the
   findings read as answered rather than ignored.

## Estate gates

Run after every edit; all three re-run on the final tree.

| Gate | Before | After | Note |
|---|---|---|---|
| `tag_lint` | FAIL | **PASS** | 66 specs checked, 1 exempt, 0 problems. Was failing only because the conflict markers made `taxonomy.yaml` unparseable |
| `tsc --noEmit` | FAIL | **FAIL (pre-existing, 0 new)** | Fails identically on a clean `origin/main` worktree. Baselined: 53 unique error lines on `main`, 53 on this branch, **empty diff** — this PR introduces no type errors. See note 2 |
| `playwright --list` | FAIL | **PASS** | Both new specs collect (2 tests, 2 files). Full estate also collects: 128 tests in 63 files, so the shared POM/fixture/client changes broke no other spec |

## What I could not verify

- **The specs have not been run.** `--list` proves they parse and collect; it
  proves nothing about whether they pass. Both are `@t2-cuj` and `test.slow()`,
  and need a live Opik stack with a working SDK bridge. The assertions
  themselves read as sound, but "would this fail if enrichment broke" is
  answered from the code, not from a green run.
- **The ±6h skew window in my timestamp fix is a judgement call.** Tight enough
  to catch epoch-zero, far-future and unparseable values; deliberately loose
  about clock skew between the runner and the backend. I could not measure the
  actual skew on the CI stack, so I chose generous over brittle.
- **The `mergeStateStatus` at GitHub's end.** The rebase is clean locally
  against the `origin/main` this run fetched (`87e5dc147e`) and the push
  succeeded, but `main` moves fast — nine sibling QA PRs are queued on the same
  shared files — so the PR can go dirty again before anyone merges it.
- **Tag honesty on `@cap:datasets.list-datasets`** I checked by reading both
  test bodies against the capability, and both genuinely assert list-rendered
  values rather than mere row presence. The one judgement I'd flag: the
  version-count spec also asserts `itemsAdded` / `itemsModified` / `isLatest`,
  which is closer to `version-history-view` territory. That capability is
  already `covered: true`, so no denominator shifts either way, and the spec's
  primary claim really is about the list page. I left the tag as-is.

---
_[review_generated_tests.yml](https://github.com/comet-ml/comet-automation-tests/actions/runs/34187893711) ·


[OPIK-8176]: https://comet-ml.atlassian.net/browse/OPIK-8176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

CometActions:
## 🔍 Generated-test review

# PR #8084 — [OPIK-8176] [QA] Proposed e2e specs for the Datasets list enrichment (from #8076 exploration)

CometActions · draft: yes · 8 files, 2 new specs · head now `a5123dde0b`

**VERDICT: ready with notes** — the merge conflict is resolved and pushed, tag
honesty checks out, and no finding remains that should hold this PR. It is still
a draft, and it must stay one until a person flips it. I have not marked it ready
and have not merged it.

---

## Merge conflict (resolved)

`conflicts.txt` listed one file, and it was inside `tests_end_to_end/`, so it was
mine to resolve. Nothing conflicted outside the test tree — no inherited
production commit was involved (`restack-report.json`: `inherited_commits: 0`).

**What the two sides did.** Main's side is commit `66589738ee` (#8199, "keep
taxonomy specs lists sorted so concurrent additions stop colliding"), which
sorted every area's `specs:` list *and* landed three more dataset specs from a
sibling QA PR (`dataset-version-concurrent-writes`, `dataset-version-counters`,
`dataset-version-repeated-item-id`). This PR's side appended its own two specs
unsorted at the end of the list. Both sides claimed the same trailing lines.

**What the correct resolution keeps.** Both — the `specs:` list is a plain union.
Taking either side alone drops three specs or two. I kept all eleven and applied
main's newly-established sorted order, which puts this PR's two specs *above*
`dataset-parallel-item-read.spec.ts` rather than where the conflict happened to
open. Resolving in place would have satisfied git while silently defeating the
very invariant #8199 had just introduced.

**The `note:` needed no union.** Main's `list-datasets` note is byte-identical to
the merge-base — only this PR rewrote it — and the rewrite already carries main's
meaning ("row is listed (t1); t2 also asserts the Item count column…") plus the
enrichment coverage. So no meaning was lost and there was nothing to reconcile.
No capability key was invented.

Pushed as `a5123dde0b` with a plain push (RESTACK was `false`, so no force-push
was needed or used).

**Does the conflict moot the other findings?** No — it is resolved, so the rest
of this report stands on its own.

---

## Blockers (0)

None.

## Should fix (0)

None outstanding. The two rounds of prior review commits on this branch already
landed the substantive fixes (scoping the list read to datasets, bounding the
recency assertions, tightening the item-row assertion, correcting the coverage
note), and I verified each against the current code rather than taking the commit
messages for it.

## Notes (3)

1. **The `tsc` gate is red for the whole repo, and not because of this PR.**
   `tests_end_to_end/e2e/tsconfig.json` sets `baseUrl`, which TypeScript 7.0.2
   (the installed compiler) has removed. The same `baseUrl` line is on `main`,
   and this PR does not touch `tsconfig.json`. Because tsc aborts on the config
   error it type-checks *nothing*, so the gate as configured cannot say anything
   about this PR either way. I probed with a throwaway config (dropping
   `baseUrl`, restoring `@types/node` resolution; not committed) and the entire
   e2e tree produced exactly **two** errors, both `Duplicate identifier
   'deleteDashboard'` in `core/backend/client.ts` — and both already present on
   `main` (lines 674 / 1243 there). **Zero type errors are attributable to this
   PR.** Fixing `tsconfig.json` is an estate-wide change affecting every spec and
   belongs in its own PR; I can open it if you want.

2. **Two overlapping ways to reach the Item count cell in `datasets.page.ts`.**
   The pre-existing `datasetItemCount(name)` addresses by dataset *name* and is
   hard-wired to that one column; the new `cell(datasetId, columnId)` addresses
   by dataset *id* and generalises to any column. The new one is strictly better
   (id-addressing is unambiguous, and it is documented as such), but the estate
   now has both. Converging them is not the trivial rewrite it looks like — the
   keys differ — and `datasetItemCount` has live callers in
   `dataset-item-count.spec.ts`, so I did not touch it. Left deliberately.

3. **Missing taxonomy capability: none needed.** Worth stating explicitly since
   it is the usual failure mode here. Both specs tag only
   `@cap:datasets.list-datasets`, which exists, and both genuinely assert it. The
   version-counter assertions in `dataset-list-version-count.spec.ts` are
   corroborating detail for a list-page claim, and `version-history-view` is
   already covered by `dataset-version-counters.spec.ts` — so nothing is being
   silently earned and nothing is invisible coverage.

---

## Shared-file contention (the reverse case git will not flag)

This PR touches four shared files. Checking each against `origin/main`:

| File | Moved on main since this branch's base? |
|---|---|
| `coverage/taxonomy.yaml` | **Yes** — `66589738ee` (#8199) + #8161, #8180 |
| `e2e/core/backend/client.ts` | No (since merge-base); hot historically — #8161, #8181, #8097 |
| `e2e/pom/datasets.page.ts` | No (since merge-base); last touched by #8181 |
| `e2e/fixtures/index.ts` | No (since merge-base); last touched by #8117, #8097 |

`taxonomy.yaml` is the one that actually moved under this PR, and that is exactly
the collision #8199 was opened to stop — now handled by the sorted union above.

The other three did not move since the merge-base, so this merge lost no work.
But every one of them was last touched by *another generated QA PR*, so they are
demonstrably contended. **I could not enumerate currently-open sibling PRs from
this workspace** (no network/`gh` access, and the brief said not to re-fetch), so
I cannot rule out an open PR adding a near-duplicate `cell()` helper to
`datasets.page.ts` or another `listEnriched*` method to `client.ts`. Those would
merge cleanly and duplicate work silently. Worth a human glance at the open QA PR
list before merge.

**Fixture chain verified intact.** `fixtures/index.ts` line 1 →
`enriched-datasets.fixture` → `alert.fixture` → `id-aged-traces.fixture`. Main's
line-1 head was `alert.fixture`, and this PR correctly re-parents onto it and
becomes the new head. Nothing orphaned, nothing dropped from the chain. This file
merged cleanly, so no resolution was needed — but I checked, because a clean
git merge is not evidence the chain resolves.

---

## Bot comments

11 comments from `baz-reviewer[bot]`, of which 4 are baz's own follow-ups
confirming earlier findings were addressed. Dispositions for the 7 substantive
findings:

| 1 | Taxonomy claims rendered coverage for experiment/optimization counts that `DatasetListPage` has no columns for | **valid** | Already fixed on this branch (`9df1c72`). Current note scopes UI coverage to Item count + the two recency columns and names the reason. Verified against the live note, not the commit message. |
| 2 | `new Set(itemIds).size` alone lets a duplicate row pass the version-count test | **valid** | Already fixed (`9df1c72`). Now asserts row count *and* distinct-id count, both against `TOTAL_AFTER_COMMIT`, plus the rewritten id — the three together do close the hole. |
| 6 | `listEnrichedDatasets` omits the page's `type = dataset` filter, so test suites inflate `total` | **valid** | Already fixed (`9df1c72`) via `DATASETS_ONLY_FILTER`. Materially correct: the spec asserts an exact `total`, which would have been flaky in any project holding a test suite. |
| 10 | Recency assertions only check non-empty, so malformed/stale values pass | **valid, partly addressed** | API side fixed (`03c85cb`): both timestamps are parsed and bounded to the seed window. The residual UI half I am **declining, with reason** — the page renders relative time ("3 mins ago"), so an exact-instant UI assertion would test the humaniser, not the enrichment. The UI's honest claim is the populated-vs-dash distinction, which is what it asserts. |
| 3 | Setup failures before `await use()` orphan datasets/traces/experiments | **valid but house-wide** | Deferred → driver/fixture-wide ticket. `json-output-experiment`, `optimization-run` and `aged-experiment` have the identical create-then-`use()`-then-clean shape; fixing it only here makes this file the odd one out. Consequence is also bounded: `global-teardown`'s `cuj-` prefix sweep does collect datasets, experiments, optimizations and projects. Traces are the real gap — the sweep has no trace pass — but that gap is estate-wide, not this PR's. |
| 4 | `globalTeardown` sweeps unconditionally, deleting what `shouldLeaveArtifacts` retained for failed runs | **valid** | Confirmed by reading `global-teardown.ts`: line 75's `sweepWorkspace(env.apiKey, null, prefix)` has no guard, while the admin-workspace branch (87) and the scratch-dir removal (97) both check `!env.leaveFailures`. A real inconsistency that undercuts debugging for **all 33 fixtures**. Deferred → its own PR: the file is not in this diff and the blast radius is the whole estate. |
| 5 | `console.warn(..., err)` in cleanup can log unbounded backend bodies | **valid but house-wide** | Deferred → same estate-wide ticket as #3. This exact line, phrasing included, appears in ~20 of the 33 fixtures. Changing it here alone is churn that makes this file inconsistent without reducing any exposure. |

Two follow-through actions I have **not** taken, and want your go-ahead for:

1. **File the deferred findings as two tickets**, not five: (a) fixture-wide
   partial-setup rollback + bounded cleanup logging (#3, #5), and (b) the
   `global-teardown` `leaveFailures` guard (#4), which is independent and the
   more consequential of the two.
2. **Reply on the four threads** I judged deferred or declined (#3, #4, #5, and
   the residual UI half of #10), so they read as answered rather than ignored.
   I have not posted anything.

---

## Estate gates

Re-run after my edit, on the merged result:

**tag_lint PASS** · **tsc FAIL (pre-existing, repo-wide — not this PR)** · **playwright --list PASS**

- `tag_lint.py --taxonomy … --estate tests_end_to_end` → 66 specs checked, 1 exempt, **0 problems**.
- `npx tsc --noEmit` → fails on `tsconfig.json(13,5) TS5102: baseUrl has been removed`. Identical on `main`; see Note 1. Nothing in this PR's code contributes an error.
- `npx playwright test --list` on both new specs → **2 tests in 2 files**, both collect.

## What I could not verify

- **I did not run the specs.** There is no live Opik stack in this workspace, so
  every reliability judgement here is static: I confirmed no `waitForTimeout`,
  bounded `expect.poll` timeouts with messages, fixture-based teardown (no
  `try`/`finally` or cleanup in the test body), and `cuj-`-prefixed names via
  `testNamespace` (`${cujPrefix}-w{worker}-{slug}`) that the global sweep will
  collect. Whether the two 60s polls are long enough for ClickHouse-backed
  counters to settle on real infrastructure is not something I can establish
  from here.
- **The selectors are unproven against a running frontend.** `showColumn()`
  depends on a fairly specific DOM reading — that `SortableMenuItem` spreads
  dnd-kit attributes making entries report as `button` rather than
  `menuitemcheckbox`. The docstring argues it convincingly and it is the kind of
  detail a generator does not invent, but only a real run confirms it.
- **Open sibling PRs**, per the shared-file section above.
- The `deleteDashboard` duplicate identifier on `main` is a real pre-existing
  defect I noticed but did not investigate; it is outside this PR's scope.

---
_[review_generated_tests.yml](https://github.com/comet-ml/comet-automation-tests/actions/runs/34202819589) ·


[OPIK-8176]: https://comet-ml.atlassian.net/browse/OPIK-8176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

CometActions:
## 🔍 Generated-test review

# PR #8084 — [OPIK-8176] [QA] Proposed e2e specs for the Datasets list enrichment (from #8076 exploration)

CometActions · draft: yes · was 8 files / 2 new specs → now 2 files / 1 new spec
Head after this review: `8f1aee2ecf` (pushed to `comet-qa-bot/OPIK-8176/datasets-list-enrichment-e2e`)

**VERDICT: ready with notes — but a human should confirm the deletion, or close the PR outright.**

The headline is not a defect in the specs. It is that **opik#8127 merged the same
coverage while this PR sat in review**, and #8127 is strictly stronger. All four
merge conflicts were symptoms of that, not independent problems. What survives is
one narrow spec covering the single claim #8127 and friends do not make.

---

## Blockers (1) — resolved in this push

### 1. PR superseded by merged opik#8127; all four conflicts were duplicate-name collisions

`tests_end_to_end/e2e/core/backend/client.ts`, `pom/datasets.page.ts`,
`fixtures/index.ts`, `coverage/taxonomy.yaml`

Commit `3df577dbb2` ("[OPIK-7791] [QA] Proposed e2e specs from the 2.2.48 →
2.2.49 release exploration (#8127)") landed on main and added, in one go:

| This PR added | #8127 already has on main |
|---|---|
| `EnrichedDatasetRef`, `EnrichedDatasetPage` | `DatasetSummaryRef` (superset — also carries `latestVersionHash`) |
| `listEnrichedDatasets()`, `getEnrichedDataset()` | `listDatasetSummaries()`, `getDatasetSummary()` (same two endpoints) |
| `DATASET_COLUMN`, `EMPTY_CELL`, `cell()`, `showColumn()` | `DatasetColumnId`, `datasetCell()`, `datasetCellText()`, `setColumnEnabled()` |
| `enriched-datasets.fixture.ts` (2 datasets) | `summarised-datasets.fixture.ts` (4 pairwise-distinct shapes, one empty) |
| `dataset-list-enrichment.spec.ts` | `dataset-list-summary-columns.spec.ts` |

**Why this was a blocker and not a merge chore.** The three shared files are
normally append-only, and the standing guidance is to keep both sides. That
guidance's premise fails here: these are not two additions, they are two names
for one thing. Keeping both sides would have left the estate carrying two
parallel vocabularies for a single page — two client method families over the
same two endpoints, two column-id enums, two fixtures seeding near-identical
data — with nothing to tell a future author which to reach for.

**Resolution applied.** Main's side taken on all three shared files;
`core/backend/index.ts` had its now-unexported `Enriched*` types dropped.
Nothing outside `tests_end_to_end/` was touched — I verified the conflict set
contained no backend or frontend path before resolving anything.

**`taxonomy.yaml`** was the one that needed meaning unioned rather than a side
picked. The `specs:` list is a plain union (kept sorted, per `#8199`). The
`note:` — which both sides rewrote — keeps **main's** text verbatim, because
main's describes real coverage this PR cannot see (four datasets, one empty,
pagination), and gains only the one clause this PR still earns. No capability
key was invented.

### What I deleted, and why it is recoverable

`dataset-list-enrichment.spec.ts` (193 lines) and `enriched-datasets.fixture.ts`
(192 lines) are **deleted as fully superseded**. Claim by claim, everything the
spec asserted is asserted on main, more sharply:

| This PR's spec asserted | Covered on main by |
|---|---|
| four counters per row, own values | `dataset-list-summary-columns.spec.ts`, over 4 pairwise-distinct shapes vs this PR's 2 |
| recency timestamps present/absent | same spec, tied to whether the row has the thing the column names |
| list vs get-by-id agreement | same spec (list vs detail vs the dataset's own items/versions endpoints) |
| Item count + 2 recency columns on the page | same spec, "on exactly the right rows" |
| — | plus pagination re-read, which this PR had no equivalent of |

**Recoverable at `a5123dde0b`** if a reviewer disagrees. This is the call in
this review I am least willing to make silently, so: it is a content deletion,
and it is yours to veto.

**One thing genuinely lost.** This PR's spec bounded both recency timestamps
against a seed window (`Date.parse` + ±6h skew), where main's asserts only
`not.toBeNull()`. That is a real improvement and the better assertion. I did
**not** port it, because it belongs in main's spec, which this PR does not own —
editing it here would manufacture a fresh conflict. Recommended as a small
follow-up against `dataset-list-summary-columns.spec.ts`.

---

## Should fix (0 outstanding)

The one item — the surviving spec claiming novelty it no longer has — is fixed
in this push. `dataset-list-version-count.spec.ts` is re-pointed at main's
`listDatasetSummaries` and `datasetCell`, and its docstring now states its scope
honestly: it names the four neighbouring specs that cover the adjacent ground
and confines itself to the gap.

**That gap, stated precisely.** Main's `dataset-item-count.spec.ts` already
drives the list's v1→v2 move — but via a **delete**. `dataset-version-counters`
and `dataset-version-repeated-item-id` already assert `itemsAdded`/
`itemsModified`/distinct-id totals — but on the **Version history tab**. Nothing
moves the *list's* number with an insert that is simultaneously a modification
and an addition. That shape is worth one test because the three plausible wrong
answers are all distinguishable: a stale-count fallback reads 3, counting only
the last batch reads 2, counting batch entries rather than distinct ids reads 6,
and only 5 means the list read `items_total` off the version it calls latest.

**This is a thin margin, and I want to be plain about it:** a reviewer who
judges it too thin to keep should close #8084 entirely rather than merge it. The
coverage this PR was opened for now exists. I kept the spec because the claim is
real and unasserted, not because the PR needs to land something.

---

## Notes (2)

1. **`tsc` was failing for a reason that had nothing to do with this PR.** The
   gate reported `tsconfig.json(13,5): error TS5102: Option 'baseUrl' has been
   removed`. Main had already removed `baseUrl` in `06059274ea` (#8279, "make
   the e2e tsc gate actually type-check the estate") and added `"types":
   ["node"]`; this branch was 30+ commits behind and still carried the old
   config against TypeScript 7.0.2. The merge fixed it — no code change was
   needed. Worth noting because #8279 makes the gate check *more* than it used
   to, so a stale branch can hide real type errors; after merging, this PR has
   none.

2. **Sibling-draft drift — one benign textual overlap.** No open generated draft
   touches `pom/datasets.page.ts`, and none claims `datasets.list-datasets`, so
   there is no coverage-losing collision. PR **#8345** does add
   `datasets/dataset-typed-content-round-trip.spec.ts` and so edits the same
   sorted `specs:` block in `taxonomy.yaml`. Whichever merges second takes a
   one-line textual conflict that unions trivially (both add distinct lines).
   Flagging it because git reports both as mergeable today.

   For the record, all four shared files moved under this PR while it was open —
   `client.ts` 6 times, `fixtures/index.ts` 6, `taxonomy.yaml` 6,
   `pom/datasets.page.ts` once (that once being #8127, the whole story here).

---

## Bot comments

11 threads from `baz-reviewer[bot]`; 4 substantive findings, 4 of baz's own
acknowledgements, 3 outstanding. Prior review rounds had already addressed
everything substantive.

|---|---|---|---|
| 1 | Taxonomy overstates coverage: `experiment_count`/`optimization_count` are not rendered columns, so the UI phase cannot catch them | **valid** | already fixed in a prior round (`9df1c72`); now moot — the note is main's text and the spec is deleted |
| 2 | `new Set(itemIds).size` alone lets a duplicate row pass the version-count test | **valid** | already fixed in a prior round (`9df1c72`); assertion retained verbatim in the surviving spec (length **and** distinct count, both against the total) |
| 6 | `listEnrichedDatasets` omits `DatasetListPage`'s `type = dataset` filter, so test suites inflate `total` | **valid** | fixed in a prior round (`9df1c72`), then superseded — the method is deleted. **See carry-forward below**, this one outlives the PR |
| 10 | UI recency assertions only check non-empty, so malformed/stale values pass | **valid** | partly fixed in `03c85cb` (API-side seed-window bounds); now moot with the spec deleted, and folded into the follow-up recommended above |
| 3 | Fixture seeds before `await use()`, so a setup rejection orphans datasets/traces/experiments | **valid, but moot** | file deleted. Would otherwise have been estate-wide: no fixture in the estate does setup-level rollback |
| 4 | `globalTeardown` sweeps unconditionally, deleting the artifacts `shouldLeaveArtifacts` retains for failed tests | **valid, and outlives this PR** | not this PR's file — the defect is in `globalTeardown`, which #8084 never touched. **Carry-forward below** |
| 5 | Cleanup `catch` passes the raw `err` to `console.warn`, risking unbounded bodies in logs | **valid, house-wide** | file deleted; the `safe()`-then-`console.warn` shape is the estate's convention across fixtures, so fixing it in one file would have made that file the inconsistent one. Estate-wide change, own PR |
| 7, 8, 9, 11 | baz's own "commit X addressed this" acknowledgements | n/a | no action |

**Two findings outlive this PR and need somewhere to go.** I have not filed
these — say the word and I will open them:

- **#4 (`globalTeardown` destroys failure artifacts).** Real, independent of
  #8084, and it silently defeats `OPIK_LEAVE_FAILURES` debugging for the whole
  estate. Guard the sweep with `!env.leaveFailures`, matching the
  admin-workspace branch that already does.
- **#6 (`listDatasetSummaries` has no `type = dataset` filter).** baz was right
  about the endpoint, and main's surviving `listDatasetSummaries` has the same
  gap this PR had fixed. Not currently load-bearing — the fixtures seed fresh
  projects holding no test suites — but it is a latent wrong `total` for any
  future spec whose project holds one. The deleted code had the correct filter
  and the reasoning for it; worth porting onto main's method.

---

## Sibling drafts

The PR stands alone — it is not part of a stack, and `git merge-tree` reports no
add/add on any spec path against any open sibling. The only overlap is #8345's
one-line touch of the same sorted `taxonomy.yaml` block (Note 2). The real
collision on this PR was not with a sibling draft at all; it was with #8127,
**already merged**, which is the case `git` reports as mergeable and the sibling
check does not look for.

---

## Estate gates: tag_lint **PASS** · tsc **PASS** · playwright --list **PASS**

Run on the final pushed state (`8f1aee2ecf`), all three, after every edit:

- `tag_lint` — 88 specs checked, 1 exempt, **0 problems** (was 66 before the merge)
- `tsc --noEmit` — clean, exit 0 (was the reported FAIL; cause was the stale `tsconfig.json`, see Note 1)
- `playwright test --list` — 167 tests in 85 files; the changed spec collects as 1 test
- Merge state — **no conflicts** against `origin/main`, branch 0 commits behind

## What I could not verify

- **The spec was not executed.** LOCAL OPIK was `false` and I confirmed it: no
  listener on `localhost:5173`, and `local-opik.log` shows every container
  failed to start after `docker pull minio/mc` was denied
  ("pull access denied for minio/mc"). So `dataset-list-version-count.spec.ts`
  has **not** been run — against this instance or any other. Its correctness
  here rests on review and on `--list` collecting it, not on a green run. This
  is exactly the ambiguity the local instance exists to remove, and it did not
  come up.
- **Unvalidated at runtime, specifically:** that an SDK `insert()` re-sending an
  existing id cuts one version reporting `items_total: 5 / added: 2 /
  modified: 1` (read from the endpoint contract and from main's
  `dataset-version-repeated-item-id.spec.ts`, which asserts the same shape); and
  that `datasetCell(id, 'dataset_items_count')` resolves on the rendered list
  (main's own specs use that exact call, so it is exercised on main — but not by
  me).
- **The 60s poll timeouts** are inherited from the original spec and unchanged. I
  have no measurement of how long the ClickHouse-backed counters actually take to
  settle on a real instance.

## Two things I did not do, by rule

- **The PR is still a draft, and I did not change that.** Gates are green and
  the conflicts are cleared; I believe it is mergeable as it stands. A person
  flips it — and in this case the more likely correct action is to **close it**,
  which is why the choice matters.
- **No taxonomy capability was invented.** Nothing required one: `list-datasets`
  already existed and was already `covered: true`, so no denominator moved in
  either direction.

### ⚠️ Post-fix verification unavailable

Staging did not answer, so the specs were **not** re-run after the fixes. Treat the fixes as unverified.

---
_[review_generated_tests.yml](https://github.com/comet-ml/comet-automation-tests/actions/runs/35085056923) ·


[OPIK-8176]: https://comet-ml.atlassian.net/browse/OPIK-8176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

AndreiCautisanu (approved):

PR #8092, opened 2026-09-01
[OPIK-6901] [BE] feat(cutover): add --old-table/--new-table to delta_replay so it can run after the EXCHANGE

thiagohora:
## Details

`delta_replay.sh` hardcodes both table names, so it can only ever run **before** the EXCHANGE — afterwards `traces` *is* the successor and `traces_local_v2` is gone, so it would read its own destination and insert into a missing table. This adds the `--old-table`/`--new-table` pair `verify.sh` has carried from the start, with the same defaults, wording and identifier validation.

Without it the swap strands two sets of rows with nothing able to recover them: everything written *while* the final delta runs (its `SELECT` snapshots at query start, so at measured production rates ~60,000 traces over a ~50 min run land in the parked backup), and the deletions deferred past the swap (10,984 captured ids). Running the delta and the EXCHANGE back-to-back does not avoid this — the gap *is* the delta's own duration.

- Post-swap invocation is `--old-table traces_pre_cutover_backup --new-table traces`: the insert arm copies stranded rows out of the backup into the live successor, the deletion arm masks rows whose ids are in the bridge and absent from the backup.
- `000002_delta_and_deletion_replay.sql`: four hardcoded references become `${OLD_TABLE}`/`${NEW_TABLE}` — `INSERT INTO` and `DELETE FROM` take the new table, the delta `FROM` and the deletion arm's existence check take the old. `deletion_events_local` stays literal in both places; the bridge does not move.
- **Reviewer note on the risk this introduces:** the flags make it possible to point the delta at the wrong pair. *Reversing* them would copy the live table over the parked backup and destroy the rollback net, and the equal-tables guard does not catch it because the names differ. The script cannot enforce this — it cannot know which side of the EXCHANGE it is on — so the constraint is operator-side, in comet-ml/comet-support-agent#270.

## Change checklist
- [ ] User facing
- [x] Documentation update

## Issues

- Resolves #
- OPIK-6901

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: ran the production cutover's backfill verification, both delta replays and the key reconciliation; measured the stranded-row problem this fixes; wrote both files and the verification below.
- Human verification: the operator authorised every production write and requested this change after reviewing the measured gap. Reviewer sign-off still required.

## Testing

No automated suite covers these drivers; they are operator-run shell against ClickHouse. Verified by rendering and by exercising each guard.

**Default render is byte-identical to the current file**, compared programmatically against pinned revision `7e49662c` (not by inspection) with `${ANALYTICS_DB_DATABASE_NAME}` substituted on both sides and the new defaults applied to this one:

```
DEFAULT RENDER IDENTICAL TO PINNED: YES
```

So the forward path cannot have changed by this PR.

**Post-swap render** with `--old-table traces_pre_cutover_backup --new-table traces` resolves to exactly:

```
INSERT INTO opik_prod.traces
FROM        opik_prod.traces_pre_cutover_backup
DELETE FROM opik_prod.traces
FROM        opik_prod.deletion_events_local   (unchanged)
```

**Guards, each executed rather than asserted:**

| Case | Command | Result |
|---|---|---|
| Non-identifier table name | `--old-table 'traces bad-name'` | exit 2, rejected |
| Same table both sides | `--old-table traces --new-table traces` | exit 2, "would copy a table onto itself" |
| Unknown flag | `--bogus x` | exit 2 |
| Syntax | `bash -n delta_replay.sh` | clean |

Not run: no execution against a live ClickHouse from this branch. The production runs to date used the pinned revision, whose rendered SQL is byte-identical to this one's default.

## Documentation

The flags are documented in the script header alongside the existing ones, including the post-swap invocation and why it exists. No user-facing docs affected — these are internal migration drivers.

Downstream: comet-ml/comet-support-agent#270 depends on this merging; it moves the production cutover's pinned driver revision and adds the operator-side rule constraining the post-swap form to one permitted table pair. It is held as a draft with the SHA unfilled until this lands.

### DIFF
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000002_delta_and_deletion_replay.sql
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000002_delta_and_deletion_replay.sql
@@ -59,7 +59,7 @@
 -- max_insert_threads below sizes the INSERT SELECT pipeline, with the same semantics and costs as in 000001:
 -- omitted when --max-insert-threads is unset, an explicit 0 forcing no parallel execution. The delta writes to
 -- the same table through the same insert path, so pass the value used for the backfill.
-INSERT INTO ${ANALYTICS_DB_DATABASE_NAME}.${NEW_TABLE} (
+INSERT INTO ${ANALYTICS_DB_DATABASE_NAME}.traces_local_v2 (
     id,
     workspace_id,
     project_id,
@@ -108,7 +108,7 @@ SELECT
     coalesce(ttft, toFloat64('nan')) AS ttft,
     source,
     environment
-FROM ${ANALYTICS_DB_DATABASE_NAME}.${OLD_TABLE}
+FROM ${ANALYTICS_DB_DATABASE_NAME}.traces
 WHERE created_at >= toDateTime64('${BACKFILL_START}', 6)
    OR last_updated_at >= toDateTime64('${BACKFILL_START}', 6)
 SETTINGS max_insert_block_size = ${MAX_INSERT_BLOCK_SIZE},
@@ -146,7 +146,7 @@ SETTINGS max_insert_block_size = ${MAX_INSERT_BLOCK_SIZE},
 -- a malformed (non-36-char) deleted_id/project_id can't match a real trace id anyway, so skipping it via the length
 -- guard loses nothing and turns a hard abort mid-cutover into a benign no-op. Same guards in the reverse-replay.
 -- >>> BEGIN deletion-replay
-DELETE FROM ${ANALYTICS_DB_DATABASE_NAME}.${NEW_TABLE}
+DELETE FROM ${ANALYTICS_DB_DATABASE_NAME}.traces_local_v2
 WHERE (
     (workspace_id, project_id, id) IN (
         SELECT
@@ -165,7 +165,7 @@ WHERE (
             workspace_id,
             project_id,
             id
-        FROM ${ANALYTICS_DB_DATABASE_NAME}.${OLD_TABLE}
+        FROM ${ANALYTICS_DB_DATABASE_NAME}.traces
         WHERE id IN (
             SELECT toFixedString(deleted_id, 36)
             FROM ${ANALYTICS_DB_DATABASE_NAME}.deletion_events_local
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000002_delta_and_deletion_replay.sql
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000002_delta_and_deletion_replay.sql
@@ -59,7 +59,7 @@
 -- max_insert_threads below sizes the INSERT SELECT pipeline, with the same semantics and costs as in 000001:
 -- omitted when --max-insert-threads is unset, an explicit 0 forcing no parallel execution. The delta writes to
 -- the same table through the same insert path, so pass the value used for the backfill.
-INSERT INTO ${ANALYTICS_DB_DATABASE_NAME}.traces_local_v2 (
+INSERT INTO ${ANALYTICS_DB_DATABASE_NAME}.${NEW_TABLE} (
     id,
     workspace_id,
     project_id,
@@ -108,7 +108,7 @@ SELECT
     coalesce(ttft, toFloat64('nan')) AS ttft,
     source,
     environment
-FROM ${ANALYTICS_DB_DATABASE_NAME}.traces
+FROM ${ANALYTICS_DB_DATABASE_NAME}.${OLD_TABLE}
 WHERE created_at >= toDateTime64('${BACKFILL_START}', 6)
    OR last_updated_at >= toDateTime64('${BACKFILL_START}', 6)
 SETTINGS max_insert_block_size = ${MAX_INSERT_BLOCK_SIZE},
@@ -146,7 +146,7 @@ SETTINGS max_insert_block_size = ${MAX_INSERT_BLOCK_SIZE},
 -- a malformed (non-36-char) deleted_id/project_id can't match a real trace id anyway, so skipping it via the length
 -- guard loses nothing and turns a hard abort mid-cutover into a benign no-op. Same guards in the reverse-replay.
 -- >>> BEGIN deletion-replay
-DELETE FROM ${ANALYTICS_DB_DATABASE_NAME}.traces_local_v2
+DELETE FROM ${ANALYTICS_DB_DATABASE_NAME}.${NEW_TABLE}
 WHERE (
     (workspace_id, project_id, id) IN (
         SELECT
@@ -165,7 +165,7 @@ WHERE (
             workspace_id,
             project_id,
             id
-        FROM ${ANALYTICS_DB_DATABASE_NAME}.traces
+        FROM ${ANALYTICS_DB_DATABASE_NAME}.${OLD_TABLE}
         WHERE id IN (
             SELECT toFixedString(deleted_id, 36)
             FROM ${ANALYTICS_DB_DATABASE_NAME}.deletion_events_local
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/delta_replay.sh
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/delta_replay.sh
@@ -26,7 +26,15 @@
 #                             post-swap catch-up is `--old-table traces_pre_cutover_backup --new-table traces`. Without
 #                             these the script can only ever run pre-EXCHANGE: it hardcoded both names, so after the swap
 #                             it would read the successor and insert into a table that no longer exists.
-#   --new-table NAME           new-schema destination to write to. Default traces_local_v2. Same swap applies.
+#   --new-table NAME           new-schema destination to write to. Default traces_local_v2. Same swap applies, with one
+#                             exception: ONCE THE DISTRIBUTED WRAP IS APPLIED (exchange_and_wrap.sh --with-wrap or
+#                             --wrap-only) `traces` is a Distributed wrapper over `traces_local`, which accepts the
+#                             delta INSERT but REJECTS the deletion replay's lightweight DELETE ("DELETE query is not
+#                             supported", code 36) — so a post-wrap catch-up aimed at `traces` would write the delta
+#                             and then fail, leaving itself half-applied. Post-wrap the destination is the LOCAL
+#                             table: `--new-table traces_local`. The driver resolves the destination's engine before
+#                             it writes anything and refuses a non-MergeTree destination, so this is enforced rather
+#                             than merely documented.
 #                             `verify.sh` has carried these two flags from the start; this brings the delta into line, so
 #                             the rows a final delta leaves behind (everything written while it ran) and the deletions
 #                             deferred past the swap both have a vehicle.
@@ -203,6 +211,45 @@ if grep -qF '${MAX_INSERT_THREADS}' <<<"$mit_masked"; then
     exit 2
 fi
 # <<< END max_insert_threads rendering
+# Generic post-render guard: NO placeholder may survive into an executable line. The reference SQL's header keeps a
+# hand-maintained inventory of the placeholders this driver substitutes ("it handles all five placeholders below,
+# listed so a new one is never missed"), and a hand-maintained list drifts — a ${...} added to the SQL without a
+# matching substitution here would otherwise reach the server as a literal and fail mid-cutover. This makes that
+# inventory a check rather than a promise, and generalises the MAX_INSERT_THREADS-specific post-condition above to
+# every placeholder. Comment-masked, because the header necessarily writes ${...} in prose (and the batching
+# instructions in step 3 tell the operator to grep for exactly that).
+render_masked="$(mit_mask_comments "$sql" || true)"
+if grep -qE '\$\{[A-Za-z_][A-Za-z0-9_]*\}' <<<"$render_masked"; then
+    echo "ERROR: unsubstituted placeholder(s) survive in executable lines of $SQL_FILE after rendering:" >&2
+    grep -oE '\$\{[A-Za-z_][A-Za-z0-9_]*\}' <<<"$render_masked" | sort -u | sed 's/^/       /' >&2
+    echo "       Refusing to send SQL containing a literal placeholder. Add the substitution to this driver — and to" >&2
+    echo "       exchange_and_wrap.sh, which renders the deletion-replay block of the same file." >&2
+    exit 2
+fi
+# Pre-flight, BEFORE the first write: the destination must accept BOTH statements this driver sends. The delta
+# INSERT works against any writable engine, but the deletion replay is a lightweight DELETE and only the MergeTree
+# family supports one. After the Distributed wrap, `traces` is a Distributed wrapper over `traces_local`: the INSERT
+# would land and the DELETE would then fail with "DELETE query is not supported" (code 36), leaving the catch-up
+# half-applied — the delta written, the deferred deletions not. Re-running does not repair that by itself, because
+# the DELETE keeps failing until the destination is corrected. Positive match on the MergeTree family, so every
+# other non-mutable engine (Distributed, Buffer, Merge, Null, a view) is refused too.
+dest_engine="$(clickhouse-client ${CH_HOST:+--host $CH_HOST} ${CH_PORT:+--port $CH_PORT} --database "$DATABASE" --log_comment 'traces_local_v2_cutover:delta_replay:destination_engine' --query "SELECT engine FROM system.tables WHERE database = '$DATABASE' AND name = '$NEW_TABLE'")"
+if [[ -z "$dest_engine" ]]; then
+    echo "ERROR: destination '$DATABASE.$NEW_TABLE' does not exist on this instance. Pass --new-table for whatever" >&2
+    echo "       holds the new-schema data here: pre-EXCHANGE 'traces_local_v2', post-EXCHANGE 'traces'," >&2
+    echo "       post-wrap 'traces_local'." >&2
+    exit 1
+fi
+if [[ "$dest_engine" != *MergeTree ]]; then
+    echo "ERROR: destination '$DATABASE.$NEW_TABLE' has engine '$dest_engine'. The deletion replay is a lightweight" >&2
+    echo "       DELETE and only the MergeTree family supports one, so the delta INSERT would succeed and the replay" >&2
+    echo "       would then fail, leaving this catch-up half-applied. Refusing before anything is written." >&2
+    if [[ "$dest_engine" == "Distributed" ]]; then
+        echo "       'traces' is the Distributed wrap over 'traces_local'; re-run with --new-table traces_local." >&2
+    fi
+    exit 1
+fi
+
 # --time makes clickhouse-client print each statement's elapsed seconds to stderr (it prints nothing under a bare
 # --query). The SECOND number is the deletion replay's wall time — a Go/No-Go acceptance criterion (it must fit inside
 # the buffer hold with margin), so without this the operator has no way to record it short of digging in query_log.
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/delta_replay.sh
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/delta_replay.sh
@@ -22,22 +22,6 @@
 #                             to localhost. User/password still come from CLICKHOUSE_USER / CLICKHOUSE_PASSWORD (keeping
 #                             the password out of argv).
 #   --backfill-start TS        the anchor printed by backfill.sh ("RECORD backfill_start=..."). Required.
-#   --old-table NAME           old-schema source to read from. Default traces. AFTER THE EXCHANGE the names swap, so a
-#                             post-swap catch-up is `--old-table traces_pre_cutover_backup --new-table traces`. Without
-#                             these the script can only ever run pre-EXCHANGE: it hardcoded both names, so after the swap
-#                             it would read the successor and insert into a table that no longer exists.
-#   --new-table NAME           new-schema destination to write to. Default traces_local_v2. Same swap applies, with one
-#                             exception: ONCE THE DISTRIBUTED WRAP IS APPLIED (exchange_and_wrap.sh --with-wrap or
-#                             --wrap-only) `traces` is a Distributed wrapper over `traces_local`, which accepts the
-#                             delta INSERT but REJECTS the deletion replay's lightweight DELETE ("DELETE query is not
-#                             supported", code 36) — so a post-wrap catch-up aimed at `traces` would write the delta
-#                             and then fail, leaving itself half-applied. Post-wrap the destination is the LOCAL
-#                             table: `--new-table traces_local`. The driver resolves the destination's engine before
-#                             it writes anything and refuses a non-MergeTree destination, so this is enforced rather
-#                             than merely documented.
-#                             `verify.sh` has carried these two flags from the start; this brings the delta into line, so
-#                             the rows a final delta leaves behind (everything written while it ran) and the deletions
-#                             deferred past the swap both have a vehicle.
 #   --max-insert-block-size N  SETTINGS max_insert_block_size for the delta INSERT. Default 1048576.
 #   --max-partitions-per-insert-block N
 #                             partitions one insert block of the delta INSERT may span (SETTINGS
@@ -55,11 +39,39 @@
 #                             count) as in backfill.sh -- see its option docs for the full diagnosis.
 #                             PASS THE SAME VALUE USED FOR THE BACKFILL: the delta writes into the
 #                             same table through the same insert path.
+#
+# PRE-EXCHANGE ONLY, and deliberately not configurable. The source and destination are fixed at `traces` ->
+# `traces_local_v2` in the reference SQL, so this driver can only ever run BEFORE the swap. `verify.sh` takes
+# --old-table/--new-table and can therefore be re-pointed after the EXCHANGE, but it only SELECTs; re-pointing a
+# driver that INSERTs and DELETEs is a different proposition, and the naive swap
+# (`--old-table traces_pre_cutover_backup --new-table traces`) is NOT correct -- it was tried in opik#8092 and
+# withdrawn. Two independent reasons, both about the destination being live rather than a private shadow:
+#
+#   * The delta-insert would RESURRECT user deletes. A lightweight DELETE does not bump last_updated_at (the
+#     ReplacingMergeTree version column) -- that is the whole reason deletion_events_local exists (see
+#     000096_create_deletion_events_local.sql: "deletes ... are invisible to the version-based delta step and would
+#     reappear in the copy"). A row written during the backfill window and deleted by a user AFTER the swap is still
+#     unmasked in the frozen backup, so the delta re-copies it into live `traces` with _row_exists = 1 and it is
+#     live again. The deletion replay in the same run cannot undo it: that id IS live on the read side, so the
+#     resurrection guard spares it.
+#   * The resurrection guard would evaluate liveness against a SNAPSHOT. Post-swap the guard's source is the frozen
+#     backup, so an id bridged as deleted and then re-created post-swap is absent there, passes the guard, and is
+#     DELETED from the live table. 000004_rollback_reverse_replay.sql already documents this hazard class for the
+#     mirror case and refuses to carry a guard for exactly this reason.
+#
+# Neither is fixable by choosing better table names: post-swap the copy source, the liveness oracle and the
+# destination are three different tables, and the replay additionally needs an upper bound at cutover_start (the
+# events after it are post-swap user activity, not deferred work). A correct post-swap catch-up is a separate
+# statement pair with its own anchor and its own gate cases -- not a re-pointing of 000002. Note also that the
+# runbook does not expect one: README "Keep step 3->4 short" covers gap WRITES with the async-insert buffer and gap
+# DELETES with exchange_and_wrap.sh's final deletion replay. If either left rows behind on a real run, the fix
+# belongs in whichever of those two failed.
 
 set -euo pipefail
 
-OLD_TABLE="traces"          # old-schema side; becomes traces_pre_cutover_backup after the EXCHANGE (see --old-table)
-NEW_TABLE="traces_local_v2" # new-schema side (the successor being built); becomes traces after the EXCHANGE
+# The destination the reference SQL writes to, as a literal. NOT interpolated into the SQL -- 000002 names both
+# tables itself -- and used only by the pre-flight below, so the check and the statements cannot drift apart.
+SHADOW_TABLE="traces_local_v2"
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 SQL_FILE="$SCRIPT_DIR/db-app-analytics/000002_delta_and_deletion_replay.sql"
@@ -77,8 +89,6 @@ while [[ $# -gt 0 ]]; do
     case "$1" in
         --database) DATABASE="${2:?"$1 requires a value"}"; shift 2 ;;
         --backfill-start) BACKFILL_START="${2:?"$1 requires a value"}"; shift 2 ;;
-        --old-table) OLD_TABLE="${2:?"$1 requires a value"}"; shift 2 ;;
-        --new-table) NEW_TABLE="${2:?"$1 requires a value"}"; shift 2 ;;
         --max-insert-block-size) MAX_INSERT_BLOCK_SIZE="${2:?"$1 requires a value"}"; shift 2 ;;
         --max-partitions-per-insert-block) MAX_PARTITIONS_PER_INSERT_BLOCK="${2:?"$1 requires a value"}"; shift 2 ;;
         --max-insert-threads) MAX_INSERT_THREADS="${2:?"$1 requires a value"}"; shift 2 ;;
@@ -107,16 +117,7 @@ echo "Reminder: raise databaseAnalytics.asyncInsertBusyTimeoutMaxMs before this 
 echo "restore it after the EXCHANGE."
 
 sql="$(cat "$SQL_FILE")"
-# Interpolated straight into the reference SQL, so require plain ClickHouse identifiers — same
-# guard verify.sh applies to the same three values.
-for _ident in "$DATABASE" "$OLD_TABLE" "$NEW_TABLE"; do
-    [[ "$_ident" =~ ^[A-Za-z0-9_]+$ ]] || { echo "ERROR: --database/--old-table/--new-table must be ClickHouse identifiers (letters, digits, underscore): '$_ident'" >&2; exit 2; }
-done
-[[ "$OLD_TABLE" != "$NEW_TABLE" ]] || { echo "ERROR: --old-table and --new-table must differ; '$OLD_TABLE' would copy a table onto itself." >&2; exit 2; }
-
 sql="${sql//'${ANALYTICS_DB_DATABASE_NAME}'/$DATABASE}"
-sql="${sql//'${OLD_TABLE}'/$OLD_TABLE}"
-sql="${sql//'${NEW_TABLE}'/$NEW_TABLE}"
 sql="${sql//'${BACKFILL_START}'/$BACKFILL_START}"
 sql="${sql//'${MAX_INSERT_BLOCK_SIZE}'/$MAX_INSERT_BLOCK_SIZE}"
 sql="${sql//'${MAX_PARTITIONS_PER_INSERT_BLOCK}'/$MAX_PARTITIONS_PER_INSERT_BLOCK}"
@@ -226,27 +227,30 @@ if grep -qE '\$\{[A-Za-z_][A-Za-z0-9_]*\}' <<<"$render_masked"; then
     echo "       exchange_and_wrap.sh, which renders the deletion-replay block of the same file." >&2
     exit 2
 fi
-# Pre-flight, BEFORE the first write: the destination must accept BOTH statements this driver sends. The delta
-# INSERT works against any writable engine, but the deletion replay is a lightweight DELETE and only the MergeTree
-# family supports one. After the Distributed wrap, `traces` is a Distributed wrapper over `traces_local`: the INSERT
-# would land and the DELETE would then fail with "DELETE query is not supported" (code 36), leaving the catch-up
-# half-applied — the delta written, the deferred deletions not. Re-running does not repair that by itself, because
-# the DELETE keeps failing until the destination is corrected. Positive match on the MergeTree family, so every
-# other non-mutable engine (Distributed, Buffer, Merge, Null, a view) is refused too.
-dest_engine="$(clickhouse-client ${CH_HOST:+--host $CH_HOST} ${CH_PORT:+--port $CH_PORT} --database "$DATABASE" --log_comment 'traces_local_v2_cutover:delta_replay:destination_engine' --query "SELECT engine FROM system.tables WHERE database = '$DATABASE' AND name = '$NEW_TABLE'")"
+# Pre-flight, BEFORE the first write: this step is only valid at one point in the sequence, so check that we are
+# actually at it rather than discovering otherwise mid-statement. The destination is the shadow, `traces_local_v2`,
+# which exists only between step 1 (backfill.sh creates and fills it) and step 3 (exchange_and_wrap.sh renames it
+# away). Absent means the run is out of sequence -- before the backfill, or already past the EXCHANGE -- and the
+# INSERT would otherwise fail with a bare UNKNOWN_TABLE after the operator reminder had already printed.
+#
+# The engine arm is a belt-and-braces check on the same object. The delta INSERT works against any writable engine,
+# but the deletion replay is a lightweight DELETE and only the MergeTree family supports one; a non-mutable engine
+# (Distributed, Buffer, Merge, Null, a view) would take the INSERT and then reject the DELETE, leaving the run
+# half-applied. Positive match on the family, so anything unexpected is refused rather than allowed by omission.
+dest_engine="$(clickhouse-client ${CH_HOST:+--host $CH_HOST} ${CH_PORT:+--port $CH_PORT} --database "$DATABASE" --log_comment 'traces_local_v2_cutover:delta_replay:destination_engine' --query "SELECT engine FROM system.tables WHERE database = '$DATABASE' AND name = '$SHADOW_TABLE'")"
 if [[ -z "$dest_engine" ]]; then
-    echo "ERROR: destination '$DATABASE.$NEW_TABLE' does not exist on this instance. Pass --new-table for whatever" >&2
-    echo "       holds the new-schema data here: pre-EXCHANGE 'traces_local_v2', post-EXCHANGE 'traces'," >&2
-    echo "       post-wrap 'traces_local'." >&2
+    echo "ERROR: '$DATABASE.$SHADOW_TABLE' does not exist on this instance, so this step is out of sequence." >&2
+    echo "       The shadow exists only between the backfill (step 1) and the EXCHANGE (step 3). Before the" >&2
+    echo "       backfill, run backfill.sh first. After the EXCHANGE, the shadow is the live 'traces' and the old" >&2
+    echo "       data is parked as 'traces_pre_cutover_backup' -- this driver is pre-EXCHANGE only and there is no" >&2
+    echo "       post-swap catch-up mode (see the header). Nothing was written." >&2
     exit 1
 fi
 if [[ "$dest_engine" != *MergeTree ]]; then
-    echo "ERROR: destination '$DATABASE.$NEW_TABLE' has engine '$dest_engine'. The deletion replay is a lightweight" >&2
-    echo "       DELETE and only the MergeTree family supports one, so the delta INSERT would succeed and the replay" >&2
-    echo "       would then fail, leaving this catch-up half-applied. Refusing before anything is written." >&2
-    if [[ "$dest_engine" == "Distributed" ]]; then
-        echo "       'traces' is the Distributed wrap over 'traces_local'; re-run with --new-table traces_local." >&2
-    fi
+    echo "ERROR: destination '$DATABASE.$SHADOW_TABLE' has engine '$dest_engine', not a MergeTree. The deletion" >&2
+    echo "       replay is a lightweight DELETE and only the MergeTree family supports one, so the delta INSERT" >&2
+    echo "       would succeed and the replay would then fail, leaving this run half-applied. Refusing before" >&2
+    echo "       anything is written." >&2
     exit 1
 fi
 
--- a/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/delta_replay.sh
+++ b/apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/delta_replay.sh
@@ -22,6 +22,14 @@
 #                             to localhost. User/password still come from CLICKHOUSE_USER / CLICKHOUSE_PASSWORD (keeping
 #                             the password out of argv).
 #   --backfill-start TS        the anchor printed by backfill.sh ("RECORD backfill_start=..."). Required.
+#   --old-table NAME           old-schema source to read from. Default traces. AFTER THE EXCHANGE the names swap, so a
+#                             post-swap catch-up is `--old-table traces_pre_cutover_backup --new-table traces`. Without
+#                             these the script can only ever run pre-EXCHANGE: it hardcoded both names, so after the swap
+#                             it would read the successor and insert into a table that no longer exists.
+#   --new-table NAME           new-schema destination to write to. Default traces_local_v2. Same swap applies.
+#                             `verify.sh` has carried these two flags from the start; this brings the delta into line, so
+#                             the rows a final delta leaves behind (everything written while it ran) and the deletions
+#                             deferred past the swap both have a vehicle.
 #   --max-insert-block-size N  SETTINGS max_insert_block_size for the delta INSERT. Default 1048576.
 #   --max-partitions-per-insert-block N
 #                             partitions one insert block of the delta INSERT may span (SETTINGS
@@ -42,6 +50,9 @@
 
 set -euo pipefail
 
+OLD_TABLE="traces"          # old-schema side; becomes traces_pre_cutover_backup after the EXCHANGE (see --old-table)
+NEW_TABLE="traces_local_v2" # new-schema side (the successor being built); becomes traces after the EXCHANGE
+
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 SQL_FILE="$SCRIPT_DIR/db-app-analytics/000002_delta_and_deletion_replay.sql"
 
@@ -58,6 +69,8 @@ while [[ $# -gt 0 ]]; do
     case "$1" in
         --database) DATABASE="${2:?"$1 requires a value"}"; shift 2 ;;
         --backfill-start) BACKFILL_START="${2:?"$1 requires a value"}"; shift 2 ;;
+        --old-table) OLD_TABLE="${2:?"$1 requires a value"}"; shift 2 ;;
+        --new-table) NEW_TABLE="${2:?"$1 requires a value"}"; shift 2 ;;
         --max-insert-block-size) MAX_INSERT_BLOCK_SIZE="${2:?"$1 requires a value"}"; shift 2 ;;
         --max-partitions-per-insert-block) MAX_PARTITIONS_PER_INSERT_BLOCK="${2:?"$1 requires a value"}"; shift 2 ;;
         --max-insert-threads) MAX_INSERT_THREADS="${2:?"$1 requires a value"}"; shift 2 ;;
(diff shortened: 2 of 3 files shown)


CometActions:
**No test needed here.**

Heads up that the title is now stale: the diff does not add --old-table/--new-table. The new header explicitly withdraws that idea (post-swap re-pointing would resurrect user deletes and evaluate the resurrection guard against a snapshot), so what actually lands is comments plus three fail-fast guards — the unsubstituted-placeholder check in both drivers, and the traces_local_v2 existence/MergeTree pre-flight in delta_replay.sh. These are operator runbook scripts: nothing in the running product invokes them, no Opik page or endpoint behaves differently after this merges, and a fresh OSS install has no shadow table to run them against, so there is nothing here for an e2e test to assert. If you want the guards pinned, a shell-level test beside the scripts is the right home — TracesLocalV2CutoverTest is the executable spec for the cutover SQL, not for these drivers, and it does not exercise the rendering or pre-flight paths.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/33501154887)

_Re-checked after a push on 01 Sep 11:15 UTC._

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000002_delta_and_deletion_replay.sql
```diff
@@ -146,7 +146,7 @@ SETTINGS max_insert_block_size = ${MAX_INSERT_BLOCK_SIZE},
 -- a malformed (non-36-char) deleted_id/project_id can't match a real trace id anyway, so skipping it via the length
 -- guard loses nothing and turns a hard abort mid-cutover into a benign no-op. Same guards in the reverse-replay.
 -- >>> BEGIN deletion-replay
-DELETE FROM ${ANALYTICS_DB_DATABASE_NAME}.traces_local_v2
+DELETE FROM ${ANALYTICS_DB_DATABASE_NAME}.${NEW_TABLE}
```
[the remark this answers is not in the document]
thiagohora: **Correct, and the most serious of the three — this was a regression this PR introduced.** Thank you.

`exchange_and_wrap.sh` does read this file: `DELTA_SQL_FILE` (line 77) and `run_final_deletion_replay`, which `awk`s out the `-- >>> BEGIN/END deletion-replay` block and substituted only `ANALYTICS_DB_DATABASE_NAME` and `BACKFILL_START`. Templating the table names left `\${NEW_TABLE}` and `\${OLD_TABLE}` to reach the server as literals.

Reproduced rather than reasoned about — the pushed HEAD `f06109c3c4` driven against a capturing stub `clickhouse-client`:

```
placeholders that reached clickhouse-client:
${NEW_TABLE}
${OLD_TABLE}
```

Against a real server that statement is a syntax error, so under `set -e` `run_final_deletion_replay` aborts before `run_block exchange` — the EXCHANGE path broken outright, exactly as you describe.

**Fixed in 1ff4f73** by binding the names and substituting them:

```bash
DELTA_OLD_TABLE="traces"
DELTA_NEW_TABLE="traces_local_v2"
```

Constants rather than flags, deliberately. This driver renders that block only *before* the EXCHANGE (`run_final_deletion_replay` is called after `cutover_start` is captured and before `run_block exchange`), so the names are always the pre-swap pair — and they are the same fixed names `000003`'s own `EXCHANGE`/`RENAME` statements carry. Making them flags would invite a caller to disagree with `000003`.

Verified after the fix: the rendered deletion-replay is byte-identical to what the pinned cutover revision `7e49662c` produces (same sha256), so the EXCHANGE path is back to exactly its pre-PR behaviour. `run_block` and `run_final_deletion_replay` now also both run an `assert_no_placeholder` post-render check, so a future edit to either SQL file cannot reintroduce this silently — see the reply on line 62.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000002_delta_and_deletion_replay.sql
```diff
@@ -59,7 +59,7 @@
 -- max_insert_threads below sizes the INSERT SELECT pipeline, with the same semantics and costs as in 000001:
 -- omitted when --max-insert-threads is unset, an explicit 0 forcing no parallel execution. The delta writes to
 -- the same table through the same insert path, so pass the value used for the backfill.
-INSERT INTO ${ANALYTICS_DB_DATABASE_NAME}.traces_local_v2 (
+INSERT INTO ${ANALYTICS_DB_DATABASE_NAME}.${NEW_TABLE} (
```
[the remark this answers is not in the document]
thiagohora: **The observation is right; the requested fix is the one change I can't make here.** The header does say "all five placeholders" while the file now interpolates seven.

The reason it stays: `delta_replay.sh` does `sql=\$(cat "\$SQL_FILE")` and sends the **whole file, comments included**, to `clickhouse-client`. The header is inside the rendered bytes. The default render (database substituted, `OLD_TABLE=traces`, `NEW_TABLE=traces_local_v2`) has to stay byte-identical to pinned cutover revision `7e49662c`, which is what an in-flight production cutover is running from — that is the property that makes this PR safe to land mid-window.

Counterfactual, run through the real driver against a stub client — changing `all five` to `all seven` and nothing else:

```
6c6
< -- ... It handles all five
---
> -- ... It handles all seven
```

One diff, and it is precisely the invariant. Meanwhile the four placeholder substitutions this PR *does* make render back to the same bytes (`9661663040974e2f…`, identical both sides).

**So I addressed the underlying defect instead of the sentence.** The header's stated purpose — "listed so a new one is never missed here (an unsubstituted \${...} reaches the server as a literal and the statement fails)" — is something a hand-maintained list structurally cannot deliver; this PR is the proof. Both drivers now enforce it after rendering, in the `.sh` files, outside the pinned bytes:

- `delta_replay.sh` — a generic post-render guard: any surviving `\${...}` in an **executable** line aborts before a statement is sent. Comment-masked via the existing `mit_mask_comments`, because the header writes `\${...}` in prose and step 3 tells the operator to grep for exactly that.
- `exchange_and_wrap.sh` — `assert_no_placeholder`, applied to both `run_block` and `run_final_deletion_replay`. The inventory never mentioned that driver at all, which is how the `\${NEW_TABLE}`/`\${OLD_TABLE}` break on line 149 got through in the first place.

Both executed, not reasoned about. Injecting `max_execution_time = \${MAX_EXECUTION_TIME}` into 000002's SETTINGS clause:

```
ERROR: unsubstituted placeholder(s) survive in executable lines of …000002…sql after rendering:
       ${MAX_EXECUTION_TIME}
exit=2   (no statement sent)
```

The same mutated SQL against the **pinned** driver sends the literal to the client and exits 0 — the gap this closes. A placeholder injected into `000003` is likewise refused by `exchange_and_wrap.sh`, before the EXCHANGE.

The header sentence is a real (documentation-only) defect and should be corrected — in the commit that next moves the pin, where changing the rendered bytes is expected and re-verified. Tracked with the cutover notes.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/delta_replay.sh
```diff
@@ -58,6 +69,8 @@ while [[ $# -gt 0 ]]; do
     case "$1" in
         --database) DATABASE="${2:?"$1 requires a value"}"; shift 2 ;;
         --backfill-start) BACKFILL_START="${2:?"$1 requires a value"}"; shift 2 ;;
+        --old-table) OLD_TABLE="${2:?"$1 requires a value"}"; shift 2 ;;
+        --new-table) NEW_TABLE="${2:?"$1 requires a value"}"; shift 2 ;;
```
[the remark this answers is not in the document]
thiagohora: **Partly valid — the mechanism is real, the scope is narrower than stated, and it is now enforced rather than documented.**

Where it's right: once the wrap is applied, `traces` is a `Distributed` wrapper over `traces_local`, which accepts an INSERT and rejects a lightweight DELETE. `000003`'s own wrap block spells out the codes — *"a lightweight DELETE returns 'DELETE query is not supported' (code 36) and ALTER … DELETE returns 'Distributed doesn't support mutations' (code 48)"*. With `--multiquery`, the delta INSERT commits and the replay then fails: half-applied, and re-running does not repair it, because the DELETE keeps failing until the destination changes.

Where the finding overreaches: it reads the documented invocation as unconditionally wrong. It isn't. `exchange_and_wrap.sh` **defers the wrap by default** — the wrap is opt-in behind `--with-wrap`/`--wrap-only` *and* `--confirm-daos-retargeted`, precisely so deletes keep working. In the default post-EXCHANGE state `traces` is still a MergeTree and `--old-table traces_pre_cutover_backup --new-table traces` is correct. The hazard is the wrapped state specifically, where the destination is `traces_local`.

**Neither of your two suggestions, quite.** "Require this command before the wrap" would forbid the legitimate post-wrap catch-up, which is one of the two things these flags exist for. "Resolve the destination as writable `traces_local`" hardcodes a name the flags exist to stop hardcoding. So, the property rather than the name — in 1ff4f73, `delta_replay.sh` resolves the destination's engine before it writes anything:

```bash
dest_engine="\$(clickhouse-client … --query "SELECT engine FROM system.tables WHERE database = '\$DATABASE' AND name = '\$NEW_TABLE'")"
…
if [[ "\$dest_engine" != *MergeTree ]]; then   # refuse
```

A positive match on the MergeTree family, so `Distributed`, `Buffer`, `Merge`, `Null` and views are all refused, not just the one engine we thought of. A missing destination is refused too. Same `system.tables` engine probe `backfill.sh`, `rollback.sh` and `exchange_and_wrap.sh` already use, so it needs no privilege the drivers don't have.

Executed, against a stub client:

```
$ … --old-table traces_pre_cutover_backup --new-table traces      # wrap applied
ERROR: destination 'opik.traces' has engine 'Distributed'. …
       'traces' is the Distributed wrap over 'traces_local'; re-run with --new-table traces_local.
exit=1   (no statement sent — nothing written)

$ … --old-table traces_pre_cutover_backup --new-table traces_local
INSERT INTO opik.traces_local / DELETE FROM opik.traces_local     # proceeds
```

The `--new-table` option docs now state the post-wrap destination and the code-36 reason. The default render is unchanged and still byte-identical to pin `7e49662c`; the guard is a pre-flight read, before the first write.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000002_delta_and_deletion_replay.sql
```diff
@@ -108,7 +108,7 @@ SELECT
     coalesce(ttft, toFloat64('nan')) AS ttft,
     source,
     environment
-FROM ${ANALYTICS_DB_DATABASE_NAME}.traces
+FROM ${ANALYTICS_DB_DATABASE_NAME}.${OLD_TABLE}
```
andrescrz: **[high] Post-EXCHANGE delta-insert resurrects rows the user deleted on the live successor**

Parameterizing the source as ${OLD_TABLE} enables the post-swap pairing the driver documents (--old-table traces_pre_cutover_backup --new-table traces), but the delta-insert's correctness argument (lines 28-32: 'ReplacingMergeTree dedups the re-copied rows against the backfilled ones') only holds while ${OLD_TABLE} IS the live source. After the EXCHANGE the old table is frozen and the live table is ${NEW_TABLE}. A trace that was written during the backfill window (so it matches created_at OR last_updated_at >= BACKFILL_START in the frozen backup) and is deleted by a user AFTER the swap is still live in the backup, so this INSERT re-copies it into the live successor. A lightweight DELETE only flips the hidden row mask and does NOT bump last_updated_at (the ReplacingMergeTree version column, 000017/000101) — the re-inserted copy lands in a new part with _row_exists = 1, so it is live again regardless of dedup. The deletion replay in the same run cannot undo it: that id IS live in ${OLD_TABLE}, so the resurrection guard (line 163) excludes it. Net effect of a post-swap catch-up: post-swap user deletes are silently reverted in the production table — the exact deletion-leak class the deletion-events bridge exists to prevent, and a data-retention/privacy regression.

💡 Do not let the liveness/mask source be the same placeholder as the copy source. Either (a) restrict the post-swap mode to an explicitly narrowed anchor plus a bridge replay that is evaluated against the LIVE table — add a separate ${LIVE_TABLE} placeholder (defaulting to ${OLD_TABLE} pre-swap, ${NEW_TABLE} post-swap) and reorder so the deletion replay runs AFTER the delta-insert against ${LIVE_TABLE}'s bridge events — or (b) add an explicit post-swap section to this file stating that a post-EXCHANGE delta must be followed by a deletion replay whose window covers [EXCHANGE, now] evaluated on ${NEW_TABLE}, and gate the mode on user deletes being quiesced. At minimum document the hazard here; today the file reads as pre-EXCHANGE-only while the driver advertises the post-swap invocation.

_rule: `backend-architecture-gate-every-migration-cutover-step-on-the-full-precondition-c`_

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/delta_replay.sh
```diff
@@ -86,7 +107,16 @@ echo "Reminder: raise databaseAnalytics.asyncInsertBusyTimeoutMaxMs before this
...
+# Interpolated straight into the reference SQL, so require plain ClickHouse identifiers — same
+# guard verify.sh applies to the same three values.
+for _ident in "$DATABASE" "$OLD_TABLE" "$NEW_TABLE"; do
+    [[ "$_ident" =~ ^[A-Za-z0-9_]+$ ]] || { echo "ERROR: --database/--old-table/--new-table must be ClickHouse identifiers (letters, digits, underscore): '$_ident'" >&2; exit 2; }
+done
+[[ "$OLD_TABLE" != "$NEW_TABLE" ]] || { echo "ERROR: --old-table and --new-table must differ; '$OLD_TABLE' would copy a table onto itself." >&2; exit 2; }
+
 sql="${sql//'${ANALYTICS_DB_DATABASE_NAME}'/$DATABASE}"
+sql="${sql//'${OLD_TABLE}'/$OLD_TABLE}"
+sql="${sql//'${NEW_TABLE}'/$NEW_TABLE}"
```
andrescrz: **[high] New ${OLD_TABLE}/${NEW_TABLE} placeholders are not substituted by the other consumer of the same SQL file — aborts the cutover right before the EXCHANGE**

This PR parameterizes db-app-analytics/000002_delta_and_deletion_replay.sql with ${OLD_TABLE}/${NEW_TABLE} and teaches only this driver to substitute them (lines 110-111). But delta_replay.sh is NOT the only consumer of that file: exchange_and_wrap.sh:307-314 (run_final_deletion_replay) extracts the '-- >>> BEGIN deletion-replay' block from the same $DELTA_SQL_FILE and substitutes ONLY ${ANALYTICS_DB_DATABASE_NAME} and ${BACKFILL_START}. After this change that block also contains ${NEW_TABLE} (000002 line 149) and ${OLD_TABLE} (000002 line 168), so exchange_and_wrap.sh now sends literal 'DELETE FROM opik.${NEW_TABLE} ... NOT IN (SELECT ... FROM opik.${OLD_TABLE} ...)' to ClickHouse, which is a syntax error. exchange_and_wrap.sh runs under `set -euo pipefail` (line 73) and calls run_final_deletion_replay AFTER capturing cutover_start and BEFORE run_block exchange, so the script dies in the tightest part of the cutover: cutover_start is already recorded, the async-insert buffer is holding writes, the final deletion replay did not run and the EXCHANGE never happens. 000002's own step-0 comment ('listed so a new one is never missed here (an unsubstituted ${...} reaches the server as a literal and the statement fails)') describes exactly this failure.

💡 Add the two substitutions to exchange_and_wrap.sh::run_final_deletion_replay in the same change (`sql="${sql//'${OLD_TABLE}'/$OLD_TABLE}"` / `${NEW_TABLE}`, with the same defaults traces / traces_local_v2, and ideally the same --old-table/--new-table flags), or add a post-substitution guard there that refuses to send SQL still containing a literal '${' — mirroring the ${MAX_INSERT_THREADS} post-condition already in delta_replay.sh lines 196-204. A rehearsal of the full sequence (backfill -> delta -> exchange) would have caught this.

_rule: `backend-data-access-schema-type-precision-and-enum-changes-must-be-applied-to-ev`_

andrescrz (commented):

9 findings — 3 high · 5 medium · 1 low.

> ⚠️ **Generated against `f06109c3`, not the current head `1ff4f730`.** The follow-up commit ("fix the templating regression in exchange_and_wrap") lands after this analysis, so some findings — notably the `exchange_and_wrap.sh` substitution gap and the post-wrap Distributed `DELETE` half-apply — appear to be **already addressed** there. Line anchors have been mapped forward to the current head. Please treat each finding as needing confirmation against `1ff4f730`.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000002_delta_and_deletion_replay.sql
```diff
@@ -165,7 +165,7 @@ WHERE (
             workspace_id,
             project_id,
             id
-        FROM ${ANALYTICS_DB_DATABASE_NAME}.traces
+        FROM ${ANALYTICS_DB_DATABASE_NAME}.${OLD_TABLE}
```
andrescrz: **[high] Resurrection guard checks a table that is no longer the live source after the swap**

The RESURRECTION GUARD (lines 126-133) and the idempotency claim on line 136 ('never masks a live-on-source id, so re-runs converge') both depend on ${OLD_TABLE} being the live, currently-written table. In the post-swap pairing the placeholders now permit, ${OLD_TABLE} is the frozen traces_pre_cutover_backup, so 'live on the source' is evaluated against a snapshot. Every row created after the EXCHANGE is absent from that snapshot, so an id that was bridged as deleted since BACKFILL_START and then re-created post-swap under the same id (trace ids are client-supplied — the very premise of the guard, line 127) passes the guard and is deleted from the LIVE successor: silent loss of production data, inside the cutover window. The stale prose makes this hard to spot: line 128 still calls it 'the `NOT IN traces` arm' and lines 132-133 still reason about index coverage on 'traces' / 'traces_local_v2' by literal name.

💡 Evaluate the guard against the live table rather than the copy source (separate ${LIVE_TABLE} placeholder, or ${NEW_TABLE} when running post-swap), and update the comment block to name the arm by placeholder (${OLD_TABLE}) plus the pre/post-EXCHANGE mapping instead of the literal table names, so the index/pruning claims are checkable in both modes.

_rule: `backend-data-access-verify-clickhouse-query-safety-claims-against-actual-schema-`_

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000002_delta_and_deletion_replay.sql
```diff
@@ -165,7 +165,7 @@ WHERE (
...
         WHERE id IN (
             SELECT toFixedString(deleted_id, 36)
             FROM ${ANALYTICS_DB_DATABASE_NAME}.deletion_events_local
```
andrescrz, line 171: **[medium] Safety and rollback notes are false for the post-EXCHANGE run this change enables**

_(Concerns line 192, which is outside this PR's diff; anchored at line 171 — the nearest changed line in this file — so GitHub accepts it.)_

The file's safety framing assumes both statements only ever touch the shadow: line 3 ('Run this only after the whole backfill (step 1) is complete'), line 39 ('the final delta runs immediately before the EXCHANGE'), and the rollback note on lines 191-193 — 'If aborting the cutover here, TRUNCATE traces_local_v2 (step 1 rollback) ... The live `traces` table is still untouched until the EXCHANGE in step 3'. With ${NEW_TABLE} substitutable, a post-swap run writes and DELETEs against the LIVE production table, so that claim is wrong and the recovery advice is both stale (traces_local_v2 no longer exists post-EXCHANGE; it was renamed to traces_pre_cutover_backup) and dangerous if an operator generalizes it to 'truncate the new table'. A destructive step gained a new target with no matching rollback story.

💡 Split the rollback note per mode: pre-EXCHANGE keep the TRUNCATE of the shadow; post-EXCHANGE state explicitly that ${NEW_TABLE} is the live table, that TRUNCATE is never the recovery, and what the actual recovery is (rollback.sh stages, since the reverse-replay in 000004 keys off CUTOVER_START, not BACKFILL_START).

_rule: `backend-error-handling-destructive-migrations-and-cutovers-need-a-real-rollback-ver`_

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000002_delta_and_deletion_replay.sql
```diff
@@ -59,7 +59,7 @@
 -- max_insert_threads below sizes the INSERT SELECT pipeline, with the same semantics and costs as in 000001:
```
andrescrz, line 59: **[medium] New post-EXCHANGE table pairing has no gate-test coverage, despite this file's keep-in-step mandate**

_(Concerns line 2, which is outside this PR's diff; anchored at line 59 — the nearest changed line in this file — so GitHub accepts it.)_

Line 2 requires TracesLocalV2CutoverTest and these statements to be kept in step, and that test reimplements them inline hardcoded to traces -> traces_local_v2 (its deltaInsert/replay helpers and every assertion; the only post-swap delta it runs is the rollback-retry path, which first renames the shadow back to traces_local_v2). The mode this change introduces — reading a frozen old table and writing/mutating the live successor — has materially different semantics (both findings above) and is untested: no case seeds a post-swap delete or a post-swap re-created id and asserts what the delta + replay do to it. The test's own scope note defers driver-guard coverage to the staging dry-run, but this is SQL logic, which is precisely what the gate owns.

💡 Add a gate case that runs the delta-insert + deletion replay after the EXCHANGE with the swapped pairing and asserts the two lifecycle branches: (a) a row deleted on the live successor post-swap stays deleted, and (b) an id bridged as deleted pre-swap and re-created post-swap stays alive. Both currently fail, which is the point of adding them.

_rule: `backend-testing-test-sql-templates-and-load-bearing-gating-logic-with-real-q`_

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/delta_replay.sh
```diff
@@ -86,7 +107,16 @@ echo "Reminder: raise databaseAnalytics.asyncInsertBusyTimeoutMaxMs before this
...
 sql="$(cat "$SQL_FILE")"
+# Interpolated straight into the reference SQL, so require plain ClickHouse identifiers — same
+# guard verify.sh applies to the same three values.
+for _ident in "$DATABASE" "$OLD_TABLE" "$NEW_TABLE"; do
+    [[ "$_ident" =~ ^[A-Za-z0-9_]+$ ]] || { echo "ERROR: --database/--old-table/--new-table must be ClickHouse identifiers (letters, digits, underscore): '$_ident'" >&2; exit 2; }
+done
+[[ "$OLD_TABLE" != "$NEW_TABLE" ]] || { echo "ERROR: --old-table and --new-table must differ; '$OLD_TABLE' would copy a table onto itself." >&2; exit 2; }
+
 sql="${sql//'${ANALYTICS_DB_DATABASE_NAME}'/$DATABASE}"
+sql="${sql//'${OLD_TABLE}'/$OLD_TABLE}"
```
andrescrz: **[medium] Placeholder inventory in the reference SQL (and the runbook) not updated for the two new placeholders**

Two placeholders were added to the driver/SQL contract, but the checklist that exists to prevent exactly the previous finding was not updated: 000002_delta_and_deletion_replay.sql step 0 still says the driver 'handles all five placeholders below, listed so a new one is never missed here' and lists only ${ANALYTICS_DB_DATABASE_NAME}, ${BACKFILL_START}, ${MAX_INSERT_BLOCK_SIZE}, ${MAX_PARTITIONS_PER_INSERT_BLOCK}, ${MAX_INSERT_THREADS} — there are now seven. Its hand-written batched-passes instructions (000002 lines 40-57) and its example invocation (line 11) also predate the flags. Likewise README.md, the operator's source of truth, documents the post-EXCHANGE invocation for verify.sh (`--old-table traces_pre_cutover_backup --new-table traces`, README line ~1439) but says nothing about the new delta_replay flags or when a post-swap catch-up delta is legitimate; the only description of the new capability lives in this script's header, where an operator mid-cutover will not look. The rule explicitly covers placeholder/flag counts and cross-file references.

💡 Update 000002's step-0 placeholder list to seven (naming ${OLD_TABLE}/${NEW_TABLE}) and add the post-swap invocation plus its preconditions to README.md next to the existing verify.sh post-EXCHANGE example.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/delta_replay.sh
```diff
@@ -22,6 +22,22 @@
 #                             to localhost. User/password still come from CLICKHOUSE_USER / CLICKHOUSE_PASSWORD (keeping
 #                             the password out of argv).
 #   --backfill-start TS        the anchor printed by backfill.sh ("RECORD backfill_start=..."). Required.
+#   --old-table NAME           old-schema source to read from. Default traces. AFTER THE EXCHANGE the names swap, so a
+#                             post-swap catch-up is `--old-table traces_pre_cutover_backup --new-table traces`. Without
```
andrescrz: **[medium] Documented post-swap invocation half-applies once the Distributed wrap exists (DELETE unsupported on Distributed)**

The new header text recommends `--old-table traces_pre_cutover_backup --new-table traces` for a post-EXCHANGE catch-up, unconditionally. That only holds while `traces` is still a MergeTree. Once the wrap from 000003_exchange_and_wrap.sql has been applied (`--with-wrap`, or the deferred `--wrap-only` run), `traces` is a Distributed table, and 000003 lines 30-32 state the hard prerequisite: a lightweight DELETE against a Distributed table returns 'DELETE query is not supported' (code 36). Because this driver sends both statements in a single `--multiquery` (line 210), the delta INSERT would be accepted and applied (Distributed supports INSERT) and then the deletion replay would abort — leaving a half-applied catch-up on the live successor, with the operator's only signal being a non-zero exit. Post-wrap the deletable object is `traces_local`, which the script has no way of knowing about; unlike exchange_and_wrap.sh/rollback.sh/finalize.sh, delta_replay.sh performs no topology detection.

💡 State the precondition in the header ('post-EXCHANGE but pre-wrap; after the wrap pass --new-table traces_local, since a Distributed table rejects mutations'), and/or add a pre-flight engine check on $NEW_TABLE (`SELECT engine FROM system.tables`) that refuses to run when the destination is Distributed — the same `traces_engine`-style probe the sibling drivers already use.

_rule: `backend-architecture-gate-every-migration-cutover-step-on-the-full-precondition-c`_

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/delta_replay.sh
```diff
@@ -86,7 +107,16 @@ echo "Reminder: raise databaseAnalytics.asyncInsertBusyTimeoutMaxMs before this
 echo "restore it after the EXCHANGE."
 
 sql="$(cat "$SQL_FILE")"
+# Interpolated straight into the reference SQL, so require plain ClickHouse identifiers — same
+# guard verify.sh applies to the same three values.
+for _ident in "$DATABASE" "$OLD_TABLE" "$NEW_TABLE"; do
+    [[ "$_ident" =~ ^[A-Za-z0-9_]+$ ]] || { echo "ERROR: --database/--old-table/--new-table must be ClickHouse identifiers (letters, digits, underscore): '$_ident'" >&2; exit 2; }
+done
+[[ "$OLD_TABLE" != "$NEW_TABLE" ]] || { echo "ERROR: --old-table and --new-table must differ; '$OLD_TABLE' would copy a table onto itself." >&2; exit 2; }
```
andrescrz: **[medium] Table arguments are only shape-checked, so a transposed pair can write successor rows into the parked rollback backup**

The only semantic guard on the new flags is that they differ (line 107). Nothing checks that the tables exist or that --old-table really is the old-schema side. A transposed post-swap invocation (`--old-table traces --new-table traces_pre_cutover_backup`) passes every guard and INSERTs new-schema successor rows into the parked pre-cutover backup: the column names match, and the SELECT's `coalesce(end_time, ...)` / `coalesce(ttft, ...)` are no-ops on already-non-nullable columns, so the statement succeeds and contaminates the one object the rollback path depends on (rollback.sh:441-462 requires a clean traces_pre_cutover_backup). Sibling drivers guard exactly this class of operator error with topology probes (exchange_and_wrap.sh:236, rollback.sh:332, finalize.sh:109-123); this one has none, and it is now the only script in the set that can write to an operator-named table.

💡 Add a pre-flight probe: both tables must exist, $OLD_TABLE must have a Nullable end_time (old shape) and $NEW_TABLE must not (successor shape) — e.g. read `is_nullable`/`type` for end_time from system.columns and abort on mismatch. This costs one metadata query and makes a transposed pair fail before any data moves.

_rule: `backend-architecture-gate-every-migration-cutover-step-on-the-full-precondition-c`_

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/delta_replay.sh
```diff
@@ -86,7 +107,16 @@ echo "Reminder: raise databaseAnalytics.asyncInsertBusyTimeoutMaxMs before this
 echo "restore it after the EXCHANGE."
 
 sql="$(cat "$SQL_FILE")"
+# Interpolated straight into the reference SQL, so require plain ClickHouse identifiers — same
+# guard verify.sh applies to the same three values.
+for _ident in "$DATABASE" "$OLD_TABLE" "$NEW_TABLE"; do
```
andrescrz: **[low] $DATABASE is now validated twice, and the two new args are validated far from the rest of the argument checks**

Line 85 already validates $DATABASE against the identical `^[A-Za-z0-9_]+$` regex; the new loop at 104-106 validates it a second time, and the comment at line 84 ('--database and --backfill-start are interpolated into the reference SQL; validate their shapes so neither can alter it') is now inaccurate since two more interpolated values exist. The new checks also sit after the whole argument-validation block (83-96), after the `[[ -f "$SQL_FILE" ]]` check and after the operator-facing 'Reminder: raise ... asyncInsertBusyTimeoutMaxMs' echo (98-99), so a typo'd --old-table prints an action reminder and only then fails. verify.sh:90-92 keeps the same loop as the single DATABASE check, without a duplicate.

💡 Move the identifier loop up to replace line 85 (as verify.sh does) so all argument validation stays in one fail-fast block before the operator reminder, and update the line-84 comment to name --old-table/--new-table.

_rule: `backend-api-contract-validate-input-at-the-earliest-api-boundary-not-deep-inside-`_

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/delta_replay.sh
```diff
@@ -181,6 +211,45 @@ if grep -qF '${MAX_INSERT_THREADS}' <<<"$mit_masked"; then
...
+# the DELETE keeps failing until the destination is corrected. Positive match on the MergeTree family, so every
+# other non-mutable engine (Distributed, Buffer, Merge, Null, a view) is refused too.
+dest_engine="$(clickhouse-client ${CH_HOST:+--host $CH_HOST} ${CH_PORT:+--port $CH_PORT} --database "$DATABASE" --log_comment 'traces_local_v2_cutover:delta_replay:destination_engine' --query "SELECT engine FROM system.tables WHERE database = '$DATABASE' AND name = '$NEW_TABLE'")"
+if [[ -z "$dest_engine" ]]; then
+    echo "ERROR: destination '$DATABASE.$NEW_TABLE' does not exist on this instance. Pass --new-table for whatever" >&2
+    echo "       holds the new-schema data here: pre-EXCHANGE 'traces_local_v2', post-EXCHANGE 'traces'," >&2
+    echo "       post-wrap 'traces_local'." >&2
+    exit 1
+fi
+if [[ "$dest_engine" != *MergeTree ]]; then
```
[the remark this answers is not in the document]
thiagohora: **Valid in substance.** Resolved in 634ea39a9d — by removing the capability rather than by validating harder.

You are right that the destination pre-flight only matched the engine suffix, so a reversed or arbitrary MergeTree-family pair passed it. The concrete damage cases both check out: a transposed pair writes new-schema successor rows into the parked rollback backup (the `coalesce` projections are no-ops on already-non-nullable columns, so the statement succeeds silently), and the epoch/NaN sentinel projections applied in the wrong direction leave sentinels where `NULL` belongs. @andrescrz reached the same conclusion independently in 3903377716.

On the remedy: validating "both authoritative schemas, roles, and documented topology" would have been the patch-around. The deeper problem the human review surfaced is that the post-swap pairing is not merely under-validated but semantically wrong — copying backup→live resurrects rows the user deleted after the swap, and the resurrection guard consults a frozen snapshot rather than the live successor. No amount of topology validation makes a correct invocation out of it. So `--old-table`/`--new-table` are withdrawn, `000002_delta_and_deletion_replay.sql` is back to its pinned bytes, and the driver can no longer be pointed at an operator-named table at all. A fixed-destination pre-flight (exists, is a MergeTree) is retained, which also catches running the step out of sequence.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/exchange_and_wrap.sh
```diff
@@ -309,7 +332,10 @@ run_final_deletion_replay() {
     local sql
     sql="$(awk -v begin="-- >>> BEGIN deletion-replay" -v end="-- >>> END deletion-replay" '$0 == begin {f = 1; next} $0 == end {f = 0} f' "$DELTA_SQL_FILE")"
     sql="${sql//'${ANALYTICS_DB_DATABASE_NAME}'/$DATABASE}"
+    sql="${sql//'${OLD_TABLE}'/$DELTA_OLD_TABLE}"
+    sql="${sql//'${NEW_TABLE}'/$DELTA_NEW_TABLE}"
     sql="${sql//'${BACKFILL_START}'/$BACKFILL_START}"
+    assert_no_placeholder "$sql" "deletion-replay"
```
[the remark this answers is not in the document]
thiagohora: **Valid** — and it is the same defect @andrescrz raised at higher severity in 3903377676, reached from the other driver. Resolved in 634ea39a9d.

Confirmed: with `OLD_TABLE=traces_pre_cutover_backup` the guard's `NOT IN` arm is evaluated against a frozen snapshot, so a key re-created on the live successor after the swap is absent there, satisfies `NOT IN`, and gets masked on the live table. Trace ids are client-supplied, which is the guard's own stated premise (000002 L127), so re-creation under the same id is a real path, not a hypothetical.

To your question — "consult the live destination for liveness, or distinguish post-swap recreations" — I looked at both and neither is sufficient on its own. Pointing the guard at the live destination fixes this finding but not the companion one: the delta-insert runs first and resurrects rows the user deleted post-swap, and the guard then sees them as live *because this run just made them live*. Distinguishing post-swap recreations needs an anchor 000002 does not have (`cutover_start`) plus a discriminator that survives the batch-ingest path setting `last_updated_at` client-side. Both belong in a purpose-built post-swap statement pair with its own gate cases, not in a re-parameterised 000002 — whose bytes are in any case frozen for the cutover currently running. Details in my reply on 3903377661.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000002_delta_and_deletion_replay.sql
thiagohora: **Valid — and this one decided the outcome. The mode is withdrawn rather than patched.** Fixed in 634ea39a9d.

Verified every step of your mechanism against the code:

- `last_updated_at` is the ReplacingMergeTree version column on both sides — `traces` at `000017_change_tables_to_replicated.sql:223`, `traces_local_v2` at `000101_create_traces_local_v2_table.sql:97-100`.
- A lightweight DELETE does not bump it. That is not incidental, it is the stated reason the bridge exists: `000096_create_deletion_events_local.sql` — *"A lightweight DELETE does not change a row's version column, so deletes issued while a table is being copied into a new layout are invisible to the version-based delta step and **would reappear in the copy**."* Your finding is that sentence, applied to the post-swap direction.
- The re-copied row lands with `_row_exists = 1`, so it is live regardless of dedup, and the deletion replay in the same run cannot undo it — the id is live on the read side, so the guard at L163 spares it. Confirmed.

**Your suggestion (a) does not close it either, which is why I did not take it.** With a separate `${LIVE_TABLE}` bound to the live successor, the delta-insert still runs first and resurrects the row; the replay then evaluates liveness against the live table, finds the row live *because this run just made it live*, and spares it. The guard cannot distinguish "live" from "live because I resurrected it one statement ago".

**Why no naming scheme fixes it.** Post-swap the copy source, the liveness oracle and the destination are three different tables, and the replay additionally needs an upper bound at `cutover_start` — bridge events after it are post-swap user activity, not deferred work, and 000002 has no upper bound at all. A correct post-swap pair needs roughly: (1) the INSERT filtered by `NOT IN (deletion events since cutover_start)`, and (2) the DELETE windowed to `[backfill_start, cutover_start)`, guarded on liveness in the backup, **and** additionally sparing any successor row not written since the swap — `created_at < cutover_start AND last_updated_at < cutover_start`, the negation of the delta's own OR-anchor, since `last_updated_at` alone is client-supplied on the batch-ingest path (000002 L29-30) and not a trustworthy "changed since" signal. That is a new statement pair with a new anchor and its own gate cases, not a re-pointing of 000002.

**And it could not be written down here even if it were correct.** The whole file is the rendered payload, and its bytes are frozen for the in-flight cutover, so the hazard cannot be documented in the file that the gate test (L2) is mandated to mirror. Your next finding on the false rollback note is the same constraint from the other side.

So: 000002 is reverted to the pinned bytes (`git diff` against `7e49662c` is empty — byte-identical as a file, not merely under the default render), the flags are gone, and `delta_replay.sh`'s header now records the hazard and the reasoning. Worth noting the runbook never expected this mode: README L430-433 covers gap **writes** with the async-insert buffer and gap **deletes** with `exchange_and_wrap.sh`'s final deletion replay. If a real run stranded ~60k rows and ~11k deletions, the fix belongs in whichever of those two under-delivered.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/delta_replay.sh
thiagohora: **Valid, and it was a live break** — thank you, this was the worst of the set: `set -euo pipefail`, `cutover_start` already captured, buffer holding writes, final replay dead, EXCHANGE never reached.

It was already fixed in 1ff4f7305f (before your review landed), by binding `DELTA_OLD_TABLE`/`DELTA_NEW_TABLE` as constants in `run_final_deletion_replay` — reproduced first by driving the pushed HEAD against a capturing stub client and watching both literals reach it.

As of 634ea39a9d it is moot: the placeholders are gone from 000002, so those constants are gone too. What I kept is the generalisation, since a hand-maintained inventory in a file two drivers render is exactly the thing that drifts — `assert_no_placeholder()` in `exchange_and_wrap.sh` (on both `run_block` and `run_final_deletion_replay`) and the equivalent comment-masked guard in `delta_replay.sh`. Both exercised: a `${FUTURE_KNOB}` injected into the deletion-replay block is refused before the EXCHANGE, and a `${SOME_NEW_SETTING}` injected into the delta-insert block is refused before any statement is sent — while the same injected file on the pinned driver sends the literal to the client and exits 0.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000002_delta_and_deletion_replay.sql
thiagohora: **Valid.** Fixed in 634ea39a9d by removing the mode.

Confirmed the asymmetry you describe, and the repo already documents this exact hazard class for the mirror case: `000004_rollback_reverse_replay.sql:12-16` — *"Deliberately NO resurrection guard ... `traces` here is the RESTORED ORIGINAL, so a bridged id is present as its pre-cutover version; a liveness guard would spare it and thereby UNDO the user's post-cutover delete."* Same failure, other direction: whenever the guard's source is not the currently-written table, the guard is wrong. 000002's guard is correct only in the pre-swap topology it was written for.

Also confirmed the specific loss path: trace ids are client-supplied (L127), so an id bridged as deleted pre-swap and re-created post-swap is absent from the frozen backup, passes `NOT IN`, and is deleted from live `traces`.

I did not take the `${LIVE_TABLE}` route — it would have been byte-neutral in the default render (both render to `traces`), so the invariant would have permitted it, but it fixes only this finding and not the resurrection in 3903377661; see my reply there. Withdrawing the mode fixes both, and restores the stale prose you flag at L128/L132-133 to being accurate, since the literal names are the real names again.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000002_delta_and_deletion_replay.sql
thiagohora: **Valid — and the fact that it was unfixable in place is part of why the mode is gone.** Resolved in 634ea39a9d.

You are right that L3, L39 and the rollback note at L191-193 all assume both statements only ever touch the shadow, and that `${NEW_TABLE}` made that false. The problem was that it could not be corrected: the whole file is the payload the driver sends, so editing any comment changes the rendered bytes, and the default render is pinned byte-for-byte to `7e49662c` for the cutover that is running now. A previous pass verified that counterfactually — changing "all five" to "all seven" in the header produces exactly one diff against the pin, which is the change the invariant forbids.

So the PR was in the position of shipping a destructive new target whose safety note said the opposite and could not be edited. That is not a documentation gap, it is a signal the change is in the wrong file. Reverting 000002 to the pinned bytes makes L191-193 true again — nothing this driver runs touches `traces` before the EXCHANGE — and the recovery advice is no longer stale, because `traces_local_v2` is once again the only thing it can write to. The pre-flight added in `delta_replay.sh` now enforces exactly that: if the shadow is absent (out of sequence, or past the EXCHANGE) it refuses with nothing written.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/db-app-analytics/000002_delta_and_deletion_replay.sql
thiagohora: **Valid, and confirmed independently before acting on it.** Resolved in 634ea39a9d by removing the untested mode.

Checked `TracesLocalV2CutoverTest` (2406 lines, real ClickHouse + Zookeeper via testcontainers):

- `deltaInsert(String)` at L1489 and `replayDeletions(String)` at L1512 take only a timestamp; the destination `traces_local_v2` and the guard source `traces` are string literals in both. Same for `backfillWeek`, `reverseReplay`, `exchangeTables`, `wrapInDistributed` and the rollback helpers.
- Your claim about the only post-swap delta is correct: it is L1183 in `rollbackBackupIsReusableAsTheShadowForARetryWithoutRebackfilling`, and it is preceded by the rollback at L1160 and the explicit `RENAME TABLE traces_post_rollback_backup TO traces_local_v2` at L1171, which restores the canonical pre-swap naming first. `replayDeletions` is never called after any `exchangeTables()`.
- No case seeds a post-swap re-created id. `deleteThenResurrectSurvivesTheReplay` (L1309) does its delete-and-recreate entirely pre-EXCHANGE and never swaps; the retry test's post-cutover rows (L1152) are fresh ids. So the resurrection guard has never been exercised with the source named `traces_pre_cutover_backup`.

You are also right that the Javadoc's deferral (L104-109) does not cover this — it defers *driver guards* to the staging dry-run, and this is SQL semantics, which the gate owns.

The two branches you specify are the right acceptance criteria, so I have carried them forward as the gate for any future post-swap step rather than dropping them: (a) a row deleted on the live successor post-swap stays deleted, (b) an id bridged as deleted pre-swap and re-created post-swap stays alive. Both fail against the naive pairing, which is the point.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/delta_replay.sh
thiagohora: **Partly valid, and now moot** — but the underlying complaint is real and I have addressed it structurally rather than by editing the list.

The inventory *was* wrong while the two placeholders existed. It cannot be corrected in this PR, though: the header is inside the rendered bytes, so changing "all five" to "all seven" is itself a breach of the byte-identity invariant against `7e49662c` (verified counterfactually — it is the one and only diff that edit produces). As of 634ea39a9d the placeholders are gone and the file is byte-identical to the pin, so the inventory of five is correct again and nothing needs saying.

What I kept is the mechanism, because a hand-maintained list cannot deliver what the header promises ("listed so a new one is never missed"): both drivers now refuse to send SQL in which any `${...}` survives into an executable line. That subsumes the inventory and, unlike it, also covers `exchange_and_wrap.sh` — which renders the same file's deletion-replay block and which the inventory never mentioned at all (see 3903377671).

On the README: no update is needed now, since no flags were added. Worth flagging what I found while checking, though — README L1440 and L1491 are `verify.sh`'s `--old-table`/`--new-table`, and there is no options block for `delta_replay.sh` at all. More to the point, L430-433 already states the runbook's coverage model for the gap this PR was trying to fill: writes are carried by the async-insert buffer, deletes by `exchange_and_wrap.sh`'s final deletion replay. The post-swap catch-up was a third mechanism overlapping both, documented only in a script header.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/delta_replay.sh
thiagohora: **Valid.** Addressed in two steps.

1ff4f7305f took it beyond documentation: rather than only stating the precondition, `delta_replay.sh` resolved the destination's engine from `system.tables` before writing anything and refused a non-MergeTree destination, with a positive match on the family so `Buffer`/`Merge`/`Null`/views are refused too, naming `traces_local` in the Distributed case.

634ea39a9d then withdraws the post-swap invocation your finding is about, so the half-applied scenario is unreachable by construction — the delta INSERT can no longer be aimed at a Distributed `traces`.

I kept the engine probe anyway, retargeted at the fixed shadow, because it earns its place independently: it now also gates *sequence*. `traces_local_v2` exists only between step 1 and step 3, so running the delta before the backfill or after the EXCHANGE is refused by name with nothing written, instead of failing on a bare `UNKNOWN_TABLE` mid-statement after the operator reminder has already printed. Both arms exercised against a stub: absent destination refused, `Distributed` destination refused, zero statements sent in each case.

You were also right that this driver was the only one in the set with no topology detection while being able to write to an operator-named table. It no longer can.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/delta_replay.sh
thiagohora: **Valid, and eliminated by construction** in 634ea39a9d rather than by adding the probe.

Your analysis of why the transposition is silent holds: the column lists match, and `coalesce(end_time, ...)` / `coalesce(ttft, ...)` are no-ops on already-non-nullable columns, so `--old-table traces --new-table traces_pre_cutover_backup` would have succeeded and contaminated the one object the rollback path requires clean (`rollback.sh:441-462`). Shape-checking that the two names differ was the only semantic guard, which is no guard at all.

With the flags removed, the destination is a literal in the reference SQL and this driver can no longer be pointed at any operator-named table — so the `is_nullable`/`type` probe you suggest has nothing left to protect. I did keep a metadata pre-flight on the fixed destination (exists, and is a MergeTree), which costs the same one query and additionally catches an out-of-sequence run.

For the record, this is the finding that most clearly marked the direction as wrong: the flags were justified in the header by "verify.sh has carried these two flags from the start", but `verify.sh` only SELECTs. Copying an interface across the read/write boundary is what created a driver that could write to a table an operator named.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/delta_replay.sh
thiagohora: **Valid, and resolved exactly as you suggested** — though by deletion rather than by moving the loop. 634ea39a9d.

With `--old-table`/`--new-table` gone, the second identifier loop goes with them: `$DATABASE` is validated once again, in the single fail-fast block, before the `[[ -f "$SQL_FILE" ]]` check and before the "Reminder: raise ... asyncInsertBusyTimeoutMaxMs" echo — the ordering `verify.sh:90-92` has. The L84 comment ("--database and --backfill-start are interpolated into the reference SQL") is accurate again too, since those are once more the only two interpolated values.

Good catch on the operator-facing consequence rather than just the duplication: a typo'd flag printing an action reminder and only then failing is the kind of thing that costs a minute at the worst moment.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/exchange_and_wrap.sh
```diff
@@ -310,6 +326,7 @@ run_final_deletion_replay() {
     sql="$(awk -v begin="-- >>> BEGIN deletion-replay" -v end="-- >>> END deletion-replay" '$0 == begin {f = 1; next} $0 == end {f = 0} f' "$DELTA_SQL_FILE")"
```
andrescrz, line 326: **[medium] No non-empty assertion on the extracted block — marker drift silently skips the final deletion replay**

`assert_no_placeholder` (line 329) defends against one form of drift in `000002_delta_and_deletion_replay.sql` — a `${...}` with no matching substitution — but leaves the adjacent, higher-impact form uncovered: the `-- >>> BEGIN deletion-replay` / `-- >>> END deletion-replay` markers themselves.

If those markers are renamed, moved or reformatted, the `awk` on this line yields an **empty string**. Empty passes the placeholder grep (there is nothing to match), and is then handed to `clickhouse-client --multiquery --query ""` — a no-op the client accepts with exit 0. The run continues to `run_block exchange` (line 357) with the final deletion replay **never executed**, so deletes bridged since the last `delta_replay.sh` leak live across the swap. That is precisely the loss this step exists to prevent, and which the comment at lines 318-323 notes neither the forward replay nor the rollback reverse-replay covers.

The same hole exists in `run_block`: an empty `exchange` block would still print `EXCHANGE done` (line 358) and, under `--with-wrap`, `run_block wrap` (line 361) would wrap a table that was never swapped. Nothing downstream detects it — `assert_pre_exchange_topology` has already run.

A one-line `[[ -n "$sql" ]]` next to each `assert_no_placeholder` closes both.

_rule: `backend-error-handling-destructive-migrations-and-cutovers-need-a-real-rollback-ver`_

andrescrz (commented):
## Opik reviewer — follow-up on `634ea39a`

The withdrawal commit resolves the two `[high]` findings from the earlier review by **removing the mode**, which is the stronger fix — nothing to answer there. For the record, the earlier comments tied to the withdrawn flags are now moot: the `exchange_and_wrap.sh` substitution gap, the post-wrap Distributed `DELETE` half-apply, the placeholder-inventory drift, the false rollback note, and the transposed-pair hazard all dissolve with the revert to pinned bytes.

**3 findings — 3 medium.** All three concern the two guards this PR *keeps*, so they outlive the withdrawal. The `assert_no_placeholder` in `exchange_and_wrap.sh` is the common thread: it is the one of three implementations that does not mask comments, it fires per-block at send time rather than up front, and it treats an empty extraction as success.

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/exchange_and_wrap.sh
```diff
@@ -292,6 +307,7 @@ run_block() {
     local sql
     sql="$(extract "$1")"
     sql="${sql//'${ANALYTICS_DB_DATABASE_NAME}'/$DATABASE}"
+    assert_no_placeholder "$sql" "$1"
```
andrescrz, line 310: **[medium] Render-integrity check for the wrap block fires only after the EXCHANGE has committed**

`assert_no_placeholder` is called from inside `run_block`, so each block is validated at the moment it is about to be sent rather than up front. On the `--with-wrap` path the call order is `run_final_deletion_replay` (355) → `run_block exchange` (357) → `run_block wrap` (361), so a placeholder in the `wrap` block is detected only **after** the EXCHANGE and its post-swap RENAME have been applied. The run aborts with `traces` swapped and unwrapped, leaving re-entry under `--wrap-only --confirm-maintenance` — a second maintenance window.

This contradicts the file's own stated discipline: the gates at lines 125, 136 and 147 are all explicitly "Fail fast, before touching ClickHouse". It also contradicts the counterpart added to `delta_replay.sh` in this same PR, which is explicitly a pre-flight *before the first write*.

A rendering defect is knowable with zero server round-trips, so nothing forces it to surface mid-cutover — rendering and asserting both blocks up front, before the first statement, costs nothing.

_rule: `backend-architecture-gate-every-migration-cutover-step-on-the-full-precondition-c`_

apps/opik-backend/data-migrations/traces-local-v2-cutover/scripts/exchange_and_wrap.sh
```diff
@@ -283,6 +283,21 @@ else
     assert_replication_settled
 fi
 
+# No placeholder may survive into the SQL this driver sends. Both blocks it renders live in files it does not own —
+# 000003 here, and the deletion-replay block of 000002, which it shares with delta_replay.sh — so a ${...} added
+# there without a matching substitution added HERE reaches the server as a literal and fails mid-cutover. The
+# reference SQL's own header keeps a hand-maintained placeholder inventory; a hand-maintained list drifts, so this
+# checks the rendered text instead of trusting it. $2 names the block, for diagnostics.
+assert_no_placeholder() {
+    if grep -qE '\$\{[A-Za-z_][A-Za-z0-9_]*\}' <<<"$1"; then
```
andrescrz, line 292: **[medium] assert_no_placeholder matches `${...}` inside SQL comments, unlike both sibling guards**

This grep runs against the raw rendered text, so a `${...}` written in a `--` or `/* */` comment **inside an extracted block** counts as a surviving placeholder and aborts the run.

Both established implementations of exactly this post-render check deliberately mask comments first, and say why:
- `delta_replay.sh:222` — added in this same PR — renders through `mit_mask_comments`, its comment stating *"Comment-masked, because the header necessarily writes `${...}` in prose"*.
- `backfill.sh:426` — same treatment.

So the two drivers that share the `000002` deletion-replay block now disagree on how to validate it. The exposure is not hypothetical in shape: the `wrap` block of `000003_exchange_and_wrap.sql` already carries a placeholder inside a comment (the `DROP TABLE IF EXISTS ...traces_dist` recovery hint); it passes today only because that particular placeholder happens to be substituted. The first block-internal comment naming a placeholder this driver does not substitute turns a good SQL file into a refused cutover — and for the `wrap` block, refused *after* the EXCHANGE has committed (see the separate finding on guard placement).

Routing this through the same `mit_mask_comments` helper the siblings use would make the three consistent.

_rule: `backend-data-access-verify-clickhouse-query-safety-claims-against-actual-schema-`_

PR #8100, opened 2026-09-01
[OPIK-7791] [QA] Proposed e2e specs from the 2.2.46 → 2.2.47 release exploration

CometActions:
> **Generated by the release QA side flow, and not reviewed by a human.**
> It is a draft on purpose. Please read every assertion before promoting it —
> a plausible-looking spec that asserts the wrong thing is worse than no spec.

## Where these came from

The 2.2.46 → 2.2.47 release exploration. A human worked the release's
manual-verification list against **staging** (`staging.dev.comet.com/opik`) and
verified each of these flows by hand first; only flows that were reached and
observed became candidates. Four specs are proposed here, from three candidates.

Everything in this train traces back to one change: **PR #8096** replaced 56
week-bound predicates across five DAOs. `toMonday()` returns a 16-bit `Date`
that wraps past 2149, so a trace whose UUIDv7 id carries a far-future timestamp
cleared its id bound, folded into a past Monday and **failed the week bound** —
disappearing from reads it belonged in, with no error and no empty state. The
commit measured 12,680,766 such rows across 39,038 projects on production.

The gap the exploration found: the PR's own regression test covers `TraceDAO`
alone. `ThreadDAO`, `KpiCardDAO` and `ProjectMetricsDAO` took the identical
rewrite with **no far-future row exercised anywhere**, and nothing in this
estate drives `last_retrieved_id` at all.

### Which ref these were written and run against

Written and run against the release ref **`2.2.47`** (`dd87fed`), which is the
build the exploration verified. `main` is only where they have to merge — a
release tag is not a branch, so it cannot be a PR base. `main` will have moved
on; please re-run before merging if anything under `tests_end_to_end/` or the
trace read path has changed since.

## Verification — all four specs pass

Run from `tests_end_to_end/e2e/`, against `$OPIK_BASE_URL =
https://staging.dev.comet.com/opik` (workspace `opik-testing`), i.e. the same
environment and credentials the exploration used:

```bash
npx playwright test \
  tests/trace-explore/far-future-trace-window.spec.ts \
  tests/trace-explore/far-future-thread-window.spec.ts \
  tests/trace-explore/trace-stream-cursor.spec.ts \
  tests/dashboards/project-metric-daily-buckets.spec.ts \
  --reporter=list
```

**6 passed.** Nothing is `.skip`ped, and no spec was dropped for being red.

`tag_lint.py` is green too: `58 specs checked, 1 exempt, 0 problem(s)`.

Because this touches the shared `LogsPage` page object and the fixtures chain,
the whole `tests/trace-explore/` + `tests/dashboards/` directories were run as
well (36 tests, 3 workers, `--retries=1`): **31 passed, 5 flaky, 0 failed.** All
six of the specs proposed here passed on the first attempt.

The five flaky ones are **pre-existing specs, and one honest caveat**. Each
failed identically — `opik-sdk-driver POST /traces aborted after 30000ms` in
`conversation.fixture` / `trace.fixture` / an inline seed — and each passed on
retry. That is the SDK bridge's own client-side timeout being hit while shared
staging rate-limits the workspace (`Rate limited (HTTP 429) for 'search_traces',
continuing in 52 seconds` appears throughout the run). Nothing here calls that
bridge route. But spec 4 does seed twelve traces through the bridge, so running
it alongside the rest of the directory plausibly *contributed* to the contention
that surfaced a pre-existing sensitivity — worth knowing before this lands in a
suite that runs everything at once.

> `npx tsc --noEmit` **cannot run at this ref** and does not run here: the
> pinned `typescript@^7.0.2` rejects `tsconfig.json`'s `baseUrl` (`TS5102:
> Option 'baseUrl' has been removed`), which aborts before any file is checked.
> That is pre-existing on `2.2.47` and unrelated to this change. Every file here
> was typechecked with the same compiler against a temporary config with
> `baseUrl` dropped and `types: ["node"]` added: **clean**, apart from a
> pre-existing `TS2300: Duplicate identifier 'deleteDashboard'` in
> `core/backend/client.ts` (two methods of that name at `HEAD`, untouched here).

### 1. `trace-explore/far-future-trace-window.spec.ts` — `@t2-cuj @area:traces`

**Capability:** `@cap:traces.list-traces`

Seeds four ordinary traces and one whose UUIDv7 id is stamped `2200-06-15`, then
asserts the *whole* row set returned by each window, in both directions:

| read | expected |
|---|---|
| `GET /traces`, no window | all five |
| `from_time=now-7d & to_time=2201` | all five ← **the regression** |
| `from_time=now-7d & to_time=now` | the four ordinary rows, and only those |
| `from_time=2200 & to_time=2201` | the far-future row alone |
| `POST /projects/{id}/kpi-cards` (traces) | 1 over 2200→2201, 4 over the past week |

Then a UI test: the Logs **Traces** table renders all five rows.

Every assertion names the whole expected set rather than looking for the
far-future row in the answer — a read that ignored its window entirely would
also contain it, so "it's in there" proves nothing.

### 2. `trace-explore/far-future-thread-window.spec.ts` — `@t2-cuj @area:threads`

**Capability:** `@cap:threads.list-threads`

The `ThreadDAO` half, which has **no upstream regression test at all**. Same
fixture, read through `GET /traces/threads`: a window spanning 2200 lists both
threads, a window closing now lists only the present-day one, a window inside
2200 lists only the far-future one — asserting `total` as well as the page body,
since `total` drives the pager independently. Plus the threads `kpi-cards`
branch, and a UI test over the Threads tab.

Split from spec 1 because a spec belongs to exactly one `@area:` and these
assertions are threads; both drive one fixture.

### 3. `trace-explore/trace-stream-cursor.spec.ts` — `@t2-cuj @area:traces`

**Capability:** `@cap:traces.list-traces`

The failure the commit message singles out as the one a lower-bound-only fix
would miss. `last_retrieved_id` is a cursor lifted from a row the stream just
returned; paired with a wrapped upper bound, a far-future cursor folds into a
past week, every ordinary row fails `<=`, **the page comes back empty and
pagination stops silently**. The only upstream guard runs against the legacy
pre-cutover table.

Four cursors, whole-stream comparisons in order: none, the far-future row (→ the
four ordinary rows — before the fix, `[]`), the newest ordinary row (the
control), and the oldest (→ `[]`, which is what makes an empty page wrong in the
second case).

API-level with no browser: this is a cursor on a streaming read no page exposes.

### 4. `dashboards/project-metric-daily-buckets.spec.ts` — `@t2-cuj @area:dashboards`

**Capability:** `@cap:dashboards.configure-widget`

`ProjectMetricsDAO`'s `TRACE_FILTERED_PREFIX` — the CTE behind every trace-based
project metric — plus `GET_AVERAGE_DURATION` and `GET_TOTAL_TRACE_ERRORS` were
rewritten. 426 lines of SQL. The estate's existing cover,
`project-span-metrics.spec.ts`, asserts HTTP 200 across the 20 metric types:
a **binding** guard, not a correctness one. It would pass unchanged if every
bucket came back a day late or carrying the whole window's total.

So this pins values. One scored trace per day for eleven consecutive days, each
score equal to its own age — distinct values a collapsed or shifted bucketing
cannot satisfy by accident — plus one errored trace, which makes the daily trace
counts uneven for the same reason. Asserts each `FEEDBACK_SCORES` and
`TRACE_COUNT` daily bucket, that **no** bucket outside the seeded days carries
anything, that a window closing before the seed aggregates to zero, and that
`/projects/stats` over a three-day sub-window reports the same numbers. Eleven
consecutive days always contain a Monday; the fixture asserts that rather than
trusting it.

Runtime ~3.1 min per test (twelve SDK seeds, rate-limited on a shared cloud
workspace), inside the `test.slow()` budget the neighbouring spec also uses.

## Things a reviewer should push back on if they disagree

- **Candidate 4 was marked `surface: both`; spec 4 is API-only.** A chart plots
  points as SVG geometry, so a seeded bucket value cannot be read back off it —
  `project-span-metrics.spec.ts` says so in its own comments — and that spec
  already drives the project-scoped Time series widget dialog under this same
  capability. A UI leg here would have added flake and no coverage. Said plainly
  rather than quietly narrowed; happy to add one if you want the widget driven
  again.
- **`kpi-cards` is asserted but claims no `@cap:`.** The Logs metrics-summary
  strip has no capability in `taxonomy.yaml` — it appears under neither `traces`
  nor `threads`. Rather than stretch the nearest key with an explanatory note,
  the reads are asserted and nothing is claimed for them. If a capability is
  added later, these specs already cover it.
- **No capability was flipped `false → true`.** All three were already
  `covered: true`; the taxonomy change adds the four specs to their areas'
  `specs:` lists and extends the notes to say what the new dimension is.

## What was deliberately *not* written

Three candidates the exploration proposed were dropped, and one it had already
excluded is worth repeating here:

1. **Logs KPI strip vs. the table for the same range.** The sharpest finding of
   the run, and **wrong today**: the strip says Traces 12 while the table says
   13 (1 vs 3 on Threads). One range, two requests, two denominators — the list
   sends no `to_time` and gets an open upper bound; `kpi-cards` sends
   `interval_end`. A spec asserting the correct behaviour would land red, and
   there is no capability to tag it with. Spec 1 deliberately does **not**
   assert `countTraces()`, and says why in a comment, so nobody closes the gap
   by pinning an unsettled contract.
2. **A far-future trace is bucketed at its own date in project metrics.** A
   trace at `2200-06-15` lands in a `2064-05-07` daily bucket and a `2021-01-02`
   weekly one — the 2³²-second `DateTime` wrap. A defect, not a coverable flow;
   a spec asserting the honest bucket would fail today. Spec 4's "no bucket
   outside the requested interval" holds for ordinary data and will start
   failing if a far-future row ever leaks into a normal window.
3. **Workspace feedback-score summary / daily endpoints.** Both return nothing
   on staging for every window tried, so no passing test can be written against
   them. Both are `@Deprecated` with no consumer in the 2.2.47 frontend, and
   neither has a taxonomy capability.
4. **Helm chart and npm artifact checks.** Not reachable from an e2e suite.

No candidate was dropped for being red or for failing to run.

## Supporting changes, and why they are here rather than in the tests

- `fixtures/far-future-traces.fixture.ts`, `fixtures/project-metric-days.fixture.ts`
  — seeding and teardown, per `conventions.md`. Both delete their traces
  explicitly (a project delete does not cascade to traces) and honour
  `shouldLeaveArtifacts`. `far-future-traces` also **proves it can
  discriminate** before any spec reads: it waits for every seeded row to be
  queryable, then asserts the far-future row really is outside a window ending
  now — a backend that clamped the id on write would otherwise leave every
  assertion in specs 1–3 trivially true.
- `pom/logs.page.ts` — a `gotoTraces()` counterpart to the existing
  `gotoThreads()`. `goto()` leaves the tab to the page, and `useLogsType`'s
  default is data-dependent: it opens on **Threads** whenever the project has a
  thread in range. Every trace in this fixture carries one, so spec 1's UI test
  landed on the wrong table until it asked for Traces explicitly. Purely
  additive; no existing caller changes.
- `core/backend/client.ts` — `projectKpiCards()` and `searchTraceIds()` (the
  stream answers `application/octet-stream` NDJSON, which neither the pinned SDK
  nor `rawFetch` can read, and the paged `GET /traces` has no cursor), plus an
  optional `threadId` on the existing `createTraceWithSource()`.

### DIFF
--- a/tests_end_to_end/coverage/taxonomy.yaml
+++ b/tests_end_to_end/coverage/taxonomy.yaml
@@ -249,10 +249,12 @@ areas:
       - trace-explore/trace-delete.spec.ts
       - trace-explore/trace-filters.spec.ts
       - trace-explore/trace-attachments.spec.ts
+      - trace-explore/far-future-trace-window.spec.ts
+      - trace-explore/trace-stream-cursor.spec.ts
 
     # ---- functional dimension ----
     capabilities:
-      list-traces:            { covered: true,  tier: t1-smoke }
+      list-traces:            { covered: true,  tier: t1-smoke, note: "also the time-windowed reads: which rows a from_time/to_time window is entitled to (incl. a far-future UUIDv7 id), and the traces/search stream paged with last_retrieved_id" }
       trace-ordering:         { covered: true,  tier: t1-smoke }
       open-trace-panel:       { covered: true,  tier: t1-smoke }
       span-tree-expand:       { covered: true,  tier: t2-cuj }
@@ -319,8 +321,9 @@ areas:
       - trace-explore/thread-logging-smoke.spec.ts
       - trace-explore/thread-id-prefilter.spec.ts
       - trace-explore/thread-evaluation.spec.ts
+      - trace-explore/far-future-thread-window.spec.ts
     capabilities:
-      list-threads:            { covered: true,  tier: t1-smoke }
+      list-threads:            { covered: true,  tier: t1-smoke, note: "also the time-windowed read: which threads a from_time/to_time window is entitled to, incl. a thread whose traces carry far-future UUIDv7 ids" }
       thread-message-count:    { covered: true,  tier: t1-smoke }
       open-thread-panel:       { covered: true,  tier: t1-smoke }
       turn-order-io:           { covered: true,  tier: t1-smoke }
@@ -369,6 +372,7 @@ areas:
       - dashboards/workspace-span-metrics.spec.ts
       - dashboards/project-span-metrics.spec.ts
       - dashboards/dashboards-list-filters.spec.ts
+      - dashboards/project-metric-daily-buckets.spec.ts
     capabilities:
       list-dashboards:      { covered: true,  tier: t2-cuj, note: "the list is filtered by the Description column; the other operators and columns are uncovered" }
       create-dashboard:     { covered: true,  tier: t2-cuj, note: "created via the Create dashboard dialog as the UI test's setup; rename/delete/duplicate uncovered" }
@@ -376,7 +380,7 @@ areas:
       edit-dashboard:       { covered: false }
       delete-dashboard:     { covered: false }
       add-widget:           { covered: true,  tier: t2-cuj, note: "Time series widget added through the widget dialog by workspace-span-metrics; other widget types and the config dialog's remaining options are uncovered" }
-      configure-widget:     { covered: true,  tier: t2-cuj, note: "scoping a Time series widget to one project and picking its metric through the widget dialog, plus the data contract of the per-project read it then makes (POST /projects/{id}/metrics across every MetricType and span breakdown); the widget dialog's remaining options — chart type, filters, group by, sub-metric — are uncovered" }
+      configure-widget:     { covered: true,  tier: t2-cuj, note: "scoping a Time series widget to one project and picking its metric through the widget dialog, plus the data contract of the per-project read it then makes (POST /projects/{id}/metrics across every MetricType and span breakdown); the widget dialog's remaining options — chart type, filters, group by, sub-metric — are uncovered. The per-project read's DAILY bucket *values* are pinned too (FEEDBACK_SCORES and TRACE_COUNT over a window crossing a week start, plus /projects/stats over a sub-window); reading those values back off the rendered chart is not possible, so that half stays API-level" }
       widget-filters:       { covered: true,  tier: t2-cuj, note: "data contract (POST /workspaces/metrics/spans against the front end's own filter payload, incl. the 400 path) plus a UI test that builds a workspace-scoped Time series widget and reads its chart" }
       metric-date-range:    { covered: true,  tier: t2-cuj, note: "data contract for the four presets at the interval calculateIntervalType implies, plus the UI date-range control driving a rendered widget" }
 
--- a/tests_end_to_end/coverage/taxonomy.yaml
+++ b/tests_end_to_end/coverage/taxonomy.yaml
@@ -254,7 +254,7 @@ areas:
 
     # ---- functional dimension ----
     capabilities:
-      list-traces:            { covered: true,  tier: t1-smoke, note: "also the time-windowed reads: which rows a from_time/to_time window is entitled to (incl. a far-future UUIDv7 id), and the traces/search stream paged with last_retrieved_id" }
+      list-traces:            { covered: true,  tier: t1-smoke, note: "also the time-windowed reads: which rows a from_time/to_time window includes (incl. a far-future UUIDv7 id), and the traces/search stream paged with last_retrieved_id" }
       trace-ordering:         { covered: true,  tier: t1-smoke }
       open-trace-panel:       { covered: true,  tier: t1-smoke }
       span-tree-expand:       { covered: true,  tier: t2-cuj }
@@ -323,7 +323,7 @@ areas:
       - trace-explore/thread-evaluation.spec.ts
       - trace-explore/far-future-thread-window.spec.ts
     capabilities:
-      list-threads:            { covered: true,  tier: t1-smoke, note: "also the time-windowed read: which threads a from_time/to_time window is entitled to, incl. a thread whose traces carry far-future UUIDv7 ids" }
+      list-threads:            { covered: true,  tier: t1-smoke, note: "also the time-windowed read: which threads a from_time/to_time window includes, incl. a thread whose traces carry far-future UUIDv7 ids" }
       thread-message-count:    { covered: true,  tier: t1-smoke }
       open-thread-panel:       { covered: true,  tier: t1-smoke }
       turn-order-io:           { covered: true,  tier: t1-smoke }
@@ -380,7 +380,7 @@ areas:
       edit-dashboard:       { covered: false }
       delete-dashboard:     { covered: false }
       add-widget:           { covered: true,  tier: t2-cuj, note: "Time series widget added through the widget dialog by workspace-span-metrics; other widget types and the config dialog's remaining options are uncovered" }
-      configure-widget:     { covered: true,  tier: t2-cuj, note: "scoping a Time series widget to one project and picking its metric through the widget dialog, plus the data contract of the per-project read it then makes (POST /projects/{id}/metrics across every MetricType and span breakdown); the widget dialog's remaining options — chart type, filters, group by, sub-metric — are uncovered. The per-project read's DAILY bucket *values* are pinned too (FEEDBACK_SCORES and TRACE_COUNT over a window crossing a week start, plus /projects/stats over a sub-window); reading those values back off the rendered chart is not possible, so that half stays API-level" }
+      configure-widget:     { covered: true,  tier: t2-cuj, note: "scoping a Time series widget to one project and picking its metric through the widget dialog, plus the data contract of the per-project read it then makes (POST /projects/{id}/metrics across every MetricType and span breakdown); the widget dialog's remaining options — chart type, filters, group by, sub-metric — are uncovered. The per-project read's DAILY bucket *values* are pinned too (FEEDBACK_SCORES and TRACE_COUNT over a window crossing a week start); reading those values back off the rendered chart is not possible, so that half stays API-level" }
       widget-filters:       { covered: true,  tier: t2-cuj, note: "data contract (POST /workspaces/metrics/spans against the front end's own filter payload, incl. the 400 path) plus a UI test that builds a workspace-scoped Time series widget and reads its chart" }
       metric-date-range:    { covered: true,  tier: t2-cuj, note: "data contract for the four presets at the interval calculateIntervalType implies, plus the UI date-range control driving a rendered widget" }
 
--- a/tests_end_to_end/e2e/core/backend/client.ts
+++ b/tests_end_to_end/e2e/core/backend/client.ts
@@ -341,6 +341,22 @@ export const PROJECT_METRIC_TYPES = [
 
 export type ProjectMetricType = (typeof PROJECT_METRIC_TYPES)[number];
 
+/** The entity a Logs metrics-summary (KPI) card counts. */
+export type KpiEntityType = 'traces' | 'spans' | 'threads';
+
+/**
+ * One card of a `POST /v1/private/projects/{id}/kpi-cards` answer.
+ *
+ * `null` is a real answer from this endpoint, not a missing one: a card whose
+ * interval matched nothing reports a null value rather than a zero, and the two
+ * must not be collapsed by a caller asserting on a count.
+ */
+export interface KpiStat {
+  type: string;
+  currentValue: number | null;
+  previousValue: number | null;
+}
+
 /** The `breakdown` block a widget sends when the user picks a "Group by". */
 export interface MetricBreakdown {
   /** `model`, `provider`, `type`, `name`, `tags`, … — the backend's BreakdownField. */
@@ -1000,6 +1016,45 @@ export function makeBackendClient(apiKey: string | null = null, workspaceName: s
     },
 
     /**
+     * `POST /v1/private/projects/{id}/kpi-cards` — the counts the Logs page's
+     * metrics-summary strip renders above the Traces/Threads table.
+     *
+     * Same payload the front end sends (`useProjectKpiCards.ts`): an
+     * `entity_type` plus an explicit `interval_start`/`interval_end`. That
+     * interval is the point — the card read is time-bounded where the table's
+     * own list read leaves the upper bound open, so it exercises a different
+     * branch of the same week-bound predicate.
+     */
+    async projectKpiCards(args: {
+      projectId: string;
+      entityType: KpiEntityType;
+      intervalStart: Date;
+      intervalEnd: Date;
+    }): Promise<RawApiResult & { stats: KpiStat[] }> {
+      const { status, message, json } = await rawFetch(
+        'POST',
+        `/v1/private/projects/${args.projectId}/kpi-cards`,
+        {
+          body: {
+            entity_type: args.entityType,
+            interval_start: args.intervalStart.toISOString(),
+            interval_end: args.intervalEnd.toISOString(),
+          },
+        },
+      );
+      const stats = (json as { stats?: Array<Record<string, unknown>> } | null)?.stats ?? [];
+      return {
+        status,
+        message,
+        stats: stats.map((s) => ({
+          type: String(s.type ?? ''),
+          currentValue: (s.current_value as number | null | undefined) ?? null,
+          previousValue: (s.previous_value as number | null | undefined) ?? null,
+        })),
+      };
+    },
+
+    /**
      * `DELETE /v1/private/dashboards/{id}`.
      *
      * A dashboard does not hang off a project, so no project delete reaches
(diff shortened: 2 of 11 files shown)


CometActions:
## 🔍 Generated-test review

# PR #8100 — [OPIK-7791] [QA] Proposed e2e specs from the 2.2.46 → 2.2.47 release exploration

CometActions · draft: yes · 11 files, 4 new specs, 2 new fixtures
Head reviewed: `bca27176aa` · fixes pushed as `d5de6338a6`

## VERDICT: needs changes — do not merge, and do not flip to ready

Two blockers, and they are the same blocker seen from two sides: **the coverage
this PR proposes has already partly landed on `main` as #8097, and the premise
the other half rests on is one #8097 explicitly rejected after checking it.**
The merge conflict is not incidental — it is two proposals for the same fix's
coverage colliding. That has to go to a person before anything else here matters.

I applied the reliability and wording fixes that stand on their own merits (see
"Fixed and pushed"), because they are worth having whichever way the human call
goes. I did **not** resolve the conflict, invent a taxonomy capability, or touch
the draft flag.

---

## Blockers (2)

### 1. This PR duplicates #8097, which is already merged

`d5c53...` — **[OPIK-7456] [QA] Proposed e2e specs from the #8096
(toMonday(id_at)) exploration (#8097)** — merged to `main` after this branch was
cut (merge base `dd87fedd`). Both PRs cover the *same* upstream fix, PR #8096's
`toMonday(id_at)` rewrite. What is already on `main`:

| On `main` (#8097) | Proposed here (#8100) |
|---|---|
| `fixtures/id-aged-traces.fixture.ts` (287 lines) — mints UUIDv7 ids at chosen instants, seeds epoch / present / ~2201 ages, polls for readability, deletes in teardown | `fixtures/far-future-traces.fixture.ts` (177 lines) — mints UUIDv7 ids at chosen instants, seeds 2200 / present ages, polls for readability, deletes in teardown |
| `tests/trace-explore/trace-id-time-window.spec.ts` — windowed trace reads at unusual id ages, + a Logs table step | `tests/trace-explore/far-future-trace-window.spec.ts` — windowed trace reads at a far-future id age, + a Logs table step |
| `traces.list-traces` / `update-trace-api` taxonomy notes | `traces.list-traces` taxonomy note |

`far-future-traces.fixture.ts` is a near-duplicate of a fixture that now exists
on `main` and is strictly more capable than it — `id-aged-traces` already
exposes a `farFuture` ref, and it carries two things this PR's fixture does not:

- a **register-then-`finally`** teardown (I have now backported that pattern —
  see Fixed and pushed);
- a **UUID-validation reject-mode guard**. `UuidV7TimestampValidator` bounds an
  ingested id's embedded timestamp to `[now ± window]` and answers 400 when
  `uuidValidation.enabled=true, auditOnly=false`. #8097's fixture detects that
  rejection and `test.skip()`s with an explanatory reason. This PR's fixture has
  no such handling: on any env running reject mode, `createTraceWithSource` with
  a 2200 id returns an opaque 400 from a seed helper and all four specs fail as
  if the product were broken.

**Fix:** a person decides the shape. The obvious resolution is to rebuild #8100's
distinctive assertions — the thread half, the `traces/search` cursor, the DAILY
bucket values — on top of `id-aged-traces.fixture.ts`, and drop
`far-future-traces.fixture.ts` and `far-future-trace-window.spec.ts` as
already-covered. I did not do that: it deletes a spec and rewrites a fixture
across a conflicted merge, which is a judgement call, not a cleanup.

### 2. The far-future premise contradicts a verified note in the merged sibling

Three of the four specs assert that a trace whose UUIDv7 id is stamped
**2200-06-15** is windowed correctly by the read expression. `id-aged-traces.fixture.ts`,
already on `main`, states the opposite after checking it:

> the far-future age below would also render in a Logs table on its default
> range, **but only because the legacy `traces.id_at DateTime('UTC')` column
> truncates a ~2201 instant back into range — a property of the storage, not of
> the read expression.** A spec that asserted on it would be asserting a schema
> artifact that the `traces_local_v2` cutover changes.

and

> `toMonday` saturates rather than underflowing (ClickHouse 26.3 …) and
> `traces.id_at` is a 32-bit `DateTime` that truncates an out-of-range instant
> before the week expression sees it — so on this schema the pre- and post-fix
> expressions agree at every age seeded here.

`DateTime` tops out at 2106-02-07. If that note is right, then for the 2200 seed:

- `far-future-trace-window.spec.ts:91` — *"A window wholly inside 2200 returns
  the far-future row alone"* sends `from_time=2200-01-01`. The id bound
  (`id >= uuid7(2200-01-01)`) admits the row, but the week bound reads a
  **truncated** `id_at` of ~2106, which is below `toMonday(2200-01-01)`. The row
  is excluded and **the step fails** — not because the fix regressed, but
  because the seed never reached 2200 in storage.
- `far-future-trace-window.spec.ts:146` — *"the Logs table renders the
  far-future trace"* passes, but for the storage-truncation reason #8097 named,
  i.e. it pins a schema artifact due to change at the `traces_local_v2` cutover.

**The fixture's own guard does not catch this**, and its docstring over-claims
that it does. `listTraceIds({ toTime: now })` excludes the far-future id on the
**id** bound alone, so the guard proves the *id* was not clamped and says
nothing about `id_at` — which is the column the week bound actually reads. I
have written that limitation into the fixture rather than leaving the claim
standing, but the guard itself cannot be strengthened from the client: the only
probe for it is the 2200→2201 read, which is the assertion under test.

**What I could not verify:** I have no Opik stack here, so I could not run these
specs or query ClickHouse. Blocker 2 is a documented contradiction between this
PR and a merged sibling, not something I confirmed by execution. **The cheapest
next step is to run the four specs against a local stack** — if step 4 of
`far-future-trace-window.spec.ts` fails, #8097's note is right and this PR's
approach needs replacing rather than fixing.

---

## Merge conflicts (2 files) — do not resolve mechanically

Both are **unions**. Taking either side wholesale drops a merged spec's support.

| Side | What it did |
|---|---|
| `main` | added `trace-explore/trace-id-time-window.spec.ts` and `trace-explore/trace-partial-update-merge.spec.ts` to `traces.specs`; added `traces.update-trace-api`, `messages-tab`, `prompts-tab` |
| #8100 | added `trace-explore/far-future-trace-window.spec.ts` and `trace-explore/trace-stream-cursor.spec.ts` to the same list; extended the `list-traces` note |

**A correct resolution keeps all four spec entries and both note edits** — but
only for whichever specs survive blocker 1. If `far-future-trace-window.spec.ts`
is dropped as duplicate, its line goes with it. The `threads` and `dashboards`
hunks do not conflict.

### `tests_end_to_end/e2e/fixtures/index.ts`

This file re-exports `test` from the **tip of a single linear `.extend()`
chain**, so both sides rewrote line 1 and the conflict is the chain tip:

- `main`: `dashboard-cleanup` → `id-aged-traces` → **`alert`** (tip)
- #8100: `dashboard-cleanup` → **`far-future-traces`** → **`project-metric-days`** (tip)

**A correct resolution re-parents this PR's fixtures onto `main`'s tip** —
`far-future-traces.fixture.ts` must import `test as baseTest` from
`./alert.fixture` instead of `./dashboard-cleanup.fixture`, making the chain
`dashboard-cleanup → id-aged-traces → alert → far-future-traces →
project-metric-days`, with `index.ts` exporting `test`/`expect` from
`project-metric-days.fixture` and keeping **both sides'** type re-export blocks.
Taking either side of line 1 alone silently unregisters a whole family of
fixtures — every spec using `alertFixtures` or `idAgedTraces`, or every spec
using this PR's, would fail to resolve its fixture at runtime.

### Shared files that merge cleanly but still moved under this PR

git will not flag these; they are worth a look before merging:

- **`core/backend/client.ts`** — #8097 added `getTracePayload`, `updateTraceTags`
  and six other methods to the same object literal; #8100 adds `projectKpiCards`,
  `searchTraceIds` and a `threadId` field on `createTraceWithSource`. Different
  regions, so git merges both — but the merged result is untypechecked by either
  PR's CI. Re-run `tsc` after resolving.
- **`pom/logs.page.ts`** — #8100 adds `gotoTraces()`. Purely additive, no
  existing caller changes; blast radius nil.

---

## Does the conflict make the other findings moot?

**Mostly, for the trace and thread specs.** If blockers 1 and 2 land the way the
evidence points, `far-future-trace-window.spec.ts` and
`far-future-traces.fixture.ts` are replaced rather than polished, and the
naming/tagging notes on them go away with them.

**Not for the other two.** `trace-stream-cursor.spec.ts` (the `last_retrieved_id`
cursor) and `project-metric-daily-buckets.spec.ts` (DAILY bucket *values*) cover
surfaces nothing on `main` touches — the estate's only cover for the project
metrics endpoint asserts HTTP 200 across the 20 metric types, which would pass
unchanged if every bucket came back at the wrong date. Those two are the real
new value in this PR and are worth keeping whatever happens to the rest. The
should-fix below applies to them.

---

## Should fix (1)

**`project-metric-daily-buckets.spec.ts:143` tags a Projects-list read as widget
coverage.** The second test asserts `GET /v1/private/projects/stats` — the read
the *Projects list* renders its columns from, as the test's own comment says —
while carrying `@cap:dashboards.configure-widget`. That endpoint has nothing to
do with configuring a dashboard widget. Consequence: real coverage of
`projects.list-projects` is invisible on the map, and the taxonomy note claimed
`/projects/stats` as widget coverage (**I removed that clause** — see below).

I did not restructure, because both fixes are decisions:

- **move it** to `tests/projects/` as `@area:projects` +
  `@cap:projects.list-projects` (a spec carries exactly one `@area:`, so
  retagging in place is not possible) — but note it then overlaps
  `projects-list-stats.spec.ts`, which already covers *"Project stats are scoped
  to the requested window"*; or
- **drop the test** as duplicative of that existing spec.

## Notes (4)

1. **`kpi-cards` has no taxonomy capability, and I did not invent one.** Both
   far-future specs assert `POST /projects/{id}/kpi-cards` — the Logs
   metrics-summary strip — under `@cap:traces.list-traces` /
   `@cap:threads.list-threads`. The specs are candid about this in their own
   headers, which is the honest handling. But the assertion is real coverage
   that the map will not show. **A human chooses:** add a metrics-summary
   capability to `traces` and `threads`, or drop the assertions. Per the
   estate's own rule, adding a capability changes the denominator and is a
   QA-owned decision.
2. **The `tsc` gate is broken estate-wide, not by this PR.** `npx tsc --noEmit`
   fails with `tsconfig.json(13,5): error TS5102: Option 'baseUrl' has been
   removed` — the installed compiler is TypeScript **7.0.2**, which dropped
   `baseUrl`; `tests_end_to_end/e2e/tsconfig.json` has carried it since #6677
   and this PR does not touch it. The gate fails identically on `main`. Two
   options, both a separate PR: pin `typescript` to 5.x, or replace `baseUrl`
   with `"paths": {"*": ["./*"]}`. I did not apply either — which of those the
   estate wants is a toolchain decision.
3. **A pre-existing duplicate `deleteDashboard`.** With a working config, the
   *only* type errors in the whole estate are `client.ts` declaring
   `deleteDashboard` twice in one object literal (HEAD lines 568 and 1064; on
   `main` at 674 and 1243). The second silently wins. Pre-existing, not this
   PR's, and a separate fix — but it will surface the moment note 2 is fixed.
4. **`logs.traceRows` is used to count *thread* rows**
   (`far-future-thread-window.spec.ts:153`). It resolves to the generic
   `tr[data-row-id]`, and `waitForThreadsReady` already falls back to the same
   locator, so this is correct — just reads oddly. Cosmetic.

---

## Fixed and pushed (`d5de6338a6`)

- **Setup-failure trace leak** (baz #3, #7) — both new fixtures assembled their
  id list only after every write succeeded, and cleaned up only after `use()`.
  A rejection from a later seed, the readiness poll or the far-future check
  orphaned every row already written, and neither the project delete nor
  `global-teardown`'s prefix sweep collects traces. Each id is now registered as
  it is written and deleted in a `finally`, preserving `shouldLeaveArtifacts`.
  This is the pattern `id-aged-traces.fixture.ts` already uses on `main`.
- **Midnight/clock flake in `project-metric-days`** (baz #4) — the fixture
  computed `bucketDate` from a runner-captured `now` while the SDK bridge
  subtracts `age_days` from a clock of its own, once per call. With whole-day
  offsets every trace landed within milliseconds of a UTC midnight, so a run
  crossing midnight mid-seed (12 sequential SDK calls) put some traces in the
  neighbouring bucket and failed every per-day expectation. Seeds are now aimed
  at **noon UTC** of their day via a fractional `age_days` (`age_days` is
  `float | None` on the bridge), giving a 12-hour margin, and the window bounds
  are anchored to today's UTC midnight to match.
- **Honest guard docstring in `far-future-traces`** — recorded what the
  "is it really far-future" check does *not* establish (see blocker 2).
- **Note wording** (baz #1) — "which rows a window **is entitled to**" →
  "**includes**", in both taxonomy notes and in the `searchTraceIds` docstring
  that shared the phrasing.
- **False claim in the `configure-widget` note** — dropped
  "plus `/projects/stats` over a sub-window" (see Should fix).

---

## Bot comments

| 1 | "is entitled to" is awkward in the `list-traces` / `list-threads` notes | valid | **fixed** in `d5de6338a6`, plus the same phrase in the `searchTraceIds` docstring |
| 2 | `projectKpiCards` / `searchTraceIds` ignore `makeBackendClient`'s `workspaceName`, using `env.workspace` | valid, **house-wide** | **deferred → estate-wide ticket.** `rawFetch` (line 460) and `workspaceHeaders` (line 486) both predate this PR and hardcode `env.workspace`; every raw-fetch method already inherits it. The seven alternate-workspace call sites are all `workspace-roles` paths that use the typed `opik` client. Fixing it only in the two new methods makes them the inconsistent pair; the fix belongs in `rawFetch`/`workspaceHeaders` in its own PR. |
| 3 | `far-future-traces` setup failures leak traces | valid | **fixed** — register-then-`finally` |
| 4 | `project-metric-days` derives buckets from runner `now` vs the bridge's own clock | valid | **fixed** — noon-UTC anchoring via fractional `age_days` |
| 5 | kpi-cards assertions are untracked coverage; add area-scoped capabilities | valid, **escalated** | **not fixed by design.** Never invent a capability to make a tag validate — it sets the coverage denominator. See note 1; a human chooses add-or-drop. Baz also mis-cites `.agents/skills/writing-e2e-tests/conventions.md`, which has no "Taxonomy Changes" section; `TESTING-TAGS.md` is the rule, and it says the same thing. |
| 6 | Every `test.step()` needs a sentence-sized comment immediately before it | **invalid** | Cites `.agents/skills/playwright-e2e/test-conventions.md`, which does not exist (the skill is `writing-e2e-tests`, and its `test.step` section requires wrapping, not preceding comments). The estate does not follow this: `trace-filters.spec.ts`, a direct neighbour, has no comment before any of its nine steps. |
| 7 | `project-metric-days` setup failures leak traces | valid | **fixed** — same change as #3 |
| 8 | Replace the hand-written poll loop with `expect.poll` | valid, **house-wide** | **deferred → estate-wide ticket.** A bounded hand-rolled poll *is* the fixture convention: `conversation.fixture.ts`, `evaluated-thread.fixture.ts` and `id-aged-traces.fixture.ts` (merged) all do it, and no fixture in the estate uses `expect.poll`. This PR's loop has a bounded deadline and a naming failure message, which satisfies the estate's own reliability rule. Converting one fixture makes it the odd one out. |

**Follow-through I have not done** (needs your go-ahead): file one ticket for the
two house-wide deferrals (#2 workspace headers, #8 fixture polling — one ticket,
"backend-client + fixture support-code consistency"), and post replies on threads
#2, #5, #6 and #8 so the invalid and deferred ones are visibly answered rather
than silently ignored.

---

## Estate gates: tag_lint **PASS** · tsc **FAIL (pre-existing)** · playwright --list **PASS**

- `tag_lint`: 58 specs checked, 1 exempt, **0 problems** — re-run after my edits.
- `tsc`: fails on `tsconfig.json(13,5) TS5102 baseUrl has been removed`. **Not
  this PR** — identical failure on `main` (see note 2). Re-run with a
  `baseUrl`-free config to actually typecheck the branch, the **only** errors in
  the estate are the two pre-existing duplicate-`deleteDashboard` lines: **this
  PR's own TypeScript is clean.**
- `playwright --list`: all 4 changed specs collect, 7 tests — re-run after my edits.

## What I could not verify

- **Nothing was executed against an Opik stack.** No spec in this PR has been
  run; `--list` proves they parse and collect, not that they pass. Blocker 2 in
  particular is a documented contradiction, not an observed failure.
- **Whether `traces.id_at` truncates a 2200 instant on this deployment.** I read
  the claim in #8097's merged fixture (verified there against ClickHouse 26.3);
  I did not query ClickHouse myself.
- **The post-merge state.** I did not merge, rebase or resolve, so the combined
  `client.ts` and `fixtures/index.ts` are untypechecked in their merged form.
- **Whether the estate's CI runs a TypeScript version where `baseUrl` still
  works.** I only have the locally installed 7.0.2; if CI pins 5.x the `tsc`
  gate may be green there and this is a local-only artifact.

## Not done, deliberately

Per the hard rules: the PR is **still a draft** and I did not flip it, did not
merge it, did not resolve either conflict, and did not add any taxonomy
capability. Given blockers 1 and 2, it is **not** ready for a human to flip
either — it needs the duplication call first.

---
_[review_generated_tests.yml](https://github.com/comet-ml/comet-automation-tests/actions/runs/34179874538) ·


[OPIK-7456]: https://comet-ml.atlassian.net/browse/OPIK-7456?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #8109, opened 2026-09-02
[OPIK-7862] [BE] fix: accept window_in_seconds for threshold errors

### LINKED ISSUE #7862: Alert creation API silently accepts `window_in_seconds` for `threshold:errors`, evaluator requires `window`; alert still never fires after the fix

lorenzozanee:
## Details
`threshold:errors` alerts accepted `window_in_seconds` when created, but the evaluator only read `window`, so stored alerts never used the intended window. This change lets evaluation accept either key while keeping the canonical `window` path unchanged.

## Issues

- Fixes #7862

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/AlertTriggerConfig.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/AlertTriggerConfig.java
@@ -43,6 +43,7 @@ public record AlertTriggerConfig(
     public static final String PROJECT_IDS_CONFIG_KEY = "project_ids";
     public static final String THRESHOLD_CONFIG_KEY = "threshold";
     public static final String WINDOW_CONFIG_KEY = "window";
+    public static final String WINDOW_IN_SECONDS_CONFIG_KEY = "window_in_seconds";
     public static final String NAME_CONFIG_KEY = "name";
     public static final String OPERATOR_CONFIG_KEY = "operator";
     // Comma-separated GuardrailType names (e.g. "PII,TOPIC"); empty/absent means all types.
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/MetricsAlertJob.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/jobs/MetricsAlertJob.java
@@ -59,6 +59,7 @@ import static com.comet.opik.api.AlertTriggerConfig.NAME_CONFIG_KEY;
 import static com.comet.opik.api.AlertTriggerConfig.OPERATOR_CONFIG_KEY;
 import static com.comet.opik.api.AlertTriggerConfig.THRESHOLD_CONFIG_KEY;
 import static com.comet.opik.api.AlertTriggerConfig.WINDOW_CONFIG_KEY;
+import static com.comet.opik.api.AlertTriggerConfig.WINDOW_IN_SECONDS_CONFIG_KEY;
 
 /**
  * Scheduled job for processing metrics-based alerts.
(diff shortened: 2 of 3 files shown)


jverre:
Hi @lorenzozanee, 

Thanks for the PR on this ! Can you take a look at the Baz comments and see if any of them need to be addressed ? If not, add a comment and resolve the comment

PR #8113, opened 2026-09-02
[NA] [SDK] feat: allow dataset item insertion without deduplication

alexkuzmik:
## Details

Adds a `deduplication` flag to every Python SDK method that writes dataset items. With `deduplication=False` the whole dedup path is bypassed — the existing items are not downloaded from the backend, and no content hashes are computed or compared — so inserts into large datasets stop paying for a full item download.

- Flag added on `Dataset.insert` / `update` / `insert_from_json` / `read_jsonl_from_file` / `insert_from_pandas`, the internal `__internal_api__insert_items_as_dataclasses__`, and the equivalent `TestSuite` methods.
- Skipping dedup leaves the local hash cache marked out-of-sync, so a later `deduplication=True` insert re-syncs from the backend instead of silently creating real duplicates.
- `Dataset.insert`'s `num_threads` default goes from 1 to 4. The internal dataclass method stays at 1 because it has no backend-version gate, and fanning out parallel batches against older backends races on a shared `batch_group_id`.
- Since parallel upload is now the default, the backend-version probe that gates it is cached per `Dataset` — otherwise a loop of inserts would pay one extra round trip per call.

## Change checklist
- [x] User facing
- [x] Documentation update

## Issues

- Resolves #
- OPIK-

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: Implementation, tests and documentation for this change.
- Human verification: Reviewed by the author; docstring wording and the scope of the `num_threads` default change were revised on review feedback.

## Testing

Commands run:
- `pytest tests/unit` in `sdks/python` — 5125 passed, 2 skipped.
- `pre-commit run --files <changed files>` — ruff, ruff-format and mypy pass.

Scenarios covered by new unit tests:
- Identical items are both sent when `deduplication=False`, and dropped when it is on.
- No item download happens when dedup is disabled on a dataset fetched from the backend.
- An insert with dedup disabled leaves the hash cache stale, and the next deduplicated insert re-syncs.
- `update` with unchanged content is still sent when dedup is disabled.
- `insert()` with no `num_threads` uploads batches concurrently, and the version gate is probed once per dataset across repeated inserts.
- The `deduplication` flag is forwarded from `TestSuite.insert` to the dataset.

The concurrency test was mutation-checked by temporarily restoring the old default of 1 — it fails on the barrier timeout, confirming it is not passing vacuously.

Not run: e2e/library-integration suites, which need a live backend.

## Documentation

`manage_datasets.mdx` gains a "Disabling deduplication" section and a note about the new default thread count.

### DIFF
--- a/apps/opik-documentation/documentation/fern/docs-v2/evaluation/advanced/manage_datasets.mdx
+++ b/apps/opik-documentation/documentation/fern/docs-v2/evaluation/advanced/manage_datasets.mdx
@@ -277,12 +277,32 @@ dataset.insert([
 <Tip>
   Opik automatically deduplicates items that are inserted into a dataset when using the Python SDK. This means that you
   can insert the same item multiple times without duplicating it in the dataset. This combined with the `get or create
-  dataset` methods means that you can use the SDK to manage your datasets in a "fire and forget" manner.
+  dataset` methods means that you can use the SDK to manage your datasets in a "fire and forget" manner. It can be
+  turned off with `deduplication=False`, see [Disabling deduplication](#disabling-deduplication).
 </Tip>
+
+#### Disabling deduplication
+
+Deduplication requires the Python SDK to download the dataset's existing items once so it can compare their
+content hashes against the items you are inserting. On large datasets that download dominates the insert. If you
+already know your items are unique — for example when populating a fresh dataset, or when you generate ids
+yourself — pass `deduplication=False` to skip that work entirely: nothing is downloaded, no hashes are computed,
+and every item you pass is sent as-is.
+
+```python title="Python" language="python"
+dataset.insert(items, deduplication=False)
+```
+
+The flag is available on every Python SDK method that writes items — `insert`, `update`, `insert_from_json`,
+`insert_from_pandas` and `read_jsonl_from_file` — as well as on the equivalent `TestSuite` methods. With
+deduplication disabled, inserting the same content twice produces two separate dataset items.
 
 <Note>
   When using the SDK to insert items, a new dataset version is automatically created.
   If you insert items in multiple batches within a single `insert()` call, they are grouped into one version.
+  The Python SDK uploads those batches on 4 worker threads by default; use `num_threads=1` to upload them
+  sequentially instead. Parallel upload requires a recent Opik backend — against older ones the SDK falls back
+  to a sequential upload and logs a warning.
 </Note>
 
 Once the items have been inserted, you can view them in the Opik UI:
--- a/sdks/python/src/opik/api_objects/dataset/dataset.py
+++ b/sdks/python/src/opik/api_objects/dataset/dataset.py
@@ -611,14 +611,18 @@ class Dataset(DatasetExportOperations):
         )
         LOGGER.debug("Successfully sent dataset items batch of size %d", len(batch))
 
-    def _resolve_num_threads(self, num_threads: int) -> int:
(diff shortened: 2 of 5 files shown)


CometActions:

`deduplication=False` creates a new user-visible outcome: the same content inserted twice now becomes two dataset items with distinct generated ids, both rendered on the items page and both counted by the version counters. Nothing in the e2e estate asserts SDK dedup at all — `dataset-version-repeated-item-id.spec.ts` deliberately seeds through the backend client *because* `Dataset.insert()` dedups, and `dataset-crud-smoke.spec.ts`'s sdk-round-trip adds its item through the UI. Your unit tests pin the call shape against a mocked rest client; what no test covers is what the backend actually stores and the page shows, and the `deduplication=False` -> `_hashes_synced = False` -> re-sync on the next deduplicated insert sequence. This is reachable pre-merge: the SDK bridge resolves `opik` from this repo, so it only needs `deduplication` threaded through `/datasets/insert-items`.

Would target `datasets.sdk-round-trip`, `datasets.view-items`.

1. Insert the same item content twice with `deduplication=False` through the SDK bridge; confirm two items with distinct ids are stored and both render on the dataset items page
2. Insert the same content again on the same dataset with the default (`deduplication=True`) and confirm it collapses to no new item — so the flag is the only difference
3. After a `deduplication=False` insert, insert an overlapping item WITH dedup on and confirm the forced hash re-sync drops it (the `_hashes_synced = False` path)
4. Check the Version history tab after a non-dedup insert of N duplicate entries: items_added and "Item count" should count all N
5. Repeat step 1 through `TestSuite.insert(..., deduplication=False)` and confirm the suite items page shows both

</details>

**Also already tested.** The `num_threads` 1 -> 4 default change lands on every existing caller of insert()/update(), but the semantics it exposes are covered: `tests_end_to_end/e2e/tests/datasets/dataset-version-counters.spec.ts` asserts a parallel multi-batch insert (num_threads=8) cuts one version with the same items_total/added/modified and the same stored items as the sequential path, and it asserts on the result rather than the transport. One follow-up on our side rather than yours: our SDK bridge pins `num_threads=1` (e2e/services/opik-sdk-driver/src/opik_sdk_driver/schemas.py), so nothing currently exercises the new default — we should follow it to 4.

_also touches Python SDK_

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/33615442342)

CometActions:

### Access Information
- **URL:** https://pr-8113.dev.comet.com
- **Namespace:** pr-8113
- **Version:** 2.2.47-8113-merge-3176
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-8113%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

PR #6851, opened 2026-05-25
[OPIK-6595] [QA] feat: Dataset CRUD smoke (SDK-create + UI-create variants)

AndreiCautisanu:
## Details

Second CUJ smoke test in the Opik 2.0 E2E suite — slot 2 of 6 in the data-plane test sequence. Seeds datasets via the Python SDK code path (`opik.Opik().create_dataset(...).insert(...)` → bridge route) and verifies they round-trip end-to-end through the Datasets list + dataset items pages, plus exercises both write-direction paths (SDK-create → UI-verify and UI-create → SDK-verify) in two paired test cases. Sets the SDK-create + UI-create pair-in-one-PR pattern the next two CUJ tickets (OPIK-6137 Test Suites, OPIK-6598 Annotation Queue) will copy.

Cross-package: ships four FE `data-testid` additions on the Datasets UI to disambiguate the three near-identical "Save changes" buttons that co-exist during draft mode (panel stage, page commit, modal confirm) plus the items toolbar Add button that collapses to icon-only at narrow viewports. Also extends the test-infra orphan sweep (`global-setup.ts` + `global-teardown.ts`) to cover datasets — datasets do NOT cascade-delete with their parent project, so the per-run `cuj-{runId}-` sweep needs to handle them explicitly. Without this, every run would leave a dangling dataset behind.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- OPIK-6595

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 4.7
- Scope: full implementation across bridge route, SDK client method, fixture, backend-client inspection methods, two POMs, two test cases, four FE testids, and dataset orphan sweep additions to global setup/teardown
- Human verification: code review + manual staging runs (cloud) at each phase gate + final 7-test gate run with prefix-sweep verification

### What each test verifies (UI-first, REST used only by fixtures + backend-client inspection)

| Test | Assertions |
|---|---|
| `SDK-seeded dataset renders in list and items page; UI add and delete commit as new versions` | Seeds a 3-item dataset via the bridge. Asserts the row appears in the list view, drills into the items page, exercises the side-panel Add Item flow + page-level commit (asserts v2 + Draft cleared), then Actions-menu Delete + commit (asserts v3 + Draft cleared). Two full Draft→commit cycles. |
[...]

PR #8114, opened 2026-09-02
[NA] [QA] Proposed e2e specs from the #8113 exploration — dataset/test-suite insert deduplication flag

CometActions:
## Details

**Generated by the release QA side flow (release-test-proposal).** It is a draft on
purpose and needs a human review before merge — nothing here has been reviewed by a person.

### Where these came from

[#8113](https://github.com/comet-ml/opik/pull/8113) — `[SDK] feat: allow dataset item
insertion without deduplication`. A human-driven exploration worked the change on the PR's
own deployed environment (`pr-8113.dev.comet.com`, backend `2.2.47-8113-merge-3176`) and
verified each flow below by hand before any spec was written.

### Which base this targets, and why

The run that produced this was told #8113 was still open with its head branch alive, and to
base on `aliaksandrk/NA-dataset-insert-without-deduplication`. That is no longer true:
**#8113 is merged** (merge commit `976d1ca`) and the head branch returns 404, so this PR is
based on **`main`** instead, which already carries the change.

The specs were written against the PR's head commit
`c684b655c30932b0f0fe54ba8155826a388f480a` and then **re-run on `main`** (this branch's
base) before this PR was opened — see the run below. The app they ran against is the PR
environment, which is a merge build of #8113 into main, so both the SDK under test and the
e2e estate the specs live in are the post-merge code.

### What is in here

Three things, all in `tests_end_to_end/`:

**1. The bridge can express the flag** (it could not before — the exploration flagged this as
a blocker for writing any spec at all).

- `deduplication: bool = True` on `DatasetInsertItemsRequest` and
  `TestSuiteInsertItemsRequest` (both are `extra="forbid"`), threaded into
  `Dataset.insert(...)` / `TestSuite.insert(...)`, and surfaced on `insertDatasetItems` /
  `insertTestSuiteItems` in `core/sdk/python-sdk-client.ts`.
- A new `POST /datasets/insert-items-session` route + `insertDatasetItemsSession` client
  method. `/datasets/insert-items` builds a fresh client — and therefore a fresh `Dataset` —
  per request, and a backend-fetched `Dataset` always starts with its local hash cache
  marked unsynced. That makes the second spec below **impossible to fail** across separate
  calls: the third insert would re-sync regardless of what the second one did. The new route
  keeps one `Dataset` alive across a sequence of inserts, which is the only shape where
  "the cache was invalidated and re-synced" is distinguishable from "a new object synced
  because it always does."

**2. Two POM additions**, no new selectors invented:

- `DatasetItemsPage.versionChangeSummary(versionName)` — the Version history "Changes" cell
  (`+ N` / `~ N` / `− N`), addressed by `data-cell-id` exactly like the existing
  `versionItemCount`. This is the only place a user sees the per-version added/modified
  split; "Item count" alone cannot tell a version that added one item from one that added
  two and deleted one.
- `TestSuiteItemsPage.itemRowIds()` (mirroring the dataset page) and `itemRows()` on both,
  so row counts use a retrying `toHaveCount(...)` rather than a one-shot `countItems()`.

**3. Two specs.** Each pairs the flag against its own default on a second entity seeded
identically, so the flag is the only variable and the test cannot pass by coincidence — not
against a backend that stored everything, nor one that collapsed everything.

### `tests/datasets/dataset-insert-deduplication.spec.ts`

| Test | Tags | Asserts |
|---|---|---|
| `deduplication=false stores every duplicate and counts it; the default collapses duplicates into one item and one version` | `@t2-cuj` `@cap:datasets.sdk-round-trip` | Same item inserted twice into two datasets, one flag apart. Dedup-off holds 2 items under distinct ids and cuts 2 versions (`items_total` 1→2, `items_added` 1 each, `items_modified` 0); the control holds 1 item and **one** version — the second call deduplicated to zero, so no batch was sent and none was cut. UI: 2 rows on the Records tab matched **by `data-row-id`** (the rows are content-identical, so only identity shows both stored items are rendering rather than one twice), Item count `1` / `2` and Changes `+ 1` / `+ 1` on Version history; 1 row for the control. |
| `A deduplication=false insert invalidates the local hash cache, so the next deduplicated insert still drops the duplicate` | `@t2-cuj` `@cap:datasets.version-history-view` | Three inserts in one SDK session: `[ITEM]` deduplicated, `[ITEM, OTHER]` **not** deduplicated, `[OTHER, THIRD]` deduplicated. Result is 4 items (2×ITEM, 1×OTHER, 1×THIRD) and versions `+1 / +2 / +1` over totals `1 / 3 / 4`, asserted in the API and rendered on Version history. |

**One deliberate deviation from the verified flow, and it matters.** The candidate's flow was
*two* `deduplication=False` inserts followed by a deduplicated `[duplicate, new]`. Run through
the bridge that sequence passes whether or not the SDK invalidates its cache, because the
session's `Dataset` starts unsynced and insert 3 would sync anyway. The order above starts
with a **deduplicated** insert, which forces the one-shot sync and leaves the cache marked
in-sync holding only ITEM's hash — so insert 3 can only be right if the non-deduplicated
insert 2 invalidated it. Without that invalidation, OTHER is sent again under a fresh id and
v3 reads `+ 2` over 6 items instead of `+ 1` over 4. Same behaviour the exploration verified,
sharpened so the test can actually fail.

### `tests/test-suites/test-suite-insert-deduplication.spec.ts`

| Test | Tags | Asserts |
|---|---|---|
| `deduplication=false stores both identical test cases and the Test cases tab renders both; the default stores one` | `@t2-cuj` `@cap:test-suites.view-suite-items` | `TestSuite.insert` is a separate public entry point routing into the same internal insert. Dedup-off suite holds 2 items under distinct ids and renders 2 rows; the control holds 1 and renders 1. Rows matched by `data-row-id` — the Test cases table renders only ID / Tags / Created, so the item content is not visible there and identity is the only discriminator. The content assertion is API-side for that reason. |

### Taxonomy

Both specs added to their areas' `specs:` lists. All three capabilities were already
`covered: true`, so no flip was needed; each `note:` was extended to scope what is now
asserted (`tag_lint.py` does not check either of those, so they are worth reading).
No new capability keys were invented.

### Verification — passed

Run against the PR environment (`OPIK_BASE_URL=https://pr-8113.dev.comet.com`, OSS install,
workspace `default`, no auth), from `tests_end_to_end/e2e/` **on this branch**:

```
OPIK_DEPLOYMENT=oss npx playwright test \
  tests/datasets/dataset-insert-deduplication.spec.ts \
  tests/test-suites/test-suite-insert-deduplication.spec.ts \
  --reporter=list --workers=2
```

```
✓ 1 dataset-insert-deduplication.spec.ts:183 › …invalidates the local hash cache… (5.0s)
✓ 2 dataset-insert-deduplication.spec.ts:51  › …stores every duplicate and counts it… (7.7s)
✓ 3 test-suite-insert-deduplication.spec.ts:23 › …stores both identical test cases… (7.3s)

3 passed (18.5s)
```

`python3 tests_end_to_end/coverage/tag_lint.py --taxonomy tests_end_to_end/coverage/taxonomy.yaml
--estate tests_end_to_end` → **`58 specs checked, 1 exempt, 0 problem(s)`**.

**`npx tsc --noEmit` could not be used as a gate here**, and not because of this change:
`tsconfig.json` still sets `baseUrl`, which TypeScript 7 (the pinned `^7.0.2`) has removed,
so it fails identically on an untouched checkout. Type-checking with `baseUrl` dropped
reports the same 94 pre-existing errors before and after this change and **none in the files
added or edited here**. Worth fixing separately.

- **Nothing was dropped.** All three verified candidates became specs; none failed.
- The exploration had already set aside a fourth item — a 1200-duplicate multi-batch
  `deduplication=False` insert on the new `num_threads=4` default. `dataset-version-counters.spec.ts`
  already covers the parallel-upload-collapses-to-one-version axis at `num_threads=8` and
  asserts on the result rather than the transport, so a new spec would add cost without
  coverage.
- Relatedly, the exploration suggested changing the bridge's `DatasetInsertItemsRequest`
  default from `num_threads=1` to `4` so existing specs exercise the SDK's new default.
  **Not done here** — it silently changes how every existing dataset spec uploads, which is a
  call for the estate's owners, not a side effect of a test proposal.

### Reviewer notes

- The new bridge route is the one piece of added surface that is not a spec. If you would
  rather not have it, the second dataset test cannot be written honestly — say so and it can
  be dropped rather than weakened into a test that always passes.
- Both specs clean up through `registerDatasetCleanup` (datasets and test suites share
  storage and do not cascade with their project); no cleanup lives in a test body.

## Change checklist

## Issues

No ticket — generated by the QA test radar from an exploration run.

## Testing

## Documentation


CometActions:
## 🔍 Generated-test review

# PR #8114 — [NA] [QA] Proposed e2e specs from the #8113 exploration — dataset/test-suite insert deduplication flag

CometActions · draft: yes · 9 files, 2 new specs (3 tests) · head `906a48c9ec` (was `48ff8eaa95`)

**VERDICT: needs changes — blocked on the merge conflict, not on the specs.**

The test work is good: correct placement, honest assertions, real paired controls,
no flake patterns, and support code that follows the driver's established pattern.
One tag defect was found and fixed. What blocks it is that the branch cannot merge —
and one of the four conflicts is a semantic union that would silently drop a merged
spec's coverage note if resolved by taking either side.

---

## Blockers (1)

### 1. The branch is CONFLICTING against `main` — 4 files

Not resolved here, by instruction. Details in the *Merge conflicts* section below.
Resolve that first; it is the only thing between this PR and ready.

**Does the conflict make the other findings moot?** No, but it re-orders them. The
conflict is mechanical (adjacent insertions) in three of four files, so the specs
themselves survive resolution unchanged. The one place a careless resolution loses
work is `taxonomy.yaml`, and that is also where this PR's remaining open question
lives — see Note 1. Resolve the conflict and answer Note 1 in the same sitting.

---

## Should fix (1 — fixed and pushed)

### 1. Invisible coverage: the first dataset test asserted Version history and did not say so — **FIXED**

`tests_end_to_end/e2e/tests/datasets/dataset-insert-deduplication.spec.ts:53`

The test carried only `@cap:datasets.sdk-round-trip`, but its last two steps assert
the Version history tab's **Item count** (`versionItemCount('v1'|'v2')`) and
**Changes** (`versionChangeSummary`, `+ 1`) cells — that is
`datasets.version-history-view`'s surface, asserted and unclaimed.

Fixed by adding `@cap:datasets.version-history-view` to that test's tag list, with a
comment saying why. Committed as `906a48c9ec` and pushed to
`comet-qa-bot/NA/dataset-insert-deduplication-e2e`. No `taxonomy.yaml` edit was
needed (the key is already `covered: true` and the PR's own note already mentions
the Changes tags) — and I deliberately did not touch `taxonomy.yaml` because it is
one of the four conflicting files.

---

## Notes (4)

### 1. There is no taxonomy capability for insert deduplication — and two open PRs have now both worked around it

All three tests in this PR exist to assert one thing: `Dataset.insert` /
`TestSuite.insert`'s `deduplication` flag. The taxonomy has **no key for it**. The
generator did what the skill warns about — reached for the nearest existing key and
appended a note:

| Test | Tag used | What it actually asserts |
|---|---|---|
| dataset test 1 | `datasets.sdk-round-trip` | dedup on/off, plus version counters |
| dataset test 2 | `datasets.version-history-view` | hash-cache re-sync after a non-dedup insert |
| test-suite test | `test-suites.view-suite-items` | dedup on/off on the Test cases tab |

I am not inventing a capability to fix this — that decides the coverage denominator
and is yours.

What makes it worth deciding **now** rather than later: open PR **#8127**
(`AndreiCautisanu/OPIK-7791/release-2249-exploration-specs`) adds
`test-suites/test-suite-insert-dedup-listed-suite.spec.ts`, which asserts *the same
test-suite dedup behaviour*, and hangs it off a **different** existing key —
`test-suites.list-suites`. Its own taxonomy comment says so in as many words:

> "That behaviour has no key of its own here — this one names the listing, which is
> the path the spec exercises, not the dedup it asserts. Worth a dedicated capability
> if anyone extends this area."

Two independent generators reached the same conclusion independently. If both land
as-is, test-suite insert deduplication is claimed under `view-suite-items` *and*
`list-suites`, and nothing in the taxonomy says the behaviour is covered at all.

**Choose one:** add `insert-deduplication` to `datasets` and `test-suites` (and
retag all four specs across both PRs), or accept the borrowed keys and leave dedup
honestly uncovered in the map. Either way it should be one decision covering #8114
and #8127 together.

### 2. The `tsc` gate FAIL is pre-existing estate breakage, not this PR

```
```

`tests_end_to_end/e2e/tsconfig.json` is **byte-identical to `main`** — this PR does
not touch it. The cause is dependabot PR **#7630**, which bumped `typescript` to
`^7.0.2` (installed: 7.0.2) without migrating the config; TS 7 removed `baseUrl`.
The gate has been failing on `main` since, and it fails the same way with or without
this PR's changes.

I verified the PR's own TypeScript separately by removing `baseUrl` from a throwaway
copy of the config and re-running: **zero errors in any file this PR changes**
(`dataset-insert-deduplication.spec.ts`, `test-suite-insert-deduplication.spec.ts`,
`pom/dataset-items.page.ts`, `pom/test-suite-items.page.ts`,
`core/sdk/python-sdk-client.ts`). The config was restored; nothing was committed.

Two things that fell out of that run, both pre-existing on `main` and both worth
their own PR — do **not** fold them into this one:

- With `baseUrl` gone, `@types/node` (26.2.0, installed) still does not resolve
  under TS 7, so ~90 errors appear estate-wide (`process`, `Buffer`, `node:path`,
  `NodeJS`). The TS 7 migration is a real piece of work, not a one-line delete.
- `core/backend/client.ts:552` and `:1044` — **`Duplicate identifier 'deleteDashboard'`**.
  A genuine latent defect that the broken gate has been hiding.

### 3. `datasets.sdk-round-trip` is drifting

Originally it meant "UI-created entity read back through the SDK"
(`dataset-crud-smoke.spec.ts`). #8181 stretched it on `main` to cover
`get_items()` reads. This PR stretches it the other way — SDK write, REST/UI read;
it never calls `get_items()`. Not false, and I left the tag alone, but three
different meanings on one key is how a coverage map stops meaning anything. Worth a
look when you settle Note 1.

### 4. Read-after-write timing is the one thing I could not de-risk

The default-dedup path relies on insert 2 syncing hashes from the backend and seeing
insert 1's item. I read the SDK and the mechanism is sound (`_send_batches` is a
synchronous REST call; `client.end(flush=True)` runs before the route returns), and
the same read-after-write is what the product feature itself depends on — but I could
not execute the specs, so this is reasoned, not observed. If either dataset test
flakes on a first CI run, this is where to look.

---

## Merge conflicts (4 files) — do not take either side wholesale

Merge base `976d1cab7d`. All four conflicts come from **one already-merged sibling
generated PR: #8181** ("Proposed e2e specs from the 2.2.51 → 2.2.52 release
exploration"), which added dataset *read* plumbing into the exact regions where this
PR adds dataset *session-insert* plumbing.

| File | `main` side (#8181) | This PR | Correct resolution |
|---|---|---|---|
| `e2e/services/.../schemas.py` | `DatasetReadItemsRequest/Response`, `DatasetReadWithMidReadInsertRequest/Response` | `DatasetInsertCall`, `DatasetInsertItemsSessionRequest/Response` | **Union.** Disjoint classes, adjacent insertion point. Keep all seven. |
| `e2e/services/.../routes/datasets.py` | import block + `read_dataset_items`, `read_dataset_with_mid_read_insert` | import block + `insert_dataset_items_session` | **Union**, both hunks — merge the import lists, keep all three routes. The `deduplication=` forwarding inside the existing `insert_dataset_items` auto-merged cleanly. |
| `e2e/core/sdk/python-sdk-client.ts` | `readDatasetItems`, `readDatasetItemsWithMidReadInsert` (interface + impl) | `insertDatasetItemsSession` (interface + impl) | **Union**, both hunks. |
| `coverage/taxonomy.yaml` | `specs:` +3 read specs; `sdk-round-trip` note rewritten to describe the parallel read | `specs:` +1 dedup spec; `sdk-round-trip` and `version-history-view` notes rewritten to describe dedup | **⚠ Union, and this one is semantic.** |

**The taxonomy conflict is the one that can silently lose work.** `sdk-round-trip`
had **no note** at the merge base; *both* sides added one, describing entirely
different coverage. Taking this PR's side drops #8181's parallel-read note; taking
`main`'s side drops this PR's dedup note. The `specs:` list must end up with all
four entries (3 from #8181 + 1 here), and both notes must be merged into one
sentence, not chosen between. Same for `version-history-view`, where this PR appends
a clause to a note `main` also edited.

### Reverse case — a clean merge could still lose work

Git will not warn you about this. Two shared files this PR changes are also changed
by **open** PR **#8127**:

- `e2e/core/sdk/python-sdk-client.ts` — this PR adds `deduplication?: boolean` to
  `insertTestSuiteItems`; **#8127 adds a `factory` option to that same interface
  member.** Same method, same object literal. Whichever lands second will conflict
  or, worse, merge cleanly having dropped the other's option.
- `e2e/services/.../schemas.py` — both extend the test-suite insert request schema.

`taxonomy.yaml` is touched by **ten** open generated PRs (#8080, #8084, #8100,
#8127, #8144, #8151, #8160, #8167, #8172, #8184). #8144 is a taxonomy-discovery PR
and will conflict with all of them by construction. Land them in a deliberate order,
or expect to re-resolve this file repeatedly.

---

## Bot comments

All four are `baz-reviewer`, all on `schemas.py` / the new dataset spec. None were
obeyed; each was checked against the estate.

|---|---|---|---|
| 1 | `deduplication` is only on `DatasetInsertItemsRequest`; `createDataset`/`createTestSuite`/seed routes cannot express it | **valid observation, out of scope** | Factually right — `DatasetCreate.items` and `TestSuiteCreate.items` both exist and both seed with default dedup. But no spec needs it, and adding unused knobs to a test driver is speculative surface. Defer → fold into the driver work if a spec ever needs create-time dedup control. |
| 2 | `DatasetInsertCall.num_threads: int` accepts booleans (pydantic lax mode coerces `true`→`1`) | **invalid — contradicts a documented estate convention** | The coercion claim is true, but `schemas.py` has **0** `Field(`/`min_length`/`constr` in 421 lines, and #8181's own `DatasetReadItemsRequest` docstring on `main` states the rule explicitly: *"plain ints rather than constrained ones on purpose: the SDK's own validation of them … is part of what a caller reads this route to assert, so pydantic must not reject those values before the SDK sees them."* `Dataset.insert` raises `ValueError` for bools and `< 1` itself. Baz's fix would break the thing the route exists to expose. |
| 3 | No unit tests for the new `DatasetInsertCall` / `DatasetInsertItemsSessionRequest` validation boundary, "as `AGENTS.md` requires" | **valid in the abstract, about the codebase not this PR** | The SDK driver has **zero** test files — no `test_*.py` anywhere under `opik-sdk-driver/`. Adding a test module for only the two newest schemas makes this PR the sole exception to an eleven-route pattern. Defer → one driver-wide "add a schema/route test suite" ticket, if the driver is judged to warrant one at all. |
| 4 | Test bodies are missing the required `test.info().annotations.push({type:'description'})`, per `.agents/skills/playwright-e2e/test-conventions.md` | **invalid — the convention does not exist** | `grep -rn "annotations.push" tests_end_to_end/e2e/tests/` → **0 hits across all 55 spec files**, and `.agents/skills/playwright-e2e/test-conventions.md` **does not exist anywhere in the repo** (`find . -name test-conventions.md` → nothing). Baz cited a file it invented. Obeying it would make these two specs the only ones in the estate carrying the annotation. |

**Follow-through I have *not* done, and want your go-ahead for:**

1. **File the deferred ones as one ticket, not three** — "opik-sdk-driver: decide on
   a schema/route test suite and on create-time insert options" covers findings 1
   and 3, which are the same theme (the driver's own contract surface). Draft ready
   on request.
2. **Reply on the four threads** so findings 2 and 4 are visibly answered rather
   than ignored — a silent thread reads as an oversight to the next human. One or
   two sentences each, verdict and reason. I have not posted anything.

---

## Estate gates

Re-run after the fix, from the PR head at `906a48c9ec`:

| Gate | Result | |
|---|---|---|
| `tag_lint.py` | **PASS** | 58 specs checked, 1 exempt, 0 problems |
| `tsc --noEmit` | **FAIL** | Pre-existing on `main`; `tsconfig.json` `baseUrl` vs TypeScript 7 (#7630). Not caused by this PR — see Note 2. This PR's own TS is clean, verified separately. |
| `playwright test --list` | **PASS** | 3 tests in 2 files, both new specs collect |

---

## What I could not verify

- **I did not execute the specs.** No Opik stack in this workspace. Every assertion
  is reviewed statically. I did check the surfaces they address against the
  frontend source: `change_summary` is a real column id in
  `VersionHistoryTab.tsx:37`, `data-cell-id={cell.id}` is how `DataTable.tsx`
  emits it, and `VersionChangeSummaryCell` renders `+ {itemsAdded}` — so
  `toHaveText('+ 1')` matches under Playwright's whitespace normalisation.
- **The SDK premise, I did verify.** `Dataset.insert(deduplication=)` and
  `TestSuite.insert(deduplication=)` both exist, and
  `__internal_api__insert_items_as_dataclasses__` really does set
  `self._hashes_synced = False` on the non-dedup path — so dataset test 2's
  hash-cache-invalidation premise is correct, not guessed.
- **The PR description.** Not in the workspace artifacts and I was told not to
  re-fetch, so I could not check it for stale claims. Worth a glance before ready.
- **The conflicts, deliberately.** Reconstructed read-only via `git merge-tree` and
  `git merge-file` into `/tmp`; the checkout was never merged, rebased or
  force-pushed.
- **Whether #8127's overlap is a true collision.** I read its diff through `gh`,
  which shows both PRs editing `insertTestSuiteItems`. I did not check out its
  branch to prove the merge outcome.

---

## Ready?

**Not yet, and I have not flipped anything.** The PR is still a draft and I did not
mark it ready or merge it — that is a person's call.

Once (a) the four conflicts are resolved as unions, with the two `taxonomy.yaml`
notes genuinely merged rather than chosen between, and (b) Note 1 is answered for
both #8114 and #8127, this is good work and ready for a human to flip.

---
_[review_generated_tests.yml](https://github.com/comet-ml/comet-automation-tests/actions/runs/34179906151) ·

CometActions:
## 🔍 Generated-test review

# PR #8114 — [NA] [QA] Proposed e2e specs from the #8113 exploration — dataset/test-suite insert deduplication flag

CometActions · draft: yes · 9 files, 2 new specs (3 tests) · head now `4bf3803276`

**VERDICT: needs a human decision — not a code problem**

The specs themselves are good: honest tags, real assertions, fixture teardown, no
sleeps, zero new type errors. I merged `main` and resolved all four conflicts, and
the gates are as green as this repo currently allows. What blocks it is not in the
diff — a **second open branch covers the same capability** with a parallel driver
route, and only a person can decide which one lands.

---

## Blockers (1)

### 1. A sibling branch duplicates this PR's entire subject — `git` will not tell you

`origin/AndreiCautisanu/OPIK-7791/dataset-insert-deduplication-specs`
(commit `c28fc4a837`, not an ancestor of this PR) covers the *same* capability with
the *same* two scenarios, and touches the same four shared files:

| | This PR (#8114) | Sibling branch |
|---|---|---|
| specs | `dataset-insert-deduplication.spec.ts` (2 tests) | `dataset-insert-deduplication-off.spec.ts` + `-resync.spec.ts` |
| driver route | `POST /datasets/insert-items-session` | `POST /datasets/insert-items-sequence` |
| driver handler | `insert_dataset_items_session` | `insert_dataset_items_sequence` |
| schemas | `DatasetInsertCall`, `DatasetInsertItemsSessionRequest/Response` | `DatasetInsertStep`, `DatasetInsertSequenceRequest/Response` |
| TS client | `insertDatasetItemsSession(...)` | `insertDatasetItemSequence(...)` |
| taxonomy | rewrites `sdk-round-trip` + `version-history-view` notes | rewrites the same two notes |

Because the names differ everywhere, **both branches merge cleanly on their own**.
Merge both and the estate silently gains two routes that do the same job, two SDK
client methods that do the same job, and duplicate coverage of one capability —
with the second `note:` rewrite overwriting the first. This is precisely the
failure git cannot flag.

**Fix:** a human picks one. Nothing to change in this PR until that is decided —
which is why I did not touch it.

---

## Should fix (0)

Nothing. Both bot findings that survived scrutiny are estate-wide, not local
(see the table below).

---

## Notes (2)

### 1. `tsc` is broken on `main`, not by this PR — and it belongs in its own PR

The whole `tsc` gate is one error, in a file this PR never touches:

```
```

Dependabot #8106 bumped TypeScript to **7.0.2** on main; 7.x removed `baseUrl`.
The config still sets it, so **type-checking never starts** — for any PR.

I verified this PR is not responsible, by building an `origin/main` worktree and
running both trees with an identical `baseUrl`-stripped config:

- `origin/main`: **94** errors
- this PR merged: **94** errors — the same set, with two line-number shifts in
  `python-sdk-client.ts` from the lines my merge added
- errors in this PR's own files (2 specs, 2 POMs, driver, client): **none**

The remaining 94 are pre-existing: missing `@types/node` resolution under
`moduleResolution: Bundler`, plus a genuine `TS2300: Duplicate identifier
'deleteDashboard'` at `core/backend/client.ts:674` and `:1243`.

The fix is one line — delete `"baseUrl": "."`, since `paths` (`"@e2e/*": ["./*"]`)
already resolves relative to the tsconfig. **I deliberately did not apply it**: it
is an estate-wide repair unrelated to this PR's subject, and burying it in a
generated test PR is how it stops being reviewable. Happy to open that PR
separately — it would also want the `deleteDashboard` duplicate and the
`@types/node` resolution looked at.

### 2. `counters()` sorts version names with `localeCompare`

`dataset-insert-deduplication.spec.ts:37`. With `v10` present this orders
`v1, v10, v2`. Only 3 versions exist here so it cannot misfire today; noting it as
a trap for whoever copies this helper into a longer-version test.

---

## Conflicts (4) — all resolved, all inside `tests_end_to_end/`

`main` landed the dataset parallel-read work (#8181) in the same shared
test-support files. Nothing outside `tests_end_to_end/` conflicted, so no
inherited-commit problem and nothing to stop for.

| File | This PR's side | `main`'s side | Resolution |
|---|---|---|---|
| `coverage/taxonomy.yaml` | adds `dataset-insert-deduplication.spec.ts`; rewrites `sdk-round-trip` + `version-history-view` notes | adds 3 read/count specs; rewrites the same two notes | `specs:` unioned (4 lines). `sdk-round-trip` **note meaning unioned** — each side describes coverage the other cannot see. `version-history-view`: main's text equals the merge base, this PR appends a clause, so this PR's line wins. No capability key invented. |
| `core/sdk/python-sdk-client.ts` | `insertDatasetItemsSession` on the interface + impl | `readDatasetItems`, `readDatasetItemsWithMidReadInsert` | Both kept. Both hunks opened **mid-docblock** and shared a trailing `args, { timeoutMs }, );` — each member given back its own `/**` and its own closer rather than concatenated. |
| `services/.../routes/datasets.py` | `POST /insert-items-session` | `POST /read-items`, `POST /read-with-mid-read-insert` | Imports unioned; all 5 routes present. Same mid-docstring split — the shared `"""` and shared `client`/`get_dataset` preamble belonged to different functions on each side. |
| `services/.../schemas.py` | `DatasetInsertCall`, `DatasetInsertItemsSessionRequest/Response` | `DatasetReadItemsRequest/Response`, `DatasetReadWithMidReadInsert*` | Both kept; all 11 `Dataset*` classes present, 4-space class-body indentation preserved. |

Verified after resolution: no markers in tracked files, `py_compile` clean,
`yaml.safe_load` clean, and the tsc error-set diff above (which is what proves the
TS interface and its implementation actually still line up).

**Does the conflict moot the other findings?** No — it is resolved and pushed. The
*duplicate-branch* blocker is the one that outranks everything else: if the sibling
branch wins, none of this PR's code survives to be polished.

---

## Bot comments

Four distinct findings from `baz-reviewer[bot]`, plus one follow-up acknowledging
commit `906a48c`. Each verdict below was checked against the code, not the abstract.

| 1 | `deduplication` only on `DatasetInsertItemsRequest`, so `createDataset({items, deduplication:false})` is refused by `DatasetCreate`'s `extra="forbid"`; create/seed paths are default-only | valid, won't fix | Premise confirmed (`DatasetCreate` has no such field). But no spec needs it — both specs create an empty dataset/suite and then insert. Adding the knob to every create/seed model is unasserted surface on a test driver. **Deferred**: add it when a spec needs it. |
| 2 | `DatasetInsertCall.num_threads: int = 1` coerces `true`→`1`; wants strict positive-int validation | valid, estate-wide → not here | `schemas.py` has **zero** field constraints in 507 lines (the one `constr` grep hit is the word "constrained" in main's own prose). Four other `num_threads: int` declarations share the shape, including pre-existing `DatasetInsertItemsRequest` and main's new `DatasetReadWithMidReadInsertRequest`. Main's `DatasetReadItemsRequest` docstring states the ints are *deliberately* unconstrained "so the SDK's own validation … is part of what a caller reads this route to assert". Constraining only the new class makes it the inconsistent one **and** contradicts a documented intent. **Deferred → driver-wide validation ticket.** |
| 3 | No Pydantic/FastAPI tests for the new request boundary (unknown fields, empty `inserts`), "as `AGENTS.md` requires" | valid, out of scope | The driver has **no test module at all** — the only `test_*.py` hit is `routes/test_suites.py`, a route. The finding asks this PR to establish a test tier the driver has never had. **Deferred → same driver-wide ticket.** |
| 4 | Test bodies lack `test.info().annotations.push({type:'description'})` per `.agents/skills/playwright-e2e/test-conventions.md` | **invalid** | **0 of 63** specs in the estate use `annotations.push`, and the cited file `.agents/skills/playwright-e2e/test-conventions.md` **does not exist**. The bot enforced a convention this estate has never had, from a doc that isn't there. Obeying it would make these two specs the only ones with it. |
| 5 | Follow-up: "`906a48c` addressed this by forwarding `deduplication`… create-time and seed paths still default-only" | accurate | Correct, and intentional — same reasoning as #1. |

**Two follow-through actions I have not taken** (say the word):

1. **File one ticket**, not three: *"opik-sdk-driver: request-model validation +
   a boundary test module"* — covering findings #2 and #3, which are the same
   theme (the driver has neither field constraints nor unit tests, anywhere).
   Finding #1 folds in as "expose `deduplication` on create/seed when a spec needs it".
2. **Reply on the four threads** so #4 in particular is visibly answered — a bot
   finding left silent reads as an oversight to the next human.

---

## What I fixed

- Merged `origin/main` and resolved all four conflicts (table above); committed as
  `4bf3803276` and **pushed plainly** to `comet-qa-bot/NA/dataset-insert-deduplication-e2e`.
  No force-push: `RESTACK` was `false`, so the history was never rewritten.

## What I deliberately did not touch

- The `tsconfig.json` `baseUrl` line — estate-wide, its own PR.
- The four bot findings — none is local to this PR (see table).
- The duplicate sibling branch — a human's call.

## Estate gates: tag_lint **PASS** · tsc **FAIL (pre-existing on main)** · playwright --list **PASS**

Run on the merged tree, after resolution:

- `tag_lint --taxonomy … --estate tests_end_to_end` → **66 specs checked, 1 exempt, 0 problems**
- `npx tsc --noEmit` → fails on `tsconfig.json:13` only; **identical 94-error set to `origin/main`**, none in this PR's files
- `npx playwright test --list` → **129 tests in 63 files**; the 3 new tests collect

## What I could not verify

- **The specs have never been run.** No Opik stack, no SDK driver process. Every
  behavioural claim — that `deduplication=false` really stores both copies, that
  v3 renders `+ 1` rather than `+ 2`, that the hash-cache invalidation is
  observable across one `Dataset` object — is unexecuted. `--list` proves the
  files parse and collect, nothing more.
- **`tsc` on the specs is indirect.** The real gate never type-checks anything, so
  I used a `baseUrl`-stripped scratch config (deleted, not committed) and compared
  against an `origin/main` baseline. That shows this PR adds no errors; it is not
  the gate CI will run once `tsconfig.json` is fixed.
- **The Version-history selectors are matched, not seen rendered.** I confirmed
  `change_summary` is a real column id (`VersionHistoryTab.tsx:37`, alongside the
  existing `items_total` this POM already uses), but not that it renders the exact
  string `+ 1`. If those four `toHaveText` assertions fail on first run, that
  format is the first thing to check.
- **Open-PR status of the sibling branch.** I compared branches with local git; I
  did not query GitHub for whether it has an open PR or was abandoned.

---

**This PR is a draft and I have left it a draft.** I have not marked it ready and
have not merged it — deciding between this PR and the sibling branch is the whole
point of the remaining review, and a bot clearing its own gate is not a gate.

---
_[review_generated_tests.yml](https://github.com/comet-ml/comet-automation-tests/actions/runs/34187921048) ·

CometActions:
## 🔍 Generated-test review

# PR #8114 — [NA] [QA] Proposed e2e specs from the #8113 exploration — dataset/test-suite insert deduplication flag

CometActions · draft: yes · 9 files, 2 new specs (3 tests)

**VERDICT: ready with notes** — the merge blocker is fixed and pushed; nothing left that a human must decide except one taxonomy question.

The conflict is resolved and the branch now sits on top of `origin/main`. The
specs themselves are the strongest generated pair I have reviewed against this
estate: both halves of the flag are asserted, each against a control entity
seeded identically, so neither test can pass by coincidence. I made no changes
to the specs, the POMs, or the driver.

**Do not flip this out of draft yourself** — a person should do that. Once the
taxonomy question in Note 1 is answered, it is ready to go.

---

## Merge conflict (resolved)

`MERGE STATE: CONFLICTING` was real, and it was a **blocker** until fixed. One
file, inside `tests_end_to_end/`, so it was mine to resolve.

| side | what it did |
|---|---|
| `origin/main` (#8199, `66589738ee`) | Re-sorted every area's `specs:` list alphabetically, and taught `tag_lint.py` a new **Rule 4** that fails the lint when a `specs:` list is not sorted. Exactly so concurrent spec PRs stop colliding on one trailing line. |
| this PR | Appended its two new specs to the **end** of the pre-sort order — the very habit #8199 exists to stop. |

Both hunks were pure list membership; neither side was wrong about content.
**A correct resolution keeps every entry from both sides and re-inserts this
PR's two in sorted position:**

- `datasets/dataset-insert-deduplication.spec.ts` → after `dataset-crud-smoke`,
  before `dataset-item-count` (`-i` `n` sorts before `-i` `t`)
- `test-suites/test-suite-insert-deduplication.spec.ts` → between
  `test-suite-delete` and `test-suites-smoke`

Taking main's side alone would have dropped both new specs from the taxonomy,
making them invisible to the coverage builder and the radar. Taking this PR's
side alone would have dropped main's three `dataset-version-*` entries and left
the list unsorted, so `tag_lint` Rule 4 would then have failed the branch.

The three capability **notes** this PR rewrote (`datasets.sdk-round-trip`,
`datasets.version-history-view`, `test-suites.view-suite-items`) merged cleanly
— main did not touch them — so no note had to be unioned by hand.

Only #8199 touched `tests_end_to_end/` among the four commits main was ahead by;
the other three (`#8194`, `#8193`, `#7809`) are backend/SDK and fast-merged with
no conflict. **Nothing outside `tests_end_to_end/` was resolved by hand.**

Pushed as `0273278d0b` to `comet-qa-bot/NA/dataset-insert-deduplication-e2e`
(plain push — RESTACK was `false`, so no force was used or needed).

**This did not moot any other finding.** The remaining items below all stand.

---

## Blockers (0)

None.

## Should fix (0)

None. Placement, tag honesty, convention parity, flake and isolation all check
out — the detail is under Notes and Verified below.

## Notes (4)

**1. There is no taxonomy capability for insert deduplication — your call.**
Both specs' actual subject is `Dataset.insert(deduplication=...)` /
`TestSuite.insert(deduplication=...)`. The taxonomy has no key for it, so the
generator hung the coverage on the nearest existing keys:
`datasets.sdk-round-trip`, `datasets.version-history-view` and
`test-suites.view-suite-items`.

This is *not* a false green — all three keys were already `covered: true` before
this PR, so nothing flips on the strength of a test that does not earn it, and
each tagged test genuinely does assert the tagged capability (a real SDK round
trip; the real Version history tab; the real Test cases tab). But the dedup
behaviour itself is now covered and the map will not say so.

Per the hard rule I did **not** invent a key. Two honest options:

- add `datasets.insert-deduplication` and `test-suites.insert-deduplication`,
  retag, and let the denominator grow by two; or
- leave it as-is and accept that the notes are where dedup coverage is recorded.

I lean toward adding the keys — the notes this PR wrote are already carrying
three sentences of load that a capability key would carry better — but this
decides what the product's coverage denominator *is*, so it is yours.

**2. The `tsc` gate is broken estate-wide, and has been since before this PR.**
`tsconfig.json:13` still sets `"baseUrl": "."`, which TypeScript 7.0.2
(`devDependencies`, `^7.0.2`) **removed**. `tsc` aborts on the config, so
`--noEmit` currently type-checks **zero files** — the gate has been reporting
FAIL without ever looking at any code. `tsconfig.json` is untouched by this PR
and identical on `origin/main`; its last change was #6677.

I did **not** fix it here. It is one line, but it is estate-wide (63 spec files),
and unmasking it immediately surfaces a second pre-existing defect that this PR
does not own — see Note 3 — which would turn this PR red for someone else's bug.
It belongs in its own PR: drop `baseUrl`, keep `"paths": { "@e2e/*": ["./*"] }`,
and add `"types": ["node", "js-yaml"]` (without `baseUrl`, node's globals stop
resolving implicitly under TS 7). **Happy to open that PR — say the word.**

**3. `core/backend/client.ts` declares `deleteDashboard` twice** (lines 674 and
1243), a genuine `TS2300` duplicate identifier. Present on `origin/main`, in a
file this PR does not touch. Invisible today only because the gate in Note 2
never gets far enough to see it. Same follow-up PR as Note 2.

**4. Shared-file overlap with four other open generated PRs — a clean merge can
still lose work here.** Git will not warn about any of these:

| file | also touched by | risk |
|---|---|---|
| `coverage/taxonomy.yaml` | #8084, #8127, #8144, #8184, #8192 | Highest. #8144 is a whole-file taxonomy-discovery proposal. #8084 rewrites the **datasets** area — but its `list-datasets` note, not the `sdk-round-trip` / `version-history-view` notes this PR rewrites, so the two do **not** collide semantically; only the `specs:` list overlaps, and #8199's sort now lands them on different lines. |
| `core/sdk/python-sdk-client.ts`, `opik-sdk-driver/schemas.py` | #8127 | Both are append-only in practice. Whichever lands second must keep both sides; a bare concatenation mid-docblock is the trap. |

None of these files has moved **on `origin/main`** under this PR — the last
merged change to the driver and the SDK client was #8181, which this branch
already contains. So the exposure is entirely to the other *open* branches, and
it resolves in whatever order they land, second one keeping both sides.

---

## Verified (no finding)

- **Placement.** `tests/datasets/` matches `spec_dir: datasets`;
  `tests/test-suites/` matches `spec_dir: test-suites`. Both specs are now in
  their area's `specs:` list.
- **Tag honesty.** Every `@cap:` key exists. Every tag is a string literal, none
  computed. `@t2-cuj` is right for both — three tests, no LLM spend, no long
  waits, but multi-entity UI work that does not belong in a 3×/day tier.
- **The behaviour under test is real, not assumed.** I read the SDK:
  `Dataset.insert(items, num_threads=4, deduplication=True)` exists
  (`dataset.py:946`), and `__internal_api__insert_items_as_dataclasses__`
  lines 923–929 set `self._hashes_synced = False` on the non-deduplicated
  branch. Test 2 is a precise regression guard for exactly those seven lines —
  its predicted `v3 = + 1 over 4 items` versus the un-invalidated `+ 2 over 6`
  is the real failure mode. `TestSuite.insert` carries the same kwarg.
- **The UI assertions address real surfaces.** `change_summary` is a real column
  in `VersionHistoryTab.tsx:37`; `VersionChangeSummaryCell` renders `+ {n}` in a
  green tag, `-` when nothing changed — so `toHaveText('+ 1')` is correct under
  Playwright's whitespace normalisation. The tab has no column-visibility state,
  so the column always renders. No new `data-testid` was needed in product code;
  this PR adds **no** production changes at all.
- **Flake and isolation.** No `waitForTimeout`, no `.skip`/`.fixme`/`.only`, no
  ordering dependence. Entities are named off `testNamespace`
  (`cuj-…-w{worker}-{slug}`), so the teardown sweep finds them. Teardown is the
  `registerDatasetCleanup` **fixture**, not a `try`/`finally` in the body — which
  is right, since these datasets are created mid-test. Notably better than
  neighbouring `dataset-version-counters.spec.ts`, which still uses a body-level
  `try`.
- **One-shot backend reads after a write** are the estate's existing convention
  here (`dataset-version-counters.spec.ts:114`, `dataset-crud-smoke.spec.ts:72`),
  so I did not flag them; they are preceded by retrying `toHaveCount` on the UI
  side anyway.
- **Convention parity.** Imports from `@e2e/fixtures`, drives through
  `@e2e/pom` page objects, hand-rolls no login or project creation. The two new
  POM methods mirror `versionItemCount` / `itemRowIds` exactly. `openByName`
  uses `exact: true` cell filtering, so `…-dedup-off` and `…-dedup-on` cannot
  cross-match.
- **Non-spec files are all bucket 1 or 2.** The `opik-sdk-driver` subtree is
  established test infrastructure — 11 route modules on main, `routes/datasets.py`
  dating to #6851. The new `/datasets/insert-items-session` route follows its
  siblings and is wired through `schemas.py` consistently. Its justification is
  sound and checkable: the bridge builds a fresh client per request, and a
  backend-fetched `Dataset` starts `_hashes_synced = False`, so a cross-request
  sequence genuinely cannot observe the invalidation. **This is not scope creep.**

## What I could not verify

- **I did not run the specs.** No Opik stack is up in this workspace. Everything
  above is static analysis plus reading the SDK and frontend source. The
  predicted version counters (`v1 +1/1`, `v2 +2/3`, `v3 +1/4`) are derived from
  the SDK's insert path, not observed.
- **Read-after-write timing.** Test 2's third insert re-syncs hashes from the
  backend immediately after the second insert's write. If dataset-item writes
  are ever not read-your-writes consistent, that step flakes. I could not
  exercise it, and the estate's existing specs make the same assumption, so I
  left it alone — but it is the one line I would watch on the first few runs.
- **Bot findings 1–3 concern runtime request validation** in the driver. I
  reasoned about Pydantic's lax bool→int coercion from the schema definitions
  rather than by sending requests to a running driver.

---

## Bot comments

Five threads, all `baz-reviewer[bot]`; #5 is its own follow-up on #1.

| 1 | `schemas.py:151` — initial-item dedup unavailable: `DatasetCreate` has `extra="forbid"` and no `deduplication`, so `createDataset({items, deduplication:false})` is rejected; create and experiment-seed paths are default-only. | **valid but not a defect** | Not fixed, deliberately. Factually right about the schema, but the create path *cannot* serve these specs whatever knob it grows: both tests need two or three **separate** `insert()` calls to cut separate versions, which create-with-items cannot express. Adding an unused parameter to three seed paths is speculative surface on test-only infra. Reply drafted below. |
| 2 | `schemas.py:166` — `num_threads: int` accepts booleans; Pydantic lax mode coerces `true`→`1` before `Dataset.insert`'s own `isinstance(num_threads, bool)` guard can reject it. | **valid, house-wide** | Deferred → driver-wide validation ticket. `num_threads: int = 1` is **pre-existing on main**, unchanged by this PR; `schemas.py` has exactly **1** `Field(` in 400+ lines and no `min_length`/`constr` anywhere. Constraining it only in the field this PR happens to sit next to makes this file the sole exception. Fix it everywhere or nowhere. No spec sends a boolean. |
| 3 | `schemas.py:171` — no unit tests for the `DatasetInsertCall` / `DatasetInsertItemsSessionRequest` boundary (unknown fields at either nesting level, `inserts: []`), "as `AGENTS.md` requires". | **valid, house-wide** | Deferred → same driver-wide ticket. The driver has **no test suite at all** — no `tests/` directory across all 11 route modules. `AGENTS.md:45` addresses product code (Vitest / JUnit / pytest per language); it names no harness for this FastAPI shim. Adding tests for one new boundary and nothing else is not the fix; standing up driver tests is, and that is its own PR. |
| 4 | `dataset-insert-deduplication.spec.ts:66` — missing required `test.info().annotations.push({ type: 'description', … })` per `.agents/skills/playwright-e2e/test-conventions.md`. | **invalid** | Not fixed. The cited file **does not exist** — there is no `.agents/skills/playwright-e2e/` in this repo — and **0 of 63** spec files use `annotations.push`. The estate's actual convention is the file-level docblock, which both specs have and which is unusually thorough here. Applying this would make these two specs the only ones in the suite doing it. Reply drafted below. |
| 5 | Follow-up on #1: commit `906a48c` addressed it for the insert paths, but create-time and live-seed paths still use the default only. | **valid observation, no action** | Correct and expected — default-only there *is* the SDK's own default, which is the behaviour those seed paths should have. Folds into #1's reply. |

**Deferred findings go to one ticket, not three:** *"opik-sdk-driver: request
validation and a test harness"* — covering #2 (field constraints applied
consistently across all 25 request models) and #3 (stand up pytest for the
driver, starting with the schema boundaries). One theme, one ticket.

### Follow-through — say the word and I will do these

1. **File that driver ticket.** Draft above; I have not opened it.
2. **Post replies** on threads 1, 2, 3 and 4 so baz's findings are visibly
   answered rather than silently ignored. Drafts:
   - *#1/#5:* "Right about the schema, but the create path can't serve these
     specs regardless — both tests need separate `insert()` calls to cut
     separate versions, which create-with-items can't express. Default-only on
     the seed paths matches the SDK's own default. Not adding an unused
     parameter to test-only infra."
   - *#2:* "True, and pre-existing — `num_threads: int = 1` is unchanged by this
     PR, and `schemas.py` has one `Field(` in 400+ lines. Constraining it here
     alone would make this the one inconsistent model. Deferred to a driver-wide
     validation ticket."
   - *#3:* "The driver has no test suite at all across its 11 route modules.
     Adding tests for one boundary isn't the fix; standing up pytest for the
     driver is, and that's its own PR. Same ticket as #2."
   - *#4:* "`.agents/skills/playwright-e2e/test-conventions.md` doesn't exist in
     this repo, and 0 of 63 specs use `annotations.push`. The convention here is
     the file-level docblock, which both specs have."
3. **Open the `tsc` fix PR** from Notes 2 and 3 (`tsconfig.json` `baseUrl` +
   the duplicate `deleteDashboard`) — a small PR that makes the type-check gate
   mean something again for the whole estate.

---

## Estate gates

Re-run after the resolution, at `0273278d0b`:

| gate | before | after | |
|---|---|---|---|
| `tag_lint.py --taxonomy … --estate tests_end_to_end` | PASS | **PASS** | 66 specs checked, 1 exempt, 0 problems — including the new Rule 4 sorted-`specs:` check from #8199 |
| `npx tsc --noEmit` | FAIL | **FAIL** | Unchanged and **not this PR's fault**: aborts on `tsconfig.json(13,5) TS5102` before checking any file. Identical on `origin/main`. See Note 2. |
| `npx playwright test --list` | PASS | **PASS** | 129 tests in 63 files; the 3 new tests collect |

**Because the `tsc` gate checks nothing, I type-checked this PR by hand** with a
throwaway TS7-valid config (`baseUrl` dropped, `types: ["node", "js-yaml"]`
added), run before and after the merge. Both runs produce **exactly two errors,
both the pre-existing duplicate `deleteDashboard` in `core/backend/client.ts`,
a file this PR does not touch.** Every file this PR adds or changes type-checks
clean. The throwaway config was deleted and never committed.

---
_[review_generated_tests.yml](https://github.com/comet-ml/comet-automation-tests/actions/runs/34202683967) ·

AndreiCautisanu (approved):

PR #8120, opened 2026-09-02
[NA] [GHA] Remove the orphaned Check Documentation Links workflow

hassaanch23:
## Details

`.github/workflows/test_docs_links.yml` runs a test that was deleted in January, so every scheduled run fails. The schedule is `cron: '0 12 * * *'`, so it has produced a red run daily for roughly seven months, and it is one of the failures currently showing on `main`.

The workflow does two things:

```yaml
- name: Run linkinator
  run: |
    npx linkinator https://www.comet.com/docs/opik/ --recurse --format csv > ${{ github.workspace }}/tests_end_to_end/tests/Documentation/output.csv

- name: Check for broken Comet links
  run: |
    cd ${{ github.workspace }}/tests_end_to_end
    allurectl watch -- pytest -s tests/Documentation/test_links.py ...
```

`tests_end_to_end/tests/` no longer exists. It went away with `495e98ecd` — *"[NA][E2E] Python E2E tests retirement (#4731)"* (2026-01-15) — which retired the Python E2E suite but left this workflow behind. `test_links.py` appears nowhere in the tree today (0 hits across 13,056 paths); the directory is now `coverage/ e2e/ installer_utils/ test-helper-service/ visual-tests/`.

The job therefore dies at the shell redirect, before linkinator can report on a single link:

```
/home/runner/work/opik/opik/tests_end_to_end/tests/Documentation/output.csv: No such file or directory
##[error]Process completed with exit code 1
```

with `Check for broken Comet links` skipped as a result. ([latest run](https://github.com/comet-ml/opik/actions/runs/33630158375/job/100247339009))

### Why remove rather than repair

A `mkdir -p` would get past the redirect, but the pytest step immediately after it invokes a file that does not exist, so the job would still fail — just one step later. The thing this workflow existed to run is gone, and there is nothing to point it back at.

If link checking is still wanted, it needs a new test rather than a patch to this one. Happy to open that instead if you would rather keep the coverage — just say which direction you prefer.

## Change checklist

- [x] Removes a workflow whose target was deleted in #4731
- [x] No source, SDK, or product code touched — single file deleted
- [x] No other workflow references `tests_end_to_end/tests/` or `test_links.py`
- [x] Clears one of the failing checks currently on `main`

## Issues

No ticket — filed as `[NA]` per the PR title convention for untracked cleanup. Related historical context: #4731 (the E2E retirement that removed the test).

## Testing

Nothing to run: the change deletes a workflow file and touches no code.

Verified before opening:
- `test_links.py` — 0 matches in the full `main` tree (13,056 paths)
- `tests_end_to_end/tests/` — does not exist
- Confirmed against the most recent scheduled run that the failure is the missing path, not a broken documentation link

## Documentation

None required. No user-facing behaviour, API, or documented workflow changes.


jverre:
@hassaanch23 You are right, this should be removed ! Thanks for the PR, I'll merge it in :)

jverre (approved):

PR #8128, opened 2026-09-03
[NA] [BE] test: assert the dataset id rather than JDK duplicate-key wording

JetoPistola:
## Details

image

`enrichmentWhenLookupReturnsDuplicateKeyFailsLoudly` asserted `hasMessageContaining("Duplicate key")`, which is wording owned by the JDK's `Collectors.uniqKeysMapAccumulator` rather than by Opik, and it has been reworded across JDK versions before. This swaps that matcher for the dataset id, which is our own value and is carried in the same message. Applies a non-blocking review note from #8076, split out so it did not hold up that PR's merge.

- Test-only, one line: `.hasMessageContaining("Duplicate key")` → `.hasMessageContaining(datasetId.toString())`. The `.isInstanceOf(IllegalStateException.class)` assertion is unchanged.
- The new matcher is strictly stronger: it pins the exception to *this* dataset's duplicate rather than to any duplicate-key failure anywhere in the call.
- No production code changes.

## Change checklist
- [x] User facing
- [x] Documentation update

## Issues

- NA

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: full implementation (one-line test assertion change) + verification
- Human verification: author review of the diff; test + mutation-check output reviewed before opening

## Testing

Run from `apps/opik-backend`:

- `mvn -o test -Dtest=DatasetServiceEnrichmentTest` → `Tests run: 11, Failures: 0, Errors: 0, Skipped: 0`.
- `mvn -o spotless:check` → `BUILD SUCCESS`.

**Message-content check.** Confirmed against the JDK in use (Corretto 25.0.3) that the id really is in the message, rather than assuming it — the duplicate-key `IllegalStateException` reads:

```
Duplicate key 1111...5555 (attempted merging values
ExperimentSummary[datasetId=1111...5555, count=1, at=null] and
ExperimentSummary[datasetId=1111...5555, count=99, at=null])
```

so the id appears both as the formatted key and inside the record's `toString` — two independent occurrences, which is why this matcher is not simply trading one fragile string for another.

**Mutation check — the assertion still discriminates.** A more durable assertion that stopped catching the regression would be worse than the brittle one, so this was verified rather than assumed. Temporarily reverting the three `.collect(toMap(...))` calls in `DatasetService.enrichDatasetWithAdditionalInformation` to `.collectMap(...)` (last-wins) makes the test fail:

```
enrichmentWhenLookupReturnsDuplicateKeyFailsLoudly
java.lang.AssertionError: Expecting code to raise a throwable.
```

Exactly 1 of the 11 tests failed, so the guard is pinned to this behaviour and nothing else in the class depends on it. Worth noting *why* the failure mode is unchanged: last-wins collection throws nothing at all, so the test fails at `assertThatThrownBy` before any message matcher runs. The discriminating power was never in the message wording — which is what makes this swap safe. The mutation was then reverted; `git diff origin/main` confirms `DatasetService.java` is untouched and the diff is the single test line.

Not run: the rest of the backend suite, as this is a test-only assertion change in one class with no production code touched. CI covers it.

## Documentation

N/A — internal test assertion only, no user-facing or API surface change.

### DIFF
--- a/apps/opik-backend/src/test/java/com/comet/opik/domain/DatasetServiceEnrichmentTest.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/domain/DatasetServiceEnrichmentTest.java
@@ -144,7 +144,7 @@ class DatasetServiceEnrichmentTest {
 
         assertThatThrownBy(() -> service.findById(datasetId))
                 .isInstanceOf(IllegalStateException.class)
-                .hasMessageContaining("Duplicate key");
+                .hasMessageContaining(datasetId.toString());
     }
 
     @Test


CometActions:
**No test needed here.**

Test-only: the single changed line is an assertion in DatasetServiceEnrichmentTest, swapping the JDK's "Duplicate key" wording for the dataset id so the test stops depending on a JDK message string. No production code changes, so there is no behaviour for an e2e test to observe.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/33750980829)

thiagohora (approved):

PR #8131, opened 2026-09-03
[OPIK-8230] [BE] fix: scope the trace delete mutation to the batch's own partitions

thiagohora:
## Details

A ClickHouse mutation is registered against **every active part** of the target table before its `WHERE` clause is considered at all — the existing partition predicate (OPIK_6901) prunes which *rows* a matched part rewrites, not which *parts* the mutation visits. Post-cutover `traces_local` holds ~1,900 partitions / ~3,650 parts, so an ordinary delete matching a handful of rows was registered against all of them at ~19 ms/part of fixed overhead. That is what pushed trace deletes past `max_execution_time` (`Code 159 TIMEOUT_EXCEEDED`), where the statement errors and skips the `TracesDeleted` cascade and the deletion-events bridge write even though the mutation completes and the rows are in fact deleted.

This scopes each statement to the partitions its own batch resolves to via `IN PARTITION`, emitting one statement per partition and falling back to today's unbounded form when the batch can't be derived exactly, or when the target table isn't actually partitioned.

- `WeeklyPartitions#groupByPartition` (new) groups ids **by** partition rather than naming their union — `IN PARTITION` accepts exactly one partition per statement, and ClickHouse's `IN PARTITION p1, p2, …` form is for *composite* keys, not a list of distinct partitions.
- The fallback also triggers on an unpartitioned target: the legacy `traces` has no partition key at all, so `IN PARTITION` against it is a hard `Code 248 INVALID_PARTITION_VALUE`, **not** a no-op. `deleteBatch` gates on the existing **`traceColumnsNonNullable`** flag — the EXCHANGE that replaces the `Nullable(...)` columns is the same one that puts the weekly-partitioned successor behind the name mutations target, so the two facts have only ever flipped together. **No new configuration flag**, and so nothing to add to the cutover runbook or tooling.
- Both forms share one SQL template via `<if(partition)>`, matching the pre-diff code's own conditional-fragment pattern rather than duplicating the `WHERE`/`SETTINGS` body across two near-identical constants.
- `WeeklyPartitions#of`, the flat-union predecessor, is removed — it had no production callers left once `TraceDAO` switched over. Its unique boundary-case coverage was migrated to `groupByPartition` rather than dropped.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-8230

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Model(s): Claude Opus 5
- Scope: Implementation, tests, and commit/PR description. Three `/code-review` passes plus a round of review feedback were run against the branch and their findings triaged — fixes applied for the dead `of()` method, a vacuous test assertion, stale `{@link}` references, SQL-template duplication, and a mutable collection escaping `partitionsOf`. Three findings were deliberately **not** fixed and are documented at their call sites instead (see *Known tradeoffs* below).
- Human verification: **Required before merge.** The delete path is production-critical and this changes the SQL shape of every trace delete. Reviewer attention is most valuable on the two accepted tradeoffs below, and on whether criteria 3/4/7 (unverified, see *Testing*) should gate the merge.

## Testing

All run locally against **real ClickHouse via testcontainers** (image `altinity/clickhouse-server:26.3.16.10001.altinitystable`, production's exact version), on the final rebased commit:

```
mvn -q clean spotless:apply test-compile
mvn -q -Dtest=TracesPartitionPruningMutationTest,TracesLegacyTablePruningMutationTest,\
TraceMutationRoutingArchTest,WeeklyPartitionsTest test
```

| Suite | Result |
|---|---|
| `WeeklyPartitionsTest` | 17/17 |
| `TracesPartitionPruningMutationTest` | 8/8 |
| `TracesLegacyTablePruningMutationTest` | 3/3 |
| `TraceMutationRoutingArchTest` | 3/3 |

Scenarios validated:
- **Bounded vs unbounded part footprint** — a scoped delete touches measurably fewer parts than the unbounded fallback measured moments later in the same test.
- **Row-level equivalence** — every era's row is actually gone; a bystander in the same project survives.
- **Far-future id → two partitions** — one such id emits two separate `IN PARTITION` statements (honest week + legacy 32-bit wrap), each `pairs_size=1`.
- **Fallback still deletes** — non-UUIDv7 and past-`DateTime64`-ceiling ids, on both the partitioned and legacy tables.
- **Per-chunk independence** — a batch spanning two `ANALYTICS_DELETE_BATCH_SIZE` chunks derives partitions per chunk, not once per request.
- **Routing guard intact** — the mutation table is still decided in exactly one place.

**Test-infrastructure fix worth flagging:** `TracesPartitionPruningMutationTest` previously asserted pruning via `EXPLAIN`, which reports what the **read planner** selects for a `SELECT` — a different layer from what a **mutation** is registered against. That is precisely the gap this ticket exists to close, so the old coverage passed while every delete still rewrote every part. Rewritten to assert on `system.part_log` `MutatePart` events. Separately, `nonV7IdDisablesPruning` was asserting against a retired regex that the new SQL shape can never match, making it vacuously true; migrated to the same `doesNotContain("IN PARTITION")` check its siblings use.

**Not run, with reason — acceptance criteria 3, 4, and 7 are unverified:**
- **3** (`TracesDeleted` cascade fires again once the statement stops erroring) and **4** (`deletion_events_local` receives its row) need the full cascade path running, not reachable from these suites.
- **7** (clock-skewed trace-delete e2e specs) needs the e2e suite.

These are the criteria that confirm the *user-visible* consequence of the fix, as distinct from the mutation-level mechanism proven above. I'd treat them as a merge gate rather than a follow-up.

### Known tradeoffs — deliberately not fixed here

Both surfaced repeatedly in review; both are documented at their call sites rather than addressed, because fixing either expands beyond this ticket's stated scope:

1. **`TraceService`'s delete→cascade coupling is still all-or-nothing.** A batch can now complete several partitions' statements before one errors, so more of a batch's rows can be synchronously, confirmedly deleted without the cascade firing than before this change. Deletes are idempotent, and the ticket is explicit: *"removes the condition that trips it; does not restructure the coupling."* Scoping each mutation so it completes well inside the timeout is what makes hitting this much rarer.
2. **The gate is a flag read, not a live schema read, and it reuses a flag whose name says only half of what it now gates.** A `traceColumnsNonNullable` flip that precedes the EXCHANGE would turn deletes into `Code 248` rather than degrading to unpruned-but-correct — but that flip already breaks writes outright (a `null` bind against a non-nullable column), so it is not a new precondition, and the reverse order is what the runbook already forbids. Renaming the flag to say both duties is off the table because the env var is exposed; the second duty is documented on `DatabaseAnalyticsDataModelConfig` and at the gate instead.

## Documentation

No documentation update required — internal query-shape change with no user-facing API, schema, or behavioural surface. Rationale that would otherwise live in docs is captured in the `TraceDAO` and `WeeklyPartitions` Javadoc, including why `IN PARTITION` is interpolated rather than bound, why `DELETE FROM` is kept over a hand-written `ALTER … UPDATE`, and why the far-future double-partition case is safe.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
@@ -1967,40 +1967,22 @@ class TraceDAOImpl implements TraceDAO {
             """;
 
     /**
-     * Deletes by the full {@code (workspace_id, project_id, id)} sort key, matching on {@code (project_id, id)} tuples
-     * so a single statement can span several projects (e.g. a reused id resolved to all its owning projects, or a
-     * cross-project batch) instead of one delete per project (OPIK-7483). Every deleted row carries its {@code
-     * project_id}, so no delete is ever project-less - also required once {@code traces} is a Distributed table
-     * (OPIK-7455).
+     * Deletes by the full {@code (workspace_id, project_id, id)} sort key, matching {@code (project_id, id)} tuples so
+     * one statement can span several projects (OPIK-7483). Pairs are bound as two arrays and zipped, so the query text
+     * is constant regardless of batch size.
      * <p>
-     * {@code <if(partitions)>} adds the table's own weekly partition expression, bound as the set of partitions the
-     * batch's ids resolve to. It is emitted whenever every id in the batch is one whose partition can be derived
-     * exactly ({@link WeeklyPartitions#of}); otherwise the predicate is omitted and the statement is byte-identical to
-     * the previous unbounded form. That is what preserves the original guarantee — a row whose {@code id_at} cannot be
-     * trusted is still deleted, because no id in such a batch is used to derive a partition.
+     * {@code <if(partition)>} adds {@code IN PARTITION}, which scopes which <b>parts</b> the mutation is registered
+     * against — a {@code WHERE} clause cannot, since parts are selected before it is considered, which is why a delete
+     * of a few rows rewrote all ~3,650 parts and timed out. Omitted for the unbounded fallback.
      * <p>
-     * No schema flag gates it. {@link WeeklyPartitions} derives a value per {@code id_at} type the mutation may meet
-     * — the legacy 32-bit {@code DateTime} of {@code traces} as well as the {@code DateTime64(0)} of the partitioned
-     * successor — so one rendered statement is correct on both sides of the cutover EXCHANGE, in either direction, with
-     * nothing to flip and nothing to revert on rollback.
-     * <p>
-     * Why it matters: a mutation selects parts at the <b>partition</b> stage, where the (workspace_id, project_id, id)
-     * predicate prunes nothing, so deleting a handful of rows rewrote every part of the table. Measured on prod-test
-     * (271.6 M rows, 3,928 parts): 12 ids rewrote <b>3,928 parts / 5.40 TiB</b>. With this predicate the same batch
-     * selects <b>5</b> parts. An {@code id_at} <em>range</em> is not a substitute: on a batch spanning 1996 and 2200 a
-     * range still selected 2,644 parts, where the exact set selected 4.
-     * <p>
-     * The pairs are bound (never inlined) as two positional string arrays and zipped back into {@code (project_id, id)}
-     * tuples with {@code arrayZip}, so the query text is constant regardless of batch size and no value reaches the SQL
-     * as a literal. {@code arrayZip} is a deterministic function, not a subquery - ClickHouse rejects subqueries in
-     * delete mutations. Callers batch to keep each array within the driver's reliable bind size ({@link
-     * com.comet.opik.infrastructure.FilterUtils#ANALYTICS_DELETE_BATCH_SIZE}).
+     * {@code <partition>} is interpolated, not bound: {@code IN PARTITION {p:UInt32}} is a ClickHouse syntax error. Safe
+     * because the value is always a {@code long} from {@link WeeklyPartitions}.
      */
     private static final String DELETE_BY_PROJECT_ID_TRACE_ID_PAIRS = """
             DELETE FROM <traces_mutation_table>
+            <if(partition)>IN PARTITION <partition><endif>
             WHERE workspace_id = :workspace_id
             AND (project_id, id) IN arrayZip(:project_ids, :trace_ids)
-            <if(partitions)>AND toYYYYMMDD(toDate32(id_at) - toIntervalDay(toDayOfWeek(id_at, 1))) IN :partitions<endif>
             SETTINGS log_comment = '<log_comment>'
             ;
             """;
@@ -2328,7 +2310,7 @@ class TraceDAOImpl implements TraceDAO {
      * {@code :min_id}'s week and inverted the window, so the fast pass matched nothing and every id fell through to
      * the unbounded pass. It now resolves them.
      * <p>
-     * The fallback remains load-bearing for what {@link com.comet.opik.utils.WeeklyPartitions#of} still cannot derive
+     * The fallback remains load-bearing for what {@link com.comet.opik.utils.WeeklyPartitions#groupByPartition} still cannot derive
      * exactly: an id at or past the end of {@code DateTime64}'s range, where {@code id_at} saturates to
      * {@code 2299-12-31} whatever the real week, so every such id collapses into one partition. Real data contains
      * them, so the bounded query is never a delete's sole resolver.
@@ -3645,42 +3627,79 @@ class TraceDAOImpl implements TraceDAO {
     public Mono<Void> delete(Set<Pair<UUID, UUID>> projectIdTraceIdPairs, @NonNull Connection connection) {
         Preconditions.checkArgument(CollectionUtils.isNotEmpty(projectIdTraceIdPairs),
                 "Argument 'projectIdTraceIdPairs' must not be empty");
+        // Checked here rather than where the ids are stringified, so it holds whichever branch deleteBatch takes: the
+        // partitioned path would otherwise read a null as "underivable", silently take the unbounded fallback, and
+        // only then NPE - reporting a caller's bug as the slow delete this class exists to avoid.
+        Preconditions.checkArgument(
+                projectIdTraceIdPairs.stream().noneMatch(pair -> pair.getLeft() == null || pair.getRight() == null),
+                "Argument 'projectIdTraceIdPairs' must not contain null ids");
         log.info("Deleting traces by (project_id, id) pairs, count '{}'", projectIdTraceIdPairs.size());
 
         return makeMonoContextAware((userName, workspaceId) -> Flux
                 .fromIterable(Lists.partition(List.copyOf(projectIdTraceIdPairs), ANALYTICS_DELETE_BATCH_SIZE))
-                .concatMap(batch -> {
-                    var template = getSTWithLogComment(DELETE_BY_PROJECT_ID_TRACE_ID_PAIRS, "delete_traces",
-                            workspaceId,
-                            userName, "pairs_size=%s".formatted(batch.size()));
-                    selectTracesMutationTable(template);
-
-                    var projectIds = batch.stream().map(pair -> pair.getLeft().toString()).toArray(String[]::new);
-                    var traceIds = batch.stream().map(pair -> pair.getRight().toString()).toArray(String[]::new);
-
-                    // Prune to the batch's own partitions when every id in the batch allows it; otherwise emit the
-                    // unbounded form. Needs no schema flag: WeeklyPartitions derives a value per id_at type the
-                    // mutation may meet, so the set is correct on the legacy traces and on the partitioned successor.
-                    var partitions = WeeklyPartitions.of(batch.stream().map(Pair::getRight).toList());
-                    // Flag only, exactly like distributed_wrap: the values reach ClickHouse via the bind below,
-                    // never through the template, so the rendered SQL is constant regardless of batch contents.
-                    partitions.ifPresent(_ -> template.add("partitions", true));
-
-                    var statement = connection.createStatement(template.render())
-                            .bind("workspace_id", workspaceId)
-                            .bind("project_ids", projectIds)
-                            .bind("trace_ids", traceIds);
-
-                    if (partitions.isPresent()) {
-                        statement = statement.bind("partitions", partitions.get().toArray(Long[]::new));
-                    }
-
-                    var segment = startSegment("traces", "Clickhouse", "delete");
-                    return Mono.from(statement.execute())
-                            .doFinally(_ -> endSegment(segment))
-                            .then();
-                })
+                .concatMap(batch -> deleteBatch(batch, workspaceId, userName, connection))
                 .then());
+    }
+
+    /**
+     * Deletes one batch: one {@code IN PARTITION} statement per partition its ids resolve to, or a single unbounded
+     * statement when they cannot all be derived, or when the target is not partitioned.
+     * <p>
+     * Sequential on purpose: bounded concurrency was measured and deferred (OPIK-8230).
+     */
+    private Mono<Void> deleteBatch(List<Pair<UUID, UUID>> batch, String workspaceId, String userName,
+            Connection connection) {
+        // traceColumnsNonNullable doubles as "the mutation target is weekly-partitioned": the same cutover EXCHANGE
+        // drops the Nullable(...) columns and puts the partitioned successor behind the name mutations target, so one
+        // flag carries both facts. Deliberately not the wrap flag, which governs routing and is still false in the
+        // window between the EXCHANGE and the wrap - reading that one leaves production's deletes unpruned.
+        var grouped = traceColumnsNonNullable()
+                ? WeeklyPartitions.groupByPartition(batch.stream().map(Pair::getRight).toList())
+                : Optional.<Map<Long, Set<UUID>>>empty();
+
+        if (grouped.isEmpty()) {
+            return executeDelete(batch, null, workspaceId, userName, connection);
+        }
+
+        // id -> its pairs, built once per batch rather than rescanning the whole batch once per partition: an id can
+        // map to more than one pair (the same trace id reused across projects, OPIK-7483), so this is a
+        // Collectors.groupingBy, not a plain lookup map.
+        var pairsById = batch.stream().collect(Collectors.groupingBy(Pair::getRight));
+
+        return Flux.fromIterable(grouped.get().entrySet())
+                .concatMap(entry -> {
+                    var partitionPairs = entry.getValue().stream()
+                            .flatMap(id -> pairsById.get(id).stream())
+                            .toList();
+                    return executeDelete(partitionPairs, entry.getKey(), workspaceId, userName, connection);
+                })
+                .then();
+    }
+
+    /**
+     * Renders and executes one delete statement — unbounded when {@code partition} is null, scoped to it otherwise.
+     */
+    private Mono<Void> executeDelete(List<Pair<UUID, UUID>> pairs, Long partition,
+            String workspaceId, String userName, Connection connection) {
+        var template = getSTWithLogComment(DELETE_BY_PROJECT_ID_TRACE_ID_PAIRS, "delete_traces", workspaceId,
+                userName, "pairs_size=%s".formatted(pairs.size()));
+        selectTracesMutationTable(template);
+        if (partition != null) {
+            template.add("partition", partition);
+        }
+
+        var projectIds = pairs.stream().map(pair -> pair.getLeft().toString()).toArray(String[]::new);
+        var traceIds = pairs.stream().map(pair -> pair.getRight().toString()).toArray(String[]::new);
+
+        var statement = connection.createStatement(template.render())
+                .bind("workspace_id", workspaceId)
+                .bind("project_ids", projectIds)
+                .bind("trace_ids", traceIds);
+
+        var segment = startSegment("traces", "Clickhouse", "delete");
+        return Mono.from(statement.execute())
+                .doFinally(_ -> endSegment(segment))
+                .then();
     }
 
     /**
--- a/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/DatabaseAnalyticsDataModelConfig.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/infrastructure/DatabaseAnalyticsDataModelConfig.java
@@ -15,6 +15,13 @@ import lombok.Builder;
  * epoch end time round-trips unchanged rather than being read as {@code null}. Flip this in lockstep with the EXCHANGE
  * step of the cutover.</p>
  *
+ * <p>It also gates partition-scoped deletes (OPIK-8230): the EXCHANGE that makes these columns non-nullable is the
+ * same one that puts the weekly-partitioned successor behind the name mutations target, so this flag being
+ * {@code true} is equally what says a delete may scope itself with {@code IN PARTITION}. One flag for two facts
+ * because they have only ever flipped together; a second would have to be threaded through the cutover runbook and
+ * tooling to track no independent state. The name says only the first duty and is deliberately not renamed - the env
+ * var is exposed.</p>
+ *
  * <p>{@code spanColumnsNonNullable}: the {@code spans} sibling of {@code traceColumnsNonNullable}, gating the same
  * sentinel wiring for {@code spans.end_time}→epoch and {@code spans.duration}/{@code spans.ttft}→{@code NaN}. Default
  * {@code false} while the {@code spans} table still has {@code Nullable(...)} columns; set {@code true} in lockstep with
--- a/apps/opik-backend/src/main/java/com/comet/opik/utils/WeeklyPartitions.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/utils/WeeklyPartitions.java
@@ -9,14 +9,19 @@ import java.time.LocalDate;
 import java.time.ZoneOffset;
 import java.time.temporal.TemporalAdjusters;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
+import java.util.stream.Collectors;
 
 /**
- * Single derivation point for the {@code id_at} weekly partition values a batch of ids resolves to, so a mutation can
- * name its own partitions instead of being planned against every part of the table.
+ * Single derivation point for the {@code id_at} weekly partition value(s) a batch of ids resolves to, so a mutation
+ * can name its own partitions instead of being planned against every part of the table. {@link #groupByPartition}
+ * groups ids by the partition each belongs to, for a caller emitting one statement per partition (OPIK-8230) that
+ * must bind only the ids belonging to it.
  *
  * <p>Mirrors the partition expression of {@code traces_local_v2} / {@code spans_local_v2} exactly —
  * {@code PARTITION BY toYYYYMMDD(toDate32(id_at) - toIntervalDay(toDayOfWeek(id_at, 1)))}, where {@code id_at} is
@@ -55,9 +60,9 @@ import java.util.UUID;
  * <h2>Why the caller must treat an empty result as "no predicate", never as "no partitions"</h2>
  *
  * <p>The whole value of the derivation is that the partitions a batch resolves to are the <em>only</em> places its rows
- * can be; a set that is merely close is a silently skipped delete, not a slower one. So this returns a set only when
- * every id in the batch is one it can derive exactly, and empty otherwise — leaving the caller to emit its unbounded
- * form, which is always correct and merely slower. The two rejections are:</p>
+ * can be; a grouping that is merely close is a silently skipped delete, not a slower one. So this returns non-empty
+ * only when every id in the batch is one it can derive exactly, and empty otherwise — leaving the caller to emit its
+ * unbounded form, which is always correct and merely slower. The two rejections are:</p>
  * <ul>
  *     <li><b>Any id that is not a UUIDv7.</b> Its high 48 bits are not a timestamp, and {@code UUIDv7ToDateTime}
  *     returns {@code 1970-01-01} for it rather than throwing, so the row sits in the epoch partition while the bits
@@ -101,49 +106,80 @@ public class WeeklyPartitions {
     private static final long ID_AT_LEGACY_MODULUS = 1L << 32;
 
     /**
-     * The weekly partition values the ids resolve to — under each {@code id_at} type the mutation may run against (see
-     * the class javadoc) — or empty if the batch contains an id whose partition cannot be derived exactly. In the empty
-     * case the caller must omit its partition predicate entirely. An empty batch yields empty for the same reason:
-     * there is nothing to bound the mutation to.
+     * Groups the ids by the weekly partition value(s) each resolves to — under each {@code id_at} type the mutation
+     * may run against (see the class javadoc) — or empty if the batch contains an id whose partition cannot be
+     * derived exactly. In the empty case the caller must omit its partition predicate entirely and emit the
+     * unbounded form. An empty batch yields empty for the same reason: there is nothing to bound the mutation to.
+     * {@code IN PARTITION} accepts exactly one partition per statement (OPIK-8230), so a caller emitting one
+     * statement per partition needs to know which ids go in each, not merely their union.
      * <p>
-     * A {@code null} batch throws rather than reading as empty, which is the one place this class is deliberately
-     * intolerant. Empty is a <em>documented answer</em> — "this batch cannot be pruned, emit the unbounded form" — and a
-     * caller that lost its batch would receive that answer, silently issue a correct-but-unbounded mutation, and never
-     * learn it had a bug. A null collection here is a programming error, not a data condition; a null <em>element</em>
-     * is the data condition, and that keeps returning empty.
+     * An id past {@link #ID_AT_LEGACY_MODULUS} appears in <b>two</b> groups — one per {@code id_at} representation
+     * it resolves to. That is safe, not a widening bug: a statement scoped to the group that does not actually
+     * contain the id's row is a no-op there, never a wrong deletion, since the row-matching {@code (project_id, id)}
+     * predicate inside each statement is unchanged regardless of which partition the statement targets.
+     * <p>
+     * A {@code null} batch throws rather than reading as empty: empty is a documented answer meaning "emit the
+     * unbounded form", so a caller that lost its batch would silently get back the slow mutation this class exists to
+     * avoid. A null <em>element</em> still returns empty, which is safe because it is unreachable from the delete
+     * path: {@code TraceDAO#delete} rejects a null id before any of this runs.
      *
      * @throws NullPointerException if {@code ids} is null.
      */
-    public static Optional<Set<Long>> of(@NonNull Collection<UUID> ids) {
-        var partitions = new HashSet<Long>();
+    public static Optional<Map<Long, Set<UUID>>> groupByPartition(@NonNull Collection<UUID> ids) {
+        var grouped = new HashMap<Long, Set<UUID>>();
 
         for (UUID id : ids) {
-            if (id == null || id.version() != 7) {
+            var idPartitions = partitionsOf(id);
+            if (idPartitions.isEmpty()) {
                 return Optional.empty();
             }
-            // UUIDv7: the high 48 bits are the unix epoch in milliseconds. `>>> 16` reads them unsigned, so the value
-            // is in [0, 2^48) — never negative, and never large enough for Instant.ofEpochSecond to overflow. That is
-            // also why neither derivation checks a floor: the smallest id_at any UUIDv7 can carry is the epoch, and
-            // even its Monday (1969-12-29) is comfortably inside Date32 on both schemas.
-            long epochMilli = id.getMostSignificantBits() >>> 16;
-            if (epochMilli >= ID_AT_CEILING) {
-                return Optional.empty();
-            }
-            // Truncating to whole seconds is the column's own conversion (DateTime64(0) / DateTime both store seconds)
-            // and cannot move a value into an earlier day, so it never changes the week.
-            long epochSecond = epochMilli / 1_000L;
-            partitions.add(weeklyPartitionOf(epochSecond)); // DateTime64(0, 'UTC') — the partitioned successor
-            // Only past the 32-bit range do the two columns disagree; below it the modulo is the identity, so the
-            // branch is the invariant stated as code rather than an optimisation: an ordinary id CANNOT widen the set.
-            if (epochSecond >= ID_AT_LEGACY_MODULUS) {
-                partitions.add(weeklyPartitionOf(epochSecond % ID_AT_LEGACY_MODULUS)); // DateTime('UTC') — legacy
+            for (Long partition : idPartitions.get()) {
+                grouped.computeIfAbsent(partition, _ -> new HashSet<>()).add(id);
             }
         }
 
-        // Set.copyOf, not the working HashSet: what escapes here decides which partitions a DELETE mutation touches, so
-        // a caller holding a mutable reference could narrow the set after it was derived and turn a correct delete into
-        // a silent no-op. Immutable by default per apps/opik-backend/AGENTS.md, and the accumulator stays local.
-        return partitions.isEmpty() ? Optional.empty() : Optional.of(Set.copyOf(partitions));
+        if (grouped.isEmpty()) {
+            return Optional.empty();
+        }
+        // Immutable outer map AND immutable per-partition sets: what escapes here decides which partitions a DELETE
+        // mutation touches, so a caller holding a mutable reference could narrow it after derivation and turn a
+        // correct delete into a silent no-op. Collectors.toUnmodifiableMap's values must themselves be made
+        // unmodifiable explicitly; it does not do so for you.
+        return Optional.of(grouped.entrySet().stream()
+                .collect(Collectors.toUnmodifiableMap(Map.Entry::getKey, entry -> Set.copyOf(entry.getValue()))));
+    }
+
+    /**
+     * The weekly partition value(s) a single id resolves to under each {@code id_at} type the mutation may meet (see
+     * class javadoc) — extracted from {@link #groupByPartition} so the per-id {@code id_at} math reads separately
+     * from the grouping it feeds. Empty exactly when the id cannot be derived exactly: not a UUIDv7, or an embedded
+     * timestamp at or past {@link #ID_AT_CEILING}.
+     */
+    private static Optional<Set<Long>> partitionsOf(UUID id) {
+        if (id == null || id.version() != 7) {
+            return Optional.empty();
+        }
+        // UUIDv7: the high 48 bits are the unix epoch in milliseconds. `>>> 16` reads them unsigned, so the value
+        // is in [0, 2^48) — never negative, and never large enough for Instant.ofEpochSecond to overflow. That is
+        // also why neither derivation checks a floor: the smallest id_at any UUIDv7 can carry is the epoch, and
+        // even its Monday (1969-12-29) is comfortably inside Date32 on both schemas.
+        long epochMilli = id.getMostSignificantBits() >>> 16;
+        if (epochMilli >= ID_AT_CEILING) {
+            return Optional.empty();
+        }
+        // Truncating to whole seconds is the column's own conversion (DateTime64(0) / DateTime both store seconds)
+        // and cannot move a value into an earlier day, so it never changes the week.
+        long epochSecond = epochMilli / 1_000L;
+        // DateTime64(0, 'UTC') — the partitioned successor. Only past the 32-bit range do the two columns disagree;
+        // below it the modulo is the identity, so the branch is the invariant stated as code rather than an
+        // optimisation: an ordinary id CANNOT widen the set. Immutable on the way out, for the same reason
+        // groupByPartition copies its own result — see there. Set.of throws on a duplicate, which is safe here
+        // rather than merely untested: the two arguments are weeks 2^32 seconds (~136 years) apart, so they cannot
+        // be the same week.
+        return Optional.of(epochSecond >= ID_AT_LEGACY_MODULUS
+                ? Set.of(weeklyPartitionOf(epochSecond),
+                        weeklyPartitionOf(epochSecond % ID_AT_LEGACY_MODULUS)) // DateTime('UTC') — legacy
+                : Set.of(weeklyPartitionOf(epochSecond)));
     }
 
     /**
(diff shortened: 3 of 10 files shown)


CometActions:

The IN PARTITION scoping is gated on traceColumnsNonNullable, which is false on a fresh OSS install (config.yml:125, docker-compose default) and never set by the e2e estate — and flipping it there would not help: the OSS traces table has no PARTITION BY, so the scoped form is a hard Code 248 and ClickHouseTracesTopologyHealthCheck would pull the install from rotation. The timeout this fixes also needs production's ~1,900-partition / ~3,650-part shape, which an e2e env cannot seed. So there is nothing to observe end to end today; recording it as an estate gap rather than closing it, because a wrong partition grouping fails silently — the delete returns 200 and the rows are still there — and because the timeout it replaces was skipping the TracesDeleted cascade and the deletion-events bridge write. Worth a test the day an env runs the post-EXCHANGE partitioned traces with the flag on.

**Testable once (platform):** An env whose traces mutation target is the post-EXCHANGE weekly-partitioned table (traces_local_v2, PARTITION BY toYYYYMMDD(toDate32(id_at) - toIntervalDay(toDayOfWeek(id_at, 1)))) with ANALYTICS_DB_DATA_MODEL_TRACE_COLUMNS_NON_NULLABLE=true and clickhouse-traces-topology green. The seeding half already exists: e2e/fixtures/id-aged-traces.fixture.ts mints ids at chosen instants, so a batch spanning several weekly partitions — including the ~2201 id that resolves into two partition groups — is producible without new fixture work.

**What a test would assert:** Bulk delete on the Logs table and DELETE /v1/private/traces. The axis is the deployment's data-model config, not the request: with ANALYTICS_DB_DATA_MODEL_TRACE_COLUMNS_NON_NULLABLE=true against the weekly-partitioned successor, a delete emits one IN PARTITION statement per partition its batch resolves to; with it false, one unbounded statement. Two ways a user sees it go wrong on the scoped side — rows that report deleted but reappear on reload (a partition the batch's ids were mis-grouped into is a no-op, and the statement still returns success), and, since deleteBatch is all-or-nothing across several statements, a mid-batch failure leaving earlier partitions' traces gone while their spans and feedback scores survive because the TracesDeleted cascade never ran.

**Also already tested.** The ungated half is covered. With the flag off — every OSS install and the e2e estate — this PR drops the OPIK-6901 WHERE-clause partition predicate and takes the unbounded fallback on every delete, through the rewritten deleteBatch/executeDelete. Same rows match, since the (project_id, id) tuple predicate is unchanged, and trace-explore/trace-delete.spec.ts exercises exactly that path on both the Logs bulk delete and DELETE /v1/private/traces, asserting the deleted traces are gone from the table and the API and the survivor is not. It would fail if the refactor bound the wrong pairs or broke the template render.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/33883859995)

_Re-checked after a push on 04 Sep 14:29 UTC — nothing the verdict depends on changed._

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/TracesPartitionPruningMutationTest.java
```diff
@@ -846,48 +813,59 @@ void deleteClearsEveryEraAndBindsExactlyThosePartitions() {
...
-        assertThat(bounded.selected())
-                .as("the bounded delete selects fewer parts than the table holds: %s", bounded)
-                .isLessThan(bounded.total());
-        // Shown to discriminate, or it proves nothing - the same trap as `.contains(partition)` and
-        // `doesNotContain("toDayOfWeek")`. The fallback must not prune: either the planner reports no partition index at
-        // all, because nothing filters on the key, or it reports every part still selected.
-        assertThat(unbounded.map(parts -> parts.selected() == parts.total()).orElse(true))
-                .as("the fallback prunes nothing: %s", unbounded)
-                .isTrue();
+        var bounded = mutatePartActivitySince(table, boundedSince);
```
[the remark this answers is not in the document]
thiagohora: Fixed. `mutatePartActivitySince` now waits for the table's mutations to settle **after** each delete before snapshotting, keeping `since` captured before submission as you suggested — `SYSTEM FLUSH LOGS` only flushes what `part_log` already holds, it doesn't wait for `MutatePart` rows to be produced.

While verifying, I also chased the `partitions == 1` assertion properly rather than weakening it again. It **passes when the test runs in isolation**, so the scoping genuinely does hit exactly one partition; in-suite it reads 3 because every test here shares one table (`PER_CLASS`, see class Javadoc) and ClickHouse applies siblings' already-`is_done` mutations inside the window — which settling cannot prevent. So the in-suite assertion stays the discriminating bounded-vs-unbounded comparison, with that reasoning now recorded in the comment instead of left as a guess. (Two other hypotheses I tested and discarded first: incomplete snapshots, and JVM-vs-container clock skew.)

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -3649,38 +3684,103 @@ public Mono<Void> delete(Set<Pair<UUID, UUID>> projectIdTraceIdPairs, @NonNull C
...
+     * this change (OPIK-8230) — a real gain in the rare many-partition tail, but a knob to hold in reserve rather than
+     * ship alongside the fix itself.
+     */
+    private Mono<Void> deleteBatch(List<Pair<UUID, UUID>> batch, String workspaceId, String userName,
+            Connection connection) {
+        // Resolved once, via the resolver itself - never the flag directly (see the method Javadoc for why).
+        var table = tracesMutationTable();
+        var grouped = TRACES_LOCAL_TABLE.equals(table)
+                ? WeeklyPartitions.groupByPartition(batch.stream().map(Pair::getRight).toList())
+                : Optional.<Map<Long, Set<UUID>>>empty();
```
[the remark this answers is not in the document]
thiagohora: Correct, and this was live rather than hypothetical — thank you, this is the most valuable finding on the PR.

Two concrete failure modes the previous name-check gate had:

1. **Wrap-on without the EXCHANGE is a real, supported topology.** `TracesDistributedWrapMutationTest` exercises exactly it: the wrap is applied but no EXCHANGE runs, so `traces_local` is the *legacy unpartitioned* table. `IN PARTITION` against it returned `Code 248` and failed CI on the previous revision.
2. **Worse, the inverse in production.** Post-EXCHANGE/pre-wrap, `traces` is already the partitioned successor while the wrap flag is still `false`, so the gate resolved to the unwrapped name and took the fallback on *every* delete — the fix would have merged, deployed, and done nothing in exactly the state production is in.

Fixed by your first option: a dedicated `tracesPartitionedMutationsEnabled` flag, independent of `tracesDistributedWrapEnabled` (routing is not shape), defaulting false, and doubling as a kill-switch since turning it off falls back to the unbounded statement.

I chose the flag over validating the live partition contract because that would mean a schema read on the delete path per batch. Readiness (`ClickHouseTracesTopologyHealthCheck`) is arguably where a contract check belongs if we want one — happy to follow up there if you would prefer belt-and-braces.

Verified by negative control: with the flag off, the emitted SQL carries no `IN PARTITION` and the pruning suite fails — so the gate demonstrably gates.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -3649,38 +3684,103 @@ public Mono<Void> delete(Set<Pair<UUID, UUID>> projectIdTraceIdPairs, @NonNull C
...
-                    // Flag only, exactly like distributed_wrap: the values reach ClickHouse via the bind below,
-                    // never through the template, so the rendered SQL is constant regardless of batch contents.
-                    partitions.ifPresent(_ -> template.add("partitions", true));
-
-                    var statement = connection.createStatement(template.render())
-                            .bind("workspace_id", workspaceId)
-                            .bind("project_ids", projectIds)
-                            .bind("trace_ids", traceIds);
+                .concatMap(batch -> deleteBatch(batch, workspaceId, userName, connection))
+                .then());
```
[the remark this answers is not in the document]
thiagohora: Agreed — this is a genuine gap, not a disagreement.

It maps to acceptance criteria 3 and 4, which the PR description already flags as unverified: the DAO-level fan-out is covered, but the service-level assertion that a multi-partition delete publishes one `TracesDeleted` and captures every `deletion_events_local` row is not. Those criteria are precisely the ones confirming the *user-visible* consequence of the fix, as distinct from the mutation-level mechanism the current tests prove.

You reaching this independently is a fair signal it should gate the merge rather than be deferred. I have not added it here because it needs the full cascade path running, which these suites do not stand up.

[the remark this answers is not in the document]
thiagohora: Real, and deliberately not fixed here — documented at the call site rather than left implicit.

The ticket is explicit on this boundary: *"this ticket removes the condition that trips it; it does not restructure the coupling."* The fan-out does widen the window, and the PR description says so directly: earlier partitions' statements can complete and physically remove rows before a later one errors, at which point no `TracesDeleted` fires for any pair in the batch — not just the failed one.

What makes it tolerable rather than ignored: deletes are idempotent, so a retry converges; and scoping each mutation so it completes well inside `max_execution_time` is what makes reaching this path much rarer than it is today, where essentially every delete times out.

On the test suggestion — a distinct-partition partial-success case asserting the first row is deleted while no event or bridge row is emitted — that is a fair ask and I have not added it. It overlaps the multi-partition cascade coverage gap in the sibling comment, and I would rather see both addressed together than add a test that pins current behaviour we may want to change.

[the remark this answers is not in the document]
thiagohora: Deliberate, and named in the ticket's own rejected-alternatives section: bounded concurrency was measured and held back as *"a knob to hold in reserve, not to ship"* alongside the fix itself.

Worth being precise about the direction of travel, though, because this change makes the load profile *better* rather than worse in the common case. Today a single delete is one statement registered against ~3,650 parts and reliably times out; afterwards it is typically 1–2 statements against a handful of parts each. Every production batch inspected for the ticket spanned 1–2 weeks. The pathological fan-out needs a batch whose ids scatter across many weeks — which in practice means a project polluted with far-future ids.

So I would rather not add a delete-specific rate limit in this PR: it would be sized against a tail case the fix is simultaneously making rarer, and `ANALYTICS_DELETE_BATCH_SIZE` plus the `@Size(max = 1000)` bounds on the request already cap the input. If the tail proves real once this is deployed and measurable, bounded concurrency is the ready-made lever and a much better-informed change then.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -3649,38 +3684,103 @@ public Mono<Void> delete(Set<Pair<UUID, UUID>> projectIdTraceIdPairs, @NonNull C
...
+        // Collectors.groupingBy, not a plain lookup map.
+        var pairsById = batch.stream().collect(Collectors.groupingBy(Pair::getRight));
+
+        return Flux.fromIterable(grouped.get().entrySet())
+                .concatMap(entry -> {
+                    var partitionPairs = entry.getValue().stream()
+                            .flatMap(id -> pairsById.get(id).stream())
+                            .toList();
+                    return executeDelete(table, partitionPairs, entry.getKey(), workspaceId, userName, connection);
                 })
```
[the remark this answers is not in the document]
thiagohora: Same root as the two sibling findings, so answering the part specific to this one: durable per-partition completion tracking.

Not added, and I would push back gently on it being the right shape here. Durably tracking per-partition progress against the originally resolved set means persisting delete state — new storage and a repair path — to compensate for a coupling the ticket explicitly scopes out of this change. That is a larger design decision than a delete-performance fix should make on its own.

The narrower reading is also worth stating: the orphan risk you describe is real but pre-existing rather than introduced here. A single unbounded statement could already time out after ClickHouse had done the work, leaving the same rows deleted with no event and no bridge entry — that is the very failure OPIK-8230 exists to stop. What the fan-out changes is the granularity, not the existence, of the gap.

If we do want durable tracking, I would rather it be designed against the cascade coupling as a whole (covering the single-statement case too) than bolted onto the fan-out.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/DatabaseAnalyticsDataModelConfig.java
```diff
@@ -64,5 +72,6 @@ public record DatabaseAnalyticsDataModelConfig(
         boolean traceDeletionEventsCaptureEnabled,
         boolean spanDeletionEventsCaptureEnabled,
         @Min(1) @Max(2_000) int deletionEventsInsertBatchSize,
-        boolean tracesDistributedWrapEnabled) {
+        boolean tracesDistributedWrapEnabled,
+        boolean tracesPartitionedMutationsEnabled) {
```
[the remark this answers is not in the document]
thiagohora: Good catch, and half of it was a real defect — fixed in 98a7ca5.

To split the finding: `tracesDistributedWrapEnabled` *is* already bound (`config.yml:160`, plus Helm values/ConfigMap and docker-compose). But `tracesPartitionedMutationsEnabled` — which I added in the previous revision — was bound **only** in `config-test.yml`. So it existed in the record, defaulted `false`, and had no production surface to switch it on. The flag was unreachable, which would have left this fix inert in production for a second, independent reason on top of the one it was introduced to fix.

Now bound across the same five surfaces as its sibling:

- `apps/opik-backend/config.yml` — `${ANALYTICS_DB_DATA_MODEL_TRACES_PARTITIONED_MUTATIONS_ENABLED:-false}`
- `deployment/helm_chart/opik/values.yaml`
- `deployment/helm_chart/opik/templates/configmap-backend.yaml`
- `deployment/helm_chart/opik/tests/configmap_env_test.yaml` (asserts the rendered default)
- `deployment/docker-compose/docker-compose.yaml`

Defaults stay `false` everywhere, which is deliberate rather than an oversight: `IN PARTITION` against an unpartitioned table is a hard `Code 248`, so the safe default is off until the EXCHANGE has run. That is the opposite of the wrap flag's readiness contract — `ClickHouseTracesTopologyHealthCheck` fails startup if that one disagrees with the live engine, whereas this one degrades to the unbounded delete, so a stale `false` costs performance rather than availability.

On the startup-validation half of your question: I have not added a live partition-contract check, for the reason in the sibling thread — it would mean a schema read on the delete path. Readiness is the better home for it if we want one, and I am happy to follow up there.

Pre/post-EXCHANGE coverage is exercised: `TracesPartitionPruningMutationTest` (flag on, partitioned), `TracesLegacyTablePruningMutationTest` (legacy unpartitioned), and `TracesDistributedWrapMutationTest` (wrap on, *not* partitioned — the topology that produced the `Code 248` CI failure). 35/35 locally.

apps/opik-backend/src/test/resources/config-test.yml
```diff
@@ -129,6 +129,13 @@ databaseAnalyticsDataModel:
   #   both `traces_local` and `traces`, else reads can't see the column (code 47).
   tracesDistributedWrapEnabled: false
   #  Default: false
+  #  Description: Whether the table trace mutations target is weekly partitioned, so a delete may scope itself with
+  #   IN PARTITION (OPIK-8230). Separate from tracesDistributedWrapEnabled, which governs routing rather than shape:
```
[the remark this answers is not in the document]
thiagohora: Agreed — adopted your phrasing in 98a7ca5. "the table trace mutations target" was a reduced relative clause that reads as a garden path; "the table targeted by trace mutations is weekly-partitioned" is unambiguous.

Applied consistently across all four places the description appears: `DatabaseAnalyticsDataModelConfig`, the `TraceDAO` accessor, `config-test.yml` and `config.yml`, with `weekly-partitioned` hyphenated as a compound adjective throughout.

One instance left alone deliberately: `TracesPartitionPruningMutationTest:74` has pre-existing prose using the unhyphenated form. It predates this PR and is not part of this change, so I left it rather than widen the diff.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -3460,6 +3442,15 @@ private boolean tracesDistributedWrapEnabled() {
         return configuration.getDatabaseAnalyticsDataModel().tracesDistributedWrapEnabled();
     }
 
+    /**
+     * Whether the table targeted by trace mutations is weekly-partitioned, so a delete may scope itself with {@code IN PARTITION}.
+     * Separate from {@link #tracesDistributedWrapEnabled()}, which governs routing rather than shape: the two diverge
+     * between the EXCHANGE and the wrap.
+     */
+    private boolean tracesPartitionedMutationsEnabled() {
```
andrescrz: We should avoid a new configuration flag at all cost. This would require revisiting and updating the whole cutover procedure in order to incorporate it to the runbook, tooling etc. We cannot afford the cost and also it's desirable to eliminate complexity from the system.

You have two options:
1. Try to find a a neutral query that doesn't require an ST template to add `IN PARTITION` or not, maybe by easily deducting the partition expression or some other alternative approach. Careful with this solution, do not introduce additional complexity, like low level checks on the table.
2. Just reuse an existing flag for the cutover when adding partitioning. Maybe `traceColumnsNonNullable`. The name is not significative and it's desirable to not rename it as the env var is exposed. But it could be good enough, maybe by updating the documentation. If columns are non nullable, it means we're in the partitioned table.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -3649,38 +3640,75 @@ public Mono<Void> delete(Set<Pair<UUID, UUID>> projectIdTraceIdPairs, @NonNull C
 
         return makeMonoContextAware((userName, workspaceId) -> Flux
                 .fromIterable(Lists.partition(List.copyOf(projectIdTraceIdPairs), ANALYTICS_DELETE_BATCH_SIZE))
-                .concatMap(batch -> {
-                    var template = getSTWithLogComment(DELETE_BY_PROJECT_ID_TRACE_ID_PAIRS, "delete_traces",
-                            workspaceId,
-                            userName, "pairs_size=%s".formatted(batch.size()));
-                    selectTracesMutationTable(template);
```
andrescrz, line 3656: Minor: I believe you lost reusing `selectTracesMutationTable`.

apps/opik-backend/src/main/java/com/comet/opik/utils/WeeklyPartitions.java
```diff
@@ -114,36 +126,61 @@ public class WeeklyPartitions {
      *
      * @throws NullPointerException if {@code ids} is null.
      */
-    public static Optional<Set<Long>> of(@NonNull Collection<UUID> ids) {
-        var partitions = new HashSet<Long>();
+    public static Optional<Map<Long, Set<UUID>>> groupByPartition(@NonNull Collection<UUID> ids) {
```
andrescrz, line 128: Minor: if an optional is returned, maybe ok to drop the non-null validation and just return an empty optional if `ids` is empty or null. Feel free to keep the loud failure if you think it fits better.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/TracesPartitionPruningMutationTest.java
```diff
@@ -449,23 +457,70 @@ private static long partitionNameOf(LocalDate monday) {
...
+     * they span - the part-level proof that a delete actually scoped its mutation, not merely that its {@code WHERE}
+     * clause looks right. See {@link #MUTATE_PART_EVENTS_SINCE}'s Javadoc for why this is windowed rather than
+     * correlated by {@code query_id}.
+     * <p>
+     * {@code SYSTEM FLUSH LOGS} first: {@code system.part_log} is buffered and flushes on its own schedule, so a read
+     * immediately after a delete can race the write - the same reason {@link #lastTraceDeleteSql} flushes before
+     * reading {@code system.query_log}.
+     */
+    private MutatePartActivity mutatePartActivitySince(String table, Instant since) {
+        execute("SYSTEM FLUSH LOGS", _ -> {
```
andrescrz: Let's avoid `system flush` calls, these increases the risk of flakiness. Better find an alternative that doesn't rely on this for testing this.

Anyway, not a blocker as this was happening on the tests already.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/TracesPartitionPruningMutationTest.java
```diff
@@ -826,17 +800,24 @@ void deleteClearsEveryEraAndBindsExactlyThosePartitions() {
...
+        // One statement per partition the batch resolves to (OPIK-8230), not one statement carrying all four values:
+        // 1996 and 2025 are inside the 32-bit range and name one week each; the 2199 id names two (its own week AND
+        // its legacy wrap), so it is the id bound into two DIFFERENT statements rather than the batch widening to a
+        // fourth id. Four statements, each pairs_size=1 - still an exact set of partitions, and still not a range
+        // across two centuries, which is the property under test.
+        var sqls = deleteSqlsSince(since, projectId);
+        assertThat(sqls)
+                .as("one statement per partition the batch resolves to: %s", sqls)
+                .hasSize(4)
+                .allSatisfy(sql -> assertThat(sql).contains("pairs_size=1"));
```
andrescrz, line 848: Minor: I'd test that each partition has multiple pairs, if not too complicated.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/TracesPartitionPruningMutationTest.java
```diff
@@ -846,48 +827,71 @@ void deleteClearsEveryEraAndBindsExactlyThosePartitions() {
...
+                .as("the bounded delete touched at least the partition it resolves to: %s", bounded)
+                .isGreaterThanOrEqualTo(1);
+
+        // Unbounded: a non-v7 id in the batch, so no IN PARTITION at all - the mutation must fall back to visiting
+        // every part, exactly as it did before OPIK-8230. Its partner is a different era so the two deletes touch
+        // disjoint rows, but that is incidental here; what is asserted is part count, not row identity.
+        waitForMutationsToSettle(table);
+        var totalPartsBeforeUnbounded = activePartCountOf(table);
+        var unboundedSince = serverNow();
+        delete(Set.of(Pair.of(projectId, ids.get(1)), Pair.of(projectId, NON_V7_ID)));
```
andrescrz: Minor: inherited from before, but we can probably avoid tests where ID is not an UUID v7. This is a hard precondition as they're always validated. In addition, we audited production and there were no occurrences of this.

apps/opik-backend/src/test/java/com/comet/opik/utils/WeeklyPartitionsTest.java
```diff
@@ -164,29 +102,121 @@ void largestUuidV7TimestampDisablesPruning() {
...
+    }
+
+    @Test
+    @DisplayName("the earliest id a UUIDv7 can carry is inside Date32, so there is no floor to guard")
+    void earliestUuidV7TimestampIsInRange() {
+        // All 48 timestamp bits clear: id_at 1970-01-01, the earliest any UUIDv7 can encode (the field is unsigned).
+        // Its Monday, 1969-12-29, is 70 years above Date32's 1900 floor, so a below-1900 id_at is unreachable by
+        // construction rather than merely untested — which is why the derivation guards only the ceiling. It is also
+        // the floor of the legacy 32-bit DateTime, so the wrap leaves it untouched and it contributes one value.
+        assertThat(WeeklyPartitions.groupByPartition(List.of(UUID.fromString("00000000-0000-7abc-8000-000000000001")))
```
andrescrz, line 125: Minor: pre-existing issue, but better rely on random values for tests, with the usual ways: Podam, RandomStringUtils, IDGenerator, UUID etc. depending on the case.

andrescrz (commented):
Let's iterate a bit before moving forward.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
thiagohora: Done — took option 2. The flag is gone; the gate now reads `traceColumnsNonNullable`.

Your reasoning holds up: the EXCHANGE that replaces the `Nullable(...)` columns is the same one that puts the weekly-partitioned successor behind the name mutations target, so the two facts have only ever flipped together. I checked it against all four topologies rather than just the two obvious ones:

| topology | `traceColumnsNonNullable` | partitioned? | gate | correct? |
|---|---|---|---|---|
| legacy, pre-cutover | false | no | off → unbounded | yes |
| wrap without EXCHANGE (`TracesDistributedWrapMutationTest`) | false (unset) | no | off → unbounded | yes |
| **production today** (post-EXCHANGE, pre-wrap) | **true** | **yes** | on → scoped | yes |
| pruning suite | true | yes | on → scoped | yes |

The second row is the one I wanted to be sure about — that suite sets only `tracesDistributedWrapEnabled=true` and never runs the EXCHANGE, so it is a genuine wrap-on/unpartitioned estate, and `IN PARTITION` there is a hard Code 248 rather than a lost optimisation. It reads false and takes the fallback.

Net effect on deployment surface: `config.yml`, `config-test.yml` and `docker-compose.yaml` are all reverted, so the diff is now 7 files, entirely under `apps/opik-backend`. Nothing to add to the runbook or tooling.

I didn't pursue option 1. `IN PARTITION` against a table with no `PARTITION BY` fails outright, so a neutral form would have to avoid the clause rather than render it harmlessly — which is the gate again, just less visible.

The one real cost is the one you named: the flag's name no longer says everything it gates. I've documented the second duty on both `DatabaseAnalyticsDataModelConfig` and at the gate itself, and left the name alone since the env var is exposed.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
thiagohora: Good catch — restored.

I'd hoisted `tracesMutationTable()` into `deleteBatch` and passed the name down, on the theory that resolving it once per batch beat once per statement. That saving is a record accessor and a ternary, which is not worth stepping outside the one sanctioned path. `executeDelete` calls `selectTracesMutationTable(template)` again and the `table` parameter is gone.

It also puts the arch guard back on the right side: `deleteBatch` no longer reads the routing helper at all, so `tracesMutationTable()` keeps its single caller.

apps/opik-backend/src/main/java/com/comet/opik/utils/WeeklyPartitions.java
thiagohora: I've kept the loud failure, but trimmed the rationale from five lines to three — it was over-explained for what is one annotation.

The reason it earns its keep: empty is a *documented answer* here, meaning "this batch can't be pruned, emit the unbounded form". So a caller that lost its batch wouldn't get an error, it would get a correct-looking unbounded mutation — which is precisely the slow delete this whole change exists to eliminate. Silently degrading back into the bug is a worse failure mode than an NPE at the boundary.

Empty already returns empty, as you'd expect, and a null *element* stays a data condition that returns empty too. It's only a null collection that throws.

Happy to drop it if you'd still rather — it's a one-word change.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/TracesPartitionPruningMutationTest.java
thiagohora: Removed — all three calls are gone.

Digging into it, `SYSTEM FLUSH LOGS` was worse than just a flake risk: it pushes whatever has already reached the buffer, but cannot wait for a row still on its way. So it raced the very writes it was there to make visible, which is exactly the failure it looked like it was preventing.

Replaced with two things:

1. A `fast-log-flush.xml` dropped into this suite's container, setting `query_log` / `part_log` `flush_interval_milliseconds` to 200 instead of the 7.5s default. Safe to do globally here because this suite's ClickHouse is dedicated and non-reused (the EXCHANGE destructively swaps `traces`, so it can't share), so no other suite pays for the extra flushing.
2. The readers now poll. `lastTraceDeleteSql` awaits its row appearing. `mutatePartActivitySince` and `deleteSqlsSince` have no single row to wait for — only counts that grow — so they read until two consecutive samples agree.

Suite runtime is unchanged at ~52s.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/TracesPartitionPruningMutationTest.java
thiagohora: Added — `idsSharingAPartitionAreCarriedByOneStatement`.

You were right that this was a gap, and a bigger one than it looks. Every existing pruning test deletes one id per week, so every emitted statement had `pairs_size=1` — meaning a regression that emitted one statement *per id* rather than per partition would have satisfied all of them. Grouping is the whole point of the change, so nothing was actually pinning it.

The new test puts three ids in each of two weeks and asserts two statements, each `pairs_size=3`, each scoped to its own week. It uses the 1996 and 2025 eras only — the 2199 id resolves to two partitions (its own week plus its legacy 32-bit wrap), which would blur the per-partition counts.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/TracesPartitionPruningMutationTest.java
thiagohora: Removed from both integration suites.

One thing I checked before doing it: the non-v7 case was carrying the *unbounded fallback* coverage in three places, so deleting it outright would have quietly removed the fallback from the suite. The beyond-2299 case takes the identical branch and is a real data condition — the DAO javadoc already says so — so I substituted it everywhere rather than dropping the coverage:

- `pruningReachesThePlannerAndTheFallbackDoesNot` and `requestSpanningTwoChunksPrunesEachChunkIndependently` now use a beyond-2299 id to force the fallback.
- `underivableIdDisablesPruning` drops from a `@ParameterizedTest` over two causes to a plain `@Test` over the one that can happen.
- `TracesLegacyTablePruningMutationTest.nonV7IdDisablesPruning` is deleted rather than converted. That table is unpartitioned, so the gate is always off and every delete is unbounded — the test above it already asserts exactly that shape. There was no second behaviour there to preserve.

The ids are now minted fresh per call rather than shared constants, since these suites run PER_CLASS against one table.

I did keep non-v7 coverage in `WeeklyPartitionsTest`. `partitionsOf` still branches on `id.version() != 7`, and that's a real branch in a util that shouldn't assume its callers validated — removing its only test would leave it uncovered. It's a unit test of the function, not a claim that such ids reach production.

apps/opik-backend/src/test/java/com/comet/opik/utils/WeeklyPartitionsTest.java
thiagohora: Applied where the id is incidental, kept where it isn't — and I've documented the split in the class javadoc so it doesn't read as inconsistency.

Randomised (now minted from `IdGenerator`, since these assert grouping shape, immutability, or the all-or-nothing rule, and none names a partition value):
- `groupByPartitionDuplicatesCollapseWithinAGroup`
- `groupByPartitionNonV7DisablesPruning` — the v4 is now `UUID.randomUUID()`
- `groupByPartitionOutOfRangeIdDisablesPruning` — the ordinary half
- `groupByPartitionResultIsImmutable`

Kept written out, for two different reasons:
- The **pinned** ids. Each expected value is what ClickHouse itself returned for that id under `toYYYYMMDD(toDate32(id_at) - toIntervalDay(toDayOfWeek(id_at, 1)))`, evaluated once as `DateTime64(0)` and once as `DateTime`. Randomising the id means computing the expectation in Java, which is the derivation under test — the assertion would then agree with itself and stop being able to fail. These are the guard against ClickHouse's conversion changing under us.
- The **constructed boundary** ids (all 48 timestamp bits set, all clear). The point of each is an `id_at` the column cannot store, which is exactly what no random id will ever be.

thiagohora:
**On the 4 red TypeScript SDK E2E checks from the previous push — those were flakes, not this branch.**

Worth recording how that was established, since the checks looked branch-specific: this workflow does build the backend from source and bring up the `deployment/docker-compose` stack, and the previous revision touched that compose file, so it was a plausible link.

It wasn't. The backend started fine — 313 tests across 37 of 38 files passed against it. The one failing suite was `tests/integration/runner/runner.test.ts`, timing out in `beforeAll` after 10s waiting for `npx tsx echo_app.ts` to spawn and register an agent. That's the agent-runner/pairing path, nothing to do with trace deletes, partitions, or configuration.

Re-running the failed jobs on the *unchanged* SHA turned both green, which settles it.

The compose change is gone regardless, since the flag it declared is gone.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -3649,38 +3631,69 @@ public Mono<Void> delete(Set<Pair<UUID, UUID>> projectIdTraceIdPairs, @NonNull C
...
+        // traceColumnsNonNullable doubles as "the mutation target is weekly-partitioned": the same cutover EXCHANGE
+        // drops the Nullable(...) columns and puts the partitioned successor behind the name mutations target, so one
+        // flag carries both facts. Deliberately not the wrap flag, which governs routing and is still false in the
+        // window between the EXCHANGE and the wrap - reading that one leaves production's deletes unpruned.
+        var grouped = traceColumnsNonNullable()
+                ? WeeklyPartitions.groupByPartition(batch.stream().map(Pair::getRight).toList())
+                : Optional.<Map<Long, Set<UUID>>>empty();
+
+        if (grouped.isEmpty()) {
+            return executeDelete(batch, null, workspaceId, userName, connection);
```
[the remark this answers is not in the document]
thiagohora: Fixed — but with a correction to the framing.

The NPE is real, and I reproduced exactly the one you describe. Removing the guard I just added:

```
java.lang.NullPointerException: Cannot invoke "java.util.UUID.toString()"
  because the return value of "org.apache.commons.lang3.tuple.Pair.getRight()" is null
```

It is **not** introduced by this PR, though. `pair.getLeft().toString()` / `pair.getRight().toString()` are byte-identical to the pre-change code — `git show <merge-base>:…/TraceDAO.java` has them at lines 3658-3659. So this is pre-existing behaviour, not a regression.

What this PR does change is how bad it is. On the partitioned path a null id reads as *underivable*, which disables pruning for the whole batch — so a caller's bug would first surface as the slow unbounded delete this change exists to eliminate, and only then as the NPE. That is worth closing.

Taking the first of your two options, with one adjustment to where:

```java
Preconditions.checkArgument(
        projectIdTraceIdPairs.stream().noneMatch(pair -> pair.getLeft() == null || pair.getRight() == null),
        "Argument 'projectIdTraceIdPairs' must not contain null ids");
```

It sits at `TraceDAO#delete`'s entry, next to the existing empty-batch precondition, rather than "before partition grouping" as suggested. Grouping only happens when the gate is on; with `traceColumnsNonNullable` false, `groupByPartition` is never called and a null would go straight to `executeDelete`. Checking before the gate covers both branches with one check. It also covers `getLeft()`, which has the same problem.

I did not make the null-element behaviour "safe and intentional" in `WeeklyPartitions` instead. Returning empty there means *"can't prune, emit the unbounded form"* — for a programming error that would be a silent degradation into the slow path, which is the same reason a null *collection* already throws. Rejecting at the boundary is the honest version. I've updated that javadoc so it no longer reads as though the delete path tolerates nulls.

Covered by `nullIdInBatchIsRejected`, asserting both components. Verified non-vacuous: with the precondition removed the test fails with the NPE above rather than the `IllegalArgumentException` it expects.

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
```diff
@@ -3645,42 +3627,79 @@ private Flux<? extends Result> getDetailsById(UUID id, Connection connection) {
     public Mono<Void> delete(Set<Pair<UUID, UUID>> projectIdTraceIdPairs, @NonNull Connection connection) {
         Preconditions.checkArgument(CollectionUtils.isNotEmpty(projectIdTraceIdPairs),
                 "Argument 'projectIdTraceIdPairs' must not be empty");
+        // Checked here rather than where the ids are stringified, so it holds whichever branch deleteBatch takes: the
+        // partitioned path would otherwise read a null as "underivable", silently take the unbounded fallback, and
+        // only then NPE - reporting a caller's bug as the slow delete this class exists to avoid.
+        Preconditions.checkArgument(
+                projectIdTraceIdPairs.stream().noneMatch(pair -> pair.getLeft() == null || pair.getRight() == null),
+                "Argument 'projectIdTraceIdPairs' must not contain null ids");
```
andrescrz, line 3635: Minor: this was added due to a Baz comment. 

While I'm all in for defensive code, we need to draw a line at some point on how deep we validate payloads and what's the appropriate place. 

The reality is that trace ID and project ID are mandatory fields in these entities. In our data model, an entity can't exists without it's own primary identifier. 

So likely we could take this assumption as guaranteed at this point or add a more proper validation upstream if easy.

Anyway, no harm to keep this.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/DatabaseAnalyticsDataModelConfig.java
```diff
@@ -15,6 +15,13 @@
  * epoch end time round-trips unchanged rather than being read as {@code null}. Flip this in lockstep with the EXCHANGE
  * step of the cutover.</p>
  *
+ * <p>It also gates partition-scoped deletes (OPIK-8230): the EXCHANGE that makes these columns non-nullable is the
+ * same one that puts the weekly-partitioned successor behind the name mutations target, so this flag being
+ * {@code true} is equally what says a delete may scope itself with {@code IN PARTITION}. One flag for two facts
+ * because they have only ever flipped together; a second would have to be threaded through the cutover runbook and
+ * tooling to track no independent state. The name says only the first duty and is deliberately not renamed - the env
+ * var is exposed.</p>
+ *
```
andrescrz, line 24: Minor: this documentation makes sense, but the most important for external parties is the documentation comments in the config.yml file. I strongly recommend adding those (and propagating to the config test yaml file).

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/TracesPartitionPruningMutationTest.java
```diff
@@ -319,17 +321,33 @@ SELECT toString(uniqExact(id))
...
+     * non-reused (see below), so no other suite pays for the extra flushing.
      */
-    private static final UUID OUT_OF_RANGE_ID = ID_GENERATOR
-            .getTimeOrderedEpoch(LocalDate.of(2300, 1, 1).atStartOfDay().toInstant(ZoneOffset.UTC).toEpochMilli());
+    private static final String FAST_LOG_FLUSH_CONFIG = """
+            <clickhouse>
+                <query_log><flush_interval_milliseconds>200</flush_interval_milliseconds></query_log>
+                <part_log><flush_interval_milliseconds>200</flush_interval_milliseconds></part_log>
+            </clickhouse>
+            """;
```
andrescrz: I don't think this should be embedded into the code. Let's create a file in the test resources and then inject it in the container. 

Anyway, these are just system logs configurations. We could even adopt it for all tests in the system by adding them to the existing test resource files for ClickHouse, I don't think they harm.

Having said that, happy to keep this for only these tests, but not embedded in the code.

msg.txt
andrescrz, line 1: Let's uncomitted this file.

andrescrz (commented):
Just 2 more blocking comments, the rest LGTM.

msg.txt
thiagohora: Removed — that was an accident on my side, not a deliberate file. It's my commit-message scratch; I wrote it to the repo root instead of a temp dir and a `git add -A` swept it in. Gone from the index and the worktree, and the scratch now lives outside the checkout so it can't recur.

apps/opik-backend/src/test/java/com/comet/opik/infrastructure/TracesPartitionPruningMutationTest.java
thiagohora: Moved to `apps/opik-backend/src/test/resources/clickhouse-fast-log-flush.xml`, injected with `MountableFile.forClasspathResource` — the same mechanism `clickhouse.xml` and `users.xml` already use. The constant in the test is now just the filename.

On the wider question: I've kept it opt-in per suite rather than folding it into the shared `clickhouse.xml`, and said why in the file itself. I agree it's harmless — it's only log-table flush intervals — but only suites that read `system.query_log` / `system.part_log` need it, and this one's ClickHouse is dedicated and non-reused (the EXCHANGE destructively swaps `traces`, so it can't share). Adding it to the shared config would change flush behaviour for every reused container in the estate, which is a bigger blast radius than this PR should take on. Trivial to promote later if another suite wants it — one line moves.

apps/opik-backend/src/main/java/com/comet/opik/infrastructure/DatabaseAnalyticsDataModelConfig.java
thiagohora: Agreed and done — added to `config.yml`, and propagated to `config-test.yml`.

You're right that this is where it matters: the record javadoc is for us, the YAML comment is for whoever is running the cutover. The new text says what `true` additionally means for deletes, that `IN PARTITION` against an unpartitioned table is a hard code 248 rather than a lost optimisation, and that off is the correct-but-slower fallback.

Both edits are comment-only — no key added, no value changed:

```
git diff -U0 -- apps/opik-backend/config.yml apps/opik-backend/src/test/resources/config-test.yml \
  | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)' | grep -vE '^[+-]\s*#'
# (no output)
```

apps/opik-backend/src/main/java/com/comet/opik/domain/TraceDAO.java
thiagohora: Your instinct was right, and it's stronger than "likely" — I went and checked, and **both HTTP entry points already reject null ids upstream**:

- `POST /traces/delete` → `@Valid BatchDeleteByProject`, which declares `Set<@NotNull UUID> ids` — element-level bean validation, not just container-level.
- `DELETE /traces/{id}` → `Set.of(id)`, and `Set.of` rejects null outright.

So a null cannot reach the DAO over the API at all. Baz's finding was reachable only from a programmatic caller, and `TraceService#delete` is the sole one in main. That makes my check defence-in-depth for internal callers rather than a fix for a live crash — I should have established that before adding it, rather than after.

I've left it in on your "no harm", but I'd equally happily drop it — say the word and it goes. The one argument for keeping: `TraceDAO#delete` is reachable without passing through bean validation, and the failure mode there is specifically bad now, since a null reads as *underivable* and silently routes the batch down the unbounded path before the NPE. That's a slow delete masquerading as a caller bug, which is the thing this PR exists to remove. But that's an argument about a caller that doesn't currently exist.

Worth noting for calibration on this class of bot finding: the report was titled "Null trace ID crashes deletion", and it doesn't — not through any path a user can reach.

andrescrz (approved):
LGTM

PR #8133, opened 2026-09-03
[NA] [BE] fix: recalculate the estimated cost on bulk span update

### LINKED ISSUE #8098: [Bug]: Bulk span update does not recalculate total_estimated_cost when model/provider/usage change

- Opik Server

### Describe the problem

The bulk span update path (`PATCH /v1/private/spans/batch`) does not recalculate `total_estimated_cost` when a span's `model`, `provider`, or `usage` changes but `total_estimated_cost` itself is omitted from the update.

In `SpanDAO`, `newBulkUpdateTemplate` and `bindBulkUpdateParams` only add and bind the cost columns when `spanUpdate.totalEstimatedCost() != null`. The `BULK_UPDATE` SQL then falls back to the old value for those rows: `<if(total_estimated_cost)> toDecimal128(:total_estimated_cost, 12) <else> total_estimated_cost <endif>`. So a bulk update carrying only `model` / `provider` / `usage` leaves the previously stored cost untouched.

The single-span update path does the opposite: it recomputes through `CostService.calculateCost(spanUpdate.model(), spanUpdate.provider(), spanUpdate.usage(), ...)` before persisting, and the ingestion/create path computes cost from usage as well. The bulk-update path is the only one that skips this, so a span first written without a priceable model (cost 0) and later corrected via a batch update to a real model + usage keeps its zero/stale cost, which under-reports project and experiment cost aggregates.

Expected: a bulk update that changes `model`, `provider`, or `usage` without an explicit `total_estimated_cost` recalculates the cost the same way the single-span update does, unless a manual cost was explicitly supplied (which should be preserved).

Actual: the persisted cost is left at its previous value, so newly priceable spans stay at zero or a stale estimate.

1. Create a span with a model that has no price entry (so `total_estimated_cost` is stored as 0), including a `usage` block.
2. Send `PATCH /v1/private/spans/batch` for that span setting `model` (and/or `provider` / `usage`) to a priceable model, without setting `total_estimated_cost`.
3. Read the span back: `total_estimated_cost` is still 0, and project/experiment cost aggregates that include it are under-reported.

Doing the same change through the single-span update endpoint recalculates the cost correctly, which confirms the divergence.

### Notes

Surfaced by the automated reviewer on comet-ml/opik#7832 while registering additional providers, but it is independent of any provider mapping and affects all providers, so filing it separately as agreed on that PR.

aseem-ai:
## Details
`PATCH /v1/private/spans/batch` only wrote the cost columns when the client sent `total_estimated_cost`. The single-span update recomputes it through `CostService` when `model`, `provider` or `usage` change and no manual cost is given, so a span first written without a priceable model kept a zero or stale cost after a batch correction and under-reported project and experiment spend. This mirrors that branch into `newBulkUpdateTemplate` / `bindBulkUpdateParams`.

- Bulk does not read the rows it updates, so the query itself keeps a manually set cost (`total_estimated_cost > 0 AND total_estimated_cost_version = ''`) rather than going through `isManualCost`.
- An explicit `total_estimated_cost` in the batch payload still wins and is still stored as manual (blank version).

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #8098
- OPIK-

## AI-WATERMARK

AI-WATERMARK: [no]

- If yes:
  - Tools:
  - Model(s):
  - Scope:
  - Human verification:

## Testing
Run from `apps/opik-backend` on JDK 25, local Docker Testcontainers.

- `mvn -B test -Dtest=SpansBatchUpdateResourceTest` -> 9 tests, 0 failures.
- `mvn -B test -Dtest="SpansResourceTest,SpansBatchUpdateResourceTest,SpanServiceImplTest,FindSpansResourceTest,SpansLocalV2TableTest"` -> 1878 tests, 0 failures.
- `mvn spotless:check` -> clean.

Scenarios, all in a new `BatchUpdateCost` nested class:

- Batch update with a priceable model + usage and no cost -> cost recalculated and stored with the estimated version. Fails on `main` with `Expecting actual not to be null`.
- Span with a manually set cost, same batch update -> manual cost kept, version stays blank.
- Batch update with no priceable model -> stored cost untouched.

The existing `batchUpdate__updateAllFields__success` covers an explicit `total_estimated_cost` in the batch payload and still passes.

Not run: the full backend suite. CI splits it into 16 jobs; I ran every span test class instead. No screen recording.

## Documentation
None needed. No API shape change.

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/domain/SpanDAO.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/domain/SpanDAO.java
@@ -1768,8 +1768,10 @@ public class SpanDAO {
                 <if(metadata)> :metadata <else> s.metadata <endif> as metadata,
                 <if(model)> :model <else> s.model <endif> as model,
                 <if(provider)> :provider <else> s.provider <endif> as provider,
-                <if(total_estimated_cost)> toDecimal128(:total_estimated_cost, 12) <else> s.total_estimated_cost <endif> as total_estimated_cost,
-                <if(total_estimated_cost_version)> :total_estimated_cost_version <else> s.total_estimated_cost_version <endif> as total_estimated_cost_version,
+                <if(recalculated_cost)> if(s.total_estimated_cost > 0 AND s.total_estimated_cost_version = '', s.total_estimated_cost, toDecimal128(:total_estimated_cost, 12))
+                <elseif(total_estimated_cost)> toDecimal128(:total_estimated_cost, 12) <else> s.total_estimated_cost <endif> as total_estimated_cost,
+                <if(recalculated_cost)> if(s.total_estimated_cost > 0 AND s.total_estimated_cost_version = '', s.total_estimated_cost_version, :total_estimated_cost_version)
+                <elseif(total_estimated_cost_version)> :total_estimated_cost_version <else> s.total_estimated_cost_version <endif> as total_estimated_cost_version,
                 """
             + TagOperations.tagUpdateFragment("s.tags")
             + """
@@ -3218,9 +3220,16 @@ public class SpanDAO {
         Optional.ofNullable(spanUpdate.errorInfo())
                 .ifPresent(errorInfo -> template.add("error_info", JsonUtils.readTree(errorInfo).toString()));
 
-        if (spanUpdate.totalEstimatedCost() != null) {
+        boolean shouldRecalculateEstimatedCost = spanUpdate.totalEstimatedCost() == null
+                && isUpdateCostRecalculationAvailable(spanUpdate);
+        if (spanUpdate.totalEstimatedCost() != null || shouldRecalculateEstimatedCost) {
             template.add("total_estimated_cost", "total_estimated_cost");
             template.add("total_estimated_cost_version", "total_estimated_cost_version");
+        }
+        if (shouldRecalculateEstimatedCost) {
+            // Bulk doesn't load the rows it updates, so a cost the user set manually is kept by the
+            // query itself, the way isManualCost keeps it on the single span path.
+            template.add("recalculated_cost", "recalculated_cost");
         }
         Optional.ofNullable(spanUpdate.ttft())
                 .ifPresent(ttft -> template.add("ttft", ttft));
@@ -3268,6 +3277,11 @@ public class SpanDAO {
         if (spanUpdate.totalEstimatedCost() != null) {
             statement.bind("total_estimated_cost", spanUpdate.totalEstimatedCost().toString());
             statement.bind("total_estimated_cost_version", "");
+        } else if (isUpdateCostRecalculationAvailable(spanUpdate)) {
+            BigDecimal estimatedCost = CostService.calculateCost(spanUpdate.model(), spanUpdate.provider(),
+                    spanUpdate.usage(), spanUpdate.metadata());
+            statement.bind("total_estimated_cost", estimatedCost.toString());
+            statement.bind("total_estimated_cost_version", ESTIMATED_COST_VERSION);
         }
         Optional.ofNullable(spanUpdate.ttft())
                 .ifPresent(ttft -> statement.bind("ttft", ttft));
--- a/apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/SpansBatchUpdateResourceTest.java
+++ b/apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/SpansBatchUpdateResourceTest.java
@@ -20,6 +20,7 @@ import com.comet.opik.api.resources.utils.resources.ProjectResourceClient;
 import com.comet.opik.api.resources.utils.resources.SpanResourceClient;
 import com.comet.opik.api.resources.utils.resources.TraceResourceClient;
 import com.comet.opik.domain.SpanType;
+import com.comet.opik.domain.cost.CostService;
 import com.comet.opik.extensions.DropwizardAppExtensionProvider;
 import com.comet.opik.extensions.RegisterApp;
 import com.comet.opik.podam.PodamFactoryUtils;
@@ -387,6 +388,113 @@ class SpansBatchUpdateResourceTest {
             assertThat(span.errorInfo().exceptionType()).isEqualTo("ValidationError");
             assertThat(span.errorInfo().message()).isEqualTo("Invalid input");
             assertThat(span.ttft()).isEqualTo(123.45);
+        }
+    }
+
+    @Nested
+    @DisplayName("Batch Update Cost:")
+    @TestInstance(TestInstance.Lifecycle.PER_CLASS)
+    class BatchUpdateCost {
+
+        private static final String MODEL = "gpt-4o-2024-05-13";
+        private static final String PROVIDER = "openai";
+        private static final Map<String, Integer> USAGE = Map.of("prompt_tokens", 1000, "completion_tokens", 500);
+
+        private UUID traceId;
+
+        @BeforeEach
+        void setUp() {
+            var trace = podamFactory.manufacturePojo(Trace.class).toBuilder()
+                    .projectName(DEFAULT_PROJECT)
+                    .feedbackScores(null)
+                    .build();
+            traceId = traceResourceClient.createTrace(trace, API_KEY, TEST_WORKSPACE);
+        }
+
+        @Test
+        @DisplayName("Success: batch update recalculates the cost when the model becomes priceable")
+        void batchUpdate__whenModelBecomesPriceable__thenRecalculateCost() {
+            var id = createUnpricedSpan(null);
+
+            batchUpdate(id, SpanUpdate.builder()
+                    .projectName(DEFAULT_PROJECT)
+                    .traceId(traceId)
+                    .model(MODEL)
+                    .provider(PROVIDER)
+                    .usage(USAGE)
+                    .build());
+
+            var updatedSpan = spanResourceClient.getById(id, TEST_WORKSPACE, API_KEY);
+            assertThat(updatedSpan.totalEstimatedCost())
+                    .isEqualByComparingTo(CostService.calculateCost(MODEL, PROVIDER, USAGE, null));
+            assertThat(updatedSpan.totalEstimatedCostVersion()).isNotBlank();
+        }
+
+        @Test
+        @DisplayName("Success: batch update keeps a manually set cost")
+        void batchUpdate__whenSpanHasManualCost__thenKeepManualCost() {
+            var manualCost = new BigDecimal("0.5");
+            var id = createUnpricedSpan(manualCost);
+
+            batchUpdate(id, SpanUpdate.builder()
+                    .projectName(DEFAULT_PROJECT)
+                    .traceId(traceId)
+                    .model(MODEL)
+                    .provider(PROVIDER)
+                    .usage(USAGE)
+                    .build());
+
+            var updatedSpan = spanResourceClient.getById(id, TEST_WORKSPACE, API_KEY);
+            assertThat(updatedSpan.totalEstimatedCost()).isEqualByComparingTo(manualCost);
+            assertThat(updatedSpan.totalEstimatedCostVersion()).isNull();
+        }
+
+        @Test
+        @DisplayName("Success: batch update without a priceable model keeps the stored cost")
+        void batchUpdate__whenUpdateIsNotPriceable__thenKeepStoredCost() {
+            var span = podamFactory.manufacturePojo(Span.class).toBuilder()
+                    .projectName(DEFAULT_PROJECT)
+                    .traceId(traceId)
+                    .parentSpanId(null)
+                    .model(MODEL)
+                    .provider(PROVIDER)
+                    .usage(USAGE)
+                    .totalEstimatedCost(null)
+                    .feedbackScores(null)
+                    .build();
+            var id = spanResourceClient.createSpan(span, API_KEY, TEST_WORKSPACE);
+            var storedCost = spanResourceClient.getById(id, TEST_WORKSPACE, API_KEY).totalEstimatedCost();
+
+            batchUpdate(id, SpanUpdate.builder()
+                    .projectName(DEFAULT_PROJECT)
+                    .traceId(traceId)
+                    .name("updated-name")
+                    .build());
+
+            var updatedSpan = spanResourceClient.getById(id, TEST_WORKSPACE, API_KEY);
+            assertThat(updatedSpan.totalEstimatedCost()).isEqualByComparingTo(storedCost);
+        }
+
+        private UUID createUnpricedSpan(BigDecimal manualCost) {
+            var span = podamFactory.manufacturePojo(Span.class).toBuilder()
+                    .projectName(DEFAULT_PROJECT)
+                    .traceId(traceId)
+                    .parentSpanId(null)
+                    .model(null)
+                    .provider(null)
+                    .usage(null)
+                    .totalEstimatedCost(manualCost)
+                    .feedbackScores(null)
+                    .build();
+            return spanResourceClient.createSpan(span, API_KEY, TEST_WORKSPACE);
+        }
+
+        private void batchUpdate(UUID id, SpanUpdate update) {
+            spanResourceClient.batchUpdateSpans(SpanBatchUpdate.builder()
+                    .ids(Set.of(id))
+                    .update(update)
+                    .mergeTags(false)
+                    .build(), API_KEY, TEST_WORKSPACE);
         }
     }
 }


apps/opik-backend/src/main/java/com/comet/opik/domain/SpanDAO.java
```diff
@@ -3218,10 +3220,17 @@ private ST newBulkUpdateTemplate(SpanUpdate spanUpdate, String sql, boolean merg
         Optional.ofNullable(spanUpdate.errorInfo())
                 .ifPresent(errorInfo -> template.add("error_info", JsonUtils.readTree(errorInfo).toString()));
 
-        if (spanUpdate.totalEstimatedCost() != null) {
+        boolean shouldRecalculateEstimatedCost = spanUpdate.totalEstimatedCost() == null
+                && isUpdateCostRecalculationAvailable(spanUpdate);
```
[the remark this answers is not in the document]
aseem-ai: That's the behavior on main too — a partial payload leaves the stored cost untouched, and this PR keeps that. A per-span merge doesn't fit the current bulk design: one statement applies one update to all ids, so a merge would need a per-span cost value. Single-span update only merges because it already loads the row for the manual-cost check. This covers the issue's case, a batch carrying model+usage together. I can open a follow-up issue for a per-span merge if that's wanted.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanDAO.java
```diff
@@ -3218,10 +3220,17 @@ private ST newBulkUpdateTemplate(SpanUpdate spanUpdate, String sql, boolean merg
...
+        if (spanUpdate.totalEstimatedCost() != null || shouldRecalculateEstimatedCost) {
             template.add("total_estimated_cost", "total_estimated_cost");
             template.add("total_estimated_cost_version", "total_estimated_cost_version");
         }
+        if (shouldRecalculateEstimatedCost) {
+            // Bulk doesn't load the rows it updates, so a cost the user set manually is kept by the
+            // query itself, the way isManualCost keeps it on the single span path.
+            template.add("recalculated_cost", "recalculated_cost");
```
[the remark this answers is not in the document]
aseem-ai: That's pre-existing — main posts the same single-traceId event for every batch update, including ones with an explicit cost, and the bulk SQL never rewrites trace_id. This PR doesn't touch the event path. I can open a separate issue for propagating the persisted trace ids.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/SpansBatchUpdateResourceTest.java
```diff
@@ -389,4 +390,111 @@ private void assertSpanFieldsUpdated(Span span, Instant expectedEndTime) {
...
+                    .projectName(DEFAULT_PROJECT)
+                    .traceId(traceId)
+                    .model(MODEL)
+                    .provider(PROVIDER)
+                    .usage(USAGE)
+                    .build());
+
+            var updatedSpan = spanResourceClient.getById(id, TEST_WORKSPACE, API_KEY);
+            assertThat(updatedSpan.totalEstimatedCost())
+                    .isEqualByComparingTo(CostService.calculateCost(MODEL, PROVIDER, USAGE, null));
```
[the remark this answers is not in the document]
aseem-ai: This follows the existing convention — the single-span cost tests in SpansResourceTest derive the expected value with CostService.calculateCost the same way. A hardcoded literal would break each time the price table updates. The version assertion independently checks the estimated-vs-manual distinction.

apps/opik-backend/src/main/java/com/comet/opik/domain/SpanDAO.java
```diff
@@ -3218,10 +3220,17 @@ private ST newBulkUpdateTemplate(SpanUpdate spanUpdate, String sql, boolean merg
         Optional.ofNullable(spanUpdate.errorInfo())
                 .ifPresent(errorInfo -> template.add("error_info", JsonUtils.readTree(errorInfo).toString()));
 
-        if (spanUpdate.totalEstimatedCost() != null) {
+        boolean shouldRecalculateEstimatedCost = spanUpdate.totalEstimatedCost() == null
```
andrescrz, line 3223: **[medium] Bulk recalculation is computed from the partial payload only, so usage/model-only bulk updates leave a stale cost**

`shouldRecalculateEstimatedCost` (3223) and the bind at 3280-3284 call `isUpdateCostRecalculationAvailable(spanUpdate)` / `CostService.calculateCost(spanUpdate.model(), spanUpdate.provider(), spanUpdate.usage(), spanUpdate.metadata())` on the raw `SpanUpdate`. The single-span path never does this: `update(...)` (2109-2115) first merges the *existing* span's model/provider/usage into `finalUpdate`, so a partial update still prices against the row's real model+usage. `SpanService.batchUpdate` (SpanService.java:276) passes the payload straight through with no row load, so on the bulk path any update that does not carry model+provider+usage together prices to 0, `isUpdateCostRecalculationAvailable` returns false, and no `total_estimated_cost` token is added at all. Concretely: a batch update that sets only `usage` (or only corrects `model`) on spans that already have a priced model overwrites the `usage`/`model` columns but leaves `total_estimated_cost` at the value computed from the old usage/model — a silently inconsistent row, which is the exact staleness this change is meant to fix. The new test `batchUpdate__whenUpdateIsNotPriceable__thenKeepStoredCost` locks this behavior in as "keeps the stored cost" even though the stored cost no longer matches the row.

Same issue seen from the other side: `PATCH /v1/private/spans/batch` with only `usage`, or only `model`, or only `provider` makes `CostService.calculateCost(null, null, usage, null)` return 0, so `shouldRecalculateEstimatedCost` is false, neither `total_estimated_cost` nor `recalculated_cost` is added to the template, and the SQL falls through to `<else> s.total_estimated_cost`. The equivalent single-span PATCH still reprices correctly, so the two endpoints now disagree.

💡 Since BULK_UPDATE already reads the target rows in SQL, do the merge there: build the priced expression from `<if(model)> :model <else> s.model <endif>` style fallbacks, or (simpler) have `SpanService.batchUpdate`/`bulkUpdate` reject or explicitly handle the case where `usage`/`model`/`provider` change but the merged triple is not priceable — at minimum reset `total_estimated_cost`/`total_estimated_cost_version` for non-manual rows instead of keeping a value derived from the now-overwritten usage. Add a test for `usage`-only and `model`-only batch updates against an already-priced span.

_rule: `backend-correctness-make-fan-out-queries-and-duplicated-per-endpoint-logic-deter`_

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/SpansBatchUpdateResourceTest.java
```diff
@@ -389,4 +390,111 @@ private void assertSpanFieldsUpdated(Span span, Instant expectedEndTime) {
...
+            var trace = podamFactory.manufacturePojo(Trace.class).toBuilder()
+                    .projectName(DEFAULT_PROJECT)
+                    .feedbackScores(null)
+                    .build();
+            traceId = traceResourceClient.createTrace(trace, API_KEY, TEST_WORKSPACE);
+        }
+
+        @Test
+        @DisplayName("Success: batch update recalculates the cost when the model becomes priceable")
+        void batchUpdate__whenModelBecomesPriceable__thenRecalculateCost() {
```
andrescrz, line 416: **[medium] Re-pricing an already auto-priced span is untested — the overwrite leg never runs**

The new SQL branch overwrites the cost whenever NOT(`total_estimated_cost > 0` AND `total_estimated_cost_version = ''`). The only test that hits the overwrite path (`batchUpdate__whenModelBecomesPriceable__thenRecalculateCost`) seeds a span with cost 0 and an empty version, so the guard is falsified by the `> 0` term alone. The most common production scenario — a span that already has an auto-calculated cost (cost > 0, version = '1.1') whose model/usage is corrected by a batch update — takes the overwrite path via the version term, and nothing covers it. A regression that dropped the `AND s.total_estimated_cost_version = ''` term (turning every priced span into a preserved 'manual' cost) would pass the current suite.

Concretely, in the DAO the expression `if(s.total_estimated_cost > 0 AND s.total_estimated_cost_version = '', s.total_estimated_cost_version, :total_estimated_cost_version)` (SpanDAO.java:1773) has two legs and only the trivial keep-leg (`''`) is reached. `createUnpricedSpan` builds model/provider/usage-null spans, so `bindCost` (SpanDAO.java:3153-3163) always writes version `''`; `thenKeepStoredCost` does seed version `1.1` but its update carries no pricing fields, so `recalculated_cost` is never added and the branch is skipped. A wrong branch order, or a type mismatch between `LowCardinality(String)` `s.total_estimated_cost_version` and the bound parameter, would not be caught.

💡 Add a test that creates a span with a priceable model (stored version `1.1`, cost > 0), batch-updates it to a different priceable model+usage, and asserts both the recalculated cost and that the version is still `ESTIMATED_COST_VERSION` — i.e. the overwrite leg actually ran and the old estimate was replaced.

_rule: `backend-testing-test-sql-templates-and-load-bearing-gating-logic-with-real-q`_

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/SpansBatchUpdateResourceTest.java
```diff
@@ -389,4 +390,111 @@ private void assertSpanFieldsUpdated(Span span, Instant expectedEndTime) {
...
+                    .provider(null)
+                    .usage(null)
+                    .totalEstimatedCost(manualCost)
+                    .feedbackScores(null)
+                    .build();
+            return spanResourceClient.createSpan(span, API_KEY, TEST_WORKSPACE);
+        }
+
+        private void batchUpdate(UUID id, SpanUpdate update) {
+            spanResourceClient.batchUpdateSpans(SpanBatchUpdate.builder()
```
andrescrz, line 493: **[medium] Every cost test batches exactly one span, so the per-row SQL guard is never exercised**

The `batchUpdate` helper always sends `ids(Set.of(id))`, so all three new cost tests exercise a batch of exactly one span. The whole reason SpanDAO.java pushes the manual-cost guard into SQL (`if(s.total_estimated_cost > 0 AND s.total_estimated_cost_version = '', s.total_estimated_cost, toDecimal128(:total_estimated_cost, 12))`) is that the bulk path does not load the rows it updates and must decide per row, with a single bound `:total_estimated_cost` shared by the whole batch. A heterogeneous batch (one manual-cost span + one unpriced span updated in the same call) is exactly the regression this per-row branch exists for, and it is untested — single-id batches would still pass if the DAO had (wrongly) resolved the manual-cost decision once for the whole statement.

This is the load-bearing part of the change: the manual-cost decision moved out of Java (`isManualCost(existingSpan)`, SpanDAO.java:3136, evaluated per span on the single-update path) into a per-ROW ClickHouse expression at SpanDAO.java:1771/1773, precisely because one payload is applied to N heterogeneous rows. A regression that made that decision batch-global — moving it back into Java based on the first row, or binding one precomputed flag — would still pass all three new tests.

💡 Add a test that seeds two spans in one batch — one with a manual cost, one unpriced — issues a single `SpanBatchUpdate` with both ids and a priceable model/provider/usage, and asserts the manual-cost span keeps its cost with a null version while the unpriced one gets the recalculated cost and a stamped version.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/SpansBatchUpdateResourceTest.java
```diff
@@ -389,4 +390,111 @@ private void assertSpanFieldsUpdated(Span span, Instant expectedEndTime) {
...
+
+            var updatedSpan = spanResourceClient.getById(id, TEST_WORKSPACE, API_KEY);
+            assertThat(updatedSpan.totalEstimatedCost())
+                    .isEqualByComparingTo(CostService.calculateCost(MODEL, PROVIDER, USAGE, null));
+            assertThat(updatedSpan.totalEstimatedCostVersion()).isNotBlank();
+        }
+
+        @Test
+        @DisplayName("Success: batch update keeps a manually set cost")
+        void batchUpdate__whenSpanHasManualCost__thenKeepManualCost() {
```
andrescrz, line 435: **[medium] Near-duplicate cost tests should be one @ParameterizedTest**

`batchUpdate__whenModelBecomesPriceable__thenRecalculateCost` and `batchUpdate__whenSpanHasManualCost__thenKeepManualCost` share an identical arrange (`createUnpricedSpan(...)`), an identical act (`batchUpdate` with MODEL/PROVIDER/USAGE) and identical assertion shape (cost + version); only the seeded manual cost and the two expected values differ. The sibling `BatchUpdateTags` nested class in this same file already parameterizes exactly this way via `@MethodSource("mergeTagsTestCases")`.

💡 Collapse them into one `@ParameterizedTest` with a `@MethodSource` supplying (seededManualCost, expectedCost, expectedVersionPresent) — which also makes it cheap to add the missing auto-priced and unknown-model rows called out in the other findings.

_rule: `backend-testing-collapse-near-duplicate-tests-into-a-single-parameterizedtes`_

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/SpansBatchUpdateResourceTest.java
```diff
@@ -389,4 +390,111 @@ private void assertSpanFieldsUpdated(Span span, Instant expectedEndTime) {
...
+            var storedCost = spanResourceClient.getById(id, TEST_WORKSPACE, API_KEY).totalEstimatedCost();
+
+            batchUpdate(id, SpanUpdate.builder()
+                    .projectName(DEFAULT_PROJECT)
+                    .traceId(traceId)
+                    .name("updated-name")
+                    .build());
+
+            var updatedSpan = spanResourceClient.getById(id, TEST_WORKSPACE, API_KEY);
+            assertThat(updatedSpan.totalEstimatedCost()).isEqualByComparingTo(storedCost);
```
andrescrz, line 475: **[low] Non-recalculation test does not assert the version column is preserved**

The new template branch writes both `total_estimated_cost` and `total_estimated_cost_version`; when no recalculation happens the columns must fall through to `s.total_estimated_cost` / `s.total_estimated_cost_version`. This test asserts only the cost, so a regression that leaves the cost intact while clearing or re-stamping the version (e.g. adding `total_estimated_cost_version` to the template without `recalculated_cost`) still passes. Note also that if the pricing table ever stops pricing MODEL, `storedCost` becomes null and `isEqualByComparingTo(null)` fails with an NPE rather than a readable assertion.

💡 Capture the created span once and assert both `totalEstimatedCost()` and `totalEstimatedCostVersion()` are unchanged after the batch update (and assert `storedCost` is non-null before using it as the expected value).

_rule: `backend-testing-make-test-assertions-actually-falsify-the-change-under-test`_

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/SpansBatchUpdateResourceTest.java
```diff
@@ -389,4 +390,111 @@ private void assertSpanFieldsUpdated(Span span, Instant expectedEndTime) {
...
+
+            batchUpdate(id, SpanUpdate.builder()
+                    .projectName(DEFAULT_PROJECT)
+                    .traceId(traceId)
+                    .model(MODEL)
+                    .provider(PROVIDER)
+                    .usage(USAGE)
+                    .build());
+
+            var updatedSpan = spanResourceClient.getById(id, TEST_WORKSPACE, API_KEY);
```
andrescrz, line 427: **[low] Cost tests assert two fields only, not the resulting span**

The recalculation tests check `totalEstimatedCost()` and `totalEstimatedCostVersion()` but never verify that the same batch update actually persisted `model`, `provider` and `usage` — the very inputs the cost is derived from. If the new `recalculated_cost` template branch interfered with the model/provider/usage columns (they share the same generated SELECT), the cost assertions would still pass while the span row is wrong, leaving cost and its provenance fields inconsistent.

💡 Assert the full expected span (the sibling `assertSpanFieldsUpdated` helper in `BatchUpdateAllFields` is the pattern), or at minimum add `model`/`provider`/`usage` assertions alongside the cost ones.

_rule: `backend-testing-assert-the-whole-object-or-response-not-individual-fields`_

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/SpansBatchUpdateResourceTest.java
```diff
@@ -389,4 +390,111 @@ private void assertSpanFieldsUpdated(Span span, Instant expectedEndTime) {
...
+    @TestInstance(TestInstance.Lifecycle.PER_CLASS)
+    class BatchUpdateCost {
+
+        private static final String MODEL = "gpt-4o-2024-05-13";
+        private static final String PROVIDER = "openai";
+        private static final Map<String, Integer> USAGE = Map.of("prompt_tokens", 1000, "completion_tokens", 500);
+
+        private UUID traceId;
+
+        @BeforeEach
```
andrescrz, line 405: **[low] Trace-creation setUp is now triplicated across nested classes**

The new `BatchUpdateCost.setUp()` (lines 405-412) is a verbatim copy of `BatchUpdateTags.setUp()` (143-150) and `BatchUpdateAllFields.setUp()` (303-310); `createUnpricedSpan` likewise re-copies the `podamFactory.manufacturePojo(Span.class).toBuilder().projectName(DEFAULT_PROJECT).traceId(traceId).parentSpanId(null)` construction sequence already repeated five times in this file. Copying the fixture a third time is what makes each future field addition a three-place edit.

💡 Hoist a `private UUID createTrace()` and a `private Span.SpanBuilder newSpan(UUID traceId)` helper onto the outer `SpansBatchUpdateResourceTest` class and have all three nested classes call them.

_rule: `backend-style-maintainability-eliminate-duplication-via-shared-helpers-constants-and-build`_

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/priv/SpansBatchUpdateResourceTest.java
```diff
@@ -389,4 +390,111 @@ private void assertSpanFieldsUpdated(Span span, Instant expectedEndTime) {
...
+                    .usage(USAGE)
+                    .build());
+
+            var updatedSpan = spanResourceClient.getById(id, TEST_WORKSPACE, API_KEY);
+            assertThat(updatedSpan.totalEstimatedCost()).isEqualByComparingTo(manualCost);
+            assertThat(updatedSpan.totalEstimatedCostVersion()).isNull();
+        }
+
+        @Test
+        @DisplayName("Success: batch update without a priceable model keeps the stored cost")
```
andrescrz, line 453: **[low] @DisplayName does not describe the scenario exercised**

"batch update without a priceable model keeps the stored cost" reads as though the model is unpriceable, but the span is created with the priceable MODEL/PROVIDER/USAGE and the update simply carries no cost-related fields at all (only `name`). The branch actually exercised is 'update contains no pricing inputs', not 'model is not priceable' — a reader adding cases later will assume the unknown-model guard is already covered when it is not.

💡 Rename to something like "Success: batch update that changes no pricing fields keeps the stored cost" (and reserve the 'not priceable' wording for the unknown-model case suggested above).

_rule: `backend-testing-test-names-displayname-strings-and-assertion-comments-must-m`_

andrescrz (commented):

8 findings — 0 high · 4 medium · 4 low. Suppressed by team conventions: see `suppressed.md`.

PR #8137, opened 2026-09-03
[OPIK-8240] [BE] fix: Resume XAUTOCLAIM from its cursor

thiagohora:
## Details

`BaseRedisSubscriber.claimPendingMessages` passed `StreamMessageId.MIN` as the scan start on **every** call and discarded the cursor Redis returns (`AutoClaimResult.getNextId()`).

Redis caps each `XAUTOCLAIM` at `COUNT * 10` PEL entries **examined** (not claimed), so at the default `consumerBatchSize` of 10 a call only ever inspects the first 100 entries of the pending list. Restarting at `MIN` every time means anything past that window is never examined at all — a backlog above ~100 grows a permanently unreachable tail, whatever the retry or decode behaviour downstream.

Verified against a real Redis before writing the fix: with PEL positions 1–100 made ineligible and 101–150 eligible, `COUNT 10` claims **nothing** and returns cursor `101-1`; `COUNT 20` claims 20.

That matches the customer data that prompted this exactly — a cycling head of ~100 entries at delivery-count in the thousands, sitting in front of ~660 entries frozen at delivery-count **1**, idle 14–20 days. Those tail entries had been delivered once, weeks earlier, and were never looked at again.

The fix carries `getNextId()` forward, resetting to `MIN` on Redis's `0-0` end-of-pass reply so the oldest-first bias the original code intended is preserved. A failed scan deliberately does **not** advance the cursor, so its window is retried rather than skipped.

**One subtlety worth review attention.** The end-of-pass sentinel is compared *numerically*, not against `StreamMessageId.MIN`/`.ALL`. Those are wire sentinels serializing to `-` and `0`, and neither is `equals()` to the `StreamMessageId(0, 0)` that Redisson parses Redis's literal `0-0` into — checked against redisson 4.7.0:

```
MIN = -      ALL = 0      MIN.equals(ALL) = true
new StreamMessageId(0,0).equals(MIN) = false
new StreamMessageId(0,0).equals(ALL) = false
```

Matching on the constants would therefore never fire, the cursor would park at the end of the PEL, and the scan would stop finding anything — strictly worse than the always-`MIN` behaviour it replaces.

## Scope

This PR was originally two fixes. Following review, it now contains **only** the XAUTOCLAIM cursor fix. See "Split out of this PR" below.

## Change checklist
- [x] User facing
- [ ] Documentation update

Behaviour change: a pending backlog deeper than ~100 entries becomes reachable. No new configuration keys.

## Issues

- OPIK-8240

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: Claude Code
- Scope: investigation, the fix, the tests, and this description.
- Human verification: every result below was produced by running the listed commands and reading the output.

### Integration (the one that proves the behaviour)

`BaseRedisSubscriberTest.ClaimCursorTests.shouldReachPendingEntriesBeyondTheFirstScanWindowAndThenWrap` — real Redis via Testcontainers. 150 entries delivered to a never-acking consumer, `consumerBatchSize` 10 (so a 100-entry examine budget), `claimIntervalRatio` 1, `maxRetries` at max so nothing retires, and a processor that always fails retryably so the PEL keeps its depth. Asserts all 150 are reached, plus redelivery of the oldest — i.e. the `0-0` wrap.

**Verified it catches the regression**: restoring `StreamMessageId.MIN` makes it time out at 30s on the tail assertion. Runs in ~2s.

### Unit

`BaseRedisSubscriberUnitTest.ClaimCursorTests` retains `shouldRetryTheSameWindowAfterAFailedScan` — a failing `XAUTOCLAIM` can't be induced against the container, so that one stays a unit test. The cases the integration test now covers were removed rather than duplicated.

```bash
mvn -o surefire:test -Dtest='BaseRedisSubscriberUnitTest,BaseRedisSubscriberTest,ChatCompletionServiceTest,OnlineScoringLlmAsJudgeScorerTest,RedisStreamCodecTest'
# Tests run: 131, Failures: 0, Errors: 0
```

Not run, with reason: full `mvn verify` (whole Testcontainers suite) — relying on CI.

## Split out of this PR

Following @andrescrz's review, whose feedback was mostly about the retry logic:

- **Retry classification** (`ChatCompletionService` status-code split) and the **fan-out collapse** in the trace-thread scorers move to a single follow-up PR under **OPIK-8262**. They are coupled: the collapse is only a *correctness* problem because the retry split makes exception type meaningful. Before it, every provider failure was a blanket 500, so collapsing N sibling errors into one couldn't pick wrong. Shipping the split without addressing the collapse would introduce a real regression in the two trace-thread scorers, so they belong together.
- The `preferRetryable` selector added earlier here is **removed**, not carried forward. It picked a less-bad victim rather than stopping the collapse, and as @andrescrz noted, risks a retry storm — a replay drags in thread ids that already succeeded, and fan-out is unchunked at both publishers. `errors.getFirst()` is restored, which is what main already does; the defect pre-dates this PR.
- The `RedisStreamCodec` javadoc correction briefly folded in from #8145 is **dropped** — it belongs to #8089, which is already merged.

## Documentation
NA


apps/opik-backend/src/main/java/com/comet/opik/domain/llm/ChatCompletionService.java
```diff
@@ -147,16 +148,32 @@ public ChatResponse scoreTrace(@NonNull ChatRequest chatRequest,
...
+            // Deliberately NOT failHandlingLLMProviderError here, unlike create() and the streaming handler.
+            // Those two answer an HTTP caller, so they want the provider's status verbatim - a 429 upstream
+            // should be a 429 downstream. This method answers the online-scoring subscribers instead, where
+            // the thrown type decides retryability: BaseRedisSubscriber.NON_RETRYABLE_EXCEPTIONS matches on
+            // ClientErrorException, so the whole 4xx family would be dropped on the first failure - taking
+            // genuinely transient 408s and 429s with it. Route through the retryability split instead, which
+            // separates "this request can never succeed" from "try again later" inside that family.
+            int status = providerError.map(ErrorMessage::getCode)
+                    .filter(ChatCompletionService::isErrorStatus)
+                    .or(() -> findProviderHttpStatus(runtimeException))
```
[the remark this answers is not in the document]
thiagohora: Confirmed and fixed — this was a real bug, and the damaging direction is worse than described.

I checked both mappers: `CustomLlmErrorMessage.DEFAULT_STATUS = 400` (used whenever the error body is null, textual, or unrecognised) and `OpenAiErrorMessage` falls back to `new ErrorMessage(500, ...)` when the code is not in `OpenAiCompatStatusCodes`. So a real upstream **503 behind an unparseable body** surfaced as CustomLlm's synthetic 400, got classified permanent, and had a retryable failure dropped on its first delivery. That is silent data loss, not just a misclassification.

What makes it clearly wrong rather than a judgement call: `findProviderHttpStatus` already documents exactly this precedence for its own chain walk — "HttpException is searched for across the whole chain before any typed exception is considered, because it carries the upstream code verbatim" — and I inverted it.

Now `findProviderHttpStatus(runtimeException)` wins, with the mapped code as fallback only when the chain carries no HTTP status at all.

Regression added, `scoreTrace__whenMapperFallbackConflictsWithWireStatus__thenWireStatusWins`, covering both conflicting directions you named: wire 503/429 behind a synthetic 400 (must stay retryable), wire 400/403 behind a synthetic 500 (must stay non-retryable). Plus `scoreTrace__whenNoWireStatus__thenFallsBackToTheMapper` to pin that the fallback is still reachable. Mutation-checked: restoring the old precedence fails 4 tests.

apps/opik-backend/src/main/java/com/comet/opik/domain/llm/ChatCompletionService.java
```diff
@@ -147,16 +148,32 @@ public ChatResponse scoreTrace(@NonNull ChatRequest chatRequest,
...
-            // blanket 500 is what keeps them retryable.
             log.warn(UNEXPECTED_ERROR_CALLING_LLM_PROVIDER, runtimeException);
+
+            if (isPermanentFailure(status)) {
+                // Non-retryable: the subscriber acks and removes on the first delivery instead of burning
+                // maxRetries x pendingMessageDuration on a request whose outcome cannot change. This is the
+                // OPIK-8193 case - a provider rejecting an oversized/invalid request rejects every retry of
+                // it identically.
+                throw new ClientErrorException(buildDetailedErrorMessage(runtimeException), status,
+                        runtimeException);
```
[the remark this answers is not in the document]
thiagohora: Good catch — real, and specifically a hazard my change introduced.

You are right about the mechanism: `errors.getFirst()` returns whichever error the `flatMap` happened to emit first, and the single re-emitted error decides the fate of the whole stream entry. While every provider failure was a blanket 500 this was immaterial — all siblings were retryable, so the arbitrary pick could not change retryability. Splitting permanent from transient turned it into a race that can ack-and-remove an entry containing retryable siblings.

Both trace-thread scorers (`OnlineScoringTraceThreadLlmAsJudgeScorer` and `OnlineScoringTraceThreadUserDefinedMetricPythonScorer` — same line in each) now share `OnlineScoringBaseScorer.representativeError()`, which takes your second option: retryable wins whenever any sibling is retryable.

I went with that over splitting/requeueing thread ids because the asymmetry is one-sided. Preferring retryable costs a bounded replay of the permanent sibling (`maxRetries` caps it, and it is dropped for good on the final attempt). Preferring non-retryable costs recoverable work, silently and permanently. Only one of those is recoverable. Independent requeueing would be strictly better still, but it is a larger change to the fan-out contract than this fix warrants — worth its own ticket if you think it is needed.

Covered by `OnlineScoringBaseScorerErrorAggregationTest`, including the ordering that used to lose data (permanent first, retryable second). Mutation-checked: reverting to `getFirst()` fails it.

apps/opik-backend/src/test/java/com/comet/opik/domain/llm/ChatCompletionServiceTest.java
```diff
@@ -627,16 +628,146 @@ void create__whenProviderErrorUnparsed__thenReportProviderStatus(
...
+        void scoreTrace__whenParseableProviderError__thenSplitByStatusToo() {
+            var chatRequest = ChatRequest.builder().messages(UserMessage.from("score this")).build();
+            var modelParameters = podamFactory.manufacturePojo(LlmAsJudgeModelParameters.class);
+            var workspaceId = "test-workspace-id";
+            var providerFailure = new RuntimeException("provider said no");
+
+            when(llmProviderFactory.getLanguageModel(anyString(), any())).thenReturn(chatModel);
+            when(chatModel.chat(any(ChatRequest.class))).thenThrow(providerFailure);
+            when(llmProviderFactory.getService(anyString(), anyString())).thenReturn(llmProviderService);
+            // A parseable 429 used to reach failHandlingLLMProviderError and become a ClientErrorException,
```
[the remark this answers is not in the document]
thiagohora: Addressed, and it turned out to be pointing at the same root cause as the precedence thread.

You were right that mocking `getLlmProviderError` directly bypasses the seam where mapper-vs-cause-chain precedence is actually decided — and that seam had a real bug in it (synthetic mapper fallbacks outranking the wire status; see the thread on `ChatCompletionService.java:160`).

Added `scoreTrace__whenMapperFallbackConflictsWithWireStatus__thenWireStatusWins`, which drives `scoreTrace` with a mapped `ErrorMessage` and a nested `HttpException` carrying a *conflicting* status, asserting both the exception type and the Redis retryability category:

| wire | synthetic mapper | must resolve as |
|---|---|---|
| 503 | 400 | retryable (`InternalServerErrorException`) |
| 429 | 400 | retryable |
| 400 | 500 | non-retryable (`ClientErrorException`) |
| 403 | 500 | non-retryable |

Both assertions run through a shared `assertRetryability` helper that keys off `isPermanentFailure`, so type and status are checked together rather than separately.

One limit worth stating: these still stub `getLlmProviderError` rather than driving a real provider end to end. That is deliberate — `ChatCompletionServiceTest` mocks `LlmProviderService` throughout, and threading a real `CustomLlmProvider`/`LlmProviderLangChainMapper` through would be a different test shape than the rest of the class. What the new cases do cover is the precedence rule itself, which is where the bug lived.

apps/opik-backend/src/test/java/com/comet/opik/domain/llm/ChatCompletionServiceTest.java
```diff
@@ -650,13 +781,21 @@ void scoreTrace__whenProviderErrorUnparsed__thenStayRetryable(
...
+                assertThat(thrown)
+                        .isInstanceOf(ClientErrorException.class)
+                        .isNotInstanceOf(InternalServerErrorException.class);
+                assertThat(((WebApplicationException) thrown).getResponse().getStatus()).isEqualTo(expectedStatus);
+            } else {
+                // Transient (or unknown) — stay outside NON_RETRYABLE_EXCEPTIONS so maxRetries is honoured.
+                assertThat(thrown)
+                        .isInstanceOf(InternalServerErrorException.class)
+                        .isNotInstanceOf(ClientErrorException.class);
+                assertThat(((WebApplicationException) thrown).getResponse().getStatus()).isEqualTo(500);
```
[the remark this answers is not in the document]
thiagohora: Fixed — renamed to "A parseable provider error is classified the same way, not just the unparseable path". Agreed, "split" was doing double duty as both the feature name and the verb.

apps/opik-backend/src/test/java/com/comet/opik/domain/llm/ChatCompletionServiceTest.java
```diff
@@ -627,16 +628,146 @@ void create__whenProviderErrorUnparsed__thenReportProviderStatus(
...
+            var thrown = catchThrowable(
+                    () -> chatCompletionService.scoreTrace(chatRequest, modelParameters, workspaceId));
+
+            assertThat(thrown)
+                    .isInstanceOf(InternalServerErrorException.class)
+                    .isNotInstanceOf(ClientErrorException.class);
+        }
+
+        @Test
+        @DisplayName("A parseable provider error is split the same way, not just the unparseable path")
```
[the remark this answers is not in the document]
thiagohora: Fixed by taking your second option: the expected outcome is now derived once and asserted through a single unconditional flow.

Rather than duplicate the data provider, both branches delegate to a shared `assertRetryability(thrown, status)` helper. The parameterized test body is now straight-line, and the helper carries `.as(...)` descriptions naming the category ("status %d can never succeed on retry, so it must be non-retryable"), so a failure identifies which side it belongs to without reading the branch.

Kept the single `providerStatusProvider` because those rows are real production shapes and both categories should keep being exercised by the shapes that actually occurred, rather than by a hand-split list that could drift.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringTraceThreadLlmAsJudgeScorer.java
```diff
@@ -124,7 +124,7 @@ protected Mono<Void> score(@NonNull TraceThreadToScoreLlmAsJudge message) {
                         .then(Mono.<Throwable>empty())
                         .onErrorResume(Mono::just))
                 .collectList()
-                .flatMap(errors -> errors.isEmpty() ? Mono.<Void>empty() : Mono.error(errors.getFirst()))
+                .flatMap(errors -> errors.isEmpty() ? Mono.<Void>empty() : Mono.error(representativeError(errors)))
```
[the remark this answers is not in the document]
thiagohora: Taken, in both scorers — and I checked the premise rather than assuming it, because it depends on whether fan-out is actually large.

It can be. Neither call site chunks the list: `ManualEvaluationService` passes every resolved thread id for a rule, and `TraceThreadOnlineScorerPublisher` passes every thread that closed in the window, grouped by rule. So a provider outage across a large fan-out is exactly the case where `collectList()` holds every sibling `Throwable` and its cause chain at once, only to discard all but one.

Now reduced pairwise via `OnlineScoringBaseScorer::preferRetryable`, keeping a single accumulator. Two things fall out that make it cleaner than the memory point alone suggests: it removes the second pass (`representativeError` re-scanning the list), and it removes the `errors.isEmpty()` branch, because an empty `Flux` reduces to an empty `Mono`, which *is* the no-failures case.

Selection semantics are unchanged: first retryable wins, else first failure.

One honest trade: the count of failed siblings is no longer recoverable, so if we ever want to log "N of M thread ids failed" that would need a separate counter. Nothing reports it today.

Applied to `OnlineScoringTraceThreadUserDefinedMetricPythonScorer` as well — same line, same fix.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/OnlineScoringBaseScorerErrorAggregationTest.java
```diff
@@ -0,0 +1,58 @@
...
+    void retryableWinsWhenPermanentArrivesFirst() {
+        // The ordering that used to lose data: getFirst() would have returned the ClientErrorException and
+        // the retryable sibling would never have been retried.
+        assertThat(OnlineScoringBaseScorer.representativeError(List.of(PERMANENT, RETRYABLE)))
+                .isSameAs(RETRYABLE);
+    }
+
+    @Test
+    @DisplayName("A retryable sibling still wins when it arrives first")
+    void retryableWinsWhenItArrivesFirst() {
```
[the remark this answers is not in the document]
thiagohora: Consolidated, and the merge earned its keep — it exposed a gap the pair had.

Now one `@ParameterizedTest` over a `@MethodSource` of (name, siblings, expected). While writing it I realised the original two cases could not distinguish "the first retryable wins" from "any retryable wins", since with one of each the two rules agree. Added a three-sibling case (`permanent, retryable, other-retryable`) that pins order-stability, plus the empty-sequence case.

Also switched from calling the accumulator directly to driving it through `Flux.reduce(...)` with `StepVerifier`, so the tests exercise it the way the scorers actually do — which is what makes the empty case meaningful (it has to stay empty rather than error).

Mutation-checked against the consolidated suite: always keeping the incumbent (the old `getFirst()` behaviour) fails 2 cases, always preferring a retryable candidate — which passes the original pair — fails the new order-stability case.

Kept `isSameAs` on the chosen `Throwable` here deliberately, unlike the cursor thread: the actual instance has to propagate so its stack trace and cause chain reach `BaseRedisSubscriber`, which classifies on the concrete type. Noted inline so the difference between the two files does not read as an inconsistency.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/BaseRedisSubscriberCursorTest.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Test
+    @DisplayName("A mid-scan position is carried forward unchanged, which is what walks past the first window")
+    void midScanPositionIsCarriedForward() {
+        // Redis caps each XAUTOCLAIM at COUNT * 10 entries EXAMINED, so without carrying this forward the
+        // scan only ever sees the first ~100 pending entries and anything behind them is never reclaimed.
+        var midScan = new StreamMessageId(1787891543627L, 0);
+        assertThat(BaseRedisSubscriber.nextCursor(midScan)).isSameAs(midScan);
+    }
```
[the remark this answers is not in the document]
thiagohora: Agreed, changed to `isEqualTo`.

`isSameAs` passed only because the implementation happens to return the same instance on the carry-forward path. That is not the contract — the contract is that the *position* is carried forward — so an equal-but-distinct `StreamMessageId` should pass, and under `isSameAs` it would have failed spuriously. Over-specified.

Noted the reasoning inline, since the sibling test in `OnlineScoringBaseScorerErrorAggregationTest` deliberately keeps `isSameAs`: there the actual `Throwable` instance must propagate so its stack trace and cause chain survive, so identity genuinely is the contract there.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/OnlineScoringBaseScorerErrorAggregationTest.java
```diff
@@ -0,0 +1,78 @@
...
+    @ParameterizedTest(name = "no failures -> empty ({0})")
+    @MethodSource("emptyCase")
+    @DisplayName("No failures reduces to empty, which is the success path rather than an error")
+    void noFailuresReducesToEmpty(String testName, List<Throwable> failures) {
+        StepVerifier.create(Flux.fromIterable(failures).reduce(OnlineScoringBaseScorer::preferRetryable))
+                .verifyComplete();
+    }
+
+    private static Stream<Arguments> emptyCase() {
+        return Stream.of(Arguments.of("every thread id succeeded", List.<Throwable>of()));
```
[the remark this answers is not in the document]
thiagohora: Fixed — plain `@Test` now, with the empty sequence inline as `Flux.<Throwable>empty()`.

You are right that it was pure indirection. It was an artefact of consolidating the ordering cases into a `@MethodSource` and then mechanically applying the same shape to this one, which has nothing to vary — leaving a `testName` parameter that was never read. Inlining also states the case more directly than an empty list threaded through a provider.

CometActions:

The diff as it stands is only the XAUTOCLAIM cursor — the retry-classification half moved out in 5b04ebf, so the title is now stale. The cursor fix is real and its blast radius is every BaseRedisSubscriber consumer (online scoring, webhooks, experiment items, thread closure): a pending list deeper than COUNT*10 grows a tail that is never re-examined. We are not proposing an e2e spec, because reproducing it needs a backlog of 100+ retryably-failing messages and then a way to watch the tail come back — and the suite has no receiver it controls that can be flipped from failing to answering mid-test (the same gap the taxonomy already records against alerts.test-alert), nor any control of consumerBatchSize / pendingMessageDuration on a test env. BaseRedisSubscriberTest.ClaimCursorTests already pins both halves against a real 150-deep PEL, which is the level where this is actually observable. For the record: the earlier agent draft #8151 carries a thread-judge spec, but it fans one message over two threads — PEL depth 1, so it would pass just as happily with this cursor reverted.

**Testable once (platform):** The e2e estate gains an endpoint the suite controls and can flip from failing to answering part-way through a test (the receiver alerts.test-alert is already waiting on), plus an environment where consumerBatchSize and pendingMessageDuration can be set so that a >100-entry backlog and one full retry pass fit inside a spec's runtime.

**What a test would assert:** Messages that fail retryably and sit past the first COUNT*10 entries of a consumer's pending list. With the default consumerBatchSize of 10 that is ~100 entries; beyond it, the old code re-examined the same head window forever and the tail was never redelivered. A user sees it as online-evaluation scores that never appear on traces past the head of a backlog (GET /v1/private/traces feedback_scores, and the score columns on the Traces table / trace side panel), and equally as alert webhooks never retried and experiment items never aggregated. The axis a test would have to vary is backlog depth — under ~100 pending entries old and new behave identically, which is why no existing spec could catch this.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/33883128190)

_Re-checked after a push on 04 Sep 14:21 UTC — nothing the verdict depends on changed._

CometActions:

### Access Information
- **URL:** https://pr-8137.dev.comet.com
- **Namespace:** pr-8137
- **Version:** 2.2.50-8137-merge-3184
- **Application logs:** [View in Grafana](https://grafana.dev.comet.com/explore?schemaVersion=1&panes=%7B%22ai4%22%3A%7B%22datasource%22%3A%22P8E80F9AEF21F6940%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Bnamespace%3D%5C%22pr-8137%5C%22%7D%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22P8E80F9AEF21F6940%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22direction%22%3A%22forward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1)

CometActions:
**Explored this PR on its own test environment.**

Worked through all 5 ranked items plus 4 follow-ups. **5 work, 2 partly blocked, 3 skipped, 0 suspicious.** Headline: the split behaves as designed — on the same trace, a rule whose provider answers a permanent 4xx is attempted **once and dropped**, while an otherwise identical rule whose provider fails transiently is **re-attempted 10 minutes later**. Seen twice, on two different traces, at two different reclaim ticks.

<details><summary>Per-item results</summary>

| # | Item | Verdict | What I saw |
|---|---|---|---|
| 1 | Seed a custom-llm provider + LLM-judge rule, confirm it scores normally first (control) | blocked (partly) | No mock gateway reachable and no provider key on this fresh OSS install, so **the LLM-judge success path is unverified**. Control established instead with a `user_defined_metric_python` rule on the same trace: scored `1.0`, proving the project's rules fire and the scoring pipeline is alive |
| 2 | Force a permanent status; provider called exactly once, one ERROR, no score | works | Provider `→ https://example.com/v1` answers **405**. Every one of 4 traces: exactly **1** `Sending … to LLM` line, **1** ERROR, **no score** — and still 1 apiece after two reclaim ticks that did re-deliver the transient rule's entries |
| 3 | Force a transient status (429, then 503); called more than once up to `maxRetries` | works (partly) | Provider `→ http://127.0.0.1:9/v1` (connection refused) is classified **500**. Trace `…2c73`: attempt 1 at 09:24:50, **re-delivered 09:34:55**. Trace `…f215`: 09:26:45, **re-delivered 09:37:25**. Both +10m = `onlineScoring.pendingMessageDuration`. **429/408/425 specifically was not exercised** — no reachable endpoint emits one |
| 4 | Confirm the two cases differ only by status | works | Same trace, same rule shape, same `{{output}}` mapping, same schema, same `sampling_rate: 1.0`; only `base_url` differs. `POST /chat/completions` with an identical body returns **405** for one provider and **500** for the other, so the classifier's input is pinned independently of the scoring path |
| 5 | What the Online evaluation / trace panel shows in each case | works | `automation-logs` page after two reclaim ticks: **4** `Sending` lines under the permanent rule (4 traces × 1) vs **6** under the transient rule (2 of them re-attempted). Trace panel → Feedback scores: exactly one score, the python control. Rules list shows all four rules `Enabled`, 100% — no health signal either way |
| 6 | Thread-scope LLM-judge scorer (changed file: `.collectList()` → `.reduce(preferRetryable)`) | works | Two threads closed in one batch `PUT /traces/threads/close`; both evaluated, each logged exactly one `Sending`/ERROR pair, neither thread got a score. The rewritten fold still propagates a failure and loses neither sibling |
| 7 | `preferRetryable` tie-break with mixed sibling classes | skipped | One rule has one model, so every thread sibling of a message fails identically. I could not construct a message whose siblings differ in class without hand-built state |
| 8 | `XAUTOCLAIM` cursor resumption | skipped | Matches the triage's own deferral: needs >100 entries sitting delivered-and-unacked in the consumer group's PEL, which this estate cannot produce, and the assertion is Redis-level, invisible to product APIs |
| 9 | `RedisStreamCodec` | skipped | Javadoc-only; `git diff` shows no non-comment line changed |
| Rule | Provider `base_url` | Status backend recovers | Attempts on that trace | Score |
|---|---|---|---|---|
| `…-perm-rule` | `https://example.com/v1` | 405 | **1** (09:24:50) | none |
| `…-transient-rule` | `http://127.0.0.1:9/v1` | 500 (no status on the wire) | **2** (09:24:50, 09:34:55) | none |
| `…-control` (python) | — | — | 1 | `1.0` |

</details>

2 flows look worth a permanent test:
* `online-evaluation.llm-judge-scores` — A permanent provider 4xx drops an LLM-judge evaluation after one attempt, while a retryable failure is re-attempted
* `online-evaluation.rule-scope-thread-span` — A thread-scope LLM-judge rule whose provider fails reports one failure per thread and writes no thread score

Writing the spec now; a draft PR will follow.

[Test env](https://pr-8137.dev.comet.com) · [Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/33856589404)

CometActions:
**Proposed a permanent test for this change.**

# Test proposal — opik PR #8137 exploration

**Draft PR: https://github.com/comet-ml/opik/pull/8151**
(`comet-qa-bot/OPIK-8240/thread-judge-provider-failure-e2e` → `thiagohora/OPIK-8240-scoring-retry-split-and-claim-cursor`, draft, no reviewers requested.)

Based on the source PR's own head branch rather than `main`: #8137 is open and its branch still
exists, so the spec sits on top of the change it covers. Written and run against commit
`ee62711f6ea743832c521b1916745ebcfe7c176f`.

## Specs written: 1

| Spec | Capabilities | Verification |
|---|---|---|
| `online-evaluation/online-evaluation-thread-judge-provider-failure.spec.ts` — two threads closed in one batch call each report their own provider failure, and neither is scored | `@cap:online-evaluation.rule-scope-thread-span` (was uncovered), `@cap:online-evaluation.automation-logs` (was uncovered) | **PASSED** on the first run, 22.7s, against `OPIK_BASE_URL=https://pr-8137.dev.comet.com` (workspace `default`). Passed again in the parallel feature-directory run. |

Supporting, additive: `backendClient.createTraceThreadLlmAsJudgeRule`, `backendClient.closeThreads`,
`providerKeys.createUnreachable`, a new `AutomationLogsPage` POM, and the matching `taxonomy.yaml`
update (spec added to the area's `specs:` list, both capabilities flipped to `covered: true` with
`tier: t2-cuj` and scoping notes).

Also green: `tag_lint.py` — `60 specs checked, 1 exempt, 0 problem(s)`; `tsc --noEmit` clean apart
from a pre-existing `Duplicate identifier 'deleteDashboard'` in `core/backend/client.ts`.

## Candidates dropped: 1 (of 2 strong)

- **"A permanent provider 4xx drops an LLM-judge evaluation after one attempt, while a retryable
  failure is re-attempted."** The headline of #8137, and the one I most wanted. Dropped because it
  could not be made into a spec that *runs*:
  - the suite's mock gateway has no forced-status hook, and is unreachable from any remote
    deployment (`mockAuthSkipReason`), so there is no controllable way to produce a permanent 4xx
    on the environment available — the exploration reached it only via a real external endpoint
    (`example.com` → 405), which a permanent spec should not depend on;
  - the retry half only becomes observable after `onlineScoring.pendingMessageDuration`, 10 minutes
    by default and 10 minutes on this environment, making it a ~25-minute test.

  It becomes proposable once the e2e compose sets `REDIS_SCORING_PENDING_MESSAGE_DURATION` (e.g.
  `30s`) **and** the mock gateway grows a per-model forced-status hook.

The exploration's own three `not_proposed` items (XAUTOCLAIM cursor resumption, `preferRetryable`'s
mixed-class tie-break, the transient-4xx half of the split) were not revisited — I agree with each
deferral, and nothing in this environment changes them.

## Worth someone's attention (not this PR's to fix)

`online-evaluation-python-metric-errors.spec.ts` fails on `https://pr-8137.dev.comet.com`: the
python evaluator answers `500 Internal Server Error: Failed to execute code` where the spec expects
the classified `400 Bad Request: Execution failed: the metric produced no output`. Confirmed
pre-existing — it fails identically with every change in the draft PR reverted to `ee62711`.

[Run](https://github.com/comet-ml/comet-automation-tests/actions/runs/33856589404)

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/BaseRedisSubscriber.java
```diff
@@ -412,14 +433,48 @@ private Mono<Map<StreamMessageId, Map<String, M>>> claimPendingMessages() {
...
+     * The cursor to use for the next scan, given the next-id Redis returned.
+     *
+     * <p>Compared numerically rather than against {@link StreamMessageId#MIN} / {@link StreamMessageId#ALL}:
+     * those are wire sentinels that serialize to {@code -} and {@code 0}, and neither is
+     * {@code equals()} to the {@code StreamMessageId(0, 0)} that Redisson parses Redis's literal
+     * {@code 0-0} end-of-pass reply into. Matching on the constants would therefore never fire, and the
+     * scan would run off the end of the PEL and stay there instead of wrapping.
+     */
+    // Package-private for unit tests.
+    static StreamMessageId nextCursor(StreamMessageId nextId) {
```
andrescrz: Minor: use VisibleForTesting annotation from Guava, instead of a comment.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringBaseScorer.java
```diff
@@ -154,6 +155,43 @@ protected List<Span> getSpansFromPreloadAndLogOverflow(@NonNull ThreadSpanPreloa
...
+     * to report -- nothing reports it today.
+     *
+     * <p>Order-stable: the first retryable failure wins, and if none is retryable the first failure wins.
+     *
+     * @param chosen the incumbent, i.e. the representative so far
+     * @param candidate the next sibling failure
+     * @return whichever should represent the batch
+     */
+    // Package-private for unit tests.
+    static Throwable preferRetryable(@NonNull Throwable chosen, @NonNull Throwable candidate) {
```
andrescrz: Minor: same about VisibleForTesting. In that case, this would be equivalent to a private method, so you could consider removing the non-null validations from the arguments, specially if you handle them. If handled, it's probably be better to change the return type to an optional.

Nevertheless, this looks more like a protected method used by subclasses, rather than something only visible for testing. It jus happens that those classes are in the same package, so they can use it in spite of not being protected.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringTraceThreadLlmAsJudgeScorer.java
```diff
@@ -123,8 +123,11 @@ protected Mono<Void> score(@NonNull TraceThreadToScoreLlmAsJudge message) {
                 .flatMap(threadId -> processThreadScores(message, threadId)
                         .then(Mono.<Throwable>empty())
                         .onErrorResume(Mono::just))
-                .collectList()
-                .flatMap(errors -> errors.isEmpty() ? Mono.<Void>empty() : Mono.error(errors.getFirst()))
+                // Reduced pairwise, not collected: fan-out is unchunked, so a provider outage across many
+                // thread ids would otherwise retain every sibling Throwable just to discard all but one.
+                // An empty Flux yields an empty Mono here, which is the no-failures case.
+                .reduce(OnlineScoringBaseScorer::preferRetryable)
+                .flatMap(Mono::<Void>error)
```
andrescrz: This change is based on the whole premise of over-retrying: f the whole batch contains a single retryable error, it should over retry. 

However, another premise is that base subscriber ack and removes everything, but the reality is that the base subscriber can have granularity to ack and remove only permanent failures, as long as the right processing result is emitted for each message. Therefore, it's not necessary that a single retryable message drags everything into retries.

As I see it, this is a problem in subscribers where they have an implementation that doesn't leverage the capabilities of the base subscriber and how it was designed to work.

The problem was already there, you're not re-introducing it. However, dragging everything (even successful messages? I'm not sure about that) into a retry might not only fix the problem, it might even cause a retry storm which is worse than losing recoverables errors due to not retrying.

I'd lean towards refactoring these online scorer and fit them into the right base subscriber design.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/BaseRedisSubscriberCursorTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Pure-function coverage for {@link BaseRedisSubscriber#nextCursor(StreamMessageId)}, the mapping that
+ * decides where the next {@code XAUTOCLAIM} scan resumes (OPIK-8240).
+ *
+ * <p>Separate from {@link BaseRedisSubscriberUnitTest} because these need no Redis mocks at all, and
+ * living under that class's Mockito setup would trip strict stubbing.
+ */
+@DisplayName("BaseRedisSubscriber claim-cursor mapping")
+class BaseRedisSubscriberCursorTest {
```
andrescrz: We shouldn't unit test this method. Also, dedicated top level class for unit testing a method aren't a good practice in Java, generally there's a 1:1 relation between class and test class, not between test class and method.

Nevertheless, you should better cover this with integration tests or unit tests (both exists for base redis subscriber).

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/OnlineScoringBaseScorerErrorAggregationTest.java
```diff
@@ -0,0 +1,76 @@
...
+ * ({@code errors.getFirst()}) could not change the outcome. Splitting permanent from transient made that
+ * arbitrary pick a race: a {@code ClientErrorException} arriving first would ack and remove the entry,
+ * taking any retryable sibling with it, unretried and unrecorded.
+ *
+ * <p>Exercised through {@link Flux#reduce} rather than by calling the accumulator directly, so the tests
+ * pin the behaviour as the scorers actually use it -- including that an empty sequence stays empty, which
+ * is the no-failures path.
+ */
+@DisplayName("Trace-thread sibling error aggregation")
+class OnlineScoringBaseScorerErrorAggregationTest {
```
andrescrz: Same feedback about this class. Better test this as high value integration tests first. If the condition can't be simulated, then as unit tests. But we should reuse the tests of the class, not adding a test class for individual methods.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/BaseRedisSubscriberUnitTest.java
```diff
@@ -767,6 +767,129 @@ private static PendingEntry pendingEntry(StreamMessageId messageId, long deliver
...
 
+    /**
+     * OPIK-8240. Redis caps each {@code XAUTOCLAIM} at {@code COUNT * 10} PEL entries <em>examined</em>,
+     * so a scan that always restarts at {@link StreamMessageId#MIN} only ever inspects the first ~100
+     * pending entries. A backlog above that grows a tail nothing ever reclaims: verified against a real
+     * Redis, entries beyond the window sat at delivery-count 1 for weeks while the head cycled. These
+     * tests pin the cursor being carried forward, which is what walks the rest of the PEL.
+     */
+    @Nested
+    class ClaimCursorTests {
```
andrescrz, line 835: This test class looks good, but I'm wondering if we can test this with an integration test, which really proofs the behaviour.

Unit tests are reserved for conditions, mostly errors, that can't be simulated. 

If you build the integration tests, remove the overlapping unit tests. No need for duplications.

apps/opik-backend/src/test/java/com/comet/opik/domain/llm/ChatCompletionServiceTest.java
```diff
@@ -627,36 +628,228 @@ void create__whenProviderErrorUnparsed__thenReportProviderStatus(
...
                     () -> chatCompletionService.scoreTrace(chatRequest, modelParameters, workspaceId));
 
-            // Then — InternalServerErrorException is absent from NON_RETRYABLE_EXCEPTIONS, which is what keeps the
-            // evaluation retryable; expectedStatus is deliberately unused here
             assertThat(thrown)
                     .isInstanceOf(InternalServerErrorException.class)
-                    .isNotInstanceOf(ClientErrorException.class)
-                    .hasMessageContaining(expectedMessagePart);
-            assertThat(((WebApplicationException) thrown).getResponse().getStatus()).isEqualTo(500);
+                    .isNotInstanceOf(ClientErrorException.class);
```
andrescrz: Minor: do not lose the status code assertion.

apps/opik-backend/src/test/java/com/comet/opik/domain/llm/ChatCompletionServiceTest.java
```diff
@@ -627,36 +628,228 @@ void create__whenProviderErrorUnparsed__thenReportProviderStatus(
...
+        }
+
+        @Test
+        @DisplayName("A parseable provider error is classified the same way, not just the unparseable path")
+        void scoreTrace__whenParseableProviderError__thenSplitByStatusToo() {
```
andrescrz: Parseable test cases should probably exercise retryable and non-retryable paths as well.

apps/opik-backend/src/test/java/com/comet/opik/domain/llm/ChatCompletionServiceTest.java
```diff
@@ -627,36 +628,228 @@ void create__whenProviderErrorUnparsed__thenReportProviderStatus(
...
+            // No HttpException anywhere, so there is nothing to outrank -- the mapped code is all we have.
+            var thrown = whenScoreTraceFails(new RuntimeException("no HTTP status in this chain"),
+                    Optional.of(new ErrorMessage(403, "forbidden per provider body")));
+
+            assertRetryability(thrown, 403);
+        }
+
+        @Test
+        @DisplayName("isPermanentFailure: the split itself, including the family boundary")
+        void isPermanentFailureSplitsTheFamilyCorrectly() {
```
andrescrz: This should be a properly parameterised test with input vs expected value.

apps/opik-backend/src/test/java/com/comet/opik/domain/llm/ChatCompletionServiceTest.java
```diff
@@ -627,36 +628,228 @@ void create__whenProviderErrorUnparsed__thenReportProviderStatus(
...
+            assertThat(thrown).hasMessageContaining(expectedMessagePart);
+            assertRetryability(thrown, expectedStatus);
+        }
+
+        /** Drives scoreTrace to failure and returns what it threw. */
+        private Throwable whenScoreTraceFails(RuntimeException providerFailure,
+                Optional<ErrorMessage> mappedProviderError) {
+            var chatRequest = ChatRequest.builder().messages(UserMessage.from("score this")).build();
+            var modelParameters = podamFactory.manufacturePojo(LlmAsJudgeModelParameters.class);
+            var workspaceId = "test-workspace-id";
```
andrescrz: Minor: better use random values for tests.

apps/opik-backend/src/test/java/com/comet/opik/domain/llm/ChatCompletionServiceTest.java
```diff
@@ -627,36 +628,228 @@ void create__whenProviderErrorUnparsed__thenReportProviderStatus(
...
+
+            return catchThrowable(
+                    () -> chatCompletionService.scoreTrace(chatRequest, modelParameters, workspaceId));
+        }
+
+        /**
+         * Asserts the thrown type against what {@code BaseRedisSubscriber} does with it: ClientErrorException
+         * is in NON_RETRYABLE_EXCEPTIONS (ack and remove on first delivery), anything else is replayed.
+         */
+        private void assertRetryability(Throwable thrown, int status) {
```
andrescrz: Minor: code smell, better two assertion methods.

andrescrz (commented):
This PR contains two deliverables.

The stream ID seems to be in better shape. Feel free to split it to deliver it faster.

See my feedback, mostly about the retry logic problems.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringTraceThreadLlmAsJudgeScorer.java
thiagohora: Agreed on all of it — reverting the change rather than defending it.

You are right that the base subscriber already has the granularity, and that the real defect is these scorers throwing it away before the decision is made. `preferRetryable` only picks a less-bad victim; it does not stop the collapse. Choosing between "discard retryable siblings" and "replay the whole batch" is a false choice created by `collectList()` + a single re-emitted error, and neither branch is right.

Your retry-storm point is the part I got wrong, and I understated it in the PR description. I framed the trade as one-sided — bounded replay versus silent data loss — but a replay drags in the thread ids that already succeeded, not just the failed ones. And fan-out is not chunked at either publisher: `ManualEvaluationService` passes every resolved thread id for a rule, `TraceThreadOnlineScorerPublisher` every thread that closed in the window. So the amplification scales with user behaviour, which makes "bounded" much weaker than I claimed.

Taking your framing that the problem pre-dates this PR: I have reverted `preferRetryable` and the `Flux.reduce` in both trace-thread scorers and restored `errors.getFirst()`, so #8137 goes back to being only the two fixes it is named for — the status-code retryability split and the XAUTOCLAIM cursor. Neither depends on the aggregation behaviour.

The refactor is filed as OPIK-8262, with your reasoning quoted and the unbounded fan-out noted for sizing. It covers making per-thread outcomes survive to the subscriber, deciding how a partially-failed fan-out is represented on a stream that carries one entry per message rather than per thread id, and ensuring a succeeded thread id is never re-scored by a sibling's retry.

The unit test I had added for the selector goes with it.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/BaseRedisSubscriber.java
thiagohora: Made it `private` instead, because your other comment removes the reason for the annotation: `BaseRedisSubscriberCursorTest` is deleted, so `nextCursor` has no direct test left and `@VisibleForTesting` would document a relaxation that nothing uses. Its only caller is the private `advanceCursorAndExtractMessages`.

Applied the annotation where it is still true: `ChatCompletionService.isPermanentFailure` is called directly from `ChatCompletionServiceTest` and now carries `@VisibleForTesting` in place of the same comment.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/BaseRedisSubscriberCursorTest.java
thiagohora: Deleted the class. The mapping is no longer unit tested at all, and `nextCursor` is now private.

The cursor behaviour it was standing in for is covered by a new integration test instead — `BaseRedisSubscriberTest.ClaimCursorTests` — see my reply on the unit-test thread for what it does.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/OnlineScoringBaseScorerErrorAggregationTest.java
thiagohora: Deleted, and not relocated: following your point on the scorer thread, the sibling-aggregation change itself is being reverted out of this PR — `preferRetryable` and the `Flux.reduce` go away and `errors.getFirst()` comes back — so there is no longer a method for this coverage to move to.

apps/opik-backend/src/test/java/com/comet/opik/api/resources/v1/events/BaseRedisSubscriberUnitTest.java
thiagohora: Agreed, and it turned out to be testable. Added `BaseRedisSubscriberTest.ClaimCursorTests.shouldReachPendingEntriesBeyondTheFirstScanWindowAndThenWrap` against the real container.

Setup: 150 entries published, all delivered to a consumer that never acks (a crashed pod's PEL), `consumerBatchSize` 10 so Redis's examine budget is 100, `claimIntervalRatio` 1 so every poll claims, and a processor that always fails retryably so nothing is acked and the PEL keeps its depth for the whole run — that last part is what makes the window a ceiling rather than a batch size. It then asserts all 150 are reached, and that the oldest is redelivered, which is the `0-0` end-of-pass wrap.

Verified it actually catches the bug: putting `StreamMessageId.MIN` back makes it time out on the tail. Runs in ~2s.

Removed the two unit tests it subsumes (`shouldResumeTheNextScanFromTheReturnedCursor`, `shouldWrapBackToTheStartAfterAFullPass`). Kept `shouldRetryTheSameWindowAfterAFailedScan` — the cursor must not advance past a window whose scan failed, and making a real Redis fail one `XAUTOCLAIM` mid-run and succeed on the next is not something the container exposes, so that one stays mocked.

apps/opik-backend/src/test/java/com/comet/opik/domain/llm/ChatCompletionServiceTest.java
thiagohora: Restored. It now goes through the new `assertRetryable` helper, which asserts the type and that the response status is 500 — so the assertion cannot be dropped again by one case drifting from the others.

apps/opik-backend/src/test/java/com/comet/opik/domain/llm/ChatCompletionServiceTest.java
thiagohora: Done. The single 429 case is now two parameterised tests over the parseable branch: `scoreTrace__whenParseablePermanentProviderError__thenNonRetryable` (400, 401, 403) and `scoreTrace__whenParseableTransientProviderError__thenRetryable` (408, 429, 500, 503). Both drive a failure with no `HttpException` in the chain, so the status can only come from `getLlmProviderError`.

This also absorbs the old `scoreTrace__whenNoWireStatus__thenFallsBackToTheMapper`, which was a one-off 403 doing the same thing.

apps/opik-backend/src/test/java/com/comet/opik/domain/llm/ChatCompletionServiceTest.java
thiagohora: Converted to `@CsvSource` rows of status vs expected boolean:

```java
@ParameterizedTest(name = "isPermanentFailure({0}) -> {1}")
@CsvSource({"400, true", "401, true", "403, true", "499, true",
            "408, false", "425, false", "429, false",
            "500, false", "503, false", "200, false", "302, false"})
void isPermanentFailure__classifiesTheStatus(int status, boolean expectedPermanent) {
    assertThat(ChatCompletionService.isPermanentFailure(status)).isEqualTo(expectedPermanent);
}
```

apps/opik-backend/src/test/java/com/comet/opik/domain/llm/ChatCompletionServiceTest.java
thiagohora: Done — the helper now builds both from `podamFactory`: the user message content and the workspace id. Neither is matched on (the stubs use `anyString()`), so nothing depended on the literals.

The scoreTrace cases that were building their own request inline now go through this helper too, so the hardcoded values are gone from those as well.

apps/opik-backend/src/test/java/com/comet/opik/domain/llm/ChatCompletionServiceTest.java
thiagohora: Split into `assertNonRetryable(thrown, expectedStatus)` and `assertRetryable(thrown)`.

Worth flagging how this reconciles with earlier bot feedback, which asked for one unconditional flow with no `if` in the parameterised body — that is what the branching helper was working around, badly. Both are satisfied by splitting the parameterised tests instead: each case is now either permanent or transient and calls one helper, with no conditional anywhere.

For the shared `providerStatusProvider` rows that meant two tests over a partition of the same source. The partition is a literal set in the test (`PERMANENT_STATUSES`), deliberately not a call to `isPermanentFailure` — a test that derives its expectation from the classifier under test cannot fail when the classifier is wrong, which was the deeper problem with the old helper.

thiagohora:
@andrescrz thanks — taking the retry feedback, and it re-scoped this PR.

**This PR is now the XAUTOCLAIM cursor fix only** (3 files). Your comments on `ChatCompletionServiceTest`, `OnlineScoringBaseScorer` and the trace-thread scorers are on code that has moved out, so those threads will look orphaned here — not dropped, and the specific asks are carried over.

**What moved, and why together.** You said the base subscriber already has per-message ack/remove granularity and the scorers are the ones not using it. Agreed — and following that through, the retry split and the fan-out collapse are coupled, so they go into **one** follow-up PR under OPIK-8262 rather than shipping separately:

The collapse is only a *correctness* problem because of the retry split. Before it, every provider failure became a blanket 500, so `errors.getFirst()` picking one of N siblings couldn't pick wrong — they were all retryable. After the split, a fan-out can mix a permanent 400 with a transient 429, and an arbitrary pick either drops retryable work or replays succeeded threads. Shipping the split without fixing the collapse would introduce that regression, so they belong in the same change.

**`preferRetryable` is removed, not carried forward.** You were right that it picks a less-bad victim rather than stopping the collapse, and right about the storm risk — I'd understated it, since a replay drags in the thread ids that already succeeded, and fan-out is unchunked at both publishers (`ManualEvaluationService` passes every resolved id, `TraceThreadOnlineScorerPublisher` every thread that closed in the window). `errors.getFirst()` is restored, matching main; the defect pre-dates this PR.

**Already applied here** (your items A/C/E): `nextCursor` is private again rather than annotated, since the method-only test class is gone; `BaseRedisSubscriberCursorTest` and `OnlineScoringBaseScorerErrorAggregationTest` are deleted; and the cursor behaviour is now covered by a real integration test on `BaseRedisSubscriberTest` against Testcontainers Redis — 150 entries past the 100-entry examine budget, asserting the tail is reached and the `0-0` wrap redelivers the oldest. It's verified to fail (30s timeout) if the cursor is reverted to `MIN`. The overlapping unit tests were removed; only `shouldRetryTheSameWindowAfterAFailedScan` stays, since a failing `XAUTOCLAIM` can't be induced on the container.

Your items F–J (the `ChatCompletionServiceTest` ones) are implemented and travel with the follow-up PR.

apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/BaseRedisSubscriber.java
```diff
@@ -412,14 +433,47 @@ private Mono<Map<StreamMessageId, Map<String, M>>> claimPendingMessages() {
...
+     * Advances {@link #claimCursor} to where Redis says the next scan should resume, then returns the
+     * batch this scan actually claimed.
+     *
+     * <p>A {@code null} or {@code 0-0} next-id means the pass reached the end of the pending list, so the
+     * cursor goes back to {@link StreamMessageId#MIN} and the following pass starts from the oldest entry
+     * again.
+     */
+    private Map<StreamMessageId, Map<String, M>> advanceCursorAndExtractMessages(AutoClaimResult<String, M> result) {
+        claimCursor = nextCursor(result.getNextId());
+        return result.getMessages();
```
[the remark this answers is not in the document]
thiagohora: Real gap, fixed — thanks. I had reset the cursor on NOGROUP, but only at the `claimPendingMessages` call site, and `recoverFromNoGroup` is reached from `readMessages` too. The read path is in fact the *likelier* of the two to notice NOGROUP first, since reads run on every tick that is not a claim tick, so the common case was the uncovered one.

Your "not guaranteed" is the important part, and it is worse than a deferred wrap on an idle stream. A scan starting above the recreated PEL only wraps once it exhausts the list, and on a live stream entries keep arriving *above* the stale position and keep giving it work at the high end. So the wrap can be deferred indefinitely while the lowest-ID entries are never examined — which is precisely the starvation this PR exists to remove, reintroduced through the recovery path.

Moved the reset into `recoverFromNoGroup` itself rather than adding a second one at the read call site: the cursor indexes the old group's PEL, so it is meaningless the moment the group is recreated regardless of which path noticed, and one owner beats two copies that can drift.

Regression test added on the read path specifically (`shouldResetTheClaimCursorWhenTheGroupIsRecreatedViaTheReadPath`) — first claim advances the cursor to `9000-0`, reads then fail with NOGROUP, and the next claim must start from `MIN` rather than resuming into the dead group. Mutation-checked by reintroducing exactly the bug you described (reset back at the claim site only): the test fails. 39 tests green across the two subscriber suites.

andrescrz (approved):
LGTM

PR #8147, opened 2026-09-04
[issue-7860] [SDK] fix: guard streaming usage helpers against IndexError and TypeError

### LINKED ISSUE #7860: [Bug]: LangChain usage extractors drop token usage when model metadata is absent
### Describe the bug

The LangChain usage extractors (`provider_usage_extractors` in the Python SDK) discard an already-extracted usage payload whenever the model-name or provider lookup raises. In `try_extract_provider_usage_data`, the orchestrator catches exceptions from `get_llm_usage_info` and returns `None`, so the span is emitted with no usage/cost even though the tokens were extractable.

Affected lookups (all unguarded against missing keys):
- `openai_usage_extractor`: `run_dict["extra"]`, `run_dict["outputs"]["generations"][-1][-1]["generation_info"]`, and `_get_provider`'s `run_dict["extra"]`
- `groq_usage_extractor`, `vertexai_usage_extractor`, `anthropic_vertexai_usage_extractor`, `google_generative_ai_usage_extractor`: `run_dict["extra"]`
- `anthropic_usage_extractor`: streaming path raises `IndexError` on empty generations (only `KeyError` is caught)

Reproduction: a streaming-shape OpenAI run whose usage sits in the message `usage_metadata` but whose generation dict has no `generation_info` key — `try_extract_provider_usage_data` returns `None` instead of the usage.

### Expected behavior

Usage extraction should not depend on model metadata being present: extract the usage and return it with `model=None` when the model name cannot be resolved.

Aftabbs:
## What

Fixes two bugs in the LangChain provider usage extractors that caused valid token usage to be silently discarded.

### Bug 1 — `IndexError` on empty `generations` in `try_get_streaming_token_usage`

`helpers.try_get_streaming_token_usage` accessed `run_dict["outputs"]["generations"][-1][-1]` unconditionally. When the generations list was empty (`[]`), this raised `IndexError`. The error propagated to the caller (`try_get_token_usage`), then to `anthropic_vertexai_usage_extractor._try_get_token_usage`'s broad `except Exception` block, which returned `None` — discarding any extractable usage.

**Fix:** wrap the index access in `try/except (IndexError, KeyError, TypeError)` and return `None` safely.

### Bug 2 — `usage_metadata` only searched inside `message["kwargs"]`

`try_get_streaming_token_usage` searched for `message["kwargs"]["usage_metadata"]`, but recent versions of `langchain-core` set `usage_metadata` directly on `message`, not inside `kwargs`. This caused the Anthropic-VertexAI streaming path to miss usage even when it was present.

**Fix:** probe `message.get("usage_metadata")` first, then fall back to `message["kwargs"].get("usage_metadata")`.

### Bug 3 — `TypeError` not caught in `anthropic_usage_extractor._try_get_model_name`

`_try_get_model_name` caught `(KeyError, IndexError)` but not `TypeError`. If `run_dict["outputs"]` was `None` or any non-subscriptable type, the inner access raised `TypeError`, which propagated through `get_llm_usage_info` to `try_extract_provider_usage_data`'s catch block — returning `None` and discarding valid usage data.

**Fix:** extend the except tuple to `(KeyError, IndexError, TypeError)`.

## Tests

Three new unit tests in `test_model_name_resilience.py`:
- `test_try_get_streaming_token_usage__empty_generations__returns_none` — empty `generations` list returns `None`, not `IndexError`
- `test_try_get_streaming_token_usage__null_outputs__returns_none` — `None` `outputs` returns `None`, not `TypeError`
- `test_try_get_streaming_token_usage__usage_metadata_on_message__returns_usage` — `usage_metadata` at the `message` level is extracted correctly

## Related

Closes #7860

## Checklist
- [ ] I have added unit tests for the changes
- [ ] The tests are all passing
- [ ] The code follows the style guidelines of this project

### DIFF
--- a/sdks/python/src/opik/integrations/langchain/provider_usage_extractors/anthropic_usage_extractor.py
+++ b/sdks/python/src/opik/integrations/langchain/provider_usage_extractors/anthropic_usage_extractor.py
@@ -89,7 +89,7 @@ def _try_get_model_name(run_dict: Dict[str, Any]) -> Optional[str]:
                 model = run_dict["outputs"]["generations"][-1][-1]["message"]["kwargs"][
                     "response_metadata"
                 ][model_name_key]
-        except (KeyError, IndexError):
+        except (KeyError, IndexError, TypeError):
             continue
 
     if model is None:
--- a/sdks/python/src/opik/integrations/langchain/provider_usage_extractors/langchain_run_helpers/helpers.py
+++ b/sdks/python/src/opik/integrations/langchain/provider_usage_extractors/langchain_run_helpers/helpers.py
@@ -23,15 +23,22 @@ def try_get_token_usage(run_dict: Dict[str, Any]) -> langchain_usage.LangChainUs
 def try_get_streaming_token_usage(
     run_dict: Dict[str, Any],
 ) -> Optional[langchain_usage.LangChainUsage]:
-    if "message" in run_dict["outputs"]["generations"][-1][-1]:
-        usage_metadata = run_dict["outputs"]["generations"][-1][-1]["message"][
-            "kwargs"
-        ].get("usage_metadata")
+    try:
+        last_gen = run_dict["outputs"]["generations"][-1][-1]
+    except (IndexError, KeyError, TypeError):
+        return None
 
-        if usage_metadata is not None:
-            return langchain_usage.LangChainUsage.from_original_usage_dict(
-                usage_metadata
-            )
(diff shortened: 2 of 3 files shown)


andrescrz:
Hi @Aftabbs 

Thanks for the contribution!

Thanks!

petrotiurin (dismissed):
Looks good, thanks for improving the error handling here!

alexkuzmik (approved):

PR #8151, opened 2026-09-04
[OPIK-8240] [QA] Proposed e2e spec from the #8137 exploration — thread-scope LLM judge over a failing provider

CometActions:
## Where this came from

Exploratory testing of **#8137** (`[OPIK-8240] [BE] fix: split provider-error retryability by status, resume XAUTOCLAIM from its cursor`), driven by hand against that PR's own deployed environment, `https://pr-8137.dev.comet.com` serving `2.2.50-6535`. The flow below was reached and observed there before any code was written for it; this PR turns the one candidate that could be made deterministic into a permanent spec.

**Base branch.** This targets `thiagohora/OPIK-8240-scoring-retry-split-and-claim-cursor`, not `main`, because that is the build the spec was written and run against (commit `ee62711f6ea743832c521b1916745ebcfe7c176f`). To be straight about it: the behaviour asserted here is not *new* in #8137 — it is regression cover for the fold that PR rewrote (`collectList().flatMap(errors.getFirst())` → `reduce(preferRetryable)`) in `OnlineScoringTraceThreadLlmAsJudgeScorer`, so it should hold on `main` too. If a reviewer would rather have it land independently of #8137, retargeting to `main` should be safe; it is based here so the spec sits on top of the change it covers.

## The spec

`tests_end_to_end/e2e/tests/online-evaluation/online-evaluation-thread-judge-provider-failure.spec.ts` — one test, `@t2-cuj`.

**Capabilities**

| `@cap:` | Why this spec earns it |
|---|---|
| `online-evaluation.rule-scope-thread-span` | Was `covered: false`. The spec creates a `trace_thread_llm_as_judge` rule and asserts it fires per thread. Flipped with a `note:` scoping it to **thread** scope — **span** scope is still uncovered and stays that way. |
| `online-evaluation.automation-logs` | Was deliberately `covered: false`, with a comment saying the existing specs read `GET /automations/evaluators/{id}/logs` but nobody opens the `/$workspaceName/automation-logs` **page**. This spec opens the page (new `AutomationLogsPage` POM) and asserts the rendered rows, so the comment no longer applies; it is replaced rather than deleted. |

**What it asserts.** Two threads are seeded, then closed in a **single** `PUT /v1/private/traces/threads/close` — that batch call is the point, because it makes the rule fan one stream message out over both thread ids, so the two evaluations are siblings of one message. Against a provider that cannot answer, the spec then asserts:

- exactly one `Evaluating threadId` line **per closed thread and no others** — asserted as the whole set, so a thread processed twice and a fan-out over a third thread both fail;
- exactly one provider call per thread, naming the rule's own model, so "it failed" cannot be satisfied by a rule that never got as far as a provider call;
- exactly one `ERROR` per thread, each raised *while scoring* that thread — no sibling's failure swallowed, none duplicated;
- both threads end up `inactive` and **neither carries a feedback score**;
- the same two failures render on the automation-logs page, one row per thread and no others.

It also asserts, before closing, that both threads exist and are still `active` — otherwise the evaluations could have been triggered by the 15-minute inactivity job rather than by the one call the test makes, and the fan-out would not be the shape under test.

**Verification: passed, on the first run, against the environment the exploration used.**

```
$ cd tests_end_to_end/e2e
$ npx playwright test tests/online-evaluation/online-evaluation-thread-judge-provider-failure.spec.ts \
    --reporter=list --workers=1
  ✓  1 [chromium] › …-thread-judge-provider-failure.spec.ts:27:7 › … (22.7s)
  1 passed (29.4s)
```

against `OPIK_BASE_URL=https://pr-8137.dev.comet.com`, workspace `default`. Also green: `tsc --noEmit` (clean apart from a pre-existing `Duplicate identifier 'deleteDashboard'` in `core/backend/client.ts`, untouched by this PR) and `tag_lint.py` — `60 specs checked, 1 exempt, 0 problem(s)`.

Because this adds a POM and touches shared seeding code, the whole feature directory was run too — `npx playwright test tests/online-evaluation/ --workers=2` → **7 passed, 1 skipped, 1 failed**. The new spec passed again there, under parallelism. The two others:

- *skipped*: `online-evaluation-smoke.spec.ts` LLM-judge case — no `ANTHROPIC_API_KEY` / `OPENAI_API_KEY` / `OPENROUTER_API_KEY` on this runner.
- *failed*: `online-evaluation-python-metric-errors.spec.ts` — the python evaluator on this environment answers `500 Internal Server Error: Failed to execute code` where the spec expects the classified `400 Bad Request: Execution failed: the metric produced no output`. **Pre-existing, not caused by this PR**: I re-ran that spec with every change here reverted to `ee62711` and it fails identically, with the same backend message. Flagging it because it is worth someone's attention on this deployment, not as something this PR should fix.

## Supporting changes, all additive

- `backendClient.createTraceThreadLlmAsJudgeRule` — a sibling of `createAutomationRule` rather than a `type` flag on it, because the two rule shapes share no `code` field. Same raw-REST/`Location`-header treatment, for the same reasons.
- `backendClient.closeThreads` — the batch close, through the public SDK.
- `providerKeys.createUnreachable` — a `custom-llm` provider whose base URL is the discard port on the backend's own loopback, so the connect is **refused in milliseconds**. Deliberately not the suite's mock gateway with a forced status: the mock binds to the test runner, which a remote deployment cannot reach (`mockAuthSkipReason` documents exactly this), so a mock-based failure spec could only ever run locally. Deliberately not a blackholed address either — that fails by *timing out*, which can leave the provider call in flight when the message is reclaimed. Teardown is the existing provider-key fixture's; no cleanup in the test body.
- `AutomationLogsPage` — columns addressed by `data-cell-id` suffix, not position: a marker column appears only when the rendered lines carry markers, so Message is the third cell for a trace-scope rule and the fourth for a thread-scope one.

## What I deliberately did not write

The exploration produced two strong candidates. **One is proposed here; one was dropped.**

- **Dropped — "a permanent provider 4xx drops an evaluation after one attempt, while a retryable failure is re-attempted."** This is the headline of #8137 and it is the one I most wanted. Two things stopped it being a spec that could be *run*:
  1. **No controllable failure status.** It needs one provider forced to a permanent 4xx and one forced to a retryable status. The suite's mock gateway has no forced-status hook at all, and it is unreachable from any remote deployment anyway. The exploration got there with real external endpoints (`example.com` → 405), which is not something a permanent spec should depend on.
  2. **The retry half is only observable after `onlineScoring.pendingMessageDuration`** — 10 minutes by default, and 10 minutes on the environment I ran against. That is a 25-minute test.

  It becomes writable once the e2e compose sets `REDIS_SCORING_PENDING_MESSAGE_DURATION` to something like `30s` *and* the mock gateway grows a per-model forced-status hook. Both are small; neither is in this PR's scope, and shipping a spec that skips on every environment we have would be worse than saying so.
- Already excluded by the exploration, and I agree: the `XAUTOCLAIM` cursor resumption (needs >100 delivered-and-unacked PEL entries and Redis-level assertions), `preferRetryable`'s tie-break with mixed sibling classes (one rule has one model, so every sibling fails identically — the product offers no per-sibling control), and the transient-4xx half of the split (408/425/429), which nothing reachable emits on demand.

Note also what this spec **cannot** distinguish, for the same reason the tie-break was skipped: with one model every sibling fails with the same class, so it pins "no sibling's failure is lost" but not *which* Throwable `preferRetryable` chooses.

## Please review before merging

Generated by the QA test-radar side flow from the #8137 exploration. It has been run and is green, but no human has read the assertions yet — that is what this draft is for. Not marked ready, no reviewers requested.

[OPIK-8240]: https://comet-ml.atlassian.net/browse/OPIK-8240?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests_end_to_end/coverage/taxonomy.yaml
+++ b/tests_end_to_end/coverage/taxonomy.yaml
@@ -680,6 +680,7 @@ areas:
       - online-evaluation/online-evaluation-python-metric-errors.spec.ts
       - online-evaluation/online-evaluation-sampling-rate.spec.ts
       - online-evaluation/online-evaluation-smoke.spec.ts
+      - online-evaluation/online-evaluation-thread-judge-provider-failure.spec.ts
       - online-evaluation/online-evaluation-thread-scope-batch-close.spec.ts
     capabilities:
       create-llm-judge-rule:   { covered: true,  tier: t1-smoke }
@@ -695,7 +696,15 @@ areas:
       # trace_thread_user_defined_metric_python rule, reads the discriminator
       # back, and drives it end to end through a batch close. `tier` records the
       # stronger of the two.
-      rule-scope-thread-span:  { covered: true,  tier: t2-cuj, note: "span/thread scope flags, both halves covered: span scope by online-evaluation-oversized-payload.spec.ts (t3), thread scope by online-evaluation-thread-scope-batch-close.spec.ts (t2) — six threads closed in ONE PUT /threads/close, each carrying exactly one score, with a mid-batch thread whose metric raises scoring nothing while its siblings do. Scoped: since #8162 the close fans out to one stream entry PER thread id, so what the spec pins is the grouped publish reaching every sampled id, not a scorer walking a multi-id message. A duplicate score WRITE is not observable — thread feedback scores upsert by name, so 'exactly once' is asserted as exactly one stored row per thread, held across a quiet period, not as insert multiplicity. Rule filters on either scope are a different key (rule-filters) and are still uncovered" }
+      #
+      # Thread scope is covered twice over, and deliberately: batch-close pins
+      # the SUCCESS path of the python metric shape, and
+      # online-evaluation-thread-judge-provider-failure.spec.ts (t2) pins the
+      # FAILURE path of the OTHER thread-scope rule shape
+      # (trace_thread_llm_as_judge). Neither subsumes the other — a fold that
+      # kept one sibling's failure, or one that never terminated the message,
+      # is invisible to a spec whose provider answers.
+      rule-scope-thread-span:  { covered: true,  tier: t2-cuj, note: "span/thread scope flags, both halves covered: span scope by online-evaluation-oversized-payload.spec.ts (t3), thread scope by online-evaluation-thread-scope-batch-close.spec.ts (t2) — six threads closed in ONE PUT /threads/close, each carrying exactly one score, with a mid-batch thread whose metric raises scoring nothing while its siblings do. Scoped: since #8162 the close fans out to one stream entry PER thread id, so what the spec pins is the grouped publish reaching every sampled id, not a scorer walking a multi-id message. A duplicate score WRITE is not observable — thread feedback scores upsert by name, so 'exactly once' is asserted as exactly one stored row per thread, held across a quiet period, not as insert multiplicity. Thread scope is also covered for the trace_thread_llm_as_judge shape by online-evaluation-thread-judge-provider-failure.spec.ts (t2), over an unreachable provider: two threads closed in one call, each reporting its OWN provider failure exactly once and neither carrying a thread score. Rule filters on either scope are a different key (rule-filters) and are still uncovered" }
       rule-filters:            { covered: false }
       sampling-rate:           { covered: true,  tier: t2-cuj, note: "50% rule vs 100% control over one 30-trace batch, binomial band 15-85%; plus a 0%-rate rule at trigger_scope=both, which must skip every SDK trace and still score experiment/playground/optimization ones" }
       clone-rule:              { covered: false }
@@ -722,12 +731,11 @@ areas:
       edit-rule:               { covered: true,  tier: t2-cuj, note: "plain-string prompt: edit-dialog hydration + no-op re-save, byte-for-byte; content_array prompt: API-only read-then-write re-save, no dialog gesture; model custom_parameters: rule seeded over REST, saved unchanged through the edit dialog, PATCH body and persisted blob both asserted to carry the thinking block and an unrelated free-form key; editing a field through the dialog is not covered" }
       enable-disable-rule:     { covered: true,  tier: t2-cuj, note: "edit-dialog switch; control rule proves scoring stopped, then resumed" }
       delete-rule:             { covered: true,  tier: t2-cuj, note: "row kebab delete; control rule proves scoring stopped" }
-      # Deliberately still false. online-evaluation-python-metric-errors.spec.ts
-      # asserts a rule's log stream through GET /automations/evaluators/{id}/logs,
-      # but this key names the /$workspaceName/automation-logs PAGE, which no spec
-      # opens. Flipping it on the strength of an API read would mark a UI nobody
-      # tested as tested for good.
-      automation-logs:         { covered: false }
+      # Flipped by online-evaluation-thread-judge-provider-failure.spec.ts, which
+      # opens the /$workspaceName/automation-logs PAGE itself (AutomationLogsPage)
+      # and asserts the rendered ERROR rows — not, as the API-only specs do, the
+      # GET /automations/evaluators/{id}/logs stream behind it.
+      automation-logs:         { covered: true,  tier: t2-cuj, note: "rule_id-scoped log stream renders: one ERROR row per failed thread and no others" }
 
     visual:
       online-evaluation-empty: { covered: true,  state: empty,   spec: "empty-states.spec.ts E11" }
--- a/tests_end_to_end/e2e/core/backend/client.ts
+++ b/tests_end_to_end/e2e/core/backend/client.ts
@@ -2232,6 +2232,76 @@ export function makeBackendClient(apiKey: string | null = null, workspaceName: s
       }
       return id;
     },
+    /**
+     * Create a THREAD-scope LLM-as-judge rule and return its id.
+     *
+     * A sibling of `createLlmJudgeRule` rather than a flag on it: the two are
+     * the same `code` shape but not the same contract. A trace-scope judge
+     * binds `variables` to fields of one trace; a thread-scope judge is handed
+     * the rendered conversation under the scorer's own `{{context}}` and takes
+     * no `variables` at all, so one builder would carry a field that is
+     * meaningless in half its calls.
+     *
+     * Raw REST for the same two reasons as its siblings: creation answers 201
+     * with an empty body, so the id exists only in the `Location` header that
+     * the pinned SDK's `void` return discards, and reading the id back by name
+     * would silently pick up a rule an earlier run left under the same
+     * namespace.
+     */
+    async createTraceThreadLlmAsJudgeRule(args: {
+      projectId: string;
+      name: string;
+      /** Fraction in [0, 1], the backend's own units — not the dialog's percentage. */
+      samplingRate: number;
+      /** Fully-qualified model id, e.g. `custom-llm/<providerName>/<modelName>`. */
+      model: string;
+      /** Score name the judge's output schema declares, and the score it would write. */
+      scoreName: string;
+      enabled?: boolean;
+    }): Promise<string> {
+      const { status, message, location } = await rawFetch(
+        'POST',
+        '/v1/private/automations/evaluators/',
+        {
+          body: {
+            type: 'trace_thread_llm_as_judge',
+            action: 'evaluator',
+            name: args.name,
+            project_ids: [args.projectId],
+            sampling_rate: args.samplingRate,
+            enabled: args.enabled ?? true,
+            code: {
+              model: { name: args.model, temperature: 0 },
+              // `{{context}}` is the thread scorer's own variable for the
+              // rendered conversation — the judge has to reference it or the
+              // rule would be evaluating an empty prompt.
+              messages: [
+                {
+                  role: 'USER',
+                  content: `Rate this conversation from 0 to 1: {{context}}`,
+                },
+              ],
+              schema: [
+                { name: args.scoreName, type: 'DOUBLE', description: 'thread score' },
+              ],
+            },
+          },
+        },
+      );
+      if (status !== 201) {
+        throw new Error(
+          `createTraceThreadLlmAsJudgeRule: expected 201 for '${args.name}', got ${status}: ${message}`,
+        );
+      }
+      const id = location?.split('/').filter(Boolean).pop();
+      if (!id) {
+        throw new Error(
+          `createTraceThreadLlmAsJudgeRule: 201 for '${args.name}' carried no usable Location ` +
+            `header (got '${location}') — cannot address the rule.`,
+        );
+      }
+      return id;
+    },
 
     /**
      * The `code.model` block of an `llm_as_judge` rule.
--- a/tests_end_to_end/e2e/core/provider-keys.ts
+++ b/tests_end_to_end/e2e/core/provider-keys.ts
@@ -60,6 +60,12 @@ export async function createProviderKey(payload: {
   provider: string;
   provider_name: string;
   base_url: string;
+  /**
+   * Static bearer, for providers that are not in OAuth2 token-auth mode. The
+   * endpoint requires one whenever `auth_config` is absent, so a static-auth
+   * provider cannot be created without it.
+   */
+  api_key?: string;
   configuration?: Record<string, string>;
   auth_config?: ProviderAuthConfig;
 }): Promise<void> {
--- a/tests_end_to_end/e2e/fixtures/index.ts
+++ b/tests_end_to_end/e2e/fixtures/index.ts
@@ -1,6 +1,7 @@
 export { test, expect } from './summarised-datasets.fixture';
 export type {
   OauthProviderSeed,
+  UnreachableProviderSeed,
   ProviderKeysFixture,
   ProviderKeyFixtures,
 } from './provider-key.fixture';
--- a/tests_end_to_end/e2e/fixtures/provider-key.fixture.ts
+++ b/tests_end_to_end/e2e/fixtures/provider-key.fixture.ts
@@ -14,12 +14,46 @@ export interface OauthProviderSeed {
   modelName?: string;
 }
 
+export interface UnreachableProviderSeed {
+  providerName: string;
+  /** Model name; unique per test so parallel specs never collide. */
+  modelName?: string;
+}
+
+/**
+ * Base URL for a provider that can never answer.
+ *
+ * The discard port on the backend's own loopback: nothing listens there, so the
+ * connect is REFUSED immediately and every scoring call through this provider
+ * fails in milliseconds, with no dependency on the mock gateway, on an external
+ * host, or on the runner being reachable from the backend at all. That last
+ * point is why this is not `mockGatewayUrlForBackend` with a forced status: the
+ * mock binds to the test runner, which a remote deployment cannot reach (see
+ * `mockAuthSkipReason`), so a mock-based failure spec can only ever run locally.
+ *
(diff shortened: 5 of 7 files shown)


CometActions:
## 🔍 Generated-test review

# PR #8151 — [OPIK-8240] [QA] Proposed e2e spec from the #8137 exploration — thread-scope LLM judge over a failing provider

CometActions · draft: **yes** · base `main` · 1 new spec, 1 new POM, 5 shared files
Reviewed at head `588b3d725b`; **restacked + fixed and pushed as `5b0c26985b`**.

## VERDICT: ready with notes — but do not merge until the `AutomationLogsPage` collision below is decided

The spec itself is good: the tags are honest, the assertions are set-shaped rather
than `find()`-shaped, teardown is entirely in fixtures, and it **passes against the
local instance built from this PR's head**. The restack dropped four inherited
backend commits that had already squash-merged to `main`, so the branch now touches
only `tests_end_to_end/` and merges cleanly.

The one thing that should stop a merge is not in this PR's diff: **three open
generated PRs each add a different `AutomationLogsPage` class at the same path.**

---

## Blockers (1)

### 1. Three open PRs add mutually incompatible `AutomationLogsPage` at the same path
`tests_end_to_end/e2e/pom/automation-logs.page.ts`

This is the reverse case git will not warn about until the second one merges, and
then only as an add/add conflict that cannot be resolved by keeping both — it is one
class at one path.

| PR | Shape |
|---|---|
| **#8151** (this one) | `rows()` method, `rowsAtLevel(level: AutomationLogLevel)`, `rowsForThreadAtLevel(level, threadId)`, 1-arg constructor, 96 lines |
| **#8192** `[OPIK-8262] e2e: a failing LLM judge's provider error on the automation logs page` | `get rows`, `rowsWithLevel(level: string)`, extra `RenderedLogRow` interface, multi-arg constructor, 136 lines |
| **#8172** `[OPIK-8262] e2e: provider error classification on the LLM-judge path` | `get rows`, `rowsWithMessage(text, level?)`, 73 lines |

Whichever lands first defines the POM; the other two then either conflict or, if
someone resolves by taking one side, **compile-break the other PRs' specs**. Both
#8192 and #8151 additionally touch `provider-key.fixture.ts`, `core/provider-keys.ts`
and `fixtures/index.ts` — #8192 is near-certainly adding its own unreachable/failing
provider seed alongside this PR's `createUnreachable`.

**Not fixable inside this PR.** A human needs to pick one POM (and one provider-seed
fixture) and rebase the other two onto it. I did not touch the other PRs.

Five open PRs also all edit `taxonomy.yaml` in the **online-evaluation** area
(#8151, #8172, #8192, #8209, #8225, plus the wholesale #8144 taxonomy-discovery PR).
Those are plain-union conflicts and are cheap; the POM is the expensive one.

---

## Conflicts — restack and resolution

`RESTACK: partial`. Four inherited commits, six conflicting files at the start; the
restack dropped the inherited commits and left two conflicts, **both inside
`tests_end_to_end/`**, which were mine to resolve.

### Dropped inherited commits (4) — correct, verified
`2b960456fa`, `8a4a8ec673`, `d5b656fbb9`, `ee62711f6e` — the branch of PR **#8137**,
which squash-merged to `main` as `df66b23dd4 [OPIK-8240] [BE] fix: Resume XAUTOCLAIM
from its cursor (#8137)`. They were replaying already-merged production code, which
is why all four backend conflicts (`BaseRedisSubscriber.java`,
`ChatCompletionService.java` and their tests) appeared. Dropping them is right, and I
resolved nothing outside the test tree.

One consequence worth stating: `preferRetryable` — the sibling-failure fold in
`OnlineScoringBaseScorer.java` that the PR body names as the thing this spec covers —
**is not on `main`**. It was in the dropped commits and did not survive the squash.
The spec still passes without it (see the run below), so what it actually pins is the
observable contract "each sibling reports its own failure", not that particular
implementation. The PR body's claim to the contrary is now stale (see Should-fix 1).

### Resolved conflict 1 — `tests_end_to_end/e2e/core/backend/client.ts`
- **Base side:** added `createLlmJudgeRule` (trace-scope judge) in exactly the slot
  this branch used, plus `getLlmJudgeModel` and `resaveAutomationRuleFromReadBack`.
  It also added its **own `closeThreads`**, from #8202.
- **PR side:** added `createTraceThreadLlmAsJudgeRule` and a near-identical
  `closeThreads`.
- **Resolution kept:** main's file whole, plus this branch's
  `createTraceThreadLlmAsJudgeRule` re-sited after `createLlmJudgeRule` so the two
  judge builders sit together. The conflict opened mid-docblock, so the two `/**`
  runs were bridged rather than concatenated. This branch's `closeThreads` was
  **dropped, not duplicated** — main's is functionally identical and better
  documented. The resolved file differs from `origin/main` by exactly the 70-line
  new method and nothing else (verified by diff). The doc comment was updated to
  explain the split against `createLlmJudgeRule` (its real sibling now) rather than
  against `createAutomationRule`.

### Resolved conflict 2 — `tests_end_to_end/coverage/taxonomy.yaml`
- **`specs:`** — plain union, kept alphabetical.
- **`list-rules`** — base has `covered: true` (from #8127); the PR side's
  `covered: false` was just the older state, not a rewrite. Base's kept.
- **`rule-scope-thread-span`** — both sides rewrote the `note:`, and each described
  real coverage the other could not see. Base: span scope by
  `online-evaluation-oversized-payload.spec.ts` (t3) + thread scope by
  `online-evaluation-thread-scope-batch-close.spec.ts` (t2, the python-metric
  **success** path). This PR: thread scope for the *other* rule shape,
  `trace_thread_llm_as_judge`, on its **failure** path. Neither subsumes the other,
  so the note unions both and a comment says why. **No capability key was invented.**
- `automation-logs` merged cleanly and its flip to `covered: true` stands — see
  Notes 1.

---

## Should fix (2)

### 1. The PR description is stale in three places — not fixed, needs an author edit
The body was not in the fix scope I was given, and it is an outward-facing edit, so I
left it. It now says things that are false:
- *"This targets `thiagohora/OPIK-8240-scoring-retry-split-and-claim-cursor`, not
  `main`"* — the base **is** `main`, and the branch now sits on it.
- The **"Supporting changes"** list still claims `backendClient.closeThreads` as this
  PR's; main already has it. And the verification block reports `60 specs checked`
  and a green run against `https://pr-8137.dev.comet.com`, superseded by the run
  below.
- The closing note *"pins 'no sibling's failure is lost' but not which `Throwable`
  `preferRetryable` chooses"* should say that `preferRetryable` never merged.

### 2. `tsc --noEmit` cannot pass on this estate at all — estate-wide, own PR
`tsconfig.json:13` sets `baseUrl`, which **TypeScript 7.0.2** (the pinned
`^7.0.2`) removed: `error TS5102`. It aborts before compiling anything.

I verified this is **not** this PR: a clean `origin/main` worktree with the same
`node_modules` fails identically, and the PR touches neither `tsconfig.json` nor
`package.json`. The fix is a one-liner (`"paths": { "@e2e/*": ["./*"] }` without
`baseUrl`) but it belongs in its own PR, since it unblocks the gate for every open
branch at once. Happy to open it.

With `baseUrl` removed in a probe config, **this PR's own files produce zero new type
errors.** The remaining 94 are pre-existing: missing `types: ["node"]` under TS7, and
a duplicate `deleteDashboard` at `client.ts:936`/`1591` that is already on `main`.

---

## Notes (3)

1. **Tag honesty — both `@cap:` claims hold.** `rule-scope-thread-span`: the spec
   creates a real `trace_thread_llm_as_judge` rule and asserts the per-thread fan-out,
   so yes. `automation-logs`: main's taxonomy comment said this key was *deliberately*
   false because the API specs read `GET /evaluators/{id}/logs` while nobody opened the
   `/$workspaceName/automation-logs` **page**. This spec does open the page, through a
   POM, and asserts rendered rows — so the flip is earned, not the usual "API-only spec
   claims a UI capability". The old comment is replaced rather than deleted. Tier
   `@t2-cuj` is right: ~51s, no provider spend (the provider is unreachable by design).
2. **`UNREACHABLE_BASE_URL = 'http://127.0.0.1:9/v1'`** relies on nothing listening on
   the *backend's* discard port. Left as-is: if something ever did answer, no ERROR
   line would appear and the poll fails loudly with its own message — it degrades to a
   red run, never to a false green. The fixture already documents why the mock gateway
   and a blackholed address were both rejected.
3. **`client.ts` and `taxonomy.yaml` have moved three times under this PR** (#8080,
   #8127, #8202). #8202 is the sibling that caused today's conflict. These two files
   are the estate's hottest contention point for generated online-evaluation PRs.

---

## What I fixed and pushed

Commit `5b0c26985b`, force-pushed with `--force-with-lease` pinned to the pre-restack
head `588b3d725b` (the branch history was rewritten by the restack, so a plain push
would have been rejected; the lease was accepted, so nobody else had pushed).

1. **Exactly-once asserted over a settled stream, not the poll's own snapshot.**
   The poll captured `stream` at the moment both threads first showed an ERROR and the
   three count assertions ran over that stale array — while the page assertion at the
   end re-read live state. A retryable provider failure replayed from `XAUTOCLAIM`
   appends rows between the two, so the API half could pass over a stream that was
   still growing and then be contradicted by the UI half. Both halves now read one
   stream after a `QUIET_PERIOD_MS` settle, matching what
   `online-evaluation-thread-scope-batch-close.spec.ts` already does on main, with the
   same justification for why a duration is the only way to assert "nothing changed
   over an interval". This **strengthens** the assertion — and it still passes.
   Test timeout 240s → 300s to fit the settle.
2. **`AutomationLogLevel` widened** from `INFO|WARN|ERROR` to all five levels the
   product defines (backend `LogItem.LogLevel` and frontend `EVALUATOR_LOG_LEVEL` both
   carry `DEBUG` and `TRACE`).
3. **Constants renamed** `EVALUATING_LINE`/`SENDING_LINE`/`SCORING_FAILURE_LINE` →
   `EVALUATING_PREFIX`/`SENDING_PREFIX`/`SCORING_FAILURE_PHRASE`, matching how each is
   actually matched (`startsWith` vs `toContain`).
4. The two conflict resolutions above.

Nothing was skipped, weakened, or dropped from the PR.

---

## Spec run — local instance at this PR's head

`OPIK_DEPLOYMENT=oss`, workspace `default`, `http://localhost:5173`, **run after all
fixes**:

```
npx playwright test tests/online-evaluation/online-evaluation-thread-judge-provider-failure.spec.ts \
  --reporter=list --retries=0

✓ 1 [chromium] › online-evaluation-thread-judge-provider-failure.spec.ts:47:7 ›
    Two threads closed in one call each report their own provider failure, and
    neither is scored
    @t2-cuj @area:online-evaluation @cap:online-evaluation.rule-scope-thread-span
    @cap:online-evaluation.automation-logs (51.1s)

1 passed (1.1m)
```

Global teardown swept cleanly (`cuj-20260910-024833-168-`, nothing left behind), which
confirms the namespacing matches the estate's timestamp teardown contract.

The pass is meaningful beyond "green": it shows the spec's exactly-once claim holds
over a **settled** stream on a build where `preferRetryable` is absent, so the spec is
pinning the observable contract rather than a dropped implementation detail.

**What I could not verify:** the retry half of #8137 (needs
`pendingMessageDuration` ≈ 10 min, as the PR body already says); the tie-break between
sibling failures of different classes (one rule has one model, so every sibling fails
identically — the product offers no per-sibling control); and whether the three
`AutomationLogsPage` variants can be reconciled without changing the other two PRs'
specs, which I did not open.

---

## Bot comments

9 findings from `baz-reviewer[bot]`. Three are on backend files that the restack
dropped from this branch.

|---|---|---|---|
| 1 | `*_LINE` names misrepresent prefix-vs-phrase matching | **valid** | **fixed** in `5b0c26985b` — renamed to `_PREFIX`/`_PREFIX`/`_PHRASE` |
| 2 | Exact-once assertions use a pre-replay snapshot | **valid** — and the sharper form is that it contradicts the page assertion that reads later | **fixed** in `5b0c26985b` — settle + single re-read, following `online-evaluation-thread-scope-batch-close.spec.ts` |
| 3 | `createTraceThreadLlmAsJudgeRule` duplicates `createAutomationRule`'s 201/`Location` handling | **valid but estate-wide** | **deferred → own PR.** `createAutomationRule` *and* `createLlmJudgeRule` already carry the identical block on `main`; extracting it only for the new method makes this method the inconsistent one. Offer stands to open the extraction PR across all three |
| 4 | Port 9 could hit an unintended listener | **valid in the abstract, not reachable as a false green** | **not fixed, documented.** A listener would produce no ERROR line and the poll fails loudly with its own message. See Notes 2 |
| 5 | `AutomationLogLevel` omits `DEBUG`/`TRACE` | **valid** | **fixed** in `5b0c26985b` — union widened to all five |
| 6 | POM query methods omit `test.step()` | **invalid** | `rows`/`rowsAtLevel`/`rowsForThreadAtLevel` are synchronous **Locator builders**, not operations. `grep` over `pom/*.page.ts` finds **zero** of the estate's Locator getters wrapped in `test.step` — the convention is that steps wrap actions. Applying this would force them async and add noise to every trace |
| 7 | `preferRetryable` misclassifies vs `BaseRedisSubscriber` | **moot here** | The commit is dropped from this branch and `preferRetryable` never merged to `main`. If the fold is re-proposed, this finding is worth carrying to that PR — it is a reasonable read |
| 8 | Source-controlled `api_key` literal | **invalid** | `'unused-the-connection-is-refused'` is a placeholder, not a credential. The endpoint requires the field whenever `auth_config` is absent, and the connection is refused before a byte is written. Nothing is authenticated by it |
| 9 | Error aggregation duplicated across the two thread scorers | **moot here** | Backend, in the dropped commits. Same disposition as #7 |

**Two follow-through actions I have not taken, awaiting your go-ahead:**
1. **Reply on threads 3, 4, 6, 7, 8, 9** so the findings read as answered rather than
   ignored. I can draft one or two sentences each.
2. **File two tickets**: (a) the shared POST-201/`Location` helper across all three
   `create*Rule` builders (finding 3), and (b) the `tsconfig baseUrl` removal that
   unblocks `tsc` estate-wide (Should-fix 2).

---

## Estate gates

Run at `5b0c26985b`, after every edit.

| Gate | Before | After | |
|---|---|---|---|
| `tag_lint.py` | FAIL — taxonomy unparseable, conflict markers at line 679 | **PASS** | `74 specs checked, 1 exempt, 0 problem(s)` |
| `tsc --noEmit` | FAIL — 14× TS1185 conflict markers in `client.ts` | **FAIL** | Conflict markers gone. Now fails on `TS5102 baseUrl` — **pre-existing, reproduced identically on a clean `origin/main` worktree.** This PR's own files add zero type errors |
| `playwright test --list` | FAIL — `client.ts` did not parse, every spec uncollectable | **PASS** | `140 tests in 71 files` estate-wide; the new spec collects |

`gh pr view 8151` now reports `mergeable=MERGEABLE`, `isDraft=true`.

---

**The PR is still a draft, and I did not mark it ready or merge it.** Once someone
decides who owns `AutomationLogsPage` across #8151 / #8172 / #8192, and the author
corrects the description, this is ready for a person to flip.

### ✅ Post-fix verification

The specs this PR touches were re-run against staging after the fixes were pushed: **1 passed**.

---
_[review_generated_tests.yml](https://github.com/comet-ml/comet-automation-tests/actions/runs/34429256664) ·


[OPIK-8240]: https://comet-ml.atlassian.net/browse/OPIK-8240?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

AndreiCautisanu (approved):

Issue #8134, opened 2026-09-03
Evaluation pipeline: failed metrics silently dropped from averages; SUT output unescaped in judge prompts; g_eval logprob scoring misparses digit tokens; error handler imports SDKs inside except block


AUTHENSOR:
## Summary

Four defects in opik's evaluation pipeline at pin c7ffd4e.

### 1. Failed metric scores silently dropped from experiment averages

Any metric exception produces ScoreResult(0.0, scoring_failed=True), those rows are never uploaded, and every aggregate excludes them. A crashing-judge run reports hallucination mean 0.0 over 5 survivors while an honest judge reports 0.5 over 10 on the same outputs.

### 2. System-under-test output interpolated unescaped into judge prompts

Judge templates interpolate SUT output verbatim; an output embedding a JSON object with score 0.0 is parsed as the judge response itself.

### 3. g_eval logprob scoring averages digit-string token candidates

Score 0 parses as 0.2999995; score 10 split into two tokens parses as 0.05. A hardcoded token index assumes single-digit scores.

### 4. Metric-error handler imports provider SDKs inside the except block

When openai/litellm are absent (the source-tree usage case), every tolerated metric error aborts the run instead of continuing.

All four verified with deterministic stubs. Harnesses available.

feiiiiii5:
Hi — I'd like to take **defect 1** (failed metric scores silently dropped from experiment averages). I can work around the crash and produce deterministic stub tests similar to what you describe, so I'll keep to that one slice rather than bundling all four.

Proposed shape, for maintainer sign-off on the semantics before I open a PR: upload `scoring_failed` ScoreResults instead of filtering them out at upload, and have every aggregate that today shrinks its denominator keep the mean over valid scores but surface the failure count explicitly (`failed_scores`), so a crashing-judge run can no longer masquerade as a low mean over a smaller sample. Per Scout's note, OPIK-6925 already handles the item level — this would complete the aggregate side of the same gap.

One direction question: is the intended aggregate semantics (a) mean over valid scores + explicit failed count (my recommendation — never silently shrink the denominator), or (b) failed scores included in the mean as zeros? Happy to follow whichever the team wants.

tonydzi:
hi, this is Mycroft, Anton's synthetic AI cofounder — I label myself so nobody has to play detective.

[...]

PR #8152, opened 2026-09-04
[issue-8134] [SDK] fix(g-eval): locate the score token by content instead of a fixed offset

feiiiiii5:
## Details

- The G-Eval logprob scorer located the score digit at a hardcoded token offset and assumed single-digit scores. A two-digit score (`"score": 10`) tokenizes across two tokens, so only the first digit was averaged: a confident **10 parsed as ~0.099**. A tokenizer folding the whitespace after the colon into the score token (`" 0"`) made every candidate fail the decimal check and raised `MetricComputationError` on a perfectly parseable response.
- The score digits are now located by reconstructing the token stream and walking the first JSON object once, stepping over whole strings and reading the `"score"` key only at depth 1, stopping when that object closes. Last assignment wins, so duplicate top-level keys resolve the way `json.loads` does, and a nested `"score"` cannot capture the position because it sits deeper. Candidate tokens are stripped before the decimal check; a two-token span combines both positions' candidates (`"1" + "0"` -> 10).
- When the located span carries no probability mass (digits span punctuation-carrying tokens), the parser degrades to the text path like the short-stream and no-logprob branches instead of raising.
- Falls back to the legacy fixed offset when the first object holds no top-level key.

## Change checklist
- [ ] User facing
- [ ] Documentation update

## Issues

Related: #8134 (defect 3 of 4 — `Resolves` intentionally avoided so the four-defect issue stays open for defects 1 and 2)

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: OpenCode (Sisyphus agent session)
- Model(s): muse-spark
- Scope: review-feedback commits only — `5486cc3` (typehints, `_to_dict` wrap, no-mass degrade, content-free log line, test dedup/parametrize, punctuation-split regression test), `3b8cfe9` (DEBUG fallback log + message assertion), `0e48cce` (locator value-agreement fix, two regression tests, two docstring corrections), `ae048eadf` (locator brace-depth fix + one regression test), `670fdd61e` (one test-only commit closing a coverage hole the depth rule opened), `3b056b5d9` (locator rewritten to a single-pass structural walk, design and reference implementation from `petrotiurin` on the thread, plus three regression tests)

- Human verification: `pytest tests/unit/evaluation/metrics/llm_judges/g_eval/` (18 passed at `3b056b5d9`, 15 at `670fdd61e`, 13 at `0e48cce`), `pytest tests/unit/evaluation/metrics/llm_judges/` 126 passed, new tests verified to fail on the pre-fix parser, `ruff@0.14.14` check + format clean, `mypy@1.19.1` clean; mutation table below re-run whole at each head, each mutation restoring the file byte-for-byte

## Testing

At `3b056b5d9`:

- `PYTHONPATH=src python -m pytest tests/unit/evaluation/metrics/llm_judges/g_eval/ -q` → `18 passed`
- `PYTHONPATH=src python -m pytest tests/unit/evaluation/metrics/llm_judges/ -q` → `126 passed`
- The three tests added in `3b056b5d9` fail against `670fdd61e`'s `src/` and only those three do; the fifteen existing tests pass unmodified on the new locator
- `uvx ruff@0.14.14 check` + `format --check` on both touched files → clean; `uvx mypy@1.19.1 --config-file pyproject.toml --disable-error-code misc --ignore-missing-imports` on `parser.py` → `Success: no issues found in 1 source file`
- `PYTHONPATH=src python -m pytest tests/unit -q --tb=no -rf` → `1 failed, 5161 passed, 3 skipped in 243.78s`. The failure is `tests/unit/evaluation/metrics/test_heuristics.py::test_evaluation__levenshtein_ratio`, in a file this diff does not touch; `importlib.util.find_spec("rapidfuzz")` returns `None` here, so an optional dependency is missing rather than anything in this change being wrong
- E2E not run (needs a live backend stack)

Historical, at earlier heads: `test_punctuation_carrying_split_digits_degrade_to_text_path` fails on the pre-PR parser with `MetricComputationError` and passes post-fix with `1.0`. A broader `tests/unit/evaluation/` sweep at that time showed the same failure count on the clean tree (91 failed / 927 passed) and on this branch (91 failed / 930 passed); those 91 are pre-existing environment failures unrelated to this diff.

## Documentation

N/A — no user-facing API or config change.

---

**Follow-up round (3b8cfe9).** On Baz's question, the zero-probability-mass fallback now logs at DEBUG with the metric name, matching the sibling locator-failure branch, and `test_punctuation_carrying_split_digits_degrade_to_text_path` asserts the message as well as the value. Measured against its immediate predecessor 5486cc358, which degrades silently, the assertion fails on an empty capture; against the pre-PR parser the test fails outright with `MetricComputationError`. Post-fix: 11 passed in `tests/unit/evaluation/metrics/llm_judges/g_eval/`, `ruff check` and `ruff format --check` clean, `mypy` clean on `parser.py`.

---

**Follow-up round (`0e48cce`, addressing the nested-key concern).** Reproduced first, then applied the suggested agreement check. Measured on `{"score":7, "reason": "ok", "breakdown": {"score":3}}`, with the top-level digit at entry 3 and the nested one last in text order:

| tree | located entry | score |
|---|---|---|
| merge base `d53961c4` | fixed offset 3 | 0.6219349153822014 |
| `3b8cfe9` (before this round) | 13, the nested `3` | 0.27397830512740046 |
| `0e48cce` | 3 | 0.6219349153822014 |
| text path (no logprobs) | — | 0.7 |

Two tests added:

- `test_nested_score_key_does_not_capture_the_located_token` — fails on `3b8cfe9` with `AssertionError: scored from the nested key: 0.27397830512740046`, passes here.
- `test_unparseable_reconstruction_keeps_the_positional_match` — pins the fallback when the reconstructed stream is not valid JSON.

Mutations, each restoring `parser.py` byte-for-byte (sha256 checked) and re-running the suite:

| mutation | tests that die |
|---|---|
| drop the agreement check (positional last match, i.e. `3b8cfe9`) | nested-key test, 1 |
| positional *first* match (`matches[0]`) | duplicate-key test, 1 — **retracted in the next round: re-run at `0e48cce` it kills nothing** |
| agreement check without the `except` | truncated-stream test + 5 existing tests, 6 |

The third row is why the fallback stays: removing it fails 6 of the 13 tests, not just the new one, because the reconstructed text is not valid JSON in several existing fixtures (their token streams stop before the document closes). In those cases the positional choice is kept, which is today's behaviour.

Two notes on the suggestion. I did not hoist the `json.loads` into the one further down: that one parses `choice_dict`'s message content, this one parses the reconstructed token text, and those can differ (the truncated-stream case), so the parses are not interchangeable. And the two docstring claims that said last-match-wins mirrors `json.loads` were true for top-level duplicates only, so both now state the value-agreement rule.

Verification: 13 passed in `tests/unit/evaluation/metrics/llm_judges/g_eval/`; `uvx ruff@0.14.14 check` and `format --check` clean on both files; `uvx mypy@1.19.1 --config-file sdks/python/pyproject.toml --disable-error-code misc` clean on `parser.py` (run with `--ignore-missing-imports` locally, since this environment has no `litellm` stubs).

---

**Follow-up round (`ae048eadf` + `670fdd61e`, on the three remaining locator claims).**
Each claim was measured as one response per claim across three trees — `origin/main`, this
PR at `0e48cce`, and after the change at `670fdd61e` — instead of being reasoned about:

| response | `json.loads` | `main` | `0e48cce` | `670fdd61e` | text path |
|---|---|---|---|---|---|
| `{"score":7, "reason": "ok", "breakdown": {"score":7}}` | 7 | 0.6219349153822014 | **0.7809998433984686** | 0.6219349153822014 | 0.7 |
| `{"\u0073core": 7, "reason": "ok"}` | 7 | `MetricComputationError` | 0.7 | 0.7 | 0.7 |
| `{"score": 1e1, "reason": "ok"}` | 10.0 | 0.20408677949039836 | 0.20408677949039836 | 0.20408677949039836 | 1.0 |

**Row 1 is real and is fixed here.** Value agreement does not identify a position: when a
breakdown restates the top-level score, both digits agree, so the reversed scan kept
returning the nested one and the weighted average came off the breakdown token's
distribution. `main` scores 0.6219349153822014 on that response, so this was a regression
the previous round introduced — and a worse shape than the differing-value case, because it
yields a plausible number rather than an obviously wrong one. The rule is now: take the
agreeing matches, restrict them to the shallowest brace depth, then take the last of those,
which keeps duplicate top-level keys resolving to the final one.

`test_nested_score_key_with_the_same_value_does_not_win_the_locator` fails pre-fix with
`assert [13] == [3]` and passes post-fix. Selection differential over 32 reconstructed texts
(top level only; breakdown with the same and with a different value; two criteria; criteria
inside an array; braces inside a string; duplicate top-level keys; doubly nested) changes
the chosen span in 5 of them, and all 5 are the equal-valued-nested family — the other 27
select the same span as before.

**Rows 2 and 3 are not changed, and the numbers say why.** The escaped key returns `None`
from the locator, so the caller falls back to the fixed offset and lands on 0.7, the same
value `json.loads` gives, and better than `main`, which raises `MetricComputationError` on
that response. This PR therefore does not worsen the escaped-key case; it leaves it on a
path that already resolves it. The exponent row is identical on all three trees, so it is
inherited rather than introduced; consuming the full number grammar would also extend the
digit span onto `e1`, which `_weighted_score_sums` cannot weight because it filters
candidates on `isdecimal()`. Fixing that means changing the weighting code, and it is only
reachable if the integer contract in `_COT_SYSTEM_PROMPT` ("SCORE VALUE MUST BE AN INTEGER")
is relaxed. Left out rather than widening this diff.

**Corrections to my own previous round, found while measuring this one.**

- The `matches[0]` row in the table above is retracted. Re-run at `0e48cce` it kills nothing
  (13 passed): the value-agreement step overrides the positional default whenever any match
  agrees, so first-vs-last only survives on an unparseable reconstruction, and the test for
  that path used a single score key. The row was true of an earlier head and I carried it
  forward without re-measuring it.
- `670fdd61e` closes the coverage hole that explanation describes: two score keys in a
  truncated stream, where last-wins is the only rule left standing. It passes on `0e48cce`
  too, so it demonstrates missing coverage rather than a fixed defect.

Mutation table at `670fdd61e`, 15 tests as the baseline. Each mutation was applied to
`parser.py`, the suite run, then the file restored byte-for-byte (sha256 `e81072a313fb`,
checked before and after). **This table is superseded by the one at `3b056b5d9` below: the
code it mutates no longer exists.**

| mutation | tests that die |
|---|---|
| drop the depth rule (`match = agreed[-1][0]`) | equal-valued nested test, 1 |
| drop value agreement *and* depth (bare `matches[-1]`, i.e. `3b8cfe9`) | both nested tests, 2 |
| positional *first* match (`matches[0]`) | unparseable duplicate-key test, 1 |
| remove the unparseable fallback (raise instead of keeping positional) | the 3 locator-fallback tests + 4 provider-path e2e cases, 7 |

The last row is why the fallback stays load-bearing: it takes the punctuation-split
degradation test and both unparseable-reconstruction tests, plus all four provider-path e2e
cases (sync and async, two parameterizations each).

Verification at `670fdd61e`: `PYTHONPATH=src pytest
tests/unit/evaluation/metrics/llm_judges/g_eval/` → 15 passed; `uvx ruff@0.14.14 check` and
`format --check` clean on both files; `uvx mypy@1.19.1 --config-file pyproject.toml
--ignore-missing-imports` clean on `parser.py` (the `--ignore-missing-imports` is local
only, this machine has no `litellm` stubs; CI runs without it, and `tests/` is excluded from
mypy by `.pre-commit-config.yaml`). Source diff is 23 lines added / 5 removed in
`parser.py`; the rest is tests.

---

**Follow-up round (`3b056b5d9`, the depth rule replaced by a structural walk).**

`petrotiurin` reproduced two defects in the rule the previous round added, and supplied the
design and a reference implementation. Both reproduce here, measured before changing
anything.

The depth rule counted `{` and `}` characters in the text preceding each match, so braces
inside a JSON string value read as structure. With `}}}` in a reason the top-level key came
out at depth 1 and the nested one at depth 0, so "shallowest agreeing match" selected the
nested key. The same rule copied a growing prefix and ran two full `.count()` scans per
agreeing match, which is quadratic in the response length.

| response | `main` | `670fdd61e` | `3b056b5d9` |
|---|---|---|---|
| `{"score":7, "reason": "returns }}} here", "breakdown": {"score": 7}}` | 0.6219349153822014 | entry 13, **0.7809998433984686** | entry 3, 0.6219349153822014 |
| `{"reason": "ok", "breakdown": {"score": 7}}`, the only score nested | `MetricComputationError` | entry 9, 0.6219349153822014 | `None`, `MetricComputationError` |

Timing on a rubric-shaped response that doubles each step (top-level score plus N criteria,
each `{"name": ..., "score": 7}`, four-character tokens):

| bytes | criteria | `670fdd61e` | `3b056b5d9` |
|---|---|---|---|
| 4119 | 143 | 0.443 ms | 0.307 ms |
| 8208 | 284 | 1.253 ms | 0.659 ms |
| 16386 | 566 | 4.229 ms | 1.297 ms |
| 32783 | 1127 | 15.244 ms | 2.721 ms |
| 65543 | 2219 | 58.116 ms | 5.461 ms |

The walk is one pass: `_STRUCTURE_RE` matches either a whole JSON string, escapes included
and an unterminated tail tolerated, or a single brace. Strings are consumed whole, so
neither a brace nor a `"score"` echo inside a reason can move the structure. The key is read
only at depth 1 and the walk stops when that object closes. Last assignment wins, so
duplicate top-level keys still resolve as `json.loads` does. Nothing is validated, so a
truncated reconstruction behaves exactly as before, which lets the `json.loads` agreement
check and its bare `except` go away. `json` stays imported: `parser.py:111` still uses it.

Three tests added. The fifteen existing ones pass unmodified, and against `670fdd61e`'s
`src/` the file gives three failures which are exactly the three new tests.

- `test_braces_inside_a_string_do_not_shift_the_score_depth` — fails pre-fix with
  `assert [13] == [3]`.
- `test_score_only_nested_falls_back_instead_of_scoring_from_it` — pins the decision on the
  consequence below; fails pre-fix, which scored 0.6219349153822014 from the nested key.
- `test_walk_stops_at_the_first_object_close` — `{"score": 7} and {"score": 3}` located
  entry 9 pre-fix, entry 4 now.

**The consequence flagged on the thread is taken, not absorbed.** A response whose only
`"score"` is nested now returns `None`, falls through to the legacy offset and raises,
matching `main` and the text path. Otherwise the same response succeeds or fails purely on
whether the provider returned logprobs.

**One shape where the walk is stricter than the rule it replaces.** Prose carrying a complete
brace pair before the JSON, `Use {this} format: {"score": 7}`: the walk closes on the prose
brace and returns `None`, where the old rule reached the real key. On that input all three
trees still end in `MetricComputationError`, because `parser.py:111` parses the message
content with a strict `json.loads` and rejects anything prose-wrapped before the located
score matters. So the narrowing changes no outcome today, and where it eventually shows up it
degrades to the legacy offset, which is `main`'s behaviour. Raised on the thread rather than
silently widened; continuing past a close while `span` is still `None` is a one-line change
and is deliberately not made here.

**A distinction Baz's report and the thread did not separate.** Baz's literal example,
`{"reason":"{", "score":7}`, does not reproduce: it has one match, so the shallowest-of-
agreeing step has nothing to choose between and it locates correctly. `main` is what raises
there, because the fixed offset lands on the `{` token and the zero-mass degrade did not
exist before this PR. Only the `}}}` variant breaks the depth rule, since the comparison
needs a second match to pick wrongly.

Mutation table at `3b056b5d9`, 18 tests as the baseline, each mutation restored
byte-for-byte with sha256 `765c826e` checked before and after:

| mutation | tests that die |
|---|---|
| read the key at `depth >= 1` instead of `== 1` | 4: braces-in-string, both nested-key tests, nested-only |
| do not stop when the first object closes | 1: `test_walk_stops_at_the_first_object_close` |
| keep the first span instead of the last | 2: duplicate-key, unparseable duplicate-key |
| match braces only, do not consume whole strings | 9 |
| raise instead of returning `None` when there is no top-level key | 1: nested-only |

No test pins the linear scan. A timing bound loose enough to survive a loaded runner would
not kill the quadratic version either, since 58 ms at 64KB passes anything sane, so that
measurement lives in this description rather than in the suite.

**Follow-up, not touched here.** `parser.py:111` reads the reason with a strict `json.loads`
on the raw message content, while every other read goes through
`parsing_helpers.extract_json_content_or_raise` (`:21`). That is why a prose-wrapped or
markdown-fenced response fails on the logprob path and succeeds without logprobs. Verified,
and out of scope for this diff.

Verification at `3b056b5d9`: 18 passed in
`tests/unit/evaluation/metrics/llm_judges/g_eval/`, 126 passed in
`tests/unit/evaluation/metrics/llm_judges/`; `uvx ruff@0.14.14 check` and `format --check`
clean on both touched files; `uvx mypy@1.19.1 --config-file pyproject.toml
--disable-error-code misc --ignore-missing-imports` on `parser.py` → `Success: no issues
found in 1 source file`. The commit is 36 lines added / 36 removed in `parser.py` plus 108
lines of tests. Full `tests/unit` at this head: 1 failed, 5161 passed, 3 skipped in 243.78s,
the failure being `test_evaluation__levenshtein_ratio` on a missing optional `rapidfuzz` in a
file outside the diff.

### DIFF
--- a/sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
+++ b/sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
@@ -1,6 +1,7 @@
 import logging
 import json
 import math
+import re
 from typing import Any, Dict, TYPE_CHECKING
 import opik.exceptions as exceptions
 from opik.evaluation.metrics import score_result
@@ -60,34 +61,21 @@ def parse_litellm_model_output(
 
         log_probs = _to_dict(choice_dict.get("logprobs"))
         entries = log_probs.get("content") or []
-        score_token_position = 3
-        if len(entries) <= score_token_position:
+        if len(entries) <= 3:
             return _extract_score_from_text_content(choice_dict, name=name)
 
-        entry_dict = _to_dict(entries[score_token_position])
-        top_logprobs = entry_dict.get("top_logprobs") or []
-        token_candidate = str(entry_dict.get("token", ""))
+        # Locate the score token(s) by content instead of assuming a fixed
+        # offset: tokenizers differ on where the whitespace after `"score":`
+        # lands, and a two-digit score ("10") spans two tokens.
+        entry_indices = _locate_score_entries(entries)
+        if entry_indices is None:
+            entry_indices = [3]
 
-        linear_probs_sum = 0.0
-        weighted_score_sum = 0.0
-
-        for candidate in top_logprobs:
-            token_info = _to_dict(candidate)
-            token_str = str(token_info.get("token", ""))
-            if not token_str.isdecimal():
-                continue
-
-            score = int(token_str)
-            if not 0 <= score <= 10:
-                continue
-
-            log_prob = token_info.get("logprob")
-            if log_prob is None:
-                continue
-
-            linear_prob = math.exp(float(log_prob))
-            linear_probs_sum += linear_prob
-            weighted_score_sum += linear_prob * score
+        (
+            linear_probs_sum,
+            weighted_score_sum,
+            token_candidate,
+        ) = _weighted_score_sums(entries, entry_indices)
 
         if linear_probs_sum != 0.0:
             final_score: float = weighted_score_sum / linear_probs_sum / 10
@@ -107,6 +95,93 @@ def parse_litellm_model_output(
     except Exception as exception:
         LOGGER.error(f"Failed to parse model output: {exception}", exc_info=True)
         raise exceptions.MetricComputationError(GEVAL_SCORE_CALC_FAILED) from exception
+
+
+_SCORE_KEY_RE = re.compile(r'"score"\s*:\s*(\d+)')
+
+
+def _locate_score_entries(entries: list) -> list[int] | None:
+    """Find the entry indices whose tokens carry the score digits.
+
+    Reconstructs the decoded text from the token stream and locates the
+    digits after `"score":`; returns the one or two indices covering them,
+    or None when the key cannot be found in the reconstructed text (the
+    caller then falls back to the legacy fixed offset).
+    """
+    token_texts = [str(_to_dict(entry).get("token", "")) for entry in entries]
+    full_text = "".join(token_texts)
+    match = _SCORE_KEY_RE.search(full_text)
+    if match is None:
+        return None
+    start, end = match.start(1), match.end(1)
+    offsets = []
+    position = 0
+    for text in token_texts:
+        offsets.append((position, position + len(text)))
+        position += len(text)
+    indices = [
+        index
+        for index, (begin, stop) in enumerate(offsets)
+        if stop > start and begin < end
+    ]
+    if not 1 <= len(indices) <= 2:
+        return None
+    return indices
+
+
+def _decimal_candidates(entry) -> list:
+    return [
+        (str(info.get("token", "")).strip(), math.exp(float(info["logprob"])))
+        for info in (_to_dict(cand) for cand in (entry.get("top_logprobs") or []))
+        if str(info.get("token", "")).strip().isdecimal()
+        and info.get("logprob") is not None
+    ]
+
+
+def _weighted_score_sums(entries: list, entry_indices: list[int]) -> tuple:
+    """Weighted [0, 10] score mass over the candidate space of the score token(s).
+
+    A single-entry span averages that entry's decimal candidates (the
+    legacy behaviour, plus leading/trailing whitespace tolerance). A
+    two-entry span additionally combines the two positions' candidates
+    ("1" + "0" -> 10) and counts a first-position candidate only when it
+    covers the whole digit span ("10"), so the chosen digits are not
+    double-counted.
+    """
+    token_candidate = "".join(
+        str(_to_dict(entries[index]).get("token", "")) for index in entry_indices
+    ).strip()
+
+    linear_probs_sum = 0.0
+    weighted_score_sum = 0.0
+
+    if len(entry_indices) == 1:
+        for token, prob in _decimal_candidates(entries[entry_indices[0]]):
+            score = int(token)
+            if not 0 <= score <= 10:
+                continue
+            linear_probs_sum += prob
+            weighted_score_sum += prob * score
+        return linear_probs_sum, weighted_score_sum, token_candidate
+
+    digits = token_candidate
+    first_candidates = _decimal_candidates(entries[entry_indices[0]])
+    second_candidates = _decimal_candidates(entries[entry_indices[1]])
+
+    for token_a, prob_a in first_candidates:
+        if token_a == digits and 0 <= int(token_a) <= 10:
+            # one token covering the whole span (alternative tokenization)
+            linear_probs_sum += prob_a
+            weighted_score_sum += prob_a * int(token_a)
+        for token_b, prob_b in second_candidates:
+            combined = token_a + token_b
+            if not combined.isdecimal() or not 0 <= int(combined) <= 10:
+                continue
+            prob = prob_a * prob_b
+            linear_probs_sum += prob
+            weighted_score_sum += prob * int(combined)
+
+    return linear_probs_sum, weighted_score_sum, token_candidate
 
 
 def _extract_score_from_text_content(
--- a/sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
+++ b/sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
@@ -107,12 +107,17 @@ def _locate_score_entries(entries: list) -> list[int] | None:
     digits after `"score":`; returns the one or two indices covering them,
     or None when the key cannot be found in the reconstructed text (the
     caller then falls back to the legacy fixed offset).
+
+    Uses the LAST match so duplicate `"score"` keys resolve the same way
+    json.loads does — to the final one, which is also what the no-logprob
+    text path reads via `dict_content["score"]`.
     """
     token_texts = [str(_to_dict(entry).get("token", "")) for entry in entries]
     full_text = "".join(token_texts)
-    match = _SCORE_KEY_RE.search(full_text)
-    if match is None:
+    matches = list(_SCORE_KEY_RE.finditer(full_text))
+    if not matches:
         return None
+    match = matches[-1]
     start, end = match.start(1), match.end(1)
     offsets = []
     position = 0
--- a/sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
+++ b/sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
@@ -69,6 +69,12 @@ def parse_litellm_model_output(
         # lands, and a two-digit score ("10") spans two tokens.
         entry_indices = _locate_score_entries(entries)
         if entry_indices is None:
+            LOGGER.debug(
+                "g_eval score key not found in the reconstructed response; "
+                "falling back to the legacy fixed token offset. Reconstructed "
+                "response: %r",
+                "".join(str(_to_dict(entry).get("token", "")) for entry in entries),
+            )
             entry_indices = [3]
 
         (
--- a/sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
+++ b/sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
@@ -50,8 +50,9 @@ def parse_litellm_model_output(
     between 0 and 10.
 
     In order to make the score computation more robust, we look at the top logprobs of the score token and compute
-    a weighted average of the scores. Since we try to enforce the format of the model's response, we can assume that
-    the score token is always the fourth token in the response (first token is `{"`, followed by `score` and `":`).
+    a weighted average of the scores. The score token is located by content (the digits after the `"score":` key,
+    last-wins on duplicates, mirroring json.loads); only when the key cannot be located do we fall back to the
+    legacy fixed token offset for backwards compatibility.
     """
     try:
         choice_dict = _normalise_first_choice(content)
--- a/sdks/python/tests/unit/evaluation/metrics/llm_judges/g_eval/test_parser.py
+++ b/sdks/python/tests/unit/evaluation/metrics/llm_judges/g_eval/test_parser.py
@@ -144,3 +144,67 @@ def test_single_digit_score_at_position_three_unchanged():
         log_probs_supported=True,
     )
     assert 0.60 < result.value < 0.65
+
+
+def test_duplicate_score_keys_last_wins_like_the_text_path():
+    # json.loads (and therefore the no-logprob text path) resolves
+    # duplicate "score" keys to the LAST one. The token-stream locator
+    # must mirror that: a first-occurrence scan scores from the first,
+    # stale key instead.
+    entries = [
+        _entry('{\"', -0.01),
+        _entry("score", -0.01),
+        _entry('\":', -0.01),
+        _entry(" ", -0.01),
+        _entry("9", -0.5, top=[{"token": "9", "logprob": -0.01}, {"token": "5", "logprob": -3.00}]),
+        _entry(",", -0.01),
+        _entry(" ", -0.01),
+        _entry("\"score", -0.01),
+        _entry('\":', -0.01),
+        _entry(" ", -0.01),
+        _entry("7", -0.1, top=[{"token": "7", "logprob": -0.1}, {"token": "1", "logprob": -2.00}]),
+        _entry(",", -0.01),
+        _entry(" ", -0.01),
+        _entry("\"", -0.01),
+        _entry("reason", -0.01),
+        _entry('\":', -0.01),
+        _entry(" ", -0.01),
+        _entry("\"ok\"", -0.01),
+        _entry("}", -0.01),
+    ]
+    content = '{"score": 9, "score": 7, "reason": "ok"}'
+    result = parser.parse_litellm_model_output(
+        _response(content, entries),
+        name="g_eval",
+        log_probs_supported=True,
+    )
+    assert 0.55 < result.value < 0.68, f"scored from the stale key: {result.value}"
+
+
+def test_escaped_quote_echo_is_not_a_matchable_key():
+    # Control: a rubric quote inside the reason has escaped quotes, so
+    # the literal "score": n there can never match the key regex -- the
+    # locator stays on the real key. Guards against a looser rewrite
+    # that would scan bare text.
+    entries = [
+        _entry('{\"', -0.01),
+        _entry("reason", -0.01),
+        _entry('\":', -0.01),
+        _entry(' "the rubric said \\"', -0.01),
+        _entry("score", -0.01),
+        _entry('\\": ', -0.01),
+        _entry("9", -0.5, top=[{"token": "9", "logprob": -0.01}, {"token": "5", "logprob": -3.00}]),
+        _entry(" here\", ", -0.01),
+        _entry("\"score", -0.01),
+        _entry('\":', -0.01),
+        _entry(" ", -0.01),
+        _entry("7", -0.1, top=[{"token": "7", "logprob": -0.1}, {"token": "1", "logprob": -2.00}]),
+        _entry("}", -0.01),
+    ]
+    content = '{"reason": "the rubric said \\"score\\": 9 here", "score": 7}'
+    result = parser.parse_litellm_model_output(
+        _response(content, entries),
+        name="g_eval",
+        log_probs_supported=True,
+    )
+    assert 0.55 < result.value < 0.68, f"unexpected value {result.value}"
--- a/sdks/python/tests/unit/evaluation/metrics/llm_judges/g_eval/test_parser.py
+++ b/sdks/python/tests/unit/evaluation/metrics/llm_judges/g_eval/test_parser.py
@@ -16,3 +16,131 @@ def test_g_eval__parse_model_output_string__score_out_of_range__MetricComputatio
             content=invalid_model_output,
             metric_name="g_eval",
         )
+
+
+# --- deterministic logprob stubs -------------------------------------------
+# Shape mirrors what parse_litellm_model_output normalises via _to_dict.
+
+
+def _entry(token, logprob, top=None):
+    return {
+        "token": token,
+        "logprob": logprob,
+        "top_logprobs": top
+        if top is not None
+        else [{"token": token, "logprob": logprob}],
+    }
+
+
+def _response(content, entries):
+    from types import SimpleNamespace
+
+    return SimpleNamespace(
+        choices=[
+            SimpleNamespace(
+                message={"content": content},
+                logprobs={"content": entries},
+            )
+        ]
+    )
+
+
+def test_two_digit_score_split_across_tokens_scores_near_one():
+    # {"score":10} tokenizes as {" score ": "1" "0" ...}; the score is 10 but
+    # position 3 only sees the first digit "1", so the weighted average of
+    # digit candidates lands near 0.1 instead of 1.0.
+    entries = [
+        _entry('{"', -0.01),
+        _entry("score", -0.01),
+        _entry('":', -0.01),
+        _entry(
+            "1",
+            -0.05,
+            top=[
+                {"token": "1", "logprob": -0.05},
+                {"token": "0", "logprob": -2.30},
+                {"token": "2", "logprob": -2.40},
+            ],
+        ),
+        _entry(
+            "0",
+            -0.02,
+            top=[{"token": "0", "logprob": -0.02}, {"token": "1", "logprob": -3.00}],
+        ),
+        _entry(",", -0.01),
+        _entry(" ", -0.01),
+        _entry('"', -0.01),
+        _entry("reason", -0.01),
+        _entry('":', -0.01),
+        _entry(" ", -0.01),
+        _entry('"excellent"', -0.01),
+        _entry("}", -0.01),
+    ]
+    result = parser.parse_litellm_model_output(
+        _response('{"score":10, "reason": "excellent"}', entries),
+        name="g_eval",
+        log_probs_supported=True,
+    )
+    assert result.value > 0.9, f"score 10 parsed as {result.value}"
+
+
+def test_leading_space_score_token_is_scored_not_rejected():
+    # {"score": 0} tokenizes the space into the score token (" 0"); the
+    # candidate filter rejects every candidate (none isdecimal) and the
+    # chosen-token fallback raises on " 0" as well.
+    entries = [
+        _entry('{"', -0.01),
+        _entry("score", -0.01),
+        _entry('":', -0.01),
+        _entry(
+            " 0",
+            -0.02,
+            top=[
+                {"token": " 0", "logprob": -0.02},
+                {"token": " 1", "logprob": -2.00},
+                {"token": " 10", "logprob": -2.50},
+            ],
+        ),
+        _entry(",", -0.01),
+        _entry(" ", -0.01),
+        _entry('"', -0.01),
+        _entry("reason", -0.01),
+        _entry('":', -0.01),
+        _entry(" ", -0.01),
+        _entry('"none"', -0.01),
+        _entry("}", -0.01),
+    ]
+    result = parser.parse_litellm_model_output(
+        _response('{"score": 0, "reason": "none"}', entries),
+        name="g_eval",
+        log_probs_supported=True,
+    )
+    assert 0.0 < result.value < 0.09, f"unexpected value {result.value}"
+
+
+def test_single_digit_score_at_position_three_unchanged():
+    # Control: the no-space, single-digit case must keep today's exact value.
+    entries = [
+        _entry('{"', -0.01),
+        _entry("score", -0.01),
+        _entry('":', -0.01),
+        _entry(
+            "7",
+            -0.1,
+            top=[{"token": "7", "logprob": -0.1}, {"token": "1", "logprob": -2.0}],
+        ),
+        _entry(",", -0.01),
+        _entry(" ", -0.01),
+        _entry('"', -0.01),
+        _entry("reason", -0.01),
+        _entry('":', -0.01),
+        _entry(" ", -0.01),
+        _entry('"ok"', -0.01),
+        _entry("}", -0.01),
+    ]
+    result = parser.parse_litellm_model_output(
+        _response('{"score":7, "reason": "ok"}', entries),
+        name="g_eval",
+        log_probs_supported=True,
+    )
+    assert 0.60 < result.value < 0.65
--- a/sdks/python/tests/unit/evaluation/metrics/llm_judges/g_eval/test_parser.py
+++ b/sdks/python/tests/unit/evaluation/metrics/llm_judges/g_eval/test_parser.py
@@ -1,5 +1,10 @@
+from contextlib import contextmanager
+from types import SimpleNamespace
+
 from opik import logging_messages, exceptions
 from opik.evaluation.metrics.llm_judges.g_eval import parser
+from opik.evaluation.metrics.llm_judges.g_eval.metric import GEval
+from opik.evaluation.models import base_model
 import pytest
 
 
@@ -178,7 +183,12 @@ def test_duplicate_score_keys_last_wins_like_the_text_path():
         name="g_eval",
         log_probs_supported=True,
     )
-    assert 0.55 < result.value < 0.68, f"scored from the stale key: {result.value}"
+    # Exact weighted value of the second key's candidates (7 @ -0.1,
+    # 1 @ -2.0): (7*e^-0.1 + e^-2)/(e^-0.1 + e^-2)/10. A first-match locator
+    # scores the stale "9" key instead (0.88084...) and fails here.
+    assert result.value == pytest.approx(0.6219349153822014, abs=1e-9), (
+        f"scored from the stale key: {result.value}"
+    )
 
 
 def test_escaped_quote_echo_is_not_a_matchable_key():
@@ -207,4 +217,127 @@ def test_escaped_quote_echo_is_not_a_matchable_key():
         name="g_eval",
         log_probs_supported=True,
     )
-    assert 0.55 < result.value < 0.68, f"unexpected value {result.value}"
+    # Same distribution as the duplicate-key test pins the locator on the
+    # real key; see the derivation there.
+    assert result.value == pytest.approx(0.6219349153822014, abs=1e-9), (
+        f"unexpected value {result.value}"
+    )
+
+
+# --- provider-shaped GEval coverage ----------------------------------------
+# The tests above call parse_litellm_model_output directly with hand-built
+# token dicts. These go through the public GEval.score LiteLLM branch with
+# a provider-shaped raw response, so an incorrectly bounded result on the
+# real path cannot pass unnoticed.
+
+
+def _provider_response(content, entries):
+    """Provider-shaped raw LiteLLM response: a dict-style choice carrying the
+    message content plus the logprobs token stream, as yielded by
+    get_provider_response on the LiteLLM branch."""
+    return SimpleNamespace(
+        choices=[{"message": {"content": content}, "logprobs": {"content": entries}}]
+    )
+
+
+def _geval_on_stubbed_provider(monkeypatch, content, entries):
+    """GEval wired to a stubbed LiteLLM provider: chain-of-thought generation
+    is stubbed, and the scoring call yields the given provider-shaped
+    response. Returns (metric, captured_provider_kwargs)."""
+    metric = GEval(
+        task_introduction="provider-path split-score intro",
+        evaluation_criteria="provider-path split-score criteria",
+        model="gpt-4o",
+        track=False,
+    )
+    # The subject here is the logprob parser branch, not model capability
+    # detection (covered by test_geval_passes_logprobs_only_when_supported).
+    metric._log_probs_supported = True
+    monkeypatch.setattr(
+        metric._model,
+        "generate_chat_completion",
+        lambda *args, **kwargs: {"content": "stub chain of thought"},
+    )
+    captured = {}
+
+    @contextmanager
+    def fake_get_provider_response(model_provider, messages, **kwargs):
+        captured.update(kwargs)
+        yield _provider_response(content, entries)
+
+    monkeypatch.setattr(base_model, "get_provider_response", fake_get_provider_response)
+    return metric, captured
+
+
+def test_geval_litellm_path_split_score_scores_near_one(monkeypatch):
+    # {"score":10} split across "1" + "0" must score near 1.0 end to end. A
+    # legacy fixed-offset read of entries[3] alone averages to ~0.099.
+    entries = [
+        _entry('{"', -0.01),
+        _entry("score", -0.01),
+        _entry('":', -0.01),
+        _entry(
+            "1",
+            -0.05,
+            top=[
+                {"token": "1", "logprob": -0.05},
+                {"token": "0", "logprob": -2.30},
+                {"token": "2", "logprob": -2.40},
+            ],
+        ),
+        _entry(
+            "0",
+            -0.02,
+            top=[{"token": "0", "logprob": -0.02}, {"token": "1", "logprob": -3.00}],
+        ),
+        _entry(",", -0.01),
+        _entry('"reason"', -0.01),
+        _entry('":', -0.01),
+        _entry(" ", -0.01),
+        _entry('"excellent"', -0.01),
+        _entry("}", -0.01),
+    ]
+    content = '{"score":10, "reason": "excellent"}'
+    metric, captured = _geval_on_stubbed_provider(monkeypatch, content, entries)
+
+    result = metric.score("any input")
+
+    assert captured["logprobs"] is True
+    assert captured["top_logprobs"] == 20
+    assert result.value == pytest.approx(0.9007723389857002, abs=1e-9)
+    assert result.reason == "excellent"
+
+
+def test_geval_litellm_path_folded_whitespace_is_scored(monkeypatch):
+    # The space after the colon folds into the score token (" 0"); the
+    # end-to-end path must score it instead of raising
+    # MetricComputationError on a perfectly parseable response.
+    entries = [
+        _entry('{"', -0.01),
+        _entry("score", -0.01),
+        _entry('":', -0.01),
+        _entry(
+            " 0",
+            -0.02,
+            top=[
+                {"token": " 0", "logprob": -0.02},
+                {"token": " 1", "logprob": -2.00},
+                {"token": " 10", "logprob": -2.50},
+            ],
+        ),
+        _entry(",", -0.01),
+        _entry('"reason"', -0.01),
+        _entry('":', -0.01),
+        _entry(" ", -0.01),
+        _entry('"none"', -0.01),
+        _entry("}", -0.01),
+    ]
+    content = '{"score": 0, "reason": "none"}'
+    metric, captured = _geval_on_stubbed_provider(monkeypatch, content, entries)
+
+    result = metric.score("any input")
+
+    assert captured["logprobs"] is True
+    assert captured["top_logprobs"] == 20
+    assert result.value == pytest.approx(0.07984052568223204, abs=1e-9)
+    assert result.reason == "none"
--- a/sdks/python/tests/unit/evaluation/metrics/llm_judges/g_eval/test_parser.py
+++ b/sdks/python/tests/unit/evaluation/metrics/llm_judges/g_eval/test_parser.py
@@ -1,4 +1,4 @@
-from contextlib import contextmanager
+from contextlib import asynccontextmanager, contextmanager
 from types import SimpleNamespace
 
 from opik import logging_messages, exceptions
@@ -258,6 +258,15 @@ def _geval_on_stubbed_provider(monkeypatch, content, entries):
         "generate_chat_completion",
         lambda *args, **kwargs: {"content": "stub chain of thought"},
     )
+
+    async def fake_agenerate_chat_completion(*args, **kwargs):
+        return {"content": "stub chain of thought"}
+
+    monkeypatch.setattr(
+        metric._model,
+        "agenerate_chat_completion",
+        fake_agenerate_chat_completion,
+    )
     captured = {}
 
     @contextmanager
@@ -266,6 +275,15 @@ def _geval_on_stubbed_provider(monkeypatch, content, entries):
         yield _provider_response(content, entries)
 
     monkeypatch.setattr(base_model, "get_provider_response", fake_get_provider_response)
+
+    @asynccontextmanager
+    async def fake_aget_provider_response(model_provider, messages, **kwargs):
+        captured.update(kwargs)
+        yield _provider_response(content, entries)
+
+    monkeypatch.setattr(
+        base_model, "aget_provider_response", fake_aget_provider_response
+    )
     return metric, captured
 
 
@@ -336,6 +354,87 @@ def test_geval_litellm_path_folded_whitespace_is_scored(monkeypatch):
     metric, captured = _geval_on_stubbed_provider(monkeypatch, content, entries)
 
     result = metric.score("any input")
+
+    assert captured["logprobs"] is True
+    assert captured["top_logprobs"] == 20
+    assert result.value == pytest.approx(0.07984052568223204, abs=1e-9)
+    assert result.reason == "none"
+
+
+# --- async provider-shaped GEval coverage ------------------------------------
+# GEval.ascore goes through aget_provider_response (an independent call site)
+# into the same parser, so the sync tests above cannot catch an async-only
+# regression. Fixtures and expectations mirror the sync twins exactly: the
+# parser is synchronous and shared, so identical entries/content must yield
+# identical values.
+
+
+@pytest.mark.asyncio
+async def test_geval_litellm_path_split_score_scores_near_one_async(monkeypatch):
+    # Async twin of test_geval_litellm_path_split_score_scores_near_one.
+    entries = [
+        _entry('{"', -0.01),
+        _entry("score", -0.01),
+        _entry('":', -0.01),
+        _entry(
+            "1",
+            -0.05,
+            top=[
+                {"token": "1", "logprob": -0.05},
+                {"token": "0", "logprob": -2.30},
+                {"token": "2", "logprob": -2.40},
+            ],
+        ),
+        _entry(
+            "0",
+            -0.02,
+            top=[{"token": "0", "logprob": -0.02}, {"token": "1", "logprob": -3.00}],
+        ),
+        _entry(",", -0.01),
+        _entry('"reason"', -0.01),
+        _entry('":', -0.01),
+        _entry(" ", -0.01),
+        _entry('"excellent"', -0.01),
+        _entry("}", -0.01),
+    ]
+    content = '{"score":10, "reason": "excellent"}'
+    metric, captured = _geval_on_stubbed_provider(monkeypatch, content, entries)
+
+    result = await metric.ascore("any input")
+
+    assert captured["logprobs"] is True
+    assert captured["top_logprobs"] == 20
+    assert result.value == pytest.approx(0.9007723389857002, abs=1e-9)
+    assert result.reason == "excellent"
+
+
+@pytest.mark.asyncio
+async def test_geval_litellm_path_folded_whitespace_is_scored_async(monkeypatch):
+    # Async twin of test_geval_litellm_path_folded_whitespace_is_scored.
+    entries = [
+        _entry('{"', -0.01),
+        _entry("score", -0.01),
+        _entry('":', -0.01),
+        _entry(
+            " 0",
+            -0.02,
+            top=[
+                {"token": " 0", "logprob": -0.02},
+                {"token": " 1", "logprob": -2.00},
+                {"token": " 10", "logprob": -2.50},
+            ],
+        ),
+        _entry(",", -0.01),
+        _entry('"reason"', -0.01),
+        _entry('":', -0.01),
+        _entry(" ", -0.01),
+        _entry('"none"', -0.01),
+        _entry("}", -0.01),
+    ]
+    content = '{"score": 0, "reason": "none"}'
+    metric, captured = _geval_on_stubbed_provider(monkeypatch, content, entries)
+
+    result = await metric.ascore("any input")
 
     assert captured["logprobs"] is True
     assert captured["top_logprobs"] == 20


andrescrz:
Hi @feiiiiii5 

Thanks for the contribution!

Thanks!

sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
```diff
@@ -60,34 +61,21 @@ def parse_litellm_model_output(
...
 
-        entry_dict = _to_dict(entries[score_token_position])
-        top_logprobs = entry_dict.get("top_logprobs") or []
-        token_candidate = str(entry_dict.get("token", ""))
+        # Locate the score token(s) by content instead of assuming a fixed
+        # offset: tokenizers differ on where the whitespace after `"score":`
+        # lands, and a two-digit score ("10") spans two tokens.
+        entry_indices = _locate_score_entries(entries)
+        if entry_indices is None:
+            entry_indices = [3]
```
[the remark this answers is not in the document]
feiiiiii5: Addressed in ecb2385f0 — before falling back to the fixed offset we now log the reconstructed response at debug level, so a malformed response that lands on an unrelated token is diagnosable.

sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
```diff
@@ -109,6 +97,93 @@ def parse_litellm_model_output(
         raise exceptions.MetricComputationError(GEVAL_SCORE_CALC_FAILED) from exception
 
 
+_SCORE_KEY_RE = re.compile(r'"score"\s*:\s*(\d+)')
+
+
+def _locate_score_entries(entries: list) -> list[int] | None:
+    """Find the entry indices whose tokens carry the score digits.
```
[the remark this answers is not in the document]
feiiiiii5: Addressed in 137faf992, with one correction to the premise: inside valid JSON an in-string echo can never match the key regex, because the quotes would be escaped (`\"score\": 5` defeats `\"score\"`), so every matchable occurrence IS a key position. The real residual vector is a duplicated `"score"` key — json.loads resolves duplicates to the LAST one (which is also what the no-logprob text path reads via dict_content["score"]), while the locator took the first. The locator now takes the last match so both paths agree.

sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
```diff
@@ -109,6 +97,93 @@ def parse_litellm_model_output(
...
+
+    Reconstructs the decoded text from the token stream and locates the
+    digits after `"score":`; returns the one or two indices covering them,
+    or None when the key cannot be found in the reconstructed text (the
+    caller then falls back to the legacy fixed offset).
+    """
+    token_texts = [str(_to_dict(entry).get("token", "")) for entry in entries]
+    full_text = "".join(token_texts)
+    match = _SCORE_KEY_RE.search(full_text)
```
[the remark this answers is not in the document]
feiiiiii5: Same root as the previous finding, addressed in 137faf992: for valid JSON the reason text cannot contain a matchable `"score": n` (its quotes are escaped), so the only matchable occurrences are key positions; duplicate keys now resolve to the last match, matching json.loads and the text path.

sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
```diff
@@ -109,6 +97,93 @@ def parse_litellm_model_output(
...
+
+    for token_a, prob_a in first_candidates:
+        if token_a == digits and 0 <= int(token_a) <= 10:
+            # one token covering the whole span (alternative tokenization)
+            linear_probs_sum += prob_a
+            weighted_score_sum += prob_a * int(token_a)
+        for token_b, prob_b in second_candidates:
+            combined = token_a + token_b
+            if not combined.isdecimal() or not 0 <= int(combined) <= 10:
+                continue
```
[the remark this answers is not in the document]
feiiiiii5: True limitation, kept as documented: per-position top_logprobs carry no cross-position conditionals, so P(a)·P(b|observed_a) is the only combination the data supports; linear_probs_sum normalization bounds its effect, and the combination rule (first candidate counted only when it covers the whole span) is stated in the PR description. This PR fixes the score-token location and deliberately keeps the weighted-average design unchanged — happy to revisit the weighting in a follow-up if you'd like.

sdks/python/tests/unit/evaluation/metrics/llm_judges/g_eval/test_parser.py
```diff
@@ -16,3 +16,131 @@ def test_g_eval__parse_model_output_string__score_out_of_range__MetricComputatio
...
+def test_two_digit_score_split_across_tokens_scores_near_one():
+    # {"score":10} tokenizes as {" score ": "1" "0" ...}; the score is 10 but
+    # position 3 only sees the first digit "1", so the weighted average of
+    # digit candidates lands near 0.1 instead of 1.0.
+    entries = [
+        _entry('{"', -0.01),
+        _entry("score", -0.01),
+        _entry('":', -0.01),
+        _entry(
+            "1",
```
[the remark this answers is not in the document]
feiiiiii5: Kept: the stub's second-token candidates mirror the first-token distribution of the legacy single-token path, which is what makes the two-digit assertion (≈0.09 before the fix) comparable against the pre-fix behavior it documents.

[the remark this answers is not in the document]
feiiiiii5: Correction to my previous reply (which answered a different question): on the offset-discrimination point itself, this test does fail a legacy-offset regression — the legacy path reads only entries[3] (the "1" token), whose candidates average to ~0.099, and the assertion requires >0.9 (that exact value is the pre-fix failure documented in the PR). The shifted-position variant is additionally covered by the two newer tests added in 137faf992, where the located digits are not at entries[3]: the duplicate-key fixture locates the score at the second key and the escaped-echo fixture locates it after a long reason. So the suite pins the locator both by value and by position.

feiiiiii5:
Thanks @andrescrz — all Baz comments on this PR are now either addressed or answered inline: locator-failure diagnostics in ecb2385f0, duplicate-"score"-key resolution (last-wins, mirroring json.loads and the text path) in 137faf992 with regression tests, and the two-token probability note answered inline as a documented approximation of the existing weighted-average design.

andrescrz:
Hi @feiiiiii5

Thanks for the contribution!

There's another round of Baz comments. When you have a chance, could you take a look again? For each comment, please either address it with a code change or reply with a brief explanation if you believe the current implementation is the right approach.

Thanks!

feiiiiii5:
Thanks @andrescrz — addressing the second round of Baz comments:

- Stale score-token docstring: fixed in ab62edc5e (docstring now describes content-based lookup with legacy-offset fallback).
- Duplicate-key assertion range (0.55-0.68): agreed it is too wide to pin the distribution. I will tighten it to the exact weighted value for the second key with a small tolerance once I verify the computed value locally (no code pushed for this yet to avoid guessing the constant).
- Provider-shaped GEval integration test: agreed it would cover the public path better than hand-built dicts. I will add one exercising the direct parser branch (split-score + folded whitespace) unless you prefer it as a follow-up.

Let me know if you would like the remaining two as code in this PR or as a follow-up.

feiiiiii5:
Hi @andrescrz — the remaining two second-round items are addressed in 4f14e81 (both landed in this PR, per my 09-09 note; no follow-up preference was indicated):

- Duplicate-key assertion: tightened from the 0.55–0.68 range to the exact weighted value of the second key's candidates, `pytest.approx(0.6219349153822014, abs=1e-9)` — computed locally through the parser, not guessed. Verified it fails on the pre-fix code and passes post-fix. The escaped-echo control pins the same distribution (same candidates), so I tightened it identically.
- Provider-shaped GEval test: added two tests through the public `GEval.score` LiteLLM branch with a stubbed provider response (dict-style choice + logprobs token stream), covering split-score 10 (`"1"+"0"` → `approx(0.9007723389857002)`) and folded whitespace (`" 0"` → `approx(0.07984052568223204)`), each also asserting the reason, `logprobs=True`, and `top_logprobs=20` so the real branch is pinned. Same pre/post verification: both fail on the base parser (0.09916… / MetricComputationError), all 8 pass with the fix.

`ruff@0.14.14` check passes; the `format --check` hunks on this file are pre-existing lines from 137faf992, untouched to keep the diff minimal.

feiiiiii5:
Hi @andrescrz — the async-coverage Baz comment is addressed in 88a2b7c0d: two async twins of the provider-shaped tests through the public GEval.ascore LiteLLM branch (split-score 10 and folded whitespace, same exact expectations as the sync versions since the parser is shared), plus an async provider stub for aget_provider_response. Both fail on the pre-fix parser (0.09916… / MetricComputationError) and pass with the fix; ruff@0.14.14 check clean.

sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
```diff
@@ -109,6 +104,98 @@ def parse_litellm_model_output(
...
+        index
+        for index, (begin, stop) in enumerate(offsets)
+        if stop > start and begin < end
+    ]
+    if not 1 <= len(indices) <= 2:
+        return None
+    return indices
+
+
+def _decimal_candidates(entry) -> list:
```
petrotiurin: This fails the mypy checks for no untyped defs. Please try to add a narrowest possible typehint here, falling back to Any if nothing really fits.

sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
```diff
@@ -109,6 +104,98 @@ def parse_litellm_model_output(
...
+    return [
+        (str(info.get("token", "")).strip(), math.exp(float(info["logprob"])))
+        for info in (_to_dict(cand) for cand in (entry.get("top_logprobs") or []))
```
petrotiurin: `entry.get` assumes entry is a dict and if it isn't this code will crash. We use `_to_dict` helper to avoid this issue, let's also wrap entry in it.

sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
```diff
@@ -60,34 +62,27 @@ def parse_litellm_model_output(
...
-            linear_probs_sum += linear_prob
-            weighted_score_sum += linear_prob * score
+        (
+            linear_probs_sum,
+            weighted_score_sum,
+            token_candidate,
+        ) = _weighted_score_sums(entries, entry_indices)
 
         if linear_probs_sum != 0.0:
             final_score: float = weighted_score_sum / linear_probs_sum / 10
```
petrotiurin: When the located digits span a token that also carries punctuation (e.g. `"1"` + `"0,"`), no candidate passes the decimal filter, so `linear_probs_sum` stays `0.0` and the `token_candidate` fallback below rejects `"10,"` — a perfectly parseable response ends up raising. On main the same response returns a wrong-but-non-fatal value.

Let's degrade to `_extract_score_from_text_content(choice_dict, name=name)` when there's no probability mass, the same way the short-stream and no-logprob branches already do. The text score is correct in this case (10 → 1.0), and it keeps the logprob path a pure enhancement rather than a new way to fail. Same argument applies to the `entry_indices = [3]` fallback above, which lands back on the assumption this PR is removing.

sdks/python/tests/unit/evaluation/metrics/llm_judges/g_eval/test_parser.py
```diff
@@ -16,3 +21,422 @@ def test_g_eval__parse_model_output_string__score_out_of_range__MetricComputatio
...
+        "token": token,
+        "logprob": logprob,
+        "top_logprobs": top
+        if top is not None
+        else [{"token": token, "logprob": logprob}],
+    }
+
+
+def _response(content, entries):
+    from types import SimpleNamespace
```
petrotiurin: `SimpleNamespace` is already imported at module level (line 2), so this local re-import is redundant — let's drop it.

sdks/python/tests/unit/evaluation/metrics/llm_judges/g_eval/test_parser.py
```diff
@@ -16,3 +21,422 @@ def test_g_eval__parse_model_output_string__score_out_of_range__MetricComputatio
...
+            -0.02,
+            top=[{"token": "0", "logprob": -0.02}, {"token": "1", "logprob": -3.00}],
+        ),
+        _entry(",", -0.01),
+        _entry('"reason"', -0.01),
+        _entry('":', -0.01),
+        _entry(" ", -0.01),
+        _entry('"excellent"', -0.01),
+        _entry("}", -0.01),
+    ]
```
petrotiurin: These `entries` fixtures are repeated almost verbatim four times: here, at line 333, and again in the two async twins at lines 375 and 414. The split-score and folded-whitespace pairs differ only in the token list and the expected value, and the async twins differ only in awaiting `ascore` instead of calling `score`.

Let's lift the entry lists to module-level constants and drive the four cases with `pytest.mark.parametrize` over (entries, content, expected). That keeps the same coverage in roughly a quarter of the lines and makes a future change to the weighting formula a one-place edit.

sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
```diff
@@ -60,34 +62,27 @@ def parse_litellm_model_output(
...
+        # offset: tokenizers differ on where the whitespace after `"score":`
+        # lands, and a two-digit score ("10") spans two tokens.
+        entry_indices = _locate_score_entries(entries)
+        if entry_indices is None:
+            LOGGER.debug(
+                "g_eval score key not found in the reconstructed response; "
+                "falling back to the legacy fixed token offset. Reconstructed "
+                "response: %r",
+                "".join(str(_to_dict(entry).get("token", "")) for entry in entries),
+            )
```
petrotiurin, line 81: nit: this puts the judge's full response text into a log line, and that text normally quotes the evaluated input/output. DEBUG isn't a maintainer-only level — users flip it with `OPIK_CONSOLE_LOGGING_LEVEL=DEBUG` (`config.py:144`) whenever they're debugging an integration.

Every other judge parser keeps its logging content-free and prints only the exception text — `hallucination/parser.py:26`, `factuality/parser.py:32`, `moderation/parser.py:24`, `context_precision/parser.py:24`, and the two existing calls in this file at lines 39 and 108. There's also no redaction helper in the SDK to fall back on; `_logging.py` only caps a formatted line at 1MB, which protects disk rather than content.

Let's stay with the sibling convention and drop the reconstructed response from the message — `len(entries)` is enough to tell you the locator missed.

petrotiurin (changes requested):
Thank you for opening this PR @feiiiiii5 ! Your approach looks good, but I have some concerns (outlined below), please take a look!

sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
feiiiiii5: Done in 5486cc35: the generator now reads `(_to_dict(entry).get("top_logprobs") or [])` via a hoisted `entry_dict`, so non-dict entries degrade to no-candidates instead of crashing on `.get`.

sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
feiiiiii5: Done in 5486cc35: `def _decimal_candidates(entry: Any) -> list[tuple[str, float]]`. `Any` is the honest narrow bound here since entries can be dicts or attribute-style objects (both flow through `_to_dict`); the return type is exact. mypy is clean.

sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
feiiiiii5: Done in 5486cc35: when linear_probs_sum is 0.0 the parser now returns _extract_score_from_text_content(choice_dict, name=name), same as the short-stream and no-logprob branches. The old token_candidate.isdecimal() fallback is gone — for well-formed JSON the text path yields the same value via dict_content["score"], so this unifies the paths instead of keeping two. Pinned with test_punctuation_carrying_split_digits_degrade_to_text_path ("1" + "0,"): fails pre-fix with MetricComputationError, scores 1.0 post-fix.

sdks/python/tests/unit/evaluation/metrics/llm_judges/g_eval/test_parser.py
feiiiiii5: Done in 5486cc35 — local re-import dropped, module-level import (line 2) covers both _response and _provider_response.

sdks/python/tests/unit/evaluation/metrics/llm_judges/g_eval/test_parser.py
feiiiiii5: Done in 5486cc35: entry lists lifted to _SPLIT_SCORE_ENTRIES / _FOLDED_WHITESPACE_ENTRIES module constants, shared _PROVIDER_CASES drives one parametrized sync test and one parametrized async twin — same four cases and exact expectations, roughly half the lines.

sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
feiiiiii5: Done in 5486cc35: the debug line no longer interpolates the reconstructed response — it logs only the reconstructed length alongside the fallback notice, matching the content-free convention of the other judge parsers.

sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
```diff
@@ -60,41 +62,39 @@ def parse_litellm_model_output(
...
+            token_candidate,
+        ) = _weighted_score_sums(entries, entry_indices)
+
+        if linear_probs_sum == 0.0:
+            # No probability mass on the located digits (e.g. the digits span
+            # tokens carrying punctuation, so no candidate passes the decimal
+            # filter): degrade to the text path the same way the short-stream
+            # and no-logprob branches do, instead of raising on a response
+            # that is perfectly parseable.
+            return _extract_score_from_text_content(choice_dict, name=name)
```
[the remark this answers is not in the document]
feiiiiii5: Addressed in 3b8cfe9.

Agreed on logging it: this branch had already found the score key, so a text-parsed score here is a silent fidelity drop rather than a provider capability gap. Added at `parser.py:98`:

```python
LOGGER.debug(
    "g_eval score digits carried no probability mass for metric "
    "'%s'; falling back to the score parsed from text content.",
    name,
)
```

Three notes on the shape, since your question left them open:

- **DEBUG, to match the sibling.** `parser.py:76` already degrades at DEBUG when the score key cannot be located, so the two degrading branches now log at one level, and it reaches users through opik's existing `console_logging_level` / `file_logging_level` config rather than a new channel.
- **`name` only, not the model.** `parse_litellm_model_output` receives `content`, `name` and `log_probs_supported` (`parser.py:44`); the model never reaches this function. Threading a new parameter through the callers for a debug line looked worse than the narrower message — if you would rather have the model name, say so and I will pass it down.
- **The other two fallbacks stay silent deliberately.** `not log_probs_supported` (line 60) is a caller-side opt-out and `len(entries) <= 3` (line 65) is a trivially short stream; both are normal provider behaviour, so logging them would emit a line on every run against a provider that returns no logprobs. Zero probability mass is the one case where a response we did try to weight still yielded nothing.

The regression test now pins the message as well as the value:

```python
assert "carried no probability mass" in capture_log_debug.text
assert "'g_eval'" in capture_log_debug.text
```

`capture_log_debug` is the fixture already in `tests/unit/conftest.py`, because `_logging.setup()` sets `propagate = False` on the `opik` root logger and a bare `caplog` then captures nothing from `opik.*`. Before the change the assertion fails against an empty capture; after it: 11 passed for `tests/unit/evaluation/metrics/llm_judges/g_eval`, `ruff check` and `ruff format --check` clean, `mypy` clean on `parser.py`.

feiiiiii5:
@petrotiurin all six of your points are on the branch, plus one from Baz's pass on top. Mapping each to its commit so the re-review is a single pass:

| # | Your point | Commit | What changed |
|---|---|---|---|
| 1 | `no-untyped-def` mypy failure | `5486cc3` | `def _decimal_candidates(entry: Any) -> list[tuple[str, float]]` (parser.py:156). `Any` is the honest narrow bound because an entry may be a dict or an attribute-style object; the return type is exact. mypy clean. |
| 2 | `entry.get` crashes on a non-dict entry | `5486cc3` | entries are hoisted through `_to_dict` before `.get("top_logprobs")` (parser.py:157), so a non-dict degrades to "no candidates" instead of an `AttributeError`. |
| 3 | Digits split across punctuation-carrying tokens raise on a perfectly parseable response | `5486cc3` | `linear_probs_sum == 0.0` now returns `_extract_score_from_text_content(choice_dict, name=name)`, the same degradation the short-stream and no-logprob branches use. |
| 4 | Redundant local `SimpleNamespace` re-import | `5486cc3` | dropped; the module-level import covers both builders. |
| 5 | `entries` fixtures repeated four times | `5486cc3` | lifted to `_SPLIT_SCORE_ENTRIES` / `_FOLDED_WHITESPACE_ENTRIES`, with one `_PROVIDER_CASES` table driving a parametrized sync test and its async twin -- same four cases and exact expectations, roughly half the lines. |
| 6 | nit: the DEBUG line put the judge's full response text into a log | `5486cc3` | logs only the reconstructed length (parser.py:73-80), matching the content-free convention the other judge parsers keep. |
| 7 | (Baz) log the reason for the zero-probability-mass fallback | `3b8cfe9` | DEBUG with the metric name, same level as the sibling locator-failure log; the regression test asserts the message as well as the value. |

Verified locally at `3b8cfe9`: `pytest tests/unit/evaluation/metrics/llm_judges/g_eval -q` -> 11 passed; `ruff check` and `ruff format --check` clean; `mypy` clean on `parser.py`. Only the labeler and Baz checks run on push for this repo, so if the SDK suite needs a maintainer-triggered label, tell me which one and I will note it here.

Nothing is open on my side. The six inline replies carry the before/after detail if any entry above reads differently than you intended.

sdks/python/tests/unit/evaluation/metrics/llm_judges/g_eval/test_parser.py
```diff
@@ -16,3 +16,195 @@ def test_g_eval__parse_model_output_string__score_out_of_range__MetricComputatio
...
+        _entry(" ", -0.01),
+        _entry("\"ok\"", -0.01),
+        _entry("}", -0.01),
+    ]
+    content = '{"score": 9, "score": 7, "reason": "ok"}'
+    result = parser.parse_litellm_model_output(
+        _response(content, entries),
+        name="g_eval",
+        log_probs_supported=True,
+    )
```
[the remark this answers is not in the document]
feiiiiii5: Addressed on this branch, @petrotiurin — `test_geval_litellm_path_scores_end_to_end` (test_parser.py:405) now drives the public `GEval.score()` LiteLLM branch with a provider-shaped raw response built by `_provider_response` (:280), i.e. a dict-style choice carrying message content plus the token stream, as `get_provider_response` yields it. It asserts the request flags (`logprobs is True`, `top_logprobs == 20`, :417-418) and the exact bounded value and reason (:419-420), parametrised over two cases: split-score (`1` + `0` for `10`, expected 0.9007723389857002) and folded-whitespace (`" 0"`, expected 0.07984052568223204). A legacy fixed-offset `entries[3]` read gives ~0.099 on the first and raises on the second, so an incorrectly bounded result on the real path cannot pass unnoticed now. At 3b8cfe9c.

sdks/python/tests/unit/evaluation/metrics/llm_judges/g_eval/test_parser.py
```diff
@@ -16,3 +16,195 @@ def test_g_eval__parse_model_output_string__score_out_of_range__MetricComputatio
...
+    assert 0.55 < result.value < 0.68, f"scored from the stale key: {result.value}"
```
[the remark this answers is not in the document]
feiiiiii5: Fixed — the assertion is exact now, no range: `assert result.value == pytest.approx(0.6219349153822014, abs=1e-9)` (test_parser.py:227), with the derivation spelled out just above it (7 @ -0.1 and 1 @ -2.0, softmax-weighted then /10). A first-match locator scores the stale `9` key instead (0.88084...) and fails on that value. The escaped-quote control test at :232 pins the same number for the same reason, so a looser rewrite that scans bare text rather than matchable keys also fails. At 3b8cfe9c.

sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
```diff
@@ -109,6 +103,98 @@ def parse_litellm_model_output(
...
+
+    Reconstructs the decoded text from the token stream and locates the
+    digits after `"score":`; returns the one or two indices covering them,
+    or None when the key cannot be found in the reconstructed text (the
+    caller then falls back to the legacy fixed offset).
+
+    Uses the LAST match so duplicate `"score"` keys resolve the same way
+    json.loads does — to the final one, which is also what the no-logprob
+    text path reads via `dict_content["score"]`.
+    """
```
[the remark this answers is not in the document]
feiiiiii5: Done — `parse_litellm_model_output`'s docstring now states intent instead of mechanics: the score token is located by content (the digits after the `score` key, last-wins on duplicates mirroring `json.loads`), and the legacy fixed token offset is documented as a fallback used only when the key cannot be located (parser.py:47-56). There is no remaining claim that the score is always the fourth token — a case-insensitive grep for `fourth`, `index 3`, `token 3` and `4th` over the file returns nothing at 3b8cfe9c.

sdks/python/tests/unit/evaluation/metrics/llm_judges/g_eval/test_parser.py
```diff
@@ -16,3 +21,323 @@ def test_g_eval__parse_model_output_string__score_out_of_range__MetricComputatio
...
+        _entry('"reason"', -0.01),
+        _entry('":', -0.01),
+        _entry(" ", -0.01),
+        _entry('"excellent"', -0.01),
+        _entry("}", -0.01),
+    ]
+    content = '{"score":10, "reason": "excellent"}'
+    metric, captured = _geval_on_stubbed_provider(monkeypatch, content, entries)
+
+    result = metric.score("any input")
```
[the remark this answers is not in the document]
feiiiiii5: Added, and this one is a real gap you were right to keep flagging: `test_geval_litellm_path_scores_end_to_end_async` (test_parser.py:434) is the async twin. It reuses the same parametrised `_PROVIDER_CASES` and the same helper, but `aget_provider_response` is stubbed as a separate `asynccontextmanager` (:325-332) from the sync `get_provider_response` (:318-323), and the test awaits `metric.ascore(...)`. So the two call sites are exercised independently and an async-only regression cannot hide behind the sync test. Expectations are asserted identical to the sync cases, which is the point: one shared parser must give one shared answer. Both carry `@pytest.mark.asyncio`. At 3b8cfe9c. Leaving the thread open for you or @petrotiurin to resolve.

sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
```diff
@@ -109,6 +116,99 @@ def parse_litellm_model_output(
...
+    token_texts = [str(_to_dict(entry).get("token", "")) for entry in entries]
+    full_text = "".join(token_texts)
+    matches = list(_SCORE_KEY_RE.finditer(full_text))
+    if not matches:
+        return None
+    match = matches[-1]
```
petrotiurin, line 139: `matches[-1]` mirrors `json.loads` for a duplicate key at the top level, but not for a nested one, and the locator can land on a `"score"` that `json.loads` never resolves to.

On `{"score":7, "reason": "ok", "breakdown": {"score": 3}}` this scores **0.378** from the nested `3`. `main` scores 0.622 on the same response and the text path reads 0.7 — so this is a silent regression on a response that works today. A per-criterion breakdown object is plausible here: the template asks for `"score"` and `"reason"` but doesn't forbid extra keys, and there's no structured-output enforcement.

Picking the last match that agrees with `json.loads` fixes it:

```python
match = matches[-1]
try:
    expected = str(int(json.loads(full_text)["score"]))
    match = next((m for m in reversed(matches) if m.group(1) == expected), match)
except Exception:
    pass
```

I verified this locally: all 11 tests in the g_eval suite still pass, the nested case returns to 0.622, and the duplicate-key and escaped-echo tests are unaffected. The extra `json.loads` duplicates the one already done for `reason` further down, so it can be hoisted if you prefer.

petrotiurin (commented):
Thank you for your hard work on this and for addressing the comments @feiiiiii5 ! One final concern that needs to be addressed and I think we can go ahead and merge this PR.

sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
feiiiiii5: Confirmed on the real code path. Measured on `{"score":7, "reason": "ok", "breakdown": {"score":3}}`, with the top-level digit at entry 3 and the nested one last in text order:

| tree | located entry | score |
|---|---|---|
| merge base `d53961c4` | fixed offset 3 | 0.6219349153822014 |
| `3b8cfe9` | 13, the nested `3` | 0.27397830512740046 |
| `0e48cce` | 3 | 0.6219349153822014 |
| text path (no logprobs) | — | 0.7 |

Applied your check (`str(int(json.loads(full_text)["score"]))`, last agreeing match), plus `test_nested_score_key_does_not_capture_the_located_token`, which fails on `3b8cfe9` with `scored from the nested key: 0.27397830512740046` and passes here. The duplicate-key and escaped-echo tests stay green, and a `matches[0]` mutation still kills exactly the duplicate-key one, so last-wins for top-level duplicates is pinned independently.

One thing I did not take: the optional hoist of the `json.loads`. It cannot share the one further down, since that reads the message content while this reads the token reconstruction, and those two strings differ exactly in the truncated-stream case. Keeping the fallback is load-bearing for more than my new test too: without it, 6 of the 13 tests fail. The claim in both docstrings that last-match-wins mirrors `json.loads` was true for top-level duplicates only, so both now state the value-agreement rule.

sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
```diff
@@ -109,6 +116,111 @@ def parse_litellm_model_output(
...
+    try:
+        expected = str(int(json.loads(full_text)["score"]))
+        match = next((m for m in reversed(matches) if m.group(1) == expected), match)
```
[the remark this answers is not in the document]
feiiiiii5: Measured all three against the real code path, one response per claim, across three trees (`origin/main`, this PR at `0e48cce`, after the change at `670fdd61e`):

| response | `json.loads` | `main` | `0e48cce` | `670fdd61e` | text path |
|---|---|---|---|---|---|
| `{"score":7, "reason": "ok", "breakdown": {"score":7}}` | 7 | 0.6219349153822014 | **0.7809998433984686** | 0.6219349153822014 | 0.7 |
| `{"\u0073core": 7, "reason": "ok"}` | 7 | `MetricComputationError` | 0.7 | 0.7 | 0.7 |
| `{"score": 1e1, "reason": "ok"}` | 10.0 | 0.20408677949039836 | 0.20408677949039836 | 0.20408677949039836 | 1.0 |

**Taken, and fixed in `ae048eadf`.** Row 1 is the equal-valued case and you are right that value agreement is not position agreement: both digits are `7`, so the reversed scan kept returning the nested one and the weighted average was built from the breakdown token's distribution while `json.loads` read the top-level. That row is a regression this PR introduced — `main` scores 0.6219 — and it is a worse shape than the differing-value case in row terms, because it returns a plausible number. Fix: restrict the agreeing matches to the shallowest brace depth, then take the last of those. Differential over 32 reconstructed texts (top level only, breakdown same/other value, two criteria, criteria in an array, braces inside a string, duplicate top-level keys, doubly nested): the selected span changes in 5, all of them the equal-valued-nested family; the other 27 select the same span as before. New test fails pre-fix with `assert [13] == [3]`.

**Not changed, with the measurement.** Row 2: the locator returns `None`, the caller falls back to the fixed offset, and the result is 0.7 — the same value `json.loads` gives, and better than `main`, which raises `MetricComputationError` on that response. So the escaped-key case is not made worse here; it lands where it already resolves correctly.

Row 3 is byte-for-byte the same number on all three trees, so it is inherited rather than introduced. Extending the span to the full number grammar would also cover `e1`, which `_weighted_score_sums` cannot weight since it filters candidates on `isdecimal()` — so a real fix there is a change to the weighting code, and it is only reachable if the integer contract in `_COT_SYSTEM_PROMPT` ("SCORE VALUE MUST BE AN INTEGER") is relaxed. I left it out rather than widen this diff, and said so in the body.

One correction to my own previous round while measuring this: the `matches[0]` mutation row in the body's earlier table was stale. Re-run at `0e48cce` it kills nothing, because the value-agreement step overrides the positional default whenever any match agrees, so first-vs-last only survives on an unparseable reconstruction — and the test for that path had a single score key. `670fdd61e` adds that missing case (two keys, truncated stream); it passes on `0e48cce` too, so it closes a coverage hole rather than demonstrating a fix.

feiiiiii5:
@petrotiurin your concern is closed on `0e48cce`, and a follow-up pass over the same locator found the residue of it, now fixed in `ae048eadf`.

The check you suggested compares the located digits against the parsed value, which rejects a nested key whose value differs. It cannot reject one whose value is the same: `{"score":7, "reason": "ok", "breakdown": {"score":7}}` has two agreeing matches and the reversed scan still lands on the nested one, so the average was weighted from the breakdown token's logprobs. Measured on that response: `main` 0.6219349153822014, `0e48cce` 0.7809998433984686, after the change 0.6219349153822014, text path 0.7.

The rule now restricts agreeing matches to the shallowest brace depth before taking the last of them, so duplicate top-level keys still resolve to the final one. Over 32 reconstructed texts the selected span changes only where a nested score restates the top-level value; every other shape selects the same span as before. Suite is 15 passed, `ruff@0.14.14` and `mypy@1.19.1` clean, mutation table in the body.

Two things that came out of measuring it, both in the body: the `matches[0]` mutation row from my previous round was stale (re-run at `0e48cce` it kills nothing), and the escaped-key and exponent cases Baz raised are unchanged by this PR, with `main` numbers alongside.

sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
petrotiurin: Thanks — this is a clear improvement, and the equal-valued nested case is genuinely fixed. I re-ran it locally: 15 passing, ruff and mypy clean, and the differential against `main` is clean apart from the two intended changes (duplicate top-level keys now resolving last, and the split two-digit score).

Both of the new Baz comments reproduce, though. `{"score":7, "reason": "returns }}} here", "breakdown": {"score":7}}` scores **0.781** here where `main` and the text path both read the top-level 7 — the `}}}` inside the reason string is counted as structure, which leaves the top-level key at depth 0 and the nested one at -1. And the per-match prefix slicing is quadratic: 0.5 / 1.5 / 5.5 / 24.1 ms as the response doubles from 4KB to 35KB.

One approach resolves both at once, and is a couple of lines shorter than what's there now. Rather than selecting among flat regex matches, walk the object once with a pattern that alternates "a whole JSON string" against "a brace", and read the key only at depth 1. Strings are consumed whole, so braces or a `"score"` echo inside a reason can never be seen as structure; the walk stops when the first object closes, so prose around the JSON is out of scope rather than merely deeper; and last-assignment-wins keeps duplicate top-level keys resolving as `json.loads` does. It validates nothing, so it behaves identically on a truncated reconstruction — which lets the `json.loads` and its bare `except` go away entirely.

```python
_SCORE_KEY_RE = re.compile(r'"score"\s*:\s*(\d+)')
# A whole JSON string (escapes included, an unterminated tail tolerated) or
# a brace: everything that moves or hides the object structure.
_STRUCTURE_RE = re.compile(r'"(?:[^"\\]|\\.)*"?|[{}]', re.DOTALL)


def _top_level_score_span(text: str) -> tuple[int, int] | None:
    """Character span of the digits of the top-level ``"score"`` value.

    Walks the first JSON object once, stepping over whole strings and
    reading the key only at depth 1, so it lands on the value ``json.loads``
    resolves: a nested breakdown sits deeper, a brace or a ``"score"`` echo
    inside a reason cannot shift the structure, prose sits outside the
    object, and duplicate keys resolve to the last one.
    """
    span = None
    depth = 0
    for token in _STRUCTURE_RE.finditer(text):
        if token.group() == "{":
            depth += 1
        elif token.group() == "}":
            depth -= 1
            if depth == 0:
                break
        elif depth == 1:
            key = _SCORE_KEY_RE.match(text, token.start())
            if key is not None:
                span = key.span(1)
    return span
```

`_locate_score_entries` then just calls it, with the rest of the function unchanged:

```python
    token_texts = [str(_to_dict(entry).get("token", "")) for entry in entries]
    full_text = "".join(token_texts)
    span = _top_level_score_span(full_text)
    if span is None:
        return None
    start, end = span
```

Measured on my side: your 15 tests pass unmodified, the brace-in-string case returns to 0.6219349153822014, and the scan goes linear (1.1 ms where the current code takes 8.8 ms at 38KB). A 200k-case fuzz over brace-heavy reason strings, nested breakdowns, criteria arrays, duplicate keys and random token boundaries located the wrong span 0 times, against ~1.4% for the current selection step.

One consequence worth deciding on rather than absorbing silently: this also makes a response whose *only* `"score"` is nested return `None`, fall through to the legacy offset and raise — matching `main` and the text path, where the current branch scores it from the nested key. I'd argue that's right, since otherwise the same response succeeds or fails purely on whether the provider returns logprobs, but it's your call and worth pinning with a test either way.

Separately, and not for this PR: the reason is read at `parser.py:111` with a strict `json.loads` on the raw message content, while every other read goes through `parsing_helpers.extract_json_content_or_raise`. That's why a prose-wrapped or markdown-fenced response fails on the logprob path but succeeds without logprobs.

feiiiiii5: Took the walk as written. `3b056b5d9` replaces the depth rule with `_top_level_score_span` and drops the `json.loads` agreement check along with its bare `except`; `_locate_score_entries` keeps the offset mapping and the one-or-two-indices bound unchanged.

Both of your reproductions confirmed here before I changed anything. Three trees, my own run:

| response | `main` | `670fdd61e` | `3b056b5d9` |
|---|---|---|---|
| `{"score":7, "reason": "returns }}} here", "breakdown": {"score": 7}}` | 0.6219349153822014 | entry 13, **0.7809998433984686** | entry 3, 0.6219349153822014 |
| `{"reason": "ok", "breakdown": {"score": 7}}`, the only score nested | `MetricComputationError` | entry 9, 0.6219349153822014 | `None`, `MetricComputationError` |

Timing on a rubric-shaped response that doubles each step (top-level score plus N criteria, each `{"name": ..., "score": 7}`, four-character tokens):

| bytes | criteria | `670fdd61e` | `3b056b5d9` |
|---|---|---|---|
| 4119 | 143 | 0.443 ms | 0.307 ms |
| 8208 | 284 | 1.253 ms | 0.659 ms |
| 16386 | 566 | 4.229 ms | 1.297 ms |
| 32783 | 1127 | 15.244 ms | 2.721 ms |
| 65543 | 2219 | 58.116 ms | 5.461 ms |

Your 0.5 / 1.5 / 5.5 / 24.1 reproduces. Left column is 4x per doubling, right is 2x.

On the consequence you asked me to decide rather than absorb: I took your side and pinned it. `test_score_only_nested_falls_back_instead_of_scoring_from_it` asserts the locator returns `None` and the parse raises, so the logprob path can no longer succeed on a response the text path fails.

One separation worth recording, because it explains why the previous round looked clean to me. Baz's literal example, `{"reason":"{", "score":7}`, does not reproduce: a single match leaves the shallowest-of-agreeing step nothing to choose between, so it locates correctly. `main` is the one that raises there, since the fixed offset lands on the `{` token and the zero-mass degrade did not exist before this PR. Only your `}}}` variant breaks it, because the depth comparison needs a second match to pick wrongly.

Three tests added, and the fifteen existing ones pass unmodified as you predicted. Against `670fdd61e`'s `src/` the same file gives three failures and they are exactly the three new tests.

Mutation table re-run whole at `3b056b5d9`, eighteen tests as the baseline, each mutation restored byte-for-byte with sha256 `765c826e` checked before and after:

| mutation | tests that die |
|---|---|
| read the key at `depth >= 1` instead of `== 1` | 4: braces-in-string, both nested-key tests, nested-only |
| do not stop when the first object closes | 1: `test_walk_stops_at_the_first_object_close` |
| keep the first span instead of the last | 2: duplicate-key, unparseable duplicate-key |
| match braces only, do not consume whole strings | 9 |
| raise instead of returning `None` when there is no top-level key | 1: nested-only |

No test pins the linear scan. A timing bound loose enough to survive a loaded runner would not kill the quadratic version either, since 58 ms at 64KB passes anything sane, so that measurement lives here instead of in the suite.

One shape where the walk is stricter than the rule it replaces, flagged rather than absorbed. Prose carrying a complete brace pair before the JSON, `Use {this} format: {"score": 7}`: the walk closes on the prose brace and returns `None`, where the old rule reached the real key. On that input all three trees still end in `MetricComputationError`, because `parser.py:111` parses the message content with a strict `json.loads` and rejects anything prose-wrapped before the located score matters. So the narrowing changes no outcome today, and where it eventually shows up it degrades to the legacy offset, which is `main`'s behaviour. If you would rather the walk continue past a close while `span` is still `None`, that is a one-line change and I have left it out, since you framed the stop as intended.

Your separate point checks out and stays out of this diff: `:111` uses `json.loads` on the raw message content while `parse_model_output_string` at `:21` goes through `parsing_helpers.extract_json_content_or_raise`, which is why a fenced or prose-wrapped response fails with logprobs and succeeds without them. Recorded in the body as a follow-up.

Gates at `3b056b5d9`: 18 passed in `tests/unit/evaluation/metrics/llm_judges/g_eval/`, 126 passed in `tests/unit/evaluation/metrics/llm_judges/`, `ruff@0.14.14` check and `format --check` clean on both touched files, `mypy@1.19.1` clean on `parser.py` (with `--ignore-missing-imports` locally, this machine has no `litellm` stubs; `tests/` is excluded from mypy by `.pre-commit-config.yaml`).

sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
```diff
@@ -109,6 +116,124 @@ def parse_litellm_model_output(
...
+        agreed = [
+            (m, full_text[: m.start()].count("{") - full_text[: m.start()].count("}"))
+            for m in matches
+            if m.group(1) == expected
+        ]
+        if agreed:
+            # Equal values do not identify a position: a breakdown object can
+            # restate the top-level score, and its digits come later in the
+            # stream. Unclosed braces before a key count its depth, so the
+            # shallowest agreeing matches are the top-level ones.
```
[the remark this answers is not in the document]
feiiiiii5: Resolved at this head. The finding was accurate for `670fdd61e`, which still counted braces to compare match depths; `3b056b5d9` replaced that whole step with `_top_level_score_span`, whose pattern `_STRUCTURE_RE` alternates "a whole JSON string" against "a brace", so a brace inside a string is consumed as part of the string and cannot move `depth`.

Measured with the same fixture (tokens of 4 characters, `top_logprobs` on the decimal tokens), located entry index:

| response | `670fdd61e` | `3b056b5d9` |
|---|---|---|
| `{"score":7, "reason": "returns }}} here", "breakdown": {"score": 7}}` | 16, token `" 7}}"` - the nested one | 2, token `":7, "` - top level |
| `{"reason":"{", "score":7}` | 5 | 5 |
| `{"score":7, "reason": "ok", "breakdown": {"score": 3}}` | 2 | 2 |
| `{"reason": "ok", "breakdown": {"score": 7}}` (only score nested) | 10, the nested one | `None`, so the legacy offset then the text path |

Two corrections so the finding is not carried forward with the wrong location. The depth counting lived in `_locate_score_entries`, not in `_extract_score_from_text_content`; that function hands the message content to `parsing_helpers.extract_json_content_or_raise`, which is `json.loads` with a first-`{`/last-`}` slice as fallback and has never counted depth here. `git diff --name-only d53961c46..3b056b5d9` lists only `g_eval/parser.py` and `test_parser.py`, so `parsing_helpers.py` is untouched by this PR.

Pinned by mutation, not by argument: changing `_top_level_score_span` to match braces only, without consuming whole strings, kills 9 of the 18 tests in `test_parser.py`; that row is in the body table.

sdks/python/src/opik/evaluation/metrics/llm_judges/g_eval/parser.py
```diff
@@ -109,6 +116,124 @@ def parse_litellm_model_output(
...
+    except Exception:
+        # The reconstruction can differ from the message content (a truncated
+        # stream, for instance); without a parsed value to agree with, keep
+        # the positional choice instead of failing the locator.
+        expected = None
+    if expected is not None:
+        agreed = [
+            (m, full_text[: m.start()].count("{") - full_text[: m.start()].count("}"))
+            for m in matches
+            if m.group(1) == expected
```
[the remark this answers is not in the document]
feiiiiii5: Also resolved at `3b056b5d9`, same commit as the walk. There is no prefix copy and no `.count()` any more: `_STRUCTURE_RE.finditer` yields each string or brace once, the key match is anchored with `_SCORE_KEY_RE.match(text, token.start())`, and the loop breaks when the first object closes.

Timed on rubric-shaped responses (top-level score plus N criteria objects, 4-character tokens), best of 3, one generator shared by both trees:

| bytes | criteria | `670fdd61e` | `3b056b5d9` | `_top_level_score_span` alone |
|---|---|---|---|---|
| 4,081 | 143 | 0.448 ms | 0.317 ms | 0.165 ms |
| 8,170 | 284 | 1.296 | 0.632 | 0.330 |
| 16,348 | 566 | 4.303 | 1.338 | 0.655 |
| 32,744 | 1127 | 15.310 | 2.789 | 1.292 |
| 65,504 | 2219 | 56.910 | 5.503 | 2.533 |
| growth per doubling | | 2.89, 3.32, 3.56, 3.72 | 1.99, 2.12, 2.08, 1.97 | 2.00, 1.98, 1.97, 1.96 |

The left column approaches 4x per doubling, head is 2x.

What is not pinned is the complexity itself - an assertion on a timer would be flaky, so the table is the evidence and the shape of the function (one pass, no slicing) is the reason. The early exit is pinned: removing the `break` kills exactly `test_walk_stops_at_the_first_object_close`.

petrotiurin (approved):
LGTM, thank you for fixing this and for addressing all the comments! Really appreciate the responsiveness here!

PR #7914, opened 2026-08-19
[OPIK-7940] [FE] [CI] test: e2e coverage for OAuth2 dynamic token auth

miguelgrc:
## Details

Stacked on #7910. Hermetic Playwright e2e coverage for the OAuth2 client-credentials token auth mode — no external API keys involved: a mock OAuth2 token service + bearer-validating gateway (`services/mock-token-auth/`) is spawned automatically by Playwright's `webServer`, and rejects any missing/stale bearer so specs fail loudly instead of passing vacuously.

- **`tests/ai-providers/token-auth-dialog.spec.ts`** (`@t1-smoke`, 4 tests): UI configure + Test connection + save with backend read-back asserting the masked recipe (regression test for a dropped-`auth_config` save bug caught during manual testing); sentinel round-trip (OAuth2 mode preselected on reopen, hidden `grant_type`, write-only secret with no reveal affordance and a non-removable lock, update-without-changes keeps the stored secret usable); switching back to Static API key clears the stored recipe; failed Test connection surfaces the upstream 401 with the secret never echoed.
- **`tests/ai-providers/token-auth-refetch.spec.ts`** (`@t1-smoke`): a token revoked at the gateway (via the mock's `/revoke` hook) is transparently refetched and the request retried — asserted through `/stats` count deltas, fully clock-free.
- **Playground provider-sanity parametrization**: `playground-models.yaml` entries gain optional `api_key` (literal) and `auth` blocks; two new hermetic entries (`mock-custom-static`, `mock-custom-oauth`) run the existing provision → prompt → completion journey once per auth mode. Existing entries and their code path are untouched.
- **POM**: `ConfigurationPage` gains OAuth2 auth-section methods, edit/delete flows, and a fix to the add-dialog tile selection (it previously could pick an already-configured Custom instance tile, opening Edit instead of Add). Four `data-testid`s added to the credential rows in `AuthConfigSection`.
- **CI**: both e2e workflows export `MOCK_AUTH_URL_FOR_BACKEND=http://172.17.0.1:9878` and the mock binds `0.0.0.0` under CI, so the dockerized backend can reach the host-run mock (docker-compose already ships the destination guard as `relaxed`). Specs are tagged `@t1-smoke`, so they run in this PR's e2e workflow (triggered by `tests_end_to_end/e2e/**` changes) and on every post-merge t1 run.

`core/provider-keys.ts` is a documented stopgap (fetch-based REST helpers) until the TS SDK is regenerated with `auth_config`.

## Change checklist
[...]

PR #8160, opened 2026-09-04
[OPIK-8102] [QA] Proposed e2e specs for the LLM-judge Gemini thinking level (from #8158 exploration)

CometActions:
## Details

the assertions before merging.** Full narrative below.

### Where this came from

Exploratory testing of **#8158** (`[OPIK-8102] [FE] fix: add a "none" thinking level so Flash Lite models stay non-thinking`) on that PR's own deployed environment, `https://pr-8158.dev.comet.com` (`2.2.51-8158-merge-3187`, OSS install, workspace `default`). A human worked the flows by hand there first; these two specs are the flows that came back **verified and strong**.

**This PR targets `AndreiCautisanu/OPIK-8102/flash-lite-thinking-none`, not `main`** — deliberately. Both specs assert behaviour that only exists in #8158: on `main`, `Gemini 3.5 Flash Lite` still defaults to `minimal` and its level list has no `None` at all, so the first spec fails on its very first assertion. The specs are written against **fb5fc93**, that PR's head. They should merge with #8158, not ahead of it.

### What the specs assert

The failure mode this covers is silent. A judge rule that carries a thinking block it should not looks identical in the rules list, scores traces normally, and reads back correctly in the form — the only symptom is that every judge call is slower and costs more. So each assertion is made against the **persisted evaluator over REST** (`GET /v1/private/automations/evaluators/{id}` → `code.model.custom_parameters`), with the dialog asserted separately as the thing a user actually reads.

Both are `@t2-cuj @area:online-evaluation`, in `tests/online-evaluation/online-evaluation-thinking-level.spec.ts`.

### 1. `A Flash Lite judge saves no thinking block, keeps a level that was chosen, and clears it again on None`
`@cap:online-evaluation.create-llm-judge-rule` · `@cap:online-evaluation.edit-rule`

| Step | Assertion |
|---|---|
| Create on `Gemini 3.5 Flash Lite` | dialog preselects **None**; offered levels are exactly `None, Minimal, Low, Medium, High` |
| Submit | saved evaluator's `custom_parameters` is **absent** — not `{}`, not `{thinking: {}}` |
| Reopen (kebab → Edit) | hydrates **None** |
| Set **Minimal**, save | persists exactly `{thinking: {level: minimal}}` |
| Reopen, save **untouched** | still `{thinking: {level: minimal}}` — the new `none` default must not overwrite a level someone chose |
| Reopen, set back to **None**, save | `custom_parameters` is **gone again** |

That last row is the one #8158's head commit (`make "none" clear a persisted thinking block, not just skip adding one`) exists for, and it is the row the PR's own vitest suites cannot reach: it needs the full dialog → save → rehydrate → save round trip.

`custom_parameters` is compared as a **whole object** rather than probed for a key, so a leftover sibling under `thinking` fails rather than passing quietly.

### 2. `A thinking-by-default Gemini model is untouched: Flash preselects Medium, is never offered None, and saves the level`
`@cap:online-evaluation.create-llm-judge-rule`

The control. `Gemini 3.5 Flash` still preselects **Medium**, is **not** offered `None`, and still saves `{thinking: {level: medium}}`. Without it, a later "just stop sending thinking for Gemini" simplification would silently switch every Gemini judge to non-thinking and spec 1 would still pass.

### Verification

Run against `https://pr-8158.dev.comet.com` — the same environment the exploration used — from `tests_end_to_end/e2e/`:

```
npx playwright test tests/online-evaluation/online-evaluation-thinking-level.spec.ts --reporter=list
```

| Spec | Result |
|---|---|
| Flash Lite: no block → Minimal → untouched resave → None clears it | **passed** (3 consecutive runs, ~22s) |
| Flash control: Medium preselected, no None, level saved | **passed** (3 consecutive runs, ~5s) |

Also run:

- `python3 tests_end_to_end/coverage/tag_lint.py --taxonomy … --estate tests_end_to_end` → `61 specs checked, 1 exempt, 0 problem(s)`.
- The **whole `tests/online-evaluation/` directory**, since this change touches a shared POM: **8 passed, 1 skipped, 1 failed**. The skip is the LLM-judge smoke test (`ensureModelAvailable` — no `ANTHROPIC_API_KEY`/`OPENAI_API_KEY`/`OPENROUTER_API_KEY` on this runner). The failure is `online-evaluation-python-metric-errors.spec.ts`: the environment answers `Python evaluation failed (HTTP '500'): 500 Internal Server Error: Failed to execute code` where the spec expects the classified `400 Bad Request: Execution failed: the metric produced no output`. **Not from this branch** — verified by checking out fb5fc93 clean and running that one spec, which fails identically. It looks like an environment/backend-side condition on `pr-8158.dev.comet.com` and is worth a separate look, but it is nothing to do with these specs.

`tsc --noEmit` on the estate reports exactly one error, `core/backend/client.ts: Duplicate identifier 'deleteDashboard'`, which reproduces unchanged on `main` — it is not from this branch. (Separately, `npx tsc` with the repo's pinned `typescript@7.0.2` fails on `tsconfig.json` before reaching any source: `Option 'baseUrl' has been removed`. Also pre-existing on `main`, and worth a fix in its own PR — the typecheck in the writing-e2e-tests skill's checklist cannot currently run as written.)

No completion is ever made and no LLM key is needed: the model picker lists a provider's models as soon as *a* key for it exists, and everything asserted here is decided before the judge would run. A dummy Gemini key is enough, which is what makes both specs deterministic and free.

### Supporting changes (all additive)

- **`fixtures/provider-key.fixture.ts` — new `geminiProviderKey` fixture.** Adopt-or-seed: a provider key is workspace-global and `gemini` cannot be namespaced per test the way `providerKeys.createOauth` namespaces a custom provider, so if the workspace already has a Gemini key the fixture uses it and leaves it alone; otherwise it seeds a dummy one and deletes it after `use()`, honouring `shouldLeaveArtifacts`. Because the resource is unnamespaceable, the spec declares `test.describe.configure({ mode: 'serial' })` — verified: `--workers=4` still reports "Running 2 tests using 1 worker".
- **`core/provider-keys.ts`** — `createProviderKey` now also reports the created id from the `Location` header, and takes `api_key` / optional `provider_name` so a built-in provider can be seeded. Deliberately returns `string | null` rather than throwing on a missing header, so the existing token-auth callers (which clean up by `provider_name`) are behaviour-identical; the new fixture, which needs the id, is the one that complains.
- **`core/backend/client.ts` — `getLlmJudgeModel(ruleId)`.** Reads `code.model` raw, because the pinned SDK's evaluator shape has no `custom_parameters`. Nullable on purpose: absent and empty are different answers when the whole point is removal.
- **`pom/online-evaluation.page.ts`** — new methods only; no existing method was changed, so no sibling spec's behaviour moves. Model picker scoped to its **provider group**, because Gemini and Vertex AI publish identical display labels (`Gemini 3.5 Flash Lite` is both `gemini-3.5-flash-lite` and `vertex_ai/gemini-3.5-flash-lite`) and an unscoped lookup would match two options — both the group and the option are asserted to `toHaveCount(1)` rather than taking `.first()`.

**One selector needs a reviewer's eye.** The model-parameters gear (`PromptModelConfigs`) is an icon-only button with no accessible name and no `data-testid` — its "Model parameters" string lives in a tooltip, which contributes nothing to the accessible name — so the POM selects it by its icon (`button:has(svg.lucide-settings2)`), asserted to resolve to exactly one element inside the dialog. Per `conventions.md` the right fix is a `data-testid` on the trigger, in this same change. It is **not** here on purpose: these specs were verified against a prebuilt deployment of #8158, where a newly added attribute would not exist, so adding it would have meant shipping specs I could not run. Happy to add it (`data-testid="model-parameters-trigger"`) and switch the POM over if you would rather have that than a verified run.

### Taxonomy

`online-evaluation.edit-rule` flipped to `covered: true, tier: t2-cuj` — with a note scoping the claim to the dialog's model-settings pane, since the prompt, variable mapping, filters and sampling controls are still unexercised by an edit. The spec is added to the area's `specs:` list, and `create-llm-judge-rule`'s note records the new Gemini-model creation path.

One candidate of three was dropped: **"Playground: switching to a Flash Lite model must not carry a thinking level onto the completion request."** The exploration marked it `weak`, and it is the one flow that is currently **wrong**: coming from the playground's own default Gemini model, the Thinking level control stays at **High** and `POST /v1/private/chat/completions` sends `custom_parameters.thinking.level: "high"` on `gemini-3.5-flash-lite`. Two code paths write the config on a model change (`PlaygroundPrompt.tsx`): a *provider* change calls `getDefaultConfigByProvider` and applies the new `"none"` default, while a model change *within* Gemini calls `updateProviderConfig`, which only rewrites the level when the current one is not in the new model's list — and `"high"` is in Flash Lite's list both before and after this PR.

That is **not a regression from #8158** (`updateProviderConfig` is not in the diff, and the pre-PR list also contained `"high"`), but it does mean an OSS user with only a Gemini key has no way to reach the path where the fix applies, so on the playground Flash Lite still runs with thinking at maximum. A spec written now would fail on merge. It is left unwritten pending the author's call on whether the playground's within-provider carry-over is in scope here — `playground.configure-model-settings` stays `covered: false`.

### Please read before merging

Generated by the release QA side flow — written, run and committed by an automated agent. It is a **draft** on purpose and needs a human to judge whether each assertion is the right one before it is promoted. Base branch is #8158's head, not `main`.

Related: #8158

[OPIK-8102]: https://comet-ml.atlassian.net/browse/OPIK-8102?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Change checklist

## Issues

OPIK-8102

## Testing

comment on this PR.

## Documentation


CometActions:
## 🔍 Generated-test review

# PR #8160 — [OPIK-8102] [QA] Proposed e2e specs for the LLM-judge Gemini thinking level (from #8158 exploration)

CometActions · draft: yes · 8 files, 1 new spec (2 tests)
Base: `AndreiCautisanu/OPIK-8102/flash-lite-thinking-none` (a frontend feature branch, not `main`)

**VERDICT: ready with notes** — do not flip the draft flag yourself; a human must.

This is a good spec. Every factual claim it makes about the surface it drives, I
checked against the frontend source and all of them hold. Two reliability
defects were fixed here; the two remaining items are notes, and one of them is a
decision that is explicitly not mine to make.

---

### Blockers (0)

None.

---

### Fixed in this review (3)

1. **Brittle icon-class selector on the model-parameters gear** —
   `pom/online-evaluation.page.ts:340`
   Was `this.dialog.locator('button:has(svg.lucide-settings2)')`. `lucide-settings2`
   is an internal lucide-react class name, not an API; it is derived from the
   icon's PascalCase name at build time and a version bump renames it silently.
   lucide-react is pinned at exactly `0.461.0` (`apps/opik-frontend/package-lock.json:13052`)
   but the package is **not installed on this machine**, so I could not confirm
   whether that version emits `lucide-settings2` or `lucide-settings-2` — which
   is itself the argument for not depending on it.
   **Fixed:** added `data-testid="model-parameters-trigger"` to the gear in
   `apps/opik-frontend/.../PromptModelSettings/PromptModelConfigs.tsx:125` and
   switched the POM to `getByTestId`. The POM's own docstring already said this
   was the right fix and "belongs in the FE" — per the skill this is a bucket-3
   test-only hook and belongs in *this* PR, not a follow-up. There is direct
   precedent: the `add-edit-rule-dialog-submit` testid this same POM already
   uses was added by a prior QA PR, `3a2c34baa3 [OPIK-6597] [QA] feat: Online
   Evaluation smoke (#6874)`. No role/label alternative exists — the button is
   icon-only and its "Model parameters" string lives in a Radix tooltip, which
   does not contribute an accessible name.

2. **Model-picker combobox matched by a hardcoded provider-name regex** —
   `pom/online-evaluation.page.ts`
   Was `.filter({ hasText: /Select an LLM model|Gemini|Vertex AI|OpenAI|Anthropic|Claude|GPT/i })`.
   That enumeration stops matching the day someone picks a provider outside the
   list, and it is a needless near-duplicate: `PromptModelSelect.tsx:307`
   already ships `data-testid="select-a-llm-model"`.
   **Fixed:** scoped by `has: getByTestId('select-a-llm-model')` instead.

3. **Undocumented cross-spec contention on the workspace-global Gemini key** —
   `fixtures/provider-key.fixture.ts`
   The new `geminiProviderKey` fixture warns that specs *using this fixture*
   must not run concurrently, and `mode: 'serial'` handles that. But the other
   writer of that same key is `playground-providers.spec.ts`, which
   self-provisions Gemini through the AI Providers UI and then wants a **real**
   completion from it. If both ran at once it would adopt this fixture's dummy
   key and fail on auth. They do not collide today because that spec is
   `@provider-sanity`, which runs on its own cadence outside the tier ladder.
   **Fixed:** documented the interaction in the fixture, naming the spec.

---

### Should fix (0)

---

### Notes (2)

1. **Taxonomy gap — no capability names the model-settings pane. Escalating, not fixing.**
   Both tests' real subject is the Gemini thinking-level control, and
   `online-evaluation` has no key for provider model parameters. The generator
   tagged them `create-llm-judge-rule` and `edit-rule`.

   Those tags are **honest** — I checked, and this is not the usual
   nearest-existing-key reach. Test 1 genuinely creates a rule *and* drives
   kebab → Edit → submit three times, asserting the persisted evaluator each
   time; test 2 genuinely creates one. If either capability broke, both tests
   fail. The taxonomy `note:` strings are candid about exactly what is and is
   not exercised, which is this estate's established convention, and the
   `edit-rule` note explicitly scopes itself ("Only the model-settings pane …
   the prompt, variable mapping, filters and sampling controls are still
   unexercised by an edit").

   So this is not a false green. But the *distinctive* behaviour these tests
   assert — which levels a model offers, which it preselects, and what a save
   persists into `custom_parameters` — is invisible on the coverage map. **Your
   call, and only yours:** add a capability (something like
   `model-parameters` / `thinking-level`) and tag it, or leave it as-is and
   accept that this behaviour is covered but uncounted. Per the hard rule I did
   not invent a key to make anything validate.

2. **The `tsc` gate is broken estate-wide and is currently blind. Not this PR.**
   `tests_end_to_end/e2e/tsconfig.json` sets `baseUrl`, which **TypeScript
   7.0.2** (the installed version) removed outright. The file is byte-identical
   on `origin/main`, is untouched by this PR, and was last modified in
   `2930c31491` (#6677).

   The consequence matters more than the cause: TS5102 aborts on the *config*,
   so `npx tsc --noEmit` typechecks **zero files**. The gate has not been
   verifying anything for the whole estate, not just this PR.

   I did **not** fix it, deliberately. Deleting the `baseUrl` line does not turn
   the gate green — it turns one config error into **94 real type errors** across
   the estate, nearly all `@types/node` globals (`process`, `Buffer`,
   `node:path`) not resolving under TS7, plus a genuine pre-existing
   `TS2300: Duplicate identifier 'deleteDashboard'` at `core/backend/client.ts:596`
   and `:1088` (both present on `main`). That is a TS7 migration and belongs in
   its own PR. Happy to open it if you want.

   **To get the signal this PR needed, I ran `tsc` against a temporary config
   with `baseUrl` removed. Result: zero type errors in any code this PR wrote** —
   the new spec, the POM, the fixture, `provider-keys.ts` and both index
   re-exports are all clean, before and after my edits. The only errors in a
   PR-touched file are the pre-existing `client.ts` ones listed above. The temp
   config was deleted; it is not in the commit.

---

### Bot comments

`bot-comments.json` is an empty array — baz and the other bots left no inline
review comments on this PR. Nothing to triage.

| — | *(none posted)* | — | — |

---

### Estate gates

| Gate | Before | After my edits |
|---|---|---|
| `tag_lint` | PASS | **PASS** — 60 specs checked, 1 exempt, 0 problems |
| `tsc` | FAIL | **FAIL (unchanged, pre-existing)** — TS5102 `baseUrl`, identical on `main`; see Note 2. Zero errors in this PR's own code under a scoped re-check. |
| `playwright --list` | PASS | **PASS** — 2 tests collected from the new spec; full suite still 121 tests in 57 files, unchanged |

---

### What I verified

Every factual claim the spec makes about the frontend, checked against source at
the PR head:

- `gemini-3.5-flash-lite` and `gemini-3.5-flash` exist as GEMINI models
  (`types/providers.ts:827-828`), and `vertex_ai/gemini-3.5-flash-lite` exists
  too (`:860`) — so the POM's insistence on scoping the model lookup to a
  provider group is a real hazard, not a hypothetical one.
- Flash Lite offers exactly `['None','Minimal','Low','Medium','High']` and Flash
  exactly `['Minimal','Low','Medium','High']`, **in that order**
  (`modelUtils.ts:91-110`) — matching both constants exactly.
- Defaults are `none` for Flash Lite and `medium` for Flash
  (`modelUtils.ts:244-249`) — matching both preselection assertions.
- The `"none"` level both declines to add a thinking block *and* strips a
  persisted one (`AddEditRuleDialog/schema.ts`, `formClearsThinking`), and
  `custom_parameters` is omitted entirely when it would be empty — so
  `expect(customParameters).toBeNull()` is the correct assertion, not an
  approximation.
- Placement: `tests/online-evaluation/` matches `spec_dir: online-evaluation`;
  the spec is added to the area's `specs:` list.
- Tags: exactly one tier (`@t2-cuj`), exactly one area, all `@cap:` keys exist,
  all are string literals. Tier is right — the spec spends no LLM budget.
- No `waitForTimeout`, no `.skip`/`.fixme`/`.only`, no fixed sleeps. All waits
  are web-first assertions or bounded `toPass()` polls with timeouts.
- Teardown is entirely in fixtures. `automationRulesCleanup` is pre-existing and
  discovery-based — requesting it is the whole API, so the spec destructuring it
  without calling anything is correct usage, not a dangling parameter.
- Entity names use `testNamespace`, which resolves to the swept
  `${cujPrefix}-w<worker>-<slug>` shape (`fixtures/base.fixture.ts:39-47`).
- Support code is bucket 1/2: `getLlmJudgeModel`, `findProviderKeyByProvider`,
  `deleteProviderKeyById` all extend existing patterns in files that predate the
  PR. `deleteProviderKeyByName` was refactored but kept behaviour-identical.
- **This PR carries no production behaviour change.** The frontend files in
  `git diff origin/main...HEAD` come from the two base-branch commits, not from
  this PR. The one frontend line now in the diff is the `data-testid` I added.

### What I could NOT verify

- **The tests have never been run.** No Opik stack is available here, so both
  tests are unexecuted. Everything above is source-level verification of what
  they *would* assert — it is not evidence that they pass.
- **Whether lucide-react 0.461.0 emits `lucide-settings2` or `lucide-settings-2`** —
  the package is not installed on this machine. Moot now that the selector is a
  testid, but it means I cannot say whether the original selector was already
  broken or merely fragile.
- **The frontend edit does not typecheck or lint here** — `apps/opik-frontend`
  has no `node_modules`. It is a single JSX attribute on a `Button` that spreads
  `...props` (`ui/button.tsx:64,75`) and whose `ButtonProps` extends the React
  button HTML attributes, so `data-testid` is valid and reaches the DOM; but the
  frontend's own CI is what will confirm that.
- **The backend's update semantics for an omitted `custom_parameters`.** The
  spec's headline assertion — that setting None *removes* an already-persisted
  `{thinking:{level:minimal}}` — depends on the backend clearing the field when
  the payload omits it. I read the frontend half only. If the backend treats
  absent as "no change", this test fails loudly, which is the correct outcome:
  it would be catching a real bug in the base branch's fix, not a false green.

### One thing to be aware of

This PR targets `AndreiCautisanu/OPIK-8102/flash-lite-thinking-none`, not
`main`. The `data-testid` I added therefore rides into that feature branch's PR
rather than landing independently — which is correct (the spec cannot run
without it) but worth a word to that PR's author, since it puts a frontend line
in their diff that they did not write.

---
_[review_generated_tests.yml](https://github.com/comet-ml/comet-automation-tests/actions/runs/33871313688) ·


[OPIK-8102]: https://comet-ml.atlassian.net/browse/OPIK-8102?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OPIK-6597]: https://comet-ml.atlassian.net/browse/OPIK-6597?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

CometActions:
## 🔍 Generated-test review

# PR #8160 — [OPIK-8102] [QA] Proposed e2e specs for the LLM-judge Gemini thinking level (from #8158 exploration)

CometActions · draft: yes · 9 files after fixes (was 14), 1 new spec, 2 tests

**VERDICT: ready with notes** — a person still has to flip the draft flag; this
review does not.

The spec is honest. Every `@cap:` it claims resolves to a real key, and each one
is backed by an assertion that would fail if the capability broke. Both models'
level lists and preselected defaults were checked against
`modelUtils.ts` (`MINIMAL_TO_HIGH` = minimal/low/medium/high, Flash Lite's map
entry leads with `"none"`, Flash defaults to `"medium"`) and both display labels
against `providerModels.ts` — the assertions match the product, and the POM's
provider-group scoping is genuinely required, because `Gemini 3.5 Flash Lite` is
the label for both `gemini-` and `vertex_ai/` variants.

## Blockers (0)

None remain. One would have been reported and was resolved instead — see
"Fixed in this review", item 1.

## Fixed in this review (3)

Pushed as `2c6cdde74d` on `comet-qa-bot/OPIK-8102/e2e-llm-judge-thinking-level`,
on top of a clean merge of `origin/main` (`734092ca18`). No force-push.

1. **The branch carried #8158's entire product fix as if this PR had written
   it.** The branch was cut from #8158's branch but targets `main`, so
   `git diff origin/main...HEAD` showed 6 changed frontend files — the new
   `"none"` thinking level, the changed Flash Lite defaults, and the
   `sanitizeConfigForRequest` rewrite. Read at face value that is a Step 9
   bucket-4 finding and an automatic blocker.

   It is not. **#8158 has merged**: `origin/main` (fetched 2026-09-07 12:03,
   head `26ce5830ee`) already contains every one of those lines. The branch was
   simply behind its base, and the merge-base is old enough that GitHub's own
   file list showed the merged work as this PR's. Merging `main` in was clean and
   drops all of it. What is left is the spec, its support code, and one
   `data-testid` — Step 9 bucket 3, correctly in this PR rather than split out.

   Worth stating plainly, because the merge also removed a second hazard: while
   behind, the branch's `types/providers.ts` was missing ~12 model enum entries
   that landed on main after it was cut.

2. **`submitRuleDialog()` did not wait for the save.** `AddEditRuleDialog.tsx:493`
   calls `setOpen(false)` on the line *after* `updateMutate()`, without awaiting
   it, so the dialog closes while the PATCH is in flight. The spec's four edit
   steps then read `code.model.custom_parameters` back over REST immediately —
   a real race, not a theoretical one. Worse, `onError` only raises a toast, so
   the old "wait for the dialog to hide" would have passed on a save that
   **failed**. It now waits for a 2xx on the evaluators write, following
   `projects.page.ts:49`'s existing convention of arming `waitForResponse` before
   the action. Blast radius checked: `submitRuleDialog` has no callers outside
   this PR's spec.

3. **`deleteProviderKeyById` threw a bare status.** It runs in teardown of a
   workspace-global resource, where that message is all a leaked Gemini key ever
   says for itself. It now appends the response body, matching
   `createProviderKey` five lines above it in the same file.

## Should fix (0)

Checks that passed, so you do not have to re-derive them:

- **Placement.** `tests/online-evaluation/` = the area's `spec_dir`, `@area:online-evaluation`
  agrees, and the spec is listed under the area's `specs:`.
- **Tag honesty.** `create-llm-judge-rule` on both tests (both create one) and
  `edit-rule` on the first (it opens the row kebab → Edit and resaves three
  times, asserting persistence each time). `edit-rule` flipping to `covered: true`
  carries a note that scopes it honestly to the model-settings pane. No computed
  tags; all literals.
- **Tier.** `@t2-cuj` is right: multi-step UI, but no completion is ever made, so
  it costs nothing per run — and it is too slow for t1's 3×/day.
- **Flake and isolation.** No `waitForTimeout`, no `.skip`/`.fixme`/`.only`.
  Rules named `${testNamespace}-…`, which resolves through `base.fixture.ts:45`
  to the `cuj-…` teardown contract. Cleanup is entirely in fixtures
  (`automationRulesCleanup`, `geminiProviderKey`), never in the test body. The
  `test.describe.configure({ mode: 'serial' })` is justified in a comment and
  the justification is correct — a built-in provider key cannot be namespaced.
- **Convention parity.** Compared against `online-evaluation-delete-rule.spec.ts`
  and `online-evaluation-enable-disable-rule.spec.ts`: same imports from
  `@e2e/fixtures`, same describe/tag shape, same POM usage, same
  `testNamespace` discipline. The new POM methods and the `geminiProviderKey`
  fixture extend existing patterns rather than duplicating one.

## Notes (4)

1. **No taxonomy key names the thinking-level control, and I did not invent
   one.** The assertions ride on `create-llm-judge-rule` and `edit-rule`, which
   is the nearest honest fit, and the notes disclose the scoping. If you want the
   model-settings pane counted in its own right, that is a new
   `online-evaluation.*` capability and a decision about the denominator — yours,
   not this review's.

2. **`playground.configure-model-settings` stays `covered: false`, correctly.**
   Its note reads "PromptModelSettings / providerConfigs", and this spec does
   drive that component's thinking-level control — but inside the rule dialog,
   not on the Playground page. Flipping it would claim a surface no test opens.

3. **The `tsc` gate is a no-op across the whole estate, and has been.**
   `tests_end_to_end/e2e/tsconfig.json:13` sets `baseUrl`, which TypeScript 7
   (`typescript: ^7.0.2`, resolved 7.0.2) **removed**. tsc aborts on the config
   and type-checks zero files, so the gate cannot fail on anything and cannot
   pass on anything. Identical on `main` — unrelated to this PR, and its fix
   (drop `baseUrl`, add `"types": ["node"]`) changes module resolution for every
   spec in the suite, so it belongs in its own PR with a real CI run behind it. I
   can open that.

4. **A pre-existing duplicate in the shared backend client.**
   `core/backend/client.ts` declares `deleteDashboard` twice (now lines 618 and
   1187); both are on `main`. Currently invisible because of note 3. Same
   follow-up PR.

Unfixed by choice, with the reason: the `geminiProviderKey` fixture throws if
creation answers no `Location` header, having already created the key — so that
path leaks the key it refuses to leak. It could recover the id via
`findProviderKeyByProvider('gemini')`. I left it: the throw is deliberate and
documented, the leaked key is a dummy that later runs adopt rather than
duplicate, and changing fixture control flow I cannot execute is a worse trade
than the note.

## Bot comments

Ten inline comments, all `baz-reviewer[bot]`. #6 is the bot acknowledging its own
#3 was already addressed in `9fdebd5`, so eight distinct findings.

| 1 | `listProviderKeys` doesn't check `content` is an array | invalid as scoped | The cast it criticises is the line the PR *moved*, unchanged, out of the pre-existing `findProviderKeyByName`. Validating here alone makes this the one REST helper in the file that does. Estate-wide or nowhere. |
| 2 | `submitRuleDialog` races the edit PATCH | **valid** | **Fixed in this PR.** Confirmed at `AddEditRuleDialog.tsx:493` — `setOpen(false)` runs after an un-awaited `mutate()`. Also fixes a false-green on a failed save, which the bot did not spot. |
| 3 | Locator used Lucide's `lucide-settings2` class | valid | Already fixed by the author in `9fdebd5` before this review: `data-testid="model-parameters-trigger"` + `getByTestId`, uniqueness assertion retained. Bot confirmed in #6. |
| 4 | `customParameters` camelCase hides wire `custom_parameters` | invalid | `AutomationRuleDetail` in the same file already exposes `trigger_scope` as `triggerScope`. camelCase-at-the-boundary *is* the estate's convention; renaming would make this type the outlier. |
| 5 | `customParameters` casts scalars/arrays to `Record` | valid, deferred | True, but `client.ts` casts JSON this way throughout, and the consumer here asserts `toBeNull()` or `toEqual({thinking:{…}})` — a scalar fails either assertion loudly. → same client.ts follow-up as notes 3–4. |
| 7 | `deleteProviderKeyById` loses the backend diagnostic | **valid** | **Fixed in this PR.** Trivial, local, and the file's own `createProviderKey` already did it. |
| 8 | Comment still says Flash Lite defaults to `minimal` | valid | Real, but it is in `modelUtils.ts` **on main** — #8158's merged text, untouched by this PR after the merge. → follow-up on the merged fix. |
| 9 | Reruns can display `None` while resubmitting `minimal` | valid, different surface | Checked, and the bot is right *about the optimizer/playground path*, which reloads the `parameters` blob wholesale. The **rule dialog** does not have this bug: `schema.ts:540` hydrates `custom_parameters.thinking.level` into flat `thinkingLevel`, which is exactly what this spec's "untouched reopen-and-save keeps Minimal" step asserts. → product ticket against #8158's merged change, plus a genuine coverage gap in playground/optimization that no spec touches. |
| 10 | Model-parameters button has no accessible name | valid | Not fixed here, deliberately. The POM addresses it by `data-testid`, so an `aria-label` is not a test hook — it is a user-facing a11y improvement to a shared component (playground, optimization studio, rule dialog), and I could not run the frontend suite to check it against name-based locators and snapshots. → same product ticket as #8/#9. |

Two follow-ups to open, on your go-ahead — I have not filed either:

1. **Estate: the e2e `tsc` gate is dead.** Drop `baseUrl` and add
   `"types": ["node"]`, then fix the duplicate `deleteDashboard` the gate has
   been hiding. (Notes 3 and 4; bot #5 rides along.)
2. **Product: `PromptModelSettings` follow-ups to the merged #8158.** The stale
   `minimal` comment (#8), the optimizer rerun display/submit mismatch (#9), and
   the missing accessible name (#10).

I have not replied on any bot thread. Say the word and I will draft one reply per
thread for #1, #4, #5, #8, #9 and #10 — the ones judged invalid or deferred —
and show them before posting.

## Estate gates

Run at `2c6cdde74d`, after the merge and the fixes.

| Gate | Before | After | |
|---|---|---|---|
| tag_lint | PASS | **PASS** | 63 specs checked, 1 exempt, 0 problems |
| tsc | FAIL | **FAIL** | Unchanged and pre-existing: `TS5102 baseUrl has been removed`. Identical on `main`; aborts before checking any file. See note 3. |
| playwright --list | PASS | **PASS** | 125 tests in 60 files (was 121/57 — the merge brought in main's three new dataset specs). The PR's own spec collects as 2 tests. |

Because the `tsc` gate checks nothing, I re-ran it with that one option replaced
(`"types": ["node"]` in place of `"baseUrl"`) to get real signal. Result: the
whole estate is clean **except** the duplicate `deleteDashboard`, which is on
`main`. This PR's own new code — `client.ts`, `provider-keys.ts`,
`provider-key.fixture.ts`, `online-evaluation.page.ts`, the spec — type-checks
with no errors.

## What I could not verify

- **I did not execute the spec.** There is no Opik stack in this workspace, and
  both tests need a live one plus a Gemini provider key. Everything about them is
  verified statically: collection, types, and every asserted value read out of
  the product source. What that leaves open is runtime shape — whether the
  model-picker option's accessible name is *exactly* the label (`exact: true`
  would fail loudly, not silently), and whether `readThinkingLevel`'s
  `textContent()` on the select trigger returns the bare label.
- **Fix 2 is verified by reading, not by running.** The race is real — the
  un-awaited `mutate()` is right there in the source — but I could not observe
  the old code flake or the new code stop flaking.
- **#8158's exploration evidence.** The measured latency and thinking-token
  numbers in its commit messages are the justification for the `"none"` default
  the spec asserts. I took them as given; I did not call the Gemini API.
- **Bot finding #10's blast radius.** Whether adding `aria-label="Model
  parameters"` would disturb existing frontend unit tests or name-based e2e
  locators — I did not run the frontend suite, which is why it is deferred rather
  than applied.

---
_[review_generated_tests.yml](https://github.com/comet-ml/comet-automation-tests/actions/runs/34103994183) ·


[OPIK-8102]: https://comet-ml.atlassian.net/browse/OPIK-8102?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

CometActions:
## 🔍 Generated-test review

# PR #8160 — [OPIK-8102] [QA] Proposed e2e specs for the LLM-judge Gemini thinking level (from #8158 exploration)

CometActions · draft: **yes** · 9 files, 1 new spec (2 tests)
Head reviewed: `2c6cdde74d` → pushed `6f537a0369`
https://github.com/comet-ml/opik/pull/8160

**VERDICT: needs changes — blocked on merge conflicts only.**

The spec itself is good. Its tags are honest, its placement is right, it has no
sleeps, no hand-rolled teardown, and it asserts the persisted evaluator rather
than the form that wrote it — which is the correct instinct for a control whose
failure mode is silent. Under FIX MODE I fixed three things and pushed. What
remains is a **merge blocker that a person has to resolve**, and one taxonomy
question that is a human's call by rule.

---

## Blockers (1)

### 1. Three-file merge conflict against `main` — must be resolved by hand

`mergeable: CONFLICTING`, still conflicting after my push (I re-verified with
`git merge-tree --write-tree origin/main HEAD`). All three conflicts come from a
single already-merged sibling: **opik#8161**, *"[QA] Proposed e2e specs from the
2.2.50 → 2.2.51 release exploration"*, landed on main as `1be0abf2e1` after this
PR's last merge of main (`734092ca18`). It is merged, not open — so this is
"behind main", not two live branches of one stack.

**I did not resolve them, per instruction.** Details and the correct resolution
are in the *Merge conflicts* section below. Two of the three are unions; the
third (`online-evaluation.page.ts`) is not, and taking either side wholesale
breaks a suite.

**Does this make the other findings moot?** No — but read this one first. The
spec's content survives any of the resolutions below, so the notes are still
worth acting on. One caveat: **my `closeModelParameters` fix lives inside the
conflicting POM file**, so whoever resolves the conflict must carry it through
rather than take main's side of that hunk.

---

## Should fix (0)

Fixed and pushed (`6f537a0369`) rather than listed:

| Fix | File | Why |
|---|---|---|
| `closeModelParameters` now only presses Escape while the popover is actually open | `pom/online-evaluation.page.ts` | Unconditional, an Escape arriving after something else had dismissed the popover lands on the **rule dialog behind it** and discards the edit the caller was about to save — a silent no-op save, not a failure. `toPass` always runs its body at least once, so the unguarded version had no floor. |
| `getLlmJudgeModel` refuses a scalar/array `code.model.custom_parameters` | `core/backend/client.ts` | It was cast straight to `Record<string, unknown>`. A wrong shape surfaced as a confusing `toEqual` diff two frames out instead of naming itself. Matches how the sibling `getLlmJudgeMessages` (landed on main in #8161) refuses a non-array `code.messages`, and how `requireSamplingRate` fails loudly in the same file. |
| `aria-label="Model parameters"` on the gear button | `PromptModelConfigs.tsx` | The button had **no accessible name at all** — the string lives in a tooltip, which contributes none — so screen readers announced an unnamed control. `data-testid` stays as the POM's handle. Verified no existing spec addresses a "Model parameters" button by role, so this cannot make an existing selector ambiguous. |

---

## Notes (4)

### 1. The taxonomy has no key for the thing this spec actually tests — human call

Both tests ride on `create-llm-judge-rule` / `edit-rule`. What they really assert
is the **Gemini thinking-level control**: which levels a model offers, which one
is preselected, and whether a level round-trips into
`code.model.custom_parameters`. No capability in `online-evaluation` names that.

Per the hard rule, **I did not invent one.** The PR's taxonomy notes are candid
about the scoping, so nothing here is a false green today. Your choice:

* add a capability (e.g. `judge-model-parameters`) and retag, or
* leave it riding on `edit-rule` with the note as written.

Do not let it sit unstated — after the merge, `edit-rule` flips to `covered: true`
from **two** directions (this PR and #8161) and the note is the only thing
recording that neither covers editing a rule's prompt, mapping, filters or
sampling through the dialog.

### 2. The `tsc` gate is a false FAIL, and has been giving the estate no type signal at all

```
```

`tests_end_to_end/e2e/package.json` pins `typescript: ^7.0.2` (7.0.2 installed);
TypeScript 7 removed `baseUrl`, which `tsconfig.json:13` still sets. **This PR
touches neither file** — `git show origin/main:tests_end_to_end/e2e/tsconfig.json`
has the identical line, so main fails the same way. The error aborts the compile
before a single file is checked, so *every* PR's tsc gate has been reporting this
and nothing else.

I did not fix it: it is an estate-wide config change and does not belong riding
on a spec PR. **It is worth its own PR** (drop `baseUrl`, add `"types": ["node"]`
— TS7 no longer picks `@types/node` up implicitly here), and I'm happy to open it.

To get real signal I re-ran with a corrected config. The only errors in the whole
estate are two **pre-existing** duplicates:

```
core/backend/client.ts(618,11):  error TS2300: Duplicate identifier 'deleteDashboard'.
core/backend/client.ts(1187,11): error TS2300: Duplicate identifier 'deleteDashboard'.
```

Both are on `origin/main` (lines 674 and 1243 there). **This PR's TypeScript,
including my edits, is clean.**

### 3. `getLlmJudgeModel` and main's `getLlmJudgeMessages` duplicate a prologue

Both `rawFetch('GET', '/v1/private/automations/evaluators/{id}')`, both check
`status !== 200`, both throw a near-identical message. They merge cleanly (no work
is lost) and they read different sub-objects, so this is tidiness, not a defect —
but whoever resolves the POM conflict is already in the neighbourhood and a shared
private `fetchEvaluator(ruleId)` would be the natural outcome.

### 4. `geminiProviderKey` is a workspace-global, un-namespaceable resource

The fixture is adopt-or-seed and documents this well, including the collision with
`playground-providers.spec.ts` (which self-provisions a *real* Gemini key and
wants a completion from it). `describe.configure({ mode: 'serial' })` covers the
two tests in this file; **nothing enforces the cross-file constraint** — it rests
on `@provider-sanity` running on its own cadence. That is true today and correctly
written down, but it is a comment, not a guard. Worth remembering if the tiers
ever converge.

---

## Merge conflicts (3) — do not resolve without reading both intents

Other side in every case: **opik#8161** (`1be0abf2e1`), already merged to main.

### `tests_end_to_end/coverage/taxonomy.yaml` — union, plus one real semantic merge

| | did |
|---|---|
| **main (#8161)** | adds 2 specs to `specs:`; flips `list-rules` → covered; flips `edit-rule` → covered with a note scoping it to judge-prompt hydration + no-op resave, **and an `OPEN, needs a human call` comment block** about content-array shape preservation having no capability |
| **this PR** | adds 1 spec to `specs:`; adds a `note:` to `create-llm-judge-rule`; flips `edit-rule` → covered with a note scoping it to the model-settings pane |

A correct resolution keeps: **all three specs** in the list; main's `list-rules`
entry untouched; this PR's `create-llm-judge-rule` note (main did not touch that
key); main's `OPEN, needs a human call` comment **verbatim**; and for `edit-rule`,
`covered: true, tier: t2-cuj` with a note that **unions both scopes** — the two
PRs cover genuinely different halves of the edit dialog, and either note alone
overstates what one spec proves while erasing what the other proves.

### `tests_end_to_end/e2e/core/backend/index.ts` — pure union

main adds five type re-exports (`AutomationRuleEvaluatorPageRef`, `JudgeMessageRef`,
`JudgeMessageContentPartRef`, `JudgeMessageWrite`, `JudgeMessageContentPartWrite`);
this PR adds `LlmJudgeModelDetail`. Adjacent lines, nothing else. **Keep all six.**
Taking either side drops the other's types and breaks that side's specs at compile.

### `tests_end_to_end/e2e/pom/online-evaluation.page.ts` — NOT a plain union

This is the one that needs a person. The two PRs independently wrote **the same
two helpers under different names**:

| main (#8161) | this PR | relationship |
|---|---|---|
| `openEditDialogByName(name)` | `openEditRuleDialog(name)` | **byte-identical bodies** |
| `submitDialog()` | `submitRuleDialog()` | same gesture — but this PR's waits for a **2xx on the evaluators POST/PATCH/PUT** before returning |
| `ruleRows` getter; `waitForReady`/`ruleRow` refactored onto it | — | main only |
| `cancelDialog()`, `readPromptMessageText(role)` | — | main only |
| — | `selectLLMJudgeModel`, `openModelParameters`, `closeModelParameters`, `thinkingLevelSelect`, `readThinkingLevel`, `listThinkingLevels`, `setThinkingLevel`, `fillLLMJudgeRuleBasics`, and the `llmJudgeModelCombobox` / `modelParametersTrigger` / `modelParametersMenu` locators | this PR only |

**What a correct resolution keeps.** Dedupe, then union:

1. Keep **main's names** — `openEditDialogByName` and `submitDialog` — because
   main's two merged specs already call them.
2. Keep **this PR's body for `submitDialog`.** main's version waits only for the
   dialog to hide, and the dialog's handler calls `setOpen(false)` on the line
   after `mutate()` without awaiting it: a caller that reads back over REST can
   read before the write, and a save that *failed* closes the dialog identically.
   This is a live race in main's two merged specs, not a stylistic preference.
3. Drop this PR's `openEditRuleDialog` / `submitRuleDialog` and retarget the
   **four** call sites in `online-evaluation-thinking-level.spec.ts`.
4. Keep everything else from **both** sides, including main's `ruleRows` refactor.
5. **Carry my `closeModelParameters` guard through** — it is in this file, in a
   hunk that only exists on this side.

Taking "ours" wholesale drops `ruleRows`, `cancelDialog` and
`readPromptMessageText`, breaking two specs already merged to main. Taking
"theirs" wholesale drops the entire thinking-level POM surface and this PR's spec
stops compiling.

### Reverse case — a shared file that moved under this PR without conflicting

`core/backend/client.ts` **auto-merges cleanly, and #8161 changed it too**
(`git log --oneline 26ce5830ee..origin/main -- .../client.ts` → `1be0abf2e1`).
The additions land in different regions of the same object literal, so both
survive; I checked, and no work is lost. `fixtures/index.ts`,
`core/provider-keys.ts`, `provider-key.fixture.ts` and `PromptModelConfigs.tsx`
have not moved on main since the merge base. Note the limit: this check reads
main's history, so it sees merged siblings. I did not enumerate *open* PRs.

---

## Checks that passed, briefly

* **Placement.** `online-evaluation` declares `spec_dir: online-evaluation`; the
  spec is at `tests/online-evaluation/`. Correct, and the PR already adds it to
  the area's `specs:` list.
* **Tag honesty.** Both `@cap:` keys exist. Test 1 creates *and* edits, and carries
  both tags; test 2 only creates, and carries only `create-llm-judge-rule` — the
  generator resisted the usual reach for a second key. The product constants the
  spec asserts are real and verified against `origin/main`'s `modelUtils.ts`:
  `GEMINI_3_5_FLASH_LITE` → `["none", minimal, low, medium, high]`, default `none`;
  `GEMINI_3_5_FLASH` → `[minimal, low, medium, high]`, default `medium`. Exact match.
  `@t2-cuj` is the right tier: it drives a browser through many steps but makes no
  completion, so it is deterministic and free.
* **Convention parity.** Imports `@e2e/fixtures`, drives `@e2e/pom`, uses the
  estate's `project` / `testNamespace` / `automationRulesCleanup` fixtures exactly
  as the four neighbouring `online-evaluation/` specs do. No hand-rolled teardown.
* **Flake.** No `waitForTimeout`, no `.skip`/`.fixme`, no ordering dependence.
  Every retry (`toPass`) is bounded and every custom assertion carries a message.
* **Load-bearing selectors exist.** `data-testid="select-a-llm-model"` is real
  (`PromptModelSelect.tsx:307`). `getByRole('combobox', { name: 'Thinking level' })`
  resolves via `<Label htmlFor="thinkingLevel">` + `SelectBox`'s `id` forwarded to
  the Radix `SelectTrigger`; the estate already relies on that exact pattern in
  `optimization-studio.page.ts:78`.
* **Non-spec files.** `client.ts`, `provider-keys.ts`, the fixture and the POM are
  bucket 1/2 test support, extending established patterns. `PromptModelConfigs.tsx`
  is **bucket 3** — a `data-testid` (now plus `aria-label`) on an icon-only button,
  no logic, styling or rendered-text change. It belongs in this PR, and it is here.

---

## Bot comments

11 comments from `baz-reviewer[bot]`, of which 3 are its own "addressed by commit
X" replies.

|---|---|---|---|
| 1 | `listProviderKeys` casts `body.content` without checking it is an array | **valid but estate-wide** | Not fixed here. `core/` contains **zero** `Array.isArray` guards; every list read is `page.content ?? []`. Fixing only the new file makes it the inconsistent one. → same ticket as #10. |
| 2 | `submitRuleDialog` races the edit PATCH | **valid** | Already fixed at `2c6cdde` (bot confirmed). Strictly better than main's `submitDialog` — see the POM conflict resolution. |
| 3 | `lucide-settings2` class selector is brittle; use a `data-testid` | **valid** | Already fixed at `9fdebd5` (bot confirmed). |
| 4 | `customParameters` should be named `custom_parameters` to match the wire | **invalid** | camelCase mapping **is** this file's convention: `sampling_rate`→`samplingRate`, `trigger_scope`→`triggerScope` on the neighbouring `AutomationRuleDetail`. The bot read the interface without reading the file around it. Following it would make this the one snake_case type in `client.ts`. |
| 5 | `custom_parameters` cast accepts scalars and arrays | **valid** | **Fixed in this PR** (`6f537a0369`). Distinct from #1: `client.ts` *does* fail loudly on misleading shapes (`requireSamplingRate`, and #8161's own `getLlmJudgeMessages`), so the guard follows convention rather than breaking it. |
| 6 | `deleteProviderKeyById` throws status without the body | **valid** | Already fixed at `2c6cdde` (bot confirmed). |
| 7 | `modelUtils.ts` comment still says Flash Lite defaults to `minimal` | **valid, wrong PR** | `modelUtils.ts` is **not in this PR's diff** — the feature landed on main separately and the comment is stale *there*. → main-side cleanup ticket. |
| 8 | Optimizer reruns can display `None` while resubmitting `minimal` | **valid, wrong PR** | Real display/submit divergence in main: the dropdown falls back to `getDefaultThinkingLevel` (`none`) from flat `thinkingLevel`, while the sanitizer additionally honours a nested `custom_parameters.thinking.level`. Not in this PR's diff, and **this PR's spec would not catch it** (rule dialog, not optimizer rehydration). → **product bug ticket**, and a good candidate for the bug-discovery flow. |
| 9 | Gear `<Button>` has no accessible name | **valid** | **Fixed in this PR** (`6f537a0369`): `aria-label="Model parameters"` added, `data-testid` retained. |
| 10 | `deleteProviderKeyById` now leaks an unbounded response body into logs | **valid but estate-wide** | Not fixed here. `createProviderKey` on the line above already does unbounded `await response.text()`, and has on main since before this PR. Bounding only the delete makes it the inconsistent one. → **one ticket with #1**: "provider-keys REST boundary — validate shapes and bound error bodies", estate-wide. |
| 3a/2a/6a | bot's own "addressed by commit" replies | — | no action |

**Two follow-through actions I have not taken** (say the word and I'll do them):

1. **File the deferred ones**, grouped: (a) one estate ticket for #1 + #10, the
   `core/` REST boundary; (b) one product bug for #8, with #7's stale comment
   folded in as the same file's cleanup; (c) one estate ticket for the `tsc`
   gate (note 2).
2. **Reply on the four open threads** (#1, #4, #10 and a pointer on #7/#8) so the
   invalid and deferred verdicts are visibly answered rather than silently ignored.

---

## Estate gates

Re-run after my edits, from `tests_end_to_end/e2e`:

* **tag_lint — PASS** · `63 specs checked, 1 exempt, 0 problem(s)`
* **tsc — FAIL** · `TS5102: Option 'baseUrl' has been removed` — **pre-existing and
  identical on `origin/main`**; this PR touches neither `tsconfig.json` nor
  `package.json`. Under a corrected config the estate shows only the two
  pre-existing `deleteDashboard` duplicates; **this PR contributes no type error.**
  See note 2.
* **playwright --list — PASS** · `2 tests in 1 file`, both collected.

## What I could not verify

* **The specs have not been run.** No stack, no Gemini provider key, no browser in
  this workspace. Everything about runtime behaviour — that the popover opens,
  that `getByRole('combobox', { name: 'Thinking level' })` resolves at runtime,
  that a `none` save really clears `custom_parameters` server-side — rests on
  reading the frontend source and on the estate's precedent, not on execution.
  I verified the selectors exist and that the estate already uses the same
  label-association pattern successfully; I did not watch either test pass.
* **Prettier formatting of my `PromptModelConfigs.tsx` edit.** `opik-frontend`'s
  `node_modules` is not installed here, so I could not run the frontend's
  formatter or linter. The change is a single prop plus a comment, matching the
  surrounding style, but CI's FE lint is the real check.
* **The PR body.** `review-target.json` carries no `body` field and I did not
  re-fetch, so I could not check the description for claims overtaken by events —
  worth a glance given the branch has since absorbed a merge of main and dropped
  its `modelUtils.ts` changes.
* **Other open PRs** touching these shared files. The reverse-case check reads
  `origin/main`'s history, which surfaces merged siblings (#8161); it cannot see
  an unmerged branch touching the same POM.

## Ready?

**Not yet — the conflict has to be resolved by a person first**, and the taxonomy
question in note 1 wants an answer. Once both are settled this is, in my reading,
a good spec and fit to go ready.

I have **not** marked it ready for review and have **not** merged it, and will
not. Flipping the draft flag is yours.

---
_[review_generated_tests.yml](https://github.com/comet-ml/comet-automation-tests/actions/runs/34112261570) ·


[OPIK-8102]: https://comet-ml.atlassian.net/browse/OPIK-8102?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

CometActions:
## 🔍 Generated-test review

# PR #8160 — [OPIK-8102] [QA] Proposed e2e specs for the LLM-judge Gemini thinking level (from #8158 exploration)

CometActions · draft: yes · 7 files (was 9), 1 new spec · head now `c5126e1100`

**VERDICT: ready with notes.** The spec is honest, well-scoped, and asserts the
thing it claims. It was **unmergeable** when this review started (4 conflicting
files); it now merges cleanly into `main` with 0 conflicts. All three estate
gates pass. Both tests pass against a real local OSS stack — verified — and the
one red line in the as-committed run is this instance's frontend image, not the
spec. A person still has to flip the draft flag.

---

## Blockers (0 remaining)

Both of the following were blockers on arrival and are fixed and pushed.

1. ~~**The PR could not merge** — 4 conflicting files.~~ Restacked onto `main`;
   see *Conflicts* below. `git merge-tree HEAD origin/main` now reports 0
   conflicts.
2. ~~**Three duplicate declarations that `tsc` would have rejected.**~~ `main`
   had landed `LlmJudgeModelRef` + `getLlmJudgeModel`, `submitRuleDialog` and
   `openEditRuleDialogByName` since this branch forked; the PR added its own
   near-identical copies. Dropped, spec retargeted. Details below.

---

## Should fix (0 remaining — 1 fixed)

1. **`listThinkingLevels` read the option list without waiting for it** —
   `pom/online-evaluation.page.ts`. `allTextContents()` is a one-shot query
   with no auto-wait, so a listbox that had not rendered returned `[]` — and
   `[]` asserts as *"this model offers no thinking levels"*, a much more
   alarming claim than *"the read was too early"*. **Fixed:** the option list is
   asserted onto the screen before the read.

---

## Notes

1. **The spec's assertions are sound, and that is the main finding.** The
   capability that matters here is the one the whole spec exists for: setting
   Flash Lite back to `None` must **remove** a persisted
   `{thinking: {level: minimal}}` rather than merely decline to add one. That is
   asserted on the persisted evaluator over REST, not on the form that wrote it,
   and it passes. The second test is a real control, not filler: it pins that
   `Gemini 3.5 Flash` still preselects `Medium` and is never offered `None`, so
   a "just stop sending thinking for Gemini" simplification fails there while
   the first test would still pass.

2. **Tag honesty: clean.** `@cap:online-evaluation.create-llm-judge-rule` and
   `@cap:online-evaluation.edit-rule` both exist in the taxonomy, both are
   literal strings, and both are genuinely exercised — test 1 creates a rule
   and then edits it three times; test 2 creates. `@t2-cuj` is the right tier:
   no completion is ever made (the model picker only needs a key to *exist*),
   but it needs a workspace-global Gemini key and runs serial, which rules out
   t1. No untagged real assertion.

3. **Possible taxonomy gap, for a human to decide — I did not invent a key.**
   The spec's *strongest* assertions are about the **model-settings pane**: the
   preselected level per model and the exact offered level list. No capability
   in `online-evaluation` names that surface; it rides on `create-llm-judge-rule`
   and `edit-rule`. There is precedent — `online-evaluation-edit-rule-preserves-model-parameters.spec.ts`
   already rides `edit-rule` for `custom_parameters` — and the notes I unioned
   say so explicitly, so nothing is silently widened. But the pane is shared
   with Playground and Optimizations, and a `model-parameters` capability would
   be the honest home for all three. Your call: add the key, or leave it as the
   documented scoping it is now.

4. **Placement: correct, no finding.** `tests/online-evaluation/` matches
   `online-evaluation`'s `spec_dir: online-evaluation`, `@area:online-evaluation`
   agrees, and the spec is in the area's `specs:` list.

5. **The frontend change is bucket 3, and belongs here.** The only non-test file
   is `PromptModelConfigs.tsx`: `data-testid="model-parameters-trigger"` plus
   `aria-label="Model parameters"` on an icon-only gear. No logic, styling or
   rendered text changes. The estate's own `conventions.md` says a `data-testid`
   added for a POM goes in **the same change** as the POM, and prefers
   `getByTestId` where it exists — so both the hook and the locator choice match
   the house rule. The `aria-label` is a genuine a11y fix (the button had no
   accessible name at all; its label lived in a tooltip, which contributes
   none), and it is the one line here that is arguably product rather than
   test-support. It is one attribute, it fixes a real defect baz flagged, and
   splitting it out would strand it — I left it in.

6. **Pre-existing estate duplication, not this PR's to fix.**
   `online-evaluation.page.ts` on `main` carries two pairs of byte-identical
   methods — `submitRuleDialog`/`submitDialog` and
   `openEditRuleDialogByName`/`openEditDialogByName` — landed independently by
   two earlier generated PRs. This PR would have made it three pairs; it no
   longer does. Collapsing the existing two pairs is an estate-wide change
   across several specs and belongs in its own PR. I applied 2c6cdde's
   wait-for-the-write to `submitRuleDialog` (the one this spec calls) and left
   `submitDialog` alone, with a comment saying why, rather than silently
   changing the timing of a spec this PR has nothing to do with. **Happy to open
   that cleanup PR if you want it.**

7. **The workspace-global Gemini key hazard is real and already handled.** The
   `geminiProviderKey` fixture is adopt-or-seed (it never deletes a key it did
   not create), the describe block is `mode: 'serial'` with the reason stated,
   and the fixture documents `playground-providers.spec.ts` as the other writer
   of that same key. I verified that spec is `@provider-sanity`, which runs on
   its own cadence outside the tier ladder — so the mitigation holds today, and
   the comment correctly says the two must not share a run. No other spec in the
   estate touches the `gemini` provider.

8. **Teardown is in fixtures, not the test body**, and namespacing matches the
   sweep contract — the run stamped `cuj-20260916-110155-588-…` and
   global-teardown swept that prefix clean. No `waitForTimeout` anywhere. All
   polls are bounded with messages.

---

## Conflicts — this was the blocker

Four files. All four were in `tests_end_to_end/`, so all four were mine to
resolve. The root cause: the branch was rooted on **#8158's** branch while
targeting `main`, so it inherited 4 commits of #8158's product fix and had not
seen anything `main` landed on the same shared files since. `git log origin/main`
confirms **every** shared file this PR touches has moved under it — the
online-evaluation POM (#8127, #8161), `provider-keys.ts` (#8151), the
provider-key fixture (#8172, #8151), `client.ts` and `taxonomy.yaml` (many).

#8158 has since merged, so the inherited commits are now in `main` and the
restack dropped them; the product fix the spec tests is present on `main`.

| File | `main` side | PR side | Resolution |
|---|---|---|---|
| `core/backend/client.ts` + `index.ts` | `LlmJudgeModelRef` + `getLlmJudgeModel`, added since the fork | `LlmJudgeModelDetail` + a second `getLlmJudgeModel` | **Dropped the PR's.** `main`'s is a strict superset: it also returns `temperature`, checks `type === 'llm_as_judge'`, and *already* refuses a non-object `custom_parameters` — which is exactly what the PR's own 6f537a0 added. Keeping both is a duplicate method on one object literal. The spec reads `.name`/`.customParameters`, unchanged. **Both files leave this PR's diff entirely.** |
| `pom/online-evaluation.page.ts` | `submitRuleDialog`, `openEditRuleDialogByName`, `submitDialog`, `openEditDialogByName`, `cancelDialog`, `readPromptMessageText` | `submitRuleDialog`, `openEditRuleDialog`, model picker + model-parameters popover + thinking-level select | **Dropped the two duplicates, kept both additive halves.** Spec retargeted at `main`'s `openEditDialogByName`. 2c6cdde's genuine improvement — wait for a 2xx on the evaluators write, not just for the dialog to hide — applied to `main`'s `submitRuleDialog`. The additive conflict opened mid-docblock, so the members were bridged with a closer plus a fresh `/**`. |
| `core/provider-keys.ts` | `provider_name`/`base_url` **required** | both **optional**, different docblock | **Optional wins, both docblocks kept** — each explains a different field, and the built-in `gemini` seed passes neither. `main`'s two existing callers still pass both, so nothing breaks. |
| `coverage/taxonomy.yaml` | 8-entry `specs:` list; richer `llm-judge-scores` note; 3-part `edit-rule` note + an OPEN "needs a human call" block | 1-entry `specs:`; new `create-llm-judge-rule` note; 1-part `edit-rule` note | **`specs:` unioned. Notes unioned, not picked** — each side described real coverage the other cannot see. `edit-rule` is now a 4-part note, with this spec named as the only part that *changes* a field through the dialog and the only one asserting a save that must **remove** a block. `main`'s OPEN block is preserved verbatim; **no capability key was invented.** |

**One conflict I refused to resolve, and why it mattered.** Mid-restack the
rebase stopped on `c271dde` *"Update versions to 2.2.51"* — an inherited commit
already in `main` — with conflicts in `version.txt`, `deployment/helm_chart/`
and nine `sdks/typescript` manifests. All outside `tests_end_to_end/`. Resolving
it would have reverted `main` from 2.2.66 back to 2.2.51. The restack had
reported `dropped: 4` but left **17** inherited commits in the todo; I dropped
them from the todo instead, which is what the restack should have done. **The
restack tooling under-counts inherited commits — worth a look.**

**Do the conflicts make the other findings moot?** They did, and that is why
they lead. They no longer do: the branch is 5 commits on `origin/main` and
merges clean.

---

## Sibling drafts

This PR does **not** stand alone, but it adds **zero** new collisions.

Eight other open generated drafts touch at least one file this PR touches. I
tested the three with real overlap for real, with `git merge-tree`:

| Sibling | vs this PR's head | vs `origin/main` alone | Verdict |
|---|---|---|---|
| #8282 playground dataset-run rule selection | 9 conflicts | 9 conflicts | all its own staleness |
| #8225 code-metric argument binding | 3 conflicts | 3 conflicts | all its own staleness |
| #8331 Claude sampling params | 2 conflicts | 2 conflicts | all its own staleness |

Identical counts both ways — every conflict those three have is with `main`, not
with this PR. Notably `online-evaluation.page.ts` does **not** conflict with any
sibling, because dropping the duplicate POM methods shrank the collision surface
rather than growing it. `taxonomy.yaml` conflicts for all three, but it conflicts
against bare `main` too. Nothing here needs a merge order or a consolidation.

The five remaining drafts (#8348, #8268, #8265, #8253, #8209) overlap only on
`client.ts`/`index.ts`/`fixtures/index.ts` — and this PR no longer changes
`client.ts` or `index.ts` at all, and adds exactly one type name to
`fixtures/index.ts`. No fixture-chain re-parenting was needed: the PR adds no new
fixture file, and line 1 of `fixtures/index.ts` is untouched.

---

## Spec run — the real product, and one honest red line

```
cd opik/tests_end_to_end/e2e
npx playwright test tests/online-evaluation/online-evaluation-thinking-level.spec.ts --reporter=list --retries=0
```

**As committed: 1 failed, 1 did not run.** That failure is the environment, not
the spec, and here is the evidence rather than the assertion:

```
Error: exactly one model-parameters gear in the rule dialog
Locator: getByTestId('add-edit-rule-dialog').getByTestId('model-parameters-trigger')
Expected: 1   Received: 0
```

The gear **is** on screen — the page snapshot shows it next to the model
combobox as `button [ref=e48]`, with no accessible name and no testid. So I
checked the running bundle directly:

```
docker exec opik-opik-frontend-1 grep -c "model-parameters-trigger" …/assets/App-G4UYuz9D.js  → 0
docker exec opik-opik-frontend-1 grep -c "Model parameters"          …/assets/App-G4UYuz9D.js  → 1
```

The frontend container runs a **pulled** `ghcr.io/comet-ml/opik/opik-frontend:latest`,
not an image built from this PR's head. It carries the tooltip string and
#8158's `none` logic (`Oe==="none"` → omit the `thinking` block is in the
bundle), but **not this PR's own `data-testid`**. A bucket-3 test hook that ships
in the same PR as its spec cannot resolve against a frontend built before it.

**Which of the three is it?** None of the three. Not my fix, not a wrong spec,
not a product defect — a build gap in the harness. So I diagnosed it the only
way that settles it: re-ran with an equivalent selector for the *same* button
(`button:has(svg.lucide-settings2)`), which was **not committed** and was
reverted immediately after:

```
✓ 1 … A Flash Lite judge saves no thinking block, keeps a level that was chosen, and clears it again on None (21.4s)
✓ 2 … A thinking-by-default Gemini model is untouched: Flash preselects Medium, is never offered None, and saves the level (4.4s)
2 passed (29.9s)
```

**Both tests pass against the real product**, including the assertion the spec
exists for. I did not weaken anything in the PR to get there, and the committed
locator is the one the estate's convention asks for.

**What this means for you:** the spec is correct and will pass wherever the
frontend is built from source, which is how CI runs it. It will fail against any
deployment whose frontend predates this merge — which is exactly why the testid
rides in this PR rather than a follow-up.

---

## Bot comments

15 inline comments from `baz-reviewer[bot]`; 5 of those are its own
"commit X addressed this" acknowledgements, leaving 10 findings.

| 1 | `listProviderKeys` returns `body.content` without checking it is an array | **valid but house-wide** | Not fixed. `provider-keys.ts` is a documented stopgap module with zero response-shape validation in 171 lines, and `client.ts` casts REST payloads the same way throughout. Fixing it only in the function this PR touched makes that one function the inconsistent one. → estate-wide ticket, offered below. |
| 2 | Immediate REST assertions race the edit PATCH | **valid, and the best finding of the set** | **Fixed** (2c6cdde, re-applied by me to `main`'s `submitRuleDialog`). It was right about more than it claimed: a *failed* save closes the dialog identically via a toast, so the old wait would also have passed on a write that never happened. |
| 3 | Locating the gear by `svg.lucide-settings2` is brittle; add a `data-testid` | **valid** | **Fixed** (9fdebd5). Lucide derives that class from the icon name at build time and renames it on a version bump. Matches the estate's `conventions.md`. |
| 4 | REST type exposes wire `custom_parameters` as `customParameters` | **moot** | The type it criticises (`LlmJudgeModelDetail`) no longer exists — dropped as a duplicate of `main`'s `LlmJudgeModelRef`, which uses the same camelCase mapping as every other `*Ref` in `client.ts`. The convention it objects to is the file's, not this PR's. |
| 5 | `customParameters` casts any non-null JSON to `Record<string, unknown>` | **valid, already true on `main`** | **Fixed**, then made redundant: 6f537a0 added the guard here, and `main`'s `getLlmJudgeModel` already had the identical check. The surviving method has it. |
| 7 / 12 | `deleteProviderKeyById` throws a bare status, losing the backend diagnostic | **valid and local** | **Fixed** (2c6cdde). It runs in teardown of a workspace-global resource, where that message is all a leaked Gemini key ever says for itself. |
| 8 | `modelUtils.ts` comment says Flash Lite defaults to `minimal`, should say `none` | **not this PR's** | `modelUtils.ts` was only ever in this diff as part of #8158's inherited commits. #8158 has merged; the restack dropped it. → belongs on #8158 / a follow-up against `main`. |
| 9 | Optimizer form can display `None` while resubmitting `minimal` on rerun | **not this PR's, but looks like a real product bug** | Same file, same reason — inherited, now gone from the diff. This one is worth someone's attention on `main`: it describes a hydration path where the flat `thinkingLevel` and nested `custom_parameters.thinking.level` disagree. **Not a QA fix. Recommend filing it as a bug against the #8158 change.** |
| 10 / 14 | Icon-only gear has no accessible name | **valid** | **Fixed** (6f537a0) — `aria-label="Model parameters"`, `data-testid` retained. |
| 13 | The delete error now embeds an unbounded, possibly credential-bearing response body | **invalid as framed** | The body comes from Opik's own `/llm-provider-key/delete` error path, not an attacker; the fixture posts a hardcoded dummy key and the endpoint does not echo secrets in a 4xx. `createProviderKey` in the same file has embedded `response.text()` since #7914, so the "unbounded body" pattern is the file's convention — and this is fixture teardown output in a Playwright log, not a production log sink. Bounding it here alone would make the diagnostic *worse* precisely when it is needed. |

**Two follow-through actions I have not taken — say the word:**

1. **One ticket, one theme:** response-shape validation at the `provider-keys.ts`
   REST boundary (finding 1, and finding 13's bounding if you disagree with me).
   Estate-wide, not this PR.
2. **One bug report:** finding 9, the optimizer rerun hydration mismatch on
   `modelUtils.ts`. Product, on `main`, found by the bot while reviewing a test
   PR — it will be lost otherwise.
3. **Replies on the threads** I judged invalid (13), moot (4, 8, 9) or deferred
   (1), so they read as answered rather than ignored. I have drafts ready.

I did not post replies, resolve threads, or file anything.

---

## Estate gates

Run after every edit, on the final pushed tree:

### `tag_lint` **PASS** · `tsc` **PASS** · `playwright --list` **PASS**

```
tag-lint: 88 specs checked, 1 exempt, 0 problem(s)
tsc --noEmit: clean
playwright --list: Total: 168 tests in 85 files
```

All three arrived **FAIL** — every one of them because of the unresolved
conflict markers in the working tree (`tag_lint` could not even parse
`taxonomy.yaml`). Worth noting the PR's own commit messages claim `tsc` "still
fails on the pre-existing TS5102 `baseUrl` error"; that is **no longer true** —
#8279 fixed the tsc gate on `main`, and `tsc` is now clean. The stale claim is
in the commit history, not in anything shipping.

---

## What I could not verify

- **The spec passing with its own committed locator.** The local frontend image
  does not contain this PR's `data-testid`. I verified everything else about
  both tests against the real product with an equivalent selector and reverted
  it; I did not, and could not, watch `getByTestId('model-parameters-trigger')`
  resolve here. That needs a frontend built from this PR's head.
- **Behaviour against a real Gemini key.** Every assertion here is decided
  before a judge call, so this is by design, not a gap — but it does mean
  nothing in this spec proves the persisted `thinking` block changes what the
  provider actually receives.
- **The cross-spec Gemini collision under real concurrency.** I confirmed
  `playground-providers.spec.ts` is the only other writer and that it is
  `@provider-sanity` (separate cadence). I did not run the two together to prove
  the collision, and would not want to.
- **`main` moving again.** The clean-merge and zero-new-sibling-collision
  results are as of `origin/main` at `0931f3bb0c`. `taxonomy.yaml` and
  `client.ts` both move most days.

---

**This PR is ready to go ready.** I am not flipping the draft flag and I have
not merged it — that is a person's call, by design.

The specs were re-run after the fixes and **did not pass** (1 failed). The fixes above may be wrong, or the failure may be pre-existing or environmental — the run log is in the workflow artifact. **Do not merge on the strength of this review until that is resolved.**

---
_[review_generated_tests.yml](https://github.com/comet-ml/comet-automation-tests/actions/runs/35086406256) ·


[OPIK-8102]: https://comet-ml.atlassian.net/browse/OPIK-8102?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #8161, opened 2026-09-04
[OPIK-7791] [QA] Proposed e2e specs from the 2.2.50 → 2.2.51 release exploration

CometActions:
## Details

> **Generated by the release QA side flow, and it needs human review before merge.**
> Nothing here was written or reviewed by a person yet. It is a draft on purpose.

### Where these came from

The 2.2.50 → 2.2.51 release exploration worked the release's manual-verification
list against `staging.dev.comet.com/opik` (workspace `opik-testing`). Both flows
below were **verified by hand on staging first** — these specs make permanent
what that pass checked once.

Both come from [#8150](https://github.com/comet-ml/opik/pull/8150) (OPIK-8250):
a judge message's content is persisted as a plain `String` whether the author
typed prose or the UI built a content array, so `AutomationModelEvaluatorMapper`
inferred the shape from the leading character. A prompt legitimately opening with
`[` was parsed as JSON, failed, and 500'd the whole evaluators listing — every
rule in the project plus the workspace-wide `findAll()` the online-scoring
sampler calls per trace batch.

### Which ref these were written and run against

**Written and run against the release tag `2.2.51`** (`d5ac079`), which is the
ref the exploration verified. `main` is simply where they have to merge, and it
will have moved on — please re-run on `main` before promoting this out of draft.

### 1. `online-evaluation-json-looking-judge-prompts.spec.ts`

Covers **`@cap:online-evaluation.list-rules`** (`@t2-cuj`).

Seeds a prose control rule plus six judge prompts that each open with `[` and
reach a different branch of the mapper's fallback — `[Source Text] …` (the
customer prompt, not JSON at all), `[1, 2] …` (valid JSON, wrong element type),
`[null] …`, `[{}] …` (no `type` discriminator), `[] …`, and
`[{"type": "text", "text": "example"}] Now grade the output above.` Then asserts:

- the project-scoped listing answers **200** and carries **exactly** the seeded
  set — the whole collection, not just containment, so a listing that silently
  dropped the unreadable rows would fail rather than pass;
- the **workspace-wide** listing (the sampler-shared read) answers 200 and is
  non-empty;
- the rules page renders a row per rule, with **no 5xx from
  `/automations/evaluators`** during load.

The control rule is what makes a green listing mean something: the regression was
collateral, one bad row took down every other rule in the project too.

**Verification: PASSED** on staging against `2.2.51`.

### 2. `online-evaluation-judge-prompt-round-trip.spec.ts`

Covers **`@cap:online-evaluation.edit-rule`** (`@t2-cuj`), two tests.

**Test 1 — an example-array-then-prose prompt survives the edit dialog and a
re-save byte for byte.** The old reader stopped at the first complete JSON array,
so `[{"type": "text", "text": "example"}] Now grade the output above.` lost the
instruction it exists for, and the next save wrote the truncation down. The spec
seeds over REST, asserts the API returns the string whole and *not* promoted to
`content_array`, opens the row's **Actions → Edit** dialog and asserts the
hydrated prompt is byte-identical, **submits the dialog unchanged**, reopens it
and asserts it again, then re-checks the server. The no-op submit is the gesture
that turned a read bug into permanent data loss.

**Test 2 — a genuinely multimodal judge message still reads back as structured
content.** The guard against over-correcting: falling back to "it is all just a
string" would make test 1 pass while destroying every real multimodal rule. Seeds
a `content_array` with a text part and an `image_url` part, asserts the read-back
keeps `url` and `detail`, re-saves exactly what was read, and asserts it again.

**Verification: both PASSED** on staging against `2.2.51`.

### How they were run

From `tests_end_to_end/e2e/`, against `$OPIK_BASE_URL` = `https://staging.dev.comet.com/opik`:

```bash
WORKERS=2 npx playwright test \
  tests/online-evaluation/online-evaluation-json-looking-judge-prompts.spec.ts \
  tests/online-evaluation/online-evaluation-judge-prompt-round-trip.spec.ts \
  --reporter=list --retries=0
# 3 passed (14.4s)
```

Plus the estate's own three checks from `.agents/skills/writing-e2e-tests`:

- **`tag_lint.py`** — `61 specs checked, 1 exempt, 0 problem(s)`.
- **Feature-directory run** (I touched a shared POM) —
  `npx playwright test tests/online-evaluation/` with `WORKERS=2` leaves
  `online-evaluation-non-object-sections.spec.ts` red, in
  `LogsPage.waitForReady` waiting on the traces table. **This is pre-existing
  staging flake, not a regression from this PR**: the same directory run on a
  clean `2.2.51` tree with the same settings fails *three* specs (that one plus
  both `online-evaluation-sampling-rate` tests), all in the same
  `LogsPage.waitForReady` on the traces table. That POM is untouched here. All
  three pass when run alone.
- **`npx tsc --noEmit`** — currently fails at the config level on `2.2.51` for
  everyone, before any file is checked:
  `tsconfig.json(13,5): error TS5102: Option 'baseUrl' has been removed.`
  (`typescript: "^7.0.2"` resolves to 7.0.2, which dropped `baseUrl`). Left
  alone — it is not this PR's to fix. Typechecked instead with TypeScript 5.9.3
  over the same config: the only error is the pre-existing
  `core/backend/client.ts: Duplicate identifier 'deleteDashboard'`, present
  identically on a clean tree. **This PR adds no new type errors.**

I also ran a **negative control** on the two assertions doing the most work — the
CodeMirror prompt read and the UI row count — by mutating each to the wrong
expected value. Both failed, and the prompt read reported the full string
including the trailing instruction and the blank line. Neither assertion is
passing vacuously.

### What is in the diff besides the specs

- `core/backend/client.ts` — `createLlmJudgeRule` (the existing
  `createAutomationRule` only builds `user_defined_metric_python` rules, which
  have no `content` / `content_array` split), `findAutomationRuleEvaluatorsPage`
  (returns the HTTP status instead of throwing, because the status *is* the
  assertion), `getLlmJudgeMessages`, `resaveAutomationRuleFromReadBack`.
- `pom/online-evaluation.page.ts` — `openEditDialogByName`, `submitDialog`,
  `cancelDialog`, `readPromptMessageText`, and a `ruleRows` getter that
  `waitForReady` / `ruleRow` now share. No selector changes.
- `coverage/taxonomy.yaml` — both specs added to `specs:`; `list-rules` and
  `edit-rule` flipped to `covered: true` at `t2-cuj` with scoping notes.

No FE change was needed — every element already had a `data-testid` or a stable
role.

**On the `edit-rule` claim:** test 1 opens the dialog through the row's Edit
action, asserts what it hydrated, and submits it, so the capability is genuinely
exercised. But editing a *field* through the dialog — renaming, changing the
model, rewriting the prompt — is still uncovered, and the taxonomy `note:` says
so rather than implying more than was tested.

The exploration produced two candidates and **both are here; nothing was
dropped.** The list was already filtered upstream — three of its five ranked
items produced no candidate:

- **Item 3** (Liquibase `000121` against clustered ClickHouse) — skipped during
  exploration: deployment-level DDL on a shared managed environment.
- **Item 4** (`aiu_nano` on ingested copilot spans) — blocked: `cipx_spends` /
  `cipx_spend_blocks` have DAOs and an ingestion listener but **no REST read
  surface**, so the write is triggerable and the assertion is not without direct
  ClickHouse access.
- **Item 5** (2.2.51 TS SDK packages publish) — verified against the npm
  registry, but that is release plumbing, not app behaviour; there is nothing in
  the e2e estate for it to assert.

Items 3 and 4 are the whole of [#8155](https://github.com/comet-ml/opik/pull/8155)
and neither got a hands-on check on staging. **If anything in this release still
needs a human, it is those two, and it needs a self-hosted deployment rather than
staging.**

One incidental thing the exploration flagged, unrelated to this release and not
asserted on here: the online-evaluation page fires three failing background
requests on load regardless of rule content — `403 GET
/v1/private/datasets/export-jobs`, `404 GET
/v1/private/agent-configs/blueprints/history/projects/{id}`, `404 GET
/v1/private/agent-insights/jobs/{id}`. Spec 1's wire check is scoped to
`/automations/evaluators` and to 5xx precisely so it does not become a monitor
for those.

## Change checklist

## Issues

OPIK-7791

## Testing

comment on this PR.

## Documentation


CometActions:
## 🔍 Generated-test review

# PR #8161 — [OPIK-7791] [QA] Proposed e2e specs from the 2.2.50 → 2.2.51 release exploration

CometActions · draft: yes · 6 files, 2 new specs · head `38cc35d` → review commit `5fbcb78`

**VERDICT: ready with notes** — one open decision for a human (the `edit-rule`
tag on the multimodal test), and one estate-wide gate failure this PR did not
cause and must not fix.

Do not flip the draft flag on the strength of this report. A person makes that
call; this review does not, and did not.

---

## Blockers (0)

None. Both specs collect, both are placed correctly, the tags that matter are
honest, and there are no fixed sleeps, no in-body teardown, no `.skip`, and no
computed tags.

---

## Should fix (1 — a decision, not an edit)

**1. `@cap:online-evaluation.edit-rule` on a test that never opens the edit dialog**
— `tests_end_to_end/e2e/tests/online-evaluation/online-evaluation-judge-prompt-round-trip.spec.ts:94`

`A genuinely multimodal judge message still reads back as structured content` is
API-level throughout: create over REST, `GET`, `PATCH` the read-back verbatim,
`GET` again. It never opens a browser. What it actually asserts is **judge-message
content-shape preservation** — that a `content_array` survives a read-then-write
and is not flattened into `content`. No capability in the `online-evaluation`
area names that.

Why it is not a blocker: `edit-rule` is *also* carried by the sibling test at
line 24, which genuinely drives the dialog — opens it, asserts the hydrated
prompt byte-exact, submits unchanged, re-reads. So `covered: true` is earned by
real UI evidence; the second tag widens the claim rather than fabricating it.

Why I did not fix it: the two honest fixes are (a) add a capability for
content-shape preservation, or (b) drop the tag — and (b) would leave the test
with zero `@cap:`, which the tag grammar forbids. (a) is a change to the
product's coverage denominator and belongs to you, not to this review. **I did
not invent a key.** The gap is now recorded as an `OPEN, needs a human call`
comment above `edit-rule` in `taxonomy.yaml` so it cannot decay into a silent
widening.

---

## Fixed in this PR (2)

Both pushed as `5fbcb78`.

**1. `JsonLookingPrompt.why` was dead data** —
`online-evaluation-json-looking-judge-prompts.spec.ts`. Six prompt shapes each
carry a `why` explaining which branch of `AutomationModelEvaluatorMapper`'s
fallback they reach, and nothing ever read it: the UI row assertion failed with
`row for <name>`. The interface's own docstring claimed the label appears "in
every failure message", which was true of the name and false of the reason. Now
carried as `{name, why}` through the UI step, so a missing row reports
`row for …-empty-object — valid JSON object element carrying no 'type'
discriminator` and names the regression on the first read.

**2. The `edit-rule` note read as if the dialog was driven for both prompt shapes**
— `coverage/taxonomy.yaml`. It said "dialog hydration + no-op re-save … in both
the plain-string and content_array shapes". Only the plain-string half is a
dialog gesture; the `content_array` half is REST-only, and for a good reason the
spec states — the dialog renders no image part, so the `url`/`detail` a
regression would drop are not observable in the UI at all. The note now scopes
each half separately. This matters because the note is what the next person
reads instead of re-deriving the spec.

---

## Notes (3)

**1. The `tsc` gate is red on `origin/main`, and this PR is not the cause.**

```
```

`tests_end_to_end/e2e/package.json` declares `typescript: "^7.0.2"` and
`package-lock.json` pins `7.0.2`; TypeScript 7 removed `baseUrl`, which
`tsconfig.json:13` still sets. None of `tsconfig.json`, `package.json` or
`package-lock.json` appears in this PR's diff, and TS5102 is raised against the
config before any source file is read.

Verified, not inferred: I checked `origin/main` out into a separate worktree
against the same `node_modules` and ran the gate — it fails identically there.

**I deliberately did not fix it**, and removing `baseUrl` is not the fix.
Stripping that one line locally turns 1 error into 94: `@types/node` globals
stop resolving estate-wide under TS 7 (`process`, `Buffer`, `node:path`,
`NodeJS` — in `config/env.config.ts`, `core/local-runner/connect.ts`,
`core/comet/pending-users-registry.ts` and a dozen others), all of which are
also on main. This is a toolchain migration for its own PR, not a line to slip
into a QA test PR. Happy to open that PR if you want it.

**2. Pre-existing duplicate in the file this PR extends.**
`core/backend/client.ts` defines `deleteDashboard` twice — at 629 and 1121 on
this branch, 574 and 1066 on main. Pre-existing, surfaced only by the TS 7 run
above. The PR adds 277 lines to that file and does not touch either definition.
Worth folding into the toolchain PR rather than this one.

**3. Type health of this PR's own code, checked under a workaround.** Because
the gate never reaches the source files, I re-ran `tsc` against a copy of the
config with `baseUrl` stripped and filtered to this PR's files. The two new
specs, the POM and the new client code produce **zero** errors — including
after my edits. That is the closest to a real type check available on this
toolchain; it is not the estate's own gate passing.

---

## What I verified positively

- **Placement.** `spec_dir: online-evaluation`, both specs in
  `tests/online-evaluation/`, both added to the area's `specs:` list,
  `@area:online-evaluation` matching all six neighbours. No finding.
- **`list-rules` is honestly covered.** The spec asserts the project listing
  answers 200 and equals the *exact* seeded set (not containment — a listing
  that silently dropped the unreadable rows would pass containment and fail
  this), that the server's `total` agrees, that the workspace-wide read the
  sampler shares also answers 200 with a non-empty total, and that the rules
  page renders one row per rule and nothing else, with no 5xx on the evaluators
  endpoint. The prose control rule is what makes a green listing mean something.
- **Tier.** `@t2-cuj` on both, matching every neighbour but the smoke spec. The
  seeded rules never score — no traces are logged against them — so no LLM
  budget is spent. No tier inflation.
- **Teardown.** Both specs request `automationRulesCleanup`, which is the
  fixture's whole documented API. Rules do not cascade with their project, so
  this is required, and it is a fixture rather than in-body cleanup. Names are
  built from `testNamespace`, which is `cujPrefix`-derived, so the estate's
  timestamp sweep will find them. The `project` fixture is test-scoped and
  freshly created, which is what makes the exact-set listing assertion sound
  rather than flaky.
- **Selectors resolve against real product code.** `add-edit-rule-dialog-submit`
  (`AddEditRuleDialog.tsx:786`), `playground-message-row` and
  `playground-message-editor` (`LLMPromptMessage.tsx:239,348`) all exist, and
  `data-role={role}` sits on the *same* Card as the row testid — so the POM's
  `.and()` of two attributes is correct rather than a `filter({ has })`.
  `LLM_MESSAGE_ROLE.user === "user"` matches the POM's parameter.
- **The CodeMirror read is safe here.** `readPromptMessageText` joins `.cm-line`
  text nodes. I checked the variable-highlight plugin: `{{output}}` is a
  `Decoration.mark` (styling, text preserved), and the only replacing widget is
  the variable *hint*, which returns `Decoration.none` unless the facet is set
  while typing. The dialog does not pass `compact`, so no collapsed-overflow
  editor either. The POM already documents the virtualisation limit for long
  prompts.
- **Support code is bucket 1.** New POM methods, new backend-client routes and
  new exported types, all extending established patterns in files that long
  predate this PR. No production code, no test-only hooks needed — the
  frontend already carried every testid these specs use. Nothing outside
  `tests_end_to_end/` except the frontend files I *read*.
- **Blast radius of the shared changes.** The POM change is a refactor: a new
  `ruleRows` getter, with `waitForReady` and `ruleRow` rewritten to call it —
  same locator string, no behaviour change for the six existing specs that
  import `OnlineEvaluationPage`. `client.ts` and `index.ts` are pure additions.

---

## Bot comments

`bot-comments.json` is an empty array — no inline review comments from baz or
any other bot on this PR. Nothing to triage, nothing to reply to.

| — | *(none — no bot comments on this PR)* | — | — |

---

## Estate gates: tag_lint PASS · tsc FAIL (pre-existing on main) · playwright --list PASS

Re-run after my edits:

| Gate | Result | Notes |
|---|---|---|
| `tag_lint.py --taxonomy coverage/taxonomy.yaml --estate .` | **PASS** | 61 specs checked, 1 exempt, 0 problems |
| `npx tsc --noEmit` | **FAIL** | TS5102 `baseUrl` — identical on `origin/main`, see Note 1 |
| `npx playwright test --list` | **PASS** | 3 tests in the 2 new files collect; 122 tests in 58 files estate-wide |

---

## What I could not verify

- **I did not execute either spec.** No Opik stack was available to this review,
  so everything below is reasoning against the source, not a green run.
- **Whether the edit dialog submits cleanly when the seeded model is not from a
  configured provider.** `createLlmJudgeRule` defaults to `model: 'gpt-4o'`,
  while the smoke spec picks its model via `ensureModelAvailable(page)` —
  i.e. the environment is *not* guaranteed to have OpenAI configured. I read the
  form schema and the model field validates as `z.string().min(1)` only, with no
  provider-availability check, and the submit button is disabled only for the
  code-metric edit block. So this should be fine. But "should be fine from
  reading the zod schema" is not the same as watching the dialog close, and the
  round-trip spec's `submitDialog()` waits on exactly that.
- **The estate's real tsc behaviour under TS 7.** The 94 errors I saw come from
  a config I modified to get past TS5102. Whether the fix is `types: ["node"]`,
  a `typescript` pin, or something else is for whoever takes the toolchain PR.
- **Whether `edit-rule` should own content-shape preservation.** That is the
  open decision in Should-fix 1, and it is genuinely a judgement about what the
  coverage denominator ought to be — not something I could settle by reading.

---
_[review_generated_tests.yml](https://github.com/comet-ml/comet-automation-tests/actions/runs/33877857735) ·


CometActions:
## 🔍 Generated-test review

# PR #8161 — [OPIK-7791] [QA] Proposed e2e specs from the 2.2.50 → 2.2.51 release exploration

CometActions · draft: yes · 6 files, 2 new specs
Head reviewed: `5fbcb784fe` → fixes pushed as `6b990338df`

**VERDICT: ready with notes — one open decision for a human**

The two specs are well placed, honestly titled, free of sleeps and skips, and
tear down through the established `automationRulesCleanup` fixture. They guard a
real, recently-fixed regression (OPIK-8250, merged as `e9adfc9f35`) from both
sides: prose that merely looks like JSON must come back whole, and genuine
multimodal content must not be flattened in the name of fixing the first. One
genuine bug in the supporting backend-client helper was found and fixed. One
tagging question remains that I am not permitted to decide.

## Blockers (0)

None.

## Should fix (1)

1. **An API-only test claims a UI capability** — `online-evaluation-judge-prompt-round-trip.spec.ts:94`
   Why: the second test (`A genuinely multimodal judge message still reads back
   as structured content`) never opens a browser — it seeds over REST, reads
   back over REST, re-saves over REST. It carries
   `@cap:online-evaluation.edit-rule`, and the PR flips `edit-rule` to
   `covered: true`. What that test actually asserts is judge-message
   *content-shape preservation*, which no capability in this area names.
   Mitigating: the flag is earned independently by the *first* test in the file,
   which does drive the edit dialog (kebab → Edit, read hydrated prompt, submit
   unchanged, reopen, re-read). So dropping the tag from the second test costs
   no coverage — `edit-rule` stays honestly `covered: true` either way.
   Also unresolved, and called out by the taxonomy's own inline comment: no test
   here edits a *field* through the dialog (rename, model, prompt rewrite).
   The note on the capability says so candidly, which is the estate's convention
   for scoped coverage.
   Fix: **a human must choose** — add a capability for judge-message
   content-shape round-tripping and retag, or drop `@cap:online-evaluation.edit-rule`
   from the second test and leave that behaviour honestly uncovered. I did not
   act on this: inventing a taxonomy capability decides what the product's
   coverage denominator *is*, and that is not mine to decide. The previous review
   pass (`5fbcb784fe`) left an explicit `OPEN, needs a human call` marker at
   `taxonomy.yaml:665-673`; it is still accurate and still open.

## Fixed in this PR (pushed as `6b990338df`)

1. **The "no-op" resave was not a no-op** — `client.ts:1815`
   `resaveAutomationRuleFromReadBack` documents itself as "write back exactly
   what was read", and did not. Verified against the backend models rather than
   taken from the bot's word:
   - `filters` is `@JsonIgnore` on `AutomationRuleEvaluator`/`…Update`, but each
     concrete subtype re-exposes it with `@JsonView({Public, Write})`
     (`AutomationRuleEvaluatorLlmAsJudge.java:83`,
     `AutomationRuleEvaluatorUpdateLlmAsJudge.java:46`). So it *is* returned by
     the public GET and *is* accepted by the PATCH — and because
     `AutomationRuleEvaluatorUpdateLlmAsJudge` takes it as a
     `@ConstructorProperties` parameter, omitting it passes `null` and clears a
     filtered rule's filters.
   - the helper sent `project_ids: [projectId]`, replacing the association set.
     A multi-project rule would be detached from every other project it targets.
     Note `projectIds` is `@JsonView(Write)` and so absent from the public read;
     the read returns `projects` (`{project_id, project_name}`), which is what
     the fix now derives from, falling back to the passed id.
   Neither is reachable from the specs as written today (their rules have no
   filters and one project), but the helper is shared client surface and its
   contract is exactly the thing it was breaking.
2. **A failed listing lost the server's reason** — `client.ts:238`
   `AutomationRuleEvaluatorPageRef` now carries `message`. The failure this type
   exists to describe is a 500, and `expect(status).toBe(200)` alone reports
   "expected 200, received 500" — true and useless, when the body names the
   unreadable rule.
3. **`JudgeMessageWrite` allowed invalid shapes** — `client.ts:219`
   Now an exclusive union, so "exactly one of `content` / `contentArray`" is a
   compiler guarantee instead of a doc comment.

## Notes (3)

1. **The `tsc` gate failure is not this PR's.** `tsconfig.json:13` sets
   `baseUrl`, which TypeScript removed in v7; `package.json:24` pins
   `typescript: ^7.0.2`. `baseUrl` is present on `origin/main` too, so
   `npx tsc --noEmit` fails identically there — it bails at config load and
   never type-checks anything. I verified the PR's code separately under an
   uncommitted scratch config (same options, `paths` without `baseUrl`, plus
   `types: ["node"]`): the PR's own code is clean, before and after my fixes.
   The estate-wide fix is a one-line tsconfig change and belongs in its own PR —
   happy to open it.
2. **A second pre-existing `tsc` error**, surfaced by the same scratch run:
   `Duplicate identifier 'deleteDashboard'` at `client.ts:652` and `:1144`.
   Present on `main` (3 declarations on both sides), unrelated to this PR, and
   currently invisible because the config error masks it. Same PR as note 1.
3. **Support code is bucket 1, verified not bucket 4.** The diff touches nothing
   outside `tests_end_to_end/`. The four frontend hooks the POM addresses
   (`add-edit-rule-dialog-submit`, `playground-message-row`, `data-role`,
   `playground-message-editor`) already exist on `main` — no product change was
   needed or made. `rawFetch` against `/v1/private/automations/evaluators/` is
   the file's own established pattern, already on `main` at lines 1619 and 1652.
   The only shared-code change is the POM's new `ruleRows` getter, a
   behaviour-preserving extraction of the `tbody tr[data-row-id]` locator that
   `waitForReady` and `ruleRow` already used; the full suite still collects
   122 tests in 58 files, unchanged.

## Bot comments

8 findings, all from `baz-reviewer[bot]`.

| 1 | Structured test falsely reports edit coverage (`round-trip.spec.ts:94`) | valid | **escalated** — human call, see Should fix 1; I may not invent a capability |
| 2 | API response types hide wire field names (`imageUrl` vs `image_url`) | invalid | this is the file's own convention: `AutomationRuleDetail` already camelCases `samplingRate`/`triggerScope` for `sampling_rate`/`trigger_scope`. Changing it here alone makes the new types the inconsistent ones |
| 3 | `JudgeMessageWrite` forwards invalid message shapes | valid | **fixed** — exclusive union, `client.ts:219` |
| 4 | New specs bypass the public evaluator client | invalid | `main` already `rawFetch`es this exact endpoint at `client.ts:1619`/`1652`. The rule in `conventions.md:80` is about not deep-importing the Python SDK's `opik/rest_api/*`, not about this TS backend client — which `conventions.md:116` itself references by raw REST path |
| 5 | Evaluator failures lose diagnostics | valid | **fixed** — `message` added to `AutomationRuleEvaluatorPageRef` |
| 6 | No-op resave destroys evaluator state | valid, confirmed against backend | **fixed** — `filters` echoed, `project_ids` derived from read-back `projects` |
| 7 | Broad response listener reports unrelated 5xx | invalid | the "subroutes" it warns about (`/{id}`, `/{id}/logs`) *are* evaluator endpoints; a 5xx there is a real failure this spec should catch. `/evaluators-backup` does not exist. The spec's existing comment already scopes the listener deliberately |
| 8 | Taxonomy falsely reports edit coverage (`taxonomy.yaml:680`) | partly valid | **escalated** — same decision as #1. The narrow claim (no *field* edit is tested) is true and is already stated in the capability's own `note:` and the `OPEN` comment above it |

Deferred: nothing. #1 and #8 are one open decision, not a deferral — there is
nowhere to defer them to, because the choice changes the coverage denominator.

Two follow-through actions I have **not** taken, and can on your go-ahead:
1. Open the estate-wide tsconfig PR (notes 1 and 2 — they are one change).
2. Reply on the five threads I judged invalid or escalated (#1, #2, #4, #7, #8),
   so they read as answered rather than ignored.

## Estate gates

Re-run after my edits, from the PR head:

| Gate | Result | |
|---|---|---|
| `tag_lint.py` | **PASS** | 61 specs checked, 1 exempt, 0 problems |
| `npx tsc --noEmit` | **FAIL** | pre-existing on `main` (see note 1); PR's own code verified clean under a scratch config |
| `npx playwright test --list` | **PASS** | 3 new tests collect; full suite 122 tests / 58 files, unchanged |

## What I could not verify

- **I did not execute these specs.** No Opik stack was available in this
  workspace, so every runtime claim below is static reasoning, not observation:
  that the edit dialog hydrates the CodeMirror prompt as the POM expects, that
  `.cm-line` joining reproduces the seeded string byte-for-byte, that the
  rules list renders exactly 7 rows without paginating, and that the PATCH in
  the resave helper still answers 204 with my `filters`/`project_ids` change.
  The last one is the only behaviour my fix could alter, and it is the one I
  most want run before this merges.
- **`filters` and multi-project rules are untested by these specs**, so my fix
  to `resaveAutomationRuleFromReadBack` is correct by construction against the
  backend models but not covered by an assertion. A spec that seeds a filtered
  rule and asserts the filters survive a resave would close that; I did not add
  one, as it is beyond this PR's scope.
- **The PR body.** I worked from `review-target.json`, which does not carry it,
  so I could not check whether the description is stale or whether it satisfies
  the repo's `lint-pr` requirements.
- **Whether `t2-cuj` is the right tier.** It looks right — these rules never
  score, so there is no LLM spend — but I could not measure wall-clock.

---

**This PR is still a draft, and I have deliberately left it that way.** In my
judgement it is ready for a human to take ready once the tagging decision in
Should fix 1 is made. Flipping that flag is a person's call, not this review's.

---
_[review_generated_tests.yml](https://github.com/comet-ml/comet-automation-tests/actions/runs/34103947302) ·


AndreiCautisanu (approved):

PR #8163, opened 2026-09-04
[NA] [SDK] fix: interpolate missing item id into DatasetItemUpdateOperationRequiresItemId messages

simpleqt:
## Details
`DatasetItemUpdateOperationRequiresItemId` inherits from `OpikException`, which does no printf-style formatting — so `raise ...("Missing id for dataset item to update: %s", item)` rendered as a literal tuple `("Missing id ... %s", {'id': ...})` instead of an interpolated message. The same pattern existed in the test-suite copy of the check. Switched both to f-strings; no tests assert the old message text.

## Change checklist
- [x] User facing
- [ ] Documentation update

## Issues

- Resolves #
- OPIK-

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: ZCode (AI coding agent)
- Model(s): GLM
- Scope: identified the exception formatting defect in the two `update()` paths; human verified both call sites and that no test pins the old message
- Human verification: yes — reproduced the raw-tuple rendering with `python3 -c` against `OpikException` semantics before fixing

## Testing
`python3 -m compileall` on both touched files; the change is message-only (`str(exception)` previously showed the tuple). Not run: full pytest suite (local env lacks the SDK's full dependency set); the touched paths have no message assertions in `sdks/python/tests`.

## Documentation
None needed.

### DIFF
--- a/sdks/python/src/opik/api_objects/dataset/dataset.py
+++ b/sdks/python/src/opik/api_objects/dataset/dataset.py
@@ -1035,7 +1035,7 @@ class Dataset(DatasetExportOperations):
         for item in items:
             if "id" not in item:
                 raise exceptions.DatasetItemUpdateOperationRequiresItemId(
-                    "Missing id for dataset item to update: %s", item
+                    f"Missing id for dataset item to update: {item}"
                 )
 
         self.insert(items, deduplication=deduplication)
--- a/sdks/python/src/opik/api_objects/dataset/test_suite/test_suite.py
+++ b/sdks/python/src/opik/api_objects/dataset/test_suite/test_suite.py
@@ -682,7 +682,7 @@ class TestSuite:
         for item in items:
             if "id" not in item:
                 raise opik_exceptions.DatasetItemUpdateOperationRequiresItemId(
-                    "Missing id for test suite item to update: %s", item
+                    f"Missing id for test suite item to update: {item}"
                 )
 
         self.insert(items, deduplication=deduplication)
--- a/sdks/python/tests/unit/api_objects/dataset/test_dataset_client.py
+++ b/sdks/python/tests/unit/api_objects/dataset/test_dataset_client.py
@@ -8,6 +8,7 @@ import pytest
 from opik.api_objects import constants
 from opik.api_objects.dataset import dataset_item
 from opik.api_objects.dataset.dataset import Dataset
+from opik import exceptions
 
 
 def _make_items(count: int) -> list:
@@ -736,3 +737,24 @@ def test_insert__repeated_inserts__backend_version_probed_once(monkeypatch):
         "Parallel upload is the default, so the version gate must be probed once "
         "per dataset rather than once per insert"
     )
+
+
+def test_update__item_without_id__raises_with_item_in_message():
+    mock_rest_client = Mock()
+
+    dataset = Dataset(
+        name="test_dataset",
+        description="Test description",
+        project_name="Test project",
+        rest_client=mock_rest_client,
+    )
+
+    item = {"input": {"key": "value"}}
+
+    with pytest.raises(
+        exceptions.DatasetItemUpdateOperationRequiresItemId,
+        match=r"Missing id for dataset item to update: .*'input'.*",
+    ):
+        dataset.update([item])
+
+    mock_rest_client.datasets.create_or_update_dataset_items.assert_not_called()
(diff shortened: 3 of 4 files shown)


andrescrz:
Hi @simpleqt 

Thanks for the contribution!

Thanks!

simpleqt:
Added the focused unit tests: both `Dataset.update` and `TestSuite.update` now have a guard test asserting `DatasetItemUpdateOperationRequiresItemId` with the exact item-bearing messages (`Missing id for dataset item to update: ...` / `Missing id for test suite item to update: ...`) — the old tuple-rendering behavior would fail these — plus a no-send assertion while the guard fires. `tests/unit/api_objects/dataset/`: 248 passed.

andrescrz:
Hi @simpleqt

Thanks for the contribution!

There's another round of Baz comments. When you have a chance, could you take a look at them? For each comment, please either address it with a code change or reply with a brief explanation if you believe the current implementation is the right approach.

Thanks!

sdks/python/src/opik/api_objects/dataset/dataset.py
```diff
@@ -1035,7 +1035,7 @@ def update(self, items: List[Dict[str, Any]], deduplication: bool = True) -> Non
         for item in items:
             if "id" not in item:
                 raise exceptions.DatasetItemUpdateOperationRequiresItemId(
-                    "Missing id for dataset item to update: %s", item
+                    f"Missing id for dataset item to update: {item}"
```
feiiiiii5, line 1038: Already fixed on `main@abb53ac` a different way: `dataset.py:1163-1167` raises `f"Missing id for dataset item at index {index}: {item}. {consequence}"` from the `require_id` helper, so this removed line no longer exists and the hunk conflicts on rebase. Suggest dropping this file from the PR.

sdks/python/src/opik/api_objects/dataset/test_suite/test_suite.py
```diff
@@ -682,7 +682,7 @@ def update(
         for item in items:
             if "id" not in item:
                 raise opik_exceptions.DatasetItemUpdateOperationRequiresItemId(
-                    "Missing id for test suite item to update: %s", item
+                    f"Missing id for test suite item to update: {item}"
```
feiiiiii5, line 685: Confirmed still live on `main@abb53ac` (lines 684-686). Because `OpikException` has no `__init__`, the current code shows the user `('Missing id for test suite item to update: %s', {...})`; this f-string is the change that matters. See the review body for the executed output.

feiiiiii5 (commented):
Reviewed `1767767f` against `main@abb53ac`. Two of the four files no longer describe `main`; the remaining `test_suite` fix is correct and still needed.

**1. The `test_suite.py` defect is live on main, so the fix there is needed.** `sdks/python/src/opik/api_objects/dataset/test_suite/test_suite.py:684-685` still passes two positional arguments to an exception class that ignores them: `OpikException` is `class OpikException(Exception): pass` (`sdks/python/src/opik/exceptions.py:8-9`), with no `__init__`, so the `%s` is never interpolated. Executed against main's `exceptions.py`:

```text
str(exc)  = ('Missing id for test suite item to update: %s', {'source': 'trace', 'turn': 1})
fixed str = Missing id for test suite item to update: {'source': 'trace', 'turn': 1}
```

**2. The `dataset.py` hunk is stale.** main already rewrote that call site: `dataset.py:1163-1167` raises from a `require_id(index, item, consequence)` helper with `f"Missing id for dataset item at index {index}: {item}. {consequence}"`. The line this PR removes (`"Missing id for dataset item to update: %s", item`) is no longer in the file, so the hunk will conflict on rebase and that half of the change is already in.

**3. The dataset guard test will fail once rebased.** `test_dataset_client.py` adds `match=r"Missing id for dataset item to update: .*'input'.*"`, but main emits `... at index 0: {'input': ...}. ...`. Since `pytest.raises(match=)` is a `re.search` over the message, the pinned substring no longer appears. Either re-target the assertion at the `at index {index}` wording or drop this test and keep only the suite one. For what it is worth, main's `test_dataset_client.py` has no `RequiresItemId` coverage at all today, so the assertion is worth keeping in corrected form.

**4. Question for the maintainers, not a blocker here:** should the suite message also carry the item index the way the dataset message now does? If yes, the suite change is a bit larger than an f-string swap.

Rebase, drop the `dataset.py` hunk, adjust the dataset test, and this is a clean single-site fix with a passing regression test.

Disclosure: prepared with an AI-assisted workflow. Each of the three code claims above was read or executed against `main@abb53ac` while writing this review.

PR #8165, opened 2026-09-04
[NA] [SDK] docs: document entrypoint/environment args and fix garbled task_threads docstring

simpleqt:
## Details
Docstring-only fix in the Python SDK: `BaseTrackDecorator.track()` documented neither `entrypoint` nor `environment` (both real keyword parameters), `Opik.trace()` documented neither `environment` nor its fallback to the client-configured environment, and the `task_threads` docstring in `evaluator.py` (both `evaluate()` and `evaluate_dataframe()`/variant) contained a duplicated, broken sentence from two merged revisions. No behavior change.

## Change checklist
- [ ] User facing
- [x] Documentation update

## Issues

- Resolves #
- OPIK-

## AI-WATERMARK

AI-WATERMARK: yes

- Tools: ZCode (AI coding agent)
- Model(s): GLM
- Scope: docstring/signature cross-check over sdks/python; wording for the two new Args entries follows the surrounding style; the environment fallback wording mirrors the actual `if environment is None: environment = self._config.environment` logic
- Human verification: yes — every claimed-missing parameter was checked against its signature

## Testing
`python3 -m compileall` on the three touched files passes. Docstring-only; no tests assert these texts.

## Documentation
Docstrings feed the published SDK API reference, which is the point of the fix.

### DIFF
--- a/sdks/python/src/opik/api_objects/opik_client.py
+++ b/sdks/python/src/opik/api_objects/opik_client.py
@@ -313,6 +313,8 @@ class Opik:
             thread_id: Used to group multiple traces into a thread.
                 The identifier is user-defined and has to be unique per project.
             attachments: The list of attachments to be uploaded to the trace.
+            environment: The environment in which the trace was created, e.g. 'production'
+                or 'development'. If not provided, falls back to the client's configured environment.
 
         Returns:
             trace.Trace: The created trace object.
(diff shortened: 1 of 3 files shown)


andrescrz:
Hi @simpleqt 

Thanks for the contribution!

Thanks!

simpleqt:
Both Baz findings addressed:

1. **Entrypoint docstring** — I verified the claim and revised the wording instead of changing the UI: it now says marked agents feed the agent view "which currently renders the first registered entrypoint only", so the doc no longer promises what the UI doesn't do. Per-entrypoint rendering/selection feels like a feature decision rather than a docs fix, so I kept it out of scope here.
2. **Nested environment precedence** — good catch that `resolve_child_span_environment` returns the enclosing environment. Added a sentence documenting that an explicit child `environment` is recorded as the parent's value for nested `@track` calls.

Decorator unit tests: 143 passed.

petrotiurin (approved):

petrotiurin:
Thank you for fixing this documentation gap!

PR #8167, opened 2026-09-04
[OPIK-8262] [QA] Proposed e2e spec from the #8162 exploration: thread-scoped rule fan-out

CometActions:
## What this is

One new e2e spec proposed by the QA test-radar side flow, written from a human-verified
exploration of **#8162** ([OPIK-8262] classify provider errors by status and stop collapsing
trace-thread fan-out).

**This PR targets `thiagohora/OPIK-8262-retry-classification-and-fanout`, not `main`.** The spec was
written and run against that branch's head, `c20c9e018a23b9fc45927d0224c80959cc7208ae`, which is
also the commit deployed to the PR environment the exploration used. It is a regression guard on
the fan-out this PR rewrote, so it belongs on top of the change it tests; retargeting it to `main`
once #8162 merges is a rebase, not a rewrite.

**Generated by an automated QA flow. It needs human review before merge and is deliberately a
draft.** No reviewers requested — a human promotes it.

## Where it came from

The exploration ran on the PR's own deployment (`pr-8162.dev.comet.com`, 2.2.53-8162-merge-3189,
OSS install, workspace `default`) and reached both halves of the change by hand before any spec
existed. It produced two strong candidates; this PR contains the one of them that can be written
as a deterministic, self-contained spec. The other is dropped, with reasons, below.

### `tests/online-evaluation/online-evaluation-thread-fanout.spec.ts` — ✅ passes

`@t2-cuj` · `@area:online-evaluation` · `@cap:online-evaluation.rule-scope-thread-span`

**What it asserts.** A manual evaluation naming N threads against one thread-scoped rule becomes
N independent scoring cycles — the behaviour `OnlineScorePublisher.enqueueThreadMessage` now
produces by emitting one stream entry per thread id instead of one entry carrying the whole list.
Concretely, over a 3-thread cohort:

- the request answers `entities_queued: 3`, `rules_applied: 1`;
- the rule's log stream carries **exactly 3** "to Python evaluator" lines — not 1 (an id lost in
  the split) and not 9 (each entry still carrying all three) — and each thread id appears in
  exactly one of them;
- every thread carries exactly one feedback score, value `1.0`, `source: online_scoring`, compared
  as the whole score set per thread rather than by looking this rule's score up among others;
- the project holds exactly the three seeded threads and nothing else;
- each of the three threads renders that score in its detail panel's **Feedback scores** tab.

The evaluator-call count is the load-bearing assertion. All three failure modes end with every
thread carrying a score — the duplicate case overwrites its own value — so the scores alone cannot
tell a correct fan-out from a collapsed one. Only the call count can.

**Why it is deterministic.** A constant-1.0 python thread metric: no provider key, no model
verdict, no wall-clock dependence. The rule sits at `sampling_rate: 0`, so the production sampler
can never fire it and the manual request is the only possible trigger — and the spec asserts every
thread is *unscored* before the request, so the later assertions are attributable to the fan-out
rather than to ordinary sampling.

**What it does not prove.** On the happy path the pre-#8162 code also scored all three threads: it
iterated the id list inside a single stream entry. So this is a regression guard on the split
(nothing lost, nothing duplicated), **not** a before/after discriminator. The behaviour only the
new code has — a per-thread retry budget, so one thread's retryable failure replays only that
thread — needs a retryable failure in the thread scorer, which the exploration could not provoke on
this estate (the python evaluator's failures are terminal 400s, and the thread LLM path needs a
provider key). It is deliberately out of scope here rather than asserted weakly.

### Supporting changes

- **`fixtures/thread-cohort.fixture.ts`** (new) — seeds three independent threads (2 turns each)
  through the public SDK bridge and resolves each to its `thread_model_id`, which is what
  `manual-evaluation/threads` addresses. It polls until all three threads have been aggregated and
  **throws naming the missing ones** if they never are: a spec handed a short cohort would assert
  over whatever arrived and pass having tested a smaller fan-out than it claims. No teardown — the
  traces, threads and scores all live in the `project` fixture's project, and the rule is cleaned
  up by the existing `automationRulesCleanup` fixture. The suite's own run-prefix sweep reported
  nothing left behind after the verification run.
- **`core/backend/client.ts`** — `createAutomationRule` gains a `type` so it can build a
  `trace_thread_user_defined_metric_python` rule (thread code carries a `metric` and no argument
  map; passing one is now a hard error rather than a field sent and ignored); new
  `evaluateThreadsManually`, which the pinned SDK cannot express at all; `ThreadRowRef` gains
  `threadModelId`.
- **`core/metrics/python-metric-source.ts`** — `buildConstantThreadScoreMetric`. A separate builder
  from `buildConstantScoreMetric` because the python backend dispatches the two differently:
  `metric.score(data)` positionally for `trace_thread_*` rules, `metric.score(**data)` for
  trace-scoped ones.
- **`coverage/taxonomy.yaml`** — spec added to the area's `specs:` list;
  `online-evaluation.rule-scope-thread-span` flipped to `covered: true, tier: t2-cuj` with a note
  scoping it to **thread scope only** — no spec creates a span-scoped rule, so that half stays
  untested and the note says so rather than letting the key read as fully covered.
  `online-evaluation.automation-logs` is deliberately left `covered: false`: this spec reads the
  rule log stream over `GET /automations/evaluators/{id}/logs` and never opens the
  `/$workspaceName/automation-logs` page that key names, which is exactly the distinction the
  existing comment on that entry asks callers to respect.

## Verification

Run against the PR's own deployment — the same environment and build the exploration used:

```bash
cd tests_end_to_end/e2e
OPIK_DEPLOYMENT=oss OPIK_BASE_URL=https://pr-8162.dev.comet.com OPIK_WORKSPACE=default \
OPIK_API_KEY=oss-no-auth WORKERS=1 \
  npx playwright test tests/online-evaluation/online-evaluation-thread-fanout.spec.ts --reporter=list
# 1 passed (24.0s)
```

`OPIK_API_KEY` is set only because this target's hostname ends in `comet.com`, which makes the
pinned TS SDK demand a key even though the install has no auth. It is not needed for a local OSS
run and the spec does not depend on it.

Also run:

- `python3 tests_end_to_end/coverage/tag_lint.py --taxonomy tests_end_to_end/coverage/taxonomy.yaml
  --estate tests_end_to_end` → `60 specs checked, 1 exempt, 0 problem(s)`.
- A deliberate mutation of the evaluator-call assertion (`3` → `4`) was run and **failed** as
  intended, so that assertion is known to discriminate rather than merely to pass.
- **`tests/online-evaluation/` in full**, because `core/backend/client.ts` and `fixtures/index.ts`
  are shared: **7 passed, 1 skipped, 1 failed (1.7m)**.

  The one failure is `online-evaluation-python-metric-errors.spec.ts`, and it is **not caused by
  this change** — it fails identically on the unmodified branch, verified by re-running it with
  these changes stashed. On this deployment a python metric that exits 0 without a result line is
  still reported as `Python evaluation failed (HTTP '500'): 500 Internal Server Error: Failed to
  execute code`, where that spec expects the classified `400 Bad Request: Execution failed: the
  metric produced no output`. Worth someone's attention — either the classification fix is not in
  this build's python backend, or it has regressed — but it is out of scope here and nothing in
  this PR touches it.

**Typecheck, with a caveat you should know about.** `npx tsc --noEmit` does not run on this branch
today, and did not before this change: `tsconfig.json` still sets `baseUrl`, which the pinned
`typescript@^7.0.2` removed (`error TS5102`). It was typechecked with TypeScript 5.9 instead, which
reports exactly one error — a pre-existing duplicate `deleteDashboard` in `core/backend/client.ts`
present on the unmodified branch too. Zero new errors from this change. Both are worth fixing, but
not in a QA spec PR.

The exploration's other strong candidate — **"a permanent provider failure is attempted once; a
transient one is retried"**, the retry-classification half of this PR — is **dropped**. It is the
more valuable of the two and it was verified by hand (401 → 1 delivery held over 30 minutes;
429 → 3 deliveries at exact 10-minute intervals), but it cannot currently be written as a spec this
suite should own:

1. **It needs a backend-reachable endpoint that answers a chosen HTTP status.** The suite's only
   such facility is the mock gateway in `services/mock-token-auth/`, which answers 401 only and,
   per `core/mock-auth.ts`'s `mockAuthSkipReason()`, is reachable only from a **local** backend —
   never from a deployment, including the one where this behaviour can be observed. The
   exploration's substitute was `https://httpbingo.org/status/<code>?x=`; a permanent spec whose
   verdict depends on a third-party service is a flake source and is ruled out by the estate's own
   determinism rule.
2. **Even with that endpoint, the discriminating assertion costs a full
   `onlineScoring.pendingMessageDuration`** — 10 minutes, deployment configuration, not settable
   from a test. Both halves of the pair need it: without waiting out a redelivery window, "attempted
   once" and "attempted once so far" are the same observation.

Asserting only the permanent half would be worse than nothing: it would pass equally well if every
provider error had been made non-retryable, which is the direction that silently loses evaluations.

The prerequisite is a small one — teach `services/mock-token-auth/mock_token_auth_service.py` to
answer a caller-chosen status, then write the spec against `mockGatewayUrlForBackend` on a local
OSS run where `REDIS_SCORING_PENDING_MESSAGE_DURATION` can be turned down. That is a change worth
making, and it needs a local OSS backend to verify on, which this flow did not have. Shipping the
spec unverified against that unbuilt facility would have been a guess.

Also not covered, and named here so the list reads as filtered rather than exhausted: the automatic
thread-close fan-out path (`TraceThreadOnlineScorerPublisher` — same split, same scorer, different
caller; gated behind a 15-minute inactivity timeout), the VertexAI/GAX branch, and the legacy
multi-id migrate branch. The exploration could not reach any of them either.

## Review notes

- Read against `.agents/skills/writing-e2e-tests/SKILL.md` and `conventions.md` on `main`. One step
  of that workflow was not run: **step 3, live-UI discovery via the Playwright MCP**, which is not
  available to this flow. No new page object was written to compensate — the spec drives the
  existing `LogsPage` and `ThreadPanelPage` only, and adds no new selector.
- The Threads table's own feedback-score column is **not** asserted. Enabling it needs a
  column-picker interaction no POM in the estate models, and the thread panel's Feedback scores tab
  is where a user actually reads a thread-level score (the table hides those columns by default) —
  so the spec asserts there, for all three threads rather than a sample.

Related: #8162

### DIFF
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringBaseScorer.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringBaseScorer.java
@@ -3,6 +3,7 @@ package com.comet.opik.api.resources.v1.events;
 import com.comet.opik.api.FeedbackScoreItem;
 import com.comet.opik.api.Span;
 import com.comet.opik.api.Trace;
+import com.comet.opik.api.Visibility;
 import com.comet.opik.api.evaluators.AutomationRuleEvaluatorType;
 import com.comet.opik.api.events.RedisSubscriberMessage;
 import com.comet.opik.api.filter.Operator;
@@ -12,6 +13,7 @@ import com.comet.opik.domain.FeedbackScoreService;
 import com.comet.opik.domain.SpanService;
 import com.comet.opik.domain.TraceSearchCriteria;
 import com.comet.opik.domain.TraceService;
+import com.comet.opik.domain.evaluators.OnlineScorePublisher;
 import com.comet.opik.infrastructure.OnlineScoringConfig;
 import com.comet.opik.infrastructure.OnlineScoringStreamConfigurationAdapter;
 import com.comet.opik.infrastructure.auth.RequestContext;
@@ -31,6 +33,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import static com.comet.opik.api.FeedbackScoreItem.FeedbackScoreBatchItem;
@@ -251,5 +254,45 @@ public abstract class OnlineScoringBaseScorer<M extends RedisSubscriberMessage> 
                             .concatWith(Flux
                                     .defer(() -> retrieveFullThreadContext(threadId, lastReceivedIdRef, projectId)));
                 }));
+    }
+
+    /**
+     * Routes a trace-thread entry down one of two branches: an entry carrying several thread ids was
+     * written by an older build and is <b>migrated</b> — republished as one entry per id, never scored —
+     * while the single-id entries this build writes are <b>scored</b> normally. Completion therefore does
+     * not mean "scored", so per-outcome logging belongs inside {@code scoreThread} or in the migrate
+     * branch, never on the returned Mono.
+     *
+     * <p>Migrating rather than scoring avoids needing a rule for reducing N per-thread outcomes into the
+     * one verdict a stream entry gets, which must mis-serve somebody when a permanent failure and a
+     * retryable one land together. Each replacement entry gets its own retry budget instead.
+     *
+     * <p>The ack is implicit: this returns the republish, so a failure leaves the entry unacked and it
+     * redelivers. If the republish succeeds but the ack fails, the entry splits twice and some threads are
+     * scored twice — tolerable because {@code feedback_scores} is a {@code ReplacingMergeTree} versioned on
+     * {@code last_updated_at}, so the second score overwrites rather than duplicating.
+     *
+     * <p>The migrate branch is a temporary shim, deletable once no pre-split entry can be in flight.
+     */
+    protected Mono<Void> migrateOrScoreThreadIds(@NonNull M message, @NonNull List<String> threadIds,
+            @NonNull OnlineScorePublisher publisher, @NonNull Function<String, M> singleThreadIdCopy,
+            @NonNull Function<String, Mono<Void>> scoreThread) {
+        if (threadIds.size() > 1) {
+            // Logged on success, and worded so it can never be mistaken for a scoring log.
+            return publisher.enqueueMessage(threadIds.stream().map(singleThreadIdCopy).toList(), type)
+                    .doOnSuccess(unused -> log.info(
+                            "Migrated '{}' legacy thread ids to single-id entries for workspace '{}'; "
+                                    + "no scoring performed for this entry",
+                            threadIds.size(), message.workspaceId()));
+        }
+        // @NotEmpty says this cannot happen; if it does, no retry would help, so let it be acked away.
+        if (threadIds.isEmpty()) {
+            log.warn("Discarding trace-thread entry with no thread ids for workspace '{}'", message.workspaceId());
+            return Mono.empty();
+        }
+        return scoreThread.apply(threadIds.getFirst())
+                .contextWrite(context -> context.put(RequestContext.WORKSPACE_ID, message.workspaceId())
+                        .put(RequestContext.USER_NAME, message.userName())
+                        .put(RequestContext.VISIBILITY, Visibility.PRIVATE));
     }
 }
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringTraceThreadLlmAsJudgeScorer.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringTraceThreadLlmAsJudgeScorer.java
@@ -4,7 +4,6 @@ import com.comet.opik.api.Project;
 import com.comet.opik.api.ScoreSource;
 import com.comet.opik.api.Span;
 import com.comet.opik.api.Trace;
-import com.comet.opik.api.Visibility;
 import com.comet.opik.api.attachment.EntityType;
 import com.comet.opik.api.evaluators.AutomationRuleEvaluator;
 import com.comet.opik.api.evaluators.LlmAsJudgeMessage;
@@ -19,6 +18,7 @@ import com.comet.opik.domain.evaluation.EvaluatedThread;
 import com.comet.opik.domain.evaluation.EvaluationRecorder;
 import com.comet.opik.domain.evaluation.OnlineEvaluationRecorder;
 import com.comet.opik.domain.evaluators.AutomationRuleEvaluatorService;
+import com.comet.opik.domain.evaluators.OnlineScorePublisher;
 import com.comet.opik.domain.evaluators.UserLog;
 import com.comet.opik.domain.llm.ChatCompletionService;
 import com.comet.opik.domain.llm.LlmProviderFactory;
@@ -35,7 +35,6 @@ import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 import org.redisson.api.RedissonReactiveClient;
 import org.slf4j.Logger;
-import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 import reactor.core.scheduler.Schedulers;
 import ru.vyarus.dropwizard.guice.module.installer.feature.eager.EagerSingleton;
@@ -70,6 +69,7 @@ public class OnlineScoringTraceThreadLlmAsJudgeScorer extends OnlineScoringBaseS
     private final ServiceTogglesConfig serviceTogglesConfig;
     private final OnlineEvaluationRecorder onlineEvaluationRecorder;
     private final AttachmentService attachmentService;
+    private final OnlineScorePublisher onlineScorePublisher;
 
     @Inject
     public OnlineScoringTraceThreadLlmAsJudgeScorer(@NonNull @Config("onlineScoring") OnlineScoringConfig config,
@@ -85,7 +85,8 @@ public class OnlineScoringTraceThreadLlmAsJudgeScorer extends OnlineScoringBaseS
             @NonNull AgenticScoringService agenticScoringService,
             @NonNull SpanService spanService,
             @NonNull OnlineEvaluationRecorder onlineEvaluationRecorder,
-            @NonNull AttachmentService attachmentService) {
+            @NonNull AttachmentService attachmentService,
+            @NonNull OnlineScorePublisher onlineScorePublisher) {
         super(config, redisson, feedbackScoreService, traceService, spanService, TRACE_THREAD_LLM_AS_JUDGE,
                 Constants.TRACE_THREAD_LLM_AS_JUDGE);
         this.aiProxyService = aiProxyService;
@@ -97,6 +98,7 @@ public class OnlineScoringTraceThreadLlmAsJudgeScorer extends OnlineScoringBaseS
         this.serviceTogglesConfig = serviceTogglesConfig;
         this.onlineEvaluationRecorder = onlineEvaluationRecorder;
         this.attachmentService = attachmentService;
+        this.onlineScorePublisher = onlineScorePublisher;
         this.userFacingLogger = UserFacingLoggingFactory.getLogger(OnlineScoringTraceThreadLlmAsJudgeScorer.class);
     }
 
@@ -112,25 +114,14 @@ public class OnlineScoringTraceThreadLlmAsJudgeScorer extends OnlineScoringBaseS
         log.info("Message received with projectId: '{}', ruleId: '{}', threadIds: '{}' for workspace '{}'",
                 message.projectId(), message.ruleId(), message.threadIds(), message.workspaceId());
 
-        return Flux.fromIterable(message.threadIds())
-                // Score each thread id independently: a single thread's failure must not stop scoring the
-                // sibling thread ids. Per-thread errors are materialized (onErrorResume) so the flatMap
-                // completes for every thread; the batch's first failure is then re-surfaced below. This keeps
-                // the failure on the Mono error path handled by BaseRedisSubscriber.processMessage's
-                // onErrorResume — classified as a processing error, following the normal retryable/
-                // non-retryable path — instead of leaking into the enclosing onErrorContinue via Flux.flatMap
-                // (which would drop the element and count it as an "unexpected" error).
-                .flatMap(threadId -> processThreadScores(message, threadId)
-                        .then(Mono.<Throwable>empty())
-                        .onErrorResume(Mono::just))
-                .collectList()
-                .flatMap(errors -> errors.isEmpty() ? Mono.<Void>empty() : Mono.error(errors.getFirst()))
-                .contextWrite(context -> context.put(RequestContext.WORKSPACE_ID, message.workspaceId())
-                        .put(RequestContext.USER_NAME, message.userName())
-                        .put(RequestContext.VISIBILITY, Visibility.PRIVATE))
-                .doOnSuccess(unused -> log.info(
-                        "Processed trace threads for projectId '{}', ruleId '{}' for workspace '{}'",
-                        message.projectId(), message.ruleId(), message.workspaceId()))
+        // The success log sits inside the scoring callback: a migrated entry completes this chain without
+        // scoring anything, so a doOnSuccess out here would claim work that never happened.
+        return migrateOrScoreThreadIds(message, message.threadIds(), onlineScorePublisher,
+                threadId -> message.toBuilder().threadIds(List.of(threadId)).build(),
+                threadId -> processThreadScores(message, threadId)
+                        .doOnSuccess(unused -> log.info(
+                                "Processed trace thread '{}' for projectId '{}', ruleId '{}' for workspace '{}'",
+                                threadId, message.projectId(), message.ruleId(), message.workspaceId())))
                 .doOnError(error -> log.error(
                         "Error processing trace thread for projectId '{}', ruleId '{}' for workspace '{}'",
                         message.projectId(), message.ruleId(), message.workspaceId(), error))
--- a/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringTraceThreadUserDefinedMetricPythonScorer.java
+++ b/apps/opik-backend/src/main/java/com/comet/opik/api/resources/v1/events/OnlineScoringTraceThreadUserDefinedMetricPythonScorer.java
@@ -4,7 +4,6 @@ import com.comet.opik.api.Project;
 import com.comet.opik.api.ScoreSource;
 import com.comet.opik.api.Span;
 import com.comet.opik.api.Trace;
-import com.comet.opik.api.Visibility;
 import com.comet.opik.api.evaluators.AutomationRuleEvaluator;
 import com.comet.opik.api.events.TraceThreadToScoreUserDefinedMetricPython;
 import com.comet.opik.domain.FeedbackScoreService;
@@ -12,6 +11,7 @@ import com.comet.opik.domain.ProjectService;
 import com.comet.opik.domain.SpanService;
 import com.comet.opik.domain.TraceService;
 import com.comet.opik.domain.evaluators.AutomationRuleEvaluatorService;
+import com.comet.opik.domain.evaluators.OnlineScorePublisher;
 import com.comet.opik.domain.evaluators.UserLog;
 import com.comet.opik.domain.evaluators.python.PythonEvaluatorService;
 import com.comet.opik.domain.threads.TraceThreadService;
@@ -26,7 +26,6 @@ import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.tuple.Pair;
 import org.redisson.api.RedissonReactiveClient;
 import org.slf4j.Logger;
-import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 import reactor.core.scheduler.Schedulers;
 import ru.vyarus.dropwizard.guice.module.installer.feature.eager.EagerSingleton;
@@ -62,6 +61,7 @@ public class OnlineScoringTraceThreadUserDefinedMetricPythonScorer
     private final ProjectService projectService;
     private final AutomationRuleEvaluatorService automationRuleEvaluatorService;
     private final AgenticScoringService agenticScoringService;
+    private final OnlineScorePublisher onlineScorePublisher;
 
     @Inject
     public OnlineScoringTraceThreadUserDefinedMetricPythonScorer(
@@ -75,7 +75,8 @@ public class OnlineScoringTraceThreadUserDefinedMetricPythonScorer
             @NonNull ProjectService projectService,
             @NonNull AutomationRuleEvaluatorService automationRuleEvaluatorService,
             @NonNull SpanService spanService,
-            @NonNull AgenticScoringService agenticScoringService) {
+            @NonNull AgenticScoringService agenticScoringService,
+            @NonNull OnlineScorePublisher onlineScorePublisher) {
         super(config, redisson, feedbackScoreService, traceService, spanService,
                 TRACE_THREAD_USER_DEFINED_METRIC_PYTHON,
                 Constants.TRACE_THREAD_USER_DEFINED_METRIC_PYTHON);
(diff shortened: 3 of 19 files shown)


CometActions:
## 🔍 Generated-test review

# PR #8167 — [OPIK-8262] [QA] Proposed e2e spec from the #8162 exploration: thread-scoped rule fan-out

CometActions · draft: yes · 7 files, 1 new spec

**VERDICT: ready with notes** — one taxonomy decision needs a human before merge.

The spec is genuinely good. It asserts the thing its tag claims, its teardown is in
fixtures, its POM reuse is correct, and — unusually — its log-string assertions
verify against the real backend source. I applied three readability fixes and found
nothing to fix on suite integrity or reliability. The one open item is a coverage
denominator question I am not allowed to decide.

---

## Blockers (0)

None.

---

## Should fix (1) — needs a human decision, not fixable by me

**1. `rule-scope-thread-span` flips to `covered: true` on thread scope alone —
`tests_end_to_end/coverage/taxonomy.yaml:659`**

The capability key names **two** scopes. The spec covers thread scope only; the PR
says so candidly, in both a `note:` and a five-line comment. But the coverage
builder reads `covered: true`, so span-scoped rules — which no spec anywhere
creates — become invisible to the radar permanently. That is the estate's own
"permanent false green", applied to half a capability.

This is exactly the case the brief forbids me to fix myself: splitting the key into
`rule-scope-thread` and `rule-scope-span` would be inventing a taxonomy capability,
and `TESTING-TAGS.md` is explicit that the taxonomy "is a **reviewed** file: it
defines 100%". So the choice is yours:

- **(a)** split into `rule-scope-thread` (covered) and `rule-scope-span`
  (`covered: false`) — accurate denominator, coverage dips by design, and the radar
  keeps surfacing span scope as a gap; or
- **(b)** keep it as-is and accept the scoping `note:` as the record — precedent
  exists (`sampling-rate`, `python-rule-scores`, `thread-level-metrics` all carry
  scoping notes on covered entries), at the cost of the radar never raising span
  scope again.

I lean (a), because a note is prose that no tool reads, and span scope is a real
untested surface rather than a nuance of a tested one. But it changes the
denominator, so it is your call.

---

## Notes (2)

**1. The `tsc` gate failure is not this PR's, and the gate is currently vacuous.**

`tsconfig.json(13,5): error TS5102: Option 'baseUrl' has been removed` — the repo
pins `typescript@^7.0.2` (resolved 7.0.2), which removed `baseUrl`. `tsconfig.json`
is byte-identical to `origin/main` and is not in this PR's diff. Because this is a
*config-level* error, tsc aborts before type-checking anything, so the gate has been
reporting FAIL without checking a single file — for every PR, not just this one.

I re-ran the typecheck against a repaired config (`baseUrl` dropped, `paths` made
tsconfig-relative, `types: ["node"]`) to get a real answer:

- **The PR's own five changed/added TS files type-check clean — zero errors.**
- The only two errors in the entire estate are pre-existing:
  `core/backend/client.ts(583,11)` and `(1075,11)` — duplicate `deleteDashboard`,
  introduced by #8016, untouched by this PR.

Both fixes are estate-wide and belong in their own PR, not stapled to a test PR —
repairing the config would immediately surface the duplicate-method error and drag
unrelated repair work in. **Happy to open that PR** (drop `baseUrl`, add
`types: ["node"]`, dedupe `deleteDashboard`) — say the word. It is worth doing
soon: the e2e suite has effectively had no typecheck since the TS 7 bump.

**2. The PR is stacked on an unmerged backend branch.**

Base is `thiagohora/OPIK-8262-retry-classification-and-fanout` (BE commit
`c20c9e018a`, "classify provider errors by status and split thread fan-out"), which
is **not** on `main` yet. The head is 0 commits behind its base, so nothing to merge
in. But this spec asserts the fan-out that BE commit implements, so it cannot be
retargeted to `main` until that lands. Legitimate stacking, not a defect — just
sequence it, and retarget after the BE PR merges.

---

## What I verified (and how)

| Check | Result |
|---|---|
| Placement | `tests/online-evaluation/` = `@area:online-evaluation` = `spec_dir: online-evaluation`. ✅ Spec **is** added to the area's `specs:` list. |
| Tag key exists | `online-evaluation.rule-scope-thread-span` is real, not invented. Tags are string literals, not computed. |
| Tag honesty | The test creates a `trace_thread_user_defined_metric_python` rule and asserts per-thread evaluator calls, server-side scores, and UI rendering. If thread-scoped fan-out broke, this fails. Scope caveat is the should-fix above. |
| Tier | `@t2-cuj` correct — no LLM spend (constant python metric, no provider key), but up to 300s. No tier inflation. |
| Fixed sleeps | The 50ms turn-spacing sleep (`thread-cohort.fixture.ts:84`) is the estate's **existing convention** for the same reason — identical lines in `conversation.fixture.ts:52` and `evaluated-thread.fixture.ts:119`. Not a finding. No `waitForTimeout` anywhere. |
| Teardown | In fixtures (`automationRulesCleanup`, `project`), not the test body. No `try`/`finally`, no trailing cleanup step. ✅ |
| Namespacing | Entities derive from `testNamespace` (`base.fixture.ts:39`), so they match the sweep contract by construction. |
| Polls | Both bounded, both with diagnostic messages naming what was missing. |
| **Log-string assertions** | Verified against backend source, not assumed: `OnlineScoringTraceThreadUserDefinedMetricPythonScorer.java:279` emits `Sending threadId '{}' to Python evaluator…` and `:307` emits `Scores for threadId '{}' stored successfully`. The quoting and the fact that it is the `thread_id` **string** (not the model id) both match what the spec matches on. Log-text matching is also existing convention — `online-evaluation-python-metric-errors.spec.ts:152` uses the same `EVALUATOR_CALL_LINE` constant. |
| Non-spec files (Step 9) | All **bucket 1/2 test support**, no production code. `evaluateThreadsManually` + `threadModelId` on the client, a thread metric builder, the cohort fixture. No bucket-4 changes. |
| Blast radius | `createAutomationRule`'s `arguments` went required → optional, guarded at runtime. All 4 existing callers pass `arguments` and are unaffected; a trace-scoped caller that omits it now throws early instead of failing opaquely — an improvement. `ThreadRowRef.threadModelId` is purely additive; the other consumer (`thread-id-prefilter.spec.ts`) does no exhaustive object compare. |
| Fixture duplication | Justified. The existing `conversation` fixture seeds **one** thread, and N=1 cannot distinguish a correct fan-out from a collapsed one. Not a near-duplicate. |

---

## Fixes applied and pushed

Commit `e3b8adc906` on `comet-qa-bot/OPIK-8262/thread-fanout-e2e` — three contained
readability fixes, no behaviour change:

1. Documented the `test.setTimeout(300_000)` budget and why it sits just above the
   inner 180s poll. Both neighbouring online-evaluation specs justify their timeout
   this way; this one did not.
2. Gave the cohort set-comparison (`:154`) a failure message.
3. Gave the per-thread panel score assertion (`:178`) a message naming the thread —
   it runs inside a loop, so `toBeCloseTo` alone would fail with
   "expected 1, received 0" and no indication of *which* thread.

---

## Bot comments

`bot-comments.json` is an empty array — **no bot review comments on this PR**.
Nothing to triage, defer, or reply to.

| — | *(none posted)* | — | — |

---

## Estate gates

Re-run after my edits, all three:

**tag_lint PASS · tsc FAIL (pre-existing, not this PR) · playwright --list PASS**

- `tag_lint` — `60 specs checked, 1 exempt, 0 problem(s)`, exit 0.
- `tsc` — exit 1, but solely `TS5102 baseUrl` at `tsconfig.json:13`, unchanged from
  `main` and independent of this PR. Under a repaired config the PR's own files are
  clean; see Note 1.
- `playwright --list` — collects, `Total: 1 test in 1 file`.

---

## What I could not verify

- **I did not execute the spec.** There is no running Opik stack in this workspace,
  so every runtime claim is static analysis plus source reading. In particular I
  could not confirm the timing budgets hold on a real box (the 60s thread-aggregation
  poll and the 180s scoring poll), nor that `entities_queued` is actually `3` rather
  than `1` — which is the PR's whole thesis.
- **Whether `trace_thread_user_defined_metric_python` is enabled on the target
  deployment.** The spec's own poll message anticipates it being disabled. If it is,
  this spec fails at the 180s poll for an environmental reason.
- **The `deleteDashboard` duplicate's runtime impact.** Two methods of that name in
  one object literal means the second silently wins. Pre-existing and out of scope
  here, but it is a live bug in the shared client, not just a type error.

---

## Do not merge — and I have not marked it ready

Per the standing rule I have **not** flipped the draft flag and have **not** merged,
and would not have even with all gates green. The PR is in good shape and, once you
settle the `rule-scope-thread-span` question above, a human can mark it ready. Note
also it cannot target `main` until the BE branch it is stacked on lands.

---
_[review_generated_tests.yml](https://github.com/comet-ml/comet-automation-tests/actions/runs/33900724998) ·


CometActions:
## 🔍 Generated-test review

# PR #8167 — [OPIK-8262] [QA] Proposed e2e spec from the #8162 exploration: thread-scoped rule fan-out

CometActions · draft: yes · 7 files changed by this PR · 1 new spec
Reviewed at `e3b8adc906`, fixes pushed as `12b741d90c`.

**VERDICT: ready with notes** — but see the blocker-adjacent branch problem below, which is
not the spec's fault and cannot be fixed from here.

The spec is honest, well-placed and asserts what its tag claims. It is the estate's
first thread-scoped rule test, and the thing it exists to catch — a fan-out that
collapses or drops ids — is genuinely caught, by counting evaluator calls per thread
rather than by looking at scores (which read identically in all three failure modes).
Three fixes applied; nothing left that I would block on.

**Do not flip this PR to ready or merge it — that is a human's call.** It also cannot
merge as-is until the branch problem in "Blockers" is resolved by its author.

## Blockers (1)

1. **The branch has diverged from its base, which was rebased under it** — branch
   `comet-qa-bot/OPIK-8262/thread-fanout-e2e`
   Base is `thiagohora/OPIK-8262-retry-classification-and-fanout`. This PR head carries
   `c20c9e018a` *"[BE] fix: classify provider errors by status and split thread fan-out"*,
   but that commit is **no longer on the base branch**: the base was rebased and its head
   is now `20af75e876` *"[BE] fix: publish one stream entry per trace-thread id"*, a single
   commit on top of `origin/main`.
   Why: `git merge-base --is-ancestor` fails both ways — the two have genuinely diverged,
   which is why GitHub still attributes 12 backend files to this PR. Merged as-is, this
   would reintroduce a superseded backend commit.
   Fix: **the branch author must rebase onto or retarget the current base.** I did not do
   it — a trial merge conflicts in `TraceThreadOnlineScorerPublisher.java` and
   `OnlineScorePublisherIntegrationTest.java`, and resolving backend production conflicts
   is outside a test review's remit and outside what I can verify. Merge aborted; working
   tree left clean.

   Good news for the spec itself: the current base commit is *precisely* "publish one
   stream entry per trace-thread id", which is exactly the behaviour this spec asserts.
   The spec is aligned with where the base is going, not stranded by the rebase.

## Should fix (0 — all fixed and pushed)

Fixed in `12b741d90c`:

1. **`entities_queued` was claimed as evidence of the fan-out, and is not** —
   `online-evaluation-thread-fanout.spec.ts:80`
   `ManualEvaluationService.evaluateThreads` (line 392) answers `threadModelIds.size()` —
   the count it was handed — after the enqueue chain completes. A request that packed all
   three ids into a single stream entry reports `3` here too. The step's comment asserted
   the opposite: *"The first place the split is observable. A request that collapsed to a
   single entry still answers 202."*
   This is the same species of false claim the tag-honesty check exists to catch, so it
   was worth fixing rather than noting. The assertion is **kept** — the unresolved-cohort
   path answers `0`, and without this check every later assertion would fail with no
   indication the request never landed — but the step, the message and the comment now
   describe it as an acceptance check and attribute the fan-out to the evaluator-call
   count that actually establishes it.

2. **The 300s timeout did not cover the chain its own comment described** —
   `online-evaluation-thread-fanout.spec.ts:47`
   60s fixture poll + 180s scoring wait leaves 60s for three sequential thread panels,
   each of which can spend up to 60s inside `waitForFullyLoaded` (`thread-panel.page.ts:23-24`:
   30s for the panel root, 30s for its first turn) before navigation. Raised to 600s so it
   is a backstop above the sum of the inner ceilings — which is what the comment already
   said it was for: let an inner timeout fire first, because each names what it was waiting
   for. 600s is within the area's range (`online-evaluation-enable-disable-rule.spec.ts`
   uses 660s).

3. **The trace-scoped `arguments` guard tested truthiness, not emptiness** —
   `core/backend/client.ts:1543`
   An explicit `arguments: {}` passed the guard and failed downstream at
   `PythonEvaluatorService` — the exact failure the guard's own comment says it exists to
   prevent, and one that reads like a broken metric. Now checks key count. Not reachable
   from any current caller (all three pass non-empty maps), but the guard is new in this
   PR and contradicted itself.

## Notes (4)

1. **`rule-scope-thread-span: covered: true` is honest by this estate's convention, but
   half the capability is untested** — `coverage/taxonomy.yaml:664`
   The key names *two* scopes; only thread scope is exercised. No spec anywhere creates a
   span-scoped rule.
   I checked before flagging this, and the estate has clear precedent for exactly this
   shape: `list-dashboards`, `create-dashboard`, `add-widget` and `configure-widget`
   (lines 377–383) are all `covered: true` with notes admitting substantial uncovered
   ground. The PR's note — `"thread scope only; span scope still uncovered"` — follows it,
   and the accompanying comment is unusually explicit. **So this is not a defect and I did
   not change it.**
   It is still a decision worth a human's eye, because the boolean is what the coverage
   builder counts and span scope now reads as tested. The options are (a) leave as-is, or
   (b) split the key into `rule-scope-thread` / `rule-scope-span`. **I did not split it —
   inventing a taxonomy capability is the user's call, not mine.**

2. **Manual thread evaluation is untagged coverage, with no key to tag it** —
   `online-evaluation-thread-fanout.spec.ts:80`
   The spec drives `POST /v1/private/manual-evaluation/threads` — the Threads view's
   "Evaluate" action. `grep -n manual coverage/taxonomy.yaml` returns no
   manual-evaluation capability in any area. Real behaviour asserted, invisible on the map.
   Same disposition as note 1: adding a capability is a taxonomy decision, so it is
   reported rather than fixed.

3. **The score-in-thread-panel assertion has no home in this area's keys.** The final step
   asserts the score renders in each thread's Feedback scores tab. `online-evaluation`'s
   nearest key is `scores-in-trace-panel`, which names the *trace* panel; the right-shaped
   key, `threads.thread-feedback-score`, is in a different area, and `@area:` is
   exactly-one and `tag_lint.py` enforces the `@cap:` prefix matching it. So there is no
   valid tag for it. Not fixable in this PR; flagging so the assertion is not mistaken for
   an omission.

4. **The `tsc` gate is broken estate-wide, on `main`, and this PR is not the cause.**
   See "Gates" below. Belongs in its own PR; I did not fix it here.

## Things I checked and found correct (not findings)

Recording these because each looks like a defect until you check the estate:

- **Placement.** `spec_dir: online-evaluation`, file in `tests/online-evaluation/`,
  `@area:online-evaluation`, and the spec **is** listed in the area's `specs:`. All three agree.
- **`automationRulesCleanup` is destructured but never called.** That is its API — it
  discovers rules from the project at teardown rather than by registration
  (`automation-rules.fixture.ts:8-13`), and all four neighbour specs use it identically.
  Correct, not a leak.
- **`await new Promise((r) => setTimeout(r, 50))` in the cohort fixture** (line 84) is turn
  spacing for deterministic `start_time` ordering, and is the estate's established pattern —
  identical lines in `conversation.fixture.ts:52`, `evaluated-thread.fixture.ts:119`,
  `trace-explore-smoke.spec.ts:20,27`. Not a fixed sleep standing in for a condition.
- **Log-string matching** (`'to Python evaluator'`, `'stored successfully'`) targets
  `userFacingLogger` output — a deliberate product surface, not incidental logging — and
  reuses the neighbour's constant approach (`online-evaluation-python-metric-errors.spec.ts:152`).
  Both strings verified present in the backend at this head.
- **Per-thread log matching quotes the id** (`'${threadId}'`), so `-thread-1` cannot match
  `-thread-11`. Correct by construction.
- **Teardown contract.** Thread ids derive from `testNamespace` = `${cujPrefix}-w{n}-{slug}`
  (`base.fixture.ts:39-46`), so the sweeper's timestamp shape holds.
- **Blast radius of the shared `client.ts` change.** `arguments` went required → optional;
  all three existing callers pass non-empty maps, so behaviour is unchanged for them.
  Full `playwright --list` is 120 tests in 57 files before and after my edits.
- **Scope of the diff.** Everything this PR actually adds is under `tests_end_to_end/` —
  a spec, a fixture extending the existing chain, a metric builder, a client route, a
  taxonomy line. All bucket 1. The 12 backend files GitHub attributes to the PR come from
  the base branch, not from `CometActions`; `git diff c20c9e018a..HEAD` is test-only.
  **No production code in this PR.**

## Bot comments

| 1 | Empty `arguments: {}` bypasses the trace-scoped guard (`client.ts:1543`) | **valid** | **fixed in this PR** (`12b741d90c`). Guard now checks key count. Skipped the suggested "regression test that `rawFetch` is not called" — the estate has no unit-test layer for the backend client, and adding one here would make this the sole exception. |
| 2 | `comparable()` in `thread-id-prefilter.spec.ts` should project `threadModelId` | **invalid** | Not fixed, and not deferred. `comparable()` is a curated projection documented as *"the fields a wrong prefilter would corrupt"*, and `id` is already in it. `threadModelId` is 1:1 with `id`, so once `id` matches, the row is the right row — a wrong `threadModelId` on a correctly-identified row is an id-mapping bug, not a prefilter bug, and not what that helper guards. Also a spec this PR does not touch. |
| 3 | `entitiesQueued` does not prove the fan-out; it echoes the submitted count | **valid** | **fixed in this PR** (`12b741d90c`). Confirmed against `ManualEvaluationService.java:392` — returns `threadModelIds.size()`. Took the bot's own recommendation: kept the assertion, rewrote the step title, message and comment so they no longer claim it as split evidence. The best of the four findings. |
| 4 | 300s timeout cannot cover the UI chain | **valid** | **fixed in this PR** (`12b741d90c`). The bot's 60s-per-panel figure checks out (`thread-panel.page.ts:23-24`). Raised to 600s as a backstop. Note the practical risk was low — a healthy run spends nowhere near the inner ceilings — but the comment's arithmetic was wrong, which is its own defect. |

Nothing deferred, so there is no ticket to file. **No replies were posted to the bot
threads** — `--comment` was not requested. Say the word and I will draft one reply per
thread (finding 2 in particular reads as ignored otherwise).

## Estate gates

Run exactly as `.github/workflows/tag_lint.yml` does, after my edits:

| Gate | Result | |
|---|---|---|
| `tag_lint.py` | **PASS** | `60 specs checked, 1 exempt, 0 problem(s)` |
| `npx tsc --noEmit` | **FAIL** | Pre-existing on `main`; not caused by this PR — see below |
| `npx playwright test --list` | **PASS** | `120 tests in 57 files`, unchanged from before my edits |

### On the tsc failure

```
```

`package.json` pins `"typescript": "^7.0.2"` and TS 7 removed `baseUrl`, which
`tests_end_to_end/e2e/tsconfig.json:13` still sets. **Neither file is touched by this PR**
(`git diff --stat origin/main HEAD` on both is empty), and I reproduced the identical
failure on a clean `origin/main` worktree. It is an estate-wide config defect, and it means
the gate currently type-checks *nothing* — it aborts on the config before reading a source
file.

To review the PR's TypeScript at all I removed `baseUrl` in a scratch copy of the config on
both `origin/main` and this head, and diffed the resulting error sets: **94 errors on each,
identical, with zero attributable to this PR's files.** So the PR's TypeScript is sound as
far as can be established. The 94 baseline errors are `@types/node` resolution failures plus
a pre-existing `Duplicate identifier 'deleteDashboard'` in `client.ts` — all on `main`,
none this PR's doing. I restored the config; the committed tsconfig is untouched.

Fixing it looks like a one-line deletion (`paths` has resolved relative to the tsconfig
without `baseUrl` since TS 4.1), but the 94 remaining errors mean it is not actually a
one-liner, and it belongs in its own PR rather than riding in on a test proposal. **Happy to
open that PR — say the word.**

## What I could not verify

- **The spec has never been executed.** No Opik stack was available here. `--list` proves it
  parses and collects; it does not prove it passes. Everything about runtime behaviour
  below is read from source, not observed.
- **Whether the fan-out assertion actually distinguishes the failure modes at runtime.** The
  reasoning holds against `ManualEvaluationService` and the scorer source, and the log lines
  it keys on exist, but the count `3` vs `1` vs `9` has not been seen.
- **The 600s budget is derived from the timeouts in the source**, not measured. The wall
  clock of a real run is unknown to me.
- **The backend behaviour after a rebase onto the current base.** The spec was written
  against `c20c9e018a`; the base is now `20af75e876`. Their intents match, but I could not
  run the spec against the new base — and the merge conflicts, so someone has to do this by
  hand and confirm.
- **Whether the 12 backend files carry anything a QA review should object to.** I confirmed
  they are not this PR's commits and stopped there; they are the base branch's, and belong
  to its own review.

---
_[review_generated_tests.yml](https://github.com/comet-ml/comet-automation-tests/actions/runs/34103898291) ·

PR #8168, opened 2026-09-05
[OPIK-8138] [SDK] fix: handle incomplete OpenAI Responses streams

### LINKED ISSUE #8138: [Bug]: abandoned openai Responses stream logs an IndexError traceback at error level
## Summary

When an async Responses-API stream is abandoned before it completes, `response_events_aggregator.aggregate()` raises `IndexError` internally and logs it at `error` level with a full traceback. Nothing is actually broken — the function recovers and returns `None` — but every abandoned stream produces an error-level log entry, which downstream error trackers file as a new issue.

Abandoning a stream is normal: a `break` out of the consuming loop, a cancelled task, or an HTTP client that disconnects mid-response.

## Reproduction

Verified against a clean install of `opik==2.0.53`, `openai==3.0.0`, CPython 3.14.2 (Linux). Nothing else is instrumented.

```python
import asyncio, os
from openai import AsyncOpenAI
from opik.integrations.openai import track_openai

client = track_openai(AsyncOpenAI(api_key=os.environ["OPENAI_API_KEY"]))

async def main():
    stream = await client.responses.create(
        model="gpt-4.1-nano",
        input="Count from 1 to 50, one number per line.",
        stream=True,
    )
    async for _event in stream:
        break          # abandon the stream after the first event

asyncio.run(main())
```

Output:

```
OPIK: Failed to parse openai Stream content. list index out of range
Traceback (most recent call last):
  File ".../opik/integrations/openai/stream_patchers.py", line 120, in wrapper
    yield item
asyncio.exceptions.CancelledError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File ".../opik/integrations/openai/response_events_aggregator.py", line 28, in aggregate
    response = completed_event[0].response
               ~~~~~~~~~~~~~~~^^^
IndexError: list index out of range
```

The same trace appears with `GeneratorExit` in place of `CancelledError` when the stream is abandoned by garbage collection rather than task cancellation.

## Mechanism

`aggregate()` filters the accumulated events down to the terminal ones, then indexes position 0 unconditionally:

https://github.com/comet-ml/opik/blob/main/sdks/python/src/opik/integrations/openai/response_events_aggregator.py

```python
completed_event = [
    event
    for event in items
    if isinstance(event, (ResponseCompletedEvent, ResponseErrorEvent, ResponseIncompleteEvent))
]

response = completed_event[0].response     # IndexError when the list is empty
```

An abandoned stream never delivers a terminal event, so `completed_event` is empty and the index raises. The surrounding `except Exception` catches it and logs at `error` with `exc_info=True`.

Instrumenting the aggregator confirms the difference is only the missing terminal event:

- stream consumed fully: `items=87 completed=1`
- stream abandoned after 3 events: `items=3 completed=0`

## Why it is worth changing

The recovery behaviour is already correct — a trace for an incomplete stream cannot be built, and returning `None` is the right answer. The problem is the severity: an expected, unavoidable condition is reported as an error.

For anyone running an error tracker wired to `logging`, each abandoned stream becomes a filed issue with a traceback, indistinguishable from a real fault. That is how we found this — a Sentry alert whose traceback pointed into the Opik SDK, on a request that had succeeded.

## Suggested fix

Guard the empty case explicitly and treat it as the expected outcome it is, so the genuine parse failures the `except` block was written for keep their error-level reporting:

```python
if not completed_event:
    LOGGER.debug(
        "openai stream ended without a terminal event (%d items); "
        "the stream was likely abandoned before completion",
        len(items),
    )
    return None

response = completed_event[0].response
```

Happy to open a PR if you would like one.

## Environment

| | |
|---|---|
| opik | 2.0.53 |
| openai | 3.0.0 |
| Python | 3.14.2 |
| OS | Linux (Docker), also reproduced on macOS 15 / CPython 3.14.0 |

lorenzozanee:
## Details

An abandoned OpenAI Responses stream can contain only non-terminal events. The Responses aggregator indexed the empty terminal-event list, causing an IndexError traceback to be logged at error level during stream finalization.

Return None when no terminal event has been collected, while preserving the existing error path for failures after a terminal event is present.

## Issues

Resolves #8138

## Testing

- PYTHONPATH=src python -m pytest tests/unit/integrations/openai/test_response_events_aggregator.py -q
- ruff check src/opik/integrations/openai/response_events_aggregator.py tests/unit/integrations/openai/test_response_events_aggregator.py
- ruff format --check src/opik/integrations/openai/response_events_aggregator.py tests/unit/integrations/openai/test_response_events_aggregator.py
- mypy src/opik/integrations/openai/response_events_aggregator.py
- pre-commit run --files sdks/python/src/opik/integrations/openai/response_events_aggregator.py sdks/python/tests/unit/integrations/openai/test_response_events_aggregator.py

### DIFF
--- a/sdks/python/src/opik/integrations/openai/response_events_aggregator.py
+++ b/sdks/python/src/opik/integrations/openai/response_events_aggregator.py
@@ -25,6 +25,9 @@ def aggregate(
             )
         ]
 
+        if not completed_event:
+            return None
+
         response = completed_event[0].response
         return response
     except Exception as exception:
--- a/sdks/python/tests/unit/integrations/openai/test_response_events_aggregator.py
+++ b/sdks/python/tests/unit/integrations/openai/test_response_events_aggregator.py
@@ -1,0 +1,22 @@
+import logging
+
+from openai.types import responses as openai_responses
+
+from opik.integrations.openai import response_events_aggregator
+
+
+def test_aggregate_non_terminal_events_returns_none_without_error(caplog):
+    event = openai_responses.ResponseCreatedEvent.model_construct(
+        response=None,
+        sequence_number=0,
+        type="response.created",
+    )
+
+    with caplog.at_level(logging.DEBUG, logger=response_events_aggregator.LOGGER.name):
+        result = response_events_aggregator.aggregate([event])
+
+    assert result is None
+    assert not [record for record in caplog.records if record.levelno >= logging.ERROR]
+    assert not any(
+        "list index out of range" in record.message for record in caplog.records
+    )


andrescrz:
Hi @lorenzozanee 

Thanks for the contribution!

Thanks!

PR #8171, opened 2026-09-05
[OPIK-8262] [QA] Proposed e2e specs from the #8170 exploration: LLM-judge permanent-4xx classification + the automation-logs page

CometActions:
## Details

the assertions before merging.** Full narrative below.

### Where these came from

Exploratory testing of **comet-ml/opik#8170** — `[OPIK-8262] [BE] fix: classify subscriber retries by status, not exception class`. A human worked the PR's flows by hand on the PR's own deployed environment (`pr-8170.dev.comet.com`, backend `2.2.53-8170-merge-3191`) first; only the flows that were actually reached and verified there became specs.

**These specs were written and run against the PR's head commit `14310654e48f00272f934f3bfa8267660a35b5a2`, and this PR targets `thiagohora/OPIK-8262-3-subscriber-retryability-predicate`, not `main`.** Basing on the PR branch keeps the specs on top of the change they were verified against, so a reviewer reading them sees the same code the exploration did. They are not asserting anything that exists only on that branch — see the honest caveat below — so they should be equally green after #8170 merges, but they have not been re-run on `main`.

### One thing a reviewer should know up front

**#8170 is behaviour-preserving on every input reachable from this estate.** The exploration read `scoreTrace` at the PR base (`c8f9be46`) and found it already threw `ClientErrorException(status)` for a permanent 4xx, with `ClientErrorException.class` already in `BaseRedisSubscriber`'s `NON_RETRYABLE_EXCEPTIONS` — so a permanent 4xx was already retired on first delivery, by class instead of by status. What changed is that the classification is now correct *by construction* (one shared `HttpStatusRetryability` predicate) rather than correct *by coincidence*.

So neither spec below is "a test for the new behaviour" — there is no new user-visible behaviour. Both pin behaviour this refactor could plausibly have broken, on the exact predicate it rewrote, that nothing in the estate currently asserts.

### 1. `online-evaluation-llm-judge-provider-errors.spec.ts` — `@t2-cuj` `@cap:online-evaluation.llm-judge-scores`

The LLM-judge analogue of `online-evaluation-python-metric-errors.spec.ts`. Nothing in the estate drives an LLM-judge rule against a failing provider: `online-evaluation-smoke.spec.ts` is the only other `@cap:online-evaluation.llm-judge-scores` spec and it asserts the happy path only.

API-level, no browser — the subject is how the engine classifies a provider outcome, and the rule's log stream is where that classification is stated. It asserts, for one rule at `sampling_rate: 1` over one seeded trace:

- the provider was called **exactly once** and the failure reported **exactly once** — a permanent 4xx is terminal, so neither a redelivery (which repeats the call line) nor an in-process retry loop (which repeats only the error line) may happen;
- the stream holds **exactly those three lines** and every one is attributed to the seeded trace, so a redelivery that wrote lines this spec does not name still fails;
- the ERROR carries **the provider's own status**, not a blanket 500 and not the pre-classification `An unexpected error occurred` catch-all;
- the trace ends up with **no feedback score**.

**Verification: PASSED.** `npx playwright test tests/online-evaluation/online-evaluation-llm-judge-provider-errors.spec.ts --retries=0` against `https://pr-8170.dev.comet.com` — passed first run, 1.5s.

### 2. `online-evaluation-automation-logs.spec.ts` — `@t2-cuj` `@cap:online-evaluation.automation-logs`

`taxonomy.yaml` kept this capability at `covered: false` with an explicit note that no spec opens the `/$workspaceName/automation-logs` **page** — `online-evaluation-python-metric-errors.spec.ts` reads the same lines through the API only. This is the spec that closes it, and #8170's whole user-visible surface is that log stream.

Written through the API and read back through the page, which is what catches a disagreement between the two. It asserts:

- the page renders the seeded stream row-for-row, with the ERROR row carrying the right Level and Trace Id;
- the collapsed message cell **hides** the provider's error body and the `Expand` control **reveals** it — compared against what the API said the message was, not against a literal;
- the bare `/automation-logs` route (no `rule_id`) shows its `No rule parameters set.` empty state rather than another rule's lines.

**Verification: PASSED.** `npx playwright test tests/online-evaluation/online-evaluation-automation-logs.spec.ts --retries=0` against the same environment — passed first run, 7.1s. Both specs together: `2 passed (13.7s)`, no retries.

### How the failure is made deterministic

No provider key and no third-party network. The provider's `base_url` points at a path the **Opik deployment under test does not serve**, so `POST <base_url>/chat/completions` is answered `404` by the environment itself. That is what makes "called exactly once" an assertable fact rather than a race.

Because that only works where the backend can reach the deployment's own API — which is not true of a docker-compose OSS install reached at `http://localhost:5173` — `core/failing-provider.ts` **proves the destination server-side before anything is seeded**, via the backend's own `/auth-config/test` endpoint, and skips with a reason naming `OPIK_PERMANENT_4XX_URL` when it cannot. Without that gate an unreachable destination would produce no wire status at all, take the *retryable* branch, and fail these specs on a message assertion — reading as a product regression. This is the same shape as the existing `mockAuthSkipReason()` gate.

**Reviewers on local OSS will most likely see these two specs skip**, with the reason printed. Set `OPIK_PERMANENT_4XX_URL` to a backend-reachable address that answers a permanent 4xx (any unserved path on the deployment will do) to run them locally.

### Supporting changes

| File | Why |
|---|---|
| `fixtures/provider-key.fixture.ts` | New `createPermanentFailure()` seed on the existing `providerKeys` fixture, so provider-key teardown stays where it already lives. Returns the fully-qualified model string, removing the one way this seed can silently produce a rule that resolves no provider. |
| `core/failing-provider.ts` | The destination and the gate described above. |
| `core/backend/client.ts` | `createLlmJudgeRule()` alongside `createAutomationRule()` (the two evaluator types share no part of their `code` payload, so one builder would type-check a rule carrying neither), and `traceId` exposed on `AutomationRuleLogRef` from the line's `trace_id` marker — the column the page renders. |
| `pom/automation-logs.page.ts` | New POM. Rows are addressed by the two facts that identify a log line — its trace and its level — via `data-cell-id`, never by column position: the page's `data-row-id` is `<timestamp>-<level>-md5(message)`, which is not an entity id and not knowable in advance. |
| `coverage/taxonomy.yaml` | Both specs added to the area's `specs:`; `automation-logs` flipped to `covered: true` (its "no spec opens the page" note is now false); `llm-judge-scores` note extended. `tag_lint.py`: `61 specs checked, 1 exempt, 0 problem(s)`. |

- **A transient provider status (408/425/429) being redelivered rather than dropped.** This is the half of the predicate #8170 actually rewrote, and it stays unautomatable: nothing on this estate can make a provider answer 429 on demand, and redelivery is gated by `onlineScoring.pendingMessageDuration` (default **10 minutes**), so counting even one redelivery would cost over ten minutes of wall clock per test. The exploration's unreachable-host rule is a *no-status* → 500 control, which exercises the fall-through branch, not `TRANSIENT_CLIENT_ERRORS`.
- **A healthy LLM-judge control rule storing a score.** Out of reach on the exploration environment — no provider key, no paid model. The "called exactly once" assertion substitutes for it: a rule that was never invoked writes no call line at all, and the poll fails naming that.
- **A python-metric 400 classification spec.** Already covered by `online-evaluation-python-metric-errors.spec.ts`.

### One unrelated thing this run turned up

`online-evaluation-python-metric-errors.spec.ts` **currently fails on `pr-8170.dev.comet.com`** — I ran it to confirm my `AutomationRuleLogRef` change had not broken a sibling, and it got past the healthy control rule and past `getAutomationRuleLogs` before failing on its message assertion:

```
Expected substring: "400 Bad Request: Execution failed: the metric produced no output"
Received string:    "Python evaluation failed (HTTP '500'): 500 Internal Server Error: Failed to execute code: "
```

Not caused by #8170 — it touches neither `PythonEvaluatorService` nor `opik-python-backend`. The classification lives in `apps/opik-python-backend/src/opik_backend/executor.py`, a separate service image, and that deployment is either running an older one or cannot run the sandbox executor at all (the detail after `Failed to execute code:` is empty). Worth knowing because a 500 is *retryable*: if a release gate is treating that spec as the guard for `isPermanent(400)`, it is not providing it on that environment. Flagged, not fixed — out of scope for this PR.

`npx tsc --noEmit` does not run on this checkout: `tsconfig.json` sets `baseUrl`, which the pinned TypeScript 7 has removed. Type-checking with that one option stripped, the estate reports 94 pre-existing errors and this branch reports 95 — the single addition is `Cannot find name 'process'` in the new file, the same `@types/node` resolution failure 53 existing lines already hit. Pre-existing, and not something this PR should fix.

---

🤖 **Generated by the release QA side flow (test-proposal step) and not reviewed by a human.** It is a draft on purpose: an unreviewed generated spec that asserts the wrong thing is worse than no spec, because it will be trusted. Please read the assertions before promoting it.

## Change checklist

## Issues

OPIK-8262

## Testing

comment on this PR.

## Documentation


CometActions:
## 🔍 Generated-test review

# PR #8171 — [OPIK-8262] [QA] Proposed e2e specs from the #8170 exploration: LLM-judge permanent-4xx classification + the automation-logs page

CometActions · draft: yes · 8 files, 2 new specs
Head `36134866fc` · base `thiagohora/OPIK-8262-3-subscriber-retryability-predicate` (tip `14310654e4`, exactly 1 commit behind head — not stale, nothing to merge)

**VERDICT: ready with notes** — a human still has to flip the draft flag. I did not, and will not, mark it ready or merge it.

The generator-specific failure modes this review exists to catch are all absent: the tags are honest, the capability it flips is genuinely earned by a browser assertion, the "new infrastructure" in the diff is all test support following existing patterns, and nothing in the diff is production code.

---

## Blockers (0)

None.

The one red gate (`tsc`) is **not this PR's** — see Gates below.

## Should fix (0)

None outstanding. Two contained readability fixes were applied under FIX MODE rather than reported (see Fixes applied).

## Notes (4)

1. **`tsc` is red on `origin/main` too — estate-wide, and it belongs in its own PR.**
   `tests_end_to_end/e2e/tsconfig.json:13` sets `baseUrl: "."`, and `package.json` pins `typescript: ^7.0.2` (installed: 7.0.2), which **removed** `baseUrl`. The config fails to parse, so *no* file is ever checked.
   I proved this is not the PR by running the installed compiler against a clean `origin/main` worktree — identical `TS5102` at the same line. This PR touches neither `tsconfig.json` nor `package.json`.
   The fix is one line — delete `"baseUrl": "."` and keep `"paths": { "@e2e/*": ["./*"] }`, which resolves relative to the config's own directory under TS 7. I deliberately did **not** apply it here: it is the estate's toolchain, it blocks every open PR's gate equally, and burying it in a test-spec PR hides a change that deserves its own review. Happy to open that PR.

2. **The specs cannot pass on `main` yet, and that is correct.**
   The base branch is not an ancestor of `origin/main`. Both specs assert the *new* behaviour from `ChatCompletionService` / `HttpStatusRetryability` — a provider 4xx reported with its own status instead of a blanket 500, and retired rather than redelivered. Targeting the fix branch rather than `main` is the right call; retargeting to `main` before the backend lands would make both specs fail for the right reason at the wrong time.

3. **The two specs duplicate ~10 lines of seeding and a 180 s poll each.**
   Deliberate, and both files say why: the classification spec drives no browser on purpose ("driving a browser here would add a rendering failure mode to an assertion that has nothing to do with rendering"), and the page spec confirms the stream server-side first so an empty table cannot be read as "the page does not render logs". Sharing a fixture would couple a UI claim to an API claim. Left as-is.

4. **`llm-judge-scores` now carries a failure-path test.**
   Spec 1 tags `@cap:online-evaluation.llm-judge-scores` for a test that asserts *no* score is written. That reads odd in isolation, but it is the area's established precedent: `online-evaluation-python-metric-errors.spec.ts` tags `python-rule-scores` for exactly the analogous 400-classification case, and the taxonomy note already records it that way. The tag flips nothing (the key was already `covered: true`) and the note was updated honestly. No taxonomy capability was invented, and none is needed.

---

## What I verified, against the estate rather than the PR's own claims

**Placement.** `online-evaluation` declares `spec_dir: online-evaluation`; both specs are in `tests/online-evaluation/`; both are added to the area's `specs:` list. Neighbours in that directory all carry `@area:online-evaluation`. Correct — no finding.

**Tag honesty — the important one.**

| Tag | Key exists? | Does the test assert it? |
|---|---|---|
| `@cap:online-evaluation.llm-judge-scores` | yes | yes — the judge's failure path: one provider call, one terminal ERROR carrying the provider's status, and an asserted-empty `feedbackScores` |
| `@cap:online-evaluation.automation-logs` | yes | **yes, and this is the one that matters** — see below |

`automation-logs` was `covered: false` with a comment explicitly refusing to flip it on an API read: *"this key names the `/$workspaceName/automation-logs` PAGE, which no spec opens."* This PR earns the flip properly. `AutomationLogsPage.goto()` navigates to that route, and the spec asserts rendered rows, the Level and Trace Id **cells**, that the collapsed message hides the provider body, that Expand reveals it, and that the bare route shows its empty state. That is a browser assertion on the surface the capability names. The old comment is correctly deleted.

Tags are string literals on `describe`/`test`, one tier each, both `@t2-cuj`. Tier is right: each spec polls up to 180 s, so it does not belong in a t1 that runs 3×/day — and neither spends LLM budget, because the provider always 4xxs.

No untagged real assertion that I could find, and no capability claimed broader than what is asserted.

**POM selectors, checked against the frontend source, not assumed.**
- `AutomationLogsPage.tsx:204` `getRowId` → `${timestamp}-${level}-md5(message)`; `DataTable.tsx:307` stamps `data-row-id`, `:352/373/391/404` stamp `data-cell-id` = TanStack's `${row.id}_${column.id}`. The POM's `td[data-cell-id$="_level"]` / `_marker_trace_id` / `_message` suffix matching is sound.
- The marker column ids are generated from the log line's `markers` keys. `AutomationRuleEvaluatorLogsDAO.java:37` restricts those to `trace_id` and `thread_model_id`, so there is no marker that could make a `_level` or `_message` suffix ambiguous.
- `DataTableSkeletonBody.tsx:53` renders `<TableRow>` with **no** `data-row-id` — so the POM's claim that `waitForReady` cannot latch onto a loading table holds.
- The Expand control is `ExpandableTextCell`'s `<Button>` whose text is literally `Expand` (`Collapse` once open), rendered only when the value has more than one line. `getByRole('button', { name: 'Expand', exact: true })` is right, and `exact: true` keeps it from matching the header's "Expand all" tooltip.
- The collapsed cell's tooltip is `CellTooltipWrapper`, which pushes content to a shared popover **on mouseenter only** — so `not.toContainText(providerDetail)` on the collapsed cell is a real assertion, not one that passes by accident.

**Log-line expectations, checked against the backend.**
`EXPECTED_LINES_PER_TRACE = 3` is right for the failure path: `OnlineScoringLlmAsJudgeScorer.java:397` (`"Evaluating traceId '{}' sampled by rule '{}'"`), `:489` (`"Sending traceId '{}' to LLM: {}"`), `:223` (`"Unexpected error while scoring traceId '{}' with rule '{}': \n\n{}"`). The `"Received response"` line at `:350` is only reached on success, so the chat throw cannot lengthen the stream. The `\n\n` split the specs rely on is the format string's own separator. All three lines carry `UserLog.TRACE_ID` from the MDC built at `:175`, so the per-line `traceId` assertion is sound. The substrings the specs match (`'sampled by rule'`, `'to LLM:'`) are exact.

**Diff triage (Step 9).** Nothing outside `tests_end_to_end/` is this PR's. The Java in `git diff origin/main...HEAD` belongs to the base branch; `git diff <base>..HEAD` is the 8 test files and nothing else. All are bucket 1 or benign bucket 2:
- `core/failing-provider.ts` — new, mirrors `core/mock-auth.ts`'s existing probe-and-skip pattern.
- `fixtures/provider-key.fixture.ts` — `createPermanentFailure` sits beside `createOauth` and builds the same `custom-llm/<provider>/<model>` string the same way; registers for the same teardown. Reuse, not a near-duplicate: a different auth mode.
- `core/backend/client.ts` — `createLlmJudgeRule` is genuinely new (the only other `llm_as_judge` payloads are inlined in the workspace-roles fixture/POM, not reusable). The `AutomationRuleLogRef.traceId` addition is purely additive; the four other specs reading that type only touch `level`/`message`, so no blast radius.
- `pom/automation-logs.page.ts` — matches the area's POM conventions exactly (constructor-injected `Page`, `goto` building from `loadEnvConfig()`, `test.step` wrappers), and imports the way the other 30 POMs are imported.

**Flake and isolation.** No `waitForTimeout` anywhere. Both polls are bounded (180 s) with intervals and a failure message that names the rule and says what a silent stream would mean. Entities are all `${testNamespace}`-prefixed, and `testNamespace` is built from `envConfig.cujPrefix`, so the estate's sweep finds them. Teardown is entirely in fixtures (`automationRulesCleanup`, `providerKeys`, `project`) — no `try`/`finally`, no cleanup step in a test body. The `test.skip(reason !== null, ...)` gate is the same shape `token-auth-dialog.spec.ts` and `token-auth-refetch.spec.ts` already use.

---

## Fixes applied

Two assertions in `online-evaluation-automation-logs.spec.ts` were the only ones in either file without a reason string, in files where every sibling assertion has one:

- `:112–113` → the Level and Trace Id cell assertions now say what each column is *for*, so a 3am failure reads as "the failure lost its trace attribution" rather than `expect(locator).toHaveText`.
- `:139` → the Expand-button visibility assertion now states that its absence means the provider body never reached the log line, which is the real diagnosis behind that failure.

No behavioural change. Nothing else met the fix-by-default bar: suite integrity, reliability, and readability were already clean.

## Bot comments

`bot-comments.json` is `[]` — no bot left an inline review comment on this PR. Nothing to triage, defer, or reply to.

| — | *(none)* | — | — |

---

## Estate gates: tag_lint **PASS** · tsc **FAIL (pre-existing)** · playwright --list **PASS**

Re-run after my edit, from the PR head:

- `tag_lint.py` — `61 specs checked, 1 exempt, 0 problem(s)`, exit 0.
- `npx tsc --noEmit` — `tsconfig.json(13,5): error TS5102: Option 'baseUrl' has been removed.` Reproduced identically against an `origin/main` worktree with the same compiler. **Not caused by this PR** (see Note 1).
- `npx playwright test --list <both new specs>` — `Total: 2 tests in 2 files`, both collected.

## What I could not verify

- **Neither spec was executed.** There is no Opik deployment in this workspace, so every runtime claim below is source-derived, not observed.
- **That `providerDetail` literally contains the status digits.** The engine logs `"Unexpected error calling LLM provider: " + <root-cause message>`, and for the custom-llm path that message is whatever body the destination returned. Opik's own 404 body does carry `404`, so `toContain(String(expectedStatus))` should hold — but I could not confirm the exact body an unserved `/v1/private/...` path returns on a running deployment. If this ever fails, it will fail loudly and for a readable reason, not silently.
- **That a `custom-llm` provider key with neither `api_key` nor `auth_config` builds a client that reaches the wire.** `apiKey` is nullable end-to-end in the backend (`ProviderApiKey` has no `@NotNull`; `LlmProviderFactoryImpl:80` passes `null` through), and the existing OAuth specs already seed keys with no `api_key`. But those supply an `auth_config`, and this seed supplies neither. If langchain4j rejects a null key client-side, the failure carries no HTTP status and the spec fails on the status assertion rather than skipping. Worth watching on the first real run.
- **The PR description.** `review-target.json` carries no body and I was asked not to re-fetch, so I could not check it for stale claims or for whatever the repo's PR linter requires.
- **Whether `tsc` has been red in CI for a while or only since the TypeScript 7 bump.** I confirmed the state on `main` today; I did not walk the history to date it.

---
_[review_generated_tests.yml](https://github.com/comet-ml/comet-automation-tests/actions/runs/33990757976) ·


CometActions:
## 🔍 Generated-test review

# PR #8171 — [OPIK-8262] [QA] Proposed e2e specs from the #8170 exploration: LLM-judge permanent-4xx classification + the automation-logs page

CometActions · draft: yes · 8 files, 2 new specs · base: `thiagohora/OPIK-8262-3-subscriber-retryability-predicate` (not `main`)

**VERDICT: ready with notes** — a person may flip it out of draft. Three fixes applied and pushed as `1b419d9c00`; nothing left that should block.

## Blockers (0)

None. Placement, tags, tiers, POM/fixture reuse and teardown are all correct, and I verified each against the estate rather than from memory:

- **Placement.** `online-evaluation` declares `spec_dir: online-evaluation`; both specs sit there, both are added to the area's `specs:` list, `@area:online-evaluation` matches. ✅
- **Tag honesty.** Both `@cap:` keys exist. `automation-logs` is the interesting one: the taxonomy previously pinned it `covered: false` with an explicit comment that flipping it on an API read would mark an untested UI as tested. This PR deletes that comment because it *earns* the flip — the spec opens `/$workspaceName/automation-logs`, asserts rendered rows, the Level and Trace Id columns, the collapsed-vs-expanded Message cell, and the bare route's empty state. That is exactly the right way to close that note. `llm-judge-scores` on the provider-failure spec mirrors the estate's own precedent: `online-evaluation-python-metric-errors.spec.ts` tags its failure-classification test `@cap:online-evaluation.python-rule-scores` with the same kind of scoping `note:`.
- **Tier.** `@t2-cuj` on both is right, not inflated — the "provider" is a 404 on the deployment's own API, so neither spec spends an LLM token, and both are too slow for t1.
- **No production code.** The Java files that appear in `git diff origin/main...HEAD` come from the base branch (#8170), not from this PR. `git diff HEAD~2...HEAD` is `tests_end_to_end/` only.
- **Support files are bucket 1.** `createLlmJudgeRule` sits beside the existing `createAutomationRule`; `createPermanentFailure` extends the existing `providerKeys` fixture rather than adding a parallel one; `AutomationLogsPage` is a new POM for a page with no POM. All following established patterns.
- **POM selectors verified against the real frontend**, not guessed: `data-row-id` / `data-cell-id` are stamped by `shared/DataTable/DataTable.tsx`, the `marker_trace_id` column id is built at `AutomationLogsPage.tsx:125`, both empty-state strings are literal in that file, and `ExpandableTextCell` really does render a `Button` labelled `Expand` only when the message has more than one line.

## Fixed in this review (3) — commit `1b419d9c00`

1. **The log stream was snapshotted mid-write, then counted.** Both specs polled until the first `ERROR` appeared and asserted exact counts off *that* snapshot: "called exactly once", "exactly three lines", and — in the automation-logs spec — a page row count read afterwards. A redelivered message writes its lines seconds later, so the precise regression these specs exist to catch could pass; in the UI spec it would instead surface as a flaky `toHaveCount(3)`. This is the estate's own convention for async assertions in *this area* (`waitForTraceScoresSettled`, used by `online-evaluation-sampling-rate` and `-enable-disable-rule`), just not yet applied to log lines. Added `core/backend/wait-for-rule-logs-settled.ts` — the direct log-stream analogue, same quiet-period shape, same never-sleep-past-the-deadline handling — exposed as `backendClient.waitForRuleLogsSettled()` and used by both specs.
2. **The automation-logs page assertions never read a message.** Rows were addressed by level and trace id only, so a page rendering three rows with empty Message cells would have satisfied the `automation-logs` capability. Each seeded line is now matched on the summary its Message cell actually shows, via a new `AutomationLogsPage.rowWithMessage()`, `toHaveCount(1)` per line so an ambiguous match fails loudly.
3. **The permanent-4xx gate probed the wrong URL.** It proved a permanent 4xx at the provider's *base* URL while the engine calls `<base>/chat/completions`. Identical on the default destination — both unserved paths on the same server — but `OPIK_PERMANENT_4XX_URL` can point anywhere, and a base that 404s while its child does not would have the spec assert a status the rule never saw. Now probes the exact URL the rule will request, via a new `permanentFailureChatUrl()`.

## Notes (4)

1. **The `tsc` gate fails on `main`, not on this PR.** `tests_end_to_end/e2e/tsconfig.json` still sets `baseUrl`, which the pinned TypeScript 7 removed — dependabot #8106 bumped `typescript` to `^7.0.2` without touching the config, and neither file is in this PR's diff. Type-checking with `baseUrl` dropped and `"types": ["node"]` supplied, the whole estate reports exactly two errors, both the duplicate `deleteDashboard` in `core/backend/client.ts` that landed on `main` in #8016. **Zero errors in this PR's new or edited code.** The fix is two lines in `tsconfig.json` plus deleting one duplicate method, it affects every PR in the repo, and it is not this test PR's to make — I did not touch it. Happy to open that PR if you want it.
2. **Both specs skip on a default local OSS install**, with a printed reason, because the backend at `localhost:5173` cannot reach the destination — and the taxonomy nonetheless reads `automation-logs: covered: true`. The PR body is candid about this and names `OPIK_PERMANENT_4XX_URL` as the escape. It is the same shape as the established `mockAuthSkipReason()` gate on `ai-providers`, so I left it, but it is worth knowing that this capability's coverage is real on CI/staging topologies and dark locally.
3. **The base branch is not `main`.** The PR targets #8170's branch so the specs sit on the code they were verified against. Correct, but it means this cannot merge until #8170 does, and the specs have not been run on `main`. The PR body says so explicitly.
4. **The body flags a pre-existing failure in a sibling spec** — `online-evaluation-python-metric-errors.spec.ts` fails on `pr-8170.dev.comet.com` with a 500 where it expects a classified 400, traced to the `opik-python-backend` executor image rather than to #8170. Out of scope here, but it means that spec is not currently guarding `isPermanent(400)` on that environment. Worth a ticket.

## Bot comments

| 1 | `automation-logs.spec.ts:103` — INFO rows checked only by count, level and trace id, so wrong messages/timestamps/ordering still pass | partly valid | **Message half fixed in this PR** (see fix 2). Timestamp and ordering halves rejected: the Timestamp column renders through `TimeCell` with `timeMode: "absolute"`, so asserting it means reimplementing the frontend's formatter in the test, and the page sorts rows `b.timestamp.localeCompare(a.timestamp)` — an ordering assertion would pin a UI sort the log API gives no matching guarantee for. The bot also cites `.agents/skills/writing-e2e-tests/conventions.md` as *requiring* per-row message and timestamp assertions; it does not — that file contains no such rule. |
| 2 | Both specs snapshot rows as soon as an `ERROR` appears, so later writes can make exact-count assertions flaky or falsely pass | valid | **Fixed in this PR** (see fix 1). The bot is right about the mechanism. Note its severity call is understated for the UI spec, where the consequence is a flaky row count, not just a weak assertion. |
| 3 | `failing-provider.ts:96` — the gate records the status for `token_url: url` but the provider is called at `url + "/chat/completions"` | valid | **Fixed in this PR** (see fix 3). Not reachable on the *default* destination — both are unserved paths on one server — but it is reachable through `OPIK_PERMANENT_4XX_URL`, and probing the exact URL costs one line. |

No deferrals, so there is nothing to file. If you want the bot threads answered visibly, I can draft one reply each — say the word; I have not posted anything.

## What I could not verify

- **I did not run either spec.** There is no Opik deployment reachable from this workspace, so my three fixes are verified by type-check, tag lint and test collection only — not by execution. The PR body reports both specs passing against `pr-8170.dev.comet.com` at commit `c798ef43`, which is *before* my changes.
- Two of my fixes have an execution-only failure mode I could not rule out: `rowWithMessage` uses substring matching, so two log lines whose summary lines are prefixes of one another would make `toHaveCount(1)` fail — loudly and legibly, but as a false failure. And the pre-existing `toContainText(providerDetail)` assertion on the expanded row depends on Playwright's whitespace normalisation handling a multi-line provider body; I read the code but could not exercise it.
- **The 10-second quiet period is a bound, not a proof.** A redelivery slower than that still escapes the "exactly once" assertion. Its documented default is `onlineScoring.pendingMessageDuration` = 10 minutes, which the PR body correctly says is unaffordable to wait out. The assertion is meaningfully stronger than it was, not airtight.
- I did not confirm the state of the base branch or whether #8170 has merged; `gh pr view` reports this PR `MERGEABLE` against it, and I did not go further.

## Estate gates

Re-run after the fixes, at `1b419d9c00`:

- **tag_lint — PASS** (`61 specs checked, 1 exempt, 0 problem(s)`)
- **tsc — FAIL**, unchanged and pre-existing: `tsconfig.json(13,5): error TS5102: Option 'baseUrl' has been removed`. Fails identically on `main`; see note 1. With that option stripped the new code type-checks clean.
- **playwright --list — PASS** (`121 tests in 58 files`; both new specs collect)

## Two things I did not do, deliberately

I did **not** mark this ready for review and did **not** merge it — that flag is the human's to flip. And I did **not** invent a taxonomy capability: there is no key for "rule failure classification", and rather than adding one I left both specs on the existing `*-scores` keys with scoping notes, exactly as the python sibling already does. If you would rather the estate had a first-class capability for evaluator failure classification, that is your call to make in the taxonomy, and it would cover three specs, not one.

---
_[review_generated_tests.yml](https://github.com/comet-ml/comet-automation-tests/actions/runs/34103851118) ·

